PR #17, opened 2015-07-14
The babel example is now correct.

oed:
The example for adding babel in the README.md was incorrect. When buildning this outputed `[object Object]` in the beginning of the app.js. Babel.transform outputs an object where the key `code` has the code we want.


tcoulter:
Thanks @oed!

PR #27, opened 2015-08-14
Helpful error message for failed contract deployment

skmgoldin:
I figure if four people who do this full-time barely realized that this might be the case, it will probably help other people.


tcoulter:
Thanks @skmgoldin!

PR #34, opened 2015-08-30
Clean folder on build, don't recreate it

jdmaresco:
Based on our chat on Gitter, I changed the build process to make Truffle clean the inside of the build folder, not replace it. Worked for me.

From what @tcoulter shared, I realized having a `--keep` flag to toggle this behavior may be unnecessary if there's not an actual case where the folder itself needs to be actually recreated.

### DIFF
--- a/lib/build.es6
+++ b/lib/build.es6
@@ -129,7 +129,7 @@ var Build = {
 
   base: function(config, key, callback) {
     // Clean first.
-    rimraf(config[key].directory).then(function() {
+    rimraf(config[key].directory + '/*').then(function() {
       return mkdirp(config[key].directory);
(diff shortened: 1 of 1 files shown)


tcoulter:
Thanks @jdmaresco. Appreciate it.

tcoulter:
@jdmaresco FYI: I'll be releasing a new version of Truffle either today or tomorrow that includes `truffle serve`, which should be a stand in for using SimpleHTTPServer. As well, `truffle serve` will act like `truffle watch` in addition to serving your built files on port 8080.

jdmaresco:
@tcoulter great, thanks!

PR #39, opened 2015-09-13
Handle contract name not matching file name.

vishakh:
Added handling for cases where contract names are not case-sensitive identical to their corresponding file names, e.g. fooBar.sol contains a contract called 'FooBar'.

I made this change because I spent quite some time trying to get Truffle to deploy before I realized that the problem was my contract name and the file name in the contracts directory were not identical.

Let me know if you want me to make it more generic and handle all cases where the contact and files names do not match.

### DIFF
--- a/lib/contracts.es6
+++ b/lib/contracts.es6
@@ -154,6 +154,14 @@ var Contracts = {
               return;
             }
 
+            if(result[key] == null && Object.keys(result).length > 0) {
+              var contractName = Object.keys(result)[0];
+              if ((contractName.toUpperCase() == key.toUpperCase()) && (contractName != key)) {
+                callback(new Error("Could not get compiler results because contract name '" + contractName + "' is not identical to file name '" + key + "'."));
+                return;
+              }
+            }
+
             contract["binary"] = result[key].code;
             contract["abi"] = result[key].info.abiDefinition;
             finished(null, contract);


tcoulter:
Thanks @vishakh. This is a good change. I wonder if since we're able to detect the problem, we might as well return the correct result. 

Also, there's one issue here. The first result from compile_solidity (i.e., `var contractName = Object.keys(result)[0];`) isn't necessarily the right contract. For instance, in the case where the code submitted to compile_solidity contains more than one contract (which is possible), the last one will be the one that represents all the compiled code. If we're going to detect this case, we should likely go through the list if keys, convert them to lowercase and see if there's matching one. If so, either show an error (or better) return the right result.

vishakh:
Thanks for the feedback, @tcoulter. I definitely need to add support for code with multiple contracts.

Before I made the change, I thought about whether to simply detect the problem or to also fix it. When two contracts with the same name but mixed cases (e.g. Foo and fOO) are submitted together, it's not clear what should be done. Do you want me to return the correct result on a best effort basis, i.e. return a result where possible and throw an error in boundary cases like the one above? Another option would be simply to warn the user since mismatches between contract and file names will likely cause them issues down the line anyway.

jgostylo:
I did not see this before I submitted my pull request: https://github.com/ConsenSys/truffle/pull/86

Please see my solution although it does not handle the lowercase instance.  It actually ignores what the filename is altogether.

PR #42, opened 2015-09-24
Closes #41: makes inject-contracts post-processor work in node env

### LINKED ISSUE #41: inject-contracts post-processor don't work in node environment
Currently, if an empty file is post processed with "inject-contracts" it works in the browser, but not in the node environment, because the window dependent code is not wrapped:

Currently the output file looks like this:
```
// If we're in the browser, call the helpers using the global
// (window) scope. If we're in node, export the helpers as a module.
if (module != null) {
  module.exports = __provisioner;
}
__provisioner.provision_contracts(window);

/* code */

; __provisioner.set_provider(window);
```

When it should look like this:
```
// If we're in the browser, call the helpers using the global
// (window) scope. If we're in node, export the helpers as a module.
if (module != null) {
  module.exports = __provisioner;
} else {
  __provisioner.provision_contracts(window);
}


; if( module == null ) {
  __provisioner.set_provider(window);
}
```

mhhf:
This PR will work with node, still I don't think its a good solution: 
To set up a web3 provider, web3 has to be passed as an object or wrapped.

### DIFF
--- a/lib/insert_contracts.js
+++ b/lib/insert_contracts.js
@@ -30,5 +30,13 @@ var __provisioner = {
 // If we're in the browser, call the helpers using the global
 // (window) scope. If we're in node, export the helpers as a module.
 if (module != null) {
-  module.exports = __provisioner;
-}
+  var obj = {};
+  __provisioner.set_provider( obj );
+  __provisioner.provision_contracts( obj ); 
+ 
+  module.exports = {
+    web3,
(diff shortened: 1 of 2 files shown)


mhhf:
I screw it up.
I don't think inject-contracts post-processor should be used as a node module. I'll write another post-processor for this purpose.

PR #47, opened 2015-10-26
Testing improvments

chetan:
### Changes in this PR:

* added watch:tests task - uses gaze to re-rerun tests automatically. this helps reduce the startup lag between test runs for more immediate feedback
* allow redeploying contracts from within tests
* use evm snapshot/revert before/after tests instead of redeploying
* display Event logs on test failures

### Possible TODOs:
* refactor test code
* readme updates
* ~~only use snapshot/revert if the server supports it (can test for it before suite)~~ (implemented)

Logs look something like this:

```
    1) should be able to transfer records

    Events emitted during test:
    ---------------------------
[DEBUG] Initializing account
[DEBUG] Adding value to account
{ event: 'Deposit',
  data: 
   { _from: '0x82a978b3f5962a5b0957d9ee9eef472ee55b42f1',
     _id: '0x666f6f6261720000000000000000000000000000000000000000000000000000',
     _value: '10000',
     status: 'success' } }
{ event: 'Deposit',
  data: 
   { _from: '0x82a978b3f5962a5b0957d9ee9eef472ee55b42f1',
     _id: '0x666f6f6261720000000000000000000000000000000000000000000000000000',
     _value: '10000',
     status: 'success' } }
[DEBUG] Adding value to account
{ event: 'Deposit',
  data: 
   { _from: '0x82a978b3f5962a5b0957d9ee9eef472ee55b42f1',
     _id: '0x666f6f6261720000000000000000000000000000000000000000000000000000',
     _value: '10000',
     status: 'success' } }

    --------------------------- END EVENTS
    ✓ should be able to be killed (636ms)


  10 passing (7s)
  1 failing

  1) Contract: Bank should be able to transfer records:
     AssertionError: assert.fail()
    at /tmp/truffle-115926-24254-166e646-bank.coffee:111:16
    at tryCatcher (node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (node_modules/bluebird/js/main/promise.js:507:31)
    at Promise._settlePromiseAt (node_modules/bluebird/js/main/promise.js:581:18)
    at Promise._settlePromises (node_modules/bluebird/js/main/promise.js:697:14)
    at Async._drainQueue (node_modules/bluebird/js/main/async.js:123:16)
    at Async._drainQueues (node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (node_modules/bluebird/js/main/async.js:15:14)
    at Function.module.exports.loopWhile (node_modules/deasync/index.js:64:21)
    at node_modules/deasync/index.js:36:18
    at runTask (truffle.es6:55:5)
    at Object.<anonymous> (truffle.es6:376:14)
    at normalLoader (node_modules/babel/node_modules/babel-core/lib/api/register/node.js:199:5)
    at Object.require.extensions.(anonymous function) [as .es6] (node_modules/babel/node_modules/babel-core/lib/api/register/node.js:216:7)
    at Object.<anonymous> (node_modules/babel/lib/_babel-node.js:144:25)
    at node.js:961:3
```

### DIFF
--- a/lib/test.es6
+++ b/lib/test.es6
@@ -2,10 +2,8 @@ var Mocha = require("mocha");
 var chai = require("chai");
 var dir = require("node-dir");
 var path = require("path");
-var web3 = require("web3");
 var fs = require("fs");
 var temp = require("temp");
-var Promise = require("bluebird");
 
 var Contracts = require("./contracts");
 
@@ -52,7 +50,7 @@ var Test = {
 
         Pudding.setWeb3(web3);
         PuddingLoader.load(config.environments.current.directory, Pudding, global, function(err, contract_names) {
-          for (var name in contract_names) {
+          for (var name in config.contracts.classes) {
             var contract = global[name];
             var inst = contract.at(contract.deployed_address);
             Truffle.log_filters.push(inst.allEvents({fromBlock: 0, toBlock: 'latest'}, function(err, results) {
--- a/lib/test.es6
+++ b/lib/test.es6
@@ -171,7 +171,7 @@ var Test = {
         // instead of the original to avoid getting cached.
         files = files.map(function(f) {
           var src = fs.readFileSync(f);
-          f = temp.path({prefix: "truffle-", suffix: path.extname(f)})
+          f = temp.path({prefix: "truffle-", suffix: "-"+path.basename(f)})
           fs.writeFileSync(f, src);
           return f;
         });
--- a/lib/test.es6
+++ b/lib/test.es6
@@ -53,9 +53,7 @@ var Test = {
           for (var name in config.contracts.classes) {
             var contract = global[name];
             var inst = contract.at(contract.deployed_address);
-            Truffle.log_filters.push(inst.allEvents({fromBlock: 0, toBlock: 'latest'}, function(err, results) {
-              // no-op, don't worry abuot tracking events just now
-            }));
+            Truffle.log_filters.push(inst.allEvents({fromBlock: 0, toBlock: 'latest'}));
           }
           done();
         });
--- a/lib/test.es6
+++ b/lib/test.es6
@@ -5,6 +5,7 @@ var path = require("path");
 var web3 = require("web3");
 var fs = require("fs");
 var temp = require("temp");
+var Promise = require("bluebird");
 
 var Contracts = require("./contracts");
 
@@ -56,6 +57,7 @@ var Test = {
 
     global.Truffle = {
       redeploy: redeploy_contracts,
+      handle_errs: (done) => { Promise.onPossiblyUnhandledRejection(done); }
     };
 
     global.contract = function(name, opts, tests) {
--- a/lib/test.es6
+++ b/lib/test.es6
@@ -55,22 +55,43 @@ var Test = {
       });
     };
 
+    var rpc = function(method, arg, cb) {
+      var req = {
+        jsonrpc: "2.0",
+        method: method,
+        id: new Date().getTime()
+      };
+      if (arguments.length == 3) {
+        req.params = [arg];
+      } else {
+        cb = arg;
+      }
+      web3.currentProvider.sendAsync(req, cb);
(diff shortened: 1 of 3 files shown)


tcoulter:
First impression (still need to dive deep):

1. "added watch:tests task" - awesome.
1. "allow redeploying contracts from within tests" - awesome. Need to see how you did it.
1. "use evm snapshot/revert before/after tests instead of redeploying" and "only use snapshot/revert if the server supports it" - the last bit is important. Truffle used to use evm_snapshot and evm_revert, but this prevented you from running your tests against geth. Hence the redeploy model. Still - is the time of deployment slowing down tests by that much? In my experience compiling takes much longer.
1. "display Event logs on test failures" - very interesting. Will take a look.

chetan:
>  Still - is the time of deployment slowing down tests by that much? In my experience compiling takes much longer.

The current implementation only redeploys contracts before() the suite is run. This means that your tests carry their state forward which can be good in some cases but also makes it harder to write them when you don't need that feature. It's better to simply start with a clean slate before each test. 

I initially tried changing the redeploy to a beforeEach() but it was too slow to be usable -- takes about 3sec to deploy, adding about 30sec to my suite.

So a snapshot/revert is far better and adds much less overhead. I guess it should also be noted that this is a breaking change from the current functionality, when enabled. You can pass a flag to disable it like so `contract("Foo", {reset_state: false}, function(accounts) { ... });`

lib/test.es6
```diff
@@ -137,11 +308,43 @@ var Test = {
           return;
         }
 
+        // if running via the 'watch:tests' task, we want to be able to run
+        // (require) our test files repeatedly, so this is a hack to make it
+        // work. we copy each test file to a temp filename and load that
+        // instead of the original to avoid getting cached.
+        files = files.map(function(f) {
```
tcoulter: Hey @chetan. Was running your branch against my apps test, and found that this behavior here causes the tests to break. Specifically, when running files from a temp directory, all relative `require` statements within those tests files will no longer resolve (i.e., `var EternalJS = require("../app/eternaljs/eternaljs.es6");`. This is a problem because my tests rely on a dependency in order to function properly, and that dependency isn't installed globally.

I think we need to find a different solution to the caching issue (though I appreciate your comment; it helped a lot). Either we write temp files to the associated test file's directory, which is messy especially if Ctrl+C is pressed while the test is running; or we find some other solution. To clarify: Mocha runs files using require rather than through other means?

tcoulter:
@chetan This branch now exists which has your changes, and my riff on them. https://github.com/ConsenSys/truffle/tree/chetan-testing_improvments

As I mentioned, I couldn't get your tests to run in my context, for two reasons:

1. Moving test files to different directories, as mentioned above.
2. Snapshotting. Snapshotting is currently only supported by TestRPC, and Truffle is (or at least, I am) moving away from that.

There are a few code changes I have questions about. You'll notice I made a few refactoring changes, but given the following questions I assume there are more to be made:

1. Nothing sets `Truffle.can_revert` to true (unless I deleted the code to do that...). Since this is the case, why support the functionality? Are you setting it yourself within your own tests? https://github.com/ConsenSys/truffle/blob/chetan-testing_improvments/lib/test.es6#L76
1. Why does both `reset_state` and `can_revert` exist? To me they're synonymous, as you mentioned that redeploying between each test is too time consuming, so you'll only want to reset state if you also can revert. Wanting to redeploy between each test seems like an edge case. https://github.com/ConsenSys/truffle/blob/chetan-testing_improvments/lib/test.es6#L111

More on that last point: I've been thinking hard about reverting and/or redeploying between each test case, and I don't think we should, even though we can with `evm_snapshot/revert`. My main reasoning is geth can't do it, so if we provide that functionality, we'll let users write a suite of tests that likely won't run properly on geth. Instead, I think we should standardize on the assumption that within a `contract` block, all state is shared.

PR #56, opened 2015-11-15
test file whitelist, filter out irrelavent/temporary files. fix #40

### LINKED ISSUE #40: Truffle should ignore non-development files.
Truffle picks up VIM buffer files (*.swp) and other files such as (.DS_Store) when running tests, which causes it to break. Only sol & se should probably be allowed at this point.

Similarly with the tests themselves, it also does the same. Perhaps only javascript files should be allowed in that scenario.

janx:


tcoulter:
Should have merged this much earlier. Thanks!

PR #73, opened 2016-01-10
Added comment to Metacoin.sol about token standards.

simondlr:
Just to mitigate the potential that someone might use the example MetaCoin.sol code, thinking it would be compatible with the rest of the network. :)


tcoulter:
Finally did something about this. Thanks @simondlr!

PS: I changed up the comment a bit, but I think it has the same message.

PR #74, opened 2016-01-12
Bump web3 to 0.15.1 so that dapps can use .getData()

simondlr:
I wanted to use .getData(), but Truffle is on 0.14. Thus I needed to update it to 0.15 in order to use it. Nothing failed/broke. But more testing may be needed with 0.15.1.


tcoulter:
This will require a change in `truffle-default-builder` as well.

tcoulter:
Fixed in https://github.com/ConsenSys/truffle/commit/848dd70b425572421c4d5bd32b21a48216bd7746. Cheers!

PR #84, opened 2016-02-04
Upgraded solc as previous version could hang in soljson-latest.js, wh…

dominicwilliams:
solc 0.2.0-1 can hang inside some valid solidity files. 
version 0.2.1-1 fixes the problem.


tcoulter:
Thanks!

PR #85, opened 2016-02-04
Implementation of solidity library linking

aaronstjohn:
Added the ability to link solidity libraries to the contract deploy process. The process follows the following flow:

1. Builds a dependency graph for libraries of all the contracts in the deploy list. This also handles libraries that are dependent on other libraries. We do this by inspecting the compiled binaries for the `__[LibName]_____________________________` marker as documented in the solidity compiler docs.
https://solidity.readthedocs.org/en/latest/contracts.html?highlight=library 

2. Deploys the contracts and libraries based on the dependency graph such that contracts and libraries that have dependencies wait until their dependencies have been deployed and a deploy address has been captured.

3. Inject the deploy address of dependencies into the bytecode using the above marker

There is also a new example library contract and unit test in the example folder to illustrate how to use this feature in practice.

NOTE: Added a dependency for the project to a graph library used to build the dependency tree: https://github.com/cpettitt/graphlib

### DIFF
--- a/example/contracts/ConvertLib.sol
+++ b/example/contracts/ConvertLib.sol
@@ -1,0 +1,6 @@
+library ConvertLib{
+	function convert(uint amount,uint conversionRate) returns (uint convertedAmount)
+	{
+		return amount * conversionRate;
+	}
+}
--- a/example/contracts/MetaCoin.sol
+++ b/example/contracts/MetaCoin.sol
@@ -1,3 +1,5 @@
+import "ConvertLib.sol";
+
 contract MetaCoin {
 	mapping (address => uint) balances;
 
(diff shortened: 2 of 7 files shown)


tcoulter:
Little bit of a heads up: Things look great so far. I'm going to convert a fairly large project to use libraries, as it's ripe for using them and it'd be a great test of the new feature. Will have an update and experience report after that.

tcoulter:
PS: Thanks again!

aaronstjohn:
Sounds good.  We've been using this in production and it seems to be working for our needs. Let me know how it goes with your project.

tcoulter:
Merged! Thanks for your work on this, I really appreciate it. It's in `master` now. An official release with these features will go out in the next few weeks.

tcoulter:
Oh, by the way: I cleaned up a little bit of the code formatting to match the rest of the code, and fixed a few edge cases related to when deploy information was output (for instance, deploy information shouldn't be printed during test runs). Cheers!

aaronstjohn:
Awesome! I'm glad it was helpful.

-A

PR #88, opened 2016-02-10
Take mocha config from truffle.json

area:
If present in truffle.json, the contents of the 'mocha' key will be
used as the arguments passed to `new Mocha()`. This is 
useful for e.g. using a different test reporter in CI.


tcoulter:
Good call. Thanks!

PR #94, opened 2016-02-23
Add notes about pudding making contract function calls

Gustav-Simonsson:
Just a suggestion as it may not be immediately clear for new users if not already familiar with pudding.


tcoulter:
This is all taken care of in the new Truffle docs: http://truffle.readthedocs.org. Cheers!

PR #111, opened 2016-03-17
Exit with nonzero code on failure

### LINKED ISSUE #102: truffle test returns 0 statuscode on failing tests
This breaks continious integration usecases. Which is a dealbreaker for orgs/teams doing continous deployment (autodeploy when tests pass).

Standard for mocha runners is to return the number of failing tests. But anything non-zero would be fine.

This may be caused by `babel-node` executable, so this may be fixed by #80 and/or the changes due to Windows support (#81).

area:
This follows the example at https://github.com/mochajs/mocha/wiki/Using-mocha-programmatically

Fixes #102, though I haven't figured out how to write a test to check for the correct behaviour.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -442,6 +442,9 @@ var Test = {
       var runMocha = function() {
         // TODO: Catch any errors here, and fail.
         mocha.run(function(failures) {
+          process.on('exit', function () {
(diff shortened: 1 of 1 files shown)


tcoulter:
Add via slightly different implementation on master. Cheers!

PR #118, opened 2016-03-23
Fix inappropriate chars for quotation.

simondlr:


simondlr:
My fault. Had the wrong node version and it threw errors.

tcoulter:
Good point though - those characters shouldn't be there. If you want to get the PR submitted though, which I'll accept, you'll have to change those strings like this:

```
`>> File ${display_path} changed.`
```

Turns into:

```
">> File " + display_path + " changed."
```

Issue #114, opened 2016-03-21
dockerize truffle


hermanjunge:
Let's dockerize truffle. The developer mounts their files into the container everytime they run a command using `make`. Not so crazy, is not only popular in the circles this poster hangs around, but they built a whole infrastructure that way in deis ([example](https://github.com/deis/router/blob/master/Makefile), [example](https://github.com/deis/controller/blob/master/Makefile)).

I'm taking this task over, of course.

jtremback:
Why?

hermanjunge:
I don' want to install anything in my machine, except _docker_

hermanjunge:
PR #134

PR #134, opened 2016-04-12
Add docker support

hermanjunge:
- Have already the image for `1.0.2` at @consenSys [docker hub account](https://hub.docker.com/r/consensysllc/truffle/).
- Regarding the point above: It'd be nice to add somewhere an automated task to build the image on new versions or pushes to `master`.
- A new directory `docker` had to be made, as two files were needed: A `Dockerfile` and a `entrypoint.sh` file.
- This solves #114

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,37 @@
+FROM node:5.10.1-slim
+
+MAINTAINER Herman Junge "herman.junge@consensys.net"
+
+################################################################################
+#
+# So. This is what is going on:
+#
+# * We are using a slim version of the node.js docker image.
+# * git is needed during the process of installing truffle.
+# * After truffle is installed, we get rid of a number of directories.
+# * And we remove git and the programs we installed as its dependencies.
+# * The result? Just one moderate layer of files over node slim.
+#
+# * NOTE: Will install the _latest_ version of truffle. If you want to
+#         install an earlier version just add @x.y.z (version numbers)
+#         after truffle.
+#
+################################################################################
+
+RUN apt-get update -y && apt-get install -y git \
+    && npm install -g truffle \
+    && rm -rf /tmp/* /root/.npm/* /var/lib/apt/lists/* \
+    && apt-get purge -y --auto-remove git
+
+################################################################################
+#
+# We need to setup the workspace: The idea is mounting your current
+# directory here. Also, the entrypoint scripts will do all their magic
+# over this directory, if needed.
+#
+################################################################################
+
+WORKDIR "/workspace"
+
+ADD entrypoint.sh /entrypoint.sh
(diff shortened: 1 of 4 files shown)


tcoulter:
Hey @hermanjunge, can you take a look at what the String team is doing here? https://github.com/dominicwilliams/instant-dapp-ide

It's a full dapp development docker image with TestRPC and Truffle. Would love to get your opinion. I'd like to evaluate supporting their effort as that pushes maintenance of docker images over to them.

hermanjunge:
**Done the evaluation**

Cons first, as I don't want to give the idea that `I don't want other people to do the docker stuff`. Actually I love the ole' `Give a man a fish` / `Teach him to fish` saying. That way, we have more bandwith to learn more things and solve other problems. So, here is the thing:

**Cons:**

* Running more than one process inside a container is an _anti-pattern_.
* They need help to make the image lighter.
* Forces the dev to develop using Cloud9. What if the dev just wants truffle without anything (they should not be using testrpc, but a real node)? What if they use _vim_ (or, god forbid me, _emacs_), _atom_, _sublime_, etc?
* Regarding the above: A much better pattern is an array of containers orchestrated using _docker-compose_.
* This is nitpicking, but they need help in the design of the _Dockerfile_, every line is a layer, and layers must be compact.
* Finally, they should rely more on external files (and directories), they could be easily updated at build time, adding the right instruction to the build script, in order to keep a smaller _Dockerfile_.

**Pros:**

* We won't have, in the future, to maintain anything about docker. (**HUGE PRO**)
* We can support their effort to modularize their dev environment, should they _want our help_, of course. (See above what I think of fishes and code).

**TLDR:** Cool. If they let us help them, we can have this excellent dev env: powerful slim, fit, adaptable. And we can focus on solving other problems.

tcoulter:
Which would you want to do? I just spent a few days with them, and I'm sure
they'd be very open to us helping. I think if we do it right we can make it
a team effort.

Totally understand both your pros and cons. Good points all around.

hermanjunge:
I can work with them, my biggest complain is that the development environment must be modularised into several containers orchestrated using _docker compose_.

tcoulter:
I'm sure they'd love your input. Let me reach out to them to see if they're
interested and we'll go from there.

hermanjunge:
Good. @tcoulter, is your call to keep this PR open or closed. I got this very implementation in my `dockerfiles` collection :-)

oluies:
Herman, I am using this here https://github.com/combient/truffle
I guess you havent progresed with this somewhere else?

hermanjunge:
@oluies Here https://github.com/hermanjunge/dockerfiles/blob/3d6b00279f044505b576560bede0b7836e4e0133/truffle/Dockerfile

PR #146, opened 2016-04-26
add babel-polyfill

jtremback:
I want to use async/await again

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
-require("babel-register")
+require("babel-register")({
+    "presets": ["es2015", "stage-2"]
+})
 require("babel-polyfill")
 
 var web3 = require("web3");
--- a/cli.js
+++ b/cli.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
-require("babel-register");
+require("babel-register")
+require("babel-polyfill")
 
 var web3 = require("web3");
 var path = require("path");
--- a/package.json
+++ b/package.json
@@ -6,6 +6,7 @@
   "dependencies": {
     "async": "^1.4.2",
     "babel-cli": "^6.4.0",
+    "babel-polyfill": "^6.7.4",
     "babel-preset-es2015": "^6.3.13",
     "babel-preset-stage-2": "^6.3.13",
     "babel-register": "^6.7.2",


lucasconstantino:
Meanwhile you can re-configure the polyfill in your "truffle.js" file, like so:

```
require("babel-polyfill");
require("babel-register")({
  "presets": ["es2015"],
  "plugins": ["syntax-async-functions","transform-regenerator"]
});

module.exports = {
  build: {
    "index.html": "index.html",
    "app.js": [
      "javascripts/app.js"
    ],
    "app.css": [
      "stylesheets/app.css"
    ],
    "images/": "images/"
  },
  deploy: [
    // "Contract"
  ],
  rpc: {
    host: "localhost",
    port: 8545
  }
};
```

(ref: http://masnun.com/2015/11/11/using-es7-asyncawait-today-with-babel.html)

tcoulter:
I've hand-merged this into the `develop` branch, which includes the `migrations` branch changes. Can you guy ensure it's working as you expect? Note the `migrations` branch includes significant changes to Truffle and requires changes in your dapp. See here for more info: https://github.com/ConsenSys/truffle/pull/143

tcoulter:
Please close if this works for you so it updates the milestone.

tcoulter:
I've found that this addition breaks testing (and likely other things):

```
Error: Couldn't find preset "es2015" relative to directory "/Users/tim/Documents/workspace/Consensys/test"
```

This is likely because of this code added:

```
require("babel-register")({
  "presets": ["es2015", "stage-2"]
})
require("babel-polyfill")
```

I've changed it to this, and left it up to the user to specify their presets in their project's .babelrc.

```
require("babel-register");
require("babel-polyfill");
```

tcoulter:
@lucasconstantino I *really* like your solution instead. Perhaps that should be the solution for all truffle projects that want to use babel. The Truffle requirement on babel is overkill since Truffle doesn't use babel internally.

ethers:
thanks @lucasconstantino solution works for me

cdyfng:
You can update your node to the newest version by following step. 
```
1. node -v     #may display version less then v8.0.0
2. sudo npm cache clean -f 
3. sudo npm install -g n
4. node -v  #check the version , mine is v8.0.0 now 
```
Test ok on Mac Os. 
Then you can use async/await in your js code.

PR #163, opened 2016-05-20
Update testing.md

raineorshine:
You can return the promise directly and not have to deal with the callback. Mocha will wait for the promise to resolve or reject.


redsquirrel:
Nice, thanks for pointing this out. I just used it to simplify the live-libs tests: https://github.com/ConsenSys/live-libs/commit/4f6d246b907ebf51d6476bf4ece0898a393f18a0

tcoulter:
What, go mocha! I didn't know this either. Thanks!

PR #164, opened 2016-05-21
Update contracts.js

FrankMN:
when deployed my project. I got the error "Could not find contract 'MetaCoin' for linking. Check truffle.json." , then I change config file truffle.js updating the file name to my new file which works. Once I was trying to locate the file "truffle.json" while not found. I think if the error could indicate the exact config file name it would be better.


tcoulter:
Good catch. Thanks!

PR #171, opened 2016-06-06
[WIP] Updates example contract to the latest token standard

GregoryBetz:
Running `truffle test` works without problem, though the speed could be improved, I think. The frontend files app.js and index.html still need some slight work.
# **DO NOT MERGE** 
_yet_

:fire: 🚒 💣


example/app/javascripts/app.js
```diff
@@ -1,16 +1,28 @@
...
+  var _initialAmount = 10000;
+  var _tokenName = 'Truffle Token';
+  var _decimalUnits = 2;
+  var _tokenSymbol = 'TRFT';
+
+  truf.initiateToken(_initialAmount, _tokenName, _decimalUnits, _tokenSymbol {from: account})
+}
+
 function refreshBalance() {
-  var meta = MetaCoin.deployed();
```
tcoulter: Can you rename this variable to a full word? `truffle` would be the obvious choice, but then seems like a misnomer (because it's not Truffle itself). How about `token`?

example/app/javascripts/app.js
```diff
@@ -1,16 +1,28 @@
...
+  var truf = TruffleToken.deployed();
```
tcoulter: Can you rename this variable to a full word? truffle would be the obvious choice, but then seems like a misnomer (because it's not Truffle itself). How about token?

example/test/truffleToken.js
```diff
@@ -0,0 +1,66 @@
+contract('TruffleToken', function(accounts) {
+  it("should put 10000 TruffleTokens in the first account", function(done) {
+    var truf = TruffleToken.deployed();
```
tcoulter: Same here. This is code coders will see. I want to remove any source of confusion.

GregoryBetz: Okay sounds good. I'll change it to `var token` .

PR #172, opened 2016-06-07
Spelling fix.

zygisx:


redsquirrel:
That's not an error. That looks correct: http://www.dictionary.com/browse/ample

PR #184, opened 2016-06-29
Fix library linking for contracts that depends on multiple libraries

thiagodelgado111:
If we have a contract that uses more than 1 library, the "link" from the first libraries are overwritten by the last ones

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -36,6 +36,7 @@ var Config = {
           build: {},
           include_contracts: true,
           deploy: [],
+          libraries: [],
           after_deploy: [],
           rpc: {},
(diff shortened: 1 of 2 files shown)


tcoulter:
Hi @thiagodelgado111. You should check out the `develop` version of Truffle. Specifically, the migrations feature removes the need for this PR. Give it a try! http://truffle.readthedocs.io/en/develop/getting_started/migrations/

tcoulter:
FYI: There was a similar issue on Truffle v2.0.0. Upgrade to the (just released) version, v2.0.2, to support contracts with multiple libraries. Cheers!

PR #224, opened 2016-08-04
Have a nonzero exit code on test failures

### LINKED ISSUE #223: Exit code is 0 even with tests failing
This is affecting v2.0.4
when running `truffle test` which has a failing test it incorrectly returns `0` as the exit code

area:
It looks like this is all that's needed to fix #223.

### DIFF
--- a/lib/test.js
+++ b/lib/test.js
@@ -326,7 +326,7 @@ var Test = {
 
         mocha.run(function(failures) {
           console.warn = warn;
-          callback(null, failures);
+          callback(failures);
         });
       });
     });
--- a/lib/test.js
+++ b/lib/test.js
@@ -326,7 +326,10 @@ var Test = {
 
         mocha.run(function(failures) {
           console.warn = warn;
-          callback(failures);
+          if (failures) {
+            return callback(new ExtendableError(failures.toString() + " test(s) failed"));
+          }
+          callback();
         });
       });
     });


tcoulter:
Interesting typo. :clap: Have you tested for any repercussions? i.e., does it return a 0 exit code when there are no failures?

area:
Yeah, it does, I've been testing with a very basic test, and the exit code is zero/nonzero as appropriate based on the test failing or passing:

```
/* eslint-env node, mocha */

contract('Contract', function (accounts) {
  describe('test', function(){
    it('should pass or fail', function(done) {
      assert.equal(1,2,'what did you expect');
      done();
    });
  });
});
```

When there's a failure, though, I've just noticed that the exit code seems to get printed to the console as the final output, which I haven't tracked down yet.

area:
Ah, the extra output is due to https://github.com/ConsenSys/truffle/blob/master/cli.js#L20. So given that it's an error that we know about, maybe something like:

```
mocha.run(function(failures) {
  console.warn = warn;
  if (failures){
    return callback(new ExtendableError(failures.toString() + " tests failed"))
  }
  callback();
});
```

is more appropriate?

tcoulter:
@area Apologies for the late response, just saw your message (I get way too much email). `truffle test` prints out enough error messages, so instead of throwing a printable error message, I added the ability for tasks to return a number as an error. If a number is returned, that will be the command line's exit code. In that case, it's the task's responsibility to print out the error message (or not, if they so choose). This should give you the non-zero exit code without an extraneous error message.

Thanks again!

area:
No worries, thanks for getting this merged in.

Issue #233, opened 2016-08-10
ReferenceError: network_id is not defined


mpolci:
When I give a wrong network name I get this error for an undefined `network_id` variable in the code:

```
$ truffle migrate --network wrong_name
/usr/lib/node_modules/truffle/lib/config.js:109
    throw new ConfigurationError("Cannot find network '" + network_id + "'");
                                                           ^

ReferenceError: network_id is not defined
    at Config.getRPCConfig (/usr/lib/node_modules/truffle/lib/config.js:109:60)
    at Config.props.provider (/usr/lib/node_modules/truffle/lib/config.js:83:26)
    at Config.Object.defineProperty.get [as provider] (/usr/lib/node_modules/truffle/lib/config.js:97:35)
    at /usr/lib/node_modules/truffle/node_modules/ether-pudding/index.js:206:18
    at tryToString (fs.js:455:3)
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:442:12)

```

eduardonunesp:
Maybe it's because you haven't specificated that network id in your `truffle.js` file, take a look at http://truffle.readthedocs.io/en/latest/advanced/configuration/#networks for more information.

tcoulter:
This has been fixed in Truffle 3.0. The new error message looks like this:

```
Tims-MacBook-Pro:test3 tim$ ../truffle/cli.js migrate --network notconfigured
Unknown network "notconfigured". See your Truffle configuration file for available networks.
```

Thanks!

cmycaro:
I also got the same error~ the undefined thing is the contract 
`Caros-MacBook-Pro:~ caro$ truffle migrate
Running migration: 2_deploy_contracts.js

/Users/caro/migrations/2_deploy_contracts.js:5
   deployer.deploy(MyAdvancedToken);
                   ^
ReferenceError: MyAdvancedToken is not defined
    at module.exports (/Users/caro/migrations/2_deploy_contracts.js:5:20)
    at /usr/local/lib/node_modules/truffle/lib/migrate.js:109:7
    at /usr/local/lib/node_modules/truffle/lib/require.js:82:7
    at tryToString (fs.js:455:3)
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:442:12)`

tcoulter:
[...]

PR #234, opened 2016-08-10
Fix for the #233

eduardonunesp:
The errors happens when the user tries to migrate an no specified network,
the fix covers to test the network configuration before compiler

Also includes the ConfigurationError file into tasks.js


tcoulter:
This has been fixed in Truffle 3.0. The new error message looks like this:

```
Tims-MacBook-Pro:test3 tim$ ../truffle/cli.js migrate --network notconfigured
Unknown network "notconfigured". See your Truffle configuration file for available networks.
```

Thanks!

PR #238, opened 2016-08-13
Fix truffle console commands output and tab completion

mpolci:


tcoulter:
Was just going to ask what this solved, but then ran into issues which this fixed. Thanks!

PR #258, opened 2016-09-02
[beta] Solidity Unit Tests

### LINKED ISSUE #141: Truffle can't find contract outside of `contracts` directory
If you import a contract that lives outside of the `contracts` directory, Truffle will have trouble finding it.

tcoulter:
This PR adds Solidity unit tests alongside your Javascript unit tests! 👏  🎉 💃 
## Installation

```
$ npm uninstall -g truffle
$ git clone https://github.com/ConsenSys/truffle.git
$ cd truffle
$ git checkout solidity-tests
$ npm install -g .
```
## Notable Changes
- Solidity unit tests! (of course :) )
- You can now import contracts outside of the `contracts` directory (fixes #141). Example:
  
  ```
  import "../node_modules/somemodule/Contract.sol";
  ```
- **BREAKING CHANGE:** Because of the above change, all imports for all your contracts need to be explicitly relative (or absolute; see below). 
  
  Good:
  
  ```
  import "./SomeDependency.sol";
  ```
  
  **NOT** good:
  
  ```
  import "SomeDependency.sol";
  ```
  
  Note that in the second case there was no `./` in front of `SomeDependency.sol;`. Truffle and the solidity compiler can no longer assume you meant `./` if not specified. 
  
  Though rarely needed, this change now allows imports to be absolute paths where it didn't in the past.
- An example solidity unit test is now included with `truffle init`.
## Solidity Unit Testing

Solidity test contracts live alongside Javascript tests as `.sol` files. When `truffle test` is run, they will be included in the mocha run with a separate test suite per test contract. These contracts maintain all the benefits of the Javascript tests: namely a clean slate per test suite, access to deployed contracts via migrations, runnable on any Ethereum client, and usage of snapshot/revert features (if supported by your Ethereum client) for increased speed. An example solidity unit test looks like this:

```
import "truffle/Assert.sol";
import "truffle/DeployedAddresses.sol";
import "../contracts/MetaCoin.sol";

contract TestMetacoin {
  function testInitialBalanceUsingDeployedContract() {
    MetaCoin meta = MetaCoin(DeployedAddresses.MetaCoin());

    uint expected = 10000;

    Assert.equal(meta.getBalance(tx.origin), expected, "Owner should have 10000 MetaCoin initially");
  }

  function testInitialBalanceWithNewMetaCoin() {
    MetaCoin meta = new MetaCoin();

    uint expected = 10000;

    Assert.equal(meta.getBalance(tx.origin), expected, "Owner should have 10000 MetaCoin initially");
  }
}
```

And when running the test:

```
$ truffle test
Compiling ConvertLib.sol...
Compiling MetaCoin.sol...
Compiling ../test/TestMetacoin.sol...

  TestMetacoin
    ✓ testInitialBalanceUsingDeployedContract (61ms)
    ✓ testInitialBalanceWithNewMetaCoin (69ms)

  2 passing (3s)
```

There are some very important structural items to discuss: 
- First, your assertion functions are provided by the `truffle/Assert.sol` library. This is the default assertion library, however you can include your own assertion library so long as the library outputs the correct events to the test runner. You can find all available assertion functions in [Assert.sol](https://github.com/ConsenSys/truffle/blob/solidity-tests/lib/testing/Assert.sol). These functions are meant to mimic those available in your Javascript tests, and more documentation will be written about each function in the future. 
- The addresses of your deployed contracts (contracts that were deployed in your migration) are available through the `truffle/DeployedAddresses.sol` library. This is provided by Truffle and is recompiled and relinked before each suite is run to provide your tests with a clean slate deployment. This library provides functions for all of your deployed contracts, in the form of:
  
  ```
  DeployedAddresses.<contract name>();
  ```
  
  This will return an address that you can then use to access that contract.
- In order to use the deployed contract, you'll have to import the contract code first. Notice `import "../contracts/MetaCoin.sol";` in the example. This import is relative to the test contract, which exists in the `./test` directory, and it goes outside of the test directory in order to find the MetaCoin contract. It then uses that contract in a test function below to cast the address to the `MetaCoin` type.
- All test contracts must start with `Test`, using an uppercase `T`. This distinguishes this contract apart from test helpers and project contracts, letting the test runner know it represents a test contract.
- Like test contract names, all test functions must start with `test`, all lowercase. Each test function is executed as a single transaction, in order of appearance in the test file (like your Javascript tests). Assertion functions provided by `truffle/Assert.sol` trigger events that the test runner evaluates to determine the result of the test. Assertion functions return a boolean representing the outcome of the assertion which you can use to return from the test early to prevent execution errors (i.e, errors that the TestRPC will expose). 
- You are provided many test hooks, shown in the example below. These hooks are `beforeAll`, `beforeEach`, `afterAll` and `afterEach`, which are the same hooks provided by mocha in your Javascript tests. You can use these hooks to perform setup and teardown actions before and after each test, or before and after each suite is run. Like test functions, each hook is executed as a single transaction. Note that some complex tests will need to perform a significant amount of setup that might overflow the gas limit of a single transaction; you can get around this limitation by creating many hooks with different suffixes, like in the example test below:
  
  ```
  import "truffle/Assert.sol";
  
  contract TestHooks {
    uint someValue;
  
    function beforeEach() {
      someValue = 5;
    }
  
    function beforeEachAgain() {
      someValue += 1;
    }
  
    function testSomeValueIsSix() {
      uint expected = 6;
  
      Assert.equal(someValue, expected, "someValue should have been 6");
    }
  }
  
  ```
  
  This test contract also shows that your test functions and hook functions all share the same contract state. You can setup contract data before the test, use that data during the test, and reset it afterward in preparation for the next one. Note that just like your Javascript tests, your next test function will continue from the state of the previous test function that ran. 
- You **do not** need to extend from any `Test` contract like in other solidity testing frameworks. This is to make writing solidity unit tests easier, and should allow for more extensibility in the future with less hassle.
- Like Javascript tests, failing tests return the events that were fired (minus assertion events) so you can get a better sense of how your contracts-under-test performed.
## Further Work / Testing

I still need to verify that these tests will run on a geth private network. As well, I need to verify you can bring in your own assertion library. If you'd like to do this verification on your own and report back that would be much appreciated.

It would also be a good exercise for people to run their existing Javascript tests and determine that they still function as expected. Compilation and migrations got minor internal changes as well, so verifying your contracts compile and deploy correctly would be very helpful. 
## Special Thanks

Special thanks goes out to @PeterBorah for letting me ask him extensive questions about the framework. As well, the assertion library was provided by [SolUnit](https://github.com/smartcontractproduction/sol-unit) that, although deprecated, significantly reduced the amount of work needed to implement this feature. 

Cheers! Feedback welcome!


simondlr:
Thanks for this!

Something to add here. This currently doesn't support 0.4.x. But I assume it will when all is merged. To add it, I had to add pragma Assert.sol & add this line to deployed.js when creating DeployeAddresses.sol (before starting with "library ...." )

``` source += "pragma solidity ^0.4.3;" + "\n\n";```

Also. An aside. Solc 0.4.3 introduced warnings when inline assembly doesn't balance. I've been using Nick Johnson's string utils and they throw these warnings now.

Example of a warning.

```
/Users/simondlr/code/ethereum/ujo-contracts/contracts/StringUtils.sol:468:26: Warning: Inline assembly block is not balanced. It leaves 3 item(s) on the stack.
                assembly {
```

This has to be fixed in stringutils.sol. However, I thought if it would be possible to have an option to try and continue on warnings, and not just err if that's the case (probably another issue for the future).

PR #263, opened 2016-09-08
Beta Features (2.0 -> 3.0)

tcoulter:
Notable Changes:
- Solidity Unit Tests
- Package Management ([docs](http://truffle.readthedocs.io/en/beta/getting_started/packages/))
- `import` statements in solidity code must be explicitly relative or absolute (must start with `./`, `../` or `/`) otherwise will be interpreted as a package.
- EtherPudding 4.0 updates: 
  - `myContract.someFunction()` now resolves to a result object instead of a transaction hash. This result object provides the transaction hash, deserialized logs and transaction receipt.
  - `.deployed()` is now thennable, and it'll detect the correct network before creating an instance.
  - `*` network is now the fallback network, and any network can be set as the default. The default network will be use when requiring the contract for the first time.

More to come.


maraoz:
Looking at http://truffleframework.com/tutorials/package-management

I feel like a better interface to `import "example-truffle-library/contracts/SimpleNameRegistry.sol";` could be `from "example-truffle-library" import SimpleNameRegistry;`

or just `import "example-truffle-library/SimpleNameRegistry.sol";`, removing the `contracts` and assuming we're looking there for contracts, by convention

Issue #239, opened 2016-08-14
`truffle test` doesn't allow network selection with `--network`


mrbodoia:
The Truffle documentation suggests that `truffle test` can be run with the `--network` option to specify a network from the configuration file on which tests should be run. However, when running the `truffle test` command, Truffle explicitly sets the network to be the one named "test" after loading the configuration file: https://github.com/ConsenSys/truffle/blob/48bd90d06fef9a90ec058159f01e192ea8c36e51/lib/tasks.js#L151.

This may be intended behavior, since it adds a safeguard to ensure that a battery of tests (potentially costing lots of gas or ether) isn't accidentally run on the main network. However, it also prevents the user from easily testing on multiple networks (e.g. TestRPC for development and Morden prior to deployment). I think it would be more useful if users could specify the network for testing, and in either case it would be good to make sure the documentation for `truffle test` matches the actual behavior.

chevdor:
I noticed the same and agree with the comment from @mrbodoia 
As far as security is concerned, running the tests requiring ETH will require the account to be unlocked.  It is probably a very bad idea on to unlock accounts on mainnet.

chevdor:
The fact that it is not implemented may be related to #261: when using a real net, you will often run into timeout issues and the timeouts are currently not configurable.

tcoulter:
This should be fixed in Truffle 3.0. Cheers!

pragmaticadam:
It does not work in version 
Truffle v5.0.3 (core: 5.0.3)
Solidity v0.5.0 (solc-js)
Node v11.6.0

PR #267, opened 2016-09-20
Fix #261 + #239

### LINKED ISSUE #261: Make test runner's BEFORE_TIMEOUT configurable
I've received some complaints that the pre-deploy steps aren't completing within 120s on some slow test nets. We should make this value configurable.

chevdor:
Fixes for:
- #261 
- #239


chevdor:
@tcoulter you closed this PR without merging while it fixes issues still pending.
Could you comment please?

PR #268, opened 2016-09-20
Open url

chevdor:
Before this PR, calling "serve" shows something like:
`... Serving on port 8080`

After this PR:
- it shows `Serving at http://localhost:8080/` (which makes it easier to open/copy the url)
- it opens the URL


tcoulter:
This is convenient, but I don't want to make `serve` more complex. If you'd like a more complex build system for web apps, consider webpack integration, which will give you a bunch more features than `serve` can ever provide. Thanks for the report!

PR #277, opened 2016-09-22
Fix #276 : Use Example.deployed() to retrieve deployed contract in tests

### LINKED ISSUE #276: Tests created with 'truffle create:test' result in an error
The template for Mocha tests attempts to retrieve the address of the contract using:
```
var example = Example.at(Example.deployed_address);
```

However, this results in an error:
```
Error: Invalid address passed to Example.at(): undefined
```

adklempner:
Fixes #276 
Retrieving the deployed contract this way does not result in an error.

### DIFF
--- a/templates/example.js
+++ b/templates/example.js
@@ -1,6 +1,6 @@
 contract('Example', function(accounts) {
   it("should assert true", function(done) {
-    var example = Example.at(Example.deployed_address);
+    var example = Example.deployed();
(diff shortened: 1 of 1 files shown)


tcoulter:
Good catch!

danielyaa5:
What caused this? I see this `Example.at(Example.deployed_address);` online many places.

Issue #281, opened 2016-09-26
truffle create:test should create a Solidity test as well


adklempner:
With the addition of Solidity tests in beta, the create:test command should be updated to create both a js and sol test file

adklempner:
A full implementation of this feature will probably require finding a solution to #141

gnidan:
@adklempner do you still want this? If not, please feel free to close. Thanks!

PR #283, opened 2016-09-28
Create a Solidity test from template when running "truffle create:test"

adklempner:
Adds feature #281 

Also creates an actual test, not an auto-pass.


gnidan:
Hi @adklempner, sorry for ignoring this. If you're interested, could you please re-open? With the monorepo change, this would need to be rebased. Thank you!

PR #284, opened 2016-09-29
Support --network argument for `truffle test`

area:
I have a `truffle.js` that looks like this:

```
module.exports = {
  rpc: {
    host: 'localhost',
    gasPrice: 20e9,
  },
  mocha: {
    reporter: 'spec',
  },
  networks: {
    "coverage":{
        network_id:1919191,
        port:  8546,
    }
  }
};
```

And while it works for `truffle migrate --network coverage`, it failed to work for `truffle test --network coverage`. I feel like if the user is specifying the network explicitly, that should be respected. Have I overlooked a knock-on consequence o not running the 'test' command on the 'test' network?

This PR leaves the current functionality for just the `truffle test` command i.e. network is set to 'test'.

### DIFF
--- a/lib/tasks.js
+++ b/lib/tasks.js
@@ -157,7 +157,9 @@ createTask('exec', "Execute a JS file within truffle environment", function(opti
 // More to come.
 createTask('test', "Run tests", function(options, done) {
   var config = Truffle.config.detect(options);
-  config.network = "test";
+  if (config.network === "default") { 
+    config.network = "test";
+  }
 
   var files = [];
 


area:
Fixes #239. This implementation (at first glance) seems to be less intrusive than the proposed fix in #267.

area:
Is there any reason why this isn't being considered, or anything that can be done to improve its chances of being merged? It would really help people to use [Solcover](https://github.com/JoinColony/solcover) seamlessly!

tcoulter:
This has been fixed in Truffle 3.0. Please reopen if you continue to run into the issue.

Apologies for the delay. This wasn't considered directly because Truffle was in the middle of a large refactoring. Thanks, as always!

PR #288, opened 2016-10-14
Adding ability to change RPC protocol

alex-miller-0:
Needed to switch RPC to HTTPS and thought this might be beneficial for the public package.
- `protocol` can be added to RPC options. If none is provided, `http` is used.

### DIFF
--- a/lib/provider.js
+++ b/lib/provider.js
@@ -33,11 +33,12 @@ module.exports = {
 
   create: function(options) {
     var provider;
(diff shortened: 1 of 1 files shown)


tcoulter:
This is more configuration then I'd like to add to this feature. Instead, you should create your desired provider and pass in that provider to your Truffle configuration instead of `host` and `port`. See configuration docs here: http://truffleframework.com/docs/advanced/configuration#networks

PR #289, opened 2016-10-17
Update tasks.js

standerson:
Update tasks.js to allow "serve" and "watch" commands from the console.


standerson:
Not sure if there is a reason why one is not suppose to run the serve and watch commands from the console, so I created this pull request.

tcoulter:
This PR is now out of date. My guess is the functionality it was providing is now available. Please open a new ticket if there's still an issue. Thanks!

PR #292, opened 2016-10-28
Fix error handling in truffle console

federicobond:
`done` is not defined in this context. It's probably a leftover from a previous refactor.

These kinds of errors can be trivially recognized by a linter. If you are interested, I can configure [eslint](http://eslint.org/) for this project in another pull request.

### DIFF
--- a/lib/repl.js
+++ b/lib/repl.js
@@ -22,7 +22,10 @@ TruffleInterpreter.prototype.start = function() {
   web3.setProvider(options.provider);
 
   this.provision(function(err) {
-    if (err) return done(err);
+    if (err) {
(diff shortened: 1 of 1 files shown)


tcoulter:
Good catch, and good suggestion. Using `process.exit()` may not be the way to solve this specific error handling case, as ideally the console should never quit. I'll take a closer look at this soon.

federicobond:
The most common case for this is starting the console without an RPC client up. I didn't see much of a case for running the console without connecting to a client, so I figured it was OK to error out.

federicobond:
Closing this since it's fixed in the latest version.

PR #293, opened 2016-10-31
Change `truffle deploy` to `truffle migrate` in docs. 

raineorshine:
Change `truffle deploy` to `truffle migrate`.

### DIFF
--- a/docs/getting_started/contracts.md
+++ b/docs/getting_started/contracts.md
@@ -159,7 +159,7 @@ Each contract abstraction has a method called `deployed()`, which you saw used a
 var meta = MetaCoin.deployed();
 ```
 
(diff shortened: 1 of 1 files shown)


raineorshine:
I assume I can commit to master and then you can merge the change into beta? Let me know if there is a different procedure for patches that need to be applied to both branches.

tcoulter:
Good catch. 

It looks like I can merge directly from your master branch, which I didn't realize until just now (the merge button is nice and green). Either way, I just made you a contributor so you can create branches in the main truffle repository. Hopefully that'll make things just a bit easier, especially if you have other branches you'd like to merge in. 

Of course, the PR process is preferred (I'm sure I don't have to say that). Thanks, as always.

PR #295, opened 2016-11-01
Only select migrations files that end in .js

intuitmike:
I'm using VIM with default swap file behavior. I was getting tired of running into migration failures when Truffle picked up the .swp files for processing. Hence, filtering for .js files.


tcoulter:
This has been fixed in `truffle-migrate` 1.0.0, and will go out with the next release of Truffle. Thanks!

PR #299, opened 2016-11-08
Fix broken exec command

ted-piotrowski:
## Issue

Running `truffle migrate` produces error

## Steps to reproduce

1. Create a script according to https://truffle.readthedocs.io/en/latest/getting_started/scripts/
2. Add `deployer.exec` command to migration jobs

```
module.exports = function(deployer) {
  deployer.exec('path/to/script');
}
```
3. Run `truffle migrate`

## Expected behavior

Script runs and exits successfully

## Actual behavior

```
Error encountered, bailing. Network state unknown. Review successful transactions manually.
Error: Expected parameter 'contracts_build_directory' not passed to function.
    at /Users/teddy/Projects/btl/truffle-webpack-demo-clone/node_modules/truffle/lib/expect.js:5:15
    at Array.forEach (native)
    at Object.options (/Users/teddy/Projects/btl/truffle-webpack-demo-clone/node_modules/truffle/lib/expect.js:3:19)
    at Object.exec (/Users/teddy/Projects/btl/truffle-webpack-demo-clone/node_modules/truffle/lib/require.js:89:12)
    at /Users/teddy/Projects/btl/truffle-webpack-demo-clone/node_modules/truffle/lib/deployer.js:101:17
    at /Users/teddy/Projects/btl/truffle-webpack-demo-clone/node_modules/truffle/lib/deployer.js:100:14
    at /Users/teddy/Projects/btl/truffle-webpack-demo-clone/node_modules/truffle/lib/deferredchain.js:20:15
```

## Environment

- Truffle 2.1.1
- Node 7.0.0
- TestRPC 3.0.0
- macOS Sierra
- npm 3.10.8

### DIFF
--- a/lib/deployer.js
+++ b/lib/deployer.js
@@ -100,7 +100,7 @@ var Actions = {
       return new Promise(function(accept, reject) {
         Require.exec({
           file: file,
-          contracts: Object.keys(deployer.known_contracts).map(function(key) {
+          contracts_build_directory: Object.keys(deployer.known_contracts).map(function(key) {
             return deployer.known_contracts[key];
           }),
           network: deployer.network,


lib/deployer.js
```diff
@@ -100,7 +100,7 @@ var Actions = {
       return new Promise(function(accept, reject) {
         Require.exec({
           file: file,
-          contracts: Object.keys(deployer.known_contracts).map(function(key) {
+          contracts_build_directory: Object.keys(deployer.known_contracts).map(function(key) {
```
igetgames: I think this should be the other way around. According to the require.exec() function in require.js, it accepts an optional array of contracts. If that array is specified, it is used instead of the contracts under options.contracts_build_directory (implicitly inside contracts.provision).

So the fix would be to add "contracts" to the expected options array used in require.exec().

tcoulter:
`deployer.exec()` is no longer available and hasn't been for awhile, due to changes in Truffle. Your best bet to replace it is to write a separate node module that exports a function, and calling that function from your migrations.

Let me know if this is currently still an issue for you. Cheers.

PR #306, opened 2016-11-26
added bytecode and abi generation to compiling contracts

phillyfan1138:
I find it very helpful for the abi and the bytecode to be generated as well as the javascript functions on contract compilation.  This PR generates the abi and bytecode every time the contract is compiled.


tcoulter:
This has been completely revamped in Truffle 3.0. You should have what you need. Thanks!

PR #310, opened 2016-12-01
Upgrade to solc 0.4.6.

### LINKED ISSUE #308: Update solc to minimum 0.4.5.
0.4.5+ fixes a issue in the Solidity compiler where it incorrectly would throw inline assembly warnings. I had to build a custom Truffle to ignore this.

I asked Christian to update solc on npm to 0.4.6 (https://github.com/ethereum/solc-js/issues/74), but somehow stopped short at 0.4.5. However, 0.4.5 has this fix so that I don't have to run a custom Truffle anymore to those warnings.

If npm updates to 0.4.6, I'll update the beta to 0.4.6 and submit a PR as that is the branch I'm currently using.

simondlr:
This is primarily to go above 0.4.5 which caused incorrect inline assembly warnings to be thrown upon compilation, which required a custom Truffle to ignore.

This fixes it.

Fixes #308

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "truffle",
   "namespace": "consensys",
-  "version": "3.0.0-3",
+  "version": "3.0.0-4",
   "description": "Truffle - Simple development framework for Ethereum",
   "dependencies": {
     "async": "^1.4.2",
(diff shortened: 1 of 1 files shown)


simondlr:
Any desire to add this? Have to link to custom Truffle installations, which is not ideal. :)

tcoulter:
FYI: This only updates `beta`. Was that your intent?

tcoulter:
v3.0.0-4 published.

simondlr:
@tcoulter yeah, only beta for now, since that is what I was using. But, yes, master should probably also get 0.4.6.

PR #311, opened 2016-12-01
Fix "Error: Expected parameter 'contracts_build_directory' not passed to function." 

tramonex-chris:
When truffle migrate, I got this error:

`Contract: Test "before all" hook: prepare suite:
     Error: Expected parameter 'contracts_build_directory' not passed to function.
      at /home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/lib/expect.js:5:15
      at Array.forEach (native)
      at Object.Expect.options (/home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/lib/expect.js:3:19)
      at Object.Require.exec (/home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/lib/require.js:89:12)
      at /home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/lib/deployer.js:101:17
      at new Promise (/home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/node_modules/core-js/modules/es6.promise.js:191:7)
      at /home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/lib/deployer.js:100:14
      at /home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/lib/deferredchain.js:20:15
      at run (/home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/node_modules/core-js/modules/es6.promise.js:87:22)
      at /home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/node_modules/core-js/modules/es6.promise.js:100:28
      at flush (/home/chris/.nvm/versions/node/v5.12.0/lib/node_modules/truffle/node_modules/core-js/modules/_microtask.js:18:9)
      at _combinedTickCallback (internal/process/next_tick.js:67:7)
      at process._tickDomainCallback (internal/process/next_tick.js:122:9)`

This is caused because 'contracts_build_directory' is not set in deploy.js library.

PR #313, opened 2016-12-05
Fix Deployer.exec()

igetgames:
Here is what I believe to be the correct fix to the attempted `Deployer.exec()` fixes in #299 and #311.

When `Require.exec()` is called via `Deployer.exec()`, it will be supplied a contracts options that contains the list of contracts known to the Deployer. If not, then the contracts_build_directory is expected to specify where the contract artifacts exist. This PR makes that behavior explicit and restores the `Deployer.exec()` function.

### DIFF
--- a/lib/expect.js
+++ b/lib/expect.js
@@ -22,15 +22,11 @@ var Expect = {
       return t + value;
     });
 
-    if (total == 1) return;
+    // If this doesn't work in all cases, perhaps we should
+    // create an expect.onlyOne() function.
+    if (total >= 1) return;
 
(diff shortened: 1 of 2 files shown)


igetgames:
I cherry-picked the commits that add the Expect.one() method.

tcoulter:
Hi there.

`deployer.exec()` is no longer supported. The best way to work around this issue is to create a node module that exports a function you can run within your migration. 

Please fill out a new ticket if this continues to be an error, or catch us in our gitter channel! 

Cheers.

PR #332, opened 2017-01-26
Be more explicit when testing bytes32 and address.

christoph2806:
When testing bytes32 or address, the tested parameters should be reported.


tcoulter:
Hi @christoph2806. This looks like a great addition. Can you provide me with some examples about what the output used to look like before and what it now looks like after this change? 

Appreciate it. Thanks!

gnidan:
Closing this since it's now rather out of date. If you're still interested, please re-open. Thank you!

PR #335, opened 2017-01-28
Fix config export to truffle-config instead of ./lib/config.js

GNSPS:
Just some clean up: this was just forgotten behind as lib/config.js was removed.


tcoulter:
This has been fixed on master. Instead of adding `truffle-config`, I removed all references to modules that can be required by themselves (i.e., `truffle-config`, `truffle-migrate`, etc.). Thanks for the PR!

PR #345, opened 2017-02-05
Fix minor typo

fabioberger:
Encountered this error message while running `truffle console` and noticed the typo :)


tcoulter:
Good catch. Thanks!

PR #350, opened 2017-02-07
gentlejack

gentlejack:
consensys-truffle


tcoulter:
I'm not sure what to make of this PR. It contains changes that are old as well as what looks like commits we don't want. Can you tell me what you're trying to do here and why the commit log is so long?

cicorias:
I'd be very afraid

tcoulter:
I didn't notice this: `ConsenSys:ethers-patch-1 from gentlejack:master` - so they're asking to merge into a branch. But I wonder why.

cicorias:
The contributor's history seems nonsensical https://github.com/ethereum/go-ethereum/pull/3652/files

PR #373, opened 2017-03-12
Fix #365: Event logging on test failure

### LINKED ISSUE #365: Events are no longer output on test failure

masonforest:
Hey @tcoulter and co!

Here’s a first pass at fixing #365

It looks like this broke when [these lines](https://github.com/ConsenSys/truffle/blob/master/lib/testing/testrunner.js#L47-L70) got commented out. It also looks like we no longer have access to `project_contracts` in that scope :/.

What I did here was read in all the contract files in place. I’m not thrilled about this because we are reading the contract files in again in `testsource.js`.

I see a few options:

1. Move all the file reading stuff into `testrunner.js`and pass the `mapping` object to TestSource’s constructor

2. Move all of the file reading stuff into the constructor of TestSource and add a getter method that returns the contracts that could be called from `testrunner.js`

3. Some combination of the two

4. Something I haven’t thought of

Currently I’m leaning toward #2 but I’d love hear your thoughts.

I picked up this bug because I thought it seemed like it had the highest work to reward ratio. Let me know if there’s something else that’s higher priority that’s available.

Thanks, and looking forward to helping out and chatting more! 😄

### DIFF
--- a/lib/testing/testrunner.js
+++ b/lib/testing/testrunner.js
@@ -4,6 +4,12 @@ var Migrate = require("truffle-migrate");
 var TestResolver = require("./testresolver");
 var TestSource = require("./testsource");
 var expect = require("truffle-expect");
+var contract = require("truffle-contract");
+var SolidityCoder = require("web3/lib/solidity/coder.js");
+var path = require("path");
+var _ = require("lodash");
+var Promise = require("bluebird");
+var fs = Promise.promisifyAll(require("fs"));
 
 function TestRunner(options) {
   options = options || {};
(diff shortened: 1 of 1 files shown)


masonforest:
Also, I added `bluebird` as a dependency. Let me know if you're trying to limit adding additional dependencies and I can do without. It does have some cool features though which make working with promises much nicer.

tcoulter:
Hey there, thanks for the PR! I removed bluebird as I didn't want to add the extra dependency. This has been merged to master. Thanks again!

PR #379, opened 2017-03-22
Specify packager info `x-via` property in ethpm manifest

gnidan:
For example, `"truffle v3.1.9"`

Include packager information in manifest when publishing to ethpm.

Additionally, make ethpm tests conform better to the extant requirements of the package lib, including using `ropsten` network name, and include test for publishing.


gnidan:
CI currently failing because this PR is dependent on https://github.com/ethpm/ethpm-js/pull/1

PR #395, opened 2017-04-26
Explicit proccess.exit after successful command run

gagarin55:
Why it needs explicitly execute proccess.exit ? Because truffle loads its configuration file. Somebody can put in his config some code which prevent finishing execution. For instance, if we use custom web3-provider-engine it will run timer to track new blocks.

### DIFF
--- a/cli.js
+++ b/cli.js
@@ -34,5 +34,7 @@ command.run(process.argv.slice(2), options, function(err) {
       }
     }
     process.exit(1);
+  } else {
+    process.exit(0);
   }
 });


tcoulter:
Hi there. Thanks for writing.

For specific design reasons, we don't run `process.exit()` on a successful command. That said, I do see why you might want it. 

If you're writing a timer that keeps the process open, the best way to handle it is by dereferencing the timer in node. This will make sure the process closes whether or not your timer is still going. Check out [this link](https://stackoverflow.com/questions/5924160/node-js-settimeout-that-does-not-keep-the-process-running) for more. 

Cheers! Hope this helps!

PR #404, opened 2017-05-14
Fix typo in GettingStarted/testing.md

rstormsf:
in the docs it states that `tests` dir should be used while the correct one would be `test`


tcoulter:
These docs in this repository are old, unfortunately, so I can't merge this one, but I've made the same changes here:

http://truffleframework.com/docs/getting_started/testing
https://github.com/trufflesuite/trufflesuite.github.io/blob/master/_site/public/docs/getting_started/testing.md#location

Thanks!

rstormsf:
Np!

PR #408, opened 2017-05-17
Make filter for js files sharp

illuzen:
Generally good practice to make switch structures sharp, i.e. don't assume anything not a .sol file is a .js file. Probably not a security risk here, but this sort of error is a common exploit vector.


gnidan:
Hi, thank you for your contribution. Closing this because the monorepo switch now creates merge conflicts everywhere. If you are still interested, please re-open!

Issue #411, opened 2017-05-23
Win 10 Install error


rfikki:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

What the issue is, in broad strokes.
https://gist.github.com/rfikki/5554651b135341ac044293556ddba92c
## Steps to Reproduce
Install latest version
Please provide the shortest amount of steps to reproduce your issue.

## Expected Behavior
clean error free install
What you expected to happen.

## Actual Results
https://gist.github.com/rfikki/5554651b135341ac044293556ddba92c
What actually happened. Please give examples and support it with screenshots, copied output or error messages.

## Environment

* Operating System:  Win 10
* Truffle version: 3.2.4
* Ethereum client: geth
* node version: node v7.4.0
* npm version: npm  v4.2.0

ripper234:
+1 I'm having the same issue

bvrooman:
For one, it looks like the dependency, fs-ext, is currently failing on Windows builds: https://ci.appveyor.com/project/baudehlo/node-fs-ext/branch/master. It seems that it is currently impossible to use Truffle on Windows as long as this dependency is failing.

ripper234:
Seems to be relatively easy to resolve then.

1. fs-ext seems to be used mostly for ```copySync``` and ```readFileSync```. I opened [an issue](https://github.com/baudehlo/node-fs-ext/issues/69) on fs-ext repo, not sure how easy it is to resolve, but at least as a workaround I believe (correct me if wrong) that fs-ext functionality can be manually implemented instead of adding it as a dependency.

2. execSync module seems to be unneeded, as
> both iojs and node.js v0.12 have execSync function

bvrooman:
I agree with the solutions you propose - the dependencies could be replaced with the alternatives you suggest. The problem, unfortunately, seems that they are not dependencies of Truffle itself, but dependencies of dependencies. I will put together a list of Truffle dependencies that use these problematic libraries. However, short of replacing these dependencies, I don't have a good action plan to propose at the moment.

bvrooman:
Here's the result of a quick analysis on Truffle dependencies: 

**Dependent on fs-ext**
lock-me (dependencies and optionalDependencies)

**Dependent on execSync**
codecov@1.0.1

**Dependent on codecov@1.0.1**
request
ignore
shelljs
node-fetch
regenerate
babylon
aegir
interface-datastore

[...]

PR #419, opened 2017-05-28
Add workaround for #411

ripper234:
See #411


tcoulter:
Hi there.

Thanks for this! We now install just fine on Windows with our latest version. 

Please open a new ticket if you're still having installation issues. Thanks again!

PR #425, opened 2017-06-01
Run test command if --tests flag is specified on watch

p14n:
In issue 269 people were expecting --tests to run tests after a rebuild.  I've added this.  Note it diverts to the test task rather than rebuilding then passing to tests (where we would have to tell it not to rebuild), so that's up for discussion.

### DIFF
--- a/lib/commands/watch.js
+++ b/lib/commands/watch.js
@@ -1,3 +1,5 @@
+var testCommand = require('./test');
+
 var command = {
   command: 'watch',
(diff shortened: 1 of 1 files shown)


tcoulter:
Hi there. 

The `watch` command will eventually be removed (that's the current trajectory; this may change based on user feedback). 

We don't plan to add new features to it at this time. Please let us know if this is an issue. 

Thanks,
Tim

PR #431, opened 2017-06-05
Noob error I ran into

captDaylight:
Might save others the 20 minutes I spent trying to figure out why I was getting an error :)


tcoulter:
Thanks!

PR #547, opened 2017-08-24
Update on Assert.sol to avoid warnings with "truffle test" 

stephgou:
Using "A" as a name for a constant in Assert.sol triggered a warning "Warning: This declaration shadows an existing declaration" on each function that was using "A" as parameter during the execution of the "truffle test" command (and there were quite many).
Note : These warnings do not appear on oldest version of node and npm. For example, no warning with node version v6.11.2 and npm v3.10.10 but a warning with npm 5.3.0 and node 8.4.0. The easiest way to remove the warning is to rename this "A" constant.

### DIFF
--- a/build/Assert.sol
+++ b/build/Assert.sol
@@ -61,7 +61,7 @@ library Assert {
     string constant STRING_NULL = "";
 
     uint8 constant ZERO = uint8(byte('0'));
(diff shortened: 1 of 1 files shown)


tcoulter:
Hi there. This has been fixed in Truffle `v4.0.0-beta.0`. Thanks!

PR #563, opened 2017-09-08
Upgrade solc to 0.4.16 and node to 8

antyale:
- Upgrade solidity compiler to the latest version
- Upgrade node to v8 in travis CI. Node 8 is the next LTS version in October 2017. ([Node LTS schedule](https://github.com/nodejs/LTS))
- Increase test timeouts to avoid issues with slower executions in Travis CI

I can create different PRs if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 node_js:
   - "6"
-  - "7"
+  - "8"
 
 addons:
   apt:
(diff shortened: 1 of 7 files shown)


tcoulter:
Hi there. Thanks for the PR. `solc` will be upgraded soon, but we're going to do that as its own task.

I'm going to hold off pulling the other changes as this directly effects our CI (and we're in the middle of a complete CI overhaul at the moment). We hope to have official support for Node 8 soon. If it helps, I'm using Truffle + Node 8 now. 

Cheers!

Issue #448, opened 2017-06-14
win10


acodercat:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

What the issue is, in broad strokes.

## Steps to Reproduce

Please provide the shortest amount of steps to reproduce your issue.

## Expected Behavior

What you expected to happen.

## Actual Results

What actually happened. Please give examples and support it with screenshots, copied output or error messages.

## Environment

* Operating System: 
* Truffle version:
* Ethereum client:
* node version:
* npm version:

brakmic:
Had the same problem with `truffle@beta`. After having installed **truffle@3.2.1** the problem disappeared. However, for a successful install of truffle@3.2.1 you must have a VS2015 build chain. I know it doesn't sound like a real solution but according to this [thread](https://github.com/trufflesuite/truffle/issues/411#issuecomment-304524066) it's rather hard to run truffle properly under Windows.

Enigmatic331:
On a Windows machine. Can confirm that I am also having the problem with the latest truffle (release v3.3.0) when running truffle serve. Installed truffle@3.2.1 which resolves this issue (thanks brakmic).

I'm a bit tempted to buy a MacBook just to develop on Truffle....

SeunLanLege:
Just had this Same issue resolved with installing `truffle@3.2.1`

it's not a windows10 issue, as i was running ubuntu 16.04

@tcoulter

haraldini:
Same issue on MacOS

EdgarCloggs:
Same issue on Ubuntu 16.04 - `truffle@3.30`, workaround I found to use: 

`npm run dev to test`

tcoulter:
Confirmed issue with Truffle 3.3.0. We're on it.

liskicious:
@EdgarCloggs 

I don't understand the npm run dev workaround?

After running:
truffle init
truffle compile
truffle migrate

there is no package.json there. So no script to run?
Would you know where I can get the package.json for the truffled project?

kevupton:
can confirm happening `windows 10`, latest env.

d10r:
Is there any known workaround for the time being? Ubuntu 14.04 here.

yozef:
`npm run dev` works with truffle 3.3.0 on Mac

yellowred:
same here
`
$ truffle version
Truffle v3.4.5 (core: 3.4.5)
`
[...]

PR #579, opened 2017-09-18
use a fake fsevents that is compatible with calling code

pospi:
This resolves the crash in #448 but file watching still does not seem to work.. might be a good place to start but expect there are further issues.


gnidan:
Hi @pospi, I'm closing this because it won't make it onto `develop` cleanly. I'm not sure if this is still a problem or not. Please feel free to re-open. Thank you!

PR #575, opened 2017-09-15
Enable development with `meta`

gnidan:
Spun out from #528

PR #589, opened 2017-09-26
Enable development with `meta` (master)

gnidan:
`master`-branch version of #575


gnidan:
Ah going to have to use a different source branch

PR #591, opened 2017-09-27
Bundle the chain executable and ensure the cli refers to it properly

tcoulter:
* Introduces `BUNDLE_CHAIN_FILENAME`, which injects the filename of the bundled chain file into the bundled version of the cli.
* Uses Webpack's banner plugin for re-adding the shebangs. Note that we don't call `chmod 755 ...` anymore, but it looks like git/webpack will keep the permissions that existed on the original files


gnidan:
Hey @tcoulter I noticed this is failing because it needs a `truffle-core` release to beta. I merged your PR there, so it should be good to go.

gnidan:
(Actually you're waiting on a new release of truffle-resolver 4.x.x, will ping when good)

gnidan:
@tcoulter I have released `truffle-resolver` v4.0.1, containing the changes that made it into v3.0.1. To follow our discussed process, this should now pass CI once `truffle-core` v4.0.0-beta.1 is released

PR #599, opened 2017-10-03
Use exact version of mocha

Marak:
I've noticed some of the dependencies of `mocha` have been adding new contributors with publish rights.

For security purposes, it would probably be best to ensure `truffle` is using a semver of `mocha` which is fixed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "version": "3.4.11",
   "description": "Truffle - Simple development framework for Ethereum",
   "dependencies": {
-    "mocha": "^3.4.2",
+    "mocha": "3.4.2",
     "original-require": "^1.0.1",
     "solc": "0.4.15"
   },


gnidan:
Closing this as it's now outdated since the switch to monorepo. Will revisit once the dust settles with Lerna. Thank you!

Marak:
Alright. I just checked again, It still looks like there are a lot of dependencies in `truffle` which are using semver ranges and not exact versions.

If you start digging down in the dependency tree you'll find a few heavily depended on libraries that have given npm publish rights to new or anonymous contributors with little to no screening process. A lot of prolific Node package developers have moved onto other interests.

All it would take is one rogue publish on an innocuously package like `commander` and anyone installing `truffle` could have their local machine compromised ( crypto stolen ). Using an exact semver would guarantee a known previous release which is presumable secure.

I'd hope that most developers are not storing private keys on any machines that run `npm install`, but I also know that some are not. This could be an accident waiting to happen.

Good luck!

gnidan:
@Marak You bring up a good point. With the monorepo update, we're now committing `yarn.lock` so versions should stay consistent. I don't think semver now poses the same threat, since versions will change less frequently.

Marak:
Very good. I hadn't notice the `yarn.lock` file, I only noticed the packages like https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-box/package.json

I'll have to take a closer look at `lerna` as I've not tried using it yet.

Thanks for the additional information.

PR #658, opened 2017-11-02
update build artefacts  attributes (chmod a+x)

robert-zaremba:
The build artefacts  don't have executable attribute. This is useful if somebody is using a version directly from github.


robert-zaremba:
The travis fail doesn't seems to be related to this change set

gnidan:
Closing this since it got lost and then outdated. Thanks, and apologies!

PR #754, opened 2018-01-17
Upgrade solc to v0.4.19

vinay035:


celeduc:
I have submitted PR #784 which requires 0.4.18 as the minimum version, but allows the use of versions less than 0.5.x.

gnidan:
Thanks for this, merging!

This will be released in v4.0.6

PR #769, opened 2018-01-24
Added `contains` and `doesNotContain` assertions for list types

sabhiram:
1.  The int[], uint[], bytes32[] and address[] types now have
    `contains` and `doesNotContain` helpers.
2.  A lot of code in there is boilerplate - should we autogenerate this?
3.  The above would make allowances to have helpers for more specific
    fixed-size byte arrays like bytesN[].

### DIFF
--- a/build/Assert.sol
+++ b/build/Assert.sol
@@ -911,6 +911,58 @@ library Assert {
             _report(result, _appendTagged(_tag(arrLength, "Tested"), _tag(length, "Against"), message));
     }
 
+    /*
+        Function: contains(uint[])
+
+        Assert that the specified value exists in the 'uint[ ]'.
+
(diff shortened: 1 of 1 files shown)


gnidan:
Hi @sabhiram, thanks for this!

Mind opening these changes against `truffle-core`? I'm afraid you found built artifacts - the file you're looking for is in https://github.com/trufflesuite/truffle-core/blob/develop/lib/testing/Assert.sol

sabhiram:
Bah - I knew I goofed something up when the other artifact file was changing. My bad. I will close these are re-submit :)

sabhiram:
Thanks for the pointers - submitted the feature as a PR to truffle-core.

PR #789, opened 2018-02-02
Update for web3 1.0

cgewecke:
+ Modifies tests so they pull a web3 1.0 compatible version of the MetaCoin project and translate `uint` values to `number` with `parseInt` instead of `.toNumber`. 

+ Makes Web3 1.0 web-packable - it has a dependency which conditionally requires `electron` without using it and webpack complains about not being able to find it. 
  + Upstream [issue](https://github.com/sindresorhus/got/issues/345) suggesting the workaround in this PR
  + The [error](https://travis-ci.org/trufflesuite/truffle-contract/jobs/336367881#L2347) in CI

(Code has been added to pull in a branch of `truffle-init-default` so that everything runs clean in CI. Will need to be removed after that branch is merged, before this branch is merged)


cgewecke:
@Asone Yes, definitely agree we should try to make `truffle-contract` compatible with the new web3 wallet. Earlier this year wallet accounts only worked with `sendTransaction` and method calls needed to be encoded as data / transmitted that way. Will look into this more today and see if that's still the case.

Also if you're testing `web3-one` and see other things that aren't working or could be improved, please feel free to note them on the `truffle-contract` [PR](https://github.com/trufflesuite/truffle-contract/pull/95).  Thanks so much for your encouragement and help :)

Issue #780, opened 2018-01-29
truffle version should give adequate information for a bug report


bmaurer:
## Issue

If you file a bug report `truffle version` is not sufficient to meet the requirements of the github template

## Steps to Reproduce

File a bug on github. See the following request in the template:

```
## Environment

* Operating System: 
* Truffle version:
* Ethereum client:
* node version:
* npm version: 
```

Run `truffle version`

## Expected Behavior
The requested information is output

## Actual Results

```
Truffle v4.0.5 (core: 4.0.5)
Solidity v0.4.18 (solc-js)
```
## Environment

lol

* Operating System: MacOS
* Truffle version: Truffle v4.0.5 (core: 4.0.5)
* Ethereum client: Ganache
* node version: v8.9.4
* npm version: 5.6.0

cgewecke:
@bmaurer It looks like you're suggesting that Truffle should output all of these fields. However it's normal for a program's `--version` flag to report itself only. For example:
```
docker --version
Docker version 1.12.3, build 6b644ec
```

The bug report asks for additional information that can be retrieved by running:
```
npm --version
node --version
```

Ethereum client and OS version commands will vary depending on their identity.

Do you think it would be helpful to update the issue template with some guidance about how to generate the required fields?

bmaurer:
Yeah even just adding some guidance would be helpful though to the extent possible it’d be great to have a copy & paste line one can put in the shell.

cgewecke:
@bmaurer There are various clients and OS's and no single way of getting their versions, so it seems  like a shell line would be tricky, but definitely agree that greater clarity about how to provide that info would be helpful to everyone. Thanks for suggesting this.

lazaridiscom:
@cgewecke , @bmaurer 's idea is excellent.

You are right the usual behaviour for tools is to provide only the version.

Yet truffle gives an additional version info. 

Just implement a new command like `truffle all-versions` (or similar, lookup the naming of similar functionality within standard tools), which then does the hard work of finding out which exact versions the calling truffle instance works with.

Instructions would then be super-simple: "Please call `truffle all-versions` within your project folder and copy the output into the issue you file"

cgewecke:
Sorry for closing this - preparing for a release tomorrow and modifications to the issues template will be included.

PR #806, opened 2018-02-19
Make it easier to get env info when filing issues

wbt:
Partially addresses Issue #780.  I considered adding a line for the Solidity version since that's also output in `truffle version` but thought it might've been excluded intentionally.  I also swapped two lines' order to keep those with tips together for better visual harmony.


cgewecke:
@wbt Nicely done! Thank you. This change should start showing up in the next patch release.

PR #813, opened 2018-02-23
Rename testrpc to ganache-cli

robinboening:
Update the Readme since `testrpc` has been renamed to `ganache-cli`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,10 @@ $ truffle init
 
 From there, you can run `truffle compile`, `truffle migrate` and `truffle test` to compile your contracts, deploy those contracts to the network, and run their associated unit tests.
 
-Be sure you're connected to an ethereum client before running these commands. If you're new, install [ganache-cli](https://github.com/rufflesuite/ganache-cli) to run a local blockchain RPC server. After that, simply run `ganache-cli` in a new tab.
+Truffle comes bundled with a local development blockchain server that launches automatically when you invoke the commands above. If you're interested in configuring the blockchain server or would like a user-friendly UI for it, check out:
+
++  [ganache-cli](https://github.com/trufflesuite/ganache-cli): a command-line version of Truffle's blockchain server.
++  [ganache](http://truffleframework.com/ganache/): A GUI for the server that displays your transaction history and chain state.
 
 See [the documentation](http://truffleframework.com/docs/) for more details.
 
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ $ truffle init
 
 From there, you can run `truffle compile`, `truffle migrate` and `truffle test` to compile your contracts, deploy those contracts to the network, and run their associated unit tests.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -34,7 +34,7 @@ $ truffle init
 
 From there, you can run `truffle compile`, `truffle migrate` and `truffle test` to compile your contracts, deploy those contracts to the network, and run their associated unit tests.
 
-Be sure you're connected to an ethereum client before running these commands. If you're new, install [testrpc](https://github.com/ethereumjs/testrpc) to run a local blockchain RPC server. After that, simply run `testrpc` in a new tab.
```
cgewecke: @robinboening Thanks! This needs to be done. One issue is that since this was written, Truffle has started shipping with a blockchain server that gets launched automatically in the background if you don't specify any configuration networks. What do you think about changing this to something like:

> `Truffle` comes bundled with a local development blockchain server that launches automatically when you invoke these commands. If you'd like to install the development client separately and do [cool network configuration stuff](http://truffleframework.com/docs/advanced/configuration) with it, you can `npm install ganache-cli`. And if you'd like your client to have a beautiful graphical user interface that displays your transaction history and chain state, checkout [ganache](http://truffleframework.com/ganache/).

(Not dictating here - would love any improvement to the above.)

robinboening: @cgewecke Thanks for the fast reply!

I am not 100% sure if I understand correctly, but let me try to explain what I understand when reading your proposal, but first of all: I am new to truffle (I haven't even used more than a simple `truffle init` by now).

You are referring to "these commands", so I assume you are talking about `truffle compile/migrate/test` which are booting up the built in dev server when invoking them. If this is correct I don't understand why I want or need the server for the compilation or to run the tests?

Thanks!

cgewecke: @robinboening Excellent points. Agree that's confusing and actually you're the perfect person to decide how this should be worded since you come with no pre-conceptions about what anything means :) 

I think you're right: `these commands`  is clearer as `the commands mentioned above`.

In answer to the question, "why do I need the server?": You don't need it to start off doing basic testing. But as your project complexity grows, it's often helpful to run the client separately because it's far more configurable that way. How replacing the section that starts 'If you'd like . . . .' with something simpler and vaguer like?

Also check out:
+  [ganache-cli](https://github.com/trufflesuite/ganache-cli): a configurable command-line version of Truffle's blockchain server.
+  [ganache](http://truffleframework.com/ganache/): A GUI for the server that displays your transaction history and chain state.

The actual links for those are:
+ https://github.com/trufflesuite/ganache-cli
+ http://truffleframework.com/ganache/

robinboening: @cgewecke thanks a lot for the explanation! Its much clearer to me now. I don't actually know what enlightened me now, maybe it just took me some time 🙄

I do like your proposal and I think it explains it quite well. I have only a few more changes to propose. I will open a new PR for it since its merged already :)

Thanks again for the quick responses and the patience with a newbie 👍

robinboening:
@cgewecke I jsut submitted a new PR here: https://github.com/trufflesuite/truffle/pull/832

PR #832, opened 2018-03-04
Add some more details about ganache

robinboening:


cgewecke:
@robinboening 

Thanks for revisiting this and adding some clarity. Have edited slightly but preserved your idea of suggesting the separate client and explaining why, which definitely makes things more intelligible.  Thanks again :)

PR #854, opened 2018-03-19
Update package.json to use latest solc version 0.4.21

patidarmanoj10:
I think truffle  should use latest solc 0.4.21

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "mocha": "^3.4.2",
     "original-require": "^1.0.1",
-    "solc": "0.4.19"
+    "solc": "0.4.21"
   },
   "devDependencies": {
     "clean-webpack-plugin": "^0.1.16",


ZitRos:
Hello!

We highly need this PR accepted! In particular, the new compiler supports some syntax (like `emit EventName(...)`) and optimizations which we need to use as soon as possible.

Also, what do you think about making solc an external (replaceable) dependency somehow, to not to update it each time a new version is released? Some kind of configurable optional value in `truffle.js` for example.

Thank you!

cgewecke:
@patidarmanoj10 @ZitRos Thanks, we'll publish this change this week. 

@ZitRos Making the compiler version configurable is on our near-term roadmap. This should be possible soon.

ZitRos:
> This should be possible soon.

Great, thank you @cgewecke!

urokuta:
@cgewecke 
thank you for your work.
when will it be released on v4.1.4 ?
Our team also needs to use as soon as possible.

cgewecke:
@urokuta Apologies we have a few things in flight right before the release. It will be this week. 

Because a formal release takes a while to execute, we're starting an experiment with publishing a  'nightly' build. It's called `darq-truffle` and is `truffle` built at the current state of the develop branch, We may add other branches as needed. 

If you'd like to test it out (it contains solc 0.4.21) and is available by running:
```shell
npm install -g darq-truffle

# NB: It's not called truffle! 
darq-truffle compile    
```

PR #856, opened 2018-03-21
Add geth client in CI

cgewecke:
Adds `geth:latest` to the list of clients we scenario test against. It's set up as an 'allowed failure' because . . . that's our current position? As such this test is more a notice of breaking changes upstream rather than a verification that we haven't broken anything.  

Another (possibly more rational) approach might be to test against `geth:stable` and require that we pass. We should discuss this. 

Also: have looted @cpurta's excellent [geth-dev-net](https://github.com/cpurta/geth-devnet) for logic and technique, especially for generating geth accounts on launch. Reaching out to him on gitter to ask if this is ok and make sure we are crediting him appropriately in the file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,6 @@ before_install:
 env:
   global:
     - TEST=repo
-    - GETH_CONTAINER=geth
   matrix:
     - GANACHE=true
     - GETH=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,17 +7,19 @@ services:
 
 before_install:
   - docker pull truffle/ci
+  - docker pull ethereum/client-go:latest
 
 env:
-  - TEST=repo
+  global:
+    - TEST=repo
+    - GETH_CONTAINER=geth
+  matrix:
+    - GANACHE=true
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -7,17 +7,19 @@ services:
 
 before_install:
   - docker pull truffle/ci
+  - docker pull ethereum/client-go:latest
 
 env:
-  - TEST=repo
+  global:
+    - TEST=repo
+    - GETH_CONTAINER=geth
```
cgewecke: This is extraneous.

cgewecke:
@gnidan Ummm. No it isn't. 

For further reference: @gnidan noted at dev sync that these geth tests will really only run  when `truffle` main repo is pulled against (typically before a release). To have them run for every module PR, we'd probably need to re-design `truffle-ci` to use `docker compose` and split out the pieces like geth, node, etc into logical units. 

See @gnidan's [docker-eth-dev](https://github.com/gnidan/docker-eth-dev/blob/master/containers/docker-compose.yml) for more.

PR #904, opened 2018-04-19
Upgrade solc to 0.4.23

cgewecke:


cgewecke:
Just a note - this build validated in #905.

PR #969, opened 2018-05-29
duplicate methods equal/notEqual for uint8-uint248

wjmelements:
Currently, if I try to compare two `uint248` with truffle's Assert.equal, I get a syntax error. To avoid casting, I added support for `uint8` through `uint248` to the `equal` and `notEqual` methods in Assert.sol.


ebaizel:
looks good, but could you add a test for one of the cases `equal` and `notEqual`?

wjmelements:
I need to make this change in truffle-core instead.

wjmelements:
https://github.com/trufflesuite/truffle-core/pull/144

wjmelements:
Looks like the truffle-core repo was merged into this one. Reopening.

PR #1014, opened 2018-06-14
Remove Truffle bundle, etc. from git

gnidan:
Also, define `npm prepare` step to do build


cgewecke:
@gnidan Does this also need an empty `.npmignore` added to `packages/truffle`? 

There was a weird issue with the debugger once where npm defaulted to the gitignore and wouldn't include the `dist`.

gnidan:
Seems like it couldn't hurt?

gnidan:
screen shot 2018-06-14 at 4 57 47 pm

Looks fine

PR #1020, opened 2018-06-15
Fix npm test for Node 10

cgewecke:
Another Node 10 crash - something is different about `fs` and it's not happy with the old callbacks. Example error:
```shell
1) NPM integration
       "before all" hook: Create a sandbox:
     TypeError [ERR_INVALID_CALLBACK]: Callback must be a function
      at maybeCallback (fs.js:159:9)
      at Object.fs.writeFile (fs.js:1278:14)
      at test/npm.js:31:10
      at /Users/cgewecke/code/consensys/truffle/packages/truffle-box/box.js:47:11

```

### DIFF
--- a/packages/truffle-core/test/npm.js
+++ b/packages/truffle-core/test/npm.js
@@ -28,7 +28,8 @@ describe('NPM integration', function() {
       };
       config.network = "development";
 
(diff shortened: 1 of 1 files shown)


cgewecke:
I know. This bodes ill.

PR #1025, opened 2018-06-16
Add lerna publishConfigs

cgewecke:
Adds to all package.jsons within `packages`
```javascript
"publishConfig": {
    "access": "public"
  }
```
This  is what lerna uses to identify publication candidates. More in their docs [here](https://github.com/lerna/lerna#publish).

Do we want to publish all of this?


cgewecke:
@gnidan Well I *think* we could set some of the keys to private if we don't intend to publish and they'll be ignored by the lerna publishing wizard. Just an idle thought.

Issue #713, opened 2017-12-12
Deployed Contracts have no Network when using async/await


jleeh:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

When my contracts are deployed using the `truffle migrate --reset` command, none of them have a network in the contract JSON even though the contracts are deployed. This causes the unit tests to fail saying the contract hasn't been deployed to the network.

This only happens with the contracts I'm working on as it works fine with different solidity projects.

I've tested the contracts in question on Remix and they deploy fine.

## Steps to Reproduce

- `truffle migrate --reset`
- `truffle test`

## Expected Behavior

- Contract has network assigned in JSON
- Can test against the contract

## Actual Results

- Contract is deployed after a successful migration:
```
Running migration: 1_initial_migration.js
  Replacing Migrations...
  ... 0xf893631620cee218dfdac5a5b118e950b39c411add85ccf5ce0891abc594b451
  Migrations: 0x01afadc80ac64129f3100633b556f1a40a15a9c0
Saving successful migration to network...
  ... 0x0af95e5b3d45c4b37b56287395f0815b4e315d5b3a5bdc0bdcdc5134e444aeb4
Saving artifacts...
Running migration: 2_deploy_linkpool.js
  Replacing Arithmetic...
  ... 0x5667313901deae25a423b2f5c9bc5554ef691491bf1d8d7cd8e71d40948f416d
  Arithmetic: 0x9a3be86baaf7b84f60860b7be146963f5bc8a21b
Saving successful migration to network...
  Deploying PoolStorageLocator...
  ... 0x6ed6122d0851c169df7d156341845ca884e232d455b91825a7e495196d54d6ae
Saving artifacts...
  ... 0x19ca0c89a2c079c8f7dd1c44db88e3520c4ac3781f907db58480e364a31d812f
  PoolStorageLocator: 0xaed391873b70b273393ffc8cce46cfa34aeedc03
  Deploying Node...
  ... 0xbc09e15f761c95071533db92a3900bcdfdac68762bfde30abfa23e89a6434748
  Node: 0xd17e00d08da375b61a10d36a25fb20fde1c8bc29
  Deploying LinkPool...
  ... 0xb6f889a1d8a8b9bd04fc8ecbdf58e58dac33dc014e940c266e15d6e2b181d050
  LinkPool: 0x02961e046dfa5ed124013dac3893cc88af45d371
[...]

PR #1028, opened 2018-06-18
New Migrations

cgewecke:
Opening to keep track of what still needs to be done this week.
- [ ] Color
- [x] Remove support for batch deployments 
- [ ] Fix E2E CI at `wild-truffle` for monorepo and check colonyNetwork / aragon migrations
- [ ] Should we have a classic migrations output option? To minimize noise in CI?
- [ ] Publish hd-wallet-provider V5 (fork provider-engine to Truffle / merge cgewecke's in)

**Gists of the output** (So we can start to make this look better)

+ [simple migration](https://gist.github.com/cgewecke/129b77c9eb51b1140675631964ec2aea)
+ [production migration w/ dry-run](https://gist.github.com/cgewecke/500e0470d0f9515a3fc57630f4b2da29) (against `geth --dev`)
+ [error case: revert](https://gist.github.com/cgewecke/d2325422bc51f2c05965d6e4f6d06127)
+ [error case: oog](https://gist.github.com/cgewecke/472e095484dded1c0b0d415d6537d5f5)

**E2E Testing vs. Rinkeby / Ropsten / MintTestNet (Settlemint/Parity POA network)**

Everything is set up [here](https://github.com/cgewecke/rinkeby-migrations) to deploy Consensys/PLCRToken to these testnets. Just clone repo (recursively), set up a `secrets.js` with a funded mnenomic, and follow README instructions to install and run at this branch's current commit. 

**Known Problems**
+ ~~geth ws [is broken](https://github.com/ethereum/go-ethereum/issues/16846) for large data sends & web3 1.0 has broken confirmations polling over rpc.~~ (Fix w/ block-polling)
+ HDWallet uses web3-provider-engine which isn't web3 1.0 compatible because the latter stopped supporting the `sendAsync` method, renaming it `send`.  `provider-engine` hates `send` though because it used to be sync. Believe this will require we temporarily fork provider-engine.  
+ ~~Account are [not really forking](https://github.com/trufflesuite/ganache-core/issues/128).~~ (Fixed at ganache, awaiting publication) 

**Goals**
+ Add substantive tests / fixtures for this feature so it can be maintained & debugged.
+ Make async/await behave intuitively without breaking the current 'then-able' pattern.
+ Meaningful error handling. Should detect constructor reverts/OOG/nonce etc and offer advice.
+ Expose revert reason strings (ganache only atm)
+ Configurable confirmation delays (by network)
+ Configurable block timeouts (by network)
+ Display more data, more clearly
+ `--dry-run`
   + Auto-runs when it detects that a migration targets a public network
   + Interactive prompt before running real migration
+ Decouple logging and migrations logic with an async Reporter

**Example command**
```shell
$ truffle migrate --network rinkeby --interactive
```

**Example config**
```javascript
networks: {
  rinkeby: {
    provider: () => new HDWalletProvider(mnemonic, `https://rinkeby.infura.io/${infura}`),
    network_id: 4,
    gas: 6000000,
    gasPrice: 20000000000,
    confirmations: 2,     // # confirmations to wait between each deployment
    timeoutBlocks: 75   // # of blocks to wait for network to mine a deployment before erroring
  },
}
```

**GIF**

migrations_a

Visual Review (future improvements):
+ @gnidan thinks migrations have too much data.
+ dry-run should be clearly distinguished
+ 4 second increments are weird (we should lie)
+ more interactivity 
+ Make sure block limits are same dry-run and real.
+ Move error handling to truffle-contract / truffle-error
+ `--only` flag
+ Spinner for saving migrations


packages/truffle-core/cli.js
```diff
@@ -47,7 +47,9 @@ command.run(process.argv.slice(2), options, function(err) {
     if (typeof handle.close === 'function'){
       handle.close();
     } else if (handle.readable && !handle._isStdio){
-      handle.destroy();
+      //This is causing problems over RPC
+      //with hd-wallet-provider V5
+      //handle.destroy();
```
cgewecke: NB: this was put in because there was crashing when `config.websockets === true` in `darq-truffle`. Needs further investigation.

packages/truffle-deployer/test/helpers/reporter.js
```diff
@@ -0,0 +1,292 @@
...
+
+      default:
+        message = this.messages('default', payload.contract, error.message);
+        this.deployer.logger.error(message);
+    }
+  }
+}
+
+module.exports = Reporter;
+
```
cgewecke: I think we should pull this out, move the migrations reporter to a new modules (perhaps `truffle-reporters`) and have these test consume the production reporter.

package.json
```diff
@@ -14,5 +15,6 @@
     "packages": [
       "packages/*"
     ]
-  }
+  },
+  "dependencies": {}
 }
```
cgewecke: ?

packages/truffle-contract/lib/execute.js
```diff
@@ -65,6 +87,30 @@ var execute = {
...
+    if (isObject){
+      const data = res.error.data;
+      const hash = Object.keys(data)[0];
+
+      if (data[hash].return && data[hash].return.includes('0x08c379a0')){
+        return web3.eth.abi.decodeParameter('string', data[hash].return.slice(10))
+      }
+    }
+
+    if (isString && res.result.includes('0x08c379a0')){
```
gnidan: else

packages/truffle-contract/lib/execute.js
```diff
@@ -145,10 +192,21 @@ var execute = {
         execute
           .getGasEstimate
           .call(constructor, params, network.blockLimit)
-          .then(gas => {
-            params.gas = gas
+          .then(result => {
+            (result.error)
+              ? context.reason = result.error.reason
+              : context.reason = null;
+
```
cgewecke: Is this getting attached at the receipt handler? Check

packages/truffle-core/lib/commands/migrate.js
```diff
@@ -85,26 +110,72 @@ var command = {
...
         if (dryRun) {
-          networkMessage += " (dry run)";
+
+          try {
+            await setupDryRunEnvironmentThenRunMigrations(conf);
+            done();
+          } catch(err){
+            done(err)
+          };
+          return;
```
cgewecke: Bad.

packages/truffle-deployer/src/deployment.js
```diff
@@ -0,0 +1,404 @@
...
+  _errors(name){
+    return `Migrations failure`
+  }
+
+  /**
+   * Helper to parse a deploy statement's overwrite option
+   * @param  {Arry}    args        arguments passed to deploy
+   * @param  {Boolean} isDeployed  is contract deployed?
+   * @return {Boolean}             true if overwrite is ok
+   */
```
cgewecke: add private doc string.

packages/truffle-deployer/src/deployment.js
```diff
@@ -0,0 +1,404 @@
...
+  _close(){
+    this.emitter.clearListeners();
+    this.promiEventEmitters.forEach(item => {
+      item.removeAllListeners();
+    });
+  }
+};
+
+module.exports = Deployment;
```
gnidan: No new line at end of file

For those of us old-heads that like to `cat` their files on the terminal

gnidan: But really I came here to say that `_methods` should **not** be called outside of this class

packages/truffle-deployer/src/deployment.js
```diff
@@ -0,0 +1,404 @@
...
+      // Wait for `n` blocks
+      if(self.confirmations !== 0 && shouldDeploy){
+        await self._waitBlocks(self.confirmations, state, contract.web3);
+      }
+      // Finish: Ensure the address and tx-hash are set on the contract.
+      contract.address = instance.address;
+      contract.transactionHash = instance.transactionHash;
+      return instance;
+    };
+  }
```
gnidan: This method is a *bit* long? Maybe worth chopping it into pieces? Up to you

packages/truffle-deployer/src/deployment.js
```diff
@@ -0,0 +1,404 @@
...
+        }
+
+        return self._deploy(contract, params)();
+      });
+
+      await self.emitter.emit('preDeployMany', arr);
+      await Promise.all(deployments);
+      await self.emitter.emit('postDeployMany', arr);
+    };
+  }
```
cgewecke: Delete this.

packages/truffle-migrate/migration.js
```diff
@@ -0,0 +1,188 @@
...
+      // deploy to use await
+      if (migrateFn && migrateFn.then !== undefined){
+        await deployer.then(() => migrateFn);
+      }
+
+      // Migrate without saving
+      if (options.save === false) return;
+
+      // Write migrations record to chain
+      const Migrations = resolver.require("./Migrations.sol");
```
gnidan: `resolver.require("Migrations")`

cgewecke: 👍

packages/truffle-contract/lib/execute.js
cgewecke: Yes it is - we also have tests for reason string handling across a matrix of the following:
+ `deploy` and `send`  
+ both states of vmErrorsOnResponse
+ non-existence of reason / geth 
+ ganache as provider and server (the former at truffle-contract / the latter at scenario)

packages/truffle-contract/lib/execute.js
cgewecke: Fixed.

packages/truffle-deployer/src/deployment.js
cgewecke: Fixed

packages/truffle-migrate/migration.js
cgewecke: Hmmm. Need to investigate this a bit. Is the hard-coding of the path (carried over from original) the cause of the bug identified [here](https://github.com/trufflesuite/truffle/issues/713#issuecomment-403031560). Or.....?

packages/truffle-contract/lib/handlers.js
```diff
@@ -104,6 +104,7 @@ var handlers = {
     // .method(): resolve/reject receipt in handler
     if (parseInt(receipt.status) == 0 && !context.onlyEmitReceipt){
       var error = new StatusError(context.params, receipt.transactionHash, receipt);
+      error.reason = context.reason;
```
gnidan: I'm a little bit worried about attaching `reason` to these error objects like this. It feels like an anti-pattern, but maybe I'm just afraid of statefulness. Let's keep an eye on this in the beta, in case it causes unforeseen complications.

Pardon me if I haven't seen it yet, but it might be worth making a special note somewhere in a comment, so we can encourage future work to be mindful of pre- vs. post-reason-attach errors. I know you've written a good matrix of tests around this, so it's probably safe enough, esp. since it's not likely to go without `undefined` checks anyway.

packages/truffle-contract/lib/subscriptions.js
```diff
@@ -1,3 +1,8 @@
+/**
+ * This code is temporarily unused pending resolution of bugs at Web3 & Geth that make
+ * subscriptions for the block head unreliable.
+ */
+
```
gnidan: Let's remove it, then!

You can hold me accountable to find it for you in commit history later.

packages/truffle-contract/scripts/test.sh
```diff
@@ -3,7 +3,7 @@
 set -o errexit
 
 if [ "$GETH" == true ]; then
-  mocha --timeout 50000 --grep @geth --colors
+  mocha --timeout 50000 --grep @geth --colors --exit
 else
-  mocha --no-warnings --timeout 7000 --colors
+  mocha --no-warnings --timeout 7000 --colors --exit
```
gnidan: Might be worth a comment so we remember why `--exit` is included in the future

packages/truffle-core/cli.js
gnidan: Maybe I've just been overly-scolded in the past for even the tiniest amounts of commented-code, but now I feel the need to pass on this sentiment.

Idea: instead of commenting out code, we make sure to remove it via PR with a reference to the upstream issue, so that way GH make it easy for us when the issue closes, and we can add it back in with the revert button.

packages/truffle/test/scenarios/migrations/production.js
```diff
@@ -0,0 +1,78 @@
...
+      const location = path.join(config.contracts_build_directory, "Example.json");
+      const artifact = require(location);
+      const network = artifact.networks[networkId];
+
+      assert(output.includes(network.transactionHash));
+      assert(output.includes(network.address));
+      assert(output.includes('2 confirmations'));
+
+      // Geth automines too quickly for the 4 sec resolution we set
+      // to trigger the output.
```
gnidan: 👌

packages/truffle/test/scenarios/commandrunner.js
```diff
@@ -4,7 +4,15 @@ var path = require("path");
 
 module.exports = {
   run: function(command, config, callback) {
-    var execString = "node " +
+    var execString;
+
+    (process.env.NO_BUILD)
+      ? execString = "node " +
```
gnidan: Sweet

packages/truffle-deployer/src/actions/exec.js
```diff
@@ -1,27 +0,0 @@
...
-//         network: deployer.network,
-//         network_id: deployer.network_id,
-//         provider: deployer.provider
-//       }, function(err) {
-//         if (err) return reject(err);
-//         accept();
-//       });
-//     });
-//   };
-// };
```
gnidan: Yeah, good stuff!!

packages/truffle-contract/test/sources/Example.sol
```diff
@@ -88,6 +89,10 @@ contract Example {
     assert(false);
   }
 
+  function triggerRequireWithReasonError(){
+    require(false, 'reasonstring');
+  }
```
gnidan: Can we also test a reason string that exceeds 32 bytes? Pardon my having peeked behind the Solidity curtain... ([gist](https://gist.github.com/gnidan/b1890c68c8e0825d4a699929ccb2018e#string-smaller-than-wordsize), if you recall)

packages/truffle-core/lib/commands/migrate.js
```diff
@@ -32,51 +37,71 @@ var command = {
...
+              // Ensure directory cleanup.
+              temp.cleanup(function(err) {
+                (args.length && args[0] !== null)
+                  ? reject(args[0])
+                  : resolve();
+              });
+            };
+
+            copy(config.contracts_build_directory, temporaryDirectory, function(err) {
+              if (err) return reject(err);
```
gnidan: If this happens, the dir won't get cleaned up, right? Do we want that?

packages/truffle-deployer/index.js
```diff
@@ -1,84 +1,88 @@
...
-  throw new Error("deployer.exec() has been deprecated; please seen the truffle-require package for integration.")
-};
+  deploy() {
+    const args = Array.prototype.slice.call(arguments);
+    const contract = args.shift();
 
-Deployer.prototype.then = function(fn) {
-  var self = this;
+    if (Array.isArray(contract)){
+      const msg = 'Support for batch deployments has been deprecated. ' +
```
gnidan: `s/has been deprecated/is no longer supported`

packages/truffle-deployer/src/deployment.js
gnidan: Ah but we should at least move the public methods above the private ones.

packages/truffle-deployer/src/linker.js
```diff
@@ -1,78 +1,55 @@
...
-  //   unlinked_libraries.forEach(function(name) {
-  //     var library = available_contracts[name];
-  //
-  //     if (library == null) {
-  //       throw new Error("Cannot auto link " + contract.contract_name + "; " + contract.contract_name + " unknown dependency " + name + ".")
-  //     }
-  //
-  //     self.link(library, contract, logger);
-  //   });
-  // }
```
gnidan: Sorry I complained earlier

packages/truffle-deployer/test/helpers/reporter.js
gnidan: Definitely into this

packages/truffle-deployer/test/helpers/utils.js
```diff
@@ -0,0 +1,62 @@
...
+    const json = require(`../build/${name}`);
+    const contract = TruffleContract(json)
+    contract.setProvider(provider);
+    contract.setNetwork(networkId);
+    contract.defaults({ from: account });
+    return contract;
+  },
+};
+
+module.exports = utils;
```
gnidan: Also are these useful more generally, outside the deployer?

packages/truffle-migrate/migration.js
```diff
@@ -0,0 +1,190 @@
...
+const Require = require("truffle-require");
+const Emittery = require('emittery');
+const async = require("async");
+const Web3 = require("web3");
+
+const ResolverIntercept = require("./resolverintercept");
+
+const util = require('util');
+
+class Migration {
```
gnidan: Oh nice, one of these!

packages/truffle-migrate/migration.js
```diff
@@ -0,0 +1,190 @@
...
+        error: e
+      };
+
+      await self.emitter.emit('error', payload);
+      deployer.finish();
+      callback(e);
+    };
+  }
+
+  // ------------------------------------- Public -------------------------------------------------
```
gnidan: Oh wait you're doing `private` first? That's cool as long as we're consistent.

packages/truffle-migrate/reporter/indentedSpinner.js
```diff
@@ -0,0 +1,16 @@
...
+    "   ⠸",
+    "   ⠼",
+    "   ⠴",
+    "   ⠦",
+    "   ⠧",
+    "   ⠇",
+    "   ⠏"
+  ]
+}
+
```
gnidan: I feel like there's gotta be a better way to get indents under control.

packages/truffle-migrate/reporter/messages.js
```diff
@@ -0,0 +1,279 @@
...
+      listMany: () =>
+        `   * ${data.contractName}`,
+
+    }
+
+    return kinds[kind]();
+  }
+}
+
+module.exports = MigrationsMessages;
```
gnidan: I dig the messages separation of concerns

packages/truffle-migrate/reporter/reporter.js
```diff
@@ -0,0 +1,459 @@
...
+  }
+
+  async postDeployMany(){
+    this.deployingMany = false;
+  }
+
+}
+
+module.exports = Reporter;
+
```
gnidan: :+1:

packages/truffle-core/cli.js
cgewecke: Oh totally that's a good idea. . . This comment is really for me - like DO NOT forget this is a serious problem right now. Needs resolution.

packages/truffle-deployer/test/helpers/reporter.js
cgewecke: Ok cool I'll do that...

packages/truffle-contract/lib/execute.js
```diff
@@ -20,23 +21,44 @@ var execute = {
...
+      }
+
+      // This rpc call extracts the reason string
+      web3.currentProvider.send(packet, (err, response) => {
+        if (response && (response.error || response.result)) {
+          reason = execute.extractReason(response, web3)
+        }
+
+        web3
+          .eth
```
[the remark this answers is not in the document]
cgewecke: 👍

packages/truffle-contract/lib/execute.js
```diff
@@ -65,6 +87,29 @@ var execute = {
...
+
+    if (isObject) {
+      const data = res.error.data;
+      const hash = Object.keys(data)[0];
+
+      if (data[hash].return && data[hash].return.includes('0x08c379a0')){
+        return web3.eth.abi.decodeParameter('string', data[hash].return.slice(10))
+      }
+
+    } else if (isString && res.result.includes('0x08c379a0')){
```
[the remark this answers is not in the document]
cgewecke: Mmm. yes

packages/truffle-migrate/migration.js
gnidan: It's just more canonical. Artifacts are by `contract` name, not filename.

packages/truffle-contract/scripts/test.sh
cgewecke: This only affects us running in CI against Geth because we can't really shut it down from the tests. Since mocha v4 (I thought it was 5) mocha [doesn't automatically exit](https://boneskull.com/mocha-v4-nears-release/#otherbreakingchanges) if there are open connections even if it *thinks* it's done. Everything works ok with ganache because we close it in an `after`.

packages/truffle-contract/lib/handlers.js
cgewecke: I've moved this into the StatusError file where it (possibly) looks less arbitrary. Truffle-contract has actually attached quite a bit of stuff to the error traditionally - like the txHash for example (which we continue to do for backwards compatibility).

packages/truffle-contract/lib/execute.js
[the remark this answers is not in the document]
cgewecke: I guess derived from this axic [comment](https://github.com/ethereum/EIPs/issues/838#issuecomment-379238333) in the EIP ? 🙂 
> Now we could instead just assume that there is a generic error with a signature of Error(string), which hashes to 08c379a0 followed by an ABI encoded string, therefore revert("oh noes") would result in the data 08c379a0 00...20 00..07 6f68206e6f6573..00

packages/truffle-contract/lib/execute.js
```diff
@@ -20,23 +21,44 @@ var execute = {
...
+        method: "eth_call",
+        params: [params],
+        id: new Date().getTime(),
+      }
+
+      // This rpc call extracts the reason string
+      web3.currentProvider.send(packet, (err, response) => {
+        if (response && (response.error || response.result)) {
+          reason = execute.extractReason(response, web3)
+        }
```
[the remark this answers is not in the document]
cgewecke: 👍

packages/truffle-migrate/migration.js
cgewecke: The older migrate tests at `truffle-core` crash when this change is made. I want to deal with separately from this PR as a bug in the resolver - not sure what's going here....

packages/truffle-contract/test/sources/Example.sol
cgewecke: Good idea! Added a test for this - it turns out that it 'just works' but . . . this is all ganache and more or less improvised. We'll need to revisit the implementation when the clients agree how to *actually* pass this info back.

cgewecke:
Aiiii!!!! It's over.

PR #1029, opened 2018-06-18
Update Solidity sources to the latest constructor syntax (requires Solidity 0.4.22)

axic:
`SafeSend` found by @bit-shift.


cgewecke:
@axic Thanks!

Issue #1030, opened 2018-06-19
 Contract send should accept a tx_params object


sebastien-kr:
This is partially copied off of the old repo issue 101 : https://github.com/trufflesuite/truffle-contract/issues/101

## Expected behaviour

`send()` should optionally consume a `tx_params` object.

## Actual behaviour

The contract.js `send()` function doesn't accept a `tx_params` object.

This is unintuitive since almost every other function provided does accept a `tx_params` object. Due to the nature of JS, no errors or warnings are given, making this a potential source of nasty bugs.

In Web3.js the `send()` function takes a `tx_params` object which makes truffle's `send()` behavior really confusing. 

## Documentation

`send()` and `sendTransaction()` (which takes a  `tx_params` !) are not documented. 
This makes the issue really frustrating and hard to fix. Basically you have to find someone who already had the problem.

cgewecke:
@sebastien-kr Thanks for opening this. 

We're in the process of upgrading `truffle-contract` to `web3 1.0` and that work should available in the next few weeks. I believe this will no longer be an issue then - will add a regression test to our suite to make sure it behaves as expected.

gnidan:
Looks like this is done. Let us know if that's not true. Thanks!

PR #1037, opened 2018-06-20
Make send accept txParams and prefer user-defined send

cgewecke:
Addresses #1030.  `instance.send` should accept tx params so you can set the `from`, etc. 

Also added a conditional to skip definition of this method if user has already defined it in their contract. Believe this was problematic for [ERC-721](https://github.com/ethereum/eips/issues/721) although now I can't find any reference to a `send` method in the interface.


mg6maciej:
Might have been problematic to https://github.com/ethereum/EIPs/issues/777 where `send` is defined.

PR #1039, opened 2018-06-20
Support reading package name with /

makoto:
This is to resolve https://github.com/trufflesuite/truffle-artifactor/issues/74


cgewecke:
@makoto Thanks for re-opening! Very sorry this was misplaced in the project re-org. 

Hopefully doing a release today and will get this in there.

PR #1046, opened 2018-06-20
Instructions for adding dependencies

cgewecke:
Self-explanatory. Should we address how someone might make a dependency uniform across the packages? Or ask them to do that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ cd packages/<truffle-package>
 yarn add <npm-package> [--dev] # Use yarn
 ```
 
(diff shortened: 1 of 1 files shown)


gnidan:
> Should we address how someone might make a dependency uniform across the packages? Or ask them to do that?

We could. I figured I was using [`lernaupdate`](https://www.npmjs.com/package/lerna-update-wizard) non-canonically and not part of "Truffle's development process", but it's so darn easy that maybe it's worth recommending.

cgewecke:
@gnidan Ok I think I'll add a little link to that in case someone finds it useful.

PR #1052, opened 2018-06-21
Add pre-release script

cgewecke:
We'd like to publish experimental builds - e.g. webpacks of `packages/truffle` checked out to a branch. Lerna doesn't let you publish a single scoped package. 

This adds a script to `packages/truffle` which checks out a branch and runs `npm version <version-tag>` where <version-tag> is a tagged prerelease increment of the current version. e.g
```shell
4.1.13-next.0
```
Command is:
```shell
node ./scripts/prereleaseVersion.js <branch> <tag>
```
Needs `npm publish` and `git push` added to the end of it. Those are left out so it can be evaluated without causing damage.

### DIFF
--- a/packages/truffle/package.json
+++ b/packages/truffle/package.json
@@ -37,8 +37,8 @@
     "build": "npm run build-cli",
     "build-cli": "webpack --config ./cli.webpack.config.js",
     "test": "npm run build-cli && mocha --colors",
-    "test:ci": "./scripts/test.sh",
-    "test:geth": "MAIN_REPO_CI=true npm test"
+    "version:byoc-safe": "node ./scripts/prereleaseVersion.js byoc-safe byoc",
+    "version:next": "node node ./scripts/prereleaseVersion.js next next"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


packages/truffle/package.json
```diff
@@ -37,8 +38,8 @@
     "build": "npm run build-cli",
     "build-cli": "webpack --config ./cli.webpack.config.js",
     "test": "npm run build-cli && mocha --colors",
-    "test:ci": "./scripts/test.sh",
-    "test:geth": "MAIN_REPO_CI=true npm test"
+    "version:byoc-safe": "node ./scripts/prereleaseVersion.js byoc-safe byoc",
+    "version:next": "node node ./scripts/prereleaseVersion.js next next"
```
gnidan: Too many nodes!

packages/truffle/scripts/prereleaseVersion.js
```diff
@@ -0,0 +1,39 @@
...
+exec(`git checkout ${branch}`, {stdio:[0,1,2]});
+console.log();
+
+// Read package
+console.log('Loading package');
+let pkg = fs.readFileSync('./package.json');
+pkg = JSON.parse(pkg);
+
+// Get semver increment string
+const version = semver.inc(pkg.version, step, tag);
```
gnidan: Does this also bump the major version?

packages/truffle/package.json
cgewecke: 🦅 👁

PR #1055, opened 2018-06-22
truffle-provider: allow async provider creation

benjamincburns:
Allows user to specify a `Promise` which resolves to a `provider`, or an asynchronous function which returns a provider as the `provider` field of the `options` object.


benjamincburns:
Crap - for code consistency I'm wondering if I should've just used the callback pattern rather than returning promises... I at least avoided use of the `async` and `await` keywords.

gnidan:
It's messy to have to add this in every place, but I'm not sure if I can think of a better solution. I'll think about it and see if I can't come up with an alternative.

gnidan:
@benjamincburns could you rebase this against `next`? I'm in favor of this solution for the time being.

gnidan:
@benjamincburns shoot, this is causing `develop` -> `next` to fail. I'm going to revert for now so as not to risk the branches getting more out of sync

Issue #1051, opened 2018-06-21
Truffle Debug: TypeError: Cannot read property 'sourcePath' of undefined


jleeh:
### Environment
- OS: W10 Pro 1803
- Truffle v4.1.12 (core: 4.1.12)
- Solidity v0.4.24 (solc-js)
- Ganache CLI v6.1.3 (ganache-core: 2.1.2)
- NodeJS v8.9.3

### Issue
Every time I try to debug a tx, I get the following:
```
-> truffle debug 0x703ae6c18730a773f16b289b5277305eee9eb9441c371cf403137e9268c61402

Compiling ./contracts/ERC20.sol...
Compiling ./contracts/ERC20Basic.sol...
Compiling ./contracts/ERC677.sol...
Compiling ./contracts/LinkPool.sol...
Compiling ./contracts/LinkToken.sol...
Compiling ./contracts/Migrations.sol...
Compiling ./contracts/Node.sol...
Compiling ./contracts/NodeFactory.sol...
Compiling ./contracts/NodeStaking.sol...
Compiling ./contracts/Ownable.sol...
Compiling ./contracts/PoolOwners.sol...
Compiling ./contracts/PoolStorage.sol...
Compiling ./contracts/PoolStorageFactory.sol...
Compiling ./contracts/SafeMath.sol...
Compiling ./contracts/interface/NodeArithmeticInterface.sol...
Compiling ./contracts/interface/NodeFactoryInterface.sol...
Compiling ./contracts/interface/NodeInterface.sol...
Compiling ./contracts/interface/PoolStorageFactoryInterface.sol...
Compiling ./contracts/interface/PoolStorageInterface.sol...
Compiling ./contracts/lib/NodeArithmetic.sol...

Gathering transaction data...

redux-saga error: uncaught at session.saga
at session.saga
 TypeError: Cannot read property 'sourcePath' of undefined
    at recordSources$ (/usr/lib/node_modules/truffle/build/webpack:/.-debugger/dist/debugger.js:3734:1)
    at tryCatch (/usr/lib/node_modules/truffle/build/webpack:/Users/gnidan/src/work/truffle/~/regenerator-runtime/runtime.js:62:15)
    at Generator.invoke [as _invoke] (/usr/lib/node_modules/truffle/build/webpack:/Users/gnidan/src/work/truffle/~/regenerator-runtime/runtime.js:296:1)
[...]

Issue #998, opened 2018-06-10
darq-truffle@next debugger not working


RobertoC27:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Using the debug command is not working with `darq-truffle`, it exits the debug mode automatically. Trying the same command with ```truffle``` is actually working.

## Steps to Reproduce

```darq-truffle debug txhash``` enters debuging mode and then it closes it without user interaction.

## Expected Behavior

Expected to be able to debug using ```darq-truffle```

## Actual Results

Debugger exits automatically.
```darq-truffle debug thxash``` 
![image](https://user-images.githubusercontent.com/8262009/41203489-f3de0f4c-6c94-11e8-8b2d-d3dfcdd28a2d.png)

Had to use ```truffle debug txhash``` instead to debug
![image](https://user-images.githubusercontent.com/8262009/41203510-3c9d66e2-6c95-11e8-8dde-f7c8acb9b22d.png)

## Environment

* Operating System:  Windows 10
* Ethereum client: Ganache CLI v6.1.0 (ganache-core: 2.1.0)
* darq-truffle version (`darq-truffle version`):  Truffle v4.1.4-next.8 (core: 4.1.8)
* Truffle version (`truffle version`): Truffle v4.1.8 (core: 4.1.8)
* node version (`node --version`): v8.11.1
* npm version (`npm --version`): 5.6.0

cgewecke:
@RobertoC27 Thanks! Will check this out.

cgewecke:
Update: This turned out to be a publishing error. The debugger is a built project and the way we've been publishing `darq-truffle` failed to take that into account.  

It looks like starting Monday we'll be publishing anything/everything from this repository since we've moved all the modules here and have Lerna installed. This bug should be fixed then and `darq-truffle` will just be called `truffle@next`. 

Will ping this issue when that happens.

RobertoC27:
Just to clarify, i should prepare to use `truffle@next` from Monday on.

cgewecke:
@RobertoC27 Sorry for the delay. Just published `truffle@next`. If you have a chance to report back about whether the debugger works that would be great, thanks!

RobertoC27:
@cgewecke sorry for the delay, i'm a bit short on time but will try to do it during the weekend. Any installation instructions that may have changed? or just `npm install truffle@next`

cgewecke:
[...]

PR #1060, opened 2018-06-22
Ensure truffle-debugger targets Node 6

gnidan:
Via babel preset-env's `targets` directive.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -5,5 +5,6 @@
   ],
   "version": "independent",
   "npmClient": "yarn",
+  "npmClientArgs": ["--ignore-engines"],
   "useWorkspaces": true
 }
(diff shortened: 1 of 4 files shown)


gnidan:
So it looks like we need Node 8 to build the debugger, but we should still be able to target Node 6. Will need to update Travis to run integration tests on Node 6 after building on 8

cgewecke:
Oh also!  The webpack build is hard-coding our own computer's paths into the build. . . It's happening on the Lerna release and the experimental releases. See:
+ [here](https://github.com/trufflesuite/truffle/issues/998#issuecomment-399711287)
+ [here](https://github.com/trufflesuite/truffle/issues/1051#issue-334429832)

PR #1068, opened 2018-06-26
Move ISSUE_TEMPLATE to root

gnidan:


gnidan:
@cgewecke, let me know if you think there are any useful updates to make in the course of this move.

cgewecke:
@gnidan I like this one - it's simple and clear. 

The only thing I kind of wish for is something that tells people to open box related issues at the box in question. We just don't have the resources to support those and their code isn't here.

gnidan:
@cgewecke that's a good point. let me see if I can word that in.

gnidan:
Actually I'm not sure the best way to do that, so I'm just gonna merge this now and revisit later.

PR #1073, opened 2018-06-27
Revert "truffle-provider: allow async provider creation"

gnidan:
Reverts trufflesuite/truffle#1055

(Since #1072 `develop` -> `next` merge is a non-trivial conflict)

cc @cgewecke @benjamincburns

### DIFF
--- a/packages/truffle-config/index.js
+++ b/packages/truffle-config/index.js
@@ -175,16 +175,6 @@ function Config(truffle_directory, working_directory, network) {
   });
 };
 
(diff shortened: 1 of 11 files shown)


cgewecke:
@benjamincburns Yes, sorry - I need to sync to develop before this because `next` is a branch we're publishing and someone has an open issue about the debugger being broken on it. Unfortunately we have to make sure we can continue to patch develop and merge it into next. . . a little confusing.

Issue #1058, opened 2018-06-22
Upgrade to 4.1.12 - Truffle Exception


djneely:
I had truffle-4.1.8 up and running fine. Did an update today to 4.1.12 and now getting the following error just trying to check the version. All other commands fail as well.

```
$ truffle version
/usr/local/lib/node_modules/truffle/build/cli.bundled.js:60353
!function(e,a){ true?module.exports=a():"function"==typeof define&&define.amd?define("ganache",[],a):"object"==typeof exports?exports.ganache=a():e.ganache=a()}(global,function(){return function(e){var a={};function t(c){if(a[c])return a[c].exports;var n=a[c]={i:c,l:!1,exports:{}};return e[c].call(n.exports,n,n.exports,t),n.l=!0,n.exports}return t.m=e,t.c=a,t.d=function(e,a,c){t.o(e,a)||Object.defineProperty(e,a,{configurable:!1,enumerable:!0,get:c})},t.r=function(e){Object.defineProperty(e,"__esModule",{value:!0})},t.n=function(e){var a=e&&e.__esModule?function(){return e.default}:function(){return e};return t.d(a,"a",a),a},t.o=function(e,a){return Object.prototype.hasOwnProperty.call(e,a)},t.p="",t.w={},t(t.s=1036)}([function(e,a,t){try{var c=t(1);if("function"!=typeof c.inherits)throw"";e.exports=c.inherits}catch(a){e.exports=t(1021)}},function(e,a){e.exports=__webpack_require__(3)},function(e,a,t){"use strict";var c="function"==typeof Symbol
SyntaxError: Unexpected identifier
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:393:7)
    at startup (bootstrap_node.js:150:9)
```

djneely:
If it helps just uninstalled 4.1.12 and re-installed 4.1.8 and everything is back in order and working.

cgewecke:
[...]

Issue #1053, opened 2018-06-21
Syntax error on Node 6, (Node 8 is ok)


etsangsplk:

cgewecke:
@etsangsplk Node 6. Please let us know if you're seeing problems with that version.

 Node's LTS is currently 8 and it should work on that as well. 

We've seen a few unusual problems/crashes with the latest Node version (10).

cgewecke:
Closing, lmk if you are having trouble with this an I will re-open.

etsangsplk:
@cgewecke 
```
nvm install v6.12.2
v6.12.2 is already installed.
Now using node v6.12.2 (npm v3.10.10)
```

```
/usr/lib/node_modules/truffle/build/cli.bundled.js:60353
!function(e,a){ true?module.exports=a():"function"==typeof define&&define.amd?define("ganache",[],a):"object"==typeof exports?exports.ganache=a():e.ganache=a()}(global,function(){return function(e){var a={};function t(c){if(a[c])return a[c].exports;var n=a[c]={i:c,l:!1,exports:{}};return e[c].call(n.exports,n,n.exports,t),n.l=!0,n.exports}return t.m=e,t.c=a,t.d=function(e,a,c){t.o(e,a)||Object.defineProperty(e,a,{configurable:!1,enumerable:!0,get:c})},t.r=function(e){Object.defineProperty(e,"__esModule",{value:!0})},t.n=function(e){var a=e&&e.__esModule?function(){return e.default}:function(){return e};return t.d(a,"a",a),a},t.o=function(e,a){return Object.prototype.hasOwnProperty.call(e,a)},t.p="",t.w={},t(t.s=1036)}([function(e,a,t){try{var c=t(1);if("function"!=typeof c.inherits)throw"";e.exports=c.inherits}catch(a){e.exports=t(1021)}},function(e,a){e.exports=__webpack_require__(3)},function(e,a,t){"use strict";var c="function"==typeof Symbol&&"sym

SyntaxError: Unexpected identifier
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:383:7)
[...]

PR #1075, opened 2018-06-29
Update ganache-cli to 6.1.4 accepting minor and patch updates

facuspagnuolo:
This PR unfixes the `ganache-cli` version of all the repositories included in truffle-core and updates it to the latest released version. 

Honestly, I couldn't understand how you guys are dealing with this main repo regarding sub-repositories' dependencies. Please let me know if I'm doing right.

### DIFF
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -15,7 +15,7 @@
   "license": "MIT",
   "devDependencies": {
     "chai": "4.1.2",
-    "ganache-cli": "6.1.0-beta.4",
+    "ganache-cli": "^6.1.4",
     "mocha": "5.2.0",
     "require-nocache": "^1.0.0",
     "solc": "0.4.24",
--- a/packages/truffle-core/package.json
+++ b/packages/truffle-core/package.json
@@ -16,7 +16,7 @@
     "ethpm-registry": "0.0.10",
     "finalhandler": "^0.4.0",
     "fs-extra": "6.0.1",
-    "ganache-cli": "6.1.0-beta.4",
+    "ganache-cli": "^6.1.4",
     "lodash": "4.17.10",
     "mkdirp": "^0.5.1",
     "mocha": "5.2.0",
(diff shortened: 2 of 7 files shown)


cgewecke:
Thanks @facuspagnuolo :)

This looks right - we're trying to keep a single version of each dep and let `yarn` manage the workspaces for us. It should hoist everything common up to the root the project. And we've been using [this](https://www.npmjs.com/package/lerna-update-wizard) cli 'wizard' to make multi-package dependency changes (because it's easier).

Unfortunately . . . we did this last week (bumping to `6.1.3`)  and it breaks Truffle for Node 6 users. See [here](https://github.com/trufflesuite/truffle/issues/1053) and [here](https://github.com/trufflesuite/truffle/issues/1058). Newer ganache-core has a dependency  (eth-block-tracker) that uses `async`. 

We don't feel we can make this change in a minor version and `ganache` believes it's not really viable to undo its changes either.  
 
Some possibilities:
+ we're also now publishing `truffle@next` and `truffle@byoc`
+ We could re-target this PR against `next` (which will be Truffle V5 and contains some breaking changes from web3 1.0) 
+ We could re-target it against `byoc` - which just has a new feature that lets you use any solc compiler version (usage docs [here](https://github.com/trufflesuite/truffle-compile/tree/byoc-docs#truffle-compiler)). 
+ You could publish your own truffle webpack as a stopgap using [this script](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle/scripts/prereleaseVersion.js) at `packages/truffle`

FWIW Truffle V4 is not going to have more work done on it - if something is super broken we'll fix it, but otherwise the project is focussed on the next release - we're hoping to publish a beta in the next couple weeks.

facuspagnuolo:
Thanks for your response @cgewecke! If your planning to release Truffle V5 soon, I think it would be better to point to `next`, do you agree?

cgewecke:
@facuspagnuolo Yes . . . Actually - I will merge this into both and republish them today. 

The issue with `next` is that: 
+ it has more changes and probably more bugs. I wrote a +/- 1500 line test suite in it last month and it was ok - however you might run into weird issues. (I'm in favor of this because it will help us to find problems, but you might be less excited about it.)
+ There are breaking changes with web3 1.0 - it outputs strings instead of BigNumber, and uses BN, etc etc. There's a full list of things we've seen [here](https://github.com/trufflesuite/truffle-contract/tree/web3-one-readme#truffle-contract-for-web3-10).

PR #1083, opened 2018-07-04
Add compilers key to config

cgewecke:
+ modifies config key to support multiple compilers
+ moves existing solc defaults there
+ grandfathers in the old solc key at `truffle-compile`

**Example** (updated)
```javascript
compilers: {
  solc: {
    version: "0.4.24",
    docker: true,
    settings: {
      optimizer: {
        enabled: true,
        runs: 1000
      }
    }
  },
  vyper: {
    version: "0.1.0-beta.1",
    snap: true
  }
}
```

### DIFF
--- a/packages/truffle-compile/index.js
+++ b/packages/truffle-compile/index.js
@@ -41,8 +41,8 @@ var compile = function(sources, options, callback) {
 
   // Grandfather in old solc config
   if (options.solc){
-    compilers.solc.evmVersion = options.solc.evmVersion;
-    compilers.solc.optimizer = options.solc.optimizer;
+    compilers.solc.settings.evmVersion = options.solc.evmVersion;
+    compilers.solc.settings.optimizer = options.solc.optimizer;
   }
 
   // Ensure sources have operating system independent paths
@@ -102,8 +102,8 @@ var compile = function(sources, options, callback) {
     language: "Solidity",
     sources: {},
     settings: {
-      evmVersion: options.compilers.solc.evmVersion,
-      optimizer: options.compilers.solc.optimizer,
+      evmVersion: options.compilers.solc.settings.evmVersion,
+      optimizer: options.compilers.solc.settings.optimizer,
       outputSelection: outputSelection,
     }
   };
--- a/packages/truffle-compile/index.js
+++ b/packages/truffle-compile/index.js
@@ -36,10 +36,14 @@ var compile = function(sources, options, callback) {
 
   expect.options(options, [
     "contracts_directory",
-    "solc"
+    "compilers"
   ]);
 
-
+  // Grandfather in old solc config
+  if (options.solc){
+    compilers.solc.evmVersion = options.solc.evmVersion;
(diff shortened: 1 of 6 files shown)


gnidan:
I think `evmVersion` and `optimizer` should maybe be put in their own `settings` object, to avoid the potential for future namespace collisions... e.g.:
```javascript
compilers: {
  solc: {
    version: "0.4.24",
    docker: true,
    settings: {
      optimizer: {
        enabled: true,
        runs: 1000
      }
    }
  },
  vyper: {
    version: "0.1.0-beta.1",
    settings: {
      snap: true
    }
  }
}
```

packages/truffle-compile/profiler.js
```diff
@@ -189,7 +189,7 @@ module.exports = {
       var compilationTargets = [];
 
       // Load compiler
-      var supplier = new CompilerSupplier(options.compiler)
+      var supplier = new CompilerSupplier(options.compilers)
       supplier.load().then(solc => {
```
gnidan: Should this be `options.compilers.solc`, like in `truffle-compile/index.js`?

packages/truffle-compile/index.js
```diff
@@ -98,8 +102,8 @@ var compile = function(sources, options, callback) {
     language: "Solidity",
     sources: {},
     settings: {
-      evmVersion: options.solc.evmVersion,
-      optimizer: options.solc.optimizer,
+      evmVersion: options.compilers.solc.settings.evmVersion,
+      optimizer: options.compilers.solc.settings.optimizer,
```
gnidan: Do we default `compilers.solc.settings` anywhere? Or are those guaranteed non-null because of the grandfathering above?

Regardless, maybe this should be more explicit, in case that grandfathering goes away.

packages/truffle-compile/profiler.js
cgewecke: Oh my god, good catch.

packages/truffle-compile/index.js
cgewecke: They're defaulted in the `truffle-config` [here](https://github.com/trufflesuite/truffle/pull/1083/files#diff-ea3ccf54753c9aaae2ba020420166f18R40).  I'm going to double-check that those are coming through but I think the integration tests would crash on the grandfathering line if that wasn't an object.

gnidan: Oh cool, I missed that.

cgewecke: Yes it seems be working . . . this what the Supplier config looks like from a scenario test after it's been copied in the constructor:
```javascript
{
 "version": null,
 "versionsUrl": "https://solc-bin.ethereum.org/bin/list.json",
 "compilerUrlRoot": "https://solc-bin.ethereum.org/bin/",
 "dockerTagsUrl": "https://registry.hub.docker.com/v2/repositories/ethereum/solc/tags/",
 "cache": true,
 "settings": {
  "optimizer": {
   "enabled": false,
   "runs": 200
  },
  "evmVersion": "byzantium"
 }
}
```

PR #1084, opened 2018-07-04
Ensure not-null binary before adding instance

jleeh:
This is for issue #1051 

Simple check that ensures binary is not equal to "0x0" before searching for it and adding the instance to an address.

### DIFF
--- a/packages/truffle-debugger/lib/evm/sagas/index.js
+++ b/packages/truffle-debugger/lib/evm/sagas/index.js
@@ -27,11 +27,13 @@ export function *addContext(contractName, binary) {
  */
 export function *addInstance(address, binary) {
   let search = yield select(evm.info.binaries.search);
-  let { context } = search(binary);
+  if (binary != "0x0") {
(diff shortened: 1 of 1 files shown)


packages/truffle-debugger/lib/evm/sagas/index.js
```diff
@@ -27,11 +27,13 @@ export function *addContext(contractName, binary) {
...
   let search = yield select(evm.info.binaries.search);
-  let { context } = search(binary);
+  if (binary != "0x0") {
+    let { context } = search(binary);
 
-  yield put(actions.addInstance(address, context, binary));
+    yield put(actions.addInstance(address, context, binary));
 
-  return context;
+    return context;
```
HarryR: Surely you need to return `context` regardless of if an instance is emitted?

This will break method chaining if it returns `null`

gnidan: I don't think those return values are actually used... I think returning `undefined` is fine.

PR #1085, opened 2018-07-05
Migrate NPM dependencies

cgewecke:
Hi @cag! Hope you're well. Finally getting this queued up to go into the new migrations work here. Have rebased and moved your (fantastic) tests into slightly different format - we now have mock project folders under `test/sources` and a little utility to load them into a sandbox.

Does everything look ok?

I also need your review and advice . . . still having trouble getting the dry-run test to pass - I'm not sure if the cause is the one you note in the test or an error I've introduced while merging your work into the newer migrate command code at truffle-core [here](https://github.com/trufflesuite/truffle/compare/async-deployer...migrate-deps?expand=1#diff-352d250c5a139d985a5bb44252cd90d4). 

Maybe comparing the outputs below could shed some light on this? 
+ [gist](https://gist.github.com/cgewecke/2350da15138ab01ed4419e156c4969c7) of successful migration test-case
+ [gist](https://gist.github.com/cgewecke/1713b5821c673288ff1b7f029030f5e5) of failing dry-run test-case

Additionally - I'm wondering it it's possible to get an example of a contract suite that's consuming Gnosis so I can run a simulation of this against Rinkeby. Am doing that with `consensys/plcrvoting` [here](https://github.com/cgewecke/rinkeby-migrations). I'd just like to get a sense of whether the output makes sense, if there's signal we could add about deploying deps, etc. 

[Edit - We have new mode we're calling 'production' that automatically does a dry / then a real run, and that section of the code definitely needs scrutiny. It passes our non-npm-dep test (on the Geth run) but we need a dep test - pretty sure the code I wrote [here](https://github.com/trufflesuite/truffle/blob/1138099cb50f5e8febee4549dd92a321506df97f/packages/truffle-core/lib/commands/migrate.js#L117-L140)  and [here](https://github.com/trufflesuite/truffle/blob/1138099cb50f5e8febee4549dd92a321506df97f/packages/truffle-core/lib/commands/migrate.js#L181-L183) would fail.)

Thanks again for this!! Just great.


cag:
Hey @cgewecke , so sorry I haven't responded yet! :sob: It's been kinda crazy, so I haven't had a chance to really look over this, but I promise I haven't forgotten about you! I still need to make some time to sit down and really go through this.

In fact, I've been meaning to do something for TruffleCon related to this ;) This is just a rumor though, to make up for the fact that I have not really looked at this PR yet :'(

Anyway, just wanted to bump this so you know I am alive and intending to resolve this soon.

cgewecke:
@cag Oh no worries at all, this can safely sit staged here for as long as you'd like.  🙂

cag:
So this is something unrelated to the test note, I think. This is also something which I've had some trouble with doing successfully, so I'm not surprised that it is so brittle.

Off the cuff, I'd guess one of two things are happening. During a dry run, either the deployment data isn't persisted between dry runs, or the resolvers for the runs aren't set up to find the temporary artifacts created by earlier dependencies in a dry run.

Anyway, I'll give this a more thorough look soon.

cgewecke:
@cag Great! 

Ok just one last thing - do you happen to have an example at hand I could do more of an E2E test with? If you just point me in the right direction that's more than enough...

Is there anything over at Gnosis - like an example that consumes some Gnosis migrations? NBD if not - I will mock something up, I'm really excited to try this out 🙂

cag:
No... I've been doing this work so that some stuff can finally be
refactored to use this

cgewecke:
@cag Quick status note: synced this to the latest changes and have published the branch to npm as: 
```
truffle@gnosis
```
Will play around with it a bit this week and work on some documentation, but I think this is close to final review. If you see anything you'd like to add/change please lmk.

cag:
@cgewecke Okay so this didn't work with `truffle test` out of the box with a default configuration. That omission has been addressed here: #1310

gnidan:
Closing this as duplicate now!

Issue #1093, opened 2018-07-10
solidity will remove ``constant`` from their json ABI with version ``0.5.0``


ekpyron:
With the next breaking release solidity will remove the ``constant`` field from the JSON ABI for contract functions, since it is redundant (``constant`` is the same as ``stateMutability === 'view' || stateMutability === 'pure'``).

I'm in the process of implementing this change in the solidity repo, but we want to make sure that everything that depends on the ``constant`` field can be updated first.

I think the main change that is required on your end, should be in ``truffle-compile/index.js``. However, I'm not an expert for your code base and any help would be highly appreciated. @gnidan @cgewecke?

ekpyron:
And to already give you a heads-up: We plan to remove the ``payable`` field as well, since it is the same as ``stateMutability == 'payable'``.

cgewecke:
Thanks @ekpyron :)

The one place I see is actually at [truffle-contract/lib/contract.js](https://github.com/trufflesuite/truffle/blob/next/packages/truffle-contract/lib/contract.js#L41) where Truffle automates whether or not web3 methods are invoked as calls or sends. This should be simple to update and be backwards compatible as well using the `stateMutability` alternative you suggest above.

@gnidan?

gnidan:
@ekpyron is removing `payable` part of v0.5.0 also? I'd like to clean this up all in one fell swoop.

ekpyron:
@gnidan (Already mentioned on gitter, but I'll comment here as well for reference):

If we manage to get external projects to update to be compatible soon enough, we'd like to remove both ``payable`` and ``constant`` in 0.5.0.

gnidan:
Truffle will support either format in v5 onward. Closing this for issue maintenance. Please let us know if you run into any trouble.

(Also please let us know when you plan to proceed with that ABI format change!)

Thanks!

PR #1098, opened 2018-07-11
Stop expecting `constant` and `payable` in ABIs

gnidan:
Ref: #1093 

Solidity is moving away from usage of `constant` and `payable` properties in the ABI. This PR removes those fields as being any kind of required, and instead now relies on the `stateMutability` property.

- Update schema accordingly
- Change tests to reflect new requirements
- Also (for good measure) put `pragma experimental "v0.5.0"` in the schema `solc` tests
- Fix a readme

### DIFF
--- a/packages/truffle-contract/lib/contract.js
+++ b/packages/truffle-contract/lib/contract.js
@@ -38,7 +38,10 @@ var contract = (function(module) {
 
       switch(item.type) {
         case "function":
-          var isConstant = item.constant === true;
+          var isConstant =
+            ["pure", "view"].includes(item.stateMutability) ||  // new form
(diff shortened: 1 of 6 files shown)


packages/truffle-contract/lib/contract.js
```diff
@@ -38,7 +38,10 @@ var contract = (function(module) {
 
       switch(item.type) {
         case "function":
-          var isConstant = item.constant === true;
+          var isConstant =
+            ["pure", "view"].includes(item.stateMutability) ||  // new form
+            item.constant;  // deprecated case
+
```
cgewecke: 👍

gnidan:
Thanks!

PR #1099, opened 2018-07-12
Add truffle-reporters

cgewecke:
Opening this as separate PR into `async-deployer` since this is a big structural change.

+ Creates a `truffle-reporters` module and moves migrations reporter there.
+ Adds `truffle-reporters` as a dep to `truffle-migrate`
+ Adds `truffle-reporters` as a dev dep to `truffle-deployer`
+ Updates deployer tests to consume production reporter
+ Removes some remnant batch deployment logic in the reporter


packages/truffle-deployer/test/deployer.js
```diff
@@ -51,7 +56,10 @@ describe("Deployer (sync)", function() {
       }
     }
     deployer = new Deployer(options);
-    reporter = new Reporter(deployer);
+    reporter = new Reporter();
+    reporter.setDeployer(deployer);
+    reporter.setMigration(mockMigration);
+    reporter.listen();
```
gnidan: Suggestion: what if we used a builder pattern for this?

```javascript
reporter = new Reporter()
  .setDeployer(deployer)
  .setMigration(mockMigration)
  .listen();
```

packages/truffle-deployer/test/helpers/reporter.js
```diff
@@ -1,292 +0,0 @@
...
-
-      default:
-        message = this.messages('default', payload.contract, error.message);
-        this.deployer.logger.error(message);
-    }
-  }
-}
-
-module.exports = Reporter;
-
```
gnidan: This is deleted just because it's now available to deployer as its own package? It's not losing anything?

cgewecke: Yes, exactly - this was a prototype for the production one.

packages/truffle-deployer/test/deployer.js
cgewecke: The Reporter is instantiated once for a complete migrations cycle, which allows us to summarize their aggregate result. But it needs to be consume individual deployer and migration instances as they arise, hence the decoupling here. Definitely looks like we could simplify by having an `init` method or something that did all this in one place. 

Possibly make this change at `async-deployer`.

Issue #690, opened 2017-11-25
Deploying Migrations... Stuck forever - implement reasonable timeout with a message "check for other stuff running on port 8545"


stefek99:
Wasted a lot of effort.

Didn't realise I had Parity running.

Some guys can have Ethereum Wallet / Mist / geth running as well as `testrpc`.

The `Deploying Migrations...` message was stuck forever.

Killed Parity, works like a charm.

Provide a reasonable timeout, I wasn't sure if something was happening... Or maybe another message (if simple enough):

> connecting to BLAH BLAH running on port BLAH BLAH...

PR #1106, opened 2018-07-15
Fix hanging on migrate command

mpetrunic:
After hours of looking trough code, I found the reason why migrate hangs even though everything executes successfully. Command callback only exits process if there is an err object in callback to enable usage of `truffle console` command.

I've added optional close param to callback so commands like `truffle migrate` can exit process from within.

This should resolve issue #690 and also issues on old repo:
https://github.com/trufflesuite/truffle-migrate/issues/31
https://github.com/trufflesuite/truffle-migrate/issues/14

PR #1107, opened 2018-07-15
Fix hanging on migrate command

mpetrunic:
New PR from #1106 

This should resolve issue #690 and also issues on old repo:
trufflesuite/truffle-migrate#31
trufflesuite/truffle-migrate#14

### DIFF
--- a/packages/truffle-core/cli.js
+++ b/packages/truffle-core/cli.js
@@ -10,6 +10,12 @@ var OS = require("os");
 
 var command = new Command(require("./lib/commands"));
 
+// Hack to suppress web3 MaxListenersExceededWarning
+// This should be removed when issue is resolved upstream:
+// https://github.com/ethereum/web3.js/issues/1648
+var listeners = process.listeners('warning');
+listeners.forEach(listener => process.removeListener('warning', listener));
+
 var options = {
   logger: console
 };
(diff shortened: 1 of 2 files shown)


gnidan:
Hi @mpetrunic could you also please cherry-pick this commit: https://github.com/trufflesuite/truffle/pull/1108/commits/6b9523f4bf794e47d8b6a84890c0013b7587d475

I noticed the debugger wasn't working with this `process.exit(0)` in place, that commit fixes it. We shouldn't merge this PR without that.

Also, feel free to add the `--no-exit` flag, I think it might be handy in some cases.

Thanks!

mpetrunic:
This is weird, I pushed commits to my forks develop branch but they ain't showing here...

gnidan:
Try again maybe?

mpetrunic:
Amending last commit seems to fix the bug. Here you go :)

gnidan:
Sorry for the delay on this @mpetrunic, but hadn't gotten the chance to try this out yet. I just tried it out and it seems like `--no-exit` isn't getting passed through. Upon considering it, though, I don't think we need this behavior (and we can add it later if we do). I'm going to just revert your latest commit and merge this. Thanks!

gnidan:
(Once Travis finishes, for good measure)

PR #1108, opened 2018-07-16
Support debugging mappings

gnidan:
Add support for mapping types in the debugger.

High-level approach:
- Begin tracking decoded values for all expressions. That is, every time the debugger steps through a given AST node, attempt to decode the current runtime value for that node and store it for reference later
- Upon stepping through an index-access AST node, use that tracked expression value to determine the key (`someMapping[x]` will use runtime value of the _Identifier_ node for `x`... note: this is not necessarily the scoped variable value, or at least it's computed differently)
- Track all keys for a given mapping, and use those when computing all known storage locations
- Fix storage slot lookup and implement mapping pointer dereferencing.

Missing from this PR:
- ~~Decoding mapping values in structs. I'm leaving this for another effort to keep this PR from growing too much.~~ Merged into this PR from #1110


packages/truffle-debugger/lib/data/decode/utils.js
```diff
@@ -267,7 +267,7 @@ export function toHexString(bytes, length = 0, trim = false) {
 }
 
 export function toBytes(number, length = 0) {
-  if (number < 0) {
+  if (number < 0 || number === "") {
```
gnidan: (This might not be necessary anymore, might be worth checking)

packages/truffle-debugger/lib/data/decode/utils.js
```diff
@@ -109,16 +109,20 @@ function allocateDeclaration(declaration, refs, slot, index) {
...
-  let struct = refs[definition.typeName.referencedDeclaration];
-  debug("struct: %O", struct);
+  if (typeClass(definition) == "mapping") {
+
+  }
 
-  let result =  allocateDeclarations(struct.variables || [], refs, slot, index);
-  debug("struct result %o", result);
-  return result;
+  return allocateValue(slot, index, byteSize);
```
gnidan: Oops this commit is worthless. Note to self: remove

packages/truffle-debugger/lib/data/decode/utils.js
gnidan: Seems this is, in fact, still necessary.

PR #1110, opened 2018-07-17
Support debugging mappings of structs of mappings of structs...

gnidan:
Sequel to #1108 

This supports nested mappings by updating calculation of storage slots and struct allocations.

- Update `slot` data type to encode offset natively, to represent arbitrarily nested storage variable declarations. When computing recursively described slots, respect offset values in the inner-slot.
- Move known mapping keys outside assignments state and into separate field.
- Pass known mapping keys as info to decode functions
- Compress decode function signatures and remove annoying `...args` splats.


gnidan:
Thanks @cgewecke!

Issue #1104, opened 2018-07-15
Use errors rather than logging to report migrations errors


cgewecke:
The current strategy in next only works in the context of verbose reporting. At present if a migrations fails when logging is silent (e.g. the tests) we get no information about what happened. 

We need to patch this as stop-gap pending efforts to move all the deployment error reporting over to `truffle-contract`

cgewecke:
#1113

PR #1113, opened 2018-07-17
Throw migration errors instead of logging them

cgewecke:
#1104

Patch to wire the reporter error output into thrown errors. Underlying issue is that migrations failures in the test suite don't report anything useful since migrations logging is turned off there.

Have left the logging open as an unused option in the reporter but all this error management  actually needs to be moved into `truffle-contract` and rationalized there in a later PR. 

Also reworked linking so that's it's fully async. Does *not* require that anyone write 
```
await deployer.link()
```
in a classic migration - just makes sure we can run async logic inside the linker and have errors propagate appropriately.


cgewecke:
cc @gnidan I need this in today.

cgewecke:
@gnidan Oh no worries thank you!!

PR #1117, opened 2018-07-18
Listen to repl reset event

sc0Vu:
Why
When I test contract with `truffle console`, I cannot find command to clear console.
So I just add this to native repl `.clear` command.

Change
* Listen to repl reset event and bubble up the event.
* Clear console when user type .clear.


packages/truffle-core/lib/repl.js
```diff
@@ -74,6 +74,12 @@ ReplManager.prototype.start = function(options) {
     self.emit("exit");
   });
 
+  // Bubble the internal repl's reset event
+  this.repl.on("reset", function() {
+    process.stdout.write('\u001B[2J\u001B[0;0f');
```
gnidan: Does this only work for POSIX-compliant terminals? Do you know if this works on Windows as well?

sc0Vu: Hi @gnidan
Thanks for your reply, I'd test on azure window 10, it worked!
Looking forward this commit to be merged:)

eggplantzzz:
Sorry this took so long to merge @sc0Vu!

Issue #1126, opened 2018-07-20
Find elegant upgrade solution for Web3 1.0 number strings


cgewecke:
Problem: Web3 0.x returned numbers as BigNumber (or BN) objects. Web3 1.0 returns them as strings, IIRC they made this change because they thought the lib should be agnostic about how large numbers are managed and strings are common to all the solutions. 

Unfortunately this is moderately catastrophic for large existing test suites. Was not enthusiastic about coercing the new outputs back into the old format because:

+ it's complicated, we'd have to read the ABI and handle all the different data structures that might have numbers nested in them. 
+ truffle-contracts behavior should be isomorphic to web3's - it's injected into the test context and its methods can be called independently.
+ web3 is likely right about handling numbers in this simpler way? 

**Damage Survey**

Zeppelin uses chai-bignumber / BigNumber and has ~550 occurrences in their suite.
```
balance.should.be.bignumber.equal(amount);
```

Colony's `colonyNetwork` has ~400 occurrences of the `.toNumber()` method (which is common to BN and BigNumber). 
```
task1[5].toNumber()
```
The colony case might have a simple resolution - we should be able modify the String prototype for the test context to be something like:
``` 
> String.protoype.toNumber = function() { return parseInt(this) };
> '5'.toNumber();
5
>
```

cgewecke:
On reflection I'm going to look into how difficult it really is to return BNs from a call. It might not be that bad.

gnidan:
Sorry for not getting back to you on this!

I'm on board with returning BN instances if it's not too hard. I think it's better than returning strings, although I expect this might not be very popular.

cgewecke:
@gnidan Yeah, there aren't a lot of good choices here. . . we should probably talk about this more. I'll work up a PR and ask Tim what he thinks too.

cgewecke:
#1138 (Still need to fix events)

PR #1127, opened 2018-07-20
Add toNumber method to String prototype in test

cgewecke:
Not expecting this PR to get a warm reception FWIW 🙂 

Motivation is detailed in #1126. It would make transitioning from V4 to V5 significantly less tedious for people who have hundreds of `toNumber()` calls in their test suites.

Don't have a great place to put a test for this but it works. Running `parseInt` on a huge number string converts it to exponential notation, otherwise normal.

Drawbacks:
+  it muddies the water around whether the return value is a BN or not, although we could address this by being super clear in the release notes that this was added as an expedient.
+ It's being done for the `test` command only, so behavior in other contexts will be variant. Would like to minimize the pain while encouraging a transition to a new order? 

Weigh in!

### DIFF
--- a/packages/truffle-core/lib/test.js
+++ b/packages/truffle-core/lib/test.js
@@ -207,6 +207,10 @@ var Test = {
         }
       };
 
+      // Hack to give web3 1.0 string numbers a convenience method
+      // that BigNumber has - widely used in truffle tests prior to V5.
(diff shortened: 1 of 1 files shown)


cgewecke:
Closing - going to open a PR reformatting number strings as BN because this solution is way too superficial.

PR #1129, opened 2018-07-20
[WIP] V5 release notes

cgewecke:
**DO NOT MERGE** - this is here so people can do review stuff to it if they want to.
Good day! We're pleased to announce the first beta release of Truffle v5. With this release, we're excited to bring you some new features and improvements that will make your life easier. 

What's included, you ask? Well, get ready to use **whatever Solidity version you want** (Truffle will even automatically download it for you.) Or how about an **improved migrations system**, which dry-runs by default and provides lots of information to understand what is going right or wrong. Maybe you've been waiting for **Web3.js 1.0**. Perhaps you're sick of typing `.then()` and want to use **`await` in the console**. Truffle v5 has all of this, and more!

If you're as excited as we are and can't wait, here's how to upgrade:
```bash
npm uninstall -g truffle
npm install -g truffle@beta
```

Or keep reading for more information!

## Release Contents

* [Bring your own compiler](#bring-your-own-compiler)
  + [List available versions at the command line](#list-available-versions-at-the-command-line)
  + [Specify a solcjs version](#specify-a-solcjs-version)
  + [Advanced](#advanced)
  + [Speed comparison](#speed-comparison)
  + [Contract Profiling](#contract-profiling)
* [🐇 🐇 Web3.js 1.0 🐇 🐇](#web3js-1.0)
  + [⚠️ Breaking Changes ⚠️](#breaking-changes)
  + [🍨 Features 🍨](#web3-features)
  + [Methods / `.new` have an EventEmitter interface *and*  return a promise.](#methods-eventemitter)
  + [Events have an EventEmitter interface](#events-eventemitter)
  + [Reason strings!! Find out the reason.](#reason-strings)
  + [Overloaded Solidity methods (credit to @rudolfix and @mcdee)](#overloaded-solidity-methods)
  + [Configure number return format](#configure-number-return-format)
  + [Call methods at any block using the `defaultBlock` parameter.](#default-block)
  + [Automated fueling for method calls and deployments.](#automated-fueling)
  + [Contract deployments & transactions allowed to take as long as they take](#contract-deployments-transactions)
  + [Gas estimation for `.new`](#gas-estimation-for-new)
  + [Config](#config)
* [🐦 🐦  New Migrations 🐦 🐦](#new-migrations)
    - [Features](#new-migrations-features)
    - [Configuration and use](#configuration-and-use)
* [Even More!](#even-more)
  + [**truffle-console** now supports async/await](#truffle-console-now-supports-asyncawait)
[...]

PR #1138, opened 2018-07-22
Add string number reformatting for .call return values

cgewecke:
#1126, [relevant suggestion](https://github.com/trufflesuite/truffle/pull/1129#issuecomment-406517580) 

+ Adds logic to transform Web3 1.0's string number outputs to BigNumber|BN|string before returning value from a `.call`. Defaults to BigNumber  for backwards compatibility with Web3 0.x
+ Adds a `numberFormat` setting to the contract abstraction that allows you to configure this at the costructor level
+ Adds tests for these return cases:
  + primitive
  + number array
  + tuple (named and unnamed elements)
  + tuple with a number array as an element

I will add event return value reformatting in a second PR, after this one is reviewed.

### DIFF
--- a/packages/truffle-contract/lib/execute.js
+++ b/packages/truffle-contract/lib/execute.js
@@ -5,6 +5,7 @@ var StatusError = require("./statuserror");
 var Reason = require("./reason");
 var handlers = require("./handlers");
 var override = require("./override");
+var reformat = require("./reformat");
 
 var util = require('util');
 var execute = {
@@ -84,11 +85,11 @@ var execute = {
 
   /**
    * Executes method as .call and processes optional `defaultBlock` argument.
-   * @param  {Function} fn      method
-   * @param  {Array}    inputs  ABI segment defining the methods inputs.
-   * @return {Promise}          Return value of the call.
+   * @param  {Function} fn         method
+   * @param  {Array}    methodABI  ABI segment defining the methods inputs.
+   * @return {Promise}             Return value of the call.
    */
-  call: function(fn, inputs, address) {
+  call: function(fn, methodABI, address) {
     var constructor = this;
 
     return function() {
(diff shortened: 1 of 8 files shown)


packages/truffle-contract/lib/execute.js
```diff
@@ -84,11 +85,11 @@ var execute = {
 
   /**
    * Executes method as .call and processes optional `defaultBlock` argument.
-   * @param  {Function} fn      method
-   * @param  {Array}    inputs  ABI segment defining the methods inputs.
-   * @return {Promise}          Return value of the call.
+   * @param  {Function} fn         method
+   * @param  {Array}    methodABI  ABI segment defining the methods inputs.
```
gnidan: This should change to `{Object}` I think.

One note: do we want to check for whether it's an array, to accept the array format too?

packages/truffle-contract/lib/execute.js
```diff
@@ -110,9 +111,19 @@ var execute = {
...
-      return constructor
-        .detectNetwork()
-        .then(() => fn(...args).call(params, defaultBlock));
+      return new Promise(async (resolve, reject) => {
+        let result;
+        try {
+
+          await constructor.detectNetwork();
+          result = await fn(...args).call(params, defaultBlock);
+          result = reformat.callOutput.call(constructor, result, methodABI.outputs);
```
gnidan: Does this work? (won't it look for `methodABI.outputs.inputs`?)

cgewecke: I *think* this is right. It's dereferencing the `outputs` keys of the object noted in the comment above . . .

cgewecke: There are a couple changes upstream from here in `contract.js` that pass in a larger chunk of the ABI.
https://github.com/trufflesuite/truffle/pull/1138/files#diff-79802128c9314316abc8de8ceaa5bddcR51

PR #1143, opened 2018-07-23
Use levenshtein distance for context binary search

gnidan:
Since context binary searching has been a problem, this sidesteps the problem a bit by finding the closest matching binary string.

Instead of attempting to remove link-references and whatnot, just use the levenshtein distance to determine the number of inserts/deletes/character changes required to change sought binary to possible found binary. Return the closest match as the result.

Filter the results so that a result is only returned in the case where the distance is <=25% the length of the sought binary.

Wikipedia: [Levenshtein Distance](https://en.wikipedia.org/wiki/Levenshtein_distance)


gnidan:
Oh yeah Levenshtein distance is mad cool :)

Thanks @cgewecke

Issue #1134, opened 2018-07-20
Support for "generic external" compiler


axic:
This has been a discussion with @gnidan with goal of trying to support ewasm contracts written in Rust (in particular the https://github.com/ewasm/ewasm-precompiles) and having the ability to the test them.

It seems like that a good default behaviour in regard to ABI is to assume that if no ABI is supplied then it just accepts the fallback function, e.g. the ABI JSON is as follows:
```
    {
      "payable": true,
      "stateMutability": "payable",
      "type": "fallback"
    },
```

This not only applies to Rust ewasm-precompiles, but any other random compiler, which may not adhere to the ABI standard. Test cases can have their hand written encoded call data.

It would still make sense to utilise the `truffle compile` command to generate all the "build artifacts" for us. In order to achieve that, it seems to be possible extending the "compilers configuration" with something like:
```
"compilers": {
  "external": {
    "command": "cargo build --release --target wasm32-unknown-unknown",
    "targets": [{
      "path": "target/wasm32-unknown-unknown/release/ewasm_precompile_ecadd.wasm",
      "contractName": "ecadd",
      "abi": "<path>"
    }]
  }
}
```

The `abi` field is optional and if missing it could fall back to the above single entry "fallback" ABI.

gnidan:
@axic Are you content to use JS testing for this? Things are getting tricky trying to get Solidity testing working with external compilers; I'm wondering whether or not this is a priority to fix.

axic:
Yes I would still be very happy to use truffle for this.

gnidan:
@axic Excellent! This work is coming along nicely in #1148, nearly ready-for-review. I've had to adapt the config you specified a bit. What do you think of this:

```javascript
"compilers": {
  "external": {
    "command": "cargo build --release --target wasm32-unknown-unknown",
    "targets": [{
      "path": "target/wasm32-unknown-unknown/release/*.wasm",
      "command": "./scripts/process-wasm-output-to-artifact.sh"
    }]
  }
}
```

In this example, you'd define a script `process-wasm-output-to-artifact` that would read each matching `*.wasm`, piping to your script as stdin, and your script would print JSON to stdout that conforms to the Truffle artifact format.

I believe you would only need to output `{ contractName, abi, bytecode, deployedBytecode? }`.

This makes things a bit more general, although it requires a bit of extra work on the end-user's part. I wasn't satisfied with the solution we talked about; it forces too much on Truffle's config.

axic:
[...]

PR #1148, opened 2018-07-25
Support external compilers

gnidan:
Ref: #1134 

# Support external compilers

## Motivation

It'd be nice to use Truffle for workflows that don't operate nicely in the context of the current `truffle compile` solc pipeline. This compilation step can be greatly generalized by way of offering configuration for two components:
- Allow specifying arbitrary command to run, for compilation
- Allow specifying targets, with optional arbitrary command to process those targets into Truffle artifacts.

This PR makes changes to `truffle-workflow-compile` to support multiple compilers (at all), and defines a new package `truffle-external-compile`, which implements the above.

## Approach

In `truffle-workflow-compile`:

- Define `SUPPORTED_COMPILERS`, mapping name to node module

- Detect keys included in `config.compilers` (NOTE: solc is defaulted!)

- Run all compile modules for detected supported compilers

- Add `--compiler` option to the `truffle compile` command, to specify a particular compiler instead.

- Update truffle-workflow-compile's interface to return an object of the form:
  ```javascript
  {
    contracts, // aggregate of all contracts by name
    outputs: {
      compilerName: compilerSpecificResult
      /* for instance: */
      solc: paths
    }
  }
  ``` 
  This is instead of having multiple return values via `callback(err, contracts, paths)`.

New package `truffle-external-compile`:

- See [Configuration Example](https://github.com/trufflesuite/truffle/blob/d6b957ba9d7a7468de3559bf5e387e9b2e483b06/packages/truffle/test/sources/external_compile/truffle.js#L2-L22)

- Define method for processing compilation targets via arbitrary commands:

  1. Run `compilers.external.command`, expect some target output

  2. For each defined target, glob-expand `target.path` and pipe to `target.command` via stdin.

  3. Use stdout from each `target.command` as JSON description of artifact

- Define method for processing compilation targets via explicit property listing.

### re: globbing in first commit
The first commit here changes the way `truffle-contract-sources` works, so that compiler configuration can do things like:
```javascript
compilers: {
  solc: {
    sources: "**/*.sol",
    /* ... */
  }
}
```

## Dependent on `truffle-workflow-compile`'s external interface

These have all been updated:

- `truffle-core/lib/test` - resolves `contracts` and `paths` in `compileContractsWithTestFilesIfNeeded`
- `truffle-debugger/test/helpers` - resolves `contracts` and `paths` in `compile`
- `truffle-core/test/compile` - resolves `contracts` in three tests
- `truffle-core/test/migrate` - resolves `contracts` in one test
- `truffle-core/test/ethpm` - resolves `contracts` in one test
- `truffle-core/test/npm` - resolves `contracts` in one test

## Known Limitations

- Solidity tests currently don't work properly when using `truffle compile` as external

### DIFF
--- a/packages/truffle-external-compile/index.js
+++ b/packages/truffle-external-compile/index.js
@@ -42,18 +42,20 @@ const runCommand = promisify(function (command, options, callback) {
 });
 
 function decodeContents (contents) {
-  // HACK to test if contents has binary data
-  // from https://stackoverflow.com/a/49773659
-  if (/\ufffd/.test(contents.toString())) {
-    return web3.utils.bytesToHex(contents);
-  }
+  // accepts JSON, a hex string, or raw binary
 
+  // JSON
   try {
     return JSON.parse(contents);
-  } catch (e) {
-    return contents;
+  } catch (e) { /* no-op */ }
+
+  // hex string
+  if (contents.toString().startsWith("0x")) {
+    return contents.toString();
   }
 
+  // raw binary
+  return web3.utils.bytesToHex(contents);
 }
 
 async function processTargets (targets, cwd) {
--- a/packages/truffle-external-compile/index.js
+++ b/packages/truffle-external-compile/index.js
@@ -8,6 +8,12 @@ const fs = require("fs");
 const expect = require("truffle-expect");
 const debug = require("debug")("external-compile");
 
+const DEFAULT_ABI = [{
+  payable: true,
+  stateMutability: "payable",
+  type: "fallback"
+}];
+
 const runCommand = promisify(function (command, options, callback) {
   const { cwd, logger, input } = options;
   const child = exec(command, { cwd, input });
@@ -119,6 +125,11 @@ async function processTarget (target, cwd) {
     if (!contract.contractName) {
       throw new Error("External compilation target must specify contractName");
     }
+
+    if (!contract.abi) {
+      contract.abi = DEFAULT_ABI;
+    }
+
     return { [contract.contractName]: contract };
   }
 }
@@ -148,7 +159,12 @@ const compile = callbackify(async function(options) {
   return await processTargets(targets, cwd);
 });
 
+// required public interface
 compile.all = compile;
 compile.necessary = compile;
+
+// specific exports
+compile.DEFAULT_ABI = DEFAULT_ABI;
+compile.processTarget = processTarget;
 
 module.exports = compile;
--- a/packages/truffle-external-compile/index.js
+++ b/packages/truffle-external-compile/index.js
@@ -1,12 +1,14 @@
 'use strict';
 
+const debug = require("debug")("external-compile");
 const { exec, execSync } = require('child_process');
 const resolve = require("path").resolve;
 const { callbackify, promisify } = require("util");
 const glob = promisify(require("glob"));
 const fs = require("fs");
 const expect = require("truffle-expect");
-const debug = require("debug")("external-compile");
+const web3 = {};
+web3.utils = require("web3-utils");
 
 const DEFAULT_ABI = [{
   payable: true,
@@ -38,6 +40,21 @@ const runCommand = promisify(function (command, options, callback) {
     callback();
   });
 });
+
+function decodeContents (contents) {
+  // HACK to test if contents has binary data
+  // from https://stackoverflow.com/a/49773659
+  if (/\ufffd/.test(contents.toString())) {
+    return web3.utils.bytesToHex(contents);
+  }
+
+  try {
+    return JSON.parse(contents);
+  } catch (e) {
+    return contents;
+  }
+
+}
 
 async function processTargets (targets, cwd) {
   const contracts = {};
@@ -111,13 +128,8 @@ async function processTarget (target, cwd) {
     const contract = Object.assign({}, target.properties || {});
 
     for (let [key, path] of Object.entries(target.fileProperties || {})) {
-      const contents = fs.readFileSync(resolve(cwd, path)).toString();
-      let value;
-      try {
-        value = JSON.parse(contents);
-      } catch (e) {
-        value = contents;
-      }
+      const contents = fs.readFileSync(resolve(cwd, path));
+      const value = decodeContents(contents);
 
       contract[key] = value;
     }
--- a/packages/truffle-external-compile/index.js
+++ b/packages/truffle-external-compile/index.js
@@ -21,12 +21,12 @@ const runCommand = promisify(function (command, options, callback) {
   const child = exec(command, { cwd, input });
 
   child.stdout.on('data', function(data) {
-    data = data.toString().replace(/\n$/, '');
+    data = data.toString().replace(/(\r|\n)+$/, '');
     logger.log(data);
   });
 
   child.stderr.on('data', function(data) {
-    data = data.toString().replace(/\n$/, '');
+    data = data.toString().replace(/(\r|\n)+$/, '');
     logger.log(data);
   });
 
--- a/packages/truffle-external-compile/index.js
+++ b/packages/truffle-external-compile/index.js
@@ -1,0 +1,72 @@
+'use strict';
+
+const exec = require('child_process').exec;
+const path = require("path");
+const { callbackify, promisify } = require("util");
+const glob = promisify(require("glob"));
+const fs = require("fs");
+const expect = require("truffle-expect");
+const debug = require("debug")("external-compile");
+
+const runCompile = promisify(function (command, cwd, logger, callback) {
+  const child = exec(command, { cwd });
+
+  child.stdout.on('data', function(data) {
+    data = data.toString().replace(/\n$/, '');
+    logger.log(data);
+  });
+
+  child.stderr.on('data', function(data) {
+    data = data.toString().replace(/\n$/, '');
+    logger.log(data);
+  });
+
+  child.on('close', function(code) {
+    // If the command didn't exit properly, show the output and throw.
+    if (code !== 0) {
+      var err = new Error("Unknown exit code: " + code);
+      return callback(err);
+    }
+
+    callback();
+  });
+});
+
+const compile = callbackify(async function(options) {
+  if (options.logger == null) {
+    options.logger = console;
+  }
+
+  expect.options(options, [
+    "compilers",
+    "working_directory"
+  ]);
+  expect.options(options.compilers, ["external"]);
+  expect.options(options.compilers.external, [
+    "command",
+    "targets"
+  ]);
+
+  const { command, targets } = options.compilers.external;
+
+  debug("running compile command: %s", command);
+  await runCompile(command, options.working_directory, options.logger);
+
+  const contracts = {};
+  for (let target of targets) {
+    const pattern = path.join(options.working_directory, target.path);
+    for (let artifact of await glob( pattern, { follow: true })) {
+      debug("processing artifact: %s", artifact);
+
+      let contract = JSON.parse(fs.readFileSync(artifact));
+      contracts[contract.contractName] = contract;
+    }
+  }
+
+  return contracts;
+});
+
+compile.all = compile;
+compile.necessary = compile;
+
+module.exports = compile;
--- a/packages/truffle-external-compile/index.js
+++ b/packages/truffle-external-compile/index.js
@@ -57,15 +57,23 @@ const compile = callbackify(async function(options) {
 
   const contracts = {};
   for (let target of targets) {
-    const pattern = path.join(cwd, target.path);
+    expect.one(target, [ "path", "command" ]);  // also allows both
 
-    for (let preprocessed of await glob( pattern, { follow: true })) {
-      debug("processing target: %s", preprocessed);
-      const input = fs.readFileSync(preprocessed).toString();
+    if (target.path != undefined) {
+      const pattern = path.join(cwd, target.path);
 
-      const output = (target.command)
-        ? execSync(target.command, { cwd, input })
-        : input;
+      for (let preprocessed of await glob( pattern, { follow: true })) {
+        debug("processing target: %s", preprocessed);
+        const input = fs.readFileSync(preprocessed).toString();
+
+        const output = (target.command)
+          ? execSync(target.command, { cwd, input })
+          : input;
+        const contract = JSON.parse(output);
+        contracts[contract.contractName] = contract;
+      }
+    } else {
+      const output = execSync(target.command, { cwd });
       const contract = JSON.parse(output);
       contracts[contract.contractName] = contract;
     }
--- a/packages/truffle-external-compile/index.js
+++ b/packages/truffle-external-compile/index.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const exec = require('child_process').exec;
+const { exec, execSync } = require('child_process');
 const path = require("path");
 const { callbackify, promisify } = require("util");
 const glob = promisify(require("glob"));
@@ -8,8 +8,9 @@ const fs = require("fs");
 const expect = require("truffle-expect");
 const debug = require("debug")("external-compile");
 
-const runCompile = promisify(function (command, cwd, logger, callback) {
-  const child = exec(command, { cwd });
+const runCommand = promisify(function (command, options, callback) {
+  const { cwd, logger, input } = options;
+  const child = exec(command, { cwd, input });
 
   child.stdout.on('data', function(data) {
     data = data.toString().replace(/\n$/, '');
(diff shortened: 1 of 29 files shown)


gnidan:
@cgewecke Let me know if I should separate the first "globbing" commit into its own PR. It's not really used right now, so it's not really part of this scope.

packages/truffle-external-compile/index.js
```diff
@@ -0,0 +1,80 @@
+'use strict';
+
+const { exec, execSync } = require('child_process');
+const path = require("path");
+const { callbackify, promisify } = require("util");
```
cgewecke: Didn't know callbackify existed!

packages/truffle-external-compile/index.js
```diff
@@ -0,0 +1,80 @@
...
+      contracts[contract.contractName] = contract;
+    }
+  }
+
+  return contracts;
+});
+
+compile.all = compile;
+compile.necessary = compile;
+
```
cgewecke: 👍

packages/truffle-workflow-compile/index.js
```diff
@@ -19,69 +64,68 @@ var Contracts = {
...
+
+  writeContracts: async function(contracts, options) {
     var logger = options.logger || console;
 
-    mkdirp(options.contracts_build_directory, function(err, result) {
-      if (err != null) {
-        callback(err);
-        return;
-      }
+    const result = await promisify(mkdirp)(options.contracts_build_directory);
```
cgewecke: Nice.

packages/truffle/test/scenarios/external_compilers/truffle-compile.js
```diff
@@ -0,0 +1,99 @@
...
+  it("will run tests", function(done) {
+    this.timeout(70000);
+    CommandRunner.run("test", config, function(err) {
+      var output = logger.contents();
+      if (err) {
+        console.log(output);
+        return done(err);
+      }
+
+      assert(output.indexOf("3 passing") >= 0);
```
cgewecke: 👍

packages/truffle/test/sources/external_compile/truffle.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  compilers: {
+    external: {
+      command: "truffle compile " +
+        "--compiler=solc " +
+        "--contracts_build_directory=external",
+      targets: [{
+        path: "external/*.json",
+        command: "cat"
```
cgewecke: What does `command` do? Is it a post-processing step?

packages/truffle/scripts/test.sh
```diff
@@ -3,7 +3,7 @@
 set -o errexit
 
 if [ "$GETH" == true ]; then
-  npm run build-cli && mocha --timeout 50000 --grep @ganache --invert --colors
+  npm run build-cli && mocha --timeout 50000 --grep @ganache --invert --colors $@
 else
-  npm run build-cli && mocha --no-warnings --timeout 7000 --grep @geth --invert --colors
+  npm run build-cli && mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
```
cgewecke: What's being added here?

packages/truffle/test/scenarios/external_compilers/truffle-compile.js
```diff
@@ -0,0 +1,99 @@
...
+      config.mocha = {
+        reporter: new Reporter(logger)
+      }
+    });
+  });
+
+  it("will compile", function(done) {
+    this.timeout(20000);
+
+    CommandRunner.run("compile --compiler=external", config, function(err) {
```
cgewecke: I like this test. Without this flag compile will run both solc compilation *and* whatever is specified by external?

packages/truffle/test/scenarios/external_compilers/truffle-compile.js
```diff
@@ -0,0 +1,99 @@
...
+      var output = logger.contents();
+      if (err) {
+        console.log(output);
+        return done(err);
+      }
+
+      assert(fs.existsSync(path.join(config.contracts_build_directory, "MetaCoin.json")));
+      assert(fs.existsSync(path.join(config.contracts_build_directory, "ConvertLib.json")));
```
cgewecke: The write here is to a folder named `external`? Can we check the path for that string?

packages/truffle/test/sources/external_compile/truffle.js
gnidan: Yep - takes whatever files specified by `path` and pipes them to the `command`. The command is expected to output a JSON artifact on stdout.

packages/truffle/scripts/test.sh
gnidan: It forwards arguments; it's just for doing `yarn test -b` so that the `-b` gets forwarded to mocha.

packages/truffle/test/scenarios/external_compilers/truffle-compile.js
gnidan: Correct

packages/truffle/test/scenarios/external_compilers/truffle-compile.js
gnidan: `truffle compile --compiler=external` calls `truffle compile --compiler=solc --contracts_build_directory=external`, which puts the contracts in `external/`, but the project is configured to still use `build/contracts/` for artifacts. So the external compiler takes what's in `external/`, `cat`s them, and saves them in `build/contracts/`

cgewecke: Ahhh ok thanks makes sense.

packages/truffle/test/sources/external_compile/truffle.js
cgewecke: Sweet.

gnidan:
@cgewecke agreed about all this. I'll add better testing when I get back; I felt rushed trying to add all these options in. It's messy code as a result :) Holding off merging, awaiting required extra love.

(Just published `truffle@external-compiler` for anyone who wants to try this out)

axic:
I have tried this now with my Rust environment:
```
module.exports = {
  // See <http://truffleframework.com/docs/advanced/configuration>
  // to customize your Truffle configuration!
  compilers: {
    external: {
      command: "cargo build --release --target wasm32-unknown-unknown",
      targets: [{
        properties: {
          contractName: "ecadd"
        },
        fileProperties: {
          bytecode: "target/wasm32-unknown-unknown/release/ewasm_precompile_ecadd.wasm"
        }
      }]
    }
  }
};
```

Good news is that the command gets invoked and I do get an artifact file.

One note: initially we've discussed that if the `abi` field is omitted  then it will consider it as a ABI with a single fallback function. That is not the case right now.

And also there don't seem to be a validation that the resulting artifact is correct, in my case it only contains `contractName`, `bytecode`, `networks`, `schemaVersion` and `updatedAt` fields.

There seems to be one issue though: the bytecode kind of looks like weirdly escaped.
```
  "bytecode": "0x\u0000asm\u0001\u0000\u0000\u0000\u0001�\u0003@`\u0003^?^?^?\u0000`\u0002^?^?\u0001^?`\u0000\u0000`\u0003^?^?^?\u0001^?`\u0000\u0001^?`\u0001~\u0000`\u0004^?^?^?^?\u0000`\u0001^?\u0000`\u0001^?\u0001^?`\u0004^?^?^?^?\u0001^?`\u0002^?^?\u0000`\u0004^?^?^?~\u0000`\u0005^?^?^?^?^?\u0001^?``
```

It has unicode escapes and unprintable characters. The input is raw binary.

axic:
Without having set an alternative RPC, trying to deploy it via ganache of course fails, but there's no useful error message:
```
$ truffle test
Using network 'test'.

    Finished release [optimized] target(s) in 0.09s
Error
    at /Users/alex/.nvm/versions/node/v10.5.0/lib/node_modules/truffle/build/webpack:/packages/truffle-deployer/src/deployment.js:354:1
```

Another tiny issue is that the output from the compile command seems to be weirdly cut up at places. It may be down to the fact that it is coloured (the colours are also removed via truffle):
```
$ truffle compile
   Compiling semver-parser v0.7.0
   Compiling serde v1.0.70
   Compiling heapsize v0.4.2
   Compiling crunchy v0.1.6
   Compiling num-traits v0.2.5
   Compiling num-integer v
0.1.39
   Compiling crunchy v0.2.1
   Compiling rand v0.4.2
   Compiling 
rustc-hex v1.0.0
   Compiling byteorder v1.2.3
   Compiling num-bigint v
0.2.0
   Compiling parity-bytes v0.1.0 (https://github.com/paritytech/parity-common#4481687a)
   Compiling semver v0.9.0
   Compiling tiny-keccak v1.4.2
   Compiling rustc_version
 v0.2.2
   Compiling fixed-hash
 v0.2.1
   Compiling bn v0.4.4 (https://github.com/paritytech/bn#786c0d56)
   Compiling uint v
0.2.1
   Compiling ethereum-types v
0.3.2
   Compiling ewasm_api v
0.1.0 (https://github.com/ewasm/ewasm-rust-api#17b4da5c)
   Compiling ethereum-types-serialize v
0.2.1
   Compiling ethbloom v0.5.0
   Compiling ethereum-bn128
 v0.0.1 (https://github.com/ewasm/ethereum-bn128.rs#4eaba2ce)
   Compiling ewasm-precompile-ecadd v
0.1.0 (file:///Users/alex/Projects/ewasm.rs/ewasm-precompiles/ecadd)
    Finished 
release [optimized] target(s) in 1m 5.26s
Writing artifacts to ./build/contracts
```

axic:
Here is my working copy: https://github.com/ewasm/ewasm-precompiles/tree/ecadd-truffle-tests/ecadd

gnidan:
Ah thanks for trying it out @axic.

When I get back, I'll:
- fix ABI to have that fallback-only default
- fix bytecode interpretation (I didn't realize the output was actual binary, plain text, silly me!)
- sort out those weird extra newlines

Hopefully you can work around these in the meantime and get a bit further along.

packages/truffle-external-compile/index.js
```diff
@@ -57,15 +57,23 @@ const compile = callbackify(async function(options) {
...
-    for (let preprocessed of await glob( pattern, { follow: true })) {
-      debug("processing target: %s", preprocessed);
-      const input = fs.readFileSync(preprocessed).toString();
-
-      const output = (target.command)
-        ? execSync(target.command, { cwd, input })
-        : input;
+    expect.one(target, [ "path", "command" ]);  // also allows both
+
+    if (target.path != undefined) {
```
jklepatch: this is safer to test for undefined: `typeof target.path !== "undefined")`

gnidan: Yep, except this is on purpose to allow `null` as a valid non-value :)

packages/truffle-external-compile/index.js
```diff
@@ -0,0 +1,154 @@
...
+    let targets = paths.map(path => Object.assign({}, target, { path }));
+    return await processTargets(targets, cwd);
+  }
+
+  if (usesProperties) {
+    // contract properties listed individually
+    const contract = Object.assign({}, target.properties || {});
+
+    for (let [key, path] of Object.entries(target.fileProperties || {})) {
+      const contents = fs.readFileSync(resolve(cwd, path)).toString();
```
lrettig: If you change `toString()` to `toString('hex')` here, it (at least partially) fixes the bytecode issue.

lrettig: Here is a slightly more sophisticated approach: https://github.com/lrettig/truffle/commit/acfc8b35e46b89554df1689d936a7829d068c5e5

lrettig:
Another thought: if we want to be thorough about adding support for Ewasm, then instead of using evm properties like `bytecode`, `sourceMap`, etc. it probably makes more sense to use the Wasm properties laid out here: https://github.com/ethereum/solidity/commit/4b5639bf63f2c86604cb55387324e00759a0b1f2#diff-d7b1fb9ae448539124dcfd3d09ad6e60R124 e.g. `wasm`, `wast`, etc. and maybe add these to the Truffle contract schema under `wasm`.

gnidan:
@lrettig that's a good point, but out of scope for this effort.

Those properties are the properties that truffle-contract-schema uses, which can be considered coincidentally the same as the Solidity ones. This rename should probably be considered when we get to implementing an overhaul of Truffle's artifacts format.

Would you mind opening an issue about that here? https://github.com/trufflesuite/artifact-updates

Thanks!

packages/truffle-external-compile/index.js
axic: I don't think that is a good idea. Why make a distinction for wasm?

axic:
@lrettig @gnidan note that isn't fully specified in Solidity, because ewasm isn't fully specified in that regard (whether it would have a deployer code with a constructor or not).

packages/truffle-external-compile/index.js
```diff
@@ -39,6 +41,21 @@ const runCommand = promisify(function (command, options, callback) {
   });
 });
 
+function decodeContents (contents) {
+  // HACK to test if contents has binary data
+  // from https://stackoverflow.com/a/49773659
+  if (/\ufffd/.test(contents.toString())) {
```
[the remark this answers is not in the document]
gnidan: Hm, so I was thinking that `contents.toString()` was "correctly" rendering those unreadable characters. Is that not the case in all situations?

Do you have any other ideas as to how to detect this, @axic? I'm trying to avoid making these files parsed in a property-specific way, and also trying to avoid requiring the user to specify the encoding on a per-property basis.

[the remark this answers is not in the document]
axic: I would suggest that if the file starts with `0x` (and passes the hex test for the rest, e.g. has only hex characters) then read it as a hex string otherwise consider it a binary.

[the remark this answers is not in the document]
gnidan: Mmm. OK so how about:

- If it parses as JSON, it's JSON
- If it starts with `0x`, it's hex
- Otherwise, it's binary

axic:
To summarise my initial reported issues:
- [x] optional abi field (falls back to a default one)
- [x] binary bytecode is handled wrongly
- [ ] compile command output is mangled
- [ ] validate that the resulting artifact file has all the necessary fields
- [ ] no real error message if deployment fails

Should I open issues for some of these or is it planned to be handled in this PR?

gnidan:
@axic I'm going to include those fixes in this PR.

Should have those resolved tomorrow and I'll put out a new build for you to try. Thanks!

axic:
Great, thank you!

gnidan:
@axic I just published `truffle@5.0.0-external-compiler.1`. I think I addressed all your concerns. Could you please try it out and let me know? Thanks!

N.B. when you install this version, you will likely need to uninstall first:
```
npm uninstall -g truffle
npm install -g truffle@5.0.0-external-compiler.1
```

axic:
Test works nicer:
```
$ truffle test
Could not connect to your Ethereum client. Please check that your Ethereum client:
    - is running
    - is accepting RPC connections (i.e., "--rpc" option is used in geth)
    - is accessible over the network
    - is properly configured in your Truffle configuration file (truffle.js)
```

The line mangling still appears:
```
$ truffle compile
   Compiling semver-parser v0.
7.0
   Compiling serde v1.0.70
   Compiling crunchy v0.1.6
   Compiling heapsize v0.4.2
   Compiling num-traits v0.2.5
   Compiling rustc-hex v1
.0.0
   Compiling rand v
0.4.2
   Compiling num-integer
 v0.1.39
```

The default ABI is populated and the binary seems to be picked up properly. Nice!

gnidan:
@cgewecke any idea why the `deployFailed` event error isn't showing up nicely?

cgewecke:
@gnidan Ai it's a bug. I thought I fixed it in #1113 :/ 

Tracking down cause rn but it affects all tests so it shouldn't block this PR. Will open something shortly.

packages/truffle-external-compile/index.js
```diff
@@ -0,0 +1,279 @@
...
+      // reset buffer
+      buffer = [last];
+
+    } else {
+      // nothing to see yet
+      yield [];
+    }
+  }
+}
+
```
cgewecke: Nice

packages/truffle-external-compile/test/test_targets.js
```diff
@@ -0,0 +1,122 @@
...
+        }
+      };
+
+      const processed = await processTarget(target, cwd);
+
+      assert.equal(processed[contractName].bytecode, bytecode);
+    });
+
+    it("reads UTF-8 file properties", async () => {
+      const contractName = "My😀Contract";
```
cgewecke: lol!

cgewecke:
@gnidan Looks good!

axic:
> Output is now properly buffered and should appear without extraneous linebreaks :)

Confirmed, it looks nice!

axic:
Also regarding validation this is what I've got when I've removed the bytecode field from the config:
```
 Warning: contract ecadd does not specify bytecode. You won't be able to deploy it.
```

I guess validation works 👍

gnidan:
@axic that's some explicit validation, because technically `bytecode` isn't required as part of truffle-contract-schema. But the actual schema validation should work, too. (Try using a malformed ABI ;)

Anything else before I merge and call this a wrap for now?

Thanks for all the good feedback!

axic:
I haven't gotten around to test it with a native ewasm client with RPC (it will be cpp-ethereum/aleth) - I may have some more things to report then.

However that may not happen before Tuesday, so I'd say go ahead.

gnidan:
Great! I'll merge and get you a new `truffle@next` release to use for your testing.

PR #1160, opened 2018-08-01
Update contracts for compatibility with solidity 0.5.0

ekpyron:
This PR *almost* makes it possible to replace ``soljson.js`` with the current nightly build (that is to become solidity 0.5.0) while still having the truffle test suite pass.

What still needs to be adjusted is ``packages/truffle-compile/test/test_supplier.js``, resp. ``packages/truffle-compile/test/sources/NewPragma.sol``, but this is tested against different versions of solidity, so some more involved logic is needed here.

Apart from that ``packages/truffle/test/scenarios/contract_names/test_imports.js``, ``packages/truffle/test/scenarios/happy_path/happypath.js`` and ``packages/truffle/test/scenarios/solidity_testing/with_balances.js`` pull some solidity files (presumably from a truffle box) that still use outdated syntax, but I don't really know where those files are coming from (``https://github.com/trufflesuite/truffle-init-default`` seems to be up to date - maybe a fixed commit or another branch or an entirely different repo is used here?)

The changes are backwards compatible in the sense that the tests still pass without updating ``soljson.js``.

### DIFF
--- a/packages/truffle-contract/test/sources/Example.sol
+++ b/packages/truffle-contract/test/sources/Example.sol
@@ -10,7 +10,7 @@ contract Example {
   event ContractAddressEvent(address _contract);
   event SpecialEvent();
 
-  constructor(uint val) {
+  constructor(uint val) public {
     // Constructor revert
     require(val != 13);
     require(val != 2001, 'reasonstring');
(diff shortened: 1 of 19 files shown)


ekpyron:
@gnidan I finally got around to put this together.

cgewecke:
@ekpyron Thanks for this :)

I *think* @gnidan has actually moved the same project we consume from the truffle-box into our local test sources files in #1148 (still WIP but close). We can circle back and swap those out in a later step. 

I'll look into the tests at CompilerSupplier and see what we could do there....

gnidan:
Thanks @ekpyron!

packages/truffle-contract/test/sources/Example.sol
```diff
@@ -86,6 +86,10 @@ contract Example {
     emit ContractAddressEvent(address(this));
   }
 
+  function triggerNumberEvent(int a, int b, address c, uint d, uint e){
```
ekpyron: This needs ``public`` as well.

Issue #1123, opened 2018-07-19
Add `startStage`, `stopStage` events to migrations reporter 


cgewecke:
Idea is that within the migrations file you should be able to tell the reporter to start collecting transaction / gas data for arbitrary web3 calls (like the kind you might make while staging things around your deployments). `stopStage` would trigger the reporter to output accumulated costs & add them to the deployment totals.

```javascript
deployer.emit.startStage()
deployer.emit.endStage()
```

cgewecke:
This was implemented as `startTransaction`, `endTransaction` in `next.7`

PR #1166, opened 2018-08-02
Add startTransaction / endTransaction reporter hooks

cgewecke:
Targets #1123

These have a really simple implementation at the moment - they just start/stop a spinner with a message. They should be expanded to aggregate gas data.

+ Integrated them into the "Saving Migrations" step so that doesn't seem like it's hanging
+ Also: stopped printing `Saving artifacts` when script doesn't actually deploy any contracts. It's not uncommon for people to use a migration file to run setup and initialization logic separately from the actual deployments.

### DIFF
--- a/packages/truffle-migrate/migration.js
+++ b/packages/truffle-migrate/migration.js
@@ -94,7 +94,7 @@ class Migration {
       const Migrations = resolver.require("./Migrations.sol");
 
       if (Migrations && Migrations.isDeployed()) {
-        const message = `Saving migration (#${self.number}) to chain.`;
+        const message = `Saving migration to chain.`;
 
         if (!this.dryRun){
           const data = { message: message }
--- a/packages/truffle-migrate/migration.js
+++ b/packages/truffle-migrate/migration.js
@@ -94,9 +94,20 @@ class Migration {
       const Migrations = resolver.require("./Migrations.sol");
 
       if (Migrations && Migrations.isDeployed()) {
-        await self.emitter.emit('saveMigration', self.number);
+        const message = `Saving migration (#${self.number}) to chain.`;
+
+        if (!this.dryRun){
+          const data = { message: message }
+          await self.emitter.emit('startTransaction', data);
+        }
+
         const migrations = await Migrations.deployed();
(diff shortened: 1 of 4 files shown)


packages/truffle-migrate/migration.js
```diff
@@ -94,9 +94,20 @@ class Migration {
       const Migrations = resolver.require("./Migrations.sol");
 
       if (Migrations && Migrations.isDeployed()) {
-        await self.emitter.emit('saveMigration', self.number);
+        const message = `Saving migration to chain.`;
+
+        if (!this.dryRun){
+          const data = { message: message }
+          await self.emitter.emit('startTransaction', data);
+        }
```
gnidan: Why not during dry-run?

packages/truffle-reporters/reporters/migrations-V5/messages.js
```diff
@@ -250,9 +256,10 @@ class MigrationsMessages{
 
       postMigrate: () => {
         let output = '';
+        let deployments = reporter.summary[reporter.currentFileIndex].deployments;
 
-        if (!reporter.migration.dryRun)
-          output += `   * Saving artifacts\n`;
+        if (!reporter.migration.dryRun && deployments.length)
+          output += `   > Saving artifacts\n`;
```
gnidan: 👍

packages/truffle-migrate/migration.js
cgewecke: Hmmm maybe this should be removed. The dry-run has had abbreviated output and was just preserving here but now this logic is outside the reporter so maybe shouldn't do that. 

On the other hand it's all disposable, like the dry-run doesn't really pick up from a migration that left off or write artifacts that get preserved etc.

gnidan: Hm, then I think this is fine for now. I think eventually we will want a "dry run reporter" that is separate from the regular reporter. That way we can get rid of this conditional logic.

cgewecke: Ah yes nice! That's a cool idea.

Issue #1165, opened 2018-08-02
Reformat event numbers as BN


cgewecke:
Companion to #1145

cgewecke:
Fixed in `next.7`

PR #1167, opened 2018-08-03
Reformat event numbers as BN

cgewecke:
#1165 

Companion to #1145.  

+ Renamed the reformatter method so it's more general.
+ Added test for numerical event outputs

### DIFF
--- a/packages/truffle-contract/lib/reformat.js
+++ b/packages/truffle-contract/lib/reformat.js
@@ -38,7 +38,7 @@ const _convertNumberArray = function(arr, format){
  *   - array
  *   - single primitive
  * @param  {String|Object|Array} result      web3 call result
- * @param  {Array}               abiSegment  event inputs OR .call outputs
+ * @param  {Array}               abiSegment  event params OR .call outputs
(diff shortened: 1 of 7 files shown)


packages/truffle-contract/lib/reformat.js
```diff
@@ -37,14 +37,14 @@ const _convertNumberArray = function(arr, format){
  *   - object (with index keys and optionally, named keys)
  *   - array
  *   - single primitive
- * @param  {String|Object|Array} result  web3 call result
- * @param  {Array}               outputs ABI outputs array
+ * @param  {String|Object|Array} result      web3 call result
+ * @param  {Array}               abiSegment  event inputs OR .call outputs
```
gnidan: I think these are called `parameters` in the ABI spec

cgewecke: Why are they so normie over there?

gnidan: Ugh, normies!

cgewecke:
@gnidan Ah! Ok. Well now it's just total muddle.

PR #1169, opened 2018-08-03
Inherit synchronization_timeout=0 during deployments

naure:
`synchronization_timeout` is set on the base class, but is not inherited by instances from `new()` and `at()`.
Therefore, later calls remain subject to the low default timeout.

Fix using the prototype inheritance of JavaScript.

### DIFF
--- a/packages/truffle-migrate/resolverintercept.js
+++ b/packages/truffle-migrate/resolverintercept.js
@@ -24,7 +24,7 @@ ResolverIntercept.prototype.require = function(import_path) {
   // During migrations, we could be on a network that takes a long time to accept
   // transactions (i.e., contract deployment close to block size). Because successful
   // migration is more important than wait time in those cases, we'll synchronize "forever".
-  resolved.synchronization_timeout = 0;
+  resolved.prototype.synchronization_timeout = 0;
 
   return resolved;
 };


cgewecke:
@naure I *think* the current `truffle-contract` [code](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-contract/contract.js#L163-L164) works as expected? It looks at the value set on the constructor rather than the instance. Additionally this this parameter has limited use - the real timeout constraint is 50 blocks and is set downstream from `truffle-contract` at the web3 dependency. 

Please correct me if I'm wrong about this.

We're also quite close to publishing a `beta` of Truffle V5 which now exposes a `timeoutBlocks` setting on the artifact that allows you to override the web3 limit (see PR #1129 for usage notes). It can also be set on a per network basis in the config for migrations scripts.

```javascript
ropsten: {
  provider: () => new HDWalletProvider(mnemonic, `https://ropsten.infura.io/${infuraKey}`),
  network_id: 3,       // Ropsten's id
  gas: 5500000,        // Ropsten has a lower block limit than mainnet
  confirmations: 2,    // # of confs to wait between deployments. (default: 0)
  timeoutBlocks: 200,  // # of blocks before a deployment times out  (minimum/default: 50)
},
```

A prerelease of the beta is available by running:
```
npm install truffle@next
```

naure:
Indeed it's using `constructor` instead.

Looking forward to the timeoutBlocks limit too.

PR #1174, opened 2018-08-09
Make debugger fail less hard

gnidan:
Might fix some problems in trufflesuite/truffle-debugger#75

- Continue stepping through Solidity code if `controller.current.location` causes an error
- Use more defaults to prevent unhandled "cannot read whatever of undefined" exceptions

### DIFF
--- a/packages/truffle-debugger/lib/controller/sagas/index.js
+++ b/packages/truffle-debugger/lib/controller/sagas/index.js
@@ -69,10 +69,15 @@ function* stepNext () {
     yield* advance();
 
     // and check the next source range
-    upcoming = yield select(controller.current.location);
+    try {
+      upcoming = yield select(controller.current.location);
+    } catch (e) {
+      upcoming = null;
+    }
 
     // if the next step's source range is still the same, keep going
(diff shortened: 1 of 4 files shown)


gnidan:
@seesemichaelj Adding you for review just in case you're interested in getting your eyes on the truffle-debugger. Feel free to disregard.

Thoughts: definitely need to come up with a better strategy for dealing with the debugger when it's in the state of "I don't know where I am!!!" This class of failure case has been entirely disregarded and that needs to be rectified in a more substantial way, instead of these `if (!something) return whatever;` that just shift the unknown down the line. One idea would involve a conversion to TypeScript, then making use of a `Maybe<T>` or `Option<T>` type.

gnidan:
Note: using the reproduction steps in https://github.com/uport-project/ethr-did-registry/issues/9, I was able to successfully step through the debugger without a single exception (using `;` for highest granularity).

gnidan:
Thanks @cgewecke!

@seesemichaelj I am merging this, just wanted you to have the notification should you be interested.

PR #1178, opened 2018-08-09
Throw real Error instead of list for schema errors

gnidan:
Currently, truffle-contract-schema's `validate()` function throws a list of objects representing validation errors.

This PR:
- Converts that exception to a real `Error` object, with lightly formatted message text.
- Appends the original `errors` list as a property on the Error object
- Removes the `useDefaults` option from the validation, since it mutates the input object. This was causing problems with the polymorphic ABI item schema.
- Use `verbose` when validating, to ensure that the errors contain the original data value.


gnidan:
Danke!

PR #1190, opened 2018-08-14
Upgrade to web3 beta.35

cgewecke:
+ Upgrades web3 to beta.35
+ Adds code to extract receipt from new error thrown by web3 and route it to truffle-contract's receipt handler where it get's rejected per the previous pattern. 
+ Check status values as booleans.
+ Updates tests.

### DIFF
--- a/packages/truffle-contract/lib/override.js
+++ b/packages/truffle-contract/lib/override.js
@@ -1,10 +1,31 @@
 var Reason = require('./reason');
+var handlers = require('./handlers');
 
 var override = {
 
   timeoutMessage: 'not mined within', // Substring of timeout err fired by web3
   defaultMaxBlocks: 50,               // Max # of blocks web3 will wait for a tx
   pollingInterval: 1000,
+
+  /**
+   * Attempts to extract receipt object from Web3 error message
+   * @param  {Object} message       web3 error
+   * @return {Object|undefined} receipt
+   */
+  extractReceipt(message){
+    const hasReceipt = message &&
+                       message.includes('{')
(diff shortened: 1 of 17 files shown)


packages/truffle-contract/lib/override.js
```diff
@@ -1,11 +1,32 @@
...
+
+    if (hasReceipt){
+      const receiptString =  '{' + message.split('{')[1].trim();
+      try {
+        return JSON.parse(receiptString)
+      } catch (err){
+        // ignore
+      }
+    }
+  },
```
gnidan: What was the old behavior? Should we make this backwards compatible to beta.33?

cgewecke: The old web3 behavior was: no error is thrown from a conventional client returning a receipt with status zero. Our strategy was to inspect the receipt in the receipt event handler and throw. 

The new web3 behavior is: an error *is* thrown from a conventional client that returns a bad receipt. Web3 stringifies the receipt into the error message. Our strategy is to catch the error, extract the receipt from the string and route it to the receipt handler as before. 

So our behavior around this hasn't really changed - our tests remain the same, we emit a receipt, etc. 

Is this what you're wondering about? In what context would we need to be backward compatible with `.33`?

This PR does contain a (potentially) breaking change because of `.34`'s switch from hex to bool status results. But we throw an error anyway so it's relatively unlikely people are inspecting that field themselves.

cgewecke: Merging but please feel free weigh in further...

PR #1192, opened 2018-08-14
Add stuff to publish truffle-contract@next 

cgewecke:
Josh asked about getting a version of `truffle-contract@next` out for the boxes. This PR updates package to:

+ build ES6 correctly
+ build on publish / install
+ ignore `dist/` in the git (and remove current dist - hopefully no one is depending on this but they can always pull from an earlier commit I guess).


cgewecke:
@gnidan I guess manually only because it's possible that there are next changes that don't touch tc?

Issue #500, opened 2017-07-16
Testing Framework


0xcaff:
Will truffle ever allow testing contracts in another testing framework or use a different assertion library with `truffle test`? I really like `jest`'s snapshot testing and it's BDD style assertions. When writing contract tests, I often find myself setuping up contract state, running some oprations and comparing the expected state to the actual state. This seems like the perfect job for snapshot testing.

I know this can be accomplished using `truffle-contract` and writing tests without using truffle directly but then the `artifact` global along with the standard `truffle test` convention is lost.

adrianmcli:
I also think this is worth considering. Perhaps we could make it possible to plug-in your own testing library in the future?

code-chimp:
Would it be possible to at least enable Chai's [BDD style assertions](http://chaijs.com/api/bdd/)?

EdsonAlcala:
I have seem Jest and I like the Immersive Watch Mode approach, is there a way to do that in truffle or mocha?

Mischi:
We just released a first version of https://github.com/AsureFoundation/truffle-jest. It allows you to run your truffle tests with *Jest* instead of `truffle test`. Give it a try if you like. This is mostly a PoC and has some rough edges 😉

We played with *Mocha / Mocha CLI* too but in the end *Jest* had better support for "Async Initialization before executing tests" as *Mocha*. 

PS: I just stumbled upon `truffle-solidity-loader` and wondering if it can be used together with *Jest* to rerun the tests once  the smart contracts changed ...

Cheers,
Fabian

Mischi:
We've published a blog post about `truffle-jest`. Take a look if you like 😃 

https://medium.com/@ASURE.io/running-truffle-unit-tests-with-jest-f59b5049b66a

Issue #403, opened 2017-05-12
Explore better (more standard) integration with Mocha


tcoulter:
See this account of integrating IntelliJ Idea: https://charterhouse.github.io/2017/05/12/running-truffle-unit-tests-in-intellij-idea.html

cedricwalter:
@tcoulter 
I've tried to follow the tutorial and it does not work (Intellij 2017.3). I would be happy to help on this one (Intellij) even helping writing a Karma-runner for truffle. What do you think?

gnidan:
Closing this as wont-fix. I think there is more investigation to be done on generalizing/improving Truffle's test system. Will hopefully find a general solution meet this kind of need. Thanks!

thomasmodeneis:
anyone knows if this was ever made better ?

PR #1197, opened 2018-08-18
use web3 instance of TestRunner instead of global

Mischi:
Hi

while trying to come up with a possible solution for #500, and #403 I noticed that the TestRunner relies on the global web3 instance which is exposed to the unit tests. I think it would be better to reuse the web3 instance of the TestRunner instance.

Cheers,
Fabian


Mischi:
Thanks 😃

PR #1219, opened 2018-08-23
Add "public" to Example.sol constructor

matt9ucci:
To suppress the following warning:
```
C:\Users\username\projectname\contracts\MyContract.sol:5:3: Warning: No visibility specified.
Defaulting to "public".
  constructor() {
  ^ (Relevant source part starts here and spans across multiple lines).
```


gnidan:
Thanks @matt9ucci!

PR #1223, opened 2018-08-23
don't try to set the from address

Mischi:
Hi.

The `config.from` property is readonly since some time. 
The from address will either be set through the network configuration or 
will be loaded by Environment.detect.

I think we can remove this statement here.

Cheers, 
Fabian


Mischi:
👍

PR #1227, opened 2018-08-24
Update package.js for solc contract json changes

jeichel-miovision:
The newer solc version produces a different JSON output with contractName and bytecode fields.


gnidan:
Thanks for this @jeichel-miovision!

PR #1242, opened 2018-08-29
Add support to generate types file

seesemichaelj:
This PR add's an `index.d.ts` using an npm `prepare` script leveraging `json-schema-to-typescript` module.

For the `index.d.ts` file to be correct, some changes were necessary to the `spec.json` files.

If you'd rather the typing file not be version controlled, I can be up for that as well.


packages/truffle-contract-schema/index.d.ts
```diff
@@ -0,0 +1,177 @@
...
+            indexed: boolean;
+            [k: string]: any;
+          }[];
+          anonymous: boolean;
+        };
+      };
+      [k: string]: any;
+    };
+  };
+}
```
gnidan: Can we put this file in `.gitignore` instead?

It's not necessary to commit because `prepare` runs even if you're doing something weird with yarn workspaces or whatever.

gnidan: Ah I missed your original comment. Yep, please leave out!

packages/truffle-contract-schema/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "JSON schema for contract artifacts",
   "main": "index.js",
   "scripts": {
+    "prepare": "cd spec && json2ts -i contract-object.spec.json -o ../index.d.ts",
```
gnidan: Was the `cd` necessary to get the cross-file references working?

seesemichaelj: Yes, per https://github.com/bcherny/json-schema-to-typescript#options there is an option to specify `CWD` through the node module, but the CLI doesn't have that option, and it defaults to `process.cwd()`

cgewecke:
I'm completely out of the loop on what this is and why.

seesemichaelj:
@gnidan I have made the recommended changes

Issue #1230, opened 2018-08-25
Where is `getData` or `encodeABI()` in  truffle v5


rstormsf:
```js
const receiver = await ERC677Recipient.new();
console.log(receiver.doSomething)
{ [Function]
  call: [Function],
  sendTransaction: [Function],
  estimateGas: [Function],
  request: [Function] }
```

I'd expect to  have `encodeABI()` function from web3 1.0

I had to use workaround:
```js
const ERC677RecipientInstance = new web3.eth.Contract(ERC677Recipient.abi, receiver.address);
var callDoSomething123 = await ERC677RecipientInstance.methods.doSomething(123).encodeABI();
```

cgewecke:
It should be [available](https://github.com/trufflesuite/truffle/blob/next/packages/truffle-contract/lib/contract.js#L23-L34) via the `contract` object:
```javascript 
receiver.contract  // <--- The underlying web3 instance. 
receiver.contract.methods.doSomething(123).encodeABI()
```

gnidan:
Hi @rstormsf, can this be closed? Thanks!

rstormsf:
Ok, perfect! thanks a lot

Issue #1229, opened 2018-08-25
Why  bytes data returns null in v5-beta? 


rstormsf:
in truffle v4 test would look like this:
solidity:
```
bytes public data;
```

```js
        (await erc677r.data()).should.be.equal('0x');
```

now in truffle v5, it returns `null`, not `0x`

rstormsf:
Second issue is that events in ganache-cli does not emit if it was called by  
```
_to.call(abi.encodeWithSignature("onTokenTransfer(address,uint256,bytes)",  msg.sender, _value, _data));
```

cgewecke:
Hi @rstormsf 

Yes - just saw the same decoding problem with `boolean` [here](https://github.com/trufflesuite/truffle/pull/1200) last week. Will investigate some more to see if it's on this side but I *think* it's at `web3-eth-abi` and might be resolved by a pending release over there. 

Will also look at problem #2. . . you're note suggests you think it's at ganache-cli - are you seeing this fire if you use `geth --dev` or do you mean that Truffle isn't detecting the event?

PR #1255, opened 2018-09-05
Upgrade web3 to beta.36

cgewecke:
Upgrade web3 to beta.36 which has several very useful [features/improvements](https://github.com/ethereum/web3.js/releases/tag/v1.0.0-beta.36). 

+ [ethers](https://github.com/ethers-io/ethers.js) is now being used to encode/decode params. PR includes some logic to convert incoming BN / BigNumbers to the ethers big number format because it looks like they're not compatible. 
+ adds some tests for the foregoing as well as a (skipped) regression test for #1229.

This needs to be run E2E at wild-truffle before publication to see if there are other decoding problems.

[EDIT - I'm seeing crashes on ~80% of Colony's tests because they [reach into](https://github.com/JoinColony/colonyNetwork/blob/develop/helpers/test-data-generator.js#L31) the underlying web3 contract to do their own `encodeABI` in some cases. This is really common (#1230 for example). Have opened an [issue](https://github.com/ethereum/web3.js/issues/1920) at web3. We might have to close this and wait for a fix...)

### DIFF
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -21,7 +21,7 @@
     "solc": "0.4.24",
     "temp": "^0.8.3",
     "truffle-blockchain-utils": "^0.0.5",
-    "web3": "1.0.0-beta.35"
+    "web3": "1.0.0-beta.36"
   },
   "dependencies": {
     "async": "2.6.1",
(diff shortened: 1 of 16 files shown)


Amxx:
+1

davy42:
+

ethereummayster:
Can anyone merge?

I need to be able to install truffle with web3 beta.36 via npm install -g command without any other actions...

Amxx:
This should be closed since the features it proposed were merged from PR 1409
https://github.com/trufflesuite/truffle/pull/1409

gnidan:
@Amxx you're correct!

Issue #1224, opened 2018-08-23
Better help on command line


eduardonunesp:
## Issue

Currently the terminal still a good source of help when the user needs to get a depth of a command, I wish that too for `truffle`, the main help is `truffle --help` command, but there's no more than that.

## Steps to Reproduce

`truffle --help`

## Expected Behavior

When the user wants to know more options available about a command the user can get help like `truffle help deploy` which can describe the options. Example

```bash
$ truffle help deploy
Truffle v5.0.0-next.10 - a development framework for Ethereum

 --reset         Restart the migrations from the beginning
 --network       Selects the network to be used on the deploy process
 --interactive   After run the simulation asks for the user if the deploy can continue
```

## Actual Results

Always is the same list of help without sub commands help, there's a link to the online documentation which helps a little, however the help on terminal should be a quick reference instead an online page.

```bash
Truffle v5.0.0-next.10 - a development framework for Ethereum

Usage: truffle <command> [options]

Commands:
  init      Initialize new and empty Ethereum project
  compile   Compile contract source files
  migrate   Run migrations to deploy contracts
  deploy    (alias for migrate)
  test      Run JavaScript and Solidity tests
  debug     Interactively debug any transaction on the blockchain (experimental)
  opcode    Print the compiled opcodes for a given contract
  console   Run a console with contract abstractions and commands available
  develop   Open a console with a local development blockchain
  create    Helper to create new contracts, migrations and tests
  install   Install a package from the Ethereum Package Registry
  publish   Publish a package to the Ethereum Package Registry
  networks  Show addresses for deployed contracts on each network
  exec      Execute a JS module within this Truffle environment
  unbox     Download a Truffle Box, a pre-built Truffle project
  version   Show version number and exit

See more at http://truffleframework.com/docs
```

## Environment

* Operating System: 10.13.6
* Truffle version (`truffle version`): Truffle v5.0.0-next.10
* node version (`node --version`): 8.11.4
* npm version (`npm --version`):  6.2.0

cgewecke:
Yes this is a really good idea.

PR #1262, opened 2018-09-10
Add help command

eggplantzzz:
Adds a userHelp property to each command file. This contains information related to displaying its usage and parameters.

~~This changes the previous behavior of typing `truffle help` without arguments.  Typing the help command by itself now displays the user help for the help command itself as opposed to the more general displaying of all commands and their descriptions.~~

Issue #1224 prompted this change.


packages/truffle-core/lib/commands/help.js
```diff
@@ -0,0 +1,41 @@
...
+  displayHelpInformation: function (selectedCommand) {
+    var commands = require("./index");
+    var commandHelp = commands[selectedCommand].userHelp;
+    console.log(`\n  COMMAND NAME: ${commands[selectedCommand].command}`);
+    console.log(`  DESCRIPTION:  ${commands[selectedCommand].description}`);
+    console.log(`  USAGE:        ${commandHelp.usage}`);
+
+    if (commandHelp.parameters.length > 0) {
+      console.log(`  PARAMETERS: `);
+      commandHelp.parameters.forEach((parameter) => console.log(`                ${parameter.parameter}: ${parameter.description}`));
```
cgewecke: Could we use a variable for the large space here? The line runs outside the github window.

eggplantzzz: Haha yeah, sounds good to me

packages/truffle-core/lib/commands/index.js
```diff
@@ -9,6 +9,7 @@ module.exports = {
   console: require("./console"),
   develop: require("./develop"),
   create: require("./create"),
+  help: require("./help"),
   install: require("./install"),
   publish: require("./publish"),
   //digest: require("./digest"),
```
eggplantzzz: Also, can we remove this line and the corresponding file from the project?  Tim said it was related to EthPM but did not know if it was relevant with EthPM v2.

gnidan: Yep, totally remove this!

eggplantzzz: Feels good

eggplantzzz:
For this, it now seems like there are two notions of running 'help'.  There is one that gives general help, in other words it displays all the truffle commands and a brief description of what each does.  When an argument is provided it displays more in-depth information about the command provided as an argument (i.e. `truffle help compile`).  Do we want to make these two accessed in different ways like `--help` and `help <command>`?

packages/truffle-core/lib/commands/console.js
```diff
@@ -2,6 +2,18 @@ var command = {
   command: 'console',
   description: 'Run a console with contract abstractions and commands available',
   builder: {},
+  userHelp: {
```
gnidan: How do you feel about just calling this `help` instead of `userHelp`?

eggplantzzz: Sure, I originally did this because there was a conflict (there is already a property called 'help' in  `compile.js` for example) but now it looks like this work makes the old property obsolete.  Changing this sounds good to me.

cds-amal:
Have you considered implementing something like what [docopt](http://docopt.org/) does? It's a tool that generates a parser from doc string. One advantage is that the help and definition of the commands are maintained in one place.

The author of the package describes the motivation for this pattern/tool [here](https://youtu.be/pXhcPJK5cMc)

Their example shows how multiple commands can be described in one doc string
```
Naval Fate.

Usage:
  naval_fate ship new <name>...
  naval_fate ship <name> move <x> <y> [--speed=<kn>]
  naval_fate ship shoot <x> <y>
  naval_fate mine (set|remove) <x> <y> [--moored|--drifting]
  naval_fate -h | --help
  naval_fate --version

Options:
  -h --help     Show this screen.
  --version     Show version.
  --speed=<kn>  Speed in knots [default: 10].
  --moored      Moored (anchored) mine.
  --drifting    Drifting mine.
```
There is a [Node](https://github.com/felixSchl/neodoc/blob/development/README.md) port of docopt as well!

eggplantzzz:
Thanks for the suggestion @cds-amal, I'll look into it.

eduardonunesp:
That was super cool! thanks for the great PR ! @eggplantzzz

PR #1265, opened 2018-09-11
Secret Decoder Ring

seesemichaelj:
**@seesemichaelj's notes:**

This PR is currently a WIP, and more details will be added later. However, in the meantime, the gist is below:
- Add a module/library for easy-to-use decoding for TruffleContract object's
  - Get the contract state variables at a certain block
  - Get all of the occurrences of an event type during a certain block, including the decoded arguments
  - Subscribe to an event type being emitted
  - Decode a transaction to a contract (function and arguments)
- Add capability for decoding variables given an EVM state (stack, memory, storage, etc.) to allow the decoding functionality to be removed from `truffle-debugger`
- Use TypeScript to try to type as much as possible

---

**@gnidan addendum:**

- Besides truffle-decoder and truffle-decode-utils, mostly the work herein affects the debugger by way of making decode an `async` operation.

  The debugger now uses a semaphore to track outstanding decode requests and requires that it resolves before `data.current.identifiers.*` selectors are available for use. Notably, this adds a new `session.variables()` method as the canonical means to get the current scope of variables.

- This also includes a poor excuse for a translation later between the decoder and the debugger, in order exclusively [for now] to convert mappings coming back from the decoder into JS `Map` objects.

- Since the decoder returns BNs, and since @seesemichaelj makes a compelling argument against `number`s, the debugger no longer cleans BNs, and instead we inject a custom inspect method for BNs for use in the debugger repl context.

- **Continuous Integration now contains a script for bootstrap instead of the inline invocation**. This is to account for the decoder and decode utils requiring a `tsc` build.

- The decoder in its current form includes a draft for tests in the form of a Truffle project. Much of the decoder logic is covered by the existing decoding integration tests. The decode utils don't have tests right now.


scripts/bootstrap.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# The below ` || exit 1;` is added to stop the script if a command fails
```
gnidan: You can do `set -e` for this exact behavior!

I also like including `-x` (`set -ex`), which prints each command before it runs.

seesemichaelj: Do you mean `set -ex insert_command_here`?

gnidan: Nope, that's the whole line.

```bash
#!/usr/bin/env/bash

set -ex

npm run lerna:bootstrap
# ...
```

seesemichaelj: oh cool, thanks! sorry for asking instead of googling :sweat_smile:

gnidan:
This has been hammer-tested significantly.

There are outstanding problems with decoding, mostly in the form of unsupported data types, or broken data types. At this point, none of the issues are regressions, but it's unclear whether the underlying brokenness is the same.

Merging now before it gets any bigger.

gnidan:
@seesemichaelj thank you for getting these new packages going. exciting days ahead as they mature!

Issue #1189, opened 2018-08-13
Truffle init --force, please


eduardonunesp:
## Issue

When run `truffle init` on a directory that already exists `truffle` gives an error.

```Something already exists at the destination. `truffle init` and `truffle unbox` must be executed in an empty folder. Stopping to prevent overwriting data.```

## Steps to Reproduce

`yarn add truffle`
`yarn truffle init`

## Expected Behavior

Truffle should have an option `--force`, to force the creation of the directory, because install `node` dependencies `globally` are simple so bad behavior.

And if the directory is just created as "empty", why not let the user just `--force` and avoid some workaround to bypass that limitation.

## Actual Results

Truffle just stop the creation of the directories expected from `init` command

## Environment

* Operating System: macOS 10.13.6
* Truffle version (`truffle version`): 4.1.14
* node version (`node --version`): 8.11.3
* npm version (`npm --version`):  5.6.0
* yarn version (`yarn -v`): 1.9.4

cgewecke:
@eduardonunesp Yes this is a very reasonable suggestion. . . thanks for proposing.

gnidan:
Note: the work that originally forced the empty directory is here: https://github.com/trufflesuite/truffle-box/pull/15

Might be useful to see how to undo this.

PR #1270, opened 2018-09-15
Add --force option to both init and unbox commands

eggplantzzz:
This PR adds the ability to include `--force` as a parameter when using the unbox or init commands.

Typically the directory in which these commands are executed needs to be empty or the process is halted before completion.  When the `--force` parameter is included, truffle ignores the state of the directory and completes the init or unbox action.  It will overwrite any files that conflict.

This PR was inspired by issue #1189


eggplantzzz:
It may be smart to include a more dangerous looking message with the documentation that tells users about `--force` in case people get upset when their files are overwritten.

Issue #1222, opened 2018-08-23
Add code coverage


cgewecke:
To see our own internal code coverage

gnidan:
Might be tricky re: monorepo, might need to check another monorepo to see how they do it

gnidan:
Probably want to hook into something like https://coveralls.io/

PR #1272, opened 2018-09-17
Add code coverage

eggplantzzz:
This PR implements nyc and coveralls to add a code coverage report.

Inspired by issue #1222


eggplantzzz:
@gnidan 
https://gist.github.com/eggplantzzz/4cf8ee78ce3dfa3609b7ee416bde9fc7

cgewecke:
@eggplantzzz I think we can also get the right coverage readings for the scenario tests if we add an `elseif` statement testing for the `COVERAGE` environment [here](https://github.com/trufflesuite/truffle/blob/next/packages/truffle/scripts/test.sh) and use this as the command. (Don't build):
```
NO_BUILD=true mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
```

Maybe there's a webpack based solution too but that should work, we already have some logic that lets us run the tests bypassing the build since it's faster.

eggplantzzz:
@cgewecke Thanks!  Actually I'm trying to sort through why it is only giving reports for `truffle-debugger` when it is run remotely...when I run it locally it gives me reports for the whole project and not just that directory.

cgewecke:
@eggplantzzz Weird. Is it possible to run the debugger tests skipping the build step?

eggplantzzz:
@cgewecke Do you know offhand which tests require the build?

.nycrc
```diff
@@ -0,0 +1,16 @@
+{
+  "all": true,
+  "reporter": "lcovonly",
+  "exclude": [
+    "**/node_modules/**",
+    "**/.tmp"
+    ,"**/scripts/**",
```
gnidan: Rogue comma!

scripts/ci.sh
```diff
@@ -36,8 +36,15 @@ elif [ "$GETH" = true ]; then
   lerna run --scope truffle test --stream -- --exit
   lerna run --scope truffle-contract test --stream -- --exit
 
-else
+elif [ "$PACKAGES" = true ]; then
 
-  lerna run --scope truffle-* test --stream --concurrency=1
+  lerna run --scope truffle-* test --stream
```
gnidan: Did you need to remove `concurrency=1`? I seem to remember this being necessary

scripts/ci.sh
```diff
@@ -36,8 +36,15 @@ elif [ "$GETH" = true ]; then
...
+
+elif [ "$COVERAGE" = true ]; then
+
+  cd packages/truffle-debugger && npm run test:coverage && \
+  cd ../../ && nyc lerna run --ignore truffle-debugger test && \
+  cat ./packages/truffle-debugger/coverage/lcov.info >> ./coverage/lcov.info && \
+  cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js
```
gnidan: Hooray, a hack! ✨

eggplantzzz: Is there a different way that you would recommend doing this?

gnidan: Oh not at all, that's a grade A hack :)

PR #1277, opened 2018-09-18
solc 0.4.25

H34D:


gnidan:
I think this also needs to be updated in a couple other packages, incl. `truffle`. Could you please search and make sure it's changed everywhere?

gnidan:
(This addresses #1292)

H34D:
here we go

gnidan:
Thanks @H34D!

PR #1280, opened 2018-09-18
Stackframes

hjaltman:
Augments variable id's with a stack depth so that if a function appears more than once in the call stack you will see the variables for the appropriate stack frame (more specifically, `data.current.identifiers.refs` will point to the appropriate copy of the variable).  (This augmenting is only done in certain contexts -- specifically, contexts that feed into `data.current.identifiers.refs`, so be aware that there are both augmented and non-augmented IDs still used.)  Variables that are not local to a function get depth 0; depths are started counting from 1 (since that is what the existing code already does).  There is a slight potential bug in that, if `f` (at depth `d`) calls `f` (at depth `d+1`) and then later calls `f` (at depth `d+1`) again, the last of these may see the values from the earlier call to `f` at depth `d+1` for variables that have not been declared yet.  Note that support for mappings required making an alteration to the `decodeMapping` function in `decode`; this does not feed into `data.current.identifiers.refs`, but does feed into `data.current.identifiers.decoded`; without this change, mappings had the correct refs yet would not be correctly display to the user.

### DIFF
--- a/packages/truffle-debugger/lib/data/sagas/index.js
+++ b/packages/truffle-debugger/lib/data/sagas/index.js
@@ -113,7 +113,7 @@ function *tickSaga() {
       debug("currentAssignments %O", currentAssignments);
       debug("currentDepth %d varId %d", currentDepth, varId);
       yield put(actions.assign(treeId, {
-        [utils.augmentWithDepth(currentDepth,  varId)]: {"stack": top}
+        [utils.augmentWithDepth(varId, currentDepth)]: {"stack": top}
       }));
       break;
 
@@ -130,7 +130,7 @@ function *tickSaga() {
       //augment declaration Id w/0 to indicate storage
       let augmentedDeclarationId = utils.augmentWithDepth(baseDeclarationId);
       //indices, meanwhile, use depth as usual
-      let augmentedIndexId = utils.augmentWithDepth(currentDepth, indexId);
+      let augmentedIndexId = utils.augmentWithDepth(indexId, currentDepth);
       debug("Index access case");
       debug("currentAssignments %O", currentAssignments);
       debug("augmentedDeclarationId %s", augmentedDeclarationId)
@@ -180,7 +180,7 @@ function *tickSaga() {
       debug("currentAssignments %O", currentAssignments);
       debug("currentDepth %d node.id %d", currentDepth, node.id);
       yield put(actions.assign(treeId, {
-        [utils.augmentWithDepth(currentDepth, node.id)]: { literal }
+        [utils.augmentWithDepth(node.id, currentDepth)]: { literal }
       }));
       break;
   }
--- a/packages/truffle-debugger/lib/data/sagas/index.js
+++ b/packages/truffle-debugger/lib/data/sagas/index.js
@@ -35,7 +35,6 @@ function *tickSaga() {
   let scopes = yield select(data.info.scopes);
   let definitions = yield select(data.views.scopes.inlined);
   let currentAssignments = yield select(data.proc.assignments);
-  let mappingKeys = yield select(data.proc.mappingKeys); //just for debugging
   let curDepth = yield select(solidity.current.functionDepth); //for pairing with ids
 
   let stack = yield select(data.next.state.stack);
@@ -77,7 +76,6 @@ function *tickSaga() {
         .reduce( (acc, assignment) => Object.assign(acc, assignment), {} );
       debug("Function definition case");
       debug("currentAssignments %O", currentAssignments);
-      debug("mappingKeys %O", mappingKeys);
       debug("assignments %O", assignments);
 
       yield put(actions.assign(treeId, assignments));
@@ -90,6 +88,8 @@ function *tickSaga() {
       debug("storage vars %o", storageVars);
 
       let allocation = utils.allocateDeclarations(storageVars, definitions);
+      debug("Contract definition case");
+      debug("allocation %O",allocation);
       assignments = Object.assign(
         {}, ...Object.entries(allocation.children)
           .map( ([id, storage]) => ({
@@ -99,9 +99,7 @@ function *tickSaga() {
             }
           }) )
       );
-      debug("Contract definition case");
       debug("currentAssignments %O", currentAssignments);
-      debug("mappingKeys %O", mappingKeys);
       debug("assignments %O", assignments);
 
       yield put(actions.assign(treeId, assignments));
@@ -111,7 +109,6 @@ function *tickSaga() {
       let varId = jsonpointer.get(tree, pointer).id;
       debug("Variable declaration case");
       debug("currentAssignments %O", currentAssignments);
-      debug("mappingKeys %O", mappingKeys);
       debug("curDepth %d varId %d",curDepth,varId);
       yield put(actions.assign(treeId, {
         [curDepth + ":" + varId]: {"stack": top}
@@ -122,7 +119,6 @@ function *tickSaga() {
       // to track `mapping` types known indexes
       let {
         baseExpression: {
-          id: baseId, //not currently used
           referencedDeclaration: baseDeclarationId,
         },
         indexExpression: {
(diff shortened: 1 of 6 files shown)


packages/truffle-debugger/lib/data/sagas/index.js
```diff
@@ -10,6 +10,7 @@ import { TICK } from "lib/trace/actions";
 import * as actions from "../actions";
 
 import data from "../selectors";
+import solidity from "../../solidity/selectors";
```
gnidan: You can do `from "lib/solidity/selectors"` instead of the relative path. (Not using relative paths makes it easier to rename files, etc.)

packages/truffle-debugger/lib/data/sagas/index.js
```diff
@@ -69,8 +71,12 @@ function *tickSaga() {
 
       assignments = returnParameters.concat(parameters).reverse()
         .map( (pointer) => jsonpointer.get(tree, pointer).id )
-        .map( (id, i) => ({ [id]: {"stack": top - i} }) )
+        .map( (id, i) => ({ [curDepth + ":" + id]: {"stack": top - i} }) )
+		//depth may be off by 1 but it doesn't matter
```
gnidan: `s/\t\t/ {8}/` :)

packages/truffle-debugger/lib/data/selectors/index.js
```diff
@@ -235,13 +235,23 @@ const data = createSelectorTree({
...
-        (assignments, identifiers) => Object.assign({},
+        (assignments, identifiers, curDepth) => Object.assign({},
           ...Object.entries(identifiers)
             .map( ([identifier, id]) => {
-              let { ref } = (assignments[id] || {})
+              let matchIds = Object.keys(assignments).filter(
+                  (longId) => longId.endsWith(":"+id) //get cases of that var
+                  ).map(
+                  (longId) => longId.split(":")[0] //get just the stack frame
+                  );
```
gnidan: Style suggestion:

```javascript
let matchIds = Object.keys(assignments)
  .filter( ... )
  .map( ... );
```

packages/truffle-debugger/lib/data/selectors/index.js
```diff
@@ -235,13 +235,23 @@ const data = createSelectorTree({
...
+              //want innermost but not beyond current depth
+              let maxMatch=Math.min(curDepth,Math.max(...matchIds));
+                  //note: if no matches, will return -Infinity
+                  //however the return value in this case is irrelevant
```
gnidan: Style suggestion: 

Typically for comments that go on their own line, I'd say put them like this:

```javascript
    // above line or block of code with same indendation level
    // note: side node
    let maxMatch = Math.min(...);
    const otherLine = alsoRelatedToComment;

    // another block of code with a comment
    return { ... }
```

Note empty-line spacing to visually separate commented blocks of code.

Oh yeah, also whitespace rules:
- Whitespace on both sides of binary operators
- Whitespace after `,` in argument lists
- Whitespace at start, end, and in-between items, for array and object literals

packages/truffle-debugger/lib/data/sagas/index.js
```diff
@@ -34,6 +35,7 @@ function *tickSaga() {
   let scopes = yield select(data.info.scopes);
   let definitions = yield select(data.views.scopes.inlined);
   let currentAssignments = yield select(data.proc.assignments);
+  let curDepth = yield select(solidity.current.functionDepth); //for pairing with ids
```
gnidan: Style suggestion: don't abbreviate variable names. Preferably ever, but occasionally it's fine if it's very clear. 

Yep, it's annoying to write; I know—I typed all those long names out everywhere :).

packages/truffle-debugger/lib/data/sagas/index.js
```diff
@@ -82,46 +88,60 @@ function *tickSaga() {
...
       let allocation = utils.allocateDeclarations(storageVars, definitions);
+      debug("Contract definition case");
+      debug("allocation %O",allocation);
       assignments = Object.assign(
         {}, ...Object.entries(allocation.children)
           .map( ([id, storage]) => ({
-            [id]: {
-              ...(currentAssignments[id] || { ref: {} }).ref,
+            ["0:" + id]: { //using depth 0 to indicate storage
+              ...(currentAssignments["0:" + id] || { ref: {} }).ref,
```
gnidan: Part of me thinks it might be better if we made a utils function:
```javascript
function atDepth (id, depth = 0) {
  return `${depth}:${id}`;
}
```

Then we wouldn't even need the auxiliary variables like `augDeclarationId` or literal string concatenations everywhere, we could instead just call the function directly where it's used, like:

```javascript
    ...(currentAssignments[atDepth(id)] || { ref: {} }).ref,
```

Although I want a better name than `atDepth`. Thoughts?

packages/truffle-debugger/lib/data/sagas/index.js
hjaltman: OK, changed.

packages/truffle-debugger/lib/data/sagas/index.js
hjaltman: Oops!  Forgot to turn on `expandtab`...

packages/truffle-debugger/lib/data/selectors/index.js
hjaltman: OK, done.

packages/truffle-debugger/lib/data/selectors/index.js
hjaltman: Heh -- a number of my whitespace omissions were to get things (with their comments) to fit on one line.  A lot of comments are going to have to become separate lines now, I guess...

packages/truffle-debugger/lib/data/sagas/index.js
hjaltman: Oy, and I was going to change `longId` to `augId` for clarity -- actually I thought I *did* do that but apparently not -- but I guess it should be `augmentedId`? Yuck.  Is `cur`, as in `curDepth`, at least an acceptable abbreviation?  That one should be common enough, right?

packages/truffle-debugger/lib/data/sagas/index.js
hjaltman: Good idea -- and the reverse functions (i.e. extract id, extract depth) should exist too.  As for a name... well, I'd probably call it `augment` or `augmented`, but I'll admit that might not be very clear.  Hm.  Not sure what a good name is.

packages/truffle-debugger/lib/data/selectors/index.js
gnidan: Yeah, I play that game often, trying to get comments to fit on the same line, etc.

I like lots of short lines with empty lines in between logical sections.

When I use inline comments, I try to make sure it's just a tiny note.

packages/truffle-debugger/lib/data/sagas/index.js
gnidan: Maybe I am extreme in my views and just stubbornly refuse to abbreviate anything. I used `currentAssignments`, after all (I did consider `curAssignments`)

packages/truffle-debugger/lib/data/sagas/index.js
gnidan: I think `augment` works for now, provided the function has a docstring. I'm into domain-specific short words, even though they require that future developers take a bit of time digesting the lingo in a particular module.

So I think if `lib/data` wants to define a new term "augment" to mean "identify some entity in the context of a given function depth", then that's totally fine. It's succinct and useful to name such a concept, given this work.

packages/truffle-debugger/lib/data/sagas/index.js
hjaltman: Anyway, for now I've de-abbreviated all cases of "cur" and "aug".  I don't *think* there were any others I used...

packages/truffle-debugger/lib/data/sagas/index.js
hjaltman: I'm currently calling it `augmentWithDepth`, and the inverses `idFromAugmented` and `depthFromAugmented`.

packages/truffle-debugger/lib/data/selectors/index.js
hjaltman: OK, done, except that I have some comments which go on their own line but don't seem substantial enough to warrant whitespace before, so I didn't put whitespace before those.

Issue #1273, opened 2018-09-17
Notify user when using unsupported version of Node


gnidan:
Truffle v5 is going to require Node v8 or later.

When users are using a version of Node lower than that, an error message should be raised to let the user know.

Thanks to @axic for raising this concern.

PR #1287, opened 2018-09-20
Validate node version

eggplantzzz:
Validate that the user is running at least Node 8.

Inspired by issue #1273

### DIFF
--- a/packages/truffle-core/cli.js
+++ b/packages/truffle-core/cli.js
@@ -22,7 +22,7 @@ const command = new Command(require("./lib/commands"));
 const listeners = process.listeners('warning');
 listeners.forEach(listener => process.removeListener('warning', listener));
 
-const options = {
+let options = {
   logger: console
 };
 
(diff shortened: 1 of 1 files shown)


packages/truffle-core/cli.js
```diff
@@ -1,20 +1,28 @@
...
 
-var Config = require("truffle-config");
-var Command = require("./lib/command");
-var TaskError = require("./lib/errors/taskerror");
-var TruffleError = require("truffle-error");
+const TaskError = require("./lib/errors/taskerror");
+const TruffleError = require("truffle-error");
 
-var command = new Command(require("./lib/commands"));
+const nodeMajorVersion = parseInt(process.version.slice(1));
```
gnidan: Not to say this is worth changing, but FYI the "fancy" way to do this might be to use the [semver](https://www.npmjs.com/package/semver) NPM package to get the version number.

Actually you might want to change to this because Node `"v10.x.x"` will slice incorrectly.

eggplantzzz: Why won't it slice correctly?  The slice should just remove the leading 'v'.  Will the version number not always be prefaced by 'n'?

gnidan:
```
> parseInt('v7.0.0'.slice(1)) < 8
true
> parseInt('v10.13.0'.slice(1)) < 8
false
```

Hm. Seems fine. JS you slay me

PR #1288, opened 2018-09-20
Update readme

eggplantzzz:
Add a note to the readme explaining the purposes of the master, develop and next branches.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,13 @@ Truffle is a development environment, testing framework and asset pipeline for E
 * Instant rebuilding of assets during development.
 * External script runner that executes scripts within a Truffle environment.
 
+### A Note To Developers
+
+The current active and stable version of Truffle can be found on the `master` branch.  There are two branches that are designated for developers to make pull requests against: the `develop` and `next` branches.
+
++   `develop`: Make your pull requests against the `develop` branch if you would like to make a change for the current major version.
++   `next`: Make your pull requests against the `next` branch if you would like to make a change for the next major version.
+
 ### Install
 
 ```
@@ -59,7 +66,7 @@ npm run bootstrap
 npm test
 
 # Adding dependencies to a package
-cd packages/<truffle-package> 
+cd packages/<truffle-package>
 yarn add <npm-package> [--dev] # Use yarn
 ```
 
--- a/README.md
+++ b/README.md
@@ -19,12 +19,8 @@ Truffle is a development environment, testing framework and asset pipeline for E
 * Instant rebuilding of assets during development.
 * External script runner that executes scripts within a Truffle environment.
 
-### A Note To Developers
-
-The current active and stable version of Truffle can be found on the `master` branch.  There are two branches that are designated for developers to make pull requests against: the `develop` and `next` branches.
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,6 +19,13 @@ Truffle is a development environment, testing framework and asset pipeline for E
 * Instant rebuilding of assets during development.
 * External script runner that executes scripts within a Truffle environment.
 
+### A Note To Developers
+
+The current active and stable version of Truffle can be found on the `master` branch.  There are two branches that are designated for developers to make pull requests against: the `develop` and `next` branches.
+
++   `develop`: Make your pull requests against the `develop` branch if you would like to make a change for the current major version.
++   `next`: Make your pull requests against the `next` branch if you would like to make a change for the next major version.
```
gnidan: Thoughts:

I vaguely feel like the README ought to keep any kind of detailed development instructions at the bottom, after "Install" and such.

So I like the thought of maintaining this sort of outline:

1. Introduction / benefits / etc.
2. Installation
3. Quick usage
4. Documentation
5. Development

This way it provides a clear path for users who are just getting started.

BUT! We of course want to make sure contributors make it all the way to number 5. So I am thinking that maybe we should move "A Note to Developers" into the "Development" section, and add some kind of callout very close to the top that says "HEY! READ THIS THING OVER HERE! PLEZ"

How about putting something like this right above "Install":

| ℹ️ **Contributors**: Please see the [Development](#development) section of this README. |
| --- |

README.md
```diff
@@ -19,6 +19,13 @@ Truffle is a development environment, testing framework and asset pipeline for E
 * Instant rebuilding of assets during development.
 * External script runner that executes scripts within a Truffle environment.
 
+### A Note To Developers
+
+The current active and stable version of Truffle can be found on the `master` branch.  There are two branches that are designated for developers to make pull requests against: the `develop` and `next` branches.
+
++   `develop`: Make your pull requests against the `develop` branch if you would like to make a change for the current major version.
```
gnidan: Also, since you expressed interest in improving your technical writing, let me dig from the bottom of the barrel of suggestions:

How about:
+ `develop`: For pull requests that must target the current major release
+ `next`: For pull requests that target the upcoming major release.

It might be also worth noting that targeting `next` is strongly preferred.

gnidan: Although this raises the question: will this information need to be updated right away when we get rid of `next`?

README.md
eggplantzzz: Hmm, yeah I can get on board with that.  My initial concept was to give a brief overview explaining how things are laid out before getting down to it but yes I think you're right to push it towards the end.  I don't know if we really even need to give a shoutout at the top.  If people are going to be working on this code they will probably read section 5, right?

gnidan: You'd think that! But no: lots of people (myself included) will take any liberty they can get when it comes to avoiding reading things.

README.md
gnidan: (also, disclaimer: I claim no real expertise in matters of clarity. I just have enough experience being bad at writing things clearly that I've gotten ideas along the way... feel free to exercise judgment whenever.)

gnidan:
(Merging)

PR #1291, opened 2018-09-21
add Configstore node module and use to add user-level default config …

fainashalts:
…file


packages/truffle-config/index.js
```diff
@@ -19,6 +21,8 @@ function Config(truffle_directory, working_directory, network) {
     from: null,
   };
 
+  const default_user_config = new Configstore(DEFAULT_USER_CONFIG, {}, {globalConfigPath: true});
```
gnidan: Style tips:

- Use camelCase for variable names (`defaultUserConfig`)... this can actually probably even be renamed `userConfig` for brevity? But that doesn't matter much
- Nowadays it's recommended that object literals should have copious spacing `{ globalConfigPath: true }`

PR #1295, opened 2018-09-22
[WIP] Allow init/unbox to check & prompt user for file/dir overwrites

CruzMolina:
Added inquirer package to truffle-box & modified checkDestination().
Refactored copyTempIntoDestination() to check and compare files in current directory to tmp and prompt user to keep or overwrite.

TODO test coverage and finalize code


gnidan:
Oh shoot! This exposes a very big requirement that we've overlooked!

We can't go prompting the user no matter what, with inquirer. We can **only** do this kind of prompting if we are certain that the user is running Truffle "interactively".

I did a bit of quick research on this: can you try only running the prompts if and only if `process.stdin.isTTY` is true? And see if that makes the tests pass.

cc @fainashalts because I also just told you to be super prompt-happy. Apologies!

gnidan:
OK I did some testing to figure out if it's safe to add these kinds of prompts or not (#1297)... looks like Truffle is very likely to run in a TTY/interactive environment and so we really shouldn't add these kinds of prompts without careful consideration.

For this effort, it's probably fine, because people are more likely to run `truffle init` or `truffle unbox` in their own local terminals, vs. running them in CI, but we should probably discuss a more general plan for this.

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -81,13 +76,45 @@ function fetchRepository(url, dir) {
...
-    });
-  });
+async function copyTempIntoDestination(tmpDir, destination) {
+    const currentContent = fs.readdirSync(destination);
+      if (currentContent.length > 1) {
+        const tmpContent = fs.readdirSync(tmpDir);
+        for (let file of tmpContent) {
+           if (currentContent.includes(file)) {
+             var overwriting = [
+  		{
```
eggplantzzz: Indentation...also I personally would prefer that this is written like
```
let variableName = [
  {
    type: 'blah,
    otherType: 'otherBlah',
  }
]
```
To me this is much more readable.

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -81,13 +76,45 @@ function fetchRepository(url, dir) {
...
+    		name: 'overwrite',
+    		message: `Overwrite ${file}?`,
+    		default: false
+  		}]
+             await inquirer.prompt(overwriting)
+               .then(async answer => {
+                 if (answer.overwrite) {
+                   try {
+                     await fs.remove(file);
+                     await fs.copy(tmpDir+"/"+file, destination+"/"+file);
```
eggplantzzz: It might be more readable if you put spaces between your strings and variable names like this 
```fs.copy(tmpDir + "/" + file, destination + "/" + file);```
Or you could use interpolation like
```fs.copy(`${tmpDir}/${file}`, `${destination}/${file}`);```

File: packages/truffle-box/lib/utils/unbox.js [hunk 3]
```diff
@@ -81,13 +76,45 @@ function fetchRepository(url, dir) {
...
-function copyTempIntoDestination(tmpDir, destination) {
-  return new Promise(function(accept, reject) {
-    fs.copy(tmpDir, destination, function(err) {
-      if (err) return reject(err);
-      accept();
-    });
-  });
+async function copyTempIntoDestination(tmpDir, destination) {
+    const currentContent = fs.readdirSync(destination);
+      if (currentContent.length > 1) {
```
eggplantzzz: All this code inside this if statement is very difficult to read for me.  I wonder if we need all the try catch statements, is it possible to use a catch statement on the outside of this method?  And is there any reason to use the async fs methods as opposed to the synchronous ones?  We could get rid of all the awaits then.

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -81,13 +76,45 @@ function fetchRepository(url, dir) {
...
+    		type: 'confirm',
+    		name: 'overwrite',
+    		message: `Overwrite ${file}?`,
+    		default: false
+  		}]
+             await inquirer.prompt(overwriting)
+               .then(async answer => {
+                 if (answer.overwrite) {
+                   try {
+                     await fs.remove(file);
```
eggplantzzz: I'm not sure we need to manually remove the file, it looks like copy overwrites the file if it already exists.  This might mean we can get rid of more logic in this method to clean it up.  Or maybe I'm missing something.

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -15,13 +16,7 @@ function checkDestination(destination) {
...
       const contents = fs.readdirSync(destination);
       if (contents.length) {
-        const err = "Something already exists at the destination. " +
-                  "`truffle init` and `truffle unbox` must be executed in an empty folder. " +
-                  "If you are absolutely sure you want to proceed, add " +
-                  "the `--force` parameter to the init or unbox command. Be careful though, " +
-                  "adding `--force` will cause truffle to ignore existing files in the current " +
-                  "directory and potentially overwrite them. Stopping to prevent overwriting data."
-        throw new Error(err);
+        console.log("Something already exists in this folder...");
```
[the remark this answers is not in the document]
CruzMolina: Agreed. I'll move this into the check in `copyTemp...`

packages/truffle-box/lib/utils/unbox.js
CruzMolina: I made `copyTemp...` into an async func so I could have `inquirer` pause the processes while waiting for the user to respond. I'll take a look at this again though to see If using callbacks or promises can help clear things up.

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -7,6 +7,7 @@ var parseURL = require('url').parse;
 var tmp = require('tmp');
 var exec = require('child_process').exec;
 var cwd = require('process').cwd();
+var inquirer = require('inquirer');
```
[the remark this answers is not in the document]
CruzMolina: I agree with using `const` for requiring modules. Don't really have any strong feelings on whether we should only use `let` or const` for variable declarations or not. I like using `var` for tmp declarations scoped within functions, but `let` would be fine as well.

packages/truffle-box/lib/utils/unbox.js
eggplantzzz: Oh sorry, it wasn't the fact that you made it async.  I was trying to think of a way to simplify the indentation mountain so that the code narrative was more clear.

CruzMolina:
Thanks for the review @eggplantzzz , let me know your thoughts on the latest changes. Adding some test coverage 😎

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -81,13 +66,36 @@ function fetchRepository(url, dir) {
...
+              fs.copySync(`${tmpDir}/${file}`, `${destination}/${file}`);
+            }
+          });
+      } else {
+        fs.copySync(`${tmpDir}/${file}`, `${destination}/${file}`);
+      }
+    }
+  } else {
+    fs.copySync(tmpDir, destination);
+  }
```
gnidan: General thoughts: this function seems to be outgrowing itself. I recommend re-organizing this by splitting it into multiple functions, or at least decreasing the number of levels of nesting.

I think you might be able to use `.filter()` to assign discrete ideas into specific variables. Possible approach, feel free to use as a starting point or not:

```javascript
const boxContents = fs.readdirSync(tmpDir);
const destinationContents = fs.readdirSync(destination);

const newContents = boxContents.filter(
  (filename) => !destinationContents.includes(filename)
);

const contentCollisions = boxContents.filter(
  (filename) => destinationContents.includes(filename)
);

let shouldCopy;
if (force) {
  // write all box contents
  shouldCopy = boxContents;
} else {
  // promptOverwrites returns list of files that the user says "yes" to overwrite
  const overwriteContents = await promptOverwrites(contentCollisions);
  shouldCopy = [...newContents, ...overwriteContents];
}

for (let file of shouldCopy) {
  fs.copySync(tmpDir, destination);
}

**Comment by CruzMolina** [comment]
Decided against testing the flow and outcome of the prompts for now, since `sinon` currently doesn't natively support testing `inquirer` prompt input. There does appear to be a way to test `inquirer` prompts using `jest` and `ava`, but I'm not sure I should spend any more time on this. I can also potentially just mock test the inner methods in `copyTemp...` and `promptOver...`

@gnidan @eggplantzzz @fainashalts

**Comment by gnidan** [comment]
@CruzMolina sounds good to me! If you want to proceed using `sinon` to mock `promptOverwrites`, that seems reasonable! Then at least we can say we have our first mock in our tests :)

File: packages/truffle/cli.webpack.config.js [hunk 8]
```diff
@@ -30,7 +30,8 @@ module.exports = {
   devtool: 'source-map',
   module: {
     rules: [
-      { test: /\.js$/, use: "shebang-loader" }
+      { test: /\.js$/, use: "shebang-loader" },
+      { test: /rx\.lite\.aggregates\.js/, use: 'imports-loader?define=>false' }
```
gnidan: Where do these files come from? Generated by sinon?

packages/truffle/package.json
```diff
@@ -4,6 +4,7 @@
   "version": "5.0.0-next.12",
   "description": "Truffle - Simple development framework for Ethereum",
   "dependencies": {
+    "imports-loader": "^0.8.0",
```
gnidan: This should be a devDependency

packages/truffle/cli.webpack.config.js
gnidan: (Just wondering about the context for this + the imports-loader thing)

packages/truffle-box/test/prompt.js
```diff
@@ -0,0 +1,111 @@
...
+  it('...can choose to overwrite redundant files/folders', () => {
+    const result = promptMocks.mockPromptOverwrites(collisions, true);
+    assert.deepEqual(result, collisions);
+  });
+
+  it('...can choose to not overwrite redundant files/folders', () => {
+    const result = promptMocks.mockPromptOverwrites(collisions, false);
+    assert.deepEqual(result, []);
+  });
+});
```
gnidan: Exciting to see this foray into mocking!

packages/truffle/cli.webpack.config.js
CruzMolina: turns out some of the checks were failing when I added `inquirer` because `rx-lite-aggregates` (part of the `rxjs` dep for `inquirer`) was calling for `'./rx.lite` instead of `rx-lite`.

packages/truffle-box/test/box.js
```diff
@@ -1,81 +1,104 @@
...
+          assert(fs.existsSync(truffleConfigPath), "truffle.js wasn't recreated!");
+          const newConfig = fs.readFileSync(truffleConfigPath, "utf8");
+          assert(newConfig !== mockConfig, "truffle.js wasn't overwritten!");
+          done();
+        });
+    }).timeout(5000);
+
+  });
+
+  describe("init/unbox prompt", () => {
```
eggplantzzz: I feel like you could probably just get rid of the last two tests in this group and just test that when 4 name conflicts exist, inquirer gets called 4 times.  But you probably should start the test by confirming that you do indeed have 4 name conflicts.

PR #1296, opened 2018-09-23
Get coverage reading for scenario tests

cgewecke:
Routes execution through the repo instead of the bundle when running the scenario tests for coverage.


eggplantzzz:
Sorry @cgewecke, I know you requested something like this but I wasn't sure about which packages relied on a build step if any...thanks for the edit!

PR #1297, opened 2018-09-24
[do not merge!] Test interactivity hypothesis

gnidan:
Opening this PR just for CI purposes as a place to test hypotheses about determining if a shell is interactive or not.

### DIFF
--- a/packages/truffle-core/cli.js
+++ b/packages/truffle-core/cli.js
@@ -29,6 +29,7 @@ if (userWantsGeneralHelp) {
 
 /** TEST check for interactivity in Travis */
 console.debug("stdin.isTTY %o", process.stdin.isTTY);
+process.exit(1);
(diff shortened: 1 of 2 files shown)


gnidan:
Ah second test shows something not very promising: https://travis-ci.org/trufflesuite/truffle/jobs/432287551#L606

Running node inside `ci.sh` has a TTY.

gnidan:
Hm so that last one fails.

So it looks like Travis gives an interactive shell by default, but Truffle's tests are getting run inside a non-interactive environment.

I just checked locally to see if `yarn`/`npm` give a TTY for package.json `"scripts"`, and they do. So the normal case of a user running Truffle in CI means we can't just start prompting all over the place.

gnidan:
`</science>`

Issue #1274, opened 2018-09-17
Compiler configuration could support not only paths but https URLs (V5)


axic:
According to the [0.5.0 release notes](https://github.com/trufflesuite/truffle/releases/tag/v5.0.0-beta.0) there are different ways to select a compiler version.

Using a version number
```
  solc: {
    version: "0.4.22",
  }
```
only seems to work with releases, and `truffle compile --list --all` also only lists them.

It would be nice to support nightly releases (in some cases), though it may be a bad idea supporting that easily via the above feature.

However, truffle also supports specifying a path, such as `version: "/home/test/mycompiler.js"`, which I think could be extended to be a URL:
- `file:///home/test/mycompiler.js`
- `https://solc-bin.ethereum.org/bin/soljson-v0.4.26-nightly.2018.9.13+commit.8b089cc8.js`

In this case it could be used for nightlies and specific compiler versions (such as one with Z3 included), yet it requires a bit more work from the user compared to just dropping in a number and as a result should discourage randomly selecting a prerelease.

gnidan:
From conversation with @axic offline: there might also be an issue with how CompilerSupplier is wrapping `soljson`. It seems that truffle-compile is not working correctly with raw `soljson` values, and instead only works with the full `solcjs` pre-built wrapper.

This should not be necessary because of the `solcWrap` functionality in that package (invoked [here](https://github.com/trufflesuite/truffle/blob/next/packages/truffle-compile/compilerSupplier.js#L411), definition [here](https://github.com/trufflesuite/truffle/blob/next/packages/truffle-compile/solcWrap.js)).

Beyond that, it might be possible to use the wrapping functionality provided by `solcjs` natively, by calling `require('solc/wrapper')(require('/path/to/soljson'))`, although this might not work within the confines of the bundled version of Truffle. (@cgewecke would know why the `solcWrap` file exists in the first place.)

[This project](https://github.com/ewasm/sentinel-governance) has a Truffle project set up with the config pointing to the desired `soljson` URL (`https://`). This config is not working, and instead [this branch](https://github.com/ewasm/sentinel-governance/tree/truffle-workaround) has a workaround.

cgewecke:
@axic @gnidan 

Not sure if this is helpful but it *is* currently possible to get a list of nightlies and to pull one from solc-bin:
```shell
truffle compile --list prereleases --all
[
 ...
 "0.4.25-nightly.2018.9.10+commit.86d85025",
[...]

PR #1299, opened 2018-09-25
Download spinner

eggplantzzz:
This adds ora to the truffle-compile package.  It provides the user with a status update while downloading the solc version list and whenever it needs to download a compiler from solc-bin.

Inspired by issue #1274

### DIFF
--- a/packages/truffle-compile/package.json
+++ b/packages/truffle-compile/package.json
@@ -8,6 +8,7 @@
     "colors": "^1.1.2",
     "debug": "^3.1.0",
     "find-cache-dir": "^1.0.0",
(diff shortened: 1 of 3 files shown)


packages/truffle-compile/package.json
```diff
@@ -8,6 +8,7 @@
     "colors": "^1.1.2",
     "debug": "^3.1.0",
     "find-cache-dir": "^1.0.0",
+    "ora": "^3.0.0",
```
gnidan: Can you also upgrade `truffle-reporters`'s dependency? Or just use the same one that package is using (`^2.1.0`)

Issue #1294, opened 2018-09-21
Custom build pipeline stderr semantics


andrewosh:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

When using a custom build process, any output on stderr is assumed to be a "build error," and the CLI messages reflect this. I'd imagine Truffle should instead assume that the custom build process interprets its own output correctly, and limits itself to interpreting the return code.

As an example, I have a custom build script that uses `cargo` to compile a few Rust contracts, and my build output is:
```
build error:     Finished release [optimized] target(s) in 0.03s

build error:     Finished release [optimized] target(s) in 0.04s

build error:     Finished release [optimized] target(s) in 0.05s
```

In the Rust world, these informational messages are emitted on `stderr`. 

## Steps to Reproduce

Successfully compile any Rust contract with `cargo build` using a custom build process.

## Expected Behavior

Perhaps just omit the "build error" and let the build script handle errors! With this approach, stderr messages from the builder should just be forwarded to the wrapper's stderr (with `console.error` instead of `console.log`).

Relevant line is [here](https://github.com/trufflesuite/truffle/blob/99d00337477b17e516ebae91f6a53012e1500dc2/packages/truffle-core/lib/build.js#L38)

Thoughts?

## Actual Results

(Described above)

## Environment

(Not relevant to issue)

gnidan:
Thanks for bringing this up! That's a really good point. Feel free to open a PR against `develop` with the change.

PR #1305, opened 2018-09-27
Reclaim watch and build commands

eggplantzzz:
This reverts the commits responsible for removing the build and watch commands.  There is also some refactoring, syntax updating, test adding and reordering of an index file so the help is in alphabetical order.


eggplantzzz:
I'm not sure if it is necessary to have the continuous function calling in `continuouslyCheck...`, if it is not there could be some simplification.  I also have some ideas about extracting that function as well as the `print..` methods to make `run` clearer but I don't know how beneficial that is.

Also, do we want to test this?

packages/truffle-core/lib/build.js
```diff
@@ -0,0 +1,140 @@
...
+
+function CommandBuilder(command) {
+  this.command = command;
+};
+
+CommandBuilder.prototype.build = function(options, callback) {
+  console.log("Running `" + this.command + "`...")
+
+  const args = spawnargs(this.command);
+  const ps = args.shift();
```
eggplantzzz: By the way, what does `ps` stand for?

packages/truffle-core/lib/commands/help.js
```diff
@@ -1,6 +1,6 @@
 var command = {
   command: "help",
-  description: "Display a list of all commands or information about a specific command",
+  description: "Display a list of all commands or info about a specific command",
```
fainashalts: why shorten this to "info"? "information" may be more understandable to people for whom English is not their first language since info is slang.

packages/truffle-core/lib/build.js
```diff
@@ -0,0 +1,133 @@
+const async = require("async");
+const mkdirp = require("mkdirp");
+const del = require("del");
+const fs = require("fs");
+const Contracts = require("truffle-workflow-compile");
```
fainashalts: is there a reason the Contracts and BuildError constants are capitalized in the first letter or should they start with a lowercase?

packages/truffle-core/lib/build.js
```diff
@@ -0,0 +1,133 @@
...
+const BuildError = require("./errors/builderror");
+const child_process = require("child_process");
```
fainashalts: should this be childProcess for consistency?

packages/truffle-core/lib/build.js
```diff
@@ -0,0 +1,133 @@
...
+              return callback(new BuildError(err));
+            }
+            return callback(err);
+          });
+        }
+      });
+    });
+  },
+
+  // Deprecated: Specific to default builder.
```
fainashalts: Is this actually deprecated? If so might be a good time to remove.

packages/truffle-core/lib/build.js
gnidan: That's just convention. Typically it depends what gets returned... if you're requiring a class or something that looks like a class, UpperCamelCase seems appropriate

packages/truffle-core/lib/build.js
adrianmcli: The only usage of it is `spawn`, and in the official [documentation](https://nodejs.org/api/child_process.html), they deconstruct directly to it. So we should probably do the same. For example:

```js
const { spawn } = require('child_process');
```

gnidan: It's annoying that this is what Node calls it. I usually sidestep the issue by requiring specific functions, not all of `child_process`.

packages/truffle-core/lib/build.js
gnidan: Process? Search me.

packages/truffle-core/lib/errors/builderror.js
```diff
@@ -0,0 +1,12 @@
+var colors = require("colors");
+var TruffleError = require("truffle-error");
+var inherits = require("util").inherits;
```
fainashalts: can we change these to const instead of var?

packages/truffle/test/scenarios/commands/build.js
```diff
@@ -0,0 +1,65 @@
+const assert = require("assert");
+const CommandRunner = require("../commandrunner");
+const MemoryLogger = require("../memorylogger");
```
fainashalts: should these be commandRunner and memoryLogger for snake case? I'm not sure if there's a special reasons for capitalizing the first letter, haven't had my coffee yet!

packages/truffle-core/lib/build.js
fainashalts: Ah, that makes sense. Will keep that in mind, thanks!

packages/truffle-workflow-compile/index.js
```diff
@@ -70,7 +70,7 @@ var Contracts = {
     const compilers = (config.compiler)
       ? [config.compiler]
       : Object.keys(config.compilers);
-
+
```
CruzMolina: Can you remove the whitespace here?

adrianmcli: Should really integrate prettier into the workflow. Someone with more bandwidth should get on that.

packages/truffle-core/lib/build.js
eggplantzzz: @adrianmcli Yes, I like deconstructing it.

packages/truffle-core/lib/build.js
eggplantzzz: @gnidan Can you comment on this?

packages/truffle-core/lib/commands/help.js
eggplantzzz: I changed this to make it fit on one line.  When `truffle help` is used, there previously were too many characters in the string and it wrapped to the next line.  We could try to shorten it some other way...

gnidan: Maybe `List all commands or provide information about a specific command`

packages/truffle-core/lib/build.js
gnidan: Is it used? I vote delete

eggplantzzz: It seems like it is unused, I'll get rid of it.

eggplantzzz:
Closes #1294

CruzMolina:
It's been connected, but for clarity, marking #1340 as containing the test steps.

Issue #1276, opened 2018-09-17
Improve debugger breaking


gnidan:
# Overview

Breakpoints in Truffle's debugger are kind of useless.

# Background

Currently, breakpoints can be set at a particular location in the code, but not set ahead of time. So a user has to step through the trace to the point of execution where they wish the debugger to `(b)`reak, and then `(c)`ontinue. The debugger then steps until it hits that point in the code on subsequent times.

Internally, truffle-debugger library permits setting breakpoints at a given AST node or at a given line number. Setting breakpoints at lines is currently a bit wonky because of how EVM operations map to line number in practice. So the workaround is that breakpoints are toggled at the current AST node in the step, since this results in a more intuitive break after continuing.

# Expected

1. `truffle-debug` **should** support breakpoints at future code locations. Users should be able to specify a line number in either the current file or in other files used in the transaction.

    Possible interface:

    - `debug(...)> b`: Toggle (same as now)
    - `debug(...)> b <line-number>`: Set a breakpoint at a given line in the current source file
    - `debug(...)> b <filename>:<line-number>`: Set a breakpoint in another file. `<filename>` should be easy to type, so maybe it should accept a contract-name? Or at least it should accept the _basename_ for a given file, e.g. `MyContract` for `./contracts/MyContract.sol`.
    - `debug(...)> ?`: Show watch expressions (current) **and** breakpoints (new)
    - (maybe?) `debug(...)> b +<relative-line>`: Set a breakpoint for some number of lines ahead
    - (maybe also?) `debug(...) b -<relative-line>`

2. The debugger should be smart about translating line numbers to actual AST node breakpoints. Open question: How do we pick which AST node to use as the breakpoint for a given line? Do we want to set the breakpoint to left-most AST node in a line? To the AST node that most completely comprises a given line? This might involve playing with the operation in practice to see where it makes sense to break from a UI standpoint.

itstehkman:
@gnidan I think this is a good question to look into. Without looking into it much, my intuition says the breakpoint should match with the left-most AST node, if that corresponds to the left-most part of the line of code. I am probably oversimplifying the problem and am wrong, but would love to hear some more about it.

PR #1306, opened 2018-09-27
Improved breakpoint functionality

hjaltman:
This branch expands the functionality of breakpoints as requested in #1276.  Breakpoints can now be set on a line of the user's choice in a source file of their choice; they can also be set at the current location as before.  Breakpoints are no longer toggled with `b` but rather set with `b` and cleared with `B`.  A breakpoint set with `b` with no arguments is set at the current location; this is the current location *within* the line, like the old breakpoints.  Breakpoints may also be set at a given line (including in another source file), which may also be specified relative to the current line; these breakpoints will stop execution at the *first* encountered location within that line.  Breakpoints in another source file need only have enough of the filename specified to unambiguously identify it; if what's given is ambiguous, the user will be prompted with a list of possibilities.

This branch does not include expanded functionality for `?`.

### DIFF
--- a/packages/truffle-core/lib/commands/debug.js
+++ b/packages/truffle-core/lib/commands/debug.js
@@ -312,7 +312,9 @@ var command = {
 
           var currentNode = currentLocation.node.id;
           var currentLine = currentLocation.sourceRange.lines.start.line;
-          var currentSource = currentLocation.source;
+          var currentSourceId = currentLocation.source.id;
+
+          var sourceName; //to be used if a source is entered
 
           var breakpoint = {};
 
@@ -323,7 +325,7 @@ var command = {
             debug("node case");
             breakpoint.node = currentNode;
             breakpoint.line = currentLine;
-            breakpoint.source = currentSource;
+            breakpoint.sourceId = currentSourceId;
           }
 
           //if the argument starts with a "+" or "-", we have a relative
@@ -340,7 +342,7 @@ var command = {
               return;
             }
 
-            breakpoint.source = currentSource;
+            breakpoint.sourceId = currentSourceId;
             breakpoint.line = currentLine + delta;
           }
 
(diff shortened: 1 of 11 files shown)


packages/truffle-core/lib/commands/debug.js
```diff
@@ -303,46 +305,188 @@ var command = {
...
+
+          //finally, report back to the user on what happened
+          if(setOrClear)
+          {
+            config.logger.log(`Breakpoint added at ${locationMessage}.`);
+          }
+          else
+          {
+            config.logger.log(`Breakpoint removed at ${locationMessage}.`);
+          }
```
gnidan: Ah so I just tried this feature out and realized something that might be nice: an extra blank line after these added/removed messages.

This looks a bit cluttered:
```
debug(develop:0x91c817a1...)> b SquareLib:5
Breakpoint added at line 5 in SquareLib.sol.
debug(develop:0x91c817a1...)> b +10
Breakpoint added at line 23.
debug(develop:0x91c817a1...)> B +10
Breakpoint removed at line 23.
```

Compared with:
```
debug(develop:0x91c817a1...)> b SquareLib:5
Breakpoint added at line 5 in SquareLib.sol.

debug(develop:0x91c817a1...)> b +10
Breakpoint added at line 23.

debug(develop:0x91c817a1...)> B +10
Breakpoint removed at line 23.
```

packages/truffle-debugger/test/data/decode/helpers.js
```diff
@@ -65,12 +66,23 @@ async function prepareDebugger(testName, sources) {
...
+
   let breakpoint = {
-    address: instance.address,
-    line: lastStatementLine(sources[fileName(testName)])
+    sourceId,
+    line: lastStatementLine(source)
   };
 
-  session.continueUntil(breakpoint);
+  session.setOrClearBreakpoint(breakpoint,true);
```
gnidan: Missed this!

packages/truffle-debugger/test/solidity.js
```diff
@@ -103,11 +103,14 @@ describe("Solidity Debugging", function() {
     let session = bugger.connect();
 
     // at `second();`
-    let breakpoint = { "address": instance.address, line: 16 }
+    let source = await session.view(solidity.current.source);
+    let breakpoint = { sourceId: source.id, line: 16 }
     let breakpointStopped = false;
 
+    session.setOrClearBreakpoint(breakpoint,true);
```
gnidan: Also this!

PR #1307, opened 2018-09-28
Wrap getAddress function in provider

sc0Vu:
Due to truffle can use another provider like [truffle-hdwallet-provider](https://github.com/trufflesuite/truffle-hdwallet-provider) or [truffle-wallet-provider](https://github.com/crypedit/truffle-wallet-provider), user can call getAddres() in provider but not in wrapped provider.

I try to add `getAddress()` in wrapped provider that user can call `deployer.getAddress()` to get provider address.

### DIFF
--- a/packages/truffle-migrate/index.js
+++ b/packages/truffle-migrate/index.js
@@ -224,6 +224,19 @@ var Migrate = {
 
           callback(err, result);
         });
+      },
+      getAddress: function() {
+        var result = "";
+
+        if (provider.getAddress === undefined) return result;
+
+        if (arguments.length === 1) {
+          result = provider.getAddress(arguments[0]);
+        } else {
+          result = provider.getAddress();
+        }
+
+        return result;
       }
     };
   },


packages/truffle-migrate/index.js
```diff
@@ -224,6 +224,19 @@ var Migrate = {
...
+
+        if (provider.getAddress === undefined) return result;
+
+        if (arguments.length === 1) {
+          result = provider.getAddress(arguments[0]);
+        } else {
+          result = provider.getAddress();
+        }
+
+        return result;
```
gnidan: Can't we just simplify this?

Something like:
```javascript
getAddress: function() {
  return provider.getAddress.apply(provider, arguments);
}
```

Not sure if we need the additional undefined check, since I think we want an exception if the underlying provider doesn't have `.getAddress()`. And the arguments can just be forwarded en-masse via `.apply`

sc0Vu: @gnidan `apply` throw error if `getAddress` function is undefined: Uncaught TypeError: Cannot read property 'apply' of undefined.

Maybe we can change like:

```JS
getAddress: function() {
  if (provider.getAddress === undefined) return '';
  return provider.getAddress.apply(provider, arguments);
}
```

gnidan: Don't we want an exception to happen, though? Or is `''` a reasonable return value in this situation? My thinking is that if the user does `truffleProvider.getAddress` but `provider.getAddress` does not exist, then there should be an exception.

gnidan:
Since the above linked PR removes this wrapper, I think we can now close this (since not having a wrapper obviates the need to expand the wrapper :)

PR #1327, opened 2018-10-09
Add debugger support for precompiled contracts

gnidan:
This prevents the debugger from failing when a contract uses an EVM precompile.

Roughly, this includes:
- Allowing the debugger to track EVM "contexts" that don't have a binary
- Fixing a bunch of `undefined` return values that should be objects
- Change missing binary behavior in `*evm.addInstance` saga (so it adds both binary-less context and instance)
- Add special case handling of CALLs to precompiles

### DIFF
--- a/packages/truffle-debugger/lib/evm/reducers.js
+++ b/packages/truffle-debugger/lib/evm/reducers.js
@@ -14,27 +14,55 @@ function contexts(state = DEFAULT_CONTEXTS, action) {
      * Adding a new context
      */
     case actions.ADD_CONTEXT:
-      let { contractName, binary } = action;
+      const addContext = ({ contractName, raw }) => {
+        let context = keccak256(raw);
 
-      if (state.byBinary[binary]) {
-        return state;
+        return {
+          ...state,
+
+          byContext: {
+            ...state.byContext,
+
+            [context]: {
+              ...(state.byContext[context] || {}),
+
+              contractName, context
+            }
+          },
+        };
       }
 
-      let context = keccak256(binary);
+      return addContext(action);
 
-      return {
-        byContext: {
-          ...state.byContext,
-
-          [context]: { context, binary, contractName }
-        },
-
-        byBinary: {
-          ...state.byBinary,
-
-          [binary]: { context: context }
+    /*
+     * Adding binary for a context
+     */
+    case actions.ADD_BINARY:
+      const addBinary = ({ context, binary }) => {
+        if (state.byBinary[binary]) {
+          return state;
         }
-      };
+
+        return {
+          byContext: {
+            ...state.byContext,
+
+            [context]: {
+              ...state.byContext[context],
+
+              binary
+            }
+          },
+
+          byBinary: {
+            ...state.byBinary,
+
+            [binary]: { context: context }
+          }
+        };
+      }
+
+      return addBinary(action);
 
     /*
      * Default case
--- a/packages/truffle-debugger/lib/evm/reducers.js
+++ b/packages/truffle-debugger/lib/evm/reducers.js
@@ -93,7 +93,7 @@ function instances(state = DEFAULT_INSTANCES, action) {
         byAddress: {
           ...state.byAddress,
 
-          [address]: { context, binary }
+          [address]: { address, context, binary }
         },
 
         byContext: {
--- a/packages/truffle-debugger/lib/evm/sagas/index.js
+++ b/packages/truffle-debugger/lib/evm/sagas/index.js
@@ -56,13 +56,32 @@ export function* begin({ address, binary }) {
 }
 
 export function* callstackSaga () {
+  let contexts;
+  let instances;
+
   while (true) {
     yield take(TICK);
-    debug("got TICK");
+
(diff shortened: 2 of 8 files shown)


packages/truffle-debugger/lib/evm/reducers.js
```diff
@@ -14,27 +14,55 @@ function contexts(state = DEFAULT_CONTEXTS, action) {
...
-        },
+              contractName, context
+            }
+          },
+        };
+      }
 
-        byBinary: {
-          ...state.byBinary,
+      return addContext(action);
```
gnidan: @hjaltman said:
> On evm/reducers.js: The style here is confusing. For both the ADD_CONTEXT case and the ADD_BINARY case, you define a function in the body of the case and then immediately use it, just once. Why define a function at all, if it's never used again? If it's going to be used again, why define it just here? I think this would be clearer just written out without the extra functions.

I was just trying this out. We have a couple switch statements that declare a bunch of variables, and I'm sick of having to move the variable declarations up outside the switch statement. So here I'm trying out the pattern for a particular case:

```javascript
case something:
  const inlineFunc = ({ /* action... */ }) => {
    /* ... */
  }
  
  return inlineFunc(action)
```

Not sure. What do you think?

packages/truffle-debugger/lib/evm/reducers.js
```diff
@@ -14,27 +14,55 @@ function contexts(state = DEFAULT_CONTEXTS, action) {
      * Adding a new context
      */
     case actions.ADD_CONTEXT:
-      let { contractName, binary } = action;
+      const addContext = ({ contractName, raw }) => {
```
hjaltman: The style here is confusing.  For both the `ADD_CONTEXT` case and the `ADD_BINARY` case, you define a function in the body of the case and then immediately use it, just once.  Why define a function at all, if it's never used again?  If it's going to be used again, why define it just here?  I think this would be clearer just written out without the extra functions.

packages/truffle-debugger/lib/evm/sagas/index.js
```diff
@@ -45,14 +56,33 @@ export function* begin({ address, binary }) {
...
 
 export function* callstackSaga () {
+  let contexts;
+  let instances;
+
   while (true) {
     yield take(TICK);
-    debug("got TICK");
+
+    // HACK contexts and instances won't be available until after first TICK
```
hjaltman: I don't understand the point of this block; I thought we just got a `TICK`?  What's going on here?

hjaltman: I don't understand the point of this block; I thought we just got a TICK? What's going on here?

packages/truffle-debugger/lib/evm/reducers.js
hjaltman: The style here is confusing. For both the ADD_CONTEXT case and the ADD_BINARY case, you define a function in the body of the case and then immediately use it, just once. Why define a function at all, if it's never used again? If it's going to be used again, why define it just here? I think this would be clearer just written out without the extra functions.

packages/truffle-debugger/lib/evm/sagas/index.js
gnidan: I originally had this assignment before the `while` loop, because the contexts/instances never change. But rather than stringing in `READY`, I'm just doing that `if` statement because I know if we have just gotten a TICK, then the session is READY

packages/truffle-debugger/lib/evm/reducers.js
gnidan: I was just trying this out. We have a couple switch statements that declare a bunch of variables, and I'm sick of having to move the variable declarations up outside the switch statement. So here I'm trying out the pattern for a particular case:

```javascript
case something:
  const inlineFunc = ({ /* action... */ }) => {
    /* ... */
  }
  
  return inlineFunc(action)
```

Not sure. What do you think?

packages/truffle-debugger/lib/evm/sagas/index.js
hjaltman: Oh!  I get it now.  The hack is not those lines or the if block, but that it's inside the while-loop.  I'd suggest adding a comment to the code with something like what you just said -- e.g., `//only make this assignment the first time through the loop`, or something to that effect.

packages/truffle-debugger/lib/session/sagas/index.js
```diff
@@ -98,7 +98,7 @@ function* fetchTx(txHash, provider) {
 
 function* recordContexts(...contexts) {
   for (let { contractName, binary, sourceMap } of contexts) {
-    yield *evm.addContext(contractName, binary);
+    yield *evm.addContext(contractName, { binary });
```
gnidan: Note: this is the first time a saga is defined that tags named arguments like this. I don't know if this break in pattern is a bad idea.

hjaltman: Oh, interesting. Personally I'm fine with it, but I guess it's worth noting?  Anyway yeah all this passes my review at least.

Issue #862, opened 2018-03-21
Truffle compiler option 'contracts_build_directory' inconsistent behavior


barakman:
Truffle compiler option `contracts_build_directory` is described [here](http://truffleframework.com/docs/advanced/configuration#contracts_build_directory).

I'm having trouble with Truffle's inconsistent behavior on that configuration parameter.

I'm setting it as follows:

    contracts_build_directory: "../../artifacts"

When I execute `truffle compile`, all artifacts are created in a folder named "artifacts" which is located two levels above the project's root directory.

When I execute `truffle migrate --network=development`, I get the following error message:

    Running migration: 1_initial_migration.js
    fs.js:904
      return binding.readdir(pathModule._makeLong(path), options.encoding);
                     ^
    
    Error: ENOENT: no such file or directory, scandir '<see_below>\artifacts'

Where `<see_below>` is a full path which ends only one level above the project's root directory.

I can therefore resolve this problem as follows:

1. Run `truffle compile`
2. Change `contracts_build_directory` from `"../../artifacts"` to `"../../../artifacts"`
3. Run `truffle migrate --network=development`

This solution is inappropriate of course.

My project directory structure is as follows:

    project folder:
        contracts folder:
            sol files
        migrations folder:
            1_initial_migration.js file
            2_deploy_contracts.js file
        test folder:
            js files
        truffle-config.js file

My truffle configuration (in truffle-config.js) is as follows:

    // See <http://truffleframework.com/docs/advanced/configuration>
    module.exports = {
        contracts_build_directory: "../../artifacts",
        networks: {
            development: {
                host:       "127.0.0.1",
                network_id: "*",
                port:       8545,
                gas:        4712388,      // Gas limit used for deploys
                gasPrice:   100000000000, // Gas price used for deploys
            },
        },
        mocha: {
            useColors: true,
            bail:      true,   // Abort after first test failure
            reporter:  "list", // See <https://mochajs.org/#reporters>
        },
        solc: {
            optimizer: {
                enabled: true,
                runs:    5000000,
            },
        },
    };

To my understanding, the problem stems from the fact that `truffle migrate` "steps into" the `migrations` folder and runs from there (hence an additional `..\` is required in the value of the `contracts_build_directory` parameter).

UPDATE:

[...]

PR #1331, opened 2018-10-11
Support configuring relative paths in `_directory` properties

gnidan:
#862 (among others)

This PR expands the capability for truffle-config's property descriptions, to enable preprocessing config values. In addition to the normal Node `get` and `set`:
  - Define `default` option for specifying default value
  - Define `transform` option for preprocessing set values

`default` is necessary because existing directory properties are getting their default values via a function literal.

### DIFF
--- a/packages/truffle-config/index.js
+++ b/packages/truffle-config/index.js
@@ -243,12 +243,32 @@ function Config(truffle_directory, working_directory, network) {
 };
 
 Config.prototype.addProp = function(key, obj) {
+  // possible property descriptors
+  //
+  // supports `default` and `transform` in addition to `get` and `set`
+  //
+  // default: specify function to retrieve default value (used by get)
+  // transform: specify function to transform value when (used by set)
   Object.defineProperty(this, key, {
+    // retrieve config property value
     get: obj.get || function() {
-      return this._values[key] || obj();
+      // value is specified
+      if (key in this._values) {
+        return this._values[key];
+      }
+
+      // default getter is specified
+      if (obj.default) {
+        return obj.default()
+      };
+
+      // obj is a function
+      return obj();
     },
     set: obj.set || function(val) {
-      this._values[key] = val;
+      this._values[key] = (obj.transform)
+        ? obj.transform(val)
+        : val;
     },
     configurable: true,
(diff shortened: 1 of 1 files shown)


packages/truffle-config/index.js
```diff
@@ -67,6 +67,9 @@ function Config(truffle_directory, working_directory, network) {
     }
   };
 
+  const resolveDirectory = (value) =>
+    path.resolve(self.working_directory, value);
```
gnidan: I don't really know a good place for this. I was going to avoid abstracting it, but I didn't like copying and pasting it everywhere.

Thoughts?

fainashalts: I like it abstracted -- makes it clearer what's happening. I wonder if there is any way to move this logic into the addProp function, or would that just make things messier? I'm just thinking since you are already checking there whether you need to account for the path anyway?

packages/truffle-config/index.js
```diff
@@ -243,12 +252,32 @@ function Config(truffle_directory, working_directory, network) {
 };
 
 Config.prototype.addProp = function(key, obj) {
```
eggplantzzz: This is already existing code, but I really dislike the variable names...I'm a little bit slow and when trying to read this, it destroys the narrative for me.  What about something like `function (propName, propValue)`?  What do you think?

eggplantzzz: I also really dislike abbreviated variables like `str`.  That is just a subjective stylistic thing so feel free to tell me to go fly a kite.

packages/truffle-config/index.js
gnidan: The reason it's not part of `addProp` is because certain properties are directories, and that feels like it should be explicitly indicated as such. If I were to write this from scratch, I'd define a bunch of helper methods for various data types, and `directory` or (`path`) might be one of them.

gnidan:
@CruzMolina can I call the test steps the ones listed in the original issue? It looks like there are decent steps there.

CruzMolina:
@gnidan removing the label and marking #862 again as containing the test steps.

gnidan:
Thanks!

PR #1337, opened 2018-10-13
Prevent debugger from quitting when trace is finished

hjaltman:
This pull request prevents the debugger from quitting when it finishes a trace, allowing for inspection of variables even after the trace is done.  It also adds storing the transaction object and receipt for later use; it then uses this to determine whether the transaction failed by inspecting the receipt, rather than by checking whether the stack is nonempty at finish.  This pull request also moves checking whether the trace is finished from the getter `session.finished` to the selector `trace.finished`; everything using `session.finished` has been updated to the new system (`session.failed` is similarly gone).  Note that at finish the trace is left in the state it was in just before finish -- so, e.g., the call stack is still nonempty -- and attempting to advance the trace in this state will simply result in it staying in place.  (At the command line, entering a command to advance will just display a message, and the interface will not even attempt to advance to advance the trace.)


packages/truffle-debugger/lib/controller/sagas/index.js
```diff
@@ -5,7 +5,8 @@ import { put, call, race, take, select } from 'redux-saga/effects';
 
 import { prefixName } from "lib/helpers";
 
-import * as trace from "lib/trace/sagas";
+import * as traceSagas from "lib/trace/sagas";
+import trace from "lib/trace/selectors";
```
gnidan: The convention is that sagas don't import selectors from an external app module. (This is why the controller selectors exist in the first place.)

packages/truffle-debugger/lib/evm/reducers.js
```diff
@@ -127,7 +127,9 @@ export function callstack(state = [], action) {
       return state.concat([ {binary} ]);
 
     case actions.RETURN:
-      return state.slice(0, -1); // pop
+      //pop the stack, unless that would leave it empty (this will only happen
+      //at the end when we want to keep the last one around)
```
gnidan: This looks like a candidate for a `HACK` marker

packages/truffle-debugger/lib/session/index.js
```diff
@@ -42,10 +42,10 @@ export default class Session {
   ready() {
     return new Promise( (accept, reject) => {
       this._store.subscribe( () => {
-        if (this.state.session == "ACTIVE") {
+        if (this.state.session.status == "ACTIVE") {
           accept()
-        } else if (typeof this.state.session == "object") {
-          reject(this.state.session.error);
+        } else if (typeof this.state.session.status == "object") {
```
gnidan: I would advise against making a state variable polymorphic, but in this case I can see why you did it.

packages/truffle-debugger/lib/data/selectors/index.js
```diff
@@ -279,6 +282,9 @@ const data = createSelectorTree({
         )
       ),
 
+      /**
+       * data.current.identifiers.native
+       */
```
gnidan: Thanks for adding the missing docstrings!

packages/truffle-debugger/lib/session/reducers.js
```diff
@@ -11,24 +11,45 @@ import * as actions from "./actions";
 export const WAITING = "WAITING";
 export const ACTIVE = "ACTIVE";
 export const ERROR = "ERROR";
-export const FINISHED = "FINISHED";
 
-export function session(state = WAITING, action) {
+export function status(state = WAITING, action) {
+  //theState to avoid name collision
```
gnidan: what's this comment about?

packages/truffle-debugger/lib/trace/sagas/index.js
```diff
@@ -25,7 +25,9 @@ function *waitForTrace() {
 export function *advance() {
   yield put(actions.next());
 
-  yield take(actions.TOCK);
+  debug("TOCK to take");
+  yield take([actions.TOCK,actions.END_OF_TRACE]);
```
gnidan: whitespace after comma

packages/truffle-debugger/test/transaction.js
```diff
@@ -0,0 +1,98 @@
...
+    let bugger = await Debugger.forTx(txHash, {
+      provider,
+      contracts: artifacts
+    });
+
+    let session = bugger.connect();
+
+    assert.ok(session.view(sessionSelector.transaction.receipt).status);
+  });
+});
```
gnidan: It might be good to test the new behavior here, namely that you can inspect variables after the trace finishes (in either success or failure case)

packages/truffle-debugger/lib/controller/sagas/index.js
hjaltman: I see, didn't realize, will fix!

packages/truffle-debugger/lib/evm/reducers.js
hjaltman: Heh, yeah.

packages/truffle-debugger/lib/session/index.js
hjaltman: I didn't do that -- that's preexisting!  This line is only marked as changed because it's now under `session.status` rather than just `session`.

packages/truffle-debugger/lib/data/selectors/index.js
hjaltman: No problem. :)

packages/truffle-debugger/lib/session/reducers.js
hjaltman: Oops.  Outdated comment from intermediate version, looks like.  Will fix.

packages/truffle-debugger/lib/trace/sagas/index.js
hjaltman: Oh, thanks.

packages/truffle-debugger/test/transaction.js
hjaltman: Hm, yeah, that could be worth doing.  So, a third test, then...

packages/truffle-debugger/lib/evm/reducers.js
hjaltman: I can add that to the other instance of this pattern as well...

packages/truffle-debugger/lib/session/index.js
gnidan: It's so much easier to criticize than it is to implement -___-

packages/truffle-debugger/test/transaction.js
gnidan: You could also just add the asserts to the existing two tests. Check that the status is correct _and_ check that the variables are what you expect them to be.

packages/truffle-debugger/lib/session/reducers.js
hjaltman: (Originally I had it as `session.state`, but that left the problem of what to call the variable `state`, so I renamed it to `theState`.  Eventually I realized that `status` was a suitable synonym for `state, avoiding the collision problem.)

packages/truffle-debugger/lib/controller/sagas/index.js
hjaltman: (I'm guess you probably told me that at one point while explaining the architecture, but I'd entirely forgotten...)

packages/truffle-debugger/test/transaction.js
hjaltman: A thought: Should the file be renamed then?  I named it `transaction.js` on the idea that, hey, this is where the tests related to transactiong-saving go.  Whereas this is rather different.

hjaltman: I guess I'll call it `endstate.js`.

hjaltman: Yeah in fact I think I'm going to make this a separate test, because it's actually rather different -- the two tests so far *don't* depend on advancing the debugger to the end.  (It took me a while to realize why this new test wasn't working!)  Better to make it separate.

gnidan:
:+1: you bet!

PR #1346, opened 2018-10-16
Adds step counter display to the debugger output

ccolorado:
* Helps keeping track of how far the debugger is into the execution.
* Helps out for finding a good place to put breakpoints after hitting a
  revert.

### DIFF
--- a/packages/truffle-core/lib/commands/debug.js
+++ b/packages/truffle-core/lib/commands/debug.js
@@ -54,6 +54,7 @@ var command = {
 
       var txHash = config._[0];
 
+      var jumpCount = 0;
       var lastCommand = "n";
       var enabledExpressions = new Set();
       var breakpoints = [];
@@ -151,7 +152,7 @@ var command = {
           var lines = splitLines(source);
 
           config.logger.log("");
-
+          config.logger.log("Jump Count #"+jumpCount++);
           config.logger.log(
             DebugUtils.formatRangeLines(lines, range.lines)
           );


ccolorado:
Should I have forked from 
e8a985212b23768a3217cd8e1960acebfc80f7c1 
(HEAD -1) instad of 9b84e8d633b13156f98833061c4f24c19df747fc (test are failing on this commit) 
?

gnidan:
@ccolorado We're looking at the build status right now. Once that's fixed, we can pull in the fixes from upstream onto your branch.

Also, requesting @hjaltman for review on this, since he might have more bandwidth than me right now.

packages/truffle-core/lib/commands/debug.js
```diff
@@ -54,6 +54,7 @@ var command = {
 
       var txHash = config._[0];
 
+      var jumpCount = 0;
```
hjaltman: So, this PR doesn't seem like a bad idea, but I'm leery of the implementation.  This variable you've added is basically reimplementing the `trace.index` selector, right?  I think it would make more sense to get the info from that existing variable (in this context, it should be accessible as `session.view(trace.index)`) rather than adding a second counter that does essentially the same thing.

hjaltman:
Sorry, two additional comments:

1. I don't think we want to actually print this every time.  It's a little too fine-grained, you know?  On the other hand, I get why you don't want to print it only on `;`.  Talking to @gnidan, his suggestion was to print it only on `p`; that way you don't need to do a `;` to see it, but we also don't print it out with every single thing.

2. Another suggestion from @gnidan -- when we print this, we probably want to print how many steps it's out of, as well.  That can be accessed at `session.view(trace.steps).length`.

gnidan:
Oh, another thing! If you want this, it's actually pretty easy to turn on and off:

```
debug> +!trace.index
```

```
debug> -!trace.index
```

ccolorado:
@gnidan @hjaltman I agree with the suggestions. I rather a flag for truffle debug. Something like:

`truffle debug [-t | --trace ] prints the trace steps for a debug run. `

This was you can store it on package.json.scripts like.
The only thing that I am not a fan is that if the debugger finishes, it just quits. So you would have to be very careful, ( or rather lucky ) to catch a glimpse of the last trace. 

I would ultimately love to have something like this :
 `truffle debug [-a | --auto-jump-to <NUMBER> ]  auto jumps to the step trace <NUMBER>. `

Thanks for all your comments.

gnidan:
> I would ultimately love to have something like this :
> `truffle debug [-a | --auto-jump-to <NUMBER> ] auto jumps to the step trace <NUMBER>.`

@ccolorado want to open an issue for this feature request?

On the topic of CLI flags, I think maybe we should generalize this...

```
truffle debug <txHash> --watch trace.index --watch trace.steps.length
```

_(**Note: `trace.steps.length` is not an existing selector right now... but that would be trivial to implement.)_

This way users can start the debugger, watching any selector(s) automatically. 

(This might need to be split up... `--watch-selector` vs. `--watch-expression`, but maybe not. Maybe we add all three, and `--watch` tries to guess if you are talking about a selector vs. an arbitrary JS expression? CLI option to watch a selector is probably the more common use case.)

Not sure if you want to restructure your PR to add this flag instead, or if you want to open this as an issue also, or whatever. Go to town :)

Thanks for your feedback!

hjaltman:
Interesting!  Based on @ccolorado's suggestion I have to wonder, maybe this should more generally be a possible new way to specify a breakpoint?

haltman-at:
Hey, so, this seems to have been left to languish, but I incorporated some of the ideas here into PR #1744 which has now been merged.  I'd forgotten about the idea of a step-count based breakpoint, though.  I'll need to write that down.  Thanks again, hope you find this helpful!

haltman-at:
Hello again @ccolorado!  I've now also submitted PR #1766, which adds a multiple-advance command.  Once that's merged, you'll (assuming you're willing to do some subtraction) be able to skip directly to the step you want.  (Note that steps are 1-indexed now, not 0-indexed, so watch out for that.)  So, I hope that addresses your remaining problems.  Thanks again for the suggestion!  Let us know if that suffices and perhaps we can close down this old one.

gnidan:
@ccolorado I'm closing this since we implemented something similar in #1744. Thanks for starting the discussion that resulted in that PR! Hopefully it meets your needs.

Issue #1090, opened 2018-07-06
Allow `truffle develop` to be configurable


gnidan:
Right now, running `truffle develop` uses default parameters defined [here](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/lib/environment.js#L120-L125)

It would be nice if this were configurable in `truffle.js`, perhaps under the `networks.develop` section.

fainashalts:
@CruzMolina is this issue resolved now? Wondering if we can close it.

CruzMolina:
Thanks for the reminder! Closing.

PR #1348, opened 2018-10-16
Configurable Truffle Develop

CruzMolina:
## Feature

Addresses Feature Request #1090.

Now users can configure the number of accounts generated at startup (`accounts`), amount of ether assigned to each account on startup (`defaultEtherBalance`), and specify the block time in seconds for automatic mining (`blockTime`) for `truffle develop` in their `truffle.js` or `truffle-config.js` file.

Example:

```
module.exports = {
  networks: {
    develop: {
      accounts: 5,
      defaultEtherBalance: 500,
      blockTime: 3
    }
  }
};
```

### DIFF
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -1,9 +1,9 @@
-var emoji = require('node-emoji');
-const mnemonicInfo = require('truffle-core/lib/mnemonics/mnemonic');
+var emoji = require("node-emoji");
+const mnemonicInfo = require("truffle-core/lib/mnemonics/mnemonic");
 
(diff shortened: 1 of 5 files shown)


packages/truffle-core/lib/commands/develop.js
```diff
@@ -38,26 +33,43 @@ var command = {
...
+      );
 
       c.start(done);
       c.on("exit", function() {
         process.exit();
       });
     });
   },
-  run: function (options, done) {
+  run: function(options, done) {
```
gnidan: Random note (not worth fixing now), but I believe most JS style-guides suggest `function (` instead of no-space.

PR #1349, opened 2018-10-16
Edit test for null build config

eggplantzzz:
Change the way truffle checks for a lack of build configuration.  The config now returns a `null` value instead of `undefined` when it is not set in the config.

Also update the travis build badge to indicate the build status of `next` and not `develop`.

### DIFF
--- a/packages/truffle-core/lib/build.js
+++ b/packages/truffle-core/lib/build.js
@@ -74,7 +74,7 @@ const Build = {
     // Duplicate build directory for legacy purposes
     options.destination_directory = options.build_directory;
 
-    if (builder === null) {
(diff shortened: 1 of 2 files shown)


packages/truffle-core/lib/build.js
```diff
@@ -74,7 +74,7 @@ const Build = {
     // Duplicate build directory for legacy purposes
     options.destination_directory = options.build_directory;
 
-    if (typeof builder === "undefined") {
+    if (builder === null) {
```
gnidan: Should we also allow `undefined`? Just in case? (ooh, a new github feature)
```suggestion
    if (builder === null || typeof builder === "undefined") {
```

PR #1351, opened 2018-10-17
Pluggable Config Packages

CruzMolina:
## Feature

One small step towards pluggability...

Now users can bring their own config packages to return their own custom config objects and (theoretically) add support for non-native truffle commands and features.

Example:

```
module.exports = {
  config: require("advanced-truffle-config"),
  ...
};
```

Potentially overkill, but mainly added to test out making workflows pluggable.

### DIFF
--- a/packages/truffle-config/index.js
+++ b/packages/truffle-config/index.js
@@ -327,6 +327,7 @@ Config.default = function() {
   return new Config();
 };
 
+
 Config.detect = function(options, filename) {
   var search;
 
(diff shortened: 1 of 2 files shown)


packages/truffle-config/index.js
```diff
@@ -344,14 +344,17 @@ Config.detect = function(options, filename) {
 };
 
 Config.load = function(file, options) {
-  var config = new Config();
+  var static_config = originalrequire(file);
```
gnidan:
```suggestion
  var staticConfig = originalrequire(file);
```

packages/truffle-config/index.js
```diff
@@ -344,14 +344,17 @@ Config.detect = function(options, filename) {
...
+
+  var config = static_config.config
```
gnidan:
```suggestion
  var config = staticConfig.config
```

packages/truffle-config/index.js
```diff
@@ -344,14 +344,17 @@ Config.detect = function(options, filename) {
...
+    ? new static_config.config()
```
gnidan:
```suggestion
    ? new staticConfig.config()
```

Sorry I'm just playing around with this new button.

packages/truffle-config/index.js
```diff
@@ -344,14 +344,17 @@ Config.detect = function(options, filename) {
...
+    : new Config();
 
   config.working_directory = path.dirname(path.resolve(file));
 
   // The require-nocache module used to do this for us, but
   // it doesn't bundle very well. So we've pulled it out ourselves.
   delete require.cache[Module._resolveFilename(file, module)];
-  var static_config = originalrequire(file);
```
gnidan: We need to be sure that the `delete` happening before the `originalrequire` is not necessary. Might be safer just to do the loading after the `delete`, like it is currently.

PR #1356, opened 2018-10-19
Opt in analytics

fainashalts:
This pull request does two things: 
1) Provides a "truffle config (--enable-analytics or --disable-analytics optional) command for users to set analytics preferences at the user level. If passed without options, user is prompted to choose whether they would like to enable analytics.
2) Sends analytics events to Google Analytics for users who have enabled this feature, in the following circumstances: 
- truffle init
- truffle unbox
- process exiting due to an error. Possible errors this will cover: wrong node version, general help/task error, missing configuration file/truffle error, a specific numbered error, and other errors that may be thrown such as "No network specified."

The version is included in the label of the event being sent to Google Analytics.


packages/truffle-core/lib/services/google-analytics.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+    if(visitor) {
+      visitor.event(eventObject, function(err) {
+      });
+    }
+    return true;
+  }
+}
+
+process.on('message', function(eventObject) {
```
fainashalts: I put this here instead of making a separate executable file. If this is a bad idea and should be put elsewhere please let me know.

packages/truffle-core/lib/commands/config.js
```diff
@@ -0,0 +1,41 @@
+const command = {
+  command: "config",
+  description: "set user-level configuration options",
```
eggplantzzz: Can you capitalize the first S?  This will display in the help.

packages/truffle-core/lib/commands/config.js
```diff
@@ -0,0 +1,41 @@
...
+  config: {
+    usage: "truffle config --enable-analytics",
```
eggplantzzz: The value for usage should probably be something like `truffle config <analytics_option>` if we are following the conventions in the command files, right?  Are we going to eventually add other options in here for this command?  We might want to consider having a format like `truffle config --analytics off|on`...what do you think?

packages/truffle-core/lib/commands/config.js
```diff
@@ -0,0 +1,41 @@
...
+    if(options.enableAnalytics) {
+      setAnalytics = googleAnalytics.setAnalytics(true);
+    } else if (options.disableAnalytics){
+      setAnalytics = googleAnalytics.setAnalytics(false);
+    } else {
+      setAnalytics = googleAnalytics.setUserConfigViaPrompt();
+      setAnalytics.then(() => done()).catch((err)=> err);
+    }
+
+    if(setAnalytics === true) {
```
eggplantzzz: You could leave this if statement off and just put a return before the last else's setAnalytics?  On first reading it it feels a little bit cryptic.  Like you could do
```
else {
  setAnalytics = ....
  return setAnalytics.then(() =>... => done());
}
return done();
```

packages/truffle-box/box.js
```diff
@@ -8,7 +8,7 @@ var Box = {
   unbox: function(url, destination, options) {
     options = options || {};
     options.logger = options.logger || { log: () => {} };
-
+
```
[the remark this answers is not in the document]
fainashalts: I know! I tried to remove this like 3 times... :/

packages/truffle-core/lib/commands/config.js
fainashalts: I actually had it that way originally, but when the prompt gets run the "return done()" gets called before  it has time to resolve. Setting the if statement avoids that. Is there a different solution maybe?

packages/truffle-core/lib/commands/init.js
```diff
@@ -17,6 +17,11 @@ var command = {
     var Config = require("truffle-config");
     var OS = require("os");
     var UnboxCommand = require("./unbox");
+    const path = require('path');    
+    const cp = require('child_process');
+    const child = cp.fork(path.join(__dirname,  "../services/google-analytics"));
+
+    child.send({ec: "initialization", ea: "truffle init"})
```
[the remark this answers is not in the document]
fainashalts: Love it! I feel like if I look at my own code too long I'm bound to miss stuff like this.

packages/truffle-core/test/google-analytics.js
```diff
@@ -0,0 +1,98 @@
...
+const configCommand = require('../lib/commands/config');
+const Configstore = require('configstore');
+
+
+describe("googleAnalytics", function() {
+  beforeEach(() => {
+    sinon.stub(inquirer, "prompt").returns({ then: () => 1 });
+  });
+  afterEach(() => {
+    inquirer.prompt.restore();
```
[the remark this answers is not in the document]
fainashalts: So I am, actually. In the sense that I don't want the inquirer prompt to run, which happens in the background of several of these tests. This was a blanket way to ensure it didn't happen.

packages/truffle-core/lib/services/google-analytics.js
```diff
@@ -0,0 +1,157 @@
...
+  * set user-level options for analytics
+  * @param {bool} analyticsBool
+  * @param {Object} userConfig
+  */
+  setAnalytics: function(analyticsBool) {
+    if(analyticsBool === true) {
+      this.setUserId();
+      userConfig.set({ 'enableAnalytics': true, 'analyticsSet': true, 'analyticsMessageDateTime': Date.now() });
+    } else {
+      userConfig.set({ 'enableAnalytics': false, 'analyticsSet': true, 'analyticsMessageDateTime': Date.now() });
```
eggplantzzz: How about something like
```
setAnalytics: function(enableUsersAnalytics) {
  if (enableUsersAnalytics) this.setUserId();
  userConfig.set({ 
    enableAnalytics: enableUsersAnalytics,
    blah: blah,
    blah....
  })...
```?

File: packages/truffle-core/lib/services/google-analytics.js [hunk 9]
```diff
@@ -0,0 +1,157 @@
...
+        this.setAnalytics(true);
+      }
+    } else if(userConfig.get('analyticsSet') && !userConfig.get('enableAnalytics') && process.stdin.isTTY === true) {
+      let answer = await inquirer.prompt(analyticsEnable)
+      if(answer.analyticsEnable) {
+        this.setAnalytics(true);
+      } else {
+        this.setAnalytics(false);
+      }
+    } else {
```
eggplantzzz: You can probably drop this else.

packages/truffle-core/lib/services/google-analytics.js
```diff
@@ -0,0 +1,157 @@
...
+  appVersion = "(unbundled) " + "version " + version.core;
+  truffleAnalyticsId = "UA-83874933-7";
+}
+
+
+const analyticsInquiry = [
+  {
+    type : "list",
+    name : "analyticsInquiry",
+    message : "Would you like to enable analytics for your Truffle projects? Doing so will allow us to make sure Truffle is working as expected and address any bugs more efficiently.",
```
eggplantzzz: The grammar is slightly weird near the end...maybe `Doing so will allow us to make sure Truffle is working as expected and help us address any bugs more efficiently.`

packages/truffle-core/lib/services/google-analytics.js
```diff
@@ -0,0 +1,157 @@
...
+  */
+  setUserConfigViaPrompt: async function() {
+    if(!userConfig.get('analyticsSet') && process.stdin.isTTY === true) {
+      let answer = await inquirer.prompt(analyticsInquiry)
+      if(answer.analyticsInquiry === analyticsInquiry[0].choices[0]) {
+        this.setAnalytics(true);
+      } else {
+        this.setAnalytics(false);
+      }
+    } else if (userConfig.get('analyticsSet') && userConfig.get('enableAnalytics') && process.stdin.isTTY === true) {
```
eggplantzzz: If ya like you could do something like 
```
} else if (userConfig.get('analyticsSet') && process.stdin.isTTY) {
  const analyticsEnabled = userConfig.get('enableAnalytics');
  let answer = analyticsEnabled ? await inquirer.prompt(analyticsDisable) : await inquirer.prompt(analyticsEnable);
  answer.analyticsDisable ? this.setAnalytics(false) : this.setAnalytics(true);
}
```
It would be a little slimmer.

packages/truffle-core/test/google-analytics.js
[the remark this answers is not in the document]
CruzMolina: Oh, that's right. Had to do the same for my truffle-box tests.

fainashalts:
I see the failed Travis tests, pretty sure this is happening because there's a constructor that needs to be stubbed/mocked. Working on that now!

packages/truffle-core/lib/commands/config.js
eggplantzzz: Ahh you are right, I missed that the callback will get called before the async method is done.

fainashalts: Yeah, I tried that but then my test throws "TypeError: done is not a function" -- thoughts on how to address? This was a workaround for that issue.

packages/truffle/cli.webpack.config.js
```diff
@@ -66,6 +67,11 @@ module.exports = {
       "BUNDLE_CHAIN_FILENAME": JSON.stringify("chain.bundled.js")
     }),
 
+    new webpack.DefinePlugin({
+      "BUNDLE_VERSION": JSON.stringify(pkg.version),
+      "BUNDLE_ANALYTICS_FILENAME": JSON.stringify("analytics.bundled.js")
+    }),
```
gnidan: ah I think this should be merged into the one above... we can get rid of the extra `BUNDLE_VERSION` and just include `BUNDLE_ANALYTICS_FILENAME` in the same object

packages/truffle-core/lib/services/google-analytics.js
eggplantzzz: I really dislike abbreviated variable names :p

packages/truffle-core/cli.js
```diff
@@ -38,16 +42,21 @@ if (userWantsGeneralHelp) {
 command.run(inputArguments, options, function(err) {
```
eggplantzzz: I think you are probably already on it Faina, but you are going to have the version sent to GA as well, right?  Just wanted to add this comment here so we don't lose it.

fainashalts: Yes, I send it! This is currently being added on inside of google-analytics.js before sending the event (it will be part of the error string if there is one, otherwise sent as an event label for init and unbox), though I'm refactoring that file to move the child process out into its own file.

packages/truffle-core/test/user-config.js
```diff
@@ -0,0 +1,33 @@
+const assert = require("chai").assert;
+const sinon = require("sinon");
+const googleAnalytics = require('../lib/services/google-analytics');
+const configCommand = require('../lib/commands/config');
+const inquirer = require("inquirer");
+
+describe("config", function() {
+  describe("#run", function() {
+  	beforeEach(() => {
```
eggplantzzz: Ya got some indentation problems here

packages/truffle-core/test/user-config.js
```diff
@@ -0,0 +1,33 @@
...
+    	sinon.stub(googleAnalytics, "setUserConfigViaPrompt").resolves();
+	  });
+	  afterEach(() => {
+	    inquirer.prompt.restore();
+	    googleAnalytics.setAnalytics.restore();
+	    googleAnalytics.setUserConfigViaPrompt.restore();
+	    sinon.restore();
+	  });
+    it("calls googleAnalytics.setAnalytics() when provided with enableAnalytics option", function() {
+      configCommand.run({enableAnalytics: true}, ()=> {});
```
eggplantzzz: Stupid thing but can we add some breathing room? i.e. `() => {}`

packages/truffle-core/test/google-analytics.js
```diff
@@ -0,0 +1,98 @@
...
+    Configstore.prototype.get.restore();
+    Configstore.prototype.set.restore();
+    googleAnalytics.setUserConfigViaPrompt.restore();
+    ua.Visitor.prototype._enqueue.restore();
+    sinon.restore();
+  });
+  describe("#setUserId", function(){
+    it("sets a userId if one does not exist", function() {
+      googleAnalytics.setUserId();
+   		sinon.assert.calledOnce(Configstore.prototype.get);
```
eggplantzzz: indentation

packages/truffle-core/lib/services/analytics/main.js
```diff
@@ -0,0 +1,9 @@
+const googleAnalytics = require("./google.js");
+
+process.on('message', function(eventObject) {
```
eggplantzzz: indentation

packages/truffle-core/lib/services/analytics/index.js
```diff
@@ -0,0 +1,18 @@
+require("./main.js");
+const analytics = {
+	send: function(eventObject) {
```
eggplantzzz: indentation

packages/truffle-core/lib/services/analytics/google.js
```diff
@@ -0,0 +1,162 @@
...
+  */
+  sendAnalyticsEvent: function(eventObject, callback) {
+    let visitor = this.setPersistentAnalyticsData();
+    if(eventObject["el"]) {
+      eventObject["el"] = eventObject["el"] + " (" + appVersion + ")";
+    } else {
+      eventObject["el"] = appVersion;
+    }
+    if(visitor) {
+      visitor.event(eventObject, function(err) {
```
eggplantzzz: Might as well put the `});` on the same line like...
`(error) => {}`

CruzMolina:
@fainashalts to help resolve the stylistic issues that @eggplantzzz is pointing out, it might help to merge `next` into your branch again to test out using `eslint` and `prettier`

packages/truffle-core/lib/services/analytics/google.js
```diff
@@ -0,0 +1,168 @@
...
+      let visitor = ua(truffleAnalyticsId, { uid: userId });
+    }
+    return visitor;
+  },
+
+  /**
+   * send event to Google Analytics
+   * @param {Object}
+   */
+  sendAnalyticsEvent: function(eventObject) {
```
gnidan: Can we make this function `async` and return true or false based on whether the event was actually sent to GA?

Issue #1354, opened 2018-10-18
[BETA] Truffle Install fails


thekevinbrown:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Trying to run `truffle install zeppelin` errors

## Steps to Reproduce

```
$ npm install -g truffle@beta
$ truffle init
$ truffle install zeppelin
TypeError: Cannot read property 'getLockfileURI' of undefined
    at [my node install]/lib/node_modules/truffle/build/webpack:/~/ethpm/lib/installer.js:64:1
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

## Expected Behavior

Zeppelin library should be installed.

## Actual Results

```
TypeError: Cannot read property 'getLockfileURI' of undefined
    at [my node install]/lib/node_modules/truffle/build/webpack:/~/ethpm/lib/installer.js:64:1
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

## Environment

* Operating System: OS X 10.14 (18A391)
* Ethereum client: `ganache-cli`
* Truffle version (`truffle version`): Truffle v5.0.0-beta.1 (core: 5.0.0-beta.1)
* Solidity version: Solidity v0.4.25 (solc-js)
* node version (`node --version`): v10.10.0
* npm version (`npm --version`): 6.4.1

eggplantzzz:
This feature is kind of incomplete at the moment...we are currently working to revamp EthPM and hope that within the next couple of months we'll be able to have something useful in place!

eggplantzzz:
Thanks for the PR!  I talked with some other members of the team and we're planning to try and look into fixing this.  It might only be working for specific packages as some of them mentioned that there might be some people successfully using this feature.  It might work for just a selection of packages...it might be more correct to say that we are not totally sure what its current state is.  However I think we are going to try and fix this feature.

thekevinbrown:
That sounds great, is there anything I can do to help here? Does it still work in v4? I might be able to find some time before devcon to chip in.

eggplantzzz:
Hmm, I think you might take it up with @gnidan...I'm not sure we've fully taken stock of what it will take to fix.

PR #1357, opened 2018-10-20
Remove install command as per issue #1354

thekevinbrown:
#1354 

Since the command doesn't work, and is expected not to work, it'd be best to remove it so as not to confuse users who think it should.

If there's anything else that needs doing to remove this feature from the beta branch just let me know, keen to help.


gnidan:
Hey @thekevinbrown!

I don't think we should remove this! I think we should fix it! It's true that we're working on EthPM v2, but in the meantime, EthPM v1 should still work to the extent that it can.

Does `truffle install` work for you on Truffle v4? Or do you have the same problem?

thekevinbrown:
I'm not sure, I'm using v5 on the project I'm on, and if it works or not in v4 that wouldn't really help me.

Is EthPM v1 expected to work with Truffle 5, or not? I see the following potential resolutions here:

1. If EthPM v1 is expected to work in the meantime, we fix the problem in the issue.
2. If EthPM v1 is not expected to work in the meantime, we remove `truffle install` as it's broken and just confuses people being there.

Of course once the v2 implementation is done this won't be an issue, I'm just trying to arrive at a stable place for Truffle v5 in the meantime, particularly if it's a few months away.

eggplantzzz:
@thekevinbrown Sorry for not responding about helping out, it took a while to track down and figure out what was going on.  It turns out that the problem was not in truffle but in a package related to EthPM.  The install command should be fixed in the next week or so for the beta version.

PR #1364, opened 2018-10-23
Fix .eslintrc

CruzMolina:
Remove unnecessary implicit dependency (`eslint-config-standard`) and redundant disabled option declarations.

(Credits to @adrianmcli for pointing this out! 😎)


adrianmcli:
Ah I see. So you don't want to use Standard just yet I guess?

PR #1367, opened 2018-10-23
ABC w/ Minimal Config

CruzMolina:
## Feature

Allows users to ABC for community-maintained Truffle features with minimal configuration in their truffle config file.

Example:

```
module.exports = {
  commands: {
    upgrade: require("truffle-upgrade"),
    analyze: require("truffle-analyze"),
    coverage: require("truffle-solc-coverage")
  },
  ...
};
```

Expected Behavior:

```
truffle upgrade MyContract.sol

...(does whatever truffle-upgrade would do)...

truffle analyze MyOtherContract.sol

...(does whatever truffle-analyze would do)...
```


packages/truffle-core/lib/command.js
```diff
@@ -14,9 +14,13 @@ function Command(commands) {
   });
 
   this.args = args;
-};
+}
+
+Command.prototype.getCommand = function(inputStrings, config, noAliases) {
```
gnidan: We can just get rid of `noAliases` because it comes from `config` anyway.

I think we can even get away with doing `function (inputStrings, options)` and doing `Config.detect()` inside the function here. There might be a small performance loss, but we can always solve that inside `detect` itself.

CruzMolina:
Can't wait to get this merged in. Super stoked about the `Config.search` helper. Should allow me to trudge on making all the things configurable.

packages/truffle-config/index.js
```diff
@@ -326,16 +353,22 @@ Config.default = function() {
   return new Config();
 };
 
-Config.detect = function(options, filename) {
-  var search;
+Config.search = function(options, filename) {
```
gnidan: Cool!

packages/truffle-core/test/commands.js
```diff
@@ -12,29 +12,29 @@ describe("Commander", function() {
   });
 
   it("will infer commands based on initial letters entered", function() {
-    var actualCommand = commander.getCommand("m").command;
+    var actualCommand = commander.getCommand("m", _).command;
```
gnidan: Wait, won't `_` be undefined?

packages/truffle/test/scenarios/commands/help.js
```diff
@@ -1,17 +1,27 @@
 const assert = require("assert");
 const CommandRunner = require("../commandrunner");
 const MemoryLogger = require("../memorylogger");
-let config = {};
+const sandbox = require("../sandbox");
+const path = require("path");
 
-describe("truffle help", () => {
+describe.only("truffle help", () => {
```
gnidan: Gotta get rid of the `.only`!

PR #1371, opened 2018-10-24
Fix undefined sourcePath in debugger

MickdeGraaf:
Fixes: https://github.com/trufflesuite/truffle-debugger/issues/51

### DIFF
--- a/packages/truffle-debugger/lib/session/sagas/index.js
+++ b/packages/truffle-debugger/lib/session/sagas/index.js
@@ -1,7 +1,7 @@
 import debugModule from "debug";
 const debug = debugModule("debugger:session:sagas");
 
-import { cancel, call, all, fork, take, put } from 'redux-saga/effects';
+import { cancel, call, all, fork, take, put } from "redux-saga/effects";
 
 import { prefixName } from "lib/helpers";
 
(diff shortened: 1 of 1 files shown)


packages/truffle-debugger/lib/session/sagas/index.js
```diff
@@ -15,98 +15,104 @@ import * as web3 from "lib/web3/sagas";
...
-    yield *solidity.addSource(source, sourcePath, ast);
+  for (let i = 0; i < sources.length; i++) {
+    const sourceData = sources[i];
+    if (sourceData !== undefined && sourceData !== null) {
+      yield* solidity.addSource(
+        sourceData.source,
+        sourceData.sourcePath,
+        sourceData.ast
+      );
+    }
```
gnidan: But why is it undefined?

MickdeGraaf: Something to do with interfaces I believe. When importing something goes wrong with them. I dont know. Saw it somewhere as a fix and there was no fix merged anywhere so I created a pull request.

hjaltman: (Btw, I'm guessing most of these nonfunctional changes due to that linter we're now using?  Man, this is going to make reviewing PR's a bit of a pain for a while...)

hjaltman: After looking deeper at the code... I'm still pretty confused as to why this is occurring!  I could maybe see it if the `files` option was used, but it doesn't seem like the cases causing the problem involve that.

PR #1377, opened 2018-10-24
Quick bug fix in /build

eggplantzzz:
Add check to see if builder is null so as not to execute a method on it and add a test


eggplantzzz:
To make sure it doesn't come back!

fainashalts:
Nice! This is to handle undefined errors in build?

eggplantzzz:
Yes, I think it was erroring if it didn't have a build script specified in the config.

PR #1380, opened 2018-10-25
Allow configuration of allowed_extensions for the migrations (fixes #1361)

### LINKED ISSUE #1361: [v5] Allow configuration of allowed_extenstions for the migrations
I'm using https://github.com/ethereum-ts/truffle-typechain-example to get everything into typescript but `truffle migrate` cannot find any migrations ending in .ts due to thus config option: https://github.com/trufflesuite/truffle/blob/46fc7bdb9fde7cc3fa99b94fef04b1d6c04c38f7/packages/truffle-migrate/index.js#L30

There is however no option to configure this in the truffle.js file.

Crosslink: https://github.com/ethereum-ts/truffle-typechain-example/issues/2

roderik:
- I added a separate config function to truffle-config, same naming and signature scheme as the test regexp
- I added the new config regexp as second choice to truffle-migrate. While there are no usages of the option version, I did not wan to inadvertently break any code depending on this option.
- separate commits because my editor wanted to reformat every line of code, yay prettier :)

### DIFF
--- a/packages/truffle-migrate/index.js
+++ b/packages/truffle-migrate/index.js
@@ -4,7 +4,7 @@ const async = require("async");
 const expect = require("truffle-expect");
 const Config = require("truffle-config");
 const Reporter = require("truffle-reporters").migrationsV5;
-const Migration = require('./migration.js');
+const Migration = require("./migration.js");
 
 /**
  *  This API is consumed by `truffle-core` at the `migrate` and `test` commands via
@@ -14,11 +14,11 @@ const Migrate = {
   Migration: Migration,
   reporter: null,
 
-  launchReporter: function(){
+  launchReporter: function() {
     Migrate.reporter = new Reporter();
   },
 
-  acceptDryRun: async function(){
+  acceptDryRun: async function() {
     return Migrate.reporter.acceptDryRun();
   },
 
@@ -31,7 +31,9 @@ const Migrate = {
 
       let migrations = files
         .filter(file => isNaN(parseInt(path.basename(file))) == false)
-        .filter(file => path.extname(file).match(options.allowed_extensions) != null)
+        .filter(
+          file => path.extname(file).match(options.allowed_extensions) != null
+        )
         .map(file => new Migration(file, Migrate.reporter, options));
 
       // Make sure to sort the prefixes as numbers and not strings.
@@ -58,7 +60,7 @@ const Migrate = {
       "network",
       "network_id",
       "logger",
-      "from", // address doing deployment
+      "from" // address doing deployment
     ]);
 
     if (options.reset == true) {
@@ -85,7 +87,9 @@ const Migrate = {
       }
 
       if (options.to) {
-        migrations = migrations.filter(migration => migration.number <= options.to);
+        migrations = migrations.filter(
+          migration => migration.number <= options.to
+        );
       }
 
       self.runMigrations(migrations, options, callback);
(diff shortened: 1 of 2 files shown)


packages/truffle-migrate/index.js
```diff
@@ -23,10 +23,14 @@ const Migrate = {
   },
 
   assemble: function(options, callback) {
+    var config = Config.detect(options);
```
sneko: @roderik , when building on my own I get `ReferenceError: Config is not defined`

sneko: Missing `const Config = require("truffle-config");` (note the const)

sneko:
@roderik after fixing what I mentioned in my comments I'm still getting errors due to TypeScript syntax.

Here one of my files:
```
var Migrations = artifacts.require("./Migrations.sol");

module.exports = async function(deployer, network, accounts) {
  await deployer.deploy(Migrations);
} as Truffle.Migration;

// because of https://stackoverflow.com/questions/40900791/cannot-redeclare-block-scoped-variable-in-unrelated-files
export {};
```

I get the following error when migrating:
```
} as Truffle.Migration;
  ^^

SyntaxError: Unexpected identifier
```

If I remove the ` as Truffle.Migration` it shows me the next issue:
```
export {};
^^^^^^

SyntaxError: Unexpected token export
    at Object.createScript (vm.js:80:10)
```

To solve this my file needs to look like:
```
var Migrations = artifacts.require("./Migrations.sol");

module.exports = async function(deployer, network, accounts) {
  await deployer.deploy(Migrations);
};
```

Like an ES6 file...

Do you know why the parser is not using TypeScript?

Thanks,

roderik:
@sneko, you are probably missing the require("ts-node/register"); line mentioned in the setup at https://github.com/ethereum-ts/truffle-typechain-example. Thanks for catching the missing require, forgot to copy paste that line in the commit :)

sneko:
@roderik I’m actually requiring the tsnode/register... hmmm I don’t understand what could be the problem.

I’m using Truffle 5 and my settings are stored in the « truffle-confit.js ».

I will try to figure out tomorrow morning what’s going on. (Including cloning and testing the example)

sneko:
@roderik the `truffle-typechain-example` is working well for migrations if I build on my own truffle (including your PR).

But if I add this line in the `truffle.js`:
`migrations_file_extension_regexp: /.*\.ts$/,`

I get the same error than in my project:
```
} as Truffle.Migration;
  ^^

SyntaxError: Unexpected identifier
```
 😢

CruzMolina:
Thanks again for doing this @Roderik. @sneko I did some playing around with replicating your error and I believe it is a configuration issue on the ts side (typechain, tsconfig).

Since this is a simple enough feature addition, I think we're going to move ahead with supporting this after I add some tests to ensure coverage.

CruzMolina:
@sneko did a little more replication using the [truffle-typechain-example](https://github.com/ethereum-ts/truffle-typechain-example) repo.

I was able to get the example working using their current setup (package.json, tsconfig, etc). I was also able to get things working using Roderik's current workaround [here](https://github.com/ethereum-ts/truffle-typechain-example/issues/2#issuecomment-431696452).

sneko:
@CruzMolina yeah the workaround is working.

But did you succeed in replacating the issue I have?

_Building Truffle with this pull request and injecting it inside the `node_modules` directory on `truffle-typechain-example` (do not forget to add `migrations_file_extension_regexp: /.*\.ts$/,` inside the Truffle config file)_

CruzMolina:
Ah, I see what you mean @sneko. After reviewing the codebase, it looks like `truffle-migrate` exports the migration files to a js parser in `truffle-require`. Using `truffle-typings` with `typechain`, you don't have to add to or change `migrations_file_extension_regexp` to parse .ts (`tsc` will transpile it for you to .js). I was able to get the example repo working without Roderik's workaround. You don't have to specify a different `migrations_build` folder, you can just use the given `tsconfig` and the example migrations .ts file to base your own migration files.

CruzMolina:
So after diving into this, it looks like this change doesn't immediately further .ts integration or ethereum-ts support (although #1385 should help a teeny bit towards that end). However, it does make an aspect of `truffle-config` configurable in a way that could be useful more long term. Let me know your thoughts on this @gnidan . (Possibly an argument to get back to work on the `pluggable-truffle` feature branch... 😂 )

fainashalts:
+1 Looks good to me!

PR #1385, opened 2018-10-26
Native .ts support for truffle tests

CruzMolina:
## Feature

Allows native .ts support for truffle tests.

Currently, users have to add a workaround to their config file in order to support .ts test files.

For example:

```
module.exports = {
  test_file_extension_regexp: /.*\.ts$/,
  ....
};
```

Steps to Test:

https://github.com/ethereum-ts/truffle-typechain-example#setup (minus the config workaround)

### DIFF
--- a/packages/truffle-config/index.js
+++ b/packages/truffle-config/index.js
@@ -2,10 +2,10 @@ var _ = require("lodash");
 var path = require("path");
 var Provider = require("truffle-provider");
 var TruffleError = require("truffle-error");
-var Module = require('module');
+var Module = require("module");
 var findUp = require("find-up");
 var originalrequire = require("original-require");
-const Configstore = require('configstore');
+const Configstore = require("configstore");
 
 var DEFAULT_CONFIG_FILENAME = "truffle.js";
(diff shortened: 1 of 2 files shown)


packages/truffle-config/index.js
```diff
@@ -104,7 +108,7 @@ function Config(truffle_directory, working_directory, network) {
       transform: resolveDirectory
     },
     test_file_extension_regexp: function() {
-      return /.*\.(js|es|es6|jsx|sol)$/;
+      return /.*\.(js|ts|es|es6|jsx|sol)$/;
```
gnidan: This will just break if the user doesn't set up the typechain stuff, right?

Could we add error messaging when the user has `.ts` files, to point them in the right direction to set up everything they need?

CruzMolina: Ran a couple tests and it looks like we support .ts test files just fine. Will work on adding this to our automated tests to ensure coverage.

CruzMolina:
Looks like this needs some more work for the geth build and possibly one of the coverage builds.

CruzMolina:
Woohoo, I think this is ready! 🤞

PR #1398, opened 2018-10-31
Fix coverage & CompilerSupplier Method

CruzMolina:
## Bug Fix

The Travis Coverage build was failing due to timeouts. In the process of fixing the build, it was discovered that a CompilerSupplier method was missing a conditional check.

New expected Behavior:

Coverage build should pass on the current `next` branch and users should be able to successfully specify a cached version of solc to compile with.


eggplantzzz:
Sweet, code coverage report!

PR #1399, opened 2018-11-01
Fix error message

lrettig:
Add missing variable to make error clearer


eggplantzzz:
This seems like a good idea to me!

gnidan:
Thanks for the fix!

PR #1404, opened 2018-11-02
hdwallet provider: Update deps, webpack, fix test

benjamincburns:
No more native deps required.

Should make @davidburela happy :-D


benjamincburns:
Sadly the build is still friggin' huge (1.8M JS file, 5.4M source mapping file). We'll definitely need to prune the build sometime soon, but this should be a good stopgap for now.

PR #1409, opened 2018-11-05
web3 1.0.0-beta.36 - enables ABIv2 struct params

benjamincburns:
requires contracts to use `pragma experimental ABIEncoderV2;`

### DIFF
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -21,7 +21,7 @@
     "solc": "0.4.25",
     "temp": "^0.8.3",
     "truffle-blockchain-utils": "^0.0.6-beta.0",
-    "web3": "1.0.0-beta.35"
+    "web3": "1.0.0-beta.36"
   },
   "dependencies": {
     "async": "2.6.1",
--- a/packages/truffle-contract/package.json
+++ b/packages/truffle-contract/package.json
@@ -35,10 +35,10 @@
     "truffle-contract-schema": "^3.0.0-beta.1",
     "truffle-error": "^0.0.3",
     "uglify-es": "^3.3.9",
-    "web3": "1.0.0-beta.35",
-    "web3-core-promievent": "1.0.0-beta.35",
-    "web3-eth-abi": "1.0.0-beta.35",
-    "web3-utils": "1.0.0-beta.35"
+    "web3": "1.0.0-beta.36",
+    "web3-core-promievent": "1.0.0-beta.36",
+    "web3-eth-abi": "1.0.0-beta.36",
+    "web3-utils": "1.0.0-beta.36"
   },
   "devDependencies": {
     "async": "2.6.1",
--- a/packages/truffle-core/package.json
+++ b/packages/truffle-core/package.json
@@ -49,7 +49,7 @@
     "truffle-resolver": "^5.0.0-beta.1",
     "truffle-solidity-utils": "^1.1.2",
     "truffle-workflow-compile": "^2.0.0-beta.1",
-    "web3": "1.0.0-beta.35",
+    "web3": "1.0.0-beta.36",
     "yargs": "^8.0.2"
   },
   "bin": {
--- a/packages/truffle-debugger/package.json
+++ b/packages/truffle-debugger/package.json
@@ -33,8 +33,8 @@
     "truffle-code-utils": "^1.1.2-beta.0",
     "truffle-expect": "^0.0.4",
     "truffle-solidity-utils": "^1.1.2",
-    "web3": "1.0.0-beta.35",
-    "web3-eth-abi": "1.0.0-beta.35"
+    "web3": "1.0.0-beta.36",
+    "web3-eth-abi": "1.0.0-beta.36"
   },
   "devDependencies": {
     "async": "2.6.1",
--- a/packages/truffle-deployer/package.json
+++ b/packages/truffle-deployer/package.json
@@ -33,7 +33,7 @@
     "mocha": "5.2.0",
     "truffle-reporters": "^1.0.0-beta.1",
     "truffle-workflow-compile": "^2.0.0-beta.1",
-    "web3": "1.0.0-beta.35"
+    "web3": "1.0.0-beta.36"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/truffle-external-compile/package.json
+++ b/packages/truffle-external-compile/package.json
@@ -26,7 +26,7 @@
     "glob": "^7.1.2",
     "truffle-contract-schema": "^3.0.0-beta.1",
     "truffle-expect": "^0.0.4",
-    "web3-utils": "1.0.0-beta.35"
+    "web3-utils": "1.0.0-beta.36"
   },
   "devDependencies": {
     "chai": "^4.1.2",
--- a/packages/truffle-hdwallet-provider/package.json
+++ b/packages/truffle-hdwallet-provider/package.json
@@ -34,7 +34,7 @@
     "ganache-core": "^2.2.1",
     "js-scrypt": "^0.2.0",
     "mocha": "^5.1.1",
-    "web3": "1.0.0-beta.33",
+    "web3": "1.0.0-beta.36",
     "web3-provider-engine": "git+https://github.com/cgewecke/provider-engine.git#web3-one",
     "webpack": "^4.24.0",
     "webpack-cli": "^3.1.2"
--- a/packages/truffle-migrate/package.json
+++ b/packages/truffle-migrate/package.json
@@ -28,7 +28,7 @@
     "truffle-expect": "^0.0.4",
     "truffle-reporters": "^1.0.0-beta.1",
     "truffle-require": "^2.0.0-beta.1",
-    "web3": "1.0.0-beta.35"
+    "web3": "1.0.0-beta.36"
   },
   "devDependencies": {
     "mocha": "5.2.0"
--- a/packages/truffle-provider/package.json
+++ b/packages/truffle-provider/package.json
@@ -24,7 +24,7 @@
   "homepage": "https://github.com/trufflesuite/truffle-provider#readme",
   "dependencies": {
     "truffle-error": "^0.0.3",
-    "web3": "1.0.0-beta.35"
+    "web3": "1.0.0-beta.36"
   },
   "devDependencies": {
     "ganache-cli": "6.1.8",
--- a/packages/truffle-reporters/package.json
+++ b/packages/truffle-reporters/package.json
@@ -12,7 +12,7 @@
   "dependencies": {
     "node-emoji": "^1.8.1",
     "ora": "^3.0.0",
-    "web3-utils": "1.0.0-beta.35"
+    "web3-utils": "1.0.0-beta.36"
   },
   "gitHead": "07967b472112f1c771ebbc90319781454cf9331b"
 }
--- a/packages/truffle-require/package.json
+++ b/packages/truffle-require/package.json
@@ -29,7 +29,7 @@
     "original-require": "1.0.1",
     "truffle-config": "^1.1.0-beta.1",
     "truffle-expect": "^0.0.4",
-    "web3": "1.0.0-beta.35"
+    "web3": "1.0.0-beta.36"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/truffle/package.json
+++ b/packages/truffle/package.json
@@ -26,7 +26,7 @@
     "tmp": "0.0.33",
     "truffle-box": "^1.0.8-beta.1",
     "truffle-contract": "^4.0.0-beta.1",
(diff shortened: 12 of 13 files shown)


benjamincburns:
Looks like this duplicates #1255. Since this one has no conflicts, maybe it should be preferred? That said, I think there was more to Chris' commits than just bumping the package version...

DavidBurela:
Might also want to check about allowing returns of structs from functions

benjamincburns:
@DavidBurela thanks for the tip - I just realized that in my haste I didn't add tests for this!

I did however test manually and struct returns also seem to work.

benjamincburns:
Soooo [the test that's failing](https://github.com/trufflesuite/truffle/blob/33b732790b88d30348572ce1d7012d1f8c74415c/packages/truffle-debugger/test/context.js#L103) is doing so due to a bug that I didn't introduce.

If on `next` you change the above test to `result.logs.length` instead of `result.receipt.logs` it fails over there as well. This is because when we decode logs we silently drop them if we can't find their ABI in our lookup table.

I thought linking was supposed to take care of this, so I tried adding a link step to the test's migration. Unfortunately it looks to me like `deployer.link` only solves this in the case where we have a contract depending on a library. In the case of this test we have a contract depending on another contract, and it seems `deployer.link` ignores the relationship between the two.

I feel like fixing this is outside the scope of this PR. That said, I'd really like to see this PR accepted, so if you need another one to make `deployer.link` work for contract -> contract linking, let me know.

packages/truffle-contract/lib/contract.js
```diff
@@ -408,8 +500,9 @@ var contract = (function(module) {
     toJSON: function() {
       return this._json;
     },
-  };
 
+    decodeLogs: utils.decodeLogs
```
benjamincburns: Added `decodeLogs` helper as a `_constructorMethod`

packages/truffle-contract/lib/contract.js
```diff
@@ -609,21 +723,25 @@ var contract = (function(module) {
       abi.forEach(function(item) {
         if (item.type != "event") return;
 
-        var signature = item.name + "(";
+        if (item.signature) {
```
benjamincburns: It seems the new web3 events have the signature already calculated, and for the tuple event it didn't match the one we calculated below. I didn't fully understand how this all worked, so I figured I'd best play it safe and branch on whether the `signature` is precalculated.

packages/truffle-contract/lib/utils.js
```diff
@@ -117,49 +141,77 @@ var Utils = {
   },
 
   // Extracts optional tx params from a list of fn arguments
-  getTxParams: function(args){
+  getTxParams: function(methodABI, args) {
```
benjamincburns: I now pass the `methodABI` through so that it can be used in `utils.getTxParams`

packages/truffle-contract/lib/utils.js
```diff
@@ -117,49 +141,77 @@ var Utils = {
...
     var constructor = this;
 
-    var tx_params =  {};
+    var expected_arg_count = methodABI ? methodABI.inputs.length : 0;
```
benjamincburns: Reworked this logic to be a bit more robust. The only time `methodABI` should be falsy is in the case of a call out to the fallback function.

packages/truffle-contract/lib/utils.js
```diff
@@ -117,49 +141,77 @@ var Utils = {
...
+
+    tx_params = {};
     var last_arg = args[args.length - 1];
 
-    // It's only tx_params if it's an object and not a BigNumber.
-    if (Utils.is_object(last_arg) && !Utils.is_big_number(last_arg)) {
+    if (
+      args.length === expected_arg_count + 1 &&
+      Utils.is_tx_params(last_arg)
```
benjamincburns: The way this was done before caused an off-by-one if the last constructor arg was an `object` that wasn't a `BigNumber`

packages/truffle-contract/lib/utils.js
```diff
@@ -1,54 +1,78 @@
...
     return typeof val == "object" && !Array.isArray(val);
   },
+
   is_big_number: function(val) {
     if (typeof val != "object") return false;
 
     return web3.utils.isBN(val) || web3.utils.isBigNumber(val);
   },
 
+  is_tx_params: function(val) {
```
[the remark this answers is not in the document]
benjamincburns: Fixed, but I'm not super thrilled about it... Made it require at least one known tx parameter field

packages/truffle-contract/lib/execute.js
```diff
@@ -31,38 +30,42 @@ var execute = {
...
   /**
    * Prepares simple wrapped calls by checking network and organizing the method inputs into
    * objects web3 can consume.
    * @param  {Object} constructor   TruffleContract constructor
+   * @param  {Object} methodABI     Function ABI segment w/ inputs & outputs keys.
    * @param  {Array}  _arguments    Arguments passed to method invocation
    * @return {Promise}              Resolves object w/ tx params disambiguated from arguments
    */
-  prepareCall: function(constructor, _arguments){
+  prepareCall: function(constructor, methodABI, _arguments) {
```
benjamincburns: I now pass the `methodABI` through so that it can be used in `utils.getTxParams`

packages/truffle-contract/lib/execute.js
```diff
@@ -85,21 +90,21 @@ var execute = {
   /**
    * Executes method as .call and processes optional `defaultBlock` argument.
    * @param  {Function} fn         method
-   * @param  {Object}   methodABI  function ABI segment w/ inputs & outputs keys.
+   * @param  {Object}   methodABI  Function ABI segment w/ inputs & outputs keys.
    * @return {Promise}             Return value of the call.
    */
   call: function(fn, methodABI, address) {
```
benjamincburns: I now pass the `methodABI` through so that it can be used in `utils.getTxParams`

packages/truffle-contract/lib/execute.js
```diff
@@ -113,57 +118,63 @@ var execute = {
...
    * Executes method as .send
-   * @param  {Function}   fn       Method to invoke
-   * @param  {String}     address  Deployed address of the targeted instance
+   * @param  {Function} fn         Method to invoke
+   * @param  {Object}   methodABI  Function ABI segment w/ inputs & outputs keys.
+   * @param  {String}   address    Deployed address of the targeted instance
    * @return {PromiEvent}          Resolves a transaction receipt (via the receipt handler)
    */
-  send: function(fn, address) {
+  send: function(fn, methodABI, address) {
```
benjamincburns: I now pass the `methodABI` through so that it can be used in `utils.getTxParams`

packages/truffle-contract/lib/execute.js
```diff
@@ -172,92 +183,101 @@ var execute = {
...
+   * @param  {Number} blockLimit      `block.gasLimit`
+   * @return {PromiEvent}             Resolves a TruffleContract instance
+   */
   deploy: function(args, context, blockLimit) {
     var constructor = this;
+
     var abi = constructor.abi;
+    var constructorABI = constructor.abi.filter(
+      i => i.type === "constructor"
+    )[0];
```
benjamincburns: grab constructor ABI to pass to `getTxParams`

packages/truffle-contract/lib/execute.js
benjamincburns: I now pass the `methodABI` through so that it can be used in `utils.getTxParams`

packages/truffle-contract/lib/execute.js
```diff
@@ -325,47 +363,59 @@ var execute = {
...
 
   // This gets attached to `.new` (declared as a static_method in `contract`)
   // during bootstrapping as `estimate`
-  estimateDeployment : function(){
+  estimateDeployment: function() {
     var constructor = this;
+
+    var constructorABI = constructor.abi.filter(
+      i => i.type === "constructor"
+    )[0];
```
benjamincburns: ... you know the drill by now.

packages/truffle-contract/test/cloning.js
```diff
@@ -1,41 +1,21 @@
...
-
-    var compiled = result.contracts["Example"];
-
-    network_one_id = 1000;
-    network_two_id = 1001;
-    network_one = TestRPC.provider({network_id: network_one_id, seed: network_one_id, logger: log});
-    network_two = TestRPC.provider({network_id: network_two_id, seed: network_two_id, logger: log});
-
-    ExampleOne = contract(compiled);
+    ExampleOne = util.createExample();
```
benjamincburns: cleaned up test due to eslint errors

packages/truffle-contract/test/linking.js
```diff
@@ -77,26 +107,41 @@ describe("Library linking with contract objects", function() {
...
-    return web3.eth.net.getId().then(function(id){
+    return web3.eth.net.getId().then(function(id) {
       network_id = id;
     });
   });
 
   before(function() {
     this.timeout(10000);
 
+    const exampleLibraryPath = path.join(
```
benjamincburns: for some reason these paths weren't resolving correctly while debugging in vs code - this fixes that

packages/truffle-contract/test/methods.js
```diff
@@ -1,28 +1,25 @@
 var assert = require("chai").assert;
 var BigNumber = require("bignumber.js");
-var util = require('./util');
+var util = require("./util");
```
benjamincburns: A bunch of the changes in this file came over from the stuff I cherry-picked from Chris' branch

packages/truffle-contract/test/separation.js
```diff
@@ -1,54 +1,48 @@
...
         assert(events.length === 1);
         done();
       }
     });
 
     exampleB
       .triggerContractAddressEvent()
-      .then(exampleA.triggerContractAddressEvent);
+      .then(() => exampleA.triggerContractAddressEvent())
+      .catch(assert.fail);
```
benjamincburns: This test caused an unhandled promise rejection when it was failing. Now it doesn't.

packages/truffle-contract/lib/utils.js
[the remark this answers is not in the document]
gnidan: Heh, I see. Well, it's probably fine.

packages/truffle-contract/test/util.js
```diff
@@ -20,18 +21,37 @@ var util = {
 
   // Compiles and instantiates (our friend) Example.sol
   createExample: function() {
+    return util._createContractInstance(
```
benjamincburns: I just followed in the footsteps of `createExample` here - refactoring the logic of that out to `_createContractInstance`

packages/truffle-contract/test/cloning.js
gnidan: 🎉

packages/truffle-debugger/test/context.js
```diff
@@ -93,14 +89,14 @@ describe("Contexts", function () {
...
-  it("returns view of addresses affected", async function () {
+  it("returns view of addresses affected", async function() {
     let outer = await abstractions.OuterContract.deployed();
     let inner = await abstractions.InnerContract.deployed();
 
     // run outer contract method
     let result = await outer.run();
 
-    assert.equal(2, result.receipt.logs.length, "There should be two logs");
+    assert.equal(2, result.receipt.rawLogs.length, "There should be two logs");
```
benjamincburns: switch the test to use the new `rawLogs` field on the truffle receipt.

packages/truffle-contract/lib/handlers.js
```diff
@@ -84,25 +85,33 @@ var handlers = {
    * @param  {Object} context   execution state
    * @param  {Object} receipt   transaction receipt
    */
-  receipt: async function(context, receipt){
+  receipt: async function(context, receipt) {
+    // keep around the raw (not decoded) logs in the raw logs field as a
+    // stopgap until we can get the ABI for all events, not just the current
+    // contract
+    receipt.rawLogs = receipt.logs;
```
benjamincburns: per the comment, keep around the raw (not decoded logs)

packages/truffle-contract/test/abiV2.js
```diff
@@ -0,0 +1,74 @@
+var assert = require("chai").assert;
+var util = require("./util");
+
+// attribution where it's due - parts of this test are from this gist:
+// https://gist.github.com/ricmoo/e38d4d71dff7156033922d2e5de88d37
+
+describe("ABIEncoderV2", function() {
```
benjamincburns: If you want to see how ABIEncoderV2 support is meant to work from the user's perspective, this is the place to look.

packages/truffle-contract/test/cloning.js
benjamincburns: Just kidding, it was due to the path resolution failing when executing tests in the debugger.

PR #1411, opened 2018-11-06
fix: prevent sending deduped (false) events downstream.

m-schmoock:
When using truffle-contract on websocket enabled geth nodes, duplicated events got passed as `false` to the application callback handler.

I dont think this was intended behaviour, otherwise it should be documented that callbacks may occasionally be called with just `false` instead of a parsed message.

### DIFF
--- a/packages/truffle-contract/lib/execute.js
+++ b/packages/truffle-contract/lib/execute.js
@@ -252,9 +252,9 @@ var execute = {
       // As callback
       if (callback !== undefined){
         var intermediary = function(err, e){
-          if (err) callback(err);
-          var event = dedupe(e.id) && decode.call(constructor, e, true)[0];
-          callback(null, event);
+          if (err) return callback(err);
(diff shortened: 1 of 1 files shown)


m-schmoock:
Note: Same may apply to the `event.on` data/changed/error handlers. Deduped events should just always be silently dropped instead of converted to "false".

m-schmoock:
note: the test coverage measurement is clearly defunct.

m-schmoock:
What about the `event.on` data/changed/error handlers? They also seem to be called with "false" on filtered duplicates...

m-schmoock:
@gnidan two issue:
1. Whats with the `event.on` handlers that are also called with `false` on duplicates?
2. F.y.i. This duplicate stuff only happens when used with a real network on i.e. geth daemon. So the tests only partly make sens, as they are not reproducibly using ganache test.

PR #1412, opened 2018-11-06
add yml file to configure github stale bot

fainashalts:
Created a yml file to enable configuration for github bot to close stale issues.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,7 +1,7 @@
 # Configuration for probot-stale - https://github.com/probot/stale
 
 # Number of days of inactivity before an Issue or Pull Request becomes stale
-daysUntilStale: 30
+daysUntilStale: 60
 
 # Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
 # Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,54 @@
...
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 30
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 7
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
```
fainashalts: are there any labels we want to add here, that we never want to be marked as stale? maybe infrastructure? other ones?

.github/stale.yml
```diff
@@ -0,0 +1,54 @@
...
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
+
+# Label to use when marking as stale
+staleLabel: stale
+
+# Comment to post when marking as stale. Set to `false` to disable
+markComment: >
+  Thank you for raising this issue! It has been automatically marked as stale because it has not had
```
fainashalts: Happy to change the wording for this.

.github/stale.yml
eggplantzzz: Perhaps we could make a custom label that we could use for issues that need to sit around for a while.  I don't know what it might be called...

.github/stale.yml
```diff
@@ -0,0 +1,55 @@
...
+  recent activity. It will be closed in 7 days if no further activity occurs. If you would like to keep this issue open,
+  please respond with information about the current state of this problem.
+
+# Comment to post when removing the stale label.
+unmarkComment: >
+  Thanks for your response! This issue is no longer considered stale and someone from the Truffle team will try to respond as soon as we can.
```
eggplantzzz: "Truffle team will try to respond as soon as **they** can."

PR #1415, opened 2018-11-08
Add slither integration

anukul:
Closes https://github.com/trailofbits/slither/issues/18

### DIFF
--- a/packages/truffle-slither-analyze/index.js
+++ b/packages/truffle-slither-analyze/index.js
@@ -1,0 +1,19 @@
+const which = require("which");
+const { spawn } = require("child_process");
+
+function truffleSlitherAnalyze(options) {
+  try {
+    let slither = which.sync("slither");
+    let child = spawn(slither, options);
+    child.stdout.on("data", data => process.stdout.write(data.toString()));
+    child.stderr.on("data", data => process.stdout.write(data.toString()));
+  } catch (error) {
+    console.log("Slither not found in path");
+    console.log(
+      `Please install the package using \`pip install slither-analyzer\`\n`
+    );
+    throw error;
+  }
+}
+
+module.exports = truffleSlitherAnalyze;
--- a/packages/truffle-core/lib/commands/index.js
+++ b/packages/truffle-core/lib/commands/index.js
@@ -19,4 +19,5 @@ module.exports = {
   version: require("./version"),
   config: require("./config"),
   watch: require("./watch"),
+  slither_analyze: require("./slither-analyze")
 };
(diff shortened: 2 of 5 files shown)


packages/truffle-slither-analyze/index.js
```diff
@@ -0,0 +1,17 @@
...
+
+function truffleSlitherAnalyze(options) {
+  try {
+    let slither = which.sync("slither");
+    let child = spawn(slither, options);
+    child.stdout.on("data", data => console.log(data.toString()));
+    child.stderr.on("data", data => console.log(data.toString()));
+  } catch (error) {
+    console.log("Slither not found in path");
+    console.log(`Please install the package using \`pip install slither\`\n`);
```
montyly:
```suggestion
    console.log(`Please install the package using \`pip install slither-analyzer\`\n`);
```

anukul: Oh, sorry about that. I had built it from source, never actually installed from pypi.

gnidan:
Hi @anukul, thanks for this PR.

After reviewing with members of the team, we've concluded that we are not accepting PRs to add arbitrary commands to Truffle.

We're working on features to enable these kinds of integrations at a user-level, via external node modules. I'll describe our plan here, so hopefully you don't feel you've come away empty-handed:

1. First, we're working on a system for doing `truffle run <command-plugin>`. In your case, you could just take the `slither-analyze` logic, package it on NPM, and instruct users to add it to their project config. Then users could do `truffle run slither-analyze`.

This step is a **top priority** right now, and I'm scheduling it for **inclusion in the next beta**, which will release either this week or next (once the `truffle run` work is done).

2. Beyond that, and the real reason for not accepting PRs of this nature, is that we intend to identify integration points in existing commands, and possibly add some new ones in the process. We plan to define lifecycle hooks for plugins, so that plugins can hook into Truffle artifacts and output results to a Truffle reporting component. So in the future, you could define a general Slither plugin, to hook into the future `truffle analyze` command.

The idea here is that we want the top-level `truffle <command>` namespace to remain restricted to the core set of workflows that Truffle offers, so that users can ensure that each command works as expected, or to make it easy to identify gaps in documentation, etc. when things behave unexpectedly.

This effort is certainly a larger undertaking, but hopefully we'll be able to get Step 1 out very soon, which I think will meet the immediate needs of this PR.

If you have any questions, please don't hesitate to ask, either here or in the [truffle-dev Gitter](https://gitter.im/trufflesuite/truffle-dev). Apologies, and thank you again.

PR #1425, opened 2018-11-13
Add better cleanup for truffle-box sandbox()

seesemichaelj:
This PR changes the `sandbox` interface to accept an argument `options` instead of `name` to allow some extra options such as `unsafeCleanup` and `setGracefulCleanup` that interface with the `tmp` module. There is back compatibility where if `options` is a `string`, then it will use it as the `name` directly

You can find an example of using these changes, as well as the motivation for these changes in `packages/truffle-debugger/test/helpers.js` to cleanup the sandboxes when mocha finishes the tests for the debugger.

Here are some comments on the the options:
```
// options.unsafeCleanup
//   Recursively removes the created temporary directory, even when it's not empty. default is false
// options.setGracefulCleanup
//   Cleanup temporary files even when an uncaught exception occurs
```

### DIFF
--- a/packages/truffle-box/box.js
+++ b/packages/truffle-box/box.js
@@ -4,43 +4,82 @@ var path = require("path");
 
 var Config = require("truffle-config");
 
+function parseSandboxOptions(options) {
+  if (typeof options === "function") {
+    return {
+      name: "default",
+      unsafeCleanup: false,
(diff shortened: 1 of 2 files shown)


gnidan:
(Oh, please wait for Travis :)

PR #1430, opened 2018-11-14
Replace chokidar

eggplantzzz:
Chokidar doesn't seem to like to bundle well with webpack for whatever reason.  So replace it with 'sane'!


eggplantzzz:
Hmm I'm not sure @gnidan

tcoulter:
@balexander4 Would *love* if you do a platform pass on this, and ensure it works on Windows/Mac/Linux. (This has historically been a problem in the past.)

PR #1433, opened 2018-11-14
Bump solc version to 0.5.0

JChanceHud:
This pull request bumps the `solc` version in the `truffle` and `truffle-compile` packages to [`0.5.0`](https://www.npmjs.com/package/solc) which was released a day ago.

I also added a node script `solc-bump` that updates dependent subpackages in `./packages` to `solc@latest`.

To get only the package version updates cherry-pick commit 5b7c6f9f0f0d23eb82814cbe96e4da47ddd8099b.

Thanks!


gnidan:
!!! Awesome.

Reviewing now.

JChanceHud:
Actually this solidity release seems to have quite a bit of [changes](https://solidity.readthedocs.io/en/latest/050-breaking-changes.html). This may need to be a more extensive refactor than a simple bump :/

scripts/solc-bump.js
```diff
@@ -0,0 +1,56 @@
+const fs = require('fs');
+const { execSync } = require('child_process');
+const readline = require('readline');
+
+const PACKAGES_DIR = `${__dirname}/../packages`;
```
gnidan: Should use `path.resolve()` for this to make it more cross-platform.

scripts/solc-bump.js
```diff
@@ -0,0 +1,56 @@
...
+  });
+}
+
+/**
+ * Use child_process to run `npm install` in `packageName` synchronously
+ **/
+function updatePackages(packageNames) {
+  packageNames.forEach(packageName => {
+    console.log(`Updating ${packageName}`);
+    execSync('npm install solc@latest', {
```
gnidan: Are requirements for this script to save the value to the `package.json`? It seems that you would need `--save` for that.

scripts/solc-bump.js
```diff
@@ -0,0 +1,56 @@
...
+  const packages = fs.readdirSync(PACKAGES_DIR);
+  return packages.filter(packageName => {
+    // parse the package.json file for `packageName`
+    const packagePath = `${PACKAGES_DIR}/${packageName}/package.json`;
+    const packageJson = JSON.parse(fs.readFileSync(packagePath, 'utf8'));
+
+    // operate if the solc dependency exists
+    return packageJson
+      && packageJson.dependencies
+      && !!packageJson.dependencies['solc'];
```
gnidan: We might want to include `devDependencies` here; any thoughts?

gnidan:
Related, waiting for Travis: #1434

scripts/solc-bump.js
JChanceHud: Good call, updated

scripts/solc-bump.js
JChanceHud: npm install actually saves to dependencies by default, but I added explicity `--save` and `--save-dev` with `devDependencies` support

scripts/solc-bump.js
JChanceHud: Agreed, updated and refactored a bit

PR #1434, opened 2018-11-15
Fix build for Solidity v0.5.0

gnidan:
Separate the truffle-compile test contracts into v4 and v5 so we don't have to upgrade the entire test suite right now.

Since `native` is just configured to use the most recent version, just use the v5 pragma now for the native tests.


gnidan:
Fixin' the build!

PR #1444, opened 2018-11-19
Pass empty-string instead of 0x to abi.decodeLog

gnidan:
Since that is, what it seems, what that function wants

Seems related to https://github.com/ethereum/web3.js/issues/1916

Also: add a proper catch statement so this sort of exception doesn't just get completely ignored.


gnidan:
Thanks @eggplantzzz !

PR #1453, opened 2018-11-22
Update README.md

jtakalai:
You probably meant to save the instance to the outer variable instead of introducing local variable.

Added ES6 code just to show it doesn't have to be so ugly, at least if you're using non-archaic node.


gnidan:
(Huh seems the build is failing and GH won't let me merge a failing build on mobile, so this will have to wait.)

PR #1460, opened 2018-11-26
Tweak options for sane library

eggplantzzz:
Use a different option to attempt to cope with watching big codebases.

See the different modes in the docs.  
https://www.npmjs.com/package/sane

### DIFF
--- a/packages/truffle-core/lib/commands/watch.js
+++ b/packages/truffle-core/lib/commands/watch.js
@@ -20,7 +20,8 @@ const command = {
       ignored: [
         "build/**/**",
         /[\/\\]\./ // Ignore files prefixed with .
-      ]
+      ],
+      watchman: true
     };
     const watchCallback = filePath => {
       const displayPath = path.join(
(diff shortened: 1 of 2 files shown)


packages/truffle-core/lib/commands/watch.js
```diff
@@ -22,6 +23,18 @@ const command = {
...
     };
+    // Certain large codebases have trouble with the watch command.
+    // Installing watchman resolves some of these issues.
+    if (commandExistsSync("watchman")) {
+      watchOptions.watchman = true;
+    } else {
+      config.logger.log(
+        colors.yellow(
+          "If you have trouble using watch, try installing watchman."
+        )
```
gnidan: If only displaying this message in a `catch` is not an option, can we remove the `yellow` here? That way the message will be more ignorable, but still easy enough to find if you're looking at an unhandled exception.

eggplantzzz: Sounds fine to me

PR #1462, opened 2018-11-26
Fix `truffle help <command>` lag

CruzMolina:
## Bug Fix

Running `truffle help <command>` would lag due to missing callback executions.

New expected Behavior:

`truffle help <command>` executes quickly w/ minimal lag.

### DIFF
--- a/packages/truffle-core/lib/commands/help.js
+++ b/packages/truffle-core/lib/commands/help.js
@@ -16,7 +16,7 @@ var command = {
     var commands = require("./index");
     if (options._.length === 0) {
       this.displayCommandHelp("help");
-      return;
+      return callback();
     }
     var selectedCommand = options._[0];
 


packages/truffle-core/lib/commands/help.js
```diff
@@ -1,17 +1,18 @@
...
+        description: "Name of the command to display information for."
       }
-    ],
+    ]
   },
   builder: {},
-  run: function (options, callback) {
+  run: function(options, callback) {
     var commands = require("./index");
     if (options._.length === 0) {
```
gnidan: This branch `return`s when it should `return callback();`

packages/truffle-core/lib/commands/run.js
```diff
@@ -20,7 +20,7 @@ const command = {
     if (options._.length === 0) {
       const help = require("./help");
       help.displayCommandHelp("run");
-      return;
+      done();
```
gnidan: This should probably also `return`

packages/truffle-core/lib/commands/help.js
gnidan: Looks like this one is still missing

PR #1465, opened 2018-11-26
Upgrade test contracts to 0.5.0 and fix all resulting problems

haltman-at:
[Note: This branch is kind of a mess; it might not end up getting merged in as such.]

This branch upgrades all the test contracts to 0.5.0, and attempts to fix all the resulting problems.  I've already fixed one such problem, namely the lack of support for `STATICCALL`, in PR #1446.  But there's more.

One issue that's come up is that a number of my tests are quite shaky.  This isn't an 0.5.0 problem, but I in my investigations I think I've figured out what's causing it anyway -- namely, I was relying on `sourceId` to be consistent between runs, when it's not.  That's an easy fix.

But the real problems are the 0.5.0-specific problems, which are very consistent.  The failing tests (now that I've fixed the storage decoding tests, where the problem was `STATICCALL`) are:
1. The memory decoding tests
2. The variable IDs tests
3. The two mapping tests where the domain is numeric

Note that these errors exist also, in near-identical form, on the `truffle-decoder` branch; I made a separate branch for testing that, namely `decoder-0.5.0-tests`, which see.

The memory decoding tests, as best I can tell, do seem to be due to a decoding problem.  However, I'm not certain of this.

The variable ID tests seem to be due to some difference in how variable declarations are handled, except for one of them which also seems to be due to a decoding problem like above.  Again, all this is uncertain.

The two mapping tests... I'd say also a decoding problem, probably, but a different one.  However I've made no real exploration of this at all.

Basically, I don't know what's going on, but that's my current progress.  Will update.

### DIFF
--- a/packages/truffle-debugger/test/ast.js
+++ b/packages/truffle-debugger/test/ast.js
@@ -4,7 +4,6 @@ const debug = debugModule("test:ast");
 import { assert } from "chai";
 
 import Ganache from "ganache-cli";
-import Web3 from "web3";
 
 import { prepareContracts } from "./helpers";
 import Debugger from "lib/debugger";
@@ -13,10 +12,10 @@ import ast from "lib/ast/selectors";
 import solidity from "lib/solidity/selectors";
 import trace from "lib/trace/selectors";
 
-import { getRange, findRange, rangeNodes } from "lib/ast/map";
+import { getRange } from "lib/ast/map";
 
 const __VARIABLES = `
-pragma solidity ^0.4.18;
+pragma solidity ~0.5;
 
 contract Variables {
   event Result(uint256 result);
@@ -44,22 +43,19 @@ contract Variables {
 }
 `;
 
-
 let sources = {
   "Variables.sol": __VARIABLES
 };
 
 describe("AST", function() {
   var provider;
-  var web3;
 
   var abstractions;
   var artifacts;
   var files;
 
   before("Create Provider", async function() {
-    provider = Ganache.provider({seed: "debugger", gasLimit: 7000000});
-    web3 = new Web3(provider);
+    provider = Ganache.provider({ seed: "debugger", gasLimit: 7000000 });
   });
 
   before("Prepare contracts and artifacts", async function() {
@@ -93,23 +89,24 @@ describe("AST", function() {
 
         let node = session.view(ast.current.node);
 
-        let [ nodeStart, nodeLength ] = getRange(node);
+        let [nodeStart, nodeLength] = getRange(node);
         let nodeEnd = nodeStart + nodeLength;
 
         let pointer = session.view(ast.current.pointer);
 
         assert.isAtMost(
-          nodeStart, start,
+          nodeStart,
+          start,
           `Node ${pointer} at should not begin after instruction source range`
         );
         assert.isAtLeast(
-          nodeEnd, end,
+          nodeEnd,
+          end,
           `Node ${pointer} should not end after source`
         );
 
         session.stepNext();
-      } while(!session.view(trace.finished));
-
+      } while (!session.view(trace.finished));
     });
   });
 });
--- a/packages/truffle-debugger/test/context.js
+++ b/packages/truffle-debugger/test/context.js
@@ -76,6 +76,7 @@ describe("Contexts", function() {
 
   var abstractions;
   var artifacts;
+  var files;
 
   before("Create Provider", async function() {
     provider = Ganache.provider({ seed: "debugger", gasLimit: 7000000 });
@@ -87,6 +88,7 @@ describe("Contexts", function() {
     let prepared = await prepareContracts(provider, sources, migrations);
     abstractions = prepared.abstractions;
     artifacts = prepared.artifacts;
+    files = prepared.files;
   });
 
   it("returns view of addresses affected", async function() {
@@ -102,6 +104,7 @@ describe("Contexts", function() {
 
     let bugger = await Debugger.forTx(txHash, {
       provider,
+      files,
       contracts: artifacts
     });
     debug("debugger ready");
--- a/packages/truffle-debugger/test/context.js
+++ b/packages/truffle-debugger/test/context.js
@@ -11,7 +11,7 @@ import Debugger from "lib/debugger";
 import sessionSelector from "lib/session/selectors";
 
 const __OUTER = `
-pragma solidity ^0.4.18;
+pragma solidity ~0.5;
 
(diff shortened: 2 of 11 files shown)


haltman-at:
Note: The intermittent failure should now be addressed due to PR #1466.

packages/truffle-debugger/test/helpers.js
```diff
@@ -68,17 +62,21 @@ export function getAccounts(provider) {
...
-    ) {
-      if (err) return reject(err);
-      result.resolver = new Resolver(result);
-      result.artifactor = new Artifactor(result.contracts_build_directory);
-      result.networks = {};
+    Box.sandbox(
+      {
+        unsafeCleanup: true,
+        setGracefulCleanup: true,
+        name: "default#web3-one"
```
gnidan: This change is necessary for Migrations.sol

packages/truffle-config/index.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
gnidan: We should probably get rid of this

packages/truffle-debugger/test/ast.js
```diff
@@ -13,10 +12,10 @@ import ast from "lib/ast/selectors";
 import solidity from "lib/solidity/selectors";
 import trace from "lib/trace/selectors";
 
-import { getRange, findRange, rangeNodes } from "lib/ast/map";
+import { getRange } from "lib/ast/map";
 
 const __VARIABLES = `
-pragma solidity ^0.4.18;
+pragma solidity ~0.5;
```
gnidan: These of course are necessary

packages/truffle-debugger/test/context.js
```diff
@@ -11,7 +11,7 @@ import Debugger from "lib/debugger";
 import sessionSelector from "lib/session/selectors";
 
 const __OUTER = `
-pragma solidity ^0.4.18;
+pragma solidity ~0.5;
```
gnidan: :+1:

packages/truffle-debugger/test/context.js
```diff
@@ -33,7 +33,7 @@ contract OuterContract {
 `;
 
 const __INNER = `
-pragma solidity ^0.4.18;
+pragma solidity ~0.5;
```
gnidan: :+1:

packages/truffle-debugger/test/data/decode/decoding.js
```diff
@@ -5,136 +5,155 @@ import faker from "faker";
...
    */
   describeDecoding(
-    "Storage Variables", commonFixtures, evm.current.state.storage,
+    "Storage Variables",
+    commonFixtures,
+    evm.current.state.storage,
 
     (contractName, fixtures) => {
-      return `pragma solidity ^0.4.23;
+      return `pragma solidity ~0.5;
```
gnidan: :+1:

packages/truffle-debugger/test/data/decode/decoding.js
```diff
@@ -5,136 +5,155 @@ import faker from "faker";
...
 
   describeDecoding(
-    "Mapping Variables", mappingFixtures, evm.current.state.storage,
+    "Mapping Variables",
+    mappingFixtures,
+    evm.current.state.storage,
 
     (contractName, fixtures) => {
-      return `pragma solidity ^0.4.24;
+      return `pragma solidity ~0.5;
```
gnidan: :+1:

packages/truffle-debugger/test/data/decode/helpers.js
```diff
@@ -67,20 +65,20 @@ async function prepareDebugger(testName, sources) {
...
+  );
   let sourceId = matchingSources[0].id;
 
   let breakpoint = {
     sourceId,
     line: lastStatementLine(source)
   };
 
-  session.addBreakpoint(breakpoint,true);
+  session.addBreakpoint(breakpoint);
```
gnidan: Sure if you wanna include

packages/truffle-debugger/test/data/ids.js
```diff
@@ -13,7 +13,7 @@ import trace from "lib/trace/selectors";
 import solidity from "lib/solidity/selectors";
```
gnidan: All the things in this file

packages/truffle-debugger/test/endstate.js
```diff
@@ -1,5 +1,5 @@
 import debugModule from "debug";
```
gnidan: all the things here

packages/truffle-debugger/test/evm.js
```diff
@@ -1,10 +1,9 @@
 import debugModule from "debug";
```
gnidan: yep these too

packages/truffle-debugger/test/helpers.js
```diff
@@ -23,12 +23,6 @@ export async function prepareContracts(provider, sources = {}, migrations) {
   };
   config.network = "debugger";
 
-  config.compilers = {
-    solc: {
-      version: "0.4.25"
-    }
-  };
-
```
gnidan: Gotta put this back in and turn the version up, apparently!

packages/truffle-debugger/test/precompiles.js
```diff
@@ -1,10 +1,9 @@
 import debugModule from "debug";
```
gnidan: All this

packages/truffle-debugger/test/reset.js
```diff
@@ -12,14 +12,14 @@ import data from "lib/data/selectors";
 import solidity from "lib/solidity/selectors";
```
gnidan: yep

packages/truffle-debugger/test/solidity.js
```diff
@@ -12,7 +12,7 @@ import solidity from "lib/solidity/selectors";
 import trace from "lib/trace/selectors";
```
gnidan: yep

haltman-at:
OK, the optimizer has now been turned off for the tests (thanks @gnidan), solving most of the problems.

In addition, I've passed the `files` argument to `prepareDebugger` in tests that were previously missing it, which should (independently of the `0.5.0` update) fix their intermittent failures.

A word of warning, the compiler version is currently set to `0.5.0` rather than `^0.5.0`; this is because the latter doesn't work on my machine.  We'll have to figure out some way of getting the latter to work, because we don't want to have to constantly update the version every time there's a new version of `solc`.

Issue #1452, opened 2018-11-22
Truffle v5.0.0-beta.2 compilers config error


jerryji:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue
On Truffle 5.0.0-beta.2 release page ( https://github.com/trufflesuite/truffle/releases/tag/v5.0.0-beta.2 ), "Bring Your Own Compiler" links to https://github.com/trufflesuite/truffle/releases/tag/v5.0.0-beta.0#bring-your-own-compiler , which states "compilers" section config as --

```
  compilers: {
     solc: {
       version: <string>  // ex:  "0.4.20". (Default: Truffle's installed solc)
     }
  }
```

However, "optimizer" does not work _within_ the "compilers" section --

```
do-nyc3-0:~/tmp/ethwalldata$ tail -10 truffle-config.js
    compilers: {
        solc: {
            version: "0.5.0",
            optimizer: {
                enabled: true,
                runs: 1000
            }
        }
    }
};
do-nyc3-0:~/tmp/ethwalldata$ truffle compile --all
Compiling ./contracts/EthWallData.sol...
...
Writing artifacts to ./build/contracts

do-nyc3-0:~/tmp/ethwalldata$ grep \"bytecode\" build/contracts/* | awk '{print $1 " " length($3)/2}' | head -1
build/contracts/EthWallData.json: 25941.5
```

while after removing "compilers" block works (notice the smaller bytecode) --

```
do-nyc3-0:~/tmp/ethwalldata$ tail -10 truffle-config.js
        }
    },
    solc: {
        version: "0.5.0",
        optimizer: {
            enabled: true,
            runs: 1000
        }
    }
};
do-nyc3-0:~/tmp/ethwalldata$ truffle compile --all
Compiling ./contracts/EthWallData.sol...
...
Writing artifacts to ./build/contracts

do-nyc3-0:~/tmp/ethwalldata$ grep \"bytecode\" build/contracts/* | awk '{print $1 " " length($3)/2}' | head -1
build/contracts/EthWallData.json: 16238.5
```

What the issue is, in broad strokes.

## Steps to Reproduce

unzip the attached zip file, compile; update truffle-config.js and recompile to see the difference
[ethwalldata.zip](https://github.com/trufflesuite/truffle/files/2606958/ethwalldata.zip)

Please provide the shortest amount of steps to reproduce your issue.

## Expected Behavior

[...]

PR #1467, opened 2018-11-27
Fix BYOC/Solc Config Bug

CruzMolina:
## Bug Fix

As noted in #1452, there is a mismatch between the documentation for BYOC/V5 and the way `compilers.solc` is handled in the codebase. This would cause user-configured compiler settings like `optimizer`, `runs`, and `evmVersion` to be omitted and use the default Config values instead.

## New Expected Behavior

Using a new BYOC/V5 format (example below) will use the solc values specified and result in optimized code.

```
module.exports = {
  ...
  compilers: {
        solc: {
            version: "0.5.0",
            optimizer: {
                enabled: true,
                runs: 1000
            }
        }
    }
};
```

This will also work:

```
module.exports = {
  ...
  compilers: {
        solc: {
            version: "0.5.0",
            settings: {
                optimizer: {
                    enabled: true,
                    runs: 1000
                }
            }
        }
    }
};
```

The V4 solc format (example below) is also still supported.

```
module.exports = {
...
solc: {
        version: "0.5.0",
        optimizer: {
            enabled: true,
            runs: 1000
        }
    }
};
```

as well as:

```
module.exports = {
...
solc: {
        version: "0.5.0",
        settings: {
            optimizer: {
                enabled: true,
                runs: 1000
            }
        }
    }
};
```


CruzMolina:
Coverage would be nice...

gnidan:
Thoughts on this: Including `settings` is the _correct_ form, but we can _allow_ the form where `optimizer` (or other specific keys) are specified without the wrapping `settings`.

My reasoning is that:
- `optimizer` is arguably solc specific: we do not control the data format for the optimizer settings.
- The object description for "compiler configs", e.g. `compilers.solc`, should be generic.
- By defining a generic `settings` property for all compiler objects, we can maintain conformity across compilers.

Including this because I already wrote it: (maybe it provides clarity?)
```javascript
compilers: {
  "<whatever-compiler>": {  // this object should be agnostic to language
    version: "...",
    /* anotherTruffleGenericSetting: "..." */
    settings: {  // a pseudo "black box"
      /* compilerSpecificSetting: "..." */
    }
}
```

eggplantzzz:
So I'm understanding that the optimizer settings are going to go under `config.compilers[<compiler>].settings`.  What does `index.js.orig` do?

CruzMolina:
Ah, thanks for spotting that @eggplantzzz . pushed the wrong file 😆

packages/truffle-compile/index.js
```diff
@@ -35,12 +35,12 @@ var compile = function(sources, options, callback) {
 
   expect.options(options.compilers, ["solc"]);
 
-  options.compilers.solc.settings = options.compilers.solc.settings || {};
+  options.compilers.solc.settings =
+    options.compilers.solc.settings || options.compilers.solc || {};
```
gnidan: This is a bit risky, but probably fine. I think ideally we would want to grab only `optimizer` and other _specific_ keys, cause otherwise `version`, etc. will get passed as `settings`, which might cause unforeseen namespace collisions.

Issue #1461, opened 2018-11-26
Random "Could not connect to your Ethereum client"


elenadimitrova:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue
We encounter a client disconnect issue at seemingly random points when running our test suite. Tests pass when run individually.

The following issues appear related:
https://github.com/trufflesuite/truffle/issues/729
https://github.com/trufflesuite/truffle/issues/1260
https://github.com/trufflesuite/ganache-core/commit/312902f25e6581c7aee900497ff5146ea217cd4a

## Steps to Reproduce

It has been really difficult to reproduce this issue as it is likely due to a race condition.  

## Expected Behavior

Test suite to complete 

## Actual Results

```
 Error: Could not connect to your Ethereum client. Please check that your Ethereum client:
    - is running
    - is accepting RPC connections (i.e., "--rpc" option is used in geth)
    - is accessible over the network
    - is properly configured in your Truffle configuration file (truffle.js)
```
Sample CI builds failing
https://circleci.com/gh/JoinColony/colonyNetwork/4940
https://circleci.com/gh/JoinColony/colonyNetwork/4930

## Environment

* Operating System: OSX
* Ethereum client: `ganache-cli@6.1.6` https://github.com/area/ganache-cli/commits/fakefix/eth_call_selfdestruct
* Truffle version (`truffle version`): `5.0.0-beta.2`
* node version (`node --version`): `9.11.2`
* npm version (`npm --version`): `5.6.0`

gnidan:
Seems like this is resolved*. Closing, but let us know if the problem starts happening again!

* https://github.com/JoinColony/colonyNetwork/pull/461

Jonybang:
@gnidan Solved by what, downgrading and using custom release? When fix of this error will appeared in regular release? on 5.0.3 release?

dzarezenko:
I still have the same problem with Ganache.

```
ProviderError: 
Could not connect to your Ethereum client with the following parameters:
    - host       > 127.0.0.1
    - port       > 8545
    - network_id > 5777
Please check that your Ethereum client:
    - is running
    - is accepting RPC connections (i.e., "--rpc" option is used in geth)
    - is accessible over the network
    - is properly configured in your Truffle configuration file (truffle-config.js)

      at /usr/local/lib/node_modules/truffle/build/webpack:/packages/provider/wrapper.js:73:1
[...]

PR #1469, opened 2018-11-27
Memoize config.provider

gnidan:
To (hopefully) address #1461, but at the very least this should improve performance.

Memoization works via:
- Memo is stored upon `config.provider`'s get
- Memo is reset upon `config.network`'s set

Risks/downsides:
- It might be dangerous to use a memoization mechanism that spans three different parts of the code (i.e., initialization, provider get, network set)

- It's dangerous if the `config.provider` memo is theoretically invalidated but never reset. It might not be sufficient to reset the memo when `config.network = ...` happens.

  For instance, `config.network["..."].port = "..."` would result in an invalid memoized `config.provider`.

  This may not matter.

- We might want to make a `config.networks["..."].memoize` option, to allow this to be turned off.

For now though... see what Travis does.

### DIFF
--- a/packages/truffle-config/index.js
+++ b/packages/truffle-config/index.js
@@ -211,18 +211,24 @@ function Config(truffle_directory, working_directory, network) {
           return null;
         }
 
-        // use memoized provider (NOTE this requires reset if network changes)
-        if (providerMemo) {
-          return providerMemo;
-        }
-
+        // collect provider options from network config + global flags
         var options = self.network_config;
         options.verboseRpc = self.verboseRpc;
 
+        // check explicit `memoize` config value otherwise default enable
+        const shouldMemoize = "memoize" in options ? options.memoize : true;
+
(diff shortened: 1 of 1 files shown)


packages/truffle-config/index.js
```diff
@@ -77,7 +75,17 @@ function Config(truffle_directory, working_directory, network) {
     // These are already set.
     truffle_directory: function() {},
     working_directory: function() {},
-    network: function() {},
+    network: {
+      get: function() {
+        return this._values.network;
+      },
+      set: function(val) {
```
eggplantzzz: I don't want to really make a thing out of this, but I just cringe slightly when I see a variable name as bland as `val`...and that it is shortened.  I don't know if I have a better suggestion.

eggplantzzz: Any thoughts?

gnidan: We can change these later

PR #1481, opened 2018-11-30
Add !<...> syntax to debugger watch expressions

gnidan:
Where `...` is the name of some selector.

For instance, this allows doing something like:
```
  debug(...)> :!<data.current.state.memory>.slice(320, 320+32)
```


gnidan:
> OK, makes sense to me. I'm a little wary of the use of text transformation to get select called, but, I guess it makes sense in context.

I'm also not thrilled with it, but I don't have a better idea. Let me know if you do!

In the meantime, gonna merge anyway. Thanks!

PR #1484, opened 2018-12-01
Compartmentalize Assert.sol & Add Legacy SafeSend.sol

CruzMolina:
## Bug Fix

Discovered while working on #1467. Assert.sol exceeds the current block gas size limit and returns an OOG error. Additionally, due to solidity 0.5.0's change of multiple address types, two versions of SafeSend.sol are needed to support both v0.4.0 & v0.5.0 solc compilers.

## Solution

Compartmentalize Assert.sol into multiple contracts & maintain a legacy SafeSend.sol for v0.4.0 solidity testing.

Also eliminated compiler output from internal truffle .sol test helpers while using `truffle test`.


packages/truffle-compile/index.js
```diff
@@ -417,16 +415,19 @@ compile.with_dependencies = function(options, callback) {
...
     }
 
+    const blacklistRegex = /^truffle\//;
+
     paths.sort().forEach(contract => {
       if (path.isAbsolute(contract)) {
         contract =
           "." + path.sep + path.relative(options.working_directory, contract);
       }
+      if (contract.match(blacklistRegex)) return;
```
gnidan: Interesting!

packages/truffle-core/lib/testing/Assert.sol
```diff
@@ -1184,497 +986,214 @@ library Assert {
...
-        for (i = 0; i < t1l; i++)
-            newB[j++] = tagged1B[i];
-        newB[j++] = ')';
-
-        return string(newB);
+    function balanceIsNotZero(address a, string memory message) internal returns (bool result) {
+        return AssertBalance.balanceIsNotZero(a, message);
     }
-
 }
```
gnidan: Future work for this file: we can use the ABIs for the various Assert type libraries and autogenerate Assert.sol

packages/truffle-core/lib/testing/soliditytest.js
```diff
@@ -94,81 +110,163 @@ var SolidityTest = {
...
-          callback();
-        }).catch(callback);
-      });
+      compile.with_dependencies(
+        runner.config.with({
+          paths: [
+            path.join(__dirname, "Assert.sol"),
+            path.join(__dirname, "AssertAddress.sol"),
+            path.join(__dirname, "AssertAddressArray.sol"),
+            // path.join(__dirname, "AssertAddressPayableArray.sol"),
```
gnidan: Since this is commented out, we probably want a comment why

packages/truffle-core/lib/testing/soliditytest.js
```diff
@@ -94,81 +110,163 @@ var SolidityTest = {
...
+    const AssertString = runner.config.resolver.require(
+      "truffle/AssertString.sol"
+    );
+    const AssertUint = runner.config.resolver.require("truffle/AssertUint.sol");
+    const AssertUintArray = runner.config.resolver.require(
+      "truffle/AssertUintArray.sol"
+    );
+    const DeployedAddresses = runner.config.resolver.require(
+      "truffle/DeployedAddresses.sol"
+    );
```
gnidan: Can we reduce the boilerplate here a bit?

Something like:
```javascript
const assertLibraries = [
  "AssertAddress",
  "AssertAddressArray",
  "AssertBalance",
  "AssertBool",
  "AssertBytes32",
  "AssertGeneral",
  "AssertInt",
  "AssertIntArray",
  "AssertString",
  "AssertUint",
  "AssertUintArray"
];

const assertAbstractions = assertLibraries.map(
  (name) => runner.config.resolver.require(`truffle/${name}.sol`)
);
```

Then, below, we can change `deployer` to use `await` syntax:

```javascript
await deployer.deploy(Assert);
for (let abstraction of assertAbstractions) {
  await deployer.deploy(abstraction);
}
/* ... continue into dependency_paths stuff ... */
```

What do you think?

packages/truffle-core/lib/testing/testsource.js
```diff
@@ -42,52 +48,199 @@ TestSource.prototype.resolve = function(import_path, callback) {
...
+  }
+
+  if (import_path === "truffle/AssertGeneral.sol") {
+    return fs.readFile(
+      path.resolve(path.join(__dirname, "AssertGeneral.sol")),
+      { encoding: "utf8" },
+      function(err, body) {
+        callback(err, body, import_path);
+      }
+    );
```
gnidan: This is another bunch of code that could use to be replaced with a list of library names + for loop.

packages/truffle/cli.webpack.config.js
```diff
@@ -107,7 +107,161 @@ module.exports = {
...
+      },
+      {
+        from: path.join(
+          __dirname,
+          "../..",
+          "node_modules",
+          "truffle-core",
+          "lib",
+          "testing",
+          "AssertUintArray.sol"
```
gnidan: Similarly these could probably be DRYed up a bit, but less important because it's just redundantly repeated data, not redundantly repeated code.

PR #1492, opened 2018-12-05
Upgrade ganache-core & ganache-cli to latest versions

CruzMolina:
Upgrade for truffle to use:
ganache-core: v2.3.0
ganache-cli v6.2.3

### DIFF
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -15,7 +15,7 @@
   "license": "MIT",
   "devDependencies": {
     "chai": "4.1.2",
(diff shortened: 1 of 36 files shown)


gnidan:
Note: I think we'll also need to include changing to `vmErrorsOnRPCResponse: false` instead of `noVMErrorsOnRPCResponse: true`? everywhere

CruzMolina:
@gnidan confused about changing `vmErrorsOnRPCResponse` to false. don't we want it to be true so that ganache returns a reason? I'm seeing tests specifically setting the RPCResponse to true in order to get a reason response.

PR #1493, opened 2018-12-05
Make the function depth workaround conditional on using solc of version <0.5.1

haltman-at:
This pull request makes the function depth contract call workaround conditional on the contract (or library) being called into having been compiled with a version of `solc` prior to `0.5.1`.  As of version `0.5.1`, `solc` now outputs corrected jump markings that no longer exhibit the problem that made the function depth workaround necessary; however, it remains necessary for any old contracts compiled prior to that, so it has been left in conditional on the compiler version.

In order to effect this, compiler information from the artifact files is now included in the `context` objects.  A new `evm` selector has been added to get the context not of the current call, but of the call about to be made (the logic previously in `evm.current.context` has been factored out into a function since they both work the same way); and a `solidity` selector has been added to get the compiler information and check whether it's `solc` of version prior to `0.5.1` in order to determine whether to invoke the workaround.

One little quirk is worth noting -- although this is really due to `solc` `0.5.1` and not to the changes here.  When calling a function externally, its parameters will now end up in `data.proc.assignments` twice -- once in the correct stackframe, and once in an otherwise unused stackframe.  Since these ghost parameters shouldn't affect anything and would be difficult to get rid of, I've left them alone.

Also, in order to keep the `functionDepth` similar to how it was prior to `0.5.1`, I've adjusted the default `functionDepth` from `1` to `0`.  This will introduce an offset when debugging contracts compiled with an older version of `solc`, but `functionDepth` is purely an internal thing anyway; this offset has no effect.

Finally, this pull requests updates the tests from version `0.5.0` to version `^0.5.0` (so, presently, that's `0.5.1`).  It also adjusts the `functionDepth` tests to be correct with `solc`'s new jump markings.  No new tests have been added in this pull request.


packages/truffle-debugger/lib/solidity/sagas/index.js
```diff
@@ -39,16 +39,25 @@ function* functionDepthSaga() {
...
+      (yield select(solidity.current.isContractCall))
+    ) {
+      //all these parentheses are necessary
       //HACK WORKAROUND
-      //because of the solc problem where contract method calls essentially
-      //return twice, we compensate by putting *two* inward jumps for such a
-      //call.  Note that this won't work if the contract method was previously
-      //placed in a function variable!  Those will continue to screw things up!
-      //But if a contract call is being made directly, we can detect that.
+      //because of the problem in solc <5.0.1 where contract method calls
```
gnidan: minor typo: `v0.5.1`

packages/truffle-debugger/lib/solidity/selectors/index.js
```diff
@@ -321,6 +323,21 @@ let solidity = createSelectorTree({
...
+     * Determines if the solidity version used for the contract about to be
+     * called was <0.5.1, to determine whether to use the above workaround
+     * Only call this if the current step is a call or create!
+     */
+    needsFunctionDepthWorkaround: createLeaf(
+      [evm.current.step.callContext],
+      context =>
+        context.compiler !== undefined && //would be undefined for e.g. a precompile
+        context.compiler.name === "solc" &&
+        semver.satisfies(context.compiler.version, "<0.5.1")
```
gnidan: I love using the `semver` library to compare semver values/constraints

packages/truffle-debugger/test/helpers.js
```diff
@@ -25,7 +25,7 @@ export async function prepareContracts(provider, sources = {}, migrations) {
 
   config.compilers = {
     solc: {
-      version: "0.5.0"
+      version: "^0.5.0"
```
gnidan: 👌

packages/truffle-debugger/lib/solidity/selectors/index.js
haltman-at: What, you think I want to make that comparison myself? :P

PR #1494, opened 2018-12-05
Strip mapping metadata from decoder output in debugger

gnidan:
- Recursively search through all decoder output and detect decoder's encoding for mappings.
- Strip out the wrapper info and convert the mapping to a JS `Map` object
- Use BN as `Map` key type when Solidity's mapping key type is some kind of `uint` or `int`

Previously:
screen shot 2018-12-04 at 9 20 03 pm

Now:

screen shot 2018-12-04 at 10 11 51 pm

### DIFF
--- a/packages/truffle-decode-utils/src/conversion.ts
+++ b/packages/truffle-decode-utils/src/conversion.ts
@@ -148,33 +148,6 @@ export namespace Conversion {
   }
 
   /**
-   * Convert a mapping representation into a JS Map
-   *
-   * Only converts integer types to BN right now, leaving other keys else alone
-   */
-  export function toMap(
-    { keyType, members }: any,
-    convertValue?: (value: any) => any
-  ): Map<any, any> {
-    // convert integer types to BN, unless string representation is hex
-    const convertKey = (key: string) => {
-      if (keyType.match(/int/) && key.slice(0,2) != "0x") {
-        return new BN(key, 10);
-      } else {
-        return key;
-      }
-    }
-
-    convertValue = convertValue || (x => x);
-
-    // populate Map with members, using converted keys/values
-    return new Map([
-      ...Object.entries(members)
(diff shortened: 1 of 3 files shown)


packages/truffle-decode-utils/src/conversion.ts
```diff
@@ -146,4 +146,72 @@ export namespace Conversion {
...
   }
-}
\ No newline at end of file
+
+  /**
+   * Convert a mapping representation into a JS Map
+   *
+   * Only converts integer types to BN right now, leaving everything else alone
+   */
+  export function toMap({ keyType, members }: any): Map<any, any> {
```
haltman-at: I'm confused -- it doesn't look like `toMap` calls `cleanMappings`, so it looks like if you have a mapping to mappings, the inner mappings won't get cleaned.  Am I mistaken?  What's going on here?  It doesn't look like it eventually gets passed to the `object` case.

gnidan: d'oh silly me

packages/truffle-decode-utils/src/conversion.ts
```diff
@@ -146,4 +146,77 @@ export namespace Conversion {
...
\ No newline at end of file
+
+  /**
+   * Convert a mapping representation into a JS Map
+   *
+   * Only converts integer types to BN right now, leaving other keys else alone
+   */
+  export function toMap(
+    { keyType, members }: any,
+    convertValue?: (value: any) => any
```
haltman-at: Hm, I'm a little confused what the point of this generality here is, with `convertValue` passed as an (optional) input parameter.  Seems to me it's only ever called with `convertValue` equal to `cleanMappings`.  This isn't wrong, but it's a little confusing and I think over-complicated.  I mean, leave it in if you have some reason for it, but otherwise I'd recommend removing `convertValue` and just invoking `cleanMappings`.

PR #1500, opened 2018-12-07
Add command to remove all breakpoints

haltman-at:
This pull request allows the user to remove all breakpoints by entering `B all`.  (`b all` will yield an error message -- you can't *add* all breakpoints.)

Also, it marks `B` as a non-repeatable command, which (like `b`) it obviously should be.


packages/truffle-core/lib/commands/debug.js
```diff
@@ -377,6 +377,17 @@ var command = {
               breakpoint.sourceId = currentSourceId;
             }
 
+            //the special case of "B all"
+            else if (args[0] === "all") {
+              if (setOrClear) {
+                // only "B all" is legal, not "b all"
+                config.logger.log("Cannot add breakpoint everywhere.\n");
```
gnidan: yay! programmer humor!

PR #1512, opened 2018-12-10
Handle asynchronous 3rd party plugins

daniyarchambylov:

PR #1519, opened 2018-12-10
Handle asynchronous 3rd party plugins (2.0)

### LINKED ISSUE #1482: 5.0.0-beta2 run plugins with callbacks not working
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle)
---------------------------

## Issue

I am not able to write a simple plugin that uses a callback and see the results of the callback. Here is plugin code:

```
const request = require('request');
module.exports = (config) => request('http://google.com', {}, (err, res, body) => {
    console.log("Hi, rocky!")
  })
```

Running this produces no output.

## Steps to Reproduce

Go through the plugin instruction at https://github.com/trufflesuite/truffle/releases and use the above code.

## Expected Behavior

`Hi, rocky`

## Actual Results

no output.

CruzMolina:
Completes #1512 & resölves #1482.

### DIFF
--- a/packages/truffle-core/lib/run.js
+++ b/packages/truffle-core/lib/run.js
@@ -54,9 +54,12 @@ const Run = {
   },
 
   // executes command or throws user helpful error
-  run(pluginConfigs, customCommand, config) {
-    let runCommand = this.initializeCommand(pluginConfigs, customCommand);
-    runCommand(config);
+  run(pluginConfigs, customCommand, config, done) {
+    const runCommand = this.initializeCommand(pluginConfigs, customCommand);
+    const commandResult = runCommand(config, done);
+    if (commandResult && typeof commandResult.then === "function") {
+      commandResult.then(() => done());
+    }
   }
 };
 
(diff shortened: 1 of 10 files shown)


packages/truffle-core/lib/run.js
```diff
@@ -54,9 +54,9 @@ const Run = {
   },
 
   // executes command or throws user helpful error
-  run(pluginConfigs, customCommand, config) {
-    let runCommand = this.initializeCommand(pluginConfigs, customCommand);
-    runCommand(config);
+  run(pluginConfigs, customCommand, config, done) {
+    const runCommand = this.initializeCommand(pluginConfigs, customCommand);
+    runCommand(config, done);
```
gnidan: We don't need the promise check? That surprises me because I'd expect that nothing would be calling `done` in that case

CruzMolina: 🔧

gnidan:
I got rid of the GH auto-close thingy cause we will want to QA the issue before it goes into the ol' circular file

Issue #1316, opened 2018-10-03
Network id configuration doesn't protect user from deploying to different network


tcoulter:
As we saw during the retreat demo, if you specify a specific number in your truffle config (say, `3`), if the actual network id of the chain doesn't match, Truffle won't error (and it should, to protect the user from inadvertently deploying with real Ether). 

Steps to repro:
1. Set Truffle config network id to 3.
1. Deploy to rinkeby (which has network id of 4). Notice Truffle doesn't error, and performs the deployment.

Note: This bug report was written very quickly because we're at the retreat demo. Feel free to ask me any questions if the above is unclear.

gnidan:
Not stale

gnidan:
Colony's build is [failing](https://circleci.com/gh/JoinColony/colonyNetwork/5323?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link), seems to be a result of the change in the above PR.

Relevant CI output:
```
...
Summary
=======
> Total deployments:   13
> Final cost:          0 ETH

Error: [object Object]
    at detectNetworkId (/home/circleci/colonyNetwork/node_modules/truffle/build/webpack:/packages/truffle-core/lib/environment.js:73:1)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:182:7)
Truffle v5.0.0-next.20 (core: 5.0.0-beta.2)
```

Looks like the error is constructed wrong [here](https://github.com/trufflesuite/truffle/blob/next/packages/truffle-core/lib/environment.js#L65-L74), although I might just be blanking on whether `{ message: "..." }` is kosher.

Hard to say if there's a real problem, we may have caught a mismatch in the test suite :)

cc @elenadimitrova

gnidan:
This is done and released in v5.0.0. Let us know if you run into it again! Thanks!

PR #1528, opened 2018-12-12
Edit network id check

eggplantzzz:
#1316 

Compare string version of network id and edit Error instantiation in order to allow a string or number representation of the network_id.

### DIFF
--- a/packages/truffle-core/lib/environment.js
+++ b/packages/truffle-core/lib/environment.js
@@ -62,7 +62,7 @@ const Environment = {
       const providerNetworkId = await web3.eth.net.getId();
       if (network_id !== "*") {
         // Ensure the network id matches the one in the config for safety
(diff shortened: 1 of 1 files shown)


packages/truffle-core/lib/environment.js
```diff
@@ -60,16 +60,14 @@ const Environment = {
...
         // Ensure the network id matches the one in the config for safety
-        if (providerNetworkId !== network_id) {
-          const error = {
-            message:
-              `The network id specified in the truffle config ` +
-              `(${network_id}) does not match the one returned by the network ` +
-              `(${providerNetworkId}).  Ensure that both the network and the ` +
-              `provider are properly configured.`
-          };
+        if (JSON.stringify(providerNetworkId) !== JSON.stringify(network_id)) {
```
gnidan: We should just use `.toString()` here. We want the loss of precision. Also, if `providerNetworkId` is a BN then we won't be able to JSONify that

Issue #1535, opened 2018-12-12
Suppress "Could not find config" warning for `truffle version`


gnidan:
Work to fix #1516 has now caused this warning to appear. We should suppress this warning.

```
$ truffle version
Could not find suitable configuration file.
Truffle v5.0.0-next.23 (core: 5.0.0-beta.2)
Node v8.12.0
```

eggplantzzz:
Fix merged PR #1539

PR #1539, opened 2018-12-13
Truffle version bug fix

eggplantzzz:
Handle the case where the user runs truffle version outside of a Truffle project

Fixes issue #1535

### DIFF
--- a/packages/truffle-core/lib/version.js
+++ b/packages/truffle-core/lib/version.js
@@ -26,9 +26,10 @@ const logNode = (logger = console) => {
   logger.log(`Node ${process.version}`);
 };
 
-const logSolidity = (logger = console, versionInformation, config = {}) => {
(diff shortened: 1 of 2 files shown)


packages/truffle-core/lib/version.js
```diff
@@ -26,9 +26,10 @@ const logNode = (logger = console) => {
   logger.log(`Node ${process.version}`);
 };
 
-const logSolidity = (logger = console, versionInformation, config = {}) => {
+const logSolidity = (logger = console, versionInformation, config) => {
   let solcVersion;
   if (
+    config &&
```
CruzMolina: Cool! so will the else clause in line 39 just return the solc version that's shipped w/ truffle-core?

eggplantzzz: Yes!

PR #1544, opened 2018-12-15
Add legacy `truffle migrate` and `truffle test`

### LINKED ISSUE #1543: Accommodate broken support for Quorum/other chains via `--legacy` flags to use old code
Discovered by @CruzMolina and @EdsonAlcala: Quorum support in Truffle seems to be broken with the new migrations system.

This is most directly observed in the failing `web3.eth.getBlock()` call on [this line](https://github.com/trufflesuite/truffle/blob/next/packages/truffle-migrate/migration.js#L186), where web3.js fails to decode the blog gas limit due to a web3-internal `bn.toNumber()` call (`Error: Number can only safely store up to 53 bits`), but there may be other problems.

As a last-minute effort to maintain Quorum support, @CruzMolina has proposed the idea of a legacy system, to take the code on `develop` and wire it up as an alternative for `truffle migrate` and `truffle test`.

gnidan:
To fix #1543 

(this is @CruzMolina's work, I'm just opening the PR so that I can open PRs into it)


CruzMolina:
Legacy migrations "complete".

Further potential work to polish it up:
- `runLegacyMigrations` method doesn't properly return a callback error as a TruffleError (yet). (Couldn't get it to work on my end).
- `truffle-contract` could potentially be streamlined further to not use legacy lib (might not be worth it, might be easy, might require more substantial surgery). Streamlining further _might_ only require using `legacy-web3` to do BigNumber conversations if options/config legacy is true. Might be more difficult than that.
- further streamline `truffle-deployer` to not use legacy lib. Tried a couple approaches that didn't work on my end, but it seemed not worth the effort.

CruzMolina:
TODO:

Streamline legacy testing.

It's a bit of a mess, but might not be too bad once re-written to build on top of the refactored legacy migrations.

packages/truffle-contract/index.js
```diff
@@ -1,12 +1,14 @@
...
-var contract = function(options) {
-  var binary = Schema.normalize(options || {});
+var contract = function(options = {}) {
+  var binary = Schema.normalize(options);
 
   // Note we don't use `new` here at all. This will cause the class to
   // "mutate" instead of instantiate an instance.
-  return Contract.clone(binary);
+  if (options.legacy) return LegacyContract.clone(binary);
+  else return Contract.clone(binary);
```
gnidan: the lack of braces here unnerves me

packages/truffle-contract/package.json
```diff
@@ -31,6 +31,7 @@
     "bignumber.js": "^7.2.1",
     "ethereumjs-util": "^5.2.0",
     "ethjs-abi": "0.1.8",
+    "legacy-web3": "^0.20.6",
```
gnidan: This strikes me as an anti-pattern. Could this be avoided by making a `truffle-contract-legacy` package instead? Possibly even make it to depend on `truffle-contract` to re-use commonality?

packages/truffle-core/lib/commands/migrate.js
```diff
@@ -135,25 +155,25 @@ var command = {
...
+        const environment = require("../environment");
+        const NewConfig = require("truffle-config");
         const config = NewConfig.detect(options);
 
         config.contracts_build_directory = buildDir;
         config.artifactor = new Artifactor(buildDir);
         config.resolver = new Resolver(config);
 
-        environment.detect(config, function(err) {
+        environment.detect(config, () => {
```
gnidan: (We probably shouldn't ignore errors from doing `Environment.detect()`... unless these get thrown)

packages/truffle-core/lib/test.js
```diff
@@ -2,10 +2,11 @@ var Mocha = require("mocha");
 var chai = require("chai");
 var path = require("path");
 var Web3 = require("web3");
+var LegacyWeb3 = require("legacy-web3");
```
gnidan: Ah this `legacy-web3` again. Now that this effort is looking more like an actual, maintainable, possibly long-term solution, I think stuff like this is worth avoiding. We shouldn't have to mirror the package just cause we want to avoid a name collision. Why not instead leverage Node itself and group legacy stuff into separate packages that define specific version dependencies?

packages/truffle-core/lib/testing/testrunner.js
```diff
@@ -1,4 +1,5 @@
 var Web3 = require("web3");
+var LegacyWeb3 = require("legacy-web3");
```
gnidan: Here's an idea: what if, instead of my above proposed `truffle-contract-legacy`, we create a single new package `truffle-legacy-system` package, minimally containing things like `truffle-contract/lib/legacy.js`, along with exposing `Web3` from v0.20.6?

Then we could do:

```javascript
const Legacy = require("truffle-legacy-system");

this.web3 = options.legacy ? new Legacy.Web3() : new Web3();
```

And then we could avoid duplicating `web3` as `legacy-web3` just for package namespace reasons. `truffle-legacy-system` would have its own namespace, specifying the old Web3.js.

packages/truffle-core/lib/testing/testrunner.js
```diff
@@ -30,21 +29,30 @@ function TestRunner(options) {
...
-};
+}
 
 TestRunner.prototype.initialize = function(callback) {
   var self = this;
 
+  if (self.web3.sha3) {
+    self.web3.utils = {};
+    self.web3.utils.sha3 = self.web3.sha3;
+  }
```
gnidan: (The stuff about `legacy-web3` is not me nitpicking, I swear. The comment I'm about to make here is nitpicking. This one you should feel free to roll your eyes at.)

This feels like a good candidate for moving to a function that has a semantic name, like `normalizeWeb3Interface(self.web3)` or `shimWeb3(self.web3)`. We could even put such a function inside my proposed `truffle-legacy-system` package :grin:

packages/truffle-core/lib/testing/testrunner.js
```diff
@@ -124,90 +144,117 @@ TestRunner.prototype.resetState = function(callback) {
...
   });
-},
-
-TestRunner.prototype.revert = function(snapshot_id, callback) {
-  this.rpc("evm_revert", [snapshot_id], callback);
-};
+}),
+  (TestRunner.prototype.revert = function(snapshot_id, callback) {
+    this.rpc("evm_revert", [snapshot_id], callback);
+  });
```
gnidan: Whoa what's up with this pair?

packages/truffle-deployer/legacy.js
```diff
@@ -0,0 +1,81 @@
...
+
+LegacyDeployer.prototype.deploy = function() {
+  var args = Array.prototype.slice.call(arguments);
+  var contract = args.shift();
+
+  if (Array.isArray(contract)) {
+    return this.queueOrExec(deployMany(contract, this));
+  } else {
+    return this.queueOrExec(deploy(contract, args, this));
+  }
```
gnidan: Is the only substantive difference here that it uses `deploy` / `deployMany` directly instead of inheriting `this.executeDeployment` from the `Deployment` base class?

If so, maybe we can do away with this file.

packages/truffle-core/lib/testing/testrunner.js
CruzMolina: ah, it was the comma on line 206 after the `snapshot` method.

packages/truffle-deployer/legacy.js
CruzMolina: iirc, the `Deployment` base class starts doing web3 1.0-specific things behind the scenes which causes legacy deployment to break. I tried a number of approaches to not have to pull in the main legacy truffle-deployer file, but never got it to work.

seesemichaelj:
@gnidan and @CruzMolina so here are the changes that I've made to the `legacy` for quorum support thing:
- `web3@0.20` was not necessary. the legacy web3 has been removed
- I removed the legacy deployer/contract files and integrated that behavior in the v5 deployer/contracts
- I kept the `runLegacyMigrations` function within `migration.js`
- I added a `truffle-interface-adapter` package which currently only has a web3 shim to allow us to monkey web3 functions. I replaced all `new Web3()` calls with this `new Web3Shim(provider)` layer
- I added a `--quorum` flag which enables `--legacy` and it also enables overloading web3's `outputFormatter` for the `getBlock` function to essentially ignore the formatted timestamp and use the original

some of this I'm not super fond of, but I needed to ship something, and this stuff will likely get removed during the adapter layer work I'll be doing shortly

what are your thoughts?

seesemichaelj:
I verified that the `--quorum` flag works for both `IBFT` and `RAFT` consensus algorithms. `IBFT` works without this PR.

CruzMolina:
As a side note, forgot to mention that I've been unable to get `IBFT` to work w/o this PR.

packages/truffle-interface-adapter/lib/web3-shim.js
```diff
@@ -0,0 +1,33 @@
...
+        block.timestamp = "0x" + timestamp.toString(16);
+        let result = _oldFormatter.call(web3.eth.getBlock.method, block);
+        result.timestamp = _oldTimestamp;
+        return result;
+      };
+    }
+
+    return web3;
+  }
+}
```
gnidan: I love me a good shim. So much utility! It's beautiful.

shim

packages/truffle-interface-adapter/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "truffle-interface-adapter",
+  "version": "1.0.0",
```
gnidan: Can we make this "0.1.0"?

packages/truffle-migrate/index.js
```diff
@@ -148,6 +148,35 @@ const Migrate = {
     };
   },
 
+  wrapLegacyProvider: function(provider, logger) {
```
gnidan: Looks like `wrapLegacyProvider` is no longer used anywhere. I think we can delete it.

packages/truffle-migrate/migration.js
```diff
@@ -97,50 +96,151 @@ class Migration {
...
+      await self.emitter.emit("error", payload);
       deployer.finish();
       callback(e);
+    }
+  }
+
+  // ------------------------------------- Private -------------------------------------------------
+  /**
+   * Runs the legacy migration sequence by instantiating a
+   * legacy deployer using a legacy web3 provider.
```
gnidan: I don't think the last bit about legacy web3 provider is true anymore? (I'm assuming this code used to reference that `wrapLegacyProvider` function)

gnidan:
Oh, another note: -0.8% code coverage decrease is quite significant. Coveralls fails for -0.5% or worse. I think we should respect that failure and not merge this without some additional tests.

packages/truffle-contract-schema/spec/contract-object.spec.json
```diff
@@ -54,14 +54,23 @@
...
-    "userdoc": { "$ref": "#/definitions/NatSpec" }
+    "userdoc": { "$ref": "#/definitions/NatSpec" },
+    "legacy": {
+      "type": "boolean",
+      "description": "Pass-through flag that specifies if this contract should use Truffle's legacy migrations"
+    },
+    "quorum": {
+      "type": "boolean",
+      "description": "Pass-through flag that specifies if this contract should use Quorum workarounds"
+    }
```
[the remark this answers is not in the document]
seesemichaelj: I believe its not worth the time as my upcoming work to implement `truffle-interface-adapter` will deprecate this interface regardless, so I don't want to spend too much time making this proper and future-proof when it'll get scrapped near term anyway.

packages/truffle-contract/package.json
```diff
@@ -30,6 +30,7 @@
     "truffle-blockchain-utils": "^0.0.8",
     "truffle-contract-schema": "^3.0.2",
     "truffle-error": "^0.0.4",
+    "truffle-interface-adapter": "^1.0.0",
```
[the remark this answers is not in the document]
seesemichaelj: Sure :)

packages/truffle-contract-schema/spec/contract-object.spec.json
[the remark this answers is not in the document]
seesemichaelj: Further, I've already spent a good amount of time on this and would like to move forward since this PR is a bandaid in principal

seesemichaelj:
Closing this PR in favor of a new PR which shifts this code to accept the following interface:
in `truffle(-config)?.js`:
```
...,
networks: {
  development: {
    type: "quorum" | ...,
    ...
  }
}
```
rather than having it at the command-line level (i.e. `--quorum`)

PR #1572, opened 2018-12-20
Update unbox process

eggplantzzz:
Use truffle-box.json to verify boxes rather than truffle.js


CruzMolina:
Looks like the box.js sandbox method needs to be updated here: https://github.com/trufflesuite/truffle/blob/update-unbox/packages/truffle-box/box.js#L44

eggplantzzz:
Yessir

CruzMolina:
At, looks like it's the `truffle-debugger` tests at this point. I think whatever Migrations.sol file they're pulling down need to be updated to have the inclusive pragma: `pragma solidity >=0.4.21 <0.6.0;`

PR #1573, opened 2018-12-20
Move the debugger to preallocation of structs, along with other small allocation and decoding fixes

haltman-at:
The primary purpose of this pull request is to move the debugger to using preallocation of structs, like Ganache does, so that the same decoding code can be used for both; this pull request deletes the debugger-specific code for decoding of structs, as the debugger will now use the code that was previously used only for Ganache.  Also, in addition to preallocating structs, contracts are now also preallocated as well.  All of this is, of course, a purely non-functional change (except perhaps when it comes to performance; I don't know that this will really make much difference but it should ideally frontload any costs there and make them one-time); but I also included several functional changes along with it that I deemed easy to incorporate.

(@seesemichaelj also made some changes on this branch, most of which I'm not entirely familiar with; I'll let him comment with his changes below.)

Firstly, the debugger now supports decoding for contracts with inheritance -- inherited state variables will be passed to the allocator along with those from the contract itself.

Secondly, the debugger now properly handles allocation for contracts with constant variables; constant variables are now skipped during allocation, as they should be (this is done by filtering them out before passing them to the allocator; this isn't handled by the allocator itself).  Note that there is presently no support in the debugger for *decoding* of constant variables; I'm intending to add this at a later time.  (Although, this isn't high-priority -- there isn't much need to decode constants, because they don't change.)

Thirdly, enums can now be decoded in any location rather than just in storage.   (Thanks to @seesemichaelj for catching a bug here.)  Currently they decode to an internal object that the user shouldn't really see, but, hey, that's what they did already in storage.  I'm intending to fix this in a later PR.

Fourthly, I did some tinkering with the struct decoding logic, because I noticed what appeared to be some errors in it.  And hey, @seesemichaelj says it's working better now, so that's good. :)

Also, in another nonfunctional change, I deleted the now-defunct file `mapping.ts`.

Finally it is worth noting that I did not add any new automated tests in this PR.  Probably there should be some, but they would be decoding tests, and I think it would make more sense to redo the decoding tests in a later PR.

I have a number of additional notes on the changes above, mostly regarding the preallocation change:

1. Let's first note what this PR does *not* do -- it does not rewrite the debugger's allocator.  The point of this PR was that it was something I could do *without* substantially altering the allocator (there was one small alteration, see next point).  This means that the debugger still cannot handle static arrays.  I'm intending to rewrite the allocator, and fix static arrays, in my next PR, along with some other improvements.
2. In order to avoid recomputation, we need to somehow memoize the allocations for structs we've already allocated.  Again, I didn't want to alter the allocator much, and I also didn't want to go adding a bunch to the state here; so instead what I did is this.  I added an extra argument to the allocator which allows it to take an object containing allocations already made.  If it goes to allocate a struct, first it looks it up in there to see if it's already been allocated; if so, it just uses the allocation it just looked up rather than recomputing it.  Now, this could still lead to recomputation; but to prevent this, in addition, I topologically sorted the structs and contracts to be allocated, and allocated them in that order.  (I think this is different than the approach Ganache uses, which is to simply memoize *everything* as it allocates it, and not to worry about order; @seesemichaelj can correct me if I'm wrong there.)  This actually means that the recursive case in the allocator will now never be called.  However, I've left it in for now as something of a failsafe.  However, in my next PR, when I rewrite the allocator, I won't bother to include such a failsafe.
3. Regarding that topological sort -- I used a fairly naïve, unoptimized topological sort.  I don't think this should be a performance problem, but if it becomes one, well, we can address it then.  (Getting any performance improvement from using a better one would have required making more of an explicit graph structure rather than just using lists, which I judged not worth the trouble.)
4. I added to the state `data.info.userDefinedTypes` (along with corresponding selector).  This lists all the user-defined types, such as structs, contracts, and enums.  The thing is, I just stored the ID of the type, not the definition or the entry from `data.info.scopes`.  Because the thing we care about here is mostly the `variables` member from `data.info.scopes`; but I'm not sure `data.info.scopes` is ready yet when `data.info.userDefinedTypes` is set up (although that entry might be?), so I chose to store just the ID, and fetch related information later from it.
6. Speaking of which, `data.info.scopes` has now been changed to account for inheritance and constants; the old one is now at `data.info.scopes.raw`.  Similarly with `data.views.scopes.inlined`.  Now, this means that `data.info.scopes` now contains substantial computation, meaning it probably really belongs in `data.views` now, not `data.info`; however since I intended the new one as something of a drop-in replacement to this commonly-used selector, I didn't want to do that.  I'm intending to reorganize this though to how it should be in a later PR (probably when I handle decoding constants or globally available variables).
7. Finally, one note that is not about allocation but rather about decoding -- as mentioned, I made some changes to struct decoding.  Let's say we have a struct; we want to decode it, which means decoding each of its members, which means determining pointers for each of its members.  Each slot in each of these child pointers has a path and an offset.  Now, the path is obviously the slot of the parent's `from`; but what about the offset?  Here's why this is an issue -- the offset should be the offset of that member, as determined by the allocator; except, that the member might itself have a nontrivial `path`.  The `offset` field is supposed to be a number, so we can't just stick this all this information in `offset`; we have to collapse it to a single overall offset with `slotAddress`, which I do.  This loses information, despite @gnidan's intent to not lose information that led to the `path` system in the first place.  Despite that, I think this is OK -- because, well, it's a struct offset, not a path to the base.  I think keeping that information isn't really very necessary for something that's ultimately just going to used as an offset; certainly there's no hashing going on here that would confuse things!  So, I went ahead and collapsed it with `slotAddress`, as needed.


seesemichaelj:
I made a comment with my main contribution. I also modified the decoder tests to test more features.

haltman-at:
OK, so, there's been a lot of confusion about the `slotAddress` thing.  I don't think people have properly understood in just what situation this becomes relevant or why I think it's the correct thing to do there.

Note that it's very possible that this situation never occurs in Ganache at all -- in the debugger, it's due to the allocator behaving in a way that I think is not good, and which I'm hoping to address in my next PR; and it's very possible that Ganache just doesn't do this and so will never encounter this situation.  In any case, I'm going to talk about things from a debugger-centric point of view, as that's what I'm more familiar with and what I tested with.

So.  Let's imagine we have some big, complex structure we want to decode.  Some horrible multidimensional mix of static arrays, dynamic arrays, and mappings.  (We'll leave structs out of this for now.)  Maybe it's a `mapping(string => uint[][5])`, or something.  We can imagine this structure as a tree.  The structure itself is the root node; its elements are the children, with the edges labeled by the keys; and then the children have children of their own, and so on, down until the leaves which are value types (in this case, `uint`).  If we want to decode a specific leaf, such as `structure["hello"][4][37]`, we do this by walking down the tree, starting at the root.  The address in storage of this leaf is represented, then, by a pointer with a `path` depth of 3, because it's 3 levels down; each link in the chain of `path`s represents going down some edge, with the `offset` at that level (or `key`, for mappings) telling us just which edge we're going down (obviously `hashPath` plays a role too).  So far, so simple.

Now, let's introduce structs to the mix.

There's a famous way of representing arbitrary trees as binary trees; if you've ever, say, implemented a trie in C, you're familiar with it.  It looks something like this:
```
typedef struct trie {
  trie *first;
  trie *next;
} trie;
```
Right?  The `first` pointer points to the node's first child, while the `next` pointer points to the parent'sn next child.  So if you want all of a given node's children, you follow its `first` pointer, and then the `next` pointers give the children in a linked list.  We've represented an arbitrary tree as a binary tree, with each node storing just two pointers.  But note that this means now, that if we want to step down the tree from a particular node to a particular child, we are, in this representation, not just taking a single step down -- we're taking one step down, followed by a number of steps to the side.  But also note that these steps to the side exist only in the binary tree representation; conceptually, in the original tree, we're still just taking a single step down.

So what does this have to do with structs?  Well, when the debugger's allocator allocates a struct (or contract), it *doesn't* just use `path` for representing steps down, as you might expect.  It *also* uses `path` for representing these steps to the side.  This muddies things up considerably.  Conceptually, each of a struct's members is just a single step down from the struct itself; there shouldn't be anything nontrivial going on in the path there.  But the debugger's allocator, rather than returning instructions for how to step directly downward from the struct to a given member, in the form of a single offset, instead returns instructions for how to step sideways from member to member.

Unsurprisingly, I don't think this is the right way of doing things.  These sideways steps fundamentally do not exist in the conceptual picture we want to draw, that of stepping down from parent node to child node; they're purely an artifact of the representation.  And so I don't think we should be storing them -- each link in the `path` chain should always represent a real step downwards, not a fictional step sideways from member to member.  Individual members within a struct should be represented by a single step downward from the struct.  But, while I don't think we should have these fictional sideways steps in our struct allocation, for the moment we do, and addressing that will be a separate PR.

So why does this cause a problem, and why is `slotAddress` the solution?  Well, imagine once again that we're walking down our structure.  At some level we hit a struct, and we want to walk down one of its edges and form a pointer to the child node below.  How do we form that pointer?  If it truly were just one direct step down -- as it is for static arrays, dynamic arrays, and mappings -- we'd just make a new pointer with the parent node as our `path` and the offset telling us which edge we're following, as usual.  But instead, when we go to look up the way to this child node, we find something with nontrivial `path` -- we find a chain of *multiple* steps from the struct to its member, even though conceptually there's only one, most of these steps being fictional sidways steps.

So what do we do here?  How do we encode this into our `Slot` datatype?  We have several options.

One possibility would be to say, well, maybe we should allow such things as offsets, rather than just simple integers.  I think this is complicated and unnecessary.  We are, conceptually, taking a single step down from the struct to its member -- a single integer should suffice for that fine.  There are no real intermediate steps worth storing, only fictional sideways ones from member to member.

A second possibility would be to take this path information that we've found and put it in the `path`.  But this would put things in the `path` that don't really belong there, putting in fictional sidweays steps -- to members that aren't even the one we're looking for -- alongside real downward ones.  As I said above, the depth of the `path` chain should be equal to our current depth in the tree.  Taking this `path` information and putting it in our `path` would destroy that invariant, muddying up the `path` with steps that conceptually don't exist.

The third possibility is the one I took: Take these fictional sideways steps and collapse them (along with the one downwards one), with `slotAddress`, into a single offset, corresponding to the single, real, step downwards that we are making, the step from struct to member.  This is why I said earlier that even though the use of `slotAddress` makes it appear that we are losing information, any loss is purely illusory.  The point of `path` is to make sure we do not lose information about the actual steps downward, but these are fictional sideways steps, ones that often go to members unrelated to the one we're looking for, and which don't belong in the picture in the first place; and as I said, I intend to rewrite the allocator to eliminate them.  But as things currently are, where a single conceptual step downwards can be represented internally by one step downwards to the wrong member followed by many steps sideways until we reach the right one, the best thing to do is to collapse all these into the single downards step, from the struct to the correct member, that should actually exist in our picture; and not include these sideways steps or steps to the wrong member at all.

In short, despite the apparent loss of information, the use of `slotAddress` loses no actual information, and is, in cases such as these, the correct way to get the internal picture to line up with the intuitive picture; the information it loses is purely a representational fiction that should never have existed in the first place.  None of the actual path information we do care about is lost.

gnidan:
(IIRC, tests should pass once this branch pulls from upstream)

packages/truffle-decode-utils/src/conversion.ts
```diff
@@ -149,19 +149,29 @@ export namespace Conversion {
...
+   * (eventually it will probably also have to clean arrays, but currently
+   * those don't require cleaning)
    *
    * detects int and uint Solidity types and uses BNs as keys in such
    * situations
    *
    * all other keys are just kept as strings right now
    */
-  export function cleanMappings(value: any): any {
+  export function cleanContainers(value: any): any {
```
gnidan: 🎉 better names yay

packages/truffle-decoder/lib/decode/storage.ts
```diff
@@ -201,20 +202,20 @@ export default async function decodeStorageReference(definition: DecodeUtils.Ast
...
-        //debug("pointer %o", pointer);
-        //debug("refPointer %o", refPointer);
+        debug("pointer %O", pointer);
+        debug("refPointer %O", refPointer);
         const childRange = <Allocation.Range>{
           from: {
             slot: {
               path: pointer.storage.from.slot,
-              offset: refPointer.storage.from.slot.offset.clone()
+              offset: slotAddress(refPointer.storage.from.slot)
```
gnidan: the controversy

packages/truffle-debugger/lib/data/reducers.js
```diff
@@ -59,8 +59,25 @@ function scopes(state = DEFAULT_SCOPES, action) {
...
 }
 
+//a note on the following reducer: solidity assigns a unique AST ID to every
+//AST node among all the files being compiled together.  thus, it is, for now,
+//safe to identify user-defined types solely by their AST ID.  In the future,
+//once we eventually support having some files compiled separately from others,
+//this will become a bug you'll have to fix, and you'll have to fix it in the
+//decoder, too.  Sorry, future me! (or whoever's stuck doing this)
+
+function userDefinedTypes(state = [], action) {
```
gnidan: How do you feel about just calling this `types`? `userDefined` seems semi-extraneous, but I could go either way.

packages/truffle-debugger/lib/data/selectors/index.js
```diff
@@ -13,6 +13,10 @@ import solidity from "lib/solidity/selectors";
 import * as TruffleDecodeUtils from "truffle-decode-utils";
 import { forEvmState } from "truffle-decoder";
 
+function shallowClone(object) {
+  return Object.assign({}, object);
+}
```
gnidan: How bout `{ ...object }` instead of defining this helper function?

packages/truffle-decode-utils/src/allocation.ts
```diff
@@ -107,18 +108,29 @@ export namespace Allocation {
...
+    //all any structs it refers to have been allocated.  Nonetheless, I'm going
+    //to leave this case in for now as something of a failsafe.
+
+    //debug("Warning!  Struct failsafe invoked!")
 
     let struct = refs[definition.typeName.referencedDeclaration];
     // debug("struct: %O", struct);
 
-    let result =  allocateDeclarations(struct.variables || [], refs, slot, index);
+    let result = allocateDeclarations(struct.variables || [], refs, allocations, slot, index);
```
gnidan: I would be in favor of removing this "dead" code, and instead: reverse the `if` on L119, throwing an exception that says "Allocation not found" or somesuch. For users that have analytics enabled, the exception will get recorded there so we can see how often it happens.

packages/truffle-decoder/lib/decode/storage.ts
gnidan: (thanks for repeatedly explaining how this is benign!)

haltman-at:
At the suggestion of @gnidan, the "failsafe" mentioned above has been removed; an exception is now thrown instead.

packages/truffle-decode-utils/src/allocation.ts
haltman-at: Done.

packages/truffle-debugger/lib/data/selectors/index.js
haltman-at: Done.

packages/truffle-debugger/lib/data/reducers.js
haltman-at: Going to leave this one as-is, since it is indeed just the user-defined types.

PR #1577, opened 2018-12-21
Default Testing network_id to match GanacheUI

CruzMolina:
Per DavidBurela's feedback: https://twitter.com/DavidBurela/status/1075961475145949185

Defaults `truffle test` to use GanacheUI's default network_id when no network configuration is defined. The network will currently be defined as `test`.

If it's a must for UX/UI reasons to have the network logged as `ganache`, then probably something along the lines of setting a boolean in options during the workflow and passing that val around will need to be done.

### DIFF
--- a/packages/truffle-core/lib/commands/test.js
+++ b/packages/truffle-core/lib/commands/test.js
@@ -150,7 +150,7 @@ var command = {
           var ganacheOptions = {
             host: "127.0.0.1",
             port: 7545,
-            network_id: 4447,
+            network_id: 5777,
             mnemonic:
               "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat",
             gasLimit: config.gas,


CruzMolina:
Should pass tests and build successfully once the leftover v4 test stuff is updated.

gnidan:
(I have a reminder to come up with my thoughts on the requirements for this. Will circle back.)

CruzMolina:
Following up on this, don't recall why it would be an issue to change the network_id to match GanacheUI's, but I'm probably forgetting something.

eggplantzzz:
@CruzMolina So did we have any followup thoughts on this? It looks like it might be fine?

CruzMolina:
@eggplantzzz This still needs to be smoke-tested. It's been a while, but I believe there are a couple edge cases to consider. Will probably need to document pros/cons of how things work now vs how they work on this branch.

CruzMolina:
So, after QAing this branch a bit, I haven't been able to find any bugs for this.

Current downsides have been well-documented in #1726. i.e. running `truffle test` out of the box w/ GanacheUI results in:
```
Error: The network id specified in the truffle config (4447) does not match the one returned by the network (5777).  Ensure that both the network and the provider are properly configured.
    at Object.detectAndSetNetworkId (/Users/cruzmolina/Code/truffle-projects/truffle/packages/truffle-core/lib/environment.js:100:15)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
```

Pinging @gnidan before merging since it seems like he had a thought or two about this.

CruzMolina:
Closing this since I believe we _don't_ actually want `truffle test` to default to GanacheUI. IIRC, `truffle test` is meant to spin up a private local development testing chain by default. This behavior _can_ be overridden in the config or via `truffle test --network someNetwork`.

One possible solution is to have `truffle test --network ganache` automatically connect to GanacheUI's default settings w/o anything needing to be specified in the config. Another is to update `init` and official truffle boxes to have a `ganache` network pre-defined w/ GanacheUI's default settings.

PR #1579, opened 2018-12-21
Default Develop port to match metamask & ganache-cli default port

CruzMolina:
Per @adrianmcli's feedback

Defaults `truffle develop` to use Metamask & ganache-cli's default port (8545) when no network configuration is defined.

Should allow a user to unbox a `drizzle` box, call `truffle develop`, and connect using MM out of the box.

(Should pass tests and build successfully once v4 test leftovers are updated)

### DIFF
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -68,7 +68,7 @@ var command = {
       numAddresses
     );
 
-    var onMissing = function(name) {
+    var onMissing = function() {
       return "**";
     };
 
@@ -83,7 +83,7 @@ var command = {
 
     var ganacheOptions = {
       host: "127.0.0.1",
-      port: 9545,
+      port: 8545,
       network_id: 4447,
       total_accounts: numAddresses,
       default_balance_ether: defaultEtherBalance,
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_install:
   - docker pull ethereum/solc:0.4.22
   - sudo add-apt-repository --yes ppa:ethereum/ethereum
   - sudo apt-get update
-  - sudo apt-get install solc jq snapd
+  - sudo apt-get install -y dpkg && sudo apt-get install solc jq snapd
   - export PATH=$PATH:/snap/bin
   - sudo snap install vyper --edge --devmode
 


adrianmcli:
Gotta make sure that all the boxes are updated (config, docs, client, etc.) after this is merged.

packages/truffle-core/lib/commands/develop.js
```diff
@@ -83,7 +83,7 @@ var command = {
 
     var ganacheOptions = {
       host: "127.0.0.1",
-      port: 9545,
+      port: 8545,
```
gnidan: Can we do this! If Ganache is already running, then `truffle develop` will fail to start, no?

adrianmcli: It would be prudent to error out and let the user know something like "Looks like you already have Ganache running on port 8545. Either change the port or turn off Ganache and try the command again."

gnidan: The whole point of `truffle develop` is that it's low friction. That's why we chose 9545 for `truffle develop` in the first place and 7545 for Ganache UI

CruzMolina: @gnidan in v5/next/develop the current behavior is that `truffle develop` doesn't clash (out of the box) with ganache-cli or ganacheUI.

on this branch/PR, `truffle develop` doesn't clash with ganacheUI out of the box, but if `ganache-cli` is already running and you try to spin up `truffle develop`, it'll clash and throw an error (because of the timestamped network_id's ganache-cli is given).  However, if you spin up `truffle develop` and then start up `ganache-cli`, truffle will auto-connect to `ganache-cli`.

gnidan:
(This is awaiting discussion / requirements gathering)

CruzMolina:
Closing this for now as #1773 seems to be the current reasonable compromise. 😎

adrianmcli:
@CruzMolina Cool to see this being addressed. Was just wondering if you don't mind outlining what would be the recommended path forward for teaching beginners then?

CruzMolina:
@adrianmcli I just looked over the "Getting Started With Drizzle And React" tutorial.

There's a discrepancy w/ MM's UI (on 6.2.2 on Chrome). 

Screen Shot 2019-03-19 at 9 37 37 AM

Currently you have to set `port: 8545` in `truffle-config.js` before running `truffle develop` to get MM to connect to the local chain.

If their intent is to support `localhost 8545` then the tutorial would need to be updated to include setting your config like so:

```
module.exports = {
    develop: {
      port: 8545
     // doesn't need other network configs
     // safe localhost default params are used
    }
  }
};

CruzMolina:
Opened an issue w/ MM regarding the above noted issue at: https://github.com/MetaMask/metamask-extension/issues/6321

PR #1582, opened 2018-12-24
docs: update site url references to https

codfish:
Site links weren't working for me initially. I kept getting 5xx errors. Now they seem to be redirecting properly though, but might as well fix these links anyway.


gnidan:
Thanks for this!

PR #1594, opened 2018-12-26
Use Strict Equality Checking [Kale]

DomAmato:
Strict equality checking can help prevent unintended behavior and improves performance since no type conversion takes place when evaluating. For a more in depth discussion around the subject see this stackoverflow discussion: https://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons


DomAmato:
Indeed, the strict null equality check seems to generally break code since null == undefined is a pretty common approach. I had to change that one line manually since our tool wrongfully flagged it for change 😛

gnidan:
Great, thanks @eggplantzzz! Merging!

PR #1624, opened 2019-01-09
Refer to the default config file name in error messages and hints

shane-t:
The default filename for the truffle config is `truffle-config.js`, this changes the name in the hints and errors to reflect that and reduce confusion.


eggplantzzz:
@shane-t Ah yes this is a good idea! Thanks!

PR #1628, opened 2019-01-11
Update compiler supplier

eggplantzzz:
Restructure how the CompilerSupplier functions in Truffle.

Much of the CompilerSupplier has been extracted into loading strategies. Each inherits from a base class (LoadingStrategy) and has its own way of handling loading the solidity compiler.

Updates were also made to allow for offline compilation when the compiler is cached on a user's machine.


packages/truffle-compile/compilerSupplier/index.js
```diff
@@ -0,0 +1,96 @@
...
+        strategy = new Bundled(this.strategyOptions);
+        return resolve(await strategy.load(userSpecification));
+      }
+      if (useSpecifiedLocal) {
+        strategy = new Local(this.strategyOptions);
+        return resolve(await strategy.load(userSpecification));
+      }
+      if (isValidVersionRange) {
+        strategy = new VersionRange(this.strategyOptions);
+        return resolve(await strategy.load(userSpecification));
```
gnidan: We can move `strategy.load()` out of the block of `if` statements.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -0,0 +1,205 @@
...
+        "https://registry.hub.docker.com/v2/repositories/ethereum/solc/tags/",
+      cache: true
+    };
+    this.config = Object.assign({}, defaultConfig, options);
+    this.cachePath = findCacheDir({
+      name: "truffle",
+      cwd: __dirname,
+      create: true
+    });
+  }
```
gnidan: There should be an abstract `load` method here to indicate that it's the external interface.

Something like:

```javascript

  async load (userSpecification) {
    throw new Error("Abstract method `loadingStrategy.load` is not implemented for this strategy.");
  }
```

gnidan: There should also be docstrings for every public method here, vs. the methods that are only used by other loading strategies.

packages/truffle-compile/compilerSupplier/loadingStrategies/Docker.js
```diff
@@ -0,0 +1,86 @@
...
+    if (!fileName) throw this.errors("noVersion", version);
+
+    const url = this.config.compilerUrlRoot + fileName;
+    const spinner = ora({
+      text: "Downloading compiler",
+      color: "red"
+    }).start();
+
+    return this.getSolcByUrlAndCache(url, fileName, spinner);
+  }
```
gnidan: Can we get rid of this implementation here and instead just construct a `VersionRange` strategy inside the Docker implementation?

packages/truffle-compile/compilerSupplier/loadingStrategies/Native.js
```diff
@@ -0,0 +1,57 @@
...
+    if (!fileName) throw new Error("No matching version found");
+
+    const url = this.config.compilerUrlRoot + fileName;
+    const spinner = ora({
+      text: "Downloading compiler",
+      color: "red"
+    }).start();
+
+    return this.getSolcByUrlAndCache(url, fileName, spinner);
+  }
```
gnidan: (Same comment as with Docker: can we move this logic into VersionRange so that Native just uses that?)

packages/truffle-compile/compilerSupplier/loadingStrategies/VersionRange.js
```diff
@@ -0,0 +1,100 @@
...
+const ora = require("ora");
+const semver = require("semver");
+const LoadingStrategy = require("./LoadingStrategy");
+
+class VersionRange extends LoadingStrategy {
+  load(versionRange) {
+    const rangeIsSingleVersion = semver.valid(versionRange);
+    return rangeIsSingleVersion
+      ? this.getSolcBySingleVersion(versionRange)
+      : this.getSolcByVersionRange(versionRange);
```
gnidan: Is the reason for this distinction because of performance?

I vote for a comment here to say why we treat the special case specially.

packages/truffle-compile/compilerSupplier/loadingStrategies/VersionRange.js
```diff
@@ -0,0 +1,100 @@
...
+    const cachedCompilerFileNames = fs.readdirSync(this.cachePath);
+    const cachedVersions = cachedCompilerFileNames.map(fileName => {
+      const match = fileName.match(/v\d+\.\d+\.\d+.*/);
+      if (match) return match[0];
+    });
+    return cachedVersions.find(cachedVersion =>
+      semver.satisfies(cachedVersion, version)
+    );
+  }
+}
```
gnidan: It's hard to follow what all these methods do and how they fit together. (e.g. which ones are used for exact-version and which ones are used for version-range?). Is there a way to separate this, or make it more clear on what's common between the two and what's different?

packages/truffle-compile/compilerSupplier/loadingStrategies/VersionRange.js
eggplantzzz: Yes, for the case where a single version is specified we will go straight to the cache to check if it is there. I'll add a comment!

packages/truffle-compile/compilerSupplier/index.js
eggplantzzz: Hmm, how about a method that returns the strategy and...
```
if (strategy) {
  strategy.load(userSpecification);
} else {
  throw Error....
```?

File: packages/truffle-compile/compilerSupplier/loadingStrategies/Docker.js (hunk 3 above)
Comment (eggplantzzz) [reply packages/truffle-compile/compilerSupplier/loadingStrategies/Docker.js]: So if I understand you correctly, the idea would be to make Docker and Native a subcase of VersionRange?

File: packages/truffle-compile/compilerSupplier/index.js (hunk 1 above)
Comment (gnidan) [reply packages/truffle-compile/compilerSupplier/index.js]: Doesn't need to be a separate method, but yeah, just at the end have that `if(strategy)` check + load.

File: packages/truffle-compile/compilerSupplier/loadingStrategies/Docker.js (hunk 3 above)
Comment (gnidan) [reply packages/truffle-compile/compilerSupplier/loadingStrategies/Docker.js]: Not a subcase, but instead to make them compose the VersionRange behavior.

So inside `load`, something like:
```javascript
const solcjs = new VersionRange().load(userSpecification);

/* ... */
return {
  /* compile, etc. */
  importsParser: solcjs
}
```

packages/truffle-compile/compilerSupplier/loadingStrategies/VersionRange.js
eggplantzzz: Yes, I agree. I'll think about this.

packages/truffle-compile/scripts/test.sh
```diff
@@ -3,7 +3,7 @@
 set -o errexit
 
 if [ "$CI" = true ]; then
-  mocha --timeout 10000 $@
+  mocha ./test/** ./test/**/* --timeout 10000 $@
 else
-  mocha --invert --grep native --timeout 10000 $@
+  mocha ./test/** ./test/**/* --invert --grep native --timeout 10000 $@
```
gnidan: Why remove those args?

gnidan:
Oh that approval is pending Travis

packages/truffle-compile/scripts/test.sh
eggplantzzz: Did you mean "why not remove those args"? If that was your question, mocha doesn't seem to dig down far enough in the test directories without those.

Issue #1593, opened 2018-12-26
BigNumber configuration not working in Truffle v5


wbt:
---------------------------

## Issue
In the Truffle v5 Release Notes, the breaking changes section [says](https://github.com/trufflesuite/truffle/releases/tag/v5.0.0#user-content-breaking-changes-truffle):   

> Numeric return values are now BN objects instead of the previously default BigNumber.js. These two projects use significantly different API semantics, so you may want to review those.
>Truffle provides a compatibility mode for projects unable to make the switch wholesale. Set numberFormat to "BigNumber", "String", or "BN" to choose.  See example number format override:

    // Choices are:  `["BigNumber", "BN", "String"].
    const Example = artifacts.require("Example");
    Example.numberFormat = "BigNumber";

I had previously developed a test suite that includes `someBigNumber.plus(1)` and similar lines.
Upgrading to Truffle 5 and running `truffle test`, the tests failed with messages like `TypeError: someBigNumber.plus is not a function`. 

Adding the `MyContract.numberFormat = "BigNumber";` line after ...`require("MyContract");` per the example, I now get the error:

```
TypeError: BigNumber is not a constructor
      at _convertNumber (C:\Users\me\AppData\Local\node\node_modules\truffle\build\webpack:\packages\truffle-contract\lib\reformat.js:17:1)
      at abiSegment.forEach (C:\Users\me\AppData\Local\node\node_modules\truffle\build\webpack:\packages\truffle-contract\lib\reformat.js:80:1)
      at Array.forEach (<anonymous>)
      at Function.numbers (C:\Users\me\AppData\Local\node\node_modules\truffle\build\webpack:\packages\truffle-contract\lib\reformat.js:47:1)
      at Promise (C:\Users\me\AppData\Local\node\node_modules\truffle\build\webpack:\packages\truffle-contract\lib\execute.js:125:1)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:160:7)
```

## Steps to Reproduce
Develop tests using BigNumber.   
Attempt to use the documented compatibility mode to not have to change over everything in the test suite immediately to Truffle v5.    
Limit tests to one file and include the `.numberFormat` line as documented in release notes/above.    
Encounter errors as shown above.

## Expected Behavior
No errors. With the adaptation, the test suite runs as before. 

## Environment
* Operating System: Win10 Pro (Username changed in output above.)
* Truffle version (`truffle version`): 5.0.0
[...]

PR #1631, opened 2019-01-14
Edit the way bignumber.js is required

eggplantzzz:
For the bundled version of Truffle (cli.bundled.js), when a contract instance was created from an abstraction that had the `numberFormat` "BigNumber" set, it would throw an error. --> https://github.com/trufflesuite/truffle/issues/1593

On doing some research, it seems others were experiencing the same issue when bundling with webpack https://github.com/MikeMcl/bignumber.js/issues/166. It seems many were experiencing success with requiring the library differently and `truffle-contract` was no different.

### DIFF
--- a/packages/truffle-contract/lib/reformat.js
+++ b/packages/truffle-contract/lib/reformat.js
@@ -1,8 +1,8 @@
 /**
  * Utilities for reformatting web3 outputs
  */
-const BigNumber = require('bignumber.js');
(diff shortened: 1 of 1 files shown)


eggplantzzz:
To be clear, the only "functional" change made was on line 4 of `reformat.js`. The rest of the changes were done by the formatter.

PR #1634, opened 2019-01-17
Update Node version pre-flight check

gnidan:
- Fix to match actual minimum requirement of v8.9.4
- Use semver to check compatibility
- Reörder imports nitpickingly
- Switch to "string"+"math" in case Node version doesn't support backtick syntax
- Reword error, remove line-break.

### DIFF
--- a/packages/truffle-core/cli.js
+++ b/packages/truffle-core/cli.js
@@ -1,23 +1,29 @@
 #!/usr/bin/env node
 require("source-map-support/register");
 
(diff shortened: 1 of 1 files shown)


gnidan:
(Unrequesting review while I look into those tests)

gnidan:
Thanks for cleaning up my mess @eggplantzzz :D

eggplantzzz:
Looks great!

Issue #1636, opened 2019-01-17
"truffle create contract" uses outdated compiler version


sbrichards:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

Hey folks! Fairly new to Truffle. Thanks for this great framework. I'm happy to try to solve this issue w/ a PR but will likely need some guidance.

## Issue

The `truffle create contract` command outputs a .sol file blueprint with a different solidity compiler version vs. what `truffle version` outputs and/or what is specified in truffle-config.js.

## Steps to Reproduce

Init a new truffle project and create a contract:
`$ mkdir truffle-test && cd truffle-test`
`$ truffle init`
`$ truffle create contract MyContract`

Looking at the MyContract.sol code...
```
pragma solidity ^0.4.22;

contract MyContract {
  constructor() public {
  }
}
```

## Expected Behavior

I expected the contract blueprint to use solidity ^0.5.1, given that is the result when running `truffle version`. I tried editing the compiler configuration in truffle-config.js:
```
compilers: {
  solc: {
    version: '0.5.1'
  }
}
```
Bue re-running the `truffle create contract` command yielded the same result.

## Actual Results

The contract blueprint uses solidity ^0.4.22. Contents of MyContract.sol:
```
pragma solidity ^0.4.22;

contract MyContract {
  constructor() public {
  }
}
```

## Environment

* Operating System: macOS Mojave 10.14.2
* Ethereum client: N/A
* Truffle version (`truffle version`):
```
$ truffle version
Truffle v5.0.1 (core: 5.0.1)
Solidity - 0.5.1 (solc-js)
Node v10.13.0
```
* node version (`node --version`):
```
$ node --version
v10.13.0
```
* npm version (`npm --version`): 
```
$ npm --version
6.4.1
```

sbrichards:
Ah, looks like that solidity version is hardcoded in the template:
https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/lib/templates/Example.sol

Thinking there must be a way to extract the version (similar to the `truffle version` command) & input that into the template? I'll dig into this but any guidance would be helpful. Thanks.

eggplantzzz:
Ahh yes @sbrichards, that should be updated. Thanks for bringing this up! Let me have a chat with @gnidan and see what behavior we want to spec out for this.

eggplantzzz:
[...]

PR #1639, opened 2019-01-18
Finish #1636

sbrichards:
#1636 

Update the contract template to the current major version of solidity (^0.5.0).


sbrichards:
Note: only change to packages/truffle-core/test/create.js is the assertion on line 49. Everything else is from the pre-commit linter.

PR #1644, opened 2019-01-19
passing provider the right way

baermathias:


gnidan:
I don't think that's the right way? `provider` is the object you want, `currentProvider` is usually used in the context of `web3.eth.currentProvider`. Is this documented somewhere?

baermathias:
If I am passing `provider` into `setProvider()`, I get this error: 
> TypeError: Cannot read property 'apply' of undefined

Passing `provider.currentProvider` solved it. Other users have also solved it this way: https://github.com/trufflesuite/truffle-contract/issues/57

By the way, I am using a private Quorum network.

PR #1652, opened 2019-01-22
Supply blockNumber when doing the call for getting revert reason.

edgraaff:
Returning the revert reason did not work for me (on Quorum), it turned out it was missing parameter 1 for `eth_call` (which is used to perform the same tx to get the revert reason).

The assumption was made that `latest` would be a proper value for a block number, since new transactions are on the latest block. A rare situation could however occur in which a new block was added between the failed transaction and getting the reason, possibly altering the behavior of the smart contract. The `blockNumber` from the transaction receipt appears as the right value to me, hence it's passed to `reason.js`.

### DIFF
--- a/packages/truffle-contract/lib/handlers.js
+++ b/packages/truffle-contract/lib/handlers.js
@@ -112,7 +112,7 @@ var handlers = {
 
     // .method(): resolve/reject receipt in handler
     if (receipt.status !== undefined && !receipt.status) {
-      var reason = await Reason.get(context.params, receipt.blockNumber, context.contract.web3);
+      var reason = await Reason.get(context.params, context.contract.web3);
 
       var error = new StatusError(
         context.params,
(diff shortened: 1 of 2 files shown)


edgraaff:
Oops, will look into failing tests later today.

edgraaff:
There, fixed. `pending` makes more sense to me than `latest`.

gnidan:
Thanks for this @edgraaff! Apologies for the long radio silence.

I'm going to close this and merge #2033 instead, since I concur with @cgewecke that `latest` probably makes more sense...

My reasoning is that the only truth we know about the chain, or at least the only truth we can rely on, is at the point of the latest block. Sure, we can get the pending information faster, but it's riskier and harder to reason about. I want to avoid the question of chain re-orgs and such, at least for now, and instead just state that `"latest"` is a pretty clear definition of canon for our purposes.

Anyway, thanks again for this! I'm glad your effort investigating and authoring will be reflected in the annals of git commit history. 🙇

edgraaff:
You're welcome, it was only a tiny change :)

PR #1654, opened 2019-01-22
Update a bunch of messages in the codebase

eggplantzzz:
In the interest of moving towards primarily using 'truffle-config.js' instead of 'truffle.js' for the Truffle config file, update some references to it in messages in the codebase.


eggplantzzz:
These are meant to be purely cosmetic changes (not functional) involving comments and messages to the user.

PR #1660, opened 2019-01-24
Unify and rewrite the allocator

haltman-at:
This pull request gets rid of the separate debugger allocator, and also in the process substantially rewrites the allocator.  Most of the changes here are under the hood and won't be visible to users.  Still, a few will be, so let me summarize those first:

1. All types are now handled in allocation, even types we don't support decoding for yet, so having such types among your state variables or struct members won't screw up allrocation.  All state variables and struct members should now be allocated correctly in storage.
2. The new allocator filters out constants; the debugger already did this, but Ganache did not.  So, having constants won't screw up allocations.  (I'm planning to add support for actually handling constants, rather than just filtering them out -- it's not like they worked previously -- in a later PR.)  (Also note that -- though this is a purely under-the-hood change -- this means the debugger no longer handles filtering out of constants, that's purely the allocator now.)
3. The old debugger allocator did not properly handle static arrays.  So, the debugger will now be able to properly decode static arrays in storage.  (At least, in the cases where it could handle that same array if it were dynamic.  Storage arrays still have problems, which I'm intending to fix in my next PR[0].)
4. It seems that my last PR introduced some problems with mappings in structs that I didn't catch.  Sorry about that.  Those are fixed now.

So, what does this PR actually do, under the hood?

Well, as I said, the old debugger allocator is gone.  The old decoder allocator is basically gone, too, because I took the liberty of substantially rewriting it.  There is now one allocator.  The file containing it has been renamed; I took the old `references.ts` and split it into `storage.ts` and `general.ts` (for things not storage-related).  And it works a bit differently, too.  Let's compare.

The core of the debugger's old allocator was a function called `allocateDeclarations`.  It took a bunch of definitions -- whether struct members or state variables -- and returned an allocation for them, along with a marking of where it had left off.  It called `allocateDeclaration` on each individual definition.  For direct types and lookup types[1], it would call `allocateValue`, which would invoke `storageSize` to determine the thing's size.  For structs, it would recursively call `allocateDeclarations`.  For static arrays... well, it couldn't handle static arrays.

Meanwhile, the core of the decoder's old allocator was a function called `allocateDefinition`.  Note the singular -- this function allocated a single definition; there was no real equivalent of `allocateDeclarations`.  Rather there were `allocateStruct` and `getContractStateVariables` (the latter of which I've now renamed `allocateContract`), which would repeatedly call `allocateDefinition`.  And `allocateDefinition` was not a pure function; it returned a value not the ordinary way but rather by mutating one of its arguments.  This was to allow for memoization -- if in the process it needed to allocate a struct, it could do that without much trouble.  But like the debugger's allocator, it still broke things down based on the type of what was being allocated, with detailed logic for different cases.

The new allocator is a bit more uniform in this regard.  Its core is a function called `allocateMembers`, analogous to `allocateDeclarations` in the old debugger allocator.  The functions `allocateStruct` and `allocateContract` still exist -- the debugger didn't need separate functions for these since it had performed processing earlier to make these two cases a bit more similarly, but in the decoder we don't have that, so we do still need separate functions there; but each now calls `allocateMembers` once rather than `allocateDefinition` repeatedly.

In any case, the key new addition in the new allocator is a function called `storageSizeAndAllocate`.  This function is something of a replacement for the old `storageSize`.  What this function does is to return the storage size of a definition, and *also* perform any allocations needed along the way to determine this.  With this in hand, `allocateMembers` can treat types much more uniformly -- rather than looking at detailed type information itself, it simply hands it over to `storageSizeAndAllocate` to determine how much space a type should take up, and responds accordingly.

Note that the old `storageSize` still exists, for use by the decoder once allocation is complete, but it's now just a wrapper around `storageSizeAndAllocate`.  The allocator, of course, does not use this wrapper, because it might actually be calling it in a situation where allocations are yet to be performed, and it's going to need those new allocations that get returned.  Note that `storageSize`, along with `storageSizeAndAllocate`, now lives in `allocate/storage.ts` with the rest of the allocation logic, and not in `truffle-decode-utils`.

However `storageSize` does behave slightly differently than before.  Rather than taking a single `referenceDeclaration` as argument, it takes the set of all reference declarations, because sometimes you don't know in advance which ones you need.  And of course it takes an `allocations` argument, used to look up the sizes of any structs involved.  (Allocations for structs now include a `size` field alongside the `members` field so that their overall size may be easily looked up.  For contracts this is omitted.)

The return type is a little different too.  I said above that `allocateMembers` now treats types uniformly, but that's not *quite* true -- because, after all, some types are sized in bytes, while some types are sized in words, and these cases act differently.  While `allocateMembers` doesn't go checking type information itself, it does check what `storageSizeAndAllocate` returns; and that return value, instead of simply being a number of bytes, will be an object of either the form `{bytes: number}` or `{words: number}`, so that `allocateMembers` (as well as `storageSizeAndAllocate`, and anything else that might care) can easily treat these cases separately.  (A function is provided to convert these objects to simple numbers of bytes for when that's what's needed instead.)

So, really, the core of the allocator is now this pair of mutually-recursive functions -- `allocateMembers` on the one hand, and `storageSizeAndAllocate` on the other.    (Although what `storageSizeAndAllocate` actually calls is `allocateStruct`, but that's basically a wrapper around `allocateMembers`.)  Memoization is handled simply by having all of these taking an `existingAllocations` argument, and always having, as the return value (or in the return value), an object including *all* allocations made so far, including those passed in.  When `allocateMembers` attempts to allocate something, first it checks whether it's already been allocated; if so, it just returns the existing allocations that were passed in.  If not, then it will perform actual allocations.  (Similarly, `storageSizeAndAllocate` will return the same allocations it was passed in unless it's dealing with a struct or static array.)  Because of this memoization, I was able to remove the topological sorting in the debugger that I added in my last PR.

For what it's worth, `allocateMembers` also takes an optional `suppressSize` argument; this is used by `allocateContract` so that, when allocating a contract, it doesn't store a (potentially misleading) `size` field.  (Of course, `storageSizeAndAllocate` wouldn't look at it anyway -- it knows that the storage size of a contract is always 20 bytes -- but I still didn't want misleading information included in the structures.)

Note by the way that the new allocator does not use the `path` field in its storage slots, which is something the old debugger allocator did.  As I've discussed before, struct members (and state variables) are each just a single step down the conceptual storage access tree.  So the slots here do not use paths but rather use raw `BN` offsets.  Again, nothing is getting pre-hashed here; it's all just addition.  But this does mean that the contentious use of `storageSlot` in the decoder can now be removed (which I've done).

In the process of doing the above, I made a number of other changes as well.  (And this PR was originally going to be even bigger, can you believe it!)  For instance:
1. I got rid of the use of `scopes` in the decoder, converting the debugger to use `referenceDeclarations` instead, removing some annoying duplication.
2. I moved the debugger over to using the same allocation format as the decoder, removing the shim I had there previously to convert.
3. I rewrote various types that the decoder used, and the interface surrounding the allocator.  In particular I moved contract definitions inside `referenceDeclarations`, rather than having them be something stored separately.
4. I factored out the logic for detecting the length of a static array into a new function in `definition.ts`.
5. I changed the decoder's contract interface to store the ast nodes for all contracts involved, rather than potentially looking them up repeatedly.
6. I moved the allocation table in the debugger from a selector into the state; in the process, I had to rewrite a bunch of the `ast` sagas.  There's now no longer any `ast` listener or any `ast` actions at all (the file is deleted).
7. I, uh, changed our `eslint` settings to allow function arguments beginning with `_` to go unused.  That's not really related to allocation at all, but it happened along the way.

Let me now make some remarks on what this PR does *not* do, and other bits of potential ugliness:
1. First off, this PR does *not* fix storage decoding and array decoding in general like I had originally intended -- this ended up being quite big enough as it is so I decided to push that off to my next PR[0], as mentioned above.  Storage arrays whose base types are larger than a word still will not decode properly; neither will static arrays in memory; and mappings still have a number of problems, not all of which I'm even planning to get to in my next PR.
2. It turns out the biggest source of annoying duplication in the decoder isn't `scopes`, but rather things being written so that the definition passed in can be either a variable definition or a type definition.  I'm hopiong to handle this in a later PR, but for now that's left alone.
3. The file `allocation.ts`, in `truffle-decode-utils`, though gutted, has not been deleted.  I'm intending in my next PR to delete `normalizeSlot` and move the rest into `allocate/storage.ts`, but for now I haven't done that (moving the rest without deleting `normalizeSlot` would result in some backwards dependencies).
4. I kind of junked up the `AstDefinition` type with a bunch of question marks.  Really that type seems to be doing a bit too much at the moment, I think we need a greater variety of more structured AST types, but I wasn't doing that in this PR.  I guess there is the `Ast` type -- I'll admit I don't really know what that's doing, but I replaced one use of it with `AstDefinition` because it wasn't convertible.  I don't think that should be a problem.
5. We are now doing a double-compute of getting inherited variables for each contract when using the debugger -- the debugger does it once, and then the allocator does it again.  I couldn't find any easy way to avoid this while keeping our current structure.  This doesn't exactly strike me as an expensive computation -- it's just concatenating some arrays -- but I thought it was worth noting.
6. That said, here are two things that might actually contribute to the startup time.  First, as I mentioned, I fixed the problem above where the debugger wasn't properly waiting before starting, so now it might take longer to start.  I haven't noticed any problem, but, again, worth noting (and I haven't tried any huge many-contract projects).
7. Also, the debugger doesn't presently have any good system to determine just what contracts are being used in a given transaction, so it just allocates *everything*.  Again, I haven't noticed any problems, but it could slow down debugger start-up time for large many-project contracts.  I'm hoping this doesn't actually cause too much of a problem.
8. Finally... the decoder is written in this style where having really long lines of code is just fine, and, well, since that's the way it's written, my additions are written that way too.  It still bugs me though -- maybe we should turn on `prettier` for TypeScript?  (And use Nick's history-rewriting trick too, probably?)

Anyway, likely I missed one or two things, but that's what this PR is doing (and not doing).  Next time: Fixing arrays, and somewhat fixing mappings!

[0]Note: When I speak of "my next PR", I mean my next PR related to the decoder.  I'm probably going to handle some minor other stuff first before I get back to this.
[1]I'm using here the terminology I set forth in [this paper](https://ethdebug.github.io/solidity-data-representation/).


haltman-at:
So, a note: In order to get this merge to work out, I had to delete a number of unused variables in the unrelated `truffle-compile` in order to satisfy `eslint`.  I am hoping this is not screwing anything up.

PR #1665, opened 2019-01-25
Fix misnamed property in config.js

remilb:
Attempting `truffle help config` causes truffle to crash due to a misnamed property in config.js. PR is simply a small fix for the typo.  :)


eggplantzzz:
Oh whoops, thanks for catching this!

Issue #1640, opened 2019-01-18
Truffle testing with Solidity ^0.5; DeployedAddresses not supporting payable addresses 


ryanwdavies:
## Issue
In Solidity 0.5 addresses are of the type "payable" and "non-payable".

Truffle's DeployedAddresses appears to only support non-payable addresses, which does not allow it to be used with contracts which have a payable fallback function.

## Steps to Reproduce

example contract:
````javascript
pragma solidity 0.5.2;

contract ThamesBoats {
  address owner;

  constructor() public {
      owner = msg.sender;
  }

  /// @notice fallback funtion, payable
  function() external payable { }
}
````

Test.sol
````javascript
pragma solidity 0.5.2;

contract TestThamesBoats {

  ThamesBoats public thamesBoats;
  thamesBoats = ThamesBoats(DeployedAddresses.ThamesBoats());

  function () external payable {}
}
````

Running the test script yields:
````javascript
truffle test

TypeError: Explicit type conversion not allowed from non-payable "address" to "contract ThamesBoats", which has a payable fallback function.
      thamesBoats = ThamesBoats(DeployedAddresses .ThamesBoats());
                    ^------------------------------------------^
````

## Expected Behavior

If I repeat the same except use "pragma solidity 0.4.25", all is well and the contract address is returned by DeployedAddresses in the normal way.

## Actual Results

Error above.

## Environment

* Operating System: Ubuntu 18.04.1 LTS
* Ethereum client: 
* Truffle version (`truffle version`):
Truffle v5.0.0 (core: 5.0.0)
Solidity - 0.5.2 (solc-js)
Node v8.12.0

* node version (`node --version`): v8.12.0
* npm version (`npm --version`): 6.4.1

eggplantzzz:
I think we should have a fix for this with the release of `0.5.3`. We plan on releasing today (January 30).

eggplantzzz:
truffle@5.0.3 was just released and should solve this issue! Closing this but let me know if you still experience problems.

PR #1670, opened 2019-01-28
Make DeployedAddresses payable

gnidan:
As a stop-gap solution to #1640, make all addresses in DeployedAddresses payable.


gnidan:
(We will probably want to split these up based on the fallback function, but in the meantime this should be fine because whatever consuming contract uses the address will specify either payable or not, and payable is implicitly convertible to not)

Issue #1667, opened 2019-01-28
Error: Missing parameter: provider


princesinha19:
```
node_modules/web3-core/dist/web3-core.cjs.js:148
      throw new Error("Missing parameter: ".concat(name));

Error: Missing parameter: provider
```

I am getting this error when using web3 version "web3": "^1.0.0-beta.39"
truffle version: ^4.0.2

Can anyone help me with this error?

filips123:
Package Web3 have many changes in last version (see ethereum/web3.js#2000). It looks like that Truffle currently doesn't support them.

This should really be fixed, because it is currently impossible to use `truffle-contract` with three latest `web3` versions (previous versions also caused errors in #1617 and #1614).

Temporary solution may be do downgrade Web3, possibly to `1.0.0-beta.33`.

princesinha19:
Thanks, @filips123 for your quick response. 
So, you guys are downgrading web3 in your package?
How much time it will take to do this?

princesinha19:
`Temporary solution may be do downgrade Web3, possibly to 1.0.0-beta.33.
`

How to do this inside truffle-contract @filips123?

princesinha19:
@filips123 I have asked the same question on web3 but they are saying that you guys are using unofficial API.
![web3 issue](https://user-images.githubusercontent.com/28489437/51824638-d1ea1400-2308-11e9-8513-83e8ce3398e3.png)

daniyarchambylov:
I'm experiencing this issue too and downgraded Web3 package, but it's blocking from publishing a package. No end user should do workarounds.

rocky:
For what it is worth:

```
        "web3": "^1.0.0-beta.37",
        "web3-core-promievent": "^1.0.0-beta.37",
```

work for me

Levino:
This is the most urgent bug of all here I would say. I just created a new truffle project. Wont run.

gnidan:
We are looking into this issue as a matter of top priority.

Can someone provide reproduce steps so we can validate potential fixes?

Heads up, we are planning to release today without web3 fixes, but we will look into getting another release shortly after that.

alcuadrado:
A smaller reproduction script is:

* `npm init -y`
* `npm install truffle-contract`
* `node --eval 'require("truffle-contract")'`

It results in this error:

```sh
/private/tmp/repro/node_modules/web3-core/dist/web3-core.cjs.js:148
      throw new Error("Missing parameter: ".concat(name));
      ^

[...]

Issue #1617, opened 2019-01-05
Cannot find module 'xmlhttprequest'


filips123:
## Issue

I have code that uses `truffle-contract` for interacting with smart contract. It should work in Node.js and browser so I use [AEgir](https://www.npmjs.com/package/aegir) to build it for browser (with Webpack).

It works normally in Node.js but there are problems in the browser.

When I include my built JS file, I get error `Cannot find module 'xmlhttprequest'` and none of my code works.

It looks like request for module `xmlhttprequest` come from package `ethers`, more specifically [`ethers/utils/web`](https://github.com/ethers-io/ethers.js/blob/master/utils/web.js#L10), and Webpack didn't resolve module into browser's code.

I think that worked in previous versions of `truffle-contract`.

Please let me know how to fix this or if this is issue with AEgir, Webpack or Ethers.

## Screenshots

![error](https://user-images.githubusercontent.com/16626308/50728588-a1112780-112c-11e9-8ea0-ac727300174e.png)

## Environment

* Browser: Firefox and Chrome
* Truffle Contract: 4.0.1
* AEgir: 18.0.2

filips123:
This is now fixed in `web3@1.0.0-beta.38`. This package should now also be updated.

filips123:
Fixed in #1646. See https://github.com/trufflesuite/truffle/issues/1614#issuecomment-459075959 for details.

Issue #1614, opened 2019-01-04
truffle-contract@4.0.1 - React app - "ethers" -  Module not found: Can't resolve 'xmlhttprequest'


gjeanmart:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Importing `truffle-contract@4.0.1` in a React app breaks the web-app compilation

```
Failed to compile.

./node_modules/ethers/utils/web.js
Module not found: Can't resolve 'xmlhttprequest' in '/home/gjeanmart/workspace/sandbox/myapp-react-truffle-contract-issue/node_modules/ethers/utils'
```

## Steps to Reproduce
1. Create a new React app calle `myapp`
```
$ npx create-react-app myapp
```

2. Go to the project folder 
```
$ cd myapp
```

3. Install` truffle-contract` version `4.0.1` as a dependency
```
$ npm install truffle-contract@4.0.1 --save
```

4. Edit `src/App.js` to import `truffle-contract`
```
$ vi src/App.js
import React, { Component } from 'react';
import logo from './logo.svg';
import './App.css';
import TruffleContract from 'truffle-contract'; // <<<<<<<<<<<<<<<<<<<< IMPORT BREAKING THE BUILD

class App extends Component {
  render() {
    return (
      <div className="App">
        <header className="App-header">
          <img src={logo} className="App-logo" alt="logo" />
          <p>
            Edit <code>src/App.js</code> and save to reload.
          </p>
          <a
            className="App-link"
            href="https://reactjs.org"
            target="_blank"
            rel="noopener noreferrer"
          >
            Learn React
          </a>
        </header>
      </div>
    );
  }
}

export default App;
```

5. Start the React App
```
$ npm start
```

## Expected Behavior

Should start the app nicely on http://localhost:3000/

```
Compiled successfully!

You can now view react-truffle-contract-4.0.1 in the browser.

  Local:            http://localhost:3000/
  On Your Network:  http://192.168.0.59:3000/

Note that the development build is not optimized.
To create a production build, use npm run build.


```

## Actual Results

The compilation fails and throws the following exception:
```
Failed to compile.

./node_modules/ethers/utils/web.js
Module not found: Can't resolve 'xmlhttprequest' in '/home/gjeanmart/workspace/sandbox/myapp-react-truffle-contract-issue/node_modules/ethers/utils'
```

## Environment

* Operating System: Ubuntu 18
[...]

PR #1675, opened 2019-01-29
Updated package.json of truffle-contract

princesinha19:
Downgraded **web3, web3-eth-abi, web3-core-promievent, web3-utils** to version **^1.0.0-beta.33** from ^1.0.0-beta.37.
Due to issues like #1667 with current truffle-contract which works on ^1.0.0-beta.37.

### DIFF
--- a/packages/truffle-contract/package.json
+++ b/packages/truffle-contract/package.json
@@ -33,10 +33,10 @@
     "truffle-blockchain-utils": "^0.0.7",
     "truffle-contract-schema": "^3.0.1",
     "truffle-error": "^0.0.3",
-    "web3": "^1.0.0-beta.37",
-    "web3-core-promievent": "^1.0.0-beta.37",
-    "web3-eth-abi": "^1.0.0-beta.37",
-    "web3-utils": "1.0.0-beta.37"
+    "web3": "^1.0.0-beta.33",
+    "web3-core-promievent": "^1.0.0-beta.33",
+    "web3-eth-abi": "^1.0.0-beta.33",
+    "web3-utils": "1.0.0-beta.33"
   },
   "devDependencies": {
     "async": "2.6.1",


gnidan:
Hi, thanks for taking the time to open this.

Unfortunately, we won't be able to accept this downgrade, as the newer versions of web3 provide much-needed functionality, i.e., support for structured ABI parameters. A lot of users are already making heavy use of this, and we can't just take that away.

Hopefully there is another solution to the above issue.

gnidan:
Yep, we probably should pin to an exact version. At the moment, it'll have to be beta.37, until we can look into the more recent beta (to make sure it works with Truffle)

gnidan:
@freeatnet we're looking into it as a matter of top priority. We're doing a patch release today with outstanding merged changes, then we will look at upgrading web3 to the latest beta to see if that fixes the problems here.

In the meantime, can you or someone else provide clear reproduction steps so we can validate the problem locally?

princesinha19:
Steps to reproduction:
1. Install truffle-box using npm
     npm install truffle-box --save
2. Compile and migrate your contracts using truffle migrate command.
3. Import truffle-box inside your project (In nodejs app.js)

```
const contract = require("truffle-contract");
var contract_json_data = require('../build/contracts/ContractName');
var MyContract;
MyContract = contract(contract_json_data);
MyContract.setProvider(web3.currentProvider);
```

4. Use contract method inside your app. Then, you will get this error.
eg..
```
MyContract.deployed().then(function(instance){
    ContractInstance = instance;
     instance.function();
});

```

gnidan:
Thanks @princesinha19! I just opened #1685 to upgrade Web3.js to latest, also pinning the version specifically. We'll see how that goes. If it passes our tests, we can proceed with the upgrade, otherwise I think we can try pinning to `beta.37` for our release today and seeing if that solves the problem.

filips123:
@gnidan Web3 `beta.37` has problems with Webpack and React in #1617 and #1614, so maybe you will need to downgrade Web3 even more.

gnidan:
@filips123 Downgrading web3 is out of the question, because versions older than beta.37 do not support structured function parameters / the new ABI encoder. I am hoping that beta.38, which includes significant changes to their webpack config, will address these issues.

filips123:
@gnidan If you keep `beta.37`, you should also manually require the latest version of `ethers`. See https://github.com/trufflesuite/truffle/issues/1614#issuecomment-452158490 for details.

gnidan:
@filips123 Ah, yep. So #1646 was merged and is getting released today. I believe that fixes the `xmlhttprequest` problem, so that should be fine. We'll see about the rest of this.

Fun day :)

gnidan:
I'm closing this for now. If we don't go forward with upgrading web3 immediately, @princesinha19 I will cherry-pick your commits for your contributor credit.

Let's continue the conversation in #1667 so it's all in one place. Thanks again everyone. We'll get this sorted out today.

princesinha19:
Okay no problem @gnidan. Let's hope after upgrade the issue will get resolved.

PR #1677, opened 2019-01-29
Fix various decoding errors with reference types

haltman-at:
This PR fixes a number of decoding errors with reference types.  Specifically:

1. Storage arrays whose base type is longer than a word now work properly.
2. Mappings whose value type is longer than a word now work properly (if they didn't already?).
3. Mappings whose value type is a signed integer shorter than a full word now work properly.
4. Statically-sized arrays in memory now work properly.
5. Struct types with mappings in them now work properly in memory.  Actually, now structs in memory work at all!  Turns out there was a bug causing errors when you attempted to decode any struct in memory.
6. Decoding of storage local variables is now supported.

Also, the file `truffle-decode-utils/allocation.ts` has now been deleted; its contents have been moved into `truffle-decoder/allocate/storage.ts`, except for `normalizeSlot`, which is gone.  (Turns out it was already unused, and I could have deleted it in my last PR like I wanted to, but I didn't notice that then.)

Note that in order to add support for local variables, I ended up rewriting `decode/index.ts` to change how the dispatching works there, and added a new file `decode/stack.ts`, along with some new entry points to storage and memory decoding (`decodeStorage`, `decodeMemory`, `decodeStorageByAddress`).

I also added a test for these fixes; it doesn't fit in well with the other decoding tests, but I figured it was better than not including it.  Also incidentally I turned down the timeout on the variable ID tests from 30 seconds (!) to just 4 seconds.  Hopefully that's not too short and this PR passes Travis without me having to turn that back up!


packages/truffle-decoder/lib/allocate/storage.ts
```diff
@@ -23,6 +23,24 @@ export function storageLengthToBytes(size: StorageLength): number {
...
+  length?: number;
+}
+
+export interface StoragePosition {
+  slot: Slot;
+  index: number;
+};
+
+export interface Slot {
+  key?: any; // TODO:
```
gnidan: what's this TODO?

packages/truffle-decoder/lib/decode/storage.ts
```diff
@@ -5,16 +5,55 @@ import read from "../read";
...
-import { Allocation } from "truffle-decode-utils";
-import { storageSize, storageLengthToBytes } from "../allocate/storage";
+import * as Allocation from "../allocate/storage";
 import BN from "bn.js";
 import Web3 from "web3";
 import { EvmStruct, EvmMapping } from "../interface/contract-decoder";
 import clonedeep from "lodash.clonedeep";
 
-export default async function decodeStorageReference(definition: DecodeUtils.AstDefinition, pointer: StoragePointer, info: EvmInfo, web3?: Web3, contractAddress?: string): Promise<any> {
+export default async function decodeStorage(definition: DecodeUtils.AstDefinition, pointer: StoragePointer, info: EvmInfo, web3?: Web3, contractAddress?: string): Promise <any> {
```
gnidan: Gosh this line is long

packages/truffle-decoder/lib/decode/storage.ts
```diff
@@ -39,79 +75,94 @@ export default async function decodeStorageReference(definition: DecodeUtils.Ast
       }
       debug("length %o", length);
 
-      const baseDefinition = DecodeUtils.Definition.baseDefinition(definition);
+      const baseDefinition = definition.baseType || definition.typeName.baseType;
```
gnidan: What happened here? Shouldn't this use the function you just fixed?

packages/truffle-decoder/lib/decode/storage.ts
```diff
@@ -269,7 +351,7 @@ export default async function decodeStorageReference(definition: DecodeUtils.Ast
           }
 
           result.members[memberName] =
-            await decode(definition.typeName.valueType, valuePointer, info, web3, contractAddress);
+            await decode(valueDefinition, valuePointer, info, web3, contractAddress);
         }
       }
```
gnidan: gosh, changes to this file are sure getting difficult to read about. it all seems reasonable though. we should think about getting this more organized so it doesn't require as large a brain-cache size to look at PRs.

packages/truffle-decoder/lib/allocate/storage.ts
haltman-at: That's preexisting -- I assume @seesemichaelj put that there.  I assume it's about the fact that `key` is of type `any`, which should be narrowed down to something more specific.  You're seeing it anew due to it having moved to a different file.

packages/truffle-decoder/lib/decode/storage.ts
haltman-at: Then turn on prettier :P

packages/truffle-decoder/lib/decode/storage.ts
haltman-at: I didn't fix it -- I removed it.  I didn't see a need for a separate function.  Later when I go through and make it so that we're always assuming `definition` is a type definition and not a variable definition, I'll remove all the duplication.

packages/truffle-decoder/lib/allocate/storage.ts
haltman-at: Or maybe I'll wait until the new decoder output format, actually.  In any case the point is I'll handle it eventually.

haltman-at:
Let me add a comment on this most recent commit, because it changes the structure somewhat.  Literal pointers are now no longer passed to `decodeValue`, since they might not be values (the old code got this right, silly me), but rather to `decodeLiteral`.  Since literal pointers come from the stack (unless they're string literals), `decodeLiteral` resides in `stack.ts`, with `decodeStack` now calling it.  String literals will be dispatched by `decodeLiteral` to `decodeValue`, but this is kind of a hack, and I'm intending to change this in a later PR.

Issue #1664, opened 2019-01-24
Improvement in "Could not connect" error


wbt:
## TL;DR
This proposes making an error message more helpful with additional context information.

## Issue
I get the "Could not connect to your Ethereum client" error frequently, especially in load testing, regardless of Ethereum client being used.  I assume that what's happening (also in [issue 1461](https://github.com/trufflesuite/truffle/issues/1461)) is that the client is just getting overloaded with too many pending requests, and stops responding to new ones.  

If the issue were **not** an overloaded server that had just successfully responded to several transactions, and really was a connectivity issue, it would be helpful to know what configuration information was being read and used. 

## Steps to Reproduce
Set up a basic Truffle project to work with Ganache, don't start Ganache, and run `truffle migrate --reset`.

## Actual Results
Error message: 
>Could not connect to your Ethereum client. Please check that your Ethereum client:
    - is running
    - is accepting RPC connections (i.e., "--rpc" option is used in geth)
    - is accessible over the network
    - is properly configured in your Truffle configuration file (truffle.js)

## Expected Results
The same error, but with more specific information about where it was trying to connect to.  
>Could not connect to your Ethereum client at http://127.0.0.1:7545 (Network ID: 5777).
Please check that your Ethereum client:
    - is running
    - is accepting RPC connections (i.e., "--rpc" option is used in geth)
    - is accessible over the network
    - is properly configured in your Truffle configuration file (truffle.js), especially if the server information in the first line of this error does not match what you expect.

I would only expect "Network ID" to be there if it had been specified. 

## Environment
* Operating System: Windows 10 pro
* Ethereum client: For this Issue, none
* Truffle version (`truffle version`): 4.1.15
* node version (`node --version`): 9.3.0
* npm version (`npm --version`):  6.5.0

wbt:
Per your comment [here](https://github.com/truffle-box/metacoin-box/pull/7#issuecomment-452820812), that last line should probably also refer to truffle-config.js instead of truffle.js.

eggplantzzz:
Yep, I changed that line about 2 days ago! It should be out in the next patch release.

eggplantzzz:
@wbt Would you like to open a PR for this? Let me know if you want to tackle it, otherwise I'll take it on next week.

wbt:
Go for it, and thanks!

gnidan:
This was merged! And released! Closing. Thanks!

bater53:
[...]

PR #1678, opened 2019-01-29
Provide more feedback with connection error

eggplantzzz:
Pass the options to ProviderError to provide more information to the user when encountering a connection error


gnidan:
This fixed #1664

PR #1693, opened 2019-02-03
Update projects' package.json repository fields

zachlysobey:
I noticed that many of the repository fields for sub-packages here were pointed at the now defunct original repositories (pre-monorepo). I actually had some initial confusion when I was looking into a `truffle-foo` dependency I'm using, and was directed to a now-defunct github repo with old code.

They now point at the individual packages here in a standard way.

**For example:**

```diff
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/trufflesuite/truffle-compile.git"
-  },
+  "repository": "https://github.com/trufflesuite/truffle/tree/master/packages/truffle-compile",
```

This field can apparently be configured in several different ways, and given this mono-repo setup, its not totally clear to me what the *best* setup is, but I just copied [what *babel* uses](https://github.com/babel/babel/blob/master/packages/babel-parser/package.json#L19) because that seems reasonable.

---

#### Notes:

In my opinion, the [`package.json` repository field](https://docs.npmjs.com/files/package.json#repository) is super useful.

- it populates the *repository* link for [npmjs.com](https://www.npmjs.com) package pages
- from the terminal you can type `npm docs truffle-core` or `npm repo truffle-core` and be directed to the right GH page in your browser
- if you're in the directory of a package here, you can just type `npm docs` or `npm repo` and it will do the same

### DIFF
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -8,10 +8,7 @@
   "scripts": {
     "test": "mocha"
   },
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/trufflesuite/truffle-artifactor.git"
-  },
+  "repository": "https://github.com/trufflesuite/truffle/tree/master/packages/truffle-artifactor",
   "license": "MIT",
   "devDependencies": {
     "chai": "4.2.0",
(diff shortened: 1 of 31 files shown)


eggplantzzz:
Thanks for submitting this! We will review it.

PR #1701, opened 2019-02-06
Improved decoding for value types and for mappings, and partial support for constant state variables

haltman-at:
This PR does three main things.  (It also contains assorted cleanup but I won't go into that here.)

Firstly, it improves decoding for value types (or more generally, direct types).  Specifically:

1. Variables of type `bytesN` will now decode to hex strings that are just N bytes long, regardless of their location, whereas previously they were, except in storage, always 32 bytes long regardless of N.
2. Decoded addresses will now be in checksum case rather than lowercase, and will always decode to hex strings that are 20 bytes long, whereas previously they could be shorter if they began with leading 0s (any leading 0s would be trimmed).

In addition, rudimentary support for decoding contracts has been added; a contract will just decode to its address for now.

Secondly, it adds partial support for decoding of constant state variables.  Previously these were supported in allocation (in that they were skipped), but there was no decoding of them.  Now it is possible to decode *some* constant state variables -- basically, ones defined by string literals (including `hex"..."` literals) and ones defined by simple arithmetic expressions (this includes ones of type `bytesN` defined by hex literals, though not addresses defined by hex literals).  Constant state variables that cannot be decoded remain excluded from the variables list.  Note that in order to accomplish this, a new type of `DataPointer` has been added, called `ConstantDefinitionPointer`.  Rather than pointing to any EVM location, it holds the AST node giving the constant's definition.

Why only partial support?  Well, since a constant state variable is not actually stored anywhere during execution, its value must, as mentioned above, be determined from its AST node.  Since it is a constant, this is necessarily possible!  However, it is not so easy, because the AST does not always directly encode the value; rather, in general, it encodes the AST for the definition.  So, decoding arbitrary constants would require writing a function to evaluate Solidity expression ASTs.  Obviously this is doable, but not really the best use of my time right now -- in particular it makes no sense to do this before redoing the decoder output format, since it would be very helpful to have a more structured way to represent Solidity values before attempting such a thing -- so instead I have only added support for some particularly simple cases, which also happen to be cases we need (see below).

Finally, it improves decoding of mappings and, especially, their keys.  This improvement consists of a number of parts and is worth discussing in more detail.  Note that this PR does *not* yet rework how mapping keys are stored, which I intend to do in an upcoming PR; that remains the same.  However, substantial parts of how mapping keys are handled has been redone.

Firstly, the use of `soliditySha3` has been corrected; previously this was handled incorrectly for a number of types.  There are two ways it was handled incorrectly.  Firstly, `soliditySha3` provides the hash of the packed representation; however, when the mapping key is of value type (i.e. not `bytes` or `string`), Solidity uses the padded representation to locate the corresponding storate slot.  Therefore, for certain types, padding needed to be added.  Secondly, when passed a string, `soliditySha3` attempts to autodetect whether that string might represent a hex string or a decimal number, and if so treats it as such rather than as a UTF-8 string; this meant that mapping keys such as `"0xdeadbeef"` or `"12345"` would fail to hash correctly.

Fortunately, one may avoid this second problem by passing instead to `soliditySha3` an argument of the form `{type: ..., value: ...}`; by explicitly specifying `type` as `"string"`, we may safely pass strings to `soliditySha3`.  This same solution is used to achieve the needed padding as well -- we now always pass such objects to `soliditySha3`, rather than raw values, but sometimes we lie about the type in order to achieve padding.  Types requiring right zero-padding get labeled `bytes32`; those requiring left zero-padding get labeled `uint`; and those requiring left sign-padding get labled `int`.  To make this easier, `Slot` objects with keys now contain a new field called `keyEncoding` that lists which `type` to pass on to `soliditySha3`.  (Note that the `keyEncoding` field is a temporary measure, which will be rendered unnecessary by the new decoder output format.)

The one exception to the above is `bool`s, because `soliditySha3` simply refuses to pad these correctly no matter what you pass in.  So, for these we leave `keyEncoding` undefined and handle the problem a bit more manually.  It's a hack, but it works.

But this is just one of the problems with mappings that has been addressed!  Let's move on to the next: Keys were often decoded as the wrong type.

Previously, when decoding a mapping key, the definition used for decoding was that of the index expression, rather than the mapping's key definition.  However due to implicit type conversions, these might not always agree; this could result in mapping keys being decoded as the wrong type (e.g. negative mapping keys decoding as large positive mapping keys, or `bytesN` mapping keys decoding as numbers, or bytestring mapping keys being interpreted as UTF-8).  Oh, and attempting to use a `bool` as a mapping key would cause an error (not sure if this is why, but it would).  Mapping keys are now always decoded using the mapping's key definition rather than the index definition.

The next improvement to mapping keys relates to mapping keys which are not sourcemapped to.  The existing code for mapping keys already covered one such case, namely, string literals.  However, it turns out that there are more.  For instance, the mapping key could be a constant state variable.  This case, of course, we cannot completely handle, as we do not yet support decoding arbitrary constant state variables.  So, users beware; if you use constant state variables as mapping keys, it is not currently guaranteed that we can keep track of them.  That said, if it's one of the types of constant state variables we *can* currently decode, we will now do so.

But there's one more way that a mapping key can fail to appear in the image of the sourcemap.  (Or at least, one more that I know of!  I'm pretty sure that this is all, but it's hard to say.  I will say that I was a little surprised to find that keys of the form `c.length`, for `c` a constant state variable of type `bytes`, did not require any additional support on my part, but rather worked already.)  And this is when the key is what I'll call a silent (explicit) type conversion.

Type conversions are actually worth discussing in more detail -- I briefly mentioned implicit type conversions above, but didn't say much more than that.  So, let's dissect this.

Type conversions come in four types.  Firstly, a type conversion can be implicit or explicit.  Secondly, a type conversion can be "silent" or not.  We'll say a type conversion is silent when it does not alter the underlying representation, e.g., a conversion from `uint8` to `uint`.  An example of a non-silent conversion would be a conversion from `uint8` to `byte`, since the two types have different padding.  So, then, going over each type of conversion:

1. An explicit, non-silent conversion poses no problem.  Because the conversion is not silent, the type conversion corresponds to an actual machine instruction, i.e. will appear in the image of the sourcemap, and so the existing code will correctly record its stack value.

2. An implicit, silent conversion also poses no problem.  Now the thing about implicit conversions is that there's no AST node corresponding to the conversion, only to the value being converted.  So our index AST node that we look at and fetch the stack value for isn't the conversion, but rather the value to be converted.  But because it's a silent conversion, this is fine -- we're looking at the same stack value either way!

3. An implicit, non-silent conversion, however, poses a problem.  Because now, when we fetch the stack value corresponding to the index expression, we get the stack value for the value to be converted, rather than the converted value.  And since it's not a silent conversion, that's the wrong stack value!  Fortunately, I'm pretty sure there's only one way to have an implicit but not silent conversion; namely, the implicit conversion of a hex constant to type `bytesN`.  Fortunately, there's a simple way to handle this.  Remember how the existing code handles string literals?  We just need to expand that logic to include not just type `stringliteral` but also type `rational` -- which, hey, are also the two types of constants I added logic for -- and then move that check up so that it comes *before* the check for a stack value.  This way we can read the correct value right out of the AST definition, rather than plucking the wrong value off the stack.

4. This leaves the last case, though: The explicit, silent conversion.  Because this is an explicit conversion, there *is* an AST node for the conversion.  But because the conversion is silent, it doesn't correspond to any actual machine instruction, and so is not sourcemapped to.  So, this case needs another way of handling it.  So, after attempting all other ways of decoding keys, if none have worked, we check if the index expression is an (explicit) type conversion.  If it is, we fetch its operand, and retry with that expression in place of the index expression, repeating as necessary until we have either decoded the key or concluded that we cannot.  (Which is what would happen if, say, you wrapped a constant state variable we presently can't handle inside a bunch of explicit-but-silent type conversions.)

Note, by the way, that this same issue could notionally occur for any operation that acts as the identity on the underlying representation.  Fortunately, I don't think there are any others!  For instance, unary `+` would presumably be a problem if it existed, but Solidity thankfully doesn't have any such thing.

The resulting code for decoding keys is, I'll admit, fairly ugly.  I was hoping to factor it out into its own function, but when I tried this it stopped working, presumably due to something about the interaction between generators and `async`.  It'll become even more complicated when I redo how mapping keys are stored, though, so it'll basically *have* to be factored out then, and I'll just have to figure out how to make that work!

But, in short, we should now be able to handle any mapping key except for certain constant state variables (possibly wrapped in silent type conversions).  Whew!  I think that covers all the major changes, so I'll leave it at that.  There's still more that needs to be done to improve mappings -- there's a reason I intend to redo how mapping keys are stored -- but I'd say the most glaring problems are now handled.


packages/truffle-debugger/lib/helpers/index.js
```diff
@@ -23,7 +23,7 @@ export function keccak256(...args) {
  * stringify operation before hashing
  */
 export function stableKeccak256(obj) {
-  return keccak256(stringify(obj));
+  return keccak256({ type: "string", value: stringify(obj) });
```
gnidan: huh neat

haltman-at:
Thanks!  While looking through this though I actually noticed some unused or commented-out code I forgot to delete, and one series of checks I think might be in the wrong order, so I'm going to go fix that before merging this.  (I also noticed one case that I think I missed regarding mapping keys, but I'll fix that in my more general mapping keys PR, not attempt to redo this to include it.)

haltman-at:
OK, I did make one substantive change after approval: I removed the special case for assignment expressions, since they're perfectly good expressions and could be used as mapping keys.

However, there is still a problem that `IndexAccess` expressions won't necessarily work as mapping keys.  Oops.  Missed that above.  However, I'll fix that in my next PR, not this one.

PR #1706, opened 2019-02-08
Fix stack allocation

haltman-at:
The primary purpose of this PR is to fix stack allocation for input parameters to functions.  The existing allocation algorithm assumed all stack variables take up a single word; however, some -- specifically, external functions and some calldata pointers -- take up two.  Since we process input parameters from right to left, this could cause decoding errors for any input parameter which had a two-word input parameter somewhere to the right of it, as its stack allocation would be incorrect.

To accomplish this, the `StackPointer` type has been redefined; instead of containing just a single number, it now contains `from` and `to` numbers.  A new file, `read/stack.ts`, with a function `readStack`, has been added to the decoder, and the `read` function now calls this instead of reading from the stack manually.  The `readStack` function is now also used in the debugger itself for the stack read when assigning literal pointers to nodes, in anticipation of future improvements where we will need this for calldata variables used as mapping keys.  (Other stack reads have been left as raw stack reads; I didn't see any need to change them.)

Note that all stack pointers have been adjusted to be appropriately sized now, but this only currently matters in the parameter case mentioned above.  The only two-word stack variables are, as mentioned above, external functions and certain calldata pointers, and there is still no decoding support for calldata variables or functions of any kind.  This PR covers allocation and reading only, not decoding.

Additional notes:

1. This PR does contain a change which is redundant with one of the changes in the `value-types` PR.  It's a very small change, so it's not redoing substantial work, don't worry.
2. I did various cleanup of the data sagas and actions -- for instance, I found that the `ASSIGN` action had a component which was unused, so I removed it.  I also found that the data saga was doing a lot of unnecessary operations with JSON pointers when it could simply use the AST nodes instead, so I changed it to do that.
3. It turns out that the existing stack allocation code had an interesting error -- it put output parameters below input parameters rather than above them.  It worked anyway, because in fact the stack position in the `FunctionDefinition` case is appropriate for input, not output, parameters; and because output parameters, unlike input parameters, have their declarations sourcemapped, and so ended up overwritten by the `VariableDeclaration` case, which got them right.  This PR alters the `FunctionDefinition` case to only cover input parameters, and leaves the output parameters to the `VariableDeclaration` case.
4. You may be wondering why I did `StackPointer` as `from` and `to`, instead of a start and a length.  The reason is that we typically process the stack from the top down, so `start` and `length` would have been less natural than `top` and `length`, and, well, end-and-length just seems awkward.  (And maybe we'll want to go bottom-up someday.)  So I went with from-and-to.

### DIFF
--- a/packages/truffle-debugger/lib/data/actions/index.js
+++ b/packages/truffle-debugger/lib/data/actions/index.js
@@ -18,10 +18,9 @@ export function declare(node) {
 }
 
 export const ASSIGN = "ASSIGN";
-export function assign(context, assignments) {
+export function assign(assignments) {
   return {
     type: ASSIGN,
-    context,
     assignments
   };
 }
--- a/packages/truffle-debugger/lib/data/sagas/index.js
+++ b/packages/truffle-debugger/lib/data/sagas/index.js
@@ -2,7 +2,6 @@ import debugModule from "debug";
 const debug = debugModule("debugger:data:sagas"); // eslint-disable-line no-unused-vars
 
 import { put, takeEvery, select, call, putResolve } from "redux-saga/effects";
-import jsonpointer from "json-pointer";
 
 import { prefixName, stableKeccak256 } from "lib/helpers";
 
@@ -11,9 +10,9 @@ import * as actions from "../actions";
 
 import data from "../selectors";
 
-import * as TruffleDecodeUtils from "truffle-decode-utils";
+import * as DecodeUtils from "truffle-decode-utils";
 
-import { getStorageAllocations } from "truffle-decoder";
+import { getStorageAllocations, readStack } from "truffle-decoder";
 
 export function* scope(nodeId, pointer, parentId, sourceId) {
   yield putResolve(actions.scope(nodeId, pointer, parentId, sourceId));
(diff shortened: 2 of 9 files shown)


packages/truffle-debugger/lib/data/actions/index.js
```diff
@@ -18,10 +18,9 @@ export function declare(node) {
 }
 
 export const ASSIGN = "ASSIGN";
-export function assign(context, assignments) {
```
gnidan: `context` wasn't being used anymore, was it?

packages/truffle-debugger/lib/data/sagas/index.js
```diff
@@ -62,35 +62,37 @@ function* tickSaga() {
 
   switch (node.nodeType) {
     case "FunctionDefinition":
-      parameters = node.parameters.parameters.map(
-        (p, i) => `${pointer}/parameters/parameters/${i}`
-      );
+      let parameters = node.parameters.parameters;
+      //note that we do *not* include return parameters, since those are
+      //handled by the VariableDeclaration case (no, I don't know why it
+      //works out that way)
```
gnidan: huh that's new?

gnidan: or at least news

packages/truffle-debugger/lib/data/actions/index.js
haltman-at: Yup.  Not sure it ever was while I was here...

packages/truffle-debugger/lib/data/sagas/index.js
haltman-at: Well, it can't be new -- if it were, the old code, which put them below rather than above the input parameters -- wouldn't have ever worked...

PR #1709, opened 2019-02-11
Update stale bot

fainashalts:
Increases daysUntilStale to 120 and daysUntilClose to 14. Will not mark issues with an assignee as stale.


.github/stale.yml
```diff
@@ -17,6 +17,9 @@ exemptLabels:
   - optimization
   - "needs roadmapped"
```
eggplantzzz: Why does this one have ""s and the other one doesn't?

fainashalts: These are based on what our labels are in Github, so not sure why one has an s and one doesn't!

PR #1712, opened 2019-02-12
Add check for unboxing in a non-empty directory [un-tested]

adrianmcli:
Pretty self-explanatory, it just warns the user that the directory is non-empty. We probably want to do this before the download step though.

Either way, I haven't had a change to test this, I just quickly wrote it before I have to board my flight. Sorry! I just wanted to get the ball rolling.


packages/truffle-box/lib/utils/unbox.js
```diff
@@ -114,6 +114,23 @@ async function promptOverwrites(contentCollisions, logger = console) {
 
 async function copyTempIntoDestination(tmpDir, destination, options) {
   fs.ensureDirSync(destination);
+  if (!fs.emptyDirSync(destination)) {
```
eggplantzzz: This method deletes the contents of the directory if it is not empty. Actually you need to do this check before the temporary directory is created anyways.

adrianmcli: This is so weird, how come you need a temp directory?

eggplantzzz: Perhaps it is unnecessary, this is just the architecture that was handed to me on a silver platter.

adrianmcli:
@CruzMolina Can you quickly describe what this PR does? I'm not too familiar with the inquirer package.

CruzMolina:
Sure, this PR makes it so when using `truffle unbox` or `truffle init` in a non-empty directory, truffle will inform the user of that fact and ask if they wish to move forward with the unbox/init.

This will occur before `truffle-box` attempts to do anything else (create a tmp dir, download a truffle box, etc).

packages/truffle-box/box.js
```diff
@@ -30,15 +32,18 @@ function parseSandboxOptions(options) {
...
     const unpackBoxOptions = {
       logger: options.logger,
       force: options.force
     };
 
     try {
-      options.logger.log("");
+      logger.log("");
+      await Box.checkDir(options, destination);
+      await Box.checkDir(options);
```
CruzMolina:
```suggestion
```

adrianmcli:
Sounds great, I'm all for it!

eggplantzzz:
She passed! @adrianmcli looks like you are good to go with this one.

packages/truffle-box/box.js
```diff
@@ -66,13 +70,35 @@ const Box = {
...
+            type: "confirm",
+            name: "proceed",
+            message: `Proceed anyway?`,
+            default: true
+          }
+        ];
+        const answer = await inquirer.prompt(prompt);
+        if (!answer.proceed) {
+          logger.log("Unbox cancelled");
+          process.exit();
```
eggplantzzz: Is this how we should we handle the case where they do not want to proceed? We don't pass the callback down this far so maybe this is best.

CruzMolina: Yeah, thought about that when moving the logic. I don't see the harm in handling it like this, but I'm not opposed to it being resolved otherwise.

eggplantzzz: Well, breaking the entire command flow this far away from `cli.js` just feels like a code smell to me. It probably won't hurt anything, like you said. I'll take another look and see if there is something that can be done that isn't too involved.

eggplantzzz:
@adrianmcli @CruzMolina Can I get a thumbs up on this PR if you think it can be merged. I'm going to merge it when I get some approval.

adrianmcli:
I have trouble understanding how this code works. It's probably better if @CruzMolina reviews this.

eggplantzzz:
Didn't you write the thing @adrianmcli ??

adrianmcli:
I only wrote the first commit in a hurry before boarding a plane. I had no idea what I was doing, just wanted to get the ball rolling.

PR #1714, opened 2019-02-14
Compiler version output

eggplantzzz:
Include solc version in the output during compilation and update some syntax.

Additionally, this restructures the compilation code so that when errors occur they are not swallowed.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -8,7 +8,10 @@
     "sourceType": "module"
   },
   "rules": {
-    "no-unused-vars": ["error", {"argsIgnorePattern": "^_"}],
+    "no-unused-vars": [
+      "error",
+      { "argsIgnorePattern": "^_", "varsIgnorePattern": "debug" }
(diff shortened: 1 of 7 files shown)


eggplantzzz:
screen shot 2019-02-14 at 3 35 11 pm

fainashalts:
Looks good to me! Cool use of promisify and I don't feel like I see whilst too often either!

gnidan:
Actually... the "writing artifacts" line makes sense to be block-quoted like that.

What if instead, we did something like:

```
  > compiling Foo.sol
  > compiling Bar.sol
  > writing artifacts to blah/blah

Compiled successfully using solc v0.5.4+commit.blah
```

eggplantzzz:
Sure @gnidan, I like the way that output looks.

eggplantzzz:
New and improved output sample.
screen shot 2019-02-15 at 3 16 08 pm

PR #1715, opened 2019-02-14
Add post install script to download solc

eggplantzzz:
This is the first of a series of PRs that will be made in order to remove solc from Truffle's dependencies.

In this installment, a `--download-solc` flag was added to the compile command. It takes a solc version as an argument and uses the `CompilerSupplier` to download and cache that version of the solidity compiler.

In addition, a post install script has been added to the package.json that downloads the "default" version of the solidity compiler that eventually will be used to replace the bundled version.


package.json
```diff
@@ -7,6 +7,7 @@
     "build:truffle-contract-schema": "cd packages/truffle-contract-schema && npm run build",
     "lerna:bootstrap": "lerna bootstrap",
     "bootstrap": "./scripts/bootstrap.sh",
+    "postinstall": "./scripts/downloadSolc.sh",
```
gnidan: Can we just inline the script here? I'm worried about non-UNIX operating systems where that file might not be compatible.

Should be fine if the entirety of the `postinstall` is just `truffle compile --download-solc 0.5.4`

If it turns into more than one line later, where a shell script becomes more warranted, we'll have to look into cross-platform solutions then.

gnidan: Can we just inline the script here? I'm worried about non-UNIX operating systems where that file might not be compatible.

Should be fine if the entirety of the `postinstall` is just `truffle compile --download-solc 0.5.4`

If it turns into more than one line later, where a shell script becomes more warranted, we'll have to look into cross-platform solutions then.

scripts/downloadSolc.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+truffle compile --download-solc 0.5.4
```
gnidan: (Note from above: let's remove this file and inline it into the package.json)

package.json
eggplantzzz: Sounds fine to me

PR #1719, opened 2019-02-17
Increased unbox test timeout

maxsam4:
It takes about 3 seconds (default timeout is 2 seconds) on my network to fetch the default package from GitHub. I have a 100mbit line but the peering between GitHub and my ISP just sucks. I suppose there must be others with not so great internet connections that will benefit from the increased timeout.


gnidan:
Huh weird that this is failing CI. Let me see if `develop` is currently broken.

gnidan:
Alright it's passing! Merging. Thanks @maxsam4!

PR #1731, opened 2019-02-20
Analytics sanitize user info

fainashalts:
Identifying information in the form of the user's working directory name is found in the error stack is passed when a non-numbered error is passed to Google Analytics. This PR should remove that information while preserving the rest of the stack trace.


packages/truffle-core/cli.js
```diff
@@ -83,6 +83,14 @@ command.run(inputArguments, options, function(err) {
         process.exit(err);
       } else {
         let error = err.stack || err.message || err.toString();
+        //remove identifying information if error stack is passed to analytics
+        if (error === err.stack) {
+          let directory = __dirname;
+          let identifyingInfo = directory.split("truffle")[0];
```
gnidan: So, theoretically, if the user's username has `truffle` in it, this would cause incorrect results. (`/Users/truffler/.nvm/8.11.0/node_modules/truffle/packages/truffle-core` or somesuch)

But we probably don't need to bother accounting for this. It might be worth a comment though, just to reduce head scratching in 2 years when someone gets bit by this weird case.

eggplantzzz: We could be a bit more specific...like we could just split it by something like `/truffle/packages`

fainashalts: @gnidan I will add a comment. I realized this, but also felt like it wasn't worth adding more code to account for that. 
@eggplantzzz I did originally make it more specific but it is not always /truffle/packages. Sometimes it is truffle/build and possibly other iterations as well. This seemed like the most efficient way to catch the vast majority of cases.

eggplantzzz: Oh yeah, just an idea...maybe `/truffle/` is enough to minimize conflicts.

fainashalts: @eggplantzzz I looked into this and apparently on some machines the file path uses backslashes instead of forward slashes? Just based on the errors I see in Analytics right now. So I think it's probably better to leave this alone and try to keep it simple if we can. Happy to revisit after we see some more errors in case this solution is woefully inadequate!

eggplantzzz: Oh yeah Windows systems us "\" instead of "/"

PR #1734, opened 2019-02-21
Improve recording of mapping keys (and make session functions async)

haltman-at:
The purpose of this PR is to keep better track of mapping keys; the tracking of mapping keys was already much improved by PR #1701, but keys were still being stored by AST ID, which has a number of problems, as I'll detail below.  In the process of making this change, it was found that to make this work, a number of session functions had to be made `async`.  This introduces a breaking change to the debugger (when used a library; the CLI remains the same).  As succopy h this PR targets `next` rather than `develop`, though @gnidan might decide not to bother with that and just do the breaking change.  (Yeah I know you *already* said not to bother with `next`, but I figure may as well leave the option open.)

Also, a warning: This PR includes one change that could technically be considered a regression.  It has basically no real consequence, so I think it's OK -- you'll see why when I go into detail about it later -- but @gnidan you are warned.

So, then, what does this PR actually do?

Previously, mapping keys were stored by AST ID.  This has two problems:

1. For the most part, only top-level mappings will have an AST ID.  Mappings inside arrays, and mappings inside other mappings, have no AST ID and so we were previously completely unable to keep track of their keys; they would decode as empty no matter what.  (For mappings inside structs, see the next point.)
2. In some cases, multiple mappings can have the same AST ID, leading to us treating their keys as shared.  One way this can happen is if you have multiple instances of the same contract; if the contract has a state variable mapping `map`, its keys would be shared across all instances.  Another way is if you have a mapping inside a struct; if the struct has a mapping member `map`, again, its keys would be shared across all instances of that struct.

Now, mapping keys are stored by address and storage slot.  An address and a slot uniquely determines a mapping, so each mapping -- regardless of whether it's top-level or inside something else -- will have its mapping keys tracked and tracked separately.  Note that slots are not stored as pure offsets, but rather as the `Slot` data type, tracking the whole path to the mapping's storage slot.  Remarkably, it seems to be the case that we are always able to track the full path.  Nonetheless, if I've gotten something wrong and there is a case where we can't, the system will still work, just with some information loss in the path.

To effect the above, the way that `mappingKeys` works has been changed.  Indeed, `mappingKeys` is no longer part of the state; it's been replaced by `mappedPaths`, and the selector has been moved from `data.proc.mappingKeys` to `data.views.mappingKeys`.  This selector, `data.views.mappingKeys`, will return an array of `Slot` objects representing all known mapping keys for the current address.  (It could, I suppose, instead return an object mapping addresses to arrays of Slots; I'll likely change to that in the future, but it wouldn't be a good idea right now.)  The decoder, when decoding a mapping, instead of looking up all mapping keys for its AST ID, will filter the list of mapping keys for those whose `path` matches the slot of the mapping currently being decoded, and use the `key` and `keyEncoding` from such slots in generating the child slots to decode.  (Note that for robustness, slots are compared using `slotAddress`, rather than a deep equality check, just in case I'm wrong above and path information does get lost somehow.)  (OK, in reality it was because I originally thought we *would* sometimes necessarily lose path information, and it didn't occur to me until just now that I could change it, but I think I'll leave it as for the robustness reason, like I said.)

So what does `mappedPaths` contain?  Let's start with an abstract description -- `mappedPaths` keeps track of storage slots (as full paths) that have been accessed and which might be relevant to tracking of mapping keys.  What does that mean concretely?  It means we keep track of all storage slots accessed via an `IndexAccess` or a `MemberAccess`, with some exceptions for cases that can be easily seen to be irrelevant (because there's no way they could lead to a mapping somewhere down the line -- e.g., if a `bytes` is being indexed, or if what's being accessed is not of reference type).

The main part of `mappedPaths`, `byAddress`, contains these paths; they are stored by address, slot address, and data type.  Note that address may be a dummy address, and that the `mappedPaths` reducer does respond to `LEARN_ADDRESS` actions; learning addresses here doesn't presently help much because of how bad we presently are at keeping track of storage across different EVM calls in the same transaction, but hey, it's there.  In any case, I'm planning to take the dummy address system out soon, and then that'll be gone.

By "data type", I mean type identifier, since that's really the only way we have at the moment to represent a data type in a way that can be used as a key in an object.  Also, they're convenient, because they can be fetched right out of the AST for basically any node.  (Although we do mangle them by putting `_ptr` on the end of any reference types that don't have it, since we don't want to distinguish here between pointer and non-pointer versions of those types.)  Them being stored by data type and slot address is of course irrelevant for what gets passed to the decoder, which is just the stripped-down `mappingKeys` object above, but it's an important part of the mechanism by which we keep track of them.

(Why do we pass only this stripped-down version to the decoder?  Because I don't want other things that use the decoder to have to replicate this complicated structure.)

The other part of `mappedPaths` is `decodingMappingKeys`, which retains its existing function.  Maybe I should have renamed it, since now it can be array indices as well.  Oh well.  An earlier version of this branch contained a third part, `byId`, which contained auxiliary info that allowed much of the path to be determined in the saga rather than in the reducer; I've sinced removed this, and all logic for converting slot addresses to actual paths is now handled by the reducer.

So how does the tracking mechanism work, and how do we avoid losing path information?  When a `MemberAccess` or `IndexAccess` occurs, if it's not excluded and (in the index case) if we successfully decoded the index -- note that array indices are now decoded as well, not just mapping keys! -- it gets the slot address for the base node from the base node's associated `StackLiteralPointer`.  It constructs a temporary slot, whose `path` is just a raw `offset` with the base node's slot address, and sends this slot -- along with the current address and any relevant type information -- to the reducer.  The reducer looks up the information for the base node from the given slot's `path`, as well as the given address and type info, and uses the existing slot it finds as the slot's new `path`, replacing the temporary one.  If nothing is found then the temporary one stays; but this should only happen when the index access is of a top-level state variable, so no path information should be lost.

So why don't we lose paths?  What if local `storage` pointers are used?  Well, here's the thing -- that local `storage` pointer must have been initialized from somewhere.  And when that earlier access occurred, we'll have mapped it, meaning we'll recognize that slot address later, even if it's coming from a pointer rather than an index access.

Ah, but what about ways of accessing storage that we don't map?  Because there are ways of accessing storage arrays and structs that we do not keep track of.  Or rather there's one way, which is a copy operation (all other apparent ways actually fall under this, I'm pretty sure).  And, sure, you can set storage this way, deep down within multiple layers of arrays and structs, and none of it will be tracked by our system.  But none of it has to be, because none of this helps you access a mapping element.  As best I can tell, there is no way to access a mapping element other than honest index accesses (possibly split up with pointers but we can handle those).

(Why does `mappedPaths` contain data type info?  Because in constructing our tree of tracked paths, we need to distinguish between a struct or static array on the one hand, and the first element of that struct or array on the other hand; these start in the same storage slot, but being at different levels on the tree, should have distinct `Slot` objects.  Note that the existence of circular structs poses no problem here, since any circularity must be mediated by a dynamic array or mapping, which would introduce hashing.  For similar reasons, we do not need to preserve this information when passing the mapping keys to the decoder.)

So, that's big change number one.  What's big change number two?

Well, you see, the serial nature of how we track mapping keys exposed some weaknesses in the clock we use to keep our sagas in sequence.  It turned out that there was very little keeping them in sequence other than good fortune, and with the addition of this subsystem, that luck has run out.  So, now, to keep things in sequence, each `tickSaga` will call `trace.signalTickSagaCompletion` when done; this will put an action, `SUBTOCK`, and the trace saga will not put a `TOCK` until the appropriate number of `SUBTOCK`s has been received.  (I've also reorganized all `tickSaga`s to work the same way, which includes removing the error-swallowing from the data sagas.)

But that's not enough -- because that keeps the sagas in sequence, but not the CLI.  So I've added a system to do that as well; control actions will now put a `DONE_STEPPING` action when complete, and the `session` functions that invoke them will listen for it before returning.  But to do this, they have to be `async`, and that's a breaking change.  At @gnidan's advice, in fact, I made *all* `session` functions that have side effects `async`, even if they don't currently need to be.  So... breaking change ahoy!

What else has changed that is worth noting?

I fixed an error in the `visibility` function in `truffle-decode-utils`.  I realize we're probably going to merge this straight into `develop`, but just in case we don't and keep it on `next`, I'll do a separate PR later on `develop` so that gets fixed ASAP on `develop`.

I added back in the `baseDefinition` function that I removed earlier, because it turns out that now we occasionally *do* need to fake up a base definition rather than just getting it from `baseType`.  Using `baseType` works fine when you're dealing with a variable declaration... but not when you're dealing with an `expression` node.  Note that I've only used this function where necessary, though -- the existing places where I got rid of it, I've stil left it out.  I don't want to call a function that potentially does definition-faking unless there's a need.

Also, because of this, some of the existing `Definition` functions that previously worked by getting information from the `definition` object directly, now work by parsing the `typeIdentifier`.  This is necessary to handle the case of a faked definition, as I just noted.  Note that I've only made this change where relevant; functions that presently should never encounter a faked definition remain as they were.

(Getting base definitions isn't the only place we're doing fakery in this PR, actually.  Remember I said we need to decode array indices now?  That requires faking up a `uint256` definition!)

Finally, I do have to mention that one change that is *technically* a regression.  It is the following situation: Consider a struct that contains a mapping, passed as a `storage` pointer to a (`public` or `external`) library function.  (Or instead of being passed as the pointer itself, it could be nested below other layers of structs and arrays.)  Since previously mapping keys were stored by AST ID, mapping keys would work across the EVM call boundary here.  (They'd be jumbled together with other keys from other mappings, but the real keys would at least be there among the chaff.)  After this PR, however, this won't work.

Now obviously this is a bug.  The *reason* it won't work right now, after all, is because currently we don't have any system in place for noting that a `DELEGATECALL` EVM stackframe should use the address of the caller (or most recent non-delegate caller, rather), rather than its own address.  Obviously, I'm planning to add such a system and fix this bug later.  However, I didn't think this PR was the place to do it.  That does result in the above regression.

Like I said, though, I think this is OK, because given how poorly storage tracking works across EVM call boundaries at the moment, the fact that we fail to properly track the mapping *keys* in such a case will have essentially no effect.  If we did properly track the mapping keys, we'd still fail to properly decode what was stored at those keys, because we'd just see a bunch of zeroes there.  (And if we got the location to show up in the state by actually accessing it, we'd then have the key in our tracking after all.)  So this regression will have essentially no effect, as mentioned.

(It's also worth noting that this is kind of an edge case, and also that we only started supporting decoding for `storage` locals very recently; if I'd done this PR before I had implemented decoding for `storage` locals, this wouldn't have been a regression at all.  But I think the stronger reason it's OK is the one mentioned above.)

Oh, one last note: The test I added for this PR takes so fricking long.  I just gave it a really long timeout, because I didn't want to not have a test, but man, it takes so long, and this is the gutted version that only tests really basic things.  Oh well?  I don't know what else to do, other than possibly just remove it or gut it further.

I think that basically covers it!  Whew, this took too long...


packages/truffle-debugger/lib/data/reducers.js
```diff
@@ -187,30 +196,104 @@ function mappingKeys(state = DEFAULT_MAPPING_KEYS, action) {
       );
       return {
         decodingStarted: state.decodingStarted + (action.started ? 1 : -1),
-        byId: { ...state.byId }
+        byAddress: state.byAddress
       };
-    case actions.MAP_KEY:
-      let { id, key } = action;
-      debug("mapping id and key: %s, %o", id, key);
+    case actions.MAP_PATH_AND_ASSIGN:
```
gnidan: This case is difficult to reason about at a high-level. Might be worth a couple sentence comment explaining the general approach.

packages/truffle-debugger/lib/data/reducers.js
```diff
@@ -187,30 +196,104 @@ function mappingKeys(state = DEFAULT_MAPPING_KEYS, action) {
...
+
+      return newState;
+
     case actions.RESET:
-      return DEFAULT_MAPPING_KEYS;
+      return DEFAULT_PATHS;
+
+    case actions.LEARN_ADDRESS:
+      return {
+        decodingStarted: state.decodingStarted,
```
gnidan:
```suggestion
        ...state,
```

packages/truffle-debugger/lib/session/index.js
```diff
@@ -117,63 +118,85 @@ export default class Session {
...
+        if (!isStepping && hasStarted && !hasResolved) {
+          hasResolved = true;
+          debug("heard step stop");
+          unsubscribe();
+          resolve(true);
+        }
+      });
+      this.dispatch(stepperAction);
+    });
+  }
```
gnidan: 👌 good abstractin'

packages/truffle-decode-utils/src/definition.ts
```diff
@@ -21,14 +22,24 @@ export namespace Definition {
...
+
   /**
    * For function types; returns internal or external
    * (not for use on other types! will cause an error!)
    * should only return "internal" or "external"
    */
   export function visibility(definition: AstDefinition): string {
     return definition.typeName ?
-      definition.typeName.visibility : definition.typeName.visibility;
+      definition.typeName.visibility : definition.visibility;
```
gnidan: heh

gnidan:
In seriousness though... this was clearly a substantial effort. Nice job!

packages/truffle-debugger/lib/data/reducers.js
haltman-at: Done

haltman-at: Also made the analogous change elsewhere in that reducer

packages/truffle-debugger/lib/data/reducers.js
haltman-at: OK, I added one!  Hopefully you find it sufficiently explanatory.  (Honestly it's possible that the code in this case can be simplified now that I took out the relinking system -- as distinct from the initial linking system that's still in place -- but I didn't want to think through it again. :P )

haltman-at:
OK, this commit addresses the above, and also fixes the `LEARN_ADDRESS` case, which I realized I, uh, entirely forgot to test.  It had a problem.  I fixed it now...

packages/truffle-debugger/lib/data/reducers.js
gnidan: looks good!

PR #1741, opened 2019-02-21
Use vyper beta instead of bleeding edge

gnidan:
Since edge seems to be causing a segfault.


gnidan:
Thanks @jacqueswww for investigating this and offering this solution!

eggplantzzz:
Looks good to me

PR #1742, opened 2019-02-22
Added fallback solc download URL

maxsam4:
- Added fallback URL to fetch version and compiler binary.
- Made ethereum URL default to reduce trust required on truffle.

ps. I am not sure how to test `versionsFallbackUrl` in tests. For some reason, it was using cached information rather than accessing the URL when I tried. It works fine if you manually replace the version url but need to add a test case for it.


gnidan:
This was discussed in Gitter, but including it here for completeness:

Could we switch from using those additional two variables and instead just make a list? Maybe something like:
```javascript
{
  compilerRoots: ["https://relay.trufflesuite.com/solc/bin/", "https://solc-bin.ethereum.org/bin/", "https://ethereum.github.io/solc-bin/bin/"]
}
```

This way we don't need a separate variable for the location of `list.json` (it's just appended), and rather than an `if` statement for a specific fallback, we can just iterate through the list to find the first URL that works.

maxsam4:
Force push to trigger rebuild.

packages/truffle-compile/compilerSupplier/index.js
```diff
@@ -13,9 +13,19 @@ const {
 class CompilerSupplier {
   constructor(_config) {
     _config = _config || {};
-    const defaultConfig = { version: null };
+    const defaultConfig = {
+      version: null,
+      compilerRoots: [
+        "https://solc-bin.ethereum.org/bin/",
+        "https://ethereum.github.io/solc-bin/bin/",
+        "https://relay.trufflesuite.com/solc/bin/"
```
gnidan: Could we make relay.trufflesuite.com first?

packages/truffle-compile/compilerSupplier/index.js
```diff
@@ -13,9 +13,19 @@ const {
...
+      ]
+    };
     this.config = Object.assign({}, defaultConfig, _config);
-    this.strategyOptions = { version: this.config.version };
+    this.strategyOptions = {
+      version: this.config.version,
+      compilerRoots: this.config.compilerRoots
+    };
```
gnidan: Adding this here now suggests to me that maybe we should move all of the `compilerRoots` stuff to VersionRange, since that's the only place where it's used.

packages/truffle-compile/compilerSupplier/index.js
```diff
@@ -76,7 +86,7 @@ class CompilerSupplier {
 
   getReleases() {
     return new VersionRange(this.strategyOptions)
-      .getSolcVersions()
+      .getSolcVersions(0)
```
gnidan: (And if we move it to VersionRange then we don't need the argument here)

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -4,8 +4,11 @@ const fs = require("fs");
 class LoadingStrategy {
   constructor(options) {
     const defaultConfig = {
-      versionsUrl: "https://relay.trufflesuite.com/solc/bin/list.json",
-      compilerUrlRoot: "https://relay.trufflesuite.com/solc/bin/",
+      compilerRoots: [
+        "https://solc-bin.ethereum.org/bin/",
+        "https://ethereum.github.io/solc-bin/bin/",
+        "https://relay.trufflesuite.com/solc/bin/"
+      ],
```
gnidan: Do we need this to be duplicated here?

packages/truffle-compile/compilerSupplier/loadingStrategies/VersionRange.js
```diff
@@ -90,18 +90,18 @@ class VersionRange extends LoadingStrategy {
     const solcFileName = this.getCachedSolcFileName(commit);
     if (solcFileName) return this.getCachedSolcByFileName(solcFileName);
 
-    const allVersions = await this.getSolcVersions(this.config.versionsUrl);
+    const allVersions = await this.getSolcVersions(0);
```
gnidan: I think we should make the `0` be the default value for the argument, so we don't need to change the interface for that method.

packages/truffle-compile/test/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -8,17 +8,19 @@ describe("LoadingStrategy base class", () => {
   beforeEach(() => {
     instance = new LoadingStrategy();
     expectedDefaultConfig = {
-      versionsUrl: "https://relay.trufflesuite.com/solc/bin/list.json",
-      compilerUrlRoot: "https://relay.trufflesuite.com/solc/bin/",
+      compilerRoots: [
+        "https://solc-bin.ethereum.org/bin/",
+        "https://ethereum.github.io/solc-bin/bin/",
+        "https://relay.trufflesuite.com/solc/bin/"
```
gnidan: (Same as above: can we move relay.trufflesuite to the top?)

packages/truffle-compile/compilerSupplier/index.js
maxsam4: `compilerRoots` is configurable by the user to I need to pass the user config to the `VersionRange`. 
I have removed the default `compilerRoots` from `compilerSupplier`.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
maxsam4: Removed the duplicate from `compilerSupplier`.

gnidan:
@maxsam4 this is looking good! I was about to merge it for you but it looks like there's now a failing test. Mind looking into that? I'd love to get this out in the next release

PR #1756, opened 2019-02-25
Fix profiler

eggplantzzz:
Make method async to get rid of some indentation and to cause errors that occur in that block of code to not get swallowed up whole.


eggplantzzz:
Well, the thing is, I'm not sure what the strategy would be. It was a problem with the architecture and not some weird input. If you have any ideas, leave them here.

PR #1765, opened 2019-02-26
Create event manager system with reporters

eggplantzzz:
WORK IN PROGRESS ---- DO NOT MERGE YET

This PR creates an event manager system which allows Truffle to emit events during command flows. Different reporters etc. can subscribe to certain events and handle them in whatever way is useful. The first proof of concept will be reporting for the compile and unbox commands.


packages/truffle-reporters/compile/index.js
```diff
@@ -0,0 +1,63 @@
...
+  },
+
+  finishJob(options, config) {
+    const { globalConfig } = options;
+    const { compilersInfo } = config;
+    const logger = options.logger || console;
+    if (Object.keys(compilersInfo).length > 0) {
+      logger.log(
+        `${OS.EOL}    > artifacts written to ${
+          globalConfig.contracts_build_directory
```
CruzMolina: Looks like for `truffle compile`, `globalConfig` is passed w/o a `contracts_build_directory` (might be happening in `truffle-workflow-compile/index.js`?)

screen shot 2019-02-26 at 2 40 24 pm

packages/truffle-compile/index.js
```diff
@@ -367,6 +365,9 @@ compile.necessary = function(options, callback) {
     if (err) return callback(err);
 
     if (updated.length === 0 && options.quiet !== true) {
+      if (options.eventManager) {
+        options.eventManager.emitEvent("compile:nothingToCompile");
```
CruzMolina: Looks like if nothing is updated AND there are no contracts to compile the `nothingToCompile` event gets emitted/reported twice.

screen shot 2019-02-26 at 2 48 04 pm

packages/truffle-reporters/compile/index.js
CruzMolina: relevant line in screenshot: `> artifacts written to undefined`

packages/truffle-compile/index.js
CruzMolina: (Also liked the ========= line stuff in the earlier code)

packages/truffle-reporters/unbox/index.js
```diff
@@ -0,0 +1,112 @@
...
+
+class UnboxReporter {
+  constructor(options) {
+    this.ora = ora;
+    this.initializeListeners(options);
+  }
+
+  cleaningTempFiles(isStart) {
+    if (isStart) {
+      this.spinner = this.ora("cleaning up temporary files").start();
```
CruzMolina:
```suggestion
      this.spinner = this.ora("Cleaning up temporary files").start();
```

packages/truffle-reporters/compile/index.js
```diff
@@ -0,0 +1,63 @@
...
+    emitter.on("compile:finishJob", this.finishJob.bind(this, options));
+    emitter.on("compile:warnings", this.warnings.bind(this, options));
+    emitter.on(
+      "compile:compiledSources",
+      this.compiledSources.bind(this, options)
+    );
+    emitter.on(
+      "compile:nothingToCompile",
+      this.nothingToCompile.bind(this, options)
+    );
```
gnidan: This comment probably belongs someplace more general, but I guess I'll pick on this spot.

This block of code, this whole module, all the similar modules that encapsulate the listener setup and the reporting behavior... having these laid out all in a PR like this is foundational. Thank you for taking the time to write this initial implementation @eggplantzzz; this code and whatever it becomes is going to teach us many lessons.

This feels like our primary resource in designing/validating the entire "subscribe" side of the eventing system. It shows off actual usage when it comes to building a reporter; we can plainly see this new, emerging, external interface to Truffle. If we do our jobs right, there will be thousands or even tens of thousands of files like this one, each of them carving out its own little window into Truffle's operations.

I'll quit with the oratory and summarize: getting this right feels both to be of substantial importance and to pose a significant challenge.

To start with some thoughts about design:

- Encapsulating subscription and handling: yes! I like that these modules define the handling behavior as well as hooking them up as listeners.

- `.on()`s/`.bind()`s for days: intimidating! That's a lot of manual set-up we're requiring, and it's all the same, everywhere!

- These `.on()`s are very far away in the code from the corresponding handler methods! It's hard to look at this collection at a glance and see how things connect. Can we find a way to make an object of event names? Something along the lines of: 
  ```javascript
  {
    "compile:startJob": (data) => { /* ... */ },
    /* ... */
  }
  ```
  Since that could more succinctly represent the important distinctions.

- Magic strings! I strongly urge us to avoid using string literals for these event names. They are risky, difficult to change, and afford no protection against typos.

- Moreover, this whole file relates to `/^compile:/`. Can we make it possible to avoid having to write this everywhere? I was sketching out something like this, but it's difficult to make it not-ugly:
  ```javascript
  module.exports = defineListeners("compile", {
    "startJob": () => {
      this.logger.log("Starting compilation...");
    },

    /* ... */
  });
  ```
  I recognize that this is at odds with the point about magic strings. Was looking for a way to exemplify both in one example, but it's tricky and I've already been at it for an hour.

- Can we avoid passing `options` to every handler? Can we just turn `options` into `this`? (This might add complications the design...)

Anyway, these are just some of my initial thoughts. Hopefully they are useful to the discussion.

packages/truffle-box/box.js
```diff
@@ -28,7 +27,8 @@ function parseSandboxOptions(options) {
 }
 
 const Box = {
-  unbox: async (url, destination, options = {}) => {
+  unbox: async (url, destination, options = {}, config) => {
```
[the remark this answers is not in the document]
eggplantzzz: No that options is the `unboxOptions`, perhaps just passing the `eventManager` itself is enough as I think that is all that is used in this method.

packages/truffle-core/lib/EventManager.js
```diff
@@ -0,0 +1,45 @@
...
+    const initializationOptions = {
+      globalConfig,
+      emitter,
+      logger,
+      muteReporters
+    };
+    this.initializeSubscribers(initializationOptions);
+  }
+
+  emitEvent(event, data) {
```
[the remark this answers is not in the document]
eggplantzzz: Yes, I wanted to ask everyone to weigh in on what they think these things should be called. Emit would cause a lot less noise.

packages/truffle-core/lib/commands/compile.js
```diff
@@ -48,13 +48,16 @@ const command = {
...
+    const { CompilerSupplier } = require("truffle-compile");
     const supplier = new CompilerSupplier();
+    const eventManager = require("../eventManager");
 
-    const Contracts = require("truffle-workflow-compile");
-    const Config = require("truffle-config");
     const config = Config.detect(options);
 
+    config.eventManager = eventManager(config);
+
```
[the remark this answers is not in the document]
eggplantzzz: Sure, without too much thought on the matter it seems like a good idea to just attach `eventManager` by default to the config. One thing though, will there ever come a time when we might want to initialize the listeners based on what command is run? Currently they all get set unless you use an option like `quiet=true`. I suppose at that time we might be able to add more options and edit `eventManager` to subscribe/initialize based on that?

packages/truffle-reporters/Reporters.js
```diff
@@ -0,0 +1,15 @@
+const compile = require("./compile");
+const UnboxReporter = require("./unbox");
+
+class Reporters {
+  constructor(initializationOptions) {
```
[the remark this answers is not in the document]
eggplantzzz: One thing that comes to mind is that I feel like there is some amount of confusion when I see options/config and inconsistencies in the naming of that throughout a command flow. I wanted it to be clear that we were talking about the options for this particular thing and not the global config object that gets passed. It seems ideal to me that that global config object has a proper name that remains consistent in Truffle though I know we have to deal with what we have right now....just a thought.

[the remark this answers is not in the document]
gnidan: I'm convinced!

packages/truffle-reporters/compile/index.js
gnidan: Another idea (not very considered): what about regex/globbed event listeners?
```javascript
defineListeners("compile", {
  ".*": (data, event) => { console.log("event %s", event) }
});
```

gnidan: Another question: how does initialization work for common things across methods (e.g. `logger`)
```javascript
module.exports = defineListeners({
  namespace: "compile",
  initialize: (config) => {
    this.logger = config.logger || console
  }
}, {
  "startJob": () => {
    this.logger.log("start job")
  },
  /* ... */
});
```

packages/truffle-config/index.js
```diff
@@ -370,6 +382,27 @@ Config.detect = (options = {}, filename) => {
   return Config.load(file, options);
 };
 
+// When new options are passed in, a new eventManager needs to be
+// attached as it might override some options (e.g. { quiet: true })
+const attachNewEventManager = (config, newOptions) => {
+  const currentEventManagerOptions = config.eventManager.initializationOptions;
+  const { quiet, logger, globalConfig } = newOptions;
+  const optionsToMerge = {};
+
```
eggplantzzz: Can we do this differently? I would love to just pull out the values but in the case that the new values are undefined I don't want it to overwrite the old stuff that we want to keep. Let me know if you have any ideas.

packages/truffle-reporters/compile/index.js
```diff
@@ -0,0 +1,46 @@
+const OS = require("os");
+
+module.exports = {
+  handlers: {
+    "compile:startJob": function() {
```
gnidan: If we go with standardizing on verbs like `start` and `finish`, then "job" becomes redundant... the entirety of "compile" is between `compile:start` and `compile:finish` (or `compile:error`)... profiling happens during that process and would be between `compile:profile:start` and `compile:profile:finish`.

If we furthermore stick with only imperative mood, it's shorter to type and consistent: `compile:profile:warn`, etc., and then easier to understand at a glance when looking at a streamed list of event names.

packages/truffle-box/box.js
```diff
@@ -53,15 +52,16 @@ const Box = {
...
-      cleanupSpinner.succeed();
+      eventManager.emit("unbox:cleaningTempFiles:end");
 
-      await utils.setUpBox(boxConfig, destination);
+      await utils.setUpBox(boxConfig, destination, eventManager);
 
       return boxConfig;
     } catch (error) {
       if (tempDirCleanup) tempDirCleanup();
+      eventManager.emit("unbox:jobFailed");
```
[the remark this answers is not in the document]
eggplantzzz: Sure, should we do `succeed` and `fail` instead of `end`?

eggplantzzz:
When thinking about the globbing for event names, I start to wonder how far we want to take that. It might be sufficient to have a "pseudo globbing" system...in other words, it seems like overkill to have actual globbing. It might seem reasonable, however, to just use the "*" as a wildcard for one term like you mentioned above. Let me know if you have different ideas about how far you want to take the globbing.

packages/truffle-reporters/Reporter.js
```diff
@@ -1,11 +1,13 @@
 class Reporter {
-  constructor({ options, emitter }) {
+  // const supportedEvents = { compile: compileEvents, unbox: unboxEvents };
```
gnidan: Challenge: can we make it so that we don't have to know the possible events in advance?

eggplantzzz: Did you have any ideas for this? It seems like we'll have to know what events exist in order to match against to set up the handlers, right?

eggplantzzz: So on initialization of a reporter, the event names are split up into two categories: one for names without wildcard characters (`*` or `**`) and another for names that do contain them.
The names without wildcards subscribe directly to whatever event they are named. The names with wildcards have a table built for them in which a regex can be looked up that will match the appropriate events. Whenever any event is fired, the reporter looks through the table of regexes to see if any of the events match the fired event. If it does, it runs the appropriate handler.

PR #1766, opened 2019-02-26
Add multiple-advance command

haltman-at:
This PR allows one to advance the state a specified number of steps by specifying the number after the `;` command.  I've also updated the help text to mention this.  This PR was inspired by the discussion in PR #1346, and I'm hoping it successfully addresses the problems raised there.

Note that this is our first (intentional) example of a command with an argument being repeatable.  The way I set it up is so that if you repeat it, the argument is lost on repeats; so that `; 3`, followed by a repeat, yields 4 steps, not 6.  Not sure if this is the best way, just what I figured made sense.

### DIFF
--- a/packages/truffle-debug-utils/index.js
+++ b/packages/truffle-debug-utils/index.js
@@ -9,7 +9,7 @@ var commandReference = {
   "i": "step into",
   "u": "step out",
   "n": "step next",
-  ";": "step instruction",
+  ";": "step instruction (include number to step multiple)",
   "p": "print instruction",
   "h": "print this help",
   "v": "print variables and values",
@@ -125,8 +125,9 @@ var DebugUtils = {
     ];
 
     var commandSections = [
-      ["o", "i", "u", "n", ";"],
-      ["p", "h", "q", "r"],
+      ["o", "i", "u", "n"],
+      [";", "p"],
+      ["h", "q", "r"],
       ["b", "B", "c"],
       ["+", "-"],
       ["?"],
--- a/packages/truffle-debugger/lib/controller/sagas/index.js
+++ b/packages/truffle-debugger/lib/controller/sagas/index.js
@@ -18,7 +18,6 @@ import controller from "../selectors";
 //reducers.js as well!
 const CONTROL_SAGAS = {
   [actions.ADVANCE]: advance,
-  [actions.MULTIADVANCE]: multiadvance,
   [actions.STEP_NEXT]: stepNext,
   [actions.STEP_OVER]: stepOver,
   [actions.STEP_INTO]: stepInto,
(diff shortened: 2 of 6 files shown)


packages/truffle-debugger/lib/controller/sagas/index.js
```diff
@@ -14,8 +14,11 @@ import * as actions from "../actions";
 
 import controller from "../selectors";
 
+//NOTE: when updating this don't forget to update CONTROL_ACTIONS in
+//reducers.js as well!
 const CONTROL_SAGAS = {
   [actions.ADVANCE]: advance,
+  [actions.MULTIADVANCE]: multiadvance,
```
gnidan: Here's an idea: what do you think about making `actions.ADVANCE` just optionally take an argument for the number of steps to advance? This way we won't need the extra action/saga/etc.

haltman-at: I thought of that, but I liked having `advance` as a separate more primitive thing.  It would throw off some of the correspondence, at least, to do that.  Like, all the other existing sagas that currently call `advance` -- would they still call `advance`, but with an argument?  Would they directly call `trace.advance`?  Would what's currently `advance` be renamed to something else?  I decided to leave it alone and just add it as something separate.

gnidan: > would they still call `advance`, but with an argument?

Default values for arguments would solve this concern.

I see the argument for having a separate action, but I'm not sure that `advance(n)` is any less primitive than `advance(1)`? At least in the controller.

Anyway my opinion isn't too strong here; I just think it's cleaner to modify `advance` rather than make a new action. But I leave it up to you to decide what you prefer.

haltman-at: Oh hm -- yeah doing it with a default argument would make sense.  OK, I think I'll change it to that way after all.

packages/truffle-debugger/lib/controller/sagas/index.js
```diff
@@ -44,12 +46,14 @@ export function* saga() {
...
- * Advance the state by one instruction
+/*
+ * Advance the state by the given number of instructions (but not past the end)
  */
-function* advance() {
-  // send action to advance trace
-  yield* trace.advance();
+function* advance(action = actions.advance()) {
+  let { count } = action;
+  for (let i = 0; i < count && !(yield select(controller.finished)); i++) {
```
gnidan: I miss C

packages/truffle-debug-utils/index.js
```diff
@@ -9,7 +9,7 @@ var commandReference = {
   "i": "step into",
   "u": "step out",
   "n": "step next",
-  ";": "step instruction",
+  ";": "step instruction (include number to step multiple)",
```
gnidan: Space here is kind of limited, hard to describe this feature clearly in the help like that.

I was trying to find a way to describe it by way of example, but I'm coming up short. Maybe we can just remove it? Or maybe you have other ideas.

My main concern is mostly that the following `(p) print instruction` note is harder to spot:

screen shot 2019-02-26 at 6 51 07 pm

haltman-at: Hmm, I dunno.  I think I'm going to just leave this as-is.  I figure we can rearrange it later maybe...

PR #1768, opened 2019-02-26
Implement variable() and add interface for watching mapping keys

haltman-at:
This PR implements the `variable` function in the contract decoder, and also adds an interface for watching and unwatching mapping keys in the contract decoder.

The `variable` function works just like the `state` function, but it takes an argument for what variable to decode.  I allowed it to take either a name (as string) or an id (as number).  Like `state`, it presently ignores the `blockNumber` argument, but I'm intending to implement this in a later PR.

The big addition is `watchMappingKey` and `unwatchMappingKey`.  These both use the same argument form.  Specifically, one calls `watchMappingKey(variable, index1, ....)` where `index1`, `index2`, etc, are a sequence of indices into it.  So, e.g., if `arr` is an array of mappings which have strings as keys, then you could write `watchMappingKey("arr", 1, "hello")` to watch `a[1]["hello"]`.  Or if `structure` is a struct containing a mapping `str.map` with strings as keys, you could write `watchMappingKey("structure", "map", "hello")`.

In more detail:

* Variables and struct members can be specified by name (string) or ID (number)
* Numbers can be given as BN, number, or numeric string
* Bytes and addresses should be given as hex strings (of appropriate length) beginning with `0x`; checksum case is not required
* Strings should be given as strings, obviously
* Booleans can be given either as booleans or as strings (in which case `"false"` is considered false and everything else is true)

Beyond that, input is not checked, so please don't expose these directly to GUI users!  Often invalid input will result in a failure to watch anything, but that's it; and sometimes entering things in the wrong form (e.g., including extra trailing zero bytes on a statically-sized `bytes`) will successfully watch that key but mess things up internally a bit.

Note that `watchMappingKey` will also watch all mapping keys above -- e.g., if `map` is a mapping of strings to mappings of strings to strings, then `watchMappingKey("map", "string1", "string2")` will watch both `map["string1"]` and `map["string1"]["string2"]`.  Conversely, `unwatchMappingKey` will unwatch everything below.  I wasn't sure if the latter was the right decision, but it's what I decided to go with.

Note that a fair amount of the logic in the mapping keys function more or less duplicates what can be found in the debugger's data saga; I suppose it could be factored out, but, well, I decided not to in this case.

It's possible that not all of this is as tested as well as it should be.  I don't really use the contract decoder; I added some basic tests in the `truffle-decoder` tests but, like I said, only very basic ones.


haltman-at:
So turns out that we can't use `util.isDeepStrictEqual` because that's a Node 9 feature.  So, I've replaced it with lodash's `isEqual`.  While I was at it I also fixed the incorrectly-done installation of `lodash.range` I did earlier.

seesemichaelj:
Hey @haltman-at, this seems like an interesting way for watching mapping keys. I'm not super fond of by-name lookup but it seems you still support by-id which is cool. I do like the support for multi-level watching in a single call!

While you may not use this feature, I strongly encourage adding more detailee tests (I.e. multi-level mappings, mappings of structs (multi-level)) since it's pretty easy to do a quick discovery if they work.

Also the large switch statement of constructSlot may seem that it could be split up (i.e. call functions for each case instead of several statements). Also is it possible it is a more generic function that belongs in another file? Not sure, just food for thought.

Not going to approve/disapprove this since I no longer own this, just wanted to provide feedback since you requested it.

PR #1771, opened 2019-02-27
Update `truffle obtain`

eggplantzzz:
Update `truffle obtain` to use '--compiler=version' syntax. Also update the postinstall script to use this syntax to be more clear to the user when installing Truffle.

### DIFF
--- a/packages/truffle-core/test/lib/commands/obtain.js
+++ b/packages/truffle-core/test/lib/commands/obtain.js
@@ -1,0 +1,40 @@
+const assert = require("chai").assert;
+const command = require("../../../lib/commands/obtain");
+const sinon = require("sinon");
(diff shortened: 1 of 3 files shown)


eggplantzzz:
No there aren't any tests! But I will remedy that.

packages/truffle-core/test/lib/commands/obtain.js
```diff
@@ -0,0 +1,40 @@
+const assert = require("chai").assert;
+const command = require("../../../lib/commands/obtain");
+const sinon = require("sinon");
+const CompilerSupplier = require("truffle-compile").CompilerSupplier;
+let options, done, solc;
+
+describe("obtain", () => {
```
CruzMolina: 😎

PR #1773, opened 2019-02-28
Enhance truffle develop configurability

CruzMolina:
This PR modernizes, streamlines, & enhances the configurability of `truffle develop`.

It enables the following `ganache-core` options to be pre-configured:

- `host`
- `port`
- `network_id`
- `gasLimit` (gas)
- `gasPrice`
- `hardfork`

Example `truffle-config.js`:

```
module.exports = {
  networks: {
    develop: {
      host: "127.0.0.1",
      port: 7545,
      network_id: 666,
      gas: 1000,
      gasPrice: 1,
      hardfork: "constantinople"
    }
  }
};
```

### DIFF
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -51,7 +51,7 @@ const command = {
       customConfig.accounts || 10
     );
 
-    const onMissing = name => "**";
+    const onMissing = () => "**";
 
     const warning =
       ":warning:  Important :warning:  : " +
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -15,7 +15,7 @@ const command = {
     usage: "truffle develop",
     options: []
   },
-  runConsole: async (config, ganacheOptions, done) => {
+  runConsole: (config, ganacheOptions, done) => {
     const Console = require("../console");
     const Environment = require("../environment");
 
@@ -36,11 +36,11 @@ const command = {
 
     const c = new Console(console_commands, config.with({ noAliases: true }));
     c.start(done);
-    c.on("exit", async () => {
+    c.on("exit", () => {
       process.exit();
     });
   },
-  run: async (options, done) => {
+  run: (options, done) => {
     const Config = require("truffle-config");
     const Develop = require("../develop");
 
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -83,17 +83,11 @@ const command = {
         config.logger.log();
 
         config.logger.log(`Accounts:`);
-        for (var i = 0; i < accounts.length; i++) {
-          const account = accounts[i];
-          config.logger.log(`(${i}) ${account}`);
-        }
+        accounts.forEach((acct, idx) => config.logger.log(`(${idx}) ${acct}`));
         config.logger.log();
 
         config.logger.log(`Private Keys:`);
-        for (var i = 0; i < privateKeys.length; i++) {
-          const privateKey = privateKeys[i];
-          config.logger.log(`(${i}) ${privateKey}`);
-        }
+        privateKeys.forEach((key, idx) => config.logger.log(`(${idx}) ${key}`));
         config.logger.log();
 
         config.logger.log(`Mnemonic: ${mnemonic}`);
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -75,6 +75,10 @@ const command = {
       noVMErrorsOnRPCResponse: true
     };
 
+    if (customConfig.hardfork !== null && customConfig.hardfork !== undefined) {
+      ganacheOptions["hardfork"] = customConfig.hardfork;
+    }
+
     Develop.connectOrStart(ipcOptions, ganacheOptions, started => {
       const url = `http://${ganacheOptions.host}:${ganacheOptions.port}/`;
 
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -1,7 +1,8 @@
-var emoji = require("node-emoji");
+const emoji = require("node-emoji");
 const mnemonicInfo = require("truffle-core/lib/mnemonics/mnemonic");
+const util = require("util");
 
-var command = {
+const command = {
   command: "develop",
   description: "Open a console with a local development blockchain",
   builder: {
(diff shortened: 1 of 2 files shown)


packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,102 +15,83 @@ var command = {
...
-    let bTime = 0;
-
-    if (customConfig) {
-      numAddresses = customConfig.accounts ? customConfig.accounts : 10;
-      defaultEtherBalance = customConfig.defaultEtherBalance
-        ? customConfig.defaultEtherBalance
-        : 100;
-      bTime = customConfig.blockTime ? customConfig.blockTime : 0;
-    }
+  async run(options, done) {
```
adrianmcli: Same as above comment:

```js
run: async (options, done) => { ... }
```

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,102 +15,83 @@ var command = {
     usage: "truffle develop",
     options: []
   },
-  runConsole: function(config, ganacheOptions, done) {
-    var Console = require("../console");
-    var Environment = require("../environment");
+  async runConsole(config, ganacheOptions, done) {
+    const Console = require("../console");
```
adrianmcli: Typically, capitalized variables are used for classes and constructor functions. Things like:

```js
const MyClass = require("../MyClass.js")
const thing = new MyClass()
```

In this case, the thing you are requiring from `../console` is just an object literal (or an object "instance", some might say). As a result, it should really be:

```js
const consoleCommand = require("../console")
```

But since this is done everywhere in the code, editing this now might make it difficult to read with the rest of the code. I'll leave this up to you, but I just wanted to point out that this is kind of non-standard.

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,102 +15,83 @@ var command = {
...
+    const Environment = require("../environment");
```
adrianmcli: Non-standard variable naming with leading capital. See above, feel free to ignore/resolve.

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,102 +15,83 @@ var command = {
...
-    if (customConfig) {
-      numAddresses = customConfig.accounts ? customConfig.accounts : 10;
-      defaultEtherBalance = customConfig.defaultEtherBalance
-        ? customConfig.defaultEtherBalance
-        : 100;
-      bTime = customConfig.blockTime ? customConfig.blockTime : 0;
-    }
+  async run(options, done) {
+    const Config = require("truffle-config");
+    const Develop = require("../develop");
```
adrianmcli: For the above 2 lines:

Non-standard variable naming with leading capital. See above, feel free to ignore/resolve.

CruzMolina:
Thanks for the review @adrianmcli ! Lovin the suggestions so far

🔬

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,102 +15,83 @@ var command = {
     usage: "truffle develop",
     options: []
   },
-  runConsole: function(config, ganacheOptions, done) {
-    var Console = require("../console");
-    var Environment = require("../environment");
+  async runConsole(config, ganacheOptions, done) {
```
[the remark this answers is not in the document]
CruzMolina: 🤔 

Strongly considering this

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,102 +15,83 @@ var command = {
...
+    const available_commands = Object.keys(commands).filter(
+      name => !excluded.includes(name)
+    );
 
-    var console_commands = {};
-    available_commands.forEach(function(name) {
+    const console_commands = {};
+    available_commands.forEach(name => {
       console_commands[name] = commands[name];
     });
```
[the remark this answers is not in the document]
CruzMolina: Boo, looks like truffle is not yet ready for this wizardry ⚡️

packages/truffle-core/lib/commands/develop.js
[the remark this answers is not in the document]
CruzMolina: Okay, convinced 👍

packages/truffle-core/lib/commands/develop.js
CruzMolina: hmmm... I can see the argument for the console command

CruzMolina: Oh, nm. it's not requiring the console cmd, it's the console lib.

CruzMolina: issue is we sort've treat all these libs you pointed out as potential classes. i think it makes more sense to keep them capitalized for now. it helps differentiate between `truffle-core/lib/console.js` and `truffle-core/lib/commands/console.js`.

CruzMolina: i see your point tho and feel there's definitely potential for a much larger restructuring

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,104 +15,79 @@ var command = {
...
-      network_id: 4447,
-      total_accounts: numAddresses,
-      default_balance_ether: defaultEtherBalance,
-      blockTime: bTime,
-      mnemonic: mnemonic,
-      gasLimit: config.gas,
+    const ganacheOptions = {
+      host: customConfig.host || "127.0.0.1",
+      port: customConfig.port || 9545,
+      network_id: customConfig.network_id || 5777,
```
gnidan: (I don't think we should change this from 4447 just because of MetaMask. We can work with them to get a better solution)

adrianmcli:
@CruzMolina ah right, so we can't use object spread, but what about:

```js
const console_commands = available_commands.reduce((acc, name) =>
  Object.assign({}, acc, { [name]: commands[name] })
, {})
```

Or more verbosely:

```js
const console_commands = available_commands.reduce((acc, name) => {
  const newCommand = { [name]: commands[name] }
  return Object.assign({}, acc, newCommand)
}, {})
```

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,104 +15,79 @@ var command = {
     usage: "truffle develop",
     options: []
   },
-  runConsole: function(config, ganacheOptions, done) {
-    var Console = require("../console");
-    var Environment = require("../environment");
+  runConsole: async (config, ganacheOptions, done) => {
```
gnidan: To avoid dual-wielding concurrency mechanisms, you can get around this pretty easily:

Firstly, remove the `done` argument and instead wrap the whole function with a `util.callbackify()`
```suggestion
  runConsole: util.callbackify(async (config, ganacheOptions) => {
```

Then, anywhere `done` is used in the function, either `return` directly (for success) or `throw` directly (for error).

This preserves the external callback interface while giving us all the benefits of `await`

adrianmcli: Also wait a second, `await` isn't even used anywhere in this function?

packages/truffle-core/lib/commands/develop.js
eggplantzzz: I think I would vote that we change them as we see them.

packages/truffle-core/lib/commands/develop.js
CruzMolina: ⚔️

packages/truffle-core/lib/commands/develop.js
CruzMolina: Pulling out the `network_id` specific commit

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,104 +15,79 @@ var command = {
...
-        ? customConfig.defaultEtherBalance
-        : 100;
-      bTime = customConfig.blockTime ? customConfig.blockTime : 0;
-    }
+  run: async (options, done) => {
+    const Config = require("truffle-config");
+    const Develop = require("../develop");
+
+    const config = Config.detect(options);
+    const customConfig = config.networks.develop || {};
```
eggplantzzz: Hmm, I don't know if this matters but we should consider the fact that this will take away the name "develop" from users and dedicate it to this functionality. What do we think about this?

packages/truffle-core/lib/commands/develop.js
```diff
@@ -14,104 +15,79 @@ var command = {
...
+      customConfig.accounts || 10
     );
 
-    var onMissing = function(name) {
-      return "**";
-    };
+    const onMissing = name => "**";
 
-    var warning =
+    const warning =
```
eggplantzzz: Is this warning message still appropriate or do we need to update it?

packages/truffle-core/lib/commands/develop.js
eggplantzzz: Get 'im outta here!

packages/truffle-core/lib/commands/develop.js
CruzMolina: It's already been taken away from users. 😈 

https://truffleframework.com/docs/truffle/getting-started/using-truffle-develop-and-the-console#configuring-truffle-develop

packages/truffle-core/lib/commands/develop.js
eggplantzzz: The same with `run`.

packages/truffle-core/lib/commands/develop.js
CruzMolina: I *think* we still need a warning msg.

What would you update it to?

packages/truffle-core/lib/commands/develop.js
CruzMolina: @gnidan thoughts?

Kinda goes against the `async` everything policy.

eggplantzzz: Yeah but if it is not an async function then there really is no reason to make it a function that returns a Promise.

packages/truffle-core/lib/commands/develop.js
eggplantzzz: Oh ok then that is that.

packages/truffle-core/lib/commands/develop.js
eggplantzzz: Like, we don't really want everything async. He's referring to methods that need to return Promises. async fundamentally changes the method in that it causes it to then return a Promise.

packages/truffle-core/lib/commands/develop.js
eggplantzzz: I'm not sure, technically that mnemonic should be secure, right? At least it should be unless users are not careful. I was thinking that this message was in reference to the "Days of Olde" with the classic 'candy maple ...' business.

packages/truffle-core/lib/commands/develop.js
CruzMolina: Ok, so two votes to revert `run` & `runConsole` back to sync funcs.

Unless @gnidan stops me, will be pushing the changes shortly.

packages/truffle-core/lib/commands/develop.js
CruzMolina: Yeah, the new mnemonic system should be secure, but maybe we need to keep this in just in case for liability reasons.

packages/truffle-core/lib/commands/develop.js
CruzMolina: Reverted 😎

adrianmcli:
lgtm!

gnidan:
@CruzMolina allow me to steal your merge button privileges so I can gather release notes at an inappropriate hour.

Issue #1791, opened 2019-03-05
artifact.require from node_modules


Amxx:
Context
======

As everyone, I have smart contract sources which path is `./contract/xxx.sol`. Once compiled, the artefact is placed in `./build/contract/xxx.json`. If I want to deploy it, I just do `artifact.require("xxx")` to import it and then use the deployer. Simple stuff!

Then I have smart contracts in dependencies. The solidity source is in `./node_modules/somerepo/contract/yyy.sol` and the artifact is precompiled in `./node_modules/somerepo/build/contract/yyy.json`. So far, I've been able to deploy them during the migration step by doing `artifact.require("somerepo/yyy")` and then use the deployer on the imported artefact.

Issue
====

After upgrading from truffle `5.0.5` to `5.0.6`, I am no longer able to deploy contracts objects in dependencies. This can be precompiled token contracts used for testing, ENS registrars and solvers (again for testing) ... you get it, they are essential to setting up test environments!

Question
=======

- Can this feature come back ?
- or can we have a better documentation of artifact.require if it's still possible but not documented how to do it.

gnidan:
Oh shoot! I broke this! I'll fix it it today and it will go out with the next release. Thanks for the heads up.

gnidan:
@Amxx we just published a fix in Truffle v5.0.7. Closing this because I'm an optimist. Could you try it out and let me know if that solves the issue? Thanks!

PR #1792, opened 2019-03-05
Remove resolver intercept cache basename logic

gnidan:
Ref: #1791 

Replace basename stripping with more flexible stripping of leading `./` and trailing `.sol`


gnidan:
Oops that's what happens when I try to escape a path separator before coffee.

gnidan:
Thanks @CruzMolina!

PR #1794, opened 2019-03-05
Add web3 version logging output

eggplantzzz:
Add web3 version to the list of things that get logged when using 'truffle version'


packages/truffle-core/lib/version.js
```diff
@@ -43,11 +44,16 @@ const logSolidity = (logger = console, versionInformation, config) => {
   }
 };
 
+const logWeb3 = (logger = console) => {
+  logger.log(`Web3 v${web3pkg.version}`);
```
gnidan:
```suggestion
  logger.log(`Web3.js v${web3pkg.version}`);
```

PR #1797, opened 2019-03-06
Add further support for modifier parameters (with some exceptions)

haltman-at:
This PR attempts to further support the stack allocation of parameters to modifiers and base constructors (invocations of base constructors are treated by Solidity essentially like invocations of modifiers, and from here on out when I speak of "modifiers" I am including base constructors unless otherwise specified or implied).  It is not entirely successful at this.  Some modifiers will fail to have their parameters allocated.  Others will have them incorrectly allocated, yielding garbage values.  And it's definitely a bundle of hacks.  But most of the time, it works.

(Note that whether or not it works depends less on the modifier being invoked, and more on what is passed for its arguments.  When I talk about whether it works for a given modifier or not, you should read that as actually being about whether it works for a given modifier invocation or not.)

(I think it's accurate to say that it works whenever none of the arguments are constant state variables, and the modifier isn't applied twice to the same function -- and I did a fair bit of work to make sure that even when some of the arguments are constant state variables it still works a lot of the time.  As for the same modifier being applied twice to the same function, well, that already causes problem with local variable allocation, so any problems here aren't really new.  In any case, if you want to know more about the other cases where it doesn't work, see the bottom.)

In addition, this PR removes most of the `ast` selectors, since they were mostly redundant with their existing duplicates in `solidity`.  I won't go into detail about that.

So, how's this thing work, basically?

Well, firstly, the `ModifierDefinition` case in the data saga is still there.  I've added an exception, though -- if the next node is a modifier invocation, and the modifier being invoked is the one whose definition node you're currently on, we skip assignment.  (Note this applies to the `FunctionDefinition` case as well -- after all, constructors have `FunctionDefinition` nodes, not `ModifierDefinition` nodes!)  This is because sometimes we hit the modifier definition node in the *middle* of setting up the modifier's arguments, and we don't want to include such intermediate garbage assignments.  (We will still sometimes get garbage assignments, unfortunately, but more on that later.)

But the data saga now has a "modifier preamble", which occurs before the type of node is checked.  (Although many types of nodes will ignore the results of this preamble; `FunctionDefinition` ignores it, for instance.)  This modifier preamble may make additional assignments if it concludes we're about to enter a modifier.  To reduce the number of actions dispatched, it doesn't actually make the assignments itself; rather it just stores them in a variable called `preambleAssignments` that will then be combined with any assignments made below before dispatching.  (To make such combination easier, I've altered the format of the `assignments` variable by stripping away the outer `byId:`.  I hope that's not unduly breaking parallelism.)

In order to determine whether we're about to jump into a modifier, the modifier preamble checks:

1. Are we currently at a node corresponding to an argument to a modifier invocation?  Or, if said argument is a type conversion, its argument (possibly nested beneath more type conversions)?
2. Is the next node none of: A modifier definition, a function definition, or inside the same invocation we're already in?

If both conditions hold, we conclude we're about to enter a modifier, and allocate its parameters.  (Note that since the next node might not actually be in the modifier, we have to look up the modifier node based on the invocation we're in.)  If we're on the *last* argument, then we can just do this like we would for a function definition, starting from `top` and working down.  If, however, we're jumping in from an earlier argument, we have to make an adjustment to `top` based on our current position to account for the arguments occurring after our current position.

In order to accomplish this I've had to add a number of selectors, primarily to `solidity` and `data`.  Let's discuss `solidity` first.  The selectors `solidity.current.instruction`, `solidity.current.source`, `solidity.current.sourceRange`, `solidity.current.pointer`, and `solidity.current.node` are now dynamic selectors, existing in both `current` and `next` versions.  Be warned, however, that the `next` versions are not always valid -- both `current` and `next` versions have to rely on `solidity.current.instructionAtProgramCounter`, so the `next` versions will not return valid results if you're on an instruction that makes an EVM call or return.  (I've added an `evm` selector, `evm.current.step.isContextChange` (well, and its `next` counterpart, as it's a dynamic step selector) that just `||`'s together `isCall`, `isCreate`, and `isHalting` to detect this.)

The bulk of this PR though is the new `data` selectors.  (Note that some of these do come in both `current` and `next` versions, but I opted not to make these dynamic, instead just factoring out common code into functions.)  Now really I think many of these selectors should really be in `solidity`, not `data`, since they deal with determining our current position in the AST.  However, `data` is where `data.info.scopes` is, and it's `data.info.scopes` (and its inlined counterpart) that allows us to easily navigate the AST.  I could move the whole `scopes` system into `solidity`, I suppose, but I didn't want to mess that here.

The big new selector is `data.current.aboutToModify`, which is what determines whether or not to invoke the modifier preamble.  It in turn relies on `data.current.modifierInvocation`, `data.next.modifierInvocation`, and `data.current.modifierArgumentIndex` (although that last one is incidental -- since I needed it anyway I included it to avoid some recomputation).

The selector `data.current.modifierInvocation` returns the node corresponding to the modifier invocation you're currently in; if none it returns the `SourceUnit` node by default.  Obviously, `data.next.modifierInvocation` is similar; however, like the other `data.next` selectors I've added, it also checks that you're not about to do a context change because the result would be invalid if you were.  (By the way, if you're wondering why `findAncestorOfType` is written to allow other lists of types, it's because originally there was also a selector `data.next.function` that determined what function the next node was in; this ended up getting removed though.  I actually thought about having `findAncestor` take a predicate that the node should satisfy, rather than a list of node types its node type should be among, but didn't see a need.)

The selector `data.current.modifierArgumentIndex` gets the index for which argument you're in, within the modifier invocation you're in (assuming, of course, you're in one at all).

Finally, the selector `data.current.modifierBeingInvoked` (as well as its `next` counterpart, which is altered similarly to `data.next.modifierInvocation`) finds the definition node for the modifier whose invocation you're currently in.  (Notice that this one works a bit differently depending on whether you're in a modifier invocation or a base constructor invocation.)  The data saga makes use of this; I'd suggest just seeing the code if you want to know how.

One odd thing: The `data` selector for getting the current node is known as `data.current.scope` rather than `data.current.node`.  I mean, I get why, but, it's really being used for more than just scopes now!  Maybe that should be renamed?  I left it alone.  (And I called its `next` counterpart that I added `data.next.scope` as well.)

So, that's the basic idea.  Let's discuss some of the weirder corners now -- and why I say it only works *most* of the time.

Basically, so long as every argument to the modifier is sourcemapped to, it works.  What isn't sourcemapped to?  String literals; constant state variables; and certain type conversions.

Type conversions aren't a problem (at least, not if their arguments aren't a problem); they just require the extra check I mentioned above.

String literals, surprisingly, are mostly not a problem.  While the string literal node itself will be skipped, when there's a string literal, the modifier definition node itself *is* hit.  So if the worst you have to worry about is string literals, there's no problem.

Unfortunately, constant state variables pose a problem.  For instance, if every argument is a constant state variable, both the argument list and definition node may be skipped over entirely, giving us nothing to work with.

A worse problem can arise from combining constant state variables and string literals (or anything else that causes the definition node to be hit) -- if one argument is a string literal, and all subsequent arguments are constant state variables, you're going to get wrong allocations with garbage values.  Unfortunately, I don't see any easy way to fix this problem at the moment.  I don't think that problem is bad enough that we shouldn't merge this, but it's definitely worth noting.

But, like I said -- as long as constant state variables aren't involved, this should work.  So we will now *mostly* support parameters for modifiers and base constructors.


packages/truffle-debugger/lib/data/selectors/index.js
```diff
@@ -510,7 +679,61 @@ const data = createSelectorTree({
...
-    }
+    },
+
+    //HACK WARNING
+    //the following selectors depend on solidity.next
+    //do not use them when the current instruction is a context change!
+
+    /**
+     * data.next.scope
+     * this should probably be named data.next.node!
```
gnidan: let's deprecate it, then! we don't have a `data.next.node`, right? We can just make that selector and use `data.next.scope` to be a deprecated alias.

haltman-at: Cool, I'll do that then!

haltman-at:
OK @gnidan, what I actually did is slightly different from your suggestion -- I renamed `data.next.scope` to `data.next.node`, renamed `data.current.scope` to `data.current.node`, and added a new `data.current.scope` which is just an alias for `data.current.node`, nothing that this use is deprecated.  (I didn't add `data.next.scope`.)

PR #1801, opened 2019-03-06
Change cache path

eggplantzzz:
This PR attempts in part to resolve a permissions issue that plagues many Truffle users. Some poor, unsuspecting folks have installed Truffle or npm using sudo and have been, as a result, experiencing issues with writing to the cache directory. This changes the cache directory to be something much less likely to experience permissions issues.

The new directory to be used is the same that the user-level config for mnemonics etc. uses, namely `~/.config/truffle/` in which a `compilers` directory will be located for cacheing compilers.

### DIFF
--- a/packages/truffle-config/index.js
+++ b/packages/truffle-config/index.js
@@ -390,4 +390,13 @@ Config.getUserConfig = function() {
   return new Configstore("truffle", {}, { globalConfigPath: true });
 };
 
+Config.getCacheDirectory = () => {
+  const configStore = new Configstore(
+    "truffle",
+    {},
+    { globalConfigPath: true }
+  );
+  return path.dirname(configStore.path);
+};
+
 module.exports = Config;
--- a/packages/truffle-config/index.js
+++ b/packages/truffle-config/index.js
@@ -390,7 +390,7 @@ Config.getUserConfig = function() {
   return new Configstore("truffle", {}, { globalConfigPath: true });
 };
 
-Config.getCacheDirectory = () => {
+Config.getTruffleDataDirectory = () => {
   const configStore = new Configstore(
     "truffle",
     {},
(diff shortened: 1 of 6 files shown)


packages/truffle-config/index.js
```diff
@@ -390,4 +390,13 @@ Config.getUserConfig = function() {
   return new Configstore("truffle", {}, { globalConfigPath: true });
 };
 
+Config.getCacheDirectory = () => {
```
gnidan: I think using the word `cache` here is a bit ambiguous. Maybe we could call this the "user config directory"? Or we could make this less general and just make the method `getCompilersCacheDirectory`. Thoughts?

eggplantzzz: It seems to me that user configuration might be a more specific use for the cache directory. In other words, Truffle has a directory that it uses to cache things. One of those things happens to be the user configuration data. And I would be totally fine with making it `Config.getCompilersCacheDirectory`. That way the loading strategies don't need to do any legwork and `Config` can worry about that stuff.

gnidan: My thinking here is predicated on my sense that the word "cache" implies something that can be deleted without losing information. E.g., the compilers are cached because they can be redownloaded. I would say that the user config isn't cached because if it's deleted, it can't be recreated automatically.

But I may be defining "cache" too specifically in my head, so I'd love a second opinion.

eggplantzzz: Oh I see what you mean. Maybe it should be something more like `truffleDataDirectory`. Maybe inside there can be a folder named something like `cachedCompilers`...how would that be?

gnidan: both those sound good!

eggplantzzz: Actually I guess `Config` shouldn't necessarily know about the compiler cacheing path...the loading strategy should handle that I think.

gnidan: I think Config should know about the "truffle data directory", but the supplier manages the compiler cache itself

PR #1802, opened 2019-03-07
Update truffle-artifactor links in README.md

linkdesu:


CruzMolina:
Thanks for this @linkdesu !

The entire README for `truffle-artifactor` needs a makeover, so thanks for pitchin in! 👍

PR #1806, opened 2019-03-10
Introduce the Network/Interface Adapter and Implement Basic Quorum Support

### LINKED ISSUE #1543: Accommodate broken support for Quorum/other chains via `--legacy` flags to use old code

seesemichaelj:
This PR replaces #1544 and fixes #1543.

#1544 was originally taken on by @CruzMolina to support issues with migrations for Quorum in Truffle v5. The interface to enable the work in #1544 were command-line arguments (`--legacy`, and implemented later on by myself `--quorum`) to the `truffle migrate` and `truffle test` commands.

After further discussion in our private Slack channels, we decided that a better interface would be letting users enable Quorum support via the `truffle(-config)?.js` config file in the `networks` section. For example:
```
...,
networks: {
  development: {
    type: "quorum" | ...,
    ...
  }
}
```

This interface allows for easier extension as well as a better user experience (by not having to specify a command-line flag constantly).

While the legacy migration support is still ingrained in other files, but network-specific fixes (i.e. how the output block formatter within web3.js needed to be monkey-patched to support the nanosecond timestamp from Quorum) are implemented in the `truffle-interface-adapter` package. This implementation is a crude/basic version, but it's overall intent is to abstract Truffle's interaction with networks/consortia/etc.

I closed #1544 for this PR as the intention of the fix had changed, however I have kept the entire commit history.


packages/truffle-interface-adapter/lib/web3-shim.js
```diff
@@ -3,30 +3,50 @@ const BN = require("bn.js");
 
 // This is a temporary shim to support the basic issues with Quorum
 
-class Web3Shim {
+class Web3Shim extends Web3 {
```
gnidan: Favor composition over inheritance, no? Why extend Web3?

packages/truffle-deployer/test/deployer.js
```diff
@@ -236,8 +241,8 @@ describe("Deployer (sync)", function() {
 
     utils.stopAutoMine();
 
-    const isLibrary = await IsLibrary.deployed();
-    const example = await Example.deployed();
+    await IsLibrary.deployed();
+    await Example.deployed();
```
gnidan: We can just remove these lines I think if we're not using the results.

packages/truffle-deployer/test/deployer.js
```diff
@@ -256,7 +261,7 @@ describe("Deployer (sync)", function() {
 
   it("emits block events while waiting for a tx to mine", async function() {
     this.timeout(15000);
-    const startBlock = await web3.eth.getBlockNumber();
+    await web3.eth.getBlockNumber();
```
gnidan: We don't need this either, right?

packages/truffle-deployer/test/errors.js
```diff
@@ -66,12 +66,17 @@ describe("Error cases", function() {
         Abstract,
         Loops
       ],
+      networks: {
+        test: {
+          // TODO:
```
gnidan: (TODO what?)

packages/truffle-interface-adapter/lib/web3-shim.js
```diff
@@ -3,30 +3,50 @@ const BN = require("bn.js");
 
 // This is a temporary shim to support the basic issues with Quorum
```
gnidan: (Risky move, calling a software component temporary!)

gnidan: I like to sign and date these sorts of comments so I can save someone the trouble of `git blame`ing me in the future.

packages/truffle-interface-adapter/lib/web3-shim.js
seesemichaelj: This is a Web3 shim. It's purpose is to shim around web3. So for consumers of it, it should look "more or less" like web3. This is all "temporary" until a better system is architected.

The inheritance was driven by stuff I'm doing for `axcore` integration, but I recently discovered that the benefits of inheritance don't translate with `Web3`'s implementation. It's not worth changing it back now.

packages/truffle-deployer/test/deployer.js
seesemichaelj: Alright, I usually do "as minimum changes as possible" and not "drive by cleanup", but I can remove these if you really want me to.

packages/truffle-deployer/test/deployer.js
seesemichaelj: Doesn't look like it

packages/truffle-deployer/test/errors.js
seesemichaelj: It looks like I forgot to remove the TODO; it's no longer necessary to add more details. I put the colon there as it highlights it in my vscode. I'll remove the todo

packages/truffle-deployer/test/deployer.js
seesemichaelj: The only reason these were altered to begin with is because of the rolling linting changes required everytime I modify a file that hasn't been through the linting process yet

packages/truffle-interface-adapter/lib/web3-shim.js
seesemichaelj: Well if this isn't temporary, then we need to revisit our roadmap and expectations for the next 4 months lol.

Git blaming takes a couple of minutes, but sure, I can date it.

packages/truffle-deployer/test/deployer.js
seesemichaelj: I'm going to go ahead and do the drive-by cleanup to not have to wait for another cycle of discussion. I'm usually not fond of this, but will make an exception.

packages/truffle-deployer/test/deployer.js
seesemichaelj: I'm going to go ahead and do the drive-by cleanup to not have to wait for another cycle of discussion. I'm usually not fond of this, but will make an exception.

packages/truffle-interface-adapter/lib/web3-shim.js
gnidan: Wait so we're planning on replacing this entirely? I thought we were going to continue morphing it to meet our needs.

seesemichaelj: Much of this code will be reused but the morphing will look like something entirety different after a better architecture is designed. The whole point of doing this PR now instead of 1-2 months from now was to get something out the door.

seesemichaelj: Further discussion happened over Zoom, and I added a comment (https://github.com/trufflesuite/truffle/pull/1806/commits/4457aa67a4692645ab34f389442a96bd21d684d5) to better explain the nature of the Web3Shim in the event that it becomes less temporary than planned.

PR #1807, opened 2019-03-10
Reduce ABI schema's type restrictiveness

gnidan:
Ref: #1774

- Allow matching array signifiers at end of `bytes` and `string` types
- Change all array matchers from `?` to `*`, to match any number repeatedly


gnidan:
Thanks @CruzMolina!

PR #1808, opened 2019-03-11
rework regex handling for errors sent to analytics to account for win…

fainashalts:
…dows file paths


fainashalts:
@cruz I had to spoof the directory name to test with, but I’m 99% sure this fix will do it!

PR #1809, opened 2019-03-12
Support Hexstrings and BNs for Gas to Support Quorum's Potential Large Block Gas Limits

seesemichaelj:
This extends on #1806 by adding support for Quorum's ability to have 54+ bit block gas limits.

This is blocked by merging #1806 just because it's based off those commits, so the changes here are going to look like more than they actually are. **You can look at the compare here in the meantime**: https://github.com/trufflesuite/truffle/compare/adapter/init-and-quorum...adapter/large-gas

### DIFF
--- a/packages/truffle-contract/lib/execute.js
+++ b/packages/truffle-contract/lib/execute.js
@@ -7,6 +7,16 @@ var Reason = require("./reason");
 var handlers = require("./handlers");
 var override = require("./override");
 var reformat = require("./reformat");
+var BN = require("bn.js");
+
+function createBN(n) {
+  if (typeof n === "string" && n.startsWith("0x")) {
+    n = n.slice(2);
+    return new BN(n, "hex");
+  } else {
+    return new BN(n);
+  }
+}
 
 var execute = {
   // -----------------------------------  Helpers --------------------------------------------------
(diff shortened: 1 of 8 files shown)


packages/truffle-contract/lib/execute.js
```diff
@@ -7,6 +7,16 @@ var Reason = require("./reason");
 var handlers = require("./handlers");
 var override = require("./override");
 var reformat = require("./reformat");
+var BN = require("bn.js");
+
+function createBN(n) {
```
CruzMolina: I think `createBN` should be moved into `truffle-contract/lib/utils.js`
(would also remove the redundancy of declaring the func twice)

seesemichaelj: Ah good catch. I wasn't even thinking that they were in the same package :sweat_smile:

packages/truffle-contract/lib/execute.js
```diff
@@ -1,6 +1,7 @@
 const debug = require("debug")("contract:execute"); // eslint-disable-line no-unused-vars
 var Web3PromiEvent = require("web3-core-promievent");
 var EventEmitter = require("events");
+var BigNumber = require("bignumber.js");
```
gnidan: Why bignumber.js instead of BN.js? We really should avoid multiple high-precision number libraries.

I know truffle-contract already has the bignumber.js dependency, but that is only for compatibility reasons, to enable `MyContract.numberFormat` to give BigNumber return values.

seesemichaelj: It was originally `bn.js`, but `constructor.gasMultiple` is/can be a decimal value and `bn.js` does not support decimals anywhere (even as an argument to `.muln()` for example). I could not think of an elegant way to turn an arbitrary decimal into a single multiply/division of two numbers.

gnidan: A hacky way to do it might be:

```javascript
let decimal; // decimal to convert
let maxPrecision; // upperbound on significant figures
let numerator;
let denominator;
const sigfigs = Math.min(
  decimal.toString().length - 1, // length less one because `.`
  maxPrecision
);

denominator = Math.pow(10, sigfigs);
numerator = decimal * denominator;
```

I strongly urge you to hack your way away from using bignumber.js for this case. Don't let decimals stop you when there are large powers of ten to be wielded.

seesemichaelj: I definitely find this hack gross, but I will implement it.

Issue #1597, opened 2018-12-29
[Truffle-Decoder] Allow option to provide contract address


adrianmcli:
For my tests, I use `truffle-compile` to compile my contract (instead of using migrations) and the resulting artifact file does not have a `networks` property (and neither does compiling with `solc-js`). Instead of migrations, I use Web3.js to deploy my contract onto a Ganache instance running in-memory.

This means that [line 134](https://github.com/trufflesuite/truffle/blob/next/packages/truffle-decoder/lib/interface/contract-decoder.ts#L134) of `contract-decoder.ts` will fail:

```js
this.contractNetwork = Object.keys(this.contract.networks)[0];
```

The next line sets the contract address based on this:

```js
this.contractAddress = this.contract.networks[this.contractNetwork].address;
```

Not everyone uses truffle migrations to deploy, but we should still allow them to use the decoder as well, I would think?

adrianmcli:
Do note that line 202 also has this issue: https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-decoder/lib/interface/contract-decoder.ts#L202

seesemichaelj:
Moved this to `new issues` from `icebox` because i think this was created when the `icebox` column in zenhub got moved to the first column

adrianmcli:
Thanks @seesemichaelj!

adrianmcli:
Another reason why this PR would be super helpful for adoption is because allowing the user to pass in a deployed contract address means that you can decode and track state for contracts that you may not own (i.e. publicly available contracts with an open ABI).

seesemichaelj:
The decoder needs more than just an open ABI; we need the compiler output which implies open source code.

I do think it's a good idea though!

adrianmcli:
Ah right, the AST as well!

haltman-at:
Hm, so, wait, if we ignore the part about being able to decode contracts you don't have artifacts for, the original request should be possible, right?  Like if I added an interface for it, so long as Ganache had the artifact it would be able to supply the artifact and address separately?  (Does Ganache still only track state for contracts it can decode?  Not sure if tracking state for undecoded contracts would have to come first.)

adrianmcli:
@haltman-at thanks for taking a look at this.

[...]

PR #1812, opened 2019-03-13
Add address parameter to contract decoder

haltman-at:
This PR addresses issue #1597, allowing an address to be provided manually (as a string) in the contract decoder.  (The particular format of the address is, anything that's good enough for Web3; we don't process this address at all.)  Note that we do not create a whole `networks` member of the contract object, we just manually set the address -- that should be good enough for our purposes.

In addition, while doing this I took a moment to eliminate the now-unused `variables` field from the `EvmInfo` type.  This should not be a functional change.

### DIFF
--- a/packages/truffle-decoder/lib/interface/contract-decoder.ts
+++ b/packages/truffle-decoder/lib/interface/contract-decoder.ts
@@ -113,7 +113,7 @@ export default class TruffleContractDecoder extends AsyncEventEmitter {
 
   private mappingKeys: Slot[] = [];
 
-  constructor(contract: ContractObject, relevantContracts: ContractObject[], provider: Provider) {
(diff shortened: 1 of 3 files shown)


packages/truffle-decoder/lib/interface/contract-decoder.ts
```diff
@@ -113,7 +113,7 @@ export default class TruffleContractDecoder extends AsyncEventEmitter {
 
   private mappingKeys: Slot[] = [];
 
-  constructor(contract: ContractObject, relevantContracts: ContractObject[], provider: Provider) {
+  constructor(contract: ContractObject, relevantContracts: ContractObject[], provider: Provider, address: string) {
```
seesemichaelj: this should be `address?: string`

seesemichaelj: Nevermind, I see you added the `?` in the public interface

PR #1814, opened 2019-03-13
Remove dummy address system in favor of storage address system

haltman-at:
This PR removes the infamous dummy address system I added a while back, replacing it with a storage address system.  It also removes that annoying test of the dummy address system that keeps timing out!  (The PR does not include a replacement for that test.)

Now, instead of dummy addresses, each entry on the callstack now contains an additional field, `storageAddress`.  While `address` (if defined) gives the address of the code that's being executed, `storageAddress` gives the address whose storage is being used.  For an ordinary call, of course, this is equal to `address`, but obviously it can differ for a `DELEGATECALL`, and of course for a creation call there is no `address`, and `storageAddress` is equal to the address of the contract that will be created.

Be aware, the selector `data.current.address` has not been renamed, but it now returns the `storageAddress` (which is always defined), not the `address`.  The selector `data.current.dummyAddress` is gone, along with `evm.current.creationDepth`, all uses of `dummyAddress`, and all the machinery for learning addresses.

So, where does the `storageAddress` come from?  For an ordinary call, of course, it's the same as the address, while for a `DELEGATECALL` it copies the `storageAddress` from the previous stackframe.  But for a creation call we have to determine the address where the contract will be created.

Now, the right way to do this would be to compute it from the nonce and such, but unfortunately we have no way to get the nonce.  (We *could* compute the address for a `CREATE2`, but I think it's simpler for now to use a unified approach.)  So, we look ahead to when the call returns and get the address off the stack.  (If the creation call is the initial call of the transaction, we instead get it from the `contractAddress` field of the receipt.)  Note that because the trace includes a `depth` field, we can do this by just looking for the next step of the same depth.  To do this, I've added a new trace selector, `trace.nextOfSameDepth`, and a new evm state selector, `evm.nextOfSameDepth.state`.  (I didn't bother with step selectors for it.)  There is of course also a new evm selector to get the address, `evm.current.step.createdAddress`, but note that this one is not dynamic and only exists for `current`, not `next`.

By the way, the `storageAddress` system fixes the "regression" introduced in #1734, since mapping keys are now stored by storage address, rather than by code address; although this still remains mostly meaningless as we still mostly don't keep track of storage across EVM calls.

There is one notable issue in this PR that will have to be addressed eventually, and that's the problem of failed creation calls.  A failed creation call puts `0` on the stack instead of the address, so for a failed creation call our `storageAddress` will be the zero address rather than the correct address.  The only problem this causes at the *moment* is that mapping keys for distinct failed creation calls will get mixed together.  However it will cause more problems in the future when I add globally available variables and yet more when we start keeping track of storage ourselves.  There are various ways we could work around the problem, but I figured the best thing for now is to leave it as-is, and address the problem later, either when we can do so without workarounds, or when we're at the point where it's causing serious enough problems that we *have* to work around it.

Note also that this is only *necessarily* a problem when the failed creation call was a call from inside the transaction; if the failed creation call was the initial call of the transaction, we may get from Web3 what the address would have been.  That's what happens in testing with Ganache anyway; I don't know to what extent EVM implementations may differ here, so that's something we should possibly be aware of.


haltman-at:
Yeah, that was what I was originally going to do, but then I was like:

1. I don't want to rename things if I don't have to, especially something that's used a lot, and
2. The storage address is really the relevant address in a data context, not the code address, so I figured it was fine (like there's no selector for the code address for it to be confused with).

IDK.  I can change it if you want, but I think it's fine as-is.

(Also I disagree that using storage addresses is cleveer, I think it's just the right thing here. :) )

gnidan:
Your argument is valid. Leave as-is!

Issue #1817, opened 2019-03-14
Truffle Test Not Working With Solidity v0.4.24


timbeiko:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue
Quoted from a [StackOverflow post I opened](https://ethereum.stackexchange.com/questions/68223/truffle-test-issue-with-address-payable-and-compiler-version), and posted here after being asked by @gnidan on Gitter.

> I am trying to run truffle test on an old contract which was written using pragma solidity 0.4.24;. While I am able to deploy the contracts using truffle migrate, I get the following error when running truffle test:

```
Using network 'development'.


Compiling your contracts...
===========================
Error: CompileError: Error parsing truffle/DeployedAddresses.sol: ParsedContract.sol:4:55: ParserError: Expected ',' but got 'payable'
  function Marketplace() public pure returns (address payable) { return 0x6395927945EA39aDD6B1E04144f1e1324B30DeFC; }
                                                      ^-----^
Compilation failed. See above.
    at async.whilst.error (/usr/local/lib/node_modules/truffle/build/webpack:/packages/truffle-compile/profiler.js:369:1)
    at /usr/local/lib/node_modules/truffle/build/webpack:/~/async/dist/async.js:969:1
    at next (/usr/local/lib/node_modules/truffle/build/webpack:/~/async/dist/async.js:5222:1)
    at Promise.all.then.results (/usr/local/lib/node_modules/truffle/build/webpack:/packages/truffle-compile/profiler.js:351:1)
    at <anonymous>
Truffle v5.0.7 (core: 5.0.7)
Node v8.12.0
```
> `truffle/DeployedAddresses.sol: ParsedContract.sol:4:55` is not code I've written, nor is it part of my project. I suspect this error may be caused because Truffle is running solidity 0.4.24 to compile my contracts but that whatever it uses to execute internal tests is running solidity 0.5. Is there a way to verify/change this?

> Here are the contents of `truffle.js` in the main directory:

```
module.exports = {
  migrations_directory: "./migrations",
  networks: {
    development: {
      host: "localhost",
      port: 8545,
      network_id: "*" // Match any network id
    }, 
 // ... Other configs specified here, removed because irrelevant
 // to SO question ... 
 compilers: {
  solc: {
    version: "0.4.24",  
    }
  },
  solc: {
    optimizer: {
      enabled: true,
      runs: 500
    }
  } 
};
```
[...]

PR #1820, opened 2019-03-14
Enable backwards-compatible 0.4.xx solc unit testing

CruzMolina:
Truffle v5's release & default support for solidity 0.5.0 introduced a number of breaking changes & necessary quick-response stop gap measures.

This PR addresses one of those stop gaps (#1817), allowing backwards-compatible solidity unit testing for users still developing with `0.4.xx`.

🍻

### DIFF
--- a/packages/truffle-core/lib/testing/deployed.js
+++ b/packages/truffle-core/lib/testing/deployed.js
@@ -1,10 +1,11 @@
 // Using web3 for its sha function...
 var Web3 = require("web3");
+const semver = require("semver");
 
 var Deployed = {
   makeSolidityDeployedAddressesLibrary: function(
     mapping,
-    removeAddressPayable
+    { solc: { version } }
   ) {
     var self = this;
 
@@ -35,8 +36,10 @@ var Deployed = {
 
     source += "}";
 
-    // if true, strip out payable!
-    if (removeAddressPayable) source = source.replace(/ payable/gm, "");
+    if (version) {
+      if (semver.lt(semver.coerce(version), "0.5.0"))
+        source = source.replace(/ payable/gm, "");
+    }
 
     return source;
   },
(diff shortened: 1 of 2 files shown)


packages/truffle-core/lib/testing/deployed.js
```diff
@@ -2,7 +2,10 @@
 var Web3 = require("web3");
 
 var Deployed = {
-  makeSolidityDeployedAddressesLibrary: function(mapping) {
+  makeSolidityDeployedAddressesLibrary: function(
+    mapping,
+    removeAddressPayable
```
gnidan: Could we make this new second argument an "`options`" and use the whole solc compiler config for input?

Something like:
```javascript
makeSolidityDeployedAddressesLibrary: function (
  mapping,
  { solc: { version } }
) { /* ... */ }
```?

This way it can just be passed wholesale, and `testsource.js` wouldn't have to know what specific information it is that `deployed.js` needs.

If future versions of Solidity affect this file, the change will be self-contained here. And we don't really have a need to generalize what's different in different versions, so there's no need for it to be outside DeployedAddresses for the time being.

CruzMolina: Sounds good! 😎

timbeiko:
Thanks for the rapid fix 🎉!

CruzMolina:
@timbeiko you should be able to test this by:

- `git clone`ing the `develop` branch
- `cd`ing into the cloned `truffle` dir
- running `yarn bootstrap` (if you don't have yarn, then `npm i -g yarn` first)
- `cd`ing into `packages/truffle/`
- running `yarn build`
- aliasing the bundle in the terminal window you wish to test truffle via `alias truffle='node /Users/cruzmolina/myPath/toTruffleRoot/truffle/packages/truffle/build/cli.bundled.js'`

🌮

PR #1822, opened 2019-03-15
Track storage across intra-transaction call boundaries

haltman-at:
This PR gives the debugger the ability to track storage across call boundaries.  The debugger still can't track storage across transaction boundaries; storage that hasn't been touched (i.e., read from or written to) within the current transaction will still show up as 0.  However, now, if one call within the transaction touches a storage slot, all calls within that transaction sharing that storage address will be able to see it.

This PR is the start of what I'm calling the "codex" -- an effort to have the debugger keep track of the state itself, instead of being totally helpless when the trace won't tell us things directly.  (Arguably, keeping track of calldata was also part of this.)  However, right now we only really need to do this for storage, so, I'm not introducing anything else into the codex yet.  That can wait until we need it.

So, now, the `evm` state contains `evm.proc.codex`.  While it will do more in the future, right now it keeps track of the current storage at each address.  It starts out empty; whenever there's a `CALL` or `CREATE` action, it adds a "page" to the codex for the call's storage address.  (If necessary; we don't add pages already there.  Also, there is no zero page; see below for more on this special case.)  The new addition is that the EVM saga now also looks for `SLOAD` and `SSTORE` instructions, and on finding one, issues a `STORE` action that instructs the codex to update its copy of storage.  This action has three parameters: `address`, `slot`, and `value`.  The slot, of course, is read off of the stack, and the value is read from `evm.next.state.storage` based on the slot.  (For an `SSTORE` the value could of course also be read off the stack, but I thought it was simpler to take a unified approach.)

So, some new selectors have been added.  There's now `evm.current.codex`, which returns the current codex; and there's `evm.current.codex.storage`, which does not return the full storage portion of the codex (that's the only portion at the moment), but rather the storage for the current storage address *based* on the codex.

There is a special case to be aware of -- as you know, currently, failed creation calls all get tagged with a storage address of zero (unless it was the initial call of the transaction).  So, if we didn't account for that, and there were multiple failed creation calls, the zero page of the codex would contain a lot of garbage, and attempting to read storage during a failed creation call would be a total crapshoot.  So instead, we don't maintain a zero page (responsibility for this is split between the reducer, which doesn't add a zero page on `CALL` or `CREATE`, and the saga, which doesn't issue a `STORE` action if the current storage address is zero).  While inside a failed creation call, `evm.current.codex.storage` does not look at the codex, but rather just returns `evm.current.state.storage`.

This is something of a hack, but note that we won't be missing anything this way -- after all, if we're dealing with a failed creation call, there's no other call (even in another transaction) that could have altered this storage, right?  So just using `evm.current.state.storage` is good enough.

(Note that mapping keys for distinct failed creation calls are still not tracked separately -- those will get mixed together.  I don't really consider that a big problem, though.)

Meanwhile, over in `data`, the selector `data.current.state.storage` is now based on `evm.current.codex.storage` rather than `evm.current.state.storage`.  Originally the plan for the codex was a much bigger reorganization, but I'm doing this the lazy way for now.

Finally, in the process of this I had to make a slight alteration to how the reset button works in the `evm` submodule.  You see, the default codex is empty; it relies on that initial `CALL` or `CREATE` to have its first page added.  But right now resetting doesn't work by fully resetting the debugger, so resetting the codex to default will result in that initial page being missing.  So now, the `reset` saga in EVM first reads the initial storage address off the callstack, and the reset action takes this as an argument, and sets up a default codex with that one page already there.  A little bit of a hack, but again, it's not like there's some case it's missing.

I also added a quick test of this new functionality.  There's still much to be done, but missing values in storage should be substantially less of a problem now!

### DIFF
--- a/packages/truffle-debugger/lib/data/selectors/index.js
+++ b/packages/truffle-debugger/lib/data/selectors/index.js
@@ -320,7 +320,7 @@ const data = createSelectorTree({
        * data.current.state.storage
        */
       storage: createLeaf(
-        [evm.current.state.storage],
+        [evm.current.codex.storage],
 
         mapping =>
           Object.assign(
--- a/packages/truffle-debugger/lib/evm/actions/index.js
+++ b/packages/truffle-debugger/lib/evm/actions/index.js
@@ -53,6 +53,16 @@ export function returnCall() {
   };
 }
 
+export const STORE = "STORE";
+export function store(address, slot, value) {
+  return {
+    type: STORE,
+    address,
+    slot,
+    value
+  };
+}
+
 export const RESET = "EVM_RESET";
 export function reset() {
   return { type: RESET };
--- a/packages/truffle-debugger/lib/evm/actions/index.js
+++ b/packages/truffle-debugger/lib/evm/actions/index.js
@@ -64,6 +64,9 @@ export function store(address, slot, value) {
 }
 
 export const RESET = "EVM_RESET";
-export function reset() {
-  return { type: RESET };
+export function reset(storageAddress) {
+  return {
+    type: RESET,
+    storageAddress
+  };
 }
--- a/packages/truffle-debugger/lib/evm/reducers.js
+++ b/packages/truffle-debugger/lib/evm/reducers.js
@@ -177,6 +177,7 @@ export function codex(state = DEFAULT_CODEX, action) {
         return state;
       }
       return {
+        ...state,
         byAddress: {
           ...state.byAddress,
           [action.storageAddress]: {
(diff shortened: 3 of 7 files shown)


packages/truffle-debugger/lib/data/selectors/index.js
```diff
@@ -320,7 +320,7 @@ const data = createSelectorTree({
        * data.current.state.storage
        */
       storage: createLeaf(
-        [evm.current.state.storage],
+        [evm.current.codex.storage],
```
gnidan: I like this name

Issue #1824, opened 2019-03-15
network_id: '*' incompatible with network 'develop'


ErikAbele:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

When explicitly defining a network named 'develop' (e.g. to be able to use `console --network develop` or `exec ... --network develop`) it is no longer possible to use network_id: '*' with this network. This was working fine up until 5.0.4 but stopped somewhere with a later version (I've upgraded to 5.0.7 directly so I am not exactly sure which version introduced this behaviour).

Now, to be able to use a 'develop' network, a numerical network_id has to be specified, e.g. '5777'.

While this is not a major issue per se, it points to some 'unclean' coding and probably shouldn't error out in the way it does (see below).

## Steps to Reproduce

Add a network named 'develop' to truffle-config.js and specify network_id: '*' with it:

```
	networks: {
		'develop': {
			host: '127.0.0.1',
			port: 9545,
			network_id: '*'
		}
	}
```
## Expected Behavior

I'd expect `truffle develop` to start without an error but it exits with the following:

## Actual Results

```
Error: Error: [number-to-bn] while converting number "*" to BN.js instance, error: invalid number value. Value must be an integer, hex string, BN or BigNumber instance. Note, decimals are not supported. Given value: "*"
    at toBN (/usr/local/lib/node_modules/truffle/build/webpack:/~/web3-utils/src/utils.js:64:1)
    at Method.hexToNumber [as outputFormatter] (/usr/local/lib/node_modules/truffle/build/webpack:/~/web3-utils/src/utils.js:234:1)
    at Method.formatOutput (/usr/local/lib/node_modules/truffle/build/webpack:/~/web3-eth/~/web3-core-method/src/index.js:163:1)
    at sendTxCallback (/usr/local/lib/node_modules/truffle/build/webpack:/~/web3-eth/~/web3-core-method/src/index.js:473:1)
    at /usr/local/lib/node_modules/truffle/build/webpack:/~/web3-eth/~/web3-core-requestmanager/src/index.js:147:1
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/truffle-provider/wrapper.js:112:1
    at XMLHttpRequest.request.onreadystatechange (/usr/local/lib/node_modules/truffle/build/webpack:/~/web3/~/web3-providers-http/src/index.js:96:1)
[...]

PR #1825, opened 2019-03-15
Handle non-integer `truffle develop` network_id values

CruzMolina:
This PR is a fix for #1824 .

Since `truffle develop` is now configurable, we need to sanitize & handle invalid `network_id` values.

☕️

### DIFF
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -84,8 +84,7 @@ const command = {
       ganacheOptions["hardfork"] = customConfig.hardfork;
     }
 
-    function sanitizeNetworkID() {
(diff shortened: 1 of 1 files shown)


packages/truffle-core/lib/commands/develop.js
```diff
@@ -84,6 +84,23 @@ const command = {
...
+            `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
+          throw new Error(error);
+        }
+      } else {
+        // We have a "*" network. Replace it w/ the default.
+        ganacheOptions.network_id = 5777;
+      }
+    }
+
+    sanitizeNetworkID();
```
gnidan: Could we make this take and return a value? It looks weird, we're not writing BASIC anymore :D

PR #1826, opened 2019-03-15
Optimize travis

CruzMolina:
Remove redundant docker start & cache node_modules.

Appears to save us ~1 min on travis jobs. 🤞 

Would have also cached yarn but couldn't figure out a quick way to have `lerna boostrap` use cached yarn instead of travis' older yucky yarn.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,6 +35,11 @@ install:
   - npm install -g yarn
   - npm run bootstrap
 
+cache:
+  directories:
+    - node_modules
+    - $(npm config get prefix)/lib/node_modules # globally installed stuff (i.e. lerna)
+
 script:
   - npm run ci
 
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -2,9 +2,6 @@ dist: trusty
 sudo: required
 language: node_js
 
-services:
-  - docker
-
```
gnidan: This isn't necessary? I thought it was necessary because we do `docker pull` and such.

CruzMolina: I thought so too!

Check out one of the jobs from the latest develop build: https://travis-ci.org/trufflesuite/truffle/jobs/507968192#L415

CruzMolina: Ah, looks like it's because our builds use `trusty` w/ `sudo: required`.

PR #1830, opened 2019-03-18
Print watch expressions with their values rather than jumbled

haltman-at:
Right now, if you have multiple watch expressions, the order of printing is jumbled so that watch expressions don't necessary appear with their corresponding values.  This is because the printing is done with the parallel `Promise.all`.  This PR replaces this with a simple `for` loop so that things print out in order and you can tell which values go with which watch expressions.


haltman-at:
Actually, hold on.  This possibly has problems I didn't realize.  Still investigating.

haltman-at:
OK can't reproduce the problem I encountered, disregard previous comment.  It was likely unrelated anyway.

Issue #1813, opened 2019-03-13
compiler crash after updating to 5.0.8


mscherer82:
## Issue

`truffle compile` causes a memory error. In version `5.0.7` compiling works fine.

## Actual Results

```
λ truffle compile

Compiling your contracts...
===========================

<--- Last few GCs --->

[14948:000001CE41908F60]    77537 ms: Mark-sweep 1693.0 (1760.7) -> 1692.9 (1729.7) MB, 1169.6 / 0.0 ms  last resort GC in old space requested
[14948:000001CE41908F60]    78715 ms: Mark-sweep 1692.9 (1729.7) -> 1692.9 (1729.7) MB, 1178.1 / 0.0 ms  last resort GC in old space requested


<--- JS stacktrace --->

==== JS stack trace =========================================

Security context: 000001EBEFB25879 <JSObject>
    1: next [C:\Projects\everdragons\everdragons-contract\node_modules\babylon\lib\index.js:~682] [pc=0000000B3126C3FF](this=000003453A6DECC1 <Parser map = 000002D5FFC88559>)
    2: /* anonymous */ [C:\Projects\everdragons\everdragons-contract\node_modules\babylon\lib\index.js:~3507] [pc=0000000B3128FB00](this=000003453A6DECC1 <Parser map = 000002D5FFC88559>,base=000002E9072D4C79 <Node map = 0...

FATAL ERROR: CALL_AND_RETRY_LAST Allocation failed - JavaScript heap out of memory
 1: node_module_register
 2: v8::internal::FatalProcessOutOfMemory
 3: v8::internal::FatalProcessOutOfMemory
 4: v8::internal::Factory::NewUninitializedFixedArray
 5: v8::internal::WasmDebugInfo::SetupForTesting
 6: v8::internal::interpreter::BytecodeArrayRandomIterator::UpdateOffsetFromIndex
 7: 0000000B310843C1
```

## Environment

* Operating System: Windows 10
* Truffle version (`truffle version`): 5.0.8
* node version (`node --version`): 8.12.0

ccolorado:
Was getting ready to submit this test:
Steps to reproduce
I created a blank virtual machine with ubuntu ubuntu-18.04.2-live-server-amd64.iso
install nvm 
nvm install v10.15.0
nvm alias default v10.15.0
npm install -g truffle
npm install -g ganache-cli
clone any solidity truffle project
npm install
truffle test
Get this result

[...]

PR #1834, opened 2019-03-19
Load @babel/register in advance of soljson

gnidan:
Ref: #1813 

To specify that soljson should be ignored by babel

(Alternate hypothesis to #1829)


gnidan:
Oooh this is failing in a useful way!

PR #1838, opened 2019-03-19
Document `truffle migrate` --to flag

CruzMolina:
Documents the `--to` flag for running migration files up to a specific number.

📝

### DIFF
--- a/packages/truffle-core/lib/commands/migrate.js
+++ b/packages/truffle-core/lib/commands/migrate.js
@@ -20,6 +20,10 @@ var command = {
       describe: "Specify a migration number to run from",
       type: "number"
     },
+    "to": {
+      describe: "Specify a migration number to run to",
+      type: "number"
+    },
     "interactive": {
       describe: "Manually authorize deployments after seeing a preview",
       type: "boolean",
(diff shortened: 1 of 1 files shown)


packages/truffle-core/lib/commands/migrate.js
```diff
@@ -28,7 +32,7 @@ var command = {
   },
   help: {
     usage:
-      "truffle migrate [--reset] [-f <number>] [--network <name>] [--compile-all] [--verbose-rpc] [--interactive]",
+      "truffle migrate [--reset] [--f <number>] [--to <number>] [--network <name>] [--compile-all] [--verbose-rpc] [--interactive]",
```
gnidan: Is it really `--f` and not `--from`?

gnidan: Could we make it both `-f` and `--from`?

CruzMolina: Wow, `--from` works, but I can't find it specified anywhere in the codebase.

CruzMolina: Oh! Found it. Hidden in `truffle-migrate/index.js` https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-migrate/index.js#L63

CruzMolina: although... I don't think that line does what it's intended to do 😂

CruzMolina: Okay, nm. Forget what I said above, `--from` doesn't work as a cmd line `truffle migrate` option.

PR #1841, opened 2019-03-20
Add & document --from flag to `truffle migrate`

CruzMolina:
Adds & documents a `--from` flag for running migration files from a specific number.

🎏


eggplantzzz:
Should we deprecate the `--f` flag with this PR?

CruzMolina:
@eggplantzzz Since we've add `--to`, makes sense to use `--from` instead of `--f`, but since it would be a breaking change, I suppose it's not something we can do until truffle v6.

PR #1842, opened 2019-03-20
Implement Basic AxCore Support in the Interface Adapter

seesemichaelj:
This PR adds AxCore support to the `truffle-interface-adapter`

~Along the way of implementing this, it made sense to move the `truffle-interface-adapter` to TypeScript, which is included in this PR.~ See #1939 

Some background of what was required to do this implementation:
AxCore requires two extra parameters to be passed into the JSON RPC calls. The actual names of these parameters are under NDA, so obfuscated names have been used instead (`param1` and `param2`). These parameters are defined at deploy time when calling `deployer.deploy(Constructor, ...constructorArgs)` or `Constructor.new(...constructorArgs)`. I have implemented this by providing and options object as the first argument prior to constructor arguments:
```
deployer.deploy(Constructor, {param1, param2}, ...constructorArgs)
or
Constructor.new({param1, param2}, ...constructorArgs)
```
This can be easily shifted to being the _last argument_, but I thought it made more sense to be the first argument.

I then take these parameters and cleverly (if I might say so myself) insert them into the payload in a `provider.send()` overload before it gets sent to the original `provider.send()`

It should be noted that `param1` and `param2` are to be inserted as part of the JSON object provided in the method. However if the method is `eth_getTransactionReceipt`, which requires 1 argument being the transaction hash, the AxCore API expects the payload to have `params: [txHash, param1, param2]`

Please don't do any rebasing/merging from `develop`. I will take care of this after more testing has been done on the current code base.


seesemichaelj:
Thanks! Seems to be resolved with the latest commit. Going to wait on merging this until I hear back from Axoni about whether or not it works for them.

CruzMolina:
Looks like we won't be needing this for the time being. Closing this for now.

Issue #1835, opened 2019-03-19
Truffle Compile re-downloads solc even after running Truffle Obtain


adrianmcli:
## Issue

Truffle Compile re-downloads solc even after running Truffle Obtain. Do note that we use `truffle` to run `truffle obtain`.

But in our tests we use `truffle-compile` by itself (as in we did `npm install truffle-compile` and used it with `require("truffle-compile")`).

## Steps to Reproduce

This is the .travis.yml used [here](https://github.com/trufflesuite/drizzle-utils/blob/master/.travis.yml):

```yml
language: node_js
node_js:
  - "node"
  - "--lts"
cache:
  yarn: true
  directories:
    - node_modules
before_script:
  - "truffle obtain --solc=0.5.2"
```

And this is the `truffle-compile` usage: https://github.com/trufflesuite/drizzle-utils/blob/master/packages/test-chain/compile.js#L3

## Expected Behavior

Since `truffle obtain` is already run, it shouldn't try to download the solidity compiler again.

## Actual Results

Check out this build:

https://travis-ci.org/trufflesuite/drizzle-utils/jobs/508220795

Note that on line 468, it already downloaded the compiler. But on line 539 it tries to do it again. It's able to fetch it in time for this build, but not in other builds.

CruzMolina:
Looking into this, I wonder if your `truffle-compile` dependency needs to be updated...

CruzMolina:
So, I updated `truffle-compile` to latest (4.0.8) on the `update-truffle-compile` branch in drizzle-utils, and I'm not sure whether it fixed the problem (the re-downloading output went away) or caused another one as the build seems to completely stall and then timeout.

https://travis-ci.org/trufflesuite/drizzle-utils/builds/508644454

CruzMolina:
Oh... I _think_ `truffle-compile` comes loaded w/ solc 0.5.0.

An earlier version of `truffle-compile` was wonky in that it either didn't cache solc properly or it continued to try and download a specified version of solc anyway (even if it was already cached).

I wonder if the fix for that compartmentalized `truffle-compile` so that by requiring and using it in the tests, there's some error somewhere being thrown, but unhandled from `truffle-compile` trying to use solc 0.5.0. Could be why the build as well as testing locally seems to hang. 🤔

adrianmcli:
@CruzMolina Thanks for looking into this. What do you suggest we do?

CruzMolina:
[...]

PR #1844, opened 2019-03-20
Fix compiler cache path

CruzMolina:
Change the `compilerCachePath` to cache compilers in a `node_modules` directory (`~/.config/truffle/compilers/node_modules/`).
This helps truffle users avoid the issue outlined in #1813.

Also indirectly related to #1835 .

### DIFF
--- a/packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
+++ b/packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
@@ -17,7 +17,7 @@ class LoadingStrategy {
 
     const compilerCachePath = path.resolve(
       Config.getTruffleDataDirectory(),
-      "compilers"
+      "compilers/node_modules" // because babel binds to require & does weird things
     );
     if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath);
     this.compilerCachePath = compilerCachePath;
--- a/packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
+++ b/packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
@@ -15,15 +15,17 @@ class LoadingStrategy {
     };
     this.config = Object.assign({}, defaultConfig, options);
 
-    let compilerCachePath = path.resolve(
+    const compilersDir = path.resolve(
       Config.getTruffleDataDirectory(),
       "compilers"
     );
-    if (!fs.existsSync(compilerCachePath)) {
-      fs.mkdirSync(compilerCachePath);
-      compilerCachePath = path.resolve(compilerCachePath, "node_modules"); // because babel binds to require & does weird things
+    const compilerCachePath = path.resolve(compilersDir, "node_modules"); // because babel binds to require & does weird things
+    if (!fs.existsSync(compilersDir)) {
+      fs.mkdirSync(compilerDir);
       fs.mkdirSync(compilerCachePath);
     }
+    if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath); // for 5.0.8 users
+
     this.compilerCachePath = compilerCachePath;
   }
 
--- a/packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
+++ b/packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
@@ -15,11 +15,15 @@ class LoadingStrategy {
     };
     this.config = Object.assign({}, defaultConfig, options);
 
-    const compilerCachePath = path.resolve(
+    let compilerCachePath = path.resolve(
       Config.getTruffleDataDirectory(),
-      "compilers/node_modules" // because babel binds to require & does weird things
+      "compilers"
     );
-    if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath);
+    if (!fs.existsSync(compilerCachePath)) {
+      fs.mkdirSync(compilerCachePath);
+      compilerCachePath += "/node_modules"; // because babel binds to require & does weird things
+      fs.mkdirSync(compilerCachePath);
+    }
     this.compilerCachePath = compilerCachePath;
   }
 
--- a/packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
+++ b/packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
@@ -21,7 +21,7 @@ class LoadingStrategy {
     );
     if (!fs.existsSync(compilerCachePath)) {
       fs.mkdirSync(compilerCachePath);
-      compilerCachePath += "/node_modules"; // because babel binds to require & does weird things
+      compilerCachePath = path.resolve(compilerCachePath, "node_modules"); // because babel binds to require & does weird things
       fs.mkdirSync(compilerCachePath);
     }
     this.compilerCachePath = compilerCachePath;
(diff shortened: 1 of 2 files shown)


packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -15,11 +15,15 @@ class LoadingStrategy {
...
 
-    const compilerCachePath = path.resolve(
+    let compilerCachePath = path.resolve(
       Config.getTruffleDataDirectory(),
       "compilers"
     );
-    if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath);
+    if (!fs.existsSync(compilerCachePath)) {
+      fs.mkdirSync(compilerCachePath);
+      compilerCachePath += "/node_modules"; // because babel binds to require & does weird things
```
eggplantzzz: Use the `path` module for this so that it will also work on Windows machines.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -15,11 +15,15 @@ class LoadingStrategy {
     };
     this.config = Object.assign({}, defaultConfig, options);
 
-    const compilerCachePath = path.resolve(
+    let compilerCachePath = path.resolve(
```
eggplantzzz: This solution will not work. If we are going to change the cache path to `~/.config/truffle/compilers/node_modules` then it needs to be done here. You can't change it later because it will only run through that if statement when it sees the directory doesn't exist. So the subsequent times it will cache things once again at `~/.config/truffle/compilers` and not in the `node_modules` folder inside `compilers`.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
CruzMolina: Good catch!

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
CruzMolina: Ah, I see. The reviewed code won't work for 5.0.8 users who've been using BYOC.

Thanks for that. Hopefully we can remove some of this code at a later date and/or find a more long-term solution.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -15,11 +15,17 @@ class LoadingStrategy {
     };
     this.config = Object.assign({}, defaultConfig, options);
 
-    const compilerCachePath = path.resolve(
+    const compilersDir = path.resolve(
```
eggplantzzz: I would say just get rid of this variable name as we don't need it. That also lets you get rid of one of the "if" statements below. All we really want to ensure is that whatever cacheing directory we choose to use exists.

eggplantzzz: I take it back. I guess we do need both to ensure both the outer and inner directory exists.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -15,11 +15,17 @@ class LoadingStrategy {
...
       Config.getTruffleDataDirectory(),
       "compilers"
     );
-    if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath);
+    const compilerCachePath = path.resolve(compilersDir, "node_modules"); // because babel binds to require & does weird things
+    if (!fs.existsSync(compilersDir)) {
+      fs.mkdirSync(compilersDir);
+      fs.mkdirSync(compilerCachePath);
```
eggplantzzz: This, however, I think you can remove. The logic is also more direct there, just two one-liners.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
CruzMolina: Yeah, it's a bummer.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
CruzMolina: I'm not sure I follow?

If the `compilersDir` doesn't already exist, then create it & then create the `node_modules` folder.

If it does already exist, but the full path doesn't already exist (5.0.8 users using BYOC will already have a `compilers` dir), then make sure it does by creating the `node_modules` folder.

CruzMolina: If I'm missing something, feel free to use the github code suggestion thingy.

eggplantzzz: I'm saying 
```suggestion
      if (!fs.existsSync(compilersDir)) fs.mkdirSync(compilerDir);
      if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath);
```
I mean yours will work but this is just simpler.

eggplantzzz: What I mean to say is that it would be simpler if the `node_modules` directory were only created in the step that checks to see if it exists. So the logic is:
  1. If the compilers directory doesn't exist, create it.
  2. If the cache directory doesn't exist, create it.

CruzMolina: Oh! I see. Cool 😎

eggplantzzz:
Looks cool to me, this will do the trick?

CruzMolina:
I think so! I've demo'd it on a couple repos requiring `babel/register` and so far so good.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -15,11 +15,14 @@ class LoadingStrategy {
...
 
-    const compilerCachePath = path.resolve(
+    const compilersDir = path.resolve(
       Config.getTruffleDataDirectory(),
       "compilers"
     );
-    if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath);
+    const compilerCachePath = path.resolve(compilersDir, "node_modules"); // because babel binds to require & does weird things
+    if (!fs.existsSync(compilersDir)) fs.mkdirSync(compilersDir);
+    if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath); // for 5.0.8 users
```
gnidan: Could you explain more about the reasoning behind this case? Not sure about what "for 5.0.8 users" means.

packages/truffle-compile/test/test_supplier.js
```diff
@@ -125,7 +125,7 @@ describe("CompilerSupplier", function() {
 
       const compilerCacheDirectory = path.resolve(
         Config.getTruffleDataDirectory(),
-        "compilers"
+        "compilers/node_modules"
```
gnidan: Is it worth de-duplicating this? Since this path join happens in two places now. Just a thought.

packages/truffle-compile/compilerSupplier/loadingStrategies/LoadingStrategy.js
CruzMolina: 5.0.8 users now have a `compilers` directory in their truffle config path if they've used BYOC.

Tyler brought up the good point of making sure the original PR logic didn't skip over them (it was a check just for `compilers` before creating `compilers` & then `node_modules`)

packages/truffle-compile/test/test_supplier.js
CruzMolina: I like this test since it sanity checks if the supplier is actually working. I don't think the test would work as it is intended w/o updating the compilerCacheDirectory path to be what BYOC is now using.

PR #1850, opened 2019-03-23
Update readme: how to use truffle-hdwallet-provider with Web3

Muhammad-Altabba:
It was not clear in the documentation that HDWalletProvider is compatible with Web3. Added code snippet for using HDWalletProvider with new and existing Web3 instance.


Muhammad-Altabba:
My pleasure!
Thanks,

PR #1854, opened 2019-03-25
Restore ability to use external list of breakpoints

haltman-at:
Background: Originally breakpoints were stored in the CLI, not the debugger state; the debugger exposed a function to stop at a specified breakpoint, but you had to supply the list of breakpoints yourself.  Later I moved them into the debugger state, exposing functions to add and remove breakpoints instead of having the list of breakpoints be an input to the `continueUntilBreakpoint` function.

However, I recently realized that if you're using the debugger as a library, this isn't necessarily a convenient interface.  So, I decided to add the old way back in as another way of doing things.  The `continueUntilBreakpoint` function now takes as an optional argument a list of breakpoints.  If supplied, it will continue to one of these breakpoints, rather than using the list stored in the state.  So, if you're using the debugger as a library, you may want to use this instead.

I also added a test.

Note: Yes, I used some redundant defaults here.

### DIFF
--- a/packages/truffle-debugger/lib/controller/sagas/index.js
+++ b/packages/truffle-debugger/lib/controller/sagas/index.js
@@ -190,12 +190,18 @@ function* stepOver() {
 /**
  * continueUntilBreakpoint - step through execution until a breakpoint
  */
-function* continueUntilBreakpoint() {
+function* continueUntilBreakpoint(action = actions.continueUntilBreakpoint()) {
   var currentLocation, currentNode, currentLine, currentSourceId;
   var finished;
   var previousLine, previousSourceId;
 
-  let breakpoints = yield select(controller.breakpoints);
+  //if breakpoints was not specified, use the stored list from the state.
+  //if it was, override that with the specified list.
+  //note that explicitly specifying an empty list will advance to the end.
+  let breakpoints =
+    action.breakpoints !== undefined
+      ? action.breakpoints
+      : yield select(controller.breakpoints);
 
   let breakpointHit = false;
 
--- a/packages/truffle-debugger/lib/controller/sagas/index.js
+++ b/packages/truffle-debugger/lib/controller/sagas/index.js
@@ -48,9 +48,11 @@ export default prefixName("controller", saga);
 
 /*
  * Advance the state by the given number of instructions (but not past the end)
+ * (if no count given, advance 1)
  */
-function* advance(action = actions.advance()) {
-  let { count } = action;
+function* advance(action) {
+  let count =
+    action !== undefined && action.count !== undefined ? action.count : 1; //default is, as mentioned, to advance 1
   for (let i = 0; i < count && !(yield select(controller.finished)); i++) {
     yield* trace.advance();
   }
@@ -190,7 +192,7 @@ function* stepOver() {
 /**
  * continueUntilBreakpoint - step through execution until a breakpoint
  */
-function* continueUntilBreakpoint(action = actions.continueUntilBreakpoint()) {
+function* continueUntilBreakpoint(action) {
   var currentLocation, currentNode, currentLine, currentSourceId;
   var finished;
   var previousLine, previousSourceId;
@@ -199,7 +201,7 @@ function* continueUntilBreakpoint(action = actions.continueUntilBreakpoint()) {
   //if it was, override that with the specified list.
   //note that explicitly specifying an empty list will advance to the end.
   let breakpoints =
-    action.breakpoints !== undefined
+    action !== undefined && action.breakpoints !== undefined
       ? action.breakpoints
       : yield select(controller.breakpoints);
 
--- a/packages/truffle-debugger/lib/session/index.js
+++ b/packages/truffle-debugger/lib/session/index.js
@@ -152,7 +152,8 @@ export default class Session {
     });
   }
 
-  async advance(count = 1) {
+  //Note: count is an optional argument; default behavior is to advance 1
+  async advance(count) {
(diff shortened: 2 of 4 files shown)


packages/truffle-debugger/lib/controller/sagas/index.js
```diff
@@ -190,12 +190,18 @@ function* stepOver() {
 /**
  * continueUntilBreakpoint - step through execution until a breakpoint
  */
-function* continueUntilBreakpoint() {
+function* continueUntilBreakpoint(action = actions.continueUntilBreakpoint()) {
```
gnidan: This gets run as a listener, right? Seems weird to construct the action here as a default value, when we can expect the type of argument to contain a `breakpoints` key.

```suggestion
function* continueUntilBreakpoint({ breakpoints }) {
```

packages/truffle-debugger/lib/controller/sagas/index.js
```diff
@@ -190,12 +190,18 @@ function* stepOver() {
...
   var previousLine, previousSourceId;
 
-  let breakpoints = yield select(controller.breakpoints);
+  //if breakpoints was not specified, use the stored list from the state.
+  //if it was, override that with the specified list.
+  //note that explicitly specifying an empty list will advance to the end.
+  let breakpoints =
+    action.breakpoints !== undefined
+      ? action.breakpoints
+      : yield select(controller.breakpoints);
```
gnidan: Then this just becomes:

```javascript
breakpoints = breakpoints || yield select(controller.breakpoints);
```

packages/truffle-debugger/lib/session/index.js
```diff
@@ -176,8 +176,12 @@ export default class Session {
     return await this.doneStepping(controller.reset());
   }
 
-  async continueUntilBreakpoint() {
-    return await this.doneStepping(controller.continueUntilBreakpoint());
+  async continueUntilBreakpoint(
+    breakpoints = this.view(controllerSelector.breakpoints)
```
gnidan: I think we can skip the default here since we're setting defaults internally, but we'll need to use a comment or an explicit `breakpoints = undefined` so it's clear that it's an optional arg.

packages/truffle-debugger/lib/controller/sagas/index.js
haltman-at: See below.

packages/truffle-debugger/lib/controller/sagas/index.js
haltman-at: Yeah, sometimes I like to do my `undefined`-checking a bit more explicitly, but I acknowledge it's not actually necessary. :P

packages/truffle-debugger/lib/controller/sagas/index.js
haltman-at: Oh wait nevermind what I was going to write was wrong.  Yeah this would work.  I was just doing redundant defaults like I said.

haltman-at: Oh wait I think I know what I was thinking.  It does run as a listener, but I was considering also the possibility that it might be called directly, since some of the others are.

packages/truffle-debugger/lib/session/index.js
haltman-at: OK the thing I was originally going to write here was wrong.  But yeah obviously we don't actually need redundant defaults like I did.

Btw, note that I also did the redundant default thing when I added multiple advancement; do you think I should take that out, have the default live in only one place?

I think you're right it's worth adding a comment at least.

packages/truffle-debugger/lib/controller/sagas/index.js
gnidan: My thinking on this is the following:

1. The only interface requirement here is an object with `breakpoints` property that may be undefined.
2. It's extra work to create the action via that function call, and potentially extra cognitive overhead because usually we create actions when putting, not taking.

packages/truffle-debugger/lib/controller/sagas/index.js
gnidan: The `||` is a common enough JS pattern for function arguments, which is why I suggest it in tandem with the interface suggestion above.

packages/truffle-debugger/lib/session/index.js
gnidan: There've been a couple instances recently where doubling-down on defaults has come back to bite. (Truffle was providing its own defaults for solc's `evmVersion` and Ganache's `hardfork`)

So now I'm of the mind that it's probably just best practice to rely on the innermost defaults. In this situation, I think it's fine to leave defaulting as a concern of the controller saga, and leave it to everything external to that with allowable undefineds.

packages/truffle-debugger/lib/controller/sagas/index.js
haltman-at: OK. Just, I would also like it to be possible to call it internally with no arguments.  But I can do that without creating an action.

packages/truffle-debugger/lib/session/index.js
haltman-at: Got it!  I'll go back and fix up the `advance` function similarly while I'm at it, then.

packages/truffle-debugger/lib/controller/sagas/index.js
gnidan: Ah right, without arguments at all. That makes the code less pretty. Well still, probably not worth making the action.

PR #1855, opened 2019-03-26
Properly handle failed calls

haltman-at:
Previously, the Truffle Debugger did nothing to handle exceptional halting.  Now, in Solidity, this usually isn't much of a problem, because if one call in the transaction reverts, the entire transaction reverts in short order.  However, this isn't necessarily the case; by using `<address>.call()` (and related functions), one may catch an exceptional halt and resume execution.  As such, we need to be able to handle this case.  This requires additional state to be tracked in both the `evm` and `solidity` submodules.

Before we get to this additional state, though, let's discuss how we detect exceptional halting in the first place.  We can't just look for the `REVERT` or `INVALID` opcodes, as there's any number of ways an exceptional halt might occur.  Thus, we detect an exceptional halt as any time the depth is about to decrease but the current opcode is *not* `STOP` or `RETURN`.

Speaking of which, I've changed to an easier method of detecting calls to precompiles (or externally-owned accounts); rather than seeing whether we're able to get the binary of the address we're calling, we just check whether the depth increases or not.  If the depth stays the same, we're calling a precompile or externally-owned account, and don't need to add to the callstack.

Also, to make such comparisons simpler, I've altered `trace.next` so that, if you're at the end of the trace, `trace.next` does not return the same thing as `trace.step`, but rather returns a spoofed trace step of depth -1.  This prevents us from having to deal with special cases at the end of the trace.

So, anyway, what's the actually-new functionality here?  Well, in `evm`, we now have a new action called `FAIL`, that's triggered on exceptional halting.  The `callstack` reducer now recognizes `FAIL` and will pop a stackframe on `FAIL` as well as on `RETURN`.  (There's also another new action: `LOAD` has been separated from `STORE`, where previously they were treated in a unified manner.  But I'll get back to that.)

However the bigger change is in the `codex` reducer.  I've had to reorganize the codex, you see.  The original plan was that it would have two members, `byAddress` for things that are stored by address (like storage) and `byStackframe` for things that are stored by stackframe (like memory -- not that that's in the codex yet).  However, given that we have to account for the possibility of reversion, *everything* needs to be by stackframe!  It's just that some things need to be by address *and* stackframe.

So, the codex is now a stack, an array of stackframe objects.  Again, currently it only tracks storage, so each stackframe object currently only holds a single member, `accounts`, which is then indexed by address, and the object for each address currently only holds a single member, `storage`.  But yeah, it's the codex.

So now, on `CALL` or `CREATE`, rather than just adding a new page to the codex if necessary, we push a new stackframe onto the codex (a clone of the old top stackframe), and then add a new page to *that* if necessary.  On `STORE`, we alter only the top stackframe of the codex.  On `FAIL`, we pop the codex -- the top stackframe has reverted, so its contents are lost!  Finally, on `RETURN`, we pop the codex, but then make the new top stackframe a clone of the old; i.e., we save the results of the stackframe that just completed, rather than discarding them.  (Of course, in the actual code, I implemented this by just dropping the stackframe second from the top.)

Now I mentioned that on `STORE` we alter only the top stackframe of the codex.  What about on `LOAD`?  Well, that depends.  If the slot we're loading is already in the codex, we no longer bother altering anything.  But if it wasn't already in the codex, that means we're loading something preexisting -- something from before this transaction.  That's not a value that will get reverted when the call fails, because it's a value that wasn't actually set in this transaction at all!  So on such a load, we update *every* stackframe.

(Btw, it's now entirely the reducer's responsibility for checking for the zero address and refusing to maintain a page for this address; the saga no longer does this.)

So now we correctly track storage when a call reverts!  Previously, we didn't distinguish stackframes like this, so a change to storage made in a failing call would not be reverted at the end of that call.

But there's another problem -- when an EVM call exceptionally halts, it could do so from inside any number of internal Solidity function calls.  So we need to be able to restore the function depth, too.

So, now, in the `solidity` state, instead of `functionDepth`, there's `functionDepthStack`.  EVM calls and returns are no longer handled with the `JUMP` action, but have new actions, `EXTERNAL_CALL` and `EXTERNAL_RETURN`.  A `JUMP` now just alters the top element of the stack.  An `EXTERNAL_CALL` grows the stack by 1, pushing a duplicate of the top element onto the stack and then incrementing *that* by 1.  And an `EXTERNAL_RETURN` (which covers both ordinary and exceptional halting) pops the stack by 1.

As a nice side benefit, this allows us to get rid of that awful function depth workaround I wrote for Solidity versions prior to 0.5.1!  This method works regardless of Solidity compiler version, and, unlike that workaround, will work just fine with function variables.  I've torn out a whole bunch of code that existed solely to support it.  I did, however, leave `compiler` in the `contexts` state, because we might need it in the future.

By the way, notice that this PR doesn't touch the `data` submodule at all.  So, any mapping keys that we start tracking during a failed call will stay tracked when the call reverts.  I don't think that's a problem.

I also added a few tests of this feature.

### DIFF
--- a/packages/truffle-debugger/lib/solidity/reducers.js
+++ b/packages/truffle-debugger/lib/solidity/reducers.js
@@ -77,14 +77,24 @@ const info = combineReducers({
   sourceMaps
 });
 
-export function functionDepth(state = 0, action) {
+export function functionDepthStack(state = [0], action) {
   switch (action.type) {
     case actions.JUMP:
+      let newState = state.slice(); //clone the state
       const delta = spelunk(action.jumpDirection);
-      return state + delta;
+      let top = newState[newState.length - 1];
+      newState[newState.length - 1] = top + delta;
+      return newState;
 
     case actions.RESET:
-      return 0;
+      return [0];
+
+    case actions.EXTERNAL_CALL:
+      return [...state, state[state.length - 1] + 1];
+
+    case actions.EXTERNAL_RETURN:
+      //just pop the stack! unless, HACK, that would leave it empty
+      return state.length > 1 ? state.slice(0, -1) : state;
 
     default:
       return state;
@@ -96,15 +106,13 @@ function spelunk(jump) {
     return 1;
   } else if (jump === "o") {
     return -1;
-  } else if (jump === "2") {
-    return 2; //HACK WORKAROUND
   } else {
     return 0;
   }
 }
 
 const proc = combineReducers({
-  functionDepth
+  functionDepthStack
 });
 
 const reducer = combineReducers({
--- a/packages/truffle-debugger/lib/evm/actions/index.js
+++ b/packages/truffle-debugger/lib/evm/actions/index.js
@@ -53,6 +53,13 @@ export function returnCall() {
   };
 }
 
+export const FAIL = "FAIL";
+export function fail() {
+  return {
+    type: FAIL
+  };
+}
+
 export const STORE = "STORE";
 export function store(address, slot, value) {
   return {
--- a/packages/truffle-debugger/lib/evm/actions/index.js
+++ b/packages/truffle-debugger/lib/evm/actions/index.js
@@ -70,6 +70,16 @@ export function store(address, slot, value) {
   };
 }
 
+export const LOAD = "LOAD";
+export function load(address, slot, value) {
+  return {
+    type: LOAD,
+    address,
+    slot,
+    value
+  };
+}
+
 export const RESET = "EVM_RESET";
 export function reset(storageAddress) {
   return {
--- a/packages/truffle-debugger/lib/evm/reducers.js
+++ b/packages/truffle-debugger/lib/evm/reducers.js
@@ -134,8 +134,9 @@ export function callstack(state = [], action) {
     }
 
     case actions.RETURN:
-      //HACK: pop the stack, UNLESS that would leave it empty (this will only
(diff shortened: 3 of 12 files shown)


packages/truffle-debugger/lib/solidity/reducers.js
```diff
@@ -77,14 +77,24 @@ const info = combineReducers({
...
 
-export function functionDepth(state = 0, action) {
+export function functionDepthStack(state = [0], action) {
   switch (action.type) {
     case actions.JUMP:
+      let newState = state.slice(); //clone the state
       const delta = spelunk(action.jumpDirection);
-      return state + delta;
+      let top = newState[newState.length - 1];
+      newState[newState.length - 1] = top + delta;
```
gnidan: While we're at it, can we make the state look like:
```javascript
functionDepthStack: [{ depth: n }]
```

Or even rename `functionDepthStack` to something like `functionStack` or `jumpStack`? Then we can leave `functionDepth` as a name for the corresponding property of the stack frame, and that property can just remain an integer.

haltman-at: I dunno, that just seems overcomplicated to me.  I'm having a hard time thinking of what else might potentially go on this particular stack.  Yes it's likely we'll eventually want a more proper Solidity stack, that we don't currently have, but this isn't the start of that -- remember, the elements of this stack represent EVM stackframes, not Solidity stackframes.

PR #1858, opened 2019-03-27
Split off watch variables from watch expressions and make the latter work better

haltman-at:
(Edit: The original verison of this PR had some mistakes, I've edited the text below to match the current version.)

This PR makes it so that watch expressions are treated differently based on whether or not they consist of just a single variable.  Now, when a watch expression does not just consist of a single variable, the `variables` object is first run through an additional `nativize` step before the watch expression is evaluated.  This step converts `Map`s to plain old objects and `BN`s to numbers (although this may cause an error).  The point of this is to allow arithmetic operations and indexing to work normally, just as it would in Solidity, without the user having to use `BN` operations or `get()`.

This does mean that the same object can print out differently depending on whether the expression you enter consists of just a single variable or not, and it does mean that we have the problems with numbers above (and also mappings with boolean keys?), but I think this is the right tradeoff to make for the moment.

Also, yes, converting things from objects to `Map`s and then later converting them back is a bit silly, but it seems like the best way with the current structure.  Things will get a bit less silly once we have the new decoder output format.

### DIFF
--- a/packages/truffle-core/lib/commands/debug.js
+++ b/packages/truffle-core/lib/commands/debug.js
@@ -283,7 +283,9 @@ var command = {
               .join(OS.EOL);
           }
 
-          function printVariables(variables) {
+          async function printVariables() {
+            let variables = await session.variables();
+
             debug("variables %o", variables);
 
             // Get the length of the longest name.
@@ -344,7 +346,8 @@ var command = {
 
             let variable = raw.trim();
             if (variable in variables) {
-              printVariables({ [variable]: variables[variable] });
+              let formatted = formatValue(variables[variable], indent);
+              config.logger.log(formatted);
               return;
             }
 
@@ -710,7 +713,7 @@ var command = {
                 printWatchExpressions();
                 break;
               case "v":
-                printVariables(await session.variables());
+                await printVariables();
                 break;
               case ":":
                 evalAndPrintExpression(cmdArgs);
--- a/packages/truffle-core/lib/commands/debug.js
+++ b/packages/truffle-core/lib/commands/debug.js
@@ -262,8 +262,6 @@ var command = {
           // TODO make this more robust for all cases and move to
           // truffle-debug-utils
           function formatValue(value, indent) {
-            value = DebugUtils.cleanConstructors(value); //HACK
-
             if (!indent) {
               indent = 0;
             }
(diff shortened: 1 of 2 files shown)


haltman-at:
OK, just pushed an update to this PR that makes it so that my `cleanConstructors` hack will now only be invoked for actual watch expressions, not for watch variables or when pressing `v` (yay!).  This is now possible because the extra `nativize` step means that `safeEval` is now altering the result of *that* rather than the object containing the actual original variables.

haltman-at:
OK, at @gnidan's request, this PR now also includes analytics for whether people are watching expressions or individual variables (I excluded selectors from consideration).

haltman-at:
Sorry, one more fix -- someone might start watching a variable before it comes into scope, so I've altered the analytics logic to go based on whether the entered expression has the form of an identifier, rather than by whether it matches an existing variable.

packages/truffle-core/lib/commands/debug.js
```diff
@@ -402,6 +423,21 @@ var command = {
...
+            if (raw.includes("!<")) {
+              //don't send analytics for watch expressions involving selectors
+              return;
+            }
+            let expression = raw.trim();
+            //legal Solidity identifiers (= legal JS identifiers)
+            let identifierRegex = /^[a-zA-Z_$][a-zA-Z_$0-9]*$/;
+            let isVariable = expression.match(identifierRegex) !== null;
+            analytics.send({
+              command: "watch expression",
```
gnidan: Is there a way to scope this within `debug`? If it's not straightforward then it's not necessary, but just curious.

haltman-at: I don't think there's any formal way, but I could add "`debug:`" or something to the beginning.  The impression I got from @fainashalts was that we don't have any current convention for this, but I guess this could be the start of one!

fainashalts: Adding `debug:` seems like a valid solution to this issue. Would definitely make it clearer where this is coming from, on the analytics side of things. Another option would be to add the watch expression part to the args -- could make args "watch expression true" and "watch expression false" since its a string anyway on the analytics side. I'm ok with either option!

haltman-at: Oh so like `command = "debug"`, `args = {type:"watch expression", isVariable}`.  That makes sense.

haltman-at: Hm, @gnidan points out that actually `command = "debug"` is probably not a good idea, as right now `command = "debug"` always means that `debug` is being invoked.  So going with the colon way after all.

PR #1859, opened 2019-03-27
Add some comments and tweak some tests

haltman-at:
This PR is just a slight update to #1855.  I decided that the `LOAD` case of the `codex` reducer wasn't commented enough so I added some comments.  (Specifically, I was worried that some future person looking at it would think it looks wrong and attempt to "fix" it, so I added some comments explaining why it works the way it does.)  Also, I added a bit more to the tests I added in that PR, so that now some of the `LOAD` case is tested as well, not just the `STORE` case.  Finally, while doing this I noticed that an old test was still lowercasing addresses before comparison, which should no longer be necessary, so I removed the conversion to lowercase.  That's all, just minor internal stuff.


haltman-at:
Also I (just now) turned up the timeouts on two of the stackframe tests because they're timing out occasionally.

PR #1862, opened 2019-03-29
Strip-down and rewrite truffle-code-utils while also updating it

haltman-at:
So, this PR came about when I noticed two things:

1. The list of opcodes in `truffle-code-utils` was quite out of date, and
2. Most of the information there wasn't being used.

So, `truffle-code-utils` has been stripped down to remove all the information we're not using -- it will now return only opcode names and push data -- and also updated to keep up with the latest EVM revisions.

There's one more thing that's been changed -- how stripping the contract metadata has been handled.

See, the existing code just always stripped off the last 43 bytes.  This has two problems.  The lesser problem is that it's not future-proof; there's a reason the metadata ends in a length field.  This problem is easy to solve.

The greater problem is that stripping off the contract metadata isn't actually what we want!  What we actually want is to strip off the entire data section, which can be longer than just the metadata at the end.  (If we're dealing with the binary of a running constructor, it could also have arguments, which come *after* the metadata, so relying on metadata stripping could go really wrong.  Admittedly, it'd go less wrong if we stuck to 43 bytes rather than reading the last two bytes as a length, but...)

Unfortunately, there's not really any marker of where the data section begins.  But, if we do know the number of instructions from elsewhere -- as can be learned from, say, the sourcemap -- we can use that to determine a stopping point.

So, `parseCode` now takes an additional argument, `numInstructions`.  If passed in, it will not strip off the metadata, but rather simply stop after the given number of instructions.  If not passed in, it will use the old behavior and attempt to strip off the metadata (well, not quite the old behavior, as now it uses the length stored in the last two bytes).

Note that `truffle-code-utils` is only used in two places right now; in the debugger and in `truffle opcode`.  I modified both of these to pass in the length rather than relying on metadata stripping, but I still left in the option to do things the old way because technically `truffle-code-utils` is its own package so I didn't want to break the interface (although maybe I should?).  (Also wondering if I should make these separate options, so that you don't have to know the length to skip metadata stripping.  Well, I think the current way is fine for now.)

Also, minor thing, but since it's technically EVM-legal (though it will never occur in a Solidity contract), I made it so that if a `PUSH` instruction runs off the end of the code, the remainder of the push data will be filled with zeroes, because that is how a `PUSH` instruction is supposed to work in that case.


packages/truffle-code-utils/index.js
```diff
@@ -4,32 +4,56 @@ module.exports = {
...
+   * If you pass in numInstructions (hint: count the semicolons in the source
+   * map, then add one) this is used to exclude metadata instead.
    *
    * @param  {String} hex_string Hex string representing the code
    * @return Array               Array of instructions
    */
-  parseCode: function (hex_string) {
-    // Convert to an array of bytes (denoted by substrings)
-    var code = hex_string.match(/(..?)/g);
+  parseCode: function(hex_string, numInstructions = null) {
```
gnidan: wanna change it to `hexString` while you're at it?

haltman-at: Sure, why not. :P

Issue #1864, opened 2019-03-29
Warn if both truffle-config.js and truffle.js is provided.


mrwillis:
- [x ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

What the issue is, in broad strokes.
There should be a warning if both truffle-config.js and truffle.js are provided when running truffle commands. 

## Steps to Reproduce

Please provide the shortest amount of steps to reproduce your issue.

## Expected Behavior

What you expected to happen.

## Actual Results

What actually happened. Please give examples and support it with screenshots, copied output or error messages.

## Environment

* Operating System: ubuntu 18
* Ethereum client: infura
* Truffle version (`truffle version`): 5.0.8
* node version (`node --version`):10.5.0
* npm version (`npm --version`): 6.7.0

adrianmcli:
Sounds pretty sensible. Would love to see a PR, @mrwillis.

mrwillis:
wrote a draft here https://github.com/trufflesuite/truffle/pull/1866

eggplantzzz:
@mrwillis This should be fixed on the latest version of Truffle. Closing this, feel free to reopen an issue if it is still a problem.

SvenMeyer:
I only have `truffle-config.js` in my project folder, but get the message  

`Warning: Both truffle-config.js and truffle.js were found. Using truffle-config.js`

$ truffle version  

Warning: Both truffle-config.js and truffle.js were found. Using truffle-config.js. \
Truffle v5.0.22 (core: 5.0.22) \
Solidity v0.5.0 (solc-js) \
Node v10.16.0 \
Web3.js v1.0.0-beta.37

ciberexplosion:
> I only have `truffle-config.js` in my project folder, but get the message
> 
> `Warning: Both truffle-config.js and truffle.js were found. Using truffle-config.js`
> 
> $ truffle version
> 
> Warning: Both truffle-config.js and truffle.js were found. Using truffle-config.js.
> Truffle v5.0.22 (core: 5.0.22)
> Solidity v0.5.0 (solc-js)
> Node v10.16.0
> Web3.js v1.0.0-beta.37

Did you find any solution? I have the same problem

J-Cook-jr:
> > I only have `truffle-config.js` in my project folder, but get the message
> > `Warning: Both truffle-config.js and truffle.js were found. Using truffle-config.js`
> > $ truffle version
[...]

PR #1866, opened 2019-03-29
Update getTruffleConfig.js to warn if two configs are found

mrwillis:
Warn if truffle-config.js and truffle.js are found. Functionality remains the same (use truffle.js on non-windows and use truffle-config.js on windows). This saves some head scratching if you run into a project with both configuration files and wondering why your project isn't changing as you change the wrong config.


fainashalts:
This looks good to me. Thanks for writing this up @mrwillis! Will leave to @CruzMolina for final approval and merging in since his review has been requested.

mrwillis:
Right, indeed. Would you like me to edit it?

CruzMolina:
@mrwillis that'd be great! I imagine you might want to move this to the `packages/truffle-config/index.js` file. Could be wrong though.

Now I'm curious about where `getTruffleConfig` is being used and why it isn't logging the output that would be expected. 😂

eggplantzzz:
Just for the record, I believe in the past `truffle-config.js` was intended to be used by Windows users to avoid a name conflict with the command to run Truffle.

Going forward I think we want Truffle to use `truffle-config.js` as the default config file for both Mac and Windows while still accepting `truffle.js` as a fallback config when it is present and `truffle-config.js` cannot be found.

@gnidan Can you confirm that this is the desired behavior?

CruzMolina:
Roger that @eggplantzzz .

This PR is specifically addressing #1864.

packages/truffle-solidity-loader/lib/getTruffleConfig.js
```diff
@@ -3,12 +3,20 @@ const Logger = require('./logDecorator');
 
 const getTruffleConfig = function () {
   const isWin = /^win/.test(process.platform);
+  let truffleJsFile = findUp.sync('truffle.js');
+  let truffleConfigJsFile = findUp.sync('truffle-config.js');
   let file;
+  
+  if (truffleJsFile && truffleConfigJsFile && isWin) {
```
eggplantzzz: So I think this can just be 
```
if (truffleJsFile && truffleConfigJsFile) {
  giveAWarning...
```

packages/truffle-solidity-loader/lib/getTruffleConfig.js
```diff
@@ -3,12 +3,20 @@ const Logger = require('./logDecorator');
...
+    Logger.log("Warning: Both truffle.js and truffle-config.js were found. Using truffle-config.js");
+  } else if (truffleJsFile && truffleConfigJsFile && !isWin) {
+    Logger.log("Warning: Both truffle.js and truffle-config.js were found. Using truffle.js");
+  }
 
   if (isWin) {
```
eggplantzzz: And here we just want to give precedence to `truffle-config.js` if they both exist. We are going to consider the Mac and Windows cases exactly the same.

eggplantzzz:
@mrwillis If you go ahead and make the suggested changes we will get this merged!

mrwillis:
So it seems that `truffle-solidity-loader` is not actually used for truffle CLI. Seems to be a webpack plugin. I think the actual code being used is `truffle-config/config.js`

CruzMolina:
Hey @mrwillis, in the process of fixing a bug for `truffle-solidity-loader`, I streamlined `getTruffleConfig.js`. Feel free to steal from that code to use in `packages/truffle-config/index.js`.

mrwillis:
I updated the code already. Check it out

CruzMolina:
Nice, will be reviewing and pushing some changes shortly.

packages/truffle-config/index.js
```diff
@@ -7,8 +7,8 @@ const findUp = require("find-up");
 const originalrequire = require("original-require");
 const Configstore = require("configstore");
 
-const DEFAULT_CONFIG_FILENAME = "truffle.js";
-const BACKUP_CONFIG_FILENAME = "truffle-config.js"; // For Windows + Command Prompt
+const DEFAULT_CONFIG_FILENAME = "truffle-config.js";
+const BACKUP_CONFIG_FILENAME = "truffle.js"; // For Windows + Command Prompt
```
gnidan: comment is outdated, might want to update it to say "old config filename" or somesuch

CruzMolina:
```suggestion
const BACKUP_CONFIG_FILENAME = "truffle.js"; // old config filename
```

PR #1876, opened 2019-04-03
 Add PC display to debugger CLI

haltman-at:
The lack of a PC display in the CLI had been bugging me, so this PR adds one.  Of course, it is only shown when pressing `p` or `;`.  I also got rid of a redundant use of an `evm` selector when we already had that same information available from the `trace` selector we were already using.

### DIFF
--- a/packages/truffle-debug-utils/index.js
+++ b/packages/truffle-debug-utils/index.js
@@ -252,7 +252,11 @@ var DebugUtils = {
   },
 
   formatPC: function(pc) {
-    return "  PC = " + pc.toString() + " = 0x" + pc.toString(16);
+    let hex = pc.toString(16);
+    if (hex.length % 2 !== 0) {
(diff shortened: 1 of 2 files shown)


packages/truffle-debug-utils/index.js
```diff
@@ -251,6 +251,14 @@ var DebugUtils = {
     );
   },
 
+  formatPC: function(pc) {
+    let hex = pc.toString(16);
+    if (hex.length % 2 !== 0) {
+      hex = "0" + hex; //ensure even length
+    }
+    return "  PC = " + pc.toString() + " = 0x" + hex;
```
gnidan: is the assumption that anyone looking for this info will know what `PC` stands for?

haltman-at: Basically, yeah.

Issue #1795, opened 2019-03-05
truffle deploy succeds when deploy dependencies, but truffle test doen't


Amxx:
Context
======

I have contracts in `./contracts/` as well has dependencies in `./node_modules/someRepo/contracts`. The dependencies are already compiles and the artifacts are in `./node_modules/someRepo/build/contracts`

When I do `truffle deploy`, I see my contracts being compiled, and the deployment works for both the local contracts (`artifact.require("xxx")`) and the dependencies (`artifact.require("someRepo/yyy")`). Once the deployment has been performed I can run my tests with `truffle test`, `yyy.json` is also copied to `./build/contracts`

Issue
====

If I start fresh (`rm -r build`) and perform `truffle test`, truffle starts by compiling the local contract and then fails during the deployment with an error `Error: Could not find artifacts for someRepo/yyy from any sources`

Concerned version
===============

- for `5.0.0`, `5.0.1`, `5.0.2`, `5.0.3`, `5.0.4`, `5.0.5`, `5.0.7` I can deploy then test, but not test directly
- for `5.0.6` I cannot deploy nor test
- haven't tested with `4.x.x` (need solidity 0.5.x)

CruzMolina:
Hey @Amxx! I've tried reproducing your error locally on a sample truffle project using `openzeppelin-solidity` dependency contracts, but so far things are working.

Could you share a truffle project repo w/ a setup similar to the one not working?

Tested on: latest Truffle v5 (`5.0.7`) & latest truffle v4 release (`4.1.15`).

gnidan:
Doing some issue maintenance: @Amxx is this still an issue on the latest version? Please let us know if that PR above fixed it for you. Thanks!

AleG94:
I'm currently having the same issue with Truffle 5.1.5. In the migration file I require artifacts from a dependency like:

`const ProxyFactory = artifacts.require('@openzeppelin/upgrades/ProxyFactory')`

If I run `truffle migrate` and then `truffle test` everything goes well but if I delete the build folder and I directly run `truffle test` (without migrate first) it will fail with error:

`Error: Could not find artifacts for @openzeppelin\upgrades\ProxyFactory from any sources`

eggplantzzz:
@AleG94 Sorry for the delay in getting back to you! Can you confirm that this is still an issue on the latest Truffle (version 5.1.6)? And if it is, do you have a test repo that I could use to reproduce the problem?

AleG94:
[...]

Issue #450, opened 2017-06-14
Dependency contracts are not available in tests unless a truffle project's contract use them.


maraoz:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Dependency contracts are not available in tests unless a truffle project's contract use them.

## Steps to Reproduce

Create a new truffle project with a npm dependency (example: `npm i zeppelin-solidity`). Create a test that uses one of the dependency contracts that's not required by any of the main project's contracts. Truffle will say it can't find the file, because it will look into the build folder and it won't find it there.

## Expected Behavior

Truffle tests should be able to use any dependency contract, not only those used by the project's contracts.

## Actual Results

Can't find the contract and test fails.

## Environment

* Operating System: Distributor ID:	Ubuntu 16.10
* Truffle version:Truffle v3.2.5
* Ethereum client: testrpc
* node version: v8.0.0
* npm version: 5.0.3

interfect:
I can reproduce this (at least in migrations) on Truffle 4.1.3.

I have a project with a dependency on the `macroverse` NPM module. In a migrations script I try to do:

```
var RealMath = artifacts.require("macroverse/contracts/RealMath.sol");
```

This does not work.

I can add a `Workaround.sol` to my project's `contracts` folder consisting of:

```
import "macroverse/contracts/RealMath.sol";
```

If I do that and a `truffle compile`, the dependency's contract is built into `build/contracts/RealMath.json` and is available to the migration script.

I *need* to be able to use the JSON files in `node_modules/macroverse/build/contracts/*.json` from my dependent project, however, as those files from the dependency contain the mainnet addresses of the deployed contracts which the dependent project's migration scripts are going to need to have available. I am pretty sure this workaround results in new artifacts being built in `build/contracts/*.json` with empty `networks` fields.

I *used* to be able to get at the dependency's artifacts as just `artifacts.require("macroverse/RealMath.sol")`, in about Truffle 3.0. Since updating to 4.1.3 that hasn't worked and I haven't been able to find anything else that does. `macroverse/build/contracts/RealMath.json` and `macroverse/build/contracts/RealMath` don't work, for example. Also, there's the syntax shown in [the documentation](http://truffleframework.com/docs/getting_started/packages-npm) for use in migration scripts:
```
// Note that artifacts.require takes care of creating an abstraction for us.
[...]

PR #1878, opened 2019-04-03
Fix solidity unit test import_paths

### LINKED ISSUE #700: Files imported from the test directory cannot find already generated artifacts
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Whenever there is a file imported into a test file when running `truffle test`, and the contract object itself doesn't match the file name, truffle complains that it cannot find the artifacts.

## Steps to Reproduce

Take your standard owned contract:

```
pragma solidity ^0.4.16;

contract owned {
    function owned() public { owner = msg.sender; }
    address owner;

    // This contract only defines a modifier but does not use
    // it - it will be used in derived contracts.
    // The function body is inserted where the special symbol
    // "_;" in the definition of a modifier appears.
    // This means that if the owner calls this function, the
    // function is executed and otherwise, an exception is
    // thrown.
    modifier onlyOwner {
        require(msg.sender == owner);
        _;
    }

    function changeOwner(address newOwner) public onlyOwner {
        owner = newOwner;
    }
}
```
- put it into the contracts directory under the filename "Utils.sol".
- use it in a contract (inherit the owned principle into some Widget contract). 
- create a TestWidget.sol contract in the test directory. Import the Widget contract. 
- Try to run `truffle test`

## Expected Behavior

It should be able to find it based on the artifacts in the build directory by scanning through each of the artifacts and looking to the sourcePath key in all of the artifacts to see if it matches...or it could just recompile. Whichever you think works best given the project needs. 

## Actual Results

```
  1) TestRoleManager "before all" hook: prepare suite:
     Error: Could not find artifacts for /Users/vor0220/itpeople/fit-repo-bitbucket/contracts/Utils.sol from any sources
```

## Environment

* Operating System: OS X High Sierra 10.13.1
* Truffle version: v4.0.1
* Ethereum client: Ganache I suppose?
* node version: v8.0.0
* npm version: 5.5.1

CruzMolina:
Currently `truffle test` turns `import_paths` of solidity unit tests into relative file basename paths.

Instead, it should use the full `test_file_path` like the rest of truffle to accurately resolve `import_paths` and return contract artifacts.

Potentially resolves #700, #450, & #1795.

### DIFF
--- a/packages/truffle-core/lib/test.js
+++ b/packages/truffle-core/lib/test.js
@@ -108,8 +108,7 @@ var Test = {
         dependency_paths = paths;
 
         testContracts = sol_tests.map(function(test_file_path) {
-          var built_name = "./" + path.basename(test_file_path);
-          return test_resolver.require(built_name);
+          return test_resolver.require(test_file_path);
         });
 
         runner = new TestRunner(config);


RickGriff:
Nice fix OP, thank you. It's good to have the option to use custom filenames for .sol test files.

PR #1880, opened 2019-04-04
Contract and contract constructor artifacts

fainashalts:


packages/truffle-db/src/artifacts/schema.ts
```diff
@@ -75,5 +96,61 @@ export const schema = mergeSchemas({
       },
       
     },
+
+    Contract: {
```
fainashalts: @gnidan below lies the heart of my problems right now. I tried a few different things here as far as fragments, etc. and I can't seem to get ast (actually now its shows up as {} but I believe it should be populated) or source data for the constructor).

packages/truffle-db/src/artifacts/schema.ts
```diff
@@ -75,5 +96,61 @@ export const schema = mergeSchemas({
...
+      name: {
+        fragment: "... on ContractObject { name: contractName }"
+      }, 
+      sourceContract: {
+        fragment: 
+        `... on ContractObject { 
+          ast
```
gnidan: The fragment here needs to include `source { contents, sourcePath }`

packages/truffle-db/src/artifacts/schema.ts
```diff
@@ -75,5 +96,61 @@ export const schema = mergeSchemas({
...
+        }
+      }, 
+    },
+
+    ContractConstructor: {
+      contract: {
+        fragment: 
+        `... on ContractObject { 
+          contents,
+          sourcePath,
```
gnidan: `contents` and `sourcePath` need to get grouped into `source`

packages/truffle-db/src/artifacts/schema.ts
```diff
@@ -75,5 +96,61 @@ export const schema = mergeSchemas({
...
+          ast
+        }`,
+        resolve: (obj, contents, sourcePath, ast) => {
```
gnidan: I think arguments are wrong here - I'd expect all the fields to be in `obj`, then on the next line we would do the destructuring.

packages/truffle-db/src/artifacts/schema.ts
gnidan: Oh we also probably need to add `name: contractName` as well, since it's read from `obj`.

packages/truffle-db/src/artifacts/schema.ts
```diff
@@ -75,5 +96,61 @@ export const schema = mergeSchemas({
...
+          const { name } = obj;
+          const contract = {
+            name: name,
+            sourceContract: {
+              name: name,
```
gnidan: Definitely prefer this new syntax wherever we can:

```suggestion
              name,
```

packages/truffle-db/src/artifacts/schema.ts
gnidan: Also likely `name: contractName`

packages/truffle-db/src/test/index.ts
```diff
@@ -0,0 +1,134 @@
...
+  const { contractConstructor } = artifacts;
+  expect(contractConstructor).toHaveProperty("contract");
+  expect(contractConstructor).toHaveProperty("createBytecode");
+
+  const { contract } = contractConstructor;
+  expect(contract).toHaveProperty("name");
+  expect(contract).toHaveProperty("abi");
+  expect(contract).toHaveProperty("sourceContract");
+});
+
```
gnidan: Could we move this file to `src/artifacts/test`?

packages/truffle-db/src/artifacts/schema.ts
fainashalts: Yeah I had it as obj before, will change it back!

packages/truffle-db/src/test/index.ts
fainashalts: done!

packages/truffle-db/src/artifacts/schema.ts
gnidan: Should be okay now I think

Issue #1317, opened 2018-10-03
Debugger: Move breakpoint down if the current line is a declaration


seesemichaelj:
Breaking/stepping along variable declarations is natural, but if a breakpoint is set on one of these lines that is being skipped, the breakpoint is currently being skipped.

We should do what most normal debuggers do: move the breakpoint down to the next valid non-skipped line (or AST).

seesemichaelj:
@haltman-at or @gnidan is this still an issue?

haltman-at:
Indeed, this remains unfixed.  I'll reopen it.

haltman-at:
OK, I've created PR #1882 to address this issue (at least partially).  It's not perfect, but I think it should be good enough for the practical cases discussed here.  Note that it only alters line-based breakpoints; if you manually advance onto a normally-skipped node with `;`, and then place a node-based breakpoint with a bare `b`, sorry, that one's not getting adjusted (at least not as the PR is currently written).  But I don't expect that to come up too often, so I thought it was OK to ignore that case.

haltman-at:
This has now been merged and released, so I'm closing it.  Let me know if it's not sufficient.

PR #1882, opened 2019-04-06
Move down breakpoints when placed on empty/skipped lines

haltman-at:
This PR addresses issue #1317, at least partially.  It introduces a selector, `controller.breakpoints.resolver`, that, given a breakpoint, returns where it should be moved down to.  The CLI now runs all breakpoints through `controller.breakpoints.resolver` before adding them (and reports back to the user where the breakpoint was *actually* added).  Note that the resolver may return `null` if there is nowhere suitable below to place a breakpoint; in this case the CLI will not add a breakpoint.

Note that the resolver does not work perfectly -- it's something of a hack -- but it should be good enough for common cases; see below for more explanation.  Also note that attempts to *remove* breakpoints are not run through the resolver.  But that's OK, because (in addition to the messages above), **also** added in this PR is (finally) a way to view all breakpoints!  Now, typing `?` will print not only all watch expressions, but also all breakpoints.

(For node-based breakpoints, the location will simply be reported as "a point in line [number]" (or "a point in line [number] in [file]" if it's in a different file), unless you're currently standing on the breakpoint, in which case it will be reported as "*this* point in line n" (emphasis added).)

One more change in this PR: I decided that the selector `controller.finished` should probably really be `controller.current.finished`.  Of course, if this is considered too much of a breaking change, that'll be an easy change to revert!  Let me know if I should.

So, how's the resolver work?

Well, for a node-based breakpoint, it simply returns the breakpoint that was entered.  (Yes, even if you're currently standing on a skipped node, due to use of `;`; perhaps such breakpoints should also be moved down, but I'm assuming that that case just isn't going to happen that often.)  For line-based breakpoints, it checks whether there any nodes starting on that line that it judges to be non-skipped.  If not, it moves it down a line and repeats, until it either finds a place to put it or runs out of lines.  (Note we go by what line the node *starts* on because that's what `continueUntilBreakpoint` goes by.)

This leaves two questions: How does it determine what nodes start on that line, and how does it determine whether a node is skipped or not?

For the first question, I altered `ast/map.js` a bit.  I took the bit where it finds which nodes overlap the range and factored that out into its own function, `findOverlappingRange`, which `findRange` now calls.  I then added `anyNonSkippedInRange`, which calls `findOverlappingRange` (here the range passed in is a full line), and checks if any of the results *start* in that range and are non-skipped.

Here's the hacky part -- what counts as a skipped node?  We can't really determine this with perfect reliability; for instance, references to constant state variables are often skipped, but I didn't think it was worth it to include that case here (especially as that's unlikely to be a full line); and certain type conversions are skipped, but it was *definitely* not worth it to include that (determining whether we're dealing with a skipped or non-skipped type conversion would just be a pain).

And what about, say, pragmas?  Surely one shouldn't be able to place a breakpoint on a line with only a pragma!  But on the other hand, probably nobody's going to attempt to do that.  So I didn't bother marking those as skipped.  One could go on in that vein -- event declarations? custom type declarations? -- but again, not worth it.

So what *did* I mark as skipped?  I decided to mark as skipped:
1. The node types we deliberately skip over: `ContractDefinition` and `VariableDeclaration`.
2. Anything else that was necessary to make skipping over variable declarations work, which meant `VariableDeclarationStatement`, and the various `[...]TypeName` nodes (and `Mapping`, because a mapping type name node is called `Mapping`, not `MappingTypeName`, for whatever reason).
3. String literals, because those are kind of infamous for their skippedness.

Note by the way that the list of deliberately skipped nodes has been moved out of `controller/sagas` and into `helpers` since that's no longer the only place that uses it (it's also been replaced with a function).

Like I said, it's not perfect, but I think it should be good enough for the common cases that originally gave rise to #1317.

(I also added a test.)

### DIFF
--- a/packages/truffle-core/lib/commands/debug.js
+++ b/packages/truffle-core/lib/commands/debug.js
@@ -578,7 +578,8 @@ var command = {
             //want to adjust to make sure we don't set it on an empty line or
             //anything like that
             if (setOrClear) {
-              breakpoint = session.adjustBreakpoint(breakpoint);
+              let resolver = session.view(controller.breakpoints.resolver);
+              breakpoint = resolver(breakpoint);
               //of course, this might result in finding that there's nowhere to
               //add it after that point
               if (breakpoint === null) {
--- a/packages/truffle-core/lib/commands/debug.js
+++ b/packages/truffle-core/lib/commands/debug.js
@@ -246,12 +246,12 @@ var command = {
             let breakpoints = session.view(controller.breakpoints);
             if (breakpoints.length > 0) {
               for (let breakpoint of session.view(controller.breakpoints)) {
-                let currentSourceId = session.view(controller.current.location)
-                  .source.id;
+                let currentLocation = session.view(controller.current.location);
                 let locationMessage = DebugUtils.formatBreakpointLocation(
                   breakpoint,
-                  false,
-                  currentSourceId,
+                  currentLocation.node !== undefined &&
+                    breakpoint.node === currentLocation.node.id,
+                  currentLocation.source.id,
                   sourceNames
                 );
                 config.logger.log("  Breakpoint at " + locationMessage);
--- a/packages/truffle-core/lib/commands/debug.js
+++ b/packages/truffle-core/lib/commands/debug.js
@@ -243,16 +243,21 @@ var command = {
                 })
               )
             );
-            for (breakpoint in session.view(controller.breakpoints)) {
-              let currentSourceId = session.view(controller.current.location)
-                .source.id;
-              let locationMessage = DebugUtils.formatBreakpointLocation(
-                breakpoint,
-                false,
-                currentSourceId,
-                sourceNames
-              );
-              config.logger.log("Breakpoint at " + locationMessage);
+            let breakpoints = session.view(controller.breakpoints);
+            if (breakpoints.length > 0) {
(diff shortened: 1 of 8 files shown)


gnidan:
> I decided that the selector controller.finished should probably really be controller.current.finished

We could always make `controller.finished` an alias. At some point it might be worthwhile to add some kind of way to mark selectors as deprecated in reselect-tree.

packages/truffle-core/lib/commands/debug.js
```diff
@@ -730,6 +771,7 @@ var command = {
                 break;
               case "?":
                 printWatchExpressions();
+                printBreakpoints();
```
gnidan: For future PRs: it's helpful to upload a screenshot of output changes so reviewers don't have to grab the branch locally just to see relevant info. Looks good here though.

haltman-at:
> > I decided that the selector controller.finished should probably really be controller.current.finished
> 
> We could always make `controller.finished` an alias. At some point it might be worthwhile to add some kind of way to mark selectors as deprecated in reselect-tree.

Good point, I'll add that back in.

packages/truffle-core/lib/commands/debug.js
haltman-at: Oh, I see!  Didn't realize I could do that.  Noted.

PR #1892, opened 2019-04-09
Add --dry-run option to migrate

eggplantzzz:


eggplantzzz:
Oh, I misunderstood the point of this option. This flag causes ONLY a dry run to be performed. There is also a flag `--skip-dry-run` which skips it. I guess there is no way to run a dry run and migration with one command in a "non-known network" environment. I'll go ahead and make an update to document both of these correctly.

eggplantzzz:
Closing this and opening PR #1901.

Issue #1895, opened 2019-04-09
Migration against infura erroring during `Reporter.postDeploy()`


adamgall:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue. -> https://gitter.im/ConsenSys/truffle?at=5caccc2225686a7dc301cab4

---------------------------

## Issue

Getting errors thrown from this line when doing migrations and using `HDWalletProvider` connected to infura nodes.

```
2_deploy_contract.js
=========================

   Deploying 'Contract'
   ----------------------------------
   > transaction hash:    0xtxH4sh
   ⠹ Blocks: 0            Seconds: 0TypeError: Cannot read property 'timestamp' of null
    at Reporter.postDeploy (/../node_modules/truffle/build/webpack:/packages/truffle-reporters/reporters/migrations-V5/reporter.js:317:1)
    at processTicksAndRejections (internal/process/next_tick.js:81:5)
   ⠦ Blocks: 0            Seconds: 0
```

Migration hangs at this point.

This was likely introduced in https://github.com/trufflesuite/truffle/pull/1868, and is probably due to infura using an array of endpoint nodes, which might not all have the same block state at any given moment.

## Steps to Reproduce

Perform a migration using `HDWalletProvider`, pointed at a public infura endpoint (such as https://ropsten.infura.io/v3/your-key), and observe the above error. It's sporadic, but has been happening very often for me.

## Expected Behavior

I'd expect that the migration does not fail.

## Actual Results

See error text above :)

## Environment

* Operating System: OSX 10.14.3
* Ethereum client: infura via https://ropsten.infura.io/v3/your-key
* Truffle version (`truffle version`):
```
Truffle v5.0.11 (core: 5.0.11)
Solidity v0.5.0 (solc-js)
Node v11.12.0
Web3.js v1.0.0-beta.37
```
* node version (`node --version`): 11.12.0
* npm version (`npm --version`): 6.9.0

CruzMolina:
Hey @adamgall , literally just noticed this error on my end while looking into other `truffle-hdwallet-provider`-related issues!

Hoping to fix this shortly 🤞

adamgall:
For anyone else out there in the meantime, downgrading to `truffle v5.0.10` allowed me to run migrations against infura without issue.

CruzMolina:
[...]

PR #1899, opened 2019-04-09
Handle `null` getBlock returns during live/testnet migrations

CruzMolina:
As discussed elsewhere (paritytech/parity-ethereum#8788, paritytech/parity-ethereum#6334, INFURA/infura/issues/43) & raised in #1895, sometimes nodes will return `null` during `getBlock` calls (usually when calling recent/latest blocks). Currently, when this happens it completely ruins the deployment experience. This PR catches falsy values & recursively calls `postDeploy` again.


gnidan:
We might need to go a bit further, since there's a potential infinite loop here? But this seems fine to get us started.

PR #1908, opened 2019-04-12
Add decoding support for external function pointers, and improved decoding for contract-type variables

haltman-at:
This PR implements decoding support for variables of `function external` type.  Notably, **this PR contains a breaking change to the contract decoder**, if a minor one -- the `init` function is now `async`.  More on this later.  There's also the non-breaking enhancement that the return value of `state` now includes `code` (as a hex string) alongside `variables`, `balance`, and `nonce`.  (Also, the decoding for variables of contract type has been improved... but this also means it's been *changed*.  More on this later.)

So, first let's discuss what external function pointers are decoded to, and what variables of contract type are decoded to, and then afterward we'll discuss how it works.

First, variables of contract type.  Previously, these just decoded as their addresses.  Now, they will decode to a string `Class(address)`, where `Class` is the contract's class and `address` is its address.  Note that when I say the contract's class, I mean the actual class of that particular instantiated contract -- not the type of the variable it's contained in.  Thus, if we write `Base base = new Derived()`, then `base` will decode as a `Derived`, not a `Base`.  We do this because the actual class is helpful information, while the type is not -- of course its type is `Base`, that's what it was declared as!  The user already knows that.  But knowing that this particular `Base` is in fact a `Derived` may be useful.

If the class of the variable cannot be determined, then it will simply decode to the address, as before.

Let's discuss that format that I chose, `Class(address)`.  This isn't really how I'd like to display them, but this was the best compromise I could think of between the two roles that decoded variables serve (standalone display and use in watch expressions).  Once we have the new output format and these uses are separated, then of course it will decode to an appropriate object, and also I'll change the display format to something better (but I'll probably leave this as the decoding in watch expressions -- at least until we redo those, too).

Worth noting is that this string format is meant to *resemble* Solidity, but it's not really quite correct Solidity; after all, in Solidity, `Class(address)` would convert the *type* to `Class`, but obviously it wouldn't change the underlying class (which is impossible).  Like I said, it was a compromise.  If you think it should be different, it could be changed.  (Maybe it could even be an object, like we currently have for enums, but I wanted to avoid that; enums shouldn't really decode that way, all this is just a stopgap until we have the new output format.)

Anyway, that's contract variables, what about function variables?  (External function variables, of course -- internal function variables, when implemented, will use a different format.)  These are decoded to the string `Class(address).functionName`, where `Class` and `address` are as above and `functionName` is the name of the function.  Now, it's possible in Solidity to construct an invalid external function pointer, so we have handling for that case as well.  If the function cannot be recognized, it will decode to `Class(address).call(selector....)`, where `selector` is the function selector (decoded as a `bytes4`).

Again, this is pseudo-Solidity; I won't point out all the various ways this is not correct Solidity.  Like I said, it was a compromise, and it can be changed if necessary.

If the class of the address can't even be recognized, then it will simply decode to `address.call(selector...)`.

OK, so that's what it does.  How does it work?

Well, fundamentally, it's pretty simple.  The `info` object now has a new field, `contexts`; this contains a `DecoderContexts` object.  The content of this is similar to that of the selector `evm.info.contexts`, but modified to be more suitable to the decoder.  In the debugger there is a new selector,`data.views.contexts`, which takes `evm.info.contexts` and transforms it in this way.

Let's skip the details of that for now and discuss function/contract decoding.  First off, the `contract` case has been separated from the `address` case in `decodeValue`, and now dispatches to a seperate function, `decodeContract`.  Meanwhile, a `function` case has been added; it has two subcases, for `external` and `internal` visibility, which dispatch to `decodeExternalFunction` and `decodeInternalFunction`, respectively.  (The latter, for now, simply always returns `undefined`.)  Note that `decodeContract` and `decodeExternalFunction` share a fair bit of code, but it's deliberately not factored for now; I'll redo this when I get to the new output format.

There's also a new special case added to `decodeLiteral` in `stack.ts`; because external function pointers are stored in a different format on the stack than they are elsewhere, when decoding an external function pointer, stack decoding will extract the address and selector itself and then dispatch *directly* to `decodeExternalFunction`, rather than going through `decodeValue`.  (This is one of only two places in the decoder that skip over `decodeValue` like this; the other is in `decodeConstant`, where literals of type `bytesN` are handled manually rather than passed to `decodeValue`.)

Sow how do `decodeContract` and `decodeExternalFunction` work?  Well, they start by `yield`ing a request for the code at that address.  Once they receive the response with the code, they check it against the `contexts` object to determine which context it is.  (If none can be determined, that case is handled as mentioned above.)  For `decodeContract`, that's basically it.  For `decodeExternalFunction`, it then checks the signature against the context's `abi` object to determine the function name.  Again, if no match is found, this case is handled as mentioned above.

For consistency, the response to a code request is given in the form of a `Uint8Array`, even though in reality that means it's likely going to get round-trip converted from text to `Uint8Array` and back (with the current context format, we need text to perform the match).  I don't think this is a big deal.

So how do the debugger and the contract-decoder handle these requests?  Let's start with the debugger.  There's now a new selector, `data.views.instances`, which contains a slightly-transformed version of `evm.info.instances` (irrelevant information is stripped out and binaries are stored as `Uint8Array`s).  When it receives a request for the code at a given address, first it checks if the address is already known.  If so, it simply responds with the known binary for that address.  If it's not already known, it calls the web3 saga `obtainBinaries` (with, yes, a single-element array for the argument) to get the binary, and then calls the evm saga `addInstance` so that this will be cached for the future.  (There's actually a third special case, but -- well, you'll read about it shortly.)

This approach does have one little quirk... the selectors `evm.info.contexts` and `evm.info.instances` -- and, indeed, the reducers that back them -- are now somewhat misnamed, as they can now have additional information added during the middle of a session.  I don't think this is too bad, because it's not *changing* information, but it is worth noting.  In particular, since the CLI displays the affected instances again when the reset command is entered, upon a reset one may see more affected instances than displayed initially!

It's because of this quirk that I added that third special case.  You see, I didn't want the zero address constantly showing up in "addresses affected" and causing a "could not find source" warning, just because someone decoded a contract or external function variable that hadn't been initialized yet.  So actually, if the debugger receives a request for the code at the zero address, it will simply respond with an empty code response, and not call `evm.addInstance` on the zero address.

By the way, I've made an alteration to `web3.obtainBinaries`, as well as the other sagas and adapter methods that back it -- they now accept a block argument, and when the decoder obtains binaries, both at startup and in the `decode` saga, it now does so passing in the transaction's block number for the `block` argument.  This is so that, in case one of these contracts self-destructed in a later block, we'll still be able to get its code.  (Of course, if it self-destructed later in this same block, or in this very transaction, we're still out of luck.)

Anyway, how does the contract decoder respond to code requests? It just makes a web3 request, converts to `Uint8Array`, and responds with that. It mostly doesn't cache anything; maybe it should (it could cache storage requests as well). (One little complication there is that people could ask for different block numbers, or they could use special strings to pick out variable blocks, so that would have to be accounted for somehow.) The one exception is that the contract decoder now requests the contract's code during `init` and stores it. (This is why `init` is now `async`.) If it receives any requests for the code at the contract's own address, it will respond with this stored code. (And yes, that's regardless of what block the user asked for, but, well, it's fairly obvious why ignoring the block number shouldn't matter in this case, because presumably they're checking out a block where the contract exists. If I'm wrong about this and Ganache starts getting complaints about it, well, we can revisit this later.)

EDIT: For a while I had a version up that *did* include caching, but it didn't work for some reason.  I guess I'll have to revisit this later; I don't really want to figure out right now why the cache wasn't working, sorry.  So for now I've just removed the cache in favor of the older version.  (Technically, one version of the caching version passed Travis, but since with any minor change it didn't, I'm worried this was spurious.)

By the way, I have to admit I'm not very clear on what the difference is supposed to be between the constructor and the `init` function.  Some of the new startup stuff I added I put in the constructor, some in the `init` function, but it was kind of arbitrary.  Hopefully it makes sense?

(Also, the actual real reason I stored the code was in preparation for internal function decoding later, but I think now that it may be the case that the contract decoder may not be able to support that right now and it may have to be a debugger-only feature. Oh well. Anyway, back to external functions.)

Now, this isn't the only new thing the contract decoder does on startup, because it also has to set up the `contexts` object.  This is a `DecoderContexts` object, not a `DebuggerContexts` object.  Due to the link reference format long name problem, though, it does still have to go through the same normalizaiton process, so the `NORMALIZE_CONTEXTS` case of the `contexts` reducer has been factored out into its own function in `truffle-decode-utils`, `normalizeContexts`.  This lives in `contexts.ts`, a new file in `truffle-decode-utils` that contains various tools for working with contexts that used to live in the debugger.  For instance, the context matching functions are now here.  It also contains the relevant type definitions, both for the decoder's contexts format and for the debuggers (which didn't *really* need it, because TypeScript can't actually check types at runtime and if we handed it something from JS that isn't in the correct format it would just chug along regardless, but I figured I ought to all the same).

So how does the decoder context format differ from the debugger context format?  Well:

1. Only non-constructor (i.e. deployed) contexts are included.
2. Contexts are indexed by contract ID, rather than by a hash.
3. Information irrelevant to the decoder is stripped out.  (Though I left in some stuff that the decoder doesn't presently use, but might in the future.)
4. The format of the `abi` object is different.

OK, how is the format of the `abi` object different?  Well:

1. Only entries of type `function` -- not `event`, `fallback`, or `constructor` -- are included.
2. Each entry is augmented with a `signature` field, containing the function's selector.
3. Rather than just being an array, `abi` is now an object, indexed by these signatures.

Thus, when the decoder wants to look up the function with a given signature for a given class, it can simply index into the `abi` object.  In order to compute this transformation, `contexts.ts` contains a function `abiToFunctionAbiWithAignatures`.  Note that since the entries of the `abi` object may *already* contain a signature, the function first checks for this and uses it if it exists.  Otherwise, it computes it with the `web3-eth-abi` function `encodeFunctionSignature`.

(There's some web3 version nastiness going on here.  It seems web3 beta 37 doesn't allow one to easily allow one to use the `web3-eth-abi` functions standalone, or with TypeScript, so I installed a later version of that specific `web3` subpackage.  I did, however, lock it to a specific version to ensure that any future web3 updates don't break anything.)

One last note: The TypeScript parts of this PR contain unfortunately contain a fair bit of nasty type coercion that I couldn't find any better alternative to.  This is for two reasons:

1. The `normalizeContexts` function has to operate both on debugger contexts and on decoder contexts, returning the appropriate type in each case... but the contract decoder doesn't know that it does that, so it needs to coerce its output to the `DecoderContexts` type.  (Again, I *could* have just submitted badly typed stuff from the debugger and not worried about it, but that seemed even worse.)
2. There are *three* different type definitions for an ABI item floating around, all trying to describe the same thing but none quite compatible, that I had to coerce between.  There's the one we have in our spec; there's the one in `web3`; and there's my own.  Maybe that last one could've been left out, IDK.  There still would've been nasty type coercion.

Oh well.  It all works, ultimately!  I also added some tests, both a new test in the debugger, and expanding the test of the contract decoder.

So that's decoding for external functions.  Next up: Decoding for internal functions.  (At least for the debugger.  For the contract decoder, it may have to wait.)  And then we'll have decoding for *every* data type!  (Unless you count constant state variables as a data type, because we'll still only be able to decode some of those.)


haltman-at:
Additional comment: I forgot to mention above that this PR also fixes a bug in the main data saga that I hadn't noticed before (because it wasn't important before).  Namely, local variables of function type would fail to appear in `data.proc.assignments`.

The reason for this, it seems, is that for whatever reason, when the debugger hit a declaration of such a variable, it would not land on the `VariableDeclaration` node as expected, but rather on the `FunctionTypeName` node.

To fix this, I added a `FunctionTypeName` case to the main data saga.  It replaces the value of `node` with the value of the parent node; just to be certain, it checks that this is a `VariableDeclaration` node and breaks out if it isn't.  If it it is, though, it then does a deliberate fall-through to the `VariableDeclaration` case, which handles the rest as usual.  Yes, I've marked this explicitly with a big all-caps comment. :P

Anyway, it's something of a hack, but it works.

haltman-at:
OK, you know what, I went ahead and added caching in the contract decoder after all (figuring out the block number problem actually wasn't too hard).  But I don't have time to write it up right now, so I'll edit the initial writeup later.

packages/truffle-debugger/lib/data/sagas/index.js
```diff
@@ -72,6 +80,25 @@ export function* decode(definition, ref) {
...
+          response = NO_CODE;
+        } else {
+          //I don't want to write a new web3 saga, so let's just use
+          //obtainBinaries with a one-element array
+          debug("fetching binary");
+          let binary = (yield* web3.obtainBinaries([address], blockNumber))[0];
+          debug("adding instance");
+          yield* evm.addInstance(address, binary);
+          response = DecodeUtils.Conversion.toBytes(binary);
+        }
```
gnidan: missing `break;`

packages/truffle-debugger/lib/data/sagas/index.js
```diff
@@ -217,6 +244,19 @@ function* variablesAndMappingsSaga() {
...
+      //we land on the FunctionTypeName instead of the VariableDeclaration,
+      //so we replace the node with its parent (the VariableDeclaration)
+      node = scopes[scopes[node.id].parentId].definition;
+      //let's do a quick check that it *is* a VariableDeclaration before
+      //continuing
+      if (node.nodeType !== "VariableDeclaration") {
+        break;
+      }
+    //otherwise, deliberately fall through to the VariableDeclaration case
+    //NOTE: DELIBERATE FALL-THROUGH
```
gnidan: 👌

packages/truffle-debugger/lib/data/sagas/index.js
haltman-at: Ooh, missed that, thanks

haltman-at: Also of course it's immediately useful, this is *why* we switched to that, remember? :P

gnidan: but it's still really neat!

PR #1915, opened 2019-04-15
Remove solc-js as an NPM dependency

CruzMolina:
PR removes `solc` as a separate bundle dep. `solc` is already bundled within `truffle-compile`.


eggplantzzz:
Yeah, hmm it doesn't look like this package is being used in `packages/truffle`

PR #1916, opened 2019-04-15
Update workspace new data model

fainashalts:
This PR updates the truffle-db workspace to match the truffle-db data model. Specifically, contract, compilation, and contractConstructor were changed to represent the new data model correctly.

### DIFF
--- a/packages/truffle-db/src/workspace/pouch.ts
+++ b/packages/truffle-db/src/workspace/pouch.ts
@@ -139,7 +139,7 @@ export class Workspace {
          const sourceIds = sources.map(source => source.id);
          const sourcesObject = Object.assign({}, sourceIds);
 
-         const id = soliditySha3(compiler.id, sourcesObject);
+         const id = soliditySha3(jsonStableStringify(compiler, sourcesObject));
 
          const compilation = await this.compilation({ id }) || { ...compilationInput, id };
  
--- a/packages/truffle-db/src/workspace/pouch.ts
+++ b/packages/truffle-db/src/workspace/pouch.ts
@@ -91,7 +91,13 @@ export class Workspace {
     return {
       contracts: Promise.all(contracts.map(
         async (contractInput) => {
-          const { name, abi, compilation, sourceContract, constructor: contractConstructor } = contractInput;
+          const {
+            name,
+            abi,
+            compilation,
+            sourceContract,
+            constructor: contractConstructor
+          } = contractInput;
           const id = soliditySha3(jsonStableStringify(name, abi, sourceContract, compilation));
 
           const contract = await this.contract( { id } );
--- a/packages/truffle-db/src/workspace/pouch.ts
+++ b/packages/truffle-db/src/workspace/pouch.ts
@@ -4,6 +4,8 @@ import PouchDBFind from "pouchdb-find";
 
 import { soliditySha3 } from "web3-utils";
 
+const jsonStableStringify = require('json-stable-stringify');
+
 const resources = {
   contracts: {
     createIndexes: [
@@ -90,8 +92,8 @@ export class Workspace {
       contracts: Promise.all(contracts.map(
         async (contractInput) => {
           const { name, abi, compilation, sourceContract, contractConstructor } = contractInput;
-          //leaving this id as just a hash of the name for now, until we figure out what it should be
-          const id = soliditySha3(name);
+          const id = soliditySha3(jsonStableStringify(name, abi, sourceContract, compilation)); 
+         
           const contract = await this.contract( { id } );
           
           if(contract) {
--- a/packages/truffle-db/src/workspace/pouch.ts
+++ b/packages/truffle-db/src/workspace/pouch.ts
@@ -98,14 +98,13 @@ export class Workspace {
             sourceContract,
             constructor: contractConstructor
           } = contractInput;
-          const id = soliditySha3(jsonStableStringify({ name, abi, sourceContract, compilation }));
-
+          const id = soliditySha3(jsonStableStringify({ name: name, abi: abi, sourceContract: sourceContract, compilation: compilation }));
           const contract = await this.contract( { id } );
 
           if(contract) {
             return contract;
           } else {
-            await this.contracts.put({
+            const contractAdded = await this.contracts.put({
             ...contractInput,
             _id: id,
             });
@@ -145,7 +144,7 @@ export class Workspace {
          const sourceIds = sources.map(source => source.id);
          const sourcesObject = Object.assign({}, sourceIds);
 
-         const id = soliditySha3(jsonStableStringify(compiler, sourcesObject));
+         const id = soliditySha3(jsonStableStringify({ compiler: compiler, sourceIds: sources } ));
 
          const compilation = await this.compilation({ id }) || { ...compilationInput, id };
 
--- a/packages/truffle-db/src/workspace/pouch.ts
+++ b/packages/truffle-db/src/workspace/pouch.ts
@@ -73,7 +73,7 @@ export class Workspace {
 
     try {
       const result = {
-        ...await this.contracts.get(id), 
+        ...await this.contracts.get(id),
 
         id
       }
@@ -87,7 +87,7 @@ export class Workspace {
     await this.ready;
 
     const { contracts } = input;
-    
+
     return {
       contracts: Promise.all(contracts.map(
         async (contractInput) => {
@@ -95,12 +95,12 @@ export class Workspace {
           const id = soliditySha3(jsonStableStringify(name, abi, sourceContract, compilation)); 
          
           const contract = await this.contract( { id } );
-          
+
           if(contract) {
             return contract;
           } else {
             await this.contracts.put({
-            ...contractInput, 
+            ...contractInput,
             _id: id,
             });
            
@@ -142,7 +142,7 @@ export class Workspace {
          const id = soliditySha3(jsonStableStringify(compiler, sourcesObject));
 
          const compilation = await this.compilation({ id }) || { ...compilationInput, id };
- 
+
           await this.compilations.put({
             ...compilation,
             ...compilationInput,
--- a/packages/truffle-db/src/workspace/pouch.ts
+++ b/packages/truffle-db/src/workspace/pouch.ts
@@ -98,7 +98,7 @@ export class Workspace {
             sourceContract,
             constructor: contractConstructor
           } = contractInput;
-          const id = soliditySha3(jsonStableStringify(name, abi, sourceContract, compilation));
+          const id = soliditySha3(jsonStableStringify({ name, abi, sourceContract, compilation }));
 
           const contract = await this.contract( { id } );
 
--- a/packages/truffle-db/src/workspace/pouch.ts
+++ b/packages/truffle-db/src/workspace/pouch.ts
@@ -138,8 +138,9 @@ export class Workspace {
     return {
       contractConstructors: Promise.all(contractConstructors.map(
         async (contractConstructorInput) => {
-          const { abi, compilation, createBytecode,linkValues, contract  } = contractConstructorInput;
-          const id = abi !== undefined? soliditySha3(abi, createBytecode.id) : soliditySha3(createBytecode.id);
+          const { createBytecode, contract  } = contractConstructorInput;
+          //just using createBytecode id for now until we figure out what the id should be comprised ofwork
+          const id = soliditySha3(createBytecode.id);
          
           const contractConstructor = await this.contractConstructor( { id } );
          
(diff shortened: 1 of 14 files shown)


packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -313,11 +358,11 @@ it("adds compilation", async () => {
     compilerName: Migrations.compiler.name,
     compilerVersion: Migrations.compiler.version,
     sourceId: soliditySha3(Migrations.source, Migrations.sourcePath),
-    id: "0xfd63faebf4e7515b2f48fdf22066869e09df9b2e91f4e0cf88d7d14ec496b515", 
+    id: soliditySha3(jsonStableStringify(Migrations.compiler, soliditySha3(Migrations.source, Migrations.sourcePath))), 
     contractId: soliditySha3(Migrations.contractName, soliditySha3(Migrations.source, Migrations.sourcePath))
   }
 
-  const compilationId = soliditySha3(variables.contractId, variables.sourceId)
+  const compilationId = soliditySha3(jsonStableStringify(Migrations.compiler, soliditySha3(Migrations.source, Migrations.sourcePath)))
```
gnidan: ID calculations getting outta hand!

Some thoughts... firstly, it feels cleaner to add wrapping braces to the IDs, a là
```suggestion
  const compilationId = soliditySha3(jsonStableStringify(Migrations.compiler, soliditySha3(Migrations.source, Migrations.sourcePath)}))
```

But moreover, it gets rather brittle to have these ID re-computations. I think we should only compute ID when that's the thing we're trying to test. So if we're testing `sourcesAdd`, then we should test that the source ID matches our expectation.

Beyond that (e.g. to avoid needing to compute compilation IDs inside the contract tests), I think we should work on getting helper methods in place so it becomes easier to run pre-cursor mutations and get the IDs that way. If you're into the work I did in #1918, we can merge that and then start looking at using `describe()` blocks with `before()` hooks to add these requisite dependent resources (e.g.. `describe("Compilation", ...)` could have `before("adds sources", () => { })`, etc.). What do you think?

packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -90,68 +87,24 @@ export class Workspace {
...
     const { contracts } = input;
-    
+
     return {
       contracts: Promise.all(contracts.map(
         async (contractInput) => {
-          const { name, source } = contractInput;
-          const id = name !== undefined ? soliditySha3(name, source.id) : soliditySha3(source.id);
+          const { name, abi, compilation, sourceContract, constructor: contractConstructor } = contractInput;
+          const id = soliditySha3(jsonStableStringify(name, abi, sourceContract, compilation));
```
gnidan: (Also noted this in the tests, but copying here in case it's easy to miss below)

I think we should wrap the IDs in objects. Here that'd be:
```suggestion
          const id = soliditySha3(jsonStableStringify({ name, abi, sourceContract, compilation }));
```

packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -90,68 +87,24 @@ export class Workspace {
...
 
     const { contracts } = input;
-    
+
     return {
       contracts: Promise.all(contracts.map(
         async (contractInput) => {
-          const { name, source } = contractInput;
-          const id = name !== undefined ? soliditySha3(name, source.id) : soliditySha3(source.id);
+          const { name, abi, compilation, sourceContract, constructor: contractConstructor } = contractInput;
```
gnidan: Stylistic note: until/unless we turn on prettier for TypeScript, I worry that it might get more difficult to read this code with long lines. General style rule of thumb here is to put each variable on its own line:
```suggestion
          const {
            name,
            abi,
            compilation,
            sourceContract,
            constructor: contractConstructor
          } = contractInput;
```

packages/truffle-db/src/workspace/pouch.ts
fainashalts: Is this stylistic, or is there a functional reason to do this? Fine with it, just want to be sure I understand the reasoning!

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -14,27 +18,74 @@ const fixturesDirectory = path.join(
...
+    sourceIds: [{ id: sourceId }] 
+  }));
+  contractId = soliditySha3(jsonStableStringify({ 
+    name: Migrations.contractName, 
+    abi: { json: JSON.stringify(Migrations.abi) } , 
+    sourceContract: { index: 0 } ,
+    compilation: { id: compilationId }
+  }));
+
+});
```
gnidan: Maybe instead of a `beforeAll` block, we could just make a helper function, something like:

```javascript
const generateId = (obj) => soliditySha3(jsonStableStringify(obj))
```

I mention this because I prefer not to generate the IDs outside the tests. I'll expand on this below per-test.

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -57,66 +109,64 @@ mutation AddSource($contents: String!, $sourcePath: String) {
...
+describe("Source", () => {
+  it("adds source", async () => {
+    const client = new WorkspaceClient();
 
-    const { data } = result;
-    expect(data).toHaveProperty("sourcesAdd");
+    const variables = {
+      contents: Migrations.source,
+      sourcePath: Migrations.sourcePath,
+      id: soliditySha3(Migrations.source, Migrations.sourcePath)
```
gnidan: Let's take this out of `variables` and make it a separate variable, something like:

```javascript
const expectedId = generateId({
  source: Migrations.source,
  sourcePath: Migrations.sourcePath
});
```

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -57,66 +109,64 @@ mutation AddSource($contents: String!, $sourcePath: String) {
...
+      const source = sources[0];
+      expect(source).toHaveProperty("id");
 
-  // ensure retrieved as matching
-  {
-    const result = await graphql(schema, GetSource, null, { workspace }, {
-      id: variables.id
-    });
+      const { id } = source;
+      expect(id).toEqual(variables.id);
```
gnidan: The idea being that we'd now just `expect(id).toEqual(expectedId)`

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -57,66 +109,64 @@ mutation AddSource($contents: String!, $sourcePath: String) {
...
+    }
 
-    const { data } = result;
-    expect(data).toHaveProperty("source");
+    // ensure retrieved as matching
+    {
+      const data = await client.execute(GetSource, { id: variables.id });
```
gnidan: And we look up by expected ID, etc.

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -129,87 +179,53 @@ mutation AddBytecode($bytes: Bytes!) {
...
-      }
-    );
-
-    const { data } = result;
-    expect(data).toHaveProperty("bytecodesAdd");
+describe("Bytecode", () => {
+  it("adds bytecode", async () => {
+    const client = new WorkspaceClient();
+    const variables = {
+      id: soliditySha3(Migrations.bytecode),
```
gnidan: Same here, `const expectedId = generateId({ bytes: Migrations.bytecode })`

Note that I'm implicitly standardizing on using objects for **all** IDs

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -249,274 +276,221 @@ mutation AddCompilation($compilerName: String!, $compilerVersion: String!, $sour
...
-    expect(contractsAdd).toHaveProperty("contracts");
+describe("Compilation", () => {
+  it("adds compilation", async () => {
+    const client = new WorkspaceClient();
+
+    const variables = {
+      compilerName: Migrations.compiler.name,
+      compilerVersion: Migrations.compiler.version,
+      sourceId: sourceId,
+      id: compilationId,
```
gnidan: And here, `compilationId` would be `expectedCompilationId` instead of a variable.

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -249,274 +276,221 @@ mutation AddCompilation($compilerName: String!, $compilerVersion: String!, $sour
...
+      contractId: contractId,
```
gnidan: `contractId` isn't used here it seems, I think we can get rid of it from this list.

packages/truffle-db/src/workspace/test/index.ts
gnidan: And instead of passing pre-generated `sourceId`, I think we should add a `before("add source", ...)` block inside `describe("Compilation", ...)`, where we assign `sourceId` to the _actual_ ID that gets generated for the source, rather than pre-generating it. Since we're testing Compilation here, not Source.

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -249,274 +276,221 @@ mutation AddCompilation($compilerName: String!, $compilerVersion: String!, $sour
...
+describe("Contract", () => {
+  const client = new WorkspaceClient();
 
   const variables = {
-    bytecodeId: soliditySha3(Migrations.bytecode), 
-    compilationId: "0xfd63faebf4e7515b2f48fdf22066869e09df9b2e91f4e0cf88d7d14ec496b515",
-    contractId: soliditySha3(Migrations.contractName, soliditySha3(Migrations.source, Migrations.sourcePath))
+    contractName: Migrations.contractName,
+    compilationId: compilationId,
+    id: contractId,
```
gnidan: In this test, `contractId` is the "expected" ID, so let's generate that in the describe block directly, and take it out of the variables here.

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -249,274 +276,221 @@ mutation AddCompilation($compilerName: String!, $compilerVersion: String!, $sour
...
-    const { contractConstructorsAdd } = data;
-    expect(contractConstructorsAdd).toHaveProperty("contractConstructors");
-
-    const { contractConstructors } = contractConstructorsAdd;
-
-    expect(contractConstructors).toHaveLength(1);
-
-    const { id } = contractConstructors[0];
-    expect(id).toEqual(constructorId);
+  beforeAll(async () => {
```
gnidan: Let's just make this `before` rather than `beforeAll`, so that it runs before each `it`, just in case.

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -249,274 +276,221 @@ mutation AddCompilation($compilerName: String!, $compilerVersion: String!, $sour
...
-    bytecodeId: soliditySha3(Migrations.bytecode), 
-    compilationId: "0xfd63faebf4e7515b2f48fdf22066869e09df9b2e91f4e0cf88d7d14ec496b515",
-    contractId: soliditySha3(Migrations.contractName, soliditySha3(Migrations.source, Migrations.sourcePath))
+    contractName: Migrations.contractName,
+    compilationId: compilationId,
+    id: contractId,
+    compilerName: Migrations.compiler.name,
+    compilerVersion: Migrations.compiler.version,
+    sourceId: sourceId,
+    bytecodeId: bytecodeId,
```
gnidan: For `compilationId`, `sourceId`, and `bytecodeId`, let's avoid calculating them for this test, and instead add source and bytecode manually to the before block and read the IDs from the result.

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -249,274 +276,221 @@ mutation AddCompilation($compilerName: String!, $compilerVersion: String!, $sour
...
+    const variables = {
+      contractName: Migrations.contractName,
+      compilationId: compilationId,
+      id: contractId,
+      compilerName: Migrations.compiler.name,
+      compilerVersion: Migrations.compiler.version,
+      sourceId: sourceId,
+      bytecodeId: bytecodeId, 
+      abi: JSON.stringify(Migrations.abi)
+    }
```
gnidan: (This overwrites `variables` from the describe block... I think we should get rid of one of these declarations, either this one or the one above.)

packages/truffle-db/src/workspace/test/index.ts
fainashalts: I see that! I'll change it for sourceId and bytecodeId.

packages/truffle-db/src/workspace/test/index.ts
fainashalts: hey @gnidan I don't see a `before()` in the Jest documentation -- just `beforeEach` and `beforeAll`. Maybe you mean I should change this to `beforeEach`, which runs before each of the tests, while `beforeAll` just runs once?

gnidan: Ah, yep, `beforeEach`!

gnidan:
@fainashalts nice job with the changes! what do you think? testing approach starting to make sense? feel free to hit the "ready for review" button and I'll give it a once-over!

PR #1918, opened 2019-04-16
Refactor workspace tests and get them passing

gnidan:
This PR introduces a `WorkspaceClient` helper class for the workspace tests, to make it easier to issue more than one request to graphql in a single test. I added this since I discovered the problem with the tests on the target branch (#1916): compilation wasn't getting added before contract, so the lookup was failing.

This also includes the addition of the `graphql-tag` package, to enable `gql` template strings, e.g.:
```javascript
const GetContractNames = gql`query GetContractNames { contractNames }`
```

Refactoring changes:
- Re-order tests so they're organized by resource type.
- Define WorkspaceClient class to abstract the graphql inner workings
- Start using gql` ... ` tag for test queries
- Switch from using `graphql()` to using graphql's `execute()` function,
  since the former only takes a string, and the tag parses automatically
- Replace test Workspace instances with WorkspaceClient instances, and
  use the client method instead of running queries manually

Changes to fix the tests:

- In "adds contracts" tests, run the AddCompilation mutation as a precursor
- Add sourceContract to GetContract query
- Add abi to GetContractQuery
- Add ast to AddContracts mutation payload request
- Uncomment now-working assertions
- Remove property check for `compilation`, since it's not in the query

### DIFF
--- a/packages/truffle-db/src/workspace/test/index.ts
+++ b/packages/truffle-db/src/workspace/test/index.ts
@@ -19,6 +19,10 @@ const fixturesDirectory = path.join(
 
 const Migrations = require(path.join(fixturesDirectory, "Migrations.json"));
 
+/*
+ * root
+ */
+
 const GetContractNames = `
 query GetContractNames {
   contractNames
@@ -36,6 +40,11 @@ it("queries contract names", async () => {
   const { contractNames } = data;
   expect(contractNames).toEqual([]);
 });
+
+
+/*
+ * Source
+ */
 
 const GetSource = `
 query GetSource($id: ID!) {
@@ -111,6 +120,11 @@ it("adds source", async () => {
   }
 });
 
+
+/*
+ * Bytecode
+ */
+
 const GetBytecode = `
 query GetBytecode($id: ID!) {
   bytecode(id: $id) {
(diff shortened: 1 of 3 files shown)


packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -17,27 +18,53 @@ const fixturesDirectory = path.join(
...
+  async execute (request, variables = {}) {
+    const result = await graphql.execute(
+      schema,
+      request,
+      null, // root object, managed by workspace
+      { workspace: this.workspace }, // context vars
+      variables
+    );
+
+    return result.data;
```
gnidan: I'm doing this for convenience but it might be a bad idea because we're going to want to test for errors eventually. I figure we can always change it later, since it won't be _too bad_ to add all the extra destructuring back in, but feel free to disagree :)

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -1,6 +1,7 @@
 import path from "path";
 
-import { graphql } from "graphql";
+import gql from "graphql-tag";
```
fainashalts: So if I understand right, the benefit of using `graphql-tag` is that it will parse our query strings into the standard graphQL AST? We were already using template literals, right? So now we just prefix with `gql`?

packages/truffle-db/src/workspace/test/index.ts
fainashalts: I'm ok with leaving it like this for now!

PR #1921, opened 2019-04-17
Add decoding for internal functions (debugger only)

haltman-at:
This PR adds support for decoding internal function pointers.  Note that this is supported only in the debugger, not in the contract decoder; attempting to decode an internal function pointer from the contract decoder will simply result in the string `"<decoding not supported>"`.

This PR also fixes a bug in my previous PR for decoding external functions.  In that PR, contexts in the decoder were indexed by contract ID rather than by hash.  However, this causes a problem in that not every context has a contract ID.  This PR changes it so that contexts in the decoder are indexed by... whatever.  (When used from the debugger, it's back to hashes.  When used from the contract decoder, I've left it as contract IDs.)

So, why is decoding for internal functions only supported from the debugger?  Fundamentally, to decode an internal function we need to take a program counter value and find the corresponding AST node.  Well, the debugger does that all the time!  But to do that, it needs to be able to associate source indices to particular source files, and to do that, it needs the whole project to be compiled at once.  The debugger handles this by recompiling the entire project at startup, but this is slow and it only does so because it has to.  That said, having done this, it's now prepared also to decode internal function pointers.

But the contract decoder mostly does not need to recompile, and it's just not worth the performance penalty to add a recompilation step just so it can decode internal function pointers.  So, decoding of internal function pointers remains a debugger-only feature for now.  Once we have truffle-db, however, it should be possible to decode internal function pointers from the contract decoder as well, and we will have to revisit this.

So, what do internal function pointers decode to?  Well, since we don't have the new output format yet, they decode to strings containing the name of the function pointed to.  The string has the form `Contract.function`, where `Contract` is the contract the function is defined in, and `function` is the name of the function.  (Note that we need to specify the contract the function is defined in due to function overriding; a derived contract can store either the derived version or the base version of an overridden function.)

There are also some special values an internal function pointer can decode to.  The case of `"<decoding not supported>"` has already been mentioned above, but there are more.  A function pointer to program counter zero will decode to `"<zero>"`.  A function pointer to the contract's designated invalid function will decode to `"assert(false)"`.  And anything else that can't be decoded will decode to `undefined`.  That's not *really* a special value, but it can come up due to a Solidity bug -- more on that later.

Anyway, how's decoding of internal functions work?  Well, because internal function decoding is only supported from the debugger, I took the liberty of having the debugger do most of the work.  Rather than the decoder computing an AST node from the program counter, the debugger precomputes a table of internal functions (for the current context) and passes it to the decoder; the decoder can then just look up the answer in the table.  The table also contains various information the decoder doesn't use, but I didn't see any need to strip it out.

The debugger now also passes the current context (in the decoder's format) to the decoder.  Very little of this is currently used by the decoder, but more will be in the future.  The one part of the current context used by the decoder is the `isConstructor` flag.  An internal function pointer contains the deployed PC value in the lower 4 bytes and (sometimes) the constructor PC value in the upper 4 bytes; the `isConstructor` flag is used to determine which of these two to look at.  (I've added a new EVM constant, `PC_SIZE`, to truffle-decode-utils.  The storage size of an internal function pointer is now defined to be `2 * PC_SIZE`, rather than just a raw 8.)

The lookup table itself is constructed in a new selector called `solidity.current.functionsByProgramCounter`.  It's indexed by the program counter and contains information on the corresponding function and its defining contract.  It's constructed by taking `solidity.current.instructions`, restricting to `JUMPDEST` instructions with file index not equal to `-1`, and then using `findRange` to find the function node.  The contract node can then be found by going up (accomplished here by manipulation of the pointer string, chopping off the end).

Among the information in the entries of this table is a field called `isDesignatedInvalid`; this is `false` for those entries constructed as described above, but we also include not only the context's defined functions but also its designated invalid function (if it has one).  This is found by looking for `JUMPDEST` instructions that don't correspond to a node, but are followed immediately by an `INVALID` instruction.  The table entry for this function has `isDesignatedInvalid` set to `true`, with the other fields left undefined.

Finally, it's worth discussing [a Solidity bug](https://github.com/ethereum/solidity/issues/6525) I discovered while working on this PR.  Internal function pointers in constructors that are supposed to point to the designated invalid function are formatted incorrectly.  The result is that such pointers, which should decode to `"assert(false)"`, will instead decode to `"<zero>"` in the constructor where they're declared, and (probably) to `undefined` if they're stored in storage and then used in a deployed context.  (They could actually decode to basically anything, but `undefined` is the most likely outcome.)  This bug is now scheduled to be fixed in Solidity 0.5.8.

I could have written this PR to handle such pointers and decode them properly, but it would have been substantially more work, and @gnidan agreed that this case was minor enough to not require such additional work.  So, be aware that when decoding uninitialized function pointers in constructors (or originating from constructors), in any presently-existing version of Solidity, this issue may come up.

This PR also adds two tests of this new functionality.

So that's internal function decoding!  With this, the debugger can finally decode **all data types**.  This statement has substantial caveats, of course -- we still can't decode many constant state variables, for instance.  But "constant state variable" is not a data type.  Nor is "circular structure".  So, I'd say we're doing pretty good! :)

### DIFF
--- a/packages/truffle-debugger/lib/solidity/selectors/index.js
+++ b/packages/truffle-debugger/lib/solidity/selectors/index.js
@@ -229,16 +229,13 @@ let solidity = createSelectorTree({
     instructionAtProgramCounter: createLeaf(
       ["./instructions"],
 
-      instructions => {
-        let map = {};
-        instructions.forEach(function(instruction) {
-          map[instruction.pc] = instruction;
-        });
-        //note: this will have gaps in it.  That's OK!  Those gaps are the data
-        //portions of push instructions, which it is illegal to jump into.  We
-        //don't need to assign instructions to illegal PC values.
-        return map;
-      }
+      instructions =>
+        Object.assign(
+          {},
+          ...instructions.map(instruction => ({
+            [instruction.pc]: instruction
+          }))
+        )
     ),
 
     ...createMultistepSelectors(evm.current.step),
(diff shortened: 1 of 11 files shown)


packages/truffle-debugger/lib/solidity/selectors/index.js
```diff
@@ -269,6 +266,71 @@ let solidity = createSelectorTree({
       }
     ),
 
+    /*
+     * solidity.current.functionsByPc
```
gnidan: This is cool!

Naming suggestion: I feel like we should spell out `programCounter`. I think it'll read better: `solidity.current.functionsByProgramCounter`, and so on.

haltman-at: Makes sense.  Will go change that.

PR #1923, opened 2019-04-18
Enable querying artifacts for contract instance contracts

gnidan:
i.e.,

```graphql
{
  artifacts {
    contractInstance(name: $name, networkId: $networkId) {
      contract {
        /* ... */
      }
    }
  }
}
```

In order to get this to work right, I had to add in an explicit `delegateToSchema` call. Not sure if there's a better way, but this should work for now.


gnidan:
The thing about it that is sub-optimal is the fact that two separate but identical calls are made to the JSON schema. Seems like it shouldn't be necessary? I don't know. Not worth fighting with it, I figure.

Gonna merge! Thanks!

PR #1926, opened 2019-04-18
Support source map option when compile using vyper.

nakajo2011:
Can switching on/off by truffle-config.js for older vyper.

example:
```
module.exports = {
  vyper: {
    source_map: true
  }
}
```

### DIFF
--- a/packages/truffle-compile-vyper/index.js
+++ b/packages/truffle-compile-vyper/index.js
@@ -83,8 +83,11 @@ function checkVyper(callback) {
 }
 
 // Execute vyper for single source file
-function execVyper(source_path, callback) {
+function execVyper(options, source_path, callback) {
   const formats = ["abi", "bytecode", "bytecode_runtime"];
+  if (options.vyper && options.vyper.source_map) {
+    formats.push("source_map");
+  }
   const command = `vyper -f${formats.join(",")} ${source_path}`;
 
   exec(command, function(err, stdout, stderr) {
@@ -114,7 +117,7 @@ function compileAll(options, callback) {
   async.map(
     options.paths,
     function(source_path, c) {
-      execVyper(source_path, function(err, compiled_contract) {
+      execVyper(options, source_path, function(err, compiled_contract) {
         if (err) return c(err);
 
         // remove first extension from filename
@@ -134,6 +137,7 @@ function compileAll(options, callback) {
           abi: compiled_contract.abi,
           bytecode: compiled_contract.bytecode,
           deployedBytecode: compiled_contract.bytecode_runtime,
+          sourceMap: compiled_contract.source_map,
 
           compiler: compiler
         };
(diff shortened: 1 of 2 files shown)


CruzMolina:
Cool @nakajo2011 ! Reviewing shortly. 🧐

packages/truffle-compile-vyper/index.js
```diff
@@ -83,8 +83,11 @@ function checkVyper(callback) {
 }
 
 // Execute vyper for single source file
-function execVyper(source_path, callback) {
+function execVyper(options, source_path, callback) {
   const formats = ["abi", "bytecode", "bytecode_runtime"];
+  if (options.vyper && options.vyper.source_map) {
```
CruzMolina: So I think we want to use camelCase for sourceMap (for consistency within truffle-config.js),
and we also probably want to have the config setting be under options.compilers, rather than start a separate key for vyper.

```suggestion
  if (options.compilers.vyper.sourceMap) {
```

packages/truffle-compile-vyper/test/test_compiler.js
```diff
@@ -1,60 +1,127 @@
...
-    compile.all(config, function (err, contracts, paths) {
-      assert.equal(err, null, 'Compiles without error');
+  it("compiles with source_map option", function(done) {
+    // this test pass when vyper 0.1.0-b9 or higher
+    exec("vyper --version", function(err, stdout) {
+      let version;
+      if (!err) version = stdout.trim();
+
+      if (version === "0.1.0-b9") {
+        config.merge({
```
CruzMolina: So then this section needs to be updated to:

```suggestion
        config.merge({
            compilers: {
              vyper: {
                sourceMap: true
              }
            }
        });
```

packages/truffle-compile-vyper/test/test_compiler.js
```diff
@@ -1,60 +1,127 @@
...
       });
 
       done();
     });
   });
 
-  it('skips solidity contracts', function (done) {
-    compile.all(config, function (err, contracts, paths) {
-      assert.equal(err, null, 'Compiles without error');
+  it("compiles with source_map option", function(done) {
```
CruzMolina: Perhaps we don't need to verify the locally installed vyper version for this test (`pip install vyper` pulls down a `source_map` option compatible vyper by default currently, and Travis CI also pulls down a compatible version).

If we _do_ want to keep the version check, versions ` >= 0.1.0b4` are compatible so we'll need to do a regex here of some sort.

CruzMolina:
Thanks for getting this started @nakajo2011! 🙏

Let me know your thoughts and if you'd like me to finish this up for you.

packages/truffle-compile-vyper/test/test_compiler.js
gnidan: Extra note on this: let's get `sourceMap` or `source_map` inside a `vyper.settings` namespace:
```javascript
  compilers: {
    vyper: {
      settings: {
        sourceMap: true
      }
    }
  }

**Comment by nakajo2011** [comment]
Thanks for your review! I will agree your review. Should I modify following your review?

**Comment by CruzMolina** [comment]
Go for it @nakajo2011 👍

**Comment by CruzMolina** [comment]
Example config for next release:

```
module.exports = {
  compilers: {
    vyper: {
      settings: {
        sourceMap: true
      }
    }
  }
}
```

PR #1929, opened 2019-04-19
Add cache to contract decoder, and lock in block number at decoding start

haltman-at:
This PR makes two related changes to the contract decoder.  One is performance-related: The contract decoder now caches its storage and code requests to avoid repeating them.  (Although, its request for the contract's own code on startup is no longer used for any sort of caching.)  The other is that now, `state` and `variable` both begin by locking in the block number that will be used for all requests, so that if decoding happens just as a new block is mined, we don't decode an inconsistent state.  (If `state` is called, though, we don't bother refetching the code based on the block number; we assume `state` is being called on a block where the contract exists with its current code.  So the code is still used for caching in that sense.)

To lock in the block number, we check if the passed in block is of type `number`; if so we leave it unaffected, otherwise we make a `web3` call to get the specified block and get its `number` field.  I suppose if we were given the block as a hash this step isn't really necessary, but, eh, whatever.  It's there to handle such cases as `"latest"` or `"pending"`, and I didn't see it as worth the extra effort to distinguish those from hashes.  (Note that `"latest"` should be the most common case; indeed, I believe at the moment it's the *only* case that's actually used.)  Also, note that the `decodeVariable` function now takes `block` as a `number`, not a `BlockType`, to make it clear that, hey, you should lock in a block number before using this.

There's a cache for code requests and a cache for storage requests.  Both of these hold the result as a `Uint8Array`, since that's what we pass to the main decoder.  The code cache is by block and address.  The storage cache is by block, address, and slot, even though at the moment the only address we'll ever request is the contract's own address.  Still, I figured I ought to be prepared there.

Anyway, that's basically it.  Pretty simple.  Locking in the block number does add an extra request (if the block wasn't given as a number), but it's only one extra request at the start of decoding, it's not a whole extra request for each variable decoded or each slot requested or anything like that, so I think it's fine.  (And obviously if things are already cached it's not a net increase.)  And, obviously, doing this allows the cache to work and gets us state coherence.

### DIFF
--- a/packages/truffle-decoder/lib/interface/contract-decoder.ts
+++ b/packages/truffle-decoder/lib/interface/contract-decoder.ts
@@ -79,6 +79,20 @@ export interface ContractMapping {
   [nodeId: number]: ContractObject;
 };
 
+export interface StorageCache {
+  [block: number]: {
+    [address: string]: {
+      [slot: string]: Uint8Array;
+    };
+  };
+}
+
+export interface CodeCache {
+  [block: number]: {
+    [address: string]: Uint8Array;
+  };
+}
+
 //note: may return undefined
 function getContractNode(contract: ContractObject): AstDefinition {
   return (contract.ast || {nodes: []}).nodes.find(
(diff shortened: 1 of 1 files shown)


haltman-at:
Yes, that's a good point.  I suggested earlier that maybe it should be its own package, but I'll admit I didn't yet have any concrete plans to actually, y'know, *do* that.  I guess I'll add it to my todo list, at least?

PR #1932, opened 2019-04-19
Refactor Travis CI

### LINKED ISSUE #1336: Run separate CI task for all tests with external (`apt-get`) dependencies
## Issue

Travis is slow and getting slower now that both `solc` and `vyper` have to get installed.

## Steps to Reproduce

1. https://travis-ci.org/trufflesuite/truffle
2. Options -> Trigger Build -> etc.
3. Wait longer than an amount of time for which it is reasonable to have patience

## Expected Behavior

The two main CI jobs (Packages & Integrations) should not run any tests that rely on any packages installed by `apt-get`.

A separate job should run for specifically these tests. (unknown requirement: should this job be marked as "allowed failure"?)

## Background

There are at least two places where CI environment changes how or which tests are run:

- The [external compile tests](https://github.com/trufflesuite/truffle/blob/46fc7bdb9fde7cc3fa99b94fef04b1d6c04c38f7/packages/truffle/test/scenarios/external_compilers/truffle-compile.js#L15) use the `jq` utility, so these only run if the `CI` environment variable is true.

- A bunch of tests are marked `[ @geth ]` ([e.g.](https://github.com/trufflesuite/truffle/blob/46fc7bdb9fde7cc3fa99b94fef04b1d6c04c38f7/packages/truffle/test/scenarios/migrations/production.js#L24-L25)), which gets indicated to mocha by way of the "test filter" option.*

*This approach is annoying because it means the main CI script for running "all the tests" consumes this CLI flag (`-k` I think but I'm too lazy to look it up). So you can't use `-k` locally, effectively at all, if you just want to do `yarn test -k ...some filter...` in the root.

For this issue, it may be worth thinking of a more explicit strategy for how CI manages collections of tests. Or not. Alternate strategy: hack it in and get tests faster by any means necessary.

## Environment

Travis-CI

CruzMolina:
This PR refactors CI tests to only install external dependencies for relevant jobs. Also ensures only relevant tests for each job are ran (i.e. geth tests in `truffle-contract` now skipped in non-geth jobs).

Specific streamlined dependencies:
`docker pull` for `solc` & `geth`, `apt-get` for `jq` & `snapd`, & `snap install vyper`.

Saves us ~4-5 min for the build.
Resolves #1336.

Also includes updated `yarn.lock` left out from #1930 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: required
 language: node_js
 
 before_install:
-  - docker pull ethereum/solc:0.4.22
   - sudo add-apt-repository --yes ppa:ethereum/ethereum
   - sudo apt-get update
   - sudo apt-get install -y dpkg solc snapd
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ before_install:
   - docker pull ethereum/solc:0.4.22
   - sudo add-apt-repository --yes ppa:ethereum/ethereum
   - sudo apt-get update
-  - sudo apt-get install -y dpkg solc jq snapd
+  - sudo apt-get install -y dpkg solc snapd
   - export PATH=$PATH:/snap/bin
   - sudo snap install vyper --beta --devmode
   - npm list -g lerna --depth=0 || npm install -g lerna
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: required
 language: node_js
 
 before_install:
-  - docker pull ethereum/client-go:latest
   - docker pull ethereum/solc:0.4.22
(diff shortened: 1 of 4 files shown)


cgewecke:
@CruzMolina Sorry for looking over your shoulder here but possibly this change should be revisited:
> geth tests in truffle-contract now skipped in non-geth jobs

Ideally all the tests should pass ganache and but only a subset of them will pass on geth, hence the geth tag.  

This is relevant wrt #1852  (upgrade web3) because it looks like many of the relevant tests are now hidden in an allowed failure on CI? Forgive me if I've misunderstood this.

CruzMolina:
Thanks for checking in on geth integration @cgewecke.

I think it's probably time to move the geth job out of it's `allowed failure` state.

PR #1934, opened 2019-04-20
change: Change maxBuffer size to 600KB.

nakajo2011:
For fixed error of `exceeded maxBuffer` when compiled too large Contract that has over 20KB bytecodes.
Maybe this error happens at apply the `sourceMap` option.

### DIFF
--- a/packages/truffle-compile-vyper/index.js
+++ b/packages/truffle-compile-vyper/index.js
@@ -87,7 +87,7 @@ function execVyper(source_path, callback) {
   const formats = ["abi", "bytecode", "bytecode_runtime"];
   const command = `vyper -f${formats.join(",")} ${source_path}`;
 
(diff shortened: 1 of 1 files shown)


CruzMolina:
Hey @nakajo2011 ! Can you share an example large vyper contract that throws the buffer error? I would like to demo the error & solution locally. Thanks! 🙏

nakajo2011:
Hi @CruzMolina .
Example large vyper is below: https://github.com/cryptoeconomicslab/plasma-chamber/blob/master/packages/contracts/contracts/RootChain.vy

I got buffer error, when compile this project with enabled sourceMap option in truffle.
thanks!

CruzMolina:
Good catch @nakajo2011 ! Thanks again for helping us improve truffle's vyper support!

PR #1948, opened 2019-04-24
Add version prop to truffle-contract

eggplantzzz:
Add a version property to the truffle contract object.

### DIFF
--- a/packages/truffle-contract/index.js
+++ b/packages/truffle-contract/index.js
@@ -1,13 +1,16 @@
-var Schema = require("truffle-contract-schema");
-var Contract = require("./lib/contract");
+const Schema = require("truffle-contract-schema");
+const Contract = require("./lib/contract");
+const truffleContractVersion = require("./package.json").version;
 
-var contract = function(options = {}) {
-  var binary = Schema.normalize(options);
+const contract = function(options = {}) {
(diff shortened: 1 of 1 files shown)


eggplantzzz:
Is there any reason that we should set this on the prototype instead of the object?

CruzMolina:
@eggplantzzz Why not both? I imagine it might be equally useful to grab it from the object or access as a prototype. I notice `web3` allows this.

eggplantzzz:
Hmm, on second thought it feels a bit weird to me to expose the library version on the contract instances themselves...

eggplantzzz:
What do you think about that @gnidan ?

gnidan:
Hm, it might be nice to have it on the contract objects as well. My thinking is that some code contexts might not `require("truffle-contract")`, instead just get passed `MyContract`?

I could go either way for this PR, though. This PR seems to meet the requirements for the connected issue, might just be worth merging.

btw @princesinha19 thanks for the reviews :)

eggplantzzz:
I'm going to go ahead and leave it off the prototype for now until I see a more definite need for the property on the contract objects. And who knows, perhaps we'll want to save that property name for something else in the future.

CplXhristianUSMC:

PR #1949, opened 2019-04-24
Retain clean artifacts by removing function signatures during schema normalization

### LINKED ISSUE #1889: ABIs in artifacts are sometimes missing signatures
## Issue

Under certain conditions, the elements of the `abi` member of the Truffle artifact will be missing their `signature` element.  This can occur even for concrete contracts (not just interfaces and abstract contracts).  (And, really, there's no reason not to include them even for contracts that aren't concrete.)

(Note that @gnidan suggests that actually the `signature` element should *always* be missing, but I'm filing this anyway for completeness.)

## Steps to Reproduce

Compile a Truffle project with the following contract file:

```
pragma solidity ^0.5.0;

contract TestAbstract {
  function doThingA() public;
}

contract TestConcrete is TestAbstract {

  function doThingA() public {
  }
}
```

## Expected Behavior

The artifact for `TestConcrete` should have signatures.  (And, IMO, as mentioned above, the artifact for `TestAbstract` should too, even though I'm guessing their omission was deliberate.)

## Actual Results

The artfact for `TestConcrete` (as well as the artifact for `TestAbstract`) did not have signatures.

CruzMolina:
Somewhere (maybe in `truffle-contract`) there's a bug that adds function `signature`s to the ABI during migrations. By adding a check to `transform` and removing any present `signature` key in an `abi` array, this PR ensures that artifacts stay clean and `signature` free!

Potentially resolves #1889 by ensuring `signature`s aren't saved to the artifacts.

### DIFF
--- a/packages/truffle-contract-schema/index.js
+++ b/packages/truffle-contract-schema/index.js
@@ -41,6 +41,8 @@ var properties = {
         } catch (e) {
           value = undefined;
         }
+      } else if (typeof value === "object") {
+        value.forEach(item => delete item.signature);
       }
       return value;
     }
--- a/packages/truffle-contract-schema/index.js
+++ b/packages/truffle-contract-schema/index.js
@@ -28,6 +28,18 @@ var abiSchema = require("./spec/abi.spec.json");
  * The optional `transform` parameter standardizes value regardless of source,
  * for purposes of ensuring data type and/or string schemas.
  */
+
+// helper that ensures abi's do not contain function signatures
+const sanitizedValue = dirtyValueArray => {
+  let sanitizedValueArray = [];
+  dirtyValueArray.forEach(item => {
+    let sanitizedItem = Object.assign({}, item);
+    delete sanitizedItem.signature;
+    sanitizedValueArray.push(sanitizedItem);
+  });
+  return sanitizedValueArray;
+};
+
 var properties = {
   contractName: {
     sources: ["contractName", "contract_name"]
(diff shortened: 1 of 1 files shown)


packages/truffle-contract-schema/index.js
```diff
@@ -41,6 +41,8 @@ var properties = {
         } catch (e) {
           value = undefined;
         }
+      } else if (typeof value === "object") {
+        value.forEach(item => delete item.signature);
```
gnidan: I was debating whether we can allow this mutation, but I don't see any other prior precedent of one of these `transform` functions mutating an object. I think it's probably a risky move, sorry. Here I'll take a stab, how's this:
```suggestion
        value = Object.entries(value)
          .filter(([key]) => key !== "signature")
          .reduce((a, b) => ({ ...a, ...b }), {});
```

gnidan: The other thought I have here is whether or not we should make this an `else`. Should we just treat it as a prerequisite that a string value here conform?

gnidan: Oh shoot my code is wrong. One sec.

CruzMolina: Ah, should have clarified in the code. We only want `abi`'s (array typed) to be returned w/o function signatures (if they exist). Pushed something a little more explicit that retains immutability. 🤞

gnidan: Oh I forgot the outer loop! I like your new solution a lot more anyway

PR #1955, opened 2019-04-26
Rework migrations

eggplantzzz:
This PR is an attempt to rework some of the legacy Truffle code. The author is preferring a Promise-based approach and attempting to minimize the number of callbacks passed.

It is also built on the work done in PR #1905


packages/truffle-compile/test/compilerSupplier/index.js
```diff
@@ -115,6 +115,13 @@ describe("CompilerSupplier", () => {
     });
 
     describe("when a user specifies the compiler url root", () => {
+      beforeEach(() => {
+        sinon.stub(VersionRange.prototype, "versionIsCached").returns(false);
+      });
+      afterEach(() => {
+        VersionRange.prototype.versionIsCached.restore();
+      });
```
gnidan: Neat!

packages/truffle-core/lib/commands/migrate.js
```diff
@@ -129,6 +129,42 @@ const command = {
...
+      const config = Config.detect(options);
+
+      config.contracts_build_directory = buildDir;
+      config.artifactor = new Artifactor(buildDir);
+      config.resolver = new Resolver(config);
+
+      try {
+        await Environment.detect(config);
+      } catch (error) {
+        throw new Error(error);
```
gnidan: By throwing this error here, we won't end up running `done(error)`. Not sure if that's intended? My hunch is that this line needs to be removed.

packages/truffle-core/lib/environment.js
```diff
@@ -154,7 +68,85 @@ const Environment = {
...
+        config.network = "development";
+      } else {
+        config.network = "ganache";
+        config.networks[config.network] = {
+          host: "127.0.0.1",
+          port: 7545,
+          network_id: 5777
+        };
+      }
+    }
```
gnidan: I like the helpers!

packages/truffle-workflow-compile/index.js
```diff
@@ -72,29 +72,37 @@ const Contracts = {
...
-      },
-      0
-    );
-
-    if (numberOfCompiledContracts === 0) this.reportNothingToCompile(options);
-
-    this.reportCompilationFinished(options, config);
-    return await this.collectCompilations(compilations);
-  }),
+  compile: async function(options, callback) {
```
gnidan: I think in order to make this polymorphic, we can't make it `async`, because it will return a promise no matter what. Not sure though, maybe this is fine as a best practice?

packages/truffle-core/lib/commands/migrate.js
eggplantzzz: Oh good catch. Actually `done` gets called on the outside of the function...`done` should be removed from the function signature. I think this was some code that got left in along with the callback to `done` in the else.

eggplantzzz: Ok so prepareConfigForRealMigrations is only called in one place and now `done` is handled on the outside of the function.

packages/truffle-workflow-compile/index.js
eggplantzzz: Ah yes I see what you mean. Let me give it a mull-over.

eggplantzzz: After working this method out in an alternate way, I'm going to go ahead and claim that it is not a problem the way it currently is in this PR. It is not too bad so if you feel strongly about changing it, then I can go ahead and make the swap no problem. Otherwise, I think it is a bit impractical write it and make sure it doesn't return anything in the case that a callback is passed.

CplXhristianUSMC:

Issue #1956, opened 2019-04-26
Specify the working directory for an external complier command


ianmorgan:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.
Discussed on slack with Truffle developers 
---------------------------

## Issue

When invoking an external compilation step (in this case some Java code) , the working directory needs to be set to the same directory as the contract code for the compiler to run, 
for example if the truffle-config has 

```
contracts_directory: "/path/to/myapp",

  compilers: {
      external: {
        command: "./gradlew build",
       targets: [{
               path: "build/libs/myapp.jar"
             }]
      }
    }
```
and I need the "./gradlew" command to be run in the '/path/to/myapp' directory 

One possible solution is 

```
contracts_directory: "/path/to/myapp",

  compilers: {
      external: {
        command: "./gradlew build",
        working_directory: "/path/to/myapp",
       targets: [{
               path: "build/libs/myapp.jar"
             }]
      }
    }
```

 
## Steps to Reproduce

See above

## Expected Behavior

Using the suggested syntax, and running on linux/OSX, then 

```
compilers: {
      external: {
        command: "pwd",
        working_directory: "/path/to/myapp",
       targets: [{
             }]
      }
    }
```
would echo '/path/to/myapp' to the console

_Note, I would also expect 'targets' to now resolve using using the working_directory specified_

## Actual Results

N/A - new feature

## Environment

Current master. Any environment

gnidan:
Looks like this already exists! The option is called `cwd`. Your provided example would be:

```
compilers: {
      external: {
        command: "pwd",
        cwd: "/path/to/myapp",
       targets: [{
             }]
      }
    }
```

I'll mark this issue as just a required documentation update. Thanks for raising the concern!

ianmorgan:
Hi - I've just tried adding this option ("cwd"), but it doesn't appear to actually change the working directory - "pwd" continues to output the root folder - this is the same if I use relative or absolute paths for "cwd" 

Note, my truffle version is returning

17LDN-MAC55-3:iteration1 ianmorgan$ truffle version 
Truffle v5.0.8 (core: 5.0.8)
Solidity v0.5.0 (solc-js)
Node v11.12.0
Web3.js v1.0.0-beta.37

eggplantzzz:
Yes, as it currently stands you would have to add the directory change into your own custom script. Thanks for the suggestion!

[...]

PR #1964, opened 2019-04-30
Add working_directory option for external compiler

gnidan:
Ref: #1956 

This adds a new `working_directory` option for external compilers, changing the default directory for external commands as well as their outputs.

### DIFF
--- a/packages/truffle-external-compile/index.js
+++ b/packages/truffle-external-compile/index.js
@@ -1,7 +1,7 @@
-'use strict';
+"use strict";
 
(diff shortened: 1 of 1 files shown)


gnidan:
Now I want to change it to `workingDirectory` since the other external compiler options are all camelCase. Thoughts?

gnidan:
> Makes sense to keep the external compiler options consistent and use camelCase.

Now it's both! Merging this, thanks @CruzMolina !

PR #1965, opened 2019-04-30
Update tsconfig files to specify explicit @types dependencies

gnidan:
(internal improvement)

Noticed this when pulling `develop` into `truffle-db` cause `@types/mocha` conflicts with `@types/jest`.

Targeting `develop` to nip this in the bud.


gnidan:
Excessive review requests are so y'all are aware we gotta be explicit about TS package @types now

PR #1966, opened 2019-04-30
Fix yarn warnings

gnidan:
Firstly, this fixes the circular dependency warnings by making the following packages into peer dependencies:
- truffle-resolver's truffle-contract
- truffle-decoder's truffle

Additionally, this fixes warnings about inconsistent node executables by finding and replacing all instances of package.json files defining scripts in terms of `npm run <script>`. `yarn <script>` is used instead in all cases now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,8 @@ matrix:
     - node_js: 8
       env: COVERAGE=true
 
-
 install:
-  - npm run bootstrap
+  - yarn bootstrap
 
 cache:
   directories:
@@ -36,7 +35,7 @@ cache:
     - $(npm config get prefix)/lib/node_modules # globally installed stuff (i.e. lerna)
 
 script:
-  - npm run ci
+  - yarn ci
 
 notifications:
   slack:
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ We welcome pull requests. To get started, just fork this repo, clone it locally,
 # Install
 npm install -g lerna@3.4.3
 npm install -g yarn
-npm run bootstrap
+yarn bootstrap
 
 # Test
 npm test
--- a/packages/truffle-contract/package.json
+++ b/packages/truffle-contract/package.json
@@ -4,7 +4,7 @@
   "description": "A better contract abstraction for Ethereum (formerly EtherPudding)",
   "main": "index.js",
   "scripts": {
-    "prepare": "npm run compile",
+    "prepare": "yarn compile",
     "publish:next": "node ../truffle/scripts/prereleaseVersion.js next next",
     "test": "./scripts/test.sh",
     "test:debug": "$(npm bin)/mocha --inspect-brk",
--- a/packages/truffle-contract/package.json
+++ b/packages/truffle-contract/package.json
@@ -7,8 +7,8 @@
     "prepare": "yarn compile",
     "publish:next": "node ../truffle/scripts/prereleaseVersion.js next next",
     "test": "./scripts/test.sh",
-    "test:debug": "$(npm bin)/mocha --inspect-brk",
-    "test:trace": "$(npm bin)/mocha --trace-warnings",
+    "test:debug": "$(yarn bin)/mocha --inspect-brk",
+    "test:trace": "$(yarn bin)/mocha --trace-warnings",
     "compile": "mkdir -p dist && browserify ./index.js -o ./dist/truffle-contract.js && uglifyjs ./dist/truffle-contract.js -o ./dist/truffle-contract.min.js"
   },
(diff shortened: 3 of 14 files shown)


packages/truffle/scripts/test.sh
```diff
@@ -3,9 +3,9 @@
 set -o errexit
 
 if [ "$GETH" == true ]; then
-  npm run build-cli && mocha --timeout 50000 --grep @ganache --invert --colors $@
+  yarn build-cli && mocha --timeout 50000 --grep @ganache --invert --colors $@
 elif [ "$COVERAGE" == true ]; then
   NO_BUILD=true mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
 else
-  npm run build-cli && mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
+  yarn run build-cli && mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
```
eggplantzzz: Does this need to have 'run' removed?

eggplantzzz:
Other than the one comment, it looks cool.

.travis.yml
```diff
@@ -26,17 +26,16 @@ matrix:
...
 
-
 install:
-  - npm run bootstrap
+  - yarn bootstrap
 
 cache:
   directories:
     - node_modules
     - $(npm config get prefix)/lib/node_modules # globally installed stuff (i.e. lerna)
```
CruzMolina:
```suggestion
    - $(yarn config get prefix)/lib/node_modules # globally installed stuff (i.e. lerna)
```

README.md
```diff
@@ -57,7 +57,7 @@ We welcome pull requests. To get started, just fork this repo, clone it locally,
 # Install
 npm install -g lerna@3.4.3
 npm install -g yarn
-npm run bootstrap
+yarn bootstrap
 
 # Test
 npm test
```
CruzMolina:
```suggestion
yarn test
```

packages/truffle-contract/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "A better contract abstraction for Ethereum (formerly EtherPudding)",
   "main": "index.js",
   "scripts": {
-    "prepare": "npm run compile",
+    "prepare": "yarn compile",
     "publish:next": "node ../truffle/scripts/prereleaseVersion.js next next",
     "test": "./scripts/test.sh",
     "test:debug": "$(npm bin)/mocha --inspect-brk",
```
CruzMolina:
```suggestion
     "test:debug": "$(yarn bin)/mocha --inspect-brk",
```
Same for `"test:trace"` below.

packages/truffle-decoder/package.json
```diff
@@ -50,5 +49,8 @@
     "truffle-decode-utils": "^1.0.11",
     "web3": "1.0.0-beta.37"
   },
+  "peerDependencies": {
```
CruzMolina: 🎉

packages/truffle/scripts/test.sh
CruzMolina:
```suggestion
  yarn build-cli && mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
```

gnidan: Yep, good catch!

.travis.yml
gnidan: Hm I just tried this locally and it doesn't seem to work. Have you tried it?

CruzMolina: Hmm, seems to work on my end?

```
➜  truffle git:(quorum-subprovider) ✗ $(yarn config get prefix)/lib/node_modules
➜  node_modules ls
yarn
➜  node_modules 
```

gnidan: What happens if you run `yarn config get prefix` by itself? I get `undefined`

CruzMolina: same, which is why we pass the expected path where the global `node_modules` would be installed/cached.

haltman-at:
Question: When did `truffle-decoder` gain a dependency on `truffle`?  I remember seeing such warnings starting to pop up, but I don't remember ever deliberately adding that.  Does it actually need to be there?  Did someone else add it?

gnidan:
@haltman-at it's a dev. dependency, really, since the decoder tests run Truffle itself, I believe.

haltman-at:
@gnidan Oh, duh.  That makes sense.  Guess that was already there then and it was something else that changed to form the loop.

.travis.yml
gnidan: This seems wonky with how it returns undefined. We'll just have to keep an eye to make sure the Travis cache behaves.

gnidan:
Thanks all!

CplXhristianUSMC:

PR #1967, opened 2019-04-30
Update truffle-migrate and truffle-require

eggplantzzz:
Update Require.file to no longer take a callback, use synchronous version of fs method, and update some syntax. Also remove callbacks from truffle-migrate.


CplXhristianUSMC:

eggplantzzz:
Hmmm, looks like those are also failing on `develop`. I guess we'll have to figure out how to resolve those.

packages/truffle-core/lib/commands/test.js
```diff
@@ -123,26 +121,29 @@ var command = {
...
+            .catch(error => {
+              done(error);
             });
-          });
         };
 
         if (config.networks[config.network]) {
           Environment.detect(config)
             .then(() => {
+              console.log("IN the then from detect!!!!!!");
```
CruzMolina:
```suggestion
```

packages/truffle-core/lib/commands/test.js
```diff
@@ -162,7 +163,7 @@ var command = {
             gasLimit: config.gas,
             noVMErrorsOnRPCResponse: true
           };
-
+          console.log("about to run Develop.connectOrStart");
```
CruzMolina:
```suggestion
```

PR #1969, opened 2019-04-30
Implement debugger transactionless mode

haltman-at:
This pull request allows the debugger to operate without a transaction loaded, and to load and unload transactions without having to recompute project-level information such as storage allocations.

On the CLI side, the debugger can now be invoked with or without an argument; invoking it without an argument will start it up with no transaction loaded.  (And invoking it with an invalid argument will, after failing to load the transaction, dump the debugger into transactionless mode, rather than quitting.)  On the library side, in addition to `forTx`, there is now also `forProject`; this does not take a transaction argument and starts the debugger with no transaction loaded.

Once the debugger is operating -- I'll just describe the CLI side of things for now, there are analogous `session` methods -- one can unload the current transaction with `T` and load a new transaction with `t`.  Note that we don't allow loading a new transaction while one is already loaded, but require manual unloading of the current transaction first; I'll get to why later, but for now take note of this limitation.

A number of error cases have been added to various things to take care of the case when no transaction is loaded.  I won't discuss those in details here.

So, what's different under the hood?  Well, a number of things, but here are what strike me as the most important changes.

The biggest change is that, whereas previously the various submodules had two levels of state, `info` and `proc`, there are now three.  What was previously `info` has been split into `info` and `transaction`; `info` contains the project-level information, while `transaction` contains the transaction-level information.  Selectors have been similarly renamed.  (Why `transaction` rather than the shorter `tx`?  Because there was already a reducer in `evm` named `tx`. :P )

The `info` state is never reset.  However, the `transaction` state is reset when an `UNLOAD_TRANSACTION` is received.  The new `session` saga `unload` sends an `UNLOAD_TRANSACTION` action to those submodules that have that action (not all need one), and a `RESET` action to the others.

Note that some of the `evm` `proc` reducers respond differently to a `RESET` than they do to an `UNLOAD_TRANSACTION`, whereas outside of `evm` both of these will just clear the `proc` state.  The reason it works differently in `evm` is that when a transaction is loaded, the `evm.begin` saga is called.  So the `UNLOAD_TRANSACTION` will completely clear the `evm` proc state, to prepare for th next call of `evm.begin`.  But a mere reset doesn't involve calling `evm.begin` again, so `RESET` just resets the `evm` state to how things are *after* `evm.begin`.  (This was already the case, just before there was no `UNLOAD_TRANSACTION` to contrast it with.)

By the way, I've simplified how the session method `reset` works -- it no longer dispatches an action which triggers a saga, but simply uses `_runSaga` to call the `reset` saga.  The other control actions have been left as they are, though, to allow them to remain interruptible.  Although I've changed the controller listener and reducers so that, instead of the `isStepping` state responding directly to stepper actions, instead the listener puts a separate `START_STEPPING` action.  This means more actions are used, but it reduces repetition in the code.

I've handled `loadTransaction` and `unloadTransaction` similarly.  The `unloadTransaction` method just uses `_runSaga` to run the `unload` saga, while the `loadTransaction` method dispatches a `LOAD_TRANSACTION` action to allow loading to be interruptible (using the same mechanism the controller uses).  This does mean once again that a synchronization method is needed; I use the session status for this, although I've reworked it a fair bit (the old system no longer made a lot of sense).

Now, instead of `session.status` being one of `"WAITING"`, `"ACTIVE"`, or an error object, (with `"FINISHED"` having been eliminated some time ago), there's `session.ready` (a boolean), and `session.lastLoadingError` (which is `null` if the last load succeeded, and an error string otherwise).  So `session.ready` now uses this new system (actually it uses it by means of a bunch of new `session` selectors I added to report on this).  So does the new synchronization method I added, `session.readyAgainAfterLoading`, which works similarly to `session.doneStepping`.  (Ideally `session.ready` and `session.readyAgainAfterLoading` would be unified, both are checking the session ready state, but I didn't see any easy way to do that.)

Also, to prevent timing problems, the `ready` promise is now set up in the constructor and stored in `session._ready`, and `session.ready()` now just awaits the stored promise.

Note that to handle the `LOAD_TRANSACTION` action, the `session` submodule now has a listener.  However I didn't want to mess with the existing default exports so unfortunately it is irregularly named and forked separately; it's not just named `saga` like the other listeners.

By the way, while I was doing all this, I simplified some things in the `trace` submodule to just pass data around normally via parameters and return values rather than via actions, as was being done.

Next, `fetchTx` no longer takes a provider; instead, there's a new `web3` saga, `init`, which takes the provider and performs the initialization.  Web3 initialization is now a separate step, that is performed whether there is a transaction or not.

Similarly, the arguments to the `session` constructor have had their order changed so that `txHash` is last, since it's now optional.  Since the constructor is meant to be private (one should use `forTx` or `forProject`), this isn't a breaking change.

Other other minor change: I added to `lib/commands/debug.js` a function to change the prompt, since the prompt needs to change when we load or unload a transaction.

Also, I added two brief tests of loading and unloading.

I think that covers the bulk of it.  So now the question: Why do you have to unload a transaction before loading a new one?  It would be easy enough to automatically unload the existing transaction when you load a new one, after all.

The reason is this: Suppose you have a transaction loaded, and you attempt to load a new one, but an error occurs.  What should happen?  Ideally your previous state would be restored, but I'm not sure how to do that.  And if we just dump you into transactionless mode, well, that'd be confusing for the user.  By requiring manual unloading of a transaction before a new one is loaded, we avoid this problem!  And then if an error occurs you're *already* in transactionless mode so there's no problem; the result (staying in transactionless mode) just seems natural and not confusing.

Anyway, we now have transactionless mode!  Which can be used both to skip recomputation right now, and which later will form an important base to start building reconstruct mode.


CplXhristianUSMC:

CplXhristianUSMC:

haltman-at:
At @gnidan's suggestion, I've added spinners to the various loading steps.  I also added a spinner for stepping, too (it goes away when it's done).

Note that doing this required... so, this isn't possible with the existing `forTx` or `forProject` interfaces, because they wait for the debugger to be fully ready, and to make the project spinner work we need to wait for the debugger to be only *partially* ready.  So, I added a new, rawer interface, `unreadyDebugger` (with warnings that you probably shouldn't use this) and had the CLI use that, instead.

A warning, this does have the side-effect that our CLI can no longer really be considered a reference implementation, since other people probably don't want to use `unreadyDebugger`, they probably just want to use `forTx` or `forProject`.

haltman-at:
OK, I've pushed a resolution to this merge conflict, but it was an absolute mess, so I'd appreciate it if the result could be reviewed for coherence (and nothing getting dropped).  Also in doing the merge I took out a `if (err) return done(err);` that was left in there because it was causing an error because `err` is not defined now; I'm assuming that was the right thing to do?  (@eggplantzzz?)  Also I'd like if it could be reviewed for making sure I didn't screw up the new structure.

(Btw @gnidan you should absolutely feel free to make those other changes you were talking about later.)

packages/truffle-core/lib/commands/debug.js
```diff
@@ -98,23 +101,65 @@ var command = {
...
           .then(function(bugger) {
+            debug("about to connect");
             return bugger.connect();
           })
           .catch(done);
 
         sessionPromise
           .then(async function(session) {
-            if (err) return done(err);
+            debug("starting session function");
```
eggplantzzz: This line was the culprit? Makes sense to me.

packages/truffle-debugger/lib/controller/selectors/index.js
```diff
@@ -48,28 +48,45 @@ const controller = createSelectorTree({
...
     /**
      * controller.current.finished
      */
-    finished: createLeaf([trace.finished], identity)
+    finished: createLeaf([trace.finished], identity),
+
+    /**
+     * controller.current.loaded
+     */
+    loaded: createLeaf([trace.loaded], identity)
```
gnidan: How do you feel about nesting `finished` and `loaded` into `controller.current.trace`? Seeing `controller.current.loaded` struck me as a bit implicit.

packages/truffle-debugger/lib/controller/sagas/index.js
```diff
@@ -108,6 +113,9 @@ function* stepNext() {
  * step.
  */
 function* stepInto() {
+  if (!(yield select(controller.current.loaded))) {
+    return;
+  }
```
gnidan: There's a lot of duplication here. Did you have any thoughts about how we might abstract this? Just curious.

packages/truffle-debugger/lib/debugger.js
```diff
@@ -44,25 +43,64 @@ export default class Debugger {
...
+      options.provider,
+      txHash
     );
 
     try {
       await session.ready();
+      debug("session ready");
     } catch (e) {
-      throw e;
+      debug("error occurred, unloaded");
```
gnidan: Maybe worth adding `unloaded transaction` or even `unloaded transaction (if any)`? What do you think? I understand that the `(if any)` would be more correct, but it might make it more confusing. Not sure - might be worth just keeping as-is for simplicity.

packages/truffle-debugger/lib/session/index.js
```diff
@@ -1,12 +1,16 @@
 import debugModule from "debug";
+//returns true on success, false on already unloadee
```
gnidan: typo

gnidan: (what is this comment even doing here?)

packages/truffle-debugger/lib/controller/selectors/index.js
haltman-at: Makes sense, I guess.  Can do that.

packages/truffle-debugger/lib/debugger.js
haltman-at: Its just a debug message, I don't think it matters.

packages/truffle-debugger/lib/session/index.js
haltman-at: Whoa. Looks like an accidental paste. Deleted.

packages/truffle-debugger/lib/controller/sagas/index.js
haltman-at: Hm... we could move the check to the listener saga.  That wouldn't affect internal calls, but, we shouldn't be calling these internally in the first place if there's no transaction loaded!  How's that sound?

PR #1970, opened 2019-05-01
Fix Geth in CI

cgewecke:
The geth CI job has been broken for a [few weeks](https://travis-ci.org/trufflesuite/truffle/builds/516498633) because geth now [requires](https://ethereum.stackexchange.com/questions/69435/error-account-unlock-with-http-access-is-forbidden-when-unlock-an-account-in-ge) you explicitly allow accounts to be unlocked via web3.  CI creates and funds several accounts using this method in [geth-accounts.js](https://github.com/trufflesuite/truffle/blob/develop/scripts/geth-accounts.js) 

There *is* a test failure in the production migrations tests [here](https://travis-ci.org/trufflesuite/truffle/jobs/526705708#L1418), but it's legit - there's actually a problem there.

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -21,6 +21,7 @@ run_geth() {
     --nodiscover \
     --dev \
     --dev.period 0 \
+    --allow-insecure-unlock \
     --targetgaslimit '7000000' \
     js ./scripts/geth-accounts.js \
     > /dev/null &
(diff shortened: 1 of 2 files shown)


gnidan:
Hmm looks like there might be more reasons than just that for why it's failing. Maybe things have changed that have gone unnoticed cause it's been failing for so long.

gnidan:
Thanks for this, by the way!

PR #1972, opened 2019-05-01
Internal improvement: truffle-artifactor

CruzMolina:
Tidy up the deps & add some code coverage.


packages/truffle-artifactor/test/methods.js
```diff
@@ -0,0 +1,40 @@
...
+    artifactor = new Artifactor("/some/path");
+
+    artifactor
+      .save({
+        "abi": [],
+        "bytecode": "0xabcdef",
+        "networks": {
+          3: { address: "0xe6e1652a0397e078f434d6dda181b218cfd42e01" }
+        },
+        "x-from-dependency": "somedep"
```
eggplantzzz: You might also want to include a `then` that contains something like `assert(false)` to guarantee that it fails. Otherwise, if it doesn't throw an error it is making no assertion.

packages/truffle-artifactor/test/methods.js
```diff
@@ -0,0 +1,40 @@
...
+
+    artifactor
+      .saveAll({
+        "contractName": "Example",
+        "abi": [],
+        "bytecode": "0xabcdef",
+        "networks": {
+          3: { address: "0xe6e1652a0397e078f434d6dda181b218cfd42e01" }
+        },
+        "x-from-dependency": "somedep"
```
eggplantzzz: Same here.

PR #1973, opened 2019-05-01
Internal improvement: Geth CI

CruzMolina:
Attempt to ensure truffle-geth integration by removing the `geth` job as an allowed failure.

Also updates err msg's and cleans up test log output.


packages/truffle-contract/test/util.js
```diff
@@ -42,6 +42,12 @@ var util = {
     };
     const options = {
       contracts_directory: path.join(__dirname, "sources"),
+      logger: {
+      log(stringToLog) {
+        this.loggedStuff = this.loggedStuff + stringToLog;
+      },
+      loggedStuff: ""
+    },
```
gnidan: Why didn't prettier get the formatting here? Strange.

Anyway, do you think it's worth respecting the `quiet` option and including that as a convenience?

CruzMolina: Oh yeah! Forgot about that.

PR #1974, opened 2019-05-01
Internal improvement:  truffle-workflow-compile tests

CruzMolina:
Updates `truffle-workflow-compile` tests to align with the reformat to migrations (#1955).

### DIFF
--- a/packages/truffle-workflow-compile/README.md
+++ b/packages/truffle-workflow-compile/README.md
@@ -23,10 +23,4 @@ const config = {
 Contracts.compile(config)
   .then(() => console.log("Compilation complete!"))
   .catch(e => console.error(e))
-
-// callback alternative
-Contracts.compile(config, (err, output) => {
-  if (err) console.error(err);
-  console.log(output)
-});
 ```
--- a/packages/truffle-workflow-compile/README.md
+++ b/packages/truffle-workflow-compile/README.md
@@ -20,13 +20,13 @@ const config = {
 
 // compiles contracts found in contracts_directory,
 // saves them in contracts_build_directory
-Contracts.compile(config, () => {
-  console.log("Compilation complete!");
-});
+Contracts.compile(config)
+  .then(() => console.log("Compilation complete!"))
+  .catch(e => console.error(e))
 
-// alternative making use of the expected callback
+// callback alternative
 Contracts.compile(config, (err, output) => {
   if (err) console.error(err);
-  else console.log(output)
+  console.log(output)
 });
 ```
--- a/packages/truffle-workflow-compile/test/index.js
+++ b/packages/truffle-workflow-compile/test/index.js
@@ -59,5 +59,5 @@ describe("Contracts.compile", () => {
           )
       );
     });
-  }).timeout(3000);
+  }).timeout(5000);
 });
--- a/packages/truffle-workflow-compile/test/index.js
+++ b/packages/truffle-workflow-compile/test/index.js
@@ -1,14 +1,25 @@
 const Contracts = require("../index");
 const assert = require("assert");
-const { existsSync } = require("fs");
+const { existsSync, removeSync } = require("fs-extra");
+const { join } = require("path");
 
 let config;
 
 beforeEach(() => {
   config = {
     contracts_directory: "./test/sources",
-    contracts_build_directory: "./test/build"
+    contracts_build_directory: "./test/build",
+    logger: {
+      log(stringToLog) {
+        this.loggedStuff = this.loggedStuff + stringToLog;
+      },
+      loggedStuff: ""
+    }
   };
+});
+
+after(() => {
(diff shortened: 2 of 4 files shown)


packages/truffle-workflow-compile/test/index.js
```diff
@@ -25,30 +36,28 @@ describe("Contracts.compileSources", () => {
...
+      contractName = Object.keys(contracts)[0];
+      assert(
+        existsSync(`${config.contracts_build_directory}/${contractName}.json`)
+      );
     });
 
     // compile again
     config.all = true;
-    Contracts.compile(config, (err, output) => {
+    await Contracts.compile(config, (err, { outputs }) => {
```
fainashalts: I think the singular `output` is used elsewhere. Any reason it is plural here?

packages/truffle-workflow-compile/test/index.js
```diff
@@ -25,30 +36,28 @@ describe("Contracts.compileSources", () => {
...
+        outputs.solc[0] ===
+          join(
             `${process.cwd()}/${config.contracts_directory}/${contractName}.sol`
-        );
-      }
+          )
+      );
     });
-  });
+  }).timeout(3000);
```
fainashalts: Was this failing without the added timeout? Seemed to be passing on my local yesterday.

packages/truffle-workflow-compile/test/index.js
```diff
@@ -25,30 +36,28 @@ describe("Contracts.compileSources", () => {
 });
 
 describe("Contracts.compile", () => {
-  it("when config.all is true, all contracts in contracts_directory are recompiled", () => {
+  it("when config.all is true, all contracts in contracts_directory are recompiled", async () => {
     let contractName;
 
     // initial compile
-    Contracts.compile(config, (err, { contracts }) => {
+    await Contracts.compile(config, (err, { contracts }) => {
```
eggplantzzz: This should either take a callback or use a Promise interface. If you use `await`, then this should be 
```
const { contracts } = await Contracts.compile(config);
```

packages/truffle-workflow-compile/test/index.js
eggplantzzz: Same here.

packages/truffle-workflow-compile/test/index.js
```diff
@@ -25,30 +36,28 @@ describe("Contracts.compileSources", () => {
...
       if (err) assert.fail(err);
```
eggplantzzz: Also, I'm not quite sure what the purpose of this `assert.fail(err)` is.

eggplantzzz:
The point of that interface stuff is that you should EITHER pass it a callback OR use it with a Promise interface...but definitely not both.

packages/truffle-workflow-compile/test/index.js
CruzMolina: The point is to fail showing `err` as the `AssertionError`, but I like your suggestion the other day to do something more like `assert(false, ${err})`.

packages/truffle-workflow-compile/test/index.js
CruzMolina: @fainashalts it looks like `output` is only used in the README in the context of logging the entire `output` object.

it's plural because of the object structure Tyler is showing above. the fancy ES6 destructuring returns the `outputs` key directly rather than having to do something like `output.outputs.solc[0]`.

packages/truffle-workflow-compile/test/index.js
CruzMolina: Same on my local. Travis is slowww.

packages/truffle-workflow-compile/test/index.js
CruzMolina: Sweet, thanks for clarifying the interface stuff @eggplantzzz !

packages/truffle-workflow-compile/test/index.js
fainashalts: Ah I see now. Thanks for clarifying!

packages/truffle-workflow-compile/README.md
```diff
@@ -20,13 +20,13 @@ const config = {
...
 // saves them in contracts_build_directory
-Contracts.compile(config, () => {
-  console.log("Compilation complete!");
-});
+Contracts.compile(config)
+  .then(() => console.log("Compilation complete!"))
+  .catch(e => console.error(e))
 
-// alternative making use of the expected callback
+// callback alternative
```
eggplantzzz: I wonder if going forward we want to even deprecate the callback interface. I'm not sure.

CruzMolina: Yeah, not sure if I should leave that example in for now or strip it out.

gnidan: Toss it. No one wants to hear about callbacks.

PR #1982, opened 2019-05-02
Internal improvement: Script release publishing

CruzMolina:
This PR builds on #1979 to streamline the release process 💥.

Now, calling `yarn publish-release` from the truffle root project directory will force us to log in to `npm`, check our npm user package permissions against trufflesuite org's package permissions (access rights as well), run `lerna publish`, and then handle the post-release git branch merging/pushing, and post-release npm installation.

⛵️


gnidan:
@CruzMolina this is awesome. Could we make it accept NPM tag arguments?

Also you might be interested in packages/truffle/scripts/preRelease or whatever it is?

CruzMolina:
@gnidan Do you mean allow a `dist-tag` to be passed so we can do something like...

```lerna publish --dist-tag next```  and more easily publish prereleases or betas?

gnidan:
Yep exactly @CruzMolina

CruzMolina:
I imagine we'd want that in a separate PR/script. For example, I don't think we'd want to do the post-git merging/pulling/pushing between `develop` and `master`. For sanity/safety purposes it also seems overall safer to have a separate script command (`yarn publish-prerelease`).

I can also take a look at the existing prerelease script (glanced over it briefly).

gnidan:
Doin' it live

PR #1990, opened 2019-05-06
Workspace artifacts loader scratch

fainashalts:


packages/truffle-db/src/loaders/artifacts/index.ts
```diff
@@ -0,0 +1,255 @@
...
+            contents: compilationsData[contract]["source"], 
+            sourcePath: compilationsData[contract]["sourcePath"]
+          };
+
+          let sourceId = generateId({
+            contents: compilationsData[contract]["source"], 
+            sourcePath: compilationsData[contract]["sourcePath"]
+          });
+          
+          sources.push({contents: compilationsData[contract]["source"], sourcePath: compilationsData[contract]["sourcePath"]});
```
gnidan: Thinking we should move this source stuff to a separate `await this.loadSources()` method, called by `loadCompilations`. (just to clean up the code)

packages/truffle-db/src/loaders/artifacts/index.ts
```diff
@@ -0,0 +1,255 @@
...
+                  json: JSON.stringify(compilationsData[contract]["ast"]),
+                }
+              }
+            ],
+            sources: [
+              { id: sourceId }
+            ]
+          }
+
+          compilationsArray.push(compilationObject);
```
gnidan: I think this isn't what we want... we should have one single compilation with multiple sources, since we only invoke truffle-workflow-compile the once.

packages/truffle-db/src/loaders/artifacts/test/index.ts
```diff
@@ -0,0 +1,173 @@
...
+          }
+        }
+      }
+    } = await db.query(GetWorkspaceCompilation, { id: expectedId });
+
+    expect(version).toEqual(Build.compiler.version);
+    expect(sources[0].contents).toEqual(Build.source);
+    expect(contracts[0].name).toEqual(Build.contractName);
+  })
+});
```
gnidan: Can we make this test check to ensure that a single compilation has exactly the correct sources? (no more, no less)

packages/truffle-db/src/loaders/artifacts/index.ts
fainashalts: I initially thought the same but workspace was expecting an array of compilations and I remembered you saying that we should write that code to accept plurals - but of course that is different than this concrete implementation of adding a compilation. Will fix this, thanks!

As a side note, it looks like compiler info is returned from truffle-workflow-compile as part of each contract. I imagine all contracts should have the same compiler, right? So would it be alright to get this info via the first contract in that array? Or perhaps you have something else in mind?

packages/truffle-db/src/loaders/artifacts/test/index.ts
fainashalts: Added `expect(sources.length).toEqual(1);` here, let me know if you had something else or more involved in mind!

packages/truffle-db/src/loaders/artifacts/index.ts
gnidan: In this case, it's safe to rely on the compiler info from the first contract, since it's all part of the same compilation!

packages/truffle-db/src/loaders/artifacts/test/index.ts
gnidan: Probably best to:
- make sure this test adds a compilation with more than one source
- check the IDs match

Can we have this test use magic-square, and then check for the three sources in that project?

fainashalts: Totally! I had to make some additional changes all around to be sure this works as expected with the mocks, and also I pushed some changes to `solidity-magic-square` so that we'll be able to test for a link reference in the future. The contracts/build files I added here are based on the branch I pushed to `solidity-magic-square`.

Thanks, let me know if this is more what you had in mind!

PR #1991, opened 2019-05-06
Throw exception if contract decoder can't locate contract's node

haltman-at:
This PR causes the contract decoder constructor to throw an exception in case it can't locate the contract node for the contract being decoded, so that this case doesn't crash Ganache.

(Note that if a contract node can't be located *other* than the one for the contract being decoded, it's simply skipped; this is already existing behavior.  But you can't just skip the contract which is supposed to be decoded!)


haltman-at:
Oh also @davidmurdoch this doesn't really change that much since you're already catching general exceptions, but figure I should point this out to you regardless.

PR #1995, opened 2019-05-08
Internal improvement: truffle-box

CruzMolina:
Tidies up the deps, adds some code coverage, misc es6 stuff, and removes outdated code.


packages/truffle-box/box.js
```diff
@@ -111,17 +110,16 @@ const Box = {
       tmp.setGracefulCleanup();
     }
 
-    tmp.dir({ unsafeCleanup }, function(err, dir) {
+    tmp.dir({ unsafeCleanup }, (err, dir) => {
```
eggplantzzz: If you are feeling ambitious, I would be interested in having the synchronous version of this method used :) If you are not interested that is also fine. Check it out --> https://www.npmjs.com/package/tmp

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -10,37 +10,33 @@ const inquirer = require("inquirer");
...
     );
 
-    var options = {
+    const options = {
       method: "HEAD",
-      uri: "https://" + configURL.host + configURL.path
+      uri: `https://${configURL.host}${configURL.path}`
     };
-    request(options, function(error, r) {
+    request(options, (error, { statusCode }) => {
```
eggplantzzz: Would we consider changing some of these uses of `request` to something that supports promises? Something like `request-promise` perhaps? 
https://www.npmjs.com/package/request#promises--asyncawait

eggplantzzz: Love the variable change by the way haha

packages/truffle-box/box.js
```diff
@@ -111,21 +110,22 @@ const Box = {
...
+        `https://github.com/trufflesuite/truffle-init-${name}`,
+        tmpDir.name,
+        options
+      ).then(() => {
+        const config = Config.load(
+          path.join(tmpDir.name, "truffle-config.js"),
+          {}
+        );
+        return callback(null, config);
+      });
```
eggplantzzz: I think you will need to keep a catch here that just throws the error, otherwise node will whine about an unhandled Promise rejection in the case of an error. So throw the error in the catch and it will be handled in the catch of the try/catch.

packages/truffle-box/lib/config.js
```diff
@@ -1,9 +1,7 @@
-var fs = require("fs-extra");
+const fs = require("fs-extra");
```
eggplantzzz: How do you feel about making all instances of the fs-extra library set to `fse` (the docs recommend this as an alternative if you don't want to use the variable fs)? I like making it clear that it is something different from the classic `fs` library but I have no problem with just leaving it as `fs` as well.

packages/truffle-box/lib/utils/index.js
```diff
@@ -29,34 +29,26 @@ module.exports = {
...
-
-        prepareSpinner.succeed();
-        resolve({
-          path: path.join(dir, "box"),
-          cleanupCallback
-        });
-      });
-    });
+    } catch (error) {
+      await prepareSpinner.fail();
```
eggplantzzz: This method is not async.

packages/truffle-box/lib/utils/index.js
```diff
@@ -29,34 +29,26 @@ module.exports = {
...
-        unsafeCleanup: true
+  setUpTempDirectory: async () => {
+    const prepareSpinner = await ora("Preparing to download").start();
+    const options = {
+      dir: cwd,
+      unsafeCleanup: true
+    };
+    try {
+      const tmpDir = await tmp.dirSync(options);
+      await prepareSpinner.succeed();
```
eggplantzzz: This method is also not async.

packages/truffle-box/lib/utils/index.js
```diff
@@ -29,34 +29,26 @@ module.exports = {
...
   },
 
-  setUpTempDirectory: () => {
-    const prepareSpinner = ora("Preparing to download").start();
-    return new Promise((resolve, reject) => {
-      const options = {
-        dir: cwd,
-        unsafeCleanup: true
+  setUpTempDirectory: async () => {
+    const prepareSpinner = await ora("Preparing to download").start();
```
eggplantzzz: Same as below.

packages/truffle-box/lib/utils/index.js
```diff
@@ -29,34 +29,26 @@ module.exports = {
...
+    return await config.read(configPath);
   },
 
-  setUpTempDirectory: () => {
-    const prepareSpinner = ora("Preparing to download").start();
-    return new Promise((resolve, reject) => {
-      const options = {
-        dir: cwd,
-        unsafeCleanup: true
+  setUpTempDirectory: async () => {
```
eggplantzzz: Check me on this, but you should make this method just plain old synchronous. It gets called from `truffle-box/box` and it looks like you can just remove the await from that file. Simplification!

packages/truffle-box/lib/utils/index.js
```diff
@@ -71,7 +63,7 @@ module.exports = {
       setUpSpinner.succeed();
     } catch (error) {
       setUpSpinner.fail();
-      throw new Error(error);
+      throw error;
```
eggplantzzz: The removal of these `new Error(error)` is something I never thought to do :)  Cool

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -1,72 +1,63 @@
 const fs = require("fs-extra");
 const path = require("path");
 const ghdownload = require("github-download");
-const request = require("request");
+const rp = require("request-promise");
```
eggplantzzz: Eww, terrible variable name. But the docs say it...

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -1,72 +1,63 @@
...
-              " doesn't exist. If you believe this is an error, please contact Truffle support."
-          )
+  const options = {
+    method: "HEAD",
+    uri: `https://${configURL.host}${configURL.path}`,
+    simple: false,
+    resolveWithFullResponse: true
+  };
+
+  await rp(options)
```
eggplantzzz: Here I would either use `async/await` syntax OR `then` chains. I think it is confusing to use both and defeats the purpose of using `await` to begin with.

packages/truffle-box/test/utils-index.js
```diff
@@ -0,0 +1,13 @@
...
+const utils = require("../lib/utils");
+
+describe("utils", () => {
+  it("setUpBox throws when passed an invalid boxConfig", () => {
+    let boxConfig = {};
+
+    utils
+      .setUpBox(boxConfig)
+      .then(() => assert(false, "didn't throw!"))
+      .catch(e => assert(e.stack.includes("Error")));
```
eggplantzzz: You might want to test against this error a little more strictly to ensure that the type of error you expect is being thrown. Currently, if this test throws for probably ANY reason, it will pass. This could lead to false positives.

packages/truffle-box/test/utils-unbox.js
```diff
@@ -0,0 +1,30 @@
...
+      utils.installBoxDependencies(boxConfig)
+      .then(() => assert(true))
+      .catch(() => assert(false, "threw an error!"))
+    })
+
+    it("promise rejects if 'post-unpack' key passed unexecutable value", () => {
+    let boxConfig = { hooks: { "post-unpack": "doBadStuff" } }
+      utils.installBoxDependencies(boxConfig)
+      .then(() => assert(false, "should have thrown!"))
+      .catch(({ stack }) => assert(stack.includes("Error")))
```
eggplantzzz: Same here I suppose.

packages/truffle-box/test/utils-unbox.js
```diff
@@ -0,0 +1,30 @@
...
+    })
+  });
+  describe("fetchRepository", () => {
+    it("throw when passed non-strings and non-objects", () => {
+      utils.fetchRepository(false)
+        .then(() => assert(false, "should have thrown!"))
+        .catch(({ stack }) => assert(stack.includes("Error")))
```
eggplantzzz: Same

packages/truffle-box/test/utils-unbox.js
```diff
@@ -0,0 +1,30 @@
...
+      utils.fetchRepository(123214)
+        .then(() => assert(false, "should have thrown!"))
+        .catch(({ stack }) => assert(stack.includes("Error")))
```
eggplantzzz: Same

packages/truffle-box/lib/config.js
CruzMolina: Looks like we use both (`fs & `fs-extra`) throughout the codebase, so makes sense to differentiate.

packages/truffle-box/lib/utils/index.js
CruzMolina: Ah, looks like it can be sync, but still need to keep the `await` in `box.js` within the `unbox` method.

packages/truffle-box/lib/utils/index.js
CruzMolina: Yeah, I've noticed there's a bunch of `new Error(err)` stuff thrown around in the codebase that messes with useful error reporting/output.

packages/truffle-box/lib/utils/unbox.js
CruzMolina: Hmm, I'm getting wonky unhandled rejection behavior that appears related to the Bluebird promises powering `request-promise`. Going to give `request-promise-native` a shot and see if native ES6 promises resolve the issue. Otherwise, it appears the current `request` callback code is better suited to handling the current various error types.

eggplantzzz: Are you handling the rejection cases?

packages/truffle-box/lib/config.js
eggplantzzz: Yeah, I mean the docs for `fs-extra` say that it is meant as a "drop in" replacement for `fs`. I'm always into the idea of making things extra clear.

packages/truffle-box/lib/utils/unbox.js
CruzMolina: Even when I catch and throw after the rejection I still get an unhandled promise rejection response.

I _think_ I found a solution, but not really a fan of it. About to find out if `request-promise-native` makes a difference.

CruzMolina: Hmm, nm. In general the `request-promise` stuff doesn't seem to be integrating well and the current callback behavior appears to fit the error handling we output to users better. Going to see if I can just `promisify` `request`.

CruzMolina: Ahh, turns out there was an issue with the structuring of the sandbox method. Still can't seem to get a promise request to play nice w/o a mix of `await` and a `then` chain. Reverting the code to use a request callback for now.

packages/truffle-box/lib/config.js
```diff
@@ -19,14 +17,14 @@ function setDefaults(config) {
 }
 
 function read(path) {
-  return fs
+  return fse
     .readFile(path)
     .catch(() => "{}")
```
eggplantzzz: Kind of a weird thing to return, huh?

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -53,20 +49,20 @@ function verifyURL(url) {
...
+      .on("end", () => {
         accept();
       });
   });
 }
 
-function prepareToCopyFiles(tempDir, boxConfig) {
-  const needingRemoval = boxConfig.ignore || [];
+function prepareToCopyFiles(tempDir, { ignore }) {
+  const needingRemoval = ignore;
```
eggplantzzz: Do we need to keep this behavior of having the default for `needingRemoval` set to an empty array?

packages/truffle-box/test/box.js
```diff
@@ -1,6 +1,6 @@
 const path = require("path");
-const regularFs = require("fs");
-const fs = require("fs-extra");
+const fs = require("fs");
+const fse = require("fs-extra");
```
eggplantzzz: `fs-extra` contains all the methods of `fs` so we won't need to use `fs` if we use `fs-extra`. Also, all the async methods return promises if you don't provide a callback.

packages/truffle-box/lib/utils/unbox.js
CruzMolina: Looks like `ignore` is either an empty array or an actual list of items to be ignored during unboxing (depending on the given truffle box being unpacked).

packages/truffle-box/lib/config.js
CruzMolina: Yeah, taking a second look at this, it looks like this is a fail-safe to protect truffle box writers/developers from themselves. i.e. cases where they botched the `truffle-box` or `truffle-init` json file. At this point in the flow, we've already identified the repo being unboxed is a truffle box (existence of `truffle-box.json`).

packages/truffle-box/test/box.js
CruzMolina: ah, `fs` is used in here as a stub since `Box.checkDir` uses `fs` and not `fse`.

Makes sense to simplify.

CruzMolina:
Closing since this has been successfully separated into different easier-to-chew-on PRs. 😋

PR #1999, opened 2019-05-09
 Prevent debugger from landing on pseudo-nodes with no id 

haltman-at:
This PR fixes a crash that could occur when debugging contracts with inline assembly.  One of the attributes of an `InlineAssembly` node is `externalReferences`, and nested inside `externalReferences` one can find certain objects I'm calling "pseudo-nodes".  They're not actual nodes -- they lack an `id` and a `nodeType`.  But they do have a `src`, which could cause the debugger to land on them.  The lack of `id` would then cause a crash.

This PR fixes the problem by excluding such pseudo-nodes from the set of nodes the debugger can land on, by only putting nodes with actual IDs into the `IntervalTree`.  This means that instead of landing on these pseudo-nodes, the debugger will land on the `InlineAssembly` node they're a part of.

Note that I'm assuming that this is the only way to get "pseudo-nodes".  If it's not, this might not be the correct solution for those other cases.  But it should be fine here.

### DIFF
--- a/packages/truffle-debugger/lib/ast/map.js
+++ b/packages/truffle-debugger/lib/ast/map.js
@@ -30,7 +30,12 @@ export function rangeNodes(node, pointer = "") {
   } else if (node instanceof Object) {
     let results = [];
 
-    if (node.src) {
+    if (node.src !== undefined && node.id !== undefined) {
+      //there are some "pseudo-nodes" with a src but no id.
+      //these will cause problems, so we want to exclude them.
+      //(to my knowledge this only happens with the externalReferences
+      //to an InlineAssembly node, so excluding them just means we find
+      //the InlineAssembly node instead, which is fine)
       results.push({ pointer, range: getRange(node) });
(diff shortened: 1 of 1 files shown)


haltman-at:
Interesting question.  We probably could, but I think I'd prefer not to; that's a bit more complex than I'd like to be doing here, I think.

haltman-at:
Actually, thinking about it further, we definitely *don't* want to do that.  Skipped nodes are ones skipped by the interface, but they're often still quite necessary to pass through internally.  Like, variable declarations are skipped, but obviously we need the debugger to pass through those!  Also, note that while I described this change as preventing the debugger from landing on these nodes, it also prevents it from considering one as the *next* node when that matters (that was actually the immediate cause of the crash).  We definitely wouldn't want skipped nodes being skipped for those purposes!

So in short, no, we can't do that here.

gnidan:
Ah of course, duh!

PR #2000, opened 2019-05-09
Internal Improvement: Migration tests

eggplantzzz:
Write a couple of unit tests for detectDryRunSettings


fainashalts:
These tests look good to me, but for the sake of completeness, I wonder if we should also add a test that runs(preferably mocks) a migration and checks whether the `determineDryRunSettings` function is run and returns the expected value?

eggplantzzz:
@CruzMolina I wrote some tests for the else clause that you referenced. I don't think it needs to be `command.determineDryRunSettings`. If you could explain why you think this, I would be happy to consider it.

CruzMolina:
Sweet! Saw the comments in #2012. Makes sense it would be a scoping bug for the `deploy` alias for `migrate`.

PR #2001, opened 2019-05-09
Internal improvement: skip non-gethy tests during geth job

CruzMolina:
This PR drops in a `@standalone` keyword into truffle scenario tests that only test truffle's bundle integration and adds the keyword to the mocha grep. Trims geth job by ~5-6 min.

~Open to having the keyword be something else (`@truffle`?) for future jobs testing other client integrations (quorum? parity? corda?), but this probably suffices for now.~


gnidan:
Agh, does this mean ALL tests need to be marked `@non-geth`? Can we look at mocha's mark/tag/whatever they call it and see if we can do it with a proper boolean "not `@geth`"?

CruzMolina:
@gnidan the `GETH` job only runs `packages/truffle` and `truffle-contract` tests. The `truffle-contract` tests run in that job are already marked as being `@geth` specific (and I imagine would work or _should_ work w/ other clients we want to integrate with [maybe we'd need to rename that eventually to something like `@client` or `@chain`]).

it looks there's only a `mocha --grep` being used for select jobs, so adding `@non-geth` to `packages/truffle` currently doesn't appear to change anything except for the `GETH` job.

We only need to mark tests in `packages/truffle` that only test truffle proper as `@non-geth`.

CruzMolina:
Ah shoot, so changing this to not tag the truffle stuff and instead tag gethy tests as `@geth` gets a little tricky because there's already an `@geth` tag or two being used and the test script already skips these tests using an `@geth` invert on the other jobs.

I think we should just rename all the `@non-geth` references to `@truffle`.

packages/truffle/scripts/test.sh
```diff
@@ -3,7 +3,7 @@
 set -o errexit
 
 if [ "$GETH" == true ]; then
-  mocha --timeout 50000 --grep @ganache --invert --colors $@
+  mocha --timeout 50000 --grep '@ganache|@standalone' --invert --colors $@
```
gnidan: This isn't just `@geth`?

CruzMolina: i guess it could be, but I figure we want to make sure anything generalized having to do with migrations works with geth on this job.

gnidan:
good branch name btw

Issue #1993, opened 2019-05-07
Tests run without a network


tcurdt:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Truffle runs and passes(?) the Tests without connecting to the specified network.

## Steps to Reproduce

test:
```
const FooContract = artifacts.require("FooContract");
contract("FooContract", async instances => {
  it("has owner when new", async () => {
    let instance = await FooContract.new()
    let owner = await instance.owner.call();
    expect(owner).to.exist;
  });
  it("has owner", async () => {
    let instance = await FooContract.deployed();
    let owner = await instance.owner.call();
    expect(owner).to.exist;
  });
});
```

truffle.js:
```
module.exports = {
    networks: {
        geth: {
            host: "127.0.0.1",
            port: 8545,
            network_id: "*"
        }
    }
};
```
Connecting to a network **not** specified in the `truffle.js` fails as expected:
```
truffle console --network foo
Unknown network "foo". See your Truffle configuration file for available networks.
Truffle v5.0.14 (core: 5.0.14)
Node v12.1.0
```

Running the test surprisingly still works
```
truffle test --network foo ./test/*.js
running tests
Using network 'foo'.
Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.
  Contract: FooContract
    ✓ has owner when new (108ms)
    ✓ has owner
```

## Expected Behavior

I don't see how the tests can run without connecting to the specified network.
It should give the same `Unknown network "foo"` error as when connecting with the console. And it of course should run the tests on the specified network.

```
truffle test --network geth ./test/*.js
```

This might be just a user error - but neither the logging nor documentation seem to cover this.

## Actual Results

The tests just run and surprisingly all pass.

## Environment

* Operating System:
```
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.6
BuildVersion:	17G6030
```
* Ethereum client:
* Truffle version (`truffle version`):
```
Truffle v5.0.14 (core: 5.0.14)
Solidity v0.5.0 (solc-js)
Node v12.1.0
Web3.js v1.0.0-beta.37
```
* node version (`node --version`): `v12.1.0`
* npm version (`npm --version`): `6.9.0`

CruzMolina:
[...]

PR #2004, opened 2019-05-10
Fix: Detect test network existence before test run

CruzMolina:
Currently users can pass whatever network they like during testing (i.e. `truffle test --network ghostNetwork`). Regardless of whether network configs are present in their truffle config file, truffle will act as if that network actually exists 👻 .

Per #1993 , this behavior can be confusing 😕 so this PR ensures if a network is passed during testing, truffle will proceed to check if the network actually exists.

In general this code should probably be cleaned up and revamped, but for now I think this will suffice.

### DIFF
--- a/packages/truffle-core/lib/commands/test.js
+++ b/packages/truffle-core/lib/commands/test.js
@@ -63,6 +63,8 @@ var command = {
 
     if (!config.network) {
       config.network = "test";
+    } else {
+      Environment.detect(config).catch(done);
     }
 
     var ipcDisconnect;


fainashalts:
What happens if the config doesn't exist in this scenario? Wondering if we should console out an error or something.

CruzMolina:
@fainashalts Should have clarified, this fix has truffle output the same error as when you try to migrate to an undetected network.

i.e.
```
truffle test --network ghostNetwork
Unknown network "ghostNetwork". See your Truffle configuration file for available networks.
Truffle (unbundled) (core: 5.0.16)
Node v11.4.0
```

Issue #2005, opened 2019-05-10
Docker compilation twice as slow, seems to hang on large contract sets.


cgewecke:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

5.0.13 significantly reduced the speed of solc docker compilation for large contract sets. Performance when compiling [colonyNetwork](https://github.com/JoinColony/colonyNetwork) on a MacBook Air 1.8 ghz:  
+ 5.0.12 time  --> real	0m59.960s
+ 5.0.17 time  --> real	2m22.203s

The command also hangs before listing the contracts for more than a minute. Believe changes to the imports parsing / profiler in #1913 are the cause. The benefit of the solcjs parsing removed in that PR was that it was fast (see this tcoulter [comment](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-compile/parser.js#L9-L18) from the original implementation).

## Steps to Reproduce

```shell
# Install colonyNetwork
git clone https://github.com/JoinColony/colonyNetwork.git
cd colonyNetwork
yarn
git submodule update --init

# Install Truffle 5.0.12 and benchmark
npm uninstall -g truffle
npm install truffle@5.0.12
time truffle compile

# Clean build
rm -rf build

# Install Truffle 5.0.17 and benchmark
npm uninstall -g truffle
npm install truffle@5.0.17
time truffle compile
```

## Expected Behavior

Using solc docker is faster than using solc-js

## Additional notes / suggestion

#1913 resolved several issues. #1720 is one but it also allowed openzeppelin-solidity to [run a solc nightly job](https://github.com/OpenZeppelin/openzeppelin-solidity/pull/1732). Previous versions of truffle had been unable to meet that use case because they couldn't match a solc-js companion to the commit referenced by the docker nightly tag. 

There might be a simple internal fix for this but one option is to go back to using a js parser for imports parsing. [solidity-parser-antlr](https://github.com/federicobond/solidity-parser-antlr) is excellent and widely used across the js eco-system. It would be nice if it received grant support and was treated as a core piece of the stack. 

The risks of depending on a parser which lags solidity development (slightly) could be mitigated by E2E testing here, having a strong relationship with the project's maintainer and maybe asking solc to ping [solidity-antrl4](https://github.com/solidityj/solidity-antlr4) with grammar changes. Perhaps they could integrate it with their own testing or even add it as a dependency to the npm published solcjs.

[...]

PR #1913, opened 2019-04-15
Refactor truffle-compile/parser & enable docker/native solc offline compilation

CruzMolina:
This PR builds on @maxsam4 's POC outlined in #1720.

Now `truffle compile` no longer requires an internet connection if you've already `docker pull`'d the solc version you'd like to use in the config, or if you've already installed native `solc`. 

Comments/documentation have been added & updated to help clarify what `parseImports` is doing, some of the logic has been streamlined, & tests have been added to verify `truffle-compile/parser.js` works when specifying native & docker solc usage.

Resolves #1567 .

PR #2007, opened 2019-05-11
Enhancement: Use bundled solc-js for speedy import parsing

CruzMolina:
As @cgewecke points out in #2005, removing `solc-js` from `parseImports` significantly slowed docker compilations(!)

This PR re-enables bundled solc usage for speedy imports (while also retaining the benefits of #1913). It also cleans up some redundant error handling during compilation. The intention in this PR is for `solc-js` to only being used to parse imports. The primary compilation is still handled by whatever the user is specifying.


packages/truffle-compile/parser.js
```diff
@@ -50,19 +47,6 @@ module.exports = {
...
-    const nonImportErrors = errors.filter(
-      ({ formattedMessage }) => !formattedMessage.includes(importErrorKey)
-    );
-
-    // Should we try to throw more than one? (aside; we didn't before)
-    if (nonImportErrors.length > 0) {
-      throw new CompileError(nonImportErrors[0].formattedMessage);
-    }
-
-    // Now, all errors must be import errors.
```
cgewecke: @CruzMolina Sadly I *think* this might not work as expected. It looks like if a file has both an import error and a syntax error, solc exits with the syntax error and doesn't report the invalid import. It's like it won't consider the imports until other basic stuff is taken care of. To see an example of this try compiling
```solidity
pragma solidity 1.2.3;

import "I/am/not/a/file";

contract Problems {
  constructor() public {
  }

  just some words;

}
```
and logging the solc output at this location in the code. Then comment out `just some words` and try again. 

If the syntax error doesn't get thrown here it means that files with invalid imports would cycle through the profiler, which could be bad. Conversely, if you throw here the compilers really have to match.

CruzMolina: Ah, so it looks like `solc-js` will throw/return the first SyntaxError it encounters.

Thanks for pointing this out! I _think_ I have a solution 🤞 .

CruzMolina: @cgewecke I pushed some failsafe code to handle "fail even faster than bad imports" scenarios. After testing various scenarios locally things _appear_ fine, but I probably should add some Solidity Import scenario tests.

cgewecke: One scenario I'm worried about is compiler mismatching. For example suppose the user has specified an `0.4.24` compiler in their config and the built-in/parsing compiler is `0.5.0`. The users code will have statements which are legal in `0.4.24` but errors in `0.5.0`. If truffle throws for these statements while imports parsing, the ability to choose any compiler is broken. 


> Ah, so it looks like solc-js will throw/return the first SyntaxError it encounters.

I'm confused by how solc behaves fwiw. To me it seems like:
+  it throws certain classes of order in sequence   
+ solc expects the user to resolve these different error classes *in a series of compilation passes*. 
+ import paths errors are not necessarily caught on the first pass. I *think* the code sample above is an example of that.

Does fail safe filtering address that if the file has both a (perceived) syntax error *and* a bad import path?

As an aside - it looks like frangio is fixing a problem with bonkers async whilsts (mea culpa) in #2009 that touches the same files as this PR - it might make sense to merge that to develop if it looks ok and rebase here since it goes nicely with the profiler simplification you're doing.

cgewecke: Apologies, I've edited the above quite a bit - I find this error injection parsing strategy difficult to reason about.

gnidan: I concur with merging #2009 first

CruzMolina: @cgewecke I definitely hear your concern, can you provide another example? It appears that imaginary statement _is_ a valid ParserError in 0.4.24 (and appears to be a ParserError at least as far back as 0.4.0).

I've been trying to reproduce any kind of bug or red flag with the current solution and haven't come up with anything yet (throwing in "uncompilable" examples from Solidity's 0.5.0 Breaking Changes {https://solidity.readthedocs.io/en/v0.5.0/050-breaking-changes.html?highlight=breaking#interoperability-with-older-contracts} into the parser while configuring truffle to use older versions of solc doesn't appear to break anything).

FWIW, solc classifies the bad file imports truffle throws in _and_ any source code not conforming to Solidity language rules as a ParserError. It appears that actual wacky code takes precedence over bad file imports. For our purposes, I think all we care about for parser compilation is that the source code is linguistically legal before passing it on to the actual supplied compiler. We let that compiler handle all the SyntaxError, DeclarationError, TypeError, and Warning error types.

cgewecke: @CruzMolina Ok I could be misunderstanding something. I'd like to see if I can create failing test case for this later today. Thanks for investigating.

CruzMolina: Np! Thanks for helping reason about this. In case I wasn't being clear, the parser injection hack _forces_ solc to fail fast, before even getting to version-specific errors. Bad file imports appear to be lower in priority compared to actual grammatical issues. Would love to know if this doesn't actually hold true! 🤞

cgewecke: @CruzMolina I think #2011 is an example. Maybe.

Also had a thought today that if mixing compilers *is* a box of worms, one *could* use solidity-parser-antlr as the default but enable slow 'expedient hack' import parsing with the user's compiler via a flag or something. For the small % of cases where the JS parser lags the solc parser.  Probably have to explain this option pretty clearly to the user via the cli when they get syntax errors though.

CruzMolina: Thanks for the PR @cgewecke!

I _think_ the solution here is to only match failing imports and then filter out any non-import matches (instead of throwing to avoid any unintended compiler mixing effects).

To contextualize the broader situation a bit, we've been actively interested in removing `solc-js` from the bundle (Windows compatibility issues). Until we're able to do so, your issue makes a strong case for using the bundled `solc-js`. I think our preference at this point is to not introduce another dependency for parsing (would be nice if solc provided a parser for us).

gnidan:
Tricky situation! FYI I just proposed #2034 as a path forward. It likely relates to the efforts in this PR.

CruzMolina:
Closing this in favor of #2067.

PR #2011, opened 2019-05-13
Failing test for #2007

cgewecke:
**DO NOT MERGE -- EXAMPLE ONLY**

#2007 

@CruzMolina Worth sanity-checking this against develop to make sure it passes there - currently has merge conflicts from #2009. 

The [tests that fail](https://travis-ci.org/trufflesuite/truffle/jobs/531623403#L753-L758) make sure the profiler selects the correct compilation set. When that doesn't happen people have to constantly `rm -rf build` to get their code code changes to persist between edits. 

Watched the profiler working through this case and it's very confusing because there are repeated compilation events, the inheritance set is complicated, etc. But I *think* the problem is the one described in #2007 - an import isn't being picked up by the error-injection strategy because another error takes precedence. Not 100% confident though. 

Also apologies for harping on about bad import paths initially - you're absolutely right about those being resolved in the 'real' compilation step.

### DIFF
--- a/packages/truffle/test/sources/inheritance/contracts/Branch.sol
+++ b/packages/truffle/test/sources/inheritance/contracts/Branch.sol
@@ -1,4 +1,4 @@
-pragma solidity ^0.5.0;
+pragma solidity 0.4.24;
 
 import "./LeafA.sol";
 import "./LeafB.sol";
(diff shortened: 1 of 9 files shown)


CruzMolina:
Thanks for this @cgewecke ! As noted in #2007, it makes more sense to not throw within the parser and instead only return actual hacky bad import matches.

cgewecke:
Ha ha! Ok you win. Sorry if this was annoying.

CruzMolina:
No worries! Thanks for helping straighten things out.

PR #2020, opened 2019-05-14
Internal improvement: truffle config (docs & coverage)

CruzMolina:
This PR adds some docs, coverage, & tidies up misc things (es6 code, package.json links, etc).

### DIFF
--- a/packages/truffle-config/README.md
+++ b/packages/truffle-config/README.md
@@ -2,22 +2,28 @@
 Utility for interacting with truffle-config.js files
 
 ### Usage
-
  ```javascript
 const TruffleConfig = require("truffle-config");
+```
 
-// instantiate a default TruffleConfig object
+#### Instantiate default TruffleConfig object
+ ```javascript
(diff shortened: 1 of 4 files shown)


packages/truffle-config/README.md
```diff
@@ -1,2 +1,23 @@
 # truffle-config
 Utility for interacting with truffle-config.js files
+
+### Usage
+
+ ```javascript
+const TruffleConfig = require("truffle-config");
+
+// instantiate a default TruffleConfig object
```
gnidan: Instead of making these all part of one code block, let's make these h4's with separate code blocks!

gnidan: We'll want to shorten the wording to be more heading-appropriate (remove `a`/`the`, etc. and generally make each succinct)

gnidan: We could group some, too. Like `#### Config.detect()` and then have 1 code block with both examples.

Basically my goal here is that I want to be able to link to specific sections of the README easily.

PR #2018, opened 2019-05-14
Internal improvement: Add basic migration tests for migrate and deploy

eggplantzzz:
This PR adds two very basic tests for `truffle migrate` and `truffle deploy`. These just ensure that both `migrate` and `deploy` run on a fresh `truffle init` without throwing an error.

PR #2022, opened 2019-05-15
 fix bug for syntax

wxy:
In my macbook (node.js 11 ),  I have runing `truffle deploy --network ropsten` : 

```
TypeError: this.determineDryRunSettings is not a function
    at Contracts.compile.then (/Users/xingyuwang/.nvm/versions/node/v11.15.0/lib/node_modules/truffle/build/webpack:/packages/truffle-core/lib/commands/migrate.js:184:10)
```

fix it.

### DIFF
--- a/packages/truffle-core/lib/commands/migrate.js
+++ b/packages/truffle-core/lib/commands/migrate.js
@@ -91,7 +91,7 @@ const command = {
     ]
   },
 
-  determineDryRunSettings: (config, options) => {
+  determineDryRunSettings: function(config, options) {
     // Source: ethereum.stackexchange.com/questions/17051
     const networkWhitelist = [
       1, // Mainnet (ETH & ETC)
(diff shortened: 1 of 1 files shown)


packages/truffle-core/lib/commands/migrate.js
```diff
@@ -91,7 +91,7 @@ const command = {
     ]
   },
 
-  determineDryRunSettings: (config, options) => {
+  determineDryRunSettings: function(config, options) {
```
eggplantzzz: I think this change is unnecessary but I suppose it doesn't hurt anything :)

eggplantzzz:
Thanks for the help @wxy!

packages/truffle-core/lib/commands/migrate.js
wxy: 1. Keep the context syntax style consistent (same as +132)
2. In the following code( +184, +194), use `this.` call is not correct, you need to use `command.` to call, otherwise it will report the error and can't run.

eggplantzzz: The solution I had was to pass the bound method to the deploy command. In that case `this` has the correct context.
This solution looks good to me. I have some basic test for these two commands at PR #2018

PR #2023, opened 2019-05-16
Leverage lerna for prepare lifecycle (remove `./scripts/bootstrap.sh`)

gnidan:
Turns out that `lerna run`, by default, sorts packages topologically based on dependencies! ([`--no-sort`](https://github.com/lerna/lerna/tree/master/core/global-options#--no-sort) turns it off). Also, `lerna bootstrap` automatically calls `yarn prepare` in the monorepo root, if it's defined.

So it looks like we can clean this up pretty easily and just rely on those two assertions:

- Add a `yarn prepare` in the monorepo root (this gets called by `lerna bootstrap`)
- Remove `./scripts/bootstrap.sh` since we no longer need it
- Add `truffle-debugger` as a dependency of `truffle` (toposort seems broken without it)
- Remove a few now-extraneous scripts (`build:*`, `lerna:bootstrap`)

(This still maintains `yarn bootstrap` as the primary command to get everything all linked and built.)


gnidan:
yay deleting things!

PR #2024, opened 2019-05-16
ContractsAdd accidentally in schema.ts file twice, this commit fixes …

fainashalts:
…this.


gnidan:
This is failing because it needs commit 043317543a02e582a2a642b7fb50d8694adfe0a7.

We can still merge regardless, I think, because the other PR is getting close to done and that will fix the build.

Thanks for this!

Issue #2025, opened 2019-05-16
Debugger doesn't cope with imports not from files


gilesbradshaw:
- [x ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Debugger doesn't cope with imports not from files.

I hope this makes sense - I am new to all this!

So the app whose contracts I am trying to debug compiles its solidity code with the solc compiler - but it's generating the solidity from templates so they are never files.

for example this is how it is compiling solidity code:

````
import _truffleCompile from 'truffle-compile'

const truffleCompile = solidity =>
  new Promise(
    (resolve, reject) =>
      _truffleCompile(
        solidity,
        {
          contracts_directory: 'i do not care',
          compilers: {
            solc: {
              settings: {
                "optimizer": {
                  "enabled": false,
                  "runs": 200
                },
                "evmVersion": "byzantium",      
                outputSelection: {
                  '*': {
                    '*': ['*']
                  }
                }
              }
            }
          }
        },
        (err, result) => err ? reject(err) : resolve(result))
      )

const contracts = await truffleCompile({
    ...sources,
    [model.id]: model.solidity,
  })  
  
````

This means the imports in the solidity looks like this:

````
import "AbstractRegistry";

````

however I can't get your debugger to compile these contracts by just passing it a directory with the contracts in as the imports aren't from .sol files.  It gives me the error:

````
Starting Truffle Debugger...
⠹ Compiling your contracts...Error: Error: Could not find AbstractRegistry from any sources; imported from /home/giles/ipbm/ipbm/src/truffle/contracts/AbstractProcess.sol
    at async.whilst.error (/usr/lib/node_modules/truffle/build/webpack:/packages/truffle-compile/profiler.js:381:1)
    at /usr/lib/node_modules/truffle/build/webpack:/packages/truffle-compile/~/async/dist/async.js:969:1
    at next (/usr/lib/node_modules/truffle/build/webpack:/packages/truffle-compile/~/async/dist/async.js:5222:1)
Truffle v5.0.17 (core: 5.0.16)
Node v11.14.0

````

I can make them compile by changing the import to:

````
import "AbstractRegistry.sol";
````

however now I can't debug them as any change to the source code changes the byte code and they won't be recognised.

I 'fixed' this by hacking your code (truffle/packages/truffle-core/lib/commands/debug.js)

````
let compilePromise = new Promise(function(resolve, reject) {
          
[...]

PR #2030, opened 2019-05-17
Tear lib/commands/debug into pieces

gnidan:
That's it. I couldn't take it anymore. #2025 was the last straw.

- Switch to `const`s and `let`s instead of `var`s
- Move a function to debug-utils (`formatValue`)
- Split up lib/commands/debug into:
  - a thing that does the compilation (DebugCompiler)
  - a thing that houses all the copious print methods (DebugPrinter)
  - a thing that inelegantly wraps the giant repl interpreter function that does all that printing (DebugInterpreter)
  - a CLIDebugger class that does compilation and sets up the interpreter
- Make the code not gosh darn impossible to read.
- Expose how the heck this could actually be organized in the future.

### DIFF
--- a/packages/truffle-core/lib/debug/compiler.js
+++ b/packages/truffle-core/lib/debug/compiler.js
@@ -6,16 +6,18 @@ class DebugCompiler {
   }
 
   async compile() {
-    const { contracts, files } = await new Promise((accept, reject) => {
-      //we need to set up a config object for the compiler.
-      //it's the same as the existing config, but we turn on quiet.
-      //unfortunately, we don't have Babel here, so cloning is annoying.
-      let compileConfig = Object.assign(
-        {},
-        ...Object.entries(this.config).map(([key, value]) => ({ [key]: value }))
-      ); //clone
-      compileConfig.quiet = true;
+    //we need to set up a config object for the compiler.
+    //it's the same as the existing config, but we turn on quiet.
+    //unfortunately, we don't have Babel here, so cloning is annoying.
+    const compileConfig = Object.assign(
+      {},
+      ...Object.entries(this.config).map(([key, value]) => ({ [key]: value }))
+    ); //clone
+    compileConfig.quiet = true;
 
+    // since `compile.all()` returns two results, we can't use promisify
+    // and are instead stuck with using an explicit Promise constructor
+    const { contracts, files } = await new Promise((accept, reject) => {
       compile.all(compileConfig, (err, contracts, files) => {
         if (err) {
           return reject(err);
--- a/packages/truffle-core/lib/debug/compiler.js
+++ b/packages/truffle-core/lib/debug/compiler.js
@@ -1,0 +1,34 @@
+const compile = require("truffle-compile");
+
+class DebugCompiler {
+  constructor(config) {
+    this.config = config;
+  }
+
+  async compile() {
+    const { contracts, files } = await new Promise((accept, reject) => {
+      //we need to set up a config object for the compiler.
+      //it's the same as the existing config, but we turn on quiet.
+      //unfortunately, we don't have Babel here, so cloning is annoying.
+      let compileConfig = Object.assign(
+        {},
+        ...Object.entries(this.config).map(([key, value]) => ({ [key]: value }))
+      ); //clone
+      compileConfig.quiet = true;
+
+      compile.all(compileConfig, (err, contracts, files) => {
+        if (err) {
+          return reject(err);
+        }
+
+        return accept({ contracts, files });
+      });
+    });
+
+    return { contracts, files };
+  }
+}
+
+module.exports = {
+  DebugCompiler
+};
--- a/packages/truffle-core/lib/debug/index.js
+++ b/packages/truffle-core/lib/debug/index.js
@@ -1,0 +1,10 @@
+const debugModule = require("debug");
+const debug = debugModule("lib:debug");
+
+const { DebugPrinter } = require("./printer");
+const { DebugInterpreter } = require("./interpreter");
+
+module.exports = {
+  DebugPrinter,
+  DebugInterpreter
+};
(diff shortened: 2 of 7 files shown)


eggplantzzz:
What a guy!

gilesbradshaw:
I've done a pr into your branch fixing #2025 expecting it to show up here... but it's made a separate pr #2031 but anyhow it's there ..

packages/truffle-core/lib/debug/interpreter.js
```diff
@@ -0,0 +1,537 @@
+const debugModule = require("debug");
+const debug = debugModule("lib:debug:interpreter");
+
+const path = require("path");
+const util = require("util");
+const ora = require("ora");
+
+const DebugUtils = require("truffle-debug-utils");
+const selectors = require("truffle-debugger").selectors;
+const { session, solidity, trace, controller } = selectors;
```
gilesbradshaw: no need for session import

packages/truffle-core/lib/debug/interpreter.js
```diff
@@ -0,0 +1,537 @@
...
+
+      //first let's get the line number as usual
+      let line = parseInt(lineArg, 10); //want an integer
+      if (isNaN(line)) {
+        this.config.logger.log("Line number must be an integer.\n");
+        return;
+      }
+
+      //search sources for given string
+      let sources = session.view(solidity.info.sources);
```
gilesbradshaw: this.session.view

gnidan: good catch, thanks!

packages/truffle-core/lib/debug/interpreter.js
gnidan: You bet not!

packages/truffle-core/lib/debug/interpreter.js
```diff
@@ -0,0 +1,537 @@
...
+    command: "debug: watch expression",
+    args: { isVariable }
+  });
+}
+
+class DebugInterpreter {
+  constructor(config, session, txHash) {
+    this.config = config;
+    this.session = session;
+    this.printer = new DebugPrinter(config, session);
```
haltman-at: Just a general comment about the whole `DebugPrinter` idea -- it's kind of confusing that, even though you have this debug printer, much of the printing doesn't go through it, and is just raw `config.logger.log`.  I mean, I dunno that it makes sense to write a whole new method for every message we want to print, but maybe at least the `DebugPrinter` could get a simple `print` method that all these uses of `config.logger.log` could be replaced with?  So that at least all printing would go through the printer?

packages/truffle-core/lib/debug/compiler.js
```diff
@@ -0,0 +1,34 @@
+const compile = require("truffle-compile");
+
+class DebugCompiler {
+  constructor(config) {
+    this.config = config;
+  }
+
+  async compile() {
+    const { contracts, files } = await new Promise((accept, reject) => {
```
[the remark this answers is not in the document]
gnidan: It is necessary to have the explicit Promise because `compile` passes _two_ success parameters to the callback, and in my observation, this fails to work through `promisify`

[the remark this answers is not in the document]
gnidan: I'll resolve this once I've added a comment explaining as such

packages/truffle-core/lib/debug/index.js
```diff
@@ -0,0 +1,103 @@
...
+    };
+
+    const bugger =
+      txHash !== undefined
+        ? await Debugger.forTx(txHash, debuggerOptions)
+        : await Debugger.forProject(debuggerOptions);
+
+    const session = bugger.connect();
+
+    // check
```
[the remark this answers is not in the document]
gnidan: oh it should probably say
```suggestion
    // check for error
```

in order to document the below `if`, for the lazy lookers.

packages/truffle-core/lib/debug/interpreter.js
gnidan: I recognized the lack of cohesion as I was putting `DebugPrinter` together, but I decided that it could be a separate effort to clean up the input/output. My thinking is that this PR is mostly a gentle structural nudge, not a full clean-up. Moving around actual `config.logger` statements seems like it's bound to result in slip-ups that are hard to account for.

We'll want to do that eventually, but it's just a question of whether it belongs in this PR. My feeling is that it's not urgent enough to block this PR, but let me know if you disagree. Your point may be sound: it might not be worth making this class without it being more cohesive and decoupled. What do you think?

packages/truffle-core/lib/debug/compiler.js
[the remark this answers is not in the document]
haltman-at: Oh huh.  Except that shouldn't apply in this case, right?  Because it only passes one (in that it wraps the two in an object).  Is there something I'm missing here?

packages/truffle-core/lib/debug/interpreter.js
haltman-at: I mean it shouldn't be hard to do a search/replace to turn `config.logger.log` into `this.printer.print` should it?  Obviously that's kind of a stopgap, but I think it's a lot better than raw calls to `config.logger.log`.

gnidan: Oh I see. Yeah I'll do that!

gnidan: OK I got rid of the `this.config.logger.log` statements in the interpreter, and even went a bit further there to get rid of `this.config`, since it was barely used.

gnidan:
(I don't know how codeclimate got turned on here but I am not planning on respecting that failure for this PR)

packages/truffle-core/lib/debug/compiler.js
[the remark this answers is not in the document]
gnidan: Cleaned this up a bit. Thanks for identifying the code smell!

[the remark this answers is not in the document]
haltman-at: (It turned out I was missing something, yes.  But @gnidan and I decided to at least take as much outside of the manual promise as possible.)

gnidan:
Woot!

Issue #1718, opened 2019-02-15
Build a complete, flexible system for the querying and mutation of Truffle's smart contract records


gnidan:
## Issue

The artifacts format isn't cutting it. It's incomplete and doesn't support all use cases [issue citation needed]. Also, the files are too big to be natively useful to users in the front-end [issue citation needed].

## Solution

Build something better.

clouds56:
Shall we have qualified name for each contract
```shell
contracts/
  a/
   a1.sol  # contract A, C
  a.sol    # contract A
  a-a1.sol # contract A (note `-` in path would be replace with `_`)
  b.sol    # contract B
```
would build to artifacts (not `-` is not valid identifier of contract name)
```shell
build/
  a-a1-A.json # from a/a1.sol
  a-a1-C.json # from a/a1.sol
  a_a1-A.json # from a.a1.sol
  a-A.json
  b-B.json
```
when we query a artifacts, we could use both `simple name` and `qualified name`
```ts
artifacts.require("C") // would load with simple name `a-a1-C.json`
artifacts.require("a::a1::A") // would load with qualified name `a-a1-A.json`
artifacts.require("A") // would raise a error for name conflict,
                       // saying all `a::a1::A, a_a1::A, a::A` would provide `A`
```

PR #2031, opened 2019-05-17
allow compile of an object

gilesbradshaw:
as per #2025 - this allows the .sol files to specify properties of an object hence allowing imports without the .sol extension.

a config property "debugObjects" if truthy controls if this happens.

### DIFF
--- a/packages/truffle-core/lib/debug/object-compiler.js
+++ b/packages/truffle-core/lib/debug/object-compiler.js
@@ -13,10 +13,7 @@ class DebugObjectCompiler {
   async compile() {
     // all the file names
     const fileNames = await new Promise((resolve, reject) =>
-      fs.readdir(`${this.config.working_directory}/contracts`, function(
-        err,
-        filePaths
-      ) {
+      fs.readdir(this.config.contracts_directory, function(err, filePaths) {
         if (err) {
           return reject(err);
         }
@@ -29,7 +26,7 @@ class DebugObjectCompiler {
       fileNames.map(file =>
         new Promise((resolve, reject) =>
           fs.readFile(
-            `${this.config.working_directory}/contracts/${file}`,
+            `${this.config.contracts_directory}/${file}`,
             (err, text) => {
               if (err) {
                 return reject(err);
--- a/packages/truffle-core/lib/debug/object-compiler.js
+++ b/packages/truffle-core/lib/debug/object-compiler.js
@@ -1,4 +1,6 @@
+const { promisify } = require("util");
 const compile = require("truffle-compile");
+const find_contracts = require("truffle-contract-sources");
 const fs = require("fs");
 
 // this class copes with a directory of .sol files each if which
@@ -12,37 +14,25 @@ class DebugObjectCompiler {
   }
   async compile() {
     // all the file names
-    const fileNames = await new Promise((resolve, reject) =>
-      fs.readdir(this.config.contracts_directory, function(err, filePaths) {
-        if (err) {
-          return reject(err);
-        }
-        return resolve(filePaths);
-      })
+    const filePaths = await promisify(find_contracts)(
+      this.config.contracts_directory
     );
     // this is an object containing solidity code as properties
     // hence imports can be without the .sol extension
     const objectToCompile = await Promise.all(
-      fileNames.map(file =>
-        new Promise((resolve, reject) =>
-          fs.readFile(
-            `${this.config.contracts_directory}/${file}`,
-            (err, text) => {
-              if (err) {
-                return reject(err);
-              }
-              resolve(text);
-            }
-          )
-        ).then(text => ({
-          file,
+      filePaths.map(filePath =>
+        promisify(fs.readFile)(filePath).then(text => ({
+          filePath,
           text
         }))
       )
     ).then(files =>
-      files.reduce((acc, { file, text }) => {
+      files.reduce((acc, { filePath, text }) => {
         // nasty mutate as no object spread operator
-        acc[file.substring(0, file.indexOf(".sol"))] = text.toString();
+        // regexes just get filename from full path and strip extension
+        acc[
+          filePath.replace(/^.*[\\/]/, "").replace(/\.[^/.]+$/, "")
+        ] = text.toString();
         return acc;
       }, {})
     );
--- a/packages/truffle-core/lib/debug/object-compiler.js
+++ b/packages/truffle-core/lib/debug/object-compiler.js
@@ -1,0 +1,76 @@
+const compile = require("truffle-compile");
+const fs = require("fs");
+
+// this class copes with a directory of .sol files each if which
+// becomes a property of an object which is then compiled
+// the imports in these files will be in the form
+// import "file"
+// rather than import "file.sol"
+class DebugObjectCompiler {
+  constructor(config) {
+    this.config = config;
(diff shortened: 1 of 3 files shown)


gilesbradshaw:
however...

````

debug(development:0xdec67418...)> b Goods_Shipement_Contract:133
Thrown:
TypeError: session.view is not a function
    at TTY.onStreamRead (internal/stream_base_commons.js:150:17)
    at ReadStream.Readable.push (_stream_readable.js:231:10)
    at readableAddChunk (_stream_readable.js:276:11)
    at addChunk (_stream_readable.js:295:12)
    at ReadStream.EventEmitter.emit (domain.js:481:20)
    at ReadStream.emit (events.js:193:13)
    at ReadStream.onData (readline.js:1097:36)
    at emitKeys.next (<anonymous>)
    at emitKeys (internal/readline.js:424:14)
    at ReadStream.EventEmitter.emit (domain.js:481:20)
    at ReadStream.emit (events.js:193:13)
    at ReadStream.onkeypress (readline.js:183:10)
    at REPLServer.self._ttyWrite (repl.js:727:7)
    at REPLServer.Interface._ttyWrite (readline.js:960:14)
    at REPLServer.Interface._line (readline.js:680:8)
    at REPLServer.Interface._onLine (readline.js:307:10)
    at REPLServer.EventEmitter.emit (domain.js:481:20)
    at REPLServer.emit (events.js:193:13)
    at REPLServer.onLine (repl.js:650:10)
    at REPLServer.runBound [as eval] (domain.js:438:12)
    at bound (domain.js:425:14)
    at ReplManager.interpret (/home/giles/truffle-2/packages/truffle-core/lib/repl.js:118:18)
    at Object.bound interpreter [as interpreter] (util.js:215:13)
    at DebugInterpreter.interpreter (/home/giles/truffle-2/packages/truffle-core/lib/debug/interpreter.js:457:20)
    at DebugInterpreter.setOrClearBreakpoint (/home/giles/truffle-2/packages/truffle-core/lib/debug/interpreter.js:119:29)

````

(not sure if this is an issue with my mod or the await-debug branch)

**Comment by gilesbradshaw** [comment]
I've fixed the issue with session.view..

File: packages/truffle-core/lib/debug/object-compiler.js [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+  async compile() {
+    // all the file names
+    const fileNames = await new Promise((resolve, reject) =>
+      fs.readdir(this.config.contracts_directory, function(err, filePaths) {
+        if (err) {
+          return reject(err);
+        }
+        return resolve(filePaths);
+      })
+    );
```
gnidan: I think you can replace this with [truffle-contract-sources](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-contract-sources/index.js). You can see how it's used [here](https://github.com/trufflesuite/truffle/blob/a1f1340c525d8a1e667386ffd8c61d948ad5d315/packages/truffle-compile/index.js#L355) for instance.

gilesbradshaw: I also changed it to use promisify that regex is a bit ugly but it works + for both / and \

gilesbradshaw:
@gnidan could you point me at the docs for this?  I'll change that too if u like..

gnidan:
Which docs @gilesbradshaw? This one: https://github.com/trufflesuite/trufflesuite.com?

BTW I think my goal for this PR is to get rid of the separate Object compiler and make it an option for the regular compiler. But I'm in meetings for the rest of the day and will likely have to circle back on Monday. Immediately, my goal is to see about how we can continue to trim your implementation and then plug it into the normal debugger startup workflow, so that way it doesn't represent a completely separate branch of code.

gilesbradshaw:
ok I'll have a ponder about that too

gnidan:
Hey @gilesbradshaw sorry for the delay. Still haven't gotten into this yet, I should be able to get my attention here again by the end of the week.

gnidan:
Hey @gilesbradshaw, I've been trying to reason about this PR and the underlying necessity, and I think that this should not be debug-specific. Assuming I have this right, you want to make `truffle debug` perform compilation differently than `truffle compile`.

If that's the premise, I'd say this is out of scope (but with promising caveats, bear with me!). An unmet requirement for the `truffle debug` is that it should follow `truffle compile` in the same workflow... i.e., `truffle debug` should just reuse a prior `truffle compile` result without the need to compile everything at start every time.

Effectively, this means that all of `truffle debug`'s own compilation phase is a hack! We're seeking to remove this compilation step from the debugger entirely, but doing so is blocked by #1718. In the meantime, I'd hate to add to the footprint of this hack... heck, I wouldn't want you to start relying on this behavior, only for it to get deleted!

Instead, I think we can solve the underlying limitation you're facing with #2025 in one or both of two directions:

- We make `truffle compile` more configurable with respect to import paths. This feels plainly unobjectionable: Truffle really ought to let you customize your import paths! But the question is: what does that configuration look like? I'm not sure, personally; would need to do further consideration. 

- Since you raised this concern originally (and prompted my rewriting lib/commands/debug), we now have a `CLIDebugger` class that can serve as a decent external component for other packages! In fact, I ended up recreating one aspect of this PR in #2050 (see [this file change](https://github.com/trufflesuite/truffle/pull/2050/commits/acf1a39868a9b3a7adf7bf1b4e29c62ccc193efa#diff-a110551c33c5e7b37090895c4755f0ae), adding a `compilation` constructor argument).

  Following this direction, I'd advocate that CLIDebugger be made into its own package (for separate semantic versioning), and that DebugObjectCompiler is out of scope to this repository, and it could be something you maintain just for your own workflow or as an NPM package.

I am curious your thoughts in response. Does the former sound like something you'd want? Are you even looking for a `truffle compile`-free workflow? Am I making sense / do I understand the premise? 

Please let me know what you think. Thank you!

packages/truffle-core/lib/debug/object-compiler.js
```diff
@@ -0,0 +1,63 @@
...
+    const { contracts, files } = await new Promise((accept, reject) => {
+      //we need to set up a config object for the compiler.
+      //it's the same as the existing config, but we turn on quiet.
+      //unfortunately, we don't have Babel here, so cloning is annoying.
+      let compileConfig = Object.assign(
+        {},
+        ...Object.entries(this.config).map(([key, value]) => ({ [key]: value }))
+      ); //clone
+      compileConfig.quiet = true;
```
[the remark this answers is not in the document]
haltman-at: Oops, that's preexisting and is actually due to me.

gilesbradshaw:
Hi @gnidan that would seem to make sense.  The contracts I was trying to debug weren't originally produced with truffle at all - although I did modify the system producing them to use truffle in the hope of making them debugable.  So FWIU you are saying that one could 'plug' precompiled contracts and source code into the debugger so it doesn't do the compile step at all.

I'm not sure how far I will get with all this as it's a bit of a 'pet' project and i get busy with work that actually pays bills..

As an aside it would also be good for the debugger to show a call stack and also go back and out

eggplantzzz:
Closing for maintenance. Feel free to bring this back up and we can re-start the conversation to get this through. Thanks!

gilesbradshaw:
hey that's fine sorry about delay in answering - not working on this right now..

PR #2032, opened 2019-05-17
Fix ethpm bug

eggplantzzz:
When implementing the Web3Shim (I think), a Web3 require statement was removed in `packages/truffle-core/lib/package.js`. This caused `truffle install` to fail. This PR adds the require statement back in. In addition it provides some pretty output at the end of the job stating that packages were installed.

The output for this command now will look like the following:
Screen Shot 2019-05-17 at 6 31 56 PM

It also adds a basic test for `truffle install`, ensuring that `truffle install zeppelin` completes successfully and a basic `truffle init` project has the "installed_contracts" directory that `truffle install xxx` creates.

### DIFF
--- a/packages/truffle-core/lib/commands/install.js
+++ b/packages/truffle-core/lib/commands/install.js
@@ -1,29 +1,30 @@
-var command = {
-  command: 'install',
-  description: 'Install a package from the Ethereum Package Registry',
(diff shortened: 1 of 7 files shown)


fainashalts:
Nice!

PR #2033, opened 2019-05-17
Get revert reason from Geth clients

cgewecke:
Continuation of #1652 (cherry picked its first commit). That PR is marked Quorum/Enterprise but this problem affects truffle-contract use with geth generally.  

Geth [requires that eth_call use a block param](https://github.com/ethereum/go-ethereum/issues/2472) and its absence means that `truffle-contract` isn't able to correctly replay  failed transactions and get a reason returned as data. 

Believe 'latest' is the right value here - should work with both instamining dev clients and regular networks. Truffle-contract fetches the reason in a call immediately following a failed tx - 'latest' will be typically be the relevant block.

Have added a test in the [ @ geth ] methods suite.

Additional notes:
+  Ganache returns a gas estimate for transactions that fail, geth doesn't. If you're using `truffle-contract` *outside the truffle cli context* (e.g. a browser), targeting geth, and relying on automatic gas estimation, you won't be able to fetch the reason string because your tx will always OOG.

+ Still doesn't seem possible to get a reason string from a failing `require` in a contract constructor with Geth. This needs more investigation but it might be an issue in web3 1.0.0-beta.37.

### DIFF
--- a/packages/truffle-contract/lib/reason.js
+++ b/packages/truffle-contract/lib/reason.js
@@ -37,9 +37,9 @@ const reason = {
    */
   get: function(params, web3){
     const packet = {
-      jsonrpc: "2.0",
-      method: "eth_call",
-      params: [params],
+      jsonrpc: '2.0',
+      method: 'eth_call',
+      params: [params, 'latest'],
       id: new Date().getTime(),
     };
 
--- a/packages/truffle-contract/test/methods.js
+++ b/packages/truffle-contract/test/methods.js
@@ -453,6 +453,29 @@ describe("Methods", function() {
       }
     });
 
+    it("errors with a revert reason", async function() {
+      const example = await Example.new(1);
+      try {
+        // At the moment, this test can't rely on truffle-contract's
+        // gas estimation when running with geth.
+        await example.triggerRequireWithReasonError({gas: 1000000});
+        assert.fail();
+      } catch (e) {
+        assert(
+          e.reason === "reasonstring",
+          "Triggered require result should have revert reason field"
+        );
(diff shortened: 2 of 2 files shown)


cgewecke:
NB It looks like there are tests failing in the geth job that are not being caught by mocha as errors? Affects the scenario tests for the simple deploy, simple migrate commands. That problem isn't specific to this PR, can be seen on unrelated PRs (e.g [2031](https://travis-ci.org/trufflesuite/truffle/jobs/533935787))

cgewecke:
Sweet, thanks @gnidan!

tcoulter:
👋 Hi @cgewecke!

cgewecke:
Hi @tcoulter :)

stevenlcf:
Could someone give an example of how to fetch the revert reason from geth with this new feature? It seems that, under this PR, only a file related to ganache revert reason and a test file were modified. I am confused how they could help get the revert reason from geth?

cgewecke:
@stevenlcf This hasn't been published yet.

Ganache and geth return the revert reason the same way. If a failing transaction is executed as a `call`, its return data encodes the reason string. Truffle-contract automatically replays failing txs to extract this info.

The reason will be available on the thrown error object as a `reason` property and will also appear in the error message. Ex:
```javascript
try {
  await myContract.myFailingMethod();
} catch (err) {
  console.log(err); // Displays the reason in addition other error msgs from the client
  console.log(err.reason) // Displays just the reason 
}
```

stevenlcf:
Thanks, @cgewecke !

PR #2036, opened 2019-05-20
Rework loader functionality and add contract

fainashalts:
This PR seeks to rework the loader functionality by:
 
1) Running a manual compilation via `truffle-workflow-compile`'s `compile()` function
2) Adding `bytecodes`, `sources`, and `contracts` based on the results from the manual compilation

This PR also reworks the manual compilation in `loaders/artifacts/index.ts` to return the compilation results without a callback. This change was made to avoid using a callback in the Jest's module mock of `truffle-workflow-compile`. Previously, when a callback was used, functions run inside the callback did not get run in the tests. This fix addresses this issue to ensure that the tests simulate as closely as possible the actual behavior of this code.


packages/truffle-db/src/loaders/artifacts/test/index.ts
```diff
@@ -9,26 +9,42 @@ import * as Contracts from "truffle-workflow-compile";
...
+        "solc": [
+          "/Users/fainashalts/solidity-magic-square/contracts/MagicSquare.sol",
+          "/Users/fainashalts/solidity-magic-square/contracts/Migrations.sol",
+          "/Users/fainashalts/solidity-magic-square/contracts/SquareLib.sol"
+        ],
+        "vyper": [
+           "/Users/fainashalts/truffle-six/testing2/contracts/VyperStorage.vy",
+        ]
+      },
+      "contracts": [{
```
gnidan: ~Something's off here. As it currently stands, this part of the return value should be an object mapping of contract name -> contract object, not an array. I'm betting the code that's using this mock is wrong as a result... I'll confirm in another pass.~

On another note about these mocks: I know `contract_name` is prevalent, but is that really how we're getting it? The correct field here is `contractName`, although it's possible that it's not coming out of t-w-c like that. I think we should do a `require("truffle-contract-schema").normalize(contractObject)` for all of these.

gnidan: Oh, well, hey! Nevermind! The only place the code looks at the result is [here](https://github.com/trufflesuite/truffle/pull/2036/files#diff-6d89eb0a6dc68a0c021900eb15ee4345R281), and it uses `Object.values(contracts)`, which works for both arrays and objects just fine.

So since this is going to change to using an array anyway, I think it's fine to keep the mock as an array. Don't mind me!

packages/truffle-db/src/loaders/artifacts/test/index.ts
```diff
@@ -39,14 +55,18 @@ const config = {
...
-    all: true
+  contracts_directory: path.join(__dirname, "compilationSources"),
+  contracts_build_directory: path.join(__dirname, "sources"),
+  all: true
 }
 
 const db = new TruffleDB(config);
 const Migrations = require(path.join(fixturesDirectory, "Migrations.json"));
-const Build = [require(path.join(__dirname, "sources", "MagicSquare.json")), require(path.join(__dirname, "sources", "Migrations.json")), require(path.join(__dirname, "sources", "SquareLib.json")) ];
+const Build = [
```
gnidan: Naming seems a bit off here: for one, it should probably be lowercase because it's a specific thing, not a type of thing or module of some kind.

Can we just call it `artifacts`? I don't think this file uses that name for something else, and that's what this list is.

packages/truffle-db/src/loaders/artifacts/test/index.ts
```diff
@@ -149,46 +165,150 @@ query getWorkspaceCompilation($id: ID!) {
...
+      contractIds.push({ id: expectedId });
+      let {
+        data: {
+          workspace: {
+            contract: {
+              id,
+              name,
+              abi: {
+                json
+              },
```
gnidan: This isn't checked for, so let's delete it and save three lines!

packages/truffle-db/src/loaders/artifacts/test/index.ts
```diff
@@ -149,46 +165,150 @@ query getWorkspaceCompilation($id: ID!) {
...
+              constructor: {
+                createBytecode: {
+                  bytes
+                }
+              },
+              sourceContract: {
+                source: {
+                  contents,
+                  sourcePath
```
gnidan: `sourcePath` isn't being tested for either; might as well remove the unused variable.

packages/truffle-db/src/loaders/artifacts/test/index.ts
```diff
@@ -149,46 +165,150 @@ query getWorkspaceCompilation($id: ID!) {
...
+                }
+              },
+              compilation: {
+                compiler: {
+                  version
+                },
+                sources,
+                contracts
```
gnidan: `sources` and `contracts` seem like they're unused

packages/truffle-db/src/loaders/artifacts/test/index.ts
```diff
@@ -149,46 +165,150 @@ query getWorkspaceCompilation($id: ID!) {
...
+        }
+      } = await db.query(GetWorkspaceContract, contractIds[index]);
+
+      expect(name).toEqual(Build[index].contractName);
+      expect(bytes).toEqual(Build[index].bytecode);
+      expect(contents).toEqual(Build[index].source);
+      expect(version).toEqual(Build[index].compiler.version);
+      expect(id).toEqual(contractIds[index].id);
+    }
+  });
```
gnidan: All-in-all, diggin' these tests! Reading them provides good assurance that the behavior is thoroughly tested, which is great! Nice job!

packages/truffle-db/src/loaders/artifacts/index.ts
```diff
@@ -152,122 +197,125 @@ export class ArtifactsLoader {
...
+      let contractObject = {
+        name: contract["contract_name"],
+        abi: {
+          json: JSON.stringify(contract["abi"])
+        },
+        compilation: {
+          id: compilationId
+        },
+        sourceContract: {
+          index: +index
```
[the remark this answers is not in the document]
fainashalts: It was to ensure that index was a number but it's not actually necessary. Will remove.

packages/truffle-db/src/loaders/artifacts/test/index.ts
fainashalts: Agree! Changing.

PR #2042, opened 2019-05-21
Enhancement: Remove solc-js compiler from the bundle

### LINKED ISSUE #2034: Remove solc NPM dependency and use Truffle's `obtain` mechanism instead
## Overview

In a nutshell, this issue proposes to:
- remove NPM solc from Truffle
- use soljson and the compiler supplier (`truffle obtain` mechanism) by default instead
- auto-run `truffle obtain` on Truffle install

### Background

An important requirement for Truffle is to provide users with solc automatically, to make it easy to get started writing smart contracts. To achieve this, Truffle has always depended on the `solc` NPM package, either baked into the bundle or as a separate dependency.

Both of these approaches have issues:
- The former creates a bundle too large for Node.js memory bounds.
- The latter has been fraught with compatibility problems on different platforms. This is at odds with the goal of the `truffle` package to be as portable as possible.

Truffle v5 introduces a mechanism to autodownload solc via the `truffle obtain --solc=<version>` command. This fetches/caches soljson, which is a portable emscripten binary, and wraps it with the NPM package `solc`'s provided `wrapper` module.

This issue proposes to meet the stated requirement of providing users with solc automatically by using this `truffle obtain` during the `postinstall` hook when installing Truffle. Since installation can be expected to require an internet connection, it seems safe to perform this download upfront.

## What this entails

- Set up postinstall hook to invoke `truffle obtain --solc=<specific-version>` in user and development contexts
- Make `compilers.solc` default to this auto-obtained soljson, instead of the current NPM solc
- Add user-configuration option for disabling auto-download
- Get `solc/wrapper` into its own package and add this as dependency of truffle-compile
- Remove `solc` dependency from all packages
- Remove `Bundled` strategy for the compiler supplier

There may be more here that I am missing. We should make sure to consider unknown UX ramifications or other reasons why this might result in a negative experience.

## Why?

- soljson is portable.
- The auto fetch/cache functionality (`truffle obtain --solc`) is a key feature and will receive proper maintenance for the foreseeable future.
- Truffle currently runs solc in a number of different ways:
  - NPM solc
  - soljson
  - native shell command
  - docker

Effectively, the NPM solc integration has been brittle and required a lot of dedicated effort to maintain over the years. That effort can be eliminated moving forward and spent on other, less annoying things ✨

CruzMolina:
This PR seeks to resolve #2034 by removing the `solc-js` compiler component completely from the bundle.

This is handled by:

- leveraging `truffle obtain` in a postinstall to download and cache a default soljson when installing truffle via npm.
- Importing only the `solc/wrapper` to parse soljson's.
- Removing the `Bundled` strategy from the CompilerSupplier.

The current implementation of this can be tested at `truffle@unbundled-solc`.

### DIFF
--- a/packages/truffle-compile/package.json
+++ b/packages/truffle-compile/package.json
@@ -13,6 +13,7 @@
     "request-promise": "^4.2.2",
     "require-from-string": "^2.0.2",
     "semver": "^5.6.0",
+    "solc": "^0.5.0",
     "truffle-config": "^1.1.12",
     "truffle-contract-sources": "^0.1.3",
     "truffle-error": "^0.0.4",
--- a/packages/truffle-compile/package.json
+++ b/packages/truffle-compile/package.json
@@ -17,8 +17,7 @@
     "truffle-config": "^1.1.12",
     "truffle-contract-sources": "^0.1.3",
     "truffle-error": "^0.0.4",
-    "truffle-expect": "^0.0.8",
-    "truffle-solc-wrapper": "https://github.com/trufflesuite/truffle-solc-wrapper"
+    "truffle-expect": "^0.0.8"
   },
   "devDependencies": {
     "babel-core": "^6.26.0",
--- a/packages/truffle-compile/package.json
+++ b/packages/truffle-compile/package.json
@@ -13,11 +13,11 @@
     "request-promise": "^4.2.2",
     "require-from-string": "^2.0.2",
     "semver": "^5.6.0",
-    "solc": "0.5.0",
     "truffle-config": "^1.1.12",
     "truffle-contract-sources": "^0.1.3",
     "truffle-error": "^0.0.4",
-    "truffle-expect": "^0.0.8"
+    "truffle-expect": "^0.0.8",
+    "truffle-solc-wrapper": "https://github.com/trufflesuite/truffle-solc-wrapper"
   },
   "devDependencies": {
     "babel-core": "^6.26.0",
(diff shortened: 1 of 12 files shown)


packages/truffle-compile/package.json
```diff
@@ -13,11 +13,11 @@
     "request-promise": "^4.2.2",
     "require-from-string": "^2.0.2",
     "semver": "^5.6.0",
-    "solc": "0.5.0",
     "truffle-config": "^1.1.12",
     "truffle-contract-sources": "^0.1.3",
     "truffle-error": "^0.0.4",
-    "truffle-expect": "^0.0.8"
+    "truffle-expect": "^0.0.8",
+    "truffle-solc-wrapper": "https://github.com/trufflesuite/truffle-solc-wrapper"
```
gnidan: Let's open an issue at solc-js to start the conversation about them maintaining this.

packages/truffle/scripts/postinstall.js
```diff
@@ -0,0 +1,20 @@
...
+
+const bundledCLI = "./build/cli.bundled.js";
+
+const postinstallObtain = () => {
+  try {
+    statSync(bundledCLI);
+    execSync(`node ${bundledCLI} obtain --solc=0.5.0`);
+  } catch ({ message }) {
+    if (message.includes("no such file")) return;
+    throw new Error(message);
```
gnidan: We should make sure that this error case looks nice on both `npm` and `yarn` (like what if the obtain fails?)

packages/truffle-compile/package.json
gnidan: Ah I am out of touch. There is already https://github.com/ethereum/solc-js/issues/360

PR #2043, opened 2019-05-21
Internal improvement: Update truffle-core and remove some unneeded code

eggplantzzz:
This PR removes a handful of legacy code and seeks to update some of the syntax in truffle-core.


packages/truffle-core/cli.js
```diff
@@ -33,15 +33,7 @@ const Command = require("./lib/command");
 
 const command = new Command(require("./lib/commands"));
 
-// Hack to suppress web3 MaxListenersExceededWarning
-// This should be removed when issue is resolved upstream:
-// https://github.com/ethereum/web3.js/issues/1648
-const listeners = process.listeners("warning");
-listeners.forEach(listener => process.removeListener("warning", listener));
```
gnidan: Are you sure we can remove this?

packages/truffle-core/lib/command.js
```diff
@@ -1,138 +1,122 @@
...
-  delete argv["$0"];
+    const newOptions = Object.assign({}, options, argv);
 
-  // Some options might throw if options is a Config object. If so, let's ignore those options.
-  var clone = {};
-  Object.keys(options).forEach(function(key) {
     try {
-      clone[key] = options[key];
-    } catch (e) {
-      // Do nothing with values that throw.
```
gnidan: Ah so this `try`/`catch` is important because of the way truffle-config works, since it throws exceptions when you try to assign to things that aren't kosher. Might not be safe to remove this.

eggplantzzz:
Yeah some of these things I wanted to remove and see if it worked afterwards. I haven't encountered any problems yet in the few tests I've done :)

packages/truffle-core/cli.js
CruzMolina: Where is the output for this supposed to occur?

CruzMolina: Maybe I'm reading Coveralls incorrectly, but it looks like on `develop` these suppression lines are still getting hit.

packages/truffle-core/lib/command.js
CruzMolina: What's confusing is that it seems like `truffle-config` does catch and throw exceptions for non-kosher assignments, but here the code catches and then does nothing? Does the error bubble up in the `catch` on line 119?

If that's the case, then I think it does make sense to keep the `try/catch`, since I believe users will see a confusing `TypeError` thrown instead.

eggplantzzz: @CruzMolina It is attempting to "sanitize" the values. In other words, it will remove all the values that `throw`. I'm not certain that it is necessary though...

packages/truffle-core/cli.js
eggplantzzz: I removed it and I haven't seen any adverse effects yet. I guess we should test everything pretty well to see if it is just leftover junk or not :)

packages/truffle-core/lib/command.js
CruzMolina: Ah, I see. `truffle-config` has it's own normalizing method (https://github.com/trufflesuite/truffle/blob/truffle-core-housekeeping/packages/truffle-config/index.js#L297-L307).

I think the comment `Some options might throw...` suggests if options is a `TruffleConfig` object containing getters, the key assignment might throw. I'm not sure if `Object.assign` will appropriately skip a getter and _not_ throw a `TypeError`.

eggplantzzz: @CruzMolina I don't think that `options` will ever be an instance of a config object. Can you think of a case where it would be?

CruzMolina: Looks like `options` includes an instance or two of a `Config` object while inside `truffle develop` and `truffle console`.

The original comment isn't very specific (would be nice to know _which_ options might throw), but I imagine it would be safer to keep a `try/catch` here to make sure we don't inadvertently break something obscure.

packages/truffle-core/cli.js
CruzMolina: Following the trail of issues/PRs...

https://github.com/ethereum/web3.js/issues/1648
https://github.com/trufflesuite/ganache-core/issues/267
https://github.com/trufflesuite/ganache-core/pull/269

It looks like this hasn't been resolved yet in the version of web3 we're using.
Also looks like the `ganache-core` PR would help when using Ganache as a provider, but there's still web3-specific event emitting issues.

packages/truffle-core/lib/command.js
eggplantzzz: Yeah but keep in mind that this is in `command.js` and before any of the code in the "commands" files themselves. `options` at this point will always just be the logger I think. What gets attached later should have nothing to do with this code.

eggplantzzz: Ahh yes I see where a command is instantiated with a config object, you are right. I wonder if it really causes any problems...

PR #2047, opened 2019-05-23
Internal improvement: update licensing

CruzMolina:
This PR updates the existing formal license documentation, adds formal license documentation to the remaining codebase, dates the licenses, and reformats the text to be more human-readable.

📝 

Pending final thoughts regarding copyright ownership.


gnidan:
Note: I think we want the filenames to be `LICENSE` for conventionality. What was the reasoning for the `MIT-LICENSE` filename?

CruzMolina:
It appeared to be the case that the [API Github uses](https://developer.github.com/v3/licenses/) to determine which [pretty license page](https://github.com/rails/rails/blob/master/MIT-LICENSE) to provide searched for a license file prefix. Looks like this is actually not the case and instead it tries to match the contents of the file.

PR #1894, opened 2019-04-09
Update truffle-artifactor (docs & codebase)

CruzMolina:
This PR updates the documentation to `truffle-artifactor`'s current behavior & API. Also burns some code bringing the package up to ES6, converts the API into async methods, & intends to make the code easier to reason while debugging.

Resolves #1096.

PR #2049, opened 2019-05-24
Fix: appropriately assign & merge artifact properties

### LINKED ISSUE #2048: Incorrect ABI generated when compiling after modifying a contract
## Issue

Since truffle 5.0.13, when compiling a specific contract after adding a variable to it, the resulting ABI is incorrect.

## Steps to Reproduce

We have narrowed down the issue to these specific two contracts. Let's call them V1:

```
pragma solidity ^0.5.0;

contract MyContract {
  uint256 public x;
  string public s;

  function initialize(uint256 _x, string memory _s) public {
    x = _x;
    s = _s;
  }
}
```

And V2:
```
pragma solidity ^0.5.0;

contract MyContract {
  uint256 public x;
  string public s;
  uint256 public t;

  function initialize(uint256 _x, string memory _s, uint256 _t) public {
    x = _x;
    s = _s;
    t = _t;
  }
}
```

Note that the difference between the two is the added `t` public field, and adding it as a parameter in the initialize function.

When compiling V2 directly, the resulting ABI is correct, and the definition for the `t` public function is:
```
$ jq .abi[2] build/contracts/MyContract.json 
{
  "constant": true,
  "inputs": [],
  "name": "t",
  "outputs": [
    {
      "name": "",
      "type": "uint256"
    }
  ],
  "payable": false,
  "stateMutability": "view",
  "type": "function"
}
```

However, **if I first compile V1, then modify the Solidity file changing it to V2, and recompile**, the resulting ABI looks like the following for `public t`:
```
{
  "constant": true,
  "inputs": [
    {
      "name": "_x",
      "type": "uint256"
    },
    {
      "name": "_s",
      "type": "string"
    }
  ],
  "name": "t",
  "outputs": [
    {
      "name": "",
      "type": "uint256"
    }
  ],
  "payable": false,
  "stateMutability": "view",
  "type": "function"
}
```

Note that, in V1, the third item of the ABI was the initialize function, which had that exact same look (except for the name). It would seem like the new function (public t) is getting mixed with the previous ABI definition.

If I clear out build/contracts and recompile, the ABI generated is the correct one.

## Environment

Have reproduced the issue with truffle 5.0.13 and 5.0.18. Version 5.0.12 seems to work fine, so the error should have been introduced in 13.

* Operating System: Ubuntu 18.04
* Truffle version (`truffle version`):
```
$ npx truffle version
Truffle v5.0.13 (core: 5.0.13)
Solidity v0.5.0 (solc-js)
Node v10.13.0
Web3.js v1.0.0-beta.37
```
* node version (`node --version`): 10.13.0
* npm version (`npm --version`): 6.9.0

CruzMolina:
This PR fixes an oversight implemented in #1894 to simply `_.merge` existing and new compiled artifacts rather than `_.merge` their known networks before `_.assign`ing other properties. Also cleans up the pkg a bit.

Resolves #2048 .

### DIFF
--- a/packages/truffle-artifactor/index.js
+++ b/packages/truffle-artifactor/index.js
@@ -34,7 +34,11 @@ class Artifactor {
       const normalizedExistingArtifact = Schema.normalize(
         existingArtifactObject
       );
-      _.merge(completeArtifact, normalizedExistingArtifact, normalizedArtifact);
+      _.assign(
+        completeArtifact,
+        normalizedExistingArtifact,
+        normalizedArtifact
+      );
       writeArtifact(completeArtifact);
     } catch (e) {
       // if artifact doesn't already exist, write new file
--- a/packages/truffle-artifactor/index.js
+++ b/packages/truffle-artifactor/index.js
@@ -10,8 +10,8 @@ class Artifactor {
   }
 
   async save(artifactObject) {
-    const normalizedArtifact = Schema.normalize(artifactObject);
-    const contractName = normalizedArtifact.contractName;
+    const normalizedNewArtifact = Schema.normalize(artifactObject);
+    const contractName = normalizedNewArtifact.contractName;
 
     if (!contractName) throw new Error("You must specify a contract name.");
 
@@ -37,12 +37,12 @@ class Artifactor {
       _.assign(
         completeArtifact,
         normalizedExistingArtifact,
-        normalizedArtifact
+        normalizedNewArtifact
       );
       writeArtifact(completeArtifact);
     } catch (e) {
       // if artifact doesn't already exist, write new file
-      if (e.code === "ENOENT") return writeArtifact(normalizedArtifact);
+      if (e.code === "ENOENT") return writeArtifact(normalizedNewArtifact);
       else if (e instanceof SyntaxError) throw e; // catches improperly formatted artifact json
       throw e; // catch all other errors
     }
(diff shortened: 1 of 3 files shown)


CruzMolina:
Looks like I need to take a look at some of the tests added/changed in #1894.

packages/truffle-artifactor/utils.js
```diff
@@ -0,0 +1,12 @@
...
+const writeArtifact = (_completeArtifact, _outputPath) => {
+  _completeArtifact.updatedAt = new Date().toISOString();
+  fse.writeFileSync(
+    _outputPath,
+    JSON.stringify(_completeArtifact, null, 2),
+    "utf8"
+  );
+};
+
+module.exports = writeArtifact;
```
gnidan:
```suggestion
module.exports = { writeArtifact };
```
 Cause "utils" is plural ^^

packages/truffle-artifactor/index.js
```diff
@@ -1,46 +1,49 @@
-var Schema = require("truffle-contract-schema");
-var fs = require("fs-extra");
-var path = require("path");
-var _ = require("lodash");
-var debug = require("debug")("artifactor");
+const Schema = require("truffle-contract-schema");
+const fse = require("fs-extra");
+const path = require("path");
+const _ = require("lodash");
+const writeArtifact = require("./utils");
```
gnidan:
```suggestion
const { writeArtifact } = require("./utils");
```

(to go with below)

PR #2050, opened 2019-05-24
Enable in-test breaking to debug transactions and calls

gnidan:
#2206 

_Work in progress. This PR is currently targeting Truffle v5.1 for maximum synergistic awesomeness with other features._

- [x] Add `truffle test --debug` flag, enabling the feature
- [x] Add `await debug(...)` global function to serve as interrupt point
- [x] Add `truffle test --debug --debug-global=<name>` to customize the name of the global (for non-clobbering reasons)
- [x] Support `debug(sendTransactionOperation)`
- [x] Support `debug(callOperation)`
- [ ] Support `debug(createOperation)`
- [ ] Support `debug(estimateGasOperation)`

Example:
```javascript
// regular call in test
const square = await instance.getSquare();

// add wrapper to debug that call
const square = await debug(instance.getSquare());
```

Voilà. Debugger instance.

Screen Shot 2019-05-25 at 2 48 55 AM

This function will passthrough the normal return value, so that when the debugger finishes, the test will proceed as normal.


gnidan:
Ah probably need to turn off mocha timeouts :)

Well it sorta works. Easier than expected!

gnidan:
Closing this in favor of broken-out atomic PRs

PR #2051, opened 2019-05-24
Internal improvement: refactor lib/contract/constructorMethods.js in truffle-contract

CruzMolina:
This PR attempts to make `lib/contract/constructorMethods.js` more maintainable and easier to reason about. During the process, some helper methods were created and added to `utils.js`, the `package.json` was tidied up a bit, and other misc changes were made.

### DIFF
--- a/packages/truffle-contract/lib/contract/constructorMethods.js
+++ b/packages/truffle-contract/lib/contract/constructorMethods.js
@@ -71,10 +71,10 @@ module.exports = Contract => ({
     }
   },
 
-  deployed() {
-    utils.checkProvider(this);
-
-    return this.detectNetwork().then(() => {
+  async deployed() {
+    try {
+      utils.checkProvider(this);
+      await this.detectNetwork();
       // We don't have a network config for the one we found
       if (this._json.networks[this.network_id] == null) {
         throw new Error(
@@ -92,9 +92,10 @@ module.exports = Contract => ({
           })`
         );
       }
-
       return new this(this.address);
-    });
+    } catch (error) {
+      throw error;
+    }
   },
 
   defaults(class_defaults) {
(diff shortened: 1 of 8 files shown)


CruzMolina:
Will attempt to add extra coverage shortly.

packages/truffle-contract/lib/contract/constructorMethods.js
```diff
@@ -1,391 +1,305 @@
...
-        }
-
-        return new constructor(constructor.address);
-      });
-    },
-
-    defaults: function(class_defaults) {
-      if (this.class_defaults == null) {
-        this.class_defaults = {};
+    } else {
```
gnidan: You can get rid of the `else` since every branch above either `return`s or `throw`s. Just if you want to remove an indent.

packages/truffle-contract/package.json
```diff
@@ -21,9 +21,9 @@
   "author": "Tim Coulter <tim.coulter@consensys.net>",
   "license": "MIT",
   "bugs": {
-    "url": "https://github.com/trufflesuite/truffle-contract/issues"
```
gnidan: Oh jeez this is why everyone thinks it's deprecated

packages/truffle-contract/test/networks.js
```diff
@@ -366,11 +366,13 @@ describe("Different networks: ", function() {
         from: ExampleTwo.defaults().from // Borrow the address from this one.
       });
 
-      NetworkExample.deployed()
-        .then(function(instance) {
-          assert.equal(NetworkExample.network_id, uri);
-          assert.equal(instance.address, json.networks[uri].address);
-          done();
+      NetworkExample.new(1)
```
gnidan: Why the extra `.new()`s?

CruzMolina: It appears there was a bug that allowed calling `.deployed()`on a new contract instance to set a `network_id` _before_ deployment. The docs are a bit out of date, but I _think_ that's not the correct behavior for that method.

gnidan: Oh, hm, seems a bit weird, then. I'll circle back on this next week and give it a think.

Issue #2028, opened 2019-05-16
Not able to debug a contract created by another contract that subsequently causes the transaction to revert


gilesbradshaw:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

I'm trying to debug  a transaction - I've got all the contracts linked to source code except one..

````
Addresses affected:
 0xe78A0F7E598Cc8b0Bb87894B0F60dD2a88d6a8Ab - ProcessRegistry
 0x7C728214be9A0049e6a86f2137ec61030D0AA964 - BindingAccessControl
 0x4bf749ec68270027C5910220CEAB30Cc284c7BA2 - TaskRoleContract_Contract
 0x3Cd8e615aF0774eE283329454962abbD4F2b427A - Order_To_Cash_Contract
 0xD833215cBcc3f914bD1C9ece3EE7BF8B14f841bb - Goods_Shipement_Factory
 0x5f8e26fAcC23FA4cbd87b8d9Dbbd33D5047abDE1 - Order_To_Cash_worklist
 0x8C6AAd0C92a48112aAa0e6e8F98A160120f17059(UNKNOWN)

````

It turns out  the one I can't link has a null binary - but it is in this one where the error is occurring.

All the others have been deployed by web3 - the one with the null binary (0x) has been created by another contract:

````
Goods_Shipement_Contract newContract = new Goods_Shipement_Contract(parent, worklist, processRegistry);
````

is there any way such contracts can be linked to their source code so they can be debugged?

For reference the null binary gets returned here:
packages/truffle-debugger/lib/session/sagas/index.js

````
let binaries = yield* web3.obtainBinaries(addresses, blockNumber);
````
## Steps to Reproduce

I can provide my source files if you'd like.. + ganache-cli data

## Environment

* Operating System: ubuntu
* Ethereum client: ganache-cli
* Truffle version Truffle v5.0.17 (core: 5.0.16)
* node version Node v11.14.0
* npm version 6.7.0

gilesbradshaw:
Looking at this a little bit more I've debugged a successful transaction where a new contract is created by the transaction:

````
Order_To_Cash_Contract newContract = new Order_To_Cash_Contract(parent, worklist, processRegistry);
````

so in this instance I DO get the binary for the created contract and I can debug it

````
Addresses affected::
[...]

PR #2052, opened 2019-05-24
Store code returned by creation calls

haltman-at:
The immediate purpose of this PR is to fix issue #2028.  However, if I just wanted to do that, I could've done it an easier way.  The broader purpose is to lay some more foundations for the reconstruct mode that's eventually coming.

So, as I said, the primary visible effect of this PR is to fix issue #2028.  Now, when a creation call successfully returns, we store in the state the code and address that it returned, so that other parts of the debugger will have access to it.  So, if you then enter a message call to a contract you just created, we will be able to sourcemap it and properly debug it, even if the transaction (or some part of it) later reverts, or even if the contract eventually selfdestructs.  (Note that selfdestructs are still a problem if you *didn't* just create the contract; that will require reconstruct mode to handle, I'm afraid.)

So, where in the state are we storing it?  Well, the relevant state has been reorganized a fair bit.  The old `evm.transaction.instances` is gone, along with its corresponding selector.  Instead, the code attached to each account is now stored in the codex.  (The selector `evm.current.codex.instances` has taken the place of the old selector.)    The `ADD_INSTANCE` action still exists, but it's the codex that responds to it.

Meanwhile, the old `evm` action `RETURN` has been split into `RETURN_CALL` and `RETURN_CREATE`.  The effect of `RETURN_CALL` is unchanged (other than an adjustment I'll get to later), but `RETURN_CREATE` first records the returned code in the codex before saving the stackframe.

(Note that there's no separate action for selfdestructs, because there's simply no need to track selfdestructs when you're only dealing with a single transaction; that will have to be split off when reconstruct mode is implemented, however.)

It addition to the code, it also records the context for easy lookup, so let's discus contexts.  Contexts for unknown contracts have been basically eliminated.  The selector `evm.current.context` will of course return something useful whether you're in a known context or not -- it won't return `null` unless no transaction is loaded -- but there are no more *stored* contexts for unknown contracts.  (Because of this, I was able to make a slight decoder alteration -- once again it is now always the case that in the decoder contexts are indexed by contract ID, rather than sometimes by ID and sometimes by hash.)

This together with the elimination of the old `instances` does mean we're no longer storing any easy way to look up, given an address, what other addresses share its context -- but, well, we weren't using that anywhere anyway.

But, like I was saying above, I didn't just do all that; I also took the time to make some other preparations for reconstruct mode.

Previously, the bottom stackframe in the codex represented the 0 stackframe, the outermost call of the transaction.  But now, the bottom stackframe represents the "-1 stackframe" -- the blockchain itself.  We don't really need this yet, because, as mentioned, the real purpose is for reconstruct mode, where at transaction's end we have to save the result to the -1 stackframe (or discard it in case of failure) but it's nice to have.

(The adjustment I mentioned to `RETURN_CALL` is that now we make sure to always leave *two* frames on the stack, not just one.  Similarly with `RETURN_CREATE` and `FAIL`, obviously.)

However, adding this necessitated some changes to how the reset button works.  Previously the reset button worked by resetting the codex to a default state.  But not a *fully* default state, like the unload transaction action; because resetting isn't followed by a call to *evm.begin*, so I had to keep around the results of that on the codex.

With these changes, that approach was no longer practical.  So, instead, the `evm.reset` saga now issues *two* actions.  The first is the `RESET` action, which pops the whole codex except for the -1 stackframe.  (Actually, because it's possible for `LOAD`s to modify the -1 stackframe, this means that now, after resetting, the debugger will remember any storage values it saw but didn't modify, rather than listing them as the default 0.  I'd consider this an improvement, but if you think this is confusing, well, we can figure something out, I guess.)

The second is a replay of the action issued by `evm.begin`.  How is that possible?  Well, we store it in the state!  There's a new component to the state, `evm.transaction.initialCall`, with corresponding selector.  It starts out as null, and when the first `CALL` or `CREATE` occurs, it becomes set to that action (yes, the action itself), and then stays as such until the transaction is unloaded.  So the `evm.reset` saga now first issues a `RESET` action, and then looks up the initial call and replays it, resetting the codex to its starting state.  (The `callstack` reducer has been appropriately adjusted for this as well.)

Let's say a bit more about the debugger startup.  There's been a sequencing change -- `evm.begin` now happens *after* recording the instances, to make all this work.  Also, since we add instances at the beginning for everything we call into, this means creating accounts for them in the codex, which means the `CALL` action no longer checks whether it needs to create an account for the address it calls into, because, well, it can be sure that one already exists.

Also during start up I changed back the fetching of binaries back to how it used to be, where it only fetches them for accounts we call into, not ones we only create.  Because why make extra RPC requests?  (Really we never needed those, I just added them for consistency, but now it strikes me as kind of dumb.)  This does mean the startup display is a bit worse, but, well, what are you going to do.  I modified the text a little to compensate?  This can maybe be changed.

Finally, I want to discuss the potential for weird interactions between this and decoding.  Because there's one other way an `ADD_INSTANCE` action can occur, other than at startup, and that's when we need to decode a variable of contract type or external function type, but we don't already have the code loaded for the corresponding contract.  Can this potentially create weird path-dependent effects?

I think the answer is basically no.  In reconstruct mode it will be definitely no, since we'd be fetching the code from the *previous* block, not the current one.  But even outside of reconstruct mode I'm pretty sure it's fine, even if this is less obvious.  Maybe if you had some crazy thing where `CREATE2` was used to create a contract on top of an old self-destructed one you could get a problem or something.  Honestly, I didn't bother worrying about that.  I don't think there's any reliable way to deal with such problems outside of reconstruct mode.

But the skeleton of that mode is starting to come together!

### DIFF
--- a/packages/truffle-core/lib/debug/printer.js
+++ b/packages/truffle-core/lib/debug/printer.js
@@ -40,7 +40,7 @@ class DebugPrinter {
     const affectedInstances = this.session.view(session.info.affectedInstances);
 
     this.config.logger.log("");
-    this.config.logger.log("Addresses affected:");
+    this.config.logger.log("Addresses called: (not created)");
     this.config.logger.log(
       DebugUtils.formatAffectedInstances(affectedInstances)
     );
--- a/packages/truffle-debugger/lib/data/sagas/index.js
+++ b/packages/truffle-debugger/lib/data/sagas/index.js
@@ -665,6 +665,10 @@ function fetchBasePath(
     astId: baseNode.id,
     stackframe: currentDepth
   });
+  debug("astId: %d", baseNode.id);
+  debug("stackframe: %d", currentDepth);
+  debug("fullId: %s", fullId);
+  debug("currentAssignments: %O", currentAssignments);
   //base expression is an expression, and so has a literal assigned to
   //it
   let offset = DecodeUtils.Conversion.toBN(
--- a/packages/truffle-debugger/lib/data/selectors/index.js
+++ b/packages/truffle-debugger/lib/data/selectors/index.js
@@ -193,10 +193,10 @@ const data = createSelectorTree({
 
     /*
      * data.views.instances
-     * same as evm.info.instances, but we just map address => binary,
-     * we don't bother with context
+     * same as evm.current.codex.instances, but we just map address => binary,
+     * we don't bother with context, and also the code is a Uint8Array
      */
-    instances: createLeaf([evm.transaction.instances], instances =>
+    instances: createLeaf([evm.current.codex.instances], instances =>
       Object.assign(
         {},
         ...Object.entries(instances).map(([address, { binary }]) => ({
@@ -222,7 +222,7 @@ const data = createSelectorTree({
         ...Object.values(contexts)
           .filter(context => !context.isConstructor)
           .map(context => ({
-            [context.context]: debuggerContextToDecoderContext(context)
+            [context.contractId]: debuggerContextToDecoderContext(context)
           }))
       )
     )
--- a/packages/truffle-debugger/lib/evm/actions/index.js
+++ b/packages/truffle-debugger/lib/evm/actions/index.js
@@ -70,10 +70,19 @@ export function create(binary, storageAddress, sender, value) {
   };
 }
 
-export const RETURN = "RETURN";
+export const RETURN_CALL = "RETURN_CALL";
 export function returnCall() {
   return {
-    type: RETURN
+    type: RETURN_CALL
+  };
+}
+
+export const RETURN_CREATE = "RETURN_CREATE";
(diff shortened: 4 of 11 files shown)


haltman-at:
Before I merge, probably worth explaining the "two stackframes" thing (recapping a discussion with @gnidan) on Slack.  In reconstruct mode, when going through preliminary transactions, we *will* go down to just 1 stackframe at the end.  However, here, where we're going through the transaction you're actually debugging, we don't want to do that, because we want to essentially freeze right at the end of the transaction so you can see its end state.

Basically, if the transaction failed, we don't want to go down to just 1 stackframe, because that would mean throwing away the end state.  Whereas if it succeeded, we don't want to do that because that would *save* the end state, preventing the reset button from working.

PR #2053, opened 2019-05-25
Read file source and save to vyper compilation output

fainashalts:
This PR seeks to add the source contents for Vyper contracts to the contracts build file upon the completion of a successful compilation.

In order to await the results of the `readFile()` function, it was necessary to make `compileAll()` async, as well as making the callback from `execVyper()` async. There may be a more elegant way to do this, so please let me know if that's the case. 

Also, I assume that no additional changes to the source_path object are necessary to make this code compatible with Windows systems, but have not tested on one.

### DIFF
--- a/packages/truffle-compile-vyper/index.js
+++ b/packages/truffle-compile-vyper/index.js
@@ -1,6 +1,6 @@
 const path = require("path");
 const exec = require("child_process").exec;
-const fse = require("fs-extra");
+const fs = require("fs");
 
 const async = require("async");
 const colors = require("colors");
@@ -113,7 +113,7 @@ function execVyper(options, source_path, callback) {
 }
 
 // compile all options.paths
-async function compileAll(options, callback) {
+function compileAll(options, callback) {
   options.logger = options.logger || console;
 
   compile.display(options.paths, options);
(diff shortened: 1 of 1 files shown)


packages/truffle-compile-vyper/index.js
```diff
@@ -112,15 +113,15 @@ function execVyper(options, source_path, callback) {
 }
 
 // compile all options.paths
-function compileAll(options, callback) {
+async function compileAll(options, callback) {
```
gnidan: Functions that take a callback can't also be `async`... it causes problems.

packages/truffle-compile-vyper/index.js
```diff
@@ -112,15 +113,15 @@ function execVyper(options, source_path, callback) {
...
   options.logger = options.logger || console;
 
   compile.display(options.paths, options);
 
   async.map(
     options.paths,
     function(source_path, c) {
-      execVyper(options, source_path, function(err, compiled_contract) {
+      execVyper(options, source_path, async function(err, compiled_contract) {
```
gnidan: Not sure about this, but I'm pretty sure we can't make this `async` either. Can we use the `readFileSync` instead?

fainashalts: Yeah I wasn’t a fan of making these async- it works in the sense that we get the file contents but I figured it was the wrong way.  

Was hesitant to use readFileSync because it’s synchronous but happy to change it if you think that’s the better way!

gnidan: Definitely! Otherwise who will `.then` the promise?

fainashalts: Fair enough! Was concerned about having a blocking method in case someone has a lot of Vyper files or something but that's probably not that big a concern at this time? 

Made the change, thanks for the review!

PR #2062, opened 2019-05-28
Account for exceptional halts from normal halting instructions

haltman-at:
This PR accounts for the possibility that a normal halting instruction might, in fact, halt exceptionally.  In addition, it also fixes what I think is likely a bug I introduced in #2052.  (A `RETURN` can fail if there are insufficient stack arguments; a `SELFDESTRUCT` can fail that same way, or by running out of gas, or if it's inside a static call.  `STOP` is, in fact, the only EVM instruction that is guaranteed not to fail.)

So first off, let me mention a quick improvement over #2052 -- in that previous PR, when determining the address created by a returning creation call, I looked at the stack to determine the address.  But this is foolish, because we already have the address on our callstack as `storageAddress`.  So now we use that instead, and I got rid of the selector I added in #2052 for getting that address.  (This was the bug, too, btw -- it would have a problem if we were at the end of the trace.)

Anyway, the main change in this PR is to the selector `evm.current.step.isExceptionalHalting`.  Previously, a normal halting instruction was assumed to never halt exceptionally.  Now, if the current instruction is a normal halting instruction, we check its return status (another new selector, `evm.current.step.returnStatus`).  If we're not at the last step of the trace, we just read this off of `evm.next.state.stack`.  If we are, we have to check the transaction's status instead.

However, the status wasn't previously being stored in the state, so now it is, in `evm.transaction.status`, along with corresponding selector.  Well, OK, it was being stored in the copy of the receipt stored in the `session` state, but that doesn't count.  Speaking of which, in the places in the tests and the interpreter where we used that to check the status, I changed it to now check `evm.transaction.status` instead.

And, that's really it!

### DIFF
--- a/packages/truffle-debugger/lib/evm/selectors/index.js
+++ b/packages/truffle-debugger/lib/evm/selectors/index.js
@@ -97,7 +97,7 @@ function createStepSelectors(step, state = null) {
      * .isHalting
      *
      * whether the instruction halts or returns from a calling context
-     * NOTE: this covers only orddinary halts, not exceptional halts;
+     * NOTE: this covers only ordinary halts, not exceptional halts;
      * but it doesn't check the return status, so any normal halting
      * instruction will qualify here
      */
@@ -144,8 +144,8 @@ function createStepSelectors(step, state = null) {
       callAddress: createLeaf(
         ["./isCall", state],
 
-        (matches, { stack }) => {
-          if (!matches) {
+        (isCall, { stack }) => {
+          if (!isCall) {
             return null;
           }
 
@@ -162,8 +162,8 @@ function createStepSelectors(step, state = null) {
       createBinary: createLeaf(
         ["./isCreate", state],
 
-        (matches, { stack, memory }) => {
-          if (!matches) {
+        (isCreate, { stack, memory }) => {
+          if (!isCreate) {
             return null;
           }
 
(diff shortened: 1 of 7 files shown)


packages/truffle-debugger/lib/evm/selectors/index.js
```diff
@@ -97,7 +97,9 @@ function createStepSelectors(step, state = null) {
      * .isHalting
      *
      * whether the instruction halts or returns from a calling context
-     * (covers only ordinary halds, not exceptional halts)
+     * NOTE: this covers only orddinary halts, not exceptional halts;
```
gnidan: typo `orddinary`

packages/truffle-debugger/lib/evm/selectors/index.js
```diff
@@ -481,12 +472,46 @@ const evm = createSelectorTree({
...
+       * (returns a boolean -- true for success, false for failure)
+       */
+      returnStatus: createLeaf(
+        [
+          "./isHalting",
+          "/next/state",
+          trace.stepsRemaining,
+          "/transaction/status"
+        ],
+        (matches, { stack }, remaining, finalStatus) => {
```
gnidan: I confirmed the related issue in codeclimate for this: I think we can get around this with `s/matches/isHalting`, and it'll also be a bit more clear at a glance what "to match" means.

haltman-at: Do you think this should also be changed elsewhere?

gnidan:
(Feel free to ignore codeclimate for this PR. We have determined that having it as a red X machine is not something we want to do at this time)

PR #2063, opened 2019-05-30
Separate contract decoder into its own package

haltman-at:
This PR splits out the contract decoder into its own package.  **This is a breaking change to the contract decoder**, not in terms of the actual behavior, but in that, well, it's in a different package now. (@davidmurdoch, @adrianmcli, be aware.  Although of course this will be bundled together with the other breaking changes.)

I was going to make other changes too but I've decided to save that for later, for when I write the event decoder; for now I'm leaving everything else as it was.


haltman-at:
Change: At @gnidan's suggestion, I've renamed `truffle-decoder` to `truffle-decoder-core` and `truffle-contract-decoder` to `truffle-decoder`.  So, @davidmurdoch, @adrianmcli, false alarm, it's staying the same package after all.  Anyway this is all targeted for `next` anyway so it's not like you'll be seeing these changes yet.

PR #2067, opened 2019-05-31
Enhancement: Enable expedient solc-js parsing when using docker or native solc

CruzMolina:
Continuation of #2007 building on the efforts of #2042.

**Context:**
Removing `solc-js` from `Parser.parseImports` significantly slowed docker compilations.

**Solution:**
This PR enables leveraging the default cached soljson (currently solc 0.5.8) specifically for speedy import parsing when using docker or native solc. Primary contract compilation is still handled by whatever the user is specifying.

To enable speedy JS parsing, users will need to specify a `parser: "solcjs"` key/value pair in their config like so:

```
module.exports = {
  compilers: {
    solc: {
      parser: "solcjs"
    }
  }  
}
```


CruzMolina:
Wow, thanks for that long detour Travis.

@gnidan, I think this is finally ready.

BrendanChou:
Any plans to merge this?

gnidan:
Ah thanks for the ping @BrendanChou. I think this was left in a waiting state while #2109 was in flight, but this might be good now? @CruzMolina this can be reviewed/merged pending the conflicts, right?

Issue #2070, opened 2019-06-02
Not usable on Node.js 12


filips123:
## Issue

Truffle and its sub-packages (`truffle-contract`) are unusable on Node.js 12.

This needs to be **fixed quickly** because it currently makes any package that uses `truffle-contract` **unusable on Node.js 12**.

## Details

Node.js 12 removed some deprecated functions from the V8 add-on API which are used by some of Truffle dependencies.

One of that dependencies is `sha3@1.2.2` which is used by `eth-lib#keccakjs`. To fix this, you need to update all dependencies `yarn.lock`. This should install `sha3@1.2.3` which has fixed compatibility with Node.js 12.

Another dependency with problems is `scrypt.js@0.2.0` which is used by `web3#web3-eth#web3-eth-accounts`. It was fixed in `scrypt.js@0.3.0`. Problem with this is that the fixed version is used only in `web3@1.0.0-beta.55` which is incompatible with Truffle (#1690). To fix this, you need to update Truffle to support `web3@1.0.0-beta.55` which will use the fixed version of `scrypt.js`.

Maybe there are also some other broken dependencies.

## Logs from CI

Error with `sha3`:
https://travis-ci.com/filips123/truffle/jobs/204732468

Error with `scrypt`:
https://travis-ci.com/filips123/EthAvatar.JS/jobs/204728583

## Environment

* Operating system: Ubuntu 14.04 on Travis CI
* Node.js version: 12.3.1
* NPM version: 6.9.0

gnidan:
Hey @filips123, thanks for bringing this up. This incompatibility is a known issue but unfortunately it's not something we can "fix quickly".

Firstly, upgrading to Web3.js v1.0.0-beta.55 is not an option, since there are breaking changes that would have to wait until Truffle v6. This upgrade has been something we've been attempting for at least 6 months, without success. I won't say more now, but expect an announcement soon about Web3.js versioning. A number of projects having been working with Web3.js directly and we are in the process of outlining a path forward here.

As for the sha3 issue, we can look into upgrading that; this may be more feasible. Feel free to open a pull request if you have a clear understanding of the work entailed.

Sorry about this. I understand it's frustrating—Node v12 has lots of features that I can't wait to get my hands on! Please bear with us and the rest of the community as we seek to address the limitations preventing this from working (or please help if you can!)

filips123:
[...]

PR #2071, opened 2019-06-02
Update dependencies

filips123:
Update dependencies in `yarn.lock`. Related to #2070 and resolved `sha3` issue. There is still a problem with `scrypt` because it is used by old Web3 versions.


gnidan:
Thanks for this!

Does this also need the package.json files updated? Or do they already account for the new versions?

filips123:
This is probably enough because share is required with `^1.2.2` 2hich will also require `1.2.3`.

I didn't delete and regenerate yaen.lock. I just tried to upgrade sha3 to 1.2.3. But it looks like yarn did that 😀

PR #2073, opened 2019-06-04
Introduce new truffle-compile interface

gnidan:
#2207 #2208

- Explicitly refer to all internal uses of truffle-compile via `truffle-compile/legacy`

- Define a new interface for truffle-compile, accessible via `require("truffle-compile/new")`.
    - Return `contractName` instead of `contract_name`

    - Remove `unlinked_binary`

    - Return an object with results `{ contracts, sourceIndex, compilerInfo }`
      where `contracts` is a list, not a name mapping.

    - Use external linkReferences of the form `{ offsets, name, length }`, instead of underscores

- Update truffle-compile tests to use new format

- Define separate `"truffle-workflow-compile/legacy"` and `"truffle-workflow-compile/new"` interfaces. New interface changes:
    - Stop automatically saving upon compilation
    - New interface has only two methods: `compile` and `save`
    - Return `{ contracts, compilations }`, where `contracts` is the aggregate list of all compiled contracts, and `compilations` is the raw result in truffle-compile's new return format.

- Refactor truffle-workflow-compile a bit into a few separate modules

- Define legacy->new (lossy) shim, so that `"truffle-workflow-compile/new"` can work despite no new interface implementation for truffle-compile-vyper or truffle-external-compiler. This PR does not add the new interface to either of those packages.

- Fix missing support of the compile `files` option, which allows explicit file paths to be compiled, otherwise not found by `compile.all` or `compile.necessary`

**Note** this PR should be a minor release for truffle-compile and truffle-workflow-compile


packages/truffle-workflow-compile/shims.js
```diff
@@ -0,0 +1,88 @@
...
+    .slice(2) // remove 0x prefix
+    .replace(/__[^_]+_*/g, (linkReference, characterOffset) => {
+      const [, name] = linkReference.match(/__([^_]+)_*/);
+
+      const characterLength = linkReference.length;
+
+      const offset = characterOffset / 2;
+      const length = characterLength / 2;
+
+      linkReferences.push({
```
fainashalts: 🎉

packages/truffle-compile/test/test_supplier.js
```diff
@@ -144,45 +135,35 @@ describe("CompilerSupplier", function() {
...
-                );
-              }
+      await compile(version4PragmaSource, cachedOptions);
 
-              done();
-            });
-          })
-          .catch(done);
-      });
+      assert(fs.existsSync(expectedCache), "Should have cached compiler");
```
fainashalts: Is there a reason to use `fs.existsSync()` over `fs.exists()` or `fse.exists()`? Looks like there's no problem awaiting here so may be worthwhile removing the synchronous function?

packages/truffle-compile/test/test_supplier.js
```diff
@@ -144,45 +135,35 @@ describe("CompilerSupplier", function() {
...
+
+      // Get cached solc access time
+      initialAccessTime = fs.statSync(expectedCache).atime.getTime();
```
fainashalts: Ditto question for `fs.statSync()`.

packages/truffle-compile/test/test_supplier.js
```diff
@@ -144,45 +135,35 @@ describe("CompilerSupplier", function() {
...
+
+      // Wait a second and recompile, verifying that the cached solc
+      // got accessed / ran ok.
+      await waitSecond();
+
+      const { contracts } = await compile(version4PragmaSource, cachedOptions);
+
+      finalAccessTime = fs.statSync(expectedCache).atime.getTime();
```
fainashalts: Ditto.

fainashalts:
+1 Minor comments but overall LGTM!

packages/truffle-compile/test/test_supplier.js
gnidan: Ah that's just legacy code, I didn't add that. I'll make the change.

packages/truffle-compile/test/test_supplier.js
gnidan: I'll also make this change.

packages/truffle-compile/test/test_supplier.js
gnidan: And here.

packages/truffle-compile/test/test_supplier.js
```diff
@@ -1,10 +1,11 @@
 const debug = require("debug")("compile:test:test_supplier");
-const fs = require("fs");
+const fs = require("fs-extra");
```
eggplantzzz: How do you feel about changing all instances of `fs` to `fse`? Cruz and I have been following this convention, I don't know if you would care to participate in the party...? If you don't feel like it, then I won't fight you :)

gnidan: you got it buddy

gnidan:
Thanks! Takin' it to mergetown.

PR #2077, opened 2019-06-04
Fix link value type in contract network object schema

seesemichaelj:
I was getting the contract verification error:
```
should be object (type):
{ dataPath: '.networks[\'1559252182005\'].links[\'ConvertLib\']',
  schemaPath: '#/type',
  params: { type: 'object' },
  data: '0xA516d46A37780b71e3d5BaB2070f345967f7b8e0',
```

using `truffle-contract-schema@3.0.10` and `truffle@5.0.15`
you can see the networks object shows that the `ConvertLib` property is a `string` and not an `object`:

```json
"networks": {
  "1559252182005": {
    "events": {},
    "links": {
      "ConvertLib": "0xA516d46A37780b71e3d5BaB2070f345967f7b8e0"
    },
    "address": "0x7d2aF93e5758C578a32B2F2977A74599e12c9a95",
    "transactionHash": "0xa607a79b6d6d77d826acf1954c15449cb0fd644e83b26fdc530fe472c564f454"
  }
}
```

Going down the rabbit hole, this looks like the schema should be updated rather than the code:

- Assignment shows it using an address hex string: https://github.com/trufflesuite/truffle/blob/4bb5caa49a68ab5a68ad4345fa5339511cc2ffd2/packages/truffle-contract/lib/contract/constructorMethods.js#L201
- The usage of that getter: https://github.com/trufflesuite/truffle/blob/4bb5caa49a68ab5a68ad4345fa5339511cc2ffd2/packages/truffle-contract/lib/contract/properties.js#L243-L248
- And the utility function showing that it expects strings (specifically looking for hex strings): https://github.com/trufflesuite/truffle/blob/4bb5caa49a68ab5a68ad4345fa5339511cc2ffd2/packages/truffle-contract/lib/utils.js#L135-L144

### DIFF
--- a/packages/truffle-contract-schema/spec/network-object.spec.json
+++ b/packages/truffle-contract-schema/spec/network-object.spec.json
@@ -17,7 +17,7 @@
     "links": {
       "type": "object",
       "patternProperties": {
-        "^[a-zA-Z_][a-zA-Z0-9_]*$": { "$ref": "#/definitions/Link" }
+        "^[a-zA-Z_][a-zA-Z0-9_]*$": { "$ref": "#/definitions/Address" }
       },
       "additionalProperties": false
     }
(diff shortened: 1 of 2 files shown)


gnidan:
Good find, thanks for exercising the schema so much @seesemichaelj!

PR #2078, opened 2019-06-04
Fix interaction of continue command with unmapped code

haltman-at:
This PR fixes some issues resulting from using the `c` command when there is unmapped code.  You could get a problem if you used it from within unmapped code, or if you continued to the end when the end is unmapped.  I fixed this by adding a `break` to the loop if you're already finished (so that the rest of the loop, that could cause a problem, never runs) and by adding a `continue` to the loop if you're in unmapped code (for the same reason) and rearranged other checks so that this works.  Pretty straightforward.

### DIFF
--- a/packages/truffle-debugger/lib/controller/sagas/index.js
+++ b/packages/truffle-debugger/lib/controller/sagas/index.js
@@ -196,10 +196,6 @@ function* stepOver() {
  * continueUntilBreakpoint - step through execution until a breakpoint
  */
 function* continueUntilBreakpoint(action) {
-  var currentLocation, currentNode, currentLine, currentSourceId;
-  var finished;
-  var previousLine, previousSourceId;
-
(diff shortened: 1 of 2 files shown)


packages/truffle-debugger/lib/controller/sagas/index.js
```diff
@@ -222,17 +221,22 @@ function* continueUntilBreakpoint(action) {
...
+    debug("currentLocation: %O", currentLocation);
     finished = yield select(controller.current.trace.finished);
-    debug("finished %o", finished);
+    if (finished) {
+      break; //can break immediately if finished
+    }
 
+    currentSourceId = currentLocation.source.id;
+    if (currentSourceId === undefined) {
+      continue; //never stop on an unmapped instruction
```
gnidan: sounds like dangerous business!

PR #2081, opened 2019-06-06
Internal improvement: Update truffle test code

eggplantzzz:
In the interest of making the code for truffle test more readable and maintainable, this PR was created! It removes the callback being passed to the run method for `packages/truffle-core/lib/test.js` and instead adopts a Promise-based interface. The callback is now being handled in `packages/truffle-core/lib/commands/test.js`. The war on callbacks wages on!

### DIFF
--- a/packages/truffle-core/lib/testing/testsource.js
+++ b/packages/truffle-core/lib/testing/testsource.js
@@ -1,8 +1,8 @@
-var Deployed = require("./deployed");
-var path = require("path");
-var fs = require("fs");
-var contract = require("truffle-contract");
-var find_contracts = require("truffle-contract-sources");
+const Deployed = require("./deployed");
+const path = require("path");
+const fse = require("fs-extra");
+const contract = require("truffle-contract");
+const find_contracts = require("truffle-contract-sources");
 
 function TestSource(config) {
   this.config = config;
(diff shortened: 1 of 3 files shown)


packages/truffle-core/lib/testing/testsource.js
```diff
@@ -13,87 +13,83 @@ TestSource.prototype.require = function() {
...
+      const promises = abstraction_files.map(file => {
+        return new Promise((accept, reject) => {
+          fs.readFile(
+            path.join(self.config.contracts_build_directory, file),
+            "utf8",
+            function(err, body) {
+              if (err) return reject(err);
+              accept(body);
+            }
+          );
```
fainashalts: This seems a little clunky -- do you need the `return new Promise` part?  Since `fs.readFile` is async won't it already return a promise? Feel free to ignore me if I'm missing something here! 

You could also just wrap the map here in a `Promise.all` and chain the `then` code from below to it since line 65 seems to be the only place you use the `promises` variable, but that's just a stylistic thing, I think its fine either way.

eggplantzzz: I don't think those fs methods return promises but rather have a callback interface...HOWEVER, it looks like we can just drop in `fs-extra` instead. Those asynchronous methods DO return Promises. That way it will be more prettier ;)

PR #2095, opened 2019-06-12
Internal improvement: Clean up develop and console code

eggplantzzz:
In an effort to provide better maintainability, this PR seeks to clarify the logic used by the console and develop commands. Callback-style functions are typically converted to Promise-based structures and there are also a handful of syntax updates.


packages/truffle-core/lib/develop.js
```diff
@@ -20,18 +20,16 @@ var Develop = {
...
-    var cmd = spawn("node", [chainPath, ipcNetwork, JSON.stringify(options)], {
+    return spawn("node", [chainPath, ipcNetwork, JSON.stringify(options)], {
       detached: true,
       stdio: "ignore"
     });
-
-    callback();
   },
 
   connect: function(options, callback) {
```
gnidan: Can't get rid of this callback?

eggplantzzz: Is this an old comment from before I removed this callback?

packages/truffle-core/lib/console.js
```diff
@@ -1,248 +1,225 @@
...
-                  new Error("Cannot parse " + file + ": " + e.message)
-                );
-              }
-
-              accept(body);
-            }
-          );
-        })
+        fse
+          .readFile(
```
CruzMolina: can we use `fse.readFileSync` here?
that way we can keep `provision` a sync method and remove `Promise.all` (maybe rename the `promises` array to `jsonBlobs`).

packages/truffle-core/lib/console.js
```diff
@@ -1,248 +1,225 @@
...
+    - allow whitespace before everything else
+    - optionally capture `var|let|const <varname> = `
+      - varname only matches if it starts with a-Z or _ or $
+        and if contains only those chars or numbers
+      - this is overly restrictive but is easier to maintain
+    - capture `await <anything that follows it>`
+    */
+    let includesAwait = /^\s*((?:(?:var|const|let)\s+)?[a-zA-Z_$][0-9a-zA-Z_$]*\s*=\s*)?(\(?\s*await[\s\S]*)/;
+
+    var match = cmd.match(includesAwait);
```
fainashalts: Can these be `let` or am I missing something re: the scope?

packages/truffle-core/lib/console.js
```diff
@@ -1,248 +1,225 @@
...
+    var source = cmd;
+    var assignment = null;
+
+    // If our code includes an await, add special processing to ensure it's evaluated properly.
+    if (match) {
+      var assign = match[1];
```
fainashalts: Same question here re: `var` or `let`

packages/truffle-core/lib/console.js
```diff
@@ -1,248 +1,225 @@
...
-      displayErrors: true,
-      breakOnSigint: true,
-      filename: filename
+    const runScript = script => {
+      const options = {
+        displayErrors: true,
+        breakOnSigint: true,
+        filename: filename
+      };
+      return script.runInContext(context, options);
```
fainashalts: Are you sure this works? I don't think a function from inside `script` is accessible inside the arrow function because of the way context works in arrow functions.

packages/truffle-core/lib/console.js
eggplantzzz: Yeah I don't see any reason these can't be updated!

packages/truffle-core/lib/console.js
eggplantzzz: Your concern is about `script` and accessing it's functions? If script has a function on it then it should be fine. The thing to worry about would be the context. I think `context` should be fine here?

packages/truffle-core/lib/console.js
eggplantzzz: Sure! However, I wonder where it is especially useful to use async methods for performance perks...any thoughts @gnidan?

packages/truffle-core/lib/console.js
fainashalts: My concern is that you are passing script as an object in the arrow function parameter, right? And `runInContext()` is an object method of script's, right? That should return undefined if I'm not missing something. Since the context of the arrow function is inherited from the execution context but the arrow function doesn't bind context/this I think there's an issue with making this an arrow function.

PR #2099, opened 2019-06-14
Internal improvement: shorten save function (maintainability) 

CruzMolina:
Inspired by Code Climate, this lil PR aims to improve `truffle-artifactor`'s maintainability and reduce tech debt.

https://codeclimate.com/github/trufflesuite/truffle/compare/ccArt

### DIFF
--- a/packages/truffle-artifactor/utils.js
+++ b/packages/truffle-artifactor/utils.js
@@ -1,31 +1,31 @@
 const fse = require("fs-extra");
 const _ = require("lodash");
 
-const writeArtifact = (_completeArtifact, _outputPath) => {
-  _completeArtifact.updatedAt = new Date().toISOString();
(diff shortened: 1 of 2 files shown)


packages/truffle-artifactor/utils.js
```diff
@@ -9,4 +10,22 @@ const writeArtifact = (_completeArtifact, _outputPath) => {
   );
 };
 
-module.exports = { writeArtifact };
+const finalizeArtifact = (
+  _normalizedExistingArtifact,
+  _normalizedNewArtifact
```
gnidan: What's up with all these underscores in this file

CruzMolina: Probably not necessary now that the helpers are in a separate file and not really "private" anymore.
If it's confusing and/or inconsistent, I can remove 'em.

PR #2100, opened 2019-06-15
Internal improvement: refactor similar getBlock code

CruzMolina:
Inspired by Code Climate, this lil PR aims to improve `truffle-blockchain-utils`' maintainability and reduce tech debt.

https://codeclimate.com/github/trufflesuite/truffle/compare/ccBUtils

### DIFF
--- a/packages/truffle-blockchain-utils/index.js
+++ b/packages/truffle-blockchain-utils/index.js
@@ -1,21 +1,16 @@
-const defaultJSONQuery = {
-  jsonrpc: "2.0",
-  method: "",
-  params: [],
-  id: Date.now()
-};
-
 const Blockchain = {
   getBlockByNumber(blockNumber, provider, callback) {
-    defaultJSONQuery.params = [blockNumber, true];
-    defaultJSONQuery.method = "eth_getBlockByNumber";
-    provider.send(defaultJSONQuery, callback);
+    provider.send(
+      { params: [blockNumber, true], method: "eth_getBlockByNumber" },
+      callback
+    );
   },
 
   getBlockByHash(blockHash, provider, callback) {
-    defaultJSONQuery.params = [blockHash, true];
-    defaultJSONQuery.method = "eth_getBlockByHash";
-    provider.send(defaultJSONQuery, callback);
+    provider.send(
+      { params: [blockHash, true], method: "eth_getBlockByHash" },
+      callback
+    );
   },
 
   parse(uri) {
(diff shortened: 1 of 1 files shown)


packages/truffle-blockchain-utils/index.js
```diff
@@ -1,28 +1,22 @@
...
+  getBlockBy(NumberOrHash, value, provider, callback) {
+    defaultJSONQuery.params = [value, true];
+    if (NumberOrHash === "Number")
+      defaultJSONQuery.method = "eth_getBlockByNumber";
+    else if (NumberOrHash === "Hash")
+      defaultJSONQuery.method = "eth_getBlockByHash";
+    else
+      throw new Error(
+        `.getBlockBy() only currently supports querying a "Number" or "Hash"!`
+      );
```
adrianmcli: I'd push back on having string-denoted things that alter a function's behaviour. A bit of a code smell IMO. I actually think the duplication here is not that big a deal and it's actually more complex and less readable than before.

adrianmcli:
I'd push back on having string-denoted things that alter a function's behaviour. A bit of a code smell IMO. I actually think the duplication here is not that big a deal and it's actually more complex and less readable than before.

CruzMolina:
Thanks for sharing your thoughts @adrianmcli! What do you think about the [first commit](https://github.com/trufflesuite/truffle/pull/2100/commits/8733cfb56a6b2e4ba8fadea9050e0211e1118a2c)?

adrianmcli:
@CruzMolina I still think that would be a type of in-direction. I would prefer duplication in this case because it's clearer to an unfamiliar reader of the code what is happening.

CruzMolina:
Turns out we can simplify w/o duplication https://github.com/CruzMolina/truffle/commit/dada32741146148f5e938b7f30dd468f66773eed

adrianmcli:
@CruzMolina that's much better

packages/truffle-blockchain-utils/index.js
```diff
@@ -1,26 +1,14 @@
 const Blockchain = {
   getBlockByNumber(blockNumber, provider, callback) {
-    const params = [blockNumber, true];
     provider.send(
-      {
-        jsonrpc: "2.0",
```
gnidan: We don't need `jsonrpc` / `id`? They get added automatically?

CruzMolina: Ah, looks like we don't actually test any of this code against `geth`.

CruzMolina: Hmm, also need to consider whether an `httpProvider`, `ws` or `ipc` provider is being used...

CruzMolina:
Closing this for now. I *think* the current code suffices for supported `web3` provider types. Turns out this wasn't such an easy win after all!

PR #2105, opened 2019-06-17
remove a devDependency

gingerhot:


gingerhot:
Sorry for a careless click operation from my forked repository : (

PR #2107, opened 2019-06-18
Use current AST format instead of legacy for ordering ABI output

gnidan:
(Internal improvement)

Truffle's compilation process orders output ABI in source-defined order using the AST parse info. This PR switches to using the current `ast` compilation source property instead of the `legacyAST` compilation source property.

+some refactoring

### DIFF
--- a/packages/truffle-compile/index.js
+++ b/packages/truffle-compile/index.js
@@ -280,72 +280,38 @@ function replaceLinkReferences(bytecode, linkReferences, libraryName) {
   return bytecode;
 }
 
-function orderABI(contract) {
-  var contract_definition;
-  var ordered_function_names = [];
+function orderABI({ abi, contract_name: contractName, legacyAST }) {
+  // AST can have multiple contract definitions, make sure we have the
+  // one that matches our contract
+  const contractDefinition = legacyAST.children.filter(
+    ({ name: nodeType, attributes: { name } }) =>
+      nodeType === "ContractDefinition" && name === contractName
+  )[0];
 
-  for (var i = 0; i < contract.legacyAST.children.length; i++) {
-    var definition = contract.legacyAST.children[i];
-
-    // AST can have multiple contract definitions, make sure we have the
-    // one that matches our contract
-    if (
-      definition.name !== "ContractDefinition" ||
-      definition.attributes.name !== contract.contract_name
-    ) {
-      continue;
-    }
-
-    contract_definition = definition;
-    break;
+  if (!contractDefinition || !contractDefinition.children) {
+    return abi;
   }
 
-  if (!contract_definition) return contract.abi;
-  if (!contract_definition.children) return contract.abi;
-
-  contract_definition.children.forEach(function(child) {
-    if (child.name === "FunctionDefinition") {
-      ordered_function_names.push(child.attributes.name);
-    }
-  });
+  const orderedFunctionNames = contractDefinition.children
+    .filter(({ name: nodeType }) => nodeType === "FunctionDefinition")
+    .map(({ attributes: { name: functionName } }) => functionName);
 
   // Put function names in a hash with their order, lowest first, for speed.
-  var functions_to_remove = ordered_function_names.reduce(function(
-    obj,
-    value,
-    index
-  ) {
-    obj[value] = index;
-    return obj;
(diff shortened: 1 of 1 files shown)


packages/truffle-compile/index.js
```diff
@@ -280,72 +280,39 @@ function replaceLinkReferences(bytecode, linkReferences, libraryName) {
...
-    ) {
-      continue;
-    }
-
-    contract_definition = definition;
-    break;
+function orderABI({ abi, contract_name: contractName, ast }) {
+  // AST can have multiple contract definitions, make sure we have the
+  // one that matches our contract
+  const contractDefinition = ast.nodes.filter(
```
haltman-at: I'd suggest using `arr.find(...)` rather than `arr.filter(...)[0]`.

packages/truffle-compile/index.js
```diff
@@ -280,72 +280,39 @@ function replaceLinkReferences(bytecode, linkReferences, libraryName) {
...
-    newABI.push(item);
-  });
-
-  // Now pop the ordered functions definitions on to the end of the abi..
-  Array.prototype.push.apply(newABI, function_definitions);
-
-  return newABI;
+  const functionIndexes = orderedFunctionNames
+    .map((functionName, index) => ({ [functionName]: index }))
+    .reduce((a, b) => Object.assign({}, a, b), {});
```
haltman-at: Hahaha this is yet another way of functionally constructing objects :P

haltman-at: Actually, wait, this has a real potential problem, in that it assumes function names are unique.  But Solidity allows overloading, and constructors and fallback functions both appear as `FunctionDefinition`s named `""`.  You're going to get a bunch of collisions here.  This either needs to be done a differently, or those need to be handled separately.

packages/truffle-compile/index.js
```diff
@@ -280,72 +280,39 @@ function replaceLinkReferences(bytecode, linkReferences, libraryName) {
...
+
+  // Construct new ABI with functions at the end in source order
+  return [
+    ...abi.filter(({ name }) => functionIndexes[name] === undefined),
```
haltman-at: I gather it's intentional that this sorts events to the top, in arbitrary order.  I'm guessing it's also intentional that it sorts inherited-and-not-overridden functions to the top in arbitrary order as well?  (Given as that's what the old code also did.)

packages/truffle-compile/index.js
haltman-at: Actually I guess that's ultimately OK seeing as this is only being used for sorting?  Like, now that I look, the old code had this problem as well.  I was going to request changes here but if the old code also did that I guess holding it up based on that is a bit silly.  Still, it's worth noting that really matching a function to an ABI entry is more work than just checking the name, so it's more work to really get the correct order, rather than grouping together functions with the same name.  (I have a function for exactly this sort of thing on my `wire-decoder` branch; maybe that could get moved to a separate PR so it could be used here.  Though presumably it'd need to go somewhere other than `truffle-decode-utils`.)

packages/truffle-compile/index.js
gnidan: That's a thing!? Heck yeah!

packages/truffle-compile/index.js
gnidan: This is just how it's been implemented. I'm not sure if that's necessary; I'll find out.

packages/truffle-compile/index.js
gnidan: Apparently this is a can of worms! I move to leave this limitation as-is for the time being, until the changes you mention are merged.

packages/truffle-compile/index.js
gnidan: Doesn't look like this matters: at least, in the Solidity testing stuff, it just filters out non-functions: https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/lib/testing/soliditytest.js#L77

gnidan: (And same as above: I move that we don't bother with this kind of semantic changes until the aforementioned `wire-decoder` stuff is available)

gnidan:
Thanks @haltman-at!

PR #2108, opened 2019-06-18
Add native tag to parser test needing docker/native

gnidan:
_(Internal improvement)_

Mark a few truffle-compile tests as being `@native` because they use Docker. I've seen this done elsewhere so I assume it's kosher?


gnidan:
Thanks!

PR #2109, opened 2019-06-18
Refactor truffle-compile's solc integration

gnidan:
This PR substantially refactors `truffle-compile/index.js`, primarily by:
- Moving the compilation logic to a `"truffle-compile/run"` module, providing `async run()`
- Moving the interface `compile.all`, etc. into `"truffle-compile/legacy"`
- Shimming `run` to get legacy output

## Interface notes / Context

This PR is intended to be an implementation change only. This should not constitute a breaking change, nor does it add a new interface for use externally. The scope of this PR is restricted to the addition of internal modules, for use in future work.

As a result of this PR, `index.js` is notably pretty tiny now: the future work here will be to introduce an interface switching mechanism, exposing three ways to require truffle-compile (names may change):
a. `require("truffle-compile/legacy")` to get the old interface
b. `require(truffle-compile/new")` to get the new interface
c. `require("truffle-compile")` to continue to forward to legacy by default

The hope is to provide a low-risk method of making this breakage cutover over time.
1. We begin recommending the use of (a) as a way to ensure compatibility to anyone and everyone, including inside the rest of the codebase.
2. We offer (b) as a means to get non-lossy output wherever it's needed (#2050, for instance)
3. Later, we flip the switch to make (c) point to the new format.

## Implementation notes

How `async run()`'s interface differs from truffle-compile's:
- `contracts` are a list, not a name mapping
- `contractName` is used instead of `contract_name`
- Bytecodes are composite objects `{ bytes, linkReferences }` instead of strings
- Bytecode byte representations are different
  - Link references are now zeroed out: no more `__uNdErScorEs______` blocks
  - No `0x` prefix
- `async run()` returns  `Promise<{ sourceIndexes, contracts, compilerInfo }>`, not a list of those things passed to a callback.

## Notes for review

Hopefully the commits here are useful in making sense of such an enormously useless diff! I tried to clean them up. Also might be better to disable whitespace changes in the diff.


eggplantzzz:
I think this is looking good. Are we going to do something similar with the Vyper stuff as well?

gnidan:
That's the plan @eggplantzzz!

packages/truffle-compile/run.js
```diff
@@ -396,42 +396,50 @@ function processContracts({
...
 
-  const unprefixed = libraryLinkReferences.reduce(
-    (bytecode, { libraryName, links }) =>
-      replaceLinkReferences(bytecode, links, libraryName),
-    bytecode
-  );
-
-  return `0x${unprefixed}`;
+  // convert to { offsets, length, name } format
+  return libraryLinkReferences.map(({ name, links }) => ({
```
fainashalts: woohoo! :)

packages/truffle-compile/run.js
```diff
@@ -358,4 +344,48 @@ async function invokeCompiler({ compilerInput, options }) {
...
+  // extract messages
+  let errors = rawErrors.map(({ formattedMessage }) => formattedMessage).join();
+  const warnings = rawWarnings
+    .map(({ formattedMessage }) => formattedMessage)
+    .join();
+
+  if (errors.includes("requires different compiler version")) {
+    const contractSolcVer = errors.match(/pragma solidity[^;]*/gm)[0];
+    const configSolcVer =
+      options.compilers.solc.version || semver.valid(solc.version());
```
gnidan: Oops. Gotta fix this.

gnidan: (I'll refrain from rebasing this into the right commit)

gnidan:
Thanks for the reviews!

PR #2112, opened 2019-06-18
Internal improvement: Refactor truffle-box/utils & add some coverage

CruzMolina:
This PR modernizes & updates the `truffle-box/lib/utils` files to minimize callbacks. Adds `request-promise-native` as a dep and streamlines truffle box repo verification. Also adds some code coverage.

### DIFF
--- a/packages/truffle-box/lib/utils/unbox.js
+++ b/packages/truffle-box/lib/utils/unbox.js
@@ -10,37 +10,33 @@ const inquirer = require("inquirer");
 function verifyURL(url) {
   // Next let's see if the expected repository exists. If it doesn't, ghdownload
   // will fail spectacularly in a way we can't catch, so we have to do it ourselves.
-  return new Promise(function(accept, reject) {
-    var configURL = parseURL(
-      vcsurl(url)
+  return new Promise((accept, reject) => {
+    const configURL = parseURL(
+      `${vcsurl(url)
         .replace("github.com", "raw.githubusercontent.com")
-        .replace(/#.*/, "") + "/master/truffle-box.json"
+        .replace(/#.*/, "")}/master/truffle-box.json`
     );
 
-    var options = {
+    const options = {
       method: "HEAD",
-      uri: "https://" + configURL.host + configURL.path
+      uri: `https://${configURL.host}${configURL.path}`
     };
-    request(options, function(error, r) {
+    request(options, (error, { statusCode }) => {
       if (error) {
         return reject(
           new Error(
-            "Error making request to " +
-              options.uri +
-              ". Got error: " +
-              error.message +
-              ". Please check the format of the requested resource."
+            `Error making request to ${options.uri}. Got error: ${
+              error.message
+            }. Please check the format of the requested resource.`
           )
         );
-      } else if (r.statusCode === 404) {
+      } else if (statusCode === 404) {
         return reject(
           new Error(
-            "Truffle Box at URL " +
-              url +
-              " doesn't exist. If you believe this is an error, please contact Truffle support."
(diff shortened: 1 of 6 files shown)


packages/truffle-box/lib/utils/unbox.js
```diff
@@ -1,88 +1,73 @@
...
-          });
-        })
-    );
+    .map(async filePath => {
+      try {
+        fse.removeSync(filePath);
+      } catch (error) {
+        throw error;
+      }
+    });
```
gnidan: Hm we don't actually use those promises for anything... we can just change this to a single `.map` and an outer `for` loop

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -135,23 +119,22 @@ async function copyTempIntoDestination(tmpDir, destination, options) {
...
-    exec(postUnpack, { cwd: destination }, function(err, stdout, stderr) {
-      if (err) return reject(err);
-      accept(stdout, stderr);
-    });
-  });
+  try {
+    execSync(postUnpack, { cwd: destination });
+  } catch (error) {
+    throw error;
+  }
```
gnidan: We can get rid of this `try`/`catch` because we don't anything with the exception ourselves

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -1,90 +1,73 @@
...
-            resolve();
-          });
-        })
-    );
-
-  return Promise.all(promises);
+    .forEach(filePath => {
+      try {
+        fse.removeSync(filePath);
+      } catch (error) {
```
eggplantzzz: Can we remove this catch?

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -1,90 +1,73 @@
-const fs = require("fs-extra");
+const fse = require("fs-extra");
 const path = require("path");
 const ghdownload = require("github-download");
-const request = require("request");
+const rp = require("request-promise-native");
 const vcsurl = require("vcsurl");
 const parseURL = require("url").parse;
-const exec = require("child_process").exec;
+const execSync = require("child_process").execSync;
```
eggplantzzz: Not really a big deal or something that needs to hold this back, but how do you like the destructured version of this?
`const { execSync } = require("child_process");`

packages/truffle-box/lib/utils/unbox.js
```diff
@@ -1,90 +1,73 @@
...
+  );
+
+  const options = {
+    method: "HEAD",
+    uri: `https://${configURL.host}${configURL.path}`,
+    resolveWithFullResponse: true,
+    simple: false
+  };
+
+  try {
```
eggplantzzz: I would say that it looks like this try/catch can be removed as well as it does not seem to do anything.

packages/truffle-box/lib/utils/unbox.js
CruzMolina: I assumed `removeSync` would throw if a file needing removal wasn't there, but looks like `removeSync` will just silently do nothing.

packages/truffle-box/lib/utils/unbox.js
CruzMolina: Under certain conditions the request fails before returning a `statusCode`. I imagine we probably want to keep a catch there _just in case_.

packages/truffle-box/lib/utils/unbox.js
CruzMolina: Good idea, I think it's cleaner.

PR #2115, opened 2019-06-19
Internal improvement: remove unused devDeps & failing test import

CruzMolina:
This PR removes `go-ipfs-dep` & `ipfsd-ctl` as `devDependencies` along with an unused failing test import from `ethpm`.

### DIFF
--- a/packages/truffle-core/package.json
+++ b/packages/truffle-core/package.json
@@ -78,7 +78,6 @@
   ],
   "devDependencies": {
     "glob": "^7.1.2",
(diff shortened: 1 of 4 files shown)


eggplantzzz:
So you probably already had this conversation, but where is `go-ipfs-dep` and `ipfsd-ctl` used?

CruzMolina:
`go-ipfs-dep` is a dependency of `ipfsd-ctl`, which is `require`d by `ethpm/lib/index/github-examples` 🌀🙃

Issue #468, opened 2017-06-29
Warn/error about unsupported flags


JustinDrake:
Unsupported command line flags are silently ignored. For example, `truffle test --unsupportedgarbage` will ignore the `--unsupportedgarbage` flag.

I suggest warning about such flags, or maybe even erroring out.

CruzMolina:
## Requirements

1. When truffle is passed an unsupported command line flag, it should throw a helpful error.

Example:
```
truffle test --unsupportedFlag

Error: `truffle test` does not currently support the `--unsupportedFlag`.
Please see https://truffleframework.com/docs/truffle/reference/truffle-commands
for currently supported command flags.
```

2. Add an integration mocha test to verify proper error handling.

## Prerequisite Context

- Familiarity with the current entry point for Command handling: https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/lib/command.js. In particular [these lines](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/lib/command.js).
- Familiarity with current `truffle` command integration tests, i.e. https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/test/commands.js.

eggplantzzz:
Should this throw an error or perhaps just a warning?

eggplantzzz:
This has been fixed for a while now. Closing this for issue maintenance.

PR #2116, opened 2019-06-20
Warn/error about unsupported flags #468

skywing:
Check the command line options (flags) against the help options. If the options are not in the help option return the error with the message. This will stop the command execution. 

Error for unsupported flags will provide a clear message to the user that one of the option/flag in the command line is not correct and has no effect to the command.

### DIFF
--- a/packages/truffle-debugger/lib/data/sagas/index.js
+++ b/packages/truffle-debugger/lib/data/sagas/index.js
@@ -403,7 +403,7 @@ function* variablesAndMappingsSaga() {
           indexValue = yield* decode(splicedDefinition, indexReference);
         } else if (
           indexDefinition.referencedDeclaration &&
-          scopes[indexDefinition.referenceDeclaration]
+          scopes[indexDefinition.referencedDeclaration]
         ) {
           //there's one more reason we might have failed to decode it: it might be a
           //constant state variable.  Unfortunately, we don't know how to decode all
@@ -426,7 +426,11 @@ function* variablesAndMappingsSaga() {
               indexValue = yield* decode(keyDefinition, {
                 definition: indexConstantDeclaration.value
               });
+            } else {
+              indexValue = null; //can't decode; see below for more explanation
             }
+          } else {
+            indexValue = null; //can't decode; see below for more explanation
           }
         }
         //there's still one more reason we might have failed to decode it:
--- a/packages/truffle-debugger/test/data/more-decoding.js
+++ b/packages/truffle-debugger/test/data/more-decoding.js
@@ -97,6 +97,9 @@ contract ElementaryTest {
   mapping(string => string) stringMap;
   mapping(address => address) addressMap;
 
+  //constant state variables to try as mapping keys
+  uint constant two = 2;
+
   function run() public {
     //local variables to be tested
     byte oneByte;
@@ -114,6 +117,7 @@ contract ElementaryTest {
     bytesMap[hex"01"] = hex"01";
 
     uintMap[1] = 1;
+    uintMap[two] = two;
 
     intMap[-1] = -1;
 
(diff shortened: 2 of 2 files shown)


packages/truffle-core/lib/command.js
```diff
@@ -100,6 +100,35 @@ class Command {
       }
     });
 
+    //Check unsupported command line flag according to the option list in help
+    try {
+      const inputOptions = [];
+      inputStrings.map(i => {
+        if (i.startsWith("--")) {
```
Elaniobro: should the `if` statements maintain consistency? If so, which one is preferred, this style and or no `{}` for one liners like you have on [line 115](https://github.com/trufflesuite/truffle/pull/2116/files#diff-8e8c8e19fd9034bcaa495fe10529a3d3R115)?

eggplantzzz: My personal preference is no `{}`s for one-liners. Otherwise I always include them. I'm not sure we have established a strict convention though.

packages/truffle-core/lib/command.js
```diff
@@ -100,6 +100,35 @@ class Command {
...
+      let notValidOptions = inputOptions.filter(
+        opt => !validOptions.includes(opt)
+      );
+
+      if (notValidOptions.length > 0)
+        return callback(
+          new TaskError(
+            "Unsupported (Undocumented) command line option: " + notValidOptions
+          )
+        );
```
eggplantzzz: Hmm, so if you are up for it I think we can extend this `try/catch` and encapsulate all the code from here through the next `try/catch` since the catch's do the same thing. In other words, something like
```
      if (notValidOptions.length > 0)
        return callback(
          new TaskError(
            "Unsupported (Undocumented) command line option: " + notValidOptions
          )
        );

      const newOptions = Object.assign({}, clone, argv);

      result.command.run(newOptions, callback);
      analytics.send({
        command: result.name ? result.name : "other",
        args: result.argv._,
        version: bundled || "(unbundled) " + core
      });
    } catch (err) {
      callback(err);
    }
```
I don't think there is any reason we want to strictly not include line 132 from it.

packages/truffle-core/lib/command.js
```diff
@@ -100,6 +100,35 @@ class Command {
       }
     });
 
+    //Check unsupported command line flag according to the option list in help
+    try {
+      const inputOptions = [];
+      inputStrings.map(i => {
```
eggplantzzz: This is a really nitpicky style thing but how would you feel about being a little more verbose with this variable name...maybe name it `inputString`? I think it reads a little bit better even though it probably doesn't much matter.

packages/truffle-core/lib/command.js
```diff
@@ -100,6 +100,35 @@ class Command {
...
+        if (i.startsWith("--")) {
+          inputOptions.push(i);
+        }
+      });
+
+      const validOptions = [];
+      result.command.help.options.map(item => {
+        let opt = item.option.split(" ")[0];
```
eggplantzzz: Same with `opt` as with `i` above. Sorry again for nitpicking!

eggplantzzz:
Hey thanks @skywing! This is super cool.

packages/truffle-core/lib/command.js
Elaniobro: I agree with you @eggplantzzz  I like the one-liners without the `{}`. As for convention, I hear ya, I would say that as a rule of thumb, one should stick with the same style within a file :)

skywing: Thanks for the feedback. I will follow the style on file.

eggplantzzz:
@skywing Hmm, you might have to merge in the most recent work from the `develop` branch. There was a bug a couple of versions back where the install was failing due to an issue with `truffle obtain` which is run as part of the postinstall script. Looks like it is breaking the CI builds.

skywing:
Hi, I did a merge from develop branch to get all the latest. The build still failed. Any idea?

skywing:
I setup a Travis-CI build to test. I get the latest upstream develop branch into my fork and the build pass where it fails in the develop@unsupportedflag. I am looking into and fix the fail build.

PR #2119, opened 2019-06-20
Internal improvement: Convert `truffle-box` sandbox() from callback to promise

CruzMolina:
This PR makes the `Box.sandbox` method async and adds some code coverage. Tests using `sandbox()` updated where necessary. Also modernizes `lib/config.js` in `truffle-box`. 

**Breaking change to truffle-box**

### DIFF
--- a/packages/truffle-box/box.js
+++ b/packages/truffle-box/box.js
@@ -96,10 +96,6 @@ const Box = {
       options
     );
 
-    if (typeof options === "function") {
-      callback = options;
-    }
-
     if (setGracefulCleanup) {
       tmp.setGracefulCleanup();
     }
--- a/packages/truffle-box/box.js
+++ b/packages/truffle-box/box.js
@@ -97,7 +97,7 @@ const Box = {
   //   Recursively removes the created temporary directory, even when it's not empty. default is false
   // options.setGracefulCleanup
   //   Cleanup temporary files even when an uncaught exception occurs
-  sandbox: (options, callback) => {
+  sandbox: async options => {
     const { name, unsafeCleanup, setGracefulCleanup } = parseSandboxOptions(
       options
     );
@@ -111,19 +111,16 @@ const Box = {
     }
 
     const tmpDir = tmp.dirSync({ unsafeCleanup });
-    Box.unbox(
-      `https://github.com/trufflesuite/truffle-init-${name}`,
-      tmpDir.name,
-      options
-    )
-      .then(() => {
-        const config = Config.load(
-          path.join(tmpDir.name, "truffle-config.js"),
-          {}
-        );
-        callback(null, config);
-      })
-      .catch(callback);
+    try {
+      await Box.unbox(
+        `https://github.com/trufflesuite/truffle-init-${name}`,
+        tmpDir.name,
+        options
+      );
+      return Config.load(path.join(tmpDir.name, "truffle-config.js"), {});
+    } catch (error) {
+      throw error;
+    }
   }
 };
 
--- a/packages/truffle-box/box.js
+++ b/packages/truffle-box/box.js
@@ -7,13 +7,7 @@ const fse = require("fs-extra");
 const inquirer = require("inquirer");
 
 function parseSandboxOptions(options) {
-  if (typeof options === "function") {
-    return {
-      name: "default",
-      unsafeCleanup: false,
-      setGracefulCleanup: false
-    };
-  } else if (typeof options === "string") {
+  if (typeof options === "string") {
     // back compatibility for when `options` used to be `name`
     return {
       name: options,
--- a/packages/truffle-box/box.js
+++ b/packages/truffle-box/box.js
@@ -101,16 +101,12 @@ const Box = {
     }
 
     const tmpDir = tmp.dirSync({ unsafeCleanup });
-    try {
-      await Box.unbox(
-        `https://github.com/trufflesuite/truffle-init-${name}`,
-        tmpDir.name,
-        options
-      );
-      return Config.load(path.join(tmpDir.name, "truffle-config.js"), {});
-    } catch (error) {
-      throw error;
-    }
+    await Box.unbox(
+      `https://github.com/trufflesuite/truffle-init-${name}`,
+      tmpDir.name,
+      options
+    );
+    return Config.load(path.join(tmpDir.name, "truffle-config.js"), {});
   }
 };
 
--- a/packages/truffle-box/lib/config.js
+++ b/packages/truffle-box/lib/config.js
@@ -1,4 +1,4 @@
-const fs = require("fs-extra");
+const fse = require("fs-extra");
 
 function setDefaults(config = {}) {
   const hooks = config.hooks || {};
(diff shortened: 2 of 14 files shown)


packages/truffle-core/test/config.js
```diff
@@ -16,26 +16,21 @@ describe("config", function() {
     from: "0x1234567890123456789012345678901234567890"
   };
 
-  before("Create a sandbox with extra config values", function(done) {
+  before("Create a sandbox with extra config values", async () => {
     this.timeout(10000);
```
eggplantzzz: Does this need to be tacked on to the end because it is an arrow function?

packages/truffle-core/test/lib/commands/compile.js
```diff
@@ -14,26 +14,21 @@ describe("compile", function() {
...
-        secondary: {
-          network_id: "12345"
-        }
-      };
-      config.network = "default";
-      config.logger = { log: val => val && memStream.write(val) };
-      done();
-    });
+  before("Create a sandbox", async () => {
+    this.timeout(10000);
```
eggplantzzz: I think the same here...does this work like this in `async` functions?

packages/truffle-core/test/lib/create.js
```diff
@@ -11,15 +11,11 @@ var Artifactor = require("truffle-artifactor");
 describe("create", function() {
   var config;
 
-  before("Create a sandbox", function(done) {
+  before("Create a sandbox", async () => {
     this.timeout(10000);
```
eggplantzzz: same

packages/truffle-core/test/migrate.js
```diff
@@ -14,16 +14,12 @@ var Web3 = require("web3");
...
-    Box.sandbox("default", function(err, result) {
-      if (err) return done(err);
-      config = result;
-      config.resolver = new Resolver(config);
-      config.artifactor = new Artifactor(config.contracts_build_directory);
-      config.networks = {};
-      done();
-    });
+  before("Create a sandbox", async () => {
+    this.timeout(10000);
```
eggplantzzz: same

packages/truffle-core/test/npm.js
```diff
@@ -18,27 +18,23 @@ describe("NPM integration", function() {
...
-      config.resolver = new Resolver(config);
-      config.artifactor = new Artifactor(config.contracts_build_directory);
-      config.networks = {
-        development: {
-          network_id: 1
-        }
-      };
-      config.network = "development";
+  before("Create a sandbox", async () => {
+    this.timeout(10000);
```
eggplantzzz: same

packages/truffle-core/test/profiler.js
```diff
@@ -8,39 +8,38 @@ var Artifactor = require("truffle-artifactor");
 
 // TOOD: Move this to truffle-compile!
 
-describe('profiler', function() {
+describe("profiler", function() {
   var config;
 
-  before("Create a sandbox", function(done) {
+  before("Create a sandbox", async () => {
     this.timeout(10000);
```
eggplantzzz: same

packages/truffle/test/scenarios/cyclic_dependencies/compiles.js
```diff
@@ -12,21 +12,17 @@ describe("Cyclic Dependencies [ @standalone ]", function() {
   let options;
   const logger = new MemoryLogger();
 
-  before("set up sandbox", function(done) {
+  before("set up sandbox", async () => {
     this.timeout(10000);
```
eggplantzzz: same

packages/truffle/test/scenarios/happy_path/happypath.js
```diff
@@ -21,19 +21,15 @@ describe("Happy path (truffle unbox)", function() {
     Server.stop(done);
   });
 
-  before("set up sandbox", function(done) {
+  before("set up sandbox", async () => {
     this.timeout(10000);
```
eggplantzzz: same

packages/truffle/test/scenarios/solidity_testing/solidityTests.js
```diff
@@ -48,9 +43,9 @@ describe("Solidity Tests", function() {
   after(done => Server.stop(done));
 
   describe("test with balance", function() {
-    before(function(done) {
+    before(async () => {
       this.timeout(5000);
```
eggplantzzz: same

packages/truffle/test/scenarios/solidity_testing/solidityTests.js
```diff
@@ -66,9 +61,9 @@ describe("Solidity Tests", function() {
   });
 
   describe("tests failing", function() {
-    before(function(done) {
+    before(async () => {
       this.timeout(5000);
```
eggplantzzz: same

packages/truffle-box/box.js
```diff
@@ -97,33 +91,26 @@ const Box = {
...
-      })
-      .catch(callback);
+    try {
+      await Box.unbox(
+        `https://github.com/trufflesuite/truffle-init-${name}`,
+        tmpDir.name,
+        options
+      );
+      return Config.load(path.join(tmpDir.name, "truffle-config.js"), {});
+    } catch (error) {
```
eggplantzzz: In this case we don't need to catch this as all it does is throw.

packages/truffle-box/lib/config.js
```diff
@@ -19,14 +17,14 @@ function setDefaults(config) {
 }
 
 function read(path) {
-  return fs
+  return fse
     .readFile(path)
     .catch(() => "{}")
```
eggplantzzz: I still think this is a weird return :P

packages/truffle-box/test/sandbox.js
```diff
@@ -0,0 +1,24 @@
...
+  let options;
+
+  it("successfully unboxes and returns TruffleConfig when passed a valid options object", async () => {
+    options = { name: "default", force: true };
+    try {
+      const config = await sandbox(options);
+      assert(config);
+      assert(config.truffle_directory);
+    } catch (error) {
+      throw error;
```
eggplantzzz: Why do we need the `try/catch`?

CruzMolina: Thanks for reviewing and being so thorough!

packages/truffle-box/box.js
```diff
@@ -7,13 +7,7 @@ const fse = require("fs-extra");
...
 
 function parseSandboxOptions(options) {
-  if (typeof options === "function") {
-    return {
-      name: "default",
-      unsafeCleanup: false,
-      setGracefulCleanup: false
-    };
-  } else if (typeof options === "string") {
+  if (typeof options === "string") {
```
eggplantzzz: So are there places in the code where `typeof options === "function"`? That would mean that `sandbox` would be called with the callback being the first argument I guess? We should make sure that doesn't happen if we remove the check.

eggplantzzz: I guess it looks like you got them all :)

eggplantzzz:
It seems good.

PR #2120, opened 2019-06-21
Fix bugs relating to constant mapping keys

haltman-at:
So, it looks like when I implemented code to allow constant state variables to work as mapping keys, I must never have actually tested it very much, because due to a typo, they actually didn't work as mapping keys much at all.

Of course, that typo masked an infinite loop!  I forgot to set `indexValue = null` if we failed to decode a constant state variable being used as a mapping key, meaning (had the typo not prevented the case from ever coming up) any constant state variable we couldn't decode would throw the debugger into an infinite loop.

(Why did I forget that?  Because this is the only case with another nested `if` inside it, meaning it's the only case where it's necessary; other cases that fail will just get caught by the `else` at the end, but these ones won't and need their own `else`s.)

Anyway, I fixed both the typo and the loop it masked.  So now constant state variables can actually work as mapping keys (well, in more cases than previously, not in general), and when they don't, they won't kill the debugger.

### DIFF
--- a/packages/truffle-debugger/lib/data/sagas/index.js
+++ b/packages/truffle-debugger/lib/data/sagas/index.js
@@ -403,7 +403,7 @@ function* variablesAndMappingsSaga() {
           indexValue = yield* decode(splicedDefinition, indexReference);
         } else if (
           indexDefinition.referencedDeclaration &&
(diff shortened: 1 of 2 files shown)


haltman-at:
Good idea, I'll add a quick test before merging this.

PR #2123, opened 2019-06-22
fix(truffle-core/cli): logs now to STDERR

yxliang01:
I believe all logs are supposed to be outputted to `STDERR`, instead of `STDOUT`.

### DIFF
--- a/packages/truffle-core/cli.js
+++ b/packages/truffle-core/cli.js
@@ -38,7 +38,13 @@ const command = new Command(require("./lib/commands"));
 const listeners = process.listeners("warning");
 listeners.forEach(listener => process.removeListener("warning", listener));
 
-let options = { logger: console };
+let options = { 
+  logger: {
+    log: function(message) {
+      console.error(message);
+    }
+  }
+};
 
 const inputArguments = process.argv.slice(2);
 const userWantsGeneralHelp =


gnidan:
I don't think that's true @yxliang01. Certainly, some logs (errors) should be going to stderr, but regular output? Do you have any resources that talk about best practices here?

yxliang01:
@gnidan Regular output I think should be to STDOUT. However, by "regular output", I think it means the mocha output and it's not affected by the modification in this PR. See https://stackoverflow.com/a/3385261/3913932 , it has the prevalent usage description for each standard stream. STDERR normally includes not only the error information but also the debugging or more generally the "extra information", in my opinion. 

Actually, my use case is that I want truffle to output parsable json output for my toolchain to make use of the truffle test result. But, even if I change the mocha reporter to json, it can't be parsed due to the extra outputs like the compiling messages in `STDOUT`.

I think there should be some highly regarded resources pointing out this, but I don't have one at this moment. Maybe later I can find one if you wish.

gnidan:
Ah @yxliang01 thanks for the explanation. I see what you mean; you are correct. But I don't think this PR is sufficient. Truffle's logging right now is... rather simplistic.

There are two issues, as I see it:
- I'm not sure changing this here will get the behavior you want everywhere. This might not be the only place where stdout is configured. I would say it's very likely that you're still going to end up with unwanted stdout.
- More importantly, there may be projects that depend on the stdout behavior of Truffle. This PR runs the risk of breaking other people's work, since it's entirely possible that projects have already compensated for the problem you describe, and in fact it's rather likely that projects are set up to look through stdout for specific output.

As a result, I think this is a breaking change that will have to wait for the next major release.

A side note: there's a WIP effort to improve Truffle's logging infrastructure by way of building an event system (#2035). The plan there is to get rid of all `config.logger.log` statements inside the codebase, instead emitting events that could be handled in a central "Reporter" class. Once this work finishes, you will be able to configure custom reporters to get whatever output behavior you want, and we can revisit the stdout/stderr question in time for Truffle v6, when breaking changes are available to us.

Could you please open this as an issue instead? I don't want to lose track of this.

Thanks!

fainashalts:
Closing this for issue maintenance. @yxliang01 if you're still interested in this please make it an issue so we can discuss. Thanks for bringing this to our attention!

PR #2125, opened 2019-06-23
Warn/error about unsupported flags #468

skywing:
I opened a pull request #2116 and Travis CI build failed. I was thinking not having the latest from develop branch and the git force command cause the issue close while using it on my branch so I have to open a new pull request.

After investigation I found out that the newly implementation for unsupported option validation broke the build. The truffle command that trying to obtain the solc has a bit different help documentation due to the nature of the dynamic option.

I can update the truffle obtain command help so the CI build won't break but this is not a good solution because any future command document with a bit different format will potential break the build or prevent user to run the command.

I think the better approach is to warn the user about the potential problem when an unsupported options are detected so I change it from return TaskError to just use the options logger to warn the user.

### DIFF
--- a/packages/truffle-core/lib/command.js
+++ b/packages/truffle-core/lib/command.js
@@ -117,19 +117,18 @@ class Command {
         }
       });
 
-      let notValidOptions = inputOptions.filter(
+      let invalidOptions = inputOptions.filter(
         opt => !validOptions.includes(opt)
       );
 
-      if (notValidOptions.length > 0) {
-        console.log(
-          "Unsupported (Undocumented) command line option: " + notValidOptions
-        );
-        // return callback(
-        //   new TaskError(
-        //     "Unsupported (Undocumented) command line option: " + notValidOptions
-        //   )
-        // );
+      if (invalidOptions.length > 0) {
+        if (options.logger) {
+          const warn = options.logger.warn || options.logger.log;
+          warn(
+            "warning: possilble unsupported (undocumented in help) command line option: " +
+              invalidOptions
+          );
+        }
       }
 
       const newOptions = Object.assign({}, clone, argv);
--- a/packages/truffle-core/lib/command.js
+++ b/packages/truffle-core/lib/command.js
@@ -102,20 +102,22 @@ class Command {
 
     //Check unsupported command line flag according to the option list in help
     try {
-      const inputOptions = [];
-      inputStrings.map(i => {
-        if (i.startsWith("--")) {
-          inputOptions.push(i);
-        }
-      });
+      const inputOptions = inputStrings
+        .map(string => {
+          return string.startsWith("--") ? string : null;
+        })
+        .filter(item => {
+          return item != null;
+        });
 
-      const validOptions = [];
-      result.command.help.options.map(item => {
-        let opt = item.option.split(" ")[0];
-        if (opt.startsWith("--")) {
-          validOptions.push(opt);
-        }
-      });
+      const validOptions = result.command.help.options
+        .map(item => {
+          let opt = item.option.split(" ")[0];
+          return opt.startsWith("--") ? opt : null;
+        })
+        .filter(item => {
+          return item != null;
+        });
 
       let invalidOptions = inputOptions.filter(
         opt => !validOptions.includes(opt)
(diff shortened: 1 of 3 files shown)


eggplantzzz:
Actually it might not be a bad idea to just update the help in the `obtain` command. We could have an option for each compiler type. For example, change the `<--<compiler_name> <version>>` to `--solc <version>`. Let me talk to some others from the team and see what they think is the best solution.

eggplantzzz:
So I talked with @gnidan and I think it is ok that we create an option for each compiler in the obtain command file. Even though we'll have to update it with every new compiler I think that should be fine. I don't really expect we'll support 100 new compilers any time soon :)

packages/truffle-core/lib/command.js
```diff
@@ -100,9 +100,39 @@ class Command {
...
+
+      let invalidOptions = inputOptions.filter(
+        opt => !validOptions.includes(opt)
+      );
+
+      if (invalidOptions.length > 0) {
+        if (options.logger) {
+          const log = options.logger.log || options.logger.debug;
+          log(
+            "warning: possilble unsupported (undocumented in help) command line option: " +
```
eggplantzzz: Typo here, "possible unsupported". Also, for consistent formatting with other logging in Truffle, could we make this message something like "> Warning: possible...."?

eggplantzzz:
So @skywing, so I would say there are two more things to take care of before we can get this merged:
  1. Fix the typo/formatting for the error
  2. Edit the help for the obtain command so that `--solc` has its own entry (i.e. make it `--solc <version>` instead of the more general description)
Thanks again for doing this work!

skywing:
Thank you!!! I will work on the update based on the recommendation.

eggplantzzz:
Hmm I wonder why coverage dropped so much after that commit...

packages/truffle-core/lib/commands/obtain.js
```diff
@@ -10,6 +10,11 @@ module.exports = {
           `Download and cache a version of the specified compiler.\n` +
           `                    compiler_name must be one of the following: ` +
           `'solc'.(required)`
+      },
+      {
```
eggplantzzz: Instead of adding a new option, edit the existing one above. Change `<--<compiler_name> <version>>` to `--solc <version>`. Or just delete the first option and leave the one you added.

skywing: I think I might find the problem of low code coverage. I am testing the build to see the coverage now.

skywing:
> Hmm I wonder why coverage dropped so much after that commit...

Hi, the problem is fix and the code coverage is back to green.

packages/truffle-core/lib/command.js
```diff
@@ -100,9 +100,39 @@ class Command {
       }
     });
 
-    const newOptions = Object.assign({}, clone, argv);
-
+    //Check unsupported command line flag according to the option list in help
     try {
+      const inputOptions = [];
+      inputStrings.map(i => {
```
eggplantzzz: Instead of creating an array first and then using the `push` method, just set `inputStrings` like the following:
```
const inputOptions = inputStrings.map(string => {
  return string.startsWith("--") ? string : null;
}).filter(item => item);
```
`map` already returns an array but you also need to filter out the `null`s afterwards.

packages/truffle-core/lib/command.js
```diff
@@ -100,9 +100,39 @@ class Command {
...
+        if (i.startsWith("--")) {
+          inputOptions.push(i);
+        }
+      });
+
+      const validOptions = [];
```
eggplantzzz: Do something similar here.

skywing: Implemented based on feedback. Thanks.

eggplantzzz:
Ok @skywing, one last thing and then I think we can be ready to merge this. In the obtain command file, there should only be one `options` property. Remove the `<--<compiler_name> <version>>` option. Also, add a "(required)" to it as it is a required option right now. See the following line of code for an example from `truffle exec`. 
https://github.com/skywing/truffle/blob/develop%2Bunsupportedflag3/packages/truffle-core/lib/commands/exec.js#L24

If you can make this last edit, we'll get it merged. Thanks again!!

skywing:
Hi, can we trigger the build again. I have a separate travis-ci build and that one didn't failed.

eggplantzzz:
Looks good!

PR #2131, opened 2019-06-26
Fix a regression in decoding of dynamic structs!

haltman-at:
It seems I accidentally broke decoding of dynamic structs in calldata a while back.  Probably when I factored out struct decoding; in combining the static and dynamic cases, I left out the `startPosition` argument in the call to `decodeAbi`, which is fine in the static case but is very wrong in the dynamic case.  Oops.

(In the static case, you're always reading the member directly, not a pointer to it, so you don't need to know what base that pointer should be considered relative to, because it's not a pointer at all.  In the dynamic case, you may be reading a relative pointer, so you need to know what it's relative to.)

As a check against this happening again, I added an additional (very simple) calldata decoding test to catch this.

Note that in order to make this work I ended up updating the Solidity version used in our debugger tests to 0.5.10.  (It was on 0.5.4; dynamic structs in calldata weren't allowed yet in 0.5.4.)

### DIFF
--- a/packages/truffle-debugger/test/data/calldata.js
+++ b/packages/truffle-debugger/test/data/calldata.js
@@ -13,7 +13,7 @@ import * as TruffleDecodeUtils from "truffle-decode-utils";
 import solidity from "lib/solidity/selectors";
 
 const __CALLDATA = `
-pragma solidity ^0.5.4;
+pragma solidity ^0.5.6;
 pragma experimental ABIEncoderV2;
 
 contract CalldataTest {
@@ -23,6 +23,10 @@ contract CalldataTest {
   struct Pair {
     uint x;
     uint y;
+  }
+
+  struct StringBox {
+    string it;
   }
 
   function simpleTest(string calldata hello) external {
@@ -59,6 +63,10 @@ contract CalldataTest {
     pair.x = 321;
     pair.y = 2049;
     this.multiTest("hello", someInts, pair);
+  }
+
+  function stringBoxTest(StringBox calldata stringBox) external returns (string memory) {
+    return stringBox.it; //break stringBox
   }
 
 }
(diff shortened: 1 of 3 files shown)


packages/truffle-debugger/test/data/calldata.js
```diff
@@ -180,6 +188,40 @@ describe("Calldata Decoding", function() {
...
+    let bugger = await Debugger.forTx(txHash, {
+      provider,
+      files,
+      contracts: artifacts
+    });
+
+    let session = bugger.connect();
+
+    let sourceId = session.view(solidity.current.source).id;
+    let source = session.view(solidity.current.source).source;
```
gnidan: Can skip the doubles here and do:
```javascript
let { source, id: sourceId } = session.view(solidity.current.source)
```

gnidan: (also it looks like all the `let`s can just be `const`s)

packages/truffle-decoder/lib/decode/calldata.ts
```diff
@@ -177,7 +177,8 @@ function* decodeCalldataStructByPosition(definition: DecodeUtils.AstDefinition,
     //there also used to be code here to add on the "_ptr" ending when absent, but we
     //presently ignore that ending, so we'll skip that
 
-    let decoded = yield* decodeCalldata(memberDefinition, childPointer, info);
+    let decoded = yield* decodeCalldata(memberDefinition, childPointer, info, startPosition);
```
gnidan: Oh so this argument just got lost somewhere, hence the regression.

packages/truffle-debugger/test/data/calldata.js
haltman-at: Yeah, that's a bunch of copypaste that I think I'm going to skip changing for now.  I can redo these more generally later.  But thanks for pointing it out.

packages/truffle-decoder/lib/decode/calldata.ts
haltman-at: Right -- I didn't bother including it in the static case (because it's meaningless there), and I accidentally used the static code rather than the dynamic code when I combined them.

PR #2142, opened 2019-06-28
Internal improvement: Profiler update

eggplantzzz:
This PR reorganizes the `updated` method for the `Profiler` in truffle-compile. All uses of the `async` module are eliminated. Also much of the logic has been extracted into helper files.

Another change (not truffle-compile related) that this makes is that it allows the main method exposed by truffle-contract-sources to have a Promise interface as well as its old callback one.

### DIFF
--- a/packages/truffle-compile/package.json
+++ b/packages/truffle-compile/package.json
@@ -7,6 +7,7 @@
     "async": "2.6.1",
     "colors": "^1.1.2",
     "debug": "^4.1.0",
+    "fs-extra": "^8.0.1",
     "ora": "^3.0.0",
     "original-require": "^1.0.1",
     "request": "^2.85.0",
--- a/packages/truffle-compile/package.json
+++ b/packages/truffle-compile/package.json
@@ -4,7 +4,6 @@
   "description": "Compiler helper and artifact manager",
   "main": "index.js",
   "dependencies": {
-    "async": "2.6.1",
     "colors": "^1.1.2",
     "debug": "^4.1.0",
     "fs-extra": "^8.0.1",
--- a/packages/truffle-compile/profiler/findUpdatedFiles.js
+++ b/packages/truffle-compile/profiler/findUpdatedFiles.js
@@ -1,0 +1,38 @@
+const fse = require("fs-extra");
+
+module.exports = (sourceFilesArtifacts, sourceFilesArtifactsUpdatedTimes) => {
+  let updatedFiles = [];
+  // Stat all the source files, getting there updated times, and comparing them to
+  // the artifact updated times.
+  const sourceFiles = Object.keys(sourceFilesArtifacts);
+
+  let sourceFileStats;
+  sourceFileStats = sourceFiles.map(file => {
+    try {
+      return fse.statSync(file);
+    } catch (error) {
+      // Ignore it. This means the source file was removed
+      // but the artifact file possibly exists. Return null
+      // to signfy that we should ignore it.
+      return null;
+    }
+  });
+
+  sourceFiles.forEach((sourceFile, index) => {
+    const sourceFileStat = sourceFileStats[index];
+
+    // Ignore updating artifacts if source file has been removed.
+    if (sourceFileStat == null) return;
+
+    const artifactsUpdatedTime =
+      sourceFilesArtifactsUpdatedTimes[sourceFile] || 0;
+    const sourceFileUpdatedTime = (
+      sourceFileStat.mtime || sourceFileStat.ctime
+    ).getTime();
+
+    if (sourceFileUpdatedTime > artifactsUpdatedTime) {
+      updatedFiles.push(sourceFile);
+    }
+  });
+  return updatedFiles;
+};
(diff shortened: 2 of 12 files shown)


packages/truffle-compile/profiler/index.js
```diff
@@ -0,0 +1,266 @@
...
+const { readAndParseArtifactFiles } = require("./readAndParseArtifactFiles");
+const {
+  minimumUpdatedTimePerSource
+} = require("./minimumUpdatedTimePerSource");
+const { findUpdatedFiles } = require("./findUpdatedFiles");
+const { isExplicitlyRelative } = require("./isExplicitlyRelative");
+const { getImports } = require("./getImports");
+
+module.exports = {
+  updated(options, callback) {
```
gnidan: What do you think of making this callback-style function be a wrapper around an async function? To get rid of the `.then()`s (and same with `required_sources`)

packages/truffle-compile/profiler/findUpdatedFiles.js
```diff
@@ -0,0 +1,45 @@
...
+      return fse.statSync(file);
+    } catch (error) {
+      // Ignore it. This means the source file was removed
+      // but the artifact file possibly exists. Return null
+      // to signfy that we should ignore it.
+      return null;
+    }
+  });
+
+  sourceFiles.forEach((sourceFile, index) => {
```
gnidan: Can we turn this into a `.map` so we don't have to `updateFiles.push`?

packages/truffle-compile/profiler/getImports.js
```diff
@@ -0,0 +1,21 @@
+const { isExplicitlyRelative } = require("./isExplicitlyRelative");
+const Parser = require("../parser");
+const path = require("path");
+
+const getImports = (file, { body, source }, solc) => {
+  // No imports in vyper!
+  if (path.extname(file) === ".vy") return [];
```
gnidan: heh it's kinda silly that this is here... but vyper _is_ using the profiler, I know that.

packages/truffle-compile/profiler/index.js
```diff
@@ -0,0 +1,266 @@
...
+        return options.files;
+      } else {
+        return findContracts(contracts_directory);
+      }
+    }
+
+    let sourceFilesArtifacts = {};
+    let sourceFilesArtifactsUpdatedTimes = {};
+
+    let updatedFiles = [];
```
gnidan: We can move this declaration into the function below and make it `const`

packages/truffle-contract-sources/index.js
```diff
@@ -1,11 +1,13 @@
 const debug = require("debug")("contract-sources");
-
 const path = require("path");
 const glob = require("glob");
+const { promisify } = require("util");
+const promisifiedGlob = promisify(glob);
```
gnidan: Just skip this nonsense and do:
```javascript
const { promisify } = require("util");
const glob = promisify(require("glob"));
```

packages/truffle-contract-sources/index.js
```diff
@@ -16,5 +18,9 @@ module.exports = (pattern, callback) => {
     follow: true // follow symlinks
   };
 
-  glob(pattern, globOptions, callback);
+  if (callbackPassed) {
+    glob(pattern, globOptions, callback);
+  } else {
+    return promisifiedGlob(pattern, globOptions);
+  }
```
gnidan: And here, use the promisified version for both, and do the callback string-up with a `.then`

eggplantzzz: So like the following?
```
return promisify(glob(pattern, globOptions)
  .then(files => {
    callback(null, files);
  })
  .catch(error => {
    callback(error);
  });
} else {
  return promisify(glob(patter, globOptions);
}
```
Is that what you mean?

packages/truffle-compile/package.json
```diff
@@ -7,6 +7,7 @@
     "async": "2.6.1",
```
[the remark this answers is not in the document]
eggplantzzz: I'd like to but it is still being used by `required_sources`. I'd like to get rid of it but I haven't quite figured out how to rewrite that yet. I will give it another try today.

packages/truffle-compile/profiler/index.js
eggplantzzz: Ultimately I would just like to make it a straight-up `async` function but I guess we should have a conversation about breaking changes and Truffle in general. Otherwise I suppose making it wrap an `async` function seems like a fine idea.

packages/truffle-compile/profiler/findUpdatedFiles.js
eggplantzzz: Oh yeah you bet we can!

packages/truffle-contract-sources/index.js
eggplantzzz: Oh I see, you mean remove the if and just do...
```
return promisify(glob(patther, globOptions))
  .then(files => {
    if (callbackPassed) callback(null, files);
    return files;
  .catch(error => {
    if (callbackPassed) callback(error);
    throw error;
  });
```
Yeah?

gnidan: Either one!

packages/truffle-compile/profiler/index.js
gnidan: Up to you if you want the dual interface. I was just saying the wrapper is sufficient and will help make the code nicer to look at.

PR #2149, opened 2019-07-02
New Feature: Hyperledger Integration (chaincode-fabric-evm)

CruzMolina:
This PR builds on top of prior work in #1806 to implement `chaincode-fabric-evm` integration.

⛓ 

Example `truffle-config.js` for using Truffle with `chaincode-fabric-evm`:
```
module.exports = {
  networks: {
    development: {
      ...
      type: "fabric-evm"
      ...
    }
  }
};
```


packages/truffle/test/scenarios/migrations/fabric-evm.js
```diff
@@ -0,0 +1,69 @@
...
+    };
+
+    const provider = new Web3.providers.HttpProvider("http://localhost:8545", {
+      keepAlive: false
+    });
+    web3 = new Web3(provider);
+    networkId = await web3.eth.net.getId();
+  });
+
+  it("runs migrations (sync & async/await)", function(done) {
```
gnidan: I think we should keep all new tests written using `async function() {}` instead of `function(done) {}`

packages/truffle-interface-adapter/test/fabric-evm-getId.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        assert(typeof(networkID) === "number")
+        preparedGanache.server.close(resolve);
+      } catch (e) {
+        preparedGanache.server.close(() => {
+          reject(e);
+        });
+      }
+    });
+  });
+});
```
gnidan: (new line at end of file)

packages/truffle-interface-adapter/lib/index.ts
```diff
@@ -1 +1,4 @@
+const supportedNetworks = ["quorum", "fabric-evm"];
```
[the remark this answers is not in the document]
gnidan: To follow-up: if you're into the ideas I propose in the other comments, it'd be natural to just make this dynamic now.

packages/truffle-interface-adapter/lib/fabric-evm-overloads.ts
```diff
@@ -0,0 +1,17 @@
...
+  // @ts-ignore
+  const _oldGetIdFormatter = web3.eth.net.getId.method.outputFormatter;
+  // @ts-ignore
+  web3.eth.net.getId.method.outputFormatter = (hexId: any) => {
+    // chaincode-fabric-evm currently returns a "fabric-evm" string
+    // instead of a hex networkID. Instead of trying to decode the hexToNumber,
+    // let's just accept `fabric-evm` as a valid networkID for now.
+    return hexId;
+  };
+};
```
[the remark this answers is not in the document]
CruzMolina: Looks like this was happening from not having the nvim typescript plugin 😢

packages/truffle/test/sources/migrations/fabric-evm/migrations/2_migrations_sync.js
```diff
@@ -0,0 +1,8 @@
+const Example = artifacts.require("Example");
+const UsesExample = artifacts.require("UsesExample");
+
+module.exports = function(deployer) {
+  deployer
+    .deploy(Example)
+    .then(() => deployer.deploy(UsesExample, Example.address));
```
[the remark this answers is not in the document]
CruzMolina: `fabric-evm` currently uses the legacy migrations system, so I think we'll need this for now 😞

packages/truffle-deployer/index.js
```diff
@@ -42,8 +43,14 @@ class Deployer extends Deployment {
...
     const args = Array.prototype.slice.call(arguments);
     const contract = args.shift();
+    const networkType = this.networks[this.network].type;
 
-    if (this.networks[this.network].type === "quorum") {
+    if (
+      networkType &&
+      getSupportedNetworks().then(supportedNetworks =>
+        supportedNetworks.includes(networkType)
+      )
```
gnidan: This syntax is wrong: the `getSupportedNetworks().then(/*...*/)` bit will _always_ resolve to a truthy value

gnidan: Best make `getSupportedNetworks` synchronous, I think

packages/truffle-interface-adapter/lib/ethereum-overloads.ts
```diff
@@ -1,11 +1,21 @@
 import BN from "bn.js";
-import Web3 from "web3";
+import { Web3Shim } from "./web3-shim";
+
+export const EthereumDefinition = {
+  async initNetworkType (web3: Web3Shim) {
+    // truffle has started expecting gas used/limit to be
+    // hex strings to support bignumbers for other ledgers
+    getBlock(web3);
```
gnidan: Naming note here, not to change for this PR, but we might want to fix at some point:

For explicitness purposes, these methods should probably include the word `override` or something similar, to convey that it's not that we're getting the block, but we're overriding how we get the block.

An alternative solution would be to abstract this, so that `getBlock`, `getTransaction`, would be in a mapping, something like:
```typescript
const overrides = {
  "getBlock": (original) => {
    const _oldFormatter = original.method.outputFormatter;
    original.method.outputFormatter = (block) => {
      _oldFormatter.call(original, block);
      /* ... */
    };
  },
  /* ... */
}
```

Just if you want to keep this in the back of your mind for whenever.

packages/truffle-interface-adapter/lib/index.ts
```diff
@@ -1 +1,5 @@
+export function getSupportedNetworks() {
+  return ["quorum", "fabric-evm"];
+};
```
gnidan: Not blocking approval, but I think we should fix the name here to reflect that this specifically relates to which network types use the legacy system. Maybe `getLegacyNetworkTypes`.

Ooh, we could take this a step further! We can add `requiresLegacySystem` as a boolean field on the Definition objects, and that way it becomes a lot more cohesive when adding a new network. Basically, just these steps:
1. Create an overrides file + definition
2. Import the definition and link it up in the web3-shim module

packages/truffle-interface-adapter/lib/quorum-overloads.ts
```diff
@@ -1,13 +1,23 @@
 import BN from "bn.js";
-import Web3 from "web3";
+import { Web3Shim } from "./web3-shim";
 import {AbiCoder as EthersAbi} from 'ethers/utils/abi-coder';
 import _ from "underscore";
 
+export const QuorumDefinition = {
+  async initNetworkType (web3: Web3Shim) {
+    // duck punch some of web3's output formatters
```
gnidan: 🦆 😢

packages/truffle-interface-adapter/lib/web3-shim.ts
```diff
@@ -1,22 +1,31 @@
...
 import { Provider } from "web3/providers";
 
-import * as ethereumOverloads from "./ethereum-overloads";
-import * as quorumOverloads from "./quorum-overloads";
+import { EthereumDefinition } from "./ethereum-overloads";
+import { QuorumDefinition } from "./quorum-overloads";
+import { FabricEvmDefinition } from "./fabric-evm-overloads";
 
 // March 13, 2019 - Mike Seese:
 // This is a temporary shim to support the basic, Ethereum-based
```
gnidan: ha! ha-HA! yeah right

packages/truffle-interface-adapter/lib/web3-shim.ts
```diff
@@ -56,33 +65,13 @@ export class Web3Shim extends Web3 {
...
-    ethereumOverloads.getBlock(this);
-    ethereumOverloads.getTransaction(this);
-    ethereumOverloads.getTransactionReceipt(this);
-  }
+  async initInterface() {
+    const networkTypes: NetworkTypesConfig = new Map(Object.entries({
+      "ethereum": EthereumDefinition,
+      "quorum": QuorumDefinition,
+      "fabric-evm": FabricEvmDefinition
+    }));
```
gnidan: can we move this block to _immediately_ follow the imports, as a module-scoped constant? (Same reasoning as the other comment: make the code surface area as small as possible when adding a new network type)

packages/truffle-interface-adapter/lib/quorum-overloads.ts
CruzMolina: 👊

PR #2150, opened 2019-07-02
Load contract instances from artifacts into Truffle DB

fainashalts:
The purpose of this PR is to add contract instance and network resources to truffle DB's 
schema, queries/mutations, and loader (the loader is used to save artifact information in truffle DB's in-memory data store). For information about the Resources and other data objects being added to Truffle DB, please see the proposed data model for Truffle DB here: https://trufflesuite.github.io/artifact-updates/data-model.html

Via the loader, a new `Contract Instance` is added to Truffle DB for every `Network` in a project's built contracts (assuming that network is accessible via web3 and in the project's configuration file). Contract Instances represent contracts that have been deployed to a network. Thus, networks and contract instances are determined to exist and added per contract, as found via artifacts. 

As a part of this effort, `networksAdd`, `network`, `contractInstancesAdd`, and `contractInstance` functions have been added to `pouch.ts`. A `networksAdd` and a `contractInstancesAdd` mutation have been added to the workspace `schema.ts` file. These changes to the workspace code have then been leveraged to load network and contract instance information from the artifacts files to workspace via the loader code, using the new `loadNetworks` and `loadContractInstances` functions.


fty48jj19:
nao tem com fazer outra conta nao quero modifica nada desculpa

packages/truffle-db/src/loaders/artifacts/index.ts
```diff
@@ -287,21 +381,98 @@ export class ArtifactsLoader {
...
 
 
     const compilations = await this.db.query(AddCompilation, { compilations: compilationObjects });
 
-    //map contracts to the compilation they belong in before adding them to truffle-db
-    await Promise.all(compilations.data.workspace.compilationsAdd.compilations
-      .map(({ compiler, id }) => this.loadCompilationContracts(contracts[compiler.name], id)));
+    //map contracts and contract instances to compiler
+    await Promise.all(compilations.data.workspace.compilationsAdd.compilations.map(async ({compiler, id}) => {
+      const contractIds = await this.loadCompilationContracts(contracts[compiler.name], id, compiler.name);
```
fainashalts: @gnidan I remember you mentioned that you thought I should pull this part out into its own function, but it's using from this loadCompilations function so I wanted to double check what you mean here. Thanks!

packages/truffle-db/src/loaders/index.ts
```diff
@@ -29,13 +29,14 @@ const resolvers = {
...
           contracts_build_directory: tempDir.name,
+          artifacts_directory: artifactsDirectory,
           all: true
         }
         const loader = new ArtifactsLoader(db, compilationConfig);
         await loader.load()
         tempDir.removeCallback();
         return true;
-      } 
+      }
```
gnidan: Editor settings, nice :)

packages/truffle-db/src/loaders/test/index.ts
```diff
@@ -3,6 +3,7 @@ import path from "path";
 import gql from "graphql-tag";
 import { TruffleDB } from "truffle-db";
 import * as Contracts from "truffle-workflow-compile";
+import * as Ganache from "ganache-core"
```
gnidan: I don't think this is used in this file?

packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -24,6 +24,14 @@ const resources = {
...
     createIndexes: [
     ]
+  },
+  networks: {
+    createIndexes: [
+    ]
+  },
+  contractInstances: {
+    createIndexes: [
+    ]
```
gnidan: This will likely need an index on `network.id`

packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -162,7 +172,94 @@ export class Workspace {
...
+    const { contractInstances } = input;
+
+    return {
+      contractInstances: Promise.all(contractInstances.map(
+        async (contractInstanceInput) => {
+          const { address, network, contract, callBytecode } = contractInstanceInput;
+          // hash includes address and network of this contractInstance
+          const id = soliditySha3(jsonStableStringify({
+            address: address,
+            network: network
```
gnidan: Since we only want the `network.id` portion of this to be used, it's probably worth being explicit about this here:

```suggestion
            network: { id: network.id }
```

packages/truffle-db/src/artifacts/schema.ts
```diff
@@ -73,6 +75,20 @@ export const schema = mergeSchemas({
...
+      network: {
+        fragment: `... on ContractObject {
+          networks {
+            networkId
+          }
+        }`,
+        resolve: (obj) => {
+          const { networks } = obj;
+          const result = networks
+            .map( ({ networkId }) => ({ networkID: networkId }))[0]
```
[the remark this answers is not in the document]
fainashalts: `networkID` is how it is in `schema.graphql`, I think to indicate that it is the `NetworkID` type? Happy to change everything to `networkId` though.

packages/truffle-db/src/loaders/artifacts/index.ts
```diff
@@ -182,16 +188,100 @@ const AddContracts = gql`
...
+const AddNetworks = gql`
+  input HistoricBlockInput {
+    height: Int!
+    hash: String!
+  }
+  input NetworkInput {
+    name: String
+    networkID: NetworkID!
+    historicBlock: HistoricBlockInput!
+    fork: NetworkInput
```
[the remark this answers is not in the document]
fainashalts: That makes sense. Will change it.

packages/truffle-db/src/loaders/test/index.ts
fainashalts: Oops, you're right! Will remove.

packages/truffle-db/src/workspace/pouch.ts
fainashalts: For `fork`, right? I'll add it. But do you mean `network.id`? Or `network.networkId`?

packages/truffle-db/src/artifacts/schema.ts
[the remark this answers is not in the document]
gnidan: I probably made it `networkID` before I realized the convention was the other way. Let's stick with `networkId`

packages/truffle-db/src/workspace/pouch.ts
gnidan: Yep, `network.id`! We will want to be find contract instances by a specific network object.

packages/truffle-db/src/loaders/artifacts/index.ts
fainashalts: I made some changes here in line with this pseudo code. I can break it down some more but I don't want to have to map an extra time if it can be avoided. Let me know if this is more in line with what you imagined or if you'd like additional changes!

fainashalts:
Woohoo!!

Issue #1658, opened 2019-01-23
Audit Truffle's existing features/APIs for compatibility with Quorum


benjamincburns:
Raise new issues for gaps found and assign them to epic #1657, please.

gnidan:
Not stale.

Issue #1657, opened 2019-01-23
Quorum Support Epic


benjamincburns:
We seem to have left Quorum out of our design and test for Truffle v5. Let's use this issue as an Epic to account for the work necessary to make Truffle work with Quorum again. 

To be clear, the goal of this issue is full support in Truffle for Quorum. This may include adopting/utilizing/integrating features of Quorum which weren't adequately integrated in prior Truffle versions. As always our goal is to make support as out of the box as possible, but for the sake of pragmatism minimal user-initiated configuration steps, installation of extra dependencies, or other similar tasks are fine, provided they are straightforward. That said, any user-initiated steps required for full Quorum support should be addressed in our user docs on our website before this issue should be considered resolved.

Issues in this epic (for non-zenhub users):

- [x] [Accommodate broken support for Quorum/other chains via `--legacy` flags to use old code](https://github.com/trufflesuite/truffle/issues/1543) - **Available in [v5.0.9](https://github.com/trufflesuite/truffle/releases/tag/v5.0.9)**
- [x] [`truffle migrate` on private permissioned Quorum network returns nothing to console](https://github.com/trufflesuite/truffle/issues/1622) - **Available in [v5.0.9](https://github.com/trufflesuite/truffle/releases/tag/v5.0.9)**
- [x] [Add support for Quorum's privateFor transaction parameter](https://github.com/trufflesuite/truffle/issues/1656) - **Available in [v5.0.14](https://github.com/trufflesuite/truffle/releases/tag/v5.0.14)**
- [ ] [Add support for signing Quorum private transactions](https://github.com/trufflesuite/truffle/issues/2153)
- [ ] [Audit Truffle's existing features/APIs for compatibility with/exposure of Quorum's feature set](https://github.com/trufflesuite/truffle/issues/1658)

dgrmunch:
How is this going? Is there any deadline? Thanks!

seesemichaelj:
Thanks @dgrmunch for reaching out!

We're still working on this. This issue is an Epic for a handful of smaller issues. We're still working on the first in the list: #1543. We'll update this issue as we complete sub-issues!

Unfortunately, we don't currently have a deadline or expected finish date at this point.

seesemichaelj:
[v5.0.9](https://github.com/trufflesuite/truffle/releases/tag/v5.0.9) has been released which provides basic Quorum support again!

[...]

PR #2154, opened 2019-07-04
Fix: Re-enable decoding of Quorum private values (return zero value instead of decoding error)

CruzMolina:
This PR fixes a [web3.js](https://github.com/ethereum/web3.js/blob/v1.0.0-beta.37/packages/web3-eth-abi/src/index.js#L225-L227) decoding regression introduced in Truffle's `web3` dependency upgrade to `1.0.0-beta.37`. For completeness, it appears this line of code was first introduced within web3.js in [`1.0.0-beta.36`](https://github.com/ethereum/web3.js/blob/v1.0.0-beta.36/packages/web3-eth-abi/src/index.js#L225-L227).

Specifically, the regression fails to properly decode Quorum hidden/private values from a `privateFor` privacy transaction. Currently, `web3` returns an `Error: Returned values aren't valid, did it run Out of Gas?` error. Prior to Truffle v5 and the `web3` dependency upgrade, a hidden/private value would be decoded as a `{ [String: '0'] s: 1, e: 0, c: [ 0 ] }` `BigNumber` object. This PR allows a private value to be decoded as a `BN` object  (`BN { negative: 0, words: [ 0, <1 empty item> ],  length: 1, red: null }`).

See [Using Quorum's privacy features...](https://www.trufflesuite.com/tutorials/building-dapps-for-quorum-private-enterprise-blockchains#using-quorum-39-s-privacy-features-to-make-transactions-private) & [Interacting with contracts privately](https://www.trufflesuite.com/tutorials/building-dapps-for-quorum-private-enterprise-blockchains#interacting-with-contracts-privately) from the Truffle Quorum tutorial for current expected behavior.

Works towards #1657 & #1658.


CruzMolina:
Currently, I've removed the guard and converted any falsy `bytes` passed to `decodeParameters` to be a 0 value.

One alternative is to keep the guard but instead of throwing the `Out of Gas?` error, convert any `bytes` meeting the guard condition(s) to be a 0 value.

Another alternative is to keep the guard, but throw our own custom `Error` (or some other meaningful output).

PR #2169, opened 2019-07-09
Replace truffle-core/mnemonic require with a relative one

chebykin:
This hardcoded `truffle-core` prevents running `truffle` from my forked version of `truffle-core` package.


gnidan:
thanks for this!

PR #2180, opened 2019-07-12
Redo output format with interfaces instead of classes (also some bugfixes)

haltman-at:
This PR does two things.

Firstly, it redoes the decoder output format with interfaces instead of classes.  The format is still the same; just now it's with interfaces and functions instead of classes and method.

The reason it originally used classes was to get `util.inspect.custom` to work.  However now instead we handle this problem but having a single `ResultInspector` class which contains a single `Result`; instead of inspecting `Result`s, when we want to print them we first wrap them in a `ResultInspector`.

Similarly, the `nativize` method is now a function.  The `toSoliditySha3Input` method has been kind of split into two different functions.  There's now a mapping key encoder in addition to the ABI encoder; that's what we now use for mapping keys instead of `toSoliditySha3Input`.  However, `key.ts` also contains a debugging function returning info for printing mapping keys.  This is basically like the old `toSoliditySha3Input`, except it's freed from the constraints of having to actually work as input to that function. :)  (So don't use it for that, as it won't work for that anymore!)

Secondly, this PR contains two bug fixes regarding compiler versions, one major and one minor.  (Don't worry, these both only exist on `next`, not `develop`.)

The first bug fix is that, in order to determine what compiler a given user-defined type was compiled with, we would... do a complicated thing that was wrong, and involved looking it up in the context; this would run into a problem when the type was defined in an interface or abstract contract, as those don't get contexts.  Now we do it the right way but just looking up the source ID in `data.info.scopes` and then getting the compiler from `solidity.info.sources`.

The second (really minor) bug fix is that I altered the compiler version check so it won't mistake prerelease versions of 0.5.0 as not being 0.5.x.

### DIFF
--- a/packages/truffle-codec-utils/src/conversion.ts
+++ b/packages/truffle-codec-utils/src/conversion.ts
@@ -189,7 +189,8 @@ export namespace Conversion {
             return coercedResult.value.asString;
           case "malformed":
             // this will turn malformed utf-8 into replacement characters (U+FFFD) (WARNING)
-            return Buffer.from(coercedResult.value.asHex, 'hex').toString();
+            // note we need to cut off the 0x prefix
+            return Buffer.from(coercedResult.value.asHex.slice(2), 'hex').toString();
         }
       }
       //fixed and ufixed are skipped for now
--- a/packages/truffle-codec-utils/src/conversion.ts
+++ b/packages/truffle-codec-utils/src/conversion.ts
@@ -194,18 +194,18 @@ export namespace Conversion {
         }
       }
       //fixed and ufixed are skipped for now
-      case "array": //WARNING: circular case not handled
+      case "array": //WARNING: circular case not handled; will loop infinitely
         return (<Values.ArrayValue>result).value.map(nativize);
       case "mapping":
         return Object.assign({}, ...(<Values.MappingValue>result).value.map(
           ({key, value}) => ({[nativize(key).toString()]: nativize(value)})
         ));
-      case "struct": //WARNING: circular case not handled
+      case "struct": //WARNING: circular case not handled; will loop infinitely
         return Object.assign({}, ...(<Values.StructValue>result).value.map(
           ({name, value}) => ({[name]: nativize(value)})
         ));
       case "magic":
-        Object.assign({}, ...Object.entries((<Values.MagicValue>result).value).map(
+        return Object.assign({}, ...Object.entries((<Values.MagicValue>result).value).map(
             ([key, value]) => ({[key]: nativize(value)})
         ));
       case "enum":
--- a/packages/truffle-codec-utils/src/conversion.ts
+++ b/packages/truffle-codec-utils/src/conversion.ts
@@ -5,6 +5,7 @@ import BN from "bn.js";
 import Web3 from "web3";
 import { Constants } from "./constants";
 import { Values } from "./types/values";
+import { enumFullName } from "./types/inspect";
 
 export namespace Conversion {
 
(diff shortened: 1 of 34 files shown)


haltman-at:
OK, some last-minute additions:

1. Hell with it, I added the tuple type.  Note that use of this type is really *not* supported yet, but I added it.

2. I added optional type hints to the uint, address, and tuple types.  Note that we don't actually use these yet.  Note that I didn't bother with very strong typing for these, because, well, they're just hints.

3. I added `decodingMode` to the decoding objects, which can be `"full"` or `"abi"`.  Right now it's always `"full"`, of course.

4. Unrelated nonfunctional change: I rewrote the allocation-attempt loop in `decodeEvent` to use a labeled `continue` rather than having a giant `try` block like I had earlier.  I figure `try` blocks should be kept to just the part you expect to throw the error, if possible.

(If you really don't like the labeled `continue` we could have the `try` block encompass the whole inner `for` loop.  But I had it way too big before, I had it around the entire body of the out `for` loop for some reason...)

packages/truffle-codec-utils/src/types/errors.ts
```diff
@@ -23,10 +23,10 @@ export namespace Errors {
   //For when we need to throw an error, here's a wrapper class that extends Error.
   //Apologies about the confusing name, but I wanted something that would make
   //sense should it not be caught and thus accidentally exposed to the outside.
-  export class DecodingError extends Error {
-    error: GenericError;
-    constructor(error: GenericError) {
-      super(error.message());
+  export class DecodingError extends Error{
+    error: ErrorForThrowing;
```
gnidan: how bout name: `ThrowableError`?

although Harry, I gotta say, I like your naming style :)

packages/truffle-debug-utils/index.js
```diff
@@ -318,7 +319,7 @@ var DebugUtils = {
 
   formatValue: function(value, indent = 0) {
     return util
-      .inspect(value, {
+      .inspect(new CodecUtils.ResultInspector(value), {
```
gnidan: 👌

packages/truffle-codec-utils/src/types/types.ts
```diff
@@ -172,6 +176,19 @@ export namespace Types {
     location?: Ast.Location;
   }
 
+  export interface OptionallyNamedType {
+    name?: string;
```
gnidan: Shouldn't this be `name: string | undefined` instead?

packages/truffle-codec/lib/decode/abi.ts
```diff
@@ -259,11 +303,18 @@ function* decodeAbiStructByPosition(dataType: Types.StructType, location: AbiLoc
   const typeId = dataType.id;
   const structAllocation = allocations[typeId];
   if(!structAllocation) {
-    let error = new Errors.UserDefinedTypeNotFoundError(dataType);
+    let error = {
+      kind: "UserDefinedTypeNotFoundError" as "UserDefinedTypeNotFoundError",
```
gnidan: that's what it is!

packages/truffle-codec/lib/decode/special.ts
```diff
@@ -24,69 +24,77 @@ export function* decodeMagic(dataType: Types.MagicType, pointer: SpecialPointer,
...
+          value: <Values.UintResult> (yield* decodeValue(
+            {
+              typeClass: "uint",
+              bits: 256
+            },
+            {location: "special", special: "value"},
+            info
+          ))
+        }
+      };
```
gnidan: good thing there's type inference

packages/truffle-codec/lib/decode/value.ts
```diff
@@ -36,142 +40,238 @@ export default function* decodeValue(dataType: Types.Type, pointer: DataPointer,
 
     case "bool": {
       if(!checkPaddingLeft(bytes, 1)) {
-        let error = new Errors.BoolPaddingError(CodecUtils.Conversion.toHexString(bytes));
+        let error = {
+          kind: "BoolPaddingError" as "BoolPaddingError",
+          raw: CodecUtils.Conversion.toHexString(bytes)
+        };
```
gnidan: you know, you could make helper functions for these, so you don't have to deal with all these object literals all the time.

gnidan: (that'd be the redux-y way to do it)

packages/truffle-codec/lib/interface/decoding.ts
```diff
@@ -151,67 +155,70 @@ export function* decodeEvent(info: EvmInfo, address: string, targetName?: string
...
-        continue;
-      }
-      const id = allocation.contractId;
-      const attemptContext = info.contexts[id];
-      const contractType = CodecUtils.Contexts.contextToType(attemptContext);
-      //you can't map with a generator, so we have to do this map manually
-      let decodedArguments: AbiArgument[] = [];
-      for(const argumentAllocation of allocation.arguments) {
-        const value = <Values.Result> (yield* decode(
+  allocationAttempts: for(const allocation of possibleAllocationsTotal) {
```
gnidan: 👀

packages/truffle-codec/lib/decode/abi.ts
haltman-at: Yeah -- for some reason you need to do that for TypeScript to assign it a string constant type instead of just the `string` type.  It's annoying.

packages/truffle-codec-utils/src/types/types.ts
haltman-at: In this context, I wouldn't consider that any different.

packages/truffle-codec-utils/src/types/errors.ts
haltman-at: Well, so `ThrowableError` was my first thought, but that just sounds like the builtin `Error` class...

packages/truffle-codec-utils/src/conversion.ts
```diff
@@ -160,7 +161,90 @@ export namespace Conversion {
...
+        return enumFullName(<Values.EnumValue>result);
+      case "contract": {
+        let coercedResult = <Values.ContractValue> result;
+        switch(coercedResult.value.kind) {
+          case "known":
+            return `${coercedResult.value.class.typeName}(${coercedResult.value.address})`;
+          case "unknown":
+            return coercedResult.value.address;
+        }
+        break; //to satisfy typescript
```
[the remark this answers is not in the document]
haltman-at: I could, but that seems less helpful.

packages/truffle-codec-utils/src/conversion.ts
```diff
@@ -160,7 +161,90 @@ export namespace Conversion {
...
+      case "mapping":
+        return Object.assign({}, ...(<Values.MappingValue>result).value.map(
+          ({key, value}) => ({[nativize(key).toString()]: nativize(value)})
+        ));
+      case "struct": //WARNING: circular case not handled
+        return Object.assign({}, ...(<Values.StructValue>result).value.map(
+          ({name, value}) => ({[name]: nativize(value)})
+        ));
+      case "magic":
+        Object.assign({}, ...Object.entries((<Values.MagicValue>result).value).map(
```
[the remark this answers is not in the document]
haltman-at: Oh crap, thanks!

packages/truffle-codec-utils/src/conversion.ts
```diff
@@ -160,7 +161,90 @@ export namespace Conversion {
...
+        }
+      }
+      //fixed and ufixed are skipped for now
+      case "array": //WARNING: circular case not handled
+        return (<Values.ArrayValue>result).value.map(nativize);
+      case "mapping":
+        return Object.assign({}, ...(<Values.MappingValue>result).value.map(
+          ({key, value}) => ({[nativize(key).toString()]: nativize(value)})
+        ));
+      case "struct": //WARNING: circular case not handled
```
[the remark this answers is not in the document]
haltman-at: Yeah, as in expect infinite loop.

Mind you, currently the circular case doesn't exist -- right now if you've got circular structures you're going to get an infinite loop when you try to decode it; you'll never even reach this part.  So at the moment it's not worth addressing that case here.

But OK yeah I'll add that comment.

packages/truffle-codec-utils/src/conversion.ts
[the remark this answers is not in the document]
haltman-at: As for why the linter missed it... because this is TypeScript and we don't have a linter turned on for that!

haltman-at:
OK, made the following changes (these are unrelated to Nick's review above):

1. I added another strict-mode error, this time for an overlarge pointer in the ABI.  In this case the potential problem isn't a DOS -- it's just that we're going to convert this pointer to a `number`, and we don't want to crash when that happens!  (OK, I guess we wouldn't actually crash due to the `try` block, but still, I'd prefer to throw our own structured errors when the error is one we expect. :) )

(Mind you, we'll still crash if this happens outside of strict mode.  But the question of how to handle this sort of thing outside of strict mode is, well, a question for another time.)

2.  `ResultInspector` and `nativize` now handle malformed strings by using node's builtin UTF-8 conversion (via `Buffer`) which uses the replacement character (U+FFFD) for malformed UTF-8 rather than throwing an exception.  Mind you, we don't use this in the decoder itself, because there we don't want to lose information.

packages/truffle-codec/lib/decode/value.ts
haltman-at: Hm, I guess.  Not clear it'd save that much trouble though.  I don't really want to bother at this point, and as I said I'm not actually convinced it'd help...

packages/truffle-codec-utils/src/conversion.ts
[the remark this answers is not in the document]
gnidan: oh right

packages/truffle-codec/lib/decode/value.ts
gnidan: Yeah I'm not about to suggest doing it now that you've done all this work making them objects. In the future, maybe it's worth doing if we're doing other refactoring work.

The reason for doing it is really just for clarity, esp. if you're looking at the helper functions inside an IDE that gives you jump-to-definition functionality (or whenever GH finally gives us that feature on this repo)

Issue #2173, opened 2019-07-10
Support command-line argument `--option=value` syntax (not just `--option value`)


eggplantzzz:
When using a command line flag with an `=` sign instead of a ` `, Truffle will throw a warning that you are using an unsupported or undocumented command. This is due to the way Truffle parses out the flags [here](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/lib/command.js#L115-L119).

This should probably be updated to replace all `=` with spaces before checking. It might also be worth considering extracting this validation into its own utility file to keep `command.js` clean.

This would probably be a good first issue.

richmanrich:
I suppost

haltman-at:
Looks like this was fixed a while back.  Closing.

PR #2181, opened 2019-07-12
Support command-line argument `--option=value`

pensivej:
This PR relates to #2173 
This supports command-line argument `--option=value` syntax (not just `--option value`).
It is designed to replace the first occurence of "=" to " ".

### DIFF
--- a/packages/truffle-core/lib/console.js
+++ b/packages/truffle-core/lib/console.js
@@ -42,7 +42,7 @@ class Console extends EventEmitter {
     this.repl.on("exit", () => this.emit("exit"));
   }
 
-  async start(callback) {
+  start(callback) {
     if (!this.repl) this.repl = new Repl(this.options);
 
     // TODO: This should probalby be elsewhere.
@@ -55,8 +55,7 @@ class Console extends EventEmitter {
       this.repl.start({
         prompt: "truffle(" + this.options.network + ")> ",
         context: {
-          web3: this.web3,
-          accounts: await this.web3.eth.getAccounts()
+          web3: this.web3
         },
         interpreter: this.interpret.bind(this),
         done: callback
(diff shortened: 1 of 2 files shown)


eggplantzzz:
Hey @pensivej, thanks for submitting this PR! However, there are a couple of issues here:
  - in calculating the `inputOptions` array, we should only put in the flagged command. In other words, if the user types `--option=true`, then we should only put `--option` into that array.
  - the `validOptions` as far as I can tell is currently being calculated correctly - I don't believe we will ever need to have `=` removed from it.
  - when we warn we also should only warn about the flagged option (don't include the value) - this means the output should look like 
```
> Warning: possible unsupported (undocumented in help) command line option: --option
```
and not 
```
> Warning: possible unsupported (undocumented in help) command line option: --option=true
```
Let me know if you have time to make these corrections!

pensivej:
Sure, I will take a deeper look into it.

packages/truffle-core/lib/console.js
```diff
@@ -42,7 +42,7 @@ class Console extends EventEmitter {
     this.repl.on("exit", () => this.emit("exit"));
   }
 
-  start(callback) {
+  async start(callback) {
```
gnidan: Heads up: it's bad practice to make a `callback`-style function also `async`. Doing this has broken things in strange/unexpected ways.

Instead, I think you have to wrap the call to `this.repl.start` inside a `this.web3.eth.getAccounts().then(...)`.

pensivej:
Oops wrong commits
I will make another pull request

gnidan:
Heh so much for my quick review! Well, I'll copy my comment about the `async`+`callback` problem over to the new PR.

pensivej:
@gnidan wow that was quite a quick review! I'll submit new PRs :)

PR #2182, opened 2019-07-12
Test Web3.js 1.0 release by adding `yarn link` step to Travis install

gnidan:
To validate https://github.com/ethereum/web3.js/pull/2905

### DIFF
--- a/packages/truffle-decode-utils/src/contexts.ts
+++ b/packages/truffle-decode-utils/src/contexts.ts
@@ -3,7 +3,6 @@ const debug = debugModule("decode-utils:contexts");
 
 import { Abi } from "truffle-contract-schema/spec";
 import { AbiCoder } from "web3-eth-abi";
-import { AbiItem } from "web3-utils";
 const abiCoder = new AbiCoder();
 import escapeRegExp from "lodash.escaperegexp";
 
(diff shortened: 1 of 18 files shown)


CruzMolina:
🤞

gnidan:
(Must also double check that we don't need to do a rebuild post linking)

scripts/test-web3.sh
```diff
@@ -19,7 +19,7 @@ link_web3() {
   do
     cd $pkg
     yarn link web3
-    cd ..
+    cd ../..
```
CruzMolina: 😮

gnidan:
scrypt!!!

gnidan:
Tried [yalc](https://github.com/whitecolor/yalc) and that's causing its own problems. At least, it's either causing its own problems, or the release branch breaks types for `truffle-decode-utils`

packages/truffle-decode-utils/src/contexts.ts
```diff
@@ -97,7 +96,7 @@ export namespace Contexts {
             debug("signature read: %s", signature);
             //if not, compute it ourselves
             if(signature === undefined) {
-              signature = abiCoder.encodeFunctionSignature(<AbiItem>abiEntry);
+              signature = abiCoder.encodeFunctionSignature(<any>abiEntry);
```
gnidan: Doing this since #2214 demonstrates the same `AbiItem` failure as on this branch, and I think this shouldn't impact the web3 release validation.

gnidan:
We're green! This PR uses yalc to inject locally-built web3.js from the `release/1.0` branch. I'm loosening the type in truffle-decode-utils because the problem is happening separate from this test, as observable elsewhere.

I think we can confidently sign off on ethereum/web3.js#2905!

PR #2213, opened 2019-07-12
Upgrade lodash dependency

eggplantzzz:
Upgrade lodash in two packages to solve security vulnerability

### DIFF
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "fs-extra": "6.0.1",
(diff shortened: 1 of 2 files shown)


packages/truffle-artifactor/package.json
```diff
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "fs-extra": "6.0.1",
-    "lodash": "4.17.11",
+    "lodash": ">=4.17.13",
```
gnidan: `>=` is dangerous because this will automatically include `5.0.0`, which might have breaking changes

Issue #2157, opened 2019-07-05
Enhancement: Add `accounts` global to truffle console


gnidan:
## Issue

Spun out from #2135, this issue proposes to add a convenient global to `truffle console` and `truffle develop`, so that users can easily see their accounts.

djokicx:
Thank you for opening a separate issue for this!

TripleSpeeder:
As https://github.com/trufflesuite/truffle/pull/2218 is merged i think this can be closed.

gnidan:
Yep! Bam!

PR #2218, opened 2019-07-15
Add `accounts` global to truffle console

pensivej:
This PR relates to #2157 
Also, thanks to @gnidan for quick comments :)
I set the branch somehow complicated so I just removed the fork and made the new one 👍 so that I can submit two separate PRs.

### DIFF
--- a/packages/truffle-core/lib/console.js
+++ b/packages/truffle-core/lib/console.js
@@ -51,8 +51,9 @@ class Console extends EventEmitter {
     this.options.repl = this.repl;
 
     try {
-      const abstractions = this.provision();
       this.web3.eth.getAccounts().then(accounts => {
+        const abstractions = this.provision();
+
         this.repl.start({
           prompt: "truffle(" + this.options.network + ")> ",
           context: {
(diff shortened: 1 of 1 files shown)


eggplantzzz:
This looks nice. Nice and simple.

packages/truffle-core/lib/console.js
```diff
@@ -52,13 +52,16 @@ class Console extends EventEmitter {
...
     try {
       const abstractions = this.provision();
-      this.repl.start({
-        prompt: "truffle(" + this.options.network + ")> ",
-        context: {
-          web3: this.web3
-        },
-        interpreter: this.interpret.bind(this),
-        done: callback
+      this.web3.eth.getAccounts().then(accounts => {
```
gnidan: Oh shoot. I might be being paranoid, but I think maybe we want this entire function to get wrapped in the `.then()`. I'm not sure, but we might need to make sure that `this.resetContractsInConsoleContext` doesn't rely on some implicit ordering. _shrug_ I hate these callback-style functions everywhere, it makes the code so ugly :(

pensivej: Maybe including this.resetContractsInConsoleContext into .then() will solve the issue?
I don't like this type of callback functions everywhere also :(

fainashalts:
+1 This looks good to me!

eggplantzzz:
Cool, time to merge!

pensivej:
Thanks for the review :)

PR #2220, opened 2019-07-15
Internal Improvement: Port truffle-artifactor to TS

CruzMolina:
This PR ports the `truffle-artifactor` library (excluding tests) to TypeScript.

Adds preliminary typings to `truffle-contract` & `truffle-contract-schema`.


packages/truffle-contract-schema/typings/index.d.ts
```diff
@@ -0,0 +1,7 @@
+declare module "truffle-contract-schema" {
+  import Contract from "truffle-contract";
+  namespace Schema {
+    export function normalize(dirtyObj: object, opts?: object): Contract.TruffleContract;
```
gnidan: This returns the ContractObject, not a truffle-contract Contract...

i.e. this is more accurate:
```typescript
import { ContractObject } from "truffle-contract-schema/spec";
```

packages/truffle-artifactor/.gitignore
```diff
@@ -4,3 +4,4 @@ config/development
 config/test
 yarn.lock
 package-lock.json
+dist
```
gnidan: I don't think these nested .gitignores do anything.

packages/truffle-artifactor/README.md
```diff
@@ -3,7 +3,7 @@
 This package saves contract artifacts into JSON files
 
 ```javascript
-const Artifactor = require("truffle-artifactor");
+const { Artifactor } = require("truffle-artifactor");
```
gnidan: Breaking change?

packages/truffle-artifactor/artifactor.ts
```diff
@@ -1,15 +1,16 @@
...
-class Artifactor {
-  constructor(destination) {
+export class Artifactor {
+  destination: string;
+  constructor(destination: string) {
     this.destination = destination;
   }
 
-  async save(artifactObject) {
+  async save(artifactObject: object) {
```
gnidan: We can't type this any more specifically than `object`? It should be a truffle-contract Contract

packages/truffle-artifactor/utils.ts
```diff
@@ -1,19 +1,23 @@
-const fse = require("fs-extra");
-const _ = require("lodash");
+import fse from "fs-extra";
+import _ from "lodash";
+import Contract from "truffle-contract";
 
-const writeArtifact = (completeArtifact, outputPath) => {
+export function writeArtifact(
+  completeArtifact: Contract.TruffleContract,
```
gnidan: This should be a ContractObject from `truffle-contract-schema/spec`

packages/truffle-artifactor/utils.ts
```diff
@@ -1,19 +1,23 @@
...
-};
+}
 
-const finalizeArtifact = (
-  normalizedExistingArtifact,
-  normalizedNewArtifact
-) => {
+export function finalizeArtifact(
+  normalizedExistingArtifact: Contract.TruffleContract,
+  normalizedNewArtifact: Contract.TruffleContract
```
gnidan: Same with these

packages/truffle-contract-schema/.gitignore
```diff
@@ -1,4 +1,4 @@
 node_modules
 yarn.lock
 package-lock.json
-*.d.ts
\ No newline at end of file
+spec/index.d.ts
```
gnidan: (Same inquiry about nested .gitignores)

packages/truffle-contract-schema/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "3.0.11",
   "description": "JSON schema for contract artifacts",
   "main": "index.js",
+  "typings": "./typings/index.d.ts",
```
gnidan: Does this conflict with the types exposed by `truffle-contract-schema/spec`?

packages/truffle-contract/typings/index.d.ts
```diff
@@ -0,0 +1,11 @@
+declare module "truffle-contract" {
+  namespace TruffleContract {
+    export interface TruffleContract {
+      contractName: string;
+      contract_name: string;
```
gnidan: blech, both!?

packages/truffle-core/test/config.js
```diff
@@ -6,7 +6,7 @@ var Contracts = require("truffle-workflow-compile");
 var Ganache = require("ganache-core");
 var provision = require("truffle-provisioner");
 var Resolver = require("truffle-resolver");
-var Artifactor = require("truffle-artifactor");
+const { Artifactor } = require("truffle-artifactor");
```
gnidan: The more I see this, the more I think we should just keep our `index.js` to maintain the default export, where all that file would do is import from `./dist`

CruzMolina: Is there a way to `export default` w/o breaking things? Importing destructured like this seemed prettier than having to do `const Artifactor = require("truffle-artifactor").default` everywhere.

packages/truffle-artifactor/.gitignore
CruzMolina: I can remove them, but they do seem to make git ignore them when working locally.

packages/truffle-contract-schema/.gitignore
CruzMolina: Same here. This was making git ignore the added typings, so I set it to only ignore the output spec typings. (Looks like the pkg was setup to ignore the spec typings on purpose).

packages/truffle-artifactor/artifactor.ts
CruzMolina: I think `artifactObject` is technically not a `ContractObject` until we normalize it. I believe we allow passing any object in there. Maybe that doesn't matter and I should just type it as a `ContractObject`?

packages/truffle-contract/typings/index.d.ts
CruzMolina: yuppp, I believe we currently support both.

packages/truffle-contract-schema/package.json
CruzMolina: For whatever reason, we don't currently expose those typings. Should I do so within `typings/index.d.ts`?

packages/truffle-artifactor/.gitignore
gnidan: Oh, well, empirical observation is better than my postulatin'!

packages/truffle-artifactor/artifactor.ts
gnidan: Nope, you're probably right! Leave it!

packages/truffle-contract-schema/package.json
gnidan: Yep, I think so!

packages/truffle-contract/typings/index.d.ts
gnidan: _sigh_ ok

packages/truffle-core/test/config.js
gnidan: You have to do it in JS I believe

gnidan: Or maybe you can finagle the tsconfig to make it work the way you want.

packages/truffle-contract-schema/spec/contract-object.spec.json
```diff
@@ -11,6 +11,11 @@
       "description": "Name used to identify the contract",
       "default": "Contract"
     },
+    "contract_name": {
+      "allOf": [{ "$ref": "#/definitions/ContractName" }],
+      "description": "Name used to identify the contract",
+      "default": "Contract"
+    },
```
gnidan: Blech, can we not do this? (Is it possible?)

CruzMolina: i believe changing this [line](https://github.com/trufflesuite/truffle/pull/2220/files#diff-1e819f795cdbdcfa8915a5e598b511b2R49) to use `contractName` would be a breaking change 😞 .

gnidan: Shouldn't be breaking. That's what `normalize` is for... `contractName` is the schema-correct version, and `contract_name` should get converted via normalize.

Issue #2222, opened 2019-07-16
Plugin improvement: expose truffle-core's index.js in webpack bundle 


cgewecke:
As a plugin I might like to compile with special settings or run tests after preprocessing something. It would be convenient to have all the resources exported in truffle-core's [index.js](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-core/index.js) immediately available with
```javascript
const truffle = require("truffle");
```

Not sure but I think this is possible with webpack's [expose-loader](https://webpack.js.org/loaders/expose-loader/). Like maybe you just have to expose `truffle-core` as `index.js` and list the bundled cli.js as the package.json's main? 

**Questions**
+ Is this already possible but I've missed it?
+ If not, Is there any appetite for a PR like this? 
+ If it becomes possible, should any additional truffle modules be exposed?

gnidan:
_issue number with all 2's, nice @cgewecke_

Is this file actually useful in its current state? I barely even knew it existed.

Anyway, just as a quick response:

> Is this already possible but I've missed it?

I haven't looked, but the solution you describe sounds viable, or maybe it's simpler than that... you might only need to define another entrypoint in the webpack config.

> If not, Is there any appetite for a PR like this?

I'd be in favor of getting Truffle available as a library!

> If it becomes possible, should any additional truffle modules be exposed?

Probably! I think maybe it's worth collecting use cases here? It wouldn't be awful, but it'd be a shame to suddenly offer Truffle as a library only to break the interface quickly afterwards.

cgewecke:
@gnidan Whoa! I didn't notice that about the 2's'. Crazy!

I don't know about other use cases but mine is coverage. I need to:
+ set the provider as a ganache-core instance (so I can listen to the vm step and watch opcodes)
+ compile from a custom contracts_directory with `optimization.enabled: false` to a custom build_directory.
+ execute the tests using the artifacts written to the custom buid_directory.

I *think* if I have `workflow-compile` and `truffle-core/lib/test` I'm good, idk.

> define another entrypoint in the webpack config.

Apologies, could you explain that to me?

gnidan:
> define another entrypoint in the webpack config.

[...]

PR #2226, opened 2019-07-17
Generate truffle library webpack bundle for package main

cgewecke:
#2222 

Makes the following possible (in a plugin context, for example). 
```javascript
const truffle = require("truffle");
await truffle.contracts.compile(config);
```

**Implementation Details**
+ In #2222 I proposed using webpack [expose-loader](https://www.npmjs.com/package/expose-loader) for this because I was concerned about the bundle size growing a lot. The unpacked version *is* bigger here (25-30%), but the packed version sent over the wire from npm is still quite small and it doesn't affect installation speeds. I published my own version to test that issue & validate the POC in the code snippet above - everything worked. Also expose-loader is a nightmare.

+ Added `ganache-core` to the `truffle-core` exports  so consumer can spin up an in-process provider.

+ Added top layer API correctness tests - you'll get a scenario test failure if you delete/rename one of those methods. 

[cf: solidity-coverage 0.7.0](https://github.com/sc-forks/solidity-coverage/issues/346)

### DIFF
--- a/packages/truffle-core/index.js
+++ b/packages/truffle-core/index.js
@@ -7,5 +7,6 @@ module.exports = {
   contracts: require("truffle-workflow-compile"),
   package: require("./lib/package"),
   test: require("./lib/test"),
-  version: pkg.version
+  version: pkg.version,
+  ganache: require("ganache-core")
 };
--- a/packages/truffle/cli.webpack.config.js
+++ b/packages/truffle/cli.webpack.config.js
@@ -31,6 +31,13 @@ module.exports = {
       "services",
       "analytics",
       "main.js"
+    ),
+    library: path.join(
+      __dirname,
+      "../..",
+      "node_modules",
+      "truffle-core",
+      "index.js"
     )
   },
   target: "node",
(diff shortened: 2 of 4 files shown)


cgewecke:
Also had a thought about the API reliability question Nick raises in #2222 - perhaps the simplest thing is to document this feature in the plugin docs, declare the API all-caps unstable (except for truffle.version) and advise people to use individual semvered `truffle-` modules when possible. Happy to open a PR for that as well. 

I my case I *really* need to be able to invoke the `test.run` logic after manipulating the config. The absence `require("truffle")` would mean I'd have to include truffle-core as a dependency (which is a heavy payload for a plugin).

eggplantzzz:
Thanks for the PR @cgewecke, I'll try and look over this soon!

cgewecke:
@eggplantzzz Thanks! Just lmk if/when you want a rebase.

cgewecke:
@gnidan Awesome, thanks so much! Will open the docs PR tomorrow.

PR #2232, opened 2019-07-19
Factor out wire decoder from contract decoder

haltman-at:
This PR factors out the wire-decoder functions from the contract decoder, in order to reduce the amount of copypasted code.  In the process, I made a few minor improvements/fixes as well.

Rather than having `TruffleContractDecoder` extend `TruffleWireDecoder`, I took @gnidan's advice and had it contain one internally instead.  This way, multiple contract decoders can be made from the same wire decoder, without having to recompute ABI allocations, etc.  Basically this is potentially useful if you want to have contract decoders for multiple contracts in the same project.

The contract decoder now just delegates to the wire decoder for all wire-decoding functions; its only major function still handled internally is decoding contract state.  This also means the wire-decoding functions no longer have the restrictions I placed on them earlier -- they have the full power of the wire decoder.  (Also I realized such restrictions don't really make sense because, say, one contract can cause another to emit an event.)

The `events` method does make one change before delegating to the wire decoder -- it (by default) sets the `address` option to the contract's address.  However the user can override this, including by passing in `address: undefined` to turn off filtering by address.

The contract now also delegates much of the setup to the wire decoder -- the wire decoder now has functions to extract the info that the contract decoder will need.  Unfortunately, since the contract decoder doesn't extend the wire decoder, these have to be made public -- TypeScript doesn't have an access modifier to restrict access to the same package, like C# or Java -- but I've at least put a comment marking these functions as intended for internal use only.

The file `index.ts` now contains three functions:

1. `forProject`, which makes a wire decoder;

2. `forContract`, which makes a contract decoder just as before;

3. `forContractWithDecoder`, which makes a contract decoder, but takes a `TruffleWireDecoder` instead of a provider and array of `relevantContracts`.

Of these, (1) works just like before (except that I've eliminated `TruffleWireDecoder`'s `init` function and just moved that code into the constructor; there's no real reason for it to have a separate `init` function).  Number (3) is new, and works by just invoking the `TruffleContractDecoder` constructor, which has been changed to now take arguments in the form of (3).  (Well, and then waiting for `init()`, of course.)

Number (2), therefore, now works a little differently.  First, it makes a list of all contracts passed in -- i.e., it takes `relevantContracts` and (if it's not already there) prepends `contract` to it.  Then it creates a `TruffleWireDecoder` with the given `provider` and the just-constructed list of contracts.  Then it creates a `TruffleContractDecoder` from that (with the given `contract` and, if supplied, `address`).

Other minor changes: The decoder now excludes contracts with empty bytecode (`0x`) when setting up contexts.  Also, as with the wire-decoder, I moved into the constructor all the setup that doesn't need to be `async`; only the async stuff remains in `init`.  (Edit: Getting rid of `init` in the wire-decoder actually now seems foolish to me due to future stuff I have planned but  didn't think of that... oh well -- leaving it how it is for now; I can move it back in the future, that's easy enough.)

Actually, it's not clear that all of the contexts stuff that I included needs to actually be there -- some of it is unused at the moment -- but I decided to leave it alone as I'm planning on possibly redoing that soon, so I figured it's not worth it to change it right now.

Thought for the future: This was a good thing to do but I think we should go even further in this direction, honestly.  Why is the contract decoder restricted to a single address?  Like, you've got to make a whole new contract decoder, redo allocation and all, if you've got multiple instances of the same contract?  That seems silly.  Seems to me we ought to have some new sort of decoder class, one that works for a contract *class*, not a contract *instance*.  Alternatively, turn the `WireDecoder` into a full-blown `ProjectDecoder`, and have it decode state as well.  Do all allocations upfront and have it recognize contract class by the bytecode.  Although you'd still want something like the `ContractDecoder` for when you don't have the bytecode available, I guess.  Anyway, just idle thoughts; certainly not doing that in this PR, which is basically just some factoring.

### DIFF
--- a/packages/truffle-codec-utils/src/contexts.ts
+++ b/packages/truffle-codec-utils/src/contexts.ts
@@ -12,7 +12,7 @@ import { CompilerVersion } from "./compiler";
 
 export namespace Contexts {
 
-  export type Contexts = DecoderContexts | DebuggerContexts;
+  export type Contexts = DecoderContexts | DebuggerContexts | DecoderContextsById;
 
   export type Context = DecoderContext | DebuggerContext;
 
--- a/packages/truffle-decoder/lib/contract.ts
+++ b/packages/truffle-decoder/lib/contract.ts
@@ -8,6 +8,7 @@ import Web3 from "web3";
 import { ContractObject } from "truffle-contract-schema/spec";
 import BN from "bn.js";
 import { Definition as DefinitionUtils, AbiUtils, EVM, AstDefinition, AstReferences } from "truffle-codec-utils";
+import TruffleWireDecoder from "./wire";
 import { BlockType, Transaction } from "web3/eth/types";
 import { Log } from "web3/types";
 import { Provider } from "web3/providers";
@@ -23,16 +24,14 @@ export default class TruffleContractDecoder extends AsyncEventEmitter {
   private contractNetwork: string;
   private contractAddress: string;
   private contractCode: string;
-  private relevantContracts: ContractObject[];
-
-  private contracts: DecoderTypes.ContractMapping = {};
-  private contractNodes: AstReferences = {};
-  private contexts: CodecUtils.Contexts.DecoderContexts = {};
-  private contextsById: CodecUtils.Contexts.DecoderContextsById = {}; //deployed contexts only
   private context: CodecUtils.Contexts.DecoderContext;
   private constructorContext: CodecUtils.Contexts.DecoderContext;
   private contextHash: string;
   private constructorContextHash: string;
+
+  private contexts: CodecUtils.Contexts.DecoderContexts = {};
+  private contextsById: CodecUtils.Contexts.DecoderContextsById = {}; //deployed contexts only
+  private constructorContextsById: CodecUtils.Contexts.DecoderContextsById = {};
 
   private referenceDeclarations: AstReferences;
   private userDefinedTypes: Types.TypesById;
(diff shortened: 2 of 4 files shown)


haltman-at:
OK, actually, I decided to add a bit of extra functionality while I was at it.  The result is... a bit awkward, and it's quite possible this whole PR should have been done a different way so as to make it less awkward.  But, it should be done in a way that makes this functionality possible.  That's why I made sure to add it now, so we remember and don't accidentally do it in a way that makes this too hard to do!

So, what did I add? Well, let's say I create a `TruffleContractDecoder` for a contract that has no `deployedBytecode` field.  Then, let's say I attempt to use this decoder to decode a transaction or event containing an external function pointer whose address is that of this contract.  What should happen?

Well, the contract has no `deployedBytecode`; so if I were to do it with the wire decoder rather than a contract decoder, it will have no way of knowing that that address is an instance of this contract, and so the function pointer will come back as unknown.  But we're not using the wire decoder; we're using a contract decoder for that specific address.  We've just *told* it what sort of contract lives at at that address, and what its ABI is.  So we ought to be able to decode it regardless.

How do we implement this?  Well, one could do all sorts of tricky alterations to the external function decoder, but here's a simpler way requiring no alteration.  As part of the decoder's initialization, if the contract has no `deployedBytecode`, we instead get its code off the blockchain and add a context with that deployed bytecode.  Hooray!  Problem solved!

Of course, to do that, we somehow have to get this extra info to the wire decoder.  So, various wire decoder methods now take an optional `additionalContexts` argument to make this work.

Now what I was originally going to do was have the contract decoder actually modify the wire decoder's contexts via a special method on startup, rather than having it pass the additional ones each time it makes a call.  But this would mean modifying the wire decoder object, which doesn't work with my idea that one wire decoder could be shared by multiple contract decoders.  (I mean, unless we're OK with that shared wire decoder suddenly gaining information it shouldn't have.  But I sure don't like that.  I mean, maybe the information the user passed in was unreliable; it should be contained to that context, you know?)

Anyway, yeah, the result is pretty awkward, but, oh well.

Also I don't like having to just, like, write comments saying that certain parameters are only for internal use, but, again, TypeScript doesn't have package-level protection, so, yeah, kind of stuck there.  (I mean, I *could* separate out functions like `decodeTransaction` into two, one for public use and one for internal use, it just strikes me as kind of pointless seeing as the one for internal use would have to be `public` anyway. :-/ )

packages/truffle-codec-utils/src/contexts.ts
```diff
@@ -12,7 +12,7 @@ import { CompilerVersion } from "./compiler";
 
 export namespace Contexts {
 
-  export type Contexts = DecoderContexts | DebuggerContexts;
+  export type Contexts = DecoderContexts | DebuggerContexts | DecoderContextsById;
```
gnidan: We don't need to add this to `Context` also? (I think I understand why not, but just verifying)

packages/truffle-decoder/lib/contract.ts
```diff
@@ -43,156 +43,92 @@ export default class TruffleContractDecoder extends AsyncEventEmitter {
...
-      ([id, _]) => this.contractNodes[parseInt(id)].contractKind === "library"
-    ).
-    map(
-      ([id, { abi }]) => ({
-        abi: AbiUtils.schemaAbiToAbi(abi),
-        id: parseInt(id)
-      })
-    );
+    this.referenceDeclarations = this.wireDecoder.getReferenceDeclarations();
+    this.userDefinedTypes = this.wireDecoder.getUserDefinedTypes();
```
gnidan: Do we need to forward these onto `this`? Or can we just defer to `this.wireDecoder` when necessary?

packages/truffle-decoder/lib/contract.ts
```diff
@@ -43,156 +43,92 @@ export default class TruffleContractDecoder extends AsyncEventEmitter {
...
 
     this.allocations = {};
+    this.allocations.abi = this.wireDecoder.getAbiAllocations();
```
gnidan: Same here - is this data duplication necessary?

packages/truffle-decoder/lib/contract.ts
```diff
@@ -362,123 +282,23 @@ export default class TruffleContractDecoder extends AsyncEventEmitter {
...
-    }
-    //at this point, result.value holds the final value
-    const decodings = <Codec.LogDecoding[]>result.value;
-    
-    return {
-      ...log,
-      decodings
-    };
+  public async decodeLog(log: Log): Promise<DecoderTypes.DecodedLog> {
+    return await this.wireDecoder.decodeLog(log, {}, this.additionalContexts);
```
gnidan: nice dedupe!

packages/truffle-codec-utils/src/contexts.ts
haltman-at: Correct.

packages/truffle-decoder/lib/contract.ts
haltman-at: I guess we don't really need to, no.

packages/truffle-decoder/lib/contract.ts
haltman-at: This one I'd prefer to keep, because I'd like to be able to just set `this.allocations` once and then be done with it.

haltman-at:
OK, after talking to @gnidan some more, I'm leaving the data duplication as-is for now and just merging.

PR #2241, opened 2019-07-22
Enhancement: Enable silencing compilation output during `truffle test` runs

CruzMolina:
This PR enables silencing compilation output during `truffle test` runs. `truffle compile` && `truffle migrate` already support this so the behavior is enabled to be exactly the same.

### DIFF
--- a/packages/truffle-core/lib/commands/migrate.js
+++ b/packages/truffle-core/lib/commands/migrate.js
@@ -87,10 +87,6 @@ const command = {
       {
         option: "--skip-dry-run",
         description: "Do not run a test or 'dry run' migration."
-      },
-      {
-        option: "--quiet",
-        description: "Suppress all compilation output."
       }
     ]
   },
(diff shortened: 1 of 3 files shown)


eggplantzzz:
Do we need to add this to other commands besides `migrate` and `test`? Maybe compile as well?

packages/truffle-core/lib/commands/test.js
```diff
@@ -38,6 +38,10 @@ const command = {
       {
         option: "--show-events",
         description: "Log all contract events."
+      },
+      {
+        option: "--quiet",
+        description: "Suppress all compilation output."
```
eggplantzzz: Hmm, should this suppress the migration report as well? Seems kind of weird for this option to only pertain to compilation.

packages/truffle-core/lib/commands/migrate.js
```diff
@@ -87,6 +87,10 @@ const command = {
       {
         option: "--skip-dry-run",
         description: "Do not run a test or 'dry run' migration."
+      },
+      {
+        option: "--quiet",
```
eggplantzzz: Maybe if this option is only for compilation we should consider naming it something else? Something like `--quiet-compile`?

packages/truffle-core/lib/commands/test.js
CruzMolina: Hmmm, good point. 🤔

PR #2245, opened 2019-07-23
Make out-of-range enums not an error; put "AsBN" on state output

haltman-at:
OK, I decided to make out-of-range enums not an error after all.  I'll explain why in a moment.  Also, I realized that for consistency, the fields in `state` that are `BN`'s should have `AsBN` in their field name, which is unrelated but I decided to toss it in there.

Anyway, the reason I changed this isn't actually for decoding now, but for encoding later.  Or rather, for wrapping.  I thought it would be confusing if the wrapping-procedure *would* let you wrap an out-of-range value as an enum if it only has ABI information, but wouldn't if it has full definition information.  This might lead to people going, how do I force ABI-only encoding mode so this will work?  Etc.  I don't want to deal with that, so I made out-of-range no longer an error.

Basically, things get confusing if a type acts too differently -- in particular, has a different set of legal encodings -- from its underlying ABI type (when such a thing exists).  So I decided to make it line up.

Note that padding errors are still errors, though, because, well, the above gives no reason to make them not errors.

### DIFF
--- a/packages/truffle-codec-utils/src/conversion.ts
+++ b/packages/truffle-codec-utils/src/conversion.ts
@@ -209,6 +209,7 @@ export namespace Conversion {
             ([key, value]) => ({[key]: nativize(value)})
         ));
       case "enum":
+        //note: if it's out of range, this will just be undefined
         return enumFullName(<Values.EnumValue>result);
       case "contract": {
         let coercedResult = <Values.ContractValue> result;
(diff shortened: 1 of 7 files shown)


haltman-at:
OK, this is actually a bad PR.  Enums are more complicated than that and I think we just have to accept the fact that they don't necessarily act like the underlying ABI type.  I'm abandoning this PR (although not everything in it) as a bad idea.  Will close this later.

PR #2246, opened 2019-07-24
TruffleDB Persistence MVP

fainashalts:
Persistence here works, using jsdown as a drop-in for memdown in a roll-your-own adapter. 

The persisted data is saved in `data.json` files inside of a `.db/resource` directory in the project. Workspace tests have been updated to query a separate instance of Workspace called `persistedWorkspace` to ensure that data is truly being persisted, and `tempDir` directories have been added for persisting files during tests.


packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -58,7 +62,14 @@ describe("ContractNames", () => {
...
 
     const { contractNames } = data;
-    expect(contractNames).toEqual([]);
+    const dirExists = await fse.exists(path.join(tempDir.name, ".db", "contract"));
+
+    if(dirExists) {
+     expect(contractNames).toEqual(["MagicSquare", "SquareLib", "Migrations", "VyperStorage", "Migrations"]);
+    } else {
+       expect(contractNames).toEqual([]);
+    }
```
gnidan: I think this should be two separate test cases (one for it existing, one for it not)

packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -50,14 +53,34 @@ export class Workspace {
...
+
   private ready: Promise<void>;
 
-  constructor () {
-    PouchDB.plugin(PouchDBMemoryAdapter);
+  constructor (workingDirectory:string) {
     PouchDB.plugin(PouchDBFind);
 
+    this.jsondownpouch["valid"] = () => true;
+    this.jsondownpouch["use_prefix"] = false;
```
gnidan: I think these should be in `this.adapter`

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -20,9 +20,11 @@ tmp.setGracefulCleanup();
 
 class WorkspaceClient {
   private workspace: Workspace;
+  private persistedWorkspace: Workspace;
 
   constructor () {
     this.workspace = new Workspace(tempDir.name);
+    this.persistedWorkspace = new Workspace(tempDir.name);
```
gnidan: Maybe naming?

packages/truffle-db/src/workspace/test/index.ts
```diff
@@ -36,6 +38,18 @@ class WorkspaceClient {
 
     return result.data;
   }
+
+  async executePersisted (request, variables = {}) {
```
gnidan: Maybe make a parameter for selecting workspace :)

packages/truffle-db/src/workspace/test/source.spec.ts
```diff
@@ -37,7 +37,8 @@ describe("Source", () => {
   });
 
   test("can be queried", async() => {
-    const getSourceResult = await wsClient.execute(GetSource, { id: expectedId });
+    const getSourceResult = await wsClient.executePersisted(GetSource, { id: expectedId });
+
```
gnidan: Maybe add another test to assert that `executePersisted` behavior in fact fails if you delete the persistence directory

packages/truffle-db/src/workspace/test/utils.ts
```diff
@@ -33,6 +39,18 @@ export class WorkspaceClient {
     );
     return result.data;
   }
+
+  async executePersisted (request, variables = {}) {
```
gnidan: Maybe add an optional parameter to `execute` instead of making a separate method

packages/truffle-db/src/workspace/test/utils.ts
```diff
@@ -16,11 +17,16 @@ export const fixturesDirectory = path.join(
   "fixtures"
 );
 
+const tempDir = tmp.dirSync({ unsafeCleanup: true });
+tmp.setGracefulCleanup();
+
 export class WorkspaceClient {
   private workspace: Workspace;
+  private persistedWorkspace: Workspace;
```
gnidan: Maybe this warrants a comment specifying the purpose of the two workspaces

packages/truffle-db/src/workspace/test/utils.ts
```diff
@@ -16,11 +17,16 @@ export const fixturesDirectory = path.join(
...
 
   constructor () {
-    this.workspace = new Workspace();
+    this.workspace = new Workspace(tempDir.name);
+    this.persistedWorkspace = new Workspace(tempDir.name);
```
gnidan: Are you sure you want to create two workspaces all the time? Maybe this could be a `constructor` argument (cause we really only need the second workspace for that particular test)

packages/truffle-db/src/workspace/test/utils.ts
fainashalts: Agreed, adding.

packages/truffle-db/src/workspace/pouch.ts
fainashalts: I tried moving it these into this.adapter as we discussed and for some reason a bunch of errors got thrown about reading the json data. Not sure why - ok to leave it as is or would you rather I investigate further?

packages/truffle-db/src/workspace/test/index.ts
fainashalts: This should just actually always be blank, because we haven't added contract names when the test is originally run. We add contracts elsewhere, and test for persistence so I think its ok to leave just the test for the empty array (will still be empty even with persistence since I remove the .db folder in test cleanup each time).

packages/truffle-db/src/workspace/test/source.spec.ts
fainashalts: This is actually slightly more problematic than we thought -- I believe jsondownpouch actually saves the data in memory *and* in json format... so deleting the directory doesn't cause the test to fail. Of course, that means this way of testing persistence isn't accurate, so I'm a bit at a loss of how else to test it. I suppose I can check for the existence of the expected subdirectories?

fainashalts: This is figured out with my last commit. Main issue appeared to be that if we instantiate two workspaces at the same time there appears to be some cross-talk between them? I'm not totally sure why but if I instantiate persistedWorkspace later on my tests work as expected.

packages/truffle-db/src/workspace/test/utils.ts
fainashalts: I now only create the persisted workspace in the execute function if the appropriate parameter is passed. This way it is only done for the persistence test!

packages/truffle-db/package.json
```diff
@@ -56,10 +55,12 @@
   "dependencies": {
     "@gnd/graphql-tools": "^4.0.5-fix.0",
     "@gnd/jsonschema2graphql": "^1.0.15",
-    "@truffle/artifactor": "^4.0.31",
+    "@truffle/artifactor": "^4.0.32",
+    "@types/pouchdb-adapter-leveldb": "^6.1.3",
```
gnidan: FYI `@types/` packages can go in `devDependencies`, since they're only used during development and don't get distributed with the package

packages/truffle-db/src/loaders/index.ts
```diff
@@ -26,6 +26,7 @@ const resolvers = {
     artifactsLoad: {
       resolve: async (_, args, { artifactsDirectory, contractsDirectory, workingDirectory, db }, info) => {
         const tempDir = tmp.dirSync({ unsafeCleanup: true })
+        tmp.setGracefulCleanup();
```
gnidan: Does this also need that `removeCallback()` business?

packages/truffle-db/src/workspace/test/utils.ts
```diff
@@ -16,22 +18,38 @@ export const fixturesDirectory = path.join(
...
   private workspace: Workspace;
+  private persistedWorkspace: Workspace;
 
   constructor () {
-    this.workspace = new Workspace();
+    this.workspace = new Workspace(tempDir.name);
+  }
+
+  async destroy(resource) {
+    await this.workspace.dbApi.bytecodes.destroy();
```
gnidan: You have this hardcoded as `bytecodes` instead of `resource`

packages/truffle-db/src/loaders/index.ts
fainashalts: Yep, it's there on line 39 from before!

packages/truffle-db/src/workspace/test/utils.ts
fainashalts: Thanks! Fixed.

PR #2249, opened 2019-07-24
Internal improvement: Port truffle-box to TS

CruzMolina:
This PR ports the `truffle-box` library (excluding tests) to TypeScript.

Adds preliminary typings to `truffle-config`.


packages/truffle-box/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "paths": {
+      "web3": [
+        "../../node_modules/@types/web3/index",
+        "node_modules/web3/index"
+      ]
+    },
+    "rootDir": ".",
+    "typeRoots": [
+      "./typings",
+      "node_modules/@types",
```
gnidan: This might end up being brittle if some of those `@types` get moved to `../../node_modules`. It's probably best practice to follow the way `@types/mocha` does it below, and explicitly list all required type packages.

packages/truffle-box/lib/config.ts
```diff
@@ -1,6 +1,6 @@
-const fse = require("fs-extra");
+import fse from "fs-extra";
```
gnidan: Probably worth adding an `export interface BoxConfig { }` and making `read` return that, so people can see what they should expect.

packages/truffle-box/box.ts
```diff
@@ -1,12 +1,12 @@
...
-const tmp = require("tmp");
-const path = require("path");
-const Config = require("truffle-config");
-const ora = require("ora");
-const fse = require("fs-extra");
-const inquirer = require("inquirer");
+import utils from "./lib/utils";
+import tmp from "tmp";
+import path from "path";
+import Config from "truffle-config";
```
gnidan: Huh, we don't actually use the typings here, do we?

packages/truffle-box/lib/utils/index.ts
```diff
@@ -51,12 +49,12 @@ module.exports = {
...
   },
 
-  unpackBox: async (tempDir, destination, boxConfig, unpackBoxOptions) => {
+  unpackBox: async (tempDir: string, destination: string, boxConfig: any, unpackBoxOptions: object) => {
     unbox.prepareToCopyFiles(tempDir, boxConfig);
     await unbox.copyTempIntoDestination(tempDir, destination, unpackBoxOptions);
   },
 
-  setUpBox: (boxConfig, destination) => {
+  setUpBox: (boxConfig: any, destination: string) => {
```
gnidan: (If you agree about making a `BoxConfig` interface, we can change these `any`s)

packages/truffle-box/box.ts
CruzMolina: We don't, but I think `tsc` does? Was yelling at me before I added the types.

packages/truffle-box/tsconfig.json
CruzMolina: Wouldn't that cause issues when using pkgs independently? 

Locally it appears `tsc` knows to recursively check node mod's for the types.

packages/box/tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+    "paths": {
+      "web3": [
+        "../../node_modules/@types/web3/index",
+        "node_modules/web3/index"
+      ]
+    },
+    "rootDir": ".",
+    "typeRoots": [
+      "./typings",
+      "node_modules/@types",
```
gnidan: (I still think we should list the ones we use individually)

CruzMolina: What was the reasoning for this again?

Using the monorepo, this should direct appropriately, using the pkg individually, this should also direct appropriately.

packages/truffle-box/box.ts
gnidan: Ah well, cool I guess!

packages/box/tsconfig.json
gnidan: I ran into a problem where `@types` dependencies for package A were breaking package B. Specifically, things like `lib` being different in different tsconfigs (like maybe package A specifies `dom.d.ts`, or somesuch)

gnidan: In fact, previously, the types were included wholesale like that, and I changed it to be explicit. Here's the PR https://github.com/trufflesuite/truffle/pull/1965

CruzMolina: Ah, this is probably a good reason to set things up using the Quramy example.

Issue #2211, opened 2019-07-12
Update in-test debugging infrastructure to use new compile interface


gnidan:

gnidan:
Already donezo

PR #2251, opened 2019-07-25
Emit method information for truffle-contract operations

gnidan:
Part of #2211 

Adds syntax to listen for method names and arguments.

Example syntax:
```javascript
const instance = await MyContract.deployed();
instance.myMethod().on("execute:send:method", ({ fn, abi, args, contract, address }) => {
  // access the abi item information, the arguments used, etc.
}
```

Note: this does not appear to directly conflict with #2035, but there may be reconciliation required later, once truffle-contract is instrumented for the event system.

### DIFF
--- a/packages/truffle-contract/lib/execute.js
+++ b/packages/truffle-contract/lib/execute.js
@@ -164,20 +164,26 @@ var execute = {
 
       constructor
         .detectNetwork()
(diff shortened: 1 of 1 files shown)


gnidan:
Thanks @CruzMolina!

PR #2252, opened 2019-07-25
Build mocha hook integration for truffle-debugger

gnidan:
Builds on #2251 #2073 #2255 #2260 

This defines the class `CLIDebugHook`, intended for use inside Mocha, to provide an interactive CLI debugger via test global interrupt.

- Define CLIDebugHook to wrap CLIDebugger
- Use #2251 work to capture transaction info for send, or to replay calls as tx

Issue #2206, opened 2019-07-12
Enable in-test breaking to use Truffle's debugger


gnidan:
**Epic**. Contains issues:
- [x] ~#2204 Support debugging estimateGas operations~ calling out of scope for now
- [x] #2212 Provide new truffle-compile interface
- [x] #2205 Support debugging contract creation operations

Related PRs:
- [x] #2252 Build mocha hook for truffle-debugger
- [x] #2253 Provide in-test debugging interrupts for Solidity inside JS tests
- [x] #2255 Switch debugger to use new compile interface
- [x] #2260 Emit deploy method information in truffle-contract-lifecycle

Issue #2205, opened 2019-07-12
Support debugging contract creation operations


gnidan:
This is part of epic #2206

PR #2253, opened 2019-07-25
Provide in-test debugging interrupts for Solidity inside JS tests

gnidan:
#2206 #2205 

Builds on #2252 #2254. Replaces #2050.

- Add `truffle test --debug` flag, to enable in-test debugging mode
- Add `debug()` global for use in JS tests, to indicate breakpoint for Solidity debugging

### DIFF
--- a/packages/truffle-core/lib/commands/test.js
+++ b/packages/truffle-core/lib/commands/test.js
@@ -6,11 +6,20 @@ const command = {
       describe: "Show all test logs",
       type: "boolean",
       default: false
+    },
+    "debug": {
+      describe: "Enable in-test debugging",
+      type: "boolean",
+      default: false
+    },
+    "debug-global": {
+      describe: "Specify debug global function name",
+      default: "debug"
     }
   },
   help: {
     usage:
-      "truffle test [<test_file>] [--compile-all] [--network <name>] [--verbose-rpc] [--show-events]",
+      "truffle test [<test_file>] [--compile-all] [--network <name>] [--verbose-rpc] [--show-events] [--debug] [--debug-global <identifier>]",
     options: [
       {
         option: "<test_file>",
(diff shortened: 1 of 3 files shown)


packages/truffle-core/lib/commands/test.js
```diff
@@ -66,8 +86,14 @@ const command = {
...
     }
 
-    let ipcDisconnect;
-    let files = [];
+    // enables in-test debug() interrupt, forcing compileAll
+    if (config.debug) {
+      config.compileAll = true;
+    }
+
+    var ipcDisconnect;
```
CruzMolina:
```suggestion
    let ipcDisconnect;
```

packages/truffle-core/lib/commands/test.js
```diff
@@ -66,8 +86,14 @@ const command = {
...
+
+    var files = [];
```
CruzMolina:
```suggestion
    let files = [];
```

CruzMolina: so you're the one throwing `var`'s everywhere :trollface:

packages/truffle-core/lib/test.js
```diff
@@ -105,15 +106,22 @@ const Test = {
...
+    });
 
     // Finally, run mocha.
     process.on("unhandledRejection", reason => {
       throw reason;
     });
 
     return new Promise(resolve => {
-      mocha.run(failures => {
+      this.mochaRunner = mocha.run(failures => {
```
CruzMolina: `this.mochaRunner` isn't block scoped here?

gnidan: I don't follow. What do you mean?

packages/truffle-core/lib/commands/test.js
gnidan: oh jeez I hope not. bad merge perhaps?

packages/truffle-core/lib/test.js
CruzMolina: `this.mochaRunner` is instantiated inside `Test.run`, but then called within `setJSTestGlobals.

seems like it would be `undefined` within `setJS...`

gnidan: Oh, I see; yes that is confusing.

It's fine because `this.mochaRunner` is accessed inside the `async operation => { ... }` closure, so lookup happens when `debug()` is called, v. likely giving enough time for `Test.run` to reach this line. Ugly stateful business, isn't it?

Any ideas on how to improve this?

CruzMolina: how about a comment?

PR #2254, opened 2019-07-25
Switch `truffle test` to use new compile interface

gnidan:
Implements #2211 
Builds on #2073 #2258 

- Use `"truffle-workflow-compile/new"` inside lib/test
- ~~Remove extraneous profiler usage (truffle-workflow-compile does this)~~ turns out this isn't so extraneous
- Capture entire compilation result instead of just dependencyPaths

### DIFF
--- a/packages/truffle-core/lib/test.js
+++ b/packages/truffle-core/lib/test.js
@@ -156,7 +156,7 @@ const Test = {
       all: config.compileAll === true,
       files: updated.concat(solidityTestFiles),
       resolver: testResolver,
-      quiet: false,
+      quiet: config.quiet,
       quietWrite: true
     });
 
--- a/packages/truffle-core/lib/test.js
+++ b/packages/truffle-core/lib/test.js
@@ -2,7 +2,6 @@ const Mocha = require("mocha");
 const chai = require("chai");
 const path = require("path");
 const Web3 = require("web3");
-const { promisify } = require("util");
 const Config = require("truffle-config");
 const Contracts = require("truffle-workflow-compile/new");
 const Resolver = require("truffle-resolver");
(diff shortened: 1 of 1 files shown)


packages/truffle-core/lib/test.js
```diff
@@ -144,36 +145,33 @@ const Test = {
...
-        // Compile project contracts and test contracts
-        Contracts.compile(compileConfig)
-          .then(result => {
-            const paths = result.outputs.solc;
-            accept(paths);
-          })
-          .catch(reject);
-      });
+    const updated =
+      (await promisify(Profiler.updated)(
```
eggplantzzz: No need for this `promisify`. `Profiler.updated` now has a dual Promise/callback interface. Check it [out](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle-compile/profiler/index.js#L19).

packages/truffle-core/lib/test.js
```diff
@@ -144,36 +145,33 @@ const Test = {
...
+        config.with({ resolver: testResolver })
+      )) || [];
+
+    const compileConfig = config.with({
+      all: config.compileAll === true,
+      files: updated.concat(solidityTestFiles),
+      resolver: testResolver,
+      quiet: false,
```
eggplantzzz: Don't we want this to come from the config? Wasn't this something changed recently by @CruzMolina?

packages/truffle-core/lib/test.js
```diff
@@ -101,7 +102,7 @@ const Test = {
     await this.defineSolidityTests(
       mocha,
       testContracts,
-      dependencyPaths,
+      compilations.solc.sourceIndexes,
```
eggplantzzz: Is this the same thing as `dependencyPaths` was??

packages/truffle-core/lib/test.js
gnidan: Hot diggity!

packages/truffle-core/lib/test.js
gnidan: Oh yeah you're right!

packages/truffle-core/lib/test.js
gnidan: Yep

eggplantzzz: Rad

gnidan:
Thanks!

PR #2255, opened 2019-07-25
Switch debugger to use new compile interface

gnidan:
Builds on #2073 

- Refactor lib/debug module structure / logging
- Support pre-computed compilation results, to skip forced compilation in debugger
- Use new compile interface when compilation is not provided
- Allow DebugInterpreter to start via async invocation

Note: this currently shims the new compile bytecode format (with external link references) back into the legacy format (inlined underscores). Updating the debugger to expect the new link reference format can be a later piece of work.


packages/truffle-core/lib/debug/interpreter.js
```diff
@@ -236,25 +236,22 @@ class DebugInterpreter {
   }
 
   start(terminate) {
-    let prompt;
+    if (terminate === undefined) {
+      return util.promisify(this.start.bind(this))();
```
haltman-at: Would you mind explaining what's going on here?

packages/truffle-core/lib/debug/compiler.js
```diff
@@ -15,19 +15,9 @@ class DebugCompiler {
...
     compileConfig.quiet = true;
 
-    // since `compile.all()` returns two results, we can't use promisify
-    // and are instead stuck with using an explicit Promise constructor
-    const { contracts, files } = await new Promise((accept, reject) => {
-      compile.all(compileConfig, (err, contracts, files) => {
-        if (err) {
-          return reject(err);
-        }
+    const { contracts, sourceIndexes } = await compile.all(compileConfig);
```
haltman-at: I'm putting this comment here since I can't put it on the lines you didn't change, but... as long as you're redoing this bit, this seems like a good time to make that change to using `config.with` that you suggested earlier instead of my manual-clone-and-alter?

(Also, "indexes"? :P )

packages/truffle-core/lib/debug/interpreter.js
gnidan: Sure: `terminate` is a callback that happens to kill the process. This change makes it so that if you don't pass the callback, it just recurses into a promisified version of itself, thus returning a Promise.

packages/truffle-core/lib/debug/compiler.js
gnidan: Recently @fainashalts discovered that `config.with` is actually a bit broken right now, so I think I'd rather leave it for now until we get to the bottom of it.

And "indexes" appears to be acceptable use here? Not sure.

gnidan:
Thanks @haltman-at!

PR #2256, opened 2019-07-26
Fix missing underscore dependency

spalladino:
File `quorum-overloads` from `interface-adapter` imported the underscore library, even though it was not declared on the package.json as a dependency. This caused a `require` of the package (or any other package that depended on it) to fail.

Instead of installing underscore, since other truffle packages already use lodash, this commit changes underscore to lodash, and declares it as a dependency on the package.json.


gnidan:
(I'll merge once it goes green)

spalladino:
What if it doesn't? :-P
(I'm on it)

gnidan:
One of these days I'll add a git hook to make sure this problem doesn't happen again...

For now, merging! Thanks again!

PR #2258, opened 2019-07-26
Add missing defaults for shimBytecode

gnidan:
For case where trying to shim bytecodes that are undefined


eggplantzzz:
Looks smoov👍🏽

gnidan:
Thanks!

PR #2260, opened 2019-07-28
Emit deploy method information in truffle-contract lifecycle

gnidan:
Required for #2205 
Continues along same reasoning as #2251 

In addition to the existing `execute:send:method` and `execute:call:method`, this adds a new event `execute:deploy:method`.

The purpose here is to allow #2252 and #2253 to also offer the behavior of debugging contract deployments in tests.

**Note** Future work might involve changing the names of these events, perhaps before they end up getting used externally too much.

### DIFF
--- a/packages/truffle-contract/lib/contract/constructorMethods.js
+++ b/packages/truffle-contract/lib/contract/constructorMethods.js
@@ -1,4 +1,3 @@
-const Web3PromiEvent = require("web3-core-promievent");
 const Web3Shim = require("truffle-interface-adapter").Web3Shim;
 const utils = require("../utils");
 const execute = require("../execute");
@@ -17,8 +16,6 @@ module.exports = Contract => ({
   },
 
   new() {
-    const promiEvent = new Web3PromiEvent();
-
     utils.checkProvider(this);
 
     if (!this.bytecode || this.bytecode === "0x") {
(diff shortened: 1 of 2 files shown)


packages/truffle-contract/lib/execute.js
```diff
@@ -113,19 +112,12 @@ var execute = {
...
-      }
-
-      params.to = address;
-      params = utils.merge(constructor.class_defaults, params);
-
-      constructor
-        .detectNetwork()
-        .then(async () => {
+      execute
+        .prepareCall(constructor, methodABI, args)
```
CruzMolina: Yay refactoring!

packages/truffle-contract/lib/contract/constructorMethods.js
```diff
@@ -31,23 +28,9 @@ module.exports = Contract => ({
...
-      onlyEmitReceipt: true
-    };
-
-    this.detectNetwork()
-      .then(({ blockLimit }) => {
-        utils.checkLibraries.apply(this);
-        return execute.deploy.call(this, args, context, blockLimit);
-      })
-      .catch(promiEvent.reject);
+    var constructorABI = this.abi.filter(i => i.type === "constructor")[0];
```
CruzMolina: Yay detailed commits! 👨‍💻

gnidan:
Thanks @CruzMolina!

Issue #1865, opened 2019-03-29
HDWalletProvider exits without warning if `http/https` is not provided in front of provider URL


mrwillis:
- [ x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

What the issue is, in broad strokes.

## Steps to Reproduce

Use `HDWalletProvider` 1.0.2 and omit the `https://` for the provider URL. You'll get

```
Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.
```

and then it exits without any error. Then include the https:// and it works fine. 

## Expected Behavior

What you expected to happen.

HDWalletProvider should throw an error if the prefix is not provided.

## Actual Results

What actually happened. Please give examples and support it with screenshots, copied output or error messages.

## Environment

* Operating System: Ubuntu 18
* Ethereum client: INFURA with a v3 URL (no private token auth)
* Truffle version (`truffle version`):  Truffle v5.0.8 (core: 5.0.8)
* node version (`node --version`): v10.5.0
* npm version (`npm --version`): 6.7.0

eggplantzzz:
@mrwillis Thanks for the heads up! It seems like a good suggestions. We'll check it out.

briankostar:
Yes this would have been useful. Spent a few hours scratching my head as there's no errors either.

cds-amal:
PR #2121 resolved this.

PR #2274, opened 2019-08-01
Ensure that HDWalletProvider will throw an error when the url provider is without slash / slashes

djuanit0x:
#1865


gnidan:
Oh oops @cds-amal sorry to steal your thunder with the review. I just verified locally that `url.slashes` behaves as expected. Amal let me know if there's anything I missed and I can fix it.

PR #2282, opened 2019-08-06
Dependency update: Upgrade web3 packages to 1.2.1

gnidan:
Upgrades dependencies:
- web3
- web3-core-promievent
- web3-eth-abi
- web3-utils

Note: I've removed the carets (^) because I think we should keep these pinned at specific versions.

cc @nivida


gnidan:
Thanks Cruz!

PR #2289, opened 2019-08-12
Internal improvement: Add --log option to develop command file

eggplantzzz:
Document the previously hidden `--log` option

### DIFF
--- a/packages/truffle-core/lib/commands/develop.js
+++ b/packages/truffle-core/lib/commands/develop.js
@@ -15,7 +15,10 @@ const command = {
     options: [
       {
         option: `--log`,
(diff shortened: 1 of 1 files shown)


packages/truffle-core/lib/commands/develop.js
```diff
@@ -11,8 +11,13 @@ const command = {
     }
   },
   help: {
-    usage: "truffle develop",
-    options: []
+    usage: "truffle develop [--log]",
+    options: [
+      {
+        option: `--log`,
+        description: `Start/Connect to a Truffle develop session and log all rpc activity.`
```
gnidan: Is it too much trouble to state explicitly that there's no repl?

Issue #2288, opened 2019-08-12
ora dependency skips the logger


tcoulter:
We're using the `ora` dependency to show a nice spinner. Examples [here](https://github.com/trufflesuite/truffle/search?q=ora&unscoped_q=ora). 

`ora` goes straight to the console and skips any configured logger (e.g., `Config.logger`). In fact, it doesn't look like the logger configuration is even available in many places `ora` is used. This breaks the logger abstraction and prevents sending all output through a consumable interface. 

`ora` has a [stream](https://github.com/sindresorhus/ora#stream) feature, which may let us send its data through the same pipe. But this requires extending the logger abstraction to support streams (not a bad idea) or abstracting a stream sent to `ora` so it fits the logger abstraction (e.g., just log lines). 

Ultimately, I wouldn't be surprised if we ditch the logger abstraction entirely, requiring the TTY to take care of it, but that would make Truffle harder to run in more abstract places (like the browser). 

Happy to dive into more detail if needed. Cheers!

eggplantzzz:
Hey @tcoulter, we should be able to resolve this after the new event system is out. #2199

tcoulter:
That's what I've heard! 👏

PR #2290, opened 2019-08-12
[WIP] Truffle Live POC: 100% in-browser Truffle + Ganache

tcoulter:
**[DEMO!](https://tcoulter.github.io/truffle-live-demo/index.html)** <--- click here

------

The link above is a demo of two new packages: `browser-truffle` and `truffle-live`. 

- `browser-truffle` is a webpacked, browser version of Truffle that uses [BrowserFS](https://github.com/jvilk/BrowserFS) for mimicking node's `fs` API in LocalStorage.
- `truffle-live` is the demo above. It includes a simple app and a `TruffleTerminal` react component using using [this](https://github.com/rohanchandra/react-terminal-component) terminal emulator (more on this in a bit).

I'll describe both below and then follow up with some general thoughts on where to go next.

# Browser Truffle

All the magic features of `browser-truffle` all exist within the webpack configuration. Because we're basically packing `truffle` all over again for a different context, `browser-truffle` uses `truffle`'s webpack configuration as a base and then adds to it and extends it. Much of the webpack configuration is there to support BrowserFS or to add a lot of shims; there's a lot of functionality the browser doesn't support, so we shim out those features in various ways, either replacing features with equivalent functionality or gutting it all together. **Not everything is shimmed**. This was a weekend passion project and I desperately need some game time, so expect dragons in many places. The following is a list of commands I've gotten to work for at least basic cases: 

- `truffle init`
- `truffle compile`
- `truffle migrate`
- `truffle version`

There were some caveats to implementation of all the above: 

### truffle init

This version of `init` is shimmed from the main `truffle` init command. Instead of doing any box downloading or installing, it simply writes out hardcoded files copied into `browser-init.js`. Lots more work can definitely be done here, like figuring out how to download a box and install it, as well as pre-filling the filesystem for specific examples so init doesn't need to be called.  

### truffle compile

This was one of my favorites, though it has oddities, and you can mostly thank `solc` for that. First, the soljson files downloaded by solc are so large they can't be consumed by webpack. Second, `solc` wants you to use those files in the browser by placing them in a `<script/>` tag, and then that script, when run, will set the (you guessed it) `window.Module` variable. Actually, that was a joke, I'm pretty sure you didn't guess it. But that's exactly the point. The solution I settled with was to create a new Browser compiler supplier that leverages the existing download functionality, then "caches" the downloaded file in a `<script/>` tag. To consume the downloaded solidity, the compiler supplier returns `window.Module`, just expecting it to exist. This is a bit terrible, but I'm not sure there's much better. The downsides of this implementation are 1) the global variable, and 2) I'm currently not removing collisions on the `window.Module` variable, which means there can only be one version of solidity "cached" at any one point in time. Theoretically, we could remove those collisions pretty easily and then manage the downloaded solidity files in memory after they're parsed, effectively caching them as the CompilerSupplier intended. 

Oh, and I forgot: The reason I even took the route above was because LocalStorage has a maximum space of 10Mb. You should have seen the disappointment on my face when I couldn't let what was already built handle the downloading and installing. 

### truffle migrate

Honestly, this was one of the easiest barring a few nasty bugs. Ganache(-core) bundled pretty cleanly, and it uses the file system for data and that was already taken care of by the BrowserFS bundling. I had to make a few runtime hacks in `browser-truffle.js` to support some API differences, but all in all it works as you might expect.

### truffle version

Unsurprisingly, I did nothing special here. It "just worked" the first time I tried it. :) 

# Truffle Live

As mentioned above, the `truffle-live` package and the `TruffleTerminal` react component uses the [react-terminal-component library](https://github.com/rohanchandra/react-terminal-component) for displaying and running the terminal. **This is not a real terminal.** It just looks like one. Originally, I thought this would be the easiest path toward getting something that looks and feels and acts like a terminal (and who knows, I might have been right) but in the end it's pretty finicky. For instance, it doesn't support long running processes out of the box, and adding programmatic output to the terminal is much harder than you'd expect (_"where's console.log() g**-d@#$@#$!"_). But it ultimately worked, and it allowed me to easily add other commands like `ls`, `cd`, `cat`, etc. for inspecting the filesystem. Though I think using it was generally a win for a POC, it has some definite downsides (all of them are in user experience; it only looks and feels and acts like a terminal at first glance -- users like @gnidan will be let down). If I were to put more time into the UI, I'd integrate [xterm.js](https://xtermjs.org/), though it _looks_ like that'd take significantly more work to get looking pretty. Also, it wasn't immediately obvious how to integrate xterm into React. Such is life, I guess. 

One interesting thing about the way `TruffleTerminal` is built is that it allows me to intercept a command before the command exists. This lets me download `browser-truffle.js` as it's requested, and not as the page is loading. This significantly reduces page load time as the initial payload isn't 10+ Mb. One thing to note is that the server configuration of Github pages (the demo above) prevents seeing progress when downloading the `browser-truffle.js` binary. If we controlled the server, like we would on an official release of this thing, the download progress would be shown to the user while they waited.

# Notes on some Specific Commits

The following are some commits that I'd generally like to point out, or changes I made to core libraries to support the browser. 

- [4a7730f](https://github.com/trufflesuite/truffle/commit/4a7730f9494aadf19c38d21d1d816a90f1687541) - There was a variable in `truffle-contract` declared but not declared with (something like) `var`. This threw my bundle for a loop, and was extremely hard to debug. I added the `var` declaration and everything seems fine. 
- [d0ca9d7](https://github.com/trufflesuite/truffle/commit/d0ca9d7b21348c486063154445ccac296bedf122) - Here we were detecting the types of filesystem errors using the error message. This proved to be problematic in that my fs shim library (BrowserFS) had a slightly different case on the error message, making detection fail. Turns out the right way to do it is to check the error code instead. Who knew? 
- [16bed8f](https://github.com/trufflesuite/truffle/commit/16bed8fd14deb0851b6755871b42a11757c5b9cf) - This was a minor cleanup commit, in that we should detect configuration before doing any work, as the work may not be necessary if a config object isn't present.
- [e4df761](https://github.com/trufflesuite/truffle/commit/e4df7612161a499095e54cbdff715db554871e7f) - This was a doozy. Turns out the `Config.getUserConfig()` call actually writes a file when the file doesn't exist. This file was being written on initialization (e.g., when `require`'d) which caused issues in the browser. I moved this call to `getOrGenerateMnemonic` which seemed like the write time to write the file if it didn't exist.
- [d0e7cc6](https://github.com/trufflesuite/truffle/commit/d0e7cc675cef91421a01b7f30978bfeead9320e2) - I meant to do this one individually, but screwed up. First, I added the initial Browser compiler supplier here -- ignore it, because I changed it heavily later. Second though, was a change to using `axios` and `findup-sync`. `axios` bundles better than `request-promise`, and the original `findup` didn't work within the browser filesystem for reasons I can't discern. Both seem to work just fine in a node context as well as the browser, so those should be better choices. Oh and [this commit](https://github.com/trufflesuite/truffle/commit/e5577fe65754e716615b6bb7b9b8d84ce4f7747e) should be included here too as its an artifact of switching to `axios`. 
-  [579906e](https://github.com/trufflesuite/truffle/commit/579906e88da3a102641c4074594b3070bffe94be) - Ahh! Our relayer breaks cors! I had do default to ethereum.org in order for Solidity to be downloaded correctly.
- [0b12116](https://github.com/trufflesuite/truffle/commit/0b12116ab39de77abd4f2678ca4ce1be1242d128) - There's an official `readline` library in Node 10, which breaks bundling. They've moved the package to `linebyline`, which does the exact same thing.

# Outstanding issues and thoughts on future work

The following is a list of outstanding issues and where the code can be improved. This is definitely a POC, and there's still a lot of work that needs to be done. 

- **Missing output:** There are a lot of libraries we use that skip the logger. That, or we decide not to use the logger entirely.  For instance, any time we use [ora](https://github.com/trufflesuite/truffle/issues/2288) we skip the logger, and go straight to the `console`. This makes it hard for the browser version to intercept that logging, so a lot of it is lost. There are two things that might fix this. One, the event system I've been told about to unifying console output into on place. Or two, we could use a real TTY library like [xterm](https://xtermjs.org/) in the browser. I'm not sure which is more feasible, but likely a bit of both might do the trick.
- **`truffle help` doesn't work:** Because of the above, `truffle help` doesn't work because the help text is output by `yargs`. The same is true for `truffle opcode` too actually, as the opcodes aren't printed through the logger.
- **`truffle test` doesn't work:** I ran out of time. I'm sure there's a lot of work we can do to support testing. To do it, we'd have to not only bundle `mocha` the right way, but also copy over all the testing solidity files so the browser version can have them at runtime. 
- **No support for `truffle console` or `truffle develop`.** Currently both of these commands use the `repl` library. I'm not sure how to recreate that for the browser, and it's highly likely it'll take some interaction with the actual terminal, assuming we don't use a real TTY library.
- **You're eventually gonna run out of space.** All files are saved in local storage. This includes the project code as well as the associated chain data. Local storage isn't automatically cleared. If you use the demo enough, you'll eventually run out of local storage space.
- **But we can fix that if we let users manage it!** One of the main reasons you'll run out of space is because a new project and chain is being created for you each time. It should be relatively easy to use the same state and filesystem when the user comes back to the page. 
- **The current example could be great for interactive examples.** Imagine having documentation that lets you use the commands on the fly with no install, with a default filesystem and project in place. Wouldn't that be awesome?
- **But it can get even better.** Imagine having a GUI filesystem viewer and text editor, and some light, GanacheUI-like code to see all the transactions. This would blow remix out of the water, and would be a huge win for hackathons and example projects.
- **Oh, and get's better than that.** We could even have people share projects by making Truffle Live into a Truffle Fiddle. :boom: Using Truffle in the browser just got way more exciting!

I'm sure there's more things I wanted to discuss here, but I'm pooped. 

Feel free to ask me any questions! Oh, and I meant to say: I wouldn't be surprised if there's a host of atrocities in this code. I tried to make it as clean as possible, but I was also moving fast trying to build something I wasn't sure was doable. I'm super happy with the result regardless, but I wouldn't be surprised if there was more cleanup to do.

Cheers! :wave:


tcoulter:
Another afterthought: 

- The `ls`, `cd`, `pwd`, etc. commands should probably use ShellJS rather than be written from scratch: https://github.com/shelljs/shelljs

eggplantzzz:
Closing this for issue maintenance. Thanks for the POC! We'll revisit this when we get it on the roadmap.

PR #2296, opened 2019-08-14
Bump ganache-core version

haltman-at:
This pull request bumps the `ganache-core` version in the various `package.json` files from `2.5.7` to `^2.6.1`.  (I'm hoping there wasn't some reason it was fixed at a specific version, no caret, before...)

The specific reason I'm doing this is so that later we can finally enable the globally-available variable tests in the debugger, but I figured I'd do it in all the packages because why not.

### DIFF
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -18,7 +18,7 @@
   "devDependencies": {
     "chai": "4.2.0",
     "debug": "^4.1.1",
-    "ganache-core": "2.5.7",
+    "ganache-core": "^2.6.1",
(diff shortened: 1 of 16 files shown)


haltman-at:
OK, done.

haltman-at:
Um, 2.7.0 seems to be causing Travis problems.  I'll going to go back down to 2.6.1 and pin, I guess? :-/  (And then increase the timeout on that one failing test.)

haltman-at:
OK, @davidmurdoch says he's going to take a look at what's causing the problems.  So maybe we'll end up going to 2.7.1 instead? :P

PR #2302, opened 2019-08-15
Abstract the config command to read project-level parameters

seesemichaelj:
This PR modifies the `truffle config` command behavior to be similar to other config-getting/setting CLIs.

This PR maintains the previous behavior to prevent a breaking change:
- `truffle config` will prompt the user if they'd like to enable/disable analytics
- `truffle config --(en|dis)able-analytics` will enable/disable user-level analytics

The second flags are now deprecated in favor of the new functionality (i.e. `truffle config set analytics <true|false>`). A message to the user has been added to tell them to use the new syntax.

This PR changes the `Help Usage` to be `truffle config <get|set> <key> [<value-for-set>]`. Here is a run down of each of the different commands:
~- `enable` is an alias for `set` with an implied `value-for-set` to be `true`~
~- `disable` is an alias for `set` with an implied `value-for-set` to be `false`~
- `get` retrieves the value of a user-level setting (which is denoted in `userLevelSettings` in `commands/config.js`) or project-level setting after `truffle-config` loads the project. It logs the value to `options.logger`
~- `read` is an alias for `get`~
- `set` sets the value for a user-level setting. Project-level setting cannot be easily done with the current infrastructure (and outside the scope of this PR) to modify the `truffle(-config)?.js` file appropriately
~- `write` is an alias for `set`~

This enables you to do things like `truffle config get migrations_directory` in the truffle project's directory.

### DIFF
--- a/packages/core/lib/commands/config.js
+++ b/packages/core/lib/commands/config.js
@@ -203,8 +203,8 @@ const command = {
             options.logger.log(config[command.key]);
           }
         })
-        .catch(options.logger.logs)
-        .finally(done);
+        .then(done)
+        .catch(options.logger.logs);
     }
   }
 };
--- a/packages/core/lib/commands/config.js
+++ b/packages/core/lib/commands/config.js
@@ -1,107 +1,3 @@
-const parse = function(args) {
-  if (args.length === 0) {
-    return null;
-  }
-
-  let command = args[0];
-
-  if (typeof command !== "string") {
-    // invalid command
-    throw new Error(`Invalid config option "${command}"`);
-  }
-  command = command.toLowerCase();
-
-  let set = false;
-  let key = args[1];
-  let value = args[2];
-
-  switch (command) {
-    case "enable": {
-      set = true;
-      value = true;
-      break;
-    }
-    case "disable": {
-      set = true;
-      value = false;
-      break;
-    }
-    case "get":
-    case "read": {
-      set = false;
-      if (typeof key === "undefined" || key === null || key === "") {
-        // invalid key
-        throw new Error("Must provide a <key>");
-      }
-
-      break;
-    }
-    case "set":
-    case "write": {
-      set = true;
-      if (typeof key === "undefined" || key === null || key === "") {
-        // invalid key
-        throw new Error("Must provide a <key>");
-      }
-
-      if (typeof value !== "string" || value === "") {
-        // invalid value
-        throw new Error("Must provide a <value-for-set>");
-      }
-
-      switch (value.toLowerCase()) {
-        case "null": {
-          value = null;
-          break;
-        }
-        case "undefined": {
-          value = undefined;
-          break;
-        }
-        case "true": {
-          value = true;
-          break;
-        }
-        case "false": {
-          value = false;
-          break;
-        }
-        default: {
-          // check if number, otherwise leave as string
-          const float = parseFloat(value);
-          if (!isNaN(float) && value === float.toString()) {
-            value = float;
-          }
-          break;
-        }
-      }
-
-      break;
-    }
-    default: {
-      if (
-        command !== "--enable-analytics" &&
-        command !== "--disable-analytics" &&
-        command !== ""
-      ) {
-        // TODO: Deprecate the --(en|dis)able-analytics flag in favor for `enable analytics`
-        // invalid command!
-        throw new Error(`Invalid config option "${command}"`);
-      } else {
-        // we should not have gotten here
-        return null;
-      }
-    }
-  }
-
-  return {
-    set,
-    userLevel: module.exports.userLevelSettings.includes(key),
-    key,
-    value
-  };
-};
-
 const command = {
   command: "config",
   description: "Set user-level configuration options",
@@ -190,7 +86,7 @@ const command = {
       done();
     } else {
       Promise.resolve()
-        .then(async () => {
+        .then(() => {
           const config = Config.detect(options);
 
           if (command.set) {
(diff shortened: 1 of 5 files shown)


eggplantzzz:
How do you feel about prefixing all of the commands with `--`? In general it seems to be a pattern with Truffle to do `truffle <command> --<someOption> <specificValue>`.

For example, `truffle migrate --network candyLand` or `truffle compile --list releases`.

packages/truffle-core/lib/commands/config.js
```diff
@@ -1,8 +1,113 @@
...
+      } else {
+        // we should not have gotten here
+        return null;
+      }
+    }
+  }
+
+  return {
+    set,
+    userLevel: module.exports.userLevelSettings.includes(key),
```
eggplantzzz: Why do you access this using `module.exports` and not just `command`?

packages/truffle-core/lib/commands/config.js
```diff
@@ -12,27 +117,94 @@ const command = {
         option: "--disable-analytics",
         description:
           "Disable Truffle's ability to send usage data to Google Analytics"
+      },
+      {
+        option: "enable",
+        discription: "Enable a Truffle Attribute"
```
eggplantzzz: Should attribute be capitalized in all these cases?

packages/truffle-core/lib/commands/config.js
```diff
@@ -12,27 +117,94 @@ const command = {
...
+              "Setting project-level parameters is not supported yet."
+            );
+            // TODO: add support for writing project-level settings to the truffle config file
+            // config[command.key] = command.value;
+          } else {
+            options.logger.log(config[command.key]);
+          }
+        })
+        .then(done)
+        .catch(options.logger.logs);
```
eggplantzzz: Is `logs` a typo?

packages/truffle-core/lib/commands/config.js
eggplantzzz: Why not just `command` rather than `module.exports`? In the future I wonder if this `parse` function should be pulled out into its own utility, and then this would have to be passed in somehow. I'm in favor of these specific command files being kept as lean as possible and this file is just starting to get a bit large.

packages/truffle-core/lib/commands/config.js
```diff
@@ -12,27 +117,94 @@ const command = {
...
+            options.logger.log(googleAnalytics.getAnalytics());
+          }
+          break;
+        }
+      }
+
+      done();
+    } else {
+      Promise.resolve()
+        .then(async () => {
```
eggplantzzz: This doesn't need to be `async`? Unless we just want to do it on principle?

packages/truffle-core/lib/commands/config.js
seesemichaelj: It is

packages/truffle-core/lib/commands/config.js
seesemichaelj: :man_shrugging: Y'all can change it to whatever you want

packages/truffle-core/lib/commands/config.js
seesemichaelj: Ah, `parse` was in the `command` so I had `module.exports` to reference itself. I can change this back to `command` now that I moved `parse` into its own function

I agree the file was getting a bit large. There wasn't a pattern yet for a multi-file `commands/*.js` command yet, so I kind of was like "ehhh, they can figure out when/if they want to change the architecture".

packages/truffle-core/lib/commands/config.js
seesemichaelj: No it doesn't anymore. I had copied this chunk from `commands/debug.js` but then removed the `await Environment.detect(...)` without removing the `async`

seesemichaelj:
> How do you feel about prefixing all of the commands with `--`? In general it seems to be a pattern with Truffle to do `truffle <command> --<someOption> <specificValue>`.
>
> For example, `truffle migrate --network candyLand` or `truffle compile --list releases`.

I'm all for `--` in general. In this specific case I'm actually favoring the proposed mechanism as it feels more intuitive. I'm find with whatever yall think is best as I just care to have the functionality and hacked this together

> UPDATE: Oh, this will be a totally new pattern for Truffle since there are three parts to setting something: `truffle <verb> <key> <value>`. Would it be a bad idea to try and standardize a pattern for these 3-parters and keep the current pattern for the 2-parters?

I don't think it's a bad idea to prevent technical debt. I was trying to add some functionality quick, and probably don't have the capacity to figure out what would work out better.

> In other words, it would be `truffle --get analytics` but something else for `truffle set analytics true`. Or just some other idea: we could do something like `truffle --set analytics=true`? Or `truffle --set-analytics true`? How many of these options are we expecting? Will it get out of hand if we try and do this last one?

I'm very against `--set-analytics` specifically as it doesn't scale as you hinted at. I'd much prefer `--set analytics true` over `--set analytics=true`. Although I don't have much stake in the game, as I'll do whatever you guys make the interface to be. I think the `truffle --get|set` pattern is even more out of the box for `truffle`. We don't even have a `truffle --version` (it's `truffle version` to be consistent with the `truffle <command>` structure).

> You probably already have a couple of examples of other CLI's that do this kind of thing. Do you happen to have any examples offhand?

[npm](https://docs.npmjs.com/cli/config) follows the proposed pattern, [yarn](https://yarnpkg.com/lang/en/docs/cli/config/) follows the proposed pattern, [git](https://git-scm.com/docs/git-config) is similar, but it removes the `get/set/enable/etc` in favor of "if there is a 4th argument, it's a set, if there isn't, it's a get (see https://git-scm.com/docs/git-config#EXAMPLES) with an option of `--get`

----

I will add that this is a blocker for Teams, and depending on how long this may take for a release (for review, test, and ship), I may just duplicate this code into a Truffle Plugin

packages/truffle-core/lib/commands/config.js
eggplantzzz: Hehe yeah I think the next time this happens I'm going to just give it a folder and include helper files in there. I just wanted to note this here, I'm not going to fight now to create that architecture :) Thanks Mike!

packages/truffle-core/lib/commands/config.js
eggplantzzz: @gnidan Can you weigh in on this one? What do you think about for all these we call them "configuration options". So for example, "Enable a Truffle configuration option"? Does that sound too robotic?

packages/truffle-core/lib/commands/config.js
seesemichaelj: Fixed: https://github.com/trufflesuite/truffle/pull/2302/commits/96a54f2e50ec8e0eafb38dfb1737c6c3562bf314#diff-7deba4495724b2fd4f390796fccbf054R89

packages/truffle-core/lib/commands/config.js
seesemichaelj: Fixed using `command` instead of `module.exports` in https://github.com/trufflesuite/truffle/pull/2302/commits/96a54f2e50ec8e0eafb38dfb1737c6c3562bf314

seesemichaelj:
I'm no longer blocked by this as I'll be taking another approach with the `truffle migrate --describe` that @gnidan and I have chatted about (that I'm going to implement)

I do still think this PR is a good addition, and I'm willing to support getting this merged

eggplantzzz:
So are we good to give this guy a merge?

seesemichaelj:
:man_shrugging: up to yall if you want to change any particulars, but I'm good when you're good

eggplantzzz:
Screen Shot 2019-09-13 at 3 23 00 PM
Here is a screenshot of the deprecation warning for `--enable-analytics` and `--disable-analytics`.

PR #2304, opened 2019-08-15
Bump solc version to 0.4.26

eggplantzzz:
Co-authored-by: barakman barakman@yahoo.com

### DIFF
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -18,7 +18,7 @@
     "ganache-cli": "6.1.8",
     "mocha": "5.2.0",
     "require-nocache": "^1.0.0",
-    "solc": "0.4.25",
+    "solc": "0.4.26",
     "temp": "^0.8.3",
     "truffle-blockchain-utils": "^0.0.5",
     "web3": "0.20.6"
--- a/packages/truffle-artifactor/package.json
+++ b/packages/truffle-artifactor/package.json
@@ -27,7 +27,7 @@
     "async": "2.6.1",
     "debug": "^3.1.0",
     "fs-extra": "6.0.1",
-    "lodash": "4.17.10",
+    "lodash": "^4.17.12",
     "truffle-contract": "^3.0.7",
     "truffle-contract-schema": "^2.0.2",
     "truffle-expect": "^0.0.4"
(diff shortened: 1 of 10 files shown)


eggplantzzz:
If all goes well we will try and get this merged and released in the next week or so.

eggplantzzz:
When I pull in branch `v4` and run the test suite, it looks like it fails at the same place that the CI does.

barakman:
@eggplantzzz:

Hi.

Is this out any time soon?

Also, an I please ask that while you're at it, upgrade the `lodash` dependency to version 4.17.12 or higher?

The reason for this is a security vulnerability, as described in [this snyc report](https://snyk.io/vuln/SNYK-JS-LODASH-450202).

We would otherwise need to upgrade to Truffle 5.x, which is what we've trying to avoid (as well as the reason for this PR) in the first place.

Thanks again :)

eggplantzzz:
Hey @barakman, I think this will go out this week. Sure we can bump the lodash version while this PR is up. Just to make sure you understand though, going forward we will be discontinuing support for version 4 of Truffle. This release was considered a special case though :)

barakman:
@eggplantzzz:

Sure, I perfectly understand that.
Since there are no more upgrades on the Solc 0.4 branch, I don't see a reason why there should be any upgrades on the Truffle 4 branch (in fact, I requested this upgrade specifically for Solc 0.4.26).

I admit that I've recently alerted the Solc guy (chriseth) of a small compiler issue (see [here](https://github.com/ethereum/solidity/issues/7303)), which he said should be fixed on the next release.

But I doubt that it would be applied on the 0.4 branch (i.e., Solc 0.4.27).

Thanks again for all your help :)

barakman:
@CruzMolina:
I've just noticed this morning that v4.1.16 is available (officially released on npm I suppose).
So first of all - thank you! We have updated our repository to use the new version.
Second - I see (above) that your CI tests fail. Is that something that we should worry about?
Thanks

CruzMolina:
Hey @barakman , I believe there's a couple diff issues with the v4 CI build that _should not_ indicate any issues with the bundle itself. Def let us know if any weirdness comes up between 4.1.15 & 4.1.16.

PR #2306, opened 2019-08-15
Whitelist specific branches for Travis push builds

gnidan:
We don't really need to run two builds for every PR.

This PR should tell Travis to run only the PR build for normal PRs, running the "push build" only on branches that will be the recipient of pull requests / other merges.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,17 @@ dist: trusty
 sudo: required
 language: node_js
 
+# configure Travis to run on-push builds only on PR-recipient branches
+branches:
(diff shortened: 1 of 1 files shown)


gnidan:
There is a possible downside here, which is that our current setup does two builds:
- it builds and tests the branch itself (this is the push build), and
- it builds and tests the target branch post-merge (this is the PR build)

So if we remove the push build, we lose the pre-merge validation, which is of some value. But I think not enough value to make up for the possibility of getting the Travis containers out of the queue!

PR #2314, opened 2019-08-19
Bug fix: Update solc version getting

eggplantzzz:
Since solc has been basically removed from the Truffle bundle, it does not make sense to try and get Solidity version information from that package (only a portion of which remains in the Truffle bundle). Therefore, use the compilerSupplier to determine what version of Solidity is being used in Truffle.

### DIFF
--- a/packages/truffle-core/lib/commands/version.js
+++ b/packages/truffle-core/lib/commands/version.js
@@ -17,7 +17,7 @@ const command = {
     } catch (error) {
       // Suppress error when truffle can't find a config
       if (error.message === "Could not find suitable configuration file.") {
-        config = null;
+        config = Config.default();
       } else {
         return done(error);
       }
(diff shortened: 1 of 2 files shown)


gnidan:
Marking this "needs requirements" because there a couple edge cases here that I feel we should not gloss over...

Here's what comes to mind:
- How do caret versions behave if they're the latest matching version?
- How do caret versions behave if they're not the latest matching version, and we don't have the latest in the cache?
- How do caret versions behave if they're not the latest matching version, and we do have the latest already?

It looks like this doesn't attempt to resolve the caret at all, is that correct? Is that behavior fine? What do you think?

(Just posing the questions)

eggplantzzz:
@gnidan I remember we had this discussion the last time `truffle-core/lib/version` had a major overhaul...back in December I believe.

So this PR basically maintains the current behavior. I guess we're back to the old question of how robust do we want to make this method at the expense of requiring that a user has an internet connection to make the request for version information? We could try and make the request and if it fails use the truffle config as a fallback? It might be nice to have some sort of marker in this case when it is from the truffle config. Then again, I'm not sure how much we gain from having this method be incredibly robust; so far the current behavior has been sufficient in my view.

PR #2316, opened 2019-08-20
Update run command

eggplantzzz:
This PR updates the run command. A decision was made to bundle truffle-plugin-solhint by default. This adds a "whitelisted plugins" object that will contain what I imagine will be some default plugins. The user will then be able to load/use them without having to complete any extra configuration.


cgewecke:
@CruzMolina 
@eggplantzzz 
@gnidan 

I'm a little concerned about the way whitelisted plugins are going to be bundled with Truffle. In this case it looks they're going into `truffle-core` as a dependency and will be webpacked on publication. Doesn't this mean `solhint` will  lose all data about which projects install it and how many people are downloading it?  

Given truffle's dominance as an Ethereum development platform it's likely their own NPM numbers and "used-by" stats *will drop rapidly* with this addition, even though actually more people are using it than ever. 

If this happened to any OSS repos I help maintain I'd be a little demoralized tbh. That data is one of the ways developers translate their volunteer work into grant funding and telegraph the usefulness of their projects to other sources of support. Normally when someone writes a module that's good enough to get included in another major project, its Github and NPM metrics will meaningfully reflect that (rather than precipitously deteriorate.) 

How difficult would it be for whitelisted plugins to go in the [main Truffle repo package](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle/package.json#L28-L32) as external dependencies like `mocha`?

eggplantzzz:
@cgewecke Thanks for the concern. @gnidan was the one who arranged for this job with the solhint people. I think he said he will get back to you with some thoughts on it a bit later!

gnidan:
@cgewecke so for background, this came out of a conversation I had with @pablofullana and @fvictorio, where we discussed how cool it'd be for Truffle to include a default set of plugins, like to support solhint. (think babel's `preset-env`)

I'm interested in this because it'd be nice to move parts of Truffle out of the core codebase and into "core plugins". And solhint is cool and widely used, so it seems reasonable to start to test the architectural waters with that.

I agree 100% about the metrics, as well as other downsides (like bundle size). Don't worry, this won't get merged until we figure out a better approach!

cgewecke:
> I'm interested in this because it'd be nice to move parts of Truffle out of the core codebase and into "core plugins". And solhint is cool and widely used, so it seems reasonable to start to test the architectural waters with that.

@gnidan I'm having a bit of "commenter's remorse" today. I certainly don't want to delay this PR and what you've said here makes perfect sense. ....

eggplantzzz:
Closing this until we have a more general solution.

pablofullana:
hey @eggplantzzz, hello there. any updates on this?

eggplantzzz:
Hi @pablofullana, not yet. We've kind of set this aside for a bit but have plans to pick it up in the future!

Issue #2318, opened 2019-08-21
Truffle repl: ignore `truffle` keyword


DavidBurela:
## suggestion
If in `truffle dev` or `truffle console`.  Please ignore `truffle` if it is the first command.

## use case 1: context switching
Many times I'm in a loop of using `truffle test` or `truffle migrate --reset` while developing.
I then drop into `truffle console`, to test things with the repl, but muscle memory gets me to type `TRUFFLE test` again.

## use case 2: new users
When I am teaching new users I usually show them `truffle test`, and get them to use it as part of their dev loop. When they complain that the tests take too long to run, I show them `truffle dev` and how it keeps truffle open, and they can then just use `test` to rapidly execute the tests.
Again due to muscle memory they type `truffle test` and question why it fails.

gnidan:
Thanks @DavidBurela! Looks like this will likely go out next week.

eggplantzzz:
Released into the wild with `5.0.34`!

PR #2320, opened 2019-08-21
Enhancement: Ignore keyword truffle in console input

eggplantzzz:
Add processReplInput method which filters out the word truffle when it is the first keyword in console input. This work was inspired by https://github.com/trufflesuite/truffle/issues/2318

### DIFF
--- a/packages/truffle-core/lib/repl.js
+++ b/packages/truffle-core/lib/repl.js
@@ -113,9 +113,25 @@ ReplManager.prototype.stop = function(callback) {
   }
 };
 
-ReplManager.prototype.interpret = function(cmd, context, filename, callback) {
-  var currentContext = this.contexts[this.contexts.length - 1];
-  currentContext.interpreter(cmd, context, filename, callback);
+ReplManager.prototype.interpret = function(
+  replInput,
+  context,
+  filename,
+  callback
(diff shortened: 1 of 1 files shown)


packages/truffle-core/lib/repl.js
```diff
@@ -113,9 +113,25 @@ ReplManager.prototype.stop = function(callback) {
...
-ReplManager.prototype.interpret = function(cmd, context, filename, callback) {
-  var currentContext = this.contexts[this.contexts.length - 1];
-  currentContext.interpreter(cmd, context, filename, callback);
+ReplManager.prototype.interpret = function(
+  replInput,
+  context,
+  filename,
+  callback
+) {
+  const processedReplInput = processReplInput(replInput);
```
gnidan: Note that this approach means you can type `truffle n` to step next in the debugger, since it reuses the same ReplManager.

Not sure if this matters, but if it does, we could move this logic to `lib/console` instead. I'll think on it a bit.

eggplantzzz: Ah ok, I didn't realize they both used this. And yes, maybe you are right that it belongs in `lib/console` if that is unused by the debugger.

PR #2323, opened 2019-08-21
Update ABI utilities to account for internalType

haltman-at:
This PR updates `abi.ts` by adding `internalType` to the ABI type; and it updates `definition2abi.ts` to generate this.  To find the `internalType` from the definition, it simply takes the `typeString` and then strips off any location that may be included at the end.

Since we're now making more use of type strings, I updated the hacky case of the `keyDefinition` function in `definition.ts` to spoof the full `typeDescriptions` rather than just the `typeIdentifier`.  (I left the `baseDefinition` function alone, since it's irrelevant here.)  I also made it so that `keyDefinition` will just, well, take the key definition if there is one -- the function was originally written to be used with expressions, rather than with actual variable definitions, which is why that case was missing.

### DIFF
--- a/packages/truffle-codec-utils/src/abi.ts
+++ b/packages/truffle-codec-utils/src/abi.ts
@@ -53,6 +53,7 @@ export namespace AbiUtils {
     type: string;
     indexed?: boolean; //only present for inputs
     components?: AbiParameter[]; //only preset for tuples (structs)
+    internalType?: string;
   }
 
   export interface FunctionAbiBySelectors {
--- a/packages/truffle-codec-utils/src/definition.ts
+++ b/packages/truffle-codec-utils/src/definition.ts
@@ -19,6 +19,11 @@ export namespace Definition {
     return definition.typeDescriptions.typeString;
   }
 
+  //returns the type string, but with location (if any) stripped off the end
+  export function typeStringNoLocation(definition: AstDefinition): string {
+    return typeString(definition).replace(/ (storage|memory|calldata)$/, "");
+  }
+
   /**
    * returns basic type class for a variable definition node
    * e.g.:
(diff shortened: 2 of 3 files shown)


packages/truffle-codec-utils/src/definition.ts
```diff
@@ -19,6 +19,11 @@ export namespace Definition {
     return definition.typeDescriptions.typeString;
   }
 
+  //returns the type string, but with location (if any) stripped off the end
+  export function typeStringNoLocation(definition: AstDefinition): string {
```
gnidan: how about naming it `typeStringWithoutLocation`?

gnidan: unless, of course, you mean `typeStringのLocation`

packages/truffle-codec-utils/src/abi.ts
```diff
@@ -53,6 +53,7 @@ export namespace AbiUtils {
     type: string;
     indexed?: boolean; //only present for inputs
     components?: AbiParameter[]; //only preset for tuples (structs)
+    internalType?: string;
```
gnidan: What is "internal type"?

haltman-at: It's the new field added in 0.5.11 which I earlier went and added to `truffle-contract-schema`.  It's basically a type hint for parameters which have types which lose a bit when converted to ABI.  E.g., for a struct or an enum, it'll tell you the name of the struct or the enum.  My equivalent for it in the decoder output format is called `typeHint`.

packages/truffle-codec-utils/src/definition.ts
haltman-at: Hahaha, OK, easy change.

Oy, I really made some hasty commits/pushes on this one... oh well...

packages/truffle-codec-utils/src/abi.ts
gnidan: Ah right, I remember that now. Thanks!

PR #2326, opened 2019-08-22
Fix: Fix migrations regression for non-ETH integrations

CruzMolina:
This PR fixes a migrations regression introduced in v5.
Now non-ETH integrations will use the same new v5 migrations system.


packages/truffle/test/scenarios/migrations/fabric-evm.js
```diff
@@ -7,15 +7,15 @@ const Reporter = require("../reporter");
 const sandbox = require("../sandbox");
 const Web3 = require("web3");
 
-describe("migrate with [ @fabric-evm ] interface", function() {
+describe("migrate with [ @fabric-evm ] interface", () => {
```
gnidan: Did you know that Mocha recommends using `function()` instead of `() =>`?

CruzMolina: 🤫

PR #2329, opened 2019-08-23
Syntax highlighting in the debugger

haltman-at:
Note: This branch was actually forked off of `develop`, but this is a PR against `next`.  I see there are some merge conflicts as a result (and also oddities like that compiler supplier thing being affected for some reason).  Not sure why, as I just merged `develop` into `next`, but I guess we'll have to sort that out later.  Wanted to put this PR up first.

This PR adds syntax highlighting to the debugger CLI.

The syntax highlighting is accomplished with `highlightjs-solidity` (recently given something of an overhaul by yours truly) and a customized version of `chromafi`.  The colors used have been chosen from the Truffle style guide to give things a Truffly look.  Note that I avoided anything I judged to be too close to white or black because I didn't want to risk the colors not showing up on certain terminals.

The version of chromafi is a customized org-scoped version of chromafi because, unfortunately, chromafi seems to be unmaintained at the moment.  My customizations are as follows:

1. I fixed a bug where colors would not continue across lines.
2. I exposed the `hljs` object so that the user (us) can modify it; we modify it by using `highlightjs-solidity` to register the `solidity` language on it.

Honestly, chromafi is a mess and we may just want to reimplement it oureselves later, but for now it was simpler just to make these modifications.

Also, note that because syntax highlighting must be done to entire source files at once, rather than line-by-line, we must store the colorized sources alongside their uncolorized versions.

Anyway, beyond that, it's pretty straightforward.  One thing worth noting though is that the `formatRangeLines` function takes both the colorized *and* the uncolorized source; while the colorized source is what ultimately gets printed, the uncolorized source is used for determining where to place the carets (the line pointer), as the colorized version doesn't work for that.

Also note that this PR doesn't add any other color yet, just syntax highlighting.  Further colorization will come later.

It's possible we may want to review my choice of colors.  You'll notice I set up a `truffleColors` object  that maps color names to corresponding chalk colors, and a separate `trufflePalette` object which is what chromafi actually uses, mapping highlight.js classes to the color names.  (You'll notice that some classes got listed multiple times because I was uncertain how chromafi handles them.  Like I said, chromafi is a mess.)  Any class we don't currently use is simply given the default of `chalk` (i.e. uncolored).

Note that I didn't bother including `truffleColors` any colors that I thought were unsuitable for use in this context, though I did include some that I thought were OK but which I didn't actually use.  Also note the `comment` color; this one does not come from the style guide, which is why it doesn't have a proper name.  Rather it comes from a different syntax highlighting thing in Truffle (I am forgetting what, sorry).  It's the only one specified as HSL rather than RGB because, well, that's how it was specified in what I copied it from.

One final note: I'm not entirely satisfied with the syntax highlighting, not in terms of the colors used, but rather the classes assigned.  Problem is, we're using highlight.js, so, well, we have to work within its framework.  It has no operator class, e.g., so operators aren't highlighted.  Now we could, if we so chose, totally abuse highlight.js, using its classes as arbitrary designators... but then we couldn't use the public version of highlightjs-solidity, which as to be suitable for general use; we'd have to use our own crazy customized version that abuses the classes and which would certainly not get upstreamed.  I didn't want to go doing that at the moment, so for now this is all within the framework of properly-used highlight.js, and so e.g. operators are not highlighted.


haltman-at:
Oh, I see, it's just `yarn.lock` that has conflicts.  That's nothing.  Phew.

As for the compiler supplier thing & similar, I think that's due to `prettier` getting run when I merged `develop` into this branch due to a merge conflict?  I think?

PR #2338, opened 2019-08-26
Add list command to debugger; colorize line pointer

haltman-at:
This PR adds an `l` command to the debugger CLI to list additional source context.  It also colorizes the line pointer (the carets).  Currently the color used is watermelon, but we might want to ask a design person for suggestions there... (I'm not so happy with it personally, it just seemed better than the alternatives).

Note that this is a PR on `next`, not `develop`.  Since this originally branched off of `syntax-highlighting`.  Notionally the `l` command by itself could be backported to `develop`, but I didn't see it as worth the trouble.

Anyway, the `l` command displays additional source context; instead of the usual 2 lines before and none after, it's 5 lines before and 3 after.  I didn't include any way to customize this; I didn't think that was necessary, but maybe that could be considered for the future, IDK.

Note that `l` is not a repeatable command.

That's basically all, really.


haltman-at:
I ended up swapping purple and watermelon.  We can get design to review the colors later, I suppose.

PR #2350, opened 2019-08-30
Internal improvement: Org scoping

eggplantzzz:
1. truffle-contract-sources --> @truffle/contract-sources
  2. truffle-contract --> @truffle/contract
  3. truffle-config --> @truffle/config
  4. truffle-artifactor --> @truffle/artifactor


scripts/ci.sh
```diff
@@ -39,7 +39,7 @@ elif [ "$GETH" = true ]; then
   run_geth
   sleep 30
   lerna run --scope truffle test --stream -- --exit
-  lerna run --scope truffle-contract test --stream -- --exit
+  lerna run --scope contract test --stream -- --exit
```
CruzMolina:
```suggestion
  lerna run --scope @truffle/contract test --stream -- --exit
```

eggplantzzz: Is this not a reference to the directory and not the package?

gnidan: I think it's supposed to be a reference to the package (at least that's what the build failure seemed to suggest)

CruzMolina: I thought so! https://github.com/trufflesuite/truffle/pull/2350/checks?check_run_id=208414445#step:6:954

eggplantzzz: Changed to the package name.

PR #2351, opened 2019-08-30
Internal improvement: More org scoping

eggplantzzz:
1. truffle-interface-adapter --> @truffle/interface-adapter
  2. truffle-core --> @truffle/core
  3. truffle-deployer --> @truffle/deployer
  4. truffle-environment --> @truffle/environment


packages/truffle/cli.webpack.config.js
```diff
@@ -8,37 +8,25 @@ const outputDir = path.join(__dirname, "build");
...
 module.exports = {
   entry: {
-    cli: path.join(
-      __dirname,
-      "../..",
-      "node_modules",
-      "truffle-core",
-      "cli.js"
-    ),
+    cli: path.join(__dirname, "../..", "node_modules", "core", "cli.js"),
```
CruzMolina: Believe all of these need to be like...

```suggestion
    cli: path.join(__dirname, "../..", "node_modules", "@truffle/core", "cli.js"),
```

eggplantzzz: Yeah that totally makes sense, thanks!

packages/truffle/cli.webpack.config.js
```diff
@@ -12,21 +12,21 @@ module.exports = {
...
-      "truffle-core",
+      "@truffle/core",
       "cli.js"
     ),
     chain: path.join(
       __dirname,
       "../..",
       "node_modules",
-      "truffle-environment",
+      "environment",
```
CruzMolina:
```suggestion
      "@truffle/environment",
```

PR #2352, opened 2019-08-30
Allow absence of whitespace in import statement

MaxWMcKinley:
Solc can compile either of the below import lines (notice the lack of white space in the second one):
`import "Example.sol;`
`import"Example.sol;`

But the regex currently fails on the second one.  This update allows either to pass.


MaxWMcKinley:
@gnidan I didn't see a way to update where this PR was pointing, so I made a new one here:https://github.com/trufflesuite/truffle/pull/2363

gnidan:
Cool, closing this one, then! Thanks!

Issue #2355, opened 2019-08-31
truffle doesn't use solidity's remappings


PaulRBerg:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.
- [x] I've asked for help on [Twitter](https://twitter.com/PaulRBerg/status/1167838340021936128)

---------------------------

## Issue

When compiling contracts that import other contracts, truffle behaves differently to sol-compiler and solc. The reason is that solidity's "remappings" are not put in the metadata object, as they should. See the docs on [Layout of a Solidity Source File](https://solidity.readthedocs.io/en/v0.5.11/layout-of-source-files.html) or this [StackExchange thread](https://ethereum.stackexchange.com/questions/74448/what-are-remappings-and-how-do-they-work-in-solidity) for more info.

## Steps to Reproduce

Make sure to have truffle and [sol-compiler](https://sol-compiler.com) globally installed and check out my [sol-compiler-truffle-mismatch](https://github.com/PaulRBerg/sol-compiler-truffle-mismatch) repo.

```
1. git clone git@github.com:PaulRBerg/sol-compiler-truffle-mismatch.git
2. cd sol-compiler-truffle-mismatch
3. yarn install
4. ./compare.sh
```
The last instruction should print "Artifacts are NOT equal".

## Expected Behavior

The artifacts should be the same, even if they are compiled with two different tools.

## Actual Results

The resulted artifacts have different metadata, because truffle doesn't put solidity's remappings in there, while sol-compiler does. It follows that the bytecode of the contracts gets to be different, since the bytes at the end include the hash of the metadata (see the solidity docs on [Contract Metadata](https://solidity.readthedocs.io/en/v0.5.11/metadata.html)).

## Environment

* Operating System: macOS v10.14.6
* Ethereum client: ganache-cli v6.5.1
* Truffle version (`truffle version`): v5.0.33
* node version (`node --version`): v10.15.3
* npm version (`npm --version`): v6.10.2

eggplantzzz:
Ok so looks like we just need to whitelist a remapping property for `compilers.solc.settings` in `truffle-config.js`. Then people should be able to specify remappings there. Thanks for such a detailed report @PaulRBerg!

eggplantzzz:
https://github.com/trufflesuite/truffle/pull/2358

PaulRBerg:
No worries. Thanks for the quick fix!

PaulRBerg:
[...]

PR #2358, opened 2019-09-02
Bug fix: Add remappings to solc options

eggplantzzz:
Previously, Truffle was not passing through remapping data to the Solidity compiler. This PR creates a default value for `compilers.solc.settings.remappings` in @truffle/config. It also passes the value on to the Solidity compiler at compile time in @truffle/compile-solidity so users can provide remappings in `truffle-config.js`.

Inspired by #2355


yarn.lock
```diff
@@ -14271,6 +14271,53 @@ trim-right@^1.0.1:
   resolved "https://registry.yarnpkg.com/trim-right/-/trim-right-1.0.1.tgz#cb2e1203067e0c8de1f614094b9fe45704ea6003"
   integrity sha1-yy4SAwZ+DI3h9hQJS5/kVwTqYAM=
 
+truffle-blockchain-utils@^0.0.10:
```
gnidan: What's up with these? Shouldn't they be the new names?

eggplantzzz: Hehe, is it a dependency of a dependency?

eggplantzzz: Yeah, I don't get it

eggplantzzz: Well ethpm-registry has a dependency of truffle-contract.

gnidan: ah that would do it!

PR #2362, opened 2019-09-02
Internal improvement: Rename truffle-resolver to @truffle/resolver

eggplantzzz:
1. truffle-resolver --> @truffle/resolver


gnidan:
why this fail

eggplantzzz:
yo no sé

eggplantzzz:
Oh that's because there is some weirdness with the version control as `node_modules` are ignored and it looks like they got put in there manually.

PR #2376, opened 2019-09-09
Bug fix: Correct packages/box/tsconfig.json

eggplantzzz:
Edit paths in packages/box/tsconfig.json to point to the root-level node_modules folder instead of the one for packages/box as they are not installed there.

`./packages/box` has the types dependencies listed in the package.json but they end up being installed at the project root level instead of in `packages/box`.

### DIFF
--- a/packages/box/tsconfig.json
+++ b/packages/box/tsconfig.json
@@ -19,11 +19,11 @@
     "typeRoots": [
       "./typings",
       "node_modules/@types",
-      "../../**/node_modules/@types/mocha",
-      "../../**/node_modules/@types/fs-extra",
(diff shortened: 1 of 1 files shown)


packages/box/tsconfig.json
```diff
@@ -19,11 +19,11 @@
     "typeRoots": [
       "./typings",
       "node_modules/@types",
-      "../../**/node_modules/@types/mocha",
```
gnidan: the ol' double star not cutting it, eh

eggplantzzz: So those typing packages get installed at the root level and not in `packages/box/node_modules` because I guess they are used somewhere else in the repo as well.

PR #2379, opened 2019-09-11
ABI-only based decoding as fallback in the wire decoder

haltman-at:
This PR allows the wire decoder to fall back onto ABI-based decoding when AST-based decoding encounters an error.

Essentially, the wire decoder can now either run in "full mode" or "ABI mode".  In full mode, the decoder works based on information contained in the AST, allowing it to output more information than it could if it were going based purely on the ABI.  For instance, enums can be decoded as enums, rather than as unsigned integers.  In ABI mode, the decoder works purely based on the ABI.  A tuple type has been added to the decoder output format's type system in order to support this.

This two-mode structure applies to allocation of calldata and events as well.  The ABI allocator has been restructured to run on type objects rather than AST definitions.  Note this only applies to the ABI allocator; ideally I'd have done the same for the storage and memory allocators, but this PR was complicated enough and there wasn't any pressing need so I left them alone and they still use definitions everywhere.  In full mode, the calldata or event allocator will convert definitions to types and allocate based on those.  However, if something goes wrong, the allocator will fall back into ABI mode, and get the types from the ABI rather than a matching definition.  ~~Note that full-mode allocations contain *both* the full-mode types *and* their ABI versions so that the decoder can still fall back into ABI mode when using them.~~ Edit: This last sentence was me misremembering; there's no need to store the ABI versions since they can easily be computed when needed.

(Also, the duplicated ABI size/isDynamic functions are now gone; it's all types now, no definitions.  Again, I didn't touch storage handling, so there's still a duplicated function there.)

So when decoding with a given allocation, then, if it is a full mode allocation, the decoder will first attempt to run in full mode; if something goes wrong -- which concretely means that either the matching definition can't be found or that an exception occurs during decoding (although more on that second condition later) -- then the decoder falls back into ABI mode.  Note that when doing the full-mode decoding, the decoder passes in a new option to the decoder core, `allowRetry`; this option causes particular exceptions -- ones where it makes sense to retry -- to be thrown when they otherwise would not be.  The option name is perhaps somewhat misleading, but it *is* intended to be used when you intend to allow a retry.  Meanwhile, when decoding with an ABI-based allocation, the decoder starts out in ABI mode and does not allow retries.

~~(I'm realizing now that there's a slight inconsistency in how I wrote this for calldata vs for events.  For calldata, I handled falling back into ABI mode by redoing the entire decoding, but based on ABI types rather than full-mode types.  Whereas for events, I did it by abifying (more on that in a bit) the previously-decoded arguments, redoing the decoding of the argument that caused the problem, and then continuing on in ABI mode.  I may want to change this for calldata to be consistent with how I did it for events and avoid recomputation.  In any case, this is a purely internal concern; you can basically ignore this paragraph.)~~ Edit: This is now fixed.

The wire decoder's decoding results are all marked with a decoding mode, `"full"` or `"abi"`, so the user knows what mode was used for their decoding.  Now, the user may want a specific mode, rather than having to deal with possibly inconsistent output!  What can such a user do?  Well, if they want everything to be in full mode... not much.  We use full mode whenever we can, but, well, sometimes we can't.  They can, uh, throw an exception if they don't get full mode like they want, I guess?  And they can try to ensure that full mode is used by writing all their contracts in Solidity and compiling them all simultaneously.  But in general we cannot guarantee the use of full mode.

On the other hand, if the user wants to force the decoder to run in ABI mode... well, we don't provide that option.  But we don't really need to, because instead we provide various `abify` functions that can convert a full-mode decoding into its ABI-mode equivalent; this obviously being possible as full mode contains more information.

Note by the way that modes are always applied at the level of the decoding.  So, within a single decoding, you won't have a mix with some variables in full mode and some variables in ABI mode; it's always consistent for the whole decoding.  But conversely, if an object admits *multiple* decodings -- e.g., when decoding an ambiguous event -- the different interpretations of the event *can* have different modes.

Also note that there actually is one *slight* difference between abifying your full-mode result, and a hypothetical option that would force the decoder to run in ABI mode, which is the handling of out-of-range enums.  In full mode, when decoding events, the decoder will discard any potential interpretations where an enum would be out of range, since they're not valid; such decodings simply won't show up on the decodings list, and therefore still won't be present if the decodings list is abified.  Whereas if the decoder were actually running in ABI mode, the decoder would lack the information needed to discard such decodings -- unless they're so far out of range that they're out of the range of the corresponding ABI type -- and so they would show up.  *Maybe* there is someone out there who really wants the latter functionality all the time, either because they are truly obsessive about consistency at the cost of correctness or because they actually are generating out-of-range enums, but for now I think this is fine.

Anyway.  Let's talk about exceptions.  Right now, when running in strict mode, the decoder will throw a `StopDecodingError` if something goes wrong.  I didn't really feel like making `RetryDecodingError` a separate class, so instead I just added a flag to `StopDecodingError` that indicates whether retries are allowed or not.  As mentioned above, the retriable ones are thrown both when strict mode is on as usual, but also when the `allowRetry` option is set in the decoder options.  Retries are only allowed for exceptions resulting from the decoder's inability to locate a user-defined type.  In practice this probably just means enums, since if a struct can't be located, that should cause a problem back at allocation, meaning the decoder wouldn't even be running in full mode in the first place; but still, just in case the decoder fails at locating a struct, it will throw a retriable exception.

Meanwhile, some things *don't* throw exceptions anymore!  Most peculiarly, out-of-range enums no longer cause exceptions, *even when strict mode is enabled*.  Why?  Well, because an exception would cause us to abandon this attempt at decoding... but maybe in decoding a later variable, something will go wrong and we'll have to fall back into ABI mode!  Since modes are always applied at the decoding level, that means this out-of-range enum is now retroactively OK, so we have to be prepared for that; an exception wouldn't allow that.

If the enum isn't an indexed parameter, of course, removing the exception doesn't cause any problems, because it still has to pass the re-encoding check afterwards.  If it *is* an indexed parameter... oops!  There is no re-encoding check for those, because it wasn't necessary previously!  So, now I've added a re-encoding check for indexed parameters as well.  This is kind of overkill, but I thought it was the cleanest way.  Although, it did require me to put a bit of a hack in the `encodeAbi` function to handle `IndexedReferenceTypeError`s.  I mean, it's not actually a hack programming-wise, it's very straightforward, it's just not really what `encodeAbi` is supposed to do.  I'll change this later once we have more proper functions for handling encoding of indexed parameters.

(Speaking of overkill, encoding of tuples was handled in a naïve, recomputation-heavy way in order to keep it simple and avoid rewriting the encoder.  We can revisit this if performance actually ever becomes a problem, but, while I do generally prefer to avoid recomputation, I am fairly doubtful this will actually be a serious problem.)

You know what else doesn't throw exceptions anymore?  The allocator!  You no longer have to worry about the wire decoder throwing exceptions on startup; it'll just fall back into ABI mode instead.  Note that the contract decoder can still throw exceptions on startup; I'm intending to address those in a later PR.

In order to make all this work, I had to basically eradicate our use of contract IDs as primary keys, because not every contract *has* a contract ID.  Now there basically are no more primary keys; in place of the contract ID, context hashes are used instead, but neither of a contract's context hashes can really be considered a primary key, since a context may have only one or indeed neither.  Still, it all hangs together!

Also new in this PR: Support for fixed-point types!  Solidity doesn't have these yet, but Vyper does, so it's here.  Since `bn.js` only handles integers, fixed-point numbers are represented via `big.js`.  Note that some of the handling of these -- specifically, the handling of types of the form `[u]fixedNx0` -- should not be considered final.

I also added some tests of the above functionality.  Note that because I haven't been able to get Vyper working, and had to test with Solidity, these are some seriously hacky tests -- they involve deliberately mangling the contract artifacts so that they'll cause allocation or decoding errors when the decoder uses them.  But hey!  The tests pass!  And they revealed some real problems along the way!

Two other changes worth mentioning:

1. In the `message` case of `decodeCalldata`, the `abi` field in the decoding will now distinguish between having a non-payable fallback function (returning the fallback ABI entry in that case) and not having a fallback function at all (returning `null` in that case).
2. When creating the context objects for a given contract, if the decoder can't find a given contract's `contractKind` field in its AST, it will attempt to guess if the contract is a library by reading its bytecode (although the library signature it looks for is only present in Solidity 0.4.20 and later).  If it doesn't find it, or no bytecode is given, it assumes it's an ordinary contract.

(Note that Solidity 0.5.12 will add a contract kind field to the compiler output, separate from the AST, and we might want to start putting that in the artifacts; if we do, the decoder can start looking at that, going by the AST only if that's not there.)

~~Oh, and here's an annoying thing: I had to make `GeneratorJunk` even larger.  Well, that'll go away when I merge in `next`.~~ OK this is gone now.

There are other minor things, but I think they're skippable.  So that's ABI mode!


packages/truffle-codec/lib/types/errors.ts
```diff
@@ -49,10 +49,14 @@ export class NoDefinitionFoundForABIEntryError extends Error {
...
-  constructor(error: Errors.DecoderError) {
+  public allowRetry: boolean; //setting this to true means that, if the error occurs
+  //when decoding in full mode, we allow an ABI-mode retry.  (if we were already in
+  //ABI mode, we give up.)
+  constructor(error: Errors.DecoderError, allowRetry?: boolean) {
     const message = `Stopping decoding: ${error.kind}`; //sorry about the bare-bones message,
     //but again, users shouldn't actually see this, so I think this should suffice for now
     super(message);
     this.error = error;
+    this.allowRetry = Boolean(allowRetry);
```
gnidan: why capital-B boolean here?

packages/truffle-codec/lib/decode/abi.ts
```diff
@@ -372,3 +367,33 @@ function* decodeAbiStructByPosition(dataType: Types.StructType, location: AbiLoc
     value: decodedMembers
   };
 }
+
+//note that this function takes the start position as a *number*; it does not take a pointer
```
gnidan: why not?

packages/truffle-codec-utils/src/types/abify.ts
```diff
@@ -0,0 +1,242 @@
...
+    //First: types that do not go in the ABI
+    case "mapping":
+    case "magic":
+      return undefined;
+    //Next: address & contract, these can get handled together
+    case "address":
+    case "contract":
+      return {
+        typeClass: "address",
+        payable: false,
```
gnidan: false in all cases? really?

packages/truffle-codec-utils/src/conversion.ts
```diff
@@ -11,17 +12,21 @@ import { enumFullName } from "./types/inspect";
...
       return undefined;
     } else if (typeof bytes == "string") {
       return new BN(bytes, 16);
     } else if (typeof bytes == "number" || BN.isBN(bytes)) {
       return new BN(bytes);
-    } else if (bytes.reduce) {
+    } else if (bytes instanceof Big) {
+      return new BN(bytes.toFixed()); //warning, better hope input is integer!
+      //note: going through string may seem silly but it's actually not terrible here,
+      //since BN is binary-based and Big is decimal-based
```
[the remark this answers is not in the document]
haltman-at: Yes; the `.toFixed()` method converts to a string while guaranteeing that scientific notation is not used.  Do you think that's worth commenting?

packages/decoder/lib/contract.ts
```diff
@@ -42,30 +40,17 @@ export default class TruffleContractDecoder extends AsyncEventEmitter {
...
-      const hash = CodecUtils.Conversion.toHexString(
-        CodecUtils.EVM.keccak256({type: "string",
-          value: this.contract.bytecode
-        })
-      );
-      this.constructorContextHash = hash;
-      this.constructorContext = this.contexts[hash];
+      const unnormalizedContext = Utils.makeContext(this.contract, this.contractNode);
+      this.contextHash = unnormalizedContext.context;
+      this.context = this.contexts[this.contextHash];
```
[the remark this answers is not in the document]
haltman-at: Hm, I guess I could do it that way.

packages/truffle-codec/lib/interface/decoding.ts
```diff
@@ -164,68 +202,131 @@ export function* decodeEvent(info: EvmInfo, address: string, targetName?: string
...
+              }
+            );
+          }
+          catch(_) {
+            //if an error occurred on the retry, this isn't a valid decoding!
+            debug("rejected due to exception on retry");
+            continue allocationAttempts;
+          }
+          //4. the remaining parameters will then automatically be decoded in ABI mode due to (1),
+          //so we don't need to do anything special there.
```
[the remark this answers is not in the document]
haltman-at: Hm, I'm not sure how.  There are small differences (specifically in the error-handling), and even without that... well, this is a very imperative section of the code, you know?  I don't really see how you could do that here, not unless Javascript had macros.

packages/truffle-codec/lib/types/errors.ts
haltman-at: Because capital-B `Boolean` is how you convert to a lowercase-b `boolean` in Javascript.  Weird, right?  It's the same with `Number`, `String`, and maybe other things.  You might think that `Boolean` converts to a `Boolean` (which we definitely don't want!), not a `boolean`, but no, that's `new Boolean`.  `Boolean` used without `new` makes a `boolean`.  Again, it's the same with `Number` and `String`, although with `String` I guess you'd usually use `.toString()` instead.

Of course in the case of `Boolean` you can use `!!` instead, but I think it's silly to express as two operations what is really one operation.

Note that in this case I'm just using it to convert `undefined` to `false`, so I could also write it as `allowRetry || false`.  But that seems less direct and less clear to me.

packages/truffle-codec-utils/src/conversion.ts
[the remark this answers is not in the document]
gnidan: yeah just because the comment talks about strings, so there's potentially a disconnect (esp. cause you use `toString()` in other places)

packages/decoder/lib/contract.ts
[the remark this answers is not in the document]
gnidan: Yeah, you can use the `get` keyword for this!

packages/truffle-codec/lib/interface/decoding.ts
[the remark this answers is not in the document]
gnidan: cool, just asking!

packages/truffle-codec/lib/types/errors.ts
gnidan: Heh oh right I forgot about that

packages/truffle-codec/lib/decode/abi.ts
haltman-at: For the same reason as with structs.  So why does the struct function work that way?  Well, originally as you'll recall there was no separate function for structs, the code was simply copypasted in both the dynamic and static decoding functions.  The struct function came about by factoring that.

So why not take a pointer?  Because that would work differently in the static vs dynamic cases.  The dynamic function assumes that the pointer points to a *pointer* to the struct (or in this case tuple); the static function assumes that the pointer points to the struct (or tuple) itself.  So, this factored version just takes the start as a number.

Notionally you could have it take a pointer to the struct (or tuple) itself, but I don't like that, as it breaks the convention used elsewhere that with dynamic things you *always* use a pointer to the pointer.  You could also combine the two cases a bit more, but that would require getting rid of an optimization I have in the static case where it doesn't go back to the dispatcher but just calls the static case again directly.

(Is it worth getting rid of that optimization for the sake of clarity?  I think it's not; for one, I'd dispute the "clarity" gain as it breaks the convention I mention above, and two, even if it is worth it, it's certainly not worth the effort to change now that it's already in the code. :P )

Anyway that's why.

packages/truffle-codec-utils/src/types/abify.ts
haltman-at: Because we're converting from the ABI, and there's no such ABI type as `address payable`, only `address`.

Now you'll recall I did discuss with you a while back having address-of-unspecified-payability as being a thing in the type system in order to make ABI-mode and full-mode more compatible, but, well, ultimately that compatibility turned out to be a fool's errand and I didn't do it, so I didn't make address-of-unspecified-payability a thing either.  I admit it's maybe a little misleading, but, well, that's ABI types for you.

I mean, alternatively you could make it always `true`, but, that seems even worse.  I think it's fine as-is.

packages/truffle-codec/lib/decode/abi.ts
gnidan: alright. well, since I am clearly in over my head here, I have no grounds to argue :)

packages/truffle-codec-utils/src/types/abify.ts
gnidan: Always being true seems more correct to me? Since the ABI treats everything as payable.

haltman-at:
OK, so I just added a commit that does a few things:

1. It adds error-handling to the ABI allocator... where by "error-handling" I mean "if an error occurs it just discards that allocation and moves on".  I thought about adding a flag that would tell it, hey, don't ignore errors, but, I didn't see that as being especially useful, so I didn't.  We can add it if there's ever a reason for it later.  Note that doing things this way is a little wasteful in terms of recomputation, but as usual I'm doubtful it's actually a real problem and we can revisit it if it is.

2. It modifies the test that was *supposed* to catch things like this so that it *actually* catches things like this. :P

3. It removes storage allocation from the wire decoder.  I think that was in there with the intent that this way the contract decoder wouldn't have to redo storage allocation if you made more than one contract decoder, but ultimately this doesn't fit with the way things currently work, and it was causing problems, so I removed it.  Maybe we can find a way to put it back in later, but for now it's gone for the sake of simplicity.

packages/truffle-codec-utils/src/types/abify.ts
haltman-at: I mean, maybe?  IMO it'd be a little more confusing to have `address payable` turn up as a type in your ABI-mode decoding.  OTOH we did do something similar regarding plain uses of `address` in pre-0.5.0 Solidity.  I'm more inclined towards `false`, but if you really think `true` is the way to go we can change it.  Just hoping nobody ends up relying on the result. :P

gnidan: Alternatively, could we make `payable` nullable? Like... we don't really know!

haltman-at: Hahaha yup that possibility again!  I mean, we could?  This is basically the possibility I discussed with you before.  I'm a little leery of that, as it makes `address` (the Solidity type) and `address` (the ABI type) into different things?  (Also, we might want to add a tag to distinguish the two cases, which was actually how we discussed doing it before.)  But yeah that's definitely one more possibility?

gnidan: Hm, but aren't they different things?

I'm not sure how I feel about this. I see the trickiness. We might just have to solve this with notes in documentation that this isn't supported yet.

haltman-at: I mean, they're *kind* of different, but I didn't want to get into, like, distinguishing ABI types from Solidity types except where necessary.  Like, is a `uint8` different when it's a Solidity `uint8` vs an ABI `uint8`?  (Because the latter might secretly be an enum.)  I thought it was best to say, no it's not, and have the `typeHint` field for when we do have that information.

Not sure what you mean by "not supported yet".  It will very definitely be supported, the only question is which possibility we go with!

Actually I only just noticed that this is in `abify.ts`, that this is `abifyType`, and not `abiParameterToType` like I thought.  So, in this context, if you really want that information as to whether it was payable or not, you can check the `typeHint` field!

haltman-at: Actually, that's another slight difference between using `abify` vs actually running the decoder in ABI mode: If you use `abify`, you will get extra type information in `typeHint`, like whether the address was payable or not.  If you actually run the decoder in ABI mode, you won't.

haltman-at: OK, after talking to @gnidan we decided to make `payable` nullable, marking it with `kind: "general"` or `kind: "specific"`, like with external functions.

haltman-at:
OK, pushed a commit addressing the review comments.  The biggest change, of course, is the redoing of the address type.  Some notes on this:

1. Since we now have `kind: "general"`, there's no more need for the weird hacks with `compiler = null` or `forceNonPayable` in the debugger.  All that logic is gone.  Now, in pre-0.5.0 Solidity, all addresses are general, rather than payable (with a hack used to make them nonpayable in the case of mapping keys).

2. Note that this meant I had to add logic for determining what type to use for addresses used as part of the globals `msg`, `tx`, and `block`.  I added one function for `tx.origin` and `block.coinbase` (payable if 0.5.0 or greater, general otherwise) and another for `msg.sender` (which currently is the same but which is kept separate because `msg.sender` is likely going to become nonpayable in 0.6.x).  To do this I factored out the "what version is this?" logic from `isAddressPayable` and made it its own function.

Also while I was doing this I removed various comments from `types.ts` and `values.ts` about how various things aren't supported yet because, well, now they are.

packages/truffle-codec-utils/src/compiler.ts
```diff
@@ -1,6 +1,19 @@
+import semver from "semver";
+
 export interface CompilerVersion {
   name?: string;
   version?: string;
   //NOTE: both these should really be present,
   //but they need to be optional for compilation reasons
 }
+
+export type SolidityFamily = "pre-0.5.0" | "0.5.x";
```
gnidan: ooh I like this

packages/truffle-codec-utils/src/compiler.ts
```diff
@@ -1,6 +1,19 @@
...
+
+export function solidityFamily(compiler: CompilerVersion): SolidityFamily {
+  if(semver.satisfies(compiler.version, "~0.5 || >=0.5.0", {includePrerelease: true})) {
```
gnidan: (Of course we'll have to change `>=` post v0.6.0)

gnidan: I think you can just use `^0.5.0`, no?

packages/truffle-codec-utils/src/definition.ts
```diff
@@ -146,22 +145,6 @@ export namespace Definition {
...
-      //note that we use ~0.5 || >=0.5.0 to make sure we include prerelease versions of
-      //0.5.0 (no, that is *not* what the includePrerelease flag doesn; that allows
-      //prerelease versions to be included *at all*)
-      return typeIdentifier(definition) === "t_address_payable";
-    }
-    else {
-      return true;
-    }
-  }
-
```
gnidan: sweet delete

packages/truffle-codec/lib/decode/special.ts
```diff
@@ -97,10 +91,7 @@ export function* decodeMagic(dataType: Types.MagicType, pointer: SpecialPointer,
     case "block":
       let block: {[field: string]: Values.Result} = {
         coinbase: yield* decodeValue(
-          {
-            typeClass: "address" as const,
-            payable: true
-          },
+	  externalAddressType(info.currentContext.compiler),
```
gnidan: nice use of helpers!

packages/truffle-codec-utils/src/compiler.ts
haltman-at: The reason for the complex version condition is that I want to make sure to include prerelease versions of 0.5.0; I'm not sure that `>=0.5.0` or `^0.5.0` cover those, but I think `~0.5` does.  (Note the comment about how no, passing the `includePrerelease` flag is not sufficient for that.)

Thinking about it now, I guess it would make more sense to just use `~0.5`.  Only reason I didn't was because, well, this logic was factored from `isAddressPayable`, and it didn't occur to me to change it.  Could change it now or could save it for when we handle 0.6.0.

haltman-at: Actually likely it makes sense to leave it as-is for now as if someone somehow *is* using Solidity 0.6.0 right now, it makes more sense to return 0.5.x rather than pre-0.5.0.  Yeah I think I'll just change this when we do 0.6.0 handling.

PR #2382, opened 2019-09-11
Stop wrapping the provider for migrations

gnidan:
Looks like the provider wrapper doesn't actually do anything!

It [used to](https://github.com/trufflesuite/truffle/blob/28dca63366004258e8abf1329fbbd1be1ca9948f/packages/truffle-migrate/index.js#L206-L233), but now the wrapper is basically a no-op with restrictions.

This should remove the need for #1307 and solve other problems (@CruzMolina said something about the provider-engine fork, but not sure of specifics.)

### DIFF
--- a/packages/migrate/index.js
+++ b/packages/migrate/index.js
@@ -130,7 +130,6 @@ const Migrate = {
       clone.logger = { log: function() {} };
     }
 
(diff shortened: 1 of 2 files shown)


eggplantzzz:
Sweet, I was wondering why this was done :)

gnidan:
Thanks!

Issue #2354, opened 2019-08-30
add a command line option to specify the truffle config file to be used


naddison36:
This is an enhancement request to add a command line option to specify the truffle config file to be used on compilation.

The use case I have is I'm using the Gnosis MultiSigWallet which requires Solidity `0.4.15` while the other contracts in my app are using `0.5.11` as I need ABIEncoderV2. At the moment I put the MultiSigWallet contract in a different folder and manually modify my `truffle-config.js` file to specify the compiler version to use for the folder I want to compile.

Ideally, I'd like to have two truffle config files and just use a command line option to specify which config file to use when compiling. That means in a continuous integration build, two truffle compile commands would be run.

CruzMolina:
Related #1879.

gnidan:
Oh, maybe that other one can be closed as a duplicate of this, since it's less general? Let's figure that out when we triage this.

naddison36:
An update on how I work around this issue. I have separate folders for my Solidity 0.5 and 0.4 contracts with separate `truffle-config.js` files in each folder. I then run `truffle compile` from the folder I want to compile.

eggplantzzz:
@naddison36 This feature has been merged and released with version 5.0.38. https://github.com/trufflesuite/truffle/releases/tag/v5.0.38

PR #2383, opened 2019-09-11
Enhancement: Allow --config flag

eggplantzzz:
Allow an option `--config` whereby a user can provide a relative or absolute path to a config file of their choice. Truffle will attempt to load the file and use it during the command flow they specified.

This also adds 2 unit tests for `Config.detect`.

### DIFF
--- a/packages/config/index.js
+++ b/packages/config/index.js
@@ -132,12 +132,19 @@ Config.search = (options = {}, filename) => {
 };
 
 Config.detect = (options = {}, filename) => {
-  const configFile = Config.search(options, filename);
+  const { configPath } = options;
+  let configFile;
+  if (configPath) {
+    configFile = path.isAbsolute(configPath)
+      ? configPath
+      : path.resolve(configPath);
+  } else {
+    configFile = Config.search(options, filename);
+  }
 
   if (!configFile) {
     throw new TruffleError("Could not find suitable configuration file.");
   }
-
   return Config.load(configFile, options);
 };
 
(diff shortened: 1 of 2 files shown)


CruzMolina:
Side note: Adding support for a config key such as `includes` or `extends` that expects an array of paths to various truffle config files and runs truffle commands (certain ones? all?) in each projectDir passed to it seems like a cool way to resolve [this kind of use case](https://github.com/trufflesuite/truffle/issues/2354#issuecomment-528118574).

So in a monorepo CI/CD scenario, one `truffle compile` run from a root `truffle-config.js` which contains something like...

```
includes/extends: [
  "./projectA",
  "./projectB",
  ...
  ],
 ```

...would compile projects at each specified level using their respective truffle configs. One could have two versions of the same project (solc 0.4.x & solc 0.5.x) compiled/migrated/tested/etc in one go.

File: packages/config/index.js [hunk 1]
```diff
@@ -132,19 +132,26 @@ Config.search = (options = {}, filename) => {
...
   }
-
   return Config.load(configFile, options);
 };
 
 Config.load = (file, options) => {
   const config = new Config();
 
-  config.working_directory = path.dirname(path.resolve(file));
+  config.working_directory = process.cwd();
```
CruzMolina: I think it makes sense to do a boolean check in here or in detect to see if someone is using the config flag.

If they are, use cwd, otherwise use the current behavior.

gnidan:
Too bad Cruz

PR #2387, opened 2019-09-12
Enhancement: Feature truffle init

eggplantzzz:
This incredible PR enables the passing of a path (as a command line argument) to the init command. For example, one could execute `truffle init ./myPath` to initialize a shiny, brand new project in the `myPath` folder!

As a consequence of this cutting-edge feature, the unbox command has also been enhanced! Now you can also pass a path to the command as an argument as follows: `truffle unbox metacoin /Users/Agatha/secretStuff/myMetacoinProject`. The metacoin box will then be unboxed in the `myMetacoinProject` directory.

**NOTE:** Previously you could provide a path at the end of the URL a la `http://something:my/path`. If you provide your own path as an argument, it overrides this suffix path.

**ANOTHER NOTE:** The nightmare of a method that is known as `normalizeInput` in `packages/core/lib/commands/unbox.js` should probably be rewritten at some point in the future as it is a mess to untangle and make sense of. Liberté, egalité, fraternité!

### DIFF
--- a/packages/core/lib/commands/unbox.js
+++ b/packages/core/lib/commands/unbox.js
@@ -125,7 +125,7 @@ const command = {
 
     const config = Config.default().with({ logger: console });
 
-    ({ url, destination } = normalizeInput(options._));
+    let { url, destination } = normalizeInput(options._);
     url = normalizeURL(url);
 
     const normalizedDestination = normalizeDestination(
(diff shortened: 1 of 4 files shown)


packages/core/lib/commands/unbox.js
```diff
@@ -113,19 +121,23 @@ const command = {
...
+    const fse = require("fs-extra");
 
-    const config = Config.default().with({
-      logger: console
-    });
-
-    let [url, destination] = normalizeInput(options._[0]);
+    const config = Config.default().with({ logger: console });
 
+    ({ url, destination } = normalizeInput(options._));
```
gnidan: I think you need a `const` or `let` here

PR #2394, opened 2019-09-18
Internal improvement: Set CI env

CruzMolina:
Since Github Actions supports any number of custom workflows, it doesn't pass `CI=true` by default (which Travis does).

Makes sure docker/native solc tests (along with any other tests relying on the env) run in Actions CI.


CruzMolina:
this didn't run Actions. also, looks like something changed or broke with native solidity

gnidan:
danggg

PR #2396, opened 2019-09-18
Contract instance creation

fainashalts:
This PR adds the `creation` field to the Contract Instance resource, adding it to the schema, loaders, and workspace. 

Note: creation includes a `constructor` object which includes `createBytecode`. A forthcoming PR will be reworking how bytecode is done throughout the Truffle DB project, so this is just a stand-in.


gnidan:
adding you as reviewer @cds-amal to help get your feet wet

packages/truffle-db/src/loaders/artifacts/index.ts
```diff
@@ -433,7 +459,9 @@ export class ArtifactsLoader {
               configNetworks.push({
                 contract: contractName,
                 id: id,
-                address: filteredNetwork[0][1]["address"]
+                address: filteredNetwork[0][1]["address"],
```
gnidan: btw you can do `.find(...)` instead of `.filter(...)[0]` if you want to just filter for a single item. should clean up this code a bunch! (thanks to Harry for telling me about this :)

PR #2406, opened 2019-09-23
@truffle/hdwallet-provider added typescript declaration file

MrRefactoring:


packages/hdwallet-provider/index.d.ts
```diff
@@ -0,0 +1,27 @@
+import {
+  JSONRPCRequestPayload,
+  JSONRPCResponsePayload
+} from 'ethereum-protocol';
```
gnidan: Looks like this is missing from the deps. (Wait, where does this sort of thing go... `dependencies` or `devDependencies`?)

MrRefactoring:
> Thanks for this @MrRefactoring, that's awesome! Just the one thing about the missing dep.
> 
> Alternatively... how hard would it be to port this to TS? 😈

This would be very easy to port to typescript. If you are ready to switch to typescript, then I am ready to help

packages/hdwallet-provider/index.d.ts
MrRefactoring: `web3-provider-engine` use an `ethereum-protocol`. Better add this to `dependencies`?

MrRefactoring: I also noticed that you did not add `debug` to `dependcies`

gnidan: Yep, best to add all dependencies directly! Good catch on `debug`, that should be there too!

gnidan:
> This would be very easy to port to typescript. If you are ready to switch to typescript, then I am ready to help

Definitely! We've been converting packages to TS here and there, incrementally, so if that's something you have availability for, we'd definitely welcome and appreciate it!

gnidan:
Also if you pull upstream from `develop` I think I fixed the failing Slack notifier build so you won't have to see a red ❌

MrRefactoring:
Migrated `@truffle/hdwallet-provider` to typescript in PR #2412

PR #2407, opened 2019-09-23
Stop running Slack notifier for forks

gnidan:
To prevent build failure when the lack of `secrets` causes the action to break


gnidan:
Misaotra!

PR #2408, opened 2019-09-23
Rework interface for getting variables in the contract decoder

haltman-at:
That's right, @davidmurdoch, @adrianmcli, it's another breaking change to the decoder!  But it's getting bundled with all the others.  Hoping to write up docs for all this soon.

Anyway.  This PR does several things:

1. The `state` function in the decoder no longer includes the variables.  If you want the variables, that's now a separate `variables` function.

2. The decoder should no longer throw exceptions on startup!  I mean, I can't rule it out entirely, but, if it happens, that means I messed up.  It's no longer expected behavior, no longer something you should have to be prepared for.

3. However, the new `variables` function -- as well as the existing `variable`, `watchMappingKey`, and `unwatchMappingKey` function -- now all begin by checking to see whether allocation succeeded, and, if not, they throw an exception.  Note that `state` is not listed here; since it doesn't involve the variables anymore, it shouldn't throw exceptions.

4. Internal change: The wire decoder now performs storage struct allocations -- it doesn't need these, of course, but this way any contract decoders made with it don't have to redo this work.  In order to allow this, the `getStorageAllocations` function now has an optional third argument, where you can pass in any existing storage allocations you already know about.  Note that it's OK if some of the stuff you pass in to be allocated in the first two arguments is already there, since the allocator just skips over things that are already allocated.

Anyway, point is, you should no longer have to `try`/`catch` everything involving the decoder!  Only things that actually involve decoding variables.

### DIFF
--- a/packages/codec/lib/allocate/storage.ts
+++ b/packages/codec/lib/allocate/storage.ts
@@ -24,8 +24,8 @@ interface DefinitionPair {
 
 //contracts contains only the contracts to be allocated; any base classes not
 //being allocated should just be in referenceDeclarations
-export function getStorageAllocations(referenceDeclarations: AstReferences, contracts: AstReferences): StorageAllocations {
-  let allocations: StorageAllocations = {};
+export function getStorageAllocations(referenceDeclarations: AstReferences, contracts: AstReferences, existingAllocations: StorageAllocations = {}): StorageAllocations {
+  let allocations = existingAllocations;
   for(const node of Object.values(referenceDeclarations)) {
     if(node.nodeType === "StructDefinition") {
       try {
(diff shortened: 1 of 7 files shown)


packages/codec/lib/allocate/storage.ts
```diff
@@ -24,15 +24,29 @@ interface DefinitionPair {
 
 //contracts contains only the contracts to be allocated; any base classes not
 //being allocated should just be in referenceDeclarations
-export function getStorageAllocations(referenceDeclarations: AstReferences, contracts: AstReferences): StorageAllocations {
-  let allocations: StorageAllocations = {};
+export function getStorageAllocations(referenceDeclarations: AstReferences, contracts: AstReferences, existingAllocations: StorageAllocations = {}): StorageAllocations {
```
gnidan: Positional arguments :(

Issue #2414, opened 2019-09-24
Truffle config.load() does not pass expected parameters to new Config()


fainashalts:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

What the issue is, in broad strokes.

The Config constructor in the `config/index.js` file expects three parameters to be passed: the truffle_directory, the working_directory, and the network. When `new Config()` is called in the `load()` function in this file, these parameters are not passed. Moreover, `config.working_directory` is set after `new Config()` is called, whereas it should have been set through the constructor.

## Steps to Reproduce

Please provide the shortest amount of steps to reproduce your issue.

## Expected Behavior

I expect that the expected parameters are passed to `new Config()` if they are available.

## Actual Results

These parameters are not passed.

## Environment

* Operating System: mac
* Ethereum client: n/a
* Truffle version (`truffle version`): 5.0.32
* node version (`node --version`): 8.16.0
* npm version (`npm --version`): 6.10.3

PR #2418, opened 2019-09-25
@truffle/config: migration to typescript

MrRefactoring:
- migrated to typescript
- tests fixed (in windows, paths are broke)
- fixed context missing in `addProp`
- I tried to take into account the comments from #2414
- added missed deps


packages/config/src/index.ts
```diff
@@ -0,0 +1,183 @@
...
+          self._values[propertyName] = descriptor.transform
+            ? descriptor.transform(value)
+            : value;
+        },
+      configurable: true,
+      enumerable: true
+    });
+  }
+
+  public normalize(obj: any): any {
```
MrRefactoring: Do I understand correctly that this is just cloning? Can we use Object.assign or the spread operator instead?

gnidan: Looks like it's specifically cloning for the purpose of getting rid of the prototype.

I think `Object.assign` is fine, or the spread operator if you make sure it has the polyfill for Node v8 support.

gnidan: But! Keep `config.normalize` around, even if you don't end up using it, because otherwise it'll  be a breaking change.

MrRefactoring: > But! Keep `Config.normalize` around, even if you don't end up using it, because otherwise it'll be a breaking change.

Yes, I thought over this variant

gnidan:
Sweet @MrRefactoring, thanks so much for this! I'll get you a review today

gnidan:
Oh but @eggplantzzz merged his PR #2383. Can you do the merge?

MrRefactoring:
> Oh but @eggplantzzz merged his PR #2383. Can you do the merge?

Yes sure

MrRefactoring:
@gnidan could you restart travis? I looked, there is a strange node-gyp error, I think it's not on our side

packages/config/src/configDefaults.ts
```diff
@@ -1,15 +1,23 @@
-const path = require("path");
-const _ = require("lodash");
-const Provider = require("@truffle/provider");
+import * as path from 'path';
+import lodash from 'lodash';
```
eggplantzzz: Can we import this as `_`? This is currently the project convention.

packages/config/src/configDefaults.ts
```diff
@@ -123,15 +135,15 @@ const configProps = ({ configObject }) => {
           throw new Error("Network not set. Cannot determine network to use.");
         }
 
-        let conf = configObject.networks[network];
+        let config = configObject.networks[network];
```
eggplantzzz: Thank you for using the full name instead of `conf` :)

packages/config/package.json
```diff
@@ -2,10 +2,12 @@
   "name": "@truffle/config",
   "version": "1.1.21",
   "description": "Utility for interacting with truffle-config.js files",
-  "main": "index.js",
-  "typings": "./typings/index.d.ts",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
```
eggplantzzz: Oooh, this looks like our error. This property should be `"types"` and not `"typings"`.

eggplantzzz: For reference https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

MrRefactoring: These are synonyms, as vscode says, but I’d better leave types to not cause questions.

types

packages/config/package.json
```diff
@@ -2,10 +2,14 @@
...
   "description": "Utility for interacting with truffle-config.js files",
-  "main": "index.js",
-  "typings": "./typings/index.d.ts",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
-    "test": "mocha"
+    "build": "tsc",
+    "prepare": "yarn build",
+    "lint": "tslint src/**/*.ts",
```
eggplantzzz: Thanks for adding these but I'm not sure that we want these extra linting libraries. We already have ESLint that we use project-wide and it looks like tslint will be deprecated this year (see [the repo](https://github.com/palantir/tslint). I don't know, thoughts @gnidan?

packages/config/src/configDefaults.ts
```diff
@@ -24,10 +32,10 @@ const _values = ({ truffleDirectory, workingDirectory, network }) => {
     resolver: null,
     artifactor: null,
     ethpm: {
-      ipfs_host: "ipfs.infura.io",
-      ipfs_protocol: "https",
-      registry: "0x8011df4830b4f696cd81393997e5371b93338878",
-      install_provider_uri: "https://ropsten.infura.io/truffle"
+      ipfs_host: 'ipfs.infura.io',
```
eggplantzzz: Also, prettier changes strings to use double quotes and not single quotes wherever possible.

packages/config/package.json
MrRefactoring: I deleted it. I think it's better to postpone it. In addition, given that there is already eslint. It remains only to wait for the transition tslint to eslint

packages/config/src/configDefaults.ts
MrRefactoring: Resolved

eggplantzzz:
It seems good to me. Maybe @CruzMolina can set his eyes on this. He is more familiar than I with setting up TS projects :)

packages/config/tslint.json
```diff
@@ -0,0 +1,7 @@
+{
```
CruzMolina: Woops! Forgot to remove this.

packages/config/typings/original-require/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'original-require';
```
CruzMolina: Same here (re: EOL).

packages/config/typings/provider/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@truffle/provider';
```
CruzMolina: Also here (re: EOL).

yarn.lock
```diff
@@ -14471,11 +14760,42 @@ ts-node@^8.4.1:
...
+  version "1.10.0"
+  resolved "https://registry.yarnpkg.com/tslib/-/tslib-1.10.0.tgz#c3c19f95973fb0a62973fb09d90d961ee43e5c8a"
+  integrity sha512-qOebF53frne81cf0S9B41ByenJ3/IuH8yJKngAX35CmiZySA0khhkovshKK+jGCaMnVomla7gVlIcc3EvKPbTQ==
+
 tslib@^1.9.0, tslib@^1.9.3:
   version "1.9.3"
   resolved "https://registry.yarnpkg.com/tslib/-/tslib-1.9.3.tgz#d7e4dd79245d85428c4d7e4822a79917954ca286"
   integrity sha512-4krF8scpejhaOgqzBEcGM7yDIEfi0/8+8zDRZhNZZ2kjmHJ4hv3zCbQWxoJGz1iw5U0Jl0nma13xzHXcncMavQ==
 
+tslint@^5.20.0:
```
CruzMolina: Can you also regenerate the `yarn.lock`?

packages/config/typings/error/index.d.ts
```diff
@@ -0,0 +1 @@
+declare module '@truffle/error';
```
CruzMolina: Same here (re: EOL).

packages/config/tslint.json
MrRefactoring: Nice catch!

gnidan:
Just update your editor settings to add new lines at end of files :) (and delete trailing whitespace on other lines, too, if you don't have it!)

CruzMolina:
Thanks to #2429 , no need to worry about EOL/whitespace issues!™️ (🤞)

Issue #2419, opened 2019-09-25
truffle test error when empty folder in test folder


eggplantzzz:
When running `truffle test`, if an empty folder exists in the test directory it will fail with 
```
TypeError: Cannot read property 'files' of undefined
    at subloop (/Users/tyler/.nvm/versions/node/v8.16.0/lib/node_modules/truffle/build/webpack:/~/node-dir/lib/paths.js:92:1)
    at /Users/tyler/.nvm/versions/node/v8.16.0/lib/node_modules/truffle/build/webpack:/~/node-dir/lib/paths.js:107:1
    at onDirRead (/Users/tyler/.nvm/versions/node/v8.16.0/lib/node_modules/truffle/build/webpack:/~/node-dir/lib/path
.......
........
```

eggplantzzz:
It turns out this issue is related to the `node-dir` library and unfortunately it has not been published for the last 2 years.

https://github.com/fshost/node-dir/issues/53

gnidan:
good find!

eggplantzzz:
This issue was resolved with the release of v5.0.40. https://github.com/trufflesuite/truffle/pull/2420

PR #2420, opened 2019-09-25
Bug fix: Error with empty folder in test dir

eggplantzzz:
The node-dir library throws an error when an empty directory is found. This is what is used to find test files and created errors. I have removed that library in favor of the almighty glob!!

This PR uses glob to find all the contents of the test directory and filters out non-files. Here it is again, all hail glob!!


packages/core/test/lib/create.js
```diff
@@ -124,32 +123,29 @@ describe("create", function() {
       err
     ) {
       if (err) return done(err);
+      const files = glob.sync(`${config.migrations_directory}${path.sep}*`);
 
-      dir.files(config.migrations_directory, function(err, files) {
-        if (err) return done(err);
+      var found = false;
+      var expected_suffix = "_my_new_migration.js";
```
MrRefactoring: I think here you can use `const` instead of `var`

packages/core/test/lib/create.js
```diff
@@ -124,32 +123,29 @@ describe("create", function() {
...
 
-        var found = false;
-        var expected_suffix = "_my_new_migration.js";
+      for (var i = 0; i < files.length; i++) {
```
MrRefactoring: `let` is more suitable here instead of `var`

packages/core/test/lib/create.js
```diff
@@ -124,32 +123,29 @@ describe("create", function() {
...
+        var file = files[i];
 
-        for (var i = 0; i < files.length; i++) {
-          var file = files[i];
+        if (
+          file.indexOf(expected_suffix) ===
+          file.length - expected_suffix.length
+        ) {
+          var file_data = fse.readFileSync(file, { encoding: "utf8" });
```
MrRefactoring: I think about `const` in this place

eggplantzzz:
This fixes #2419

eggplantzzz:
Thanks for the review @MrRefactoring!

packages/core/test/lib/create.js
```diff
@@ -124,32 +123,29 @@ describe("create", function() {
...
 
-      dir.files(config.migrations_directory, function(err, files) {
-        if (err) return done(err);
+      const found = false;
+      const expectedSuffix = "_my_new_migration.js";
 
-        var found = false;
-        var expected_suffix = "_my_new_migration.js";
+      for (let i = 0; i < files.length; i++) {
+        const file = files[i];
```
gnidan: Did you know... you can just do:

```javascript
for (let file of files) {
  // ...
}
```

since it doesn't look like you're using `i` at all.

PR #2422, opened 2019-09-27
Fix debugger crash from regression in codex system

haltman-at:
This PR fixes a debugger crash resulting from a regression in the codex system.  I think what happened is this:

Originally, when I made the codex system, I made it so that, at debugger startup, it created an account for every contract that got called or created.  However, later we made it so that this didn't include creations.  I forget why, but I assume we had a good reason. :P  This then caused a problem when the `updateFrameStorage` function just assumed that the accounts it was updating already existed, and they might not, potentially crashing the debugger.

So, this PR fixes that; now, if `updateFrameStorage` finds that no account exists, it will create one.  Note that it *does* still rely on the assumption that this will only happen with contracts we create rather than contracts we call into -- when it creates a new account, it creates it with no code.  This is only valid because, as I said, this case should only arise with contracts created during the transaction, which therefore must have necessarily started the transaction with no code.

It's possible that really we should revisit the question of which contracts should have accounts created on startup, but, well, this seemed easier, and as far as I can tell is entirely correct (again, assuming we do fetch the code at startup for all contracts we don't create, which currently we do).

### DIFF
--- a/packages/debugger/lib/evm/reducers.js
+++ b/packages/debugger/lib/evm/reducers.js
@@ -198,14 +198,17 @@ function codex(state = DEFAULT_CODEX, action) {
     accounts: {
       ...frame.accounts,
       [address]: {
-        ...frame.accounts[address],
+        code: "0x", //this will get overridden if it already exists!
+        context: null, //similarly!
+        ...frame.accounts[address], //may be undefined
         storage: {
-          ...frame.accounts[address].storage,
+          ...(frame.accounts[address] || {}).storage, //may be undefined
           [slot]: value
         }
       }
     }
   });
+  //(note that {...undefined} just expands to {} and is OK)
 
   const updateFrameCode = (frame, address, code, context) => {
     let existingPage = frame.accounts[address] || { storage: {} };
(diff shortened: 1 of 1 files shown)


packages/debugger/lib/evm/reducers.js
```diff
@@ -198,14 +198,17 @@ function codex(state = DEFAULT_CODEX, action) {
...
+        ...frame.accounts[address], //may be undefined
         storage: {
-          ...frame.accounts[address].storage,
+          ...(frame.accounts[address] || {}).storage, //may be undefined
           [slot]: value
         }
       }
     }
   });
+  //(note that {...undefined} just expands to {} and is OK)
```
gnidan: huh. the more you know!

gnidan:
(Oh, in what situations would we not be able to fetch code for contracts we don't create?)

haltman-at:
> (Oh, in what situations would we not be able to fetch code for contracts we don't create?)

Not sure I understand the question.  We do fetch the code for all contracts we don't create.  The issue here involves contracts we *do* create, before we create them.  That's why it's safe to assume the code is `0x` in this case.

PR #2424, opened 2019-09-27
Hide private state variables when out of scope

haltman-at:
This PR hides out-of-scope private state variables in the debugger.  So, if you have a base class `Base` with a private state variable `x`, and are currently stepping through a function in a derived class `Derived`, the variable `x` will not appear when you press the `v` command, nor will it be accessible in watch expressions.

~~To accomplish this, I've slightly modified the `DECLARE` system.  Now, each scope maintains both `variables` and `visibles`; when a variable is declared; it's added to `variables`, but only added to `visibiles` if its visibility is not equal to `private` (i.e., is equal to `public` or `internal`).~~

~~Then, when variables from base classes are added in during computation of `data.info.scopes`, only the `visibles` from these base classes are pulled in.  Note that the last entry in the list of base classes, i.e. the class under consideration itself, now has to be treated specially, as we want all of its variables, even the `private` ones.~~

Edit: I've redone this without modifying any reducers.  Now all filtering is done in the selector, because we can use a variable's ID to look up its visibility, without needing to store it.

All this has no effect on allocation, as the allocator handles inheritance itself; so we're not screwing up the allocation by filtering some variables out in the debugger.

~~Note: I didn't bother altering the value of `scope.visibles` in `data.info.scopes`, because, well, that's not used for anything, and it'd just be way more trouble to change it.  So that's still there with its old value that doesn't account for inheritance, same as in `data.info.scopes.raw`, it just doesn't mean anything.  Hope that's OK.~~

Also note: I branched this off develop, but it's possible we might want to save this for `next`, because it kind of goes along with the other base variable stuff, which I was planning on only doing on `next`?  Probably it makes more sense to do this on develop like I did in fact do it, but just thought I'd throw that out there.

### DIFF
--- a/packages/debugger/lib/data/reducers.js
+++ b/packages/debugger/lib/data/reducers.js
@@ -16,7 +16,6 @@ const DEFAULT_SCOPES = {
 function scopes(state = DEFAULT_SCOPES, action) {
   var scope;
   var variables;
-  var visibles;
 
   switch (action.type) {
     case actions.SCOPE:
@@ -40,10 +39,6 @@ function scopes(state = DEFAULT_SCOPES, action) {
     case actions.DECLARE:
       scope = state.byId[action.node.scope] || {};
       variables = scope.variables || [];
-      visibles = scope.visibles || [];
-
-      let visible = action.node.visibility !== "private";
-      //so, public and internal are both visible
 
       return {
         byId: {
@@ -54,12 +49,9 @@ function scopes(state = DEFAULT_SCOPES, action) {
 
             variables: [
               ...variables,
-              { name: action.node.name, id: action.node.id }
-            ],
 
-            visibles: visible
-              ? [...visibles, { name: action.node.name, id: action.node.id }]
-              : visibles
+              { name: action.node.name, id: action.node.id }
+            ]
           }
         }
       };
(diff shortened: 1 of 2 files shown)


gnidan:
> Also note: I branched this off develop, but it's possible we might want to save this for next, because it kind of goes along with the other base variable stuff, which I was planning on only doing on next? Probably it makes more sense to do this on develop like I did in fact do it, but just thought I'd throw that out there.

Nah, let's keep this targeting `develop`, since it's really kind of like a bug fix.

packages/debugger/lib/data/reducers.js
```diff
@@ -49,9 +54,12 @@ function scopes(state = DEFAULT_SCOPES, action) {
...
             variables: [
               ...variables,
-
               { name: action.node.name, id: action.node.id }
-            ]
+            ],
+
+            visibles: visible
+              ? [...visibles, { name: action.node.name, id: action.node.id }]
+              : visibles
```
gnidan: How do you feel about modifying `variables` to include `visibility` (maybe even the full-form, not just`true`/`false`!), and then just filtering in the selector?

haltman-at: Hm.  Now that you bring it up, I'm realizing this can be done even more simply, without keeping track of any extra information at all.  Because we can filter in the selector just by looking up the visibility in the variable definition, because we have the variable ID and all the definitions!  I think I'm going to go back and change it to do it that way, reverting the reducer to how it was and having all the work occur in the selector.

haltman-at:
OK, I redid this without `visibles` as discussed above.

PR #2426, opened 2019-09-30
Fix: Filter console reading of contracts_build_directory for json

SeanJCasey:
`truffle develop` and `truffle console` currently die silently if using the `contracts_build_directory` config option with a directory that has a system file, e.g., `.DS_Store`.

Filtering for json files only fixes the issue.

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -87,7 +87,10 @@ class Console extends EventEmitter {
   provision() {
     let files;
     try {
-      files = fse.readdirSync(this.options.contracts_build_directory);
+      const unfilteredFiles = fse.readdirSync(
+        this.options.contracts_build_directory
+      );
+      files = unfilteredFiles.filter(file => file.match(".json") !== null);
     } catch (error) {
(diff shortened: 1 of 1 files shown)


eggplantzzz:
Hmm weird. This problem, more generally, seems to occur whenever you have anything else directly inside the `build` folder that you specify. Thanks for reporting this!

packages/core/lib/console.js
```diff
@@ -87,7 +87,10 @@ class Console extends EventEmitter {
   provision() {
     let files;
     try {
-      files = fse.readdirSync(this.options.contracts_build_directory);
+      const unfilteredFiles = fse.readdirSync(
+        this.options.contracts_build_directory
+      );
+      files = unfilteredFiles.filter(file => file.match(".json") != null);
```
eggplantzzz: Just one nit-picky thing...can you use `!==` instead of `!=`?

eggplantzzz:
Thanks for doing this @SeanJCasey!

packages/core/lib/console.js
SeanJCasey: For sure, done!

Issue #2374, opened 2019-09-08
Browser support for @truffle/hdwallet-provider 


mitar:
## Issue

It looks like `@truffle/hdwallet-provider` does not wok in the browser.

## Steps to Reproduce

Try doing `const HDWalletProvider = require('@truffle/hdwallet-provider');` in the browser, in an app bundled with Webpack.

## Expected Behavior

That import will work.

## Actual Results

When I try to do so, I get the error:

```
Uncaught TypeError: Cannot read property 'toString' of undefined
```

mbiegert:
I would like to second this issue. Or are we missing something here?

I get three module not found errors: `child_process`, `fs` and `net`, which is obviously understandable since we are not running on node but in the browser.

CruzMolina:
Hey @mbiegert & @mitar. Currently `@truffle/hdwallet-provider` is not [bundled](https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/webpack/webpack.config.js#L9) for browser usage.

You might want to give [`react-native-truffle-hdwallet-provider`](https://github.com/bpeters/react-native-truffle-hdwallet-provider) a try.

mitar:
It looks like that one requires react-native for crypto, which is not the web platform. But yes, something similar where crypto is provided by WebCrypto should be made.

mbiegert:
> It looks like that one requires react-native for crypto, which is not the web platform. But yes, something similar where crypto is provided by WebCrypto should be made.

Is there something we could possibly do to facilitate that?
Maybe it's worth stating explicitly in the readme that this is not browser compatible. The requirements list `node`, but still I was thinking that it might work at first.

If I get it correctly, one would need to drop dependencies on node only packages (child_process, fs, crypto) and come up with a replacement/use `WebCrypto`.

mitar:
So for many node-only packages you can get browser-compatible replacements which provide the same API. The question is more if some features are needed at all in the browser, like, why is `child_process` even needed?

haltman-at:
So, `@truffle/hdwallet-provider` itself does not rely on `child_process`; rather, some of its dependencies do.  As such, there's not much we can do about this at the moment, but we'll keep this open with the hope that we can fix it in the future.

mitar:
One can always mock them with dummy functions. If `hdwallet-provider` never really invokes that code path, this will work out.

gnidan:
Sounds like that's worth a shot @mitar!

[...]

PR #2427, opened 2019-09-30
@truffle/hdwallet-provider: debug removed (browser support added)

MrRefactoring:
After an investigation, I noticed that after building `webpack`, I get a broken bundle, the reason for this is the `debug` module, which cannot work with the browser.

Therefore, I decided to remove this module and add support for browsers.

Related with #2374 issue

### DIFF
--- a/packages/hdwallet-provider/package.json
+++ b/packages/hdwallet-provider/package.json
@@ -20,7 +20,6 @@
     "any-promise": "^1.3.0",
     "bindings": "^1.5.0",
     "bip39": "^2.4.2",
-    "debug": "^4.1.1",
     "ethereum-protocol": "^1.0.1",
     "ethereumjs-tx": "^2.1.1",
     "ethereumjs-util": "^6.1.0",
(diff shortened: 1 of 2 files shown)


eggplantzzz:
Thanks again @MrRefactoring! There have been some people requesting browser support from this module so I'm sure they will be happy.

MrRefactoring:
> Thanks again @MrRefactoring! There have been some people requesting browser support from this module so I'm sure they will be happy.

I also need this support. I would be happy if it were merged!

eggplantzzz:
Are we sure that `debug` is not compatible with the browser? It seems to have a section on output in the browser [here](https://www.npmjs.com/package/debug#web-browser).

MrRefactoring:
> Are we sure that `debug` is not compatible with the browser? It seems to have a section on output in the browser [here](https://www.npmjs.com/package/debug#web-browser).

Yes but webpack cannot assemble debug without additional manipulations with the config. Probably we can define when will be used package (NodeJS or Browser)? Currently, I do not know how to do it

eggplantzzz:
Ok, that makes sense. Looks good to go to me!

mbiegert:
Very nice, thanks a lot.

PR #2428, opened 2019-10-01
@truffle/hdwallet-provider: additional ignores in .npmignore

MrRefactoring:
Before:
before
After:
after


packages/hdwallet-provider/.npmignore
```diff
@@ -1,2 +1,7 @@
 .idea
 .vscode
+test
+.travis.yml
```
CruzMolina: Oh, there'a  `.travis.yml` in here?!!

Are you open to removing it as part of this PR?

MrRefactoring: Why do we need .travis.yml in the final build of the package?

CruzMolina: Ah, I miscommunicated. I was referring to the pkg itself.

I didn't know there was a `.travis.yml` file in the pkg. We no longer use it (probably leftover from before this pkg was a part of the monorepo). Are you open to removing the `travis.yml` from this pkg as part of the PR?

MrRefactoring: Yeah sure

MrRefactoring: Done

gnidan:
So lean

PR #2430, opened 2019-10-01
Enhancement (breaking for @truffle/provider): Edit and use method to validate provider

eggplantzzz:
This PR updates @truffle/provider to validate user-provided providers. There exists a `test_connection` method that is only used in the test suite and in the interest of providing clearer error messages on having a bad provider, this PR edits this method and uses it in the provider creation flow.

Currently when a bad provider is passed, Truffle migrate dies silently. This new `testConnection` method attempts to get the block number using the supplied provider. If if cannot successfully get the block number, it spits out an error telling the user to check that they've supplied a correct provider URL.


MrRefactoring:
If I understand correctly, then the creation of the provider is carried out through the create function.  If so, then this PR breaks backward compatibility and requires a major upgrade of the package, right?

eggplantzzz:
Yes @MrRefactoring, this PR would be a breaking change since `create` would now be an `async` function.

eggplantzzz:
So actually now I've made just the `testConnection` method async. I aim to add that method to the flow for `truffle migrate` in order to detect for a bad provider and give better error information.

eggplantzzz:
Are we ok with this breaking change @gnidan?

packages/provider/index.js
```diff
@@ -26,19 +29,13 @@ module.exports = {
...
-    var fail = new Error(
-      "Could not connect to your RPC client. Please check your RPC configuration."
-    );
-
-    web3.eth
-      .getCoinbase()
-      .then(coinbase => callback(null, coinbase))
-      .catch(() => callback(fail, null));
+  testConnection: async function(options) {
+    const provider = this.getProvider(options);
```
CruzMolina: Why do we need this `getProvider` call?

CruzMolina: On second thought, I think we should remove it. It couples us more closely to Web3 & EVM-like networks.

eggplantzzz: Why do you think it couples us? This method is our adapter for getting the provider. Sometimes people provide it as a function, sometimes directly. This handles getting whatever provider they give us.

CruzMolina: Most of the time users don't specify a provider in their config. With this code, they'll be hitting the `else` clause in `getProvider`.

Thinking about this some more, why not `try/catch` around the `getBlockNumber` to test the provider?

Actually, we already de facto do a number of provider checks in [`Environment.detect`](https://github.com/trufflesuite/truffle/blob/develop/packages/environment/environment.js#L11-L24). I agree the error messaging should/could be better. It looks like there could be a number of reasons why the provider fails.

eggplantzzz: This PR does not change the logic for creating a provider but just organizes it a bit differently. Are you saying we should change the fundamental logic with how the provider is currently being created in the `create` method?

Are you also saying that it would be better to change the signature of the `testConnection` method? Currently it returns `true` if it is successful and throws an error if not. Perhaps it should be name `connectionValid` and return `true` or `false depending on whether the provider seems valid or not.

I mean, currently there are many cases where the process just dies without giving the user feedback. This was the primary reason I wanted to put in this PR.

CruzMolina: I think the `create` logic is fine for now since we're able to re-route what the provider does via the `Web3Shim`/`InterfaceAdapter`.

I guess not only do I think the implementation of `testConnection` into the `migrate` flow is redundant (it won't get hit until a number of web3 calls have already been made), but having...

wait nm, for non-evm networks we're likely not going to do anything with the provider (for now).

So basically, integrating `testConnection` into the `migrate` flow seems redundant. Do we have reproduction repo's or have we been able to isolate and pinpoint where exactly users are having their process' die?

eggplantzzz: That's a good point that the provider is used before this call so maybe this call to `testConnection` method is kind of nonsensical. I haven't reproduced this in a while and if I remember correctly it was some call to `getId` which might be in `Environment.detect`. 

Let me try and see if I can reproduce that problem later today to see how it might be handled better. Currently if I feed it a bogus provider it seems to error properly with a nice error message. If I can reproduce this problem of it dying silently then maybe that will be a better time to implement this call to test if the provider is good.

eggplantzzz: So if you give it a random url ("www.google.com" for example) it will timeout and give no message. We need to figure out a way to test this. It looks like `getId` will not cut it for `testConnection`.

eggplantzzz:
So I thought 30 seconds was a bit much and gave it a 20 second period to receive a response from the provider. Here are two screenshots of the output (I added a spinner) when it succeeds and fails.
Screen Shot 2019-10-23 at 6 49 38 PM
Screen Shot 2019-10-23 at 6 48 27 PM

eggplantzzz:
I'm going to take out the spinner, it kind of looks out of place in this flow.

eggplantzzz:
Updated output.
Screen Shot 2019-10-24 at 7 20 43 AM
Screen Shot 2019-10-24 at 7 21 30 AM

packages/provider/index.js
```diff
@@ -26,20 +30,35 @@ module.exports = {
         { keepAlive: false }
       );
     }
-
-    return this.wrap(provider, options);
+    return provider;
   },
 
-  test_connection: function(provider, callback) {
-    const interfaceAdapter = new InterfaceAdapter();
```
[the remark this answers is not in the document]
CruzMolina: Yup, just drop it in right above the `Web3Shim` instantiation in the new `testConnection` method.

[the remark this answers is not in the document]
CruzMolina: Just waiting on this.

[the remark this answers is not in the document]
eggplantzzz: Ok Cruz, I just added this instantiation back in. Anything else?

PR #2432, opened 2019-10-02
Internal improvement: Update networks module

eggplantzzz:
This PR updates some code in `@truffle/core` for the `Networks` module found at `lib/networks`. It also adjusts code that calls this module.


packages/core/lib/networks.js
```diff
@@ -1,253 +1,217 @@
...
+            return reject(error);
+          }
+        })
+      );
+    });
 
-          // binaries.map(function(b) {return b.contract_name + ": " + JSON.stringify(b.networks, null, 2)}).forEach(function(b) {
-          //   console.log(b);
-          // });
+    const binaries = await Promise.all(promises);
```
gnidan: Looks like you can just get rid of a lot of this and do:
```javascript
const binaries = files.map(file => {
  const filePath = path.join(options.contracts_build_directory, file);
  const fileContents = fs.readFileSync(filePath, "utf8");
  const body = JSON.parse(fileContents);
  return body;
});  
```

We don't need the extra layer of Promises because it's not actually doing anything async, and that means we don't need the try/catch because any errors will just end up throwing anyway.

packages/core/lib/networks.js
```diff
@@ -1,253 +1,217 @@
...
+    const idsToNames = {};
+    const networks = {};
 
-          Object.keys(options.networks).forEach(function(network_name) {
-            var network = options.networks[network_name];
-            var network_id = network.network_id;
+    // binaries.map(function(b) {return b.contract_name + ": " + JSON.stringify(b.networks, null, 2)}).forEach(function(b) {
+    //   console.log(b);
+    // });
```
gnidan: delete commented code!

packages/core/lib/networks.js
```diff
@@ -1,253 +1,217 @@
...
 
-            if (network_id == null) {
-              return;
-            }
+    Object.keys(options.networks).forEach(networkName => {
```
gnidan: want to just make this a regular `for` loop?

packages/core/lib/networks.js
```diff
@@ -1,253 +1,217 @@
...
-            Object.keys(json.networks).forEach(function(network_id) {
-              var network_name = ids_to_names[network_id] || network_id;
+      idsToNames[networkId] = networkName;
+      networks[networkName] = {};
+    });
 
-              if (networks[network_name] == null) {
-                networks[network_name] = {};
-              }
+    binaries.forEach(json => {
```
gnidan: also this one could be a regular `for` loop

packages/core/lib/networks.js
```diff
@@ -1,253 +1,217 @@
...
+  },
 
-      var configured_networks = Object.keys(config.networks);
-      var promises = [];
+  clean: async function(config) {
+    let files = fs.readdirSync(config.contracts_build_directory);
+    const configuredNetworks = Object.keys(config.networks);
+    const promises = [];
+
+    files.forEach(file => {
```
gnidan: (You can probably also make this loop nicer)

packages/core/lib/networks.js
eggplantzzz: yessir

PR #2434, opened 2019-10-02
Use absolute paths in @truffle/codec

gnidan:
No one likes `import * as foo from "../../../really"`


gnidan:
@haltman-at another rubber stamp, please?

gnidan:
> it would be better to make internal imports not look like external imports.

Agreed. I'll do that on my other branch. Merging this one. Thanks!

PR #2435, opened 2019-10-02
Internal improvement: use latest hyperledger fabric binary

CruzMolina:


CruzMolina:
test again

CruzMolina:
Closing again since it looks like the latest fabric-evm binary still has the same bug as before.

PR #2437, opened 2019-10-07
Internal improvement: Update Require.exec interface

eggplantzzz:
Allow Require.exec to accept a Promise-based interface and use it in lib/commands/exec.js

### DIFF
--- a/packages/require/require.js
+++ b/packages/require/require.js
@@ -94,29 +94,32 @@ const Require = {
   },
 
   exec: function(options, done) {
-    expect.options(options, [
-      "contracts_build_directory",
-      "file",
-      "resolver",
-      "provider",
-      "network",
-      "network_id"
-    ]);
-
-    const web3 = new Web3Shim({
-      provider: options.provider,
-      networkType: options.networks[options.network].type
-    });
-
+    const callbackPassed = typeof done === "function";
     try {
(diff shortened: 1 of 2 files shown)


packages/require/require.js
```diff
@@ -94,29 +94,32 @@ const Require = {
...
+      });
+
       const fn = this.file({
         file: options.file,
         context: { web3 },
         resolver: options.resolver
       });
-      fn(done);
+      if (callbackPassed) return fn(done);
+      return fn;
```
gnidan: Hm, so there's four cases to consider here... does this cover all of them?

- `Require.exec` can be called with our without a callback
- `fn` can be written to use a callback or written to return a promise

It looks like this assumes that if `Require.exec` is not given a callback, then `fn` shouldn't be written to expect a callback. I think this is not correct; we should provide a callback to `fn` whether or not `Require.exec` was given one.

Thoughts?

eggplantzzz: I think it should cover everything as long as you know that if you use the promise interface (for `Require.exec`) you need to pass a callback to the returned function.

This is kind of a weird signature in Truffle, I'm not sure we have too many other places where this happens. Maybe this should be left alone and we can just use `promisify` on the outside.

PR #2412, opened 2019-09-24
@tuffle/hdwallet-provider: migration to typescript

MrRefactoring:
Migration `@truffle/hdwallet-provider` package to typescript

PR #2438, opened 2019-10-08
Bug fix: Downgrade dependency for Geth compatibility

eggplantzzz:
Downgrade ethereumjs-tx as the newer version was not working with geth chains (Rinkeby).

### DIFF
--- a/packages/hdwallet-provider/package.json
+++ b/packages/hdwallet-provider/package.json
@@ -21,7 +21,7 @@
     "bindings": "^1.5.0",
     "bip39": "^2.4.2",
     "ethereum-protocol": "^1.0.1",
-    "ethereumjs-tx": "^2.1.1",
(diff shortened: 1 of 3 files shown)


cgewecke:
@eggplantzzz Out of curiosity, do you know what the incompatibility introduced by the newest version was?

eggplantzzz:
The last working version was 1.0.18 I think. So when the module was ported to TS [here](https://github.com/trufflesuite/truffle/pull/2412) something must have changed. I thought I had found the problem but it seems like people are still having trouble. I cannot really find a reliable way to reproduce this issue as it happens intermittently for me. As of now, I am able to successfully migrate a toy project to both Ropsten and Rinkeby with 1.0.21 and not 1.0.20. So I'm not certain about what is going on here.

PR #2439, opened 2019-10-08
Dependency update: Remove unused dep

eggplantzzz:
Remove truffle-init which seems to be an unused dependency.

I noticed in some of the logging output that some dependencies are still using the old non-org-scoped packages. In the interest of getting rid of these warnings, truffle-init has been removed as it is not used anywhere.


eggplantzzz:
Wow, that's a random catch!

Issue #1698, opened 2019-02-05
Dry run fails (false negative) when tx success depends on deployer address


MichalZalecki:
## Issue

When transaction within the migration depends on deployer's address (e.g. ownership transfer) then dry run fails.

## Steps to Reproduce

Try to transfer ownership as part of the migration.

## Expected Behavior

The same set of `accounts` is used in the dry run as well as during the actual migration.

## Environment

* Operating System: macOS 10.14.3
* Ethereum client: Geth 1.8.22-stable
* Truffle version: v5.0.2
* node version: 10.15.0
* npm version: 6.7.0

mrwillis:
Similar issue here:

https://github.com/trufflesuite/truffle/issues/1612#issuecomment-459273927

mrwillis:
I notice also the use of network names as `network-fork` can really mess up the logic if you have network-dependent logic.

mrwillis:
The best solution I have for anyone facing this issue (which only seems to occur past v5.x truffle) is to just split up your migration files for mainnet with 1-2 transactions each. This way you won't have to restart if something goes wrong.

zulhfreelancer:
Add `skipDryRun: true` in your network config. That should help.

Example:

```
development: {
      skipDryRun: true,
      host: "127.0.0.1",
      port: 7545,
      network_id: "*" // match any network
 },
```

MichalZalecki:
@zulhfreelancer Sure, it's easy to "skip" the problem, but it doesn't solve the problem with dry run.

eggplantzzz:
Hey @MichalZalecki, do you have a specific example that I can take a look at to reproduce the problem?

AleG94:
+1

To reproduce the issue, try to deploy a simple erc20 and call the transfer function from deployer address.
Skipping dry run works.

eggplantzzz:
@seesemichaelj Did you do something related to this issue?

seesemichaelj:
Nope, it hit the back burner

eggplantzzz:
@seesemichaelj Do you have more information about this issue that you could link here?

seesemichaelj:
This is what @nicholasjpaterno wrote on our other repo about the issue:

To add more background info, echoing @seesemichaelj's comments on this issue for solidarity:

[...]

PR #2441, opened 2019-10-09
Bug fix: Don't pass Ganache an empty array for "unlocked_accounts"

eggplantzzz:
Reorganize how Ganache gets options passed to avoid passing an empty array for the `unlocked_accounts` property. This apparently is causing a bug that will be fixed on the Ganache end as well. See the discussion [here](https://github.com/trufflesuite/truffle/issues/1698) for more information.

### DIFF
--- a/packages/environment/environment.js
+++ b/packages/environment/environment.js
@@ -38,14 +38,15 @@ const Environment = {
     const upstreamNetwork = config.network;
     const upstreamConfig = config.networks[upstreamNetwork];
     const forkedNetwork = config.network + "-fork";
+    const ganacheOptions = {
+      fork: config.provider,
+      gasLimit: block.gasLimit
+    };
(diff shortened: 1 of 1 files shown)


packages/environment/environment.js
```diff
@@ -38,14 +38,15 @@ const Environment = {
     const upstreamNetwork = config.network;
     const upstreamConfig = config.networks[upstreamNetwork];
     const forkedNetwork = config.network + "-fork";
+    const ganacheOptions = {
+      fork: config.provider,
+      gasLimit: block.gasLimit
+    };
+    if (accounts.length > 0) options.unlocked_accounts = accounts;
```
CruzMolina:
```suggestion
    if (accounts.length > 0) ganacheOptions.unlocked_accounts = accounts;
```

eggplantzzz: Ugh, very silly.

Issue #2431, opened 2019-10-02
@truffle/hdwallet-provider does not work with Geth, err="invalid sender"


salanfe:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

conversation link: https://gitter.im/ConsenSys/truffle?at=5d94748c6d4b2512c1519b87

---------------------------

## Issue

I'm trying to migrate a helloworld smart contract on my local ethereum dev/test net (geth + genesis with my keys) using truffle. The migration fails when the package `@truffle/hdwallet-provider` is used but works with the old package `truffle-hdwallet-provider`. Strangely it works fine for both packages when deployed on ganache, so it looks like it's a geth specific problem for `@truffle/hdwallet-provider`

## Steps to Reproduce

Here is my `truffle-config.js` file:
```js
const HDWalletProvider = require('@truffle/hdwallet-provider');
const privateKey = "e46cae82c1e350c7b00070a2b8414dead89b2d768f1c09410b668d21cd4b4eea"; // dummy testing key
module.exports = {
  networks: {
    gethtestnet: {
      provider: () =>
          new HDWalletProvider(privateKey, "http://geth-test-net.k8s.local/rpc"),
      network_id: 3516,
    }
  },
  compilers: {
    solc: {
      version: "0.5.10",
      settings: {
       evmVersion: "petersburg"
      }
    }
  }
}
```

(if anyone wonders about my hostname, my node runs in docker-desktop kubernetes and an `nginx ingress` maps `/rpc` against `:8545`)

here is my genesis file (might this help)

```json
{
    "config": {
        "chainId": 3516,
        "homesteadBlock": 0,
        "eip150Block": 0,
        "eip150Hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
        "eip155Block": 0,
        "eip158Block": 0,
        "byzantiumBlock": 0,
        "constantinopleBlock": 0,
        "petersburgBlock": 0,
        "clique": {
            "period": 1,
            "epoch": 30000
        }
    },
    "nonce": "0x0",
    "timestamp": "0x5cacf490",
    "extraData": "0x000000000000000000000000000000000000000000000000000000000000000017f934cD8A8b086F772Bc11C42Bb7c2c0310320C0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "gasLimit": "0x989680",
    "difficulty": "0x1",
[...]

PR #2456, opened 2019-10-15
@truffle/hdwallet-provider: send method fixed

MrRefactoring:
Added missed `callback` argument (#2431)

### DIFF
--- a/packages/hdwallet-provider/src/index.ts
+++ b/packages/hdwallet-provider/src/index.ts
@@ -175,8 +175,11 @@ class HDWalletProvider {
     this.engine.start(); // Required by the provider engine.
   }
 
-  public send(payload: JSONRPCRequestPayload): void {
-    return this.engine.send.apply(this.engine, [payload]);
+  public send(
+    payload: JSONRPCRequestPayload,
+    callback: (error: null | Error, response: JSONRPCResponsePayload) => void
+  ): void {
+    return this.engine.send.apply(this.engine, [payload, callback]);
   }
 
   public sendAsync(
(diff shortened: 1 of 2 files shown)


eggplantzzz:
Oh is this the problem people are experiencing [here](https://github.com/trufflesuite/truffle/issues/2431)?

MrRefactoring:
> Oh is this the problem people are experiencing [here](https://github.com/trufflesuite/truffle/issues/2431)?

Exactly!

eggplantzzz:
Oh that is super, I've been trying to figure out what is going on. Are you able to resolve the problem in the CI? Thanks @MrRefactoring!!

MrRefactoring:
@eggplantzzz I made some corrections and type improvements, I started it, but the pipeline does not work. You can restart the pipeline because the error seems strange to me. Locally, everything works

MrRefactoring:
Should I do something or is this a problem on the part of CI?

eggplantzzz:
I just restarted it. I don't think it is related to you, I think something was up with the CI mechanisms themselves.

gnidan:
it's green!

MrRefactoring:
Hooray! 📣

MrRefactoring:
I used debugger to find the cause of the error. On a test example from @gjgd I checked the working ability of the new version

gnidan:
Can we get at least one or two more confirmations that this fixes #2431? I'm in line with Cruz's thinking that we should avoid releasing another buggy version.

gnidan:
Screen Shot 2019-10-15 at 6 44 12 PM

PR #2460, opened 2019-10-16
truffle-db/refactor/tests

cds-amal:
This PR refactors the workspace/test/index module into multiple smaller modules that can be tracked/modified independently. It also keeps the graphql queries and mutations where they are immediately tested, and exports them so they can be used to set up other tests.

jest will continue to match **/test/index, but now allows for test modules that end with '.spec.(js|ts)' or  '.test.(js|ts)'


packages/truffle-db/src/workspace/test/bytecode.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      bytecodes: [{
+        bytes: $bytes
+      }]
+    }) {
+      bytecodes {
+        id
+      }
+    }
+  }
+`
```
fainashalts: need new line

packages/truffle-db/src/workspace/test/compilation.spec.ts
```diff
@@ -0,0 +1,155 @@
...
+          }
+          ast {
+            json
+          }
+          name
+        }
+      }
+    }
+  }
+`;
```
fainashalts: need new line

packages/truffle-db/src/workspace/test/contract.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+          createBytecode {
+            bytes
+          }
+        }
+      }
+    }
+  }
+`;
+
+
```
fainashalts: extra new line

packages/truffle-db/src/workspace/test/contractInstances.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+              bytes
+            }
+          }
+        }
+      }
+    }
+  }
+`;
+
+
```
fainashalts: extra new line

packages/truffle-db/src/workspace/test/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+      request,
+      null, // root object, managed by workspace
+      { workspace: this.workspace }, // context vars
+      variables
+    );
+    return result.data;
+  }
+}
+
+export const Migrations = require(path.join(fixturesDirectory, "Migrations.json"));
```
fainashalts: need new line

fainashalts:
This looks great, @cds-amal! I made a comment about new lines just because this used to always happen to me until I changed a setting in my IDE to do it automatically. :) 

I think its awesome to pull out each resource this way and will open the path for making the tests even more modular in the future. I think this can get a bit more complicated in the loader tests so let's chat about it when/if you decide to modularize that. Thanks for this!

packages/truffle-db/package.json
```diff
@@ -89,8 +89,10 @@
       "^test/(.*)": "<rootDir>/test/$1"
     },
     "testMatch": [
-      "<rootDir>/src/**/test/*.+(ts|js)",
-      "<rootDir>/test/**/test/*.+(ts|js)"
```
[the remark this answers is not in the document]
cds-amal: The previous regex did match every test in the matched folder. This new way, allows us to specify tests we want to match by naming convention (ends with .(test|spec).ts or .(test|spec).js. For example, the workstation/test folder looks like this: 
```
test
├── bytecode.spec.ts
├── compilation.spec.ts
├── contract.spec.ts
├── contractInstances.spec.ts
├── contractNames.spec.ts
├── network.spec.ts
├── source.spec.ts
└── utils.ts
```

`utils.ts` will not be matched by the test runner. This allows us to put helpers, or refactor common logic into files that won't be matched but can be imported into the actual spec, or test files.

packages/truffle-db/src/workspace/test/bytecode.spec.ts
cds-amal: This is strange.  `truffle-db` hasn't merged in the lint-staged config setup for TS files from `develop`. My expectation is that it would be handled by that, though I'll test to verify.

PR #2462, opened 2019-10-16
@truffle/hdwallet-provider: improving performance

MrRefactoring:
I replaced `apply` with `call`, as this is a more performance solution

You can find confirmations [here](https://jsperf.com/call-apply-segu) or search for: `call vs apply performance`


MrRefactoring:
I checked it locally and everything works without errors!

PR #2465, opened 2019-10-16
Internal improvement: Refine test flow

eggplantzzz:
Clean up the test command file to increase readability and clarity.


packages/core/lib/commands/test.js
```diff
@@ -67,81 +63,27 @@ const command = {
...
+        .then(() => copyArtifactsToTempDir(config))
+        .then(({ config, temporaryDirectory }) => {
+          return prepareConfigAndRunTests({
+            config,
+            files,
+            temporaryDirectory
+          });
+        })
+        .then(numberOfFailures => {
+          done.call(null, numberOfFailures);
```
eggplantzzz: One questionable thing here; I changed `apply` to `call` and did not pass `arguments` which might allow for passing of things other than just the number of failures. I couldn't see anything else that might be passed besides the failures but maybe I am missing something? This way seemed much cleaner and self-descriptive so I favored it.

fainashalts: The difference is just in how the arguments are passed, right? I prefer `call` personally, agree that it is cleaner.

packages/core/lib/commands/test.js
```diff
@@ -42,17 +42,13 @@ const command = {
...
-    const Test = require("../test");
-    const fs = require("fs");
-    const { promisify } = require("util");
-    const promisifiedCopy = promisify(require("../copy"));
     const { Environment, Develop } = require("@truffle/environment");
+    const {
+      copyArtifactsToTempDir,
+      determineTestFilesToRun,
+      prepareConfigAndRunTests
+    } = this.helpers;
```
fainashalts: So much cleaner.

packages/core/lib/commands/test.js
```diff
@@ -67,81 +63,27 @@ const command = {
...
-    }
-
-    files = files.filter(function(file) {
-      return file.match(config.test_file_extension_regexp) != null;
-    });
-
-    let temporaryDirectory;
     try {
-      temporaryDirectory = temp.mkdirSync("test-");
+      files = determineTestFilesToRun(options, config);
```
gnidan: Think it's safe to use just the `config` here, not the `options`? We've already `detect`ed against the options.

I can't decide what I think is better; what do you think?

packages/core/lib/commands/test.js
gnidan: Hm, the only consideration I'd add here about `call` vs. `apply` is this that `call` is faster, as @MrRefactoring brought to our attention in https://github.com/trufflesuite/truffle/pull/2462

packages/core/lib/commands/test.js
```diff
@@ -162,11 +104,83 @@ const command = {
...
+              done.call(null, numberOfFailures);
+              ipcDisconnect();
+            })
             .catch(done);
         }
       );
     }
+  },
+
+  helpers: {
```
gnidan: Why not just move these to `lib/test/`? That way you can have the imports at the top of the file, all in one place, like one might expect.

packages/core/lib/commands/test.js
MrRefactoring: I think it is preferable to use call, as it works faster and is more popular among developers.

I think it makes sense to use apply only in one case if you have an array of arguments. This will be faster than using the call and spread operator at the same time.

packages/core/lib/commands/test.js
eggplantzzz: What about moving all test stuff into a `test` folder? I've wanted to start this pattern in `core`. So instead of having a folder with all the command files, you would have a bunch of directories with directories for each command. I would like to save `lib` for stuff that is used by more than one command. How does that seem?

Issue #2291, opened 2019-08-13
Contract objects made with Contract.deployed() are missing transactionHash


haltman-at:
## Issue

Truffle contract objects made with `Contract.new()` have a field `transactionHash` which gives the transaction hash of the contract's creation.  Contract objects made with `Contract.deployed()`, though, are missing this field (or rather it's equal to `undefined`).

Note: These are not the only two ways to create contract objects.  For instance, there's also `Contract.at()`.  But it's probably not possible for `Contract.at()` to find this information in arbitrary cases.  That said, perhaps it could also still do it in cases where the information is somehow available.

(Are there other ways of creating contract objects?  Those are the ones I'm familiar with...)

## Steps to Reproduce

Deploy a contract (let's call it `Contract`) with `truffle migrate` and then, in `truffle console`, type `await Contract.deployed()`.  The result will have its `transactionHash` field undefined.

## Expected Behavior

The contract object should have its `transactionHash` field equal to the transaction hash of the contract's deployment.

## Actual Results

It's undefined.

## Environment

* Operating System: Linux Mint 19.1 64-bit
* Ethereum client: Ganache CLI v6.4.3 (ganache-core: 2.5.5)
* Truffle version (`truffle version`): 5.0.31
* node version (`node --version`): 10.15.3
* npm version (`npm --version`): 6.4.1

eggplantzzz:
So I'm proposing that the answer to this is to store a property like `creationTransactionHash` in the `networks.<id>` property in the artifact.

CruzMolina:
Looks like this already part of the artifact spec as `transactionHash`: https://github.com/trufflesuite/truffle/blob/develop/packages/contract-schema/spec/network-object.spec.json#L9

I imagine the fix is to add the hash in same area where events are saved in the codebase.

CruzMolina:
Closing since I believe this has been resolved by #2466.

PR #2466, opened 2019-10-16
Enhancement: Ensure all contract objects return a txHash

CruzMolina:
This PR addresses #2291 by making sure a `transactionHash` is attached to a contract object when it's successfully returned. Ensures `MyContract.at`, `MyContract.new` & `MyContract.deployed` calls return the same expected form of contract object.


packages/contract/lib/contract.js
```diff
@@ -20,13 +20,6 @@ if (typeof Web3 === "object" && Object.keys(Web3).length === 0) {
     var instance = this;
     var constructor = instance.constructor;
```
eggplantzzz: Will this be a breaking change removing this functionality? I'm not sure if it is useful for external use or if it is even documented.

CruzMolina: Good question. It's at least an internal/module-specific breaking change. I think the benefit of normalizing the documented & exposed API behavior outweighs the rare possible usage of something like `const newInstance = new SimpleStorage(SimpleStorage.address)` vs `const newInstance = SimpleStorage.at(SimpleStorage.address)`.

I don't think we really want or expect people to do the former.

gnidan: It is a breaking change! Also, there's https://en.m.wikipedia.org/wiki/Robustness_principle, which says to be liberal in what inputs you accept.

I am against removing this.

CruzMolina: Good point. Will rework the code to be non-breaking.

eggplantzzz: By the way, I'm not against your renaming the input to this function to more accurately describe what it is. When the argument says "contract" I expect it to be a contract, not a string representation of the address. This is the kind of stuff that makes the code super confusing and hard to read. :)

packages/contract/lib/contract.js
```diff
@@ -20,13 +20,17 @@ if (typeof Web3 === "object" && Object.keys(Web3).length === 0) {
...
+    // in this case contract is a contract address string
     if (typeof contract === "string") {
-      var web3Instance = new constructor.web3.eth.Contract(constructor.abi);
-      web3Instance.options.address = contract;
-      contract = web3Instance;
+      const contractInstance = new constructor.web3.eth.Contract(
+        constructor.abi
+      );
+      contractInstance.options.address = contract;
+      if (constructor._json.networks[constructor.network_id])
```
haltman-at: May want braces here to avoid future problems

CruzMolina: >Or would it be possible to only populate transactionHash when the address passed to at matches the address in the artifact, in which case you can just look up the transaction hash there?

This makes sense!

CruzMolina: Actually, thinking about this some more. `MyContract.at('0x...')` will *only* return with a `transactionHash` if it's been deployed with Truffle to the current network (via `network_id`) that you're connected/connecting to.

So for example, if you used Truffle to migrate `MyContract` to Ropsten, but then try to get an instance of `MyContract` that you deployed to Rinkeby using something _other_ than Truffle (i.e. Remix), that contract's information won't be stored in the artifacts, so if you call `MyContract.at()` to the address on Rinkeby, the `transactionHash` won't be available.

haltman-at: But there could be multiple instances of a contract on the same network.  Suppose you used Truffle to deploy `MyContract` to Rinkeby, and then deployed another copy of `MyContract` to Rinkeby by a different means; say for instance that it was created as part of a transaction and you got the address from an event.  Using `MyContract.at()` on that second address shouldn't give you a contract with the `transactionHash` of the first.

CruzMolina: Good point!

packages/contract/lib/contract.js
```diff
@@ -20,13 +20,20 @@ if (typeof Web3 === "object" && Object.keys(Web3).length === 0) {
...
-      contract = web3Instance;
+      const contractInstance = new constructor.web3.eth.Contract(
+        constructor.abi
+      );
+      contractInstance.options.address = contract;
+      if (
+        constructor._json.networks[constructor.network_id] &&
+        constructor._json.networks[constructor.network_id].address === contract
+      )
+        contractInstance.transactionHash = constructor.transactionHash;
```
haltman-at: Again, might want braces here to prevent confusion

PR #2470, opened 2019-10-17
Bug fix: Filter out non-json files in build folder

eggplantzzz:
This ignores all non-json files in the build folder when loading artifacts during testing.

### DIFF
--- a/packages/core/lib/testing/testsource.js
+++ b/packages/core/lib/testing/testsource.js
@@ -27,6 +27,10 @@ TestSource.prototype.resolve = function(importPath, callback) {
         abstractionFiles = fse.readdirSync(
           self.config.contracts_build_directory
         );
+        // Filter out non-json files
+        abstractionFiles = abstractionFiles.filter(file => {
+          return file.match(/^.*.json$/);
+        });
       } catch (error) {
         return callback(error);
       }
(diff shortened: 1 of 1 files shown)


packages/core/lib/testing/testsource.js
```diff
@@ -24,9 +24,13 @@ TestSource.prototype.resolve = function(importPath, callback) {
 
       let abstractionFiles;
       try {
-        abstractionFiles = fse.readdirSync(
+        const buildDirFiles = (abstractionFiles = fse.readdirSync(
           self.config.contracts_build_directory
-        );
+        ));
+        // Filter out non-json files
+        abstractionFiles = buildDirFiles.filter(file => {
```
fainashalts: Could/should this be `abstractionFiles = buildDirFiles.filter(file => file.match(/^.*.json$/);`? Its not particularly important, I just find it a little cleaner to do that if the only thing in the curly braces is the return statement.

eggplantzzz: Sure, I just know that prettier is going to do something dumb-looking like
```
...buildDirFiles.filter(file =>
      file.match(....
```
heheh

fainashalts: Haha, good point!

fainashalts:
👍 This looks good to me!

packages/core/lib/testing/testsource.js
eggplantzzz: I went ahead and changed it be a one-liner. Thanks @fainashalts!

Issue #2449, opened 2019-10-13
Truffle error when executing method call. Probably, error with event parsing.


chebykin:
- [ no] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

When calling a contract method in my tests I'm getting `Error: TypeError: Cannot read property 'toString' of null Given value: "null"`.

What I've found is when two different contracts have two events with the same signatures but different argument names, truffle parses the event using a wrong argument set. In my case the event is ERC721 Transfer:

`event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);`

but truffle parses it as ERC20 Transfer:

`event Transfer(address indexed from, address indexed to, uint256 amount);`

## Steps to Reproduce

Use the following set of contracts:

```solidity
pragma solidity ^0.4.24;


contract MyERC20 {
  event Transfer(address indexed from, address indexed to, uint256 amount);

  function transferFrom(address _from, address _to, uint256 _amount) public {
    emit Transfer(_from, _to, _amount);
  }
}

contract MyERC721 {
  event Transfer(address indexed from, address indexed to, uint256 indexed tokenId);

  function transferFrom(address _from, address _to, uint256 _tokenId) public {
    emit Transfer(_from, _to, _tokenId);
  }
}


contract Foo is MyERC20 {
  MyERC721 public erc721Contract;

  constructor() public {
    erc721Contract = new MyERC721();
  }

  function myTransfer() public {
    erc721Contract.transferFrom(msg.sender, address(this), 1);
  }
}
```

and run this test:

```javascript
const Foo = artifacts.require('Foo.sol');

contract('Foo test', accounts => {
  it('should execute #myTransfer()', async function() {
    const foo = await Foo.new();
    await foo.myTransfer();
  });
});
```

## Expected Behavior

The test should pass without errors

## Actual Results

The test fails with 
```
  1) Contract: Foo test
       should execute #myTransfer():
     Error: TypeError: Cannot read property 'toString' of null Given value: "null"
      at Object.toBN (node_modules/web3-utils/src/utils.js:64:15)
      at _convertNumber (node_modules/@truffle/contract/lib/reformat.js:21:24)
      at abiSegment.forEach (node_modules/@truffle/contract/lib/reformat.js:77:33)
      at Array.forEach (<anonymous>)
      at Function.numbers (node_modules/@truffle/contract/lib/reformat.js:52:14)
[...]

PR #2474, opened 2019-10-18
Workarounds for test event errors

haltman-at:
This PR provides some temporary workarounds for #2449 until the new decoder is ready.

In the test runner, this consists of putting a try/catch around event decoding and just printing a warning if it fails.

In `contract`, this consists of moving the number formatting call into the try/catch I earlier put around event decoding (thanks @eggplantzzz for figuring out this problem).

### DIFF
--- a/packages/contract/lib/utils.js
+++ b/packages/contract/lib/utils.js
@@ -66,10 +66,14 @@ const Utils = {
         let logArgs;
         try {
           logArgs = abi.decodeLog(logABI.inputs, copy.data, copy.topics);
+          copy.args = reformat.numbers.call(
+            constructor,
+            logArgs,
+            logABI.inputs
+          );
         } catch (_) {
           return null;
         }
-        copy.args = reformat.numbers.call(constructor, logArgs, logABI.inputs);
 
         delete copy.data;
         delete copy.topics;
--- a/packages/core/lib/testing/testrunner.js
+++ b/packages/core/lib/testing/testrunner.js
@@ -206,6 +206,10 @@ TestRunner.prototype.endTest = function(mocha, callback) {
         } catch (_) {
           //temporary HACK until we're using the new decoder
           self.logger.log(`    Warning: event decoding failed`);
+          self.logger.log(
+            `    (This may be due to multiple events with same signature`
+          );
+          self.logger.log(`    or due to unsupported data types)`);
           return;
         }
 
--- a/packages/core/lib/testing/testrunner.js
+++ b/packages/core/lib/testing/testrunner.js
@@ -196,11 +196,18 @@ TestRunner.prototype.endTest = function(mocha, callback) {
           return input.type;
         });
 
-        var values = abi.decodeLog(
-          event.abi_entry.inputs,
-          log.data,
-          log.topics.slice(1) // skip topic[0] for non-anonymous event
-        );
+        var values;
+        try {
+          values = abi.decodeLog(
+            event.abi_entry.inputs,
+            log.data,
+            log.topics.slice(1) // skip topic[0] for non-anonymous event
+          );
+        } catch (_) {
+          //temporary HACK until we're using the new decoder
+          self.logger.log(`    Warning: event decoding failed`);
+          return;
+        }
 
         var eventName = event.abi_entry.name;
         var eventArgs = event.abi_entry.inputs


packages/contract/lib/utils.js
```diff
@@ -66,10 +66,14 @@ const Utils = {
         let logArgs;
         try {
           logArgs = abi.decodeLog(logABI.inputs, copy.data, copy.topics);
+          copy.args = reformat.numbers.call(
```
eggplantzzz: I don't know if it is useful to have the rest of the params decoded, but if they are useful you could put a guard in `_convertNumber` in `contracts/lib/reformat.js` to return something benign like `""`. That way you could save the rest of the info if it can be useful later.

packages/core/lib/testing/testrunner.js
```diff
@@ -196,11 +196,18 @@ TestRunner.prototype.endTest = function(mocha, callback) {
...
+        var values;
+        try {
+          values = abi.decodeLog(
+            event.abi_entry.inputs,
+            log.data,
+            log.topics.slice(1) // skip topic[0] for non-anonymous event
+          );
+        } catch (_) {
+          //temporary HACK until we're using the new decoder
+          self.logger.log(`    Warning: event decoding failed`);
```
eggplantzzz: Maybe this should be something more informative to the user. I think maybe the message should be something that let's them know "Hey, the decoding failed but it is not that big a deal. This just happens in certain circumstances when you have similarly named events."

I'm not sure how to say that succinctly haha. Maybe
```
Warning: event decoding failed. 
This sometimes happens with events with duplicate names.
```

What do you think?

eggplantzzz:
Do you think we need more robust commenting just in case the future work to replace these gets pushed way down the line?

packages/core/lib/testing/testrunner.js
haltman-at: I mean, that's kind of what "Warning" means, isn't it?  There are other reasons decoding can fail as well.

Note that this message doesn't actually come up in this case!  The problem here is in entirely in `contract`, not the test runner.  I just left this in here to be safe, basically?  <shrug>  I'm not very clear on when duplicate names cause problems and when they don't...

packages/contract/lib/utils.js
haltman-at: Hm, maybe?  I think if something goes wrong here the results are largely going to be nonsense so I'm not sure how much it matters.  This is just a temporary workaround anyway.

haltman-at:
Could add more comments if you really think it's necessary, but when this gets replaced it's going to get replaced wholesale, so not sure how much it is.

packages/contract/lib/utils.js
eggplantzzz: That's not really true, here the other values do get decoded going into `reformat.numbers.call`. In other words, the other values are properly decoded but there is just a problem with the data for the indexed/non-indexed value(s) not being where Truffle expects it. Again, I'm not sure if this is useful or not so this commend doesn't need to hold this up. Especially if this is going to be pull out soon.

packages/core/lib/testing/testrunner.js
eggplantzzz: This won't come up? If someone is using the `--show-events` flag then they will get the report at the end of the test runner feedback. I really just wanted to let people know that it is more a Truffle problem at this point rather than being something that they are doing wrong and should change. But again, if you disagree then don't let this hold up moving this forward.

haltman-at: Sorry, when I say "in this case" I don't mean it won't come up at all (otherwise there'd be no point!), I mean it doesn't come up in the reported case that led to this.  But I didn't try it with `--show-events` so uhhh I guess maybe it comes up then, I should try that.

I just don't like splitting the message over two lines honestly, or suggesting a particular thing as the cause when there are multiple possible causes.  But maybe we can have something that fits on one line and still conveys the same idea.  E.g., "Event decoding failed due to internal Truffle error"?

packages/contract/lib/utils.js
eggplantzzz: No but the thing is, check out the return values from `abi.decodeLog`. That is not where it is failing. All can be legit except the ones that it doesn't know where to get.

Like, for example the test function I was using had two events with the same signature but the third argument was indexed in one and not the other. The first two were decoded properly while the 3rd value was `null`.

haltman-at: Yes, that what happens in *this* case, but this is a particularly nice case where the only difference is at the end.  In general, one can get nonsense values.  Like imagine what would happen if the indexing difference were in the first argument rather than the third -- you'd have a shift in the arguments, yielding nonsense values.  Or imagine that not all arguments were the same type.  Etc.  I don't think it makes sense to make an effort to report values that may be nonsense, just because it's *possible* that they're valid.

packages/core/lib/testing/testrunner.js
eggplantzzz: I'm really thinking about the fact that I think people are going to start reporting this to us after we implement the message. I was just thinking if we provided them with the explanation that it might "head them off at the pass" and they wouldn't need to report it if it was the case that they named their events with the same names.

haltman-at: Oh, that's a good point.  I hadn't thought of that.  OK, yeah, I guess we do need to say something like that.  But I'll try to word it in a way that suggests the possibility of other causes. :)

packages/contract/lib/utils.js
eggplantzzz: Ok that's fine. I'm not even sure that it is that useful to decode the ones that you can haha. Looks fine to me.

packages/core/lib/testing/testrunner.js
eggplantzzz: Sounds cool, either way I won't hold it back any more. Looks good

haltman-at: Huh, testing it out with `--show-events` on, it indeed does not come up in this case.  I guess due to some difference between how `contract` and the test runner handle things.  Weird.  Well, in any case, I've updated the message.

PR #2477, opened 2019-10-21
support command-line args as `--option=value` #2173

divyanshu1610:
This PR is for #2173
Supports command line arguments also as `--option=value` (not just  `--option value`)
If argument is provided as `--option=true` then only `--option` is passed down as argument
A separate file **packages/core/utils.js** consists the code for this

### DIFF
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -10,6 +10,8 @@ const version = require("./lib/version");
 const versionInfo = version.info();
 const XRegExp = require("xregexp");
 
+const { reformatArguments } = require("./utils");
+
 // pre-flight check: Node version compatibility
 const minimumNodeVersion = "8.9.4";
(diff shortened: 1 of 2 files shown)


eggplantzzz:
@divyanshu1610 Thanks for doing this work but that format is already supported in Truffle. The `yargs` package parses and takes care of all that business. I just confirmed it with `truffle migrate --network=development`

gnidan:
@eggplantzzz Truffle was reporting that "unknown option" warning when using equals syntax... I think that's why we opened #2173

divyanshu1610:
@eggplantzzz Yes , `yargs` is working fine but truffle was showing "unknown option" warning as @gnidan said. This PR solves this issue but I've found better way to solve it. I'll fix it soon !

PR #2479, opened 2019-10-21
fixed `unknown option` when `--option=value` was used #2173

divyanshu1610:
This PR is for #2173 
With the command line args with flags passed as `--option=value`  truffle showed a warning of `unknown option`.
This is fixed in this PR.
`packages/core/lib/utils` has utility file for same.

### DIFF
--- a/packages/core/lib/utils/utils.js
+++ b/packages/core/lib/utils/utils.js
@@ -1,27 +1,3 @@
-// Regex for checking --option=value
-const argTester = /--[a-zA-Z0-9]+[-]*[a-z0-9A-Z]*=[a-zA-Z0-9.-:]+/;
-
-// Reformats the arguments --option=value to  --option value
-const reformatArguments = inputArguments => {
-  inputArguments.map((arg, i) => {
-    // If arg is in form --option=value
-    if (argTester.test(arg)) {
-      if (arg.replace(argTester, "").length === 0) {
-        let splittedArgs = arg.split("=");
-        // Check if value=true then don't include the value  in inputArguments
-        if (splittedArgs[1] === "true") {
-          inputArguments.splice(i, 1, splittedArgs[0]);
-        } else if (splittedArgs[1] === "false") {
-          // if value is  false then remove that inputArguments
-          inputArguments.splice(i, 1);
-        } else {
-          inputArguments.splice(i, 1, ...splittedArgs); // push down the splittedArgs
-        }
-      }
-    }
-  });
-};
-
 // Extracts the input flags --option from the arguments  of type `--option=value` or `--option value` or `--flag`
 const extractFlags = inputArguments => {
   // Get all the args that begins with `--`. This also includes `--option=value`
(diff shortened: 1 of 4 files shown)


packages/core/lib/utils/utils.js
```diff
@@ -0,0 +1,48 @@
...
+        flag = flag.slice(0, iOfEqual);
+        inputFlags.splice(i, 1, flag);
+      }
+    }
+  });
+  return inputFlags;
+};
+
+module.exports.reformatArguments = reformatArguments;
+module.exports.extractFlags = extractFlags;
```
gnidan: How bout this syntax instead? 😄 

```javascript
module.exports = { reformatArguments, extractFlags };
```

packages/core/lib/utils/utils.js
```diff
@@ -0,0 +1,48 @@
...
+
+  // Extract only the flags i.e `--option` from `--option=value`
+  inputFlags.map((flag, i) => {
+    if (!flag) {
+      inputFlags.splice(i, 1, flag);
+    } else {
+      let iOfEqual = flag.indexOf("=");
+      if (iOfEqual > 0) {
+        flag = flag.slice(0, iOfEqual);
+        inputFlags.splice(i, 1, flag);
```
gnidan: I try to avoid using `splice` since it mutates the variable. Would you mind rewriting this to use `slice` instead?

Also I would recommend getting rid of the `else` by using this form:

```javascript
inputFlags.map( (flag, i) => {
  if (!flag) {
    return inputFlags.slice(i, 1, flag); // check my args here, I forget if slice is different than splice
  }

  let iOfEqual = flag.indexOf("=");
  /* ... continues the same, just with less indent ... */
}
```

packages/core/lib/utils/utils.js
```diff
@@ -0,0 +1,48 @@
...
+  const inputFlags = inputArguments.filter(flag => {
+    return flag.startsWith("--") ? flag : null;
+  });
+
+  // Extract only the flags i.e `--option` from `--option=value`
+  inputFlags.map((flag, i) => {
+    if (!flag) {
+      inputFlags.splice(i, 1, flag);
+    } else {
+      let iOfEqual = flag.indexOf("=");
```
gnidan: Might be better to be explicit with variable name here... maybe `equalsIndex` or just spelling out `indexOfEqualsSign`?

packages/core/lib/utils/utils.js
```diff
@@ -0,0 +1,48 @@
...
+      if (arg.replace(argTester, "").length === 0) {
+        let splittedArgs = arg.split("=");
+        // Check if value=true then don't include the value  in inputArguments
+        if (splittedArgs[1] === "true") {
+          inputArguments.splice(i, 1, splittedArgs[0]);
+        } else if (splittedArgs[1] === "false") {
+          // if value is  false then remove that inputArguments
+          inputArguments.splice(i, 1);
+        } else {
+          inputArguments.splice(i, 1, ...splittedArgs); // push down the splittedArgs
```
gnidan: (Same comment as below... I discourage use of `splice` since it mutates)

divyanshu1610: This `reformatArguments` is redundant. I'll remove it.

packages/core/lib/utils/utils.js
divyanshu1610: I made a copy of all the flags with `--option` and stored it in `inputFlags` array. I want to convert all the flags stored as `--option=value`  into `--option` only. For this i used splice.This won't mutate the original `inputArguments`. Only the copied ones will be affected.

PR #2473, opened 2019-10-18
Internal Improvement: From Web3Shim to InterfaceAdapter – Part I

CruzMolina:
This PR renames the `Web3Shim` class to `InterfaceAdapter`, implements the first generic method (`getNetworkId`), & subs every `Web3Shim` reference & var assignment to `InterfaceAdapter`/`adapter`.

👨‍🚀

PR #2480, opened 2019-10-21
Internal Improvement: From Web3Shim to InterfaceAdapter – Step Two: getNetworkId

CruzMolina:
Continuation of #2473 that implements Step 2 for `getNetworkId` as outlined [here](https://gist.github.com/gnidan/a55bbec28800f64d487054be18730691#file-step2-ts).


packages/contract/lib/contract/properties.js
```diff
@@ -364,5 +364,13 @@ module.exports = {
   },
   devdoc: function() {
     return this._json.devdoc;
+  },
+  networkType: {
```
gnidan: Why does this exist separately from `type` in the network objects?

packages/interface-adapter/lib/web3-shim.ts
```diff
@@ -74,8 +74,12 @@ export class Web3Shim extends Web3 {
...
 
-  setNetworkType(networkType: NetworkType) {
+  public setNetworkType(networkType: NetworkType) {
     this.networkType = networkType;
     initInterface(this);
   }
+
+  public getNetworkId() {
+    return this.eth.net.getId();
+  }
```
gnidan: I think it inadvisable to paste this into the shim itself... I would imagine there'd be a Web3InterfaceAdapter that implements `getNetworkId()` and calls out to the shim via Web3.js's interface

packages/interface-adapter/lib/web3-interface-adapter.ts
```diff
@@ -0,0 +1,12 @@
+import { Web3Shim, Web3ShimOptions } from "./web3-shim";
+
+export interface Web3InterfaceAdapterOptions extends Web3ShimOptions {}
+
+export class Web3InterfaceAdapter extends Web3Shim {
```
gnidan: Making it extend the shim? Is this necessary? Can you put the shim as an instance variable and do things like `this.web3.eth.net.getId()` instead?

CruzMolina: Well, looks like if we don't extend then none of the Web3/Shim methods are available to the Class. 😕

PR #2484, opened 2019-10-22
Bug fix: Correct loop syntax

eggplantzzz:
Use in rather than of in loop

### DIFF
--- a/packages/core/lib/networks.js
+++ b/packages/core/lib/networks.js
@@ -156,7 +156,7 @@ const Networks = {
       const fileContents = fs.readFileSync(filePath, "utf8");
       const body = JSON.parse(fileContents);
 
-      for (let installedNetworkId of body.networks) {
+      for (let installedNetworkId in body.networks) {
         let found = false;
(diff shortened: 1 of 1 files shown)


packages/core/lib/networks.js
```diff
@@ -156,7 +156,7 @@ const Networks = {
       const fileContents = fs.readFileSync(filePath, "utf8");
       const body = JSON.parse(fileContents);
 
-      for (let installedNetworkId of body.networks) {
+      for (let installedNetworkId in body.networks) {
```
eggplantzzz: `in` is used for iterating over object properties. 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

eggplantzzz:
Tests added @cds-amal @CruzMolina. Also accounted for both a string or number value being input in the config.

packages/core/lib/networks.js
gnidan: I won't object strongly, but I think `for (let installedNetworkId of Object.keys(body.networks)) {` is probably cleaner syntax

PR #2485, opened 2019-10-23
Organize @truffle/codec and @truffle/decoder

gnidan:
[_description still wip, but here are some notes_]

### Approach Notes

Essentially, this PR seeks to:
- Find and eliminate all submodule imports externally into the codec
- Consequently, organize all of the codec into a root namespace
  - Favor organizing codec modules into domain-major groupings (e.g. eliminate central `types/ast` in favor of `ast/types`)
    - Each domain-major grouping may define "local" types module, with expectation that it may, in some contexts, be substituted so that  `require("<domain>")` provides a superset over `require("<domain>/types")`
      - As a result, we still have `codec/types` - this is what populates the `@truffle/codec` namespace in the API docs
    - To do: encode, decode, and allocate are still mostly undealt with
      - In particular, I think codec/decode/errors should go in codec/types/
- Design organization in terms of how it will be imported. This means, e.g. use `Format.` as a prefix rather globally.

### To-dos

- [x] Consolidate docs generation for both codec and decoder into home within codec
- [x] ~Find and eliminate one remaining import externally into the codec, which is `allocate/types`~

- [x] ~Review docs for usage of TypeDoc categories to group index listings~
  - Note for usage: `@category` must be added to the symbol declaration; marking the export doesn't work
- [x] ~Figure out how to link to the root readme~
- [x] ~Put giant warnings on `nativize` docs (and `nativizeVariables`, naturally)~
- [x] ~Consistify private vs. hidden: everything that is marked either private or hidden should be protected instead.~
- [x] ~Shore up theme's defaults around viewing private / unexported entries~
- [x] ~Finish documents listed below (look for usage of phrase "this interface")~
- [x] ~Update soon-to-be incorrect wordings that say things like "when using this interface"~
        - Update wording on `reference`
        - Internal function pointer
- [x] ~If possible: Make the theme display `async` for async functions~

_Remaining work moved to #2505_

### Readmes
On a related note, in addition to the above, I see work still remaining here (or possibly as a follow-on PR) to include organizing the readmes. I started outlining these into:

  - Package repository readmes (one for codec, one for decoder)
    - Overview of package
    - "This is a development readme" Link to usage docs
    - These should have build instructions and links to the corresponding package global in the API doc
  - Docs index readme (the codec and/vs. the decoder)
    - Introduction 
    - Usage
      - Initialization
      - Full codeblock decoder usage example, including import statemen
    - Overview of each package, including
      - Summary of package
      - Highlight key parts of interface with links+description (e.g. forProject in decoder, Format in codec)
    - Note on decoding modes
  - Decoder index (module comment in decoder/lib/index.ts)
     - Summary
     - Installation instructions
     - Usage examples
  - Codec index (module comment in codec/lib/index.ts)
     - Summary
    - Installation instructions
    - Output format: Types and Results (or maybe this should be in the module index itself?)
      - Values vs. Errors

[ ℹ️ [**Preview online**](http://gnidan.org/tmp/codec/) ]


packages/decoder/README.md
```diff
@@ -0,0 +1,223 @@
...
+a `value` field containin the decoded value.  In the latter case, there will be
+an `error` field indicating what went wrong.  *Warning*: When decoding a
+complex type, such as an array, mapping, or array, getting a kind of `"value"`
+does not necessarily mean the individual elements were decoded successfully.
+Even if the `Result` for the array (mapping, struct) as a whole has kind
+`"value"`, the elements might still have kind `"error"`.
+
+3. `value`: As mentioned, this is included when `kind` is equal to `"value"`.
+It contains information about the actual decoded value.  See
+[[Format.Values|`FOrmat.Values`]] for more information.
```
haltman-at: This is my fault but miscapitalization here :P

packages/codec/lib/types.ts
```diff
@@ -222,18 +219,29 @@ export interface AbiArgument {
    */
   indexed?: boolean; //included for event parameters
   /**
-   * The decoded value of the argument.  Note that this is a [[Format.Values.Result|Values.Result]], so it
+   * The decoded value of the argument.  Note that this is a [[Format.Values.Result|Format.Values.Result]], so it
```
haltman-at: You know, if you're making the displayed text the same as the actual link, there's no need to use the bar syntax...

packages/decoder/lib/utils.ts
```diff
@@ -0,0 +1,181 @@
...
+//numbers may be BN, number, or numeric string
+//strings should be given as strings. duh.
+//bytes should be given as hex strings beginning with "0x"
+//addresses are like bytes; checksum case is not required
+//booleans may be given either as booleans, or as string "true" or "false"
+//[NOTE: in the future this function will:
+//1. check its inputs,
+//2. take a slightly different input format,
+//3. also be named differently and... it'll be different :P ]
+export function wrapElementaryViaDefinition(
```
haltman-at: Hm, so I was originally going to comment that I don't think this function should be in `decoder`, I think it should be `codec`.  Remember that this is eventually going to be expanded into a much broader set of wrapper functions to be used for encoding.

OTOH, arguably that belongs in the high-level interface, so `decoder` is the right place after all.  Especially if wrapping is going to eventually include ENS integration!

(This does bring up once again that circular dependency problem I've been mentioning, with `contract` and `decoder` both depending on each other, since currently ENS resolution lives in `contract`...)

packages/codec/lib/abify.ts
```diff
@@ -250,19 +287,24 @@ export function abifyResult(result: Values.Result, userDefinedTypes?: Types.Type
  * but you can, uh, throw an exception if you don't, I guess.)
  */
 
-export function abifyCalldataDecoding(decoding: CalldataDecoding, userDefinedTypes: Types.TypesById): CalldataDecoding {
-  if(decoding.decodingMode === "abi") {
+/** @category ABIfication */
```
haltman-at: I can't add a comment on something that hasn't changed, so let me note that the comment above is out of date; these functions are used now.

packages/decoder/lib/decoders/contract.ts
```diff
@@ -612,13 +611,16 @@ export class ContractInstanceDecoder {
    */
   public watchMappingKey(variable: number | string, ...indices: any[]): void {
```
haltman-at: Hey you forgot to make `watchMappingKey` and `unwatchMappingKey` async like we talked about :P

packages/codec/lib/compiler/utils.ts
```diff
@@ -2,7 +2,7 @@ import debugModule from "debug";
 const debug = debugModule("codec:utils:compiler");
 
 import semver from "semver";
-import { CompilerVersion, SolidityFamily } from "@truffle/codec/types/compiler";
+import { CompilerVersion, SolidityFamily } from "./types";
 
 export function solidityFamily(compiler: CompilerVersion): SolidityFamily {
```
haltman-at: Wasn't there a comment here before about the weird version string?  ...huh, I guess not.  Well, pretty sure I had it commented at some point.  Will have to add that comment back in.

packages/codec/lib/encode/abi.ts
```diff
@@ -1,138 +1,174 @@
...
 import { abiSizeInfo } from "@truffle/codec/allocate/abi";
 import sum from "lodash.sum";
 import utf8 from "utf8";
-import BN from "bn.js";
 
 //UGH -- it turns out TypeScript can't handle nested tagged unions
 //see: https://github.com/microsoft/TypeScript/issues/18758
 //so, I'm just going to have to throw in a bunch of type coercions >_>
 
 //NOTE: Tuple (as opposed to struct) is not supported yet!
```
haltman-at: My fault, but this comment is obviously outdated and should be deleted.

packages/codec/lib/read/index.ts
```diff
@@ -2,14 +2,15 @@ import * as storage from "./storage";
...
-
+import * as Pointer from "@truffle/codec/pointer";
+import { DecoderRequest } from "@truffle/codec/types";
+import * as Evm from "@truffle/codec/evm";
+
+export default function* read(
+  pointer: Pointer.DataPointer,
+  state: Evm.EvmState
+): Generator<DecoderRequest, Uint8Array, Uint8Array> {
+  switch (pointer.location) {
```
haltman-at: Again, can't comment on things that haven't changed, but something's clearly gone wrong with the comments on this file; the comment on the `eventdata` case is meaningless.  Delete it I guess?

packages/codec/lib/format/utils/exception.ts
```diff
@@ -0,0 +1,147 @@
+import debugModule from "debug";
+const debug = debugModule("codec:utils:errors");
+
+import * as Format from "@truffle/codec/format/common";
+import * as Ast from "@truffle/codec/ast";
+import * as Storage from "@truffle/codec/storage/types";
+
+function isContractDefinedType(
```
haltman-at: What is this function doing in this file??  It belongs with `isReferenceType` and etc.

packages/codec/lib/format/utils/exception.ts
```diff
@@ -0,0 +1,147 @@
...
+        return `Can't read storage from index ${
+          error.range.from.index
+        } in ${slotAddressPrintout(error.range.from.slot)} to index ${
+          error.range.to.index
+        } in ${slotAddressPrintout(error.range.to.slot)}`;
+      }
+  }
+}
+
+function slotAddressPrintout(slot: Storage.Slot): string {
```
haltman-at: Does this function really belong here either?  I mean, I can see the logic, this is the only place it gets used, so I guess it makes sense, but it still seems a little out of place.

packages/codec/lib/format/utils/inspect.ts
```diff
@@ -0,0 +1,482 @@
...
+    ...Object.entries(variables).map(([name, value]) => ({
+      [name]: nativize(value)
+    }))
+  );
+}
+
+//HACK! Avoid using! Only use this if:
+//1. you absolutely have to, or
+//2. it's just testing, not real code
+export function nativize(result: Format.Values.Result): any {
```
haltman-at: Huh, this is an interesting place to put these two functions.  I can see the logic, but do they really go here?  Maybe they should get their own file.

packages/codec/lib/format/types.ts
```diff
@@ -285,13 +313,211 @@ export interface MagicType {
...
+  | StructType
+  | StringType
+  | BytesTypeDynamic;
 
 export interface TypesById {
   [id: string]: UserDefinedType;
-};
+}
+
+function isUserDefinedType(anyType: Type): anyType is UserDefinedType {
```
haltman-at: Why are their functions in this file?  Everywhere else we keep them separate, and here we *really* should, I think.  This file should contain types only.

packages/decoder/lib/utils.ts
haltman-at: (Maybe I should ping @eggplantzzz here?)

packages/codec/lib/encode/abi.ts
haltman-at: (The one about tuple not being supported yet, I mean.)

packages/decoder/lib/utils.ts
gnidan: Mostly what I did, when deciding to move things to the decoder, is if they are used only in the decoder, I made sure they were also defined in the decoder.

packages/codec/lib/abify.ts
gnidan: Cool I will remove

packages/decoder/lib/decoders/contract.ts
gnidan: ah cool, will do!

packages/codec/lib/format/utils/exception.ts
gnidan: It's only used here and thus I decided its definition should live, unexported, in the only place where it's used.

packages/codec/lib/format/utils/exception.ts
gnidan: (Same reasoning as `isContractDefinedType`)

Anyway, for these decisions I made to un-export things, I did so mostly out of practicality, figuring we can move and/or export it later when we need it somewhere else.

packages/decoder/lib/decoders/contract.ts
haltman-at: (Note obviously this will also require modifying tests appropriately!)

packages/codec/lib/format/utils/inspect.ts
gnidan: I'm okay with them being in their own file, but I was trying to reduce the number of "buckets" users needed to be aware of... and since we don't want people to use `nativize`, I figured it shouldn't be so prominently visible.

packages/codec/lib/format/types.ts
gnidan: I am now of the mind that it doesn't make sense to have types-only modules... like if you do

```javascript
const { Format } = require("@truffle/codec");
```

in Javascript, you get nothing!

It only seems logical that helper functions live alongside the types upon with they operate.

packages/codec/lib/format/utils/exception.ts
haltman-at: Yeah I still really think this should really go with `isReferenceType` and etc.  Putting it here is going to make it impossible to find later when it's needed somewhere else.  It does the same sort of thing as them and it has no logical connection to the point of this file.

packages/codec/lib/format/utils/inspect.ts
haltman-at: I mean, yeah, makes sense.

packages/codec/lib/format/utils/exception.ts
haltman-at: Yeah I mean this is a fine place for it for now.

packages/codec/lib/format/types.ts
haltman-at: > If you do `const { Format } = require("@truffle/codec");` in Javascript, you get nothing!

Don't do that then :P

> It only seems logical that helper functions live alongside the types upon which they operate.

So, what, are you going to combine them elsewhere?  Because so far you've kept them separate everywhere else, with a `types.ts` in each folder.  And combining them seems a bit hard to do with say `Values` (where are all the functions that operate on those? They're all over the place), and meanwhile you have other `Types` functions still living elsewhere (e.g. in `format/utils`, which is where `maketype.ts` lives -- will those functions be moved into `types.ts` as well?).  I mean, if you can make it work, go ahead, but I think splitting it is the easier solution.

haltman-at:
Also on the overall structure, you have some things -- other than core and index -- that aren't even in subdirectories at all!  *And* they're typically minor functions, not major ones!  It's confusing!  I would expect anything that gets a file directly in `lib` to be super-important, but that isn't always the case. Basically the directory structure you have right now is really failing to distinguish major functions from minor ones.

gnidan:
> 1. What is this memory folder? There's, like, nothing in there.

It's just to expose `Codec.Memory.Allocate`. I'm also imagining that we'll have `Codec.Memory.Encode` and `Codec.Memory.Decode`.

> 2. Can we perhaps find a better name for evm? I always left that name alone, but I'm not sure it really describes the contents of that directory. Doesn't need changing though if we can't think of anything better.

Feel free to suggest alternatives :)

> 3. Overall organization: I feel like the overall organization is kind of a mess. Just looking at the directory structure, it's not obvious that you have "major directories" (allocate, decode) and "minor directories" (abi, contexts). I kind of liked how before we had utils, since it separated the major functions from the minor functions. Could we introduce something like that again, to make it clear to a casual browser which directories are major and which aren't?

I want to remove `allocate/`, etc., as I feel they are the wrong grouping. It makes more sense to group according to domain, not according to what kind of component

packages/codec/lib/format/types.ts
gnidan: I've been keeping `types.ts` separate, but importing them into the same namespace. The reason for this is often to avoid circular dependencies. As stated in my PR description, my thinking is that `require("<domain>")` provides a superset over `require("<domain>/types")`

gnidan: Re: MakeType... the distinction I am making is that the functions I added here take a `Format.Types.Type` as an argument, whereas MakeType _returns_ Type objects

gnidan:
> Also on the overall structure, you have some things -- other than core and index -- that aren't even in subdirectories at all!

I will make them subdirectories. I view `module.ts` exactly the same as `module/index.ts`, and avoided making extra directories in some situations purely just out of convenience.

gnidan:
Note that I am looking at this now from the perspective, not of how things exist on disk, since that is a purely internal concern, but instead looking from the perspective of "what does a user of this library have access to"?

haltman-at:
> It's just to expose `Codec.Memory.Allocate`. I'm also imagining that we'll have `Codec.Memory.Encode` and `Codec.Memory.Decode`.

There's not going to be a `Memory.Encode`.  Encoding is going to be ABI-only.

Also... have you looked at how decode works?  Like, doing it the way you describe is workable, but splitting the various `decode` files from each other... I mean, I guess it's OK, but it's icky.  (I'm assuming the central dispatcher would just go straight in `lib`?  Similarly with `read`'s central dispatcher?)

There's also the question of how doing things this way will interact with `read`.  We don't currently have a separate `read/memory.ts` and `read/abi.ts`; they're both just `read/bytes.ts`.  Not sure how that will work under your system.

> > 1. Overall organization: I feel like the overall organization is kind of a mess. Just looking at the directory structure, it's not obvious that you have "major directories" (allocate, decode) and "minor directories" (abi, contexts). I kind of liked how before we had utils, since it separated the major functions from the minor functions. Could we introduce something like that again, to make it clear to a casual browser which directories are major and which aren't?
> 
> I want to remove `allocate/`, etc., as I feel they are the wrong grouping. It makes more sense to group according to domain, not according to what kind of component

Yeah but when working on the code I want to be able to easily say, ah, here's all the files I need to change. :P  I mean, OK, I guess this way works too, it's just going to be more confusing to me...

(Is there going to be a `stack` folder?  Should I take it the `stackSize` function is going to be moved to a new file `stack/allocate.ts`?)

packages/codec/lib/format/types.ts
gnidan: I believe we've reached consensus on this (at least until you have more desire to debate me on it, I'm guessing :)

PR #2486, opened 2019-10-23
Tdb/feature/getters

cds-amal:
This PR is the first pass implementation of plural resource getters: bytecodes, compilations, contracts, contractInstances, networks and sources.

## Testing with Playground

1. Start `ganache-cli` in one terminal

1. Prepare a truffle test project in another terminal
   ```
   # assuming you're in the root `truffle folder`
   $ cd packages/truffle-db/test/truffle-projects/drizzle-box
   $ yarn # grab node dependencies like OZ ERC contracts
   $ truffle migrate
   ```
1. Start truffle-db from the truffle-db workspace
   ```
   $ yarn start:drizzle # or start:drizzle:debug for pouch-db debug logging
   ```
1. Navigate to Playground on your browser http://localhost:4444

## Playground notes

Be sure to load the workspace once per session

```
mutation loadWorkspace {
  loaders {
    artifactsLoad {
      success
    }
  }
}
```

## Sample Queries
```
query getAllSources{
  workspace {
    sources {
      id
      sourcePath 
      contents
    }
  }
}
query getAllContractInstances {
  workspace {
    contractInstances {
      id
      address
      network {
        name
        networkId
      }
      contract {
        id
        name
      }
    }
  }
}

query getAllContracts {
  workspace {
    contracts {
      name
      sourceContract {
        name
      }
      source {
        id
      }
      abi {
        json
      }
      compilation {
        compiler {
          version
        }
      }
      sourceContract {
        source {
          sourcePath
        }
      }
    }
  }
}

query getAllCompilations {
  workspace {
    compilations {
      compiler {
        name
        version
        settings
      }
      sources {
        id
      }
    }
  }
}

query getAllBytecodes{
  workspace {
    bytecodes {
      id
      bytes
      linkReferences {
        offsets
        length
      }
      sourceMap
      instructions {
        opcode
      }
    }
  }
}
```


packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -1,10 +1,21 @@
 import PouchDB from "pouchdb";
 import PouchDBMemoryAdapter from "pouchdb-adapter-memory";
 import PouchDBFind from "pouchdb-find";
-
+import pouchdbDebug from "pouchdb-debug"
 import { soliditySha3 } from "web3-utils";
+import jsonStableStringify from 'json-stable-stringify';
+
+type PouchApi = {
+  bytecodes?: PouchDB.Database,
```
fainashalts: I'm wondering why the `?` here -- I feel like each resource's db should be required, and they all get generated when we load up Pouch in this file.

packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -37,29 +48,27 @@ const resources = {
...
-  contracts: PouchDB.Database;
-  contractInstances: PouchDB.Database;
-  networks: PouchDB.Database;
+
+  private dbApi: PouchApi;
 
   private ready: Promise<void>;
 
   constructor () {
+    this.dbApi = {}
```
fainashalts: Ah, I see. Would this work as `this.dbApi;`? We just talked about this via zoom, so just leaving this here for now!

packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -69,10 +78,50 @@ export class Workspace {
...
   }
 
+  private async fetchAll(res : IWorkspaceQueryResource): Promise<IWorkspaceQueryResourceCollection> {
+    await this.ready;
+
+    try {
+      const query = { selector: {} }
+      const { docs } : any = await this.dbApi[res].find(query);
+
+      return docs.map(doc => ({...doc, id: doc['_id']}));
```
fainashalts: I might be misunderstanding this line, but there is a difference between `id`(we generate this for each resource)  and `_id`(pouchDB generates this). Not sure if this is the one you want to return.

packages/truffle-db/src/workspace/test/contractInstances.graphql.ts
```diff
@@ -0,0 +1,94 @@
+import gql from "graphql-tag";
```
fainashalts: I think this file should be renamed to the singular -- `contractInstance.graphql.ts`

fainashalts:
👍 This looks great @amal! I made a couple of small comments, and also have a concern re: using the PR description in our documentation. I think it's fine to have the new `start:drizzle` command for testing, but that's not how I'd like to present the steps to a new user, who likely has a project of their own they wish to test with. 

Instead, it would be nice to present the steps for hooking up one's own project to Truffle DB using `yarn start /path/to/individual/project` and take the steps from there. Perhaps with an additional note re: how we've already added a sample drizzle box project and here is how you can use that to test with. I hope that makes sense, happy to discuss further but it shouldn't block merging this PR!

packages/truffle-db/src/workspace/pouch.ts
cds-amal: That line initializes an object and assigns it to `this.dbAp`. If it weren't initialized, this.dbApi remains `undefined` and we couldn't treat it as an object. Which leads to another problem :( `this.dbApi` is populated dynamically in a loop and I'm not sure TS can properly infer types and will complain.

packages/truffle-db/src/workspace/pouch.ts
cds-amal: This makes sure the returned object has an `id` property by mapping pouchDB's `_id` to `id` to conform with GraphQL schema.

packages/truffle-db/src/workspace/test/contractInstances.graphql.ts
cds-amal: Oh yea, it should be consistent. Thanks @fainashalts

packages/truffle-db/src/workspace/pouch.ts
fainashalts: Ah, never mind. Our internal `id` is actually set to `_id` when each resource is added to workspace. They are actually the same value.

packages/truffle-db/src/workspace/pouch.ts
fainashalts: Yep, I'm ok leaving it as is. I sort of liked the extra check to make sure each resource is actually a part of the object but I don't think its really necessary.

packages/truffle-db/src/workspace/pouch.ts
cds-amal: I agree with you here. Thanks for pushing for this. I modified the code to make PouchApi stricter so we can have type checks in the future; and explicitly casted the initialization `as` a PouchApi to declare the object is a PouchApi.

cds-amal:
> I think it's fine to have the new `start:drizzle` command for testing, but that's not how I'd like to present the steps to a new user, who likely has a project of their own they wish to test with.

I intended for contributors so they can code, and test truffle-db against a set of fixture truffle projects to examine in the GraphiQL playground. We can add magicSquare in there as well. 
 
> Instead, it would be nice to present the steps for hooking up one's own project to Truffle DB using `yarn start /path/to/individual/project` and take the steps from there. Perhaps with an additional note re: how we've already added a sample drizzle box project and here is how you can use that to test with. 

We should do this too! 

> I hope that makes sense, happy to discuss further but it shouldn't block merging this PR!

I incorporated your previous comments, let me know what you think of the updates

packages/truffle-db/src/workspace/pouch.ts
```diff
@@ -54,12 +54,11 @@ export class Workspace {
   private ready: Promise<void>;
 
   constructor () {
-    this.dbApi = {}
     PouchDB.plugin(pouchdbDebug);
-
     PouchDB.plugin(PouchDBMemoryAdapter);
     PouchDB.plugin(PouchDBFind);
 
+    this.dbApi = {} as PouchApi;
```
fainashalts: I think this is a good use case for `as`!

fainashalts:
Thanks for these changes Amal! Looks good to me!

PR #2488, opened 2019-10-24
Dependency update: Bump web3 to version 1.2.2

eggplantzzz:


gjgd:
Hi @eggplantzzz anything I can do to help merging this? It is blocking a fix that we want to merge in https://github.com/decentralized-identity/element/

cgewecke:
@eggplantzzz If you're seeing tsc errors triggered by web3, this config might be helpful (see [thread in web3 3134](https://github.com/ethereum/web3.js/issues/3154#issuecomment-545698735)
```
{
    "compilerOptions": {
        "skipLibCheck": true
    }
}
```
The types are a 'stapled on' there and it will likely take some time to get everything straightened out.

gnidan:
Oh yikes, I don't think that `skipLibCheck` is a good idea, unless there's a way to skip only web3. I don't think we should skip typechecking all declaration files all across Truffle!

cgewecke:
@gnidan The setting suppresses compilation errors from node_modules deps. Is this a big problem? 

When I google it, the second entry I see (after the official TS page) is titled `Why is skipLibCheck not the default setting?`. It seems like for many libs the exported types don't have an organic relationship to the underlying code and are often 'off'. And there's an endless discussion going on about type correctness in deps that's not critical to thing people are directly working on.

cgewecke:
For further reference, a couple other Ethereum monorepo projects using this in their tsconfig
+ [OxProject][1]
+ [ParadigmFoundation/kosu-monorepo][2]

[1]: https://github.com/0xProject/0x-monorepo/blob/development/tsconfig.json#L10
[2]: https://github.com/ParadigmFoundation/kosu-monorepo/blob/master/packages/tsc-config/tsconfig.json#L8

eggplantzzz:
@gjgd We just need to get this to pass the CI. Currently it looks like there are some typing issues, I'm currently unsure of how to proceed. I suppose I need to do some research.

eggplantzzz:
@cgewecke I'm not sure I understand exactly what the problem but won't that option turn off a huge piece of the functionality of TS? I mean, isn't type checking one of the big reasons to use TS?

cgewecke:
@eggplantzzz They're looking for a reproduction case as we speak in [web3 3154](https://github.com/ethereum/web3.js/issues/3154#issuecomment-547060875) because the proximate cause isn't obvious. Are you just seeing on this on the HDWallet? Have you run it locally?

AFAIK the setting does not disable type-hints, but rather smooths out compilation setting discreps. There's a pretty good discussion [here](https://stackoverflow.com/questions/52311779/usage-of-the-typescript-compiler-argument-skiplibcheck) about how selective it is.

cgewecke:
@eggplantzzz Live update: [web3](https://github.com/ethereum/web3.js/issues/3154#issuecomment-547070064) is diagnosing this as problem in the host config (e.g at Truffle) and related to your use of `typeroots`.

Does tsc work without those paths specified. Is there a design decision there you could clarify? Not a typescript expert and interested in learning more :)

gnidan:
AFAIK the manual paths are there as a stopgap. Happy to remove if it makes this work!

eggplantzzz:
PR for adjusting web3 typings. 
https://github.com/ethereum/web3.js/pull/3177

eggplantzzz:
Closing this in favor of #2558

PR #2489, opened 2019-10-24
Internal improvement: Minor contract cleanup

eggplantzzz:
This PR updates a bunch of variable declarations in `@truffle/contract` and moves some code into the proper folder.


packages/contract/lib/contract/index.js
```diff
@@ -1,5 +1,144 @@
...
-  bootstrap: require("./bootstrap"),
-  properties: require("./properties")
-};
+const debug = require("debug")("contract:contract"); // eslint-disable-line no-unused-vars
+let Web3 = require("web3");
+const webUtils = require("web3-utils");
+const execute = require("../execute");
+const bootstrap = require("./bootstrap");
+const constructorMethods = require("./constructorMethods");
+const properties = require("./properties");
```
gnidan: So we're not exporting these anymore... might that be a breaking change?

CruzMolina: Piggybacking off this, since we maintain this pkg as something we expect at least _some_ users to use independently, it seems like we should make sure breaking changes are more meaningful.

eggplantzzz: Correct me if I'm wrong but I don't think those libraries were externally exposed before. I don't think they are currently accessible, take a look.

PR #2491, opened 2019-10-24
Replace name with class and address

haltman-at:
This PR replaces the `name` field of `ContractState` with a `class` field, and also adds an `address` field.  The `class` field is a `ContractType` and contains the name among other info.  This is a breaking change.  Tests are updated appropriately.

(Also, `prettier` seems to have done some work here... oops? :-/ )

This PR also makes a slight change to handling of user-inputted addresses.  Before they weren't checked; Web3 would check them when used, but we didn't do the check, which is potentially troublesome.  Now they're checked on use (as in, either they should be checksummed, or else they should be all one case to circumvent the checksum), and converted to checksum form if not in that form already.

I also fixed some comments and some error constructors, which you shouldn't really do in the same commit, but oh well...

BTW, possibly crazy idea I had: What if instead of `ContractState`, we returned a `Format.Values.ContractValue`??  The other stuff in `ContractState` being added as optional fields in `ContractValue`??  I didn't do that in this PR as I'm not sure it's the right thing to do and I wanted this done quickly, but I thought it was worth raising the idea...

### DIFF
--- a/packages/codec/lib/format/values.ts
+++ b/packages/codec/lib/format/values.ts
@@ -35,15 +35,29 @@ export * from "./elementary";
  */
 
 //This is the overall Result type.  It may encode an actual value or an error.
-export type Result = ElementaryResult
-  | ArrayResult | MappingResult | StructResult | TupleResult | MagicResult
+export type Result =
+  | ElementaryResult
+  | ArrayResult
+  | MappingResult
+  | StructResult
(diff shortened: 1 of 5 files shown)


gnidan:
Oh BTW:
> What if instead of ContractState, we returned a Format.Values.ContractValue??

I'm into it

PR #2503, opened 2019-10-28
Redo @truffle/decoder constructor functions

haltman-at:
This PR redoes the `@truffle/decoder` constructor functions.  It also adjusts the documentation and tests to match.

Because of all the added documentation, I'm skipping a long writeup of how the new functions work and why they're organized as they are; you can just see the documentation.  (Which, uh, hopefully I haven't messed up.)  Also @gnidan you will be familiar with it from our earlier discussion. :P

This PR does contain one nasty hack (which is why I've tagged you, @CruzMolina :P ); since there don't currently exist proper types for contract constructor or abstraction objects, I've just declared them myself, containing just the parts that we need.  The intent is to get rid of this later once proper types exist.  (...or maybe not, if that leads to circular dependencies; but the intent is also to eventually get rid of the contract decoder and just move all that functionality into `contract`, so whatever.)

Note I didn't add tests of all the functions, just `forContractAbstraction` -- which calls `forContractAt`, which calls `forContract`, which calls `forArtifact`; I'm hoping leaving the others without specific tests due to how analogous they are is OK.

### DIFF
--- a/packages/decoder/lib/index.ts
+++ b/packages/decoder/lib/index.ts
@@ -35,7 +35,7 @@ import { ContractConstructorObject, ContractInstanceObject } from "./types";
 /**
  * Constructs a wire decoder for the project.
  * @param provider The Web3 provider object to use.
- * @param contracts A list of contract artifacts for contracts in the project.
+ * @param artifacts A list of contract artifacts for contracts in the project.
  *
  *   Contract constructor objects may be substituted for artifacts, so if
  *   you're not sure which you're dealing with, it's OK.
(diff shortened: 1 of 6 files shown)


packages/decoder/lib/index.ts
```diff
@@ -30,76 +30,216 @@ export {
...
- * @param contract The contract constructor object corresponding to the type of the contract.
- * @param relevantContracts A list of contract constructor objects for other contracts in the project that may be relevant
- * (e.g., providing needed struct or enum definitions, or appearing as a contract type).
- * Including the contract itself here is fine; so is excluding it.
+ * @param contract The artifact for the contract.
+ *
+ *   A contract constructor object may be substituted for the artifact, so if
+ *   you're not sure which you're dealing with, it's OK.
  * @param provider The Web3 provider object to use.
+ * @param relevantContracts A list of artifacts for other contracts in the project that may be relevant
```
gnidan: Thought we were just calling all of these `artifacts` instead?

haltman-at: Oh.  We could?  That list that I showed you in Slack wasn't intended as the actual variable names, but rather a quick notation for what types the functions took.  (There's more that would need to be changed to be consistent with that, if we really wanted to stick with that.)  I'm for leaving this as-is, personally.

gnidan: Oh I much prefer `artifacts` because it's a name widely-known

gnidan: If you change it only at the surface, it's fine for now, I think

haltman-at: OK.  Guess I'll make other changes of `contract` to `artifact` for consistency while I'm at it then, I guess.  (And yeah I have no intention of changing this in the decoders themselves...)

PR #2504, opened 2019-10-28
Bug Fix: Ensure cross-chain compatibility with Provider.testConnection

CruzMolina:
Slipped by in review, but currently `Provider.testConnection` is instantiating only the `ethereum`-compatible shim/adapter.

Also passes `testConnection` the already instantiated `web3/interfaceAdapter` in `Environment.detect` & removes the [redundant](https://github.com/trufflesuite/truffle/blob/develop/packages/config/src/configDefaults.ts#L197-L206) `getProvider` call.

### DIFF
--- a/packages/provider/index.js
+++ b/packages/provider/index.js
@@ -1,6 +1,5 @@
 const debug = require("debug")("provider");
 const Web3 = require("web3");
-const { Web3Shim, InterfaceAdapter } = require("@truffle/interface-adapter");
 const wrapper = require("./wrapper");
 
 module.exports = {
@@ -32,10 +31,7 @@ module.exports = {
     return provider;
   },
 
-  testConnection: function(options) {
-    const provider = this.getProvider(options);
-    const interfaceAdapter = new InterfaceAdapter();
-    const web3 = new Web3Shim({ provider });
+  testConnection: function(web3, interfaceAdapter) {
     return new Promise((resolve, reject) => {
       console.log("Testing the provider...");
       console.log("=======================");
--- a/packages/environment/environment.js
+++ b/packages/environment/environment.js
@@ -21,7 +21,7 @@ const Environment = {
       networkType: config.networks[config.network].type
     });
 
-    await Provider.testConnection(config);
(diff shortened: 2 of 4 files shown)


packages/provider/index.js
```diff
@@ -32,10 +31,7 @@ module.exports = {
     return provider;
   },
 
-  testConnection: function(options) {
-    const provider = this.getProvider(options);
-    const interfaceAdapter = new InterfaceAdapter();
-    const web3 = new Web3Shim({ provider });
+  testConnection: function(web3, interfaceAdapter) {
```
eggplantzzz: By removing the logic in `getProvider` you will be removing functionality. For example, what happens when someone supplies the provider as a function in the Truffle config? Am I missing something or is that handled somewhere?

eggplantzzz: I guess this just means that the provider needs to be provided while instantiating the web3 passed to this function. I guess the real question is, is that addressed in the code calling this function?

CruzMolina: [`config.provider`](https://github.com/trufflesuite/truffle/pull/2504/files#diff-69a1d20ea0c7b0784b313148d5a23169R20) will have already hit [this getter](https://github.com/trufflesuite/truffle/blob/develop/packages/config/src/configDefaults.ts#L197-L206) by the time it's passed to `Environment.detect`.

CruzMolina: I'm also in favor of going back to the drawing board and reverting the original PR to remove a potential blocker for release.

CruzMolina: (in order to provide time to re-think logging, where we want to test the connection, etc)

eggplantzzz: So I don't think we really mind testing it a bunch of times but I suppose the logging is wonky. We can remove it for now, it just seems weird to have a 20 second timeout waiting for the response and not let the user know what is going on. Let's talk about it at/after dev-sync today.

CruzMolina: I really don't think it makes sense to explicitly call `getProvider` when `config.provider` has the aforementioned getter, esp if we keep the `testConnection` call in `Environment.detect`.

Issue #2505, opened 2019-10-28
Finish preparing @truffle/codec and @truffle/decoder for release


gnidan:
## Issue

Part of #2191, this issue seeks to gather known work remaining for the upcoming releases:
- @truffle/codec v0.1.0 (new package)
- @truffle/decoder v4.0.0 (major, breaking)

This issue should represent that which is blocking release, generally related to interface concerns and documentation. This issue constitutes outstanding tasks listed in #2485 and #2467, as well as from informal discussion.

---

ℹ️ **[Docs preview](http://gnidan.org/tmp/codec/)**
_(current as of bf4a50389e7989501090ebdbae3e287015e9842c)_

---

### Docs

- [x] Determine permalink URL for codec docs (#2549)
- [x] ~Expand on bulleted list of entry functions with additional figure that provides at a glance for which decode methods are available via which entry functions.~
- [x] Ensure completeness of module comments / interface comments / etc. (anything that gets its own page). (_These can be minimal in most cases_)

  Style notes: most comments are written in subject-dropped indicative mood, where the subject is the module in question (probably best to follow that convention)
- [x] Put giant warnings on nativize docs (and nativizeVariables, naturally)
- [x] Review docs for usage of TypeDoc categories to group index listings.

  Note on usage: `@category` must be added to the symbol declaration; marking the export doesn't work

- [x] Edit values-vs-errors distinction to mention overlong arrays / overlarge pointers
- [x] Consistify private vs. hidden: everything that is marked either private or hidden should be protected instead.
- [x] Fork typedoc theme and update default to include only public, exported identifiers
- [x] Finish various docs entrypoints
  - [x] Author `packages/codec/README.md` (codec repository readme)
  - [x] Author `packages/decoder/README.md` (decoder repository readme)
  - [x] Author _Truffle Decoding and Encoding_ index documentation
  - [x] Do a once-over for @truffle/decoder index documentation
  - [x] Author @truffle/codec index documentation
  - [x] Move Format documentation and author overview section

- [x] Add usage notes/example for ResultInspector

- [x] ~Determine feasibility of modifying the TypeDoc theme's display of `async` functions (to differentiate them more clearly from regular functions~
- [x] Find home for notes on decoding modes and ensure it's linked to appropriately.

### Interface changes

- [x] Define in-house interfaces for Transaction and Log
- [x] Simplify return type for decodeTransaction / decodeLog
[...]

PR #2506, opened 2019-10-29
Mark things protected, add warning to nativize

haltman-at:
This PR does two things.  First, it replaces `@hidden` and `@private` with `@protected`.  Second, it puts giant warnings on the nativize function to make clear you shouldn't actually use it in real code.

(Partial addressing of #2505 )

### DIFF
--- a/packages/codec/lib/format/utils/inspect.ts
+++ b/packages/codec/lib/format/utils/inspect.ts
@@ -380,7 +380,7 @@ export function nativizeVariables(variables: {
  * probably have a better option than using this in real code!
  *
  * (For instance, if you just want to nicely print individual values, without
- * attempting to first operate them via Javascript expressions, we have the
+ * attempting to first operate on them via Javascript expressions, we have the
  * [[ResultInspector]] class, which can be used with Node's
  * [util.inspect()](https://nodejs.org/api/util.html#util_util_inspect_object_options)
  * to do exactly that.)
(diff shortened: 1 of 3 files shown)


packages/codec/lib/format/utils/inspect.ts
```diff
@@ -319,9 +363,27 @@ export function nativizeVariables(variables: {
...
+ *
+ * If you are not doing something as horrible as our hacked-together watch
+ * expression system, then you probably have a better option than using this in
+ * real code!
+ *
+ * Remember, the decoder output format was made to be machine-readable.  It
+ * shouldn't be too hard for you to process.  If it comes to it, copy-paste
+ * this code and dehackify it for your use case, which hopefully is saner than
+ * the one that caused us to write this.
+ */
```
gnidan: This should probably end with "if you want to just look at these values, see ResultsInspector"

packages/codec/lib/format/utils/inspect.ts
```diff
@@ -319,9 +363,27 @@ export function nativizeVariables(variables: {
...
+ * ever written to support our hacked-together watch expression system, and
+ * later repurposed to make testing easier.
+ *
+ * If you are not doing something as horrible as our hacked-together watch
+ * expression system, then you probably have a better option than using this in
+ * real code!
+ *
+ * Remember, the decoder output format was made to be machine-readable.  It
+ * shouldn't be too hard for you to process.  If it comes to it, copy-paste
+ * this code and dehackify it for your use case, which hopefully is saner than
```
gnidan: Harsh!

PR #2509, opened 2019-10-29
Categorize decoder funcs

gnidan:
Part of #2505 

Screen Shot 2019-10-29 at 11 15 00 AM

### DIFF
--- a/packages/decoder/lib/index.ts
+++ b/packages/decoder/lib/index.ts
@@ -41,6 +41,7 @@ import { ContractConstructorObject, ContractInstanceObject } from "./types";
  *   you're not sure which you're dealing with, it's OK.
  *
  *   This parameter is intended to be made optional in the future.
+ * @category Provider-based Constructor
  */
 export async function forProject(
   provider: Provider,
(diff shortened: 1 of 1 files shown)


gnidan:
Happy to entertain other categorizations. Could you list the categories you had in mind, @haltman-at? Note that TypeDoc limitations means all of the categories are going to appear with the word "Functions" at the end.

gnidan:
(BTW this is why I wanted a figure to show "this is what you have", "this is what you want to decode", and which function/method call(s) you should use)

haltman-at:
Well, there's what I suggested above -- Wire Decoder Constructor Functions, Contract Decoder Constructor Functions, Contract Instance Decoder Constructor Functions, like I listed them in the readme.

gnidan:
My problem with those categories as-is is that they don't tell you anything semantically useful without an additional dereference on the corresponding term—you have to know what the various decoders are used for. I picked "Provider-based" and "Truffle Contract-based" because I think it can be assumed that most users will be familiar with those concepts. It seems less reasonable to expect users to read the introductory paragraph the one time and remember it when they're looking at this index.

haltman-at:
OK.  Well, if you think these are the right categories I'll go ahead and approve this.

gnidan:
Maybe not the best, but we'll go with it for now. Thanks!

PR #2510, opened 2019-10-29
Internal improvement: Improve network check

eggplantzzz:
Due to some awkward logging issues, it was decided to set a timeout default of 5000 ms for waiting on a response from the network while attempting to confirm that the connection is valid. It was also decided to get rid of the "Testing provider" logging as this will get run whenever `Environment.detect` is called.

Users will now be able to configure this timeout length as well by editing the `networks.<networkName>.networkCheckTimeout` property.


packages/provider/index.js
```diff
@@ -33,33 +34,37 @@ module.exports = {
...
+    let networkCheckTimeout;
+    if (networks[network]) {
+      networkCheckTimeout =
+        networks[network].networkCheckTimeout || DEFAULT_NETWORK_CHECK_TIMEOUT;
+    } else {
+      networkCheckTimeout = DEFAULT_NETWORK_CHECK_TIMEOUT;
+    }
     const provider = this.getProvider(options);
     const interfaceAdapter = new InterfaceAdapter();
     const web3 = new Web3Shim({ provider });
```
CruzMolina: Need to pass `networkType` here to `Web3Shim`.

packages/provider/index.js
```diff
@@ -33,33 +34,37 @@ module.exports = {
...
   testConnection: function(options) {
+    const { networks, network } = options;
+    let networkCheckTimeout;
+    if (networks[network]) {
+      networkCheckTimeout =
+        networks[network].networkCheckTimeout || DEFAULT_NETWORK_CHECK_TIMEOUT;
+    } else {
+      networkCheckTimeout = DEFAULT_NETWORK_CHECK_TIMEOUT;
+    }
     const provider = this.getProvider(options);
```
CruzMolina: still think calling `getProvider` and instantiating a new `InterfaceAdapter/Web3Shim` here is redundant, esp considering the rest of `Environment.detect` uses the already instantiated adapter/shim in the [helpers](https://github.com/trufflesuite/truffle/blob/develop/packages/environment/environment.js#L25-L26).

eggplantzzz: One thing that using `getProvider` gives us is more flexibility with what `options` can be. Not using it to get the provider makes it so that the `provider` property will need to be a provider and it will not accept a `host` and `port` property etc. I'm not sure what behavior we want for this method. I'm not sure I have a strong preference, any thoughts?

CruzMolina: The getter for `config.provider` will already return [Provider.create](https://github.com/trufflesuite/truffle/pull/2510/files/58c594d5693f2927fa4769758df927672dc5f749#diff-eb621b6113cc100218116be01feae659L11) which calls `getProvider` & returns a normalized Provider. This also happens right before the `testConnection` call in `Environment.detect`.

It makes sense that `testConnection` _should_ be testing an already normalized Provider.
Also makes sense to DRY up the code in general.

eggplantzzz: Yes, I understand what you are saying that it should have already parsed and set a provider on the config object. However what I am pointing out is that `testConnection` is exposed in this package. The `getProvider` method allows for the input to be a little more flexible.

I don't know what you mean by DRY here because bye using the `getProvider` method you are re-using existing code.

CruzMolina: Ah, I see. I agree as an exposed pkg util func it's useful. It still feels repetitive to implement it in the `Environment.detect` flow.

packages/provider/index.js
```diff
@@ -33,33 +34,38 @@ module.exports = {
...
-            "to the network."
-        );
         const errorMessage =
-          "Failed to connect to the network using the " +
-          "supplied provider.\n       Check to see that your provider is valid.";
+          "There was a timeout while attempting to connect to the network." +
+          "\n       Check to see that your provider is valid.\n       If you " +
+          "have a slow internet connection, try configuring a longer " +
+          "timeout in your Truffle config. Use the " +
+          "networks.<networkName>.networkCheckTimeout property to do this.";
```
gnidan: networks["<networkName>"] is probably better syntax

eggplantzzz: Yes, it is also more consistent with notes in other parts of Truffle

packages/provider/index.js
```diff
@@ -33,33 +34,38 @@ module.exports = {
...
+      networkCheckTimeout =
+        networks[network].networkCheckTimeout || DEFAULT_NETWORK_CHECK_TIMEOUT;
+      networkType = networks[network].type;
+    } else {
+      networkCheckTimeout = DEFAULT_NETWORK_CHECK_TIMEOUT;
+    }
     const provider = this.getProvider(options);
     const interfaceAdapter = new InterfaceAdapter();
-    const web3 = new Web3Shim({ provider });
+    const web3 = new Web3Shim({ provider, networkType });
```
CruzMolina: Oh, for when you resolve the merge conflict: 
```suggestion
    const interfaceAdapter = new InterfaceAdapter({ provider, networkType });
    const web3 = new Web3Shim({ provider, networkType });
```

eggplantzzz: Should I add this now?

eggplantzzz:
@CruzMolina I committed your suggestion for the InterfaceAdapter.

packages/provider/index.js
CruzMolina: Woops, guess you can't do multi-line suggestions. You can take out the extra var instantiation in the merge conflict.

PR #2513, opened 2019-10-30
Add dedicated Codec.Format module documentation

gnidan:
Part of #2505

- Provide module information
- Move output format documentation from README to Format module docs
- Revise this documentation to serve as a more standalone introduction

[Preview (commit `8cedda5`)](http://gnidan.org/tmp/codec-format/modules/_truffle_codec.format.html)


packages/codec/lib/index.ts
```diff
@@ -16,7 +16,147 @@ require("source-map-support/register");
...
+   * // when importing entire Codec, use Codec.Format.*:
+   * import * as Codec from "@truffle/codec";
+   *
+   * // or import Format directly:
+   * import { Format } from "@truffle/codec";
+   * ```
+   *
+   * ![Example struct decoding](media://example-struct-decoding.png)
+   *
+   * ## Format information
```
haltman-at: I just want to say, I like this introduction that you've written.  It's quite helpful.

packages/codec/lib/index.ts
```diff
@@ -16,7 +16,147 @@ require("source-map-support/register");
...
+   * [Ethereum JSON RPC](https://github.com/ethereum/wiki/wiki/JSON-RPC).
+   *
+   * Objects in this format may be deeply nested and/or contain circular
+   * dependencies. As such, **do not** serialize objects in this format or
+   * otherwise attempt to display them in full without considering potential
+   * risk. **Objects in this format are for the machine to read, not humans!**
+   * This module provides utilities for inspecting objects in this format,
+   * including the **safe** [[Format.Utils.Inspect.ResultInspector]] wrapper
+   * (for [util.inspect](https://nodejs.org/api/util.html#util_util_inspect_object_options)),
+   * and the **unsafe** [[Format.Utils.Inspect.nativize]] function. For more
```
haltman-at: I'm a little wary of mentioning nativize, but if you're going to put it in the docs at all, maybe upfront and marked unsafe is the best way to do so...

packages/codec/lib/index.ts
```diff
@@ -16,7 +16,147 @@ require("source-map-support/register");
...
+   *
+   * Secondly, when decoding events, it is impossible to decode indexed parameters
+   * of reference type.  Thus, these decode to an error
+   * (`IndexedReferenceTypeError`, which see) rather than to a value.
+   *
+   * Finally, except when decoding events, we do not return an error if the pointers
+   * in an ABI-encoded array or tuple are arranged in a nonstandard way, or if
+   * strings or bytestrings are incorrectly padded, because it is not worth the
+   * trouble to detect these conditions.
+   *
```
haltman-at: This is my fault, and possibly belongs in a different PR, but I only just now realized that there's a special case I forgot about when writing this -- overlong arrays / overlarge pointers.  Probably I should take care of that, maybe in a separate PR like I said...

packages/decoder/README.md
```diff
@@ -94,99 +94,6 @@ Note that decoding of state variables is only available in full mode; attempting
 to decode state variables will result in an exception if full mode is not
 possible.
 
-### Output format: `Type`s and `Result`s
```
haltman-at: Hold on, all this is just gone from the README??  I'd suggest at least including a link.  Like, leave the section in, but rename it and just make the content a link.  I think this is too important to not appear in the README, even if it's no longer going to appear directly.

packages/codec/lib/index.ts
```diff
@@ -16,7 +16,147 @@ require("source-map-support/register");
...
+   * By contrast, the following *are* errors:
+   *   - A `bool` which is neither `false` (0) nor `true` (1);
+   *   - An `enum` which is out of range;
+   *   - ..., etc.
+   *
+   * (You may be wondering about the enum case here, because if you go sufficiently
+   * far back, to Solidity 0.4.4 or earlier, it *was* possible to generate
+   * out-of-range enums without resorting to assembly or compiler bugs.  However,
+   * enums are only supported in full mode, which only supports 0.4.9 and later, so
```
haltman-at: The mention of full mode here doesn't make sense anymore now that you've moved this outside the README.  Not sure how to fix this.

packages/codec/lib/index.ts
gnidan: Thank you!

packages/codec/lib/index.ts
gnidan: I originally had it as a snarky remark saying "there's an unsafe alternative which you may find for yourself", but I figured the reader doesn't have enough context at that point to recognize the joke.

packages/decoder/README.md
gnidan: Heh ok fine I was being lazy and saving that for another PR.

Will update!

packages/codec/lib/index.ts
gnidan: Saw that too, I'll figure it out

packages/codec/lib/index.ts
gnidan: Add to #2505?

packages/codec/lib/index.ts
gnidan: Fastest solution here is to move the notes on decoding modes to a particular symbol, likely DecodingMode... if you are not extremely opposed to this, then that can save figuring out how to link to the docs index safely.

(Leaving the breadcrumb of header + content link, of course :)

Thoughts?

haltman-at: Hm, OK.  Maybe group the two outsourced sections under a "key concepts" or "important things to be aware" of section with then like two subheadings each with its link? Or something?

gnidan: Mind if I add this task to #2505 and leave for a separate PR?

haltman-at: Sounds good to me!

PR #2520, opened 2019-10-30
Update debug namespaces for codec reorg

haltman-at:
This PR just updates the `debug` namespaces as per @gnidan's codec reorganization.

(`prettier` seems to have done some work as well...)


haltman-at:
Ugh. Failing due to timeout and won't let me restart.  @gnidan...

PR #2523, opened 2019-10-31
Author root docs page for decoding/encoding

gnidan:
Part of #2505

- Move decoder readme to module comment
- Reconfigure docs to use separate file for root page
- Provide initial overview of contents
- Also fix a wrong version string


gnidan:
Thanks @haltman-at !

Issue #2457, opened 2019-10-15
Document --quiet flag


yxliang01:
- [X] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

The very useful CLI flag `--quiet` is undocumented.

## Expected Behavior

Should be documented.

CC: @eggplantzzz

eggplantzzz:
Hmm, actually it looks like this is documented for the `compile` command. What command were you using it with @yxliang01? Was it `truffle test`?

yxliang01:
Yes `truffle test` @eggplantzzz

eggplantzzz:
So we decided to add a new flag for `truffle test`, the `--runner-output-only` option. This will suppress all output except for the runner output (see PR above).

eggplantzzz:
The option mentioned above is now a feature as of v5.0.44.

yxliang01:
Thanks! @eggplantzzz

PR #2527, opened 2019-10-31
Enhancement: Add flag `truffle test --runner-output-only`

eggplantzzz:
Add flag for truffle test to suppress all output save for the output from the test runner itself.

Inspired by https://github.com/trufflesuite/truffle/issues/2457

### DIFF
--- a/packages/core/lib/commands/test.js
+++ b/packages/core/lib/commands/test.js
@@ -38,6 +38,10 @@ const command = {
       {
         option: "--show-events",
         description: "Log all contract events."
+      },
+      {
+        option: "--quiet",
+        description: "Suppress all output except for the test report."
       }
     ]
   },
@@ -132,7 +136,9 @@ const command = {
 
       promisifiedCopy(config.contracts_build_directory, temporaryDirectory)
         .then(() => {
-          config.logger.log("Using network '" + config.network + "'." + OS.EOL);
+          if (config.quiet !== true) {
+            config.logger.log(`Using network '${config.network}'.${OS.EOL}`);
+          }
 
           run();
         })
--- a/packages/core/lib/commands/test.js
+++ b/packages/core/lib/commands/test.js
@@ -6,6 +6,11 @@ const command = {
       describe: "Show all test logs",
       type: "boolean",
       default: false
+    },
+    "mocha-output-only": {
+      describe: "Suppress all logging except for mocha output.",
+      type: "boolean",
+      default: false
     }
   },
   help: {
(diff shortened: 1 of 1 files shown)


packages/core/lib/commands/test.js
```diff
@@ -38,6 +38,10 @@ const command = {
       {
         option: "--show-events",
         description: "Log all contract events."
+      },
+      {
+        option: "--quiet",
+        description: "Suppress all output except for the test report."
```
CruzMolina: Shouldn't this be...

```suggestion
        description: "Suppress all compilation output."
```

instead?

packages/core/lib/commands/test.js
```diff
@@ -132,7 +136,9 @@ const command = {
 
       promisifiedCopy(config.contracts_build_directory, temporaryDirectory)
         .then(() => {
-          config.logger.log("Using network '" + config.network + "'." + OS.EOL);
+          if (config.quiet !== true) {
```
CruzMolina: This looks like a breaking change.

packages/core/lib/commands/test.js
eggplantzzz: Well it depends on what we want. On this first run at this task, I also removed the logging about what network to use as you see. It seems to make sense for this flag (and this was sort of requested) to suppress all output and not JUST compilation output.

Initially the person asking about this feature was using a JSON reporter and wanted to parse all the output. Just suppressing compilation output seems to me to be strange behavior. With that flow, you get "Using network <networkName>" and then the test report. That seems very weird to me, quiet seems like it should mean "As little logging as possible for this task".

packages/core/lib/commands/test.js
eggplantzzz: Yes indeed. However, one could also make the argument that this is not yet an official Truffle feature since there it is totally undocumented.

CruzMolina: Given that at least one user knows about it, it makes sense that we'd strictly document what the current behavior is, before deciding whether we want to change it later on.

packages/core/lib/commands/test.js
CruzMolina: The current behavior appears to be the same as the documented `truffle compile --quiet` flag. And I believe under the hood, `truffle test --quiet` just tells the compiler pkgs to suppress compilation output.

The issue you linked is requesting documentation. Makes sense we should simply document current behavior.

packages/core/lib/commands/test.js
eggplantzzz: Well actually one user knows about it because I told him about it. I would prefer to determine what behavior we want before we document it as it will save us a lot of time. Once it is "official" or documented, it will be harder to change.

PR #2529, opened 2019-11-01
Simplify wire decoder return types; internalize Web3 types

haltman-at:
(Part of #2505)

This PR simplifies the return types of `decodeTransaction` and `decodeLog` by having them simply return the decoding (or array of decodings) without the additional complication of `DecodedTransaction` or `DecodedLog`.  The `DecodedLog` type is still present so it can be returned by `events`, though.  Also, the `decodeLogs` function has been removed entirely.  Tests have been updated appropriately.

In addition, the `Transaction`, `Log`, and `BlockType` types from Web3 have been replaced by equivalent types we define.  Note that I renamed `BlockType` to `BlockSpecifier` for clarity.

In writing this I noticed that there may be some slight problems in how block specifiers are currently handled.  I'll fix those in a separate PR.  (Or possibly two PRs, since some of this I'd like to fix on `develop` as well.)

### DIFF
--- a/packages/decoder/lib/decoders.ts
+++ b/packages/decoder/lib/decoders.ts
@@ -22,8 +22,6 @@ import * as DecoderTypes from "./types";
 import Web3 from "web3";
 import { ContractObject } from "@truffle/contract-schema/spec";
 import BN from "bn.js";
-import { BlockType, Transaction } from "web3/eth/types";
-import { Log } from "web3/types";
 import { Provider } from "web3/providers";
 import {
   ContractBeingDecodedHasNoNodeError,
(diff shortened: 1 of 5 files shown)


packages/decoder/lib/decoders.ts
```diff
@@ -271,36 +266,47 @@ export class WireDecoder {
...
    *
-   * Takes a Web3
-   * [Log](https://web3js.readthedocs.io/en/v1.2.1/web3-eth.html#eth-getpastlogs-return)
-   * object and returns a copy of that object but with an additional decodings
-   * field.  This field holds an array of [[LogDecoding|LogDecodings]]; see the
-   * documentation on [[DecodedLog]] for more.
+   * Takes a [[Log]] object and decodes it.  Logs can be ambiguous, so this so
+   * this function returns an array of [[LogDecoding|LogDecodings]].
+   *
+   * Note that logs are decoded in strict mode, so (with one exception) none of the decodings should
```
gnidan: Is "strict mode" defined somewhere?

haltman-at: Ooh, good catch.  No, I don't think it is.  Of course this isn't a new problem with the docs (this is just moved from elsewhere), but it hadn't previous occurred to me that it *is* a problem.  Thanks.  Um, I'm thinking maybe just merge now, fix later?

PR #2533, opened 2019-11-01
Split Codec allocate/types into respective allocate/{abi|storage|memory} submodules

gnidan:
Part of #2505 

- Convert allocate/{abi|storage|memory} from files to directories
- Move abi/storage/memory-related types from allocate/types into those directories
- Remove wonky path config that remained in decoder


gnidan:
Thanks!

PR #2535, opened 2019-11-01
Internal improvement: Update tests to use the InterfaceAdapter

CruzMolina:
This PR swaps out using `Web3` in scenario & integration tests to properly test the `InterfaceAdapter` instead.

### DIFF
--- a/packages/artifactor/test/contracts.js
+++ b/packages/artifactor/test/contracts.js
@@ -8,7 +8,7 @@ const fs = require("fs");
 const requireNoCache = require("require-nocache")(module);
 const Compile = require("@truffle/compile-solidity/legacy");
 const Ganache = require("ganache-core");
-const Web3 = require("web3");
+const { InterfaceAdapter } = require("@truffle/interface-adapter");
 const { promisify } = require("util");
 
 describe("artifactor + require", () => {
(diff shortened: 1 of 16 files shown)


packages/artifactor/test/contracts.js
```diff
@@ -96,7 +95,7 @@ describe("artifactor + require", () => {
   });
 
   before(() =>
-    web3.eth.getAccounts().then(_accounts => {
+    interfaceAdapter.eth.getAccounts().then(_accounts => {
```
gnidan: Ah, this sort of thing is the culprit! If you want to do this, you should make a `getAccounts` method for real on the interface adapter

packages/artifactor/test/contracts.js
```diff
@@ -230,20 +229,20 @@ describe("artifactor + require", () => {
           "Fallback should not have been triggered yet"
         );
         return example.sendTransaction({
-          value: web3.utils.toWei("1", "ether")
+          value: interfaceAdapter.utils.toWei("1", "ether")
```
gnidan: I think these need to go somewhere else. Use web3 for this for now. The point of the interface adapter is that it removes usage of the Web3.js interface. In places where it doesn't serve to sufficiently replace the Web3.js interface adapter, we should use either the shim or Web3.js itself

packages/artifactor/test/contracts.js
CruzMolina: Ahh, so we should have started with `setProvider` then 🤦‍♂ .

CruzMolina:
Closing for now.

PR #2536, opened 2019-11-01
Organize Codec process directories

gnidan:
Part of #2505 

- Remove lib/allocate and move contents to data location directories
- Make MappingKey root namespace in `lib/mapping-key` and move `lib/encode/key` to `lib/mapping-key/encode`
- Remove lib/decode/* and move contents to data location directories
- Add `@category` tags for data locations
- Group data locations by ABI/Solidity/common
- Rename Abi to AbiData
- Split off topic into Topic


haltman-at:
Oh wait, "direct" isn't quite right either, because it also handles `string` and `bytes`.  Well, sort of.  It's a bit messy.  Blech.

haltman-at:
OK, @gnidan has asked me to suggest better names.  Hm.

I think `decodeConstant` can remain where it is for now.  I doubt improved constant decoding is actually coming anytime soon.  (Oh, hm, that reminds me of something that needs fixing in the docs...)

As for `elementary`... maybe `basic`?  I don't know, that's the best I can come up with at the moment:

* "elementary" has the problem above
* "value", what it was already called, also conflicts with how we're using that word elsewhere (and also doesn't quite match the Solidity use of the term)
* "direct" is opaque and not quite accurate
* "atomic" would make people think of, like, concurrency atomicity, which is unrelated
* "indecomposable" is just silly :P (and "unsplit" is sillier :P )

So, uh, I'll suggest "basic", I guess? <shrug>

haltman-at:
Oh, one other note, I notice the debug namespaces haven't been updated.  But I guess we can ignore that for now if you don't want to bother with it.  (I'd approve but no review has been re-requested yet... should I go ahead and do so anyway?)

gnidan:
@haltman-at I want to rename the directory for keys as well, to match the namespace name. I'll get that for you shortly and rerequest review

Also I guess I'll fix the debug modules :)

haltman-at:
Heh, keys actually *could* be called "elementary", but that would probably be more confusing than helpful...

Issue #1707, opened 2019-02-08
Add support for mocha-parallel-tests


PhABC:
It would be great if truffle could add a flag (or make it as a default) such that tests can be executed in parallel, perhaps via https://www.npmjs.com/package/mocha-parallel-tests 

There might be  need to run multiple ganache / nodes instances in parallel as well to make sure there are no transaction nonce conflicts, but maybe a ganache fork could also work. 

Running tests in parallel could drastically reduce testing times, which in many cases can take a while.

gnidan:
This would be great! But it's tricky and would constitute a breaking change, were it to be default. Have you tried configuring Mocha via Truffle's config (see docs [here](https://www.truffleframework.com/docs/truffle/reference/configuration#mocha))? It might work that way, without a change to Truffle itself.

kkx:
@gnidan 
The config used in truffle-config is for configuring mocha itself, there is no way to change using other enchanced mocha library instead of the standard mocha.
would be great if user can select the mocha library to use and import it dynamically when creating Mocha instance here in this line:
https://github.com/trufflesuite/truffle/blob/b4b26add0239bb89da0aa562e730023d5eb35341/packages/core/lib/test.js#L124

like this:
```
// mocha-parallel-tests example
// if you're using TypeScript you don't need to install @types/mocha-parallel-tests
// because package comes with typings in it
import Mocha from 'mocha-parallel-tests'; // or `const Mocha = require('mocha-parallel-tests').default` if you're using CommonJS
const mocha = new Mocha();
```

gnidan:
Sure @kkx I think that can work by adding a field to the `mocha` part of the Truffle config: https://www.trufflesuite.com/docs/truffle/reference/configuration#mocha

Feel free to open a PR for this and we can make sure it won't break anything. Seems straightforward though... good idea!

k06a:
@kkx any chance you will contribute this?

k06a:
This is my try: https://github.com/trufflesuite/truffle/pull/2540

eggplantzzz:
Merged in Truffle [v5.1.11](https://github.com/trufflesuite/truffle/releases/tag/v5.1.11)

mrice32:
[...]

PR #2540, opened 2019-11-03
Respect "package" option of mocha config to allow "mocha-parallel-tests" usage instead of "mocha"

k06a:
Related to https://github.com/trufflesuite/truffle/issues/1707


eggplantzzz:
Can you provide me with an example of this new use case?

k06a:
@eggplantzzz:

image

k06a:
@eggplantzzz this is for https://github.com/trufflesuite/truffle/issues/1707

eggplantzzz:
Will this work? Seems like you would have to require the actual package in your Truffle config (not just include the name of a package as that is not included in Truffle's dependencies).

Also, there might need to be some setup involving managing different instances of Ganache to run each set of tests against. I'll think about this.

k06a:
@eggplantzzz it should be included in project dependencies. It seems both libs work with same arguments and should work transparent.

eggplantzzz:
So I discussed this with @gnidan and decided that if we are going to include this, we don't want "mocha-parallel-tests" to be a Truffle dependency. I think that it should be something a user requires in `truffle-config.js` to set in the `package` property. Can you update this PR to allow for this? 

Also, can you use `let` instead of `var`? :)

k06a:
@eggplantzzz did you mean this?
```js
package: require('mocha-parallel-tests'),
```

eggplantzzz:
Yes, that way we don't have to bundle it into Truffle. Users who want to utilize that package can install it locally and pass it to Truffle in `truffle-config.js`.

eggplantzzz:
@k06a Sorry for letting this sit around. I will try and get attention for this this week!

eggplantzzz:
I think this should be ready! I'll get one more set of eyes on this before merging.

PhABC:
I've been trying to use this but I get the following error ; 

`TypeError: Mocha is not a constructor at Object.createMocha`

I do have

```json
mocha: {
  package: require('mocha-parallel-tests'),
  ..
}
```
in my truffle.json file and `mocha-parallel-tests` is installed.

julienmuller-forge:
@PhABC : i had the same issue, which i fixed with package: require('mocha-parallel-tests').default

However, it still doesn't work for me : whatever i do i always get 'ReferenceError: artifacts is not defined' on the first line of the first test.

leovigna:
I have found a way to run parallel mocha tests with the truffle contracts though it's not a drop-in replacement. Here is my answer on a similar issue: https://github.com/trufflesuite/truffle/issues/1707#issuecomment-748313997

PR #2543, opened 2019-11-04
Makes `config` available in the migration context.

MicahZoltu:
This will allow a migration to be able to interact with any of the Truffle packages using the same configuration that the migration is using to run.  One example use case (mine) is to compile dynamic contracts based on the result of migration (e.g., hard-code an address in a contract as a gas optimization rather than passing it as a constructor).

Along the way I also cleaned up options vs config a little bit, once `config` is resolved it should be used rather than `options` since it may have different property values once full config resolution is complete.

### DIFF
--- a/packages/migrate/index.js
+++ b/packages/migrate/index.js
@@ -31,20 +31,18 @@ const Migrate = {
     const config = Config.detect(options);
 
     if (
-      fs.existsSync(options.migrations_directory) &&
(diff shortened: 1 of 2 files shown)


gnidan:
(This stems out of a Gitter discussion that roughly concludes [here](https://gitter.im/ConsenSys/truffle?at=5dbf789ae1c5e915082b33ab))

PR #2544, opened 2019-11-04
Adds dependency that was missing.

MicahZoltu:
`resolver/index.js` references `@truffle/contract` yet the project doesn't have a dependency on it.  This means if someone tries to use this package standalone, or as a transitive dependency of some other package, they'll get a runtime failure.


eggplantzzz:
Thanks for taking care of this!

PR #2550, opened 2019-11-04
Internal improvement: From Web3Shim to InterfaceAdapter – Step Two: getTransaction

CruzMolina:
Continuation of #2473 that implements Step 2 for `getTransaction` as outlined [here](https://gist.github.com/gnidan/a55bbec28800f64d487054be18730691#file-step2-ts).


packages/interface-adapter/lib/interface-adapter.ts
```diff
@@ -46,7 +49,11 @@ export class InterfaceAdapter {
     return this.adapter.getBlock(block);
   }
 
-  public setProvider(provider: Provider) {
+  public setProvider(provider: Provider): void {
```
gnidan: Can we make this async? I know that was last PR (sorry!)

packages/interface-adapter/lib/interface-adapter.ts
```diff
@@ -46,7 +48,11 @@ export class InterfaceAdapter {
     return this.adapter.getBlock(block);
   }
 
-  public setProvider(provider: Provider) {
+  public setProvider(provider: Provider): Promise<void> {
```
gnidan: is `adapter.setProvider` not being used? we'll need to add `await` to those (this is the part that I figured you might object to :)

CruzMolina: we only use it in `@truffle/contract`'s `setProvider` method. I believe at least one of the CI jobs would be failing if it wasn't properly resolving.

CruzMolina: anddd it's gone.

PR #2553, opened 2019-11-05
Various further documentation improvements

haltman-at:
This PR makes several improvements to the documentation:

1. It marks as `protected` various decode modules with no public functions.  In some cases this led to the containing module having nothing exposed, in which case I marked that `protected` too.  Note: I didn't mark `Stack` as `protected`, which leaves its page empty.  Why did I leave it that way then?  Well, because I assume `readStack` is going to get moved in here soon enough.  But I could change it of course.

2. I put in more category orderings; in particular, I made sure "Other" is always sorted to the end.

3. I changed "Special container types" to "Special container types (debugger-only)" for additional clarity.

4. I added documentation for `ResultInspector`.

5. I expanded the `Format` documentation regarding errors to mention unsupported constants and other read errors.


gnidan:
Agree about `readStack`. Mind adding "Split up read module" to #2505?

haltman-at:
> Agree about `readStack`. Mind adding "Split up read module" to #2505?

Done; included `bytes` in that item as well.

PR #2558, opened 2019-11-05
Dependency update: Bump web3 to version 1.2.2 & implement typings

CruzMolina:
Continuation of #2488. Also adjusts & migrates typings from `@types/web3` to various `web3`@1.2.2 pkg typings.


packages/decoder/tsconfig.json
```diff
@@ -22,7 +18,10 @@
       "../../**/node_modules/@types/lodash",
       "../../**/node_modules/@types/lodash.clonedeep",
       "../../**/node_modules/@types/lodash.isequal",
-      "../../**/node_modules/@types/lodash.merge"
+      "../../**/node_modules/@types/lodash.merge",
+      "../../**/node_modules/@types/lodash.merge",
```
eggplantzzz: Duplicate of lodash.merge.

packages/interface-adapter/package.json
```diff
@@ -19,16 +19,17 @@
     "watch": "tsc -w"
   },
   "dependencies": {
+    "@truffle/blockchain-utils": "^0.0.13",
     "bn.js": "^4.11.8",
     "ethers": "^4.0.32",
     "lodash": "^4.17.13",
-    "web3": "1.2.1"
+    "web3": "1.2.2",
+    "web3-eth": "1.2.2"
```
cgewecke: @CruzMolina I *think* when web3 is installed, all the subsidiary web3- packages are auto-populated to `node_modules`. But in some truffle packages (truffle/contract is an example), some sub-packages are also individually listed as deps. 

For purely selfish reasons (an E2E test at web3 which uses Truffle as a target and needs to swap web3 in/out), it would be convenient if this pattern was normalized a bit. 

Is this possible or are you finding it necessary to have the separate dep?

cgewecke: IIRC I started adding sub-packages this way at truffle/contract because an early 1.x beta suddenly changed the way everything was published, and then reversed course. But I lost track of it...

CruzMolina: Let me look into. I _might_ need to specify the separate dep for typings, but I can probably adjust the `tsconfig`'s to find the sub-pkgs.

cgewecke: @CruzMolina Ok cool - yeah if it's not convenient for you please just ignore.

CruzMolina: Pushed the changes, seems to work locally 🤞

packages/blockchain-utils/typings/index.d.ts
```diff
@@ -1,3 +1,21 @@
+import { HttpProvider } from "web3-providers-http";
+import { IpcProvider } from "web3-providers-ipc";
+import { WebsocketProvider } from "web3-providers-ws";
+
+export type Provider = HttpProvider | IpcProvider | WebsocketProvider;
```
gnidan: Wanna move this to `@truffle/provider`?

packages/blockchain-utils/index.ts
```diff
@@ -1,8 +1,12 @@
-import { Provider, Callback, JsonRPCResponse } from "web3/providers"
-import { parsedUriObject } from "typings"
+import { parsedUriObject } from "typings";
```
eggplantzzz: Does this need to be here? Looks like it isn't used.

CruzMolina: https://github.com/trufflesuite/truffle/blob/edd533722a361872ba7a24f46b14878c1009358a/packages/blockchain-utils/index.ts#L40

eggplantzzz: Oh my mistake, I only searched the diff.

packages/decoder/tsconfig.json
```diff
@@ -10,9 +10,10 @@
     "baseUrl": ".",
     "lib": ["es2017"],
     "paths": {
-      "web3": ["../../node_modules/@types/web3/index", "node_modules/web3/index"],
-      "web3/types": ["../../node_modules/@types/web3/types", "node_modules/web3/types"],
-      "web3/eth/types": ["../../node_modules/@types/web3/eth/types", "node_modules/web3/eth/types"]
+      "web3-core": [
+        "../../node_modules/web3/node_modules/web3-core/types",
```
eggplantzzz: Why do both of these paths need to be specified and not just one?

eggplantzzz: Just a fallback?

packages/hdwallet-provider/package.json
```diff
@@ -37,7 +37,8 @@
     "ganache-core": "^2.8.0",
     "mocha": "6.2.0",
     "ts-node": "^8.4.1",
-    "typescript": "^3.6.3"
+    "typescript": "^3.6.3",
+    "web3": "1.2.2"
```
eggplantzzz: Does this really need to go to `devDependencies`?!

packages/decoder/tsconfig.json
CruzMolina: Yup. The first path is in monorepo context, second is when working w/ the `decoder` pkg independently.

packages/hdwallet-provider/package.json
CruzMolina: Yup, I think the original import was incorrect and pointing to `@types/web3` instead of `web3`.

CruzMolina: oh, nm. missed the other import!

eggplantzzz:
Ok I think this is cool with me when it passes CI.

PR #2561, opened 2019-11-06
Remove circular dependencies from Codec

gnidan:
Branch `scatter-the-pages` currently introduces circular dependencies, mostly around trouble mostly of these kinds:

- Ast was previously not a data location and thus was safer for more things to import wholesale. Ast's new Decode module introduces circular dependencies. Consequently, this PR modifies problematic Ast imports to import specific submodules

- `types` modules seem to mostly need rely only on other types modules directly

- Read modules seem to need to import other Read modules directly

- Encode modules seem to need to import other Encode modules directly

- Decode modules seem to be able to import entire other modules wholesale

Not sure if this is worth codifying, or if it is sound inductive reasoning at all.

Ah, also: revert #2554


gnidan:
The biggest cause was that data locations are more restricted as to what can safely import them wholesale, and Ast previously was not a data location

PR #2562, opened 2019-11-06
Finish reorganization of read, encode, & decode

haltman-at:
This PR (part of #2505) does several things.

First, it splits up `read` the same way that `decode` has been split up.  The `lib` directory now contains a file `read.ts` that plays the role of the old `read/index.ts`, similar to how `decode.ts` replaced `decode/index.ts`.

You may be wondering, where did `readBytes` end up going?  Well, there's now a `bytes` folder!  More on that in a bit.

It also makes consistent how the various read functions work, in that now they all take a pointer and the state, instead of the mishmash of arguments they took before.  (Note that `readConstant` and `readStackLiteral` only take a pointer, as the state is irrelevant; and that `readStorage` is of course a generator function rather than a pure function like the others.)

(Note that the debugger, when it calls `readStack`, just passes `{ stack }` as the state; it doesn't attempt to fill in the rest.  That wouldn't work very well.  Technically this is a violation of the type signature, in that storage is supposed to be required, so if we wanted we could also pass in empty storage, but it's irrelevant, so I didn't bother; I just put a HACK comment.)

Doing this has another benefit -- since `readBytes` now takes a pointer, it now knows which location is being read, which allows for improved error messaging.  The `ReadErrorBytes` error in `Format.Errors` now includes the location, and its error message is adjusted to account for this.

Note that every single case of `read` now involves a function call, even if that function does very little work; there are now functions `readSpecial`, `readTopic`, and `readStackLiteral`.

You may also be wondering where `readConstant` ended up.  It's in `ast`!  The `decodeConstant` function has been moved there too.

So let's move on to the next point, which is the reorganization of `decode`.  Decoding of the types `string` and `bytes` have been moved out of `decodeBasic` and moved to a new function `decodeBytes` which also lives in the `bytes` folder.  This does mean that some things that used to just defer to `decodeBasic` now have to do an additional check for whether they should defer to `decodeBasic` or `decodeBytes`, but the result is a lot more coherent.

Note: I didn't bother adding `DirectType` or `DirectValue` or anything like that to the type system, but quite possibly I should?  It would definitely improve the typing here and remove a coercion that's in there currently.  In fact now I'm really thinking I should've done that...

Anyway, yeah, `decodeBytes` has been split off and `decodeConstant` has been moved.  That's it for `decode`.

OK, finally, let's talk about `encode`.  The encoding functions have now been factored, so there's no longer these big blocks of copypaste, and split up in a way that better matches the new structure.

The `basic` folder now contains `encodeBasic`, which does what it sounds like, and `bytes` contains `encodeBytes`, which encodes the types `string` and `bytes` to raw bytestrings (which other encoding functions can then build on).  Various helper functions have been moved from `abi-data/encode.ts` to `bytes/encode.ts` as a result.

Note that `encodeBasic` has some weird typing as a result of me not adding that stuff to the type system I mentioned.  Yeah, I should really go back and change that...

Anyway, `encodeMappingKey` is now much simplified as it just dispatches to `encodeBasic` or `encodeBytes` as appropriate.  Meanwhile `encodeAbi` is also simplified as a result, although obviously not as simple (it has to handle other more complicated types, and also for bytestrings it has to pad them and prepend the length).

Finally, `encodeTopic` has been split off as a separate function, although it doesn't have all its functionality yet.  But that hack in `encodeAbi` where it would refuse to encode errors *except* `IndexedReferenceTypeError`s has now been moved to `encodeTopic`, which basically means it's not a hack anymore, hooray!

My intent is that `encodeTopic` will eventually be able to encode and hash reference types passed directly into it as well, but I haven't written that yet (not that it'll be hard, especially now that things have been factored).

Note: Only `encodeAbi`, `encodeTupleAbi`, and `encodeTopic` allow for `Result`s; everything else requires `Value`s.

Also note: I opted to have `encodeBasic` refuse to encode internal functions, on the basis that these never go in any location for which we're implementing encoding.  If I fix the typing though, so that returning `undefined` no longer makes sense, then perhaps I should write encoding logic for them after all (even though the result would be meaningless).


haltman-at:
Ugh, scratch what I said about the typing -- fixing that would cause other problems.  Leaving that alone for now...

haltman-at:
OK, going to probably make some minor nonfunctional changes before merging.

PR #2564, opened 2019-11-06
Remove Codec.Format.Utils.MakeType; divide into data locations

gnidan:
- Move `definitionToType` and `definitionToStoredType` to Ast.Import
- Move `abiParameterToType` to AbiData.Import
- Move `contextToType` to Contexts.Import (**side note**: should we singularize this?)


gnidan:
Question: I hate to add more work to this PR, but I just had the thought that maybe "Import" is a better name than "Convert"? Thoughts?

haltman-at:
> Question: I hate to add more work to this PR, but I just had the thought that maybe "Import" is a better name than "Convert"? Thoughts?

Interesting idea.  Makes sense I guess?

gnidan:
I do thinking splitting `Constant` out is the right move long-term, but happy to skip for now. Now that you seem to be into the idea, I'm leaning towards the sign of doing it.

I will plan to do the `Import` switch as well, and also rebase, since I think the revert is a bad first commit

haltman-at:
> I do think splitting `Constant` out is the right move long-term, but happy to skip for now... but now that you seem to be into the idea, I'm leaning towards the side of doing it.

I mean, the point of doing it now would be that we could revert more of those changes to imports that you made so that all could be nicer.  Right, remember I commented above that this PR that was supposed to revert lots of those ended up not reverting very many.

gnidan:
Ah, right! Well, regardless of your points @haltman-at, I think we should not do the Constant change in this PR, since it's out of scope. I also think it's fine to merge this _then_ do the Constant stuff, rather than wait and rebase again. LMK if you concur!

haltman-at:
Hm, I basically thought of the point of this PR as doing things that allow us to revert those import changes.  So I don't see that it's out of scope.  Having it separate just seems more confusing to me; ideally those import changes would have never gotten merged into `next` in the first place, I don't really like having them sitting around.  But do what you want, I guess?

gnidan:
Ideally, if we wanted history to be pristine, we'd have not merged your PR that moved everything to Ast, so I'm not worried about a few lingering imports that never-woulda-been.

Merging this since your objections don't seem strong enough to counteract my desire to hit the green button. Thanks! I'll make sure the Constant thing is on #2505

Issue #2539, opened 2019-11-03
Gas multiplier is never applied in truffle-contract


cgewecke:
## Issue

By default, contracts using autoGas should multiply the gas estimate by 1.25.  At the moment, the multiplying logic seems like it just returns the original estimate value.

Web3 1.2.2 [fixes][1] long-broken logic that registers a transaction as failed when the gas consumed equals the gas sent. In older versions of the eth chain and some contemporary forks ( ETC? maybe Quorum?) failing txs consume all gas and the gas comparison is a reliable indicator that something has gone wrong. 

This is relevant for #2488 because truffle-contract's tests are fueling the tx's with perfect estimates and triggering a false error.  

The question of what behavior is correct is a little complicated and will need further discussion at Web3, but Truffle could side-step this by fixing the multiplier.

Possible that #2512 is related...

cf: [web3 3175](https://github.com/ethereum/web3.js/issues/3175)

## Steps to Reproduce

```javascript
utils.multiplyBigNumberByDecimal(
  utils.bigNumberify(500),
  1.25
);
> 500
```
[1]: https://github.com/ethereum/web3.js/pull/3123/files
## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): 5.0.41
* node version (`node --version`):
* npm version (`npm --version`):

eggplantzzz:
Ahh yes I see what you mean. It looks like the logic in that method isn't sound. Thanks for pointing this out!

eggplantzzz:
A fix for this is now out in the wild with v5.0.44.

PR #2565, opened 2019-11-06
Bug fix: Properly apply gasMultiplier to gasEstimates

CruzMolina:
Fixes a bad math bug (#2539)!

### DIFF
--- a/packages/contract/lib/utils.js
+++ b/packages/contract/lib/utils.js
@@ -1,6 +1,6 @@
 const debug = require("debug")("contract:utils"); // eslint-disable-line no-unused-vars
 const web3Utils = require("web3-utils");
-const bigNumberify = require("ethers/utils/bignumber").bigNumberify;
+const { bigNumberify } = require("ethers/utils");
 const abi = require("web3-eth-abi");
 const BlockchainUtils = require("@truffle/blockchain-utils");
 const reformat = require("./reformat");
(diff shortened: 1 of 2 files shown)


packages/contract/lib/utils.js
```diff
@@ -255,10 +255,10 @@ const Utils = {
     );
 
     const denominator = bigNumberify(10).pow(significantFigures);
-    const numerator = Math.round(decimal * denominator);
-    const secondOperand = bigNumberify(numerator).div(denominator);
+    const multiplier = Math.round(decimal * denominator);
+    const numerator = bigNumberify(multiplier).mul(bignum);
 
-    return bignum.mul(secondOperand);
+    return numerator.div(denominator);
```
gnidan: So, my algebra is rusty, but I had to reason this out in vim for myself to make sense of it, so thought I'd share:

```
This PR changes this:            To this:

        round( d * 10^P )            n * round( d * 10^P )
v = n * -----------------        v = ---------------------
            10^P                              10^P
```


We're trying to do this:

```
v = n * d
```

Or, expressed as digit placeholders (with underscores as precision gaps):

```
VVVV.0 = NNNN.0 * D.DDD____
```

where:
  - v is the result we want
  - n is `bignum`,
  - d is `decimal`
  - P is `significantFigures`, i.e. some amount of precision

**To [try to?] show how this might play out, the new code does steps:**

1. Scale for desired precision (multiply+round):

   ```
   D.DDD____  ->  DDDD.0
   ```

2. Multiply by our `bingum`:

   ```
   NNNN.0 * DDDD.0 = VVVV____.0
   ```

3. Then divide by original scale:

   ```
   VVVV____.0  ->  VVVV.____
   ```

No loss of precision.


To confirm why the old one was bad, the difference is that we scaled then immediately de-scaled:

1. Scale:
   ```
   D.DDD____  ->  DDDD.0
   ```

2. De-scale:

   ```
   DDDD.0  ->  D.0
   ```

3. Then multiply:
   ```
   NNNN.0 * D.0 = VV__.0
   ```

OK, I'm convinced this is correct and that the old one was not correct.

gnidan: (Should this reasoning be a code comment? Would that spare future incredulity?)

CruzMolina: I'l leave this convo open for a future Truffler. ⏲

PR #2568, opened 2019-11-07
Rename ContractObject to Artifact for clarity

haltman-at:
(Part of #2505)

This PR renames `ContractObject` to `Artifact` on import for clarity.

Also, while I was at it, I fixed the outdated function name `forContractAbstraction` in the decoder docs.


adrianmcli:
Thanks @haltman-at and @gnidan!

PR #2574, opened 2019-11-08
Remove unused dependencies from decoder

haltman-at:
(Part of #2505)

This PR removes some unused dependencies from the decoder (or switches them to `devDependencies` if they were used only in testing).

Btw there were some other things in there I didn't recognize at all, but I assume those are being used for some internal stuff somehow or whatever?  IDK.  I guess not in here but rather in `codec/package.json`.  Like what is `source-map-support`, and is it really a runtime dependency rather than a dev dependency?  ...I guess that's really the only case.  Anyway I left that one alone.

### DIFF
--- a/packages/decoder/package.json
+++ b/packages/decoder/package.json
@@ -38,29 +38,18 @@
     "@types/bn.js": "^4.11.2",
     "@types/debug": "^0.0.31",
     "@types/lodash.clonedeep": "^4.5.4",
-    "@types/lodash.escaperegexp": "^4.1.6",
-    "@types/lodash.partition": "^4.6.6",
-    "@types/lodash.sum": "^4.0.6",
-    "@types/semver": "^6.0.0",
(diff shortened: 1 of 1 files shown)


gnidan:
`source-map-support` is a regular dep, used to make sure the Node runtime handles source mappings back to the original TS. It should be used at the top of `lib/index.ts` in both the codec and the decoder

haltman-at:
Oh, I see!  Well, it is used in `codec` like you say, but in `decoder` it's neither used (nor listed in the package.json, there's no problem there).  Should I add it then?  (And should it be in all our TS stuff...?)

gnidan:
@haltman-at yep, should be added generally everywhere. also worth someone else verifying my knowledge here.

haltman-at:
OK, maybe that should be a separate PR then?

haltman-at:
(Welp, going ahead and merging this, and we can handle that later.)

PR #2575, opened 2019-11-08
JavaScript is terrible. Use TypeScript.

MicahZoltu:


adrianmcli:
@CruzMolina shouldn't a linter have caught this?

CruzMolina:
Good point @adrianmcli. I haven't messed with this file, so not sure if `eslint` was catching the unused var.

eggplantzzz:
The linter (with current settings) wouldn't catch this because `e` could be some global. The unused var rule wouldn't catch this either since it is used :)

CruzMolina:
Ah, looks like turning on `no-undef` would have caught this.

PR #2576, opened 2019-11-08
Normalize the source paths so they match when reading artifact files.

MicahZoltu:
This is a bandaid solution that "works for me".  The problem is that the path written to the build.json files is not internally consistent on Windows (sometimes I get `/C/Path/to/file.sol` and sometimes I get `C:\Path\to\file.sol` and it is also not consistent with what `glob` returns (`c:/Path/to/file.sol`).

This is a terrible bandaid fix that addresses my specific instance of this problem, but I'm sure doesn't work for all scenarios.

### DIFF
--- a/packages/compile-solidity/profiler/index.js
+++ b/packages/compile-solidity/profiler/index.js
@@ -34,7 +34,9 @@ module.exports = {
     let sourceFilesArtifactsUpdatedTimes = {};
 
     try {
-      const sourceFiles = await getFiles();
+      const sourceFiles = (await getFiles())
+        .map(x => x.replace(/\//g, '\\'))
+        .map(x => /^[a-zA-Z]:\\.*/.test(x) ? `${x.slice(0,1).toUpperCase()}${x.slice(1)}` : x);
       sourceFilesArtifacts = readAndParseArtifactFiles(
         sourceFiles,
         contracts_build_directory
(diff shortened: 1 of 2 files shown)


eggplantzzz:
So you are saying that you get paths that look like both `some\path\artifact.json` and `some/path/artifact.json`? I'll take a look for a bug in the artifactor to try and see why this is happening.

MicahZoltu:
Looking at the files in `build/contracts/*.json`, I see the following different types of values for the `sourcePath` key:
1. `@org/package/path/to/File.sol`
1. `C:\\Path\\to\\file.sol`
1. `C:/Path/to/File.sol`

eggplantzzz:
Ok, thanks for reporting this. I'll take a look and see if I can get it resolved!

fainashalts:
HI @MicahZoltu did you happen to give any more thought to this? I see how this is a bandaid fix. Wondering if there is a better way to do more general path normalization here.

MicahZoltu:
Once I got it "working for me" I moved on.  I definitely encourage finding a better fix than this, but unfortunately I don't have one.  😢

fainashalts:
Thanks for getting back to us on this, @MicahZoltu. I'm going to close this PR and open a more general issue for us to look into this further so we can come up with a more generalizable solution.

PR #2579, opened 2019-11-08
Internal improvement: From Web3Shim to InterfaceAdapter – Step Two: getCode

CruzMolina:
Continuation of #2473 that implements Step 2 for `getCode` as outlined [here](https://gist.github.com/gnidan/a55bbec28800f64d487054be18730691#file-step2-ts).


eggplantzzz:
I think this is cool!

PR #2583, opened 2019-11-11
@truffle/require migration to typescript

MrRefactoring:
@truffle/require rewritten in typescript


eggplantzzz:
This is also cool @MrRefactoring! If you can get it to pass CI and resolve the conflicts we'd be happy to move it forward. Also, could you please not bump the versions of extraneous packages? It is probably good but we want to do those as their own separate PRs. Thanks again!

eggplantzzz:
Hey @MrRefactoring, we're going to close this PR for maintenance for now. Let us know when you get some time to return to this. Thanks for the work!

PR #2585, opened 2019-11-11
Internal improvement: From Web3Shim to InterfaceAdapter – Step Two: getBlockNumber

CruzMolina:
Continuation of #2473 that implements Step 2 for `getBlockNumber` as outlined [here](https://gist.github.com/gnidan/a55bbec28800f64d487054be18730691#file-step2-ts).

### DIFF
--- a/packages/contract/lib/override.js
+++ b/packages/contract/lib/override.js
@@ -97,10 +97,10 @@ const override = {
     };
 
     // Start polling
-    let currentPollingBlock = await constructor.web3.eth.getBlockNumber();
+    let currentPollingBlock = await constructor.interfaceAdapter.getBlockNumber();
 
     const pollID = setInterval(async () => {
-      const newBlock = await constructor.web3.eth.getBlockNumber();
+      const newBlock = await constructor.interfaceAdapter.getBlockNumber();
 
       if (newBlock > currentPollingBlock) {
         currentPollingBlock = newBlock;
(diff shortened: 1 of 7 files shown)


eggplantzzz:
There are some instances of `getBlockNumber` that were missed. There are a bunch in the tests.

For example, see https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/test/provider.test.ts#L81.

There are also some instances in the decoder. See line 823 (wow!) https://github.com/trufflesuite/truffle/blob/develop/packages/decoder/lib/decoders.ts#L823.

Also https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/testing/testrunner.js#L149. And it looks like there are a bunch more.

Did we want to get all these all at once?

CruzMolina:
@eggplantzzz , are you checking the correct branch?

Outside of what we were already testing using the `Web3Shim`, we've only been updating tests as necessary. (I think this is the last PR needed to make it easier to swap over all the relevant tests from using `Web3` directly).

Regarding the `decoder`/`codec`/`debugger` pkgs, those are intended (so far) to only work with Ethereum, so they use `Web3` directly instead of the adapter.

eggplantzzz:
We don't want to switch the tests over to using `interfaceAdapter.getBlock` instead of `web3.eth.getBlock`? It seems like that would more appropriately test things.

CruzMolina:
We def do, the `Web3Shim` has barely been tested at all. There's a branch for it & a PR I closed a while back to switch over all the relevant tests.

eggplantzzz:
Oh ok, so are you going to update the tests that currently use `web3.eth.getBlock`?

CruzMolina:
Not planning to on this specific PR.

PR #2591, opened 2019-11-13
Fix accidental breaking change to resolver

gnidan:
Surfaced from [Gitter conversation](https://gitter.im/ConsenSys/truffle?at=5dcb38ba4adf071a841b760d), this PR fixes a recent breakage caused by the introduction of a networkType argument to the root @truffle/contract factory.

Since @truffle/contract is documented to work via a separate "provision" step, this PR removes the recently-added pre-provision network configuration and aligns it with the more mature provider configuration, since they are effectively related.

This consolidates `setNetworkType` / `setProvider` inside @truffle/contract to be simple facades to a new `configureNetwork` method, and implements the newly-required `setNetworkType` method to the InterfaceAdapter (necessary for the parallelism).

### DIFF
--- a/packages/contract/index.js
+++ b/packages/contract/index.js
@@ -2,10 +2,8 @@ const Schema = require("@truffle/contract-schema");
 const Contract = require("./lib/contract");
 const truffleContractVersion = require("./package.json").version;
 
-const contract = (json = {}, networkType = "ethereum") => {
-  const normalizedArtifactObject = Object.assign({}, Schema.normalize(json), {
-    networkType
-  });
+const contract = (json = {}) => {
+  const normalizedArtifactObject = Schema.normalize(json);
 
   // Note we don't use `new` here at all. This will cause the class to
   // "mutate" instead of instantiate an instance.
--- a/packages/contract/lib/contract/constructorMethods.js
+++ b/packages/contract/lib/contract/constructorMethods.js
@@ -4,6 +4,30 @@ const execute = require("../execute");
 const bootstrap = require("./bootstrap");
 
 module.exports = Contract => ({
+  configureNetwork({ networkType, provider } = {}) {
+    // otherwise use existing value as default (at most one of these)
+    networkType = networkType || this.networkType;
+    provider = provider || this.currentProvider;
+
+    if (this.interfaceAdapter) {
+      // implies this.web3 also
+      // update existing
+      this.interfaceAdapter.setNetworkType(networkType);
+      this.interfaceAdapter.setProvider(provider);
+
+      this.web3.setNetworkType(networkType);
+      this.web3.setProvider(provider);
+    } else {
+      // create new
+      this.interfaceAdapter = new InterfaceAdapter({ networkType, provider });
+      this.web3 = new Web3Shim({ networkType, provider });
+    }
+
+    // save properties
+    this.currentProvider = provider;
+    this.networkType = networkType;
+  },
+
   setProvider(provider) {
     if (!provider) {
       throw new Error(
@@ -11,9 +35,7 @@ module.exports = Contract => ({
       );
     }
 
-    this.web3.setProvider(provider);
-    this.interfaceAdapter.setProvider(provider);
-    this.currentProvider = provider;
+    this.configureNetwork({ provider });
   },
 
   new() {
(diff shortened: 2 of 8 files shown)


packages/contract/index.js
```diff
@@ -2,10 +2,8 @@ const Schema = require("@truffle/contract-schema");
 const Contract = require("./lib/contract");
 const truffleContractVersion = require("./package.json").version;
 
-const contract = (json = {}, networkType = "ethereum") => {
-  const normalizedArtifactObject = Object.assign({}, Schema.normalize(json), {
-    networkType
-  });
+const contract = (json = {}) => {
+  const normalizedArtifactObject = Schema.normalize(json);
```
CruzMolina: ✨

packages/contract/lib/contract/constructorMethods.js
```diff
@@ -128,14 +150,12 @@ module.exports = Contract => ({
...
   setNetworkType(networkType = "ethereum") {
-    if (this.web3) {
-      this.web3.setNetworkType(networkType);
-    }
-
-    this.networkType = networkType;
+    this.configureNetwork({ networkType });
   },
 
   setWallet(wallet) {
```
CruzMolina: ~Do we really even support this method?~

[Apparently we do, just not with `geth`](https://github.com/trufflesuite/truffle/blob/develop/packages/contract/test/methods.js#L539-L541).

packages/core/lib/testing/testrunner.js
```diff
@@ -77,11 +77,7 @@ TestRunner.prototype.initialize = function(callback) {
         function(err, data) {
           if (err) return callback(err);
 
-          var contracts = data
-            .map(JSON.parse)
-            .map(json =>
-              contract(json, self.config.networks[self.config.network].type)
-            );
+          var contracts = data.map(JSON.parse).map(json => contract(json));
```
CruzMolina: need to `provision` here.

packages/contract/lib/contract/constructorMethods.js
```diff
@@ -4,16 +4,38 @@ const execute = require("../execute");
 const bootstrap = require("./bootstrap");
 
 module.exports = Contract => ({
+  configureNetwork({ networkType, provider } = {}) {
```
CruzMolina: Hurray! 👐

packages/interface-adapter/lib/interface-adapter.ts
```diff
@@ -14,31 +14,44 @@ import { Provider } from "@truffle/provider";
...
-          `Sorry, "${options.networkType}" is not supported at this time.`
-        );
-    }
-    return this.adapter;
+    this.provider = options.provider;
+
+    this.setNetworkType(options.networkType);
+  }
+
+  public setNetworkType(networkType: InterfaceAdapterOptions["networkType"]) {
```
CruzMolina: 👍

packages/core/lib/testing/testrunner.js
gnidan: I don't think so... all this is used for is to get the ABI, which doesn't need the provider

CruzMolina: Ah, so I think the change here is from a cherry-picked commit from the `alphaTez` branch.

This is resolved, but there's still an instance or two that needs to be updated. Lemme update the main review comment.

gnidan:
@CruzMolina I removed the extra arg in lib/console; for [testsource](https://github.com/trufflesuite/truffle/blob/dont-network-type-me/packages/core/lib/testing/testsource.js#L66), I don't think we need provisioning.

Looking at what it uses:
- `network_id`
- `isDeployed()`

The former is separate from provisioning, in that it tells the abstraction _which_ network, but not how to connect. Similarly, with the latter, no network connection is required, since [that method](https://github.com/trufflesuite/truffle/blob/dont-network-type-me/packages/contract/lib/contract/constructorMethods.js#L113-L123) doesn't do any kind of verification.

gnidan:
@CruzMolina it's probably worth further cleanup here so that this distinction is more clear in code. Ultimately this comes from the idea I have that being able to delay provisioning buys us architectural flexibility in the future.

_shrug_ food for thought. Thanks for catching the things I missed!

PR #2593, opened 2019-11-13
Internal improvement: add @truffle/debugger to code coverage report

CruzMolina:
~Let's see what happens...~

Cool, looks like this is ready.


CruzMolina:
AFAIK, the change to the `.nycrc` ensures `@truffle/debugger` is included in the Coveralls report for the `coverage` job.

If there's something I'm missing here, will gladly revert the commit.

PR #2601, opened 2019-11-15
Internal improvement: Update migrations

eggplantzzz:
This PR removes the `async` library from `@truffle/migrate`. It, in my opinion, also cleans up the code a bit and makes it easier to read.

### DIFF
--- a/packages/migrate/index.js
+++ b/packages/migrate/index.js
@@ -141,7 +141,7 @@ const Migrate = {
     });
 
     try {
-      for (let migration of migrations) {
+      for (const migration of migrations) {
         await migration.run(clone);
       }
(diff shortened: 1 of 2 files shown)


packages/migrate/index.js
```diff
@@ -144,30 +140,22 @@ const Migrate = {
...
-            error: error ? error.toString() : null
-          });
-
-          if (error) return reject(error);
-          return resolve();
-        }
-      );
-    });
+    try {
+      for (let migration of migrations) {
```
CruzMolina:
```suggestion
      for (const migration of migrations) {
```

just a thought. we might mutate `Migration` so maybe not a good suggestion.

PR #2605, opened 2019-11-15
Dependency Update: web3-provider-engine@14.1.0

CruzMolina:
Take `@truffle/hdwallet-provider` off the `trufflesuite` hacky fork of `web3-provider-engine@~14.0.6` and onto `web3-provider-engine@14.1.0`.

Also unleashes a necessary shim to get `Web3` working with `web3-provider-engine`.
`web3-provider-engine` no longer supports synchronous requests, so we shim Web3 & also tweak `HdWalletProvider.send` to use `engine.sendAsync`.

QA/Smoke tested locally.

### DIFF
--- a/packages/hdwallet-provider/src/index.ts
+++ b/packages/hdwallet-provider/src/index.ts
@@ -177,7 +177,7 @@ class HDWalletProvider {
     payload: JSONRPCRequestPayload,
     callback: JSONRPCErrorCallback | Callback<JsonRpcResponse>
   ): void {
-    return this.engine.send.call(this.engine, payload, callback);
+    return this.engine.sendAsync.call(this.engine, payload, callback);
   }
 
   public sendAsync(
--- a/packages/hdwallet-provider/src/index.ts
+++ b/packages/hdwallet-provider/src/index.ts
@@ -163,7 +163,6 @@ class HDWalletProvider {
         Web3.providers.HttpProvider.prototype.send;
       this.engine.addProvider(
         new ProviderSubprovider(
-          // @ts-ignore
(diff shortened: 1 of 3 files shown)


packages/hdwallet-provider/src/index.ts
```diff
@@ -177,7 +177,7 @@ class HDWalletProvider {
     payload: JSONRPCRequestPayload,
     callback: JSONRPCErrorCallback | Callback<JsonRpcResponse>
   ): void {
-    return this.engine.send.call(this.engine, payload, callback);
+    return this.engine.sendAsync.call(this.engine, payload, callback);
```
gnidan: really?

CruzMolina: Ya, have an alternative in mind?

eggplantzzz: What do you mean by really? @CruzMolina can you explain exactly what this is about?

CruzMolina: Need to look into this a bit more. Web3 might not have actually dropped `sendAsync`.

The problem is Web3 `1.x` came out and formally removed support for `sendAsync`. `web3-provider-engine` responded by removing support for sync requests & forcing providers to have a `sendAsync` method.

CruzMolina: Idk, the options here kind've suck.

We could downgrade Web3 here to `0.x` & not have to do the hack on this line (or the shimming).

CruzMolina:
Closing since there doesn't seem to be interest in making the necessary adjustments to upgrade this dependency and transition `@truffle/hdwallet-provider` off of our fork.

Issue #2555, opened 2019-11-05
NPM packages don't have pinned versions.


MicahZoltu:
- [X] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue
`npm install @truffle/workflow-compile` will result in getting different transitive dependencies depending on what is available from NPM at install time.  While this is a specific problem that can be addressed, the bigger issue is that none of the distributed NPM packages appear to include a `package-lock.json` which I believe is required by NPM in order to fetch fixed versions.  Without it, dependency versions from `package.json` will be used and any transitive dependencies will follow whatever versioning strategy they have internally.

## Steps to Reproduce
1. Create new project.
1. `npm install @truffle/workflow-compile`
1. Commit (including `package-lock.json`) and push to GitHub.
1. Wait for automated security check in the background.
1. Receive email notifying you of security issues with packages.

## Expected Behavior
All Truffle packages published to NPM include a `package-lock.json` file that includes a fixed set of dependencies (including transitives).

## Actual Results
All Truffle packages appear to be published to NPM without a `package-lock.json` file, which means dependency versions are calculated an user install time.

## Environment

* Operating System: Windows 10
* Ethereum client: N/A
* Truffle version (`truffle version`): 5.0.43
* node version (`node --version`): 12.3.0
* npm version (`npm --version`): 6.9.0

gnidan:
We definitely want to look into this, perhaps urgently.

My biggest question is: what do other projects that use Lerna/Yarn do about this?

Also cc @davidmurdoch because I may recall your mentioning this in the past? Or was that something else?

gnidan:
Also @eggplantzzz makes a strong case that this is glaringly unbelievable and warrants reproduction.

davidmurdoch:
Truffle uses `yarn`, which doesn't generate a `package-lock.json` or `npm-shrinkwrap.json` (note: only `npm-shrinkwrap.json` can be published to npm). If `npm-shrinkwrap.json` files are never published, `npm` will always respect semver for all dependencies and transitive dependencies based on the contents of 
the package.json.

Example:

If `@truffle/fictional-package@1.1.1`'s `package.json` contains:
```
"dependencies": {
  "some-package": "1.2.3"
}
```

and `some-package@1.2.3`'s `package.json` contains:

```
"dependencies": {
  "other-package": "^4.5.6"
}
```

[...]

PR #2609, opened 2019-11-16
Internal improvement: Create shrinkwrap for truffle-cli

CruzMolina:
Pins all `packages/truffle` deps & pushes shrinkwrap generated from all cli dependencies.

Inspired by #2555 .


CruzMolina:
There's two cases that come to mind:
- If we add a dep to the cli, we'll need to update the shrinkwrap.
- If we expect or want devs to use the cli as a devDep, we'll need to update the shrinkwrap & publish the cli, _after_ bumping/publishing the org pkgs.

It seems that we recommend devs/users install the cli globally and don't expect them to consume it as a devDep. Still, not sure whether we want to move forward with this.

CruzMolina:
Closing since there doesn't seem to be interest in doing this.

PR #2619, opened 2019-11-19
Internal improvement: Optimize webpack rules

CruzMolina:
Only uses `shebang-loader` where needed & removes seemingly unused `imports-loader` rule along w/ the `imports-loader` devDep.

### DIFF
--- a/packages/truffle/cli.webpack.config.js
+++ b/packages/truffle/cli.webpack.config.js
@@ -56,8 +56,14 @@ module.exports = {
   devtool: "source-map",
   module: {
     rules: [
-      { test: /rx\.lite\.aggregates\.js/, use: "imports-loader?define=>false" }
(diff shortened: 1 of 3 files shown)


packages/truffle/cli.webpack.config.js
```diff
@@ -56,8 +56,14 @@ module.exports = {
   devtool: "source-map",
   module: {
     rules: [
-      { test: /\.js$/, use: "shebang-loader" },
-      { test: /rx\.lite\.aggregates\.js/, use: "imports-loader?define=>false" }
+      {
+        test: /\.js$/,
+        include: [
+          path.resolve(__dirname, "../core"),
+          path.resolve(__dirname, "../environment")
```
gnidan: Maybe it's worth a comment above each of these to say why we're using the shebang-loader in that package.

PR #2620, opened 2019-11-19
Internal improvement: tidy up externals in webpack config

CruzMolina:
Converts function in `externals` to a regex. ~Also turns out `mocha` behaves and bundles correctly (🤞) .~


CruzMolina:
Uh oh, turns out `mocha` still doesn't bundle well.

PR #2624, opened 2019-11-20
Internal improvement: add ens-related conditional to provisioner

CruzMolina:
There should probably be a `TruffleContract` method that handles this (`abstraction.setEns`) & errors appropriately if something bad gets passed, but this seems like a good start.

### DIFF
--- a/packages/provisioner/index.js
+++ b/packages/provisioner/index.js
@@ -13,7 +13,9 @@ var provision = function(abstraction, options) {
     abstraction.setNetworkType(options.networks[options.network].type);
   }
 
-  abstraction.ens = options.ens;
+  if (options.ens) {
+    abstraction.ens = options.ens;
+  }
 
   ["from", "gas", "gasPrice"].forEach(function(key) {
     if (options[key]) {


eggplantzzz:
Why do you say that this should be handled inside Truffle contract? Note that `ens` is now a default property on a Truffle config object.

eggplantzzz:
Oh I see, you are trying to follow the pattern here. I would be interested in having a discussion some time about whether all these getters and setters in the config actually need to exist or whether they are "out-of-date" and we should move away from them to a simpler architecture.

Let me ask you a different question then, what do we gain from this PR besides consistency?

CruzMolina:
Ah, didn't notice the addition of `ens` on an initial TruffleConfig object.

I _think_ adding logic in a `getter/setter` for `ens` to check and throw or return `undefined` when there is a (network)`type` present in the config file and the `type` isn't `"ethereum"` should suffice in place of this code.

CruzMolina:
Closing since this is done on `alphaTez`.

PR #2626, opened 2019-11-20
Internal improvement: ensure proper decoding of solidity unit test event logs

CruzMolina:
This could still be working when bundled, so maybe it's just an internal improvement.


gnidan:
Not sure how this would actually fix anything, but approved regardless cause this is the direction we discussed anyway.

Issue #2599, opened 2019-11-15
Resolver.resolve doesn't make any sense


gnidan:
## Issue

[@truffle/resolver](https://github.com/trufflesuite/truffle/blob/develop/packages/resolver) provides two main methods for external use:

- `require()` (as in, `artifacts.require()`... `artifacts` just an instance of a resolver)
- `resolve()`

Now, the resolver's purpose is to load build artifacts, not sources, and so, when you do `artifacts.require("MetaCoin")`, it reads it from `build/contracts/MetaCoin.json`, not the Solidity. This is true even if you do `artifacts.require("MetaCoin.sol")`: Truffle finds the JSON file whose source was `MetaCoin.sol`, and that's what it loads.

And so you'd think, that the `resolve()` function is just a lightweight `require()` that returns the artifact _path_ instead of the fully loaded contract abstraction from that path.

But that's not what `resolve()` does! Look: [the one place we use this function inside Truffle](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-solidity/profiler/index.js#L167-L194) (the profiler). The profiler uses it to figure out Solidity `import` statements. And it looks like it expects `resolve` to give the path _and_ the loaded contents.

So, why can't we contain this functionality entirely within the profiler itself? Well, we need to provide a means to inform the profiler of additional sources at runtime, e.g. with Solidity tests' `import "truffle/DeployedAddresses.sol"`. That source is generated from a template and never lives on disk. Truffle hooks this up to the profiler by means of the [TestSource](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/testing/testsource.js), injected into the TestResolver, so that the profiler can include that in compilation.

There **is** a package associated with sources, [@truffle/contract-sources](https://github.com/trufflesuite/truffle/blob/develop/packages/contract-sources), but currently it only _lists_ available sources, and incompletely so. My first thought is that this is a better home for this functionality, although there are various limitations in the way:

- That package is just a static function; it doesn't get attached to the config, and so it doesn't have the same extant utility as the resolver (which is available throughout Truffle, already instantiated)
- How do we reconcile the duplication of "sources"? Does that get separated into its own package, out of resolver? What does that look like?

Blech. This is sort of related to #2469, and it'd probably be worth solving both this issue and that one at the same time.

gnidan:
[...]

PR #2645, opened 2019-11-25
Internal improvement: Move `Resolver.resolve` method

eggplantzzz:
This PR currently is a work in progress.

In response to #2599, the `Resolver.resolve` method will be removed from resolver and put in contract-sources. Additionally, the sources found in resolver have been separated into their own package called "sources". 

Both the resolver and contract-sources will now be requiring this new package to use. Still remaining as "work to be done" is the task of creating an actual npm package called something like `@truffle/sources` and publishing it.


packages/contract-sources/index.js
```diff
@@ -33,3 +35,27 @@ module.exports = (pattern, callback) => {
...
+      };
+    }
+  }
+
+  let message = "Could not find " + importPath + " from any sources";
+  if (importedFrom) message += "; imported from " + importedFrom;
+  throw new Error(message);
+}
+
+module.exports = {
```
gnidan: Instead of returning a collection of functions here, I think we should follow the pattern of the Resolver and use a class, then attaching it to the config.

PR #2646, opened 2019-11-27
Move Truffle Test over to new decoder

haltman-at:
This PR moves Truffle Test over to using Truffle Decoder to decode events, rather than ethers (by means of web3).  Now event decoding in Truffle Test is more comprehensive, more robust, and -- most obviously -- more colorful!

This PR has 3 commits.  (Edit: Well, OK, it originally did anyway.)  The first commit rewrites `testrunner.js` and `soliditytest.js` in async/await style, and adjusts `tests.js` to invoke them as such.  Note that the two rewritten files aren't exported and so this shouldn't break anything.

The second resolves the merge conflict resulting from @CruzMolina replacing all uses of `this.web3.utils` with `web3Utils`.

(Then there's a third inbetween, oops.)

The last actually moves these two files over to using the new decoder.  For `soliditytest.js`, this doesn't do very much other than move things over.  For `testrunner.js`, this also means adding an indicator for anonymous events (which were previously omitted entirely), adding handling for ambiguous events, and using `ResultInspector` to nicely print the decoded values (with color!).  I kept it all on a single line as it was before, regardless of length.

I did remove the type indicators that used to be printed; they seemed less necessary now.  Let me know if you think those should be added back in.

Also, I didn't do anything to handle full mode vs ABI mode; I just printed decodings in whatever form they came in.

Anyway yeah that's basically it.  Also I didn't change the compilation call over to the new compile interface, but I know that @gnidan is going to write a review telling me to do so. :P


haltman-at:
Hm, looks like this is failing CI, presumably because I didn't update tests; will have to see what's up wth that...

haltman-at:
Ugh, looks like I missed an `await` when moving things over to async/await.  I'd rebase to make it part of that commit, but there's conflicts... (I am doing a force push so it'll at least be grouped with those)

haltman-at:
Hm -- looks like at least one of the test failures was actually due to a bug in the `deployedBytecode` getter/setter in `contract`.  So, one of my commits now fixes that!

haltman-at:
And, another looks to have been due to a bug in `codec`...

packages/core/lib/testing/testrunner.js
```diff
@@ -49,231 +50,164 @@ function TestRunner(options = {}) {
...
+      case 0:
+        this.logger.log(`    Warning: Could not decode event!`);
+        break;
+      case 1:
+        this.logger.log("    " + printEvent(log.decodings[0]));
+        break;
+      default:
+        this.logger.log(`    Ambiguous event, possible interpretations:`);
+        for (const decoding of log.decodings) {
+          this.logger.log(`      Could be:`); //6 spaces to indent a little
```
gnidan: I think we should get rid of this line
```suggestion
```

packages/core/lib/testing/testrunner.js
```diff
@@ -49,231 +50,164 @@ function TestRunner(options = {}) {
...
+          this.logger.log("      " + printEvent(decoding));
```
gnidan: And just make this a bulleted list:
```suggestion
          this.logger.log("    - " + printEvent(decoding));
```

packages/core/lib/testing/testrunner.js
```diff
@@ -49,231 +50,164 @@ function TestRunner(options = {}) {
...
+      .map(({ name, indexed, value }) => {
+        let namePrefix = name ? `${name}: ` : "";
+        let indexedPrefix = indexed ? "<indexed> " : "";
+        let displayValue = util.inspect(
+          new Codec.Format.Utils.Inspect.ResultInspector(value),
+          {
+            depth: null,
+            colors: true,
+            maxArrayLength: null,
+            breakLength: Infinity,
```
gnidan: Can we make breakLength be 80?
```suggestion
            breakLength: 80,
```

(or whatever, to get the lines to be a bit shorter... I guess this might require finagling to get the indent to work right)

packages/core/lib/testing/testrunner.js
gnidan: Before:
Screen Shot 2019-11-27 at 1 24 14 PM

After:
Screen Shot 2019-11-27 at 1 25 24 PM

Looks better, I think

packages/core/lib/testing/testrunner.js
haltman-at: Sure!  I was just matching the old behavior which was printing the whole thing on one line.  As you say, though, splitting over multiple lines will require some indent finagling.

packages/core/lib/testing/testrunner.js
gnidan: Hm, might want to clean this up so multiple events are easier to read (maybe a blank line in between would suffice?):
Screen Shot 2019-11-27 at 2 25 59 PM

packages/contract/lib/contract/properties.js
```diff
@@ -269,6 +269,10 @@ module.exports = {
     get: function() {
       var code = this._json.deployedBytecode;
 
+      if (!code) {
+        return code;
```
eggplantzzz: Small stylistic thing but would it be better to explicitly return `null` rather than whatever falsy value `code` happens to be? Or is whatever falsy value code is important?

packages/core/lib/testing/soliditytest.js
```diff
@@ -1,178 +1,172 @@
...
+                case "valid":
+                  message = messageDecoding.value.asString;
+                  break;
+                case "malformed":
+                  //use buffer to convert hex to string
+                  //(this causes malformed UTF-8 to become U+FFFD)
+                  message = Buffer.from(
+                    messageDecoding.value.asHex.slice(2),
+                    "hex"
+                  ).toString();
```
eggplantzzz: Does it make sense at all to provide a fallback "default" case here?

packages/core/lib/testing/soliditytest.js
```diff
@@ -1,178 +1,172 @@
...
+    //HACK: promisify seems to lose context, so we explicitly bind compile
+    const contracts = await util.promisify(
+      compile.with_dependencies.bind(compile)
+    )(
+      runner.config.with({
+        paths: [
+          path.join(__dirname, "Assert.sol"),
+          path.join(__dirname, "AssertAddress.sol"),
+          path.join(__dirname, "AssertAddressArray.sol"),
+          // path.join(__dirname, "AssertAddressPayableArray.sol"), only compatible w/ ^0.5.0
```
eggplantzzz: Should we just delete this?

packages/core/lib/testing/soliditytest.js
```diff
@@ -1,178 +1,172 @@
...
+  async deployTestDependencies(abstraction, dependencyPaths, runner) {
+    debug("deploying %s", abstraction.contract_name);
     const deployer = new Deployer(
       runner.config.with({
         logger: { log() {} }
       })
     );
 
+    debug("starting deployer");
+    await deployer.start();
```
eggplantzzz: So this code worked before without this? Do you know why?

packages/core/lib/testing/testrunner.js
```diff
@@ -49,231 +50,164 @@ function TestRunner(options = {}) {
...
-                abi_entry: abi
-              };
-            }
-          });
-          callback();
-        }
-      );
-    });
-  };
+  let files = await fs.readdir(this.config.contracts_build_directory);
```
eggplantzzz: Maybe it is just better to use `fs.readdirSync`?

packages/core/lib/testing/testrunner.js
```diff
@@ -49,231 +50,164 @@ function TestRunner(options = {}) {
...
-          );
-        },
-        function(err, data) {
-          if (err) return callback(err);
-
-          var contracts = data.map(JSON.parse).map(json => contract(json));
-          var abis = _.flatMap(contracts, "abi");
+  if (this.first_snapshot) {
+    // // Make the initial deployment (full migration).
+    // await this.deploy();
```
eggplantzzz: Was this intended to be left in?

packages/core/lib/testing/testrunner.js
```diff
@@ -49,231 +50,164 @@ function TestRunner(options = {}) {
...
-      );
-    });
-  };
+  let files = await fs.readdir(this.config.contracts_build_directory);
 
-  if (self.first_snapshot) {
-    // // Make the initial deployment (full migration).
-    // self.deploy(function(err) {
-    //   if (err) return callback(err);
+  files = files.filter(file => path.extname(file) === ".json");
```
eggplantzzz: How do you feel about just tacking this filter call on to the call to get the files above? It makes me feel weird when variables keep on being reassigned like this. If you don't want to then I won't fight you on it though.

packages/core/lib/testing/testrunner.js
```diff
@@ -49,231 +50,164 @@ function TestRunner(options = {}) {
...
 
-    self.snapshot(function(err, initial_snapshot) {
-      if (err == null) {
-        self.can_snapshot = true;
-        self.initial_snapshot = initial_snapshot;
-      }
+  let data = await Promise.all(
+    files.map(
+      async file =>
+        await fs.readFile(
```
eggplantzzz: We can probably use the synchronous version of this as well, huh? Then we can get rid of the `Promise.all` and the other async stuff here.

packages/core/lib/testing/testrunner.js
```diff
@@ -49,231 +50,164 @@ function TestRunner(options = {}) {
...
-  this.interfaceAdapter
-    .getBlockNumber()
-    .then(result => {
-      const one = web3Utils.toBN(1);
-      const resultBN = web3Utils.toBN(result);
-
-      // Add one in base 10
-      this.currentTestStartBlock = resultBN.add(one);
+TestRunner.prototype.startTest = async function() {
+  let blockNumber = await this.web3.eth.getBlockNumber();
```
eggplantzzz: Should this call use `interfaceAdapter`?

packages/contract/lib/contract/properties.js
haltman-at: I don't know that the specific value is important, no.  If we do want to make it a single explicit thing I'd lean towards `undefined` over `null` but obviously that doesn't really matter that much.

packages/core/lib/testing/soliditytest.js
haltman-at: No, those are the only two cases, as specified by the format.

packages/core/lib/testing/soliditytest.js
haltman-at: We probably should!  I decided not to mess with the commented-out code (OK, I updated it in places :P ) but I didn't delete it because I wasn't sure what it was for.  If anyone who knows (@gnidan?) can chime in, it'd be helpful.

packages/core/lib/testing/soliditytest.js
haltman-at: No, the call to `deployer.start` was there before.  It looked like it was at the end rather than at the beginning, but I'm assuming that was a misunderstanding on my part -- control flow in callback style is *confusing*!

packages/core/lib/testing/testrunner.js
haltman-at: I wasn't aware that existed!  Yes, that would probably be better.  Thanks!

packages/core/lib/testing/testrunner.js
haltman-at: See comment above about commented-out code.

packages/core/lib/testing/testrunner.js
haltman-at: How would that work exactly?  I haven't been keeping up with that.

CruzMolina: I'm surprised we still even have the `Web3Shim` in here, since it doesn't look like it's being used.
```suggestion
  let blockNumber = await this.interfaceAdapter.getBlockNumber();
```

haltman-at: Well, `this.web3` is a `Web3Shim`, so it is being used in that one place.  But if we change this one call like you say, then we can get rid of it.  So that's nice.  (The `web3` object/shim used to also be used for getting the logs, but now that's handled by `decoder.events()`.)

haltman-at:
Quick note: I've changed this so that it will now use `definedIn` rather than `class` when available -- like when you see `Class.Event`, it'll be the class that defined it.  If that's not available it'll fall back to the class that emitted it, but this should only happen in cases where it shouldn't cause serious confusion.

packages/core/lib/testing/testrunner.js
haltman-at: OK, I did this, but in order for the result to make sense I also had to put each argument on a separate line.  And so I also ended up putting a blank line between events like you suggested elsewhere.  Also, I decided to reintroduce the type marker, because why not.

I am hoping the result is good?  I encourage you to try it out and see what you think.

packages/core/lib/testing/testrunner.js
haltman-at: I ended up adding the blank line, and also used `*` instead of `-` for clarity.

packages/core/lib/testing/soliditytest.js
gnidan: delete it. it's in the git history if we want it again

haltman-at:
Minor note, I've stuck a commit in this PR to fix a problem in `codec`'s `typeString()` function.  It was supposed to only print the mutability if the mutability was something other than `nonpayable`; instead, I reversed the test and it only printed the mutability of the mutability was equal to `nonpayable`.  Oops.  Well, fixed now.

haltman-at:
Question: While it's not due to this PR, this is where it's coming up, so I'm wondering: Is the message you get when you try to decode an indexed argument of reference type (e.g. `string`) too long?  This is basically the only place it appears at the moment, it could totally be shortened if you think it's necessary or think it can be improved.

haltman-at:
OK, after talking to @gnidan, I added some line-wrapping there.  While I was at it I changed the existing line-wrapping logic to go based on total length rather than just the length of the first line...

haltman-at:
One last quick change before I merge, I've lowered `breakLength` to account for indentation somewhat.

PR #2649, opened 2019-11-27
Enhancement: add goerli network id to the migrations network whitelist

### LINKED ISSUE #2648: Add Goerli's network id to the migrations whitelist
Noticed during the trufflesuite crash course workshop that migrations targeting `goerli` don't dry-run by default. We should add it to the migrations network whitelist.

CruzMolina:
Resolves #2648.


eggplantzzz:
Rad

PR #2652, opened 2019-12-02
Internal improvement: use workspace dependency lerna when bootstrapping

### LINKED ISSUE #2651: Leverage workspace lerna dependency when bootstrapping
For local project development consistency's sake, we should probably use the workspace `lerna` dependency instead of requiring a global version be installed and update the README.

CruzMolina:
Resolves #2651.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "npx lerna bootstrap",
     "prepare": "lerna run prepare --stream --concurrency=1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "private": true,
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "npx lerna bootstrap",
```
gnidan: I think this part isn't necessary though

CruzMolina: https://github.com/trufflesuite/truffle/pull/2683/checks?check_run_id=347793212#step:5:6 😕

Issue #2655, opened 2019-12-05
HDWalletProvider silently exits when rpc endpoint is not available


TripleSpeeder:
- [ X ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.
-> No luck there :-(

---------------------------

## Issue

Using HDWalletProvider in a node.js application. When the rpc endpoint is not available the whole node process exits silently, instead of throwing an error.

## Steps to Reproduce

Short testcase:
```javascript
onst HDWalletProvider = require('@truffle/hdwallet-provider')
const Web3 = require('web3')

const providerString = 'http://127.0.0.1:8545'

async function checkPlainProvider() {
    console.log("Setting up plain web3")
    const plainWeb3 = new Web3(providerString)
    try {
        console.log("\tTrying to get nodeInfo. Should throw an Error due to invalid provider.")
        let result = await plainWeb3.eth.getNodeInfo()
        console.log("\tResult: " + result)
    } catch (error) {
        console.log("\tError message: " + error.message)
        console.log("\tCaught error as expected. Returning.")
        return
    }
    console.log("\tShould never get here")
}

async function checkHDWalletProvider() {
    console.log("Setting up web3 with HDWalletProvider")
    const hdwalletprovider = new HDWalletProvider('0xaabbcc', providerString)
    const hdWeb3 = new Web3(hdwalletprovider)
    try {
        // This call should throw an exception. But instead the whole process dies without showing any error.
        console.log("\tTrying to get nodeInfo. Should throw an Error due to invalid provider.")
        let result = await hdWeb3.eth.getNodeInfo()
        console.log("\tResult: " + result)
    } catch (error) {
        console.log("\tError message: " + error.message)
        console.log("\tCaught error as expected. Returning.")
        return
    }
    console.log("\tShould never get here...")
}

async function runme() {
    await checkPlainProvider()
    await checkHDWalletProvider()
}

runme()
```

Output:
```
> node providertest.js 
Setting up plain web3
        Trying to get nodeInfo. Should throw an Error due to invalid provider.
        Error message: Invalid JSON RPC response: ""
        Caught error as expected. Returning.
Setting up web3 with HDWalletProvider
        Trying to get nodeInfo. Should throw an Error due to invalid provider.
>
```

## Expected Behavior
An error should be thrown instead of silent exit.

## Actual Results

The process just dies, no exit code set, no exception thrown.

## Environment

* Operating System: Arch Linux
[...]

PR #2657, opened 2019-12-06
Upgrade dependency: web3@1.2.4

haltman-at:
Hoping this works...

### DIFF
--- a/packages/artifactor/package.json
+++ b/packages/artifactor/package.json
@@ -30,7 +30,7 @@
     "temp": "^0.8.3",
     "ts-node": "^8.3.0",
     "typescript": "^3.6.3",
-    "web3": "1.2.2"
+    "web3": "1.2.4"
   },
   "publishConfig": {
     "access": "public"
--- a/packages/contract/package.json
+++ b/packages/contract/package.json
@@ -27,7 +27,7 @@
     "bignumber.js": "^7.2.1",
     "ethereum-ens": "^0.7.7",
     "ethers": "^4.0.0-beta.1",
-    "web3": "1.2.2",
+    "web3": "1.2.4",
     "web3-core-promievent": "1.2.2",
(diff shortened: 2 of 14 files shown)


haltman-at:
...and, nope.  Compile error in `hdwallet-provider`.

cgewecke:
@haltman-at Visiting from web3. First impression is that TS is catching on HDWallet's minimal interface? It's wrapping web3 in many cases but not implementing some of the public facing properties and methods.

```
test/provider.test.ts:48:22 - error TS2345: Argument of type 'HDWalletProvider' is not assignable to parameter of type 'provider'.
  Type 'HDWalletProvider' is missing the following properties from type 'WebsocketProvider': isConnecting, responseCallbacks, notificationCallbacks, connected, and 9 more.
```

Those are listed [here][1] fwiw. You should be able to get away with just implementing the HTTP Provider interface in the short term. Some details...

+ `supportsSubscriptions() : boolean` is an alias for 'using a WebsocketProvider'. Believe, the Wallet only connects with an HttpProvider atm so perhaps this can always return false. I see a url in the tests that uses `ws` but I suspect that's only working because ganache is atypical. 

+ `disconnect() : boolean`:  can probably be mapped to the provider-engine's [stop method][2].  

+ `connected: boolean` : is false on new, true after you've connected (See #2655 for a current bug here about that)

+ `host`: is the url 

[1]: https://github.com/ethereum/web3.js/blob/5b8c12ecc4abf396ce06fe999a9651380b5beb36/packages/web3-core-helpers/types/index.d.ts#L135-L149
[2]: https://github.com/trufflesuite/provider-engine/blob/3538c60bc4836b73ccae1ac3f64c8fed8ef19c1a/index.js#L65-L69

CruzMolina:
Closing for now.

PR #2659, opened 2019-12-06
Dependency update: @taquito/taquito

CruzMolina:
Updates `@taquito/taquito` to 5.2.0-beta.1.

Also updates a cameligo test contract to match new cameligo syntax.


CruzMolina:
Looks like launchpad.net is currently down. Mergin' away.

PR #2663, opened 2019-12-10
Fix: Downgrade web3 to 1.2.1

### LINKED ISSUE #2654: Investigate & fix bug surfaced in 5.0.44
Greenkeeper caught a bug released in 5.0.44 over in Colony's repo: JoinColony/colonyNetwork#734.

~Appears to be related to `InterfaceAdapter` changes, but it could be something else. Seems like this is relatively high priority.~

EDIT: So far it seems like this is actually a Web3.js issue that appeared in the past. See #1504, #1461, & ethereum/web3.js#2105.

CruzMolina:
Downgrades & pins Web3.js libraries to 1.2.1.

Resolves #2654.


haltman-at:
Would it be possible to get rid of all the stupid casts used in the decoder (and possibly elsewhere?) that were introduced to deal with the erroneous typings in 1.2.2?

PR #2668, opened 2019-12-10
Fix: bump network timeout check

CruzMolina:
It appears the default 5 second timeout is occasionally too short for live tezos networks.
Bumping to 10 seconds has proven thus far to resolve the issue.


CruzMolina:
Looks like there's a bug in the `latest` ligo docker image. Merging since this passes locally.

PR #2670, opened 2019-12-10
Add memory display, etc, to debugger CLI, and also add more color

haltman-at:
This PR does a few things.

The main thing it does is expand the `p` command in the debugger so that, instead of just printing the current instruction, stack, and gas remaining, it can also (if requested) print out memory, calldata, and (a partial view of) storage.

Before we discuss other things, let's discuss: How does one use these new displays?

To include `memory`, `storage`, or `calldata` in your printout, you can simply type `p memory`, `p storage`, or `p calldata`; each of these can also be abbreviated to a 3-letter (or longer) prefix, so `p mem` works, as does `p memo`, etc.  They can also be combined -- `p mem cal`, for instance.  Note that the order of the printout will always be the same; it doesn't depend on the order you enter these in.

Moreover, you can write things like `p +mem` to not only include memory in the printout, but add it to the defaults; from here on out, the printouts from both `p` and `;` will include memory, until you turn it off with `p -mem`.  If you really want, you can even turn off the stack printout with `p -sta`.  Again, all these can be combined; you can write things like `p cal +mem -sta` if you really want.  There is no way to turn off the file printout, instruction printout, gas printout, or current line printout.

Note that there wasn't really any way I could fit all this in the help text, so instead I just had the help text refer to the docs.  But this PR doesn't actually update the docs, since the website is a separate repo.  So, I'm going to have to make a PR for that as well, because this definitely needs documentation support.  I did think about ways to fit a partial explanation in the help text, but ultimately decided it wasn't worth it.  Let me know if you have a better idea.

As for the other things: These new displays are also colorized, and have addresses listed alongside the data; the stack display and instruction display have also been colorized, and the stack also has a sort of "addresses" alongside the data.

(FWIW: The instruction part of the instruction display is mint; the stack data is orange; memory and calldata are pink; and storage is purple.)

The storage display, of course, is necessarily partial, but I've done my best to convey this in the printouts.  I've made it so that storage addresses are only printed at the beginning or where there are gaps; an address will only be printed at the first of each run of consecutive storage slots.

There is no support for printing returndata.  We don't currently keep track of returndata, so this would require a whole new system for doing so.  Frankly, the only reason to keep track of returndata would be to allow someone to print it out, and returndata is nowhere near as relevant as calldata, so I don't expect anyone to be asking for this anytime soon.

Also, I've moved the instruction display below the stack display, so that the instruction is now actually displayed near its arguments.  I had instead wanted to invert the stack display to display the top on top, but @gnidan thought this was a better solution.

While it doesn't really relate to the main point of the PR, as long as I was colorizing things I've also added some color to the help text for clarity.

The code in this PR is not the best, but, y'know, it's the CLI. :P  Feel free to demand some cleanup if you think it's warranted.

I've attached some screenshots at the bottom to demonstrate the new colorization and new displays.

Note: I just realized the file printout can now be quite far from the current line printout.  After this PR is up I'm going to go back and add a commit to make it so those are now always next to each other.

Colorized help text display
Storage display
Memory and calldata displays
Stack and instruction displays


haltman-at:
OK, I moved the file printout next to the line printout, but also I noticed -- the defaults you set will not only affect `p` and `;`, but also `c` if you happen to land somewhere where there's no source code.  I guess that's OK?

haltman-at:
Oh yeah also worth calling out the way I did the calldata display, although you've already probably noticed it in the screenshot.  Whereas memory is shown word by word, calldata is shown with the first 4 bytes separate, right-aligned, and then word-by-word after that.  So whereas the memory addresses go 0x00, 0x20, 0x40, etc, the calldata addresses go 0x00, 0x04, 0x24, 0x44, etc.

Also, if we have nothing in storage, the storage message, instead of saying "No data in storage" (like "No data in memory", "No data in calldata", "No data on stack"), instead says "No known relevant data found in storage" to make it clear that this doesn't necessarily mean storage is actually empty.

haltman-at:
Btw I'm slightly worried that the purple I used for storage (Teams Purple, of course :) ) doesn't have enough contrast.  Let me know if you agree, I can use a different color from our palette.

haltman-at:
Oh this will need documentation for sure.  But the website is a separate repo so that'll have to be a separate PR.

Question: Do you have any ideas for how the help message could be improved so as to rely less on external documentation?  I couldn't really think of anything unfortunately...

Also before merging this I'm probably going to change that purple color to something more visible.

haltman-at:
Difficult to add tests for this; ignoring the coverage problem and merging anyway per @gnidan.

PR #2675, opened 2019-12-12
Update index.ts

AmrEShaker:


gnidan:
Thanks for opening this, but I'm afraid we can't accept this PR without a compelling argument... `keepAlive: false` was added because it was necessary to prevent exceeding the maximum number of open sockets in long-running processes. In such situations, Truffle may create many copies of the provider, and if keepAlive were enabled, these would all maintain their own open socket.

(This problem surfaced in https://github.com/trufflesuite/truffle/issues/1461, and `keepAlive: false` ended up being the fix.)

If you really think this is safe now, please explain.

PR #2691, opened 2019-12-17
Correctly disable VM errors for `truffle test`'s ganache-core

gnidan:
Previously, looks like lib/commands/test was trying to turn off Ganache VM errors (i.e., operate in standard mode)... but it was using the field name `noVMErrorsOnRPCResponse` instead of the correct `vmErrorsOnRPCResponse`.


gnidan:
Thanks!

PR #2693, opened 2019-12-17
Internal improvement: Update resolve method in test flow

eggplantzzz:
This PR eliminates a number of callbacks in the test flow in favor of a Promise-based architecture.

The sources for the resolver have also been moved into their own folder. A function returning all of these sources is now exposed and not each source independent of the others.

For the future, I start to wonder whether we should make `./packages/core/lib/testing/testresolver.js` a part of the regular resolver and include `TestSource` in the group of sources mentioned above.


eggplantzzz:
Oh yeah this is definitely breaking since it switches from callback interface to Promises. Is that a problem @gnidan?

PR #2700, opened 2019-12-18
Internal improvement: fix import/export typings bug

CruzMolina:
Also pins `@ensdomains/resolver` to v0.1.9.


eggplantzzz:
I don't think we want to pin that ens library, do we?

CruzMolina:
I had issues bundling that appeared to be specific to that package in 1.0.13 (which the caret pulls in locally)

```
ERROR in ./node_modules/@ensdomains/resolver/index.js
: Module not found: Error: Can't resolve './build/contracts/CustodialPublicResolver.json'
```

eggplantzzz:
Oh, I thought that was already pinned. Sounds good!

PR #2707, opened 2019-12-23
Rename TruffleLogger to Console

adrianmcli:
I hope I didn't miss any re-namings here, but basically this makes it so that usage requires `Truffle.log` rather than `TruffleLogger.log`.

### DIFF
--- a/packages/contract/lib/handlers.js
+++ b/packages/contract/lib/handlers.js
@@ -101,7 +101,7 @@ const handlers = {
       return context.promiEvent.reject(error);
     }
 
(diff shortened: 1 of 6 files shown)


adrianmcli:
Related: https://github.com/trufflesuite/truffle-logger-example/pull/6

adrianmcli:
Although it is kind of weird to import like this now:

```
import "truffle/Truffle.sol";
```

instead of:

```
import "truffle/TruffleLogger.sol";
```

Issue #469, opened 2017-06-29
Avoid compilation when doing tests


JustinDrake:
I am writing tests (leaving the contract code intact) but truffle will recompile every time I run `truffle test`. Ideally `truffle test` would be smart about recompiling only files that have changed. As a quick fix, a flag to skip compilation would do for me 👍

JustinDrake:
I see there's the `--compile-all` option. I wonder why the default contracts generated by `truffle init` get recompiled every time.

```
--compile-all: Compile all contracts instead of intelligently choosing.
```

jellegerbrandy:
I have the same problem:
```
jelle@:~/projects ★ truffle version
Truffle v3.3.0, bundle version: 3.3.1
Solidity v0.4.11 (solc-js)
jelle@:~/projects/daostack/play ★ mkdir play
jelle@:~/projects/daostack/play ★ cd play
jelle@:~/projects/daostack/play ★ truffle init
Downloading project...
Project initialized.

  Documentation: http://truffleframework.com/docs

Commands:

  Compile: truffle compile
  Migrate: truffle migrate
  Test:    truffle test

jelle@:~/projects/play ★ truffle compile
Compiling ./contracts/ConvertLib.sol...
Compiling ./contracts/MetaCoin.sol...
Compiling ./contracts/Migrations.sol...
Writing artifacts to ./build/contracts

jelle@:~/projects/play ★ truffle test
Using network 'development'.

Compiling ./contracts/ConvertLib.sol...
Compiling ./contracts/MetaCoin.sol...
Compiling ./test/TestMetacoin.sol...
Compiling truffle/Assert.sol...
Compiling truffle/DeployedAddresses.sol...


  TestMetacoin
    ✓ testInitialBalanceUsingDeployedContract (44ms)
    ✓ testInitialBalanceWithNewMetaCoin (38ms)

  Contract: MetaCoin
    ✓ should put 10000 MetaCoin in the first account
    ✓ should call a function that depends on a linked library (64ms)
    ✓ should send coin correctly (120ms)


  5 passing (594ms)

jelle@:~/projects/play ★ truffle test
Using network 'development'.

Compiling ./contracts/ConvertLib.sol...
Compiling ./contracts/MetaCoin.sol...
Compiling ./test/TestMetacoin.sol...
Compiling truffle/Assert.sol...
Compiling truffle/DeployedAddresses.sol...


  TestMetacoin
    ✓ testInitialBalanceUsingDeployedContract
[...]

PR #2709, opened 2019-12-26
Close #2661 - add no-compile flag

### LINKED ISSUE #2661: add no-compile flag
## Issue

It would be nice to have a `--no-compile` flag that skips compilation on commands that otherwise include it by default. This would also solve the issue here: https://github.com/trufflesuite/truffle/issues/469.

This is a command I needed and had to switch to buidler to use because I have some contracts which for legacy purposes must remain tied to solc4 and some contracts which should use solc5. In order to compile them I need to set a different config file for each compilation. Then when it comes to migrating or testing them, I want to prevent them from being compiled with the wrong compiler, so i include the `--no-compile` option.

robertmagier:
I throw an error when flags --compile-all or --debug are used together with -compile-none
Not sure if this is the correct way to do it. 
```

    if (config.compileAll && config.compileNone) {
      options.logger.log(
        "--compile-none can not be set together with --compile-all or --debug"
      );
      throw "Command line error.";
    }
```


packages/core/lib/test.js
```diff
@@ -98,26 +98,31 @@ const Test = {
...
-      return testResolver.require(testFilePath);
-    });
-
     const runner = new TestRunner(config);
+    var compilations = null;
+
+    // Do not compile contracts, test contracts, do not add test contracts to testing set
+    // when --compile-none flag is set
+    if (!config.compileNone) {
+      var { compilations } = await this.compileContractsWithTestFilesIfNeeded(
```
gnidan: Instead of declaring `compilations` twice, could you move this `compileNone` check to inside the `compileContractsWithTestFilesIfNeeded` method?

robertmagier: Ok. I can return {} if config.compileNone is set to true. Does it make sense ? 
I will still have to make check for config.compileNone flag in Test.run function to skip solidity tests if this flag is set.

robertmagier:
@gnidan Should I add test case for --compile-none flag scenario ?

robertmagier:
I added test file to test how to use --compile-none flag. In last scenario this test is failing exactly as in issue #469 Can someone verify if this scenario is correct ?

robertmagier:
@gnidan  @CruzMolina Pleeeease ;)

gnidan:
Ended up taking a different approach, but this should be done now as of #3208.

Sorry for the radio silence here, but I think this PR can now be closed, since that functionality now exists!

robertmagier:
All my work for nothing.

gnidan:
I'm really sorry about that @robertmagier! I felt awful but still decided that the alternate approach was simple enough.

Looking at this again now, however, I think your tests will provide lots of value! I'll reopen this with the plan to get these tests merged, at least!

eggplantzzz:
Some of this work was taken and used in https://github.com/trufflesuite/truffle/pull/3335. I gave coauthoring credit to @robertmagier.

robertmagier:
I am sorry for whining ;) Thank you for giving coauthoring credit to me as well. Let me know if you need any help here... or there.

eggplantzzz:
It's no problem :) I hate to have good work go to waste, it has been a busy time the last year or so but we have implemented some process to help us to stay on top of contributions like this one. Is your Windows CI PR still relevant?

eggplantzzz:
So this feature was merged in [this PR](https://github.com/trufflesuite/truffle/pull/3335) and should be released next week!

eggplantzzz:
Closing as this feature was released in version 5.1.47. Thanks!

PR #2713, opened 2020-01-02
Fix #2442 - add an option to compile only specific files

### LINKED ISSUE #2442: Question - Is there a way to compile just one contract in the contracts/ folder?
The title is pretty much my question. Is there a way to compile just one contract in the `contracts/` folder? If i have a lot of contracts in the Contracts forlder, but only need one, i stil have to wait for them all to compile when i compile from a fresh checkout.

## Expected Behavior
executing `truffle compile MyContract` should compile `contracts/MyContract.sol` only, and not all the contracts in the folder.

## Current Behavior
Truffle ignores the final argument and compiles everything, which is slow when i have a bunch of different contracts in one project.

## Possible Solution
As explained, parse the positional arguments as names of contracts to compile, and if not specified, compile everything as usual.

## Context
I'm writing a Makefile to allow me to document how to build different parts of a system my team is making, and we want to speed up some workflows, including CI. in CI, for example, when testing one contract with one subsystem, we don't want to wait for everything to compile

robertmagier:
I created function called compile.specific both in solidity and vyper compilers. They are almost identical except there is a difference in extensions names. Would be good to do some refactoring here, but I don't know where outside of compiler-solidity and compiler-vyper I should implement this function. I can think about creating function Profiler.specific in compiler-solidity which will iterate through provided file names and return list of paths according to defined extensions. 
`Profiler.specific(files,extensions)`
Let me know if it makes sense. 

I have also added one test to core/test/lib/commands/compile.js
But it is only testing solidity compilation and skipping vyper since there are no vyper files in sandbox.

### DIFF
--- a/packages/core/lib/commands/compile.js
+++ b/packages/core/lib/commands/compile.js
@@ -23,8 +23,7 @@ const command = {
     }
   },
   help: {
-    usage:
-      "truffle compile [--list <filter>] [--all] [--quiet]",
+    usage: "truffle compile [<sources>..] [--list <filter>] [--all] [--quiet]",
     options: [
       {
         option: "--all",
@@ -53,6 +52,11 @@ const command = {
         description:
           "Save the raw compiler results into <output-file>, overwriting any existing content."
       },
+      {
+        option: "sources",
+        description:
+          "List of files and its dependencies to be compiled. It will be ignored if --all flag is used."
+      }
     ],
     allowedGlobalOptions: ["network", "config"]
   },
@@ -78,6 +82,33 @@ const command = {
       );
     }
 
+    if (config._ && config._.length > 0) {
+      // set paths based on command-line inputs, transforming to absolute
+      // paths where appropriate
+      config.paths = config._.map(specifiedPath => {
+        // convert relative paths to absolute paths based on whether
+        // the naive absolute path exists on disk
+        //
+        // NOTE in case of collision where the specified path refers to some
+        // non-FS source (e.g. `truffle/Assert.sol`) and where that specified
+        // path corresponds to an existing file relative to the working dir.,
+        // this selects the latter as priority over the former.
+
+        const absolutePath = path.resolve(
+          config.working_directory,
+          specifiedPath
+        );
+
+        // i.e., pass the absolutePath if it's a real file, otherwise just
+        // pass whatever was specified.
+        if (fse.existsSync(absolutePath)) {
+          return absolutePath;
+        } else {
+          return specifiedPath;
+        }
+      });
+    }
+
     const compilationOutput = await WorkflowCompile.compile(config);
     if (options.saveIntermediate) {
       // Get the filename the user provided to save the compilation results to
@@ -86,7 +117,7 @@ const command = {
       await fse.writeFile(
         compilationOutputFile,
         JSON.stringify(compilationOutput),
-        {encoding: "utf8"}
+        { encoding: "utf8" }
       );
     }
 
@@ -96,7 +127,7 @@ const command = {
   },
 
   listVersions: async function (options) {
-    const {CompilerSupplier} = require("@truffle/compile-solidity");
+    const { CompilerSupplier } = require("@truffle/compile-solidity");
     const supplier = new CompilerSupplier({
       solcConfig: options.compilers.solc,
       events: options.events
--- a/packages/core/test/lib/commands/compile.js
+++ b/packages/core/test/lib/commands/compile.js
@@ -102,6 +102,36 @@ describe("compile", function () {
     );
   });
 
+  it("compiles one specified contract after three are updated", async function () {
+    this.timeout(10000);
+
+    const reset = await updateSources([
+      "ConvertLib.sol",
+      "MetaCoin.sol",
+      "Migrations.sol"
+    ]);
+
+    try {
+      const { contracts } = await WorkflowCompile.compileAndSave(
+        config.with({
+          all: false,
+          quiet: true,
+          specificFiles: [
+            path.resolve(config.contracts_directory, "ConvertLib.sol")
+          ]
+        })
+      );
+
+      assert.equal(
+        Object.keys(contracts).length,
+        2, //ConvertLib.sol is imported by MetaCoin.sol so there should be two files.
+        "Didn't compile specified contracts."
+      );
+    } finally {
+      reset();
+    }
+  });
+
   it("compiles updated contract and its ancestors", async function () {
     this.timeout(10000);
 
--- a/packages/core/test/lib/commands/compile.js
+++ b/packages/core/test/lib/commands/compile.js
@@ -8,26 +8,47 @@ const command = require("../../../lib/commands/compile");
 const path = require("path");
 const fs = require("fs-extra");
 const glob = require("glob");
-let config;
-let output = "";
-let memStream;
 
-function updateFile(filename, _stat = undefined) {
-  var file_to_update = path.resolve(
-    path.join(config.contracts_directory, filename)
+const makeResolveSource = config => async contractSourcePath => {
+  const { filePath } = await config.resolver.resolve(
+    contractSourcePath,
+
+    // to pretend we're inside the contracts directory
+    path.join(config.contracts_directory, "_")
   );
-  if (_stat) {
-    fs.utimesSync(file_to_update, _stat.atime, _stat.mtime);
-    return;
-  } else {
-    var stat = fs.statSync(file_to_update);
-    var newTime = new Date().getTime();
-    fs.utimesSync(file_to_update, newTime, newTime);
-    return stat;
-  }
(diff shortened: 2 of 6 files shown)


packages/compile-vyper/index.js
```diff
@@ -53,6 +53,25 @@ compile.necessary = function(options, callback) {
...
+// Function called when specific files to compile were defined.
+// It checks if files exist and if they have correct .vy extension.
+// Then it will call compile.with_dependencies.
+
+compile.specific = function(options, callback) {
+  options.logger = options.logger || console;
+  var files = [];
+  options.specificFiles.map(element => {
+    let filename = path.join(options.working_directory, element);
+    if (path.extname(filename) == ".vy") files.push(filename);
```
robertmagier: Should I think about some other extensions for vyper compilations ?

gnidan:
@robertmagier thank you for opening this. I've sat down to review this a couple times and I can imagine the wrangling through code you went through.

This raises an interesting question about where/how this should be implemented. Things are kind of a mess, with competing mechanisms (e.g. `paths` vs. `files` options). I will write up some documentation, hopefully over the next few days, and we can use that to discuss how to approach this addition.

Thanks again!

robertmagier:
@gnidan No problem. It was  kind of satisfying. I think I put more effort into #2729. This one wasn't that difficult. 
What kind of documentation do you want to write  ? Maybe I can help here ?

gnidan:
> What kind of documentation do you want to write?

Mostly, I want to have an at-a-glance view of the current interfaces for compile-solidity (legacy+new), as well as the interfaces for workflow-compile (legacy+new), with attention paid to the existing `files` and `paths` options. If we have that, I think it'll be easier to communicate what the interface change should be here.

I'll also get back to #2729, thank you for the reminder :)

robertmagier:
Do you have some existing document I can look at to see what you mean ? Do you want to make some diagram or just describe it somewhere ?

gnidan:
Unfortunately there's no existing document, only the code :(

A diagram would be useful, but I was even just thinking a couple paragraphs and bulleted lists would help facilitate discussion.

gnidan:
Ah but this might be useful, the PR for the new compile interface: https://github.com/trufflesuite/truffle/pull/2073

robertmagier:
@gnidan 

The way how options.paths and options.files are implemented doesn't say in clear way what was the idea behind it. 
- options.files are used in Profiler in getFiles function in Profiler.updated function. 

```
  async updated(options, callback) {
    const callbackPassed = typeof callback === "function";
    expect.options(options, ["resolver"]);

    const { contracts_directory, contracts_build_directory } = options;

    async function getFiles() {
      if (options.files) {
        return options.files;
      } else {
        return findContracts(contracts_directory);
      }
    }

    let sourceFilesArtifacts = {};
    let sourceFilesArtifactsUpdatedTimes = {};

    try {
      const sourceFiles = await getFiles();

```
If option.files.length === 0 then we will process all contracts from contracts directories.
If it !=0 then we will use options.files to create list of files which should be compiled. However, options.files is used again in 

```
compile.necessary = async function(options) {
  options.logger = options.logger || console;

  const paths = [
    ...new Set([
      ...(await promisify(Profiler.updated)(options)),
      ...(options.files || [])
    ])
  ];

```

So, we check options.files in Profiler.updated, but then we use it again in compile.necessary after we call Profiler.updated. It is so wrong. 

Workflow-compile doesn't do anything with options.paths nor options.files

So this is how it should be:

- Set options.files = options._ for compile command.
- Create new function compile.specific to be called when --specific flag is true
- compile.specific will call compiler.with_dependencies with options.path equal to options.files
- compile.all will call compile.with_dependencies with options.path equal to option.files plus result from findContratcs
- compile.necessary will call compile.with_dependencies with options.path equal to option.files plus result from Profiler.updated
- compile.with_dependencies calls Profiler.required_sources with options.paths only. 
- options.files to be used as source of extra files to be compiled.
- options.paths to be used only in compile.with_dependencies to generate list of sources to be send to compiler. 
- Profiler.updated does not use options.paths nor options.files
- Profiler.required use options.paths only
- options.paths is updated in compile.all, compile.necessary or compile.specific only. 

In other words compile-solidity and compile-vyper will use options.files to extend list of files to be compiled depending on value of flags --all and --specific. options.paths is used only internally in compile-solidity and compile-vyper

eggplantzzz:
We should reassess this after the merge of PR #3252

gnidan:
I rebased this and made some edits to accommodate changes to `develop` since this got opened.

I also reworked a couple things, including the `updateFile` helper.

Here's how this differs from @robertmagier's original implementation:
- Don't define `compile.specific`; instead, use `Compile.sourcesWithDependencies` (which I've added to the new compile-common interface)
- Update `config.specificPaths` to use absolute paths; convert relative paths to absolute paths in `core/lib/commands/compile`
- Separate fsstat stuff into own commit (and then revise)

Marking this as draft, since the two commits at the beginning should get to `develop` first (once reviewed)

Will add a dedicated review to highlight my changes, so they can be reviewed separately.

packages/core/test/lib/commands/compile.js
```diff
@@ -8,11 +8,47 @@ const command = require("../../../lib/commands/compile");
...
+      fs.utimesSync(filePath, newTime, newTime);
+    }
+
+    return () => {
+      for (const [filePath, stat] of Object.entries(stats)) {
+        fs.utimesSync(filePath, stat.atime, stat.mtime);
+      }
+    };
+  };
+};
```
gnidan, line 44: If this is worth keeping, we'll probably want to put it into a separate file? It's kinda ugly.

packages/compile-common/src/types.ts
```diff
@@ -54,12 +54,22 @@ export interface WorkflowCompileResult {
...
   }) => Promise<CompilerResult>;
+
+  sourcesWithDependencies: ({
+    paths,
+    options
+  }: {
+    paths: string[];
+    options: object;
+  }) => Promise<CompilerResult>;
 }
```
gnidan, line 75: This change should be a separate PR

packages/compile-vyper/index.js
```diff
@@ -119,7 +119,7 @@ const Compile = {
     const vyperFiles = sources.filter(path => minimatch(path, VYPER_PATTERN));
 
     // no vyper files found, no need to check vyper
-    if (sources.length === 0) {
+    if (vyperFiles.length === 0) {
```
gnidan, line 122: Unrelated to the work in this PR, turns out this check was just wrong. (Also should be own PR)

packages/core/lib/commands/compile.js
```diff
@@ -67,6 +72,10 @@ const command = {
     const Config = require("@truffle/config");
     const config = Config.detect(options);
 
+    config.specificFiles = (options._ || []).map(sourcePath =>
+      path.resolve(config.working_directory, sourcePath)
```
gnidan, line 76: (I modified the implementation to do this resolution here. @robertmagier originally had these in a dedicated `compile.specific` method in each of the packages)

gnidan (commented):
Here are notes on what code is mine that deviates from @robertmagier's original approach.

gnidan:
@eggplantzzz you'll see in my rebase here that I'm using `Compile.sourcesForDependencies` - maybe this is a case for adding it to the compile-common interface already? Let me know what you think and I'll split that out into its own PR

packages/compile-vyper/index.js
gnidan: (Opened #3384 for this change)

gnidan:
Alright I rebased one change out and made a separate PR - that should be pretty quick. Once that's in, I'll do the same for the more substantial change of adding `Compile.sourcesWithDependencies` to the compile-common interface.

packages/workflow-compile/index.js
```diff
@@ -33,6 +33,12 @@ async function compile(config) {
       const compileMethod =
         config.all === true || config.compileAll === true
           ? Compile.all
+          : config.specificFiles
+          ? config =>
+              Compile.sourcesWithDependencies({
+                paths: config.specificFiles,
+                options: config
+              })
```
gnidan, line 41: I may have broken this. Need to test that `truffle compile <vyper-contract>` doesn't cause compile-solidity to explode.

packages/compile-vyper/index.js
gnidan: I think it's safe to leave this sort of requirement out until someone wants to use a different file extension.

gnidan:
Oops, did not realize deleting the base branch would close this.

packages/compile-common/src/types.ts
gnidan: (Moved this to #3385)

eggplantzzz:
This was implemented in https://github.com/trufflesuite/truffle/pull/5079.

PR #2715, opened 2020-01-04
Name resource

fainashalts:
This PR creates a NameRecord resource and a Project resource that holds information about the head of the linked list of name records.


packages/truffle-db/src/loaders/artifacts/index.ts
```diff
@@ -112,19 +138,51 @@ export class ArtifactsLoader {
...
+    let previousIndex = currentNameHeads.findIndex(
+      head => type == head.type && name.toLowerCase() == head.name
+    );
+
+    if (previousIndex !== -1) {
+      previous = { id: currentNameHeads[previousIndex].id };
+    } else {
+      previous = null;
+    }
+    return previous;
```
gnidan: I believe this whole function can just be replaced with:

```typescript
return currentNameHeads.find(
  head => type === head.type && name.toLowerCase() === head.name.toLowerCase()
);
```

(also note the triple-equals and the `toLowerCase` on both sides, just for good measure)

generally best to avoid relying on array indexes whenever possible.

PR #2717, opened 2020-01-04
Ignore persisted DB in Truffle DB tests

gnidan:
In the .gitignore!


gnidan:
(I can't approve this because it's my PR, but it's @cds-amal's commit, which I approve)

PR #2720, opened 2020-01-05
[wip] Hook up Truffle DB ingest inside Truffle

gnidan:
[wip, waiting on #2746]

- Add `db: { enabled: true }` feature flag to truffle-config
- Upgrade typescript to ^3.6.3
- Start defining generator-based loaders for separation of concerns (includes sources+compilations currently)
- Move relevant loading outside the artifacts loader and into this generator-based system.

PR #2721, opened 2020-01-05
Replace Truffle DB's Contract.constructor with bytecode fields

gnidan:
Currently, the Contract object defines a `constructor` that was the same as ContractInstance's `creation.constructor`, which specified linked binary. Since ContractInstances have a network, this linked binary is available in that context (we can determine link values given a network), and so that's fine.

But, in the context of #2720, which tries to add contracts during compilation, when networks are not available, this proves impossible. We can't get link values, so we can't save linked bytecodes. This PR posits that Contract should just have unlinked bytecode fields (`createBytecode` and `callBytecode`) instead of a Constructor data object.

Changes:
- Remove all notion of contracts having a constructor
- Add createBytecode and callBytecode fields to contract (as regular,
  unlinked bytecode objects)
- Update schemas for artifacts, workspace, loaders, and the root schema
- Update mutations and queries and such
- Update tests that check contract constructor to check bytecodes instead

### DIFF
--- a/packages/truffle-db/src/artifacts/schema.ts
+++ b/packages/truffle-db/src/artifacts/schema.ts
@@ -141,23 +141,17 @@ export const schema = mergeSchemas({
           return sourceContract;
         }
       },
-      constructor: {
+      createBytecode: {
         fragment: `... on ContractObject {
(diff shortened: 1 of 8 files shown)


gnidan:
Thanks @fainashalts!

Issue #2722, opened 2020-01-05
Truffle doesn't display compilation warnings


mrwillis:
- [x ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Compilation warnings do not print

## Steps to Reproduce

Create a project with compilation warnings 

## Expected Behavior

Get compilation warnings printed out

## Actual Results

`truffle compile --all`

> ...
> > Compiling ./contracts/libraries/LibOrderAmounts.sol
> > Compiling ./contracts/libraries/LibOutcome.sol
> > Compiling ./contracts/libraries/LibString.sol
> > Compiling openzeppelin-solidity/contracts/access/Roles.sol
> > Compiling openzeppelin-solidity/contracts/cryptography/ECDSA.sol
> > Compiling openzeppelin-solidity/contracts/math/SafeMath.sol
> > Compiling openzeppelin-solidity/contracts/token/ERC20/IERC20.sol
> > Compilation warnings encountered:
> ,,,,,,,,,,,,,,,,,,,,,,,,,,,,

## Environment

* Operating System: 
* Ethereum client: Ganache CLI v6.7.0-beta.0 (ganache-core: 2.8.0-beta.0)
* Truffle version (`truffle version`): 5.1.7
* node version (`node --version`): 10.5.0
* npm version (`npm --version`): 6.13.4

haltman-at:
I'm getting this too now, quote marks are being printed instead of compilation warnings.  Not sure which particular release introduced this, but it was fine before we went on break...

eggplantzzz:
There is a fix for this in https://github.com/trufflesuite/truffle/pull/2725

haltman-at:
Oops, seems we never closed this.  (This was released in v5.1.8).  Closing!

PR #2725, opened 2020-01-07
Bug fix: Edit compile subscriber output

eggplantzzz:
#2722 

Adjust subscriber output to correctly display warning output.

Previously it looks like the raw warning output from the Solidity compiler was getting passed to the subscriber. Currently only the formatted message is and so the handler function was not adjusted to accept this.

### DIFF
--- a/packages/events/defaultSubscribers/compile.js
+++ b/packages/events/defaultSubscribers/compile.js
@@ -43,9 +43,7 @@ module.exports = {
     "compile:warnings": [
       function({ warnings }) {
         this.logger.log("> Compilation warnings encountered:");
-        this.logger.log(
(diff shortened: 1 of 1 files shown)


packages/events/defaultSubscribers/compile.js
```diff
@@ -43,9 +43,7 @@ module.exports = {
     "compile:warnings": [
       function({ warnings }) {
         this.logger.log("> Compilation warnings encountered:");
-        this.logger.log(
-          warnings.map(warning => warning.formattedMessage).join()
-        );
+        this.logger.log(`${OS.EOL}    ${warnings.join()}`);
```
gnidan: That'll do it

PR #2727, opened 2020-01-07
Add missing return statement to fix snapshot bug

### LINKED ISSUE #2710: Test Isolation in Truffle 5.1.5
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

I think truffle 5.1.5 broke test file isolation.

The same contract addresses seem to be used across multiple test files when running `$(npm bin)/truffle test`. From a deeper investigation, it looks like a fresh migration is occurring, but the addresses that are injected into the environment aren't being updated from test to test.

## Steps to Reproduce

See https://github.com/mrice32/truffle_example for a simple example (varies only slightly from `truffle init`) of the test isolation issue.

```
git clone https://github.com/mrice32/truffle_example
cd truffle_example
npm install
$(npm bin)/truffle test
```

## Expected Behavior

Tests should pass and print different addresses.

## Actual Results

Output for `$(npm bin)/truffle test` in the above repository.
```

Compiling your contracts...
===========================
> Compiling ./contracts/Migrations.sol



  Contract: 1
0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0
    ✓ Check and Set value (69ms)

  Contract: 2
0x8CdaF0CD259887258Bc13a92C0a6dA92698644C0
    1) Check and Set value
    > No events were emitted


  1 passing (271ms)
  1 failing

  1) Contract: 2
       Check and Set value:

      AssertionError: expected '1' to equal '0'
      + expected - actual

      -1
      +0
      
      at Context.<anonymous> (test/2.js:7:12)
      at processTicksAndRejections (internal/process/task_queues.js:86:5)
```

## Environment

* Operating System: Mac OS X 10.15
* Ethereum client: Default truffle test client
* Truffle version (`truffle version`): Truffle v5.1.5
* node version (`node --version`): v11.15.0
* npm version (`npm --version`): 6.13.4

haltman-at:
This PR fixes #2710... sort of.  The underlying bug that got exposed remains, of course, but this fixes the immediate cause of the problem and gets things back to how they were before.

So, what went wrong here is that I accidentally broke the test runner's use of snapshots, forcing it to rely on the buggy redeploy-based fallback.  How did I break it?  Well, uh, the test runner's `rpc` method, that performs RPC calls... I left out the `return` statement.  Oops.  With the `return` added in, snapshotting works just fine and the problem goes away.

Once this is released, I'll open an issue for the underlying redeploy bug that got exposed.  I have no idea what to do about that one.

### DIFF
--- a/packages/core/lib/testing/testrunner.js
+++ b/packages/core/lib/testing/testrunner.js
@@ -53,11 +53,15 @@ TestRunner.prototype.initialize = async function() {
   );
 
   if (this.first_snapshot) {
+    debug("taking first snapshot");
     try {
       let initial_snapshot = await this.snapshot();
       this.can_snapshot = true;
       this.initial_snapshot = initial_snapshot;
-    } catch (_) {}
+    } catch (error) {
+      debug("first snapshot failed");
+      debug("Error: %O", error);
+    }
     this.first_snapshot = false;
   } else {
     await this.resetState();
@@ -91,9 +95,11 @@ TestRunner.prototype.deploy = async function() {
 
 TestRunner.prototype.resetState = async function() {
   if (this.can_snapshot) {
+    debug("reverting...");
     await this.revert(this.initial_snapshot);
     this.initial_snapshot = await this.snapshot();
   } else {
+    debug("redeploying...");
     await this.deploy();
   }
 };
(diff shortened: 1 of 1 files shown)


haltman-at:
> What do you think about adding a test case for #2710?

Hm, it would make sense.  I don't really know the tests for that area of the code but I can take a look at them...

haltman-at:
Alright, skipping adding a test for now -- the `truffle test` tests are a bit confusing -- and just merging.

PR #2730, opened 2020-01-08
Dependency Update: use v5.2.1 mocha fork for node 12/LTS support

### LINKED ISSUE #2070: Not usable on Node.js 12

CruzMolina:
Aims to resolve #2070.

### DIFF
--- a/packages/truffle/package.json
+++ b/packages/truffle/package.json
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "app-module-path": "^2.2.0",
-    "mocha": "5.2.0",
+    "mocha": "https://github.com/trufflesuite/mocha.git#v5.2.1",
     "original-require": "1.0.1"
   },
   "devDependencies": {
--- a/packages/truffle/package.json
+++ b/packages/truffle/package.json
@@ -7,7 +7,7 @@
   "bugs": {
     "url": "https://github.com/trufflesuite/truffle/issues"
   },
-  "version": "5.1.8",
+  "version": "5.1.9-nodeLTS.0",
   "main": "./build/library.bundled.js",
   "bin": {
     "truffle": "./build/cli.bundled.js"
--- a/yarn.lock
+++ b/yarn.lock
@@ -10582,6 +10582,22 @@ mocha@^6.0.2, mocha@^6.1.4:
     yargs-parser "13.1.1"
     yargs-unparser "1.6.0"
 
+"mocha@https://github.com/trufflesuite/mocha.git#v5.2.1":
+  version "5.2.0"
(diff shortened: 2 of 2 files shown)


packages/truffle/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "dependencies": {
     "app-module-path": "^2.2.0",
-    "mocha": "5.2.0",
+    "mocha": "https://github.com/trufflesuite/mocha.git#v5.2.1",
```
gnidan: This will force users to have git to install Truffle.

Also, don't do this :)

CruzMolina: Good point. `@truffle/mocha` coming right up. 👏

gnidan: No but please don't do it this way 🙌

CruzMolina:
Closing since there doesn't seem to be interest in merging this into truffle proper.

packages/truffle/package.json
fainashalts: @CruzMolina did can this just say "5.2.1" instead of "5.2.0" or was there a reason you did it this way?

CruzMolina: @fainashalts Currently `5.2.1` only exists as a branch on our fork. Want me to publish it as `@trufflesuite/mocha`?

PR #2746, opened 2020-01-10
Add sqlite adapter for Truffle DB

gnidan:
Add another backend storage adapter for Truffle DB, to use pouchdb-adapter-node-websql to write to local sqlite3 databases.

### DIFF
--- a/packages/truffle-db/src/db.ts
+++ b/packages/truffle-db/src/db.ts
@@ -8,6 +8,12 @@ interface IConfig {
   contracts_build_directory: string;
   contracts_directory: string;
   working_directory?: string;
+  db?: {
+    adapter?: {
+      name: string;
+      settings?: any;
+    };
+  };
 }
 
 interface IContext {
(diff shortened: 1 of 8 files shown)


packages/truffle-db/src/db.ts
```diff
@@ -8,6 +8,12 @@ interface IConfig {
   contracts_build_directory: string;
   contracts_directory: string;
   working_directory?: string;
+  db?: {
+    adapter?: {
```
fainashalts: This may be a silly question but why the `?` here? We'll always have _some_ adapter, right? I assume we'd just get the default if this is empty but just putting this note here in case I misunderstood.

gnidan: Not a silly question! I made this allow empty because I wanted to keep Truffle DB's adapter defaults an internal concern for now. I can see reasons for/against moving the defaulting to @truffle/config. Happy to discuss or ignore for now :)

gnidan:
Thanks @fainashalts!

PR #2750, opened 2020-01-13
Reduce Truffle DB's workspace Databases to single type parameter

gnidan:
(Waiting on #2746)

Previously, Databases used a bunch of different types that came from different places. This PR unifies the generics to take a single param.

- Move Workspace-specific types to definitions module, to simplify adding new resources
- Move abstract Databases class to own file, separate from type-level business


gnidan:
Thanks @fainashalts !

PR #2758, opened 2020-01-14
Upgrade dependency: ganache-core@2.9.1

CruzMolina:
* @truffle/artifactor: 2.8.0 →  2.9.1
* @truffle/contract: 2.8.0 →  2.9.1
* @truffle/core: 2.8.0 →  2.9.1
* @truffle/debugger: 2.8.0 →  2.9.1
* @truffle/deployer: 2.8.0 →  2.9.1
* @truffle/environment: 2.8.0 →  2.9.1
* @truffle/hdwallet-provider: 2.8.0 →  2.9.1
* @truffle/interface-adapter: 2.8.0 →  2.9.1
* @truffle/provider: 2.8.0 →  2.9.1
* truffle: 2.8.0 →  2.9.1

### DIFF
--- a/packages/contract/test/deploy.js
+++ b/packages/contract/test/deploy.js
@@ -214,7 +214,6 @@ describe("Deployments", function() {
       Example.autoGas = true;
     });
 
-    // Constructor in this test consumes ~6437823 (ganache) vs blockLimit of 6721975.
     it("should not multiply past the blockLimit", async function() {
       this.timeout(50000);
       let iterations = 1000; // # of times to set a uint in a loop, consuming gas.
@@ -225,8 +224,8 @@ describe("Deployments", function() {
 
       assert(multiplier === 1.25, "Multiplier should be initialized to 1.25");
       assert(
-        multiplier * estimate > block.gasLimit,
-        "Multiplied estimate should be too high"
+        multiplier * estimate < block.gasLimit,
+        "Multiplied estimate should not be higher than the blockLimit"
       );
       assert(estimate < block.gasLimit, "Estimate on it's own should be ok");
 
--- a/packages/deployer/test/errors.js
+++ b/packages/deployer/test/errors.js
@@ -181,7 +181,7 @@ describe("Error cases", function() {
   });
 
   it("OOG (w/ estimate, hits block limit)", async function() {
-    this.timeout(30000);
+    this.timeout(80000);
 
     const migrate = function() {
       deployer.deploy(Loops);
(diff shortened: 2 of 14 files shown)


packages/deployer/test/errors.js
```diff
@@ -213,8 +213,8 @@ describe("Error cases", function() {
       assert.fail();
     } catch (err) {
       assert(err.message.includes("Loops"));
-      assert(err.message.includes("code couldn't be stored"));
-      assert(err.message.includes("check your gas limit"));
+      assert(err.message.includes("ran out of gas"));
+      assert(err.message.includes("network config or deployment parameters"));
```
haltman-at: I assume this is just changes in error messages?

packages/contract/test/deploy.js
```diff
@@ -225,8 +224,8 @@ describe("Deployments", function() {
 
       assert(multiplier === 1.25, "Multiplier should be initialized to 1.25");
       assert(
-        multiplier * estimate > block.gasLimit,
-        "Multiplied estimate should be too high"
+        multiplier * estimate < block.gasLimit,
```
haltman-at: Huh, what's the reason for this change?

packages/deployer/test/errors.js
CruzMolina: there was a regression that's been fixed in this release. I'll change this commit to use the original (now fixed in this release) output we were looking for https://github.com/trufflesuite/truffle/blame/be86b8eed120af9dca7d2ea1ee751a54ded200fc/packages/deployer/test/errors.js#L215-L218.

haltman-at: Oh yeah that whole thing... cool.

packages/contract/test/deploy.js
CruzMolina: This test is/was confusing because the `it` statement says the multiplied estimate *should* be less than the `blockLimit` and we were previously asserting that it *was* higher than the `blockLimit`.

After chatting with @nicholasjpaterno about this test case, it appears the gas cost for creating the `Example` contract got much cheaper (by 2/3rds) due to a dirty storage slot bug fix in `istanbul` (which ganache now defaults to). See [EIP 2200](https://eips.ethereum.org/EIPS/eip-2200) & [EIP 2028](https://eips.ethereum.org/EIPS/eip-2028).

Thus, it made sense to reverse the sign since this appears to be in line with the original intent of this specific test.

CruzMolina:
no wait!

CruzMolina:
need to pass the `hardFork` option, since the default changed. been pairing with @nicholasjpaterno, there be bugs in the `istanbul` logic.

PR #2761, opened 2020-01-15
Internal improvement: Unskip globally-available variables tests

haltman-at:
Now that we're on ganache-core 2.9.1, where the timestamp bug is fixed, we can finally turn on the globally-available variables tests and not have them fail due to timestamp problems!  Hooray!


haltman-at:
Ah damn, looks like it...

haltman-at:
(Holding off on merging this until ganache-core 2.9.1 is replaced by 2.9.2.)

Issue #2733, opened 2020-01-08
Truffle 5.1.5 and later does not handle Solidity require fail correctly in test


cfelde:
## Issue

When running a test that expects a require(..) condition to fail, the require failure is not handled correctly in Truffle v5.1.5 and later. This works correctly in v5.1.4 and earlier.

## Steps to Reproduce

Please see bug POC at https://github.com/cfelde/truffle-requirebug

## Expected Behavior

Test should pass.

## Actual Results

Test does not pass.

## Environment

* Operating System: Mac

Truffle v5.1.5 (core: 5.1.5)
Solidity v0.5.12 (solc-js)
Node v13.3.0
Web3.js v1.2.1

fainashalts:
Hi @cfelde, it looks like this is due to an accidental breaking change in 5.1.5, as of #2691. We will look into a fix for this, thanks for bringing it up!

gnidan:
Oops, sorry!

haltman-at:
This is fixed now, right?

gnidan:
@haltman-at you're correct. Process dictates this should've been closed. Closing it now!

Y'all tell me if I'm wrong; happy to reopen :)

PR #2762, opened 2020-01-15
Remove nonexistent setting that wouldn't work

gnidan:
Ganache neither defines `noVMErrorsOnRPCResponse`, nor can Truffle make use of the intended option `vmErrorsOnRPCResponse: false`, since that behavior would be a breaking change (identified thanks to #2733)

### DIFF
--- a/packages/core/lib/commands/test/index.js
+++ b/packages/core/lib/commands/test/index.js
@@ -132,7 +132,6 @@ const command = {
         mnemonic:
           "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat",
         gasLimit: config.gas,
-        noVMErrorsOnRPCResponse: true,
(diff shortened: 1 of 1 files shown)


gnidan:
Wait don't tell me that option does something after all

gnidan:
Is @haltman-at doing the time increase in #2761? Otherwise maybe it should just be a separate PR xor part of the 2.9.2 PR

haltman-at:
I only increased times on the ones I unskipped in that PR.  I didn't increase times on anything else.

eggplantzzz:
Is this still moving forward?

gnidan:
> Is this still moving forward?

It's currently stalled - just haven't gotten to doing the rebase here.

gnidan:
Thanks all!

PR #2765, opened 2020-01-16
Track collection of current names per project in Truffle DB

gnidan:
- Expand workspace databases with support for mutable collections
- Define projectNames mutable collection
- Add queries to resolve current name
- Add mutation to assign current names
- Ensure loader assigns names based on previously current names

Thanks to @fainashalts for much of this work :)

### DIFF
--- a/packages/truffle-db/src/loaders/artifacts/index.ts
+++ b/packages/truffle-db/src/loaders/artifacts/index.ts
@@ -15,7 +15,10 @@ import {
   AddContracts,
   AddContractInstances,
   AddNameRecords,
-  AddNetworks
+  AddNetworks,
+  AddProjects,
+  GetCurrent,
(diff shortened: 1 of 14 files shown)


packages/truffle-db/src/loaders/artifacts/index.ts
```diff
@@ -128,11 +138,56 @@ export class ArtifactsLoader {
...
+      })
+    );
+
+    //set new projectNameHeads based on name records added
+    const projectNamesResult = await this.db.query(AssignProjectNames, {
+      projectNames
+    });
+  }
+
+  async getCurrent(projectId: string, type: string, name: string) {
```
cds-amal: A small nit, perhaps renaming to `getCurrentProject` for clarity?

PR #2766, opened 2020-01-16
Dependency Upgrade: ganache-core@2.9.2

CruzMolina:
This PR brings truffle's defaults up to ~latest Ethereum mainnet~ Istanbul (Muir Glaicier is latest).

Truffle will now spin up `ganache-core` using the `istanbul` hardfork by default.
Truffle will also postinstall & invoke `solc@0.5.16` which targets `istanbul` by default.

Also upgrades a `truffle@contract-schema`'s `solc` devDep to 0.5.16.

### DIFF
--- a/packages/artifactor/package.json
+++ b/packages/artifactor/package.json
@@ -24,7 +24,7 @@
     "@types/node": "^12.6.2",
     "chai": "4.2.0",
     "debug": "^4.1.1",
-    "ganache-core": "2.8.0",
+    "ganache-core": "2.9.2",
     "mocha": "5.2.0",
     "require-nocache": "^1.0.0",
     "temp": "^0.8.3",
--- a/packages/contract/test/deploy.js
+++ b/packages/contract/test/deploy.js
@@ -214,9 +214,10 @@ describe("Deployments", function() {
       Example.autoGas = true;
     });
 
+    // Constructor in this test consumes ~6388773 (ganache) vs blockLimit of 6721975.
     it("should not multiply past the blockLimit", async function() {
       this.timeout(50000);
-      let iterations = 1000; // # of times to set a uint in a loop, consuming gas.
+      let iterations = 6000; // # of times to set a uint in a loop, consuming gas.
 
       const estimate = await Example.new.estimateGas(iterations);
       const block = await web3.eth.getBlock("latest");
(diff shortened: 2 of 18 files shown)


packages/artifactor/package.json
```diff
@@ -24,7 +24,7 @@
     "@types/node": "^12.6.2",
     "chai": "4.2.0",
     "debug": "^4.1.1",
-    "ganache-core": "2.8.0",
+    "ganache-core": "2.9.2",
```
nicholasjpaterno: :+1:

scripts/ci.sh
```diff
@@ -23,6 +23,7 @@ run_geth() {
     --dev.period 0 \
     --allow-insecure-unlock \
     --targetgaslimit '7000000' \
+    --override.istanbul '0' \
```
nicholasjpaterno: So much for geth honoring their "default" :joy:

packages/contract/test/deploy.js
```diff
@@ -225,8 +224,8 @@ describe("Deployments", function() {
 
       assert(multiplier === 1.25, "Multiplier should be initialized to 1.25");
       assert(
-        multiplier * estimate > block.gasLimit,
-        "Multiplied estimate should be too high"
+        multiplier * estimate < block.gasLimit,
```
nicholasjpaterno: In theory, this shouldn't be necessary anymore especially for tests; previously ganache would underestimate in rare cases (conditional branching on gasLeft).  Obviously, in reality, deploying runs the risk of state changes during the interim between estimates and transaction dispatch/propagation/block inclusion etc. and so a multiplier makes sense on any public networks.

gnidan: This is a very weird change for an upgrade PR. It looks like the old version was just semantically incorrect?

CruzMolina: @gnidan see prior convo here: https://github.com/trufflesuite/truffle/pull/2758#discussion_r366688384

gnidan: Thanks for the context @CruzMolina!

I did some digging to figure out the [original intention](https://github.com/trufflesuite/truffle/commit/5a06ecfd7b0b50dabbdca68dcf5d20812915f44c) of the test, and I think it's this:
- Create a situation where the default multiplier results in a value that exceeds the block limit
- Proceed to deploy anyway and not have an exception

If this reasoning is correct, then I think to fix this test, we'd need to update the `Example` contract or the default multiplier so that:
- estimate < block limit
- estimate * default multiplier > block limit

i.e., we want to test that multiplier is self-capping

gnidan: Re: your point @nicholasjpaterno 
> In theory, this shouldn't be necessary anymore especially for tests

We still need the multiplier because we still want to address cases where there's a state change in between the estimate call and the tx. (Although with your gas exactimation work, we may be able to change the default from 1.25 to something lower)

gnidan:
Nice job all :)

PR #2774, opened 2020-01-22
Add missing decoder dependency in core

haltman-at:
It looks like when I put the new decoder in `truffle test`, I forgot to ever actually add it as a dependency of `core`, yikes!  This PR does that.


haltman-at:
> I'm not seeing the `yarn.lock` updated, but maybe it didn't need to be modified after adding the `decoder` to `core`.

Yup, running `yarn` simply didn't change the lockfile it seems.

PR #2776, opened 2020-01-22
Internal improvement: @truffle/compile-common

CruzMolina:


CruzMolina:
Closing since there doesn't seem to be interest in moving forward with this.

Will be reverting this code on `alphaTez` and removing `@truffle/compile-common` in favor
of WET `@truffle/compile-ligo` code to target the `tezos` branch.

PR #2790, opened 2020-01-24
Fix: Hook up new workflow-compile interface to use @truffle/events

CruzMolina:
Ensures `contracts_build_directory` & `compilerInfo` is output on successful compilations.

### DIFF
--- a/packages/workflow-compile/new/index.js
+++ b/packages/workflow-compile/new/index.js
@@ -60,30 +60,50 @@ async function compile(config) {
     {}
   );
 
+  const [compilerUsed] = Object.values(compilations)
+    .map(({ compilerInfo }) => compilerInfo)
+    .filter(compilerInfo => compilerInfo);
+
   const contracts = Object.values(compilations)
     .map(({ contracts }) => contracts)
     .reduce((a, b) => [...a, ...b], []);
 
-  return { contracts, compilations };
+  return { contracts, compilations, compilerUsed };
 }
 
 const Contracts = {
   async compile(options) {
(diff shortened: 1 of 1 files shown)


packages/workflow-compile/new/index.js
```diff
@@ -60,31 +60,51 @@ async function compile(config) {
...
+    }
     return {
       contracts,
       compilations
     };
   },
 
+  reportCompilationStarted,
+  reportCompilationFinished,
+  reportNothingToCompile,
```
gnidan: Why are these getting exported here now?

CruzMolina: Not sure, saw that they were exported [here](https://github.com/trufflesuite/truffle/blob/19bf2462e51ea609267b9a8c8bc951e9c2ce09b5/packages/workflow-compile/legacy/index.js#L113-L115).

CruzMolina: Haven't checked if we even use these methods anymore. Guess we could remove them in a separate PR.

CruzMolina: The core logic getting fixed in this PR should probably go out in this week's release.

gnidan: Agreed

CruzMolina: Oh, interesting, it looks like we export these functions in `workflow-compile/legacy` because the Truffle Library API makes them available (`truffle.contracts.reportCompilationsStarted`, etc..).

PR #2807, opened 2020-01-31
Bug fix: Use console as default logger

eggplantzzz:
Previously `config.logger` was set to a dummy method.
```
logger: { log: () => {} }
```
As this was the case, whenever `config.events.updateSubscriberOptions` was executed, this method would clobber whatever logger was instantiated for the subscribers. I believe the proper behavior here is to set this property to default to `console`.


eggplantzzz:
Yep @CruzMolina, all those other commands seem to display fine on my local.

PR #2809, opened 2020-02-02
Add gzip and timeout support when downloading solc

abcfy2:
Just add gzip and set timeout when downloading solc. It makes sence for low speed network or some network traffic trouble.

### DIFF
--- a/packages/compile-solidity/compilerSupplier/loadingStrategies/VersionRange.js
+++ b/packages/compile-solidity/compilerSupplier/loadingStrategies/VersionRange.js
@@ -104,7 +104,7 @@ class VersionRange extends LoadingStrategy {
       attemptNumber: index + 1
     });
     try {
-      const response = await request.get(url);
+      const response = await request.get(url, { gzip: true, timeout: 30000 });
       events.emit("downloadCompiler:succeed");
       this.addFileToCache(response, fileName);
       return this.compilerFromString(response);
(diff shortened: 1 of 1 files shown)


packages/compile-solidity/compilerSupplier/loadingStrategies/VersionRange.js
```diff
@@ -104,7 +104,7 @@ class VersionRange extends LoadingStrategy {
       attemptNumber: index + 1
     });
     try {
-      const response = await request.get(url);
+      const response = await request.get(url, { gzip: true, timeout: 30000 });
```
gnidan: Can we make this configurable? Something like:


```javascript
module.exports = {
  compilers: {
    solc: {
      obtain: {
        gzip: true,
        timeout: 30000
      }

      /* ... rest of solc config ... */
    }

    /* ... other compiler configs ... */
  }

  /* ... rest of config ... */
}
```

I used `obtain` as a name to match the `truffle obtain` command, but would love other ideas for naming/structure.

abcfy2: I think It‘s not necessary to make this configurable.

1. Just like a browser, `gzip` and `timeout` are enabled by default.
2. `module.exports` will not affect `npm install -g truffle`(`postinstall` script will download `solc`)
3. `solc` is a `js`, `gzip` will reduce download size from `12MB` to `2MB`

gnidan: Alright you've convinced me!

gnidan:
Let's just get one more approve and we can merge this. Thanks again!

Issue #2481, opened 2019-10-22
truffle test: console.warn does not work as expected


KBriverun:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

`truffle test` command uses an [overloaded](https://github.com/trufflesuite/truffle/blob/master/packages/core/lib/test.js#L50) `console.warn()` function. 

https://github.com/syscoin/sysethereum-contracts/blob/63be76b53b9f97f35c079c275c4b45e8eaec2af1/test/utils.js#L422  - this innocent line of code prints truffle config when one [runs tests](https://travis-ci.com/syscoin/sysethereum-contracts/builds/132864645#L1184) 

## Steps to Reproduce

Run tests in branch `validateallheaders` using: https://github.com/syscoin/sysethereum-contracts/tree/validateallheaders#running-the-tests

## Expected Behavior

Expected to see warning from bitcoinjs-lib (https://github.com/bitcoinjs/bitcoinjs-lib/blob/41bf2cd03d85cd93a40992f5b9fea0af3e107734/src/transaction_builder.js#L60)

## Actual Results

Truffle config was printed (it's very verbose)

## Environment

* Operating System: /
* Ethereum client: /
* Truffle version (`truffle version`): 5.0.41
* node version (`node --version`): v11.13.0
* npm version (`npm --version`): /

cds-amal:
> This issue can be closed since @Kenadia already addresses it in https://github.com/trufflesuite/truffle/pull/2812

PR #2812, opened 2020-02-04
Fix console.warn in truffle test

Kenadia:
Addresses #2481.

Before: Calling `console.warn()` during Truffle tests would print out the Truffle config and obscure whatever message was meant to be printed.

After: `console.warn()` works as expected.

====

repro steps:

Simple Truffle test that demonstrates the bug and the fix:

```javascript
console.warn('Test console.warn()');
describe('noop', function() { it('noop', () => {}); });
```

Run with e.g. `truffle test test/basic.js`


BrendanChou:
Since an arrow function was being used...

Line 65: arguments is referring to the arguments of the function one-scope-up. This results in a ton of info being printed whenever we're trying to run tests

PR #2826, opened 2020-02-11
Bug fix: Fix assertion library

eggplantzzz:
The Solidity assertion libraries have been broken as of Truffle 5.1.5 due to them not being compiled before initializing the decoder during testing. This PR flips the order of two lines of code that causes these libraries to be compiled before decoder initialization.

After fixing this, the test summary displayed a list of events emitted during the tests which included all of the `TestEvent`s (which is the mechanism by which the assertion libraries operate). Since this is not really information relevant to the user (in that the user does not need to know about these `TestEvent`s events being emitted during testing) they are now filtered out to avoid extra noise and confusion. The failures still display in the test summary as shown below.

Screen Shot 2020-02-11 at 10 46 10 AM

This is the result of running the test suite from a new MetaCoin box with `Assert.fail("Lets make it fail")` and `Assert.fail("bummer")` placed in the two Solidity tests.

### DIFF
--- a/packages/core/lib/testing/testrunner.js
+++ b/packages/core/lib/testing/testrunner.js
@@ -176,7 +176,7 @@ TestRunner.prototype.endTest = async function(mocha) {
   });
 
   const userDefinedEventLogs = logs.filter(log => {
-    return !log.decodings.length || log.decodings[0].abi.name !== "TestEvent";
+    return log.decodings.every(decoding => decoding.abi.name !== "TestEvent")
   });
 
   if (userDefinedEventLogs.length === 0) {
--- a/packages/core/lib/testing/testrunner.js
+++ b/packages/core/lib/testing/testrunner.js
@@ -175,11 +175,11 @@ TestRunner.prototype.endTest = async function(mocha) {
     fromBlock: this.currentTestStartBlock.toNumber()
   });
 
-  const filteredLogs = logs.filter(log => {
+  const userDefinedEventLogs = logs.filter(log => {
     return !log.decodings.length || log.decodings[0].abi.name !== "TestEvent";
   });
 
-  if (filteredLogs.length === 0) {
+  if (userDefinedEventLogs.length === 0) {
     this.logger.log("    > No events were emitted");
     return;
   }
@@ -188,7 +188,7 @@ TestRunner.prototype.endTest = async function(mocha) {
   this.logger.log("    ---------------------------");
   this.logger.log("");
 
-  for (const log of filteredLogs) {
+  for (const log of userDefinedEventLogs) {
     switch (log.decodings.length) {
       case 0:
         this.logger.log(`    Warning: Could not decode event!`);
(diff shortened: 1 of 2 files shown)


packages/core/lib/testing/testrunner.js
```diff
@@ -175,7 +175,11 @@ TestRunner.prototype.endTest = async function(mocha) {
     fromBlock: this.currentTestStartBlock.toNumber()
   });
 
-  if (logs.length === 0) {
+  const filteredLogs = logs.filter(log => {
```
gnidan: Nitpick: we could name this more meaningfully than `filteredLogs`. Maybe `userDefinedEvents`?

eggplantzzz: Sure, that seems like a good idea!

eggplantzzz:
@haltman-at, if you have a better idea for filtering out those events then please let me know! I'm not super familiar with the structure of logs etc. You are correct though, some of this stuff feels ever so slightly janky though I'm not sure how much it matters.

Again, let me know if you have a better suggestion for how to test that something is indeed a `TestEvent` and I'll be happy to implement it. And using `find` also seems like a better idea.

haltman-at:
I guess I really mean `.some`, rather than `.find`.  Like, right now you're checking the 0'th decoding to see if it's a test event, and I was like, maybe just to be sure you should check to see if there *exists* such a decoding, using `Array.prototype.some`.  (And then you wouldn't have to make a special case to exclude the case of no decodings returned.)

haltman-at:
As for checking for whether something is a `TestEvent`, if you want a more detailed check, you can see my code in `soliditytest.js`.  But, again, I don't know how much that is worth it; this is fine as-is.

packages/core/lib/testing/testrunner.js
```diff
@@ -175,7 +175,11 @@ TestRunner.prototype.endTest = async function(mocha) {
     fromBlock: this.currentTestStartBlock.toNumber()
   });
 
-  if (logs.length === 0) {
+  const userDefinedEventLogs = logs.filter(log => {
+    return !log.decodings.length || log.decodings[0].abi.name !== "TestEvent";
```
haltman-at:
```suggestion
    return log.decodings.every(decoding => decoding.abi.name !== "TestEvent")
```

(I guess this uses `every` rather than `some` now that I write it out, but this is the sort of thing I was talking about; thought maybe I should illustrate...)

eggplantzzz: Ok so this ensures that none of the decodings contain `TestEvent`s. Will this give us the behavior we want? Like will there ever be any weirdness where we want to keep certain `log` objects that may contain `TestEvent`s and other values for `decoding.abi.name`?

haltman-at: Well, I mean, I guess I can't *guarantee* that won't happen, but it would require a keccak256 collision so it seems pretty unlikely?  (Either that, or someone defining their own `TestEvent`s, in which case, well...)  I think hash collisions like that are in the realm of "things we can assume won't happen".

PR #2828, opened 2020-02-12
Make it work on Powershell/cmd, and still work in Git Bash.

tcoulter:
Two important changes: 

1. Normalize all paths, then construct VM paths from the normalized versions. Also, don't use process.env.PWD, because it's not available in cmd/Powershell. 
1. Use `spawn()` instead of `exec()` to handle escaping arguments. I didn't actually test that this was required, but as you can see in [the docs](https://nodejs.org/api/child_process.html#child_process_child_process_spawn_command_args_options), `spawn()` takes care of OS-specific escaping. Ain't nobody got time for anything less.

### DIFF
--- a/packages/compile-ligo/index.js
+++ b/packages/compile-ligo/index.js
@@ -1,5 +1,5 @@
 const path = require("path");
-const { exec } = require("child_process");
+const { exec, spawn } = require("child_process");
 const fs = require("fs");
 
 const async = require("async");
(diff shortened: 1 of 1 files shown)


tcoulter:
I need someone to test if this breaks Mac/Linux. I'm willing to bet it doesn't, and works fine, because node takes care of everything for us. 

Only thing I haven't officially tested is error handling, and how this version stacks up to the previous error handling of the `exec` version.

PR #2839, opened 2020-02-20
Internal improvement: Update copy library

eggplantzzz:
Remove some callbacks from the copy library in @truffle/core and adjust the calling code.

### DIFF
--- a/packages/core/lib/copy.js
+++ b/packages/core/lib/copy.js
@@ -17,7 +17,9 @@ const copy = function(from, to, extra_options) {
 
   return new Promise((resolve, reject) => {
     cpr(from, to, options, function(err, files) {
-      reject(err);
+      if (err) {
+        reject(err);
+      }
       var new_files = [];
 
       // Remove placeholders. Placeholders allow us to copy "empty" directories,
--- a/packages/core/lib/copy.js
+++ b/packages/core/lib/copy.js
@@ -12,7 +12,7 @@ var cpr_options = {
 // won't override individual files. If a file exists, it will
 // simply move onto the next file.
 
-var copy = function(from, to, extra_options, callback) {
+var copy = function(from, to, extra_options) {
   if (typeof extra_options === "function") {
     callback = extra_options;
     extra_options = {};
@@ -20,21 +20,24 @@ var copy = function(from, to, extra_options, callback) {
 
   var options = _.merge(_.clone(cpr_options), extra_options);
 
-  cpr(from, to, options, function(err, files) {
-    var new_files = [];
+  return new Promise((resolve, reject) => {
+    cpr(from, to, options, function(err, files) {
+      reject(err);
+      var new_files = [];
 
-    // Remove placeholders. Placeholders allow us to copy "empty" directories,
-    // but lets NPM and git not ignore them.
-    files = files || [];
-    for (var file of files) {
-      if (file.match(/.*PLACEHOLDER.*/) != null) {
-        fs.unlinkSync(file);
-        continue;
+      // Remove placeholders. Placeholders allow us to copy "empty" directories,
+      // but lets NPM and git not ignore them.
+      files = files || [];
+      for (var file of files) {
+        if (file.match(/.*PLACEHOLDER.*/) != null) {
+          fs.unlinkSync(file);
+          continue;
+        }
+        new_files.push(file);
       }
-      new_files.push(file);
-    }
 
-    callback(err, new_files);
+      resolve(new_files);
+    });
   });
 };
 
--- a/packages/core/lib/copy.js
+++ b/packages/core/lib/copy.js
@@ -1,8 +1,8 @@
-var cpr = require("cpr");
-var fs = require("fs");
-var _ = require("lodash");
+const cpr = require("cpr");
+const fs = require("fs");
+const _ = require("lodash");
 
-var cpr_options = {
+const cpr_options = {
   deleteFirst: false,
   overwrite: false,
   confirm: true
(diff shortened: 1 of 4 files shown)


packages/core/lib/create.js
```diff
@@ -17,29 +17,20 @@ var templates = {
...
-      if (typeof find === "string") {
-        find = new RegExp(find, "g");
-      }
-      return data.replace(find, replacement);
-    },
-    callback
-  );
+    return;
+  } catch (error) {
+    callback(error);
```
CruzMolina:
```suggestion
    return callback(error);
```

Wanna do this instead?

packages/core/lib/create.js
```diff
@@ -17,29 +17,20 @@ var templates = {
...
+    return;
```
CruzMolina:
```suggestion
```
And rm this.

packages/core/lib/create.js
```diff
@@ -124,7 +119,12 @@ var Create = {
       );
     }
 
-    copy.file(from, to, callback);
+    try {
+      await copy.file(from, to);
+      callback();
+    } catch (error) {
+      callback(error);
```
CruzMolina:
```suggestion
      return callback(error);
```

I think this needs to be returned.

packages/core/lib/create.js
```diff
@@ -124,7 +119,12 @@ var Create = {
       );
     }
 
-    copy.file(from, to, callback);
+    try {
+      await copy.file(from, to);
+      callback();
```
CruzMolina:
```suggestion
      return callback();
```

Same here I think.

eggplantzzz:
@CruzMolina Well, the problem there is that it is possible that the callback returns something which would cause the Promise to resolve to something. With just a `return`, nothing gets returned. I thought it was safer to use the constructions I chose. What do you think?

packages/core/lib/copy.js
```diff
@@ -12,51 +12,53 @@ var cpr_options = {
...
+  const options = _.merge(_.clone(cpr_options), extra_options);
 
-  var options = _.merge(_.clone(cpr_options), extra_options);
-
-  cpr(from, to, options, function(err, files) {
-    var new_files = [];
+  return new Promise((resolve, reject) => {
+    cpr(from, to, options, function(err, files) {
+      if (err) {
+        reject(err);
```
davidmurdoch: This needs a `return;` after the `reject(err);` so the `resolve` isn't called as well.

packages/core/lib/copy.js
```diff
@@ -12,51 +12,53 @@ var cpr_options = {
...
-  writeStream.on("finish", function() {
-    callback();
-  });
+    writeStream.on("finish", function() {
+      resolve();
+    });
 
-  readStream.pipe(writeStream);
+    readStream.pipe(writeStream);
+  });
```
davidmurdoch: Instead of overwriting `reject` you should use a pattern like (untested):

```
function handleFinish() {
  resolve();
}
function handleError(err) {
   readStream.removeListener("error", handleError);
   writeStream.removeListener("error", handleError);
   writeStream.removeListener("finish", handleFinish);
   reject(err);
}
readStream.once("error", handleError);
readStream.once("error", handleError);
writeStream.once("finish", handleFinish);

readStream.pipe(writeStream);
```

This will ensure `resolve` and `reject` can never be called multiple times.

packages/core/lib/create.js
```diff
@@ -55,7 +46,7 @@ var toUnderscoreFromCamel = function(string) {
 };
 
 var Create = {
-  contract: function(directory, name, options, callback) {
+  contract: async function(directory, name, options, callback) {
```
davidmurdoch: The `await` keyword causes this function to return a new `Promise`. It's possible `Create.contract` isn't used in a context where this could cause confusion, but if it is exposed as a public API it would be a little strange.

packages/core/lib/create.js
```diff
@@ -69,14 +60,16 @@ var Create = {
...
+      return callback(error);
+    }
 
-      replaceContents(to, templates.contract.name, name, callback);
-    });
+    replaceContents(to, templates.contract.name, name, callback);
   },
 
-  test: function(directory, name, options, callback) {
+  test: async function(directory, name, options, callback) {
```
davidmurdoch: Same as above regarding the `async` keyword.

packages/core/lib/copy.js
eggplantzzz: Good catch!

packages/core/lib/create.js
```diff
@@ -92,16 +85,18 @@ var Create = {
...
-        if (err) return callback(err);
-        replaceContents(to, templates.contract.variable, underscored, callback);
-      });
+    replaceContents(to, templates.contract.name, name, function(err) {
+      if (err) return callback(err);
+      replaceContents(to, templates.contract.variable, underscored, callback);
     });
   },
-  migration: function(directory, name, options, callback) {
+  migration: async function(directory, name, options, callback) {
```
davidmurdoch: Same as above regarding the `async` keyword.

packages/core/lib/create.js
eggplantzzz: Ooh, I missed that this is exposed as a library in this package and so it would probably need to have a dual Promise/callback interface. In version 6 of Truffle all this stuff should probably be fully redone to just use Promises. Is there any reason to keep its callback api as well?

packages/core/lib/copy.js
eggplantzzz: Yeah this seems like the right kind of patter for this.

packages/core/lib/create.js
davidmurdoch: A reason to keep the callback style in version 6 would be to avoid the annoyance of breaking changes. I prefer keeping it for one major version while also shipping a deprecation notice if that style is used (e.g., `console.warn("Callback style for {package/method name} is deprecated and will be removed in v7");`). Of course, this must makes it annoying for us as the developers of truffle.

eggplantzzz:
I think I will close this for now and try and figure out what work can be done without breaking the exposed library.

PR #2842, opened 2020-02-20
Change some schema properties to match data model better

fainashalts:
This pull request contains some small changes to the schema in order to line up better with the data model in the documentation.

### DIFF
--- a/packages/db/src/schema.graphql
+++ b/packages/db/src/schema.graphql
@@ -108,11 +108,11 @@ type Bytecode {
 }
 
 type Instruction {
-  opcode: Int!,
-  programCounter: String!,
+  opcode: String!,
+  programCounter: Int!,
   meta: InstructionMeta,
   sourceRange: SourceRange,
-  pushData: String
+  pushData: Bytes
 }
 
 type InstructionMeta {
--- a/packages/db/src/schema.graphql
+++ b/packages/db/src/schema.graphql
@@ -8,7 +8,6 @@ scalar ByteOffset
 scalar FileIndex
 
 type Project {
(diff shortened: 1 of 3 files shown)


packages/db/src/schema.graphql
```diff
@@ -77,10 +76,11 @@ type AST {
 type Network implements Named {
   name: String!
   networkId: NetworkId!
-  historicBlock: HistoricBlock!
+  historicBlock: Block!
+  fork: Network
```
gnidan: Is now the time to change this name? How bout `parentFork`?

packages/db/src/schema.graphql
```diff
@@ -108,11 +108,11 @@ type Bytecode {
 }
 
 type Instruction {
-  opcode: String!,
-  programCounter: Int!,
+  opcode: Int!,
```
gnidan: I think the data model here is wrong (it should be `String!`)

packages/db/src/schema.graphql
```diff
@@ -108,11 +108,11 @@ type Bytecode {
...
+  programCounter: String!,
```
gnidan: and this is `Int!`

packages/db/src/schema.graphql
```diff
@@ -108,11 +108,11 @@ type Bytecode {
...
   meta: InstructionMeta,
   sourceRange: SourceRange,
-  pushData: Bytes
+  pushData: String
```
gnidan: Should be `Bytes`

gnidan (approved):

PR #2851, opened 2020-02-25
Allow debugger & decoder to work with multiple compilations for use with Truffle DB

haltman-at:
OK, so, this is a **giant** PR, and I am just not going to describe most of it.  Most of it after all is just a lot of plumbing and litte details to make this all work.  I am going to focus on the outward-facing changes, and the changes I think are most significant.  (Apologies for not rebasing this even smaller than the 25 commits I squashed it down into, but I didn't really want to take the time to figure out how to squash it further.)

So, first off, **this PR contains various breaking changes to `codec`**.  I'm not even going to list them; just lots of functions have been changed, frequently to require a `compilationId`.

(Hm, note: I often just wrote `compilation` where I meant `compilationId`.  Wondering if that should be changed...)

Secondly, **this PR is a breaking change to `debugger`**, because: 1., I just straight-up reworked a lot of selectors (most significantly, `solidity.info.sources`), and 2. breakpoints now require a compilation ID.

And thirdly... this PR is kind of, technically, a breaking change to `decoder`.  Except the thing I broke was never even documented, so arguably it's not a breaking change.  (When specifying struct members in `watchMappingKey` and such, it's no longer allowed to specify them by internal ID, only by name.  Not that anyone's likely to care.)

Btw, in addition to the main point of this PR, there were also a number of little bug fixes as I noticed them.

So.  I'm now going to summarize the biggest changes to each package:

1. Codec

I'm mostly going to skip codec; it's a lot of internal stuff.  But, while I was doing this, I did fix a number of little things I hadn't previously noticed were broken, so that's nice.

However, `codec` now includes the `compilations` folder, which has the `Compilation` type and the `shimArtifacts` function.  Of note: The `Compilation` type has some compatibility fields that should never appear on a real compilation but are used by the `shimArtifacts` function since the compilation it ouputs is, well, not a real compilation.  Among these is the `unreliableSourceOrder` flag.  If this is set, it indicates that, well, the order of sources is not reliable, allowing features that require reliable source order to take note.

In the case of the debugger, well, if the flag is set, the debugger will just throw an error and refuse to operate.  However the debugger *CLI* now uses this flag to determine whether it needs to recompile.  If the flag isn't set -- which is what will happen if `shimArtifacts` finds that all the sources have different IDs -- a recompile is skipped, hooray!  If the flag is set, then, oh well, a recompile is necessary after all.

As for how this flag is used in the decoder... well, presently it isn't.  But it will be eventually!  See the end.

Oh, one other thing worth mentioning here: Contract state allocation is now separate from storage struct allocation, and the latter is (finally) based purely on types, not definitions.  (Memory struct allocation is also now based on types.)  Contract state allocation remains definition-based however.  Also all of this still lives in the `storage` folder.

Also I reused an `abi-data` type in the `storage` folder for convenience.  Hope that's OK.  If `immutable` ever hits, the `storage` folder may end up looking to `abi-data` for a fair bit more...

2. Debugger (library)

Changes to invocation: The changes to invocation are non-breaking, thankfully!  You can invoke the debugger with `artifacts` and `files` as usual, or you can use `compilations` instead.  If you do, `artifacts` and `files` will be ignored.

Changes to breakpoints: Breakpoints now have a `compilation` field where the compilation ID goes.  This field must be included.

Also, lots of selector reworks, which I'm not going to cover.

3. Decoder

Although none of the changes here were breaking (barring the minor undocumented thing above), this is where I really have the most to discuss.

First off, invocation.  The old method of invocation still works, but now, instead of a list of artifacts, you can supply a `ProjectInfo`.  A `ProjectInfo` can take a few forms -- only two at the moment but the intent is that more are coming in the future.

First off, it can contain `artifacts`; this works essentially the same as if you'd just passed the list of artifacts, except it doesn't have the convenience feature where you can omit the main artifact itself from the list.

Second off, it can contain `compilations`, which is after all the whole point.

Yeah, from a user's perspective that doesn't change very much, but like I said, it'll gain more convenient options in the future.

Also, at the request of @adrianmcli, it's now possible to omit this argument entirely, in which case it'll behave as if you'd used it the old way and passed `[]`, as a further convenience when invoking it the old way.

That's really the only outward-facing change in decoder, but I want to discuss the changes to internals somewhat.

Mostly I want to ask the question... should `ContractDecoder` -- that intermediate object inbetween `WireDecoder` and `ContractInstanceDecoder` -- still exist?  I mean, I didn't want to go and just get rid of it, hard to do that without breaking things, but it doesn't have much of a point anymore, because I took the one thing it really did and moved it to `WireDecoder` for simplicity.  Admittedly I could move it back if we're really concerned that performance will be an issue, but that possibility seems unlikely to me...

Also: Even though a `ContractDecoder` is still spawned based on an `Artifact` (again: not changing the rest of the interface here), it no longer really uses it for much; instead of going based on the artifact that was passed in, it now just uses the name, bytecode, and deployed bytecode of what was passed in to *locate* the contract in the project info (which is always converted into compilations form as the first step in making a decoder).  So, yeah, it's kind of awkward.  It also raises the possibility of spawning these things in other ways... such as maybe spawning a `ContractInstanceDecoder` based on address (and `projectInfo`) alone?  (See below.)

4. Debugger CLI

Recompiles on startup may now be skipped if the whole project was compiled at once!  I already covered this up in `codec`, though, so I won't describe it more here.

Features that this PR makes possible, but will have to wait for later:

1. Allow spawning a contract instance decoder based *only* on an address (and `projectInfo`)

Since contract decoders no longer actually work based on artifacts, the idea is that we could look up the depoyed bytecode on the blockchain, then locate the appropriate contract within the project info.  Hooray!

2. Allow passing in a `config` object in `projectInfo`

The idea is that we'd use the `config` object to find all the project's artifacts, rather than the user having to pass them in manually.  Note: a proper version of this would require `artifacts.list`, but we could still make a primitive version based on the same artifact-gathering logic we have elsewhere.

3. Allow the decoder to decode internal function pointers when `unreliableSourceOrder` isn't set

Self-explanatory.  Kind of awkward to implement as it'd mean moving a bunch of code that currently lives in the debugger to, well, somewhere else.  That'd need to be figured out.  But one thing at a time.

OK!  Hopefully I didn't forget anything big there!  This is a big PR after all...


adrianmcli:
> should ContractDecoder [...] still exist

Yeah I'd agree that it should be taken out. I've never used it anywhere and actually was at first confused that there was both ContractDecoder and ContractInstanceDecoder. But since I always just use the latter, there's no point anymore.

haltman-at:
OK, one more thing, had to go back and make in-test debugging work with the new system.  Thanks to @gnidan for reminding me about that, I'd entirely forgotten!

packages/codec/lib/storage/allocate/index.ts
```diff
@@ -69,46 +80,62 @@ export function getStorageAllocations(
       }
     }
   }
-  for (const contract of Object.values(contracts)) {
+  return allocations;
+}
+
+export function getStateAllocations(
```
gnidan: this might be worth a doc comment saying what "state" means

packages/codec/lib/storage/allocate/index.ts
```diff
@@ -243,12 +255,24 @@ function getStateVariables(contractNode: Ast.AstNode): Ast.AstNode[] {
   );
 }
 
-function allocateContract(
+function allocateContractState(
   contract: Ast.AstNode,
+  compilation: string,
```
gnidan: `compilationId`?

packages/core/lib/debug/interpreter.js
```diff
@@ -56,6 +56,9 @@ class DebugInterpreter {
     const currentSourceId = currentLocation.source
       ? currentLocation.source.id
       : null;
+    const currentCompilationId = currentLocation.source
```
gnidan: Is there any way to avoid compilation being a necessary part of the definition of a breakpoint?

packages/core/lib/debug/printer.js
```diff
@@ -106,7 +111,7 @@ class DebugPrinter {
       return;
     }
 
-    const colorizedSource = this.colorizedSources[sourceId];
+    const colorizedSource = this.colorizedSources[compilation][sourceId];
```
gnidan: `compilationId`? :)

packages/debugger/lib/ast/sagas/index.js
```diff
@@ -34,35 +50,40 @@ function* handleEnter(sourceId, node, pointer, parentId) {
...
+    Object.values(compilations).map(({ byId }) => byId)
+  ).filter(x => x);
 
   yield all(
-    Object.entries(sources)
-      .filter(([_, source]) => source.ast)
-      .map(([id, { ast }]) => call(walk, id, ast))
+    sources
+      .filter(({ ast }) => ast)
+      .map(({ ast, id, compilation }) => call(walk, compilation, id, ast))
```
gnidan: `compilationId`? :)

packages/debugger/lib/controller/reducers.js
```diff
@@ -12,6 +12,7 @@ function breakpoints(state = [], action) {
       if (
         state.filter(
           breakpoint =>
+            breakpoint.compilation === action.breakpoint.compilation &&
```
gnidan: you probably know what I'm going to say here

packages/debugger/lib/controller/sagas/index.js
```diff
@@ -92,8 +88,10 @@ function* stepNext() {
     (!upcoming ||
       !upcoming.node ||
       isDeliberatelySkippedNodeType(upcoming.node) ||
-      (upcoming.sourceRange.start == startingRange.start &&
-        upcoming.sourceRange.length == startingRange.length))
+      (upcoming.sourceRange.start === starting.sourceRange.start &&
+        upcoming.sourceRange.length === starting.sourceRange.length &&
+        upcoming.source.id === starting.source.id &&
+        upcoming.source.compilation === starting.source.compilation))
```
gnidan: `upcoming.source.compilation.id`??

packages/debugger/lib/data/actions/index.js
```diff
@@ -1,19 +1,21 @@
...
-export function scope(id, pointer, parentId, sourceId) {
+export function scope(id, pointer, parentId, sourceId, compilationId) {
   return {
     type: SCOPE,
     id,
     pointer,
     parentId,
-    sourceId
+    sourceId,
+    compilationId
```
gnidan: 😎

packages/debugger/lib/data/selectors/index.js
```diff
@@ -167,43 +168,142 @@ const data = createSelectorTree({
...
        */
       inlined: {
         /**
          * data.views.scopes.inlined (selector)
-         * see data.info.scopes for how this differs from the raw version
+         * see data.views.scopes for how this differs from the raw version
+         * NOTE: grouped by compilation
          */
-        _: createLeaf(["/info/scopes", "./raw"], (scopes, inlined) =>
+        _: createLeaf(["../_", "./raw"], (scopes, inlined) =>
```
gnidan: do you need to do `../_`? I would expect just `..` to work, but I forget the implementation.

packages/debugger/lib/solidity/reducers.js
```diff
@@ -3,31 +3,24 @@ import { combineReducers } from "redux";
 import * as actions from "./actions";
 
 const DEFAULT_SOURCES = {
-  byId: {}
+  byCompilation: {} //by compilation, then in an array
```
gnidan: should this be `byCompilationId`?

packages/decoder/lib/utils.ts
```diff
@@ -183,3 +231,13 @@ export function wrapElementaryValue(
     //fixed and ufixed are not handled for now!
   }
 }
+
+//shimBytecode handles undefined, but not strings, so let's fix that
```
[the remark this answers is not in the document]
haltman-at: OK, totally going to do that then!

packages/codec/lib/ast/import/index.ts
```diff
@@ -455,3 +476,10 @@ export function definitionToStoredType(
     }
   }
 }
+
+//I am deliberately *NOT* exporting this.
+//If you have to make a type ID, instead make the type and then
+//take its ID.
+function makeTypeId(astId: number, compilationId: string): string {
+  return compilationId + ":" + astId;
```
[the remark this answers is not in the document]
haltman-at: Sure, I can change that I guess.

packages/codec/lib/storage/allocate/index.ts
haltman-at: OK, I can add that.

packages/debugger/lib/controller/sagas/index.js
haltman-at: Well, `upcoming.source.compilationId`, but yes, point taken. :P

packages/debugger/lib/data/selectors/index.js
haltman-at: I dunno.  I ran into a bunch of problems with this, this is what ultimately worked.  You're right I didn't try just `..` though.  Can try it and see what happens.  Might want to leave as `../_` for clarity regardless though.

packages/debugger/lib/solidity/reducers.js
haltman-at: I guess?

packages/codec/lib/core.ts
```diff
@@ -33,10 +33,15 @@ const Web3Utils = require("web3-utils");
 export function* decodeVariable(
   definition: Ast.AstNode,
   pointer: Pointer.DataPointer,
-  info: Evm.EvmInfo
+  info: Evm.EvmInfo,
+  compilationId: string
```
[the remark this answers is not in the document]
haltman-at: I don't think that's feasible, no.  Basically now everything that takes an AST node requires knowing what compilation it's from.  In this case it's needed for handling of user-defined types.  You do a `definitionToType` (which requires this to handle user-defined types), then you decode with that.  I don't really see a better alternative?  Honestly I think since it's in `codec` it's fine. :P

packages/core/lib/debug/interpreter.js
haltman-at: Well, without a compilation, a source index no longer uniquely specifies the source, so I think we need it, yes.  Of course we could use source *ID*s, but:

1. (The stupid reason) Introducing source IDs into the debugger would require renaming things so it doesn't call source indexes source IDs :P
2. (The better reason) We won't have source IDs if this is a shimmed compilation.

I mean, if we know that real compilations, rather than shimmed ones, will always have source IDs, then we could use a weird hybrid solution where we use source IDs when we have them, but if we don't we use source indices, and we omit the compilation because there should only be one (shimmed) compilation in that case?  That sounds even ickier though.

But honestly I don't really see anything wrong with including compilation?  Anyone coding against the debugger library can always just do what I did and, y'know, get the current compilation ID from the appropriate selector.  Yes, that's is a breaking change, but given all the selector changes this is a breaking PR anyway.  And doing it the other way -- based on source IDs rather than indexes, were we to add those -- would be breaking too.

packages/decoder/lib/utils.ts
```diff
@@ -24,28 +29,81 @@ export function nativizeDecoderVariables(
...
+    undefined
   );
 }
 
+/*
+ * extract the primary source from a source map
+ * (i.e., the source for the first instruction, found
+ * between the second and third colons)
+ * (this is something of a HACK)
+ * NOTE: duplicated from debugger, sorry
```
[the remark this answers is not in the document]
haltman-at: Hm I guess it could go in `codec`... I could put it in `compilations`, I guess?  Or perhaps `ast`?  But I'm leaning `compilations`.

packages/codec/lib/ast/import/index.ts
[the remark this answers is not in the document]
gnidan: Not required, just letting you know :) (although you could just hit the "commit suggestion" button and save yourself the trouble!)

packages/codec/lib/core.ts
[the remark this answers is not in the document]
gnidan: Yeah okee!

packages/debugger/lib/data/selectors/index.js
gnidan: yeah you're probably right!

packages/debugger/lib/controller/sagas/index.js
gnidan: ah right because `id` here means "index"? I was just commenting on the asymmetry

packages/decoder/lib/utils.ts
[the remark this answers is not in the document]
gnidan: `compilations` seems reasonable!

packages/core/lib/debug/interpreter.js
gnidan: > then we could use a weird hybrid solution where we use source IDs when we have them, but if we don't we use source indices

Could we just use source indices as source IDs in that situation?

haltman-at: Yes, I guess we could do that.  Of course then I'm going to have to go through and make source ID and source index separate concepts in the debugger. :P

Honestly I think this is OK as it is, but I guess it would be nicer to have just a single source ID instead of a pair (compilation, index).

haltman-at: Oh, here's a thought: What if we put source IDs in the debugger, but we gave them a different name, like `uniqueId` or something?  (To distinguish from the ordinary, non-unique IDs, i.e., indices...)  Yes, that's terrible, but then I wouldn't have to sort through everywhere a source index is used and change it...

(This would still probably be breaking though.  I think the only way to keep `compilation` out in a fully non-breaking way would be to do things the hard way. :-/ )

haltman-at: OK, talking things over with @gnidan, I'm leaving this as-is for now...

PR #2857, opened 2020-02-26
Support ethpm V2

njgheorghita:


packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
-
-    expect.one(options.ethpm, ["provider", "install_provider_uri"]);
-
-    // ipfs_port and ipfs_protocol are optinal.
+  install: async options => {
+    expect.options(options.ethpm, ["ipfs_host", "ipfs_port", "ipfs_protocol"]);
+    expect.options(options, ["working_directory", "ethpm_uri", "logger"]);
 
+    // Create a web3 instance connected to a blockchain
+    if (!options.provider && !options.ethpm.install_provider_uri) {
```
gnidan: I think we don't want to use the network provider, since there's no reason to believe that the user will be developing against the same network as the network they're using for ethpm

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
-    expect.options(options, ["working_directory", "ethpm"]);
-
-    expect.options(options.ethpm, ["registry", "ipfs_host"]);
-
-    expect.one(options.ethpm, ["provider", "install_provider_uri"]);
-
-    // ipfs_port and ipfs_protocol are optinal.
+  install: async options => {
+    expect.options(options.ethpm, ["ipfs_host", "ipfs_port", "ipfs_protocol"]);
+    expect.options(options, ["working_directory", "ethpm_uri", "logger"]);
```
gnidan: note `ethpm_uri` should be in `options.ethpm`

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
-    if (host instanceof EthPM.hosts.IPFS === false) {
-      host = new EthPM.hosts.IPFSWithLocalReader(
-        options.ethpm.ipfs_host,
-        options.ethpm.ipfs_port,
-        options.ethpm.ipfs_protocol
+      options.provider ||
+      new Web3.providers.HttpProvider(options.ethpm.install_provider_uri);
+
+    // Parse ethpm uri
+    const ethpmURI = new Erc1319URI(options.ethpm_uri);
```
gnidan: btw if you get sick of typing `options.ethpm_uri`, you can do this kinda business:

```javascript
const { ethpm_uri } = options
```

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
+      registries: "ethpm/registries/web3",
+      installer: "ethpm/installer/truffle",
+      storage: "ethpm/storage/ipfs"
+    }).connect({
+      provider: provider,
+      workingDirectory: options.working_directory,
+      registryAddress: ethpmURI.address,
+      ipfs: {
+        host: options.ethpm.ipfs_host,
+        port: options.ethpm.ipfs_port,
```
gnidan: y'know if we're looking at this as a breaking change, we should get rid of the underscores and make them proper camelCase to match JS style

gnidan: (but we should not do a breaking change :)

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
+        options.working_directory
+      }`
+    );
   },
 
-  publish: function(options, callback) {
-    var self = this;
-
+  release: async options => {
+    expect.options(options.ethpm, [
```
gnidan: note that for this one and the above method, you'll want to keep the check on the first line to do `expect.options(options, ["ethpm"])`

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
-    if (!ropsten) {
-      return callback(
-        new TruffleError(
-          "You need to have a `ropsten` network configured in order to publish to the Ethereum Package Registry. See the following link for an example configuration:" +
-            OS.EOL +
-            OS.EOL +
-            "    http://truffleframework.com/tutorials/using-infura-custom-provider" +
-            OS.EOL
-        )
+    if (options.network !== "mainnet") {
```
gnidan: Note that the user defines custom names for networks, so `"mainnet"` here is jut a magic string

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
+      // validate is setup with mnemonic?
+      // validate mnemonic has releasing rights for options.registry?
+      throw new TruffleError(
+        "Please add a mainnet provider with your 12 word mnemonic to truffle.js"
```
gnidan: yeah, probably don't require a mnemonic :)

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
+      }
     });
-    var host = options.ethpm.ipfs_host;
 
-    if (host instanceof EthPM.hosts.IPFS === false) {
-      host = new EthPM.hosts.IPFS(
-        options.ethpm.ipfs_host,
-        options.ethpm.ipfs_port,
-        options.ethpm.ipfs_protocol
+    let artifacts = releasable_artifacts(options);
```
gnidan: camelCase :)

packages/core/test/ethpm.js
```diff
@@ -1,314 +1,409 @@
 var assert = require("chai").assert;
```
gnidan: btw if you're looking to test this locally, I usually just create an alias in my shell:
```bash
alias truffle="path/to/truffle/packages/core/cli.js"
```

Or, to test the bundled version (the version that everyone uses)
```bash
yarn bootstrap
alias truffle="node /path/to/truffle/packages/truffle/build/cli.bundled.js"
```

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
 const expect = require("@truffle/expect");
 const TruffleError = require("@truffle/error");
-const Networks = require("./networks");
-const EthPM = require("ethpm");
-const EthPMRegistry = require("ethpm-registry");
-const Web3 = require("web3");
-const { createInterfaceAdapter } = require("@truffle/interface-adapter");
-const async = require("async");
+const { EthPM } = require("ethpm");
```
gnidan: to make this non-breaking, I would suggest creating new packages in the monorepo here, maybe `@truffle/ethpm-v1` and `@truffle/ethpm-v2` and having those packages depend on the respective versions of the ethpm Node package.

Then, you could duplicate and move this `package.js` file to those corresponding new packages

packages/config/src/configDefaults.ts
```diff
@@ -35,8 +35,10 @@ export const getInitialConfig = ({
     ethpm: {
```
gnidan: in line with the non-breaking-ness suggestion below, we could add a `version` field here that would control the delegation to the correct v1/v2 package.

gnidan: so then to switch it, users would provide this sort of configuration:

```javascript
module.exports = {
  /* ... rest of truffle-config.js ... */
  ethpm: {
    version: "2",
    /* ... rest of ethpm config ... */
  }
}
```

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
+  const fileData = [];
+  const blockchainUri = options.networks.mainnet.network_id;
+  for (let file of files) {
+    const fileContents = JSON.parse(
+      fs.readFileSync(
+        path.join(options.contracts_build_directory, file),
+        "utf8"
+      )
+    );
+    // hacky af - blockchain uri wasn't getting picked up as network_id for 2 of 3 migrations
```
gnidan: btw we use the all-caps `// HACK - ` nomenclature for the grep lulz later

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
+    // some artifacts are coming through w/ network ids that are ints
+    // the blockchain uri coming through is what is set as the network_id
+    // not a freshly created one with latest block via BlockchainUtils
+    const foundNetworkId = Object.keys(fileContents.networks)[0];
+    if (foundNetworkId !== blockchainUri) {
```
gnidan: note that in order to construct a blockchain URI, you're going to need to actually connect to the networks in the truffle-config, enumerating over each defined network, and for each of these networks, find the relevant block info

sorry

packages/core/lib/package.js
```diff
@@ -1,351 +1,214 @@
...
-        }
-      );
-    }
+    // Fetch target package manifest URI
+    const manifestURI = await ethpm.registries
+      .package(ethpmURI.packageName)
+      .release(ethpmURI.version);
+
+    // Install target manifest URI to working directory
+    await ethpm.installer.install(manifestURI, ethpmURI.address);
```
gnidan: we need to capture the result here as a list of contract types and list of deployments, so that we can pass this list of contracts to @truffle/artifactor. 

you'll need convert blockchain URIs into network IDs because that's how the artifacts format organizes contract instances

gnidan: note that the format is formally spec'd in @truffle/contract-schema

njgheorghita:
this pr has been superseded by #3247

PR #2862, opened 2020-02-28
Add Console.log with a feature flag

adrianmcli:
We should no longer hold this feature back from our users.

This PR puts `Console.log(variable)` and `Console.log("myVarLabel", variable)` behind a feature flag into Truffle so people can use it without downloading a different version of Truffle.

Simply:

1. `import "truffle/Console.sol";`
2. `Console.log(variable)`
3. Run your Truffle tests and see the result

### DIFF
--- a/packages/contract/lib/handlers.js
+++ b/packages/contract/lib/handlers.js
@@ -101,6 +101,9 @@ const handlers = {
       return context.promiEvent.reject(error);
     }
 
+    // Check logs for TruffleLogger usage & log if necessary
+    Utils.consoleLog(receipt.logs);
+
     // Emit receipt
     context.promiEvent.eventEmitter.emit("receipt", receipt);
 
--- a/packages/contract/lib/handlers.js
+++ b/packages/contract/lib/handlers.js
@@ -101,7 +101,7 @@ const handlers = {
       return context.promiEvent.reject(error);
     }
 
-    // Check logs for TruffleLogger usage & log if necessary
+    // Check logs for Console.log usage & log if necessary
     Utils.consoleLog(receipt.logs);
 
     // Emit receipt
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -80,6 +80,32 @@ const Utils = {
       .filter(log => log != null);
   },
 
+  consoleLog(receiptLogs) {
+    const TruffleLogEvents = receiptLogs.filter(x => x.event === "__Log");
+    TruffleLogEvents.forEach(event => {
+      const result = event.args;
+
+      // A mapping to show how to output into our console
+      const outputMap = {
+        num: result[0].toString(),
+        boolean: result[0],
+        str: result[0],
+        b32: result[0],
+        addr: result[0]
+      };
+
+      // A list of all "supported" types: ["num", "boolean", "str", ...]
+      const types = Object.keys(outputMap);
+
+      // Find out what type our result is, and log it out if it matches
+      types.forEach(type => {
+        if (result.hasOwnProperty(type)) {
+          console.log(outputMap[type]);
+        }
+      });
+    });
+  },
+
   toTruffleLog(events, isSingle) {
     // Transform singletons (from event listeners) to the kind of
     // object we find on the receipt
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -81,8 +81,8 @@ const Utils = {
   },
 
   consoleLog(receiptLogs) {
-    const TruffleLogEvents = receiptLogs.filter(x => x.event === "_TruffleLog");
-    TruffleLogEvents.forEach(event => {
+    const logEvents = receiptLogs.filter(x => x.event === "_TruffleLog");
+    logEvents.forEach(event => {
       const result = event.args;
 
       // A mapping to show how to output into our console
@@ -101,6 +101,34 @@ const Utils = {
       types.forEach(type => {
         if (result.hasOwnProperty(type)) {
           console.log(outputMap[type]);
+        }
+      });
+    });
+
+    const namedLogEvents = receiptLogs.filter(
+      x => x.event === "_TruffleLogNamed"
+    );
+    namedLogEvents.forEach(event => {
+      const result = event.args;
+
+      // A mapping to show how to output into our console
+      const outputMap = {
+        num: result[1].toString(),
+        boolean: result[1],
+        str: result[1],
+        b32: result[1],
+        addr: result[1]
+      };
+
+      // A list of all "supported" types: ["num", "boolean", "str", ...]
+      const types = Object.keys(outputMap);
+
+      // Find out what type our result is, and log it out if it matches
+      types.forEach(type => {
+        if (result.hasOwnProperty(type)) {
+          const label = web3Utils.toAscii(result[0]);
+          const value = outputMap[type];
+          console.log(label, value);
         }
       });
     });
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -81,7 +81,7 @@ const Utils = {
   },
 
   consoleLog(receiptLogs) {
-    const TruffleLogEvents = receiptLogs.filter(x => x.event === "__Log");
+    const TruffleLogEvents = receiptLogs.filter(x => x.event === "_TruffleLog");
     TruffleLogEvents.forEach(event => {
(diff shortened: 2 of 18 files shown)


eggplantzzz:
This looks cool!

packages/contract/lib/utils/index.js
```diff
@@ -80,6 +80,60 @@ const Utils = {
       .filter(log => log != null);
   },
 
+  consoleLog(receiptLogs) {
+    const logEvents = receiptLogs.filter(x => x.event === "_TruffleLog");
```
eggplantzzz: Should we make `_TruffleLog` more specific here? Maybe `_TruffleConsoleLog`? It's possible we may want that namespace later at some point? What do you think?

packages/contract/lib/utils/index.js
```diff
@@ -80,6 +80,60 @@ const Utils = {
...
+      // Find out what type our result is, and log it out if it matches
+      types.forEach(type => {
+        if (result.hasOwnProperty(type)) {
+          console.log(outputMap[type]);
+        }
+      });
+    });
+
+    const namedLogEvents = receiptLogs.filter(
+      x => x.event === "_TruffleLogNamed"
```
eggplantzzz: It follows that if we decide to rename to `_TruffleConsoleLog` then this should be too.

eggplantzzz:
We probably should make the error a bit sexier when we have some time though. Currently it looks like this if you omit the `--console-log` flag
Screen Shot 2020-02-28 at 11 36 52 AM

packages/contract/lib/utils/index.js
adrianmcli: I think that's a sensible suggestion. I'll make an update w/ this.

packages/contract/lib/utils/index.js
adrianmcli: Done, see latest commit.

adrianmcli:
> We probably should make the error a bit sexier when we have some time though. Currently it looks like this if you omit the `--console-log` flag
> Screen Shot 2020-02-28 at 11 36 52 AM
> For now I'll say this is good as it is clear as to what is going on.

@CruzMolina Is there a way for us to catch this easily and just add a line on top:

> Perhaps you are trying to use Truffle's Console.log feature. If so, make sure to include the `--console-log` flag when running your tests

CruzMolina:
@adrianmcli , looks like we can add a check in the `checkLibraries` method to catch usage of a `Console` library contract.
https://github.com/trufflesuite/truffle/blob/develop/packages/contract/lib/utils/index.js#L176-L207

packages/contract/lib/utils/index.js
```diff
@@ -80,6 +80,60 @@ const Utils = {
       .filter(log => log != null);
   },
 
+  consoleLog(receiptLogs) {
```
gnidan: seems like this method might could use a better name... `Utils.handleConsoleLogs` perhaps? `logConsoleLogs`?

packages/contract/lib/handlers.js
```diff
@@ -101,6 +101,9 @@ const handlers = {
       return context.promiEvent.reject(error);
     }
 
+    // Check logs for Console.log usage & log if necessary
+    Utils.consoleLog(receipt.logs);
```
gnidan: There's some nasty business with `receipt.logs` and `receipt.rawLogs`, and `receipt.logs` changes during the lifecycle of the transaction. `rawLogs` is straight from Web3.js, I believe, and might be preferable to use, unless you've done your due diligence here to make sure this doesn't have weird edge cases.

I would think to check Ganache in standard mode vs. default mode (`--noVMErrorsOnRPCResponse` flag), and also check for constructor behavior vs. regular contract call behavior. (Constructors can emit logs, right?)

cc @haltman-at because you know the perils of `receipt.logs`, if you have any insight here

packages/core/lib/commands/test/index.js
```diff
@@ -7,6 +7,11 @@ const command = {
       type: "boolean",
       default: false
     },
+    "console-log": {
```
gnidan: I know there's `truffle develop --log`, but maybe `truffle test --log` might be nice for this instead?

Alternatively, is this something to configure on a per-project basis? Did you consider adding it to the config?

packages/migrate/migration.js
```diff
@@ -198,6 +201,47 @@ class Migration {
     return { interfaceAdapter, resolver, context, deployer };
   }
 
+  async deployAndLinkLogger(options, resolver) {
```
gnidan: Hm, @truffle/migrate seems like a weird place for this. Did you think of other options for where to put this logic?

packages/resolver/sources/fs.js
```diff
@@ -57,6 +57,22 @@ class FS {
     ];
 
     let body, filePath;
+
+    if (importPath === "truffle/Console.sol") {
```
gnidan: this should be its own source, I think

gnidan:
Side-note: what Truffle commands does this support?

packages/contract/lib/handlers.js
haltman-at: For controlled logs like this, I think using `logs` should probably be fine.  Using `rawLogs` would mean having to do the decoding rather than getting them pre-decoded.  So, I think using `logs` here is fine.  As we know, the existing decoding logic that's used for `logs` (i.e.: ethers) has a number of problems, but if they used `rawLogs` they'd either just be:

1. Invoking that same decoder themselves, which wouldn't improve matters any;
2. Doing things manually, which doesn't strike me as a great idea; or
3. Using Truffle Decoder, which I don't think it currently makes sense to use here.

So, I think this way of doing things is basically fine for now.

adrianmcli:
I just realized that this has stalled and I would like to get this out the door. So I'll be responding to some of the comments above.

> Side-note: what Truffle commands does this support?

I believe it's only `test` for now.

packages/migrate/migration.js
adrianmcli: The way this works is that we have to link the logging library "behind-the-scenes" during the migration step. Unfortunately there's no way to have this outside of the migration step as far as I know.

@CruzMolina can chime in here if I got anything wrong.

packages/resolver/sources/fs.js
adrianmcli: @gnidan Not sure what "be its own source" means?

packages/core/lib/commands/test/index.js
adrianmcli: I think we want to make it so that we are sure the user deliberate wants us to link the logging library. Otherwise, it could incur additional cost and confusion.

packages/contract/lib/handlers.js
adrianmcli: Thanks @haltman-at for the input, will resolve this for now.

packages/contract/lib/utils/index.js
adrianmcli: I'm not opposed to simply re-naming this, @CruzMolina thoughts?

CruzMolina: I like `logConsoleLogs`.

adrianmcli: What about `printConsoleLogs` lol?

CruzMolina: Sure, just merge the damn thing already :trollface:

packages/resolver/sources/fs.js
gnidan: resolver defines sources (this one is `resolver/sources/fs.js`) - we can make `resolver/sources/internals.js` or somesuch... should be much cleaner that way

packages/core/lib/commands/test/index.js
gnidan: oh yeah I'm just talking about the name

packages/resolver/sources/fs.js
adrianmcli: @CruzMolina is probably more knowledgeable about this here.

CruzMolina: I consider this request out of scope for this initial PR.

@gnidan, want to open an issue and tackle it as some later point?

:shipit:

gnidan:
Also: I don't think we need a feature flag for this

davidmurdoch:
Can we ship this with the option to use either `console` or `Console`? The uppercase `C` looks weird to me. :-)

gnidan:
Closing this in favor of #3369, since that's now where the action is.

PR #2863, opened 2020-02-28
Internal improvement: Turn on source map support more broadly (I think?) (and do some other things)

haltman-at:
The original purpose of this PR was to turn on `source-map-support` in all compiled packages to enable better stack traces (currently only `codec` was using it).  I'm not sure how much that actually worked and how much it ended up being snake oil.  It should at least work for the Typescript packages?

Note that I modified the `build` command in `contract` to pass the `--debug` flag to `browserify` because apparently you need to do that for source mapping to work.

Since `debugger` uses babel, I tried using `babel-register` there instead of `source-map-support` but, uh, that caused errors.

I also ended up doing various other cleanup while I was at it:

1. `decoder` had some mistaken devDependencies; I fixed those up.

2. `artifactor` had some devDependencies listed as normal dependencies; I moved those over.

3. I noticed various packages importing `lodash` as a whole instead of just specific functions, so I changed them to just use specific functions.  In the case of `lodash.isArray`, I just replaced it with `Array.isArray`, and I replaced `lodash.isObject` checks with `typeof x === "object"`.


packages/codec/lib/index.ts
```diff
@@ -63,7 +63,7 @@
  * @module @truffle/codec
  */ /** */
 
-require("source-map-support/register");
+import "source-map-support/register";
```
gnidan: you're too cool

packages/config/src/configDefaults.ts
```diff
@@ -146,7 +146,7 @@ export const configProps = ({
           config = {};
         }
 
-        config = _.extend({}, defaultTXValues, config);
+        config = assignIn({}, defaultTXValues, config);
```
gnidan: this change name?

packages/contract/package.json
```diff
@@ -11,7 +11,7 @@
   "version": "4.1.10",
   "main": "index.js",
   "scripts": {
-    "compile": "sh -c \"mkdir -p ./dist\" && browserify ./index.js -o ./dist/truffle-contract.js && uglifyjs ./dist/truffle-contract.js -o ./dist/truffle-contract.min.js",
+    "compile": "sh -c \"mkdir -p ./dist\" && browserify --debug ./index.js -o ./dist/truffle-contract.js && uglifyjs ./dist/truffle-contract.js -o ./dist/truffle-contract.min.js",
```
gnidan: but then it's not a production build? is that a problem? would like a second opinion... @CruzMolina? @eggplantzzz?

yarn.lock
```diff
@@ -13719,7 +13726,7 @@ source-map-support@^0.4.15:
   dependencies:
     source-map "^0.5.6"
 
-source-map-support@^0.5.0, source-map-support@^0.5.1, source-map-support@^0.5.13, source-map-support@^0.5.3, source-map-support@^0.5.6:
+source-map-support@^0.5.0, source-map-support@^0.5.1, source-map-support@^0.5.16, source-map-support@^0.5.3, source-map-support@^0.5.6:
```
gnidan: can we clean up the `package.json`s so we only ever request one version of this? (it's just getting a bit out of hand)

packages/config/src/configDefaults.ts
haltman-at: Yes, `extend` is an alias: https://lodash.com/docs/#assignIn

yarn.lock
haltman-at: Hm, I'm not sure why this is happening -- the `package.json`s all say `^0.5.16`.  No idea why it's so messy here.  Is there something I missed somehow?

gnidan: oh maybe it's our dependencies that depend on `source-map-support`! nevermind then!

packages/contract/package.json
gnidan: @haltman-at want to split the contract stuff out into a separate PR so we can at least get the other packages updated while we hem and haw?

haltman-at: OK, I've combined this with the use of `exorcist` per @CruzMolina's suggestion.

haltman-at:
> Can we get a before and after bundle size comparison?

File sizes on develop:  
`truffle-contract.js`: 3145722 bytes (about 3 MiB)  
`truffle-contract.min.js`: 1880892 bytes (about 1.8 MiB)  
`truffle-contract.js.map`: (doesn't exist, obviously)

File sizes on this branch:  
`truffle-contract.js`: 3161706 bytes (still about 3 MiB)  
`truffle-contract.min.js`: 1890156 bytes (still about 1.8 MiB)  
`truffle-contract.js.map`: 3751972 bytes (about 3.6 MiB)

Differences:  
`truffle-contract.js`: Increased by 15984 bytes (about 15.6 KiB)  
`truffle-contract.min.js`: Increased by 9264 bytes (about 9 KiB)  
`truffle-contract.js.map`: Increased by the whole 3751972 bytes (about 3.6 MiB), duh

haltman-at:
> Thanks @haltman-at ! I'm glad we separated the source maps!
> 
> One last oddity I noticed with `@truffle/contract` is that we define `main` as `index.js` in the `package.json`, when it seems like we should be defining `./dist/truffle-contract.js` instead.

Huh, indeed.  However, actually changing this oddly results in the following error upon running a `yarn bootstrap`:

```
truffle: ERROR in ./packages/contract/dist/truffle-contract.js
truffle: Module not found: Error: Can't resolve './idna-map' in '/home/sniffnoy/truffle/truffle/packages/contract/dist'
truffle:  @ ./packages/contract/dist/truffle-contract.js 57560:4-57562:6
truffle:  @ ./packages/core/lib/testing/testrunner.js
truffle:  @ ./packages/core/lib/test.js
truffle:  @ ./packages/core/index.js
truffle: error Command failed with exit code 2.
truffle: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
truffle: error Command failed with exit code 2.
truffle: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
truffle: error Command failed with exit code 2.
truffle: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
lerna ERR! yarn run prepare exited 2 in 'truffle'
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
lerna ERR! yarn install --mutex network:42424 --non-interactive --ignore-engines exited 2 in 'truffle'
lerna ERR! yarn install --mutex network:42424 --non-interactive --ignore-engines exited 2 in 'truffle'
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Note that this happens when `truffle` is being prepared, not when `truffle-contract` is being prepared; that part seems to go fine.

Anybody know what's up with this?  @gnidan, @CruzMolina, @eggplantzzz?

haltman-at:
OK, per Slack discussion I'm leaving the `main` field alone for now.  To summarize for everyone else:

It seems that the browserified version has a problem; and since `main` is specifically a node.js field, it's basically OK to leave that as the source version rather than the browserified version.

To expand on the problem with the browserified version for future reference:

See that bit above about not being able to resolve `./idna-map`?  Yeah, `idna-map` isn't a package on NPM; rather, `idna-map.js` is a file in the [root directory of the] package [https://www.npmjs.com/package/idna-uts46-hx](idna-uts46-hx).  This package is required by [https://github.com/danfinlay/eth-ens-namehash](eth-ens-namehash), which is required by [ethereum-ens](https://www.npmjs.com/package/ethereum-ens), which is required by `@truffle/contract`.

It seems that `eth-ens-namehash` once ran into a similar problem, and also resorted to setting `main` to the source version rather than the bundled version.

It still needs to be determined whether the bundled version actually *works* and how to fix it.  But we can leave that for another PR.

PR #2870, opened 2020-03-04
Fix: Remove problematic debugging output

haltman-at:
My previous PR, #2851, seems to be causing the Colony tests to fail.  The reason for this seems to be some debugging output I left in that foolishly includes some dereferencing (which, obviously, can fail).  So, I'm just deleting that debugging output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,10 @@ matrix:
       env: FABRICEVM=true
     - node_js: 8
       env: COVERAGE=true
-    - node_js: 10
+    - if: |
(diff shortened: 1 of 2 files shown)


haltman-at:
(Note: I've made a commit to turn on Colony tests unconditionally on this branch. I'll revert that commit once the Colony tests pass.)

haltman-at:
OK, the Colony tests passed this time so I'm reverting that commit!  (Two other tests failed but for stupid reasons, and anyway those ones are always run anyway.)

PR #2872, opened 2020-03-05
Update debugger and decoder for Solidity 0.6.3: Contracts and enums as mapping keys

haltman-at:
This PR allows the debugger and decoder to work with enums and contracts as mapping keys, as introduced in Solidity 0.6.3.

To accomplish this, enums and contracts are now reclassified as elementary types in the codec's type system.

Also, one of the existing regexes for handling mapping AST nodes wouldn't have worked properly for enums and contracts; it's been modified appropriately (it just needed an extra underscore at the end :) ).

I also added cases for `contract` and `enum` to `wrapElementaryType` so the decoder can watch them.  They're fairly primitive and unsafe, but, well... so is that whole function.  It'll get rewritten better later...

To allow contracts as mapping keys to work better with watch expressions, contracts now just nativize to their address.

I also added tests.  In doing this I realized the decoder tests kind of needed to be reorganized.  So, I did that.


packages/codec/lib/format/errors.ts
```diff
@@ -337,7 +338,82 @@ export interface UfixedPaddingError {
 }
 
 /*
- * SECTION 3: CONTAINER TYPES (including magic)
+ * SECTION 4: User-defined elementary types
```
gnidan: Did you get rid of section 3?

packages/decoder/test/current/test/decoding-test.js
```diff
@@ -33,24 +32,26 @@ contract("DecodingSample", _accounts => {
...
     await decoder.watchMappingKey("varMapping", 3);
     await decoder.watchMappingKey("varAddressMapping", address);
+    await decoder.watchMappingKey("varContractMapping", address);
+    await decoder.watchMappingKey(
+      "varEnumMapping",
+      "DecodingSample.E.EnumValOne"
+    );
+    await decoder.watchMappingKey("varEnumMapping", "E.EnumValTwo");
+    await decoder.watchMappingKey("varEnumMapping", "EnumValThree");
+    await decoder.watchMappingKey("varEnumMapping", 4);
```
gnidan: not too shabby!

packages/codec/lib/format/errors.ts
haltman-at: Oops.  Will go fix the numbering...

Issue #2723, opened 2020-01-06
Getting a typeError when running `truffle test` after update


Amxx:
Coming back from vacations, I just updated my truffle install from v5.1.4 to v5.1.7.

When running truffle tests (that used to work) I'm now getting:

```
  "before all" hook: prepare suite:
     TypeError: Cannot read property 'typeClass' of undefined
      at abiSizeAndAllocate (node_modules/truffle/build/webpack:/packages/codec/dist/lib/abi-data/allocate/index.js:91:1)
      at abiSizeAndAllocate (node_modules/truffle/build/webpack:/packages/codec/dist/lib/abi-data/allocate/index.js:199:65)
      at allocateMembers (node_modules/truffle/build/webpack:/packages/codec/dist/lib/abi-data/allocate/index.js:60:63)
      at allocateDataArguments (node_modules/truffle/build/webpack:/packages/codec/dist/lib/abi-data/allocate/index.js:370:1)
      at allocateCalldataAndReturndata (node_modules/truffle/build/webpack:/packages/codec/dist/lib/abi-data/allocate/index.js:301:60)
      at getCalldataAllocationsForContract (node_modules/truffle/build/webpack:/packages/codec/dist/lib/abi-data/allocate/index.js:525:1)
      at Object.getCalldataAllocations (node_modules/truffle/build/webpack:/packages/codec/dist/lib/abi-data/allocate/index.js:563:1)
      at new WireDecoder (node_modules/truffle/build/webpack:/packages/decoder/dist/decoders.js:189:1)
      at Object.<anonymous> (node_modules/truffle/build/webpack:/packages/decoder/dist/index.js:213:1)
      at step (node_modules/truffle/build/webpack:/packages/decoder/dist/index.js:164:1)
      at Object.next (node_modules/truffle/build/webpack:/packages/decoder/dist/index.js:145:45)
      at /home/amxx/Work/iExec/code/PoCo-dev/node_modules/truffle/build/webpack:/packages/decoder/dist/index.js:139:1
      at new Promise (<anonymous>)
      at __awaiter (node_modules/truffle/build/webpack:/packages/decoder/dist/index.js:135:1)
      at Object.forProject (node_modules/truffle/build/webpack:/packages/decoder/dist/index.js:211:1)
      at TestRunner.initialize (node_modules/truffle/build/webpack:/packages/core/lib/testing/testrunner.js:80:1)
      at process._tickCallback (internal/process/next_tick.js:68:7)

```

[...]

PR #2873, opened 2020-03-07
Bug fix: Fail inheritance search on bad base node type

haltman-at:
This PR is a (partial?) fix for the reopened #2723.  Basically, when going through the base contracts of a given contract to find where a given function is defined, I have code that says, give up if you fail to locate one of the base contracts.  However, I didn't have code that said, give up if due to ID collisions the "base contract" you locate is not a contract at all.  This caused an error when attempting to test @Amxx's project.

I'm actually still getting errors when I try to test his project, but they're unrelated errors and it's not clear they're our fault, so I'm putting up this PR as it is to just fix this one issue.

One thing worth noting: In the particular case at hand, the base contract we're trying to look up is actually the contract itself!  Wondering if perhaps I should put in code so that for the first contract in the base contracts list (i.e. the contract itself) we don't do a lookup, so that this sort of error can't happen unless there's a collision in one of the actual base contracts *other* than the contract itself?

Also: It's really weird that there was a collision here at all!  IDs are supposed to be unique within a compilation, I thought?  @Amxx might want to report that to Solidity...

(Nonetheless this is still our fault because we ought to be able to recover gracefully from such collisions one way or another.)

### DIFF
--- a/packages/codec/lib/abi-data/allocate/index.ts
+++ b/packages/codec/lib/abi-data/allocate/index.ts
@@ -364,13 +364,17 @@ function allocateCalldataAndReturndata(
       //search through base contracts, from most derived (left) to most base (right)
       if (contractNode) {
         const linearizedBaseContracts = contractNode.linearizedBaseContracts;
+        debug("linearized: %O", linearizedBaseContracts);
         node = linearizedBaseContracts.reduce(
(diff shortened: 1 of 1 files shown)


haltman-at:
OK, I did end up making that lookup-skipping change after all.  Also, I reviewed the code a bit more and found two more places where a `nodeType` check should be added, so I did that.

PR #2880, opened 2020-03-11
Prevent some debugger crashes

### LINKED ISSUE #2879: Debugger Fails Cryptically:  Error: Compilation shimmedcompilation has unreliable source order
- [ X] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

> What the issue is, in broad strokes.
Debugger Fails Cryptically:  Error: Compilation shimmedcompilation has unreliable source order
## Steps to Reproduce
Screenshot from 2020-03-11 15-30-36

## Expected Behavior

> What you expected to happen.
Debug the transaction
## Actual Results
Screenshot ^
What actually happened. Please give examples and support it with screenshots, copied output or error messages.

## Environment

**ubuntu 16.04
Ganache CLI v6.5.0 (ganache-core: 2.6.0)
Truffle v5.1.16 (core: 5.1.16)
Solidity v0.5.16 (solc-js)
Node v12.6.0
Web3.js v1.2.1**

haltman-at:
This PR does two things:

Firstly: It fixes #2879.  Um, so, I messed up the check of whether the debugger needs to recompile, and as a result instead of recompiling when it needed to, it was crashing.  Oops!  Fixed now.

(Very odd, since I could've sworn I'd seen it recompiling in some cases...?  A little unclear what's going on there, may be worth looking at...)

Secondly: It prevents various other debugger crashes that I accidentlly introduced in 5.1.16 when source code can't be found for the transaction being debugged.

### DIFF
--- a/packages/core/lib/debug/cli.js
+++ b/packages/core/lib/debug/cli.js
@@ -38,7 +38,11 @@ class CLIDebugger {
     let artifacts = await DebugUtils.gatherArtifacts(config);
     let shimmedCompilations = Codec.Compilations.Utils.shimArtifacts(artifacts);
     //if they were compiled simultaneously, yay, we can use it!
-    if (!shimmedCompilations.unreliableSourceOrder) {
+    if (
+      shimmedCompilations.every(
+        compilation => !compilation.unreliableSourceOrder
+      )
+    ) {
       return shimmedCompilations;
     }
     //if not, we have to recompile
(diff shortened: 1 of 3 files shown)


haltman-at:
No, I didn't add any tests, I just tested it manually.  These sorts of failure conditions are a bit hard to add tests for, and as for anything involving the CLI (which is what the first part is about), well...

PR #2885, opened 2020-03-13
Prevent problems with breakpoints when location unknown

haltman-at:
Oops, sorry, one more.  Forgot to test breakpoints.  Fixed now.

### DIFF
--- a/packages/core/lib/debug/interpreter.js
+++ b/packages/core/lib/debug/interpreter.js
@@ -54,8 +54,8 @@ class DebugInterpreter {
       ? currentLocation.source.id
       : null;
     const currentLine =
-      currentSourceId != null //!= is deliberate
-        ? //sourceRange is never null, so we go by whether currentSourceId is null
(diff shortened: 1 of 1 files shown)


packages/core/lib/debug/interpreter.js
```diff
@@ -50,12 +50,14 @@ class DebugInterpreter {
...
 
     const currentNode = currentLocation.node ? currentLocation.node.id : null;
-    const currentLine = currentLocation.sourceRange
-      ? currentLocation.sourceRange.lines.start.line
-      : null;
     const currentSourceId = currentLocation.source
       ? currentLocation.source.id
       : null;
+    const currentLine =
+      currentSourceId != null //!= is deliberate
```
gnidan: instead of the comment just do `currentSourceId !== null && currentSourceId !== undefined`, no?

haltman-at: Sure, that would also work, if you think that's clearer.

PR #2887, opened 2020-03-14
Add missing chai dev dependency in decoder

haltman-at:
The decoder has been using chai in its tests, but this didn't appear in the package.json.  Now I've added it as a devDependency.


haltman-at:
Does it need updating?  I ran `yarn` again and it didn't change anything...

PR #2888, opened 2020-03-16
Internal improvement: Update api for Require.exec

eggplantzzz:
~This is a breaking change for @truffle/require!~ This is no longer breaking as `Require.exec` now has a Promise or callback api.


packages/core/lib/commands/exec.js
```diff
@@ -73,29 +73,16 @@ const command = {
...
-            if (err) return done(err);
-
-            Require.exec(
-              config.with({
-                file: file
-              }),
-              done
-            );
-          });
+          return Contracts.compile(config);
```
gnidan: But `Contracts.compile` isn't async, is it?

eggplantzzz: It sure is! It actually has both apis right now (callback or Promise).

packages/require/require.js
```diff
@@ -115,16 +115,12 @@ const Require = {
...
-      fn(done);
-    } catch (error) {
-      done(error);
-    }
+    const fn = this.file({
+      file: options.file,
+      context: { web3, interfaceAdapter },
+      resolver: options.resolver
+    });
+    await fn();
```
eggplantzzz: One thing to note here, this would disallow using callbacks in the functions provided by users. Should we worry about this?

packages/core/lib/commands/exec.js
gnidan: Want to try updating to the new compile interface? I ask because I'd like someone else besides me to get familiarity with it, but do feel free to tell me to do this work myself

eggplantzzz: Sure thing! I'll chat with you later to talk about the requirements.

packages/require/require.js
gnidan: We should preserve the ability for user scripts to be written with callbacks, cause this would break existing user code.

eggplantzzz: So is the answer here to do something like
```
if (callbackPassed) {
  return fn(callback);
} else { 
  await fn()
}
```?

**Comment by gnidan** [comment]
Just tried this out and it seems to be broken for callback-style scripts; not sure if that's what you're observing?

Compare these two exec scripts:
```javascript
const Migrations = artifacts.require("Migrations");

module.exports = async function() {
  await Migrations.new();
  console.log("Deployed Migrations");
}
```

and
```javascript
const Migrations = artifacts.require("Migrations");

module.exports = function(callback) {
  Migrations.new()
    .then(() => { console.log("Deployed Migrations"); })
    .then(callback)
}
```

When I run the async one, it works fine; I get the console output (`Deployed Migrations`). When I run the callback one, no output!

Seems like something isn't hooked up right? Not sure, what have you been trying?

eggplantzzz:
Oh, hmm. I'll investigate, nice catch.

eggplantzzz:
I'm going to solve this problem sometime in the future using `promisify`. I was hoping this would make the switch to using only Promises easier. It seems mainly like using `promisify` everywhere is easier than writing all these dual interfaces and making sure the behavior stays the same :)

PR #2892, opened 2020-03-17
Internal improvement: Remove callback api for Environment.Develop methods

eggplantzzz:
This PR is a breaking change for @truffle/environment.

Callbacks for methods in Environment.Develop are removed and Promises are implemented.


packages/environment/develop.js
```diff
@@ -97,56 +92,54 @@ const Develop = {
...
+        return {
+          started: false,
+          disconnect
+        };
+      })
+      .catch(async () => {
+        await this.start(ipcNetwork, ganacheOptions);
+        options.retry = true;
+        return this.connect(options).then(disconnect => {
+          if (connectedAlready) return;
```
gnidan: We can rewrite this with `await`s!

```suggestion
    try {
      const disconnect = await this.connect(options);
      connectedAlready = true;
      return {
        started: false,
        disconnect
      };
    } catch (_) {
      // cannot connect, must start
      await this.start(ipcNetwork, ganacheOptions);
      options.retry = true;
      const disconnect = await this.connect(options);
      if (connectedAlready) return;
```

packages/environment/develop.js
```diff
@@ -97,56 +92,54 @@ const Develop = {
...
-          callback(false, disconnect);
-        }
-      }
-    );
+          return {
+            started: true,
+            disconnect
+          };
+        });
+      });
```
gnidan: (continuing from above because GH suggestions still have strange limitations)

```suggestion
      return {
        started: true,
        disconnect
      };
    }
```

packages/environment/develop.js
eggplantzzz: Ok sounds good, I re-wrote this.

PR #2898, opened 2020-03-19
Memoize debugger's interval trees

haltman-at:
This PR is aimed at speeding up the debugger.  It looks like a lot of time is spent right now constructing the same interval trees over and over.  I want to avoid that.

So, now, the interval trees are all computed at startup, basically (they're in a big selector).  Rather than expose the trees directly, I made a selector `solidity.views.findOverlappingRange` which contains the `findOverlappingRange` functions.  There's also `solidity.current.findOverlappingRange` for when you just need the current one.

So, `ast/map.js` has now been deleted; all its functionality has been moved into either `solidity/selectors` or `controller/selectors`.  Note that we're likely going to have to move a bunch of this again later, so that the decoder can use it rather than just the debugger, but, oh well, one thing at a time.

Also, while I was at it, I changed the format of the `InternalFunctionsTable` in `codec` so that it contains `compilationId` (and renamed `source` there to `sourceIndex`).  This doesn't actually affect anything since nothing is actually using those fields right now, note.

Btw: I thought about getting rid of all the use of `jsonpointer` and just, y'know, storing the actual nodes rather than jsonpointers, but for now I left it as-is.  (There's actually a reason we wouldn't want to get rid of jsonpointer entirely, note.)  But, maybe I should go back and do that after all... it would certainly make the code cleaner to not be using `jsonpointer` everywhere, and just get the nodes directly.


haltman-at:
Yeah, why the hell not, went back and did that.  Now the actual nodes are stored in the interval tree.  The pointers are still there and still used, but we just use the node directly when we can.

PR #2901, opened 2020-03-19
Decoder output format serialization

haltman-at:
(PRing this now even though it's not quite ready so that we don't forget about it.)

This PR adds functions for converting the decoder output format from and to a serializable format.  It doesn't then actually do the serialization, like to JSON or anything, but it turns it into something serializable.  This means two things:

1. Replacing uses of `BN` and `Big` with `string` for representing numbers;
2. Removing all circularities in objects (in a reversible fashion -- thankfully the format is already set up to make this reversible).

So, the serialization occurs in two steps; first, all circularities are removed (via `sever`); then, all numbers (that aren't just of type `number`) are replaced with strings.  The deserialization turns strings back into numbers, then puts the circularities back in.  (To put the circularities in, it just uses the same function, `tie`, that creates the circularities in the first place.)

In order to avoid lots of copypaste, the types of the serialized versions are gotten by passing a configuration parameter to the type (the type now a parameter, but the default setting for this parameter just gets you the existing version).

Unfortunately:
1. Despite this, I ended up having to write the actual serialization/deserialization function via copypaste, so we didn't avoid it everywhere; and
2. Doing this to the types has done awful things to the documentation.

Right now (2) is the big problem.  So, while I'm PRing this, it shouldn't be considered ready yet; we first need some way to figure out how to make the docs work.  In the worst case, this may mean just ripping out the parameterized types, and putting in the copypaste...

(But @gnidan is currently working on the problem!)


haltman-at:
OK, I've updated this branch to account for the addition of `TypeTypeEnum`.

haltman-at:
Updated again for the `immutable` changes.  Thankfully no changes needed to the serialization/deserialization code, all changes were to `error.ts`.

eggplantzzz:
Closing this until we are closer to actually implementing this.

Issue #2469, opened 2019-10-17
Add method for getting all artifacts


haltman-at:
## Issue

It would be quite useful for the codec (or users of the codec) if there were a method that allowed one to get all the project's artifacts (say as an array).  These could just be artifacts or they could be whole constructor objects like `artifacts.require` returns; I guess the latter would be more convenient.  It could be called `artifacts.requireAll` or something similar.  This would allow one to use the codec without having to always first manually get all the artifacts you need.

## Expected Behavior

This method should exist.

## Actual Results

No such method currently exists.

haltman-at:
Note: @gnidan remarks in Slack that the method should just return a list of names, not actual artifacts; it could be called `artifacts.list`.

(And then you could do `artifacts.list().map(artifacts.require)` for what I want, I guess...)

haltman-at:
Worth noting that `truffle console` and `truffle test` both currently do primitive versions of this where they just read the `build` directory, but that isn't the only source of artifacts potentially.

PR #2907, opened 2020-03-20
Enhancement: Allow decoder to automatically find artifacts on disk

haltman-at:
This PR makes it more convenient to use the decoder by making it so that the user no longer has to manually specify all the artifacts, but instead can just pass in the config object and let the decoder find all the artifacts.  For the project info, you can now pass `{ config }` to do this.

(So, @adrianmcli, if you want to make better use of the decoder, rather than just passing in a single contract and hoping that's sufficient, I'm hoping that with this PR you'll finally be able to do that conveniently!)

Note that currently, it does this by just searching the `contracts_build_directory` for artifacts, the same way so many other things in our code do; so if there are contracts living somewhere other than the build directory, it'll miss them.  Eventually hopefully #2469 will be resolved and we'll be able to get all the artifacts in a more reliable manner; but for now we just do it the way everyone else does.

Note that I removed the artifact-gathering code from `testrunner.js` since it was only used to set up the decoder; now it simply passes in the config and lets the decoder do the work. :)

Unfortunately I couldn't really add tests of this, because the decoder tests are Truffle tests and there's not currently any way to get access to `config` in there, but because the decoder is itself used in `truffle test`, I didn't really need to add tests of this. :)  I just ran the ordinary tests with `--show-events` on. :)


haltman-at:
Also, I have upgraded `fs-extra` to 8.1.0 across the board.  Not 9.0.0, because that version doesn't work with node 8.

PR #2909, opened 2020-03-24
truffle networks only reads JSON files

### LINKED ISSUE #2906: truffle-networks fails when non artifacts in build folder
`truffle networks` fails parsing non json in build folder. We had a zip file that caused the command to fail with stack trace.

@nicholasjpaterno suggests it should only operate on .json files.

rkalis:
closes #2906
- truffle networks [--clean] used to fail with non-JSON files
  in the directory. These non-JSON files are now ignored.

### DIFF
--- a/packages/core/lib/networks.js
+++ b/packages/core/lib/networks.js
@@ -10,7 +10,10 @@ const Networks = {
   deployed: async function(options) {
     let files;
     try {
-      files = fs.readdirSync(options.contracts_build_directory);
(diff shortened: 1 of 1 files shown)


gnidan:
Hey @rkalis, our build was broken when you opened this PR, and now it is fixed! Could you please rebase so we can get the green check marks? Thanks for opening this :)

rkalis:
Done :)

PR #2910, opened 2020-03-24
[hdwallet-provider] fix detection of websockets urls

roderik:
I was trying to use a websockets url in the hdwalletprovider.

```
networks: {
    launchpad: {
      provider: () => new HDWalletProvider(bpaasconfig.mnemonic, 'wss://xxx.settlemint.com/xxx/besu/ws'),
      gasPrice: '0',
      network_id: '*',
      websockets: true,
      production: false,
      gas: "0x1dcd6400"
    },
  },
```

I figured this was possible reading the code here

https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/src/index.ts#L166-L179

Some issues I detected:

- `providerProtocol` is actually `wss:` so the case never triggers
-  `case "wss":` does not have a `break;` so is always overridden

Quick workaround is passing in my own web3 provider, but this code exists so i made it work

### DIFF
--- a/packages/hdwallet-provider/src/index.ts
+++ b/packages/hdwallet-provider/src/index.ts
@@ -164,7 +164,7 @@ class HDWalletProvider {
       // Web3.providers.HttpProvider.prototype.send;
       let subProvider;
       const providerProtocol = (
-        Url.parse(provider).protocol || "http"
+        Url.parse(provider).protocol || "http:"
       ).toLowerCase();
 
       switch (providerProtocol) {
--- a/packages/hdwallet-provider/src/index.ts
+++ b/packages/hdwallet-provider/src/index.ts
@@ -168,9 +168,10 @@ class HDWalletProvider {
       ).toLowerCase();
 
       switch (providerProtocol) {
-        case "ws":
-        case "wss":
+        case "ws:":
+        case "wss:":
           subProvider = new Web3.providers.WebsocketProvider(provider);
+          break;
         default:
           // @ts-ignore: Incorrect typings in @types/web3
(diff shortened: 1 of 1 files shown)


gnidan:
Hey @roderik, thanks for this!

Would you mind rebasing against current `develop`? Our build was failing, and now it is not.

roderik:
@gnidan done

packages/hdwallet-provider/src/index.ts
```diff
@@ -168,9 +168,10 @@ class HDWalletProvider {
       ).toLowerCase();
 
       switch (providerProtocol) {
-        case "ws":
-        case "wss":
+        case "ws:":
+        case "wss:":
```
gnidan: Huh. That's weird. Maybe the better solution would be to strip off the `:` before the `switch` statement.

But I don't care enough on this point to block approval. We'll see what @eggplantzzz says for the second approve :)

Thanks again or this @roderik!

roderik: Yeah the same : are already used here: https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/src/index.ts#L215 so i thought just to go with the flow (even though obviously nobody uses this undocumented feature :))

gnidan: sounds good!

haltman-at:
Thanks for this!  One suggestion: As long as you're doing this, would you mind fixing line 167 to use `"http:"` as well?  I don't think it really affects anything as well, but we should probably make sure to consistently do things the right way here.

roderik:
done

PR #2912, opened 2020-03-24
Internal improvement: Adjust test case for GETH job

CruzMolina:
Differentiates geth & ganache behavior when transacting with insufficient funds.


haltman-at:
If the tests pass then I guess this is good, right?

PR #2913, opened 2020-03-25
Enhancement: Allow @truffle/decoder to fully decode internal function pointers

haltman-at:
This PR adds support for full decoding of internal function pointers to `@truffle/decoder`.

In order to allow this, a lot of code that used to live in the debugger has been moved to the `solidity-utils` package so that the decoder can import it.  Note that the `solidity-utils` package is uncompiled JS, so I had to eliminate two uses of object spreads.  Also, warning, `decoder` uses it as an untyped import!  But it's all localized to one little block where the internal functions table is set up.

(I did not change the fundamental architecture of internal function pointer decoding -- i.e., `codec` still just relies on a table that's passed in to decode these; it doesn't do the work of decoding them itself.)

There was also one really awkward thing wtih `IntervalTree`.  For reasons unknown to me, using `IntervalTree` the normal way -- via the default export -- didn't work in `solidity-utils`, despite having always worked in `debugger` (where it gets compiled with Babel and Webpack); Node complained that the default export isn't a constructor.  However, the "advanced" interface worked fine, so rather than trying to set up Babel or something, I just switched over to that.

Also: I hope it's OK that `solidity-utils` imports `codec` for one little utility function? :)  I mean, it's not circular or anything...

The internal functions table is set up in the `init` function of `ContractInstanceDecoder`; it's set there rather than in the constructor because it needs to know the contract's code.  Of course, if the decoder decides there's insufficient information, it will just leave the table `undefined`, and you just won't get the extra information.

I also updated documentation and added a test.

EDIT: This is actually kind of a breaking change to `decoder`, now that I think about it, in that it allows it to output stuff it wouldn't have before, so users of it will have to account for that.  (Note to @adrianmcli, although I believe you weren't yet handling function pointers anyway.)

### DIFF
--- a/packages/codec/lib/format/values.ts
+++ b/packages/codec/lib/format/values.ts
@@ -490,13 +490,16 @@ export interface FunctionInternalValueInfoException {
 }
 
 /**
- * This type is used when decoding internal functions from the high-level
- * decoding interface, which presently doesn't support detailed decoding of
- * internal functions.  (The debugger, however, supports it!  You can get this
- * detailed information in the debugger!)  You'll still get the program counter
- * values, but further information will be absent.  Note you'll get this even
- * if really it should decode to an error, because the decoding interface
- * doesn't have the information to determine that it's an error.
+ * This type is used when decoding internal functions in contexts that don't
+ * support full decoding of such functions.  The high-level decoding interface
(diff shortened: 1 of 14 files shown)


haltman-at:
Note: I thought this PR could use an additional test, so I went back and added one.

PR #2914, opened 2020-03-25
Fix type IDs on contract types from contexts & internal fns

haltman-at:
Big oops here: When moving everything over to the new style of type ID, I fixed two of the import functions, but I forgot one: `contextToType`.

This leaves us with a problem: Where should `makeTypeId` live?  It used to live in `ast/import`, because it was needed there but not in `abi-data/import`.  But now it's needed in `context/import` as well.  I opted to put it in `context/import` and have `ast/import` import it.

Except, there's one more place type objects were being made with old-style IDs... in `basic/decode`!  See, `decodeInternalFunction` was constructing some type objects manually.  We don't want that -- all such construction (and especially, any making of type IDs) should occur in an import function.

So, I added `evm/import` and a function for converting internal function table entries to types.  Hooray, abstraction restored!

...really, a better way to handle this would probably be to just put said types in the table, rather than requiring reconstructing it later, but, eh, this is good enough for now.  And I didn't want to cause merge conflicts with my other internal functions PR. :)

### DIFF
--- a/packages/codec/lib/ast/import/index.ts
+++ b/packages/codec/lib/ast/import/index.ts
@@ -7,6 +7,7 @@ import * as Common from "@truffle/codec/common";
 import * as Compiler from "@truffle/codec/compiler";
 import * as Utils from "@truffle/codec/ast/utils";
 import { AstNode, AstNodes } from "@truffle/codec/ast/types";
+import { makeTypeId } from "@truffle/codec/contexts/import";
 
 //NOTE: the following function will *not* work for arbitrary nodes! It will,
 //however, work well enough for what we need.  I.e., it will:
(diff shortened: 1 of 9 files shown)


packages/codec/lib/ast/import/index.ts
```diff
@@ -476,10 +477,3 @@ export function definitionToStoredType(
     }
   }
 }
-
-//I am deliberately *NOT* exporting this.
-//If you have to make a type ID, instead make the type and then
-//take its ID.
```
gnidan: 👍

haltman-at: I mean, as you can see from the new comment, consumers of `codec` are still not supposed to use it -- the idea is that type IDs should be opaque and only constructed by import functions.  I'm just no longer enforcing that by not exporting it, because turns out more import functions than just these ones needed it.

PR #2915, opened 2020-03-26
Whitelist missing solc options

### LINKED ISSUE #2902: Whitelist revert strings debug compiler option
## Issue

Solidity 0.6.3 added a debug flag one can pass to the compiler, where the compiler builds in revert strings for any compiler-added reverts.  (See the [release notes](https://github.com/ethereum/solidity/releases/tag/v0.6.3).)

I don't think this is in our whitelist of options to pass along to the Solidity compiler; it should be whitelisted so users can put this in their truffle-config and compile their contracts with this debug option if they want.

haltman-at:
This PR fixes #2902 by passing the `debug` option through to solc.  I didn't add any tests because I don't know this package well enough, sorry!  But I tested it manually and it works.

By the way, I was looking through the solc documentation for compiler options, and I noticed there's two other options we don't currently pass through: `metadata` and `libraries`.  Does anyone know if there's a reason for this?  It would be easy enough to add them in.  (Would allowing the user to set the metadata format cause any problems?  I have to imagine letting them manually specify libraries probably wouldn't, right?)

### DIFF
--- a/packages/compile-solidity/run.js
+++ b/packages/compile-solidity/run.js
@@ -185,6 +185,8 @@ function prepareCompilerInput({ sources, targets, settings }) {
       optimizer: settings.optimizer,
       remappings: settings.remappings,
       debug: settings.debug,
+      metadata: settings.metadata,
+      libraries: settings.libraries,
       // Specify compilation targets. Each target uses defaultSelectors,
       // defaulting to single target `*` if targets are unspecified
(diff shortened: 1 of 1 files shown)


haltman-at:
OK, after discussing this in Slack with @gnidan, I've gone ahead and expanded this PR to whitelist the other settings we were missing as well (`metadata` and `libraries`).

PR #2916, opened 2020-03-26
Display all migration gas numbers in dec and hex formats

### LINKED ISSUE #2799: Error message when exceeding block gas limit presents comparable numbers in different bases
## Issue
When sending more gas than the network's current block limit, the error message that comes back is confusing and can/should be clarified. 

## Steps to Reproduce
In truffle-config, specify a "gas" value which is higher than the block gas limit for the network you are connected to.  Attempt to do a basic migration or any other Truffle transaction.

## Expected Behavior
The "Block limit" and "Gas sent" values are both in the same base, probably decimal.  Then the user can easily see that the "gas sent" number is larger than the "block limit" and possibly do something about it.

## Actual Results
As shown in [this issue](https://github.com/trufflesuite/truffle/issues/2278):
>"Migrations" exceeded the block limit (with a gas value you set).
>   * Block limit:  0x20175c8
>   * Gas sent:     6721975
>   ...

The inequality between these two going the wrong way is the core issue, but using different bases obscures that comparison and makes the error harder to understand. 

## Environment
* Truffle version (`truffle version`): 5.1.0

rkalis:
closes #2799
- Add `decAndHex(num)` utlity function that displays a number as decimal with hex in parenthesis
- Use this utility in all migration messages that print gas

I wasn't sure what the versioning process is, so I left the versions as is. But since the dependencies are `core` depends on `migrate` depends on `reporters`, both `reporters` and `migrate` need a version bump to be included in core.

### DIFF
--- a/packages/reporters/reporters/migrations-V5/messages.js
+++ b/packages/reporters/reporters/migrations-V5/messages.js
@@ -29,6 +29,10 @@ class MigrationsMessages {
     return `MIGRATION_STATUS:${JSON.stringify(msg)}`;
   }
 
+  decAndHex(num) {
+    return `${Number(num).toString(10)} (0x${Number(num).toString(16)})`;
+  }
+
   // ----------------------------------- Interactions ----------------------------------------------
 
   questions(kind) {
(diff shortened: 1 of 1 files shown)


rkalis:
@CruzMolina It should be in gas units. wei/gwei only come into play once you have to pay for the gas, which is unrelated to the limit or the amount of gas sent. Unless I'm misunderstanding what you mean?

gnidan:
People have enough time if they wanted to object to this :)

I'm merging. Thanks again @rkalis !

Issue #2841, opened 2020-02-20
Truffle Decoder cannot be imported in the browser


adrianmcli:
I think this has something to do with the way it's being webpacked and published. It's not outputting a module compatible with ES6 import statements. The following fails:

```js
import Decoder from "@truffle/decoder";
console.log(Decoder) // undefined
```

If you change it to CommonJS require statements, then it works:

```js
const Decoder = require("@truffle/decoder")
console.log(Decoder) // {ContractDecoder: ƒ ...}
```

But both of the above cases should work properly.

haltman-at:
Weird!  Thanks for pointing this out, this'll have to be fixed...

haltman-at:
Closing this since the issue is just that, as a TypeScript module without a default export, the way to import it is with `import *`.

adrianmcli:
I'm having this issue again, and it doesn't seem to be a lack of default import as it actually fails catastrophically (for a lack of a better word, I apologize). I've created a reproduction repo here that is based on a plain Create React App + Decoder. The moment I try to import it, it fails with:

```
index.js:107 Uncaught TypeError: Cannot read property 'native' of undefined
    at Object../node_modules/fs-extra/lib/fs/index.js (index.js:107)
    at __webpack_require__ (bootstrap:784)
    at fn (bootstrap:150)
    at Object../node_modules/fs-extra/lib/index.js (index.js:6)
    at __webpack_require__ (bootstrap:784)
    at fn (bootstrap:150)
    at Object../node_modules/@truffle/decoder/dist/index.js (index.ts:176)
    at __webpack_require__ (bootstrap:784)
    at fn (bootstrap:150)
    at Module../src/App.js (App.css?4433:45)
    at __webpack_require__ (bootstrap:784)
    at fn (bootstrap:150)
    at Module../src/index.js (index.css?f3f6:45)
    at __webpack_require__ (bootstrap:784)
    at fn (bootstrap:150)
    at Object.1 (serviceWorker.js:141)
    at __webpack_require__ (bootstrap:784)
    at checkDeferredModules (bootstrap:45)
    at Array.webpackJsonpCallback [as push] (bootstrap:32)
    at main.chunk.js:1
```

I wonder if it's something to do with the usage of `fs`.

adrianmcli:
This time it's failing (with the same error as above) with all of the following cases:

```js
// ES6 Module
import Decoder from "@truffle/decoder";

// ES6 Module w/ destructuring
[...]

PR #2919, opened 2020-03-31
Bug fix: Remove the use of fs-extra from @truffle/decoder

haltman-at:
This PR is intended to address #2841.  Turns out just importing `fs-extra` causes serious problems on web.  But also, I don't need it.  So, removing it.  Hopefully that'll fix the problem.  If not, we'll have to try hackier things...


adrianmcli:
I just tested it, it works!!!

adrianmcli:
Could we just merge this?

haltman-at:
Sure, I'll add you as a reviewer; then you can approve it and it can get merged. :D

PR #2946, opened 2020-04-08
Remove release notes folder cause gathering dust

gnidan:
Since this is causing security vulnerability alerts and we don't need this in source control anymore.


eggplantzzz:
👍

PR #2955, opened 2020-04-14
Convert @truffle/resolver to TypeScript

gnidan:
- Put code in `lib/` directory
- Change things to modern style (goodbye `require("async/whilst")`!)
- Define interface `ResolverSource` and make all the sources conform to that
- Use classes everywhere

(Practically speaking, this is non-breaking to @truffle/resolver)


gnidan:
> Unsurprisingly, I have little to say about this; I don't really know this part of the code. So, my only comments are:
> 
> 1. Since you're using TS, you should use `source-map-support`.

👍 

> 2. Kind of tangential, but, do you know why `decoder` is the only package that passes the `--declaration` option to tsc?  Should all TS packages be doing that?

Yes, they all should!

> 3. Hoping the dive into the `resolver` codebase you had to do in order to make this means you might be well equipped to write `resolver.list` now! XD

This PR's a distant pre-cursor to that effort, but yeah, that's why I requested your review @haltman-at ;)

gnidan:
Thanks!

PR #2956, opened 2020-04-14
Convert Resolver.resolve to async

gnidan:
ℹ️ **This is a breaking change to @truffle/resolver**

Besides the high-level change to promisify `resolver.resolve`, this PR also includes:

- Rename ResolverSource method `resolve_dependency_path` to `resolveDependencyPath`
- Rename `file` property to `filePath` in return value, since that's what it's called elsewhere (and it's more explicit)
- Convert various old-style classes to use `class` syntax

### DIFF
--- a/packages/core/lib/testing/testresolver.js
+++ b/packages/core/lib/testing/testresolver.js
@@ -1,42 +1,39 @@
 const path = require("path");
 
-function TestResolver(resolver, source, search_path) {
-  this.resolver = resolver;
-  this.source = source;
-  this.search_path = search_path;
+module.exports = class TestResolver {
+  constructor(resolver, source, searchPath) {
+    this.resolver = resolver;
+    this.source = source;
+    this.searchPath = searchPath;
 
-  this.seen = [];
-  this.require_cache = {};
-  this.cache_on = true;
-}
-
-TestResolver.prototype.require = function(import_path) {
-  if (this.cache_on && this.require_cache[import_path]) {
-    return this.require_cache[import_path];
+    this.seen = [];
+    this.require_cache = {};
+    this.cache_on = true;
   }
 
-  // For Windows: Allow import paths to be either path separator ('\' or '/')
-  // by converting all '/' to the default (path.sep);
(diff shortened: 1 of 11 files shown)


packages/core/lib/testing/testsource.js
```diff
@@ -4,104 +4,104 @@ const fse = require("fs-extra");
...
-    addresses.forEach((address, i) => {
-      const name = path.basename(abstractionFiles[i], ".json");
+      const addresses = filesData
+        .map(data => JSON.parse(data))
+        .map(json => contract(json))
+        .map(c => {
+          c.setNetwork(self.config.network_id);
+          if (c.isDeployed()) return c.address;
+          return null;
+        });
```
nicholasjpaterno, line 60: On the one hand this is nice, clean and easy to read. On the other hand this is :vomiting_face: since we map over the same data set 3 times..  I know this isn't new code but I wonder if it doesn't deserve a little TLC.

packages/core/lib/testing/testresolver.js
```diff
@@ -1,42 +1,39 @@
...
-    .catch(callback);
-};
+    return result;
+  }
 
-module.exports = TestResolver;
+  async resolve(importPath, importedFrom) {
+    const result = await this.source.resolve(importPath);
+    if (result && result.body) {
+      return result;
```
eggplantzzz, line 34: So it looks like the signature of this method changed from returning `result.body` and `result.resolvedPath` to just returning the object containing those two. I can't tell, do we need to adjust the calling code?

packages/core/lib/testing/testsource.js
gnidan: Want to PR what you're thinking?

packages/core/lib/testing/testresolver.js
gnidan: Nope, I investigated that and I believe we're all good!

gnidan:
Thanks @nicholasjpaterno and @eggplantzzz! Merging this. Will address followup PR if you can't get to it @nicholasjpaterno, let me know!

PR #2971, opened 2020-04-17
Update cli.js, improve help conditional

nicholasjpaterno:
The `inputArgument.length` check should be first to short circuit.  This looks like a bug.

Improve readability with `array.includes`.  Looks like there would be no benefit to amortize the creation of a `new Set(...etc)` since this is only called once per run?

### DIFF
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -41,7 +41,8 @@ listeners.forEach(listener => process.removeListener("warning", listener));
 let options = { logger: console };
 
 const inputArguments = process.argv.slice(2);
-const userWantsGeneralHelp = inputArguments.length && ['help', '--help'].includes(inputArguments[0]);
+const userWantsGeneralHelp = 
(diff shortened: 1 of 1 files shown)


packages/core/cli.js
```diff
@@ -41,9 +41,7 @@ listeners.forEach(listener => process.removeListener("warning", listener));
 let options = { logger: console };
 
 const inputArguments = process.argv.slice(2);
-const userWantsGeneralHelp =
-  (inputArguments[0] === "help" || inputArguments[0] === "--help") &&
-  inputArguments.length === 1;
+const userWantsGeneralHelp = inputArguments.length && ['help', '--help'].includes(inputArguments[0]);
```
CruzMolina, line 44: [These two tests](https://github.com/trufflesuite/truffle/pull/2971/checks?check_run_id=594035026#step:6:950) are failing because this code now returns `true` and calls `command.displayGeneralHelp` when passing `help <validOrInvalidCommand>` to the cli, instead of logging the particular commands help menu or logging the command passed is invalid.

PR #2972, opened 2020-04-17
Update repl.js improve readability & performance.

nicholasjpaterno:
No need to run `noop` functions, we can just make our own filter with a reduce to collect the `.done` methods.

### DIFF
--- a/packages/core/lib/repl.js
+++ b/packages/core/lib/repl.js
@@ -49,9 +49,9 @@ ReplManager.prototype.start = function(options) {
       // If we exit for some reason, call done functions for good measure
       // then ensure the process is completely killed. Once the repl exits,
       // the process is in a bad state and can't be recovered (e.g., stdin is closed).
-      const doneFunctions = self.contexts.reduce((fns, ctx) => {
-        if (ctx.done) fns.push(ctx.done);
-        return fns;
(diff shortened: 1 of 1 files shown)


nicholasjpaterno:
Theres always the readability route
```javascript
const doneFunctions = self.contexts.map(ctx => ctx.done ? ctx.done() : () => undefined);
```

packages/core/lib/repl.js
```diff
@@ -49,16 +49,11 @@ ReplManager.prototype.start = function(options) {
...
-        return context.done
-          ? function() {
-              context.done();
-            }
-          : function() {};
-      });
-      async.series(doneFunctions, function() {
-        process.exit();
-      });
+      const doneFunctions = self.contexts.reduce((fns, ctx) => {
```
CruzMolina, line 52: Since part of this work is improving readability,
I'd generally prefer callback arguments to be spelled out.

```suggestion
      const doneFunctions = self.contexts.reduce((functions, context) => {
```

CruzMolina: I also dig the suggestion [here] (https://github.com/trufflesuite/truffle/pull/2972#issuecomment-614998441,
although the `reduce` is pretty readable as well (and an improvement from the existing code).

PR #2978, opened 2020-04-19
Various performance improvements to `resolver` package's `require`

seesemichaelj:
This PR is a quick solution to increasing the performance of the `resolver` package's `require` function. What it does:

- Reduce the number of `JSON.parse` to only execute once per JSON file per call to `require`. Before, we would parse the file until we found what we want, return just the name, and then read the file *again* and parse it
- "Heuristically" try the most likely file first. We have a convention that `ContractName.sol` becomes `ContractName.json`. I can't think of a scenario where this wouldn't happen (perhaps when the name of the file does not match the name of the contract, but I think things error in other places [solc maybe] when that doesn't match), but we use this to leverage the fact we could almost always only do one file read and JSON parse rather than potentially search through `n` files

### DIFF
--- a/packages/resolver/lib/sources/fs.ts
+++ b/packages/resolver/lib/sources/fs.ts
@@ -31,7 +31,11 @@ export class FS implements ResolverSource {
     searchPath = this.contractsBuildDirectory
   ) {
     // most cases we can use the `{name}.sol -> {name}.json` convention
+    // console.log(sourcePath);
+    // console.log(searchPath);
     const likelyFileBasename = path.basename(sourcePath, ".sol");
+    // console.log(likelyFileBasename);
+    // console.log(path.join(searchPath, `${likelyFileBasename}.json`));
     const likelyArtifactString = fs.readFileSync(
       path.join(searchPath, `${likelyFileBasename}.json`),
       "utf8"
@@ -41,22 +45,17 @@ export class FS implements ResolverSource {
     try {
       if (likelyArtifactString) {
         likelyArtifact = JSON.parse(likelyArtifactString);
+        return likelyArtifact;
       }
     } catch (e) {
       // do nothing, we'll handle this by doing a deeper search
-    }
-
-    if (
-      likelyArtifact &&
-      path.basename(likelyArtifact.sourcePath, ".sol") ===
-        path.basename(sourcePath, ".sol")
-    ) {
-      return likelyArtifact;
+      // console.log(e);
     }
 
     // we couldn't use the `{name}.sol --> {name}.json` convention, let's search
 
     const contractsBuildDirFiles = fs.readdirSync(searchPath);
+    // console.log(contractsBuildDirFiles);
     const filteredBuildArtifacts = contractsBuildDirFiles.filter(
       // we don't care about files that don't have the json extension
       // we also don't care about the file we already tried
@@ -73,9 +72,12 @@ export class FS implements ResolverSource {
         if (artifactString) {
           const artifact = JSON.parse(artifactString);
 
+          // console.log(artifact.sourcePath, sourcePath);
+
           if (
             path.basename(artifact.sourcePath, ".sol") ===
-            path.basename(sourcePath, ".sol")
+              path.basename(sourcePath, ".sol") ||
+            artifact.contractName === path.basename(sourcePath, ".sol")
           ) {
             return artifact;
           }
--- a/packages/resolver/lib/sources/fs.ts
+++ b/packages/resolver/lib/sources/fs.ts
@@ -26,12 +26,20 @@ export class FS implements ResolverSource {
     return result;
   }
 
+  /**
+   *
+   * @param importSourcePath The string in the require/import statement
+   *   This could be artifacts.require("Migrations") => "Migrations", or even
+   *   artifacts.require("Migrations.sol") => "Migrations.sol". This function handles both
+   * @param searchPath The location to search
+   * @returns The artifact as a JSON object
+   */
   searchForArtifact(
-    sourcePath: string,
+    importSourcePath: string,
     searchPath = this.contractsBuildDirectory
   ) {
     // most cases we can use the `{name}.sol -> {name}.json` convention
-    const likelyFileBasename = path.basename(sourcePath, ".sol");
+    const likelyFileBasename = path.basename(importSourcePath, ".sol");
     const likelyArtifactString = fs.readFileSync(
       path.join(searchPath, `${likelyFileBasename}.json`),
       "utf8"
@@ -41,9 +49,7 @@ export class FS implements ResolverSource {
     try {
       if (likelyArtifactString) {
         likelyArtifact = JSON.parse(likelyArtifactString);
-        if (likelyArtifact.sourcePath === sourcePath) {
-          return likelyArtifact;
-        }
+        return likelyArtifact;
       }
     } catch (e) {
       // do nothing, we'll handle this by doing a deeper search
@@ -70,8 +76,8 @@ export class FS implements ResolverSource {
 
           if (
             path.basename(artifact.sourcePath, ".sol") ===
-              path.basename(sourcePath, ".sol") ||
-            artifact.contractName === path.basename(sourcePath, ".sol")
+              path.basename(importSourcePath, ".sol") ||
+            artifact.contractName === path.basename(importSourcePath, ".sol")
           ) {
             return artifact;
           }
--- a/packages/resolver/lib/sources/fs.ts
+++ b/packages/resolver/lib/sources/fs.ts
@@ -31,11 +31,7 @@ export class FS implements ResolverSource {
     searchPath = this.contractsBuildDirectory
   ) {
     // most cases we can use the `{name}.sol -> {name}.json` convention
-    // console.log(sourcePath);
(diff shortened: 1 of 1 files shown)


seesemichaelj:
Wow can't believe it's been more than a month since I did this work. To give some more insight @haltman-at, I made these changes during some overtime work on a weekend and didn't have/take the time to get the tests working. @gnidan offered to help, but by basically restructuring the code, so I refused saying I would take care of it so that we do it right. This is only effective if I actually do the work though. I'll try to fit in some time in the next 2 weeks

seesemichaelj:
Here's the Slack thread for posterity: https://trufflesuite.slack.com/archives/C012GQDNV2M/p1587336310034500

seesemichaelj:
Woop I got tests to pass @gnidan. I think this is ready now

packages/resolver/lib/sources/fs.ts
```diff
@@ -23,38 +22,73 @@ export class FS implements ResolverSource {
...
+    // console.log(path.join(searchPath, `${likelyFileBasename}.json`));
+    const likelyArtifactString = fs.readFileSync(
+      path.join(searchPath, `${likelyFileBasename}.json`),
+      "utf8"
+    );
+
+    let likelyArtifact;
+    try {
+      if (likelyArtifactString) {
+        likelyArtifact = JSON.parse(likelyArtifactString);
```
gnidan, line 47: It's an edge case, for sure, but we should check to make sure that `likelyArtifact.sourcePath` ends with `sourcePath`

packages/resolver/lib/sources/fs.ts
```diff
@@ -23,38 +22,73 @@ export class FS implements ResolverSource {
...
+
+          // console.log(artifact.sourcePath, sourcePath);
 
-      if (artifact.sourcePath === sourcePath) {
-        return artifact.contractName;
+          if (
+            path.basename(artifact.sourcePath, ".sol") ===
+              path.basename(sourcePath, ".sol") ||
+            artifact.contractName === path.basename(sourcePath, ".sol")
+          ) {
```
gnidan, line 81: Could you explain why it's okay to be looser with this check?

packages/resolver/lib/sources/fs.ts
```diff
@@ -23,38 +22,73 @@ export class FS implements ResolverSource {
...
-        fs.readFileSync(path.resolve(searchPath, buildArtifact)).toString()
+      const artifactString = fs.readFileSync(
+        path.resolve(searchPath, buildArtifact),
+        { encoding: "utf8" }
       );
+      try {
+        if (artifactString) {
+          const artifact = JSON.parse(artifactString);
+
+          // console.log(artifact.sourcePath, sourcePath);
```
gnidan, line 75: 👀

gnidan (commented):
Here's some comments. Generally this looks good. Love the O(1) normal case.

packages/resolver/lib/sources/fs.ts
seesemichaelj: you can't really compare the old check with the new check. the function works differently than it did before, so you really should be evaluating this PR for the current contents and not as a diff

i believe i used the `path.basename` here because `sourcePath` in this context varied. one can include `.sol` whereas the other does not

however it's been 5 months since i wrote this code and i'm not entirely sure what the driving force would be for this. i don't think this check is looser, but the correct amount of flexibility given the new execution context

by don't take my word for it and debug through the code to understand what happens. i'm not trying to sign up for maintaining this code, but just trying to put some work into performance since no one was jumping at the oppotunity

packages/resolver/lib/sources/fs.ts
seesemichaelj: I had to revert this change because `sourcePath` is not the same as `likelyArtifact.sourcePath`. The naming was confusing, so I updated it to `importedSourcePath` and added a JSDOC to explain that it's the string passed in via `artifacts.require()`.

Having the check would mismatch the following:

`likelyArtifact.sourcePath = "/tmp/tmp-3266431-Ru5Iww0nfJeN/contracts/Migrations.sol"`
`sourcePath /* now known as importedSourcePath */ = "Migrations"`

packages/resolver/lib/sources/fs.ts
seesemichaelj: > i believe i used the `path.basename` here because `sourcePath` in this context varied. one can include `.sol` whereas the other does not

I verified that this is correct. `importedSourcePath` (formally known as `sourcePath`) is the string in the `artifacts.require()` causing it to not be exactly  like `Artifact.sourcePath`

The function changed substantially causing the old school of thought of `sourcePath` being the source disk location to no longer be valid

packages/resolver/lib/sources/fs.ts
gnidan: Where this can go wrong is something like this... let's say you have `A.sol` that defines contract `B`, and `B.sol` that defines contract `A`

The existing code IIRC handles this by verifying the guess. In your proposed change, we lose this verification.

This might not seem like a problem in the abstract example, but this sort of thing can come up when people copy and paste multiple contracts into one source file, and multiple artifacts will get generated.

Now, doing this short-circuit check is fine; all I'm saying is that we should look at `likelyArtifact.sourcePath` to make sure it matches what the user entered, before blindly returning it.

packages/resolver/lib/sources/fs.ts
gnidan: Is this the part of the logic that handles the canonical case of `artifacts.require("<contract-name>")`, or is that also covered by your initial short-circuit check at the beginning?

I think it's probably worth making a separate short-circuit check at the very beginning for the basic contract name case, since that's the "correct way", so it should also be the fastest. Also would be good for clarity around the various cases.

gnidan (changes requested):
Requesting changes for the potentially erroneous return value, plus a question/suggestion

packages/resolver/lib/sources/fs.ts
seesemichaelj: > Is this the part of the logic that handles the canonical case of artifacts.require("<contract-name>"), or is that also covered by your initial short-circuit check at the beginning?

The latter.

The short-circuit covers:
- The canonical `artifacts.require("ContractName")`
- The less used `artifacts.require("ContractName.sol")`. In this case, we ignore the `.sol` the user provided and interpret it as `"ContractName"`
- The totally incorrect usage: `artifacts.require("some/relative/path/to/ContractName.sol")`. In this case, we ignore the path and `.sol` extension and interpret it as `"ContractName"`

> I think it's probably worth making a separate short-circuit check at the very beginning for the basic contract name case, since that's the "correct way", so it should also be the fastest. Also would be good for clarity around the various cases.

If I'm understanding you correctly, this suggestion is already implemented given my above answer

packages/resolver/lib/sources/fs.ts
seesemichaelj: > The existing code IIRC handles this by verifying the guess. In your proposed change, we lose this verification.

Is there a test you can point me to that demonstrates the existing code handles this?

---

I tried to give this the benefit of the doubt one last time and do something more robust than my prior attempt. I implemented:

``` typescript
      if (likelyArtifactString) {
        likelyArtifact = JSON.parse(likelyArtifactString);

        if (
          path.basename(likelyArtifact.sourcePath, ".sol") ===
            path.basename(importSourcePath, ".sol")
        ) {
          return likelyArtifact;
        }
      }
```

But this broke several `@truffle/debugger` tests. One example is https://github.com/trufflesuite/truffle/blob/develop/packages/debugger/test/context.js#L83

In my code, this would have `importSourcePath === "OuterContract"`. My short circuit would look for `"OuterContract.json"` in the `contractsBuildDirectory`. It would find it and parse it. However `likelyArtifact.sourcePath === "/tmp/tmp-2027082-yBcgRS8DiwlZ/contracts/OuterLibrary.sol"` so the basename would be comparing `OuterContract` to `OuterLibrary`. So my short circuit would fail.

Then we'd move onto my more lengthy lookup, which would search for every `json` file in the directory **except** `OuterContract.json` because we already tried it (so why try it again?). Since we filter out the correct artifact, we'll never find it.

---

However, I can add a check that looks at `likelyArtifact.contractName` instead of `likelyArtifact.sourcePath`? I tried the below code, but i got some `Unknown exit code: 1` errors in tests

``` typescript
      if (likelyArtifactString) {
        likelyArtifact = JSON.parse(likelyArtifactString);

        if (
          path.basename(likelyArtifact.contractName, ".sol") ===
            path.basename(importSourcePath, ".sol")
        ) {
          return likelyArtifact;
        }
      }
```

I believe this is a more correct check.

---

As an aside, instead of removing the `.sol` extension or looking at the `path.basename` of `importSourcePath`, I believe the FS resolver should error when it is provided a filename/path instead of an explicit contract name. Thoughts? Would this break current users?

---

Outside of this, it'd be very helpful if you could provide a failing test that explains the scenario you're talking about since this PR seems to be lower priority for both of us (and therefore we never are working on it at the same time)

eggplantzzz:
Closing this for maintenance. @tenthirtyone, @gnidan says this may be of interest to you.

PR #2980, opened 2020-04-20
Optimize testsource.js

nicholasjpaterno:
`indexOf` is bad for tests of inclusion.  In this case the set of elements was only length 2 but this is still a O(n) operation.  Let's just make it O(1).

### DIFF
--- a/packages/core/lib/testing/testsource.js
+++ b/packages/core/lib/testing/testsource.js
@@ -26,19 +26,19 @@ TestSource.prototype.resolve = async function(importPath) {
 
     const mapping = {};
 
-    const blacklist = ["Assert", "DeployedAddresses"];
+    const blacklist = new Set(["Assert", "DeployedAddresses"]);
 
     // Ensure we have a mapping for source files and abstraction files
     // to prevent any compile errors in tests.
     sourceFiles.forEach(file => {
       const name = path.basename(file, ".sol");
-      if (blacklist.indexOf(name) >= 0) return;
+      if (blacklist.has(name)) return;
       mapping[name] = false;
     });
 
     abstractionFiles.forEach(file => {
       const name = path.basename(file, ".json");
-      if (blacklist.indexOf(name) >= 0) return;
+      if (blacklist.has(name)) return;
       mapping[name] = false;
     });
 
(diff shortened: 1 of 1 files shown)


packages/core/lib/testing/testsource.js
```diff
@@ -26,19 +26,19 @@ TestSource.prototype.resolve = async function(importPath) {
 
     const mapping = {};
 
-    const blacklist = ["Assert", "DeployedAddresses"];
+    const blacklist = new Set(["Assert", "DeployedAddresses"]);
```
gnidan, line 29: In this case, O(1) is probably worse, since it's only two elements, so the extra overhead probably makes it slower.

That's just my hunch though, not sure if you've measured this sort of thing.

nicholasjpaterno:
```
Repeated 2 times.
obj: 0.078ms
indexOf: 0.012ms
includes: 0.007ms
set: 0.005ms
Repeated 2 times.

Repeated 20 times.
obj: 0.036ms
includes: 0.041ms
indexOf: 0.033ms
set: 0.020ms
Repeated 20 times.

Repeated 200 times.
obj: 0.053ms
includes: 0.120ms
indexOf: 0.089ms
set: 0.076ms
Repeated 200 times.

Repeated 2000 times.
obj: 0.153ms
indexOf: 0.296ms
includes: 0.238ms
set: 0.196ms
Repeated 2000 times.

Repeated 20000 times.
obj: 2.389ms
includes: 12.156ms
indexOf: 12.206ms
set: 0.593ms
Repeated 20000 times.
```
This was specifically with objects, arrays, and sets of size 2.  Looks like it depends on how many `sourceFiles`, `abstractionFiles`, and `addresses` there are.  Looks like its always the case that `Set` and `Object` access will be statistically faster.

gnidan: alright, color me wrong!

nicholasjpaterno: Hmm, I can't seem to find a hex color code for wrong..  :trollface:

nicholasjpaterno: Also the optimization is almost negligible.  So its really just cause I saw some low hanging fruit.

PR #2987, opened 2020-04-20
Internal improvement: Swap out github download library

eggplantzzz:
Recently there have been a number of errors related to downloading repos from github via the unbox and init command. A very short search suggested the possibility that these errors could be related to configuration of the library used to make http requests. Unfortunately, the `github-download` library is not super configurable and it has not been updated in 4 years. I found a different library that has much more activity (and has been much more recently updated) and so swapping it out seemed like it might be an easy shot at solving this problem. It also might just be good all around in general to update to a more relevant library.

A bunch of the URL validation/sanitization methods needed to be updated as the new library liked URLs of the format `https://github.com:trufflesuite/some-box` but not `https://github.com/trufflesuite/some-box` and not `git@github.com:trufflesuite/some-box`. In addition, it did not like the `.git` extension on the repo specifier and so a sanitization method for that needed to be included as well.

### DIFF
--- a/packages/box/box.ts
+++ b/packages/box/box.ts
@@ -116,7 +116,7 @@ const Box = {
     const tmpDir = tmp.dirSync({ unsafeCleanup });
     const unboxOptions = { logger, force };
     await Box.unbox(
-      `https://github.com/trufflesuite/truffle-init-${name}`,
+      `https://github.com:trufflesuite/truffle-init-${name}`,
       tmpDir.name,
       unboxOptions,
       config
--- a/packages/box/box.ts
+++ b/packages/box/box.ts
@@ -7,7 +7,61 @@ import fse from "fs-extra";
 import inquirer from "inquirer";
 import { sandboxOptions, unboxOptions } from "typings";
 
-function parseSandboxOptions(options: sandboxOptions) {
+/*
+ * accepts a number of different url and org/repo formats and returns the
+ * format required by https://www.npmjs.com/package/download-git-repo
+ * supported input formats are as follows:
+ *   - org/repo[#branch]
+ *   - https://github.com(/|:)<org>/<repo>[.git][#branch]
+ *   - git@github.com:<org>/<repo>[#branch]
+ */
+const normalizeURL = (
+  url = "https://github.com:trufflesuite/truffle-init-default"
+) => {
+  // remove the .git from the repo specifier
+  if (url.includes(".git")) {
+    url = url.replace(/.git$/, "");
+    url = url.replace(/.git#/, "#");
+    url = url.replace(/.git:/, ":");
+  }
+
+  // rewrite https://github.com/truffle-box/metacoin format in
+  //         https://github.com:truffle-box/metacoin format
+  if (url.match(/.com\//)) {
+    url = url.replace(/.com\//, ".com:");
+  }
+
+  // full URL already
+  if (url.includes("://")) {
+    return url;
+  }
+
+  if (url.includes("git@")) {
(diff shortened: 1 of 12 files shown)


haltman-at:
Oh boy, Node 8 compatibility issues looks like...

eggplantzzz:
Why do you think it is a Node 8 compatibility issue?

eggplantzzz:
This will have to wait until [this](https://github.com/trufflesuite/truffle/pull/2997) gets merged so that it can bundle correctly.

eggplantzzz:
I think this PR may now be unnecessary. Several of the issues related to init/unbox have disappeared as people seem to be resolving issues by tweaking their own network configurations?

eggplantzzz:
I'm going to go back on what I said and claim that we should move forward with this as the proposed library looks to be much better maintained. I'll work on getting the CI to pass.

gnidan:
Thanks @eggplantzzz!

eggplantzzz:
Looks like this work might wait until version 6 of Truffle.

haltman-at:
Having tried it, I can confirm that this fixes #3181.  (Unboxing fails for me on develop with the EXDEV error, but works fine on this branch.)

packages/core/test/lib/commands/unbox.js
```diff
@@ -11,29 +11,19 @@ describe("commands/unbox.js", () => {
...
-  const absolutePaths = [
-    "https://github.com/truffle-box/bare-box:/path/SubDir",
-    "truffle-box/bare-box:/path/subDir",
-    "bare:/path/subDir",
-    "git@github.com:truffle-box/bare-box:/path/subDir",
   ];
   const validBoxInput = [
-    "https://github.com/truffle-box/bare-box",
-    "truffle-box/bare-box",
     "bare",
```
eggplantzzz, line 19: So here are all the formats that we will support

packages/box/box.ts
```diff
@@ -116,7 +116,7 @@ const Box = {
     const tmpDir = tmp.dirSync({ unsafeCleanup });
     const unboxOptions = { logger, force };
     await Box.unbox(
-      `https://github.com/trufflesuite/truffle-init-${name}`,
+      `https://github.com:trufflesuite/truffle-init-${name}`,
```
gnidan, line 119: This syntax seems wrong to me - http git URLs don't use `:` the way that `git@github.com` URLs do. Is there a technical reason for this?

packages/box/lib/utils/unbox.ts
```diff
@@ -37,12 +37,15 @@ async function verifyURL(url: string) {
...
 function fetchRepository(url: string, dir: string) {
-  return new Promise((accept, reject) =>
-    // Download the package from github.
-    ghdownload(url, dir)
-      .on("err", reject)
-      .on("end", accept)
-  );
+  return new Promise((accept, reject) => {
+    // Download the package from github - download(url, dir, options, cb)
+    download(url, dir, error => {
```
gnidan, line 42: want to just `promisify` this, instead of creating a `new Promise()` manually?

packages/box/box.ts
eggplantzzz: This library wants this format and will not respond to the proper URL format above. You can check out examples [here](https://www.npmjs.com/package/download-git-repo#examples).

packages/box/lib/utils/unbox.ts
eggplantzzz: Yeah sure, let's do it!

Issue #2977, opened 2020-04-17
"truffle debug" hangs forever on start


xlab:
## Issue

After fixing this issue locally 
https://github.com/trufflesuite/truffle/issues/1752 by ignoring undefined in the code, the "truffle debug" command now just continues to spin that spinner forever:

```
⠼ Gathering information about your project and the transaction...
```

It actually consumes 100% CPU for a while, and if I run

```
DEBUG=* truffle debug
```

I can see a lot of debug info, but at some point the info stops and CPU usage drops to 0% and that spinner then spins forever.

<img width="1137" alt="Screenshot 2020-04-18 at 00 43 56" src="https://user-images.githubusercontent.com/477998/79616565-0d3f5100-810e-11ea-8335-7c1d267496c6.png">

This behaviour is the same on macOS (old macbook) and powerful Ubuntu machine.

## Environment

* Operating System: macOS 10.15.4 / Ubuntu LTS 18.04
* Ethereum client: Ganache CLI v6.9.1 (ganache-core: 2.10.2)
* Truffle version (`truffle version`): 5.1.22
* node version (`node --version`): v10.19.0 / v10.20.1
* npm version (`npm --version`): 6.14.4

haltman-at:
Ooh, hadn't seen this one before.  Do you have a reproduction example we could test this with, or do you know if this occurs with other projects as well?  I'm wondering whether the debugger might be failing on something in your particular project.  Thanks for reporting this!

xlab:
@haltman-at Hi thanks for the attention. I've prepared a minimal self-contained repo for debug purposes:

https://github.com/InjectiveLabs/0x-exchange-omnibus

The contract sources are basically 0x exchange from https://github.com/0xProject/0x-monorepo with Truffle config and migration file on top of it.

Steps 
```
yarn
yarn truffle build
yarn truffle debug
```

are enough to reproduce issue with `source === undefined` here
https://github.com/trufflesuite/truffle/blob/eb5b18d8a996ef46d869364c72c69b14362f5630/packages/debugger/lib/session/index.js#L164

I tried to understand how it works but there are too many mappings and arrays :)

xlab:
Also some tip:

Usually I see this
```
Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.
```

But in the case of that repo it builds contracts each time I try to migrate or run tests.

haltman-at:
OK, thanks a bunch!  I'll take a look.

[...]

PR #2991, opened 2020-04-21
Fix some errors reported by xlab

haltman-at:
This PR fixes some problems reported by @xlab in #2977.

The first of these is that having nonconsecutive sources can crash the debugger.  I fixed both the part that xlab pointed out, and another point in the code I found that could have trouble on nonconsecutive sources (hopefully these were indeed the only such spots).

The second of these is a bit more mysterious: If the check that xlab pointed out was fixed, the debugger would hang on startup.  On investigation, I determined that there were two particular ASTs that would never finish walking.  I was unable to determine any reason for this.  However, I replaced the uses of `yield call(walk, ...)` in the AST saga with a simple `yield* walk(...)`, and this somehow caused the problem to go away.   (No, we're not somehow losing parallelism here; `yield call` is completely synchronous.)  I have no idea why; some sort of weird redux-saga internal thing got overloaded from all the effects, maybe?  IDK.  But this fixed the problem, so, uh, I'm sticking with it.

Finally, I altered the CLI logic to force a recompile if source indices are not consecutive.  Now, you may recall that the first commit in this PR fixes problems caused by nonconsecutive indices, so why do this?  Because nonconsecutive sources, while they should no longer *cause* a problem, still *indicate* a problem.  In the case of xlab's repo, the nonconsecutivity was due to name collisions, which Truffle infamously can't really handle.  Or at least, Truffle's *artifacts* can't handle this, but the raw compiler output can, so by forcing a recompile we avoid the problem.  (At the cost of a lot of time.)

### DIFF
--- a/packages/debug-utils/index.js
+++ b/packages/debug-utils/index.js
@@ -100,7 +100,19 @@ var DebugUtils = {
       return false;
     }
 
-    //check #2: are there any AST ID collisions?
+    //check #2: are source indices consecutive?
+    //(while nonconsecutivity should not be a problem by itself, this probably
+    //indicates a name collision of a sort that will be fatal for other
+    //reasons)
+    //NOTE: oddly, empty spots in an array will cause array.includes(undefined)
+    //to return true!  So I'm doing it this way even though it looks wrong
+    //(since the real concern is empty spots, not undefined, yet this turns
+    //this up anyhow)
+    if (compilation.sources.includes(undefined)) {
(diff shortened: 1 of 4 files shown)


xlab:
> some sort of weird redux-saga internal thing got overloaded from all the effects

Oh my! Thanks for the investigation and improvements, this was a really deep rabbit hole. Will wait for Truffle DB to evolve ;)

PR #2997, opened 2020-04-22
Dependency update: Bump webpack to ^4 

eggplantzzz:
This PR bumps the version of webpack used in `./packages/truffle` and updates the configurations there. Each of the four entry points now has its own configuration file and these files have been moved into a folder named "configs".

For this, the bundle for `cli.js` passed a "smoke test" in that the basic commands, run on fresh unboxed projects, worked well. The other bundles should be tested as well and perhaps `cli.js` should be tested further as well.

One thing to note is that during the webpacking of `cli.js`, I was continuously blowing the heap and so I upped the amount of memory allotted to Node [here](https://github.com/trufflesuite/truffle/pull/2997/files#diff-37f23de1048876944c016514a88fcf27R19) which solved this problem. However, bundling now takes forever.

In addition a library (`prepend-file`) was removed as it seems to be completely unused.


gnidan:
Oh sweet, thanks for figuring that out @CruzMolina!

CruzMolina:
uhhhh, so `geth` latest is doing some weird shit, so I'm going to drop all these `geth`-related commits and re-investigate once things (hopefully) stabilize.

CruzMolina:
I still think this is generally good to go despite the failing `GETH` job(s).

PR #3001, opened 2020-04-23
Dependency update: Bump webpack to ^4 for debugger

eggplantzzz:
Update webpack for the debugger and adjust webpack config.

I did a very basic smoke test involving migrating the metacoin box (using `truffle develop`) and then debugging one of the deployment transactions. It seemed to work. @haltman-at if you think there is some more due diligence that I can help out with then let me know!


eggplantzzz:
So what's the deal @CruzMolina? Do we need to update that dependency as well or are we good?

CruzMolina:
If we care about that job still, then seems like we need to update the dep (or maybe just the webpack config script being passed to `mocha-webpack` in the `scripts` step?).  That job is failing for some other odd reason in CI that I've tried fixing (didn't try very hard), but I can still pass that job locally in the Ubuntu VM when on `develop`. Not with this change though.

What do you think @gnidan?

eggplantzzz:
Welp, looks like that dep is already on the latest version.

haltman-at:
Oh, the `mocha-webpack` issue came up when I tried to do this earlier on my branch `babel-up`, where I tried to upgrade not only webpack but also babel.  (Glad to hear upgrading webpack by itself went much better!)  Yeah `mocha-webpack` unfortunately isn't maintained, and the last stable release isn't compatible with webpack v4...

...*but* there is a beta version that is.  When @gnidan and I discussed this earlier when I was trying this, we decided using that version was OK in this case.  So, I'd suggest upgrading `mocha-webpack` to `2.0.0-beta.0`.

eggplantzzz:
@CruzMolina Do you want to try upgrading to that beta version and see what happens on your environment?

CruzMolina:
FYI, this is still failing the build cause of latest `geth`.

eggplantzzz:
Everybody's lovin it!

PR #3002, opened 2020-04-23
Increase maxBuffer to 10MB when compiling using native compiler

danhper:
I was getting a `spawnSync /bin/sh ENOBUFS` when trying to use the native `solc` command to compile.
It is because the stdin buffer size is 1MB by default and `execSync` was trying to pass more data.
I increased that to 10MB. It is far from perfect but at least it should do the job for most cases.

The other changes are because of eslint (git hook was failing). Let me know if you would prefer me to revert these.


packages/compile-solidity/compilerSupplier/loadingStrategies/Native.js
```diff
@@ -1,31 +1,26 @@
-const { execSync } = require("child_process");
-const LoadingStrategy = require("./LoadingStrategy");
-const VersionRange = require("./VersionRange");
+const { execSync } = require('child_process');
```
gnidan, line 1: Why'd you change the strings? Can you please undo this? It'll just get overwritten by prettier next commit anyway

gnidan:
If you undo the quote style changes, I'll take a stab at fixing eslint

danhper:
Thanks for the quick reply. I removed all the style changes.

gnidan:
Cool, thank you! Hopefully we'll have the build fixed today so we can merge this!

danhper:
Tests seem to be passing so hopefully it should be good to merge?

packages/compile-solidity/compilerSupplier/loadingStrategies/Native.js
```diff
@@ -6,18 +6,13 @@ class Native extends LoadingStrategy {
...
+    const maxBuffer = 1024 * 1024 * 10;
 
-    try {
-      return {
-        compile: options => String(execSync(command, { input: options })),
-        version: () => versionString
-      };
-    } catch (error) {
-      if (error.message === "No matching version found") {
-        throw this.errors("noVersion", versionString);
```
gnidan, line 17: Why'd you remove this invocation? Don't we need to handle this case? (@eggplantzzz do you know?)

danhper: Unless I am missing something, this is returning an object and not executing anything, so there is no way that it will raise an exception at this point.
`compile` might fail later on but having the `try`/`catch` here will not help.
Simplified example:

```javascript
var a;
// try/catch not helping
try {
  a = () => { thrown RuntimeError('asdf') }
} catch (e) { console.log(e) }
a() // will not be caught
```

TBH I changed this mainly to please eslint so if you think it's better to keep this for another PR I can of course revert this change.

gnidan: Hm, I prefer leaving code alone unless the change is warranted. @eggplantzzz added that line and I suspect it does something important.

It'll definitely make this an easy merge at this point if you restrict this PR to just the max buffer change. Happy to figure out the error handling business in an issue or a subsequent PR (you could very well be right, but I just don't want to keep you waiting on this particular issue... would love to get your fix here out in this week's release!)

danhper: Sure, sounds good.
Sorry for the back and forth, I think this is now as minimal as it gets in terms of changes.

eggplantzzz:
This seems like it is ok to me. Something is weird with the CI it seems

eggplantzzz:
Looks like merging the most recent stuff on `develop` got it passing again.

gnidan:
Thanks again for this, getting this into the release!

PR #3005, opened 2020-04-23
Enhancement: Mixed Solidity-and-JS stacktraces in Truffle Test with --stacktrace

haltman-at:
This PR adds mixed Solidity-and-JS stacktraces to Truffle Test!  This is an optional feature that has to be enabled with `--stacktrace`.  (This PR is kind of a horrible hack, and turning on that flag will limit what you can do in other ways, as I'll describe.  Wondering if the flag should be marked experimental.)  Note that `--stacktrace` automatically implies `--compile-all`, much like `--debug`.  Also, stacktracing only applies to *transactions* made during testing (including `Contract.new()`), not calls or gas estimates, and only works with transactions made via Truffle Contract.

(I also made some other minor improvements to Truffle Test's stacktraces that will apply even if you aren't using `--stacktrace`. :)  Also I altered our ESLint settings so that now any variable can go unused if it starts with an underscore.)

Now, it's already the case that the debugger can generate stacktraces... but how does it do it here?  How are Mocha's stacktraces altered to be what we want?

Well, the key thing to note here is that we already have a mechanism for altering stacktraces -- the `rejectHijacker` in `contract/lib/promievent.js`.  The reject hijacker hijacks the error coming from Mocha and modifies its `stack` property where the stacktrace lives.  (With `--stacktrace` on, we'll also hijack the `message` property as well, but one thing at a time.)  So, the new reject hijacker will still make those modifications (plus some additional ones I've made), but also it will, in `--stacktrace` mode, run the debugger (in light mode for speed) over the transaction, get the stack trace and revert message at the end, and use those as well in the modified stack trace.  Great!

Problem: How on earth do we get a debugger inside the reject hijacker, and how do we get it the transaction hash to use?

The first of these turns out to be the lesser problem.  In `test.js`, if `stacktrace` is set, then after getting the compilations, we set up a light-mode debugger with them.  We then, after initial deployment is done, attach this debugger -- hack warning! -- to each of the contract abstractions by saying `abstraction.debugger = bugger`.  Yup.  @gnidan tells me we should use the provisioner... but... :shrug: We then make sure in `contract/lib/contract/index.js` that this debugger is passed on to any instances they spawn, as well.

From here we can just pass the debugger into the `PromiEvent` we make in `execute.js`.  But how do we get the transaction hash in?

My original approach was setting up a `transactionHash` handler on the call to `web3.sendTransaction` to store it in the `PromiEvent`.  (Oh, I turned the `PromiEvent` into a class, by the way, so that it could get a `setTransactionHash` method.  I wrote it as an old-style class for style consistency. :P  But if you would prefer the modern, more readable way, I could do that too...)  But for whatever reason, that didn't work!  The transaction hash just wasn't getting set in time, it seemed.  I'm not very clear on why; I'm not very clear on how PromiEvents work at all, frankly.  (Well, OK -- in Ganache's standard mode it wasn't getting set in time, but in *default* mode it wasn't getting set *at all*.)  Clearly, `web3.sendTransaction` wasn't going to cut it.  I was going to have to do things a bit more manually.

So, `execute.sendTransaction` now exists.  If the `debug` flag on the `PromiEvent` isn't set -- this flag is set if a debugger was passed into it on construction -- then it simply does a `web3.sendTransaction` as normal.  If it is...

...then we have to manually mimic one ourselves.  That's right, it's manual RPC time!  But not just manual RPC time; I did my best to mimic the whole behavior of `web3.eth.sendTransaction`, including throwing the same errors it does.  (And including the fact that it sends a *signed* transaction if web3 has the `from` address in its `web3.eth.accounts.wallet`.)  I probably missed various details, but I'm hoping this is good enough.  Except, there's one respect in which it's definitely *not* good enough: It returns just an ordinary Promise, not a PromiEvent.

So: **If the --stacktrace flag is used, you cannot use PromiEvent features with any transactions and in-test deployments you do in Truffle Test**.  Be warned!

Still, I'm hoping that's a small enough limitation that people will be OK with it.  As mentioned, I made sure that this restriction *only* applies in this case.  I had to -- otherwise I'd have broken migrations!

Anyway, yeah, we've got this function that mimics `web3.eth.sendTransaction`, except, in the middle, it makes a call to `promiEvent.setTransactionHash`, and this works, regardless of whether the transaction succeeds or fails.  Hooray!

Then when the reject hijacker runs it... well, I already described this part.  But let's say a bit more about the changes to the stacktrace.

First, due to a bug, stacktraces were sometimes missing a newline.  I fixed this.  Also, I've chopped off the first two lines of each stacktrace (as they existed previously, that is) so that they don't show Truffle-internal parts the user doesn't care about.  (In the case of deployments there are actually *three* such lines, so one is still left in, but I don't see this as particularly worth fixing.)  These are both changes to the existing stacktraces, that will occur whether you have `--stacktrace` turned on or not.

With `--stacktrace` turned on, well, the Solidity stacktrace is displayed on top of that!  But its first line is chopped off and used as the `message` instead.  Note that stacktrace displays now include the revert message.  This means that now in the debugger when you get to the end of a failed transaction the revert message will display *twice*, but, eh, I figure that's acceptable.

Anyway! Solidity/JS stacktraces in tests!  They've got some limitations, but, they're here!


packages/contract/lib/promievent.js
```diff
@@ -1,36 +1,107 @@
...
+const DebugUtils = require("@truffle/debug-utils");
 const Web3PromiEvent = require("web3-core-promievent");
 
-function PromiEvent() {
-  const { resolve, reject, eventEmitter } = Web3PromiEvent.apply(
-    null,
-    arguments
-  );
+function PromiEvent(justPromise, bugger) {
+  //selectors are passed in separately because when I
```
gnidan, line 6: are they tho?

packages/contract/lib/promievent.js
```diff
@@ -1,36 +1,107 @@
...
+          let session = bugger.connect();
+          await session.load(this.txHash);
+          await session.continueUntilBreakpoint();
+          const report = session.view(
+            bugger.constructor.selectors.stacktrace.current.finalReport
+          );
+          const rawMessage = session.view(
+            bugger.constructor.selectors.evm.current.step.returnValue
+          );
+          await session.unload();
```
gnidan, line 29: Have you considered making this a new Session method? Seems like it might make sense, to avoid exposing all these details.

packages/contract/lib/promievent.js
```diff
@@ -1,36 +1,107 @@
+const debug = require("debug")("contract:promievent");
+const DebugUtils = require("@truffle/debug-utils");
 const Web3PromiEvent = require("web3-core-promievent");
 
-function PromiEvent() {
-  const { resolve, reject, eventEmitter } = Web3PromiEvent.apply(
-    null,
-    arguments
-  );
+function PromiEvent(justPromise, bugger) {
```
gnidan, line 5: You should just use new class syntax, don't worry about the style consistency for that sorta thing. We must have forward progress!

packages/contract/lib/promievent.js
gnidan: (or maybe a Debugger method instead... `bugger.getStackTrace()` or something better named)

packages/contract/lib/promievent.js
haltman-at: Oh, ha, thanks for catching this.

packages/contract/lib/promievent.js
haltman-at: This is a good idea, I think.  Could also add `getRevertMessage`.  Then I could simplify some of the logic currently sitting outside the debugger used for handling this stuff.

haltman-at: These would be session methods, not debugger methods, because that's where we put these things.  That said, I do wonder why we have the separate debugger/session mechanism at all.  Like, why doesn't `forTx` (or `forProject`) just return a session in the first place so you don't have to do `bugger.connect()`?  It just seems like a pointless extra step.  (I'd say this is a breaking change, but I guess it could be made non-breaking if we gave session a `connect` method that just returned `this`.)

gnidan: > It just seems like a pointless extra step.

I concur. Feel free to make this change!

haltman-at: Note: I think I'm actually not going to add `getRevertMessage` for now for technical reasons.  Rather I'm going to add the revert message to the stacktrace object.  Longer explanation later.

packages/contract/lib/promievent.js
haltman-at: Hm, turns out that broke things for some reason.  Leaving this alone for now actually.

haltman-at:
Btw, on testing things out, it looks like using truffle-contract calls as PromiEvents doesn't always work so great.  This is unrelated to my changes, just worth noting it does seem to mess things up a bit, so our support for it already isn't the best. :P

haltman-at:
OK, I've added `session.callstack()` and `session.stacktrace()`.  The callstack now includes the message in the bottom frame (decoded using Truffle Codec but done crudely, i.e., in a way equivalent to what other decoders would do but without the crashes), and `formatStacktrace` now uses this rather than taking `message` separately (that also keeps this change non-breaking to `debug-utils` :P ).

Also made some other adjustments while I was at it.  The `formatStacktrace` function now accounts for the case of `status == true` and `message` being defined.  And I added an `isDeploy` argument to `PromiEvent` so it can slice off that extra line. :)

haltman-at:
OK, this is unrelated to the main point of the PR but I've gotten rid of the unnecessary `Debugger` class and all the calls to `bugger.connect()`.  Now the Debugger functions just return the session directly.  For compatibility, I've given `session` a `connect()` method that just returns `this`, and also while I was at it I stuck a `selectors` getter on it. :)

haltman-at:
OK, I've added `--stacktrace-extra`, which is like `--stacktrace` but it does the compile in debug mode. :)

packages/core/lib/test.js
```diff
@@ -225,14 +251,21 @@ const Test = {
...
+    compilations,
+    bugger //for stacktracing
   }) {
     global.interfaceAdapter = interfaceAdapter;
     global.web3 = web3;
     global.assert = chai.assert;
     global.expect = chai.expect;
     global.artifacts = {
-      require: import_path => testResolver.require(import_path)
+      require: import_path => {
```
gnidan, line 262: BTW for next time, feel free to change the local variables to camelcase while you're at it in situations like this ;)

haltman-at:
OK, fixed things so that using `--stacktrace-extra` on <0.6.3 will just print a warning instead of failing.  Note this code relies on the assumption that the default version is <0.6.3.  Accidentally squashed this with my last commit so did a force push.  Anyway, rerequesting review now. :)

packages/core/lib/test.js
haltman-at: Heh, OK. :)

PR #3011, opened 2020-04-28
Internal improvement: Speed up webpack build

eggplantzzz:
Add 'mode' property to webpack config ~edit 'devtool' property for faster builds, and change the build script to not allocate extra memory for building as it is no longer needed with the updated 'devtool' setting~ and turn off minification to speed up the build.

### DIFF
--- a/packages/truffle/webpack.config.js
+++ b/packages/truffle/webpack.config.js
@@ -56,7 +56,12 @@ module.exports = {
     library: "",
     libraryTarget: "commonjs"
   },
-  devtool: "eval",
+  devtool: "source-map",
+
+  optimization: {
+    minimize: false
+  },
+
   module: {
     rules: [
       // ignores "#!/bin..." lines inside files
--- a/packages/truffle/webpack.config.js
+++ b/packages/truffle/webpack.config.js
@@ -7,6 +7,7 @@ const rootDir = path.join(__dirname, "../..");
 const outputDir = path.join(__dirname, "build");
 
 module.exports = {
+  mode: "production",
   entry: {
     cli: path.join(
       __dirname,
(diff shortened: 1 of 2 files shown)


haltman-at:
Hm, the [docs](https://webpack.js.org/configuration/devtool/) suggest `eval` isn't appropriate for production?

eggplantzzz:
Yep, I think it may be due to the size? I saw somewhere that there will be a larger built file but I'm not sure how much difference it makes. We should see what the difference is. 

Assuming that the size is not a concern I don't know what else we should worry about here. Any ideas?

eggplantzzz:
So here are the differences in files created and their sizes. The first screenshot is using "eval" and the second using "source map" as it was set beforehand. The stacktraces from "eval" reference the build JavaScript in TypeScript projects (for example from @truffle/box).
Screen Shot 2020-04-28 at 1 13 42 PM
Screen Shot 2020-04-28 at 1 16 19 PM

eggplantzzz:
Just for comparison, "cheap-source-map" takes 22 seconds and gives the following stacktrace when throwing an error in box.
```
Error: ya done killed it
    at /Users/tyler/projects/truffle/packages/truffle/build/webpack:/cli.bundled.js:1431:1
    at o (/Users/tyler/projects/truffle/packages/truffle/build/webpack:/cli.bundled.js:1431:1)
```

packages/truffle/webpack.config.js
```diff
@@ -79,6 +84,26 @@ module.exports = {
...
+      ),
+      "original-fs": path.join(__dirname, "./nil.js"),
+      "scrypt": "js-scrypt"
+    }
+  },
+
+  stats: {
+    warnings: false
+  },
+
```
eggplantzzz, line 106: All of this is not new code but rather just moved closer to the top.

PR #3016, opened 2020-04-29
Throw an error when web3-provider-engine fails to start

aantthony:
Fixes https://github.com/trufflesuite/truffle/issues/2655

Previously, the node process would just exit silently. This would happen for example if you used an invalid API key (example below).

```js
const provider = new HDWalletProvider(
  'abc',
  `https://rinkeby.infura.io/v3/FAKE_API_KEY`
);
console.log('sending...');
provider.sendAsync({ jsonrpc: '2.0', id: 1, method: 'net_version', params: [] }, (err, res) => {
  // never called
  console.log('response', err, res);
});
```

This commit adds a missing callback to throw an error when the 'web3-provider-engine fails' to start the 'eth-block-tracker'.

### DIFF
--- a/packages/hdwallet-provider/src/index.ts
+++ b/packages/hdwallet-provider/src/index.ts
@@ -183,7 +183,11 @@ class HDWalletProvider {
     } else {
       this.engine.addProvider(new ProviderSubprovider(provider));
     }
-    this.engine.start(); // Required by the provider engine.
+
+    // Required by the provider engine.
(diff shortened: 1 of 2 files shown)


aantthony:
You can just run the example code exactly as above. In the updated commit it will throw an error saying that Infura is rejecting the authentication (because the API key is not real)

eggplantzzz:
I like it

PR #3018, opened 2020-05-01
Enhancement: Add decoding in debugger for assembly variables (0.6.x only)

haltman-at:
This PR adds decoding for assembly variables in the debugger.  Well, I say "decoding" -- of course obviously the decoding is trivial; the hard part is keeping track of them.  This only works with 0.6.x or later; 0.5.x doesn't give us an AST for the inline assembly parts, so it would be impossible without parsing.

There are definitely some design questions here about what we should even be doing with these.  For instance, in Yul (which is what now goes in `assembly` blocks), the only type is `u256` -- corresponding to `uint256`.  So should assembly variables be decoded as `uint256`?  I thought it would make more sense to decode them as `bytes32`, so that's what I did.

Another question is, should we really just decode outside variables as normal while we're in assembly?  Like, maybe it would make more sense to show the value they'll act as in assembly, instead.  Should we show *both*?  Have a Solidity variables section and an assembly variables section, with Solidity variables showing up in both??

Or maybe we shouldn't *decode* assembly variables at all, and should simply add to the stack display indicators of which variables are where?  (We could do that with Solidity variables as well.)  Maybe that would be better than showing their values when the user hits `v`.  (Maybe we could even show the stack display by default while in assembly.)

But, OK -- let's talk about what I did do, instead of what I maybe should have done (plenty of time to revise this before merge if we don't like it).  So: How do we keep track of assembly variables, and why is it hard?

There are fundamentally two parts to this:

1. Determining what variables are in scope, and
2. Tracking what stack position each variable corresponds to.

Part 1 is probably the easier of the two, but is still harder for Yul variables than for Solidity variables.  The big problem here is that Yul AST nodes, unlike Solidity AST nodes, don't get IDs; and Yul variable declarations certainly don't get convenient scope fields in the AST telling us what their scope is.

So, this raises two problems:

1. We need to have some way of identifying AST nodes when they don't have IDs, and
2. We need to be able to apply Yul's scoping rules ourselves.

To solve the first problem, I identified Yul AST nodes by a combination of source index and JSON pointer, separated by a colon (think of the source index as a drive letter :) ).  Sometimes these IDs are kept separate from AST IDs; sometimes they're all mixed together.  It's all quite hacky, frankly.

Solving the second problem is pretty simple too; Yul's scoping rules aren't exactly complex.  One caveat worth noting is that functions defined in Yul have no access to any variables outside said function, so, I applied that rule.

(By the way, note that all of this required writing a separate AST walker for Yul ASTs; now in addition to `walk` there's also `yulWalk`.  That's where the work of determining variable scopes is done.  Determining what's in scope of course is still handled in `data.current.identifiers`.)

Btw, note that with no IDs there's now a lot more walking up the tree by means of text manipulation on JSON pointers.  Generally that should only come into play when Yul gets involved though; outside of Yul `scopes` is still used for this.

(Oh, crap, I just realized that `data.current.inModifier` will always report `false` when in Yul; I'll need to fix that before this is merged...)

OK, but what about keeping track of their stack positions?  I.e., `assignments`?  How do we handle that?  This is where things get more difficult, and it's largely due to Solidity's source mapping not being the best here.

Obviously, we're going to need some new cases in the data saga, to handle Yul node types.  Also: Remember I said Yul AST nodes don't have IDs?  That means that the data submodule needs to be made robust against nodes that lack IDs, whereas previously we handled this problem by excluding such nodes from the interval tree in the first place.

Speaking of the interval tree, you may notice I've changed the way that JSON pointers are compared there; they're now compared by number of slashes rather than by raw length.  I didn't want raw lengths potentially screwing things up.  Now normally there's no way it could matter, because the set of pointers matching a given interval should form a subtree, so it doesn't matter what means you use to judge length.  However, with Yul, that's not necessarily the case, because of the `externalReferences` field on the `InlineAssembly` node, which lie outside the Yul AST but have ranges that lie inside of it.  Fortunately, such nodes are always shallower in the AST than the Yul nodes they're coextensive with, so slash-counting makes it absolutely certain they won't be hit.  Probably this was the case with just checking lengths as well, but, eh, that's harder to reason about.

Anyway, back to the data saga.  We've got four new cases: `YulLiteral`, `YulIdentifier`, `YulFunctionCall`, and `YulFunctionDefinition`.  You may notice something missing: `YulVariableDeclaration`.  That's because the source mapping never lands on these!  If you do `let x := 1`, the `1` (a `YulLiteral`) will get sourcemapped to, but the statement as a whole will not.  So, instead, we have cases for the three types of Yul expressions (literals, identifiers, and function calls), and whenever we hit one we check whether its parent node is a variable declaration; if so we then process that variable declaration.  Note that one variable declaration can declare multiple variable simultaneously in Yul, so we have to account for all variables being declared there.  We also have to make sure that the function call is *done*, so in the function call case we don't do this if we're about to find ourselves in the arguments of the function call.

There's one big problem here: What about `let`s that *don't* give an explicit initial value, and just leave the initial value as the default 0?  These, it turns out, are skipped over entirely by the sourcemapping.  That doesn't make a lot of sense -- if you do `let x := 1` then yeah the `PUSH1 0x01` will correspond to that `1`, but if you do `let x`, shouldn't the resulting `PUSH 0x00` correspond to the `let x` as a whole?  It should, but it doesn't.  So, uh, for these ones, we're just out of luck; we can't keep track of them.  (With an exception you'll see below.)

This leaves the fourth case, `YulFunctionDefinition`.  We have to be a little bit careful here; in Solidity, a function definition is hit when the function is called (and again when it returns, but we don't worry about that).  In Yul, it's hit both these times... but also when it's *defined*.  So, we have to make sure we don't react to the function being defined, and only to it being called!

That at least is pretty easy -- we just check whether we're about to end up inside the function body.  (*Strictly* inside the function body, mind you; when the function is being defined we'll end up on the body node itself.  But when it's being called, we skip that and go directly inside the function.  I guess this assumption is violated when the function has no code at all, but, um, that's fine.  So we fail to allocate the parameters to (and return values from ) a function with no code, big whoop.

Yul functions actually put their parameters in a different order than Solidity functions do.  Solidity functions have the input parameters below the output parameters, with both going from the bottom of the stack upwards.  Yul functions also have the inputs below the outputs, but while the outputs go from bottom to top, the inputs go from top to bottom!  Obviously, this must be accounted for.  (Also, unlike in Solidity, we *do* have to handle outputs; the variable declaration code won't catch them for us in Yul.)

But there's one big catch that reappears to make things a pain to get right: those bare `let`s with no initializer.  What if the function starts with one of these, or even multiple of these?  Because they're skipped over, the top of the stack as we're about to leave the `YulFunctionDefinition` node will include them!  That'll cause all our parameters and return value to get allocated wrong.  We have to account for these!

So, OK, we therefore do account for these.  (Thankfully, it doesn't count if you put them, like, inside a block within the function, only if you put them directly inside the function at the beginning.)  Meaning this is the one case where we can in fact keep track of variables declared with a bare `let`: When they're declared at the beginning of a Yul function, before any other code, including `let`s with initializers.  The one case we can keep track of, is also the one case we *have* to keep track of in order to avoid screwing up our parameter allocations.

So that's how the expansions to the data saga work!  What a mess.

Oh, wait, one more caveat: Yul functions don't have jump markings in the source map.  This leads to obvious issues if one is called recursively.  But I don't really see anything I can do about that at the moment.

(My intention is to go bug the Solidity people about these source mapping issues, as well as the modifier depth issues I've talked about previously...)

Anyway, that's my quick, hacky implementation of assembly variables.  Hopefully it makes happy the people who asked about it, at least!


haltman-at:
Note: Another effect of this PR (not even intentional!) is to allow placing breakpoints inside assembly.  However this only currently works with Solidity 0.6.x; I'm going to go back and make it work so that breakpoints work inside assembly with older versions of Solidity as well.  Can't keep track of variables there, but I can at least make breakpoints work!

haltman-at:
OK, I've added a commit to this PR to improve breakpoint resolution; particularly as it relates to assembly (both old-style and new style), but also in general.  Also I added more tests of this functionality.

haltman-at:
OK, I've added one more commit; this one adjusts the Yul handling so it can handle how things will work once [this Solidity PR](https://github.com/ethereum/solidity/pull/8840) is merged.  Note I'm assuming this makes it into 0.6.8; if not we can adjust the threshold version there.

haltman-at:
OK, that Solidity PR has been merged, so 0.6.8 is now confirmed as the correct version to check for!

haltman-at:
OK, I've now actually tested this with actual 0.6.8 rather than prerelease 0.6.8 to confirm that it does indeed work.  I decided to leave the tests as they are on 0.6.6 though for now.

packages/debugger/lib/data/reducers.js
```diff
@@ -165,31 +247,59 @@ function assignments(state = DEFAULT_ASSIGNMENTS, action) {
...
+            ...acc,
+            byId: {
+              ...acc.byId,
+              [id]: assignment
+            },
+            byCompilationId: {
+              ...acc.byCompilationId,
+              [compilationId]: {
+                ...acc.byCompilationId[compilationId],
+                bySourceAndPointer: {
```
gnidan, line 288: As discussed, let's unify the state here to treat `sourceAndPointer` just like a regular opaque ID.

From that, we can unify the declare actions to take whichever info upfront (id vs. source/pointer)

haltman-at:
OK, I unified the handling of Solidity and Yul variables by replacing a lot of uses of `byAstId` or `bySourceAndPointer` with `byIdOrPath`.  But this did necessitate one substantial change -- rather than Yul variables having Solidity definitions spoofed, they now just get their actual Yul definitions, and `definitionToType` (or rather, the functions it calls) has been altered to just treat Yul definitions (node type `YulTypedName`) as type `bytes32`.

(Again, there's the question of whether `bytes32` is the correct type, but that's another matter.)

Also, I replaced all the manual string concatenation to make the source-and-pointers with a function `makePath`.

haltman-at:
Also the CircleCI build is failing but that's nothing to do with this PR in particular and @CruzMolina will fix that. :P

haltman-at:
OK, following discussion with @gnidan, I've gone and further renamed `idOrPath` to `astRef`.  I also added a comment in `data/reducers.js` explaining what this means because it's not very self-explanatory.  (Of course, neither is most of the debugger, but hey, easy to add this clarification for this new concept!)

PR #3019, opened 2020-05-01
Internal improvement: Get rid of more async library usage

eggplantzzz:
🅰️ This will be a breaking change for `@truffle/blockchain-utils`

### DIFF
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -272,22 +272,23 @@ const Utils = {
     { networks, currentProvider, setNetwork, network_id },
     gasLimit
   ) {
+    if (!networks && Object.keys(networks).length === 0) {
+      return false;
+    }
     // go through all the networks that are listed as
     // blockchain uris and see if they match
-    if (networks && Object.keys(networks).length !== 0) {
-      for (const network of networks) {
-        if (network.startsWith("blockchain://")) {
-          const networkMatches = await BlockchainUtils.matches(
-            network,
-            currentProvider
-          );
-          if (networkMatches) {
-            setNetwork(network);
-            return {
-              id: network_id,
-              blockLimit: gasLimit
-            };
-          }
+    for (const network in networks) {
+      if (network.startsWith("blockchain://")) {
+        const networkMatches = await BlockchainUtils.matches(
+          network,
+          currentProvider
+        );
+        if (networkMatches) {
+          setNetwork(network);
+          return {
+            id: network_id,
+            blockLimit: gasLimit
+          };
         }
       }
     }
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -121,28 +121,6 @@ const Utils = {
     return merged;
   },
 
-  parallel(arr, callback = () => {}) {
-    if (!arr.length) {
-      return callback(null, []);
-    }
-    let index = 0;
-    const results = new Array(arr.length);
-    arr.forEach((fn, position) => {
-      fn((err, result) => {
-        if (err) {
-          callback(err);
-          callback = () => {};
-        } else {
-          index++;
-          results[position] = result;
-          if (index >= arr.length) {
-            callback(null, results);
-          }
-        }
-      });
-    });
-  },
-
   linkBytecode(bytecode, links) {
     Object.keys(links).forEach(library_name => {
       const library_address = links[library_name];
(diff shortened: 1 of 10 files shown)


packages/core/test/npm.js
```diff
@@ -29,45 +27,36 @@ describe("NPM integration", function() {
...
-      ],
-      done
+    fse.ensureDirSync(fakeSourcePath);
+    fse.writeFileSync(path.join(fakeSourcePath, "Module.sol"), moduleSource, {
+      encoding: "utf8"
+    });
+    fse.writeFileSync(
+      path.join(fakeSourcePath, "ModuleDependency.sol"),
+      moduleDependencySource,
+      { encoding: "utf8" }
```
nicholasjpaterno, line 52: This is just a test so I wonder if it's necessary to remove async here since it can be a `devDep`?  Also, if we do choose to remove it do we want to fall back to synchronous operations?

packages/core/lib/package.js
```diff
@@ -210,169 +209,136 @@ const Package = {
...
-        .catch(error => {
-          if (callbackPassed) {
-            callback(error);
-          }
-          throw error;
-        });
+      });
     });
+
+    Promise.all(matchingPromises)
```
nicholasjpaterno, line 324: Since publishable artifacts is `async` now we can just await this `Promise.all` and simplify this whole block.

packages/core/lib/package.js
```diff
@@ -210,169 +209,136 @@ const Package = {
...
+    }
 
-      var promises = files.map(file => new Promise((resolve, reject) => {
-        fs.readFile(
-          path.join(options.contracts_build_directory, file),
-          "utf8",
-          function(err, body) {
-            if (err) return reject(err);
+    const contracts = files.map(file => {
+      const body = fs.readFileSync(
```
nicholasjpaterno, line 268: Do we want to do this synchronously?

packages/contract/lib/utils/index.js
```diff
@@ -290,38 +268,32 @@ const Utils = {
...
+    // blockchain uris and see if they match
+    const uris = Object.keys(networks).filter(
+      network => network.indexOf("blockchain://") === 0
+    );
 
-      Utils.parallel(matches, (err, results) => {
-        if (err) reject(err);
+    const results = [];
+    for (const uri of uris) {
+      results.push(await BlockchainUtils.matches(uri, currentProvider));
```
nicholasjpaterno, line 283: This can be parallelized.

packages/contract/lib/utils/index.js
```diff
@@ -290,38 +268,32 @@ const Utils = {
...
-      const uris = Object.keys(networks).filter(
-        network => network.indexOf("blockchain://") === 0
-      );
-      const matches = uris.map(uri =>
-        BlockchainUtils.matches.bind(BlockchainUtils, uri, currentProvider)
-      );
+    // go through all the networks that are listed as
+    // blockchain uris and see if they match
+    const uris = Object.keys(networks).filter(
+      network => network.indexOf("blockchain://") === 0
```
nicholasjpaterno, line 278: `startsWith` is faster/better.

packages/core/test/npm.js
eggplantzzz: I say remove it so we don't have to think about it any more. I don't think it will matter much in this case. I don't really think we will lose much making this synchronous.

packages/core/lib/package.js
eggplantzzz: This is another case where I don't think we gain much by having this be asynchronous. I don't really mind it being async but I like this for it's clarity rather than adding a bunch of junk to deal with asynchronous methods. There will be cases where the time saved will be meaningful but it seemed to me that it wouldn't matter much in this scenario. Feel free to disagree with me.

packages/core/test/npm.js
nicholasjpaterno: Fair enough!

Both operations depend on ensureDir so making that synchronous makes sense.  It would be easy enough to have both writes be asynchronous though.  Gains are negligible plus this is a test.

packages/core/lib/package.js
nicholasjpaterno: In this case we are looping through a list of contracts and doing file IO where subsequent IO doesn't depend on previous IO.  This just beckons for async - I would delegate this off to libuv and call it a day.

packages/core/lib/package.js
eggplantzzz: Yeah, sounds good to me!!

packages/contract/lib/utils/index.js
eggplantzzz: This I would argue we should just leave it for now until we need to optimize just for the sake of clarity. I'm not sure if we gain much speed-wise here. Do we?

nicholasjpaterno: I don't see the purpose in this entire loop otherwise! `parseKnownNetworks` is an async function so we can just `Promise.all(results)` or really just combine these two loops.

nicholasjpaterno: I don't see the purpose in this entire loop otherwise! parseKnownNetworks is an async function so we can just Promise.all(results) or really just combine these two loops.

eggplantzzz: I don't really follow what you are saying. What do you mean you don't see the purpose otherwise?  The purpose is to perform the check. Whether or not it is done in parallel is a separate issue. Or am I misunderstanding something?

packages/core/lib/package.js
eggplantzzz: Ok, changed back to an ugly asynchronous operation :)

packages/contract/lib/utils/index.js
nicholasjpaterno: It seems silly to loop through `uris` to collect an array of results to loop through again.  I think `await`ing defeats the purpose of looping through a first time, why not just combine these loops or collect an array of promises?

eggplantzzz: Oh I see what you mean, let me take a look and see if it is clear and makes sense to loop just once.

eggplantzzz: What do you think about that?

nicholasjpaterno:
But don't do that :rofl:

eggplantzzz:
So @nicholasjpaterno are we good to go? I know both approaches have their advantages and disadvantages but after consideration I am AOK with what we have in this PR. Waiting to hear back from you before moving this forward.

packages/contract/lib/utils/index.js
```diff
@@ -272,9 +272,12 @@ const Utils = {
     { networks, currentProvider, setNetwork, network_id },
     gasLimit
   ) {
+    if (!networks && Object.keys(networks).length === 0) {
```
nicholasjpaterno, line 275: This is one of the reasons I like `Map`'s over `Object`s.  It keeps track of its size so no need to count!

PR #3020, opened 2020-05-01
Bug fix: Fix file extension check in console startup

haltman-at:
The console on startup would look for artifacts that contained `.json` in the file name.  I've changed it to look for ones that end in `.json`.  This avoids it from picking up, say, Vim swapfiles.


eggplantzzz:
I didn't know that `endsWith` was a string method :)

PR #3021, opened 2020-05-01
Emit warnings when duplicate contract names are found during compilation

### LINKED ISSUE #387: Truffle should warn you if you have two contracts with the same name in different folders
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

I just spent a couple of hours trying to understand, why didn't the functions of my contract appear in an ABI after compilation.
The reason was - that I had two contracts with the same name in different folders.

## Steps to Reproduce

Create two different contracts with the same name in different folders

## Expected Behavior

Truffle would give you an error during compilation

## Actual Results

Compilation was successful and it just overrided the json artifacts of the previous contract

## Environment

* Operating System: OSX
* Truffle version: 3.2.1
* Ethereum client: N/A
* node version: v7.7.2
* npm version: N/A

KiChjang:
Fixes #387.

### DIFF
--- a/packages/artifactor/index.ts
+++ b/packages/artifactor/index.ts
@@ -2,11 +2,13 @@ import "source-map-support/register";
 import Schema from "@truffle/contract-schema";
 import fse from "fs-extra";
 import path from "path";
+import OS from "os";
 import { writeArtifact, finalizeArtifact } from "./utils";
 const debug = require("debug")("artifactor");
 
 class Artifactor {
   destination: string;
+
   constructor(destination: string) {
     this.destination = destination;
   }
@@ -47,6 +49,14 @@ class Artifactor {
       const tmpArtifactArray = artifactObjects;
 
       tmpArtifactArray.forEach(artifactObj => {
+        if (newArtifactObjects[artifactObj.contract_name]) {
+          console.warn(
+            `${OS.EOL}> Duplicate contract names found for ${
+              artifactObj.contract_name
+            }.${OS.EOL}` +
+              `> This can cause errors and unknown behavior. Please rename one of your contracts.`
+          );
+        }
         newArtifactObjects[artifactObj.contract_name] = artifactObj;
       });
     } else {
--- a/packages/artifactor/test/methods.js
+++ b/packages/artifactor/test/methods.js
@@ -1,4 +1,5 @@
 const assert = require("assert");
+const sinon = require("sinon");
 const Artifactor = require("../");
 
 describe("Artifactor.save", () => {
@@ -20,6 +21,40 @@ describe("Artifactor.save", () => {
 });
 
 describe("Artifactor.saveAll", () => {
+  it("throws if there are duplicate contract names", () => {
+    let consoleWarnSpy = sinon.spy(console, "warn");
+    artifactor = new Artifactor("/some/path");
+
+    artifactor
+      .saveAll([
+        {
+          "contractName": "Example",
+          "abi": [],
+          "bytecode": "0xabcdef",
+          "networks": {
+            3: { address: "0xe6e1652a0397e078f434d6dda181b218cfd42e01" }
+          },
+          "x-from-dependency": "somedep"
+        },
+        {
+          "contractName": "Example",
+          "abi": [],
+          "bytecode": "0xdeadbeef",
+          "networks": {
+            3: { address: "0xe6e1652a0397e078f434d6dda181b218cfd42e01" }
+          },
+          "x-from-dependency": "somedep"
+        }
+      ])
+      .then(() => {
+        assert.isTrue(consoleWarnSpy.called, "No warning emitted");
+        assert.isTrue(
+          consoleWarnSpy.calledWithMatch(/Duplicate contract names/),
+          "Wrong warning message"
+        );
+      });
+  });
+
   it("throws if this.destination doesn't exist", () => {
     artifactor = new Artifactor("/some/path");
 
--- a/packages/artifactor/test/methods.js
+++ b/packages/artifactor/test/methods.js
@@ -11,17 +11,17 @@ describe("Artifactor.save", () => {
         "abi": [],
         "bytecode": "0xabcdef",
         "networks": {
-          3: { address: "0xe6e1652a0397e078f434d6dda181b218cfd42e01" }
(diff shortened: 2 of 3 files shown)


KiChjang:
@gnidan Ok, I just pushed a new commit that moves the check from `@truffle/compile-solidity` to `@truffle/artifactor`, please take a look.

eggplantzzz:
I think this looks good! I wonder if we should do anything in the `else` here. I'm not sure what `artifactObject` would be if not an array. I don't think that should stop this PR from going through as I this seems like an improvement to me.

haltman-at:
Do we really want to throw an error here?  A lot of people seem to have duplicate contract names in their projects already.  This will *often* cause problems, but not *always*, so throwing an error and saying "no you can't Truffle for your project anymore" seems a bad idea, even if some things are going to work unreliably.  I think printing a warning would be a better idea.

eggplantzzz:
I wonder if it is likely that there could be bad behavior as a result of the contract name collision. It seems like it might be safer to throw an error. We could something in the direction of adding a `--dangerously-ignore-errors` kind of flag to ignore them. I'm not sure.

KiChjang:
I do believe that it's most likely a hidden bug when there are contracts that have duplicate names in your repository, so the default behavior should be to throw an error. Will explore the possibility of emitting a warning with a flag.

haltman-at:
Yes, there absolutely can be bad behavior as a result of name collisions.  It's just that (from what I've seen) name collisions are common enough that I think having Truffle simply refuse to work entirely when they occur (and note that they're often due to imported libraries rather than due to users naming their own contracts in a colliding manner) would be a really bad result.  The flag suggestion certainly does seem a solution to this.

KiChjang:
Need some help reviewing the last commit -- I'm not sure if I've managed to add the new flag to all places that matter, and I'm still unsure why the unit test is failing.

eggplantzzz:
What do you think @gnidan?

KiChjang:
@gnidan Hi there, would love to have a review on the latest commit, I'm sure I've done something inadequate/wrong there, so please have a look.

gnidan:
Sorry for the delay, will advise shortly!

KiChjang:
@gnidan Just checking up here to make sure you didn't forget about this PR.

eggplantzzz:
Hey @KiChjang, sorry to leave you hanging. Things have been busy here for a bit but now I think we have some more time.

So it is cool how you put this behind a feature flag but I think I would prefer this to be the default behavior for now. Showing a warning will probably help a lot of people save a lot of time debugging as I assume things get kind of wonky if you have 2 contracts of the same name.

KiChjang:
@eggplantzzz Ok, just to be sure, we still want that feature flag, but make it so that it warns by default, is that correct? Or do you prefer to do away with the feature flag altogether and just make it warn by default? If we're going for the former, then I'll need some guidance since my latest commit didn't add the feature flag logic at the appropriate places.

eggplantzzz:
So I would like for this to just warn when a user has contracts of the same name. That should be sufficient to let the user know that problems will probably arise and unknown behavior might occur. You can do away with the feature flags and just output a warning on compilation. I'll make some comments on the code itself. Thanks a lot @KiChjang!!

packages/artifactor/index.ts
```diff
@@ -7,8 +7,11 @@ const debug = require("debug")("artifactor");
 
 class Artifactor {
   destination: string;
-  constructor(destination: string) {
+  errorOnDuplicates: boolean;
```
eggplantzzz, line 10: Let's just turn this behavior on in general so you can get rid of the `errorOnDuplicates` in this class.

packages/artifactor/index.ts
```diff
@@ -47,6 +50,17 @@ class Artifactor {
       const tmpArtifactArray = artifactObjects;
 
       tmpArtifactArray.forEach(artifactObj => {
+        if (newArtifactObjects[artifactObj.contract_name]) {
+          if (this.errorOnDuplicates) {
+            throw new Error(
+              `Duplicate contract names found for ${artifactObj.contract_name}.`
+            );
+          } else {
+            console.warn(
```
eggplantzzz, line 59: Let's just make this call happen when a duplicate name is found. How would you feel about the following messaging and formatting?
```
`${OS.EOL}> Duplicate contract names found for ${artifactObj.contract_name}.${OS.EOL}` +
`> This can cause errors and unknown behavior. Please rename one of your contracts.`
```

eggplantzzz:
I like what you did with the feature flags but I think it will be sufficient and much cleaner if we just warn.

KiChjang:
@eggplantzzz Ok, I made the changes you requested, please re-review.

packages/artifactor/test/methods.js
```diff
@@ -20,6 +21,40 @@ describe("Artifactor.save", () => {
 });
 
 describe("Artifactor.saveAll", () => {
+  it("throws if there are duplicate contract names", () => {
```
eggplantzzz, line 24: Should be "warns if there are...."

eggplantzzz:
Hmm, so it's funny because I think that `artifactObjects` will normally be an object containing the artifacts. In other words, I unboxed the metacoin box, created a new contract named `MetaCoin`, and was unable to trigger the warning. This ended up being because `Array.isArray(artifactsObjects)` is false. So we'll have to figure out a solution for the `else` in that conditional statement. By that time, however, deduplication has already occurred since they are indexed by name. This here is probably still an improvement so maybe that work can be done in a separate PR. What do you think @gnidan?

packages/artifactor/test/methods.js
eggplantzzz: I went ahead and pushed this change because it was simple.

gnidan:
Looks like the issue is just that workflow-compile is squashing the list of artifacts into an object before the artifactor gets it. Upon conferring with @eggplantzzz on this, I think we can just remove the call to `byContractName` [here](https://github.com/trufflesuite/truffle/blob/develop/packages/workflow-compile/new/index.js#L113)

KiChjang:
Ok, rebasing after #3211 is merged.

eggplantzzz:
Rebase away!

vm06007:
I'm having this error, but there are no duplicates in my contract folder... why is it complaining about it? or how to silent it?

PR #3027, opened 2020-05-05
Bug fix: Fix (?) command in debugger

haltman-at:
The `?` command was broken if any watch expression were set due to the use of a non-lambda function.  Rather than switch to a lambda expression, I switched the `forEach` to a simple loop.  Now it works!


eggplantzzz:
This was a `this` binding issue?

haltman-at:
> This was a `this` binding issue?

Yup

PR #3032, opened 2020-05-06
Remove redundant join/split

nicholasjpaterno:
I don't know how often this code runs.  Seems unnecessary to
`.join` on line 142 only to split and rejoin again in `indent`.

`indent` now checks for the default case and should behave as before.
If the `split` is not default it will treat unindented as an array.
It will then rejoin with the `split` parameter.

### DIFF
--- a/packages/core/lib/testing/testrunner.js
+++ b/packages/core/lib/testing/testrunner.js
@@ -104,15 +104,16 @@ TestRunner.prototype.endTest = async function(mocha) {
     return;
   }
 
-  function indent(unindented, indentation, initialPrefix = "") {
-    return unindented
-      .split("\n")
-      .map((line, index) =>
-        index === 0
-          ? initialPrefix + " ".repeat(indentation - initialPrefix) + line
-          : " ".repeat(indentation) + line
+  function indent(unindented, indentation, initialPrefix = "", split = "\n") {
+    const dedent = split === "\n" ? unindented.split(split) : unindented;
+    return dedent
+      .map(
+        (line, index) =>
+          index === 0
+            ? initialPrefix + " ".repeat(indentation - initialPrefix) + line
+            : " ".repeat(indentation) + line
       )
-      .join("\n");
+      .join(split);
   }
 
   function printEvent(decoding, indentation = 0, initialPrefix = "") {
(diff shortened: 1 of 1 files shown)


haltman-at:
...huh, why weren't most of the tests run here...? (Not that it really affects this, but...)

nicholasjpaterno:
Good question..  I have no idea!

packages/core/lib/testing/testrunner.js
```diff
@@ -104,15 +104,16 @@ TestRunner.prototype.endTest = async function(mocha) {
...
 
-  function indent(unindented, indentation, initialPrefix = "") {
-    return unindented
-      .split("\n")
-      .map((line, index) =>
-        index === 0
-          ? initialPrefix + " ".repeat(indentation - initialPrefix) + line
-          : " ".repeat(indentation) + line
+  function indent(unindented, indentation, initialPrefix = "", split = "\n") {
+    const dedent = split === "\n" ? unindented.split(split) : unindented;
```
haltman-at, line 108: This condition is just kind of bizarre.  So setting `split = "\n"` is the secret trigger to cause a split?  Wouldn't it make more sense, say, to have `split = ""` be a trigger for turning *off* the split, or something?  Or hell maybe you just want a boolean flag or something.

packages/core/lib/testing/testrunner.js
```diff
@@ -104,15 +104,16 @@ TestRunner.prototype.endTest = async function(mocha) {
...
   }
 
-  function indent(unindented, indentation, initialPrefix = "") {
-    return unindented
-      .split("\n")
-      .map((line, index) =>
-        index === 0
-          ? initialPrefix + " ".repeat(indentation - initialPrefix) + line
-          : " ".repeat(indentation) + line
+  function indent(unindented, indentation, initialPrefix = "", split = "\n") {
```
haltman-at, line 107: This function is confusing now because `unindented` might now be one of two different types.  It could be a string *or* an array of strings.  If you're going to do that, I would suggest some sort of comment, and I would suggest some sort of comment explaining.

packages/core/lib/testing/testrunner.js
haltman-at: Actually, given that the input could be either an array or a string, why not go by that?  Get rid of the `split` argument altogether, and do a split only if the input is a string.

haltman-at: Btw, `dedent` looks like a synonym for `unindented`.  This isn't TypeScript, so I don't see a need for `const synonym = existingVariable`.  You could just mutate here: `unindented = typeof unindented === "string" : unindented.split() : unindented`.  Or, perhaps better, you could rename `dedent` to `unindented`, while also renaming the input from `unindented` to, I don't know, `input` or something.

packages/core/lib/testing/testrunner.js
haltman-at: Actually, if you make the changes I suggest below, a comment is probably unnecessary, because they'll make it clear enough what's going on.

packages/core/lib/testing/testrunner.js
nicholasjpaterno: The only question I have is if we get rid of the split, how do you want to handle the join at the end of the `indent` function?  Since that's passed in on line 145 in the case where its an array.

nicholasjpaterno: In the meantime i'll push up these changes and rename split to something sensible like `delimiter`.

haltman-at: Sorry, should've been more explicit there.  You can get rid of the `split` or `delimiter` variable because it's always `"\n"` (same as it was before).  This is an indenting function; we're not going to be joining with anything else.  This lets you get rid of the extra arguments that you added on one of the calls to it.

nicholasjpaterno: So, on line https://github.com/trufflesuite/truffle/pull/3032/files#diff-05d5edb3298db96eef76194fe62560dbL146 we dropped the `.join(",\n");` so we can pass `eventArgs` into `indent` as an array (so we can avoid the need to pattern match/split the string again).  But we would need to either hardcode or be able to pass that delimiter in so we can rejoin it properly on line https://github.com/trufflesuite/truffle/pull/3032/files#diff-05d5edb3298db96eef76194fe62560dbR119

haltman-at: Oh!  Now I get why you did that.  Yeah, I see.  That makes sense.  I totally missed the comma thing before, sorry.  Current resolution is still ugly though.  What if you just included the commas in `eventArgs`?  Could either omit it from the last one with the `index` and `array` arguments to the `map` callback, or could just manually remove it afterward by doing a `.slice(0,-1)` on the last element.  (Latter personally seems simpler to me? <shrug>)  Then we can finally get rid of these extra arguments! :)

nicholasjpaterno: Sounds good to me!

haltman-at: Another possibility that occurs to me, if you really prefer the delimiter argument, would be to have the delimiter argument but join on `delimiter + "\n"`, with `delimiter` defaulting to `""`.  That makes more sense to me since it's an indenting function -- there should always be a linebreak!  But I think the other way I mentioned is probably better...

haltman-at:
Hey @nicholasjpaterno would you mind if I just finished this myself and got this merged? :)

packages/core/lib/testing/testrunner.js
nicholasjpaterno: Ah, that makes sense too.  Ok, ill go with the original!

Issue #3023, opened 2020-05-03
Test fails to use new contract addresses if evm_revert isn't supported.


no2chem:
## Issue

When running `truffle test` against a client that doesn't support `evm_revert`, Truffle will try to reset by redeploying the contracts between test suites.

https://github.com/trufflesuite/truffle/blob/0effb767df26dca7aab0b1247de30fdf5b127be9/packages/core/lib/testing/testrunner.js#L81 

However, contracts referred to by the new test suite (via `artifact.require`) point to the previously deployed contracts instead of the new ones. This causes issues for tests which expect the state of the contracts to be reset in the new test suite.

As far as I can tell by debugging through this, even though the contracts are redeployed in testrunner::deploy:

https://github.com/trufflesuite/truffle/blob/0effb767df26dca7aab0b1247de30fdf5b127be9/packages/core/lib/testing/testrunner.js#L72

The artifacts are loaded when mocha *first* loads the test files, right after the inital deploy. testrunner::deploy doesn't reload the artifacts, leading to the problem. The new redeployed contract addresses need to be reinjected to the already loaded contracts somehow.

## Steps to Reproduce

On a client the doesn't support `evm_revert` run a minimal test suite:

contracts/Increment.sol
```sol
pragma solidity >=0.5.0;

contract Increment {
    uint256 count;
    
    function increment() public {
        count++;
    }
    
    function getCount() view public returns (uint256) {
        return count;
    }
```

tests/test1.js
```js
const Increment = artifacts.require("Increment");

contract("First test", async accounts => {
     it("should increment to one", async() => {
            let deployed = await Increment.deployed();
            await deployed.increment.sendTransaction();
            console.log(`Address: ${deployed.address}`);
            expect(await deployed.getCount()).to.equal(1);
     });
});
```
tests/test2.js
```js
const Increment = artifacts.require("Increment");

contract("Second test", async accounts => {
     it("should increment to one", async() => {
            let deployed = await Increment.deployed();
            await deployed.increment.sendTransaction();
            // ISSUE 1: Address is the same as previous suite!
            console.log(`Address: ${deployed.address}`);
            // ISSUE 2: TEST FAILURE - returns 2 instead of 1 due to previous state
            expect(await deployed.getCount()).to.equal(1);
     });
});
```

## Expected Behavior

The second test suite results in the issues described above.

## Actual Results

The second test suite runs with the previously deployed contracts as the first test suite, causing the issues described above.

## Environment

[...]

PR #3047, opened 2020-05-12
Reset test state before top level suite .js is called

no2chem:
This PR fixes #2375 and #3023.

It changes the test flow so that test suites are loaded one at a time and `runner.initialize()` is called *before* each test file is loaded. This fixes the issues above, where redeployment (on clients that do not support snapshots) occurs *after* the contract artifacts are loaded. 

There might be a better way to do this with Mocha by hooking into the `suite.on('pre-require')` event, but I couldn't figure out how to do and wait for the async `runner.initialize()` function.

### DIFF
--- a/packages/core/lib/test.js
+++ b/packages/core/lib/test.js
@@ -69,7 +69,7 @@ const Test = {
       }
     };
 
-    const mocha = this.createMocha(config);
+    const solidityMocha = this.createMocha(config);
 
     const jsTests = config.test_files.filter(file => {
       return path.extname(file) !== ".sol";
@@ -77,17 +77,6 @@ const Test = {
 
     const solTests = config.test_files.filter(file => {
       return path.extname(file) === ".sol";
-    });
-
-    // Add Javascript tests because there's nothing we need to do with them.
-    // Solidity tests will be handled later.
-    jsTests.forEach(file => {
-      // There's an idiosyncracy in Mocha where the same file can't be run twice
-      // unless we delete the `require` cache.
-      // https://github.com/mochajs/mocha/issues/995
-      delete originalrequire.cache[file];
-
-      mocha.addFile(file);
     });
 
     const accounts = await this.getAccounts(interfaceAdapter);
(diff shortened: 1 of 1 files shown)


no2chem:
Well, it seems that this set fails some tests because it breaks the test suites into multiple sets. Some tests explicitly search for the string "5 passing" when the output is now "3 passing" and "2 passing". Looking for a way to keep the original behavior.

gnidan:
Neat @no2chem! Let us know if you're able to recreate the existing behavior.

If not, we could consider this as a breaking change for Truffle v6 – what are the benefits of your approach?

no2chem:
@gnidan I think it'd be better to figure out how to keep the existing behavior. I *think* that this just requires figuring out how to use the Mocha API better, which I'm working on at the moment. Please bear with me as I'm in the process of trying to figure out what the testrunner classes do :).

gnidan:
For sure @no2chem. That part of the code is a mess.

If it helps: TestResolver and TestSource belong (conceptually) as part of @truffle/resolver. Compare TestSource with, e.g., the [FS source](https://github.com/trufflesuite/truffle/blob/develop/packages/resolver/lib/sources/fs.ts).

Feel free to ask any questions!

no2chem:
@gnidan Well, looked into it and it seems more difficult than I thought to get Mocha to do what I want.

I'm thinking of taking an alternative approach now. Is there a way to get artifacts to forcefully re-load the deployed address of the contract from the json file? This would solve the issue, I think.

gnidan:
Just spoke with @haltman-at and we came up with an idea to use `.deployed()` as the place to do the refresh. I think @haltman-at will be taking a stab at implementing this tonight?

gnidan:
Alright @no2chem! See https://github.com/trufflesuite/truffle/pull/3053, does that help you with this?

haltman-at:
More specifically, can we close this as superseded now that #3053 has been merged? :)  (It should solve your problem... it certainly has solved it in my tests.)

no2chem:
Excellent, happy to hear. Yes, that does seem like it would resolve the issue. Closing this PR in favor of that one!

PR #3049, opened 2020-05-13
Enhancement: Have debugger print return values on successful completion

haltman-at:
This PR adds a `returnValue` method to the `Session` class.  Running this method once the transaction is complete will decode the return value (or revert message).  This PR also adds code in the CLI to print the return value; it will print on successful completion.  It will also print if you hit `v` once the transaction is done, alongside the other variables.  Decoding of revert strings is now done this way also -- the CLI no longer invokes `Codec`'s `decodeRevert` to do the decoding itself, but just uses `session.returnValue()`.  If the transaction reverted, hitting `v` once the transaction is complete will now the revert string as well as the variables.

The debugger side of this is pretty straightforward.  We now need to allocate calldata as well, so we do.  We pass some info to the allocator that we weren't before in order to make this possible.  We determine what allocation to use by looking at the current context and (if we're not in a constructor) reading the current selector out of the calldata.  If it doesn't match any function (such as if we're in a fallback function), then, oh well, guess we're only decoding reverts.  Also, note that we do pass the return status to `Codec.decodeReturndata` so it can distinguish between successful returns and reverts.

The CLI has some fairly detailed logic to make sure that what we print makes sense.  For instance, if there's no return value, but none was expected, we don't print anything at all.  Etc.  I'm not going to list all the cases here.

(For decoding bytecode, it would be really nice if we could include the linked libraries and immutables in the decoding, but the former is a problem due to our old bytecode format, and the latter... might be doable but I haven't looked into it yet.  Not going to consider this high-priority though.)

Anyway, this was largely pretty straightforward, so I don't have a lot to say here.  Um, I factored out the code-requesting part of `decode` into its own saga, for obvious reasons.  I guess that's worth noting.

Btw, that self-destruct test that I removed because it doesn't work, does work when I do it manually, but it seems web3 poses a problem when I try to do it in automated tests, and for whatever reason, even the usual try/catch didn't work here... (I guess probably because this is purely a web3 error, rather than Ganache's default mode throwing an error like usual?)

Also, I fixed a bug in the EVM selectors -- they couldn't read past the end of memory before.  Now they can.

### DIFF
--- a/packages/core/lib/debug/interpreter.js
+++ b/packages/core/lib/debug/interpreter.js
@@ -431,12 +431,13 @@ class DebugInterpreter {
       this.printer.print("");
       //check if transaction failed
       if (!this.session.view(evm.transaction.status)) {
-        this.printer.printRevertMessage();
+        await this.printer.printRevertMessage();
         this.printer.print("");
         this.printer.printStacktrace(true); //final stacktrace
       } else {
         //case if transaction succeeded
         this.printer.print("Transaction completed successfully.");
+        await this.printer.printReturnValue();
       }
     }
 
@@ -462,6 +463,9 @@ class DebugInterpreter {
         break;
       case "v":
         await this.printer.printVariables();
+        if (this.session.view(trace.finished)) {
+          await this.printer.printReturnValue();
+        }
         break;
       case ":":
         watchExpressionAnalytics(cmdArgs);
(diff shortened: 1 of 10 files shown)


packages/core/lib/debug/interpreter.js
```diff
@@ -431,12 +431,13 @@ class DebugInterpreter {
       this.printer.print("");
       //check if transaction failed
       if (!this.session.view(evm.transaction.status)) {
```
nicholasjpaterno, line 433: Since there's an `else` we can just invert these blocks and remove the `!`.

Issue #3051, opened 2020-05-14
SPDX license identifiers missing


wbt:
As of today (Solidity 0.6.8, from [this PR](https://github.com/ethereum/solidity/pull/8907)), Solidity issues warnings on Migrations.sol (and other) contracts for lack of an "[SPDX-License-Identfier](https://solidity.readthedocs.io/en/v0.6.8/layout-of-source-files.html#spdx-license-identifier)" code comment at the top of the file. 

A proposed fix would be to add
`//SPDX-License-Identifier: MIT`
to the start of Migrations.sol and every other contract file.

haltman-at:
Thanks for pointing this out.  I've put up a PR to address the cases in this repo, but `truffle-box` will have to be addressed separately.

haltman-at:
OK, since this should be fixed on this repo, and you've opened a separate issue on `truffle-box`, I'm going to just go ahead and close this if you don't mind.

PR #3052, opened 2020-05-14
Add SPDX license identifiers to (non-test) contracts

haltman-at:
Partial fix for #3051.  Adds SPDX license identifiers to all contracts we make available to users.  Left our own internal testing contracts alone.  Note (@cds-amal?) we'll also need to get the ones in `truffle-box`.


haltman-at:
Oh, I wasn't aware of that!  OK, I'll go back and do that.

PR #3059, opened 2020-05-18
Separate compilation record-keeping in Truffle DB loaders

gnidan:
This PR refactors the artifactsLoader to expose `db.loadCompilations(workflowCompileResult)` as a standalone method. Previously, the artifactsLoader only offered the capability of forcing a recompilation and collecting all information in one fell swoop. Now, the artifactsLoader defers to `db.loadCompilations()`. This method is intended for use inside the rest of Truffle also, so that the artifactsLoader stops being the only way to load data into Truffle DB.

For reviewing this, probably best to step through the individual commits. Notes:
- The first four commits break specific resources out from the artifactsLoader and incorporate them into the new loaders framework
- The last (fifth) commit adds a pre-processing step at the very beginning, to structure the workflow-compile result into a form that more natively matches the structure we're looking for.

Let me know if you have questions, and I'll try to expand on this description with additional context!

Oh! Forgot to mention: @cds-amal was instrumental, helping me put this together several months ago when I started it 🙏

### DIFF
--- a/packages/db/src/db.ts
+++ b/packages/db/src/db.ts
@@ -49,7 +49,9 @@ export class TruffleDB {
   }
 
   async loadCompilations(result: WorkflowCompileResult) {
-    const saga = generateCompileLoad(result);
+    const saga = generateCompileLoad(result, {
+      directory: this.context.workingDirectory
+    });
 
     let cur = saga.next();
     while (!cur.done) {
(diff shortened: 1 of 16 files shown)


fainashalts:
LGTM! I really like the way this is set up now, and I think it makes Truffle DB more modular and clear. 🎉 Really nice to see generator functions in the wild, too.

gnidan:
Thanks @fainashalts!

> Really nice to see generator functions in the wild, too.

Aren't generators cool? :)

PR #3063, opened 2020-05-20
Re-organize Truffle DB loaders module

gnidan:
This work paves the way to define individual loading methods for `truffle compile`, `truffle migrate`, etc. to use. (For reference, the loaders are currently hooked up to run entirely in a single sequence, triggered by a request to a GraphQL endpoint). I've spun this PR out of #3059, which begins the actual splitting up of these workflows.

- Group loaders into submodules:
    - `loaders/schema`: for the ArtifactsLoader and GraphQL hookup
    - `loaders/resources`: for GraphQL logic (incl. queries/mutations)
    - `loaders/commands`: for dedicated compile/migrate/etc. workflows

- Organize individual GraphQL query/mutation definitions into individual
  resource submodules and use *.graphql.ts suffix

- Rename AddCompilation -> AddCompilations


packages/db/src/loaders/schema/index.ts
```diff
@@ -0,0 +1,55 @@
+import { TruffleDB } from "@truffle/db";
+import { ArtifactsLoader } from "./artifactsLoader";
+export { ArtifactsLoader };
+import { schema as rootSchema } from "@truffle/db/schema";
+import { Workspace, schema } from "@truffle/db/workspace";
+const tmp = require("tmp");
+import { makeExecutableSchema } from "@gnd/graphql-tools";
+import { gql } from "apollo-server";
+
+//dummy query here because of known issue with Apollo mutation-only schemas
```
cds-amal, line 10: Not a blocker but a link to the issue would be helpful for our future selves.

gnidan: my present self doesn't remember what this is about. was this even me? @fainashalts was this you?

gnidan: I remove the comment in the meantime. If we come up with a source, we can add it back in... might be helpful!

gnidan:
Thanks! Will merge when green :)

PR #3064, opened 2020-05-20
Permit querying compilations for related contracts in Truffle DB

gnidan:
This PR does two things:

- Rename `SourceContract` -> `ProcessedSource`. The former name turns out to be incorrect, since sources can have multiple contracts, so "processed source" more generally represents the idea here. As a consequence of this change, also remove the `name` field from that object, since it becomes meaningless.

- Add a `contracts` field to ProcessedSource for compilations. This allows the retrieval of all related contracts given only a compilation ID.

Note: it'd be ideal if processed sources could also represent other top-levels, like structs and enums, but (a) that's out of scope for an initial implementation, and (b) those aren't even exposed by Solidity (alas)

### DIFF
--- a/packages/db/src/artifacts/schema.ts
+++ b/packages/db/src/artifacts/schema.ts
@@ -126,19 +126,18 @@ export const schema = mergeSchemas({
       name: {
         fragment: `... on ContractObject { name: contractName }`
       },
-      sourceContract: {
+      processedSource: {
         fragment: `... on ContractObject {
           ast { json }
           source { contents, sourcePath }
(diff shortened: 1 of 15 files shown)


gnidan:
(Oh, I just remembered I need to update the data-model. I'll link the PR here when I do that)

gnidan:
[Here](https://github.com/trufflesuite/artifact-updates/pull/25)'s the corresponding data model changes... merging this PR to get things through the queue. Thanks @fainashalts!

PR #3069, opened 2020-05-21
Bug fix: Avoid lookbehind in AST utils regexp

### LINKED ISSUE #3068: Firefox compatibility
## Issue

Using `@truffle/contract` `v4.2.5` breaks the app on Firefox.
The console throws `SintaxError: invalid regexp group` error.
It looks like there is a regex used that is not supported on Firefox yet (https://caniuse.com/#feat=js-regexp-lookbehind).

## Steps to Reproduce

```
  "dependencies": {
    "@truffle/contract": "^4.2.5",
    "@truffle/debug-utils": "^4.1.3",
    "react": "^16.13.1",
    "react-dom": "^16.13.1",
    "react-router-dom": "^5.2.0",
    "react-scripts": "3.4.1",
    "typescript": "^3.9.3",
    "web3": "^1.2.7"
  },

const contract = require('@truffle/contract')
```

## Expected Behavior

Should work as in Chrome browser for example.

## Actual Results

The app breaks and fails to render on Firefox browser.
Console error:
```SyntaxError: invalid regexp group```

Screen Shot 2020-05-21 at 11 21 41

## Environment

* Operating System: OSX Catalina
* Ethereum client: web3 v1.2.7
* node version (`node --version`): v10.20.1
* npm version (`npm --version`): v6.14.4

### PR DESCRIPTION
**Pull request by haltman-at**
Ought to fix #3068.  See, `@truffle/contract` imports `@truffle/debug-utils` which imports `@truffle/codec`, and `@truffle/codec` used a lookbehind regexp in its AST-handling utilities, so, things broke in Firefox.

So, instead of using lookbehind, that `string.replace` call now uses a parameterized replacement instead.  It's functionally identical.  Frigging Firefox...

It does bother me that this came up at all -- like, geez, even `truffle-contract.min.js` has that code in it?  Why?  I mean, it's not using it, right?  I would hope that some part of browserify's or uglify's compilation process would strip out unused code, that `@truffle/contract` doesn't end up with literally the entirety of every dependency, direct or indirect, inside of it!  Because, uh, geez, wow, I guess I may have been incautious with adding dependencies there.

Oh well.  We can worry about that later; this is is a quick fix that ought to get things working in Firefox again.

Also, while I was at it, I fixed the fact that `exorcist` was listed as a dependency of `@truffle/contract` rather than a devDependency.

### CODE REVIEW

File: packages/codec/lib/ast/utils.ts [hunk 1]
```diff
@@ -265,8 +265,8 @@ export function spliceLocation(
  */
 export function regularizeTypeIdentifier(identifier: string): string {
   return identifier.replace(
-    /(?<=_(storage|memory|calldata))((_slice)?_ptr)?$/,
-    "_ptr"
+    /(_(storage|memory|calldata))((_slice)?_ptr)?$/,
+    "$1_ptr" //this used to use lookbehind for clarity, but Firefox...
```
gnidan, line 269: Probably worth including a link to the original issue here

PR #3070, opened 2020-05-21
Some minor internal improvements

haltman-at:
Just some assorted cleanup:

1. The decoder tests imported `@truffle/codec` in a stupid way.  Now they do it the right way.
2. Cleaned up the invocation of typescript in various packages: added `--declaration`, added a `types` field to the JSON where there wasn't one (hoping I did this right?), removed unnecessary `node_modules/.bin/`.


packages/artifactor/package.json
```diff
@@ -7,8 +7,9 @@
   "repository": "https://github.com/trufflesuite/truffle/tree/master/packages/artifactor",
   "version": "4.0.60",
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "scripts": {
-    "build": "tsc",
+    "build": "tsc --declaration",
```
gnidan, line 12: Shouldn't this go in the tsconfig?

haltman-at: Oh, huh, didn't know you could do that.  Will change that.

gnidan: You'll find it's already there in a few places!

haltman-at:
OK, got rid of `--declaration` in favor of turning on `"declaration": true` in the tsconfig.

PR #3075, opened 2020-05-26
Fix incorrect return value on Windows path resolver

### LINKED ISSUE #2834: Cannot destructure property `body` of 'undefined' or 'null' (5.1.10 to any later version)
- [ X ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue. 
Didn't see anything there. 
I didn't see anything in the release notes that called anything out. 
Looks like another related issue here: 
https://github.com/trufflesuite/truffle/issues/1574 and https://github.com/trufflesuite/truffle/issues/1365 and https://github.com/trufflesuite/truffle/issues/991 and https://github.com/trufflesuite/truffle/issues/1936
with zero info.

I've been using truffle with the same contract consistently for a while now (6 months easily). This latest upgrade 5.1.10 to 5.1.13 (I missed one) is the first time I've seen this error. Wondering if anyone else has.

## Solution

Downgrading to 5.1.10 allows me to  truffle migrate --network development without issue.

---------------------------

## Issue

I've been using the same contract for 6+ months. After upgrading from 5.1.10 to 5.1.1x , I started getting this error when attempting to launch my contract on any network:

**Cannot destructure property `body` of 'undefined' or 'null'**

Downgrading to 5.1.10 allows me to launch my contract again. Worked my way down from 5.1.13 to 5.1.10.

## Steps to Reproduce

truffle migrate --network development
or
truffle migrate --network ropsten

## Expected Behavior

My contract would compile and launch as expected.

## Actual Results

√ Fetching solc version list from solc-bin. Attempt #1
Error: TypeError: Error parsing C:/Users/Mine/Projects/project/contracts/Contract.sol: Cannot destructure property `body` of 'undefined' or 'null'.
    at Object.compile (C:\Users\Mine\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\workflow-compile\legacy\index.js:80:1)

## Environment

* Operating System: Windows
* Ethereum client:
* Truffle version (`truffle version`): 5.1.13
* node version (`node --version`): 12.15.0
* npm version (`npm --version`): 6.13.7

gorgos:
Fixes #2834

### DIFF
--- a/packages/resolver/lib/sources/fs.ts
+++ b/packages/resolver/lib/sources/fs.ts
@@ -65,9 +65,8 @@ export class FS implements ResolverSource {
     ];
 
     let body, filePath;
-    for (let i = 0; i < possiblePaths.length; i++) {
+    for (const possiblePath of possiblePaths) {
       try {
-        const possiblePath = possiblePaths[i];
         const resolvedSource = fs.readFileSync(possiblePath, {
           encoding: "utf8"
         });
@@ -78,7 +77,7 @@ export class FS implements ResolverSource {
       } catch (error) {
         // do nothing
       }
-    });
+    }
     
     throw new Error(`Path resolving failed for ${importPath}`);
   }
--- a/packages/resolver/lib/sources/fs.ts
+++ b/packages/resolver/lib/sources/fs.ts
@@ -65,18 +65,22 @@ export class FS implements ResolverSource {
     ];
 
     let body, filePath;
-    possiblePaths.forEach(possiblePath => {
+    for (let i = 0; i < possiblePaths.length; i++) {
       try {
+        const possiblePath = possiblePaths[i];
         const resolvedSource = fs.readFileSync(possiblePath, {
           encoding: "utf8"
         });
         body = resolvedSource;
         filePath = possiblePath;
+        
+        return { body, filePath };
       } catch (error) {
         // do nothing
       }
     });
-    return { body, filePath };
+    
+    throw new Error(`Path resolving failed for ${importPath}`);
   }
 
   // Here we're resolving from local files to local files, all absolute.
--- a/packages/resolver/lib/sources/fs.ts
+++ b/packages/resolver/lib/sources/fs.ts
@@ -79,7 +79,7 @@ export class FS implements ResolverSource {
       }
     }
     
-    throw new Error(`Path resolving failed for ${importPath}`);
+    return { body, filePath };
   }
 
   // Here we're resolving from local files to local files, all absolute.


packages/resolver/lib/sources/fs.ts
```diff
@@ -65,18 +65,22 @@ export class FS implements ResolverSource {
     ];
 
     let body, filePath;
-    possiblePaths.forEach(possiblePath => {
+    for (let i = 0; i < possiblePaths.length; i++) {
       try {
+        const possiblePath = possiblePaths[i];
```
gnidan, line 70: Looks like we can just use `for (const ... of ...)` syntax here (it's cleaner)

```suggestion
    for (const possiblePath of possiblePaths) {
      try {
```

packages/resolver/lib/sources/fs.ts
```diff
@@ -65,18 +65,22 @@ export class FS implements ResolverSource {
...
           encoding: "utf8"
         });
         body = resolvedSource;
         filePath = possiblePath;
+        
+        return { body, filePath };
       } catch (error) {
         // do nothing
       }
     });
```
gnidan, line 81: Looks like the tests are failing cause of this syntax error

```suggestion
    }
```

gorgos:
@gnidan Thanks for fixing it.

https://github.com/trufflesuite/truffle/blob/a0bbfcce1d351fdc5f865d17760a63459f8dd191/packages/resolver/lib/sources/fs.ts#L64

resolves to a path with `C:\\this\\is\\the\\path`. While https://github.com/trufflesuite/truffle/blob/a0bbfcce1d351fdc5f865d17760a63459f8dd191/packages/resolver/lib/sources/fs.ts#L63 resolves to `C:/this/is/the/path`. The change in https://github.com/trufflesuite/truffle/pull/2693/files#diff-741d320df0fc163eac3fec99a9e78b75R65 meant that instead of the first path, always the last will be returned.

(later on then there is a mismatch in https://github.com/trufflesuite/truffle/blob/a0bbfcce1d351fdc5f865d17760a63459f8dd191/packages/compile-solidity/profiler/index.js#L147 because `resolved` contains with paths with `/` but `currentFile` will have `\\`.

Possibly all related to Git Bash / Mingw64.

eggplantzzz:
Hey @gorgos, looks like there is some error message-related failure in the test suite. Can you make sure you are getting the behavior you expect and update the test?

gnidan:
> always the last will be returned.

I'm not sure this is the right solution... is there some way to fix the paths in the first place, so we can preserve existing behavior but not deal with Windows path weirdness?

gorgos:
@gnidan I had a look again, it comes from this line https://github.com/trufflesuite/truffle/blob/1ead25b2ec88002639354cad8a88046830e32bd8/packages/resolver/lib/sources/fs.ts#L64

`path.dirname(importedFrom)` resolves to the paths with `\\`.

gorgos:
@gnidan How do you want to proceed?

packages/resolver/lib/sources/fs.ts
```diff
@@ -65,18 +65,21 @@ export class FS implements ResolverSource {
...
-    possiblePaths.forEach(possiblePath => {
+    for (const possiblePath of possiblePaths) {
       try {
         const resolvedSource = fs.readFileSync(possiblePath, {
           encoding: "utf8"
         });
         body = resolvedSource;
         filePath = possiblePath;
+        
+        return { body, filePath };
```
CruzMolina, line 76: @gorgos , in terms of making this fix as clean as possible and getting it out the door, it looks like this line is all that needs to be added.

packages/resolver/lib/sources/fs.ts
```diff
@@ -65,18 +65,21 @@ export class FS implements ResolverSource {
...
       } catch (error) {
         // do nothing
       }
-    });
-    return { body, filePath };
+    }
+    
+    throw new Error(`Path resolving failed for ${importPath}`);
```
CruzMolina, line 82: This might be fine, but based on the test case failing in the `PACKAGES` CI job, it appears there's already error handling for the fallback (which expects an empty object returned). You _could_ update the test case assertion (and possibly others that may need to be updated), but it would be cleaner to just stay with the current fallback code.

packages/resolver/lib/sources/fs.ts
gorgos: Well, and the change to `for .. of` loop.

packages/resolver/lib/sources/fs.ts
gorgos, line 82:
```suggestion
    return { body, filePath };
```

sfefinance:
PS F:\wo_de_shi_ye\test_truffle\aave> truffle compile       
TypeError: Error parsing F:/wo_de_shi_ye/test_truffle/aave/contracts/Flashloan.sol: Cannot read property 'addFunction' of undefined
    at runWithReadCallback (C:\Users\a'su's\AppData\Roaming\npm\node_modules\truffle\build\webpack:\~\solc\wrapper.js:37:1)
    at compileStandard (C:\Users\a'su's\AppData\Roaming\npm\node_modules\truffle\build\webpack:\~\solc\wrapper.js:78:1)
    at Object.compileStandardWrapper (C:\Users\a'su's\AppData\Roaming\npm\node_modules\truffle\build\webpack:\~\solc\wrapper.js:85:1)
    at Object.parseImports (C:\Users\a'su's\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\truffle-compile\parser.js:47:1)
    at Object.getImports (C:\Users\a'su's\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\truffle-compile\profiler.js:374:1)
    at C:\Users\a'su's\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\truffle-compile\profiler.js:346:1
Truffle v5.0.1 (core: 5.0.1)
Node v14.17.3

Issue #2688, opened 2019-12-15
Upgrading Web3 >= 1.2.2


cgewecke:
This is a tracking issue for changes necessary at Truffle to upgrade Web3 above 1.2.1. 

**Resolve #2654**

In #2654, users with long-running test suites saw performance reduction and random ganache disconnection in Truffle versions which use Web3 1.2.2. 

Based on experiments with some of the projects reporting the bug,  I think the underlying issue is the way that truffle/contract manages confirmations handling in combination with a fix in Web3 1.2.2 for broken behavior in that feature.

In Web3 1.2.1 and below, confirmations over http (incorrectly) fire once per second, 24 times, without querying the client.  1.2.2 fixes this by block polling and emitting if  **a new block has been mined**. ([See code][2]). Note that polling only occurs if a conf handler is attached. The number of confirmations the handler should wait for is [configurable][4]. 

Truffle contract automatically attaches confirmation handlers for all transactions here:

https://github.com/trufflesuite/truffle/blob/99add8f5e47586030a44a999bcfb287224e304e1/packages/contract/lib/handlers.js#L44

This means that in a unit test context with thousands of txs, each one will incur a large number of additional background block polling requests. In addition, truffle `evm_reverts` between the tests suites, affecting the block number values relevant for polling. It looks like the combination of these swamps ganache with unnecessary calls, slows things down a lot, and creates instability.

One solution might be to disable confirmation handling in unit tests (where it's rarely used). It could be opt-in.

[2]: https://github.com/ethereum/web3.js/blob/1.x/packages/web3-core-method/src/index.js#L295-L317
[4]: https://web3js.readthedocs.io/en/v1.2.4/web3-eth.html?highlight=confirmations#transactionconfirmationblocks

**Resolve #2657**

#2657 has TS  errors in the HDWalletProvider unit tests. These result from HDWalletProvider not implementing some parts of the Web3 HTTPProvider interface. There are notes about the interface in [2657 comment][3].

[3]: https://github.com/trufflesuite/truffle/pull/2657#issuecomment-563010729

If the above seems unclear / incorrect / incomplete or there's anything that could be fixed at Web3 to help with this, please just ping. 

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`):
* node version (`node --version`):
* npm version (`npm --version`):

nogo10:
A few packages have not been upgraded

-- truffle-config@1.1.16
     `-- truffle-provider@0.1.16
[...]

PR #3078, opened 2020-05-27
Bump web3 1.2.1 => 1.2.5

wbt:
While not yet tested, this appears to be the minimum change needed to incorporate [this commit](https://github.com/ethereum/web3.js/commit/6432b70f7d66c0cc5a18cead7afa7d45e5bc4301) and be able to use the 'genesis' predefined block number.  If web3 did not break semver again, this should work.  However I recognize from past experience that's a questionable assumption.

### DIFF
--- a/packages/interface-adapter/package.json
+++ b/packages/interface-adapter/package.json
@@ -23,7 +23,7 @@
     "bn.js": "^4.11.8",
     "ethers": "^4.0.32",
     "source-map-support": "^0.5.19",
-    "web3": "1.2.1"
+    "web3": "1.2.5"
   },
   "devDependencies": {
     "@truffle/provider": "^0.2.11",


gnidan:
The types!

haltman-at:
Hm, if this can be made to work (types adjusted appropriately) then that's auspicious for upgrading it in our other packages as well... :)

haltman-at:
Hi, were you planning to fix the types and make this work, or should we discard it?  Thank you!  (Another alternative that may suffice to get this working may be trying a later version of web3, rather than 1.2.5 specifically, as later versions may break less.)

wbt:
I am not sufficiently skilled in Typescript to make this fast and easy at the moment.  However, in the spirit of collaborative open-source software development, I would invite anyone with available skill and bandwidth to do so. 

I would also support trying a later version of web3 so long as it doesn't break semver.

gnidan:
Hey @wbt, since this PR seems to be going nowhere fast, I propose we close this and use [this issue](https://github.com/trufflesuite/truffle/issues/2688) to track the upgrade effort. Let me know if you think there's value in keeping this open!

Thanks again!

gnidan:
Closing this for issue maintenance, then :)

Until the next attempt!

PR #3080, opened 2020-05-29
Bug fix: Make debugger robust against lack of trace steps and add warnings

haltman-at:
This PR fixes various errors that would come up if the transaction had no trace steps.  Some of these bugs were actually relevant even when there were trace steps, oops!

It also makes some changes to the CLI: it makes it so that `p` when there are no trace steps prints a message about this rather than nonsense, and it also adds warning messages on startup (or reset or transaction load) if the transaction has no trace steps.

### DIFF
--- a/packages/core/lib/debug/interpreter.js
+++ b/packages/core/lib/debug/interpreter.js
@@ -503,9 +503,14 @@ class DebugInterpreter {
           temporaryPrintouts.add(location);
         }
         if (this.session.view(selectors.session.status.loaded)) {
-          this.printer.printInstruction(temporaryPrintouts);
-          this.printer.printFile();
-          this.printer.printState();
+          if (this.session.view(trace.steps).length > 0) {
+            this.printer.printInstruction(temporaryPrintouts);
+            this.printer.printFile();
+            this.printer.printState();
+          } else {
+            //if there are no trace steps, let's just print a warning message
+            this.printer.print("No trace steps to inspect.");
+          }
         }
         //finally, print watch expressions
         await this.printer.printWatchExpressionsResults(
@@ -556,6 +561,7 @@ class DebugInterpreter {
       case "r":
         if (this.session.view(selectors.session.status.loaded)) {
           this.printer.printAddressesAffected();
+          this.printer.warnIfNoSteps();
           this.printer.printFile();
           this.printer.printState();
         }
@@ -563,6 +569,7 @@ class DebugInterpreter {
       case "t":
         if (!loadFailed) {
           this.printer.printAddressesAffected();
+          this.printer.warnIfNoSteps();
           this.printer.printFile();
           this.printer.printState();
         } else if (this.session.view(selectors.session.status.isError)) {
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -91,9 +91,9 @@ class DebugPrinter {
   warnIfNoSteps() {
     if (this.session.view(trace.steps).length === 0) {
       this.config.logger.log(
-        colors.bold(
-          "Warning: this transaction has no trace steps. This may happen if you are attempting to debug a transaction sent to an externally-owned accounts, or if the node you are connecting to failed to produce a trace for some reason. Please check your configuration and try again."
-        )
+        `${colors.bold(
+          "Warning:"
+        )} this transaction has no trace steps. This may happen if you are attempting to debug a transaction sent to an externally-owned accounts, or if the node you are connecting to failed to produce a trace for some reason. Please check your configuration and try again.`
       );
     }
   }
@@ -288,7 +288,9 @@ class DebugPrinter {
                 ).toString();
                 this.config.logger.log(`Revert message: ${revertString}`);
                 this.config.logger.log(
-                  colors.bold("Warning: This message contained invalid UTF-8.")
+                  `${colors.bold(
+                    "Warning:"
+                  )} This message contained invalid UTF-8.`
                 );
                 break;
             }
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -63,6 +63,7 @@ class DebugPrinter {
 
   printSessionLoaded() {
     this.printAddressesAffected();
+    this.warnIfNoSteps();
     this.printHelp();
     debug("Help printed");
     this.printFile();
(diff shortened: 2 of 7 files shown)


packages/core/lib/debug/interpreter.js
```diff
@@ -503,9 +503,14 @@ class DebugInterpreter {
...
-          this.printer.printInstruction(temporaryPrintouts);
-          this.printer.printFile();
-          this.printer.printState();
+          if (this.session.view(trace.steps).length > 0) {
+            this.printer.printInstruction(temporaryPrintouts);
+            this.printer.printFile();
+            this.printer.printState();
+          } else {
+            //if there are no trace steps, let's just print a warning message
+            this.printer.print("No trace steps to inspect.");
```
gnidan, line 512: Shouldn't this call that method you made?

```suggestion
            this.printer.warnIfNoSteps();
```

packages/core/lib/debug/printer.js
```diff
@@ -86,6 +87,12 @@ class DebugPrinter {
     );
   }
 
+  warnIfNoSteps() {
+    if (this.session.view(trace.steps).length === 0) {
+      this.config.logger.log("Warning: This transaction has no trace steps.");
```
gnidan, line 92: Can we start standardizing debugger messaging? Like making `Warning:` bold and punctuating with blank lines above/below? Does it makes sense to make this look like `--stacktrace-extra` warning?

If possible, it might also be a good opportunity to add some helpful info here... like, what could cause the lack of trace steps, and is there anything the user can do to fix that?

packages/core/lib/debug/interpreter.js
haltman-at: No, it's deliberately a different message.  The `warnIfNoSteps` here wouldn't make sense here.  A warning is printed in *addition* to something, whereas this is in *place* of the usual printout.

packages/core/lib/debug/printer.js
haltman-at: > Can we start standardizing debugger messaging? Like making Warning: bold and punctuating with blank lines above/below? Does it makes sense to make this look like --stacktrace-extra warning?

Hm, we don't do that with our other warning (some contracts couldn't be identified).  Should we?  Bold sounds like a good idea, I dunno about additional blank lines (I think it's OK for both warnings to appear together?).

> If possible, it might also be a good opportunity to add some helpful info here... like, what could cause the lack of trace steps, and is there anything the user can do to fix that?

I mean, the two reasons I can think of are:
1. There are legitimately no trace steps (you decided to inspect a transfer between two externally-owned accounts for some perverse reason)
2. Your ethereum client failed to return trace steps for some reason!

I'm not really sure what more there is to say about either of these?

packages/core/lib/debug/interpreter.js
gnidan: Ah, so then it should probably be more in-your-face. Is this one an error?

packages/core/lib/debug/printer.js
gnidan: Hm, well, we could say something like that. Here's a rough take:

```
Warning: this transaction has no trace steps. This may happen if you are attempting to debug a transaction between two externally-owned accounts, or if the node you are connecting to failed to produce a trace for some reason. Please check your configuration and try again.
```

packages/core/lib/debug/interpreter.js
haltman-at: I don't think so?  There's no error here, there just aren't any trace steps to inspect.  Remember the user has to deliberately hit `p` to see this.  They've already been warned there are no trace steps.  If they try to hit `p` anyway, well, they get a message telling them there's nothing to see.  It can be marked "error" if you think that's appropriate, I guess?  But to my mind it's not much different than the "no transaction loaded" messages users will get if they try to do things with no transaction loaded.  (Or should those be marked error too?)

packages/core/lib/debug/printer.js
haltman-at: OK.  Should warnings be bolded too?

gnidan: Yep! The goal is to get users to read them :)

packages/core/lib/debug/interpreter.js
gnidan: Ah I see. OK then this is probably be fine

packages/core/lib/debug/printer.js
```diff
@@ -86,6 +88,16 @@ class DebugPrinter {
     );
   }
 
+  warnIfNoSteps() {
+    if (this.session.view(trace.steps).length === 0) {
+      this.config.logger.log(
+        colors.bold(
+          "Warning: this transaction has no trace steps. This may happen if you are attempting to debug a transaction sent to an externally-owned accounts, or if the node you are connecting to failed to produce a trace for some reason. Please check your configuration and try again."
```
gnidan, line 95:
```suggestion
          "Warning: this transaction has no trace steps. This may happen if you are attempting to debug a transaction sent between externally-owned accounts, or if the node you are connecting to failed to produce a trace for some reason. Please check your configuration and try again."
```

haltman-at: Yeah, I changed that to `to an` because that's what's really relevant, is who the transaction is sent to.  Transactions are *always* sent from an externally-owned account; it seems a little silly to specify that part.

packages/core/lib/debug/printer.js
```diff
@@ -86,6 +88,16 @@ class DebugPrinter {
     );
   }
 
+  warnIfNoSteps() {
+    if (this.session.view(trace.steps).length === 0) {
+      this.config.logger.log(
+        `${colors.bold(
+          "Warning:"
+        )} this transaction has no trace steps. This may happen if you are attempting to debug a transaction sent to an externally-owned accounts, or if the node you are connecting to failed to produce a trace for some reason. Please check your configuration and try again.`
```
gnidan, line 96: Grammar error or typo here:

```suggestion
        )} this transaction has no trace steps. This may happen if you are attempting to debug a transaction sent between externally-owned accounts, or if the node you are connecting to failed to produce a trace for some reason. Please check your configuration and try again.`
```

haltman-at:
Oh ha. Plural -> singular.  I see it now.  Will fix.

PR #3081, opened 2020-05-29
Add optional `url` option to the config to use instead of host/port

seesemichaelj:
This PR adds an optional `url` field to the truffle config to allow overriding host/port. This allows endpoints that have a path (i.e. https://mainnet.infura.io/v3/<key> and https://sandbox.truffleteams.com/<id>) as well as anything that uses TLS (aka `https` or `wss`)

Ideally, the motivation here is to reduce the number of steps for users to use a Truffle Teams Sandbox

### DIFF
--- a/packages/provider/index.js
+++ b/packages/provider/index.js
@@ -20,13 +20,13 @@ module.exports = {
       provider = options.provider();
     } else if (options.provider) {
       provider = options.provider;
-    } else if (options.websockets) {
+    } else if (options.websockets || /^wss?:\/\//.test(options.url)) {
       provider = new Web3.providers.WebsocketProvider(
-        "ws://" + options.host + ":" + options.port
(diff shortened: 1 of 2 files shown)


gnidan:
@seesemichaelj [here's the tests I found](https://github.com/trufflesuite/truffle/blob/develop/packages/provider/test/create.js)

If these tests aren't the right tests, then I'll retract my request changes

seesemichaelj:
Your test architecture is bonkers. I never find the correct tests. I 100% expect these test to exist in `packages/config`, so I never find existing tests for the things I'm adding to

seesemichaelj:
> All the config options do get passed through to options here, I gather?

Ya, there's a bit of magic of options passing that happens with truffle to get the config variable here

PR #3086, opened 2020-06-02
Bug fix: Prevent debugger from landing on nodes with no node type

haltman-at:
It's the revenge of the pseudo-nodes!

Recently, for assembly handling, we had to allow the debugger to land on nodes with no ID, which previously was disallowed.  (By "land on" I'm not referring to where `n` stops, but rather to what goes into the interval tree to be counted as a node that the debugger has hit.)  But in doing so I allowed the debugger to land on nodes with no node type, which causes problems.  So, I have now explicitly disallowed that; those no longer go in the interval tree.

(What nodes, or pseudo-nodes, are these?  They're the references in assembly to outside variables, the same ones that caused us trouble initially and caused us to exclude nodes without IDs in the first place.)

Also, while I was at it, I switched our use of `IntervalTree` back to the usual way, rather than the "advanced usage" that we didn't really need.  The reason the usual way stopped working when I moved it to `solidity-utils` is because it's a default export, and I didn't realize that `require`, unlike `import`, won't get the default export by default.

### DIFF
--- a/packages/solidity-utils/index.js
+++ b/packages/solidity-utils/index.js
@@ -2,9 +2,7 @@ const debug = require("debug")("solidity-utils");
 const CodeUtils = require("@truffle/code-utils");
 const Codec = require("@truffle/codec");
 const jsonpointer = require("json-pointer");
-//NOTE: for some reason using the default export isn't working??
-//so we're going to do this the "advanced usage" way...
-const { IntervalTree } = require("node-interval-tree");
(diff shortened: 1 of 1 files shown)


haltman-at:
Btw, this also means we can go back from counting slashes to just using string lengths (for judging which pointer is the longest), but, eh, I'll just leave that alone.  Seems more robust against weird future things, like.

PR #3088, opened 2020-06-02
Oops, I misunderstood the `docker` option

haltman-at:
I thought you put a solc version there.  In fact it's a boolean that says whether to use docker.  Removing some code where I attempted to extract the solc version from the `docker` option.


haltman-at:
Also, while I'm at it, I turned on the `loose` option to the `semver` functions since we're dealing with user input here.

PR #3090, opened 2020-06-04
Dependency update: Turn off some highlighting that shouldn't be there

haltman-at:
It's a highlightjs-solidity update.  Turned off highlighting for some Yul features that are not actually available in Solidity assembly.  Oops.


haltman-at:
Oops, meant to add reviewers, not assignments...

Issue #2066, opened 2019-05-31
Error debugging transaction using Quorum client


marcosio:
- [X] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

When we try to debug a transaction on a JPMorgan Chase Quorum node with version 2.2.3 configured in an IBFT network, the debugger fails with the error `Cannot read property 'toTwos' of undefined`.

## Steps to Reproduce

I create a private network based on this [genesis file](https://raw.githubusercontent.com/alastria/alastria-node/73d89e6e438e1ad6831f6518e9e7de95dbc63e57/data/genesis.json) with the last version of (Quorum)[https://bintray.com/quorumengineering/quorum/download_file?file_path=v2.2.3%2Fgeth_v2.2.3_linux_amd64.tar.gz] using IBFT consensus and 2 nodes (one validator and one regular).

For testing the environment, I use [I dare you](https://gitlab.com/iobuilders/idareyou-contracts) repository, using the local network and from the console:

```
$ truffle console --network local
truffle(local)> IdareYouErc20.new().then((t) => tk=t)
truffle(local)> IDareYou.new(tk.address).then((i) => idy=i)
truffle(local)> judge = '0x....' // a valid address in the network.
truffle(local)> web3.eth.getBlock('latest').then((b) => tmp=web3.utils.hexToNumber(b.timestamp))
truffle(local)> id = web3.utils.utf8ToHex('test')
truffle(local)> idy.groupBet(id, 'test', judge, 1, tmp + 1000, tmp + 9000)
Error: Transaction: 0x299d3b355b1991cc0847c44eb6d6cebb84987922e3c432381b9c5b6814283838 exited with an error (status 0). 
     Please check that the transaction:
     - satisfies all conditions set by Solidity `require` statements.
     - does not trigger a Solidity `revert` statement.

    at StatusError.ExtendableError (/home/marcos/.npm-packages/lib/node_modules/truffle/build/webpack:/packages/truffle-error/index.js:10:1)
    at new StatusError (/home/marcos/.npm-packages/lib/node_modules/truffle/build/webpack:/packages/truffle-contract/lib/statuserror.js:33:1)
    at Object.receipt (/home/marcos/.npm-packages/lib/node_modules/truffle/build/webpack:/packages/truffle-contract/lib/handlers.js:117:1)
    at process._tickCallback (internal/process/next_tick.js:68:7)
```

## Expected Behavior

[...]

PR #3095, opened 2020-06-09
Bug fix: Account for Geth's not reporting storage that is read but not written

haltman-at:
This PR is in response to issue #2066.  It turns out that the problem is that Geth, in its trace steps, only displays storage that has been *written to* in this transaction, not *read from* in this transaction.

This results in storage slots showing up in the codex with a value of `undefined`.  This causes problems when we want to convert that `undefined` to a `Uint8Array`.

There are two possible solutions to this.  The better solution is probably to simply not add `undefined`  to the codex.  However, @gnidan wanted us to add a user warning when this happens, so instead I went with the other solution -- just being OK with `undefined` and accounting for it.  Fortunately our existing code all checks `storage[slot] !== undefined`, not `slot in storage`, so not much code need change there.

As mentioned, I also added a user warning when this is detected.  It's only printed once each time through the transaction.

Of course, while this fixes the crash and similar, this is still going to cause a number of decoding problems when using the debugger with Geth.  I'll file an issue with them.

### DIFF
--- a/packages/core/lib/debug/interpreter.js
+++ b/packages/core/lib/debug/interpreter.js
@@ -43,6 +43,8 @@ class DebugInterpreter {
 
     this.repl = config.repl || new ReplManager(config);
     this.enabledExpressions = new Set();
+
+    this.undefinedStorageWarningPrinted = false;
   }
 
   async setOrClearBreakpoint(args, setOrClear) {
@@ -391,6 +393,7 @@ class DebugInterpreter {
       //(but not if nothing is loaded)
       if (this.session.view(selectors.session.status.loaded)) {
         await this.session.reset();
+        this.undefinedStorageWarningPrinted = false;
       } else {
         this.printer.print("No transaction loaded.");
         this.printer.print("");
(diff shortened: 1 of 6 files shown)


haltman-at:
OK I went back and added some padding to the state for Besu compatibility.

And of course the build is failing because of Quorum.

haltman-at:
Oops, screwed up the padding, forgot to run the tests locally, and... yeah.  Anyway fixed now...

haltman-at:
Ugh, you know what?  The way I did the Besu compatibility stuff is simply too hacky, especially considering that there's an obvious much better way.  I'm going to go back and do it a different way, so don't review this yet.

haltman-at:
OK, reverted the Besu compatibility commits and did it a non-hacky way -- convert the trace on intake!  Duh!

However I still need to retest this on actual Besu just to make sure I didn't screw up.  Too tired right now, will do that tomorrow.

haltman-at:
I realized there's a better way to do all this.  I'm closing this PR and will make a new one.

PR #3101, opened 2020-06-10
Enhancement: Make Migrations optional

CruzMolina:
- try/catch require's of `Migrations.sol` in
`@truffle/migrate` to ignore non-existence of
Migrations artifact

### DIFF
--- a/packages/migrate/index.js
+++ b/packages/migrate/index.js
@@ -178,10 +178,8 @@ const Migrate = {
     try {
       Migrations = options.resolver.require("Migrations");
     } catch (error) {
-      const message = `Could not find built Migrations contract: ${
-        error.message
-      }`;
-      throw new Error(message);
+      // don't throw, Migrations contract optional
+      return 0;
     }
 
     if (Migrations.isDeployed() === false) return 0;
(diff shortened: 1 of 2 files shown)


scripts/ci.sh
```diff
@@ -49,7 +49,7 @@ elif [ "$QUORUM" = true ]; then
   sudo mv docker-compose /usr/local/bin
   git clone https://github.com/jpmorganchase/quorum-examples
   cd quorum-examples
-  QUORUM_GETH_ARGS="-allow-insecure-unlock" docker-compose up -d
+  docker-compose up -d
```
seesemichaelj, line 52: how is this change related?

seesemichaelj: I'm happy with the rest of the code, so I defer to @gnidan's review. Seems like this change was added because Quorum changed somethings in `quorum-examples`, causing tests to fail (though I see that they're passing). I'd rather have this fixed in another PR and have this one rebased after the merge of that PR

gnidan: I agree with @seesemichaelj here. Better to make separate PRs out of unrelated changes

lionelyoung:
Does this mean we can track state ourselves and not use migrations?

gnidan:
Yep, totally optional!

PR #3118, opened 2020-06-23
Enhancement: Add a Sourcify source-fetcher

haltman-at:
This PR expands adds a source-fetcher for Sourcify.

Because Sourcify doesn't have multiple formats to deal with, it's mostly more straightforward to handle.  First, get the metadata.  Then get the sources.  If they're literal sources we can get them straight out of the metadata -- Sourcify doesn't actually support this yet but they're planning it -- otherwise we look them up by path.

I copied the retry logic from the Etherscan fetcher.  I removed the throttling; hopefully that doesn't cause any problems.

However, there are a few additional wrinkles worth noting.

First, for Sourcify, if Sourcify doesn't list a certain contract, there's no special error JSON we'll get back; rather, we'll just get a 404.  So, we now check errors to see whether they're 404 errors.  At the low level, when making requests, this means that if we get a 404 error we give up and don't retry.  At the high level, when getting metadata, this means that if we get a 404 we just return `null`, we don't throw an error.

Secondly, we access Sourcify via HTTPS, not via IPFS.  This is partly because IPFS is a pain, and partly because it just doesn't seem to have as many contracts listed as the web version??

Thirdly, *because* we only use the web version, I chose not to include any logic to determine whether the given network ID is supported, because, well, Sourcify will work exactly the same regardless.  If we were doing it the IPFS way, where things are done by name instead of numerical ID, I'd have to add this.

(Unless we wanted to try to rely on the symlinks, which... only `mainnet` and `goerli` even have symlinks apparently, so...)

Finally -- and this is a change to the Etherscan fetcher as well -- just to head off any potential problems with this, I added `"Yul"` as a possible option for `language`, alongside `"Solidity"` and `"Vyper"`.  Of course, like with Vyper, we won't do anything with Yul source, but that's why I added it -- I figure that just in case Yul source exists out there on Etherscan or Sourcify, we should handle it by ignoring it, rather than it causing an error, which is what would have happened before.

Anyway, on the whole, nothing complicated here.  Yay, Sourcify support!

### DIFF
--- a/packages/source-fetcher/lib/sourcify.ts
+++ b/packages/source-fetcher/lib/sourcify.ts
@@ -49,17 +49,25 @@ const SourcifyFetcher: FetcherConstructor = class SourcifyFetcher
     if (!metadata) {
       return null;
     }
-    const sources: Types.SourcesByPath = Object.assign(
-      {},
-      ...(await Promise.all(
-        Object.entries(metadata.sources).map(
-          async ([sourcePath, { content: source }]) =>
-            source !== undefined
-              ? source //sourcify doesn't support this yet but they're planning it
-              : await this.getSource(address, sourcePath)
-        )
-      ))
-    );
+    let sources: Types.SourcesByPath;
+    if (metadata.language === "Solidity") {
+      sources = Object.assign(
+        {},
+        ...(await Promise.all(
+          Object.entries(metadata.sources).map(
+            async ([sourcePath, { content: source }]) => ({
+              [sourcePath]:
+                source !== undefined
+                  ? source //sourcify doesn't support this yet but they're planning it
+                  : await this.getSource(address, sourcePath)
+            })
+          )
+        ))
+      );
+    } else {
+      //don't bother attempting to fetch sources if it's Yul
+      sources = {};
+    }
     return {
       sources,
       options: {
--- a/packages/source-fetcher/lib/sourcify.ts
+++ b/packages/source-fetcher/lib/sourcify.ts
@@ -3,7 +3,7 @@ const debug = debugModule("source-fetcher:sourcify");
 
 import { Fetcher, FetcherConstructor } from "./types";
 import * as Types from "./types";
-import { removeLibraries } from "./common";
+import { networksById, removeLibraries } from "./common";
 import request from "request-promise-native";
 
 //this looks awkward but the TS docs actually suggest this :P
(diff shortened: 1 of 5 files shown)


packages/source-fetcher/lib/sourcify.ts
```diff
@@ -0,0 +1,136 @@
...
+  constructor(networkId: number) {
+    //we'll just treat all network IDs as valid;
+    //that wouldn't work very well if we were using the IPFS interface,
+    //which goes by network name, but it'll work fine for the HTTP interface
+    //(no idea why they're different)
+    //(technically it should be chain ID but realistically it won't matter)
+    this.networkId = networkId;
+  }
+
+  async isNetworkValid(): Promise<boolean> {
```
fainashalts, line 41: A little confused what this function is doing - it just always returns true and I don't see it being used anywhere? I see above you say "we'll treat al network IDs as valid", so is this just setting up for any network id validity checks in the future?

packages/source-fetcher/lib/sourcify.ts
```diff
@@ -0,0 +1,136 @@
...
+  ): Promise<string> {
+    return await this.requestWithRetries<string>({
+      uri: `https://${this.domain}/contract/${
+        this.networkId
+      }/${address}/sources/${sourcePath}`
+    });
+  }
+
+  private async requestWithRetries<T>(
+    requestObject: any //sorry, trying to import the type properly ran into problems
```
fainashalts, line 113: If you can't import this type, maybe you can create one inside this file (`requestObjectType` or some such) instead of calling this `any`? Not a big deal but might be worth considering.

packages/source-fetcher/lib/sourcify.ts
```diff
@@ -0,0 +1,136 @@
...
+
+import { Fetcher, FetcherConstructor } from "./types";
+import * as Types from "./types";
+import { removeLibraries } from "./common";
+import request from "request-promise-native";
+
+//this looks awkward but the TS docs actually suggest this :P
+const SourcifyFetcher: FetcherConstructor = class SourcifyFetcher
+  implements Fetcher {
+  get fetcherName(): string {
```
fainashalts, line 12: I feel like I'm missing something here, but why is there a `get fetcherName()` _and_ a `static get fetcherName()`? Is one of these in here accidentally?

packages/source-fetcher/lib/sourcify.ts
haltman-at: Hm, maybe?  Not sure it's worth the trouble in this case, bunch of weird little options.

packages/source-fetcher/lib/sourcify.ts
haltman-at: No, I want it to be available both statically and per-instance.  Strictly speaking this is a little redundant, as you could always use `.constructor`, but that seems to run into Typescript problems? :-/  (OK, to be sure, I'm only using this from JS, but ideally it should be usable from TS as well.)

You might want to see the [code](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/debug/external.js) where this is actually used -- as you can see, we do indeed use it both on the class (for determining which ones to use and in what order, and for getting the config options for each) and on the instance (for error reporting and for assigning compilation IDs).

packages/source-fetcher/lib/sourcify.ts
haltman-at: Basically, yeah.  Both the future if I decide to add that check here -- and it's possible I really should add that check here now, as it could save an HTTPs request, although realistically it's not that likely to come up -- and also the past.  The idea is that all the fetchers adhere to a uniform interface, as given in [types.ts](https://github.com/trufflesuite/truffle/blob/develop/packages/source-fetcher/lib/types.ts).  If you look at the existing [Etherscan fetcher](https://github.com/trufflesuite/truffle/blob/develop/packages/source-fetcher/lib/etherscan.ts), you'll see it does actually potentially return false!  So like I said, the future but also the past. :)

packages/source-fetcher/lib/sourcify.ts
gnidan: Lately I've been making a point to do what you describe @fainashalts, but because this is a `private` method, I think it's less important here than if it were public? Your call @haltman-at

haltman-at: OK, gonna leave it alone then.

packages/source-fetcher/lib/sourcify.ts
haltman-at: OK @gnidan seems to agree I should make this change, so I'll do it. :P

haltman-at:
Alright, going ahead and merging this. :)

PR #3131, opened 2020-06-29
Support arrays of tuples in ABI schema

gnidan:
To fix problem with ABI input type `tuple[]`

Raised on Spectrum [here](https://spectrum.chat/trufflesuite/truffle-teams/artifacts-upload-is-not-working~e6758041-1135-4857-9429-96c81e8ecfca)


gnidan:
Thanks Harry!

PR #3134, opened 2020-06-29
Enhancement: Improve network check for congested networks

### LINKED ISSUE #3133: Intermittent hanging when interacting with congested networks
Before interacting with a network, `truffle` tests the state of a configured network using a `provider.testConnection` method defined [here](https://github.com/trufflesuite/truffle/blob/develop/packages/provider/index.js#L36-L73).

For the past week or so, `mainnet` along with other production test networks have been experiencing heavy congestion, causing the default timeout setting to throw.

Additionally, when upping the timeout to some high duration (`1e9` for example), `truffle` will hang and never resolve on the `getBlockByNumber` call occurring [here](https://github.com/trufflesuite/truffle/blob/develop/packages/provider/index.js#L58-L59).

```
➜  smoke-test truffle migrate --network rinkeby --skipDryRun

Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.

```

CruzMolina:
- Wraps `getBlockNumber` call within `Provider.testConnection` within a `setInterval`

When coupled with the `web3-provider-engine` [fix](https://github.com/trufflesuite/provider-engine/pull/6), enables smoother CLI production deployments. (Smoke-tested 20 consecutive successful deployments ✅).

Resolves #3133.

### DIFF
--- a/packages/provider/index.js
+++ b/packages/provider/index.js
@@ -55,20 +55,25 @@ module.exports = {
           "networks[networkName].networkCheckTimeout property to do this.";
         throw new Error(errorMessage);
       }, networkCheckTimeout);
-      interfaceAdapter
-        .getBlockNumber()
-        .then(() => {
-          clearTimeout(noResponseFromNetworkCall);
-          resolve(true);
-        })
-        .catch(error => {
-          console.log(
-            "> Something went wrong while attempting to connect " +
-              "to the network. Check your network configuration."
-          );
-          clearTimeout(noResponseFromNetworkCall);
-          reject(error);
-        });
+
+      const networkCheck = setInterval(() => {
+        interfaceAdapter
+          .getBlockNumber()
+          .then(() => {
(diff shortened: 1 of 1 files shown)


CruzMolina:
Currently, the network check assumes there'll be _some_ kind of timely response (which falls apart when interacting with congested networks).

Using `setTimeout` here will delay the initial network check call, while using `setInterval` will slow every request down by *some* amount (both affecting the speed/performance of using `truffle` with `ganache`)... 🤔

CruzMolina:
Looks like using the default interval for `setInterval` slows the network check by ~50ms.

CruzMolina:
Thanks @nicholasjpaterno!

Exponential backoff would be trivial to add, but I don't think it's necessary at this point.
`truffle` just needs _some_ response, even if it's an error. The default 5sec timeout seems like a sensible cap on making initial network check requests.

PR #3139, opened 2020-07-01
Fix: Bump maxBuffer for Docker solc compilations

### LINKED ISSUE #3140: Colony job broken
#3138 accidentally broke the `COLONY` job: https://app.circleci.com/pipelines/github/trufflesuite/truffle/184/workflows/81f7c350-d718-46f8-8450-9f941a592cc6/jobs/79

Looks like the solution is to crank up the `maxBuffer` a bit more...

CruzMolina:
Resolves #3140.

### DIFF
--- a/packages/compile-solidity/compilerSupplier/loadingStrategies/Docker.js
+++ b/packages/compile-solidity/compilerSupplier/loadingStrategies/Docker.js
@@ -8,7 +8,7 @@ const VersionRange = require("./VersionRange");
 
 // Set a sensible limit for maxBuffer
 // See https://github.com/nodejs/node/pull/23027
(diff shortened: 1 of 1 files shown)


chmanie:
I found that 20MB should suffice (but 100 is probably on the safe side?). I'm not sure what negative implications a high number might have here

CruzMolina:
@chmanie FWIW, apparently node 10 and lower's default `maxBuffer` for `execSync` was set to `Infinity` 👀: https://github.com/nodejs/node/pull/23027#issuecomment-480330854

I think 100 should be ok, but would be fine with 20 or 50.

fainashalts:
👍 LGTM!

PR #3145, opened 2020-07-06
Correct missing variable assignment

eggplantzzz:
Reported by goncalo.sa in the consensys-truffle slack channel. Please ping him there to alert him after this fix is released.

### DIFF
--- a/packages/compile-vyper/index.js
+++ b/packages/compile-vyper/index.js
@@ -92,7 +92,7 @@ function execVyper(options, source_path, callback) {
   ) {
     formats.push("source_map");
   }
-  const command = `vyper -f${formats.join(",")} ${source_path}`;
+  const command = `vyper -f ${formats.join(",")} ${source_path}`;
 
   exec(command, { maxBuffer: 600 * 1024 }, function(err, stdout, stderr) {
     if (err)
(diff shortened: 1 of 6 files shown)


packages/core/lib/package.js
```diff
@@ -224,7 +224,7 @@ const Package = {
     // Now get the URIs of each network that's been deployed to.
     let result;
     try {
-      await Networks.asURIs(options, deployed_networks);
+      result = await Networks.asURIs(options, deployed_networks);
```
CruzMolina, line 227: 👍

packages/compile-vyper/index.js
```diff
@@ -92,7 +92,7 @@ function execVyper(options, source_path, callback) {
   ) {
     formats.push("source_map");
   }
-  const command = `vyper -f${formats.join(",")} ${source_path}`;
+  const command = `vyper -f ${formats.join(",")} ${source_path}`;
```
CruzMolina, line 95: Can you explain why you made this change here?

eggplantzzz: Because the vyper tests are failing and I wanted to see if something changed that is suddenly making this fail. I'm having trouble upgrading the vyper compiler and I happened to notice this lack of a space between the option flag and the arguments. I want to see if this will make the tests pass.

eggplantzzz: No dice, but there should be a space here anyways according to the docs. https://vyper.readthedocs.io/en/stable/compiling-a-contract.html

PR #3146, opened 2020-07-06
Dependency update: Update Mocha to version 8

eggplantzzz:
This is a PR to update all packages (except the debugger) to version 8.

Some things to note:
 - The `useColors` options looks like it was deprecated and therefore removed here (see breaking changes [here](https://github.com/mochajs/mocha/releases/tag/v8.0.0))
 - Mocha 8 is compatible with Node >= 10.12.0 and not Node 13
 - The debugger package was left at version 5.x.x because it will require a bit more work to deal with deps like `mocha-webpack` (sorry @haltman-at).

This will be a breaking change as this will force `truffle test` to require the Node versions listed above and keep Truffle from supporting Node v8.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
-        node-version: [8.x, 10.x, 12.x]
+        node-version: [10.x, 12.x]
         env: [GETH=true, QUORUM=true, PACKAGES=true, INTEGRATION=true]
     runs-on: ${{ matrix.platform }}
 
--- a/packages/core/lib/test.js
+++ b/packages/core/lib/test.js
@@ -170,6 +170,14 @@ const Test = {
     // Propagate --bail option to mocha
     mochaConfig.bail = config.bail;
 
+    // If the command line overrides color usage, use that.
+    if (config.colors != null) mochaConfig.colors = config.colors;
+
+    // Default to true if configuration isn't set anywhere.
+    if (mochaConfig.colors == null) {
+      mochaConfig.colors = true;
+    }
+
     Mocha = mochaConfig.package || require("mocha");
     delete mochaConfig.package;
     const mocha = new Mocha(mochaConfig);
--- a/packages/core/lib/test.js
+++ b/packages/core/lib/test.js
@@ -170,14 +170,6 @@ const Test = {
     // Propagate --bail option to mocha
     mochaConfig.bail = config.bail;
 
-    // If the command line overrides color usage, use that.
-    if (config.colors != null) mochaConfig.useColors = config.colors;
-
-    // Default to true if configuration isn't set anywhere.
-    if (mochaConfig.useColors == null) {
-      mochaConfig.useColors = true;
-    }
-
     Mocha = mochaConfig.package || require("mocha");
     delete mochaConfig.package;
     const mocha = new Mocha(mochaConfig);
--- a/packages/core/lib/test.js
+++ b/packages/core/lib/test.js
@@ -171,11 +171,11 @@ const Test = {
     mochaConfig.bail = config.bail;
 
     // If the command line overrides color usage, use that.
-    if (config.colors != null) mochaConfig.colors = config.colors;
+    if (config.color != null) mochaConfig.color = config.color;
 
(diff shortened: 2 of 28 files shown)


.github/workflows/nodejs.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
-        node-version: [8.x, 10.x, 12.x]
```
CruzMolina, line 23: ooooo

https://mochajs.org/#installation

>As of v8.0.0, Mocha requires Node.js v10.12.0 or newer.

eggplantzzz:
It seems to me that we can also remove all the Node 8 jobs from CI. This would be all the Travis stuff I believe.

packages/core/lib/test.js
```diff
@@ -170,14 +170,6 @@ const Test = {
     // Propagate --bail option to mocha
     mochaConfig.bail = config.bail;
 
-    // If the command line overrides color usage, use that.
-    if (config.colors != null) mochaConfig.useColors = config.colors;
-
-    // Default to true if configuration isn't set anywhere.
-    if (mochaConfig.useColors == null) {
-      mochaConfig.useColors = true;
```
CruzMolina, line 178: Looks like `useColors` got removed in `mocha` `v8` https://github.com/mochajs/mocha/pull/4178,
and is now instead `color` https://mochajs.org/api/mocha.js.html#line696.

Requesting changes to keep these lines and replace `mochaConfig.useColors` for `mochaConfig.color`.

.github/workflows/nodejs.yml
gnidan: I'm leaning towards the idea that bumping the minimum node to v10 is non-breaking.

CruzMolina: For reference, Gatsby dropped Node 8 support in a minor release. https://www.gatsbyjs.org/blog/2020-03-20-dropping-support-for-node-8/

Looks like they also had deprecation warnings in a couple releases leading up to the minor bump.

packages/core/lib/test.js
eggplantzzz: Yep that sounds good.

eggplantzzz:
It seems like maybe we could consider it for 5.2.0.

packages/core/lib/test.js
```diff
@@ -171,11 +171,11 @@ const Test = {
     mochaConfig.bail = config.bail;
 
     // If the command line overrides color usage, use that.
-    if (config.colors != null) mochaConfig.useColors = config.colors;
+    if (config.color != null) mochaConfig.color = config.color;
```
CruzMolina, line 174:
```suggestion
    if (config.colors != null) mochaConfig.color = config.colors;
```

Funky, but if we don't keep this how it currently is (`config.colors`, which was a `mocha` CLI alias for `useColors`), then this will be a breaking change.

eggplantzzz: Ok so I wrote a conditional to check both cases.

.github/workflows/nodejs.yml
CruzMolina: Interestinggggg, so it looks like the `mocha` _CLI_ requires Node v10.12.0 and up, but not the API.

After QA'ing a bit, I have not seen this to be an actual breaking change to end users.

packages/core/package.json
```diff
@@ -54,7 +54,7 @@
     "ganache-core": "2.10.2",
     "glob": "^7.1.4",
     "hdkey": "^1.1.0",
-    "mocha": "5.1.1",
+    "mocha": "8.0.1",
```
haltman-at, line 57: This `mocha` (the one in `core`, the dependency, not the dev dependency) is exposed to the user... so isn't upgrading this one a breacking change?  Should this one be left as is?

CruzMolina: There could be some edge cases when doing customizable `mocha`-things (`mocha-parallel tests`?), but so far the core CLI functionality when running `truffle test` seems about the same.

I haven't checked the Library API we expose (`truffle.test`...), but last I checked it's undocumented and still considered experimental. 🤷‍♂️ 

Maybe we should dig in and outline what exactly the breaking changes are between v5.2.0 and v8.
FWIW, I don't think `truffle` has ever fully supported the complete features of `mocha`.

Issue #3034, opened 2020-05-06
Delete one library for creating temporary directories


eggplantzzz:
Currently both `core` and `box` use npm libraries to create temporary directories (temp and tmp).

I think we should pick the more relevant one and exclusively use that one and update it.

PR #3151, opened 2020-07-08
Dependency update: Use tmp in place of temp

eggplantzzz:
PR for getting rid of `temp` in favor of `tmp`.

Also upgrades tmp in the following packages:
* @truffle/box: 0.0.33 →  0.2.1
* @truffle/compile-solidity: 0.0.33 →  0.2.1
* truffle: 0.0.33 →  0.2.1

closes https://github.com/trufflesuite/truffle/issues/3034


haltman-at:
A comment, I notice in some places you set `tmp.setGracefulCleanup();` just before `tmp` is used; I think it might make more sense to do this at the top immediately after it's imported?  Kind of like we currently do with `temp`, except that unlike with `temp` the method doesn't return the object annoyingly.

eggplantzzz:
Yeah I suppose it makes sense since it is basically boilerplate where it is used like that @haltman-at. I went ahead and made the changes!

PR #3152, opened 2020-07-08
Update highlightjs-solidity to highlight gwei

haltman-at:
Solidity 0.6.11 adds the `gwei` unit.  I updated highlightjs-solidity to include it.  This PR updates our highlightjs-solidity dependency so we can make use of this.


haltman-at:
Note: Merged despite yellow because the yellow was in error (if you clicked through it was actually green).

PR #3161, opened 2020-07-10
Internal improvement: Eliminate call to inherit from utils module

eggplantzzz:
[The Node docs recommend not using the `inherit` method.](https://nodejs.org/api/util.html#util_util_inherits_constructor_superconstructor)
:)


eggplantzzz:
Unfortunately I realized that there are more uses of this kind of syntax that will need to be dealt with. Closing this for now until I do a bit more research.

PR #3166, opened 2020-07-14
Add `artifacts.config` for use in JS tests

gnidan:
- Expand `artifacts` global object with new property `artifacts.config`
- Sanitize the config object using `normalize`, to avoid mutation and to eliminate weird JS-object-model-stuff

### DIFF
--- a/packages/core/lib/test.js
+++ b/packages/core/lib/test.js
@@ -300,7 +300,10 @@ const Test = {
           contract.debugger = bugger;
         }
         return contract;
-      }
+      },
+      
+      // convert config into plain object (remove special properties)
+      config: config.normalize()
     };
 
     global[config.debugGlobal] = async operation => {


cds-amal:
> Sanitize the config object using normalize, to avoid mutation and to eliminate weird JS-object-model-stuff

I can understand guarding against mutation but is there an example or test that demonstrates a JS-object-model concern?

gnidan:
@cds-amal I suppose I don't have a real objection to exposing the odd object behavior... I just figured that here's a reasonable place to expose an object that's less to reason about.

gnidan:
Hm, something seems up with GitHub - it's not registering new commits to this branch, nor the force-push I just did. Closing this and re-opening, to see if that fixes it.

gnidan:
Re-opened this as #3171

PR #3169, opened 2020-07-14
Add redone transaction source function (still hacky)

haltman-at:
This is a redone version of #3159, but outside `debugger` itself (it's in `debug-utils` instead); I wasn't really comfortable with it being in `debugger`.  Otherwise this is the same except for the exact invocation.  Also at @gnidan's suggestion I added more comments about how to use it.  Also, this time I didn't include the test, sorry.

Otherwise it's the same, so I'l copy in the rest of the PR text:

----

As requested by @kevinbluer and @seesemichaelj. I'd do a non-hacky way, but, well, as discussed I am busy and this was fast.

The code here is pretty straightforward, frankly. It does a full step-through, recording all the sources used, then reports them all out at the end. The horrible long-winded name is to remind you that it's a hack and should not be used for anything other than its specific intended purpose.

Caveats, of which there are many:

* This function should **only be used in light mode**. Please make sure to start the debugger in light mode if using it.
* This function **modifies the debugger state**. If you're using it for its intended purpose, though, that shouldn't be a problem.
* After calling this function, the debugger **must be switched to full mode**, using the `startFullMode` method, before use.
* Finally, **there is no way to switch back to light mode at present**, which means that if you want to switch transactions and then use this method to get the sources for the new transaction... too bad. **You'll have to restart the debugger.** Since I gather you're doing that anyway, I think that's probably fine. But if you want a way to return to light mode so you can switch transactions and still do this, I could see about adding one. But, like I said, this is was fast.

Again, reminder that this is a nasty hack and should not be used for any purposes other than its originally intended one.

----

Oh, one more thing: If you do need to use this together with transaction-switching... bug me.  I can probably come up with a way to allow the debugger to return to light mode from full mode so this can be used again.

### DIFF
--- a/packages/debug-utils/index.js
+++ b/packages/debug-utils/index.js
@@ -781,6 +781,38 @@ var DebugUtils = {
           variable === "this" ? { [replacement]: value } : { [variable]: value }
       )
     );
+  },
+
+  /**
+   * HACK warning!  This function modifies the debugger state
+   * and should only be used in light mode, at startup, in a very specific way!
+   *
+   * let bugger = await Debugger.forTx(txHash, { lightMode: true, ... });
+   * const sources = await getTransactionSourcesBeforeStarting(bugger);
+   * await bugger.startFullMode();
+   *
+   * Don't go switching transactions after doing this, because there's no
+   * way at the moment to switch back into light mode in order to re-run
+   * this function.  You do *not* want to run this in full mode.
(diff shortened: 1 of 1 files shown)


kevinbluer:
Awesome, thanks so much @haltman-at. Will look at getting integrated and reach out if questions come up 👍

Issue #3167, opened 2020-07-14
Globals are not available during migrations


frangio:
Globals like `artifacts` are not available to modules required from a migration.

```
// migrations/02_a_migration.js
const Greeter = artifacts.require('Greeter'); // ok
require('./my-module'); // ./my-module won't have access to artifacts global
```

This issue does not happen in tests.

@gnidan suggested this may be due to migrations using `@truffle/require`, which tests don't use.

### Steps to Reproduce

See https://github.com/frangio/truffle-globals-repro

```
git clone https://github.com/frangio/truffle-globals-repro.git
cd truffle-globals-repro
npm ci
npx truffle test
```

### Expected Behavior

Globals should be global for all modules.

frangio:
I think this is due to the use of the `vm` module rather than the custom require. The `artifacts` global is injected into the context of the migration as run through `script.runInNewContext`, and it doesn't get propagated to submodules in the `require` function that is injected into that context.

gnidan:
Hm @frangio if you want to make an attempt at removing the usage of `vm`, it's possible that there's no good reason for using `vm` in the first place. If CI passes, that'd be a pretty good indication that the change is safe.

Thanks for raising this and digging into it more!

PR #3171, opened 2020-07-15
Add `artifacts.config` for use in JS tests

gnidan:
(This was #3166; I re-opened it to try to get around some GH wonkiness)

- Expand `artifacts` global object with new property `artifacts.config`
- Sanitize the config object using `normalize`, to avoid mutation

### DIFF
--- a/packages/core/lib/test.js
+++ b/packages/core/lib/test.js
@@ -300,7 +300,10 @@ const Test = {
           contract.debugger = bugger;
         }
         return contract;
-      }
+      },
+      
+      // convert config into plain object (remove special properties)
+      config: config.normalize(config)
     };
 
     global[config.debugGlobal] = async operation => {


gnidan:
(Looks like GH is still misbehaving... hopefully it'll fix itself soon)

frangio:
Is it possible to get this same config object in a migration?

gnidan:
Hm @frangio... so looking at this, it's a bit less straightforward, because migrations use that weird `ResolverIntercept` component. I think that's probably worth saving for a separate PR.

gnidan:
Hey @cds-amal, I wanna get this into this week's release if possible. Are you OK with this approach, or will it conflict with the work you're doing on config?

gnidan:
And more generally, what do y'all think of this? So inside test files, you'd be able to write `artifacts.config` to get the truffle-config.

cds-amal:
It’s too early to know if there will be a conflict but it shouldn’t block this merge.

gnidan:
Thanks @cds-amal!

gnidan:
(@CruzMolina convinced me that it shouldn't be `artifacts.config`... will change this and account for #3167)

gnidan:
Closing in favor of #3178

PR #3177, opened 2020-07-17
Add `_truffle_workflow_compile` to allow Teams to save compiler output

seesemichaelj:
Since Teams uses a bash interface to run commands in the build/deployment containers, we need a CLI interface to anything we execute with Truffle. To support Teams debugger usage, we'd like to save the compiler output from `@truffle/workflow-compile::compile()`. Rather than managing our own internal script in the Teams codebase, I believe it makes sense for this to live in the Truffle codebase

This PR adds that script as a `bin` in `packages/workflow-compile`

### DIFF
--- a/packages/workflow-compile/bin/_truffle_workflow_compile
+++ b/packages/workflow-compile/bin/_truffle_workflow_compile
@@ -1,0 +1,49 @@
+#!/usr/bin/env node
+
+// This file is a simple CLI for running workflow-compile
+// while saving the artifact files **and** the compilation results
+// Usage: _truffle_workflow_compile <output-file-for-compilation-results>
+
+const path = require("path");
+const fs = require("fs");
+const Contracts = require("../new");
+const TruffleConfig = require("@truffle/config");
+
+const argIndexForFile =
+  process.argv.findIndex(arg => path.basename(arg) === path.basename(__filename));
+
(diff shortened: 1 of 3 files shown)


packages/workflow-compile/bin/_truffle_workflow_compile
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env node
+
+// This file is a simple CLI for running workflow-compile
+// while saving the artifact files **and** the compilation results
+// Usage: _truffle_workflow_compile <output-file-for-compilation-results>
+
+const path = require("path");
+const fs = require("fs");
+const Contracts = require("../new");
+const TruffleConfig = require("@truffle/config");
```
gnidan, line 10: Conventionally, we import this as `Config` instead of `TruffleConfig`, although that might not be true everywhere

CruzMolina: FYI @gnidan  https://github.com/trufflesuite/truffle/tree/develop/packages/config#usage
https://github.com/trufflesuite/truffle/blob/develop/packages/config/src/index.ts#L16
https://github.com/trufflesuite/truffle/blob/develop/packages/config/src/configDefaults.ts#L4

PR #3178, opened 2020-07-17
Ensure `config` is an available global in test and migrate

gnidan:
Spun out from #3171 and related to #3167.

### DIFF
--- a/packages/migrate/index.js
+++ b/packages/migrate/index.js
@@ -145,6 +145,7 @@ const Migrate = {
 
     try {
       global.artifacts = clone.resolver;
+      global.config = clone;
       for (const migration of migrations) {
         await migration.run(clone);
       }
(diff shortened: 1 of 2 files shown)


gnidan:
Alright and @frangio says it works, so here's a merge!

packages/migrate/index.js
```diff
@@ -145,6 +145,7 @@ const Migrate = {
 
     try {
       global.artifacts = clone.resolver;
+      global.config = clone;
```
CruzMolina, line 148: Should we `normalize` the config here?

gnidan: It's already effectively done when cloning

gnidan: I know, this stuff should be more uniform... but it didn't seem worth it to make it uniform at this time

Issue #3180, opened 2020-07-18
External compiler flags invalid bytecode for abstract contracts


shanecarey17:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

External compiler complains about abstract contracts with no bytecode.

## Steps to Reproduce

I have three config files, one top level and two which refer to projects in subdirectories p1 and p2. I omitted some parts of the config I don't believe are important.

truffle.js

```
compilers: {
        external: {
          command: 'bash truffle-compile.sh',
          targets: [{
              path: './artifacts/**/*.json'
          }]
        }
   }
```

truffle.p1.js

```
contracts_build_directory: 'artifacts/p1/',
compilers: {
    solc: {
        version: '0.4.22'
    }
  }
```

truffle.p2.js

```
contracts_build_directory: 'artifacts/p2/',
compilers: {
    solc: {
        version: '0.5.16'
    }
  }
```

truffle-compile.sh

```
npx truffle compile --config truffle.p1.js
npx truffle compile --config truffle.p2.js
```

## Expected Behavior

The subproject compilation tasks should run with their respective compiler versions and put output into `artifacts/`. Then the targets path should move all of those outputs to the top level build directory (separate from subproject build directories).

## Actual Results

There are interface classes in the subprojects which have a bytecode of `0x` in the outputted JSON. The contract schema flags these as invalid and does not move the contracts to the build folder. This validation either does not occur or does not fail with a direct call to `npx truffle compile --config truffle.p*.js`.

```
Error: Schema validation failed. Errors:

should match pattern "^0x0$|^0x([a-fA-F0-9]{2}|__.{38})+$" (pattern):
{
  dataPath: '.bytecode',
  schemaPath: '#/definitions/Bytecode/pattern',
  params: { pattern: '^0x0$|^0x([a-fA-F0-9]{2}|__.{38})+$' },
  **data: '0x',**
  parentSchema: { type: 'string', pattern: '^0x0$|^0x([a-fA-F0-9]{2}|__.{38})+$' }
}

    at Object.validate (/node_modules/truffle/build/webpack:/packages/contract-schema/index.js:265:1)
    at processTargets (/node_modules/truffle/build/webpack:/packages/external-compile/index.js:154:1)
    at processTarget (/node_modules/truffle/build/webpack:/packages/external-compile/index.js:210:1)
[...]

PR #3184, opened 2020-07-19
Allow `0x` as valid bytecode in contract schema

gnidan:
(Since this value is used for abstract contracts / interfaces.)

Raised in #3180


gnidan:
Not sure if that's true... since `0x0` is half a byte. Maybe `0x00` would be more correct, but I think the value reported by the JSON RPC would be `0x`

Anyway, thanks!

PR #3185, opened 2020-07-20
Move run plugin infrastructure to own package

gnidan:
- Define new @truffle/plugins package
  - Expose class Plugin with static entrypoint `list` method
  - Include helper methods a particular plugin, to expose behaviors:
    - To list all commands defined by a plugin
    - To check if a command is defined by a plugin
    - To load a particular command module from a plugin

- Remove most plugin logic from `lib/commands/run`
  - Leave plugin configuration check here to play it safe in case of
    implicit functional requirements
  - Replace with invocation of `@truffle/plugins` package

_(This is taken from the `preserves` branch, as part of an effort to get those changes into `develop`. This PR leaves out non-`truffle run` plugin behavior, to be introduced in a subsequent PR)_


gnidan:
Heh, CI says it wants tests for the new package. Who am I to argue? Will add those.

packages/core/lib/commands/run/run.js
```diff
@@ -1,61 +1,33 @@
...
-    pluginPath = path.dirname(pluginPath);
-    pluginPath = path.resolve(pluginPath, relativeCommandPath);
-
-    const commandModule = originalRequire(pluginPath);
-    return commandModule;
+    // customCommand not found or failed to load
+    throw new TruffleError(
+      `\nError: "${customCommand}" command not supported by any currently configured plugins. Please make sure:
+  – plugins are correctly configured in truffle-config.js
+  – the plugin supporting the command you want to use is installed\n`
```
eggplantzzz, line 22: Should we use `OS.EOL`?

packages/core/lib/commands/run/run.js
```diff
@@ -1,61 +1,33 @@
...
-        if (pluginConfigs[plugin].commands[customCommand])
-          pluginCommandObj[plugin] =
-            pluginConfigs[plugin].commands[customCommand];
-      } catch (_) {
-        // not added to pluginCommandObj
-        // empty pluginCommandObj handled in checkPluginObject
+  initializeCommand(customCommand, plugins) {
+    // look for plugin defining customCommand
+    for (const plugin of plugins) {
+      if (plugin.definesCommand(customCommand)) {
```
eggplantzzz, line 9: This reads better here. I like it!

packages/core/lib/commands/run/run.js
```diff
@@ -1,61 +1,33 @@
...
+      `\nError: "${customCommand}" command not supported by any currently configured plugins. Please make sure:
+  – plugins are correctly configured in truffle-config.js
+  – the plugin supporting the command you want to use is installed\n`
+    );
   },
 
   // executes command or throws user helpful error
-  run(pluginConfigs, customCommand, config, done) {
-    const runCommand = this.initializeCommand(pluginConfigs, customCommand);
+  run(customCommand, config, done) {
```
eggplantzzz, line 27: Could we drop `done` in this method? Or do we need to keep it to support plugins written callback-style?

packages/core/lib/commands/run/run.js
```diff
@@ -1,61 +1,33 @@
-const path = require("path");
-const originalRequire = require("original-require");
 const TruffleError = require("@truffle/error");
```
eggplantzzz, line 1: Would this file be better renamed "execute" or something? A small thing but `run/run` feels a bit a bit bad to me. What do you think?

packages/plugins/lib/plugin.ts
```diff
@@ -0,0 +1,168 @@
+const TruffleError: any = require("@truffle/error");
```
eggplantzzz, line 1: How would you feel about naming this `Plugin.ts`?

packages/plugins/lib/plugin.ts
```diff
@@ -0,0 +1,168 @@
...
+  module: string; // specified in truffle-config.js; will be require()'d
+  definition: PluginDefinition; // loaded from truffle-plugin.json
+}
+
+interface PluginDefinitions {
+  [plugin: string]: PluginDefinition;
+}
+
+interface PluginDefinition {
+  commands?: {
```
eggplantzzz, line 165: Will there be any use for a plugin without a commands property?

packages/plugins/lib/plugin.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  const normalized = plugins.map((plugin: RawPluginConfig) =>
+    typeof plugin === "string" ? { module: plugin } : plugin
+  );
+
+  for (const plugin of normalized) {
+    // fatal error if we can't load a plugin listed in truffle-config.js
+    if (!resolves(plugin.module)) {
+      throw new TruffleError(
+        `\nError: ${module} listed as a plugin, but not found in global or local node modules!\n`
```
eggplantzzz, line 44: Same about `\n` and `OS.EOL`, should we care about this? Does it matter? I forgot what the outcome of our discussion about this was.

gnidan:
Closing this in favor of #3628

Issue #2021, opened 2019-05-15
Support multiple simultaneous solc versions


amkCha:
---
tags: compiler, solidity-versioning
---

- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

In contracts folder, I have old contracts in 0.4 version of solidity and new contracts in 0.5 version. I would like to compile them in the same project but in truffle-config.js, I can only specify one version of solc. 

## Steps to Reproduce

In contracts folder, create a contract A in 0.4 and another one, contract B, in 0.5

## Expected Behavior

I would like truffle compile to work

## Actual Results

If I compile with solc 0.4.24, contract B in 0.5 fails to compile
If I compile with solc 0.5.0, contract A in 0.4 fails to compile

## Environment

* Operating System: ubuntu 18.10
* Ethereum client: ganache
* Truffle version (`truffle version`): 5.0.16
* node version (`node --version`): 11.9.0
* npm version (`npm --version`):  6.5.0

CruzMolina:
Hey @amkCha, you _might_ be able to tinker with your older pragma statements like so:
```
pragma solidity >0.4.25;
```

(assuming you want compile with 0.5.0 and above).

amkCha:
Hi @CruzMolina, thank you very much for you help !

I had to change two more things for the 0.4 contracts to work :

fainashalts:
Hi @amkCha, thanks for bringing this up! We're aware of this limitation and will take steps to resolve this in the future.

se3000:
Any progress here?

eggplantzzz:
@se3000 This problem is actually not trivial to solve so I don't know how soon it will have a solution. One solution which is a bit hacky is to build out your own "external compilation" step. You can specify your own separate compilation step manually and include it in the truffle config. You could group the contracts that require a different compiler together and compile them separately. Here are the docs for Truffle's external compilation feature https://www.trufflesuite.com/docs/truffle/reference/configuration#external-compilers

naddison36:
I've got the same problem. See issue https://github.com/trufflesuite/truffle/issues/2354 for how I'm dealing with it.

elenadimitrova:
We need to maintain contracts between 0.5 and 0.6 versions and facing the same problem. I tried using the `--config` option with no success:
```
$ truffle compile lib --config ./truffle-config-lib.js
[...]

PR #3208, opened 2020-07-28
Document `truffle compile --compiler` flag and add `none` as an option

gnidan:
This option was previously excluded from `truffle help compile`, so this PR remedies that and also adds support for the special case `truffle compile --compiler=none`, which skips compilation.

This approach surfaced from issue #2021

### DIFF
--- a/packages/core/lib/commands/compile.js
+++ b/packages/core/lib/commands/compile.js
@@ -6,19 +6,19 @@ const command = {
   builder: {
     all: {
       type: "boolean",
-      default: false
+      default: false,
     },
     compiler: {
       type: "string",
-      default: null
+      default: null,
     },
     list: {
-      type: "string"
+      type: "string",
     },
     help: {
       type: "boolean",
-      default: "false"
-    }
+      default: "false",
+    },
   },
   help: {
     usage:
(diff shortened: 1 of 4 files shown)


gnidan:
Hm, tests failing? Must've been a breaking Solidity release today :)

fainashalts:
Note: we need to change the docs re: this flag!

fainashalts:
Suggestion from @eggplantzzz -- add a separate flag that would be coming from `migrate` and then does this under the hood for compile. Something like `migrate --compile-none`. 

Note: `--compile-none` would defer to the option in this PR for `compile`

packages/core/lib/commands/compile.js
```diff
@@ -6,19 +6,19 @@ const command = {
   builder: {
```
eggplantzzz, line 6: So we're currently not sure what this field does? I think it may help yargs to parse options? I guess we should see what happens when we remove them :)

eggplantzzz: Anyways that is unrelated to this PR.

gnidan: Yeah, of course unrelated, but @eggplantzzz I spent a fair bit of time trying to find a place where it's actually used but I couldn't find anything. Maybe you can take a second look?

eggplantzzz: So it does provide defaults to the config object. So if you print the config after it is created you get defaults for all the options  listed in `builder` as it sort of gets merged in somehow. So check out the end of the config when I print it after executing `truffle compile --all`:
```
 timeoutBlocks: [Getter/Setter],
  _: [ [length]: 0 ],
  all: true,
  help: false,
  compiler: null,
  list: undefined }
```

If you remove `builder` then you don't get the defaults for `help`, `compiler`, and `list` like so:
```
 timeoutBlocks: [Getter/Setter],
  _: [ [length]: 0 ],
  all: true }
```

Maybe we should remove this stuff unless it ends up causing some problem in CI. Although maybe we should keep it because it allows us to have defaults per command.

packages/core/lib/commands/migrate.js
```diff
@@ -52,55 +57,59 @@ const command = {
...
-        description: "Do not run a test or 'dry run' migration.",
+        description: "Do not run a test or 'dry run' migration."
       },
       {
         option: "--describe-json",
         description:
-          "Adds extra verbosity to the status of an ongoing migration",
-      },
-    ],
+          "Adds extra verbosity to the status of an ongoing migration"
```
gnidan, line 110: and awaaaaay they go!

Issue #3067, opened 2020-05-21
Support mnemonic passphrase


kigawas:
Supersedes: 

https://github.com/trufflesuite/truffle-hdwallet-provider/issues/15
https://github.com/trufflesuite/truffle-hdwallet-provider/pull/17

wbt:
Undocumented maybe, but it is supported.  Set the provider to:
`new HDWalletProvider(mnemonic, url, ...)`

kigawas:
> 
> 
> Undocumented maybe, but it is supported. Set the provider to:
> `new HDWalletProvider(mnemonic, url, ...)`

Can you please elaborate? There's no password or passphrase in the constructor.

https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/src/index.ts#L34
```ts
  constructor(
    mnemonic: string | string[],
    provider: string | any,
    addressIndex: number = 0,
    numAddresses: number = 10,
    shareNonce: boolean = true,
    walletHdpath: string = `m/44'/60'/0'/0/`
  ) 
```

wbt:
The mnemonic functions in the place of a password/passphrase.  It's all you need to reconstruct the private keys corresponding to a set of accounts.

eggplantzzz:
Here are the docs where you can find how to enter in a mnemonic to generate accounts. https://github.com/trufflesuite/truffle/tree/develop/packages/hdwallet-provider

eggplantzzz:
Closing this for issue maintenance because mnemonics are supported :) If you still think there is something that we should address get back to us and we'll be happy to re-open it and help you out!

kigawas:
> Closing this for issue maintenance because mnemonics are supported :) If you still think there is something that we should address get back to us and we'll be happy to re-open it and help you out!

I mean the passphrase of the mnemonic, not mnemonic itself.

![image](https://user-images.githubusercontent.com/4182346/83088067-b5611600-a0cd-11ea-9c62-9c8f9ac83dc6.png)

josh-kean:
@kigawas I think the best, short term solution would be to provide the first private key from your salted mnemonic phrase in place of your mnemonic phrase, and then add include a waller_hd path variable when interacting with the wallet_hdprovider library.

But adding in the ability to the wallet_hdprovider library seems like a reasonable thing to do for medium/long term development. I'm going to dive into that

D-Nice:
[...]

PR #3209, opened 2020-07-28
Add Password Option HDWallet constructor

richardpringle:
This is a preliminary PR to agree on the constructor's API before moving ahead with the implementation. 

This should be a continuation of the discussion in #3067.

(don't worry about all the CI failures for now)


packages/hdwallet-provider/src/index.ts
```diff
@@ -37,13 +57,15 @@ class HDWalletProvider {
   public engine: ProviderEngine;
 
   constructor(
-    mnemonic: string | string[],
+    mnemonic: string | string[] | MnemonicArgs | PrivateKeyArgs,
```
richardpringle, line 60: There might be a cleaner way to do this with [overloads](https://www.typescriptlang.org/docs/handbook/functions.html#overloads);

(I'm not a TS expert though, please leave a comment if you have an opinion)

packages/hdwallet-provider/src/index.ts
```diff
@@ -154,7 +176,7 @@ class HDWalletProvider {
         },
         signPersonalMessage(...args: any[]) {
           this.signMessage(...args);
-        }
+        },
```
richardpringle, line 179: These changes might be your prettier rules?

eggplantzzz: Yep, it is from prettier.

gnidan:
(You'll likely want to rebase this once #3210 is sorted out + merged)

packages/hdwallet-provider/src/index.ts
richardpringle: 🤦

richardpringle:
@gnidan, #3210 is merged and I rebased...

gnidan:
Ah now it looks like the build failure is related to this PR specifically (as opposed to the previously failing build on `develop`)

gnidan:
I'll check out your branch and take a look later tonight.

richardpringle:
Haha @gnidan, I'm throwing an error in the constructor, the build isn't going to pass until I actually implement the changes. If we're on the same page for the interface (I took your suggestions from #3067), then I can move forward with the implementation, adding/fixing tests etc.

packages/hdwallet-provider/src/index.ts
```diff
@@ -37,13 +58,15 @@ class HDWalletProvider {
...
   constructor(
-    mnemonic: string | string[],
+    mnemonic: string | string[] | ConstructorOptions,
     provider: string | any,
     addressIndex: number = 0,
     numAddresses: number = 10,
     shareNonce: boolean = true,
     walletHdpath: string = `m/44'/60'/0'/0/`
   ) {
+    throw new Error("Don't merge me!");
```
gnidan, line 68: heh I missed this. went straight to looking at the red x!

packages/hdwallet-provider/src/index.ts
```diff
@@ -37,13 +58,15 @@ class HDWalletProvider {
   public engine: ProviderEngine;
 
   constructor(
-    mnemonic: string | string[],
+    mnemonic: string | string[] | ConstructorOptions,
     provider: string | any,
     addressIndex: number = 0,
     numAddresses: number = 10,
     shareNonce: boolean = true,
     walletHdpath: string = `m/44'/60'/0'/0/`
```
gnidan, line 66: probably rename this to `derivationPath` while you're at it

packages/hdwallet-provider/src/index.ts
```diff
@@ -37,13 +58,15 @@ class HDWalletProvider {
   public engine: ProviderEngine;
 
   constructor(
-    mnemonic: string | string[],
+    mnemonic: string | string[] | ConstructorOptions,
```
gnidan, line 61: can we call this `mnemonicOrOptions` and only assign to a variable `mnemonic` when we're sure it's a mnemonic?

packages/hdwallet-provider/src/index.ts
```diff
@@ -37,13 +58,15 @@ class HDWalletProvider {
...
     provider: string | any,
     addressIndex: number = 0,
     numAddresses: number = 10,
```
gnidan, line 64: let's change this to be `numberOfAddresses` as part of my crusade to unabbreviate things ;)

richardpringle:
@gnidan ... I have a feeling there's going to be a bit of back and forth on this one. I still have to implement tests, but I would like some guidance on my implementation here. The combinations of the of being type-strict while having to maintain backwards compatibility makes from some ugly code.

richardpringle:
aaaand the tests are failing, I'll have to fix that before this is ready for another round of review

gnidan:
Happy to provide guidance! Anything specifically you're looking to get answered? I'll write up some general notes later tonight after my usual block of meetings ends.

packages/hdwallet-provider/src/index.ts
```diff
@@ -28,6 +28,127 @@ import { Callback, JsonRPCResponse } from "web3/providers";
...
+
+  return { addressIndex, numberOfAddresses, sharedNonce, derivationPath, url };
+};
+
+const getMnemonic = (
+  mnemonicOrPrivateKeysOrOptions: string | string[] | ConstructorOptions
+): { phrase: string | null; password: string } => {
+  const phrase =
+    typeof mnemonicOrPrivateKeysOrOptions === "string" &&
+    mnemonicOrPrivateKeysOrOptions.includes(" ")
```
eggplantzzz, line 108: I wonder if there is a better way to do this. I'm going to just throw out an idea here, what if we `split` on " " and test the length? That would make this more lenient in the case where they have a trailing or leading space on a single private key. We could `trim` and it would be acceptable. But perhaps for something like this we should require that they input their options precisely or else fail. What do you think?

packages/hdwallet-provider/src/index.ts
```diff
@@ -37,50 +158,52 @@ class HDWalletProvider {
...
       );
 
-      if (!bip39.validateMnemonic(mnemonic, wordlist)) {
+      if (!bip39.validateMnemonic(phrase, wordlist)) {
         throw new Error("Mnemonic invalid or undefined");
       }
 
       // crank the addresses out
-      for (let i = addressIndex; i < addressIndex + numAddresses; i++) {
+      for (let i = addressIndex; i < addressIndex + numberOfAddresses; i++) {
```
eggplantzzz, line 206: This makes me happy.

eggplantzzz, line 206: I like this style change.

packages/hdwallet-provider/src/index.ts
```diff
@@ -37,50 +158,52 @@ class HDWalletProvider {
...
+      derivationPath
+    } = normalizeConstructorOptions(mnemonicOrPrivateKeysOrOptions, args);
+
+    this.walletHdpath = derivationPath;
     this.wallets = {};
     this.addresses = [];
     this.engine = new ProviderEngine();
 
-    if (!HDWalletProvider.isValidProvider(provider)) {
+    if (!HDWalletProvider.isValidProvider(url)) {
```
eggplantzzz, line 179: I just realized that `isValidProvider` is a pretty awful method. We should probably update that at some point. Or else we should just rename it something like `hasValidProtocol`.

richardpringle:
@gnidan, I merged your commit to my branch. I think that was a really good way to get this feature done 🥇 . Now let's just see if CI passes haha.

packages/hdwallet-provider/src/index.ts
richardpringle: haha agreed

packages/hdwallet-provider/src/index.ts
richardpringle: Whoops, I merged @gnidan's code into my branch and lost this comment. Haha, the only reason I had this code is that I didn't want to change the initial checks. 

I agree that splitting the string into an actual word list and checking the length might be a good idea. From skimming [the original bip39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki#generating-the-mnemonic), it appears that any multiple of 3 between 12 and 24 words is valid. 

Let me know if you want me to change what's there now.

packages/hdwallet-provider/src/index.ts
richardpringle: I'm open to renaming if you want me too (while we're here cleaning things up anyway).

gnidan: Note: this just turned into `providerOrUrl`, because that's what the types suggested previously (`string | any`!)

packages/hdwallet-provider/test/provider.test.ts
```diff
@@ -65,25 +65,7 @@ describe("HD Wallet Provider", function() {
...
-    assert.equal(addresses[0], "0xc515db5834d8f110eee96c3036854dbf1d87de2b");
-    addresses.forEach(address => {
-      assert(EthUtil.isValidAddress(address), "invalid address");
-    });
-
-    web3.eth.getBlockNumber((err, number) => {
-      assert(number === 0);
-      done();
-    });
-  });
```
gnidan, line 84: yep, I'm requesting changes for the test I deleted myself 🤦

gnidan:
Hey @eggplantzzz any ideas on how we could organize these changes into modules? Seems like your area of expertise (that doesn't have to be in this PR, but the constructor interface support code is kind of unwieldy for an `index.ts`)

packages/hdwallet-provider/src/index.ts
```diff
@@ -28,6 +28,228 @@ import { Callback, JsonRPCResponse } from "web3/providers";
...
+
+/*
+ * namespace wrapper for old-style positional arguments
+ */
+export namespace LegacyConstructor {
+  type PossibleArguments = [
+    /*
+     * required
+     */
+    MnemonicPhrase | PrivateKey[],
```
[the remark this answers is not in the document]
gnidan: > so I think there's no harm

caveat: we'll need to make sure that this branch of code is hit, even with the new-style constructor, since I don't want to accidentally lose the error message for this situation.

packages/hdwallet-provider/src/index.ts
gnidan: Missed this comment in my re-review.

Not a comment about the implementation of that check, but see [my comment below](https://github.com/trufflesuite/truffle/pull/3209#discussion_r465457068) for an idea on where this check should live (spoiler: I think it should live inside `isMnemonicPhrase`)

gnidan: Comment about implementation of that check: sounds like something that we should open an issue for instead of putting into this PR

packages/hdwallet-provider/src/index.ts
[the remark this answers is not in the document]
richardpringle: @gnidan, feel free to "steal the fun" if you get a chance. I probably won't have time to look at this PR again until next week at the earliest.

eggplantzzz:
I mean I would probably just extract the types and utility functions out into two other files. It seems like most of them are named pretty explicitly and would be easy to understand without seeing the implementation. I think that would go far in cleaning `index.ts` up. I'll go ahead and put in a PR against this branch.

eggplantzzz:
Oh I see they are kind of a bit intertwined. Let me see if I can find some sensible way of organizing this.

eggplantzzz:
You can check out one possible way to organize this [here](https://github.com/trufflesuite/truffle/pull/3235).

packages/hdwallet-provider/src/index.ts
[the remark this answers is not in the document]
gnidan: I've also got something of a busy week, so we'll see! Guess it's a race, then!

eggplantzzz:
So I think there is one more thing @gnidan wanted to update on this PR and then we'll be good for the final review/merge.

gnidan:
FYI @richardpringle @eggplantzzz I rebased and force-pushed this branch

gnidan (approved):
Looks like this is good to go!

eggplantzzz:
Thanks @richardpringle!

richardpringle:
Haha, I wanted a feature... did a _tiny_ bit of work to introduce it, and triggered @gnidan and @eggplantzzz to do a whole bunch of work. 

It was a pleasure watching this whole thing come together. I commend both of your ability to involve me as much or as little as I wanted, 👏 .

Jorgemucisalumfilho (approved):
Terra

Issue #3207, opened 2020-07-28
Update internal tests to exercise Solidity v0.7.0


gnidan:
Including in the debugger, the truffle scenario tests, and wherever else we currently test multiple versions of Solidity.

haltman-at:
Also `init` and boxes need to be updated for 0.7.0, right?  But I guess that should probably be a separate issue?

PR #3210, opened 2020-07-28
Internal improvement: Update native solc test to 0.7.x (minimally)

haltman-at:
This PR updates a few tests to 0.7.x.  This is *not* a comprehensive test update; it's just a minimal update to get the build passing again.  I've only updated what I needed to.  So, it only *partly* addresses #3207.

Note: I didn't delete the 0.6.x sources, even though they're no longer used.  Should I have?

(We have so many tests still on 0.5.x, oh my)

### DIFF
--- a/packages/compile-solidity/test/test_supplier.js
+++ b/packages/compile-solidity/test/test_supplier.js
@@ -18,7 +18,7 @@ describe("CompilerSupplier", function () {
     let oldPragmaFloatSource; // ^0.4.15
     let version4PragmaSource; // ^0.4.21
     let version5PragmaSource; // ^0.5.0
-    let version7PragmaSource; // ^0.7.0
+    let versionLatestPragmaSource; // Currently: ^0.7.0
     let compileConfig;
 
     const options = {
@@ -45,8 +45,8 @@ describe("CompilerSupplier", function () {
         "utf-8"
       );
 
-      const version7Pragma = await fse.readFile(
-        path.join(__dirname, "./sources/v0.7.x/Version7Pragma.sol"),
+      const versionLatestPragma = await fse.readFile(
+        path.join(__dirname, "./sources/v0.7.x/Version7Pragma.sol"), //update when necessary
         "utf-8"
       );
 
@@ -54,7 +54,7 @@ describe("CompilerSupplier", function () {
       oldPragmaFloatSource = { "OldPragmaFloat.sol": oldPragmaFloat };
       version4PragmaSource = { "NewPragma.sol": version4Pragma };
       version5PragmaSource = { "Version5Pragma.sol": version5Pragma };
-      version7PragmaSource = { "Version7Pragma.sol": version7Pragma };
+      versionLatestPragmaSource = { "Version7Pragma.sol": versionLatestPragma }; //update when necessary
     });
 
     it("compiles w/ default solc if no compiler specified (float)", async function () {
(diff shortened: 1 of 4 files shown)


packages/compile-solidity/test/test_supplier.js
```diff
@@ -45,19 +45,19 @@ describe("CompilerSupplier", function() {
         "utf-8"
       );
 
-      const version6Pragma = await fse.readFile(
```
CruzMolina, line 48: Seems like we continue testing the Version 6 sources, no?

haltman-at: Well, we never really explicitly tested version 6.  It looks like basically all our tests remained on 5.  The only thing updated to 6 was the test of native solc, which had to be.  And that's all I've updated here.  I figure updating other tests like they should be can be left to a later PR, presumably one written by someone who knows this code better...

gnidan: I'm not entirely upset if we skip 0.6, since there's a much larger effort of revisiting the way we organize these cases. Currently it's all ad hoc and kind of a mess.

Although I could be convinced that we need to revisit our tests for 0.6! But that's definitely a separate issue than this PR.

packages/compile-solidity/test/test_supplier.js
```diff
@@ -11,23 +11,23 @@ function waitSecond() {
...
-  describe("integration", function() {
+describe("CompilerSupplier", function () {
+  describe("integration", function () {
     this.timeout(40000);
     let oldPragmaPinSource; //  0.4.15
     let oldPragmaFloatSource; // ^0.4.15
     let version4PragmaSource; // ^0.4.21
     let version5PragmaSource; // ^0.5.0
-    let version6PragmaSource; // ^0.6.0
+    let version7PragmaSource; // ^0.7.0
```
gnidan, line 21: I think we should keep `version6PragmaSource` around for completeness

haltman-at: Husky won't allow it; unused variable.  Should I disable the check there or something?

gnidan: We already run tests for version 4 and version 5, right? Those aren't a problem with unused variables.

What I'm proposing is that we keep the test there for 0.6 and merely _add_ tests for 0.7

haltman-at: But, as I explained above, it's **not** a test of 0.6.x.  It's a test of docker/native.  It will always be the latest version.  If it were a test of 0.6.x, obviously I would have kept it and added a parallel test of 0.7.x.  But it isn't!  If it were, it wouldn't be failing in the first place!

gnidan: Oh then let's just change the variable and directory name and such

haltman-at: Hm, I suppose that works.  I'll continue to keep around the 0.6.x source for now though unless you think it should be removed.

haltman-at: Also I guess I'll keep around 0.7.x separate from latest.  Could come in handy.

haltman-at: Hm the test explicitly checks that latest is 0.n.x though (for specified n).  Will have to figure out how to handle that.

gnidan: My point is purely to avoid future head-scratching when someone sees this code and goes "why is there no version6Pragma".

This is a minor point, though, so heed this concern only if you think you see a reasonable path forward here.

haltman-at: Oh, OK.  Sure, I'll just change the variable then; I'll leave the path alone.

PR #3211, opened 2020-07-28
Internal improvement: Update calls to artifactor.saveAll

eggplantzzz:
Remove unnecessary method call and the method itself as it is unused.

All calls to `artifactor.saveAll` in Truffle will now pass an array of contracts rather than an object. In doing this work it seemed like a good idea to update `ResolverIntercept` to store contracts as an array in the cache instead of an object. This means that contracts will not be overwritten in the case where there are multiple contracts of the same name. However, this is not a fix for the problem of not being able to handle multiple contracts with the same name.

Related to [this PR](https://github.com/trufflesuite/truffle/pull/3021)


gnidan:
Do we need to update [this line in @truffle/migrate](https://github.com/trufflesuite/truffle/blob/develop/packages/migrate/migration.js#L110) also?

gnidan:
Oh yeah this is gonna need a rebase once #3210 is merged

eggplantzzz:
Yep, I'll go ahead and see about making it by default pass an array.

packages/artifactor/README.md
```diff
@@ -86,19 +86,41 @@ Save contract data as a `.json` file. Returns a Promise.
 
 Save many contracts to the filesystem at once. Returns a Promise.
 
-* `contracts`: Object. Keys are the contract names and the values are `contractData` objects, as in the `save()` function above:
+* `contracts`: (Object | Array). If an object is passed, the keys are the
```
gnidan, line 89: sweet, better docs!

packages/artifactor/test/methods.js
```diff
@@ -29,13 +29,34 @@ describe("Artifactor.saveAll", () => {
...
-        "x-from-dependency": "somedep"
+        "x-from-dependency": "somedep",
       })
       .then(() => assert(false, "didn't throw!"))
       .catch(e => {
         assert(e.message.includes("doesn't exist!"));
       });
   });
+
+  it("throws if this.destination doesn't exist(array passed)", () => {
```
gnidan, line 42: nice additional test

packages/migrate/Migration.js
```diff
@@ -0,0 +1,221 @@
...
+      number: this.number,
+      isFirst: this.isFirst,
+      isLast: this.isLast,
+      dryRun: this.dryRun,
+      interactive: this.interactive,
+    };
+  }
+}
+
+module.exports = Migration;
```
gnidan, line 221: re: the new filename for this, I feel like we should come up with a project-wide consistent approach to this sort of thing. In the meantime, I encourage this ad-hoc renaming in the direction of each contributor's personal preference :)

packages/migrate/migration.js
```diff
@@ -4,10 +4,10 @@ const Require = require("@truffle/require");
 const Emittery = require("emittery");
```
gnidan, line 4: I'm guessing you probably meant to remove this file in favor of `Migration.js`?

packages/migrate/resolverintercept.js
```diff
@@ -1,37 +1,39 @@
-function ResolverIntercept(resolver) {
-  this.resolver = resolver;
-  this.cache = {};
-}
+class ResolverIntercept {
```
gnidan, line 1: same with this file, it seems like this is duplicated by `ResolverIntercept.js`

packages/migrate/Migration.js
eggplantzzz: I think there is a pretty established convention of naming files containing classes in Pascal case, no? I wouldn't be opposed to meeting about this.

gnidan: yeah I'm not saying anything one way or another about pascal case or whatnot, only that we should at some point make it all consistent

packages/migrate/ResolverIntercept.js
```diff
@@ -0,0 +1,39 @@
...
+  require(contractName) {
+    // remove file extension if present on name
+    const sanitizedContractName = contractName
+      .replace(/^\.\//, "")
+      .replace(/\.sol$/i, "");
+
+    // there may be more than one contract of the same name which will be
+    // problematic - only return the first one found in the cache for now
+    for (const contract of this.cache) {
+      if (contract.contract_name === sanitizedContractName) {
```
gnidan, line 16: do we need to handle `contract.contractName`? I know that in some places in the case, there's risk of either casing being used. Not sure if this is one of them.

packages/migrate/Migration.js
eggplantzzz: you got it dood

packages/migrate/ResolverIntercept.js
eggplantzzz: I think this will be Truffle artifact objects so I believe we don't need to handle that case.

eggplantzzz:
`ResolverIntercept` is not exposed in this package (unless you manually dig for it) so I am claiming this is a non-breaking change.

PR #3212, opened 2020-07-29
Bug fix: Detect external contracts that create other contracts

haltman-at:
This PR fixes a problem where a contract downloaded with `source-fetcher` might not be recognized if it in turn creates another contract.

The problem of course is the metadata hashes.  I already made it so that the CBOR structures at the ends of contracts are ignored for external contexts... but what about in the middle? If a contract creates another contract, or makes use of `type(C).runtimeCode` or `type(C).creationCode`, then you'll get metadata hashes somewhere in the *middle*.

I originally had a solution for this based on starting at the end and working backward, but I failed to notice that wouldn't work if our outer contract includes multiple inner contracts.  So here's a simpler, if somewhat hackier, solution.  For every contract marked as coming from an external source, note down the CBOR portion of its bytecode, extracted in the usual way.  Then, once again from every contract marked as external, replace any occurrence of *any* of these with dots, allowing for more general matches.

Strictly speaking this could generate some false positives, but I expect in practice it won't be a big deal.

### DIFF
--- a/packages/codec/lib/contexts/utils.ts
+++ b/packages/codec/lib/contexts/utils.ts
@@ -8,20 +8,19 @@ import {
   DecoderContext,
   Context,
   Contexts,
-  DebuggerContexts,
+  DebuggerContexts
 } from "./types";
 import escapeRegExp from "lodash.escaperegexp";
+const cbor = require("borc"); //importing this untyped, sorry!
 
 //I split these next two apart because the type system was giving me trouble
 export function findDecoderContext(
   contexts: DecoderContexts,
   binary: string
 ): DecoderContext | null {
-  debug("binary %s", binary);
   let context = Object.values(contexts).find(context =>
     matchContext(context, binary)
   );
-  debug("context found: %O", context);
   return context !== undefined ? context : null;
 }
 
@@ -29,11 +28,9 @@ export function findDebuggerContext(
   contexts: DebuggerContexts,
   binary: string
 ): string | null {
-  debug("binary %s", binary);
   let context = Object.values(contexts).find(context =>
     matchContext(context, binary)
   );
-  debug("context found: %O", context);
   return context !== undefined ? context.context : null;
 }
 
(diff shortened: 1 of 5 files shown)


gnidan:
Could we go a step further and remove the CBOR structures from consideration when tracking internal contexts as well? Do we need this for anything?

haltman-at:
We could?  But I'd rather not; this method is a little hacky, it could theoretically generate some false positives, so I'd rather only apply it where necessary.

haltman-at:
OK, I actually want to go back and make some more changes here though to reduce false positives.  Also, I think I am going to make that change to using `compile` after all, because why involve temporary files unnecessarily?

haltman-at:
Oops, that commit ended up with an unhelpful message.  Um, well, what it does is, I decided to make the change after all to using `compile` and not `compile.all` after all, because why deal with the filesystem when you don't have to?

haltman-at:
So, this last commit -- I just realized there's a way to pretty much eliminate false positives.  Namely: Check whether the purported CBOR is in fact CBOR and does in fact contain a metadata hash!

I decided on the `borc` library for CBOR parsing, though it's not the only option (`cbor` being the obvious other alternative).  Basically, before adding a CBOR segment to the list of CBOR segments to replace, we check that:
1. It is in fact CBOR (for a single thing, not multiple)
2. It is CBOR for an object/dictionary
3. At least one of those object's keys is one of `"ifps"`, `"bzzr1"`, or `"bzzr0"`

Annoyingly, `borc` varies in whether CBOR for an object returns a plain old object or a `Map`, so I had to handle both those cases.  But I think the `cbor` library does this too.

Notionally we could go further -- say, checking that the value at that key is indeed a bytestring of length 32 -- but that's maybe getting a little silly.  (I totally could add that, though....)

gnidan:
> (I totally could add that, though....)

Do I sense excitement here? 😀

haltman-at:
> > (I totally could add that, though....)

> Do I sense excitement here? grinning

I mean, after posting this I actually did start on going back and adding this before deciding that this was getting silly and it wasn't really worth the (admittedly quite minimal) effort.

PR #3213, opened 2020-07-29
Update prettier to avoid trailing commas

gnidan:
They're too everywhere


fainashalts:
+1 LGTM!

gnidan:
> Too opinionated for ya?

Not really. I like trailing commas. But we already suffered through the entire codebase getting rewritten over the course of several months of PRs. I don't want to do that again, esp. with this change, which will be absolutely everywhere!

gnidan:
Thanks for the reviews!

Issue #1976, opened 2019-05-01
`truffle test` hangs if no `networks` are specified and Ganache is running on port `7545`


davidmurdoch:
From a quick look, I believe https://github.com/trufflesuite/truffle/blob/cbd741b40696d6f7f6053ae007e1fdfb22483237/packages/truffle-core/lib/develop.js#L23-L26 is part of the issue.

If the child process can't start ganache then when it rejects (here: https://github.com/trufflesuite/truffle/blob/cbd741b40696d6f7f6053ae007e1fdfb22483237/packages/truffle-core/chain.js#L237) truffle never gets notified of the `truffle.ready` event (here: https://github.com/trufflesuite/truffle/blob/cbd741b40696d6f7f6053ae007e1fdfb22483237/packages/truffle-core/chain.js#L249)

Also, if the chain.js process crashes truffle doesn't seem to be listening. Might need to listen to the close event: https://nodejs.org/api/child_process.html#child_process_event_close

I don't seen any harm in using a different port for truffle's internal ganache server instance, but I don't have much experience with using truffle myself. For reliability you could (should?) try to start the ganache-core server on a random port, if it fails due to the port already being bound, try again on another random port, and maybe a few more tries before failing.

Alternatively, you could do what `truffle migrate` does and just (try to) use the already bound 7545 port.

To get a big speed boost you could experiement with running ganache-core as a provider (Ganache.provider(options) instead of `Ganache.server(options)`.

Let me know if you have ganache-related questions!

eggplantzzz:
@davidmurdoch Is this still a relevant issue?

davidmurdoch:
Yes. I just tested with truffle v5.0.18

CruzMolina:
@davidmurdoch I don't seem able to reproduce your specific error. If fire up Ganache (`Version 2.0.1 (2.0.1.631)`), using the default settings, and I use latest truffle to run `truffle test` on a fresh metacoin box (completely commented out truffle config), I see:

```
Error: The network id specified in the truffle config (4447) does not match the one returned by the network (5777).  Ensure that both the network and the provider are properly configured.
    at Object.detectAndSetNetworkId (/Users/cruzmolina/.npm/_npx/29227/lib/node_modules/truffle/build/webpack:/packages/truffle-core/lib/environment.js:100:1)
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
Truffle v5.0.18 (core: 5.0.18)
Node v11.4.0
```

[...]

PR #3218, opened 2020-07-31
Internal improvement: Change default port for Ganache during testing

eggplantzzz:
See https://github.com/trufflesuite/truffle/issues/1976


davidmurdoch:
Using something like https://www.npmjs.com/package/get-port may be a safer option here.

eggplantzzz:
@davidmurdoch That seems like a better solution!

fainashalts:
LGTM, except the conflict in `package.json`!

Issue #3205, opened 2020-07-28
Issue using `debug` in `truffle develop`


eggplantzzz:
## Issue

There is a problem trying to debug a transaction while using `truffle develop`

## Steps to Reproduce

1. `truffle unbox metacoin`
2. `truffle develop`
3. At the prompt, run `migrate`
4. Run the following
```
const instance = await MetaCoin.deployed();
const result = await instance.sendCoin(accounts[0], 10);
```
5. Get the transaction hash from `result` and use it as an argument to `debug`

## Expected Behavior

I expect the debugger to load the transaction for `sendCoin`

## Actual Results

I receive a connection error.
```
> Something went wrong while attempting to connect to the network. Check your network configuration.

Could not connect to your Ethereum client with the following parameters:
    - host       > 127.0.0.1
    - port       > 7545
    - network_id > 5777
Please check that your Ethereum client:
    - is running
    - is accepting RPC connections (i.e., "--rpc" option is used in geth)
    - is accessible over the network
    - is properly configured in your Truffle configuration file (truffle-config.js)
```

## Environment

* Operating System: mac
* Truffle v5.1.36 (core: 5.1.36)
* Solidity v0.5.16 (solc-js)
* Node v10.20.1
* Web3.js v1.2.1

haltman-at:
Hm, I wonder if this is due to what @gnidan, @fainashalts, and @cds-amal are working on.  Like, I wonder if it's trying to connect to the wrong network due to not inheriting the config properly?

gnidan:
Hm, very strange. I don't think it's related to that work @haltman-at, but it's definitely something with the debugger not getting the correct config - since `develop` runs on port 9545, that error's a clear giveaway that something's wrong.

I don't suppose this is related to that change you made to the debugger's `--network` flag default, @haltman-at?

haltman-at:
I mean, I got *rid* of the default.  It now uses whatever means Truffle normally uses to determine the default network.

fainashalts:
Interesting! I don't believe the issue I'm dealing with is related, but I'll keep my eyes peeled in case it is!

gnidan:
Hm well this used to work, since I used this exact process to test the debugger all the time. Might be worth figuring out the first version it was broken, to check the changelog.

eggplantzzz:
[...]

PR #3219, opened 2020-07-31
Bug fix: Remove field to prevent clobbering network name in develop

eggplantzzz:
This is in response to https://github.com/trufflesuite/truffle/issues/3205.

UPDATE: Adding a field to the `builder` object for a command provides it a default when no other value is provided. In this case, the default was getting set to `undefined` and then overwriting `options.network` [here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/command.js#L134). Since it doesn't seem that we have much use for this property, this PR removes it for `debug`.

It still might be useful to keep `builder` around as it allows us to set defaults per command. However, we should be cautious about when to use those defaults.

### DIFF
--- a/packages/core/lib/commands/debug.js
+++ b/packages/core/lib/commands/debug.js
@@ -5,10 +5,6 @@ const command = {
     "_": {
       type: "string"
     },
-    "network": {
-      describe: "Network to connect to",
-      type: "string"
-    },
     "fetch-external": {
       describe: "Allow debugging of external contracts",
       alias: "x",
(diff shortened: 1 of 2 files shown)


gnidan:
Does this address the root cause, or is it just a bandaid?

eggplantzzz:
As far as I can tell it is just a configuration mismatch. :shrug:

eggplantzzz:
Ok, so I did indeed find a better place to fix this issue. See my updated description above.

Issue #3206, opened 2020-07-28
Truffle can't handle trailing ; in truffle develop


eggplantzzz:
## Issue

Truffle complains if you add a trailing semicolon in `truffle develop`

## Steps to Reproduce

1. Run `truffle unbox metacoin`
2. Run `truffle develop`
3. Run `migrate`
4. Run `await MetaCoin.deployed();`

## Expected Behavior

No error

## Actual Results

```
truffle(develop)> const i = await MetaCoin.deployed();
Thrown:
evalmachine.<anonymous>:1
  await MetaCoin.deployed();
                           ^

SyntaxError: Unexpected token ;
```

## Environment

OS = mac
Truffle v5.1.36 (core: 5.1.36)
Solidity v0.5.16 (solc-js)
Node v10.20.1
Web3.js v1.2.1

gnidan:
Good find!

eggplantzzz:
This has always bothered me but it was always such a small thing haha.

fainashalts:
Note: we think this only happens with `await`.

eggplantzzz:
[the culprit](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/console.js#L197-L214)

PR #3220, opened 2020-07-31
Bug fix: Sanitize input in console to prevent errors.

eggplantzzz:
Because of the handling of expressions with "await" in it, a trailing ";" will break it. So we strip it off!

See https://github.com/trufflesuite/truffle/issues/3206


gnidan:
Since there's no automated testing for this, it might be good to list manual test cases. I'll play around with this branch to see, myself, but did you have a list of console statements you tried out while developing this?

eggplantzzz:
I did not have a specified list but just tested a couple that came to mind at the time.

PR #3225, opened 2020-08-03
Dependency Update: Bump ethereumjs-util

### LINKED ISSUE #3175: Remove `node-gyp` dependency
`sha3` v1.x uses `node-gyp` to build itself.

Current relevant dependency tree pulling in `sha3` v1.x:

image

CruzMolina:
- Bumps `ethereumjs-util` to remove `keccakjs` -> `sha3` -> `node-gyp` dependencies

Eliminates the `node-gyp` dependency from `truffle`!

🔥🔥🔥🔥 
 
Resolves #3175.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3316,14 +3316,6 @@ browserify-rsa@^4.0.0:
     bn.js "^4.1.0"
     randombytes "^2.0.1"
 
-browserify-sha3@^0.0.4:
-  version "0.0.4"
-  resolved "https://registry.yarnpkg.com/browserify-sha3/-/browserify-sha3-0.0.4.tgz#086c47b8c82316c9d47022c26185954576dd8e26"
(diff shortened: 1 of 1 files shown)


eggplantzzz:
Unfortunately Truffle still uses `sha3` in a couple of places like [here](https://github.com/trufflesuite/truffle/blob/acf52243f0e83714f5e2bea29de84b250897cfef/packages/deployer/ens.js#L3). Maybe we can find another sha3 utility to use.

CruzMolina:
@eggplantzzz `web3`'s `sha3` method (https://github.com/ethereum/web3.js/blob/1.x/packages/web3-utils/src/utils.js#L466-L492) is not the same as `sha3` (node-sha3) https://github.com/phusion/node-sha3/tree/v1.x.

`sha3` has only been a problem because of the `node-gyp` dependency https://github.com/phusion/node-sha3/blob/v1.x/package.json#L18

PR #3229, opened 2020-08-03
Add deprecation messages for build and watch command

eggplantzzz:
Preparation for deprecation of `build` and `watch` in version 6 of Truffle.

Also a missing call to `callback` was causing build to hang when there is no build script present. This PR also fixes this.


eggplantzzz:
Works towards #3266 and #3267.

PR #3239, opened 2020-08-06
Internal Improvement: Port @truffle/error to TS

CruzMolina:
- Ports `@truffle/error` to TypeScript.


packages/error/src/index.ts
```diff
@@ -0,0 +1,16 @@
+class ExtendableError {
+  message: string;
+  stack: string | undefined;
+  name: string;
+
+  constructor(message: string) {
```
haltman-at, line 6: I'm confused why you redefine these fields (and then set them in the constructor) rather than having `ExtendableError` actually just extend `Error`?  Then again I'm kind of confused as to what the point of this class (as opposed to just using `Error`) is in the first place.  Is this some weird JS thing?

CruzMolina: It _appears_ the original intent of this module was to define a custom `Error` class that could be used to normalize non-standard RPC & `web3.js` error responses. However, it does look like it's used in a bunch of places where a normal `Error` would have sufficed.

My intent here is to just do an initial port. For whatever reason, this was the only way I could get `typescript` to define a custom `Error` class that retained the same properties as a normal `Error`, along with a custom class name.

haltman-at: Wait, really?  I've done `class SomethingOrOther extends Error` in Typescript all the time... there's a bunch of them in decoder and codec...

CruzMolina: So... if you do a check on the custom errors in `decoder/lib/errors`, for example...

```typescript
const example = new ContractBeingDecodedHasNoNodeError("ExampleContract", "someCompilationId");
console.log(example instanceof ContractBeingDecodedHasNoNodeError);
```

returns false! And there are a number of `error instanceof TruffleError` conditionals currently in the codebase.
Also, there are some weird inheritances of this module where `message` gets dropped unless the normal `Error` fields are redefined. 🤷‍♂️

haltman-at: Whaaaaat.  The `error instanceof StopDecodingError` checks in `codec/core.ts` work fine though...

haltman-at: I'm trying this now at the console and not getting any problem...

```
> class SillyError extends Error {}
undefined
> new SillyError() instanceof SillyError
true
```

CruzMolina: I'm guessing it must be a TS to JS interoperability thing?

I'm not sure. If I write that example snippet I shared earlier inside `decoder/lib/errors.ts` and then invoke the compiled JS false, `false` is returned/logged.

haltman-at: Hm, but it seems like it compiles to exactly the thing you say doesn't work...?  [(Example in playground.)](https://www.typescriptlang.org/play?#code/MYGwhgzhAECiAeYC2AHEBTWAnLB7L068ALugHYAmM2eBA3gFDTPQoCuARiAJbDRnJ00ALzRiAC24QAdMFxkIxLG2DF80gUnQBuBgF8gA)  I remain very confused here...

CruzMolina: Check the target https://github.com/trufflesuite/truffle/blob/develop/packages/decoder/tsconfig.json#L6 and compare in the playground.. 👀

haltman-at: Blurgh.  I had been wondering if it might be that but I couldn't find a way to set it in the playground...

haltman-at: Oh yup.  Found how to change it.  Yeah, that's... different...

haltman-at: OK yeah whoa I can totally reproduce this problem now -- completely within Typescript, so long as it's compiled to ES5 or older.  Newer versions seem to work.

haltman-at: Question: Could we just up the target in our `tsconfig.json`?  Looking through our other packages, the targets seem to vary quite a bit; there's a mix of `es5`, `es6`, and `es2015` (which is the same as `es6`).

haltman-at: (Interestingly, the problem is entirely reproducible inside TypeScript; it's not some weird TS/JS boundary thing.)

gnidan: Let's get everything on `es6` if we can (this appears to be supported back to Node v.7.5.0, but someone fact-check me please)

haltman-at: Node.green seems to say that everything in ES6 is supported in Node 10, except for tail call optimization, which is irrelevant.  And almost everything goes back further, the few exceptions basically being irrelevant things.

haltman-at: Like, I seriously doubt anyone is relying on caching of template strings, or the mostly-pointless `Array.values()` function.

gnidan: Cool. Let's make the `es6` change be a separate PR before getting back to the stuff here

CruzMolina: So... I tried upping the target to `es2016` and simplifying this module to be something more like...

```typescript
class ExtendableError extends Error {

  constructor(message: string) {
    super(message);
    this.message = message;
    this.stack = new Error().stack;
    this.name = this.constructor.name;
  }
}
```

but it... it looks like this module gets sometimes gets inherited weirdly so that the `message` continues to get dropped off.

What continues to work is the current code as-is. 🤷‍♂️ 
I can at least bump the target if you'd like.

haltman-at: Huh, weird.  OK, I guess that's the thing to do.  I'm going to go ahead and take this over since you're not here anymore, I guess, but I'm just going to bump the version like you say and leave the rest alone.

Thanks again for taking this on!

haltman-at:
Oh hm.  Bumping the ES version causes problems with assinging to `prototype`.  I have to assume so would `es2015`.  OK, reverting back to `es5`, but leaving in the comment.

gnidan:
Does keeping it at `es5` mean that the `instanceof` check won't work anymore?

haltman-at:
> Does keeping it at `es5` mean that the `instanceof` check won't work anymore?

Oh, crap, you're right.  Geez.  Um...?

packages/error/src/index.ts
haltman-at: Also, sorry, a little rude to ask after you've left, but uh @CruzMolina any chance you could elaborate on what these message problems were?  So I can replicate them and determine whether they're occuring or not?  Thanks!

haltman-at:
OK, I am trying doing things the direct way, even though @CruzMolina said that didn't work.  Uh, we'll see I guess?

haltman-at:
OK, that ran into the problem that a lot of things inherited from `TruffleError` using `inherits`.  So I updated that.

Note: It turns out that `compile-solidity` does something *weird* with its errors' messages.  This might be the source of the problem that @CruzMolina encountered.

haltman-at:
OK, merging this despite red because it's just a Slack notification failure.  Hopefully these supposed message problems don't surface?

PR #3240, opened 2020-08-06
Internal Improvement: Port @truffle/provisioner to TS

CruzMolina:
- Ports `@truffle/provisioner` to TypeScript.


haltman-at:
So I gather there isn't any way to put a better type than `any` on `contractAbstraction` at the moment?  I guess using `contract-schema` doesn't work here as that just covers the JSON, not the extra stuff the abstraction has...

PR #3245, opened 2020-08-07
Internal improvement: Upgrade all TS packages to ES2016

haltman-at:
As requested by @gnidan, to go along with #3239.

Note I standardized on using `"es2015"` rather than `"es6"` since that's the proper name, but they're the same thing.  Only three packages were actually upgraded: `decoder`, `provisioner`, and `source-fetcher`.  All the others were already on `"es6"` and just had the target name standardized.  (And `codec` already had `"es2015"` so that was left alone.)

### DIFF
--- a/packages/artifactor/tsconfig.json
+++ b/packages/artifactor/tsconfig.json
@@ -3,7 +3,7 @@
     "declaration": true,
     "module": "commonjs",
     "esModuleInterop": true,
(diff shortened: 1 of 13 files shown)


haltman-at:
Actually, assuming the tests pass, I'm going to go one further and go to `es2016`.  Not gonna push it any further than that, though. :P

PR #3247, opened 2020-08-10
ethpmV3 support

njgheorghita:
Truffle support for the ethpmV3 spec!

Hey @gnidan and @eggplantzzz . Thanks for your help so far in discussing this. Here's a first pass at ethpmV3 x truffle. I'm sure there's plenty of areas where my javascript is smelly, please point them out anywhere you see them. After a review, I plan to run this through another thorough round of integration tests, & I have an update to the `trufflesuite.com` docs in the pipeline to reflect these changes.


packages/core/lib/commands/packages.js
```diff
@@ -0,0 +1,24 @@
+var command = {
```
njgheorghita, line 1: We discussed separating the `truffle packages` command out into it's own pr, so that we could discuss its api in greater depth, however for the sake of getting an mvp in and across the line, I've left it in here. Happy to remove it out into it's own pr, but it's unclear right now how much more time I have available to extend the `packages` api.

packages/config/src/configDefaults.ts
```diff
@@ -33,11 +33,13 @@ export const getInitialConfig = ({
...
-      ipfs_host: "ipfs.infura.io",
-      ipfs_protocol: "https",
-      registry: "0x8011df4830b4f696cd81393997e5371b93338878",
-      install_provider_uri:
-        "https://ropsten.infura.io/v3/26e88e46be924823983710becd929f36"
+      ipfsHost: "ipfs.infura.io",
+      ipfsProtocol: "https",
+      ipfsPort: "5001",
+      registry: {
+        address: "0x8011df4830b4f696cd81393997e5371b93338878",
```
njgheorghita, line 40: I'm happy to deploy an un-authenticated registry for truffle to use as its default, unless you guys want to handle that. I'm 50/50 on whether it should be on mainnet (better reliability / authority / helps minimize spam packages) or on any testnet (much much cheaper / easier to start getting your hands dirty with publishing ethpm packages)

packages/ethpm-v1/README.md
```diff
@@ -0,0 +1,9 @@
+# `ethpm-v1`
+
+## Usage
+
+```
+const ethpmV1 = require('ethpm-v1');
+```
+
+This package contains support for the deprecated v1 version of the ethpm specification for backwards compatibility purposes.
```
njgheorghita, line 9: Just as a heads-up, I haven't done any editing / testing to the v1 code. It was just copied into its own package from the initial implementation inside the core package. The v1 tests were (and still are) set to be skipped.

packages/ethpm-v3/lib/ethpm-v3.js
```diff
@@ -0,0 +1,485 @@
...
+      .release(ethpmConfig.name, ethpmConfig.version, manifestUri.href)
+      .encodeABI();
+    const tx = await w3.eth.signTransaction({
+      from: fromAddress,
+      to: options.ethpm.registry.address,
+      gas: 4712388,
+      gasPrice: 100000000000,
+      data: encodedTxData
+    });
+    try {
```
njgheorghita, line 465: It seems like there might be a better mechanism available somewhere in truffle that manages sending txs with a more user-friendly interface? Could you point me in the right direction...

packages/core/lib/commands/install.js
```diff
@@ -3,33 +3,43 @@ const command = {
...
+        option: "--alias",
         description:
-          "When specified, will install a specific version of the package, otherwise " +
-          "will install\n                    the latest version."
+          "A different name under which this package will be installed."
       }
     ]
   },
-  run: function(options, done) {
+  run: async function (options, done) {
```
njgheorghita, line 20: When testing locally, I occasionally run into edge cases that hang indefinitely. Is there an existing mechanism to handle these cases? Would it be a good idea to implement some kind of timeout limit when running an ethpm command?

packages/config/src/configDefaults.ts
gnidan: does this field respect ENS names?

packages/core/lib/commands/install.js
```diff
@@ -3,33 +3,43 @@ const command = {
   description: "Install a package from the Ethereum Package Registry",
   builder: {},
   help: {
-    usage: "truffle install <package_name>[@<version>]",
+    usage: "truffle install <packageId> [--alias]",
```
gnidan, line 6:
```suggestion
    usage: "truffle install <package-identifier> [--alias]",
```

packages/core/lib/commands/install.js
```diff
@@ -3,33 +3,43 @@ const command = {
...
     options: [
       {
-        option: "package_name",
-        description:
-          "Name of the package as listed in the Ethereum Package Registry. (required)"
+        option: "packageId",
```
gnidan, line 9:
```suggestion
        option: "packageIdentifier",
```

packages/core/lib/commands/install.js
```diff
@@ -3,33 +3,43 @@ const command = {
...
-  run: function(options, done) {
+  run: async function (options, done) {
     const Config = require("@truffle/config");
-    const Package = require("../package");
+    const PackageV1 = require("@truffle/ethpm-v1");
+    const PackageV3 = require("@truffle/ethpm-v3");
 
-    if (options._ && options._.length > 0) options.packages = options._;
+    if (options._ && options._.length == 0) {
+      done(new Error(`Please provide a packageId.`));
```
gnidan, line 26:
```suggestion
      done(new Error(`Please provide a package identifier.`));
```

Should also probably expand on this error with additional info, like, "Please see `truffle help install` for more information on valid package identifiers."

packages/core/lib/commands/install.js
```diff
@@ -3,33 +3,43 @@ const command = {
...
+    }
+    if (options._ && options._.length > 1) {
+      done(new Error(`Only one packageId can be installed at a time.`));
+    }
+    options.packageId = options._[0];
```
gnidan, line 31:
```suggestion
    options.packageIdentifier = options._[0];
```

packages/core/lib/commands/install.js
```diff
@@ -3,33 +3,43 @@ const command = {
...
 
     const config = Config.detect(options);
-    Package.install(config)
-      .then(() => {
-        return done();
-      })
-      .catch(done);
+
+    if (config.ethpm.version == "1") {
```
gnidan, line 35: The ideal requirement I have in mind here is that `truffle install` will install whatever package, whether it's v1 or v3. Was this not possible? Are we forced to make the user decide between one or the other for installation? (I know that it's required to make this determination ahead of time for publish)

packages/core/lib/commands/packages.js
```diff
@@ -0,0 +1,24 @@
...
+    usage: "truffle packages",
+    options: []
+  },
+  run: async function (options, done) {
+    var Config = require("@truffle/config");
+    var PackageV3 = require("@truffle/ethpm-v3");
+
+    var config = Config.detect(options);
+
+    if (config.ethpm.version == "1" || config.ethpm.version == "3") {
```
gnidan, line 15: Does PackageV3 support listing packages for v1?

packages/core/lib/commands/publish.js
```diff
@@ -6,16 +6,21 @@ var command = {
     usage: "truffle publish",
     options: []
   },
-  run: function(options, done) {
+  run: async function (options, done) {
```
gnidan, line 9: Oh I should note that it's not wise to make a function like this be `async` and take a callback. I'd suggest instead using `callbackify` (`const { callbackify } = require("util")`) and not bothering with the callback arg at all. See [Node `callbackify` docs](https://nodejs.org/api/util.html#util_util_callbackify_original)

packages/core/lib/commands/packages.js
```diff
@@ -0,0 +1,24 @@
+var command = {
+  command: "packages",
+  description: "List available packages on connected EthPM Registry",
+  builder: {},
+  help: {
+    usage: "truffle packages",
+    options: []
+  },
+  run: async function (options, done) {
```
gnidan, line 9: (async functions shouldn't take callbacks also... one or the other, else things can break in subtle ways)

packages/core/lib/commands/install.js
gnidan: This might be because you're making a callback style function also async. Generally not a good idea to do that - use one or the other (see other comment about `callbackify`)

packages/config/src/configDefaults.ts
njgheorghita: Yep, though right now it only permits ens on mainnet. However, upon further investigation it looks like truffle supports ens on all testnets except kovan, so I'll make the update to suppport ens on these other testnets.

packages/core/lib/commands/packages.js
njgheorghita: Yeah good catch, it doesn't, but I've gone ahead and updated `ethpm-v1` to mimic the behavior in `ethpm-v3` to display info about the release on the v1 ethpm registry.

packages/core/lib/commands/install.js
njgheorghita: Yeah, automatic version detection of manifests is not so straightforward, there are a lot of edge cases possible here. One thing to keep in mind is that the read api differs between the v1 registry and v3 registries. I feel we should encourage users to only use v3 (except for installing existing v1 packages). And in that case, they should be encouraged to update their packages to v3, and republish.

I guess this brings up another thought, which is adding a deprecation notice whenever users use v1. Do you have any objections to that?

packages/core/lib/commands/install.js
```diff
@@ -1,35 +1,54 @@
...
   help: {
-    usage: "truffle install <package_name>[@<version>]",
+    usage: "truffle install <package-identifier> [--alias]",
     options: [
       {
-        option: "package_name",
-        description:
-          "Name of the package as listed in the Ethereum Package Registry. (required)"
+        option: "packageIdentifier",
+        description: `(required) Name of the package as listed in the Ethereum Package Registry. Accepted formats: packageName, packageName@version, ethpm URI, ipfs URI.`
```
eggplantzzz, line 12: It is convention in Truffle to put the "(required)" at the end of the description.

packages/core/lib/commands/install.js
```diff
@@ -1,35 +1,54 @@
+const { callbackify } = require("util");
+
 const command = {
   command: "install",
   description: "Install a package from the Ethereum Package Registry",
   builder: {},
   help: {
-    usage: "truffle install <package_name>[@<version>]",
+    usage: "truffle install <package-identifier> [--alias]",
```
eggplantzzz, line 8: Looks like convention in Truffle is snake case when specifying options of this format. So it should be `package_identifiers`.

packages/core/lib/commands/install.js
```diff
@@ -1,35 +1,54 @@
...
     options: [
       {
-        option: "package_name",
-        description:
-          "Name of the package as listed in the Ethereum Package Registry. (required)"
+        option: "packageIdentifier",
```
eggplantzzz, line 11: Same as the above comment

packages/core/lib/commands/install.js
eggplantzzz: I don't have any problem with a deprecation notice. @gnidan?

packages/core/lib/commands/install.js
```diff
@@ -1,35 +1,54 @@
...
     const config = Config.detect(options);
-    Package.install(config)
-      .then(() => {
-        return done();
-      })
-      .catch(done);
+    let callbackFunction;
+
+    if (config.ethpm.version == "1") {
+      callbackFunction = callbackify(PackageV1.install);
```
eggplantzzz, line 45: This should probably be named something like `install` or `installPackage` as it is really not a callback function.

packages/core/lib/commands/packages.js
```diff
@@ -0,0 +1,30 @@
...
+var command = {
+  command: "packages",
+  description: "List available packages on connected EthPM Registry",
+  builder: {},
+  help: {
+    usage: "truffle packages",
+    options: []
+  },
+  run: function (options, done) {
+    var Config = require("@truffle/config");
```
eggplantzzz, line 12: Another small thing: could you rename these `var`s `const`? Anywhere there is a variable that we won't be setting again it should be `const`. `var`s are kinda old school

packages/core/lib/commands/publish.js
```diff
@@ -6,16 +8,22 @@ var command = {
     usage: "truffle publish",
     options: []
   },
-  run: function(options, done) {
+  run: function (options, done) {
     var Config = require("@truffle/config");
```
eggplantzzz, line 12: Same with these guys.

packages/core/lib/commands/publish.js
```diff
@@ -6,16 +8,22 @@ var command = {
...
+    var PackageV1 = require("@truffle/ethpm-v1");
+    var PackageV3 = require("@truffle/ethpm-v3");
 
     var config = Config.detect(options);
-    Package.publish(config)
-      .then(() => {
-        return done();
-      })
-      .catch(done);
+    let callbackFunction;
```
eggplantzzz, line 17: Same as the above about the name of this function.

packages/ethpm-v1/lib/ethpm-v1.js
```diff
@@ -1,16 +1,76 @@
...
+      );
+    }
+
+    const ethpmV1Network = "ropsten";
+    if (!options.networks[ethpmV1Network]) {
+      throw new TruffleError(
+        `Please include a provider in your networks config for the Ropsten testnet, the testnet where the ethpmV1 registry is deployed.`
+      );
+    }
+    const provider = options.networks["ropsten"].provider();
```
eggplantzzz, line 38: Will this always be a method? I feel like this property is usually accessed like `options.networks[<name>].provider`. I could be wrong, I'll have to check this out.

packages/ethpm-v1/lib/ethpm-v1.js
```diff
@@ -1,16 +1,76 @@
...
+    const web3 = new Web3(provider);
+    const ethpmV1RegistryAddress = "0x8011df4830b4f696cd81393997e5371b93338878";
+    var contract = new web3.eth.Contract(
```
eggplantzzz, line 41: Do we want to use `@truffle/contract` instead of `web3` for this?

packages/ethpm-v3/lib/utils.js
```diff
@@ -0,0 +1,301 @@
...
+    var msg = `Could not locate any publishable artifacts in ${options.contracts_build_directory}. Run "truffle compile" before publishing.`;
+    return new TruffleError(msg);
+  }
+
+  const detectedNetworkIdsToBlockchainUris = {};
+  const detectedArtifacts = {};
+  const sourcePaths = {};
+
+  // group artifacts by network id to ensure consistency when converting networkId => blockchainUri
+  for (let file of files) {
```
eggplantzzz, line 42: You can use `const` (like `for (const file of files) {`) here as well.

packages/ethpm-v3/lib/utils.js
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  // generate blockchainURI for each detected networkId
+  for (let networkId of requiredNetworkIds) {
+    let targetProvider;
+    for (let provider of Object.keys(options.networks)) {
+      if (options.networks[provider].network_id == networkId) {
+        // handle ganache provider from tests
+        if (
+          options.networks[provider].provider.constructor.name == "Function"
```
eggplantzzz, line 76: It is probably better to use `===` which is strict equals.

You could also do 
```
if (typeof options.networks[provider].provider === "function") {
```
if you like but I suppose your way is good too. Actually, how is this indexed by `provider`? Is that really some provider value?

packages/ethpm-v3/lib/utils.js
```diff
@@ -0,0 +1,301 @@
...
+    resolvedSources: sources,
+    resolvedContractTypes: parsedArtifacts.contractTypes,
+    resolvedDeployments: parsedArtifacts.deployments,
+    resolvedCompilers: parsedArtifacts.compilers
+  };
+}
+
+async function resolveEnsName(address, provider, options) {
+  let resolvedAddress;
+  const w3 = new Web3(provider);
```
eggplantzzz, line 118: Convention in Truffle is to use `web3` as the variable name for the instance of `Web3`.

packages/ethpm-v3/lib/utils.js
```diff
@@ -0,0 +1,301 @@
...
+      fs.readFileSync(
+        path.join(workingDirectory, "_ethpm_packages", "ethpm.lock"),
+        "utf8"
+      )
+    );
+  } catch (err) {
+    return {};
+  }
+  const installedBuildDependencies = {};
+  Object.keys(ethpmLock).forEach(function (key, _) {
```
eggplantzzz, line 202: You can just omit the 2nd argument (`_`) in this case.

An arrow function would be slickest.
```
Object.keys(ethpmLock).forEach(key => installedBuildDependencies[key] = ethpmLock[key].resolved_uri)
```

packages/core/lib/commands/install.js
njgheorghita: :+1: I've made this change - but just want to note that `package-identifier` was suggested above by @gnidan.

packages/core/lib/commands/install.js
njgheorghita: :+1: I've made this change - but just want to note that `packageIdentifier` was suggested above by @gnidan.

packages/ethpm-v3/lib/utils.js
njgheorghita: I've updated it to your way - this was a bit of a hack to handle ganache provider instances in the tests, while also supporting users setting provider methods as provider in `truffle-config`. Though (as in the other provider-related comment) I could be misunderstanding how users typically set their provider in `truffle-config`.

njgheorghita:
@eggplantzzz Thanks for the review! I've changed or responded to your comments. Let me know if there's anything else glaring / obvious that needs addressing. In terms of the failing tests... Has there been any significant changes to the resolver recently? It seems to me that the `await Migrate.run(config)` step is failing in the tests - and it's not obvious the best way to fix it, but I'll be able to dig a little deeper into it tomorrow if it's not obvious to you what changed.

eggplantzzz:
Sorry about jerking you back and forth on the `package-identifier` stuff, I didn't see @gnidan's suggestion about it until afterwards haha.

eggplantzzz:
As for the CI failures, I don't think anything changed with the migrations. I'll take a quick look and see if I can find anything out about this.

eggplantzzz:
So @njgheorghita, I took a look and your tests are failing because there were a bunch of breaking changes to some of the compilation packages. `workflow-compile` had compilation and the saving of artifacts separated into two steps. I made the necessary corrections and pushed them to a branch in the Truffle repo called `fix-ethpm-tests`. I didn't want to push directly to your branch but feel free to merge it into yours or fix it yourself. The gist of the matter is that `Contracts.compile` no longer saves the artifacts. There is now a method `Contracts.compileAndSave` that you need to use to do both. We also started using `WorkflowCompile` as the name when importing that library and I made that change in there as well. Let me know if something is unclear or if you need anything else!

njgheorghita:
@eggplantzzz Nailed it! Thanks for looking into it and the fix!

packages/core/lib/commands/install.js
eggplantzzz: Can you still make this change? It should be snake case.

njgheorghita:
@eggplantzzz I can definitely update `packageIdentifier` -> `package_identifier`. Just to make sure I'm not messing up anything here - what's the reason to merge the `develop` branch into this one? Is it fine to continue building on top of it?

njgheorghita:
I've also updated the ethpm x truffle docs [here](https://github.com/trufflesuite/trufflesuite.com/pull/620) and added my blog (which is fairly similar to the docs but has a better example and more in-depth analysis) [here](https://github.com/trufflesuite/trufflesuite.com/pull/817).

eggplantzzz:
@njgheorghita Oh, there was just a merge conflict so I resolved it for you. I hope you don't mind. And thanks for the fix!

PR #3248, opened 2020-08-11
Internal improvement: Allow test suite success on Slack notification failure

haltman-at:
If the Slack notification fails, that shouldn't cause a test failure.  This change to the GHA YAML file should make that the case, if I've understood correctly.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -47,6 +47,7 @@ jobs:
   slack_notification:
     needs: build
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


haltman-at:
Well, evidently that didn't work!  Hm...

haltman-at:
OK, I've added it in a second place.  This shouldn't make any difference, but...

haltman-at:
Aha, this time it worked!  If you click through you can see that the Slack notification errored -- and yet the PR is green all the same! :)

Issue #2463, opened 2019-10-16
Problem running `test` from `truffle develop`


eggplantzzz:
## Issue

`truffle test` fails when run from `truffle develop`

## Steps to Reproduce

1. In a new directory, run `truffle unbox metacoin`
2. After unboxing, run `truffle develop`
3. When the prompt appears, run `test`

## Expected Behavior

I expect the tests to run

## Actual Results

```
TypeError [ERR_INVALID_REPL_INPUT]: Listeners for `uncaughtException` cannot be used in the REPL
    at process.<anonymous> (repl.js:224:15)
    at process.emit (events.js:215:7)
    at process.emit (/Users/tyler/.nvm/versions/node/v12.11.0/lib/node_modules/truffle/build/webpack:/~/source-map-support/source-map-support.js:465:1)
    at _addListener (events.js:236:14)
    at process.addListener (events.js:284:10)
    at Runner.run (/Users/tyler/.nvm/versions/node/v12.11.0/lib/node_modules/truffle/node_modules/mocha/lib/runner.js:868:11)
    at Mocha.run (/Users/tyler/.nvm/versions/node/v12.11.0/lib/node_modules/truffle/node_modules/mocha/lib/mocha.js:612:17)
    at /Users/tyler/.nvm/versions/node/v12.11.0/lib/node_modules/truffle/build/webpack:/packages/core/lib/test.js:116:1
    at new Promise (<anonymous>)
    at Object.run (/Users/tyler/.nvm/versions/node/v12.11.0/lib/node_modules/truffle/build/webpack:/packages/core/lib/test.js:115:1)
```

## Environment

* Operating System: Mac OS
* Ethereum client: truffle develop
* Truffle version (`truffle version`): 5.0.40
* node version (`node --version`): 12
* npm version (`npm --version`):  6

eggplantzzz:
This bug is specific to Node 12 and is present in `5.0.39` and `5.0.38`.

eggplantzzz:
Looks like it was referenced by David Burela here https://github.com/trufflesuite/truffle/issues/2070

spore-engineering:
hello there, im having the same issue with Truffle v5.1.28, same behavior, just clarify the bug is present not only in 5.0.38 and 5.0.39.

Just made a fresh instalation following those steps, all works fine but the test command :) https://www.trufflesuite.com/boxes/drizzle

bests :)

-----------------------------------------

[...]

PR #3255, opened 2020-08-13
Run each console command in its own child process

fainashalts:
This PR runs commands from inside the `truffle develop` and `truffle console` command processes. While this PR changes the way `truffle develop` and `truffle console` work under the hood, the end user CLI experience should not be different. 

These changes became necessary when we found that Node v12 does not allow `UncaughtException` listeners in a REPL within another NodeJS program, and that one of our dependencies uses such a listener. By running a separate REPL in a child process we are able to bypass this problem. The issue that brought this to our attention is here: https://github.com/trufflesuite/truffle/issues/2463

In the process of this work, it became apparent that the ReplManager class we had been using to manage multiple REPLs was no longer necessary since each REPL will now be in its own process. This PR thus also removes the `repl.js` file and directly accesses the node REPL as needed in `console.js` and `interpreter.js`. 

To test this PR, run either `truffle develop` or `truffle console` and then use these to test out any other Truffle commands that make sense: `migrate --reset`, `compile`, `test`, `debug`, etc. Everything should work as it did previously, with the exception that the `--network` flag will throw an error when used with `truffle develop` -- the only network that should be used by `truffle develop` is the `develop` network.


packages/core/index.js
```diff
@@ -4,7 +4,6 @@ var pkg = require("./package.json");
 module.exports = {
   build: require("./lib/build"),
   create: require("./lib/commands/create/helpers"),
-  console: require("./lib/repl"),
```
fainashalts, line 7: @gnidan I removed this because it wasn't being used anywhere, but now I see it is in the truffle scenario tests, which are failing in GHA. Removing this test solves it, but I wanted to check in first in case I'm missing something: https://github.com/trufflesuite/truffle/blob/develop/packages/truffle/test/scenarios/library/api.js#L26
Ok to remove?

packages/core/lib/console.js
```diff
@@ -143,16 +126,53 @@ class Console extends EventEmitter {
...
+      );
+
+      try {
+        this.provision();
+      } catch (e) {
+        console.log(e);
+      }
+    } catch (err) {
+      callback(err);
+    }
```
cds-amal, line 161: Is the intention to eat exceptions thrown by `this.provision()` and not invoke the callback on 160?

packages/core/index.js
eggplantzzz: No, I think this is being exposed so that users can pull it out and use it separately.

packages/core/lib/debug/interpreter.js
```diff
@@ -304,10 +293,7 @@ class DebugInterpreter {
     }
 
     //split arguments for commands that want that; split on runs of spaces
-    splitArgs = cmd
-      .trim()
-      .split(/ +/)
-      .slice(1);
+    splitArgs = cmd.trim().split(/ +/).slice(1);
```
cds-amal, line 296: Note this does not account for a quoted argument with spaces. I know this isn't part of the work done in the PR but wanted to surface it

packages/core/lib/console-child.js
```diff
@@ -0,0 +1,51 @@
...
+const TruffleError = require("@truffle/error");
+const Config = require("@truffle/config");
+const Web3 = require("web3");
+
+const inputStrings = process.argv[2];
+//deserialize options
+const options = JSON.parse(process.argv[3]);
+
+//deserialize networks entries
+for (var key in options.networks) {
```
eggplantzzz, line 11: Can we use `const key`?

packages/core/lib/console-child.js
```diff
@@ -0,0 +1,51 @@
...
+  options.networks[key] = JSON.parse(options.networks[key]);
+}
+
+//detect config so we can get the provider and resolver without having to serialize
+//and deserialize them
+const configDetected = Config.detect(options);
```
eggplantzzz, line 17: What about naming this `detectedConfig`? `configDetected` sort of sounds like a predicate.

packages/core/lib/console.js
```diff
@@ -143,16 +126,53 @@ class Console extends EventEmitter {
...
+
+    const spawnOptions = { stdio: ["inherit", "inherit", "inherit"] };
 
-    this.repl.setContextVars(contextVars);
+    try {
+      //serialize config networks entries
+      const configNetworks = {};
+      for (const key in options.networks) {
+        configNetworks[key] = JSON.stringify(options.networks[key]);
+      }
```
cds-amal, line 146: How many network properties are used in the child side? If it's 2-5 can this be changed to explicitly name them as arguments to spawnSync? A benefit of this approach would be to document what props are being used.
```
const networkProps = [ '--networkProp1', value1, '--networkProp2', 'value2'...];
SpawnSync("node", [...otherArgs, ...networkProps])
```

packages/core/lib/console.js
```diff
@@ -143,16 +126,53 @@ class Console extends EventEmitter {
     abstractions.forEach(abstraction => {
       contextVars[abstraction.contract_name] = abstraction;
     });
+  }
+
+  runSpawn(inputStrings, options, callback) {
+    let args;
```
cds-amal, line 132: `args` is confusing in this context because It implies some none zero number of arguments when, in this case, it is the module/entryPoint to spawn

packages/core/index.js
fainashalts: So I wound up taking out the test, after discussing with gnidan. The repl manager doesn't exist anymore so that export isn't really meaningful anymore. Let me know if you want to discuss this more!

packages/core/lib/console.js
```diff
@@ -54,35 +56,16 @@ class Console extends EventEmitter {
...
+      this.interfaceAdapter.getAccounts().then(() => {
         const abstractions = this.provision();
 
-        this.repl.start({
+        this.repl = repl.start({
           prompt: "truffle(" + this.options.network + ")> ",
-          context: {
-            web3: this.web3,
-            interfaceAdapter: this.interfaceAdapter,
-            accounts: fetchedAccounts
```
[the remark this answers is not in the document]
fainashalts: Totally right. Fixed!

packages/core/lib/console.js
fainashalts: Unless I'm misunderstanding, the callback on lime 160 is for any exceptions thrown by spawning the repl. There's a try/catch for `this.provision()` and a try/catch for the spawning. Does that make sense or am I not following?

packages/core/lib/debug/interpreter.js
fainashalts: @cds-amal can you make a separate issue for this?

packages/core/lib/console.js
fainashalts: Agree. Renamed to `childPath` to be more clear.

packages/core/lib/console-child.js
fainashalts: I removed this part of the code, so no need! 🎉

packages/core/lib/console-child.js
fainashalts: Agree. Changed.

gnidan:
Hey @fainashalts, check out the bundled version of Truffle - there's a deprecation notice in every child process:

```
truffle(develop)> test
(node:96195) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
Using network 'develop'.

```

packages/core/lib/console.js
```diff
@@ -143,16 +130,53 @@ class Console extends EventEmitter {
...
+
+    const spawnOptions = { stdio: ["inherit", "inherit", "inherit"] };
 
-    this.repl.setContextVars(contextVars);
+    try {
+      //serialize config networks entries
+      const configNetworks = {};
+      for (const key in options.networks) {
+        configNetworks[key] = JSON.stringify(options.networks[key]);
+      }
```
gnidan, line 150: `configNetworks` appears to be unused

packages/core/lib/debug/interpreter.js
gnidan: note that this is the debug repl, I'm not sure what behavior you're describing @cds-amal, but the use case might be different here.

cds-amal: Oh. If there's an expectation for no quoted arguments then all is good. The parsing behavior I tried to explain in the following block.  

```
> cmd = "command arg1 \"arg two\" arg3"
'command arg1 "arg two" arg3'
> cmd.trim().split(/ +/).slice(1)
[ 'arg1', '"arg', 'two"', 'arg3' ]

```

packages/core/lib/console.js
cds-amal: The intention is a bit hard to follow because the two catch expressions are next to each other.  As it exists, if the provision after spawn fails then the code explicitly eats the error and continues execution to the next REPL prompt. Is an exception thrown by `this.provision()` benign as the code suggests?

packages/core/lib/debug/interpreter.js
gnidan: Yep, the debugger doesn't really provide that level of sophistication! (Maybe it should? But I think we can ignore for now)

packages/core/lib/console-child.js
```diff
@@ -0,0 +1,45 @@
+const Command = require("../lib/command");
+const TruffleError = require("@truffle/error");
+const Config = require("@truffle/config");
+const Web3 = require("web3");
+
+const inputStrings = process.argv[2];
+const network = process.argv[3];
```
gnidan, line 7: Can we merge these two? i.e., if the user runs `debug --network mainnet` from inside `truffle console --network development`, then I would expect the `debug` command to connect to mainnet. This implementation suggests that behavior might be undefined (or favor the network specified by `process.argv[3]`)

Seems like we shouldn't pass `network` separately as a positional argument, and instead pass it as a named `--network` option. Two options come to mind for how to do that:

1. Add logic to the parent process to merge `--network=<network-name>` into `inputStrings`, possibly by decoding `inputStrings` into a mapping + then re-encoding with `{ network }` merged in.

2. Use a ` -- ` separator between parent-process-provided options and user-provided options, e.g.:
  ```
  truffle(develop)> debug <tx-hash> --fetch-external
     ... would run command:
     $ console-child.js --network=develop -- <tx-hash> --fetch-external
  ```

fainashalts: Hi @gnidan!  https://github.com/trufflesuite/truffle/pull/3296 addresses this in case you do want to get rid of the positional argument. Wanted to save you a little time!

gnidan: Thanks so much @fainashalts, thanks for knocking that out!

That PR looks great! But why is it failing ☹️ ? (Re-running now)

packages/core/lib/console.js
cds-amal: `this.provision();` can throw an Error (line 110) if something is awry with an artifact file. I think we should be invoke the callback in the outer catch block.

PR #3259, opened 2020-08-13
Update ethpm-js to resolve redirect issue

haltman-at:
Hoping this solves the redirect problem and gets the tests passing.  We'll see though.

(The substance of this is updating `wget-improved` from 1.4.0 to 1.5.0; unclear if that will fix things.)


haltman-at:
OK, bumping ethpm-js one more time, this time with version 3.2.1 of `wget-improved`; we'll see if this works.

haltman-at:
OK then, let's try this version, which tears out `wget-improved` in favor of `request`...

PR #3265, opened 2020-08-14
Enhancement: Broader contract recognition (treat all contracts the way we treated externally-downloaded ones)

haltman-at:
**Edit: This PR is now a breaking change to `@truffle/codec`**

This PR removes the `externalSolidity` flag from compilations that was used to indicate, hey, when performing source-matching against contracts in this compilation, you should ignore metadata hashes.  Now we just treat everything that way.  Nothing is added; I just deleted the flag, and all code that sets it or checks it, in favor of always treating it like it's `true`.  I also updated comments.

Doing this is safe because, as of #3212, if a contract isn't Solidity (or is extremely old Solidity), we won't get valid CBOR from it and thus it'll be ignored.  So despite my earlier worries, this should be fine!


haltman-at:
OK, the tests are failing here because of a test I included where I did `contract ExternalsDerived extends ExternalsBase {}`.  So the two classes are the same except for metadata, meaning that with this PR, the debugger & decoder can no longer distinguish the two.

It's not clear what the right course of action is here.  We could say that's unacceptable and scrap this PR; or we could say that's an OK price to pay in order to have broader contract recognition, and adjust the tests to make those two classes different in some way.  However it's not obvious to me right now which to do.  So I'm just leaving this here for now.

haltman-at:
OK, so after discussing with @gnidan, I've resurrected this PR.  I made the test pass just by making the two contexts distinct (adding a dummy function to the descendant).  But what about the problems with this approach?

Well, we decided that we'd handle this as follows: In case of an ambiguity, always have the decoder resolve it in favor of the base contract, rather than the derived contract.  (If that *still* leaves ambiguity, then, yeah, pick arbitrarily.)

As part of doing this, I added the `linearizedBaseContracts` field to the `Context` type, so that the matcher could judge this appropriately.

In doing this, the distinction between `DecoderContext` and `DebuggerContext` finally became too much of a pain, so I renamed `DecoderContext` to `Context` (and `findDecoderContext` to `findContext`) and eliminated the `DebuggerContext` type.  Well, OK, no --  I left it around for reference, but the type's not actually used anywhere, other than implictly in the debugger.  How did I handle the unification on the debugger sie?  Simple, I just put `//HACK` comments on places in the debugger where we invoke Codec functions that expect a `Context` even though we're actually passing in a `DebuggerContext`! :P  Hopefully that's OK...

But yeah, I think this is good to go now.

...also `prettier` made a bunch of changes in the course of merging `develop` into this...

gnidan (approved):

packages/codec/lib/contexts/types.ts
```diff
@@ -35,9 +28,10 @@ export interface DecoderContext {
   };
   compiler?: Compiler.CompilerVersion;
   compilationId?: string;
-  externalSolidity?: boolean; //please only set for Solidity contracts!
 }
 
+//NOTE: this is being kept for reference. but we don't actually use this
+//type anywhere anymore.
```
gnidan, line 34: could you note that "for reference" means "because the debugger is JS and won't import types"?

packages/truffle/test/sources/exec/script.js
```diff
@@ -7,7 +7,7 @@ const exec = function () {
       console.log(parseInt(val));
       process.exit(0);
     })
-    .catch(err => process.exit(1));
+    .catch(_err => process.exit(1));
```
gnidan, line 10: Could just do
```suggestion
    .catch(() => process.exit(1));
```

gnidan (approved):
Looks good. One minor note about a comment and a style suggestion

packages/codec/lib/contexts/types.ts
haltman-at: Sure thing!

Issue #3272, opened 2020-08-17
Cannot find @truffle/error when requiring @truffle/contract


spalladino:
## Issue

When installing @truffle/contract, the @truffle/error dependency installed is empty, so requiring @truffle/contract fails with the error:
```
Cannot find module '.../node_modules/@truffle/error/dist/src/index.js'
```

## Steps to Reproduce

- Install via `npm init -y && npm i @truffle/contract` (
- Bring up a console with `node`
- Run `require('@truffle/contract')`

## Actual Results

```
Error: Cannot find module '/home/spalladino/Projects/sdk/truffle/check-error/node_modules/@truffle/error/dist/src/index.js'. Please verify that the package.json has a valid "main" entry
    at require (internal/modules/cjs/helpers.js:72:18)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at Function.Module._load (internal/modules/cjs/loader.js:878:14)
    at Module.load (internal/modules/cjs/loader.js:985:32)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1157:10)
    at Module._compile (internal/modules/cjs/loader.js:1137:30)
    at Object.<anonymous> (/home/spalladino/Projects/sdk/truffle/check-error/node_modules/@truffle/contract/lib/statuserror.js:1:22)
    at require (internal/modules/cjs/helpers.js:72:18)
    at Module.require (internal/modules/cjs/loader.js:1025:19)
    at Function.Module._load (internal/modules/cjs/loader.js:841:27)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:952:27)
    at Function.Module._findPath (internal/modules/cjs/loader.js:682:18)
    at tryPackage (internal/modules/cjs/loader.js:321:19) {
  code: 'MODULE_NOT_FOUND',
  path: '/home/spalladino/Projects/sdk/truffle/check-error/node_modules/@truffle/error/package.json',
  requestPath: '@truffle/error'
}
```

## Notes

This could have been caused by [this changeset](https://github.com/trufflesuite/truffle/pull/3239/files#diff-dd4b6d2d8e7b2f2a754559d360072f67), which modified the `package.json` of truffle error a few days ago.

When checking the contents of @truffle/error in the node_modules folder, it seems that there are no sources included:
```
$ ll node_modules/@truffle/error
drwxr-xr-x 2 spalladino spalladino 4096 Aug 17 13:10 ./
[...]

PR #3273, opened 2020-08-17
Fix @truffle/error package.json

gnidan:
#3272 

Seems like @truffle/error wasn't getting of its files published!

This PR should fix that. Tested this using `yarn pack` and seems to work. Assuming this is good to go, will be pushing out an early Monday night release tonight.


gnidan:
Merging this to get an emergency patch release out tonight

PR #3275, opened 2020-08-18
Replace new Buffer() calls via patch to avoid deprecation warnings:

fainashalts:
This PR attempts to address node deprecation warnings for the use of `new Buffer()` via `patch-package`, which allows us to make the changes to packages we depend on but cannot directly change at this time. With this change, a diff has been created that will be applies to these packages before they are built via webpack. There is also a change in `codec/lib/conversion.ts` with the same goal of avoiding the deprecation warning.

- Replace new Buffer() in relevant node_modules
- Replace Buffer with Buffer.alloc in codec/lib/conversion.ts
- Add patch-package to prepare script

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "private": true,
   "scripts": {
     "bootstrap": "npx lerna bootstrap",
-    "prepare": "lerna run prepare --stream --concurrency=1",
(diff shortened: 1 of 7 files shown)


gnidan:
We decided not to take this approach, right?

fainashalts:
Yep, we decided to avoid patching as too messy. Will close this PR!

PR #3276, opened 2020-08-19
Change Buffer to Uint8Array to avoid deprecation warning

fainashalts:
This PR changes the Buffer type being passed to BN.js via the `toArrayLike()`method, because somehow this was triggering a deprecation warning from node in the `spawn-repl` branch. A Buffer is technically a Uint8Array, so I think this should be ok, but would love someone with a better understanding of the `codec` code to take a look and make sure this doesn't break anything.

### DIFF
--- a/packages/codec/lib/conversion.ts
+++ b/packages/codec/lib/conversion.ts
@@ -143,7 +143,7 @@ export function toBytes(
 
     //note that the argument for toTwos is given in bits
     return new Uint8Array(
-      data.toTwos(length * 8).toArrayLike(Buffer, "be", length)
+      data.toTwos(length * 8).toArrayLike(Uint8Array as any, "be", length)
     ); //big-endian
   }
 }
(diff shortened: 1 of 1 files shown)


fainashalts:
@haltman-at I thought about that same redundancy, but since `toArrayLike` needs the type to be passed in this way I didn't see an elegant way around it. Are you saying we should scrap `toArrayLike` and replace it with something else? Or are you ok with merging this as is?

haltman-at:
I'm not saying to drop the `toArrayLike`; I'm saying to drop the `new Uint8Array()` that wraps it.  Like, `.toArrayLike(Uint8Array)` (possibly with a type coercion if TS is being dumb) should suffice to convert to a `Uint8Array`, right?  Why then stick a `new Uint8Array()` around that?

But yeah if that doesn't bother you then you can go ahead and just merge this as-is, that's what "approved" means. :)

haltman-at:
Ah nice, glad Typescript didn't give any problems here!

fainashalts:
Yep! Super easy change, thanks for catching it!

PR #3278, opened 2020-08-19
Move develop network defaults to @truffle/config

gnidan:
- Specify `develop` defaults in configDefaults
- Ensure we mark `networks` as object for "deep copy"
- Remove defaults from lib/commands/develop

### DIFF
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -20,7 +20,18 @@ export const getInitialConfig = ({
     truffle_directory,
     working_directory,
     network,
-    networks: {},
+    networks: {
+      develop: {
+        host: "127.0.0.1",
+        port: 9545,
+        network_id: 5777,
+        accounts: 10,
+        defaultEtherBalance: 100,
+        blockTime: 0,
+        gas: 0x6691b7,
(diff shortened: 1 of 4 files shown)


gnidan:
Looks like this uncovered a tricky problem with `truffle networks`. For background, the current requirements of that command + `truffle develop` are that the `networks` command should show `develop` iff the user runs that command from within the `truffle develop` console.

Since we added the option of configuring the `develop` network, there's now arguably a bug in that behavior, so `develop` will show up when `truffle develop` isn't running, iff you have a custom network config.

The right way to do this is likely to mark certain networks as "managed" by Truffle (e.g. `develop`, also `test`), and then figure out a way for Truffle to reliably recognize when managed networks are currently running, either in the foreground or the background, and only then display those networks in the `networks` command output.

Now, **the reason for this PR** was to remove duplicated logic in the `spawn-repl` branch, since that branch redefines `develop` defaults in `console-child.js`. That's slightly annoying (and so is the existing bug with `truffle networks` + `develop` network), but doing anything here would require a nontrivial amount of work, where a half-solution would likely be a hack worse than those two driving-factor inconveniences combined.

Closing this PR.

Issue #3279, opened 2020-08-20
Cannot find @truffle/provisioner when requiring @truffle/resolver 


yacare555:
## Issue

When installing @truffle/resolver, the @truffle/provisioner dependency installed is empty.

## Steps to Reproduce

- yarn add @truffle/resolver 
- Open a node console
- require('@truffle/provisioner')
- You get: Error: Cannot find module

PR #3280, opened 2020-08-20
Fix @truffle/provisioner package.json

yacare555:
A possible fix to #3279, the solution it's based in #3273


eggplantzzz:
Thanks very much @yacare555! Looks good!

PR #3281, opened 2020-08-20
feat(box): allow unboxing from local folder

troggy:
Make `truffle unbox` command to accept path to local folder with box as well. This could be useful to test your boxes in the same environment as your box users will be.

Usage example:

```sh
truffle unbox ../yourLocalBox

# or 
truffle unbox /Users/doe/projects/yourLocalBox

# or
truffle unbox ~/projects/yourLocalBox
``` 

Here is the demo: https://asciinema.org/a/IC3AZnCrmyRCtjCZHEsulDyIb

Let me know if this needs to be rebased on `next` branch.


eggplantzzz:
Thanks for the PR @troggy! We'll give it a round of reviews, I think this will be a cool feature to add.

packages/box/test/mock-local-box/truffle-box.json
```diff
@@ -0,0 +1,12 @@
+{
```
eggplantzzz, line 1: Could you put the mock project in a `sources` folder? It is a convention in the Truffle codebase to include all fixtures in a `sources` folder in the `test` folder.

packages/box/lib/utils/unbox.ts
```diff
@@ -9,7 +9,14 @@ import inquirer from "inquirer";
 import { boxConfig, unboxOptions } from "typings";
 import { promisify } from "util";
 
-async function verifyURL(url: string) {
+function verifyLocalPath(url: string) {
+  const configPath = path.join(url, "truffle-box.json");
```
eggplantzzz, line 13: We probably should name this variable `path` and not `url`. What do you think?

packages/box/lib/utils/unbox.ts
```diff
@@ -37,7 +44,18 @@ async function verifyURL(url: string) {
   }
 }
 
+async function verifyURL(url: string) {
+  if (url.startsWith("/")) {
+    return verifyLocalPath(url);
+  }
+  return verifyGithubURL(url);
```
eggplantzzz, line 51: I'm not sure we should make this specific to github. I think these should work with services like bitbucket as well.

packages/box/lib/utils/unbox.ts
troggy: yes, it was my first intention. However it clashes with Node's `path` declared already in a upper scope (and used in this very same function on line 13)

packages/box/lib/utils/unbox.ts
```diff
@@ -9,7 +9,14 @@ import inquirer from "inquirer";
...
 
-async function verifyURL(url: string) {
+function verifyLocalPath(url: string) {
+  const configPath = path.join(url, "truffle-box.json");
+  fse.access(configPath).catch(e => {
+    throw new Error(`Truffle Box at path ${url} doesn't exist.`);
+  });
+}
+
+async function verifyGithubURL(url: string) {
```
eggplantzzz, line 19: I don't think we should make this particular to GitHub as I think urls for Bitbucket will work as well.

troggy: yeah, but the current implementation (I didn't touch it at all) works only for github, that's why I named it like that. It could be made to work with any repo (actually `download-git-repo` has code to verify such URL already), but it is out of scope for this PR

packages/box/box.ts
```diff
@@ -18,6 +18,9 @@ import { sandboxOptions, unboxOptions } from "typings";
 const normalizeURL = (
```
eggplantzzz, line 18: Perhaps this method should be changed to something like `normalizeTarget` to signify that it won't necessarily be just a url. Thoughts?

packages/box/lib/utils/unbox.ts
eggplantzzz: Ah yes! How about `targetPath`?

troggy: probably `sourcePath` as the target can be confused with destination?

packages/box/lib/utils/unbox.ts
eggplantzzz: I think the current implementation will actually work for Bitbucket. `normalizeUrl` will just pass a full url through [here](https://github.com/trufflesuite/truffle/blob/develop/packages/box/box.ts#L34-L37). Then that url gets passed directly to `download-git-repo` [here](https://github.com/trufflesuite/truffle/blob/develop/packages/box/lib/utils/unbox.ts#L40-L42). So it should support full Bitbucket urls as well. Maybe I missed something?

troggy: Yeah, you are right.. I got confused by the [line 24](https://github.com/trufflesuite/truffle/pull/3281/files#diff-a853c7f30d7bb0a9d440cdc429f87ee6R24) and thus thought it is only for Github

troggy:
I've addressed the naming comments + added support for ~/ like paths. I've noticed my changes break some tests in `core` package — I'm on it. Plus I'd like to make unbox to exclude `.gitignore`d files when unboxing local box

eggplantzzz:
Thanks @troggy!

PR #3297, opened 2020-08-25
Merge `truffle-db` into `develop`

gnidan:
This PR seeks to merge the long-running `truffle-db` feature branch into `develop`.

### Motivation

This branch is way behind and has too many merge commits. Rebasing has repeatedly proven impossible. Seems like we can probably get this into `develop` without any impact on the rest of Truffle. Seems wise to do that before it just keeps getting worse.

### Status of the code on the `truffle-db` branch 

The code changes on this branch are effectively already reviewed:

- Process diligence (i.e., PR + code review + CI checks) has been followed since March, 2019, when @fainashalts started contributing to this effort.
- Before that point, I authored commits directly to this branch. That code did not receive formal review, but I posit that such implementation review acceptance criteria has been met through informal review and through follow-on work between @fainashalts, @cds-amal, and myself.

(a.k.a reviewers of this PR: you don't need to read all the code if you agree with the above reasoning)

### Additional merge acceptance criteria

In additional to usual process diligence, this should meet some additional criteria. Will plan to check these off as they are verified.

- [x] No effect on code bundles outside @truffle/db. (e.g. Truffle's `cli.bundled.js` is not any bigger)
- [x] When installing any existing package, no additional dependencies are required
- [x] No Truffle DB behavior is available except via the `@truffle/db` package

### Expected status of package `@truffle/db` post-merge

That package being in `develop` means it will go out in weekly releases. Let's keep those releases as `0.1.0-<prerelease>` for now.

- [x] Make sure @truffle/db uses prerelease versioning to start (PR #3304)

### Known changes outside @truffle/db

Most of this work is in the `packages/db/` directory, but I'll be going through an enumerating other changes here.

#### New `truffle db` command. This is implemented already, but work remains to keep Truffle behavior the same.

  - [x] Stop `require()`ing the module(s) for this command so webpack doesn't find them. (PR #3328)

#### There are changes to `truffle help` to support subcommands (see [relevant part of diff](https://github.com/trufflesuite/truffle/pull/3297/files#diff-ac8e868e41d742600ae7e42271392b6c)).

-  [x] Get those changes into `develop` first (or determine another strategy) (PR #3328)

#### Packages `truffle` and `@truffle/core` depend on `@truffle/db`

  - [x] Stop that. PR: https://github.com/trufflesuite/truffle/pull/3299

#### @truffle-error and @truffle/provisioner have `typeRoots` in their tsconfig on `truffle-db` but not on `develop`

  - [x] Determine if this is still necessary and get it onto `develop` first.

#### truffle is configured to treat @truffle/db as a webpack external

  - [x] Determine if that matters and stop it if so. (PR #3328)


packages/db/bin/server.js
```diff
@@ -0,0 +1,26 @@
...
+
+const { schema, context } = db;
+
+const server = new ApolloServer({
+  tracing: true,
+  schema,
+  context
+});
+
+// This `listen` method launches a web-server.  Existing apps
```
eggplantzzz, line 22: Get rid of this comment :)

packages/db/bin/server.js
```diff
@@ -0,0 +1,26 @@
...
+const config = Config.detect({
+  workingDirectory: process.argv[2] || process.cwd()
+});
+
+const db = new TruffleDB(config);
+
+const { schema, context } = db;
+
+const server = new ApolloServer({
+  tracing: true,
```
eggplantzzz, line 17: What is `tracing`?

eggplantzzz: Nevermind! https://github.com/apollographql/apollo-server/tree/HEAD/packages/apollo-tracing

packages/db/package.json
```diff
@@ -0,0 +1,110 @@
...
+    "tsconfig-paths": "^3.7.0",
+    "typescript": "^3.6.3"
+  },
+  "dependencies": {
+    "@gnd/graphql-tools": "^4.0.5-fix.0",
+    "@gnd/jsonschema2graphql": "^1.0.15",
+    "@truffle/artifactor": "^4.0.74",
+    "@truffle/workflow-compile": "^2.1.48",
+    "apollo-server": "^2.3.1",
+    "fse": "^4.0.1",
```
eggplantzzz, line 62: Shouldn't we use `fs-extra`?

packages/db/src/artifacts/json/resolvers.ts
```diff
@@ -0,0 +1,89 @@
+import fs from "fs";
```
eggplantzzz, line 1: This should probably be `fs-extra`, no?

packages/db/src/artifacts/test/index.ts
```diff
@@ -0,0 +1,245 @@
+import fs from "fs";
```
eggplantzzz, line 1: Same here about `fs-extra`.

packages/db/src/artifacts/test/index.ts
```diff
@@ -0,0 +1,245 @@
...
+import path from "path";
+import { TruffleDB } from "@truffle/db";
+import { shimBytecode } from "@truffle/workflow-compile/shims";
```
eggplantzzz, line 4: This is going to conflict with the PR just merged changing the compilation interface (psst, [this one](https://github.com/trufflesuite/truffle/pull/3252)).

The shims got moved to `compile-common` and have a different api now. [Here, check it out!](https://github.com/trufflesuite/truffle/tree/develop/packages/compile-common/src/shims)

packages/db/src/helpers/generateId.ts
```diff
@@ -0,0 +1,13 @@
+import { soliditySha3 } from "web3-utils";
+const jsonStableStringify = require("json-stable-stringify");
+
+const removeNullyValues = obj =>
```
eggplantzzz, line 4: `nully`! I like it!

packages/db/src/loaders/commands/compile.ts
```diff
@@ -0,0 +1,144 @@
+import { TruffleDB } from "@truffle/db/db";
+import {
+  WorkflowCompileResult,
```
eggplantzzz, line 3: It looks like some of these types now exist in `compile-common`. See [here](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-common/src/types.ts).

packages/db/src/db.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    return await execute(this.schema, document, null, this.context, variables);
+  }
+
+  async loadCompilations(result: WorkflowCompileResult) {
+    const saga = generateCompileLoad(result, {
+      directory: this.context.workingDirectory
+    });
+
+    let cur = saga.next();
```
eggplantzzz, line 56: What does `cur` mean?

packages/db/src/loaders/commands/compile.ts
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  const compilations = yield* generateCompilationsLoad(loadableCompilations);
+
+  // now time to add contracts
+  //
+  // again going one compilation at a time (for impl. convenience; HACK)
+  // (@cds-amal reminds that "premature optimization is the root of all evil")
+  for (const [compilationIndex, compilation] of compilations.entries()) {
+    const resultCompilation = resultCompilations[compilationIndex];
```
eggplantzzz, line 74: Is this variable name meant to mean "the result of a compilation"? If so, why not just `compilationResult`?

packages/db/src/loaders/resources/compilations/index.ts
```diff
@@ -0,0 +1,80 @@
+import {
+  CompilationData,
+  LoadedSources,
+  IdObject,
+  WorkspaceRequest,
+  WorkspaceResponse
+} from "@truffle/db/loaders/types";
+
+import { AddCompilations } from "./add.graphql";
+export { AddCompilations };
```
eggplantzzz, line 10: Why do we export this here as well as the places we imported it from in this file?

eggplantzzz: Oh I see, this is the main index file.

packages/db/src/db.ts
fainashalts: It's the current spot at which the generator is paused. It's just a way to keep track and check whether the generator function is done or if there are more yields in there to get through. 

Each `saga.next()` call should hit the next `yield` in the generator function, and at each yield point in the generator the generator object is returned. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*

packages/db/src/artifacts/test/index.ts
fainashalts: @eggplantzzz so now I would import `forBytecode` from `compile-common` instead? Is there more that has changed that I should keep an eye out for? I may fix this as a quick follow-on after this branch gets merged but was wondering if there is more to it.

fainashalts: Merged changes to account for new `compile-common` stuff!

packages/db/src/loaders/commands/compile.ts
fainashalts: Merged changes to address this!

eggplantzzz (approved):
merge away!

PR #3298, opened 2020-08-25
Hide `truffle db` command

gnidan:
To address the first concern in #3297


eggplantzzz:
Do we need to expose this from `core`?

Issue #3292, opened 2020-08-23
truffle test failed if the version field in truffle.config is assigned with a custom path


zexxlin:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.
---------------------------

## Issue

`truffle test` failed with version field in truffle.config pointed to custom path, with following error:

> TypeError: Invalid Version: null at new SemVer

The issue seems to lie in following code segments:

[testing/deployed.js](https://github.com/trufflesuite/truffle/blob/v5.1.41/packages/core/lib/testing/deployed.js)
```
// Line 40 - 45
if (version) {
  if (version === "native") version = new Native().load().version();
  const v = semver.coerce(version);
  if (semver.lt(v, "0.5.0")) source = source.replace(/ payable/gm, "");
  source = source.replace(/0\.5\.0/gm, v);
}
```
If **version** is a custom path, `semver.coerce` will parse it as null, then `semver.lt` will throw the above error.

## Steps to Reproduce

1. Specify version field in truffle.config as a custom path, e.g ./node_modules/solc
2. Run `truffle test`

## Expected Behavior

No error is thrown.

## Actual Results

> TypeError: Invalid Version: null
>     at new SemVer (/home/zxlin/.nvm/versions/node/v12.18.2/lib/node_modules/truffle/build/webpack:/packages/core/node_modules/semver/semver.js:314:1)
>     at compare (/home/zxlin/.nvm/versions/node/v12.18.2/lib/node_modules/truffle/build/webpack:/packages/core/node_modules/semver/semver.js:647:1)
>     at Function.lt (/home/zxlin/.nvm/versions/node/v12.18.2/lib/node_modules/truffle/build/webpack:/packages/core/node_modules/semver/semver.js:688:1)
>     at Object.makeSolidityDeployedAddressesLibrary (/home/zxlin/.nvm/versions/node/v12.18.2/lib/node_modules/truffle/build/webpack:/packages/core/lib/testing/deployed.js:43:1)
>     at TestSource.resolve (/home/zxlin/.nvm/versions/node/v12.18.2/lib/node_modules/truffle/build/webpack:/packages/core/lib/testing/testsource.js:68:1)
>     at TestResolver.resolve (/home/zxlin/.nvm/versions/node/v12.18.2/lib/node_modules/truffle/build/webpack:/packages/core/lib/testing/testresolver.js:32:1)
>     at async Promise.all (index 1)

## Environment

[...]

PR #3300, opened 2020-08-25
Bug fix: Improve version checking during testing

haltman-at:
This PR is in response to #3292.  It attempts to make version-checking in `truffle test` better and more consistent.  Right now, `test.js` and `testing/Deployed.js` both attempt to check what Solidity version is being used, but they do so in different ways, each with different pitfalls.  This PR has them do it the same way, in a way that should be more robust than what either was doing before.

Note that I basically relied on copy-paste here.  However it may be worth factoring this somehow?  I figured I'd PR this initial version and we could perhaps figure that out later.

Bascailly, before, `test.js` wanted to know whether the version was at least 0.6.3, while `testing/Deployed.js` wanted to know whether it was prior to 0.5.0.  The former could handle versions and version ranges, but not the native or local cases.  The latter could handle native and local, but might have encountered problems on version ranges.  I've made the two now work basically the same by combining the methods that the two used.

So, now, both will handle native and local by using the appropriate loading strategy to determine what version that corresponds to.  Note I didn't make it work this way in general, despite how that would have cut down on code, because that might involve a network connection for a specified version or version range.  (Also, note that both of these continue to rely on the fact that the default version is 0.5.16...)

As for individual versions and version ranges, those are now handled using appropriate semver functions.  Note that I used `semver.gtr` to make sure that, for a version range, the replacement only happens if the version is *entirely* before 0.5.0.  ~~Also on the individual version check, I wrote `0.4.x || <0.4.0` rather than `<0.5.0` in order to handle prereleases better.~~ (That last bit ended up being scrapped, see comments below.)

Also, while I was at it, I changed the pragmas in `testing` to say `< 0.8.0` instead of `< 0.7.0`.  That's not related to the main point of this PR but I thew it in there.

Note: Not sure if I tested this enough?

### DIFF
--- a/packages/compile-solidity/compilerSupplier/index.js
+++ b/packages/compile-solidity/compilerSupplier/index.js
@@ -4,10 +4,11 @@ const semver = require("semver");
 
 const { Docker, Local, Native, VersionRange } = require("./loadingStrategies");
 
+const defaultSolcVersion = "0.5.16";
+
 class CompilerSupplier {
   constructor({ events, solcConfig }) {
     const { version, docker, compilerRoots, parser } = solcConfig;
-    const defaultSolcVersion = "0.5.16";
     this.events = events;
     this.parser = parser;
     this.version = version ? version : defaultSolcVersion;
@@ -118,6 +119,10 @@ class CompilerSupplier {
           latestRelease: list.latestRelease
         };
       });
+  }
+
+  static getDefaultVersion() {
+    return defaultSolcVersion;
   }
 }
 
--- a/packages/core/lib/testing/Deployed.js
+++ b/packages/core/lib/testing/Deployed.js
@@ -1,9 +1,12 @@
 const web3Utils = require("web3-utils");
 const semver = require("semver");
 const Native = require("@truffle/compile-solidity/compilerSupplier/loadingStrategies/Native");
+const Local = require("@truffle/compile-solidity/compilerSupplier/loadingStrategies/Local");
+const path = require("path");
+const fs = require("fs-extra");
 
 var Deployed = {
-  makeSolidityDeployedAddressesLibrary: function(
+  makeSolidityDeployedAddressesLibrary: function (
     mapping,
     { solc: { version } }
   ) {
@@ -15,7 +18,7 @@ var Deployed = {
       "pragma solidity >= 0.5.0 < 0.7.0; \n\n library DeployedAddresses {" +
       "\n";
 
-    Object.keys(mapping).forEach(function(name) {
+    Object.keys(mapping).forEach(function (name) {
       var address = mapping[name];
 
       var body = "revert();";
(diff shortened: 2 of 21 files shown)


haltman-at:
Ugh, I just realized there's one more place we do version-checking during testing -- in determining which `SafeSend` to use, in `SolidityTest.js`.  OK, I think rule of three demands that I factor this.  Consider this not quite done yet.

haltman-at:
OK, did the necessary factoring.  I ended up discarding the `0.4.x || <0.4.0` subtlety because it was hard to fit in.  Anyway this is all factored into a new `Version.js` now!  So that's nice.  Resulting code is quite a bit cleaner.

Still might warrant more testing though?

haltman-at:
Note: Currently this seems to be failing due to build breakage rather than due to the changes it contains.

haltman-at:
Oops, left out the new file I added, here's a force push to add that.

haltman-at:
OK, *now* it seems to be failing for reasons actually related to what I did...

packages/compile-solidity/compilerSupplier/index.js
```diff
@@ -4,10 +4,11 @@ const semver = require("semver");
 
 const { Docker, Local, Native, VersionRange } = require("./loadingStrategies");
 
+const defaultSolcVersion = "0.5.16";
```
eggplantzzz, line 7: Do we want to make this name all caps to denote that it is a constant?

packages/core/lib/testing/Deployed.js
```diff
@@ -37,18 +36,20 @@ var Deployed = {
...
 
-    if (version) {
-      if (version === "native") version = new Native().load().version();
-      const v = semver.coerce(version);
-      if (semver.lt(v, "0.5.0")) source = source.replace(/ payable/gm, "");
-      source = source.replace(/0\.5\.0/gm, v);
+    version = Version.resolveVersionString(version);
+    if (!Version.versionIsAtLeast(version, "0.5.0")) {
+      //remove "payable"s if we're before 0.5.0
+      source = source.replace(/ payable/gm, "");
```
eggplantzzz, line 42: I know you didn't write this code but is it potentially bad to just replace this? Can you use `payable` as a variable name? Do we need to make sure that it has a space at each end like `/ payable /gm`?

packages/compile-solidity/compilerSupplier/index.js
haltman-at: Hm, I was just leaving it as-is.  I think @gnidan has said we should be moving away from all-caps anyway?

packages/core/lib/testing/Deployed.js
haltman-at: I mean, `address payable varName;` would be a syntax error before 0.5.0, right?  Dunno that this search-and-replace is the best way to handle it, but I don't think we can just leave it alone?

packages/compile-solidity/compilerSupplier/index.js
gnidan: Yeah, I'm leaning towards "name constants like normal" these days

packages/core/lib/testing/Deployed.js
gnidan: Seems like we should replace `address payable` with `address`, no?

haltman-at: Hm, yeah.  I guess I'll change that.  That's an easy change to make without having to figure out the actual proper way to go about things.

packages/core/lib/testing/Version.js
```diff
@@ -0,0 +1,49 @@
+const semver = require("semver");
```
gnidan, line 1: What do you think of moving this file to somewhere inside `compile-solidity/compilerSupplier`?

haltman-at: It's an interesting idea -- what would it be called?  `versionUtils.js`?

I do feel like it's a little out of place there as these functions are a bit... weird.  Right, like, these "versions" are *sometimes* versions and sometimes ranges.  Do you think that's not a problem?

gnidan: We can use the word "constraint" instead of "version"... or I guess "range" can refer to a range of one?

I like "utils" as part of the name!

haltman-at: Well, the thing is that individual versions and version ranges are actually handled slightly separately in order to account for prerelease versions.

It's just confusing because there are three different notions of "version" flying around:

1. Anything that can go in the `version` field (i.e. a version, a range, `native`, or a file path)
2. A version or range
3. A version

IDK -- I guess functions could be renamed to be more explicit?  Instead of `versionUtils` it could be `rangeUtils`?  Does that sound good?

gnidan: Sounds good to me!

haltman-at: OK, guess I'll do that then!

PR #3302, opened 2020-08-25
Add no-op prepare script to packages missing it

gnidan:
Thanks to @seesemichaelj for verifying that `exit 0` should be fine on Windows :)

See #3306 for something of a write-up on why this is necessary.

### DIFF
--- a/packages/compile-solidity/package.json
+++ b/packages/compile-solidity/package.json
@@ -11,6 +11,7 @@
   "version": "4.3.19",
   "main": "index.js",
   "scripts": {
+    "prepare": "exit 0",
     "test": "./scripts/test.sh"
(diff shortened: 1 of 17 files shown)


gnidan:
Well that works... circular dependency, huh?

Looks like [lerna does not and will not ever](https://github.com/lerna/lerna/issues/1198) treat `devDependencies` differently for purposes of determining package dependency DAG.

Will have see about a better way of getting rid of that dependency cycle.

gnidan:
Oh huh this does expose the circular dependency problem on `develop`! Guess we'll do it in the other order :)

gnidan:
Thanks @fainashalts !

PR #3305, opened 2020-08-25
Move contract tests to new, unpublished package

gnidan:
This PR removes a circular dependency including @truffle/contract by moving @truffle/contract's tests to a new, to-remain-unpublished package @truffle/contract-tests.

Current dependency cycle:
> @truffle/compile-common -> @truffle/resolver -> @truffle/contract -> @truffle/compile-solidity -> @truffle/compile-common

<details>
<summary>Bonus: fresh package dependency graph with new @packages/contract-tests</summary>

<p>Generated via <code>npx lerna-dependency-graph | dot -Tpng > deps.svg</code>. See package <a href="https://www.npmjs.com/package/lerna-dependency-graph">lerna-dependency-graph</a></p>
</details>


gnidan:
(This is opened against `truffle-db` branch since that branch is failing for reasons perceived related to this cycle)

gnidan:
@eggplantzzz curious what you think of this

gnidan:
On Zoom earlier, @eggplantzzz seemed unenthused by this but not opposed. We discussed a couple other alternatives, but the approach in this PR seemed to be the least-bad:

- Move the specific tests that depend on compile-solidity into a package like `truffle` or `@truffle/core`.

  This seems not great because these tests are about @truffle/contract specifically, and there's value in making sure that relationship remains clear.

- Do something funky with nested packages in the monorepo (like `packages/contract/test/package.json`).

  There seemed to be consensus that it makes sense to wait on this until we have more examples of how we would want to organize this as a rule, and that this one-off is fine for now.

- Remove compile-solidity dependency from those tests.

  This would require rewriting the tests too much for right now.

I'll let this simmer for a few more hours in case opinions change or new ideas surface :)

eggplantzzz:
I don't like it, but who does? Eventually maybe we could consider putting all tests in a top-level `test` folder? 🤷

gnidan:
Alright, half-hearted blessing is good enough for me!

PR #3306, opened 2020-08-25
Fix failing `yarn bootstrap`

gnidan:
This PR includes changes from #3302 and #3305 because neither alone appears to be sufficient. These changes seek to accommodate lerna's behavior related to package dependency ordering, which appears to behave unpredictably when you have cycles (even `devDependencies` cycles ☹️).

Branch `truffle-db` is failing because `yarn bootstrap` runs the prepare script for `@truffle/db` before it runs the prepare script for @truffle/compile-common. It appears there are two separate lerna behaviors we must accommodate:

- **First problem: `lerna run prepare` skips dependency subtrees sometimes.** Given packages A _depends-on_ B _depends-on_ C, when C defines a `prepare` script but B does not, lerna appears to treat C as not strictly topologically leafier than package A. So sometimes lerna will prepare package A before preparing package C, hence the failure.

  How this corresponds to the `truffle-db` branch failure:

  - A is @truffle/db (has `prepare` script)

  - B is @truffle/compile-solidity (lacks `prepare` script)

  - C is @truffle/compile-common.

  The first set of cherry-picked changes in this PR, from #3302, ensure every package defines a `prepare` script; using `exit 0` as a pass-through for all the packages that don't already define this script. (h/t @seesemichaelj for verifying this behavior is fine on Windows!) When all packages have `prepare` scripts, it appears that lerna does not exhibit this skipped subtree problem.

- Solving the above problem exposes the **second problem: there's a dependency cycle between several packages**. This PR includes changes from #3305, which eliminates that cycle. This cycle was previously thought to be fine because it's only cyclical due to `devDependencies`. Apparently [lerna does not treat `devDependencies` any differently and will never support circular dependencies of any kind](https://github.com/lerna/lerna/issues/1198).

**Plan for this PR**: this should make it into `develop` first. I'm opening it against `truffle-db` because, well, that's the branch that's failing.


gnidan:
Both of those changes separately made it to `develop`. Closing this!

Issue #3268, opened 2020-08-14
Expand debugger `b` help to provide more info on syntax


gnidan:
Right now, there's nothing in the help output that suggests `b` takes an argument.

(Also requested by @tcoulter)

PR #3311, opened 2020-08-25
Expand debugger help message for b command

haltman-at:
Addresses #3268.

I do worry that the help printout is losing its compactness a bit, but...


packages/debug-utils/index.js
```diff
@@ -25,8 +25,8 @@ const commandReference = {
   "+": "add watch expression (`+:<expr>`)",
   "-": "remove watch expression (-:<expr>)",
   "?": "list existing watch expressions and breakpoints",
-  "b": "add breakpoint",
-  "B": "remove breakpoint",
+  "b": "add breakpoint (can specify line in several ways; see docs)",
```
gnidan, line 28: it'd be better if we provided some real hint vs. nothing. Maybe this is too much, I don't know:
```suggestion
  "b": "add breakpoint (`b [[<source-file>:]<line-number>]; see docs for more)",
```

haltman-at: Oh that's nice; yeah, I'll change that.

haltman-at:
OK, not only did I rewrite the `b` help message as suggested, I also rewrote the `p` message similarly!  Neither of these give the complete syntax, but I think that's OK.  Kind of thinking that maybe `p` should include the `+` syntax though?  IDK.

I suppose `;` could also be similarly rewritten but I don't think it needs it.

gnidan (approved):
Looks good!

Issue #3267, opened 2020-08-14
Print highlighted source range in debugger upon revert


gnidan:
Requested by @tcoulter, he wants to see the normal debugger source output when a transaction reverts, since `l` doesn't work after that point.

haltman-at:
So, to be clear, you straight up can't see where execution ended?  Huh.  I hadn't realized that was occurring...

haltman-at:
I think the simplest thing to do here would be to just fix `l` so that it works after execution has ended; I had no idea that it didn't.

haltman-at:
I initially thought this was asking for something different, that on a revert it'd show where the revert happened (based on the stacktrace).  Which is also doable!  But this is easier, certainly.

haltman-at:
Oh I see!  The problem is that, when the actual `REVERT` happens, you're in internal code.  So, OK -- I guess the way I originally understood this is the way to handle this after all!

PR #3313, opened 2020-08-26
Enhancement: Have debugger print location of error on transaction failure

haltman-at:
This PR addresses #3267.  Now, when reaching the end of a failing transaction, the debugger will print a red `Location of error:` message along with the location where the error occurred.  (As in, the innermost point in the stacktrace.)  Moreover, you can hit `l` after the transaction is done to display this message again.  I'm hoping the `Location of error:` message is noticeable enough -- I don't want people to confuse this for the actual current location.  I'd say bold red, but it doesn't seem quite important enough to warrant that?  Maybe I should just go with bold?


haltman-at:
Oh, thought -- maybe I should put this on `s` instead of on `l`... yeah I think I should do that, that'd probably be less confusing...

haltman-at:
OK, I changed it so it prints this out on `s` rather than `l`.  Of course it only does it at the end of a failing transaction, otherwise `s` continues to have the same effect.

I also revised the `s` printout so that if you're not at the end it prints `Call stack:` rather than `Stacktrace:`

I ended up sticking with red rather than bold.

@gnidan let me know what you think, this should be good to review now!

gnidan (approved):
Looks good!

PR #3315, opened 2020-08-26
Internal improvement: Shim relocation

eggplantzzz:
Move shims for converting from legacy format to newer `CompiledContract` format to `compile-common`.


gnidan:
Generally best to avoid `namespace`s, I've found.

I usually follow this pattern:
```
index.ts
shims/
  newToLegacy.ts
  legacyToNew.ts
  index.ts
```

Then inside `shims/index.ts`, do:
```
import * as NewToLegacy from "./newToLegacy";
import * as LegacyToNew from"./legacyToNew";
export { NewToLegacy, LegacyToNew };
```

And similarly, inside `index.ts`, get the shims the same way.

packages/compile-common/src/shims.ts
```diff
@@ -1,4 +1,4 @@
-import { LinkReference, CompiledContract } from "./index";
+import { Bytecode, LinkReference, CompiledContract } from "./index";
```
gnidan, line 1: We should get these types into a separate `types.ts` file, otherwise we'll run the risk of circular imports. (This can probably be on the other PR, though)

eggplantzzz:
Oh! I thought you wanted me to use the namespaces. Coming right up!

gnidan:
Yep, "namespace" but not "`namespace`" :)

Sorry to be unclear! (Or maybe I mislead you and was wrong, don't remember)

PR #3317, opened 2020-08-27
Set context vars

fainashalts:
This PR seeks to allow the re-introduction of the child process for commands from #3255, the `spawn-repl` branch, while fixing a  few issues: 

1. A bug where Artifacts were not available within the `truffle develop` and `truffle console` commands. It appears the context containing the contracts' artifacts was not properly being set for the new repl. 

To test that this bug is fixed: 
a) Run `truffle console` or `truffle develop` in an existing project that has already migrated its contracts. 
b) Enter the name of one of its contracts.
Expected: The artifact for that contract is printed in your terminal. 

a) Run `truffle console` or `truffle develop` in an existing project for which a migration has not been run. 
b) Enter the name of a contract in the project.
Expected: Error thrown
c) Run the `migrate` command
d) Enter the name of a contract in the project
Expected: The artifact is printed in your terminal.

2. This PR also updates the `stdio` handling of `stderr` in the spawn process, to keep warnings that have already been consoled in the parent process from being printed again. Specifically, this is in relation to the error about having two config files: 
```
Warning: Both truffle-config.js and truffle.js were found. Using truffle-config.js.
```

3. This PR also pull this `this.provision()` function call outside of the try/catch it was in, to ensure that if there is an error in that function that we are able to see it. 

4. This PR also fixes the use of yargs for properly getting and passing the network in the `console-child` file. (good catch on this bug @cds-amal!)

5. This work additionally ensures that assignments/scripts created in Javascript inside the child process persist between commands. You should be able to run `const [ownerAccount] = accounts` and access `ownerAccount` throughout your `develop` or `console` session.

Many thanks to @cds-amal for letting me bounce ideas off of him while figuring out this bug!

### DIFF
--- a/packages/core/lib/console-child.js
+++ b/packages/core/lib/console-child.js
@@ -11,8 +11,7 @@ const inputStrings = input[1];
 //and deserialize them
 const detectedConfig = Config.detect({ network: yargs(input[0]).argv.network });
 const customConfig = detectedConfig.networks.develop || {};
-// console.log("custom config? " + JSON.stringify(customConfig));
-// console.log("detected config " + JSON.stringify(detectedConfig));
+
 //need host and port for provider url
 const ganacheOptions = {
   host: customConfig.host || "127.0.0.1",
--- a/packages/core/lib/console-child.js
+++ b/packages/core/lib/console-child.js
@@ -9,9 +9,10 @@ const inputStrings = input[1];
 
 //detect config so we can get the provider and resolver without having to serialize
 //and deserialize them
-const detectedConfig = Config.detect({ network: yargs(input[0]).network });
+const detectedConfig = Config.detect({ network: yargs(input[0]).argv.network });
 const customConfig = detectedConfig.networks.develop || {};
-
+// console.log("custom config? " + JSON.stringify(customConfig));
+// console.log("detected config " + JSON.stringify(detectedConfig));
 //need host and port for provider url
 const ganacheOptions = {
   host: customConfig.host || "127.0.0.1",
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -70,17 +70,20 @@ class Console extends EventEmitter {
   start() {
     try {
       this.interfaceAdapter.getAccounts().then(fetchedAccounts => {
-        this.provision();
+        const abstractions = this.provision();
 
         this.repl = repl.start({
           prompt: "truffle(" + this.options.network + ")> ",
           eval: this.interpret.bind(this)
         });
 
+        this.resetContractsInConsoleContext(abstractions);
+
+        // set these repl context vars after setting contracts to
+        // make sure they don't get overwritten
         this.repl.context.web3 = this.web3;
         this.repl.context.interfaceAdapter = this.interfaceAdapter;
         this.repl.context.accounts = fetchedAccounts;
-        this.repl.context = this.context;
       });
     } catch (error) {
       this.options.logger.log(
@@ -139,6 +142,11 @@ class Console extends EventEmitter {
     });
 
     this.context = this.setContextVars(contextVars);
+
+    // make sure the repl gets the new contracts in its context
+    if (this.repl) {
+      this.repl.context = this.context;
+    }
   }
 
   runSpawn(inputStrings, options, callback) {
@@ -149,6 +157,9 @@ class Console extends EventEmitter {
       childPath = path.join(__dirname, "../lib/console-child.js");
     }
 
+    // stderr is piped here because we don't need to repeatedly see the parent
+    // errors/warnings in child process - specifically the error re: having
+    // multiple config files
     const spawnOptions = { stdio: ["inherit", "inherit", "pipe"] };
 
     const spawnInput = "--network " + options.network + " -- " + inputStrings;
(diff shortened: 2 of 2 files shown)


packages/core/lib/console-child.js
```diff
@@ -9,9 +9,10 @@ const inputStrings = input[1];
 
 //detect config so we can get the provider and resolver without having to serialize
 //and deserialize them
-const detectedConfig = Config.detect({ network: yargs(input[0]).network });
+const detectedConfig = Config.detect({ network: yargs(input[0]).argv.network });
 const customConfig = detectedConfig.networks.develop || {};
-
+// console.log("custom config? " + JSON.stringify(customConfig));
+// console.log("detected config " + JSON.stringify(detectedConfig));
```
cds-amal, line 15: These should come out.

packages/core/lib/console.js
```diff
@@ -130,6 +140,13 @@ class Console extends EventEmitter {
     abstractions.forEach(abstraction => {
       contextVars[abstraction.contract_name] = abstraction;
     });
+
+    this.context = this.setContextVars(contextVars);
+
+    // make sure the repl gets the new contracts in its context
+    if (this.repl) {
+      this.repl.context = this.context;
```
cds-amal, line 148: I suspect this will overwrite any `repl.context` variables that `this.context` doesn't know about such as user defined variables as well as the account globals.

packages/core/lib/console.js
```diff
@@ -42,6 +42,7 @@ class Console extends EventEmitter {
       "build_directory"
     ]);
 
+    this.context = {};
```
cds-amal, line 45: `this.context` and `this.repl.context` seem to have overlapping or the same concern. What do you think?

packages/core/lib/console.js
fainashalts: Oops! That's what I get for trying to get in a quick fix. :) Thanks!

packages/core/lib/console-child.js
fainashalts: Oops! That's what I get for trying to get in a quick fix. :) Thanks!

packages/core/lib/console.js
fainashalts: So, it does. I set the other context variables that are needed after `provision()` is run.

packages/core/lib/console.js
fainashalts: You, know, I think you're right. I had originally left `this.context` because I thought we might need it elsewhere but I believe we don't. Change incoming!

packages/core/lib/console.js
fainashalts: @cds-amal my last commit should address this. Let me know if you have any concerns!

packages/core/lib/console.js
fainashalts: Ok, this concern should be addressed as of my last commit!

packages/core/lib/console.js
cds-amal: Oh I missed that!

packages/core/lib/console.js
```diff
@@ -130,6 +133,14 @@ class Console extends EventEmitter {
     abstractions.forEach(abstraction => {
       contextVars[abstraction.contract_name] = abstraction;
     });
+
+    // make sure the repl gets the new contracts in its context
+    if (this.repl) {
+      this.repl.context = {
+        ...this.repl.context,
+        ...this.setContextVars(contextVars)
```
cds-amal, line 141: `setContextVars` can be replaced with an inline operation

```suggestion
        ...contextVars
```

packages/core/lib/console.js
```diff
@@ -130,6 +133,14 @@ class Console extends EventEmitter {
     abstractions.forEach(abstraction => {
       contextVars[abstraction.contract_name] = abstraction;
     });
+
+    // make sure the repl gets the new contracts in its context
+    if (this.repl) {
```
cds-amal, line 138: This guard can be removed if the repl creation is done before getting accounts in the `Console.start` method.

```
      this.repl = repl.start(...);
      this.interfaceAdapter.getAccounts().then(fetchedAccounts ...)
```

packages/core/lib/console.js
```diff
@@ -118,7 +122,6 @@ class Console extends EventEmitter {
     });
 
     this.resetContractsInConsoleContext(abstractions);
```
cds-amal, line 124: Would you consider renaming this function to `addContractReferencesToReplContext`?

packages/core/lib/console.js
```diff
@@ -58,6 +58,10 @@ class Console extends EventEmitter {
     });
   }
 
+  setContextVars(obj) {
+    return { ...obj };
+  }
+
```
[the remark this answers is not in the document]
fainashalts: Agree. And it actually needs not to use the spread operator it turns out, so I redid that too.

packages/core/lib/console.js
fainashalts: I did this a different way to ensure preservation of any assignments/scripts that a user creates within the console.

packages/core/lib/console.js
fainashalts: I think this is a reset and not an addition. It happens at the end of every command, and if the contracts have changed they get reset to the new ones.  Happy to discuss further but this was the original name if this function in the replManager and I think it is still apt.

packages/core/lib/console.js
fainashalts: Good catch! I reworked this part a bit to incorporate this suggestion.

PR #3319, opened 2020-08-27
Run each console command in its own child process

fainashalts:
Reverts trufflesuite/truffle#3314

We have a fix for this now, so we can revert the reversion and put in the fix, which is in #3317.


packages/core/index.js
```diff
@@ -4,7 +4,6 @@ var pkg = require("./package.json");
 module.exports = {
   build: require("./lib/build"),
   create: require("./lib/commands/create/helpers"),
-  console: require("./lib/repl"),
```
eggplantzzz, line 7: So this would be a breaking change for `@truffle/core`, right?
Maybe you should keep the legacy stuff in and just file it in a `legacy` folder so it won't be?

fainashalts: Hm, I guess it's possible someone was using the ReplManager externally? It was really an internal concern though, and I don't think it ever should have been exposed. Wondering what @gnidan thinks - is this a breaking change and should I do something to keep it from being such?

packages/truffle/webpack.config.js
```diff
@@ -40,26 +40,34 @@ module.exports = {
       "@truffle/core",
       "index.js"
     ),
+    consoleChild: path.join(
```
eggplantzzz, line 43: Just out of curiosity, why do we need to create a bundle for this? It it because Truffle needs to use an external bundle to run the command in a separate process? It seems like this will drastically increase the size of the stuff we need to include with Truffle, huh?

eggplantzzz: I'm understanding that the bundle will be running this other bundle using the user's input. Yeah? This will make the distributed files a lot larger, right? I'm not trying to hold this PR up, I'm just curious.

fainashalts: I'm not sure I follow. The file being bundled is `console-child.js`, which is fairly short. The `cli.bundled.js` file is a tiny bit bigger -- 22.842243 megabytes on `develop` and  22.843334 megabytes on this branch. I don't think that's particularly significant?

More to the point though, how else would you get the correct path here?

eggplantzzz: Yes, but `console-child.js` is just the entry point. So it will double the size of the distributed stuff for Truffle. Essentially this will be like a copy of Truffle's main `cli.bundled.js`...which might be ok. I just wanted to point out that this seems to be a significant increase in size. Does that make sense?

eggplantzzz: Well I checked and it is not quite double, it increases the stuff in the build folder from ~50MB to like ~72MB. Something like that.

fainashalts: Hm, yeah I think that makes sense but I don't think there's another way. The whole point is that we can run Truffle commands through the child process. So we need all of that bundled this way. If you have a suggestion for how to reduce the distributed code size I'm happy to discuss further!

eggplantzzz: No it's probably fine. I just wanted to bring it up :)

gnidan: Yeah, let's avoid having that bundle in two places :)

eggplantzzz:
So we decided that the best fix here would be to define an entry point at `core/lib/commands/index.js`. That way we don't have to bundle `cli.js` and `consoleChild.js`. Those could be just copied to the `dist` folder and each could require the bundle. This is just an estimation of what would need to be done. There may be a little bit off messiness surrounding dealing with `cli.js` and `consoleChild.js` as I'm not sure if there is anything (processing-wise) that needs to be done with these files.

packages/core/lib/console.js
```diff
@@ -144,15 +132,49 @@ class Console extends EventEmitter {
...
     });
 
-    this.repl.setContextVars(contextVars);
+    // make sure the repl gets the new contracts in its context
+    Object.keys(contextVars || {}).forEach(key => {
+      this.repl.context[key] = contextVars[key];
+    });
+  }
+
+  runSpawn(inputStrings, options, callback) {
```
eggplantzzz, line 141: We should rewrite this method to not take a callback. This does not hold up this PR from being merged, I just wanted to note that here.

packages/core/lib/debug/interpreter.js
```diff
@@ -321,7 +310,7 @@ class DebugInterpreter {
 
     //quit if that's what we were given
     if (cmd === "q") {
-      return await util.promisify(this.repl.stop.bind(this.repl))();
+      process.exit();
```
eggplantzzz, line 313: We should also eventually set the exit code here, no? Again, I don't think this should hold this PR up.

eggplantzzz (approved):
I like it!

Issue #3082, opened 2020-05-30
Multi-line import statements in Solidity don't work with contracts in npm modules.


petejkim:
## Issue

Multi-line import statements in Solidity don't work with contracts in npm modules.

## Steps to Reproduce

1. Create a new Truffle project.
2. Install OpenZeppelin: `npm install --save-dev @openzeppelin/contracts` or `yarn add --dev @openzeppelin/contracts`
3. Set solc version to 0.6.8 (to be able to use latest OpenZeppelin)
4. Add the following contract code `Foo.sol`:
```solidity
pragma solidity ^0.6.0;

import {
  ReentrancyGuard
} from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";

contract Foo is ReentrancyGuard {
}
```
5. `truffle compile`.

## Expected Behavior

Should work fine.

## Actual Results

```
contracts/Foo.sol:3:1: ParserError: Source "@openzeppelin/contracts/utils/ReentrancyGuard.sol" not found: File import callback not supported
import {
^ (Relevant source part starts here and spans across multiple lines).
```

If the import statement is in a single line, then it works fine, but it should work regardless of whether it's multi-line or single-line. Importing a local file also works fine regardless of linebreaks.
```solidity
import { ReentrancyGuard } from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";
```

Popular code formatters like `prettier` would split the statement because it's too long for the default line width of 80.

## Environment

* Operating System: 
`Linux ubuntu 5.4.0-33-generic #37-Ubuntu SMP Thu May 21 12:53:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
* Ethereum client: N/A
* Truffle version (`truffle version`): v.5.1.27
* node version (`node --version`): v12.17.0
* npm version (`npm --version`): v6.14.4

haltman-at:
Thanks for the report!  We'll take a look.  Or, if anyone else wants to take a look at this, [here's](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-solidity/parser.js) the most relevant file.

PR #3324, opened 2020-08-28
Fix for multiline import parsing

tomlinton:
Fix for https://github.com/trufflesuite/truffle/issues/3082.

Not very pretty. Both `solcjs` and `solc` only return the first line of the statement in `formattedMessage` so its not possible to grab the import. This detects multiline imports, parses the import name out of `message` and then parses the full path out of the import from `body`.

Single line import handling isn't changed.

Example of a formattedMessage for multiline import using `solcjs`:

```
  { component: 'general',
  formattedMessage:
   'ParsedContract.sol:3:1: ParserError: Source "@openzeppelin/contracts/utils/ReentrancyGuard.sol" not found: File import callback not supported\nimport { ReentrancyGuard } from "@openzeppelin/contracts/utils/ReentrancyGuard.sol";\n^----------------------------------------------------------------------------------^\n',
  message:
   'Source "@openzeppelin/contracts/utils/ReentrancyGuard.sol" not found: File import callback not supported',
  severity: 'error',
  sourceLocation: { end: 109, file: 'ParsedContract.sol', start: 25 },
  type: 'ParserError' }
```

### DIFF
--- a/packages/compile-solidity/parser.js
+++ b/packages/compile-solidity/parser.js
@@ -49,13 +49,25 @@ module.exports = {
     // Filter out our forced import, then get the import paths of the rest.
     const imports = errors
       .filter(({ message }) => !message.includes(failingImportFileName))
-      .map(({ formattedMessage }) => {
-        const matches = formattedMessage.match(
-          /import[^'"]?.*("|')([^'"]+)("|')/
-        );
-
-        // Return the item between the quotes.
-        if (matches) return matches[2];
+      .map(({ formattedMessage, message }) => {
+        // Multiline import check which works for solcjs and solc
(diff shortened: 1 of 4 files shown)


eggplantzzz:
Thanks for the PR @tomlinton! Seems super helpful, although it looks like there is a test failing. I wonder why `truffle init` would fail.

tomlinton:
I get a timeout for that test locally, it is set to 20000ms and it takes slightly longer.

eggplantzzz:
It looks like there is some problem with `truffle install` now. Pretty certain it is unrelated to your changes as I believe that test is also failing on `develop`.

eggplantzzz:
Sweet, it was probably ipfs just being flaky.

PR #3328, opened 2020-08-31
Remove subcommands architecture and mentions for merging later

fainashalts:
This PR removes the subcommands architecture and related mentions, as we want the `truffle-db` branch to avoid affecting other packages when it is initially merged into develop.


cds-amal (commented):
Noting that this behaves differently than stable for `test` within `truffle develop console`. Both stable and starts the ganache provider on the same port.  Stable executes tests and this branch cannot find the provider.

I think this might be related the yargs argument fix.

fainashalts:
> Noting that this behaves differently than stable for `test` within `truffle develop console`. Both stable and starts the ganache provider on the same port. Stable executes tests and this branch cannot find the provider.
> 
> I think this might be related the yargs argument fix.
> 
> Apart from that issue it looks good.

@cds-amal did you mean this comment for this PR? The `yargs` fix is in the spawn-repl work #3319.

cds-amal (approved):
LGTM @fainashalts !

gnidan (approved):

PR #3331, opened 2020-09-01
Enhancement: Allow debugger to step into generated sources

haltman-at:
This PR allows the debugger to make use of generated sources output by the Solidity compiler and to step into them.  (Including variable inspection and source highlighting!)  Note that this is not something the Solidity compiler actually outputs at present, but [it's something they're working on](https://github.com/ethereum/solidity/pull/9053/) and it seems like it's going to happen.

In order to do this, this PR has to touch more than just the debugger -- in fact, this PR adds new fields to the artifact to hold the generated sources.  There's `generatedSources`, which holds the generated sources for the constructor bytecode, and `deployedGeneratedSources`, which holds the generated sources for the deployed bytecode.  I guess @fainashalts and @gnidan will need to figure out how to model this in Truffle DB.

So, the first question is, how do these sources make it into the debugger?

Well, as mentioned, we get them from the compiler.  They get stored in the artifact.  Then `shimArtifacts` adds them to the sources list and marks them as `internal` (this is a new optional boolean field on the source object).  I did a bit of factoring of `shimArtifacts` to make this read cleanly.

What changed in the debugger?  Well, not that much, really.  Basically just three things:

1. Firstly, now, when stepping, if your starting location was not in an internal source, the debugger will not stop in an internal source (unless it hits the end of the trace).  If you want to enter an internal source, you'll need to use `advance()` (i.e., `;`).  Once in one, however, you can advance as normal; using `stepNext()` (`n`) will not immediately take you back to user source.
2.  Secondly, the `solidity.current.nextMapped` selector now skips over internal source steps in order to preserve its existing functionality.  This means it's now a bit misnamed.  Perhaps I should rename it?
3.  Thirdly, I fixed a bug where, when searching up the AST for a node of a given type, the debugger would go into an infinite loop if it got to the top of an all-Yul AST without finding what it was looking for.

Those are the changes to the debugger core.  In addition, there were some changes to `debug-utils` and `core`:

1. The `getTransactionSourcesBeforeStarting` function now explicitly checks for internal sources and does not include them in the output.  Now, because it uses `stepNext` -- actually, I changed it to `stepInto`, that's another change -- this shouldn't actually happen during most of the transaction, but it could happen at the end, so I checked for it and excluded it.
2. The `colorize` function in `debug-utils` now accepts a second argument `yul` indicating that the source should be colorized as Yul rather than Solidity, and `core/lib/debug/printer.js` makes use of this; note it just assumes that internal sources are Yul and user sources are Solidity, it doesn't try to do anything more advanced than that.

The problem though is, how do we colorize Yul sources?  Since we don't have a separate HLJS language for Yul, and I didn't want to take the time to make one right now, we use a hack.  We surround the source in an `assembly` block, colorize, and then strip off the parts we added.  Like I said, it's a hack, but (for now at least) it works.

In addition to the main changes in this PR, there are some others that were needed in order to support testing this PR.  The `sandbox` function in `@truffle/box` has been changed so as to properly handle box names, although you can still pass in `"default"` and get the old behavior there.  Also, the debugger tests have been upgraded to Solidity 0.7.0.  Note that this required removing some tests, as they tested things no longer legal in Solidity 0.7.0.  Specifically:

1. In the tests of globally-available variables, the tests of `now` have been removed.
2. The test of memory structs that include mappings has been removed .

Note this PR does not add any tests of its main functionality, as, again, this is based on Solidity functionality that isn't even merged yet, let alone released.


haltman-at:
OK, some further changes on this one to get tests to pass.  The problem is that my changes to `sandbox` broke some things.

Most of the uses of `sandbox` seemed to be fine, however there were two that seemed troublesome.

In one case, the sandbox used a box named `"bare"`; this was changed to `"bare-box"`.

In another case, the sandbox used a box specification `"default#typescript"`, as in, the `typescript` branch on the `default` box.  To fix this one I just changed `box.ts` further to have `sandbox` do a more complicated replacement of `default` rather than just turning it into `undefined like I had previously.

Also while I was at it I removed the `debug-utils` devDependency from `debugger` because it's no longer used there.

haltman-at:
So it turns out that right now generated sources *all* get name` #utility.yul`.  That means we can't compare them by name to determine whether two of them are the same or not, and have to compare by content instead...

haltman-at:
Note: I've opened #3336 and pulled the tests/sandbox parts into a separate PR.  This is now a PR against that.

packages/debugger/lib/solidity/selectors/index.js
```diff
@@ -330,11 +328,25 @@ let solidity = createSelectorTree({
      * HACK: this assumes we're not about to change context! don't use this if
      * we are!
      * ALSO, this may return undefined, so be prepared for that
+     * NOTE: this selector is now somewhat misnamed, as mapped instructions are
+     * now also skipped over if they point to an internal source
```
gnidan, line 332: Could we make a new, better named selector for this, and switch the code to use that? We can leave this one here as a pass-through for backwards compatibility.

packages/debugger/lib/controller/sagas/index.js
```diff
@@ -68,11 +68,16 @@ function* advance(action) {
  * Note: It might take multiple instructions to express the same section of code.
  * "Stepping", then, is stepping to the next logical item, not stepping to the next
  * instruction. See advance() if you'd like to advance by one instruction.
+ *
+ * Note that if you are not in an internal source, this function will not stop in one
+ * (unless it hits the end of the trace); you will need to use advance() to get into
+ * one.  However, if you are already in an internal source, this function will not
+ * automatically step all the way out of it.
```
gnidan, line 75: I like this behavior

gnidan (commented):
Just one minor comment, but generally this looks fine, code-wise.

gnidan:
I think we should add some way of stepping into generated sources that doesn't require `advance()`ing into it. That way is arduous, and it's hard to find the right place to start advancing.

Maybe we can have make a repl mode that turns this on or off, and have a command that turns this on/off? Not sure if this a good solution, but just a thought.

packages/debugger/lib/solidity/selectors/index.js
haltman-at: I could also just rename the selector, it's pretty internal.  For some reason I didn't think of that while writing this initially.  But yes if we really want compatibility we could do that.

haltman-at:
> I think we should add some way of stepping into generated sources that doesn't require `advance()`ing into it. That way is arduous, and it's hard to find the right place to start advancing.
> 
> Maybe we can have make a repl mode that turns this on or off, and have a command that turns this on/off? Not sure if this a good solution, but just a thought.

Hm.  Could that be saved for a later PR maybe?  I think it might be good to get this in sooner and figure out how to do that later.

packages/debugger/lib/solidity/selectors/index.js
haltman-at: OK, renamed this to `nextUserStep` instead.

haltman-at:
Note: I changed `continueUntilBreakpoint` to use `advance()` rather than `stepNext()` to ensure that it hits breakpoints placed in internal functions.  This may slow things down slightly?

gnidan (approved):
Looks good!

PR #3332, opened 2020-09-02
Environment.configure() the `truffle run` config

gnidan:
So that the config object passed to plugins has access to the resolver and artifactor


gnidan:
This has been failing because the test covers a case which we probably don't want to lose:

Right now, you can use `truffle run` without access to a network. This is useful, e.g., for truffle-plugin-verify, which doesn't talk to the network.

But clearly, some plugins would benefit from the full `Environment.detect()` experience. How do we handle this?

PR #3333, opened 2020-09-02
Expose debugger session from CLIDebugger

gnidan:
So that CLIDebugger can expose all the nice things it does without always starting a Node REPL

### DIFF
--- a/packages/core/lib/debug/cli.js
+++ b/packages/core/lib/debug/cli.js
@@ -23,6 +23,15 @@ class CLIDebugger {
   async run() {
     this.config.logger.log("Starting Truffle Debugger...");
 
+    const session = await this.connect();
+
+    // initialize prompt/breakpoints/ui logic
+    const interpreter = await this.buildInterpreter(session);
+
+    return interpreter;
+  }
+
+  async connect() {
     // get compilations (either by shimming compiled artifacts,
     // or by doing a recompile)
     const compilations = this.compilations || (await this.getCompilations());
@@ -30,10 +39,7 @@ class CLIDebugger {
     // invoke @truffle/debugger
     const session = await this.startDebugger(compilations);
 
-    // initialize prompt/breakpoints/ui logic
-    const interpreter = await this.buildInterpreter(session);
-
-    return interpreter;
+    return session;
   }
 
   async fetchExternalSources(bugger) {


haltman-at:
I'm confused; is the idea that you could use this as a sort of library thing without side effects like printing to the console?  Because right now your `connect` function is going to print a bunch of stuff to console as it starts up.

gnidan:
@haltman-at it's specifically to avoid the repl part - I noticed the console logging, but `config.with({ quiet: true })` gets around that.

haltman-at:
OK, but, you didn't set `quiet: true`?

(Also, huh, that works in general?  I had no idea.)

Also, that won't suppress the various `ora` spinners, which bypasses the usual config/logging system (see: #2288).

gnidan:
> OK, but, you didn't set `quiet: true`?

Nope, I'm just doing that where I'm using that method.

haltman-at:
Hm, I see, like we do with the call to `compile`.  I guess that makes sense.  Are your uses OK with the `ora` uses here though...?

gnidan:
Yep because those go to stderr!

haltman-at:
Huh, interesting, I didn't realize that.  But I'm surprised that's OK.  What are you using this for anyway...?

gnidan:
> Huh, interesting, I didn't realize that. But I'm surprised that's OK. What are you using this for anyway...?

transfer

Automated diagrams!!

haltman-at:
So why is it OK that stuff gets printed on stderr?  I'm still unclear on this.

gnidan:
> So why is it OK that stuff gets printed on stderr? I'm still unclear on this.

Because it's called "CLIDebugger", it lives on the CLI :) 

Why would it not be okay?

haltman-at:
Because you're not using it that way!

It's OK for the CLI debugger to print to stderr, not because it's *called* `CLIDebugger`, but because it *is* a CLI debugger.  However, your intent is to separate out the CLI parts so that you can use it *not* as a CLI debugger, but rather as a Javascript library!  So why is it still OK for it to still print to stderr?

gnidan:
I guess the answer is "it's okay because the effort to make it not print to stderr exceeds the benefit of purity in this regard"

haltman-at:
Eh, OK, I guess if you're OK with it for now then so am I.  Q: Is there a way to make this purer?  Does the events system help here?  I never really learned how that works.

haltman-at (approved):

gnidan:
The events system may help. We should look to address this problem systematically (not just for CLIDebugger)

gnidan:
Thanks for the consideration!

PR #3334, opened 2020-09-02
Enhancement: Add missing `--compile-none` flag to docs for migrate

eggplantzzz:


fainashalts (approved):
This LGTM, though I feel like I saw some discussion about what happens if you did `--compile-all` and `--compile-none` at the same time. Did we decide to ignore that since doing this would likely be an accident? I don't think this would be the place to warn about it anyway, but wanted to bring it up!

eggplantzzz:
Yeah, we can discuss that [here](https://github.com/trufflesuite/truffle/pull/3335) on that PR. This feature is just missing from the docs where I added them here.

PR #3335, opened 2020-09-02
Feature: Implement a `compileNone` option for truffle test

eggplantzzz:
This adds a `--compile-none` option to `truffle test`. Originally started by @robertmagier, a different approach to an implementation came up. So here is a mix of contribution from @robertmagier and myself.

Co-authored-by: Robert Magier <robertmagier@users.noreply.github.com>


eggplantzzz:
So far, there is no warning when a user inputs both `--compile-none` AND `--compile-all`. It was proposed that we add some sort of warning and not throw an error as that is not how this works while using `truffle migrate`. The current behavior is that `--compile-none` overrides the `--compile-all` flag.

packages/core/lib/test.js
```diff
@@ -213,8 +213,14 @@ const Test = {
     const updated =
       (await Profiler.updated(config.with({ resolver: testResolver }))) || [];
 
+    let compiler;
+    if (config.compileNone || config["--compile-none"]) {
+      compiler = "none";
+    }
```
gnidan, line 219: This will clobber if someone passes `--compiler` directly (not sure why they'd want that, but still...)

What if you did this instead:
```suggestion
    const compiler = (config.compileNone || config["--compile-none"])
      ? "none"
      : config.compiler;
```

eggplantzzz: Yeah, that is nice.

gnidan (approved):

PR #3337, opened 2020-09-03
Remove now identifier from debugger on Solidity 0.7.x

haltman-at:
Solidity 0.7.0 removed the global `now` identifier, so this PR removes it from the debugger on 0.7.x.

Also, swapped cases of `{...variables, ...builtins}` to `{...builtins, ...variables}` as I figure it's a little safer.

Didn't add any tests; tested this manually.

### DIFF
--- a/packages/debugger/lib/data/selectors/index.js
+++ b/packages/debugger/lib/data/selectors/index.js
@@ -19,6 +19,16 @@ import * as Codec from "@truffle/codec";
  */
 const identity = x => x;
 
+function solidityVersionHasNoNow(compiler) {
+  return (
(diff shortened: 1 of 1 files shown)


packages/debugger/lib/data/selectors/index.js
```diff
@@ -19,6 +19,16 @@ import * as Codec from "@truffle/codec";
  */
 const identity = x => x;
 
+function solidityVersionHasNoNow(compiler) {
```
gnidan, line 22: solidity version has only _then_! and _later_!

gnidan (approved):

PR #3338, opened 2020-09-03
Enhancement: Debugger/decoder support for free functions

haltman-at:
The main point of this PR is to support free functions in the debugger and decoder, but it ended up doing a few other things as well.  One thing this PR does *not* do, by the way, is making sure that the globally-available variable `this` isn't available in free functions... because it turned out our existing code already did that!  Hooray!  So, really, the main thing ended up just being supporting decoding of function pointers to free functions.

So, first off, this meant a change to the decoder output format.  The `definedIn` field on a `FunctionInternalValueInfoKnown` is now allowed to be `null` to indicate a free function.  Also, a new field has been added to this interface: `id`, an internal ID used for internal functions just in case two free functions somehow have the same name.  (It is constructed in the way you would expect, from compilation ID and AST ID.)

But also this meant:
1. Having `getFunctionsByProgramCounter` in `solidity-utils` (sorry for all the prettier changes -- all the real changes are to that one function) check whether the function's contract node is, in fact, a valid contract node, and make it `null` otherwise
2. Having `functionTableEntryToType` recognize when it's been passed a free function and return `null`
3. Having `decodeInternalFunction` tag all valid internal functions with an ID.

In addition, I updated `nativize` and `ResultInspector` to account for the case of free functions (sorry for all the prettier changes to this file; but you can find the real changes here by looking for `if (coercedResult.value.definedIn)`.)

Still, nothing too complicated.  I also added a test of decoding these.

...but also, some other changes snuck their way into this PR...

First, I noticed that `functionTableEntryToType` was setting the `payable` field based on a field (`contractPayable`) that was never actually set in `getFunctionsByProgramCounter`!  So, uh, I changed the latter to actually set that field.

Secondly, I noticed that some uninitialized internal function pointers weren't decoding properly.  It seems the problem is that, at some point, they changed it so that the designated invalid function has a source index of -1.  So, I made another change to  `getFunctionsByProgramCounter` to account for this case.  Instructions with source index -1 are no longer filtered out; instead they're checked to see whether they might be the designated invalid function.

The fact that this had silently broken bothered me, so I added a test of this, too.

And, that's the entirety of this PR!  Not actually that complicated.

### DIFF
--- a/packages/codec/lib/basic/decode/index.ts
+++ b/packages/codec/lib/basic/decode/index.ts
@@ -641,7 +641,8 @@ export function decodeInternalFunction(
   }
   let name = functionEntry.name;
   let mutability = functionEntry.mutability;
-  let definedIn = Evm.Import.functionTableEntryToType(functionEntry);
+  let definedIn = Evm.Import.functionTableEntryToType(functionEntry); //may be null
+  let id = Evm.Import.makeInternalFunctionId(functionEntry);
   return {
     type: dataType,
     kind: "value" as const,
@@ -651,6 +652,7 @@ export function decodeInternalFunction(
       deployedProgramCounter: deployedPc,
       constructorProgramCounter: constructorPc,
       name,
+      id,
       definedIn,
       mutability
     }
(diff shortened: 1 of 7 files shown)


haltman-at:
Note: There are no change to stack tracing because I had already set up stack tracing to handle free functions in anticipation.

gnidan (approved):
Let's get this in the release :)

PR #3339, opened 2020-09-04
Add missing typeRoots

gnidan:
For @types/node and @types/web3, since our setup is meant for everything to be listed explicitly in the tsconfig

(This caused trouble for me when I ran `yarn test`, for a reason I didn't look into [it worked before, also for a reason I didn't look into])


eggplantzzz (approved):
AOK, I like it!

gnidan:
Thanks!

PR #3340, opened 2020-09-04
Use more explicit code style for new HDWalletProvider constructor interface

gnidan:
In response to https://github.com/richardpringle/truffle/pull/4, this PR goes a bit further to refine the in-progress work started by @richardpringle, @eggplantzzz, and myself to support mnemonic passphrases (which has turned into that, plus a constructor interface redesign + legacy support effort).

This PR offers to suggest alternatives for two aspects of the upstream PR:
- Use more type explicitness to distinguish between the currently-supported options for `new HDWallet()`'s first argument (mnemonic phrase, private key, or list of private keys)
- Don't add support for `{ privateKey }` as a way of using the new interface. We don't need it.

I also noticed some missing `typeRoots`, so there's a commit for that. Opening this as draft, at least because the `typeRoots` change should target `develop` (and is [here](https://github.com/trufflesuite/truffle/pull/3339/files)). I plan to rebase this out of the commit log here once that PR is in.

In addition, this changes a variable name that I named incorrectly in the first place.

### DIFF
--- a/packages/hdwallet-provider/src/constructor/Constructor.ts
+++ b/packages/hdwallet-provider/src/constructor/Constructor.ts
@@ -16,8 +16,7 @@ export interface MnemonicSigningAuthority {
 }
 
 export interface PrivateKeysSigningAuthority {
-  privateKeys?: PrivateKey[];
-  privateKey?: PrivateKey;
+  privateKeys: PrivateKey[];
 }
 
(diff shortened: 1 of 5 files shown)


eggplantzzz (approved):

gnidan:
Thanks!

PR #3343, opened 2020-09-04
Internal improvement: Tear out safe-eval in favor of js-interpreter

haltman-at:
This PR removes the use of `safe-eval` in the debugger CLI to evaluate expressions in response to the recent security alert.  It replaces it with `js-interpreter`, which should actually be safe.

Note that because `safe-eval` is gone, so is the `cleanConstructors` function in `debug-utils`, which existed only to work around a problem with `safe-eval`.  This also means that `debug-utils` has no more tests, since that was the only function being tested, so I just deleted the test directory and test script.

I was originally going to abstract the process of expression evaluation behind a function in `debug-utils`, in which case that could have been tested, but ultimately I didn't do that because then it would be `debug-utils` that optionally depends on `js-interpreter`, and `contract` still depends on `debug-utils`, so I didn't want to potentially open a can of worms there.  So instead that logic all just lives in `core/lib/debug/printer.js`, not abstracted.

As per discussion, `js-interpreter` is an optional dependency of both `core` and `truffle`.  We wrap the `require` in a try block and print an error message if you try to use it when it doesn't exist.  It was also marked as an external in `truffle/webpack.config.js`, which also received a heavy dose of `prettier` changes, but that's the only real change.

So how does it work?

Well, if we're in the expression case, check if we succesfully imported `Interpreter`; if not we bail out.  Then we run nativize on all the variables (same as before).  We then do the various hackery to handle the special `this` variable; this part is unchanged, I didn't touch that.  However note that all the stuff for handling selectors is gone (for now -- see below).

Then, instead of running `safeEval(expr, context)`, we do the following:

1. Set up an interpreter.  We pass the interpreter code consisting of `expr` (duh), and a startup function which goes through each variable in `context` and sets it on the interpreter's global object.  Except it doesn't set `variable = value`, but rather `variable = interpreter.nativeToPseudo(value)`, which is how one converts native objects to interpreter objects.  Unfortunately this means circular objects don't work.
2. Run the interpreter until it terminates; extract the return value; and then pass that back through `interpreter.pseudoToNative` to convert it back to a native object.

(This obviously is the part I wanted to abstract out into a function but didn't because it would have lived in `debug-utils`.

Note that the switch to `js-interpreter` means that two things will no longer work:

1. Circular objects can't be used in expressions (as opposed to printed on their own).  That kind of sucks, as circular objects can be a bit hard to grasp without expressions; but I don't think people really used these much anyway.  These have to be excluded because `js-interpreter`'s `nativeToPseudo` function for importing values doesn't work on circular objects.
2. The `:!<selector>` syntax is gone.  It might be possible to recreate this later; `js-interpreter` does seem to provide a mechanism for calling external functions with `createNativeFunction`, but I don't fully understand this yet and I wanted to get this done quickly so I left it out for the initial version.  (Of course, the unusability of circular objects means that this might be of limited use even if I get it working for non-circular objects.)

Anyway, that's that...


haltman-at:
OK, I was able to figure out the `createNativeFunction` functionality, so support for selectors are back in!  (Although, selectors with circular values won't work, as before.)

Also, I stuck a try/catch block around the creation of the interpreter itself (not just the running of it), as syntax errors would show up in creation, not running.  Exceptions there are handled the same was as exceptions while running it.

haltman-at:
OK reorganized this a bit so that we only need the one try/catch block.

gnidan:
Calling this a draft pending further discussion

gnidan (approved):

PR #3350, opened 2020-09-09
Bug fix: Fix decoding memory structs containing arrays of mappings

haltman-at:
In Solidity prior to 0.7.0, it was possible to put struct types containing mappings in memory.  In the memory representation, the mapping would simply be skipped over.

However, what I did not account for is that it's possible to do the same thing with arrays of mappings!  And multidimensional arrays of mappings!  This PR fixes this oversight, which caused such structs in memory to decode incorrectly.

Now, when the allocator/decoder check for skipped types, instead of just checking for typeclass `"mapping"`, they use the new `isSkippedInMemoryStructs` function.

This PR doesn't add any tests, as all this stuff is illegal in 0.7.x and that's what the debugger tests are on.  But I tested it manually.

### DIFF
--- a/packages/codec/lib/memory/allocate/index.ts
+++ b/packages/codec/lib/memory/allocate/index.ts
@@ -23,13 +23,25 @@ export function getMemoryAllocations(
   return allocations;
 }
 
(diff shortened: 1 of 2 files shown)


gnidan (approved):
Nice!

PR #3351, opened 2020-09-09
Bundle commands separately

eggplantzzz:
Split code shared by `cli.js` and `console-child.js` out into its own bundle

### DIFF
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -31,7 +31,12 @@ if (!semver.satisfies(process.version, ">=" + minimumNodeVersion)) {
 
 const Command = require("./lib/command");
 
-const command = new Command(require("./commands.bundled.js"));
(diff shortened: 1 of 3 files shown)


fainashalts:
@eggplantzzz thanks for putting this PR together! I'm on board with this approach, but `command.bundled.js` won't be available to unbundled versions of Truffle. I think we can probably just check whether a user is using the bundled version there.

eggplantzzz:
Note: Users will never not have the bundled version. The raw JS will only be used by devs.

fainashalts:
Yes, true about users never having the unbundled versions, though I'd venture that its important for Truffle devs and community contributors to be able to use the unbundled version when needed. This fix seems to address that just fine! This LGTM, pending tests passing.

fainashalts (approved):

PR #3354, opened 2020-09-09
Enhancement: Group variables by section in debugger

haltman-at:
@kevinbluer has requested that the debugger give information on which variables are builtins vs contract variables vs local variables for use in the teams debugger.  So, this PR adds a new selector to do this; the selector is `data.current.identifiers.sections`.

This selector returns an object with three fields: `builtin`, `contract`, and `local`, each of which is an array containing all in-scope variable names that go in the given section.  It should be pretty straightforward to use.

I basically did this a quick-and-dirty way by piggybacking off of `data.current.identifiers.refs`, identifying what section a given variable goes in by where it's stored.  So, variables with a location of `"special"` are builtins; a location of "`stack`" means a local; and a location of `"storage"`, `"code"`, `"nowhere"`, `"memory"`, or `"definition"` are contract variables.  (Other locations shouldn't occur.)

(In more detail: `"storage"` is most state variables; `"code"` is deployed immutables; `"nowhere"` is deployed immutables that are never actually read; `"memory"` is immutables in the constructor; and `"definition"` is constants.)

Maybe not the most robust way, but I figured, no need to do extra work here.

Hm, thought, I could also alter the CLI to use this as well...

### DIFF
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -478,33 +478,36 @@ class DebugPrinter {
   }
 
   async printVariables() {
-    let variables = await this.session.variables();
+    const values = await this.session.variables();
+    const sections = this.session.view(data.current.identifiers.sections);
 
-    debug("variables %o", variables);
-
-    const variableKeys = Object.keys(variables);
-
-    // Get the length of the longest name.
-    const longestNameLength = variableKeys.reduce((longest, name) => {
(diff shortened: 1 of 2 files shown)


haltman-at:
Yeah, why the hell not.  I went back and added another commit so now the CLI debugger prints variables this way too. :)

gnidan (approved):
Heckin' swell

PR #3364, opened 2020-09-15
Pull name logic and projects loading into separate commands

fainashalts:
This PR pulls out adding name records, putting that functionality behind a boolean so that we can ensure that when Truffle DB is integrated with the rest of Truffle that we do not update names when `test` is being run. In the course of this work, I also pulled out the logic for adding projects, so that it is now done before we load compilations and everything else.

### DIFF
--- a/packages/db/src/db.ts
+++ b/packages/db/src/db.ts
@@ -37,6 +37,10 @@ interface ITruffleDB {
   query: (query: DocumentNode | string, variables: any) => Promise<any>;
 }
 
+type LoaderOptions = {
+  names: boolean;
+};
+
 export class TruffleDB {
   schema: GraphQLSchema;
   context: IContext;
@@ -90,7 +94,7 @@ export class TruffleDB {
   async loadCompilations(
     project: DataModel.IProject,
     result: WorkflowCompileResult,
-    names: boolean
+    options: LoaderOptions
   ) {
     const saga = generateCompileLoad(result);
 
@@ -104,7 +108,7 @@ export class TruffleDB {
       cur = saga.next(response);
     }
 
-    if (names === true) {
+    if (options && options.names === true) {
       await this.loadNames(project, cur.value.contractsByCompilation);
     }
 
--- a/packages/db/src/db.ts
+++ b/packages/db/src/db.ts
@@ -93,10 +93,11 @@ export class TruffleDB {
   }
 
   async loadCompilations(
-    project: DataModel.IProject,
     result: WorkflowCompileResult,
     options: LoaderOptions
   ) {
+    const project = await this.loadProject();
+
     const saga = generateCompileLoad(result);
 
     let cur = saga.next();
--- a/packages/db/src/db.ts
+++ b/packages/db/src/db.ts
@@ -84,16 +84,11 @@ export class TruffleDB {
     project: DataModel.IProject,
     contractsByCompilation: Array<DataModel.IContract[]>
   ) {
-    const namesLoader = generateNamesLoad(
+    return await this.runLoader(
+      generateNamesLoad,
       toIdObject(project),
       contractsByCompilation
     );
-    let curNames = namesLoader.next();
-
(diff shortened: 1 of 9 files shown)


packages/db/src/db.ts
```diff
@@ -45,26 +53,61 @@ export class TruffleDB {
...
 
-  async loadCompilations(result: WorkflowCompileResult) {
-    const saga = generateCompileLoad(result, {
+  async getWorkspaceResponse(generatorRequest: WorkspaceRequest) {
+    const { request, variables }: WorkspaceRequest = generatorRequest;
+
+    const response: WorkspaceResponse = await this.query(request, variables);
+
+    return response;
+  }
```
fainashalts, line 62:
```suggestion
  }
  
```

packages/db/src/loaders/commands/compile.ts
```diff
@@ -72,6 +57,7 @@ export function* generateCompileLoad(
   //
   // again going one compilation at a time (for impl. convenience; HACK)
   // (@cds-amal reminds that "premature optimization is the root of all evil")
+  let contractsByCompilation = [];
```
fainashalts, line 60: We had talked about not having this here, and instead using the contracts that are inside the compilations. But I sorta like doing it this way because then we are sure to only return those contracts that have been loaded, right? If I understand this correctly.

packages/db/src/db.ts
```diff
@@ -45,26 +53,62 @@ export class TruffleDB {
...
+  
+  async loadNames(
+    project: DataModel.IProject,
+    contractsByCompilation: Array<DataModel.IContract[]>
+  ) {
```
gnidan, line 67: I'm not keen on the arguments to this method here; can we figure out something cleaner?

packages/db/src/db.ts
```diff
@@ -45,26 +53,62 @@ export class TruffleDB {
...
+    }).next();
+
+    const response = await this.getWorkspaceResponse(projectRequest.value);
+    const projectResponse = response.data.workspace.projectsAdd.projects[0];
+
+    return projectResponse;
+  }
+
+  async loadCompilations(
+    project: DataModel.IProject,
```
gnidan, line 92: I'm not keen on requiring `project` here, it seems an internal concern.

Can we just have this method call `this.loadProject()` rather than requiring it as a pre-condition?

packages/db/src/db.ts
```diff
@@ -45,26 +53,62 @@ export class TruffleDB {
...
+    result: WorkflowCompileResult,
+    names: boolean
```
gnidan, line 94: I wondering if this should go into an options object. Unnamed random booleans are always worrisome to my on the caller side.

gnidan (changes requested):
Some comments / questions / things to change

packages/db/src/db.ts
fainashalts: I changed this for `this.loadProject()` to be called inside of `db.loadCompilations` but also left the call to `loadProject()` in the `load()` function because `loadNetworks` currently requires project id. I know the networks stuff is getting changed around so thought this could be a stopgap.

gnidan (approved):
I had some comments but I want to keep things moving. How bout we merge this and then address the issues separately?

packages/db/src/loaders/commands/compile.ts
```diff
@@ -91,18 +77,13 @@ export function* generateCompileLoad(
...
-      getCurrent
-    );
-
-    yield* generateProjectNamesAssign(toIdObject(project), nameRecords);
+    const loadedContracts = yield* generateContractsLoad(loadableContracts);
+    contractsByCompilation.push(loadedContracts);
   }
 
-  return { project, compilations };
+  const contracts = contractsByCompilation.flat();
```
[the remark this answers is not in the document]
fainashalts: I would have expected CI to fail on Node 10 but I'll make a note to check this on my local!

PR #3369, opened 2020-09-16
Add Solidity-based console logging in test environment

eggplantzzz:
This PR is meant to be a continuation of #2862 

Original description from @adrianmcli:

> We should no longer hold this feature back from our users.
> 
> This PR puts `Console.log(variable)` and `Console.log("myVarLabel", variable)` behind a feature flag into Truffle so people can use it without downloading a different version of Truffle.
> 
> Simply:
> 
> 1. `import "truffle/Console.sol";`
> 2. `Console.log(variable)`
> 3. Run your Truffle tests and see the result

In the future we may make `Console` available in other environments but currently this will only be available in the test environment.


packages/core/lib/testing/TestSource.js
```diff
@@ -97,6 +97,19 @@ module.exports = class TestSource {
...
+    for (const lib of loggingLibraries) {
+      const actualImportPath =
+        typeof BUNDLE_VERSION !== "undefined"
+          ? path.resolve(__dirname, path.basename(importPath))
+          : path.resolve(__dirname, "../logging", path.basename(importPath));
+      if (importPath === `truffle/${lib}.sol`) {
+        const body = fse.readFileSync(actualImportPath, { encoding: "utf8" });
+        return { body, filePath: importPath };
+      }
+    }
```
gnidan, line 112: Did I accidentally add this here when you removed it @eggplantzzz?

packages/migrate/Migration.js
```diff
@@ -202,6 +205,46 @@ class Migration {
...
+      network_id,
+      provider,
+      solidityLogging
+    } = options;
+    if (solidityLogging) {
+      let Console;
+      try {
+        Console = resolver.require("Console");
+      } catch (error) {
+        return;
```
gnidan, line 221: Is it okay that we just let this fail?

packages/resolver/lib/sources/truffle.ts
```diff
@@ -0,0 +1,37 @@
...
+  async resolve(importPath: string) {
+    let body, filePath;
+    if (importPath === "truffle/Console.sol") {
+      const actualImportPath =
+        // @ts-ignore
+        typeof BUNDLE_VERSION !== "undefined"
+          ? path.resolve(__dirname, path.basename(importPath))
+          : path.resolve(
+              __dirname,
+              "../../../../core/lib/logging",
```
gnidan, line 16: I don't like how this is reaching into @truffle/core... especially considering that we might want to do the same here with Assert+DeployedAddresses. I think those contracts need a new home. Any ideas? I was thinking we could have @truffle/solidity-logging for Console

packages/core/lib/testing/TestSource.js
eggplantzzz: No, I think this is where this was.

gnidan: Hm, then seems it's duplicated by your Truffle resolver source, no? Is there a way to remove this duplication?

eggplantzzz: Ok let me check on it again to see which is correct. Maybe I'm mistaken

packages/migrate/Migration.js
eggplantzzz: I suppose we probably should let this fail, or at the very least warn them? It seems better since they have the logging turned on and it would fail.

packages/core/lib/testing/TestSource.js
eggplantzzz: It is duplicated but I'm not sure it is worth it to deal with such a small amount of code. Anything come to mind?

packages/resolver/lib/sources/truffle.ts
gnidan: We could, it just feels gross to have them all there. Maybe the new package could be `@truffle/solidity-libs`, for all three?

packages/migrate/Migration.js
gnidan: It might be the right behavior to abort like this. I'm just wondering if there's a way we can detect certain allowed failure conditions, rather than catch-all aborting every kind of exception with no notice. Like, surely, in some situations where things break, we should alert the user, no?

packages/core/lib/testing/TestSource.js
gnidan: It just makes me wonder... is your Truffle ResolverSource even getting used?

packages/resolver/lib/sources/truffle.ts
eggplantzzz: It seems to me that all these would be at home in `resolver`. They are basically libraries that the `truffle` source resolves. That feels pretty appropriate to me. What do you think?

packages/migrate/Migration.js
eggplantzzz: So really this code should never throw unless there is an error on our side when a user is using Truffle, right? I suppose if they are using the migration library stand-alone and don't configure something correctly it might fail as well. I might be in favor of just printing a warning to the screen that "you are using the solidity logging flag but we cannot resolve the library" or something like that.

packages/core/lib/testing/TestSource.js
eggplantzzz: Currently, it looks to me like `TestSource` will always resolve this logging library and it will never get it from `resolver`.

eggplantzzz: So the conclusion here was to remove `TestSource` altogether since that is now handled by the `truffle` source.

packages/resolver/lib/sources/truffle.ts
eggplantzzz: The decision was made to move all these templates into the `resolver` package.

fainashalts (approved):
A couple of comments but overall LGTM!

eggplantzzz:
Eventually we'll want to re-organize some of this stuff to agree with the organization of https://github.com/trufflesuite/truffle/pull/3395. Perhaps we should wait until that is merged.

gnidan (changes requested):
Marking as request-changes as per comment above re: #3395

packages/contract/lib/utils/index.js
```diff
@@ -80,6 +80,60 @@ const Utils = {
...
+      });
+    });
+
+    const namedLogEvents = receiptLogs.filter(
+      x => x.event === "_TruffleConsoleLogNamed"
+    );
+    namedLogEvents.forEach(event => {
+      const result = event.args;
+
+      // A mapping to show how to output into our console
```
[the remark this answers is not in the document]
eggplantzzz: Hmm, I can't really think of an elegant way of doing this other than creating an awkward function because both of these are in closures. In addition, you'd have to pass the index as well which would be ugly. Any ideas?

packages/migrate/Migration.js
eggplantzzz: So this function got deleted and it now gets deployed and linked like all of the other testing libraries in  `packages/core/lib/testing/SolidityTest.js`.

packages/resolver/solidity/Console.sol
```diff
@@ -0,0 +1,65 @@
+pragma solidity >=0.4.21 <0.7.0;
```
gnidan, line 1: Do we need to update this to support `^0.7.0` now?

gnidan (commented):
Just went through all this again. Looks good!

Looks like the console stuff doesn't support 0.7.0; we probably want to fix that? But apart from that, this looks good to go!

packages/resolver/solidity/Console.sol
eggplantzzz: Yeah I suppose this should have a `<0.8.0`, right?

gnidan: yep and it might have some other changes required

eggplantzzz: updated  the pragma

packages/contract/lib/utils/index.js
```diff
@@ -80,6 +80,60 @@ const Utils = {
...
+
+      // A list of all "supported" types: ["num", "boolean", "str", ...]
+      const types = Object.keys(outputMap);
+
+      // Find out what type our result is, and log it out if it matches
+      types.forEach(type => {
+        if (result.hasOwnProperty(type)) {
+          const label = web3Utils.toAscii(result[0]);
+          const value = outputMap[type];
+          console.log(label, value);
```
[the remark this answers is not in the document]
eggplantzzz: Yeah I'm not sure, we should have a discussion at some point about how we want to deal with all the logging on a large scale in Truffle.

gnidan (approved):
I think we can call this one good now!

PR #3370, opened 2020-09-16
Use fs-extra in Truffle DB

fainashalts:
This PR removes a comment from the `server.js` file and removes unneeded imports. It also replaces `fs` with `fse` everywhere it is used in the Truffle DB project. Making a separate PR just to be sure nothing wonky is going on and nobody has concerns before merging into our longstanding truffle-db branch.


gnidan (approved):
👌

PR #3372, opened 2020-09-18
Fix templates for truffle create (test | contract)

cds-amal:
This PR is to fix an error with scaffolding tests and contracts which we hit during our webinar.

Template updates
  * update pragma to match bare-box
  * remove done callback
  * use arrow function for contract/it callbacks

``` sh
  1) Contract: SS
       should assert true:
     Error: Resolution method is overspecified. Specify a callback *or* return a
      at Context.<anonymous> (test/s_s.js:7:5)

```

## how to test
In a new folder
1. truffle init
1. truffle create test SS
1. truffle create contract SS
1. add a migration file
   ``` solidity
    const SS = artifacts.require("SS");
    module.exports = function (deployer) {
      deployer.deploy(SS);
    };
   ```
1. truffle test and it should not error

### DIFF
--- a/packages/core/lib/commands/create/templates/example.js
+++ b/packages/core/lib/commands/create/templates/example.js
@@ -1,7 +1,12 @@
 const Example = artifacts.require("Example");
 
-contract("Example", () => {
-  it("should assert true", async () => {
+/*
+ * uncomment accounts to access the test accounts made available by the
+ * Ethereum client
+ * See docs: https://www.trufflesuite.com/docs/truffle/testing/writing-tests-in-javascript
+ */
+contract("Example", function (/* accounts */) {
+  it("should assert true", async function () {
     await Example.deployed();
     return assert.isTrue(true);
   });
--- a/packages/core/lib/commands/create/templates/example.js
+++ b/packages/core/lib/commands/create/templates/example.js
@@ -1,9 +1,8 @@
 const Example = artifacts.require("Example");
 
-contract("Example", function() {
-  it("should assert true", async function(done) {
+contract("Example", () => {
+  it("should assert true", async () => {
     await Example.deployed();
-    assert.isTrue(true);
-    done();
+    return assert.isTrue(true);
   });
 });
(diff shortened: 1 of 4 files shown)


packages/core/lib/commands/create/templates/example.js
```diff
@@ -1,9 +1,8 @@
 const Example = artifacts.require("Example");
 
-contract("Example", function() {
-  it("should assert true", async function(done) {
+contract("Example", () => {
```
gnidan, line 3: Technically, mocha says you shouldn't use arrow functions here - cause they make use of `this` internally (`this.setTimeout()`, if you wanted to use it)

But this doesn't matter here.

gnidan (approved):
Thanks for this!

gnidan:
Oh but CI is failing?

packages/core/lib/commands/create/templates/example.js
cds-amal: Oh, I didn't consider this. I'll change it back to conform to Mocha's docs.

cds-amal:
> Oh but CI is failing?

> [@truffle/core: AssertionError: File's solidity version does not match >= 0.5.0 < 0.7.0: expected false to be true
](https://github.com/trufflesuite/truffle/pull/3372/checks?check_run_id=1135856294#step:6:1849)

We use [0>=0.4.22 <0.8.0 in bare-box ](https://github.com/truffle-box/bare-box/blob/master/contracts/Migrations.sol#L2) so the stricter check is confusing. 🤔

gnidan:
Could you update that test? Maybe we don't need to keep that kind of version check at all.

packages/core/lib/commands/create/templates/example.js
gnidan: I take it back - it does matter here. We don't want to mislead people, since this is a template.

cds-amal: adding [the mocha issue for reference](mochajs/mocha/issues/2018).

cds-amal: ++ template reference. using `function(){...}` callback.

gnidan (approved):
Looks good! Thanks for this!

PR #3373, opened 2020-09-18
Bug fix: Remove end-of-line padding in debugger CLI

haltman-at:
I didn't realize that `chromafi` was adding lots of end of padding to the end of each line.  Unfortunately chromafi has no option to turn this off.  So, I updated our fork to add such an option.  Now we pass that option.  I also made it so we pass the `tabsToSpaces: false` option (though it hasn't seemed to affect anything), since we handle the tabs-to-spaces conversion ourselves already.


gnidan (approved):
Not too shabby

Issue #3371, opened 2020-09-18
Compilation crashes when using truffle API (v5.1.45?) 


cgewecke:
## Issue

Cross-reporting from [solidity-coverage 535][1]. 

The crash stack-trace:
```
TypeError: Function.prototype.apply was called on undefined, which is a undefined and not a function
    at [repo dir]/node_modules/truffle/build/webpack:/packages/workflow-compile/utils.js:34:1
    at new Promise (<anonymous>)
    at [repo dir]/node_modules/truffle/build/webpack:/packages/workflow-compile/utils.js:27:1
    at [repo dir]/node_modules/truffle/build/webpack:/packages/workflow-compile/legacy/index.js:101:1
    at Array.map (<anonymous>)
    at Object.compileSources ([repo dir]/node_modules/truffle/build/webpack:/packages/workflow-compile/legacy/index.js:91:1)
    at Object.compile ([repo dir]/node_modules/truffle/build/webpack:/packages/workflow-compile/legacy/index.js:54:1)
    at plugin ([repo dir]/node_modules/solidity-coverage/plugins/truffle.plugin.js:101:29)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
Truffle v5.1.45 (core: 5.1.45)
Node v12.18.4
```

The code at solidity-coverage is +/-: 
```js
const truffle = require('truffle'); 
await truffle.contracts.compile(config);
```

Prior to 8c5555e798cf1eb17d0d76d8e737e6bfc0abd032 `require('truffle').contracts` was equivalent to `require('@truffle/workflow-compile')`
...since then it's been bound to `@truffle/workflow-compile/legacy`.

Happy to update solidity-coverage to use a new API if exposed fwiw. 🙂  

[1]: https://github.com/sc-forks/solidity-coverage/issues/535

## Environment

* Operating System: 
* Ethereum client: Ganache
* Truffle version (`truffle version`): 5.1.45 (may affect earlier versions too...)
* node version (`node --version`): 12
* npm version (`npm --version`):

eggplantzzz:
Oh whoops, sorry @cgewecke. Looks like there was an oversight involving the integration of the legacy package with `external-compile` and `compile-vyper`. We'll try and have a patch out next week!

eggplantzzz:
By the way, if you happen to have something I can verify against once I have a fix then let me know. I'll test and make sure it is behaving as desired before we release.

cgewecke:
@eggplantzzz 

[...]

PR #3375, opened 2020-09-21
Bug fix: Fix bugs with workflow-compile/legacy

eggplantzzz:
There were a couple of oversights while updating the compilation packages and trying to maintain backwards compatibility with the version of workflow-compile exposed in `@truffle/core`. This PR creates a couple of adapters for `compile-vyper` and `external-compile` in `workflow-compile/legacy`, edits a method name that wasn't updated, and sends data to the events system in the proper format.

This issue was surfaced in https://github.com/trufflesuite/truffle/issues/3371 by @cgewecke. These changes seem to resolve the issues experienced with `solidity-coverage` as I tested them locally (with a bit of doctoring up of the plugin code).


packages/workflow-compile/legacy/index.js
```diff
@@ -92,18 +92,53 @@ const Contracts = {
...
-        const compileFunc = multiPromisify(
-          config.all === true || config.compileAll === true
-            ? compile.all
-            : compile.necessary
-        );
-
-        let [contracts, output, compilerUsed] = await compileFunc(config);
+        config.compilersInfo = [];
+        let contracts, output, compilerUsed;
+        if (compiler === "solc") {
```
gnidan, line 97: What happens if `compiler` isn't passed?

packages/workflow-compile/legacy/index.js
```diff
@@ -92,18 +92,53 @@ const Contracts = {
...
+            }
+          );
+          contracts = result.contracts;
+          output = result.output;
+          if (compilations[0] && compilations[0].compiler) {
+            compilerUsed = {
+              name: compilations[0].compiler.name,
+              version: compilations[0].compiler.version
+            };
+          }
```
gnidan, line 137: Should we use the `NewToLegacy` shims for these?

eggplantzzz: I don't think those shims get us what we want, do they? The only shims in there are the shims for a contract and the bytecode.

gnidan: Don't we need to shim the bytecode?

eggplantzzz: Oh, I suppose you are right about that. Doh!

gnidan (approved):
Looks good!

gnidan:
Merging this so I can have the latest to rebase #2713

PR #3377, opened 2020-09-21
Internal improvement: Some adjustments for the new compile format

haltman-at:
This PR has two purposes.

The first commit redoes `core/lib/debug/compile.js`, and the other files in that directory that invoke it, so that it can actually return multiple compilations.  PR #3332 accounted for the new compile return format by mashing the different compilations together; here instead I've separated them into different compilations.

The second commit attempts to fix the types on `shimArtifacts` from `codec` to account for the fact that `bytecode` and `deployedBytecode` may be either strings or `Bytecode` objects.  To get it to compile I had to alter some of the types in compile-common, note.  This also makes `compile-common` a devDependency of codec (but not a run-time dependency).

### DIFF
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -4,10 +4,11 @@ const debug = debugModule("codec:compilations:utils");
 import * as Ast from "@truffle/codec/ast";
 import * as Compiler from "@truffle/codec/compiler";
 import { ContractObject as Artifact } from "@truffle/contract-schema/spec";
+import { CompiledContract } from "@truffle/compile-common";
 import { Compilation, Contract, Source } from "./types";
 
 export function shimArtifacts(
-  artifacts: Artifact[],
+  artifacts: (Artifact | CompiledContract)[],
   files?: string[],
   shimmedCompilationId = "shimmedcompilation",
   externalSolidity = false
@@ -20,7 +21,6 @@ export function shimArtifacts(
   for (let artifact of artifacts) {
     let {
       contractName,
-      contract_name,
       bytecode,
       sourceMap,
       deployedBytecode,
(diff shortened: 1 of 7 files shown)


packages/codec/lib/compilations/utils.ts
```diff
@@ -4,10 +4,11 @@ const debug = debugModule("codec:compilations:utils");
 import * as Ast from "@truffle/codec/ast";
 import * as Compiler from "@truffle/codec/compiler";
 import { ContractObject as Artifact } from "@truffle/contract-schema/spec";
+import { CompiledContract } from "@truffle/compile-common";
 import { Compilation, Contract, Source } from "./types";
 
 export function shimArtifacts(
-  artifacts: Artifact[],
+  artifacts: (Artifact | CompiledContract)[],
```
gnidan, line 11: Don't we need to shim one way or the other here, for bytecodes? Since artifacts use string bytecodes and `CompiledContract` uses `Bytecode` objects

gnidan: Note: these shims are now in @truffle/compile-common itself, which will have to turn into a regular `dependency`

haltman-at: No, because it's OK with outputting either, so there's no need to shim.

haltman-at:
Huh, CI is failing due to `codec` being unable to find `compile-common`... why would that happen?  Works on my machine...

Although, running things on my own machine, looks like I introduced a circular dependency... so I guess that will need fixing somehow before merge. :-/

```
lerna WARN ECYCLE @truffle/codec -> @truffle/compile-common -> @truffle/resolver -> @truffle/contract -> @truffle/debug-utils -> @truffle/codec
```

gnidan:
Oh no!!!!

haltman-at:
OK, merged in `develop` (post-#3379) to hopefully fix the problem.

gnidan (approved):

PR #3379, opened 2020-09-22
Remove @truffle/resolver dependency in @truffle/compile-common

gnidan:
- Remove a bunch of unused imports
- Replace imported `ResolverSource` type with an interface that contains only the relevant part.
- Get rid of the dependency


haltman-at (approved):
Makes sense to me!  As a way of resolving circular dependencies, at least.

gnidan:
Thanks!

Issue #3380, opened 2020-09-22
Allow overriding maxBuffer when compiling with Dockerized solc


coeniebeyers:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Exactly like https://github.com/trufflesuite/truffle/issues/3140, however 1024 * 1024 * 100 isn't enough for our instrumented files when we are trying to produce coverage reports.

## Steps to Reproduce

Run truffle coverage.

## Expected Behavior

Coverage to run.

## Actual Results

We're getting the same error as in https://app.circleci.com/pipelines/github/trufflesuite/truffle/184/workflows/81f7c350-d718-46f8-8450-9f941a592cc6/jobs/79:
```
Error: spawnSync /bin/sh ENOBUFS
    at Object.spawnSync (internal/child_process.js:1002:20)
    at spawnSync (child_process.js:614:24)
    at execSync (child_process.js:661:13)
    at Object.compile (/home/circleci/project/node_modules/truffle/build/webpack:/packages/compile-solidity/compilerSupplier/loadingStrategies/Docker.js:23:1)
    at invokeCompiler (/home/circleci/project/node_modules/truffle/build/webpack:/packages/compile-solidity/run.js:257:1)
```

## Environment

* Operating System: Ubuntu 20.04
* Ethereum client: Ganache
* Truffle version (`truffle version`): 5.1.37
* node version (`node --version`):  12.18.3
* npm version (`npm --version`): 6.14.6

Could we change maxBuffer for the docker loading strategy to be set via environment variable?

PR #3381, opened 2020-09-22
Set maxBuffer for docker loading strategy via configuration value

coeniebeyers:
Proposal for https://github.com/trufflesuite/truffle/issues/3380
Thanks to @rynobey for also debugging/fixing this on our end.


eggplantzzz:
Hey @coeniebeyers, thanks for the PR! So far Truffle doesn't have any precedent for using environment variables, is there any way you could make this something that you would configure? Check out the current configuration options [here](https://www.trufflesuite.com/docs/truffle/reference/configuration#compiler-configuration). Perhaps it could be listed under the `compilers` property? Something like `compilers.solc.spawn.maxBuffer`? What do you think?

coeniebeyers:
Hi @eggplantzzz! np! Ah, yes, then using something like `compilers.solc.spawn.maxBuffer` makes sense. 

Please take a look at the updated PR. Is there somewhere I can add this to documentation?

Please note that I'm trying to test by locally installing truffle from source in our project with `npm install <path to truffle>` but even after linking with lerna it doesn't seem to be able to do that, so there might be some things wrong still.

coeniebeyers:
@eggplantzzz thanks for the info, I was able to test locally and it looks good after my latest commit. Please let me know if anything needs to change.

eggplantzzz:
One way that I test is that I clone the repo, make my changes, and then run `packages/core/cli.js <command>`. Maybe you will find that helpful. I'll take a look at your changes, thanks!

packages/compile-solidity/compilerSupplier/loadingStrategies/Docker.js
```diff
@@ -6,12 +6,15 @@ const semver = require("semver");
...
 
-// Set a sensible limit for maxBuffer
-// See https://github.com/nodejs/node/pull/23027
-const maxBuffer = 1024 * 1024 * 100;
-
 class Docker extends LoadingStrategy {
   async load() {
+    // Set a sensible limit for maxBuffer
+    // See https://github.com/nodejs/node/pull/23027
+    let maxBuffer = 1024 * 1024 * 100;
```
eggplantzzz, line 13: Another option here is
```
const maxBuffer = this.config.spawn && this.config.spawn.maxBuffer ?
  this.config.spawn.maxBuffer :
  1024 * 1024 * 100;
```
but what you did is good too!

eggplantzzz (approved):
I think this looks good to me.

fainashalts (approved):
LGTM as well! Thanks @coeniebeyers!

coeniebeyers:
Sure, np!

PR #3385, opened 2020-09-23
Add `sourcesWithDependencies` to compile interface

gnidan:
- Add to compile-common types
- Implement as pass-through for compile-vyper and external-compile

### DIFF
--- a/packages/compile-vyper/index.js
+++ b/packages/compile-vyper/index.js
@@ -130,6 +130,13 @@ const Compile = {
     });
   },
 
+  async sourcesWithDependencies({ paths = [], options }) {
+    return await Compile.sources({
+      sources: paths,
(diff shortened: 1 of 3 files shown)


packages/compile-vyper/index.js
```diff
@@ -130,6 +130,13 @@ const Compile = {
     });
   },
 
+  async sourcesWithDependencies({ paths = [], options }) {
+    return await Compile.sources({
+      sources: paths,
```
gnidan, line 135: This exposes the fact that compile-vyper's `sources` method doesn't conform to the type defined by compile-common

eggplantzzz (approved):

gnidan:
Thanks @eggplantzzz !

PR #3386, opened 2020-09-23
Rework shim usage

fainashalts:
This PR makes changes to the Truffle DB code so that it can adequately reflect/handle changes to the output provided from` `WorkflowCompile` and the shims and types in the `compile-common` package. As part of the work to do this, `develop` was pulled in to ensure that changes are up-to-date. 

The main changes that have been made are as follows: 
1) The compilation output mock json for testing was updated to reflect the new format
2) `artifactsLoader` was changed to handle the new format -- whereas before compilations were designated by compiler, we are now using an index to load compilations. It is assumed each compilation has one compiler. 
3) Types available in compile-common were updated to be used in the db code, where needed. The  `CompiledContract` type was removed from the db types as no longer needed.


gnidan (approved):
Looks good! Love seeing those extra type definitions disappear

PR #3392, opened 2020-09-24
Allow importing ABI JSON files in Solidity

gnidan:
Define ABI resolver source that extends from FS resolver source:

- Support only `resolve()` (explicitly exclude `require()`)
- Catch imports that end in `.json` or `.abi.json`
- Defer to `abi-to-sol` to generate Solidity source

### DIFF
--- a/packages/resolver/lib/resolver.ts
+++ b/packages/resolver/lib/resolver.ts
@@ -3,7 +3,7 @@ const expect = require("@truffle/expect");
 const provision = require("@truffle/provisioner");
 
 import { ResolverSource } from "./source";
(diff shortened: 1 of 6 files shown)


eggplantzzz:
This seems really cool, what plans do you have for this?

gnidan:
> This seems really cool, what plans do you have for this?

I ended up doing the Solidity conversion in a [side-project](https://github.com/gnidan/abi-to-sol) during my week off last week, but it seems like this change might still be desirable?

I've gone and updated this PR to use that package.

eggplantzzz:
I like it!

eggplantzzz (approved):

gnidan:
Thanks!

PR #3395, opened 2020-09-24
Internal improvement: Rework test files

eggplantzzz:
This PR removes the `TestResolver` class. Instead it uses a regular resolver that includes one extra source: the `truffle` source. It also moves all the Solidity libraries into the `resolver` package.


eggplantzzz:
Just a note; I had to comment out the `development` configuration for `truffle-init-default#typescript` which is used in some of the tests. Those tests pass locally for me but in CI would consistently fail. Removing the network configuration, however, let's them pass for some reason that I'm not sure about.

packages/resolver/lib/resolver.ts
```diff
@@ -3,17 +3,26 @@ const expect = require("@truffle/expect");
...
 import { ResolverSource } from "./source";
-import { constructSources } from "./sources";
+import { EthPMv1, NPM, GlobalNPM, FS, Truffle } from "./sources";
 
 export class Resolver {
   options: any;
   sources: ResolverSource[];
 
-  constructor(options: any) {
+  constructor(options: any, includeTruffleSource: boolean) {
```
eggplantzzz, line 12: Is there a better way to denote that the resolver being instantiated should include the `Truffle` source? Do we want to eventually include the `Truffle` source in all resolvers?

packages/core/test/lib/commands/test.js
```diff
@@ -1,14 +1,16 @@
...
+} = require("../../../lib/commands/test/determineTestFilesToRun");
 const Artifactor = require("@truffle/artifactor");
 const Resolver = require("@truffle/resolver");
 const MemoryStream = require("memorystream");
 const path = require("path");
 const fs = require("fs-extra");
 const glob = require("glob");
 const WorkflowCompile = require("@truffle/workflow-compile");
-const Test = require("@truffle/core/lib/test");
+const Test = require("../../../lib/testing/Test");
```
eggplantzzz, line 13: This is kind of ugly but do we really want to require `@truffle/core`?

packages/core/lib/testing/SolidityTest.js
```diff
@@ -4,7 +4,6 @@ const Deployer = require("@truffle/deployer");
 const { Compile } = require("@truffle/compile-solidity");
 const { Shims } = require("@truffle/compile-common");
 const RangeUtils = require("@truffle/compile-solidity/compilerSupplier/rangeUtils");
-const path = require("path");
```
gnidan, line 7: Pretty cool that we don't need this anymore :)

packages/core/test/lib/commands/test.js
gnidan: not sure it matters?

packages/resolver/lib/resolver.ts
gnidan: Eventually: definitely yes.
Now: maybe yes.

If the answer to "now" is "no", then we should do this:
```suggestion
  constructor(options: any, includeTruffleSource: boolean = false) {
```

packages/resolver/lib/resolver.ts
```diff
@@ -3,17 +3,26 @@ const expect = require("@truffle/expect");
...
     expect.options(options, ["working_directory", "contracts_build_directory"]);
 
     this.options = options;
-    this.sources = constructSources(options);
```
gnidan, line 16: I was gonna use that abstraction! One of these days... I swear...

Probably good call removing it.

packages/resolver/lib/resolver.ts
```diff
@@ -3,17 +3,26 @@ const expect = require("@truffle/expect");
...
+    this.sources = [
+      new EthPMv1(options.working_directory),
+      new NPM(options.working_directory),
+      new GlobalNPM(),
+      new FS(options.working_directory, options.contracts_build_directory)
+    ];
+
+    if (includeTruffleSource) {
+      this.sources.push(new Truffle(options));
```
gnidan, line 24: Does this need to come first? Shouldn't it have higher precedent than the others?

packages/resolver/lib/sources/truffle/Deployed.ts
```diff
@@ -1,11 +1,15 @@
 const web3Utils = require("web3-utils");
-const RangeUtils = require("@truffle/compile-solidity/compilerSupplier/rangeUtils");
+import RangeUtils from "@truffle/compile-solidity/compilerSupplier/rangeUtils";
 
-var Deployed = {
+type solcOptionsArg = {
+  solc: { version: string };
+};
+
+export const Deployed = {
```
gnidan, line 8: Wanna make this a class with all `static` methods? Syntax might be nicer.

packages/resolver/lib/sources/truffle/index.ts
```diff
@@ -85,21 +82,38 @@ module.exports = class TestSource {
...
+        const body = fse.readFileSync(actualImportPath, { encoding: "utf8" });
         return { body, filePath: importPath };
       }
     }
+
+    return { body: null, filePath: null };
+  }
+
+  require(importPath: string): null {
+    return null;
```
gnidan, line 111: You lost the comment here (it's probably worth keeping... this `null` return behavior is annoyingly opaque without it.)

packages/resolver/lib/sources/truffle/index.ts
```diff
@@ -85,21 +82,38 @@ module.exports = class TestSource {
...
   }
 
-  resolveDependencyPath(importPath, dependencyPath) {
-    return dependencyPath;
+  // Here we're resolving from local files to local files, all absolute.
```
gnidan, line 114: Does it need to be absolute? Seems more correct not to be.

packages/truffle/webpack.config.js
```diff
@@ -5,6 +5,14 @@ const webpack = require("webpack");
 const pkg = require("./package.json");
 const rootDir = path.join(__dirname, "../..");
 const outputDir = path.join(__dirname, "build");
+const truffleLibraryDirectory = path.join(
+  __dirname,
+  "../..",
+  "node_modules",
+  "@truffle/resolver",
+  "lib",
```
gnidan, line 13: I think @truffle/resolver's `yarn prepare` should put these in `packages/resolver/dist/` instead of having `truffle` grab them from `lib`.

(As it is, there's an inconsistency: @truffle/resolver's `package.json` includes `dist` in `files`, but not `lib`, so this only works in the monorepo, not that we care about that here...)

packages/truffle/webpack.config.js
```diff
@@ -150,180 +158,49 @@ module.exports = {
...
       },
       {
-        from: path.join(
-          __dirname,
-          "../..",
-          "node_modules",
-          "@truffle/core",
-          "lib",
-          "testing",
-          "AssertAddressPayableArray.sol"
```
gnidan, line 193: Why'd this get removed? This sounds like something accidental. Do we care about this type?

gnidan (changes requested):
Here's a few comments, a couple small changes I think.

Overall this looks nice! Curious how you feel about removing the `includeTruffleLibraries` and just having it turned on in all cases (not just test)

packages/resolver/lib/resolver.ts
eggplantzzz: I'm not sure which is better. It probably should come first I guess.

packages/truffle/webpack.config.js
eggplantzzz: This was not an accident, I don't think it is used. It was commented out in the list [here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/testing/TestSource.js#L78). Should we add it back in?

gnidan: We can do this if someone complains

packages/resolver/lib/sources/truffle/index.ts
gnidan: Seems like tests pass without this coming back as absolute! Marking as resolved.

gnidan:
Hm, in order to get this into the release this week, let's call it out of scope to remove the `includeTruffleLibraries` flag.

@eggplantzzz when you get back, can you remember that we should remove this?

gnidan (approved):

PR #3396, opened 2020-09-24
Internal improvement: Reorganize data saga to defer selector calls and variable assignments

haltman-at:
This PR reorganizes the debugger's data saga somewhat to avoid the slew of globalish variables up front and to not make so many selector calls when they aren't needed.  Also, while I was at it, I changed a bunch of `let`s to `consts` and added some comments in some cases where I didn't do that.

See, currently, the data saga starts with a bunch of selector calls and variable declarations.  Not only is this basically making a bunch of globalish variables, but the unnecessary selector calls could potentially slow things down, too.

So, I've taken most of these calls and declarations and moved them to where they're actually used.  I did leave up front a few ones that are common enough that it would be inconvenient to move them below, but most of that clutter is now gone...

Anyway, the result should be functionally identical, and perhaps a bit faster if we're lucky.

### DIFF
--- a/packages/debugger/lib/data/sagas/index.js
+++ b/packages/debugger/lib/data/sagas/index.js
@@ -200,6 +200,7 @@ function* variablesAndMappingsSaga() {
   }
 
   let node = yield select(data.current.node);
+  //can mutate in FunctionTypeName, YulLiteral, and YulIdentifier cases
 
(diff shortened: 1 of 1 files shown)


gnidan (approved):
Easy!

PR #3398, opened 2020-09-25
Internal improvement: Turn on eslint for typescript

haltman-at:
This PR turns on linting for typescript with `eslint`, or rather with `typescript-eslint`, as `eslint` doesn't quite do it itself.

The changes to accomplish this are as follows:

1. Add `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` as dev dependencies of Truffle (in the project root), and enable them in `.eslintrc.json`
2. Certain rules in `.eslintrc.json` were on [the list](https://www.npmjs.com/package/@typescript-eslint/eslint-plugin#extension-rules) of rules that now require the `@typescript-eslint/` prefix, so I added that.
3. Finally, of course, I added an `eslint --fix` step for TS files in our pre-commit script.

One thing worth noting here: In testing whether this worked, I tried making sure the `quote-props` rule still worked, only to find that, uh, it already didn't work...?  So, uh, yeah, dunno what's going on there... (ultimately I was able to do the test I wanted with a different rule...)

Note I didn't turn on any new Typescript-specific rules or anything, just left the existing ones in place (just added the needed prefix to the two that needed that).

Anyway (assuming I've done this right) we now have Typescript linting!  In fact just testing this out it already turned up some problems in `decoder`, so, uh, we'll see what problems it turns up more generally...

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -3,17 +3,19 @@
     "es6": true,
     "node": true
   },
+  "parser": "@typescript-eslint/parser",
+  "plugins": ["@typescript-eslint"],
   "parserOptions": {
     "ecmaVersion": 9,
     "sourceType": "module"
   },
   "rules": {
-    "no-unused-vars": [
+    "@typescript-eslint/no-unused-vars": [
       "error",
       { "argsIgnorePattern": "^_", "varsIgnorePattern": "^debug$|^_" }
     ],
     "quote-props": [ "error", "consistent" ],
-    "semi": [ "error", "always" ],
+    "@typescript-eslint/semi": [ "error", "always" ],
     "no-mixed-operators": "warn",
     "no-tabs": "warn",
     "no-useless-escape": "warn"
--- a/yarn.lock
+++ b/yarn.lock
@@ -1017,10 +1017,31 @@
     call-me-maybe "^1.0.1"
     glob-to-regexp "^0.3.0"
 
+"@nodelib/fs.scandir@2.1.3":
+  version "2.1.3"
+  resolved "https://registry.yarnpkg.com/@nodelib/fs.scandir/-/fs.scandir-2.1.3.tgz#3a582bdb53804c6ba6d146579c46e52130cf4a3b"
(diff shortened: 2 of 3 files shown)


.eslintrc.json
```diff
@@ -3,17 +3,19 @@
...
   },
+  "parser": "@typescript-eslint/parser",
+  "plugins": ["@typescript-eslint"],
   "parserOptions": {
     "ecmaVersion": 9,
     "sourceType": "module"
   },
   "rules": {
-    "no-unused-vars": [
+    "@typescript-eslint/no-unused-vars": [
```
gnidan, line 13: Will this still apply to JS?

.eslintrc.json
```diff
@@ -3,17 +3,19 @@
...
       "error",
       { "argsIgnorePattern": "^_", "varsIgnorePattern": "^debug$|^_" }
     ],
     "quote-props": [ "error", "consistent" ],
-    "semi": [ "error", "always" ],
+    "@typescript-eslint/semi": [ "error", "always" ],
```
gnidan, line 18: Same: will this still apply to JS?

yarn.lock
```diff
@@ -1557,6 +1578,11 @@
   resolved "https://registry.yarnpkg.com/@types/json-schema/-/json-schema-7.0.3.tgz#bdfd69d61e464dcc81b25159c270d75a73c1a636"
   integrity sha512-Il2DtDVRGDcqjDtE+rF8iqg1CArehSK84HZJCT7AMITlyXRBpuPhqGLDQMowraqqu1coEaimg4ZOqggt6L6L+A==
 
+"@types/json-schema@^7.0.3":
```
gnidan, line 1581: huh what are you doing here

.eslintrc.json
haltman-at: Yup, it does, I tested it.

.eslintrc.json
haltman-at: Didn't explicitly test this one, but presumably given the above.

yarn.lock
haltman-at: Hell if I know!

gnidan (approved):

PR #3399, opened 2020-09-25
Dependency update: Update eslint to 7.9.0

haltman-at:
Updates `eslint`.  I was hoping this would fix the `quote-props` problem.  It didn't.  (I'll file an issue with them.)  PRing this anyway.

Edit: Oops.  There is no `quote-props` problem.  It works fine, silly me.  Anyway, updating all the same.


haltman-at:
...also apparently `quote-props` works, just not all the time...?  I'm going to need to look into this...

gnidan (approved):

PR #3406, opened 2020-10-01
Bug fix: Fix URL construction in the CompilerSupplier

eggplantzzz:
Truffle is currently constructing bad URLs for fetching the Solidity compiler version list. It is adding an extra trailing "/" to the base of the URL. This PR strips off trailing commas from the base part of the URL before adding "/list.json".


haltman-at (approved):
Makes sense.  Looks to me like the actual bug here was the use of `compilerRoots.length` instead of `compilerRoots.length - 1`?  But this is perhaps a cleaner fix!

eggplantzzz:
Yeah, I originally fixed the indexing but @davidmurdoch provided the above solution which seemed better!

Issue #2065, opened 2019-05-30
ABI Contract Schema Tuple Components/Children are not Verified for their Types


seesemichaelj:
You'll see https://github.com/trufflesuite/truffle/pull/2064/files#diff-da060bd5170ce9e90e14ece903806d3eR135 says that the `components` property has an array of `object` types. This is *supposed to be* `"items": { "$ref": "#/definitions/Parameter" }`, but the https://github.com/bcherny/json-schema-to-typescript dependency we use to generate typings fails due to the recursion  (gets a `Maximum call stack size exceeded` error when building), despite the schema supports this recursion.

To make sure that users contracts that have structs of structs, we only check if the ABI child is an object, rather than verify it is one of the `Parameter` types.

The dependency has known about this issue for 2 years now (https://github.com/bcherny/json-schema-to-typescript/issues/76 which depends on another dependency issue https://github.com/APIDevTools/json-schema-ref-parser/issues/37), but a fix has yet to come.

This issue is to track the problem, and potentially fix the upstream dependencies and validate tuple components properly.

fainashalts:
@seesemichaelj do you know the status of this issue? Is it resolved now? Thanks!

seesemichaelj:
I don't, sorry

fainashalts:
Hi @seesemichaelj just wanted to check on this issue so we can prioritize it -- this does sound like an issue, but where is this validation needed/what is the use case in which it came up? Thanks!

seesemichaelj:
Validation is something that this package provides, and therefore it should be accurate. I'm not sure of all of the use cases, but Teams validates artifacts uploaded for External Deployments. Specifically, looking at my comment and the code it points to, it seems that the lamen terms of what the issue is:

- This validation needs to allow structs of structs
- The **technically correct** way to do this is to change https://github.com/trufflesuite/truffle/pull/2064/files#diff-da060bd5170ce9e90e14ece903806d3eR135 to `"items": { "$ref": "#/definitions/Parameter" }`
- However, making this change causes circular dependency issues caused by upstream packages. If those packages supported circular dependencies (which JSON Schema *allows*), then we could make this change and call it a day
- We currently have a bandaid implemented to allow structs of structs, but it's very permissive by saying "any object will do" instead of specifically saying it has to be a struct (aka Parameter)

So the bug is that we can potentially let some non-valid contracts come off as valid because we don't have a correct mechanism to do that validation.

[...]

PR #3414, opened 2020-10-05
Validate tuple parameters in ABI schema

gnidan:
#2065 

- Ensure non-tuple parameters do not allow `components` property
- Ensure tuple parameters require `components` property
- Ensure components validate as parameters
- Ensure there is at least one component in each tuple

### DIFF
--- a/packages/contract-schema/spec/abi.spec.json
+++ b/packages/contract-schema/spec/abi.spec.json
@@ -159,8 +159,7 @@
           "type": { "$ref": "#/definitions/Type" },
           "components": {
             "type": "array",
-            "items": { "$ref": "#/definitions/Parameter" },
-            "minItems": 1
+            "items": { "$ref": "#/definitions/Parameter" }
           },
           "internalType": { "type": "string" }
         },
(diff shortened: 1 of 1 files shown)


packages/contract-schema/spec/abi.spec.json
```diff
@@ -144,37 +144,75 @@
...
-      "then": { "required": ["name", "type", "components"] },
-      "else": { "required": ["name", "type"] }
+      "then": {
+        "properties": {
+          "name": { "$ref": "#/definitions/Name" },
+          "type": { "$ref": "#/definitions/Type" },
+          "components": {
+            "type": "array",
+            "items": { "$ref": "#/definitions/Parameter" },
+            "minItems": 1
```
haltman-at, line 163: For compatibility with Solidity 0.4.x, I think you should remove this requirement; Solidity 0.4.x allows empty structs.

haltman-at (changes requested):
One change I think is necessary, otherwise looks good!

haltman-at (approved):
Seems good to me!

gnidan:
Thanks!

PR #3415, opened 2020-10-05
Support pre-stateMutability ABIs in schema

gnidan:
Older versions of solc didn't emit `stateMutability` in ABI output.

This PR makes either `stateMutability` required **or** `payable` and/or `constant` (depending on the kind of function)

### DIFF
--- a/packages/contract-schema/spec/abi.spec.json
+++ b/packages/contract-schema/spec/abi.spec.json
@@ -64,7 +64,7 @@
           "$ref": "#/definitions/StateMutability"
         },
         "constant": { "type": "boolean" },
-        "payable": { "type": "boolean", "default": false }
+        "payable": { "type": "boolean" }
       },
       "required": ["name", "inputs", "stateMutability"],
       "additionalProperties": false
@@ -85,7 +85,7 @@
           "$ref": "#/definitions/StateMutability"
         },
         "constant": { "type": "boolean" },
-        "payable": { "type": "boolean", "default": false }
+        "payable": { "type": "boolean" }
       },
       "required": ["type", "inputs", "stateMutability"],
       "additionalProperties": false
@@ -102,7 +102,7 @@
           "$ref": "#/definitions/StateMutability"
         },
         "constant": { "type": "boolean" },
-        "payable": { "type": "boolean", "default": false }
+        "payable": { "type": "boolean" }
       },
       "required": ["type", "stateMutability"],
       "additionalProperties": false
--- a/packages/contract-schema/spec/abi.spec.json
+++ b/packages/contract-schema/spec/abi.spec.json
@@ -66,7 +66,10 @@
         "constant": { "type": "boolean" },
         "payable": { "type": "boolean" }
       },
-      "required": ["name", "inputs", "stateMutability"],
+      "anyOf": [
+        { "required": ["name", "inputs", "stateMutability"] },
+        { "required": ["name", "inputs", "constant", "payable"] }
+      ],
       "additionalProperties": false
     },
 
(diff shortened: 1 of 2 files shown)


melyalex (approved):

packages/db/src/artifacts/json/schema.ts
```diff
@@ -170,8 +170,24 @@ const translations = [
...
+                    // `"type"`
+                    (itemType.required ||
+                      itemType.anyOf
+                        .map(option => new Set(option.required))
+                        .reduce((a, b) =>
+                          !a ? b : [...a].filter(x => b.has(x))
+                        )),
+                    "type"
+                  ])
+                )
```
gnidan, line 190: For context: the module containing this change is used to convert the JSON-Schema into GraphQL.

We'll soon be removing this (since Truffle DB won't need a GraphQL version of this schema anymore), so the approach I've taken is just to squash all the `anyOf`s via intersection into a single `required` property, calling everything that's not always required as just plain nullable

packages/contract-schema/spec/abi.spec.json
```diff
@@ -85,9 +88,12 @@
           "$ref": "#/definitions/StateMutability"
         },
         "constant": { "type": "boolean" },
```
haltman-at, line 90: Should `constant` be deleted here?  Like there's no way it should ever occur here, right?

packages/contract-schema/spec/abi.spec.json
```diff
@@ -102,9 +108,13 @@
...
```
haltman-at, line 110: Same comment as above

haltman-at (commented):
JSON changes basically seem OK to me, although see above about possibly removing `constant` where not warranted.  I don't really understand the GraphQL related part.

packages/contract-schema/spec/abi.spec.json
gnidan: I'm not sure what the behavior used to be in older solc - do you know offhand?

packages/contract-schema/spec/abi.spec.json
gnidan: Same question as above :)

packages/contract-schema/spec/abi.spec.json
haltman-at: Hm, checking the docs it might actually produce it after all.  I'm actually not sure offhand, worth checking.

gnidan: Compiling this:

```solidity
pragma solidity ^0.4.0;

contract Foo {
  function Foo () public {
  }
}
```

With:

```console
$ npx solc@0.4.4 --abi ./Foo.sol
 ```

I get:

```json
[{"inputs":[],"type":"constructor"}]
```

I can't seem to get `constant` to appear in the constructor ABI. But I'm not sure we want to remove it from the spec here? Especially if the docs suggest that it's allowable. WDYT?

packages/contract-schema/spec/abi.spec.json
```diff
@@ -85,9 +88,12 @@
...
-        "payable": { "type": "boolean", "default": false }
+        "payable": { "type": "boolean" }
```
gnidan, line 91: In testing the output on older solc versions, it seems that `"payable"` only appears when you specify the keyword in the Solidity source.

Commenting here cause this will need to change accordingly.

gnidan: Hm, so Truffle only supports back to 0.4.12. I tried 0.4.11 and it includes `payable` regardless of whether the constructor is payable or not, so I'm keen to ignore the case where `payable` isn't required.

@haltman-at could you resolve if you concur that we can ignore this issue?

packages/contract-schema/spec/abi.spec.json
haltman-at: OK yeah (considering the point you raised on Slack about someone could write it by hand) I guess it makes sense to not change this.

haltman-at (approved):
Approving for the schema changes, but I can't really meaningfully comment on the GraphQL changes.

gnidan:
Merging to get this into release this week. Thanks @haltman-at!

PR #3416, opened 2020-10-05
Make schemaAbiToAbi idempotent

gnidan:


haltman-at (approved):
Huh.  You can do `abiLoose as SchemaAbi`?  (Is doing that necessary, actually? I don't see why...)

Anyway yeah, makes sense.

gnidan:
It was necessary cause it wouldn't compile otherwise!

PR #3417, opened 2020-10-06
Experimental support for compiling with multiple versions of the Solidity compiler

eggplantzzz:
This PR creates a utility for analyzing pragma statements in Solidity source files. Given a source file, the utility gathers its dependencies, strips out the pragma statements, and finds the newest version of the Solidity compiler that satisfies all statements. It then compiles the group with that version.

~Currently, this implementation is pretty dumb and there is a lot of redundancy as it performs one compilation per source file present in the the user's `contracts` directory.~ It currently finds the newest version of the compiler that satisfies each source file's requirements, then groups them all by version, and does compilation. There are a number of issues that remain to be solved including how to persist the artifacts given that a given dependency will be compiled multiple times with possibly different versions of the compiler. Additionally it will also always warn the user that they have multiple contracts with the same name if they have any imports at all.


melyalex (approved):

cds-amal:
I'm getting an error running the tests. Is it possible to seed, or cache fetching the solc versions for testing?

`     Error: Timeout of 20000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/amal/work/truffle/packages/compile-solidity/test/compilerSupplier/index.js)
`

eggplantzzz:
In the past we have primarily just increased the timeout for tests. Sometimes those requests take a long time. Truffle does automatically cache each version of the Solidity compiler that it downloads and I think seeding might be too much as those compiler files have significant size. I'll just increase the timeout for now but if you have any more ideas for addressing this then I'd love to hear them!

eggplantzzz:
Currently there are timeout failures in CI for one of the tests. Checking it locally, it looks like the `Profiler` takes most of the time in the flow.

cds-amal (approved):
Looks good, tests works.

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,123 @@
...
+      throwCompilerVersionNotFound({
+        path,
+        semverExpressions
+      });
+    }
+
+    if (versionsAndSources[newestSatisfyingVersion]) {
+      versionsAndSources[newestSatisfyingVersion] = {
+        ...versionsAndSources[newestSatisfyingVersion],
+        ...allSources
```
davidmurdoch, line 93: Does `allSources` need to be deduped?

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,123 @@
...
+      compilers: {
+        solc: {
+          version: compilerVersion
+        }
+      }
+    });
+
+    const compilation = await run(
+      versionsAndSources[compilerVersion],
+      normalizeOptions(compilationOptions)
```
davidmurdoch, line 112: I think the original `options` was already normalized, and `compilationOptions` is tightly controlled by us. I'm thinking we don't need to normalize again here.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,142 @@
...
+          path.join(sourceDirectory, "noImports", "OtherSourceWith0.7.0.sol")
+        )
+      });
+      assert.equal(compilations.length, 3);
+    });
+  });
+
+  describe("solidity files with imports", () => {
+    before(() => {
+      paths = [path.join(sourceDirectory, "withImports", "C.sol")];
```
davidmurdoch, line 113: Looks like `paths` is not being used within this test. You can remove this `before` or just update the tests below to use it.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,142 @@
...
+    });
+
+    it("finds a version that satisfies all pragmas if it exists", async () => {
```
davidmurdoch, line 116: I'd like to see a more complex set of pragma combinations; multiple paths both with and without imports, in a single test.

p.s., `fastcheck`'s property-based testing may be an excellent addition here; we could generate random tests combining different pragmas.

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,123 @@
...
+};
+
+const getSemverExpressions = sources => {
+  return sources.map(source => getSemverExpression(source));
+};
+
+// takes an array of versions and an array of semver expressions
+// returns a version of the compiler or undefined if none can be found
+const findNewestSatisfyingVersion = ({ solcReleases, semverExpressions }) => {
+  // releases are ordered from newest to oldest
```
davidmurdoch, line 20: I wonder if this ordering will be guaranteed? I don't think we should risk it. The list may even be alphabetically ordered, in which case `0.10.0` will come before `0.2.0`.

Maybe `compilerSupplier` itself should sort the list? (maybe it does already?)

davidmurdoch (changes requested):
Great work on such an awesome feature!!! I've got some feedback and some questions, as well as requests for changes (and some more tests).

And a general question: what happens if the pragma version, or the entire pragma itself, is invalid? Will it error out in a different way than it used to?

packages/compile-solidity/scripts/test.sh
```diff
@@ -3,8 +3,8 @@
 set -o errexit
 
 if [ "$CI" = true ]; then
-  mocha ./test/** ./test/**/* --timeout 10000 $@
+  mocha ./test/** ./test/**/* --timeout 50000 $@
```
davidmurdoch, line 6: After making some of these changes, let's see if we can get by with removing this timeout increase.

packages/compile-solidity/index.js
```diff
@@ -74,6 +46,10 @@ const Compile = {
 
   // this takes an array of paths and options
   async sourcesWithDependencies({ paths, options }) {
+    if (options.compilers.solc.version === "analyzePragmas") {
+      return await this.sourcesWithPragmaAnalysis({ paths, options });
```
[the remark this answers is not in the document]
eggplantzzz: Yeah, I realize that using the `await` is unnecessary (I didn't know it pushed it to the end of the queue). I kind of wrestle with the idea of intentionally being explicit about the fact that it is an asynchronous function. I didn't really know that it makes a real difference in performance. I don't really feel strongly about it so I'll be happy to omit this in the future.

packages/compile-common/src/profiler/profiler.ts
```diff
@@ -69,4 +68,24 @@ export class Profiler {
       allPaths
     });
   }
+
+  async requiredSourcesForSingleFile(options: any) {
```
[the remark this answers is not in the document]
eggplantzzz: Yeah, that sounds like a good idea. I think this will always be a `@truffle/config` object.

packages/compile-solidity/index.js
```diff
@@ -74,6 +46,10 @@ const Compile = {
 
   // this takes an array of paths and options
   async sourcesWithDependencies({ paths, options }) {
+    if (options.compilers.solc.version === "analyzePragmas") {
```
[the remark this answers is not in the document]
eggplantzzz: It will, but that should never be the case unless it is explicitly set that way. The `Config` object has defaults for that value [here](https://github.com/trufflesuite/truffle/blob/develop/packages/config/src/configDefaults.ts#L46-L57).

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: I don't think so. I should be an object with the filenames as keys so I don't see how there would be duplication.

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: These options have not been normalized yet in this flow and it seems like the main point of this normalization is backwards compatibility? However we are guaranteed that it will be a `Config` object so you are right that it is probably unnecessary! I'll go ahead and remove this. There is another call to this method in `index.js` that can probably also be removed.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
eggplantzzz: Good catch!

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: This is the order that we get from requesting the list of releases and I'm not sure it is guaranteed. So I think it is probably a good idea to sort it somewhere.

packages/compile-solidity/index.js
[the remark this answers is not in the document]
davidmurdoch: I'd say that marking this function as `async` is explicit enough here (although doing that _also_ sends the return of the promise to end of the queue*). That said, I'm actually wrong*; applying `async` to a function that returns a promise is what bumps it to the back of the promise queue, `return await` is just redundant here and has no effect on performance (though the `async` keyword does).

I _kinda_ do like the `async` keyword here, so I'm torn as to what we should do here, so I'll side with what I think gnidan would prefer... which is to keep the keyword (and `await`, since it doesn't matter). You can mark this and the other related review comments as resolved if you agree. :-)

* Why I think I was wrong:

```javascript
async function fn(r){
  return Promise.resolve(r);
}
async function a(){
  return await fn("a");
}
async function b(){
  return fn("b");
}

// which ever is called first wins
await Promise.race([a(), b()]) // "a"
await Promise.race([b(), a()]) // "b"
```

packages/compile-solidity/compileWithPragmaAnalysis.js
davidmurdoch: :+1: Looks like `options.with` normalizes everything already.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
eggplantzzz: Yeah for now I just wanted to cover the basic scenarios. I'll consider using `fastcheck`! Should we implement it this pass or wait for round 2?

packages/compile-solidity/index.js
gnidan, line 49: What do you think of shortening this to just `version: "pragma"`?

packages/compile-solidity/test/compileWithPragmaAnalysis.js
gnidan: I think there's a lot of value in getting `fast-check` in place here right away. This logic feels kind of unpredictable to me, so having thorough tests seems essential.

gnidan (commented):
What's the current behavior - does this minimize number of compilations? Like if I have a contract that needs `0.7.1` and a contract that needs `^0.7.0`, will that be grouped into the same compilation?

eggplantzzz:
> What's the current behavior - does this minimize number of compilations? Like if I have a contract that needs `0.7.1` and a contract that needs `^0.7.0`, will that be grouped into the same compilation?

No, currently it takes a source and its dependencies and finds the newest version of the Solidity compiler that satisfies the pragmas. It then throws the source and its deps into an object labeled with that version. It does this with each source so they get grouped by Solidity version. It is not any smarter than that.

eggplantzzz:
So it sounds like this will probably wait at least one more week before it makes it into a release.

packages/compile-solidity/index.js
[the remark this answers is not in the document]
eggplantzzz: All I did so far was get rid of the superfluous `await`s. Is that good?

eggplantzzz:
> And a general question: what happens if the pragma version, or the entire pragma itself, is invalid? Will it error out in a different way than it used to?

Hmm that's a good point, it might have different behavior than presently exists depending on how it is invalid. The semver expressions are parsed out and it will fail on that if it is something like `prama solidity...`.

If the version range itself is bad then it will probably fail while trying to use the `semver` package. It would probably be good to first check that all are valid to be able to provide better error messaging.

packages/compile-solidity/index.js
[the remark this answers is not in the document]
davidmurdoch: I think so

packages/compile-common/src/profiler/profiler.ts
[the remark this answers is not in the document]
gnidan: Considering this more, I think I'm in favor of leaving this as `any` and converting them all at once, separately.

packages/compile-solidity/index.js
eggplantzzz: I have no big problem with it, what about you @davidmurdoch?

packages/compile-common/src/profiler/profiler.ts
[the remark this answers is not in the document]
eggplantzzz: Well, I already switched it. Do you want to switch it back? haha

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -75,6 +86,11 @@ const compileWithPragmaAnalysis = async ({ paths, options }) => {
     const semverExpressions = await getSemverExpressions(
       Object.values(allSources)
     );
+
+    // this really just validates the expressions from the contracts' imports
+    // as it has already determined the parser version for each contract
+    await validateSemverExpressions(semverExpressions);
```
davidmurdoch, line 92: no need to await this function call.

packages/compile-solidity/index.js
eggplantzzz: updated

haltman-at:
So there's a ton here and I don't expect I'll be able to make a proper review of this, but I wanted to point out something that wasn't changed and I think should be, namely, `workflow-compile/reports.js`.

Currently, it works by going through each compilation, and for each one, looking up which compiler (solc, vyper, or external) was used, and then using that to look up in a table the version used.  But with this PR, it will be possible to have compilations with *different* versions of solc used.  So that code won't work anymore -- I'm not sure what the effect will be, but it seems to me it will end up incorrectly reporting that all compilations used the same version of solc, won't it?  When instead it should show all the different versions of solc used.

Thought: It might also be a good idea to make it so that any given version is only printed once.  This would also fix a minor annoyance that #3583 will introduce for really old Vyper versions.

davidmurdoch (approved):

haltman-at:
OK, ignore my comment, seems @eggplantzzz tested it and it prints multiple versions just fine!

haltman-at:
Btw it turns out that the *reason* my comment was wrong was because that's not actually where it's printed!  (See #3590.)  The real code, as it happens, already does things the right way. :)

PR #3418, opened 2020-10-06
Internal improvement: Separate multiple compilations in testing & factor doing so

haltman-at:
(Possibly a pre-emptive bug fix although nothing's been reported!)

I noticed that, since the compile changes, `test.js` also does the thing that `debug/compiler.js` had been altered to do -- namely, to squash multiple different compilations together, rather than keeping them separate.  I'd already fixed this in `debug`, but now I wanted to do it in `test` as well.

In order to do this though I started by making a change to `Codec`.  The function `shimArtifacts` always returned a `Compilation[]` rather than a `Compilation`, and so to do what I wanted with it in `debug` was a little awkward.

So, I factored it.  Now `shimArtifacts` is just a wrapper around `shimContracts`, which returns a `Compilation`; and for actual use, I added another wrapper around it, `shimCompilations`, which directly shims the new compiler output format, so we don't have to do any more work there.

I then changed what I did in `debug/compiler.js` and `debug/external.js` to now use this new wrapper.  And finally I did what I'd originally set out to do -- use this new wrapper in `test.js` as well.

One thing worth noting though is that the squashed-together compilation was also being used for a second purpose in `test.js`, namely, for getting all the source files to pass to `defineSolidityTests`.  So, I made a separate thing where we concatenate together all of the lists of source files, separate from where we shim the compilations.

And that's it.  And then also `prettier` made a few changes...

### DIFF
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -4,16 +4,54 @@ const debug = debugModule("codec:compilations:utils");
 import * as Ast from "@truffle/codec/ast";
 import * as Compiler from "@truffle/codec/compiler";
 import { ContractObject as Artifact } from "@truffle/contract-schema/spec";
-import { CompiledContract } from "@truffle/compile-common";
+import {
+  CompiledContract,
+  Compilation as CompilerCompilation
+} from "@truffle/compile-common";
 import { Compilation, Contract, Source } from "./types";
 
+export function shimCompilations(
+  inputCompilations: CompilerCompilation[],
+  shimmedCompilationIdPrefix = "shimmedcompilation",
+  externalSolidity = false
+): Compilation[] {
+  return inputCompilations.map(
+    ({ contracts, sourceIndexes }, compilationIndex) =>
+      shimContracts(
+        contracts,
+        sourceIndexes,
+        `${shimmedCompilationIdPrefix}Number(${compilationIndex})`,
+        externalSolidity
+      )
+  );
+}
+
+/**
+ * wrapper around shimArtifactsToCompilation that just returns
+ * the result in a one-element array (keeping the old name
+ * shimArtifacts for compatibility)
+ */
 export function shimArtifacts(
   artifacts: (Artifact | CompiledContract)[],
   files?: string[],
-  shimmedCompilationId = "shimmedcompilation",
-  externalSolidity = false
+  shimmedCompilationId: string = "shimmedcompilation",
+  externalSolidity: boolean = false
 ): Compilation[] {
-  //note: always returns a one-element array (a single fictional compilation)
+  return [
+    shimContracts(artifacts, files, shimmedCompilationId, externalSolidity)
+  ];
+}
+
+/**
+ * shims a bunch of contracts ("artifacts", though not necessarily)
+ * to a compilation.  usually used via one of the above two functions.
+ */
+export function shimContracts(
+  artifacts: (Artifact | CompiledContract)[],
+  files?: string[],
+  shimmedCompilationId: string = "shimmedcompilation",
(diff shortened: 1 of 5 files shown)


packages/core/lib/test.js
```diff
@@ -116,27 +116,14 @@ const Test = {
...
-      (a, compilation) => {
-        if (compilation.compiler && compilation.compiler.name === "solc") {
-          a.sourceIndexes = a.sourceIndexes.concat(compilation.sourceIndexes);
-          a.contracts = a.contracts.concat(compilation.contracts);
-        }
-        return a;
-      },
-      { sourceIndexes: [], contracts: [] }
+    const sourcePaths = [].concat(
+      ...compilations.map(compilation => compilation.sourceIndexes) //we don't need the indices here, just the paths
```
gnidan, line 120: Hm, are we definitely saying, then, that this `sourceIndexes` field explicitly excludes internal sources?

Can you add a comment in the corresponding @truffle/compile-common types definition, to that effect? We'll probably need to add a new `internalSourceIndexes` property...

packages/codec/lib/compilations/utils.ts
```diff
@@ -4,16 +4,54 @@ const debug = debugModule("codec:compilations:utils");
...
+      )
+  );
+}
+
+/**
+ * wrapper around shimArtifactsToCompilation that just returns
+ * the result in a one-element array (keeping the old name
+ * shimArtifacts for compatibility)
+ */
 export function shimArtifacts(
```
gnidan, line 34: Do we still use this anywhere?

gnidan: Oh right Teams might use it. @kevinbluer @seesemichaelj are you using this function? Can you switch to using the new one here, `shimCompilations`?

seesemichaelj: Yup, it's used multiple times here: https://github.com/trufflesuite/teams/blob/develop/packages/api/src/schema/debugger/resolvers.ts#L93-L117

I don't see a reason why we can't switch unless the functionality is completely changing. I'll leave this one up to @kevinbluer

gnidan: Thanks for the response @seesemichaelj! Looking at usage, it does seem to be fine - you already have multiple compilations and you're compressing them just for `shimArtifacts`. This would remove the extra step.

packages/core/lib/test.js
haltman-at: It never occurred to me that it might include internal sources, but I assume it has to exclude them, as including them would take extra work that I have to assume was not done (nor has their inclusion caused any problems thus far, which it presumably would have).

I don't see why we need an `internalSourceIndexes` property.  Internal source indices are handled in the internal source objects.

I can add that comment, though.

Thought: One potential problem that might come up with internal sources is that the user sources might not be consecutive.  This could potentially cause problems in `SolidityTest.js` if it results in a call to `resolve(undefined)`.  Do you know if that causes an error?  If you don't know that it doesn't, I'll go add a guard around that to be sure.

packages/codec/lib/compilations/utils.ts
haltman-at: The debugger CLI uses it as well when actually running off of artifacts.  I've changed it to use `shimCompilations` only in the context where it's using compiler output, not artifacts.

packages/core/lib/test.js
gnidan: It seems like it might cause an error - adding that guard seems appropriate

haltman-at: I decided to add a filter instead, heh.  Same effect.

gnidan (approved):

PR #3421, opened 2020-10-07
Convert contract-schema to initial TS version

gnidan:
- Use ttypescript to use absolute paths for imports
- Move index.js to lib/index.ts
- Add dist to gitignore
- Leave tests in JS, update require() to refer only to package root
- Remove now-extra typings module
- Clean-up code style where appropriate


gnidan:
Hm, actually, going to go a different direction with this.

PR #3422, opened 2020-10-08
Define new package @truffle/abi-utils

gnidan:
See new package [README](https://github.com/trufflesuite/truffle/blob/abi-utils/packages/abi-utils/README.md) for summary of what this package includes.

- Move types + normalization from @truffle/codec
- Rename `schemaAbiToAbi` to `normalize`
- Rename types to exclude extranenous `Abi` in identifiers
- Add `EventParameter` to remove that polymorphism from `Parameter`
- Expand normalization to remove `payable` and `constant` properties in
  favor of `stateMutability`
- Remove `payable` and `constant` from relevant entry types
- Require `stateMutability` in relevant entry types
- Define fast-check arbitraries for generating random loose ABIs (taken from [gnidan/abi-to-sol](https://github.com/gnidan/abi-to-sol), with permission ;)

**Note**: This is a breaking change to @truffle/codec.


gnidan:
cc @eshaben hopefully this will be useful for what you're working on!

gnidan:
This won't be going out in this week's release, to make sure there's time for diligent reviews.

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+    )
+    .chain(([constructor, fallback, receive, entries]) =>
+      fc.shuffledSubarray([constructor, fallback, receive, ...entries])
+    );
+
+namespace Numerics {
+  // 0 < n <= 32
+  export const Bytes = () => fc.nat(31).map(k => 32 - k);
+
+  // o < n <= 256
```
haltman-at, line 178: You have an `o` instead of a `0`. :P  Also you might want to specify `0 < n <= 256, 8 | n`? :)

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+      ReceiveEntry(),
+      fc.array(fc.oneof(FunctionEntry(), EventEntry()))
+    )
+    .chain(([constructor, fallback, receive, entries]) =>
+      fc.shuffledSubarray([constructor, fallback, receive, ...entries])
+    );
+
+namespace Numerics {
+  // 0 < n <= 32
+  export const Bytes = () => fc.nat(31).map(k => 32 - k);
```
haltman-at, line 176: I think the `+1` trick that you use below would be a clearer expression of this than this subtraction trick.  Then it'd just be `fc.nat(31).map(k => k + 1)`.

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  // o < n <= 256
+  export const Bits = () => Bytes().map(k => 8 * k);
+
+  // 0 < n < 80
```
haltman-at, line 181: This should be `0 < n <= 80` (matching the code you wrote below).

Well, OK -- I say 0<n<=80, but in fact the docs aren't fully consistent as to whether 0 is allowed or not, and when I asked the Solidity people about it, they said that's not fully resolved yet.  Anyway I think you're making the right call in excluding it for now.

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+const ParameterName = () =>
+  fc.frequency(
+    { arbitrary: fakerToArb("{{hacker.noun}}"), weight: 9 },
+    { arbitrary: fc.constant(""), weight: 1 }
+  );
+const EventName = () =>
+  fakerToArb("{{hacker.verb}} {{hacker.noun}}", pascalCase);
+const FunctionName = () => fakerToArb("{{hacker.verb}} {{hacker.noun}}");
+
+const TypeRecord = (): fc.Arbitrary<any> =>
```
haltman-at, line 295: I am so confused here.  How can you put a type (other than the unit type) on *no* arguments??  Is this type signature supposed to go on `TypeRecord` itself or something?

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+
+export const Parameter = () =>
+  fc
+    .tuple(
+      fc.record({
+        name: ParameterName()
+      }),
+      TypeRecord()
+    )
+    .map(records => records.reduce((a, b) => ({ ...a, ...b }), {}));
```
haltman-at, line 15: Kind of overkill for combining two things rather than an indefinite number, don't you think?

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+export const EventParameter = () =>
+  fc
+    .tuple(
+      fc.record({
+        name: ParameterName(),
+        indexed: fc.boolean()
+      }),
+      TypeRecord()
+    )
+    .map(records => records.reduce((a, b) => ({ ...a, ...b }), {}));
```
haltman-at, line 26: Same comment as above.

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+    name: EventName(),
+    inputs: fc.array(EventParameter(), { maxLength: 10 }).filter(inputs => {
+      if (inputs.filter(({ indexed }) => indexed).length > 3) {
+        // only up to 3 params can be indexed
+        return false;
+      }
+
+      // names that are not blank should be unique
+      const names = inputs.map(({ name }) => name).filter(name => name !== "");
+      return names.length === new Set(names).size;
```
haltman-at, line 40: Hooray for finite sets :)

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+        {
+          outputs: fc.array(Parameter(), { maxLength: 10 }).filter(outputs => {
+            // names that are not blank should be unique
+            const names = outputs
+              .map(({ name }) => name)
+              .filter(name => name !== "");
+            return names.length === new Set(names).size;
+          })
+        },
+        { withDeletedKeys: true }
```
haltman-at, line 74: I'm pretty sure `outputs` is not optional, and you don't have it as optional in your TS type; why is this here?

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+        )
+    )
+    .map(records => records.reduce((a, b) => ({ ...a, ...b }), {}))
+    .filter(entry => {
+      const { inputs, outputs = [] } = entry as Types.FunctionEntry;
+      // names that are not blank should be unique
+      const names = [...inputs, ...outputs]
+        .map(({ name }) => name)
+        .filter(name => name !== "");
+      return names.length === new Set(names).size;
```
haltman-at, line 104: Question: Given that you perform this uniqueness check at the end, does it make sense to also include the two intermediate ones?  (Does including them speed things up significantly?)

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+            fc.constant("view"),
+            fc.constant("nonpayable"),
+            fc.constant("payable")
+          ),
+          fc.boolean()
+        )
+        .map(([stateMutability, includeOptionals]) =>
+          !includeOptionals
+            ? { stateMutability }
+            : {
```
haltman-at, line 89: I notice you don't include any way of generating ABIs with `stateMutability` missing (in favor of just `payable`/`constant`, as best I can tell.  (This applies to fallbacks and constructors as well, obviously.)  Should that be added, for testing normalization?

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+        type: fc.constant("fallback")
+      }),
+      fc
+        .tuple(
+          fc.oneof(fc.constant("nonpayable"), fc.constant("payable")),
+          fc.boolean()
+        )
+        .map(([stateMutability, includeOptionals]) =>
+          !includeOptionals
+            ? { stateMutability }
```
haltman-at, line 126: Same comment about stateMutability but recording it here again for good measure.

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+      fc.record({
+        type: fc.constant("constructor"),
+        inputs: fc.array(Parameter())
+      }),
+      fc
+        .tuple(
+          fc.oneof(fc.constant("nonpayable"), fc.constant("payable")),
+          fc.boolean()
+        )
+        .map(([stateMutability, includeOptionals]) =>
```
haltman-at, line 149: And, the mutability thing again.

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+  "switch",
+  "typedef",
+  "typeof",
+  "unchecked"
+]);
+
+const fakerToArb = (template: string, transform = camelCase) => {
+  return fc
+    .integer()
+    .noBias()
```
haltman-at, line 277: This function needs some comments.  Why `.noBias`?  Why `.noShrink`?  (Checking the documentation doesn't even give me a good idea what shrinking is.)  What is `faker`?  This all needs more explanation.

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+    // Primitives.Function(), // TODO re-add
+    Primitives.Bytes(),
+    Primitives.String(),
+    Primitives.Tuple()
+  );
+
+const Type: fc.Memo<string> = fc.memo(n =>
+  n === 0
+    ? Primitive()
+    : fc.oneof(Primitive(), ArrayFixed(n > 3 ? 3 : n), ArrayDynamic(n))
```
haltman-at, line 221: The fact that `ArrayFixed` has its dimension capped warrants at least a comment explaining why.

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+);
+
+const ArrayFixed = fc.memo(n => {
+  const tuple =
+    n <= 1
```
haltman-at, line 226: Isn't this redundant?  Won't this be handled automatically by the recursive call to `Type()`?

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+      ? fc.tuple(Primitive(), fc.integer(1, 256))
+      : fc.tuple(Type(), fc.integer(1, 256));
+
+  return tuple.map(([type, length]) => `${type}[${length}]`);
+});
+
+const ArrayDynamic = fc.memo(n => {
+  const item = n <= 1 ? Primitive() : Type();
```
haltman-at, line 234: Same comment about the redundant condition.

packages/abi-utils/lib/arbitrary.test.ts
```diff
@@ -0,0 +1,55 @@
+import { testProp } from "jest-fast-check";
+import * as abiSchema from "@truffle/contract-schema/spec/abi.spec.json";
+import { matchers } from "jest-json-schema";
+
+expect.extend(matchers);
+
+import * as Arbitrary from "./arbitrary";
+
+const withDefinitions = (schema: object) => ({
```
haltman-at, line 9: This function probably warrants a comment at least (I figured out what it does, but it could still use explanation).

haltman-at (commented):
So, uh, I had a lot of comments on this one.

packages/abi-utils/lib/arbitrary.ts
gnidan: The reason I didn't do this is because fast-check maintains the concept of what is a simple value vs. what is a complex value (that's what that "shrink" business relates to). So I did `32 - k` because I want 32 to be the "simple" value (since 0 is the "simple" value from `fc.nat(31)`)

haltman-at: Oh, huh.  In that case is a similar subtraction warranted for the decimal places?  80 is the default there, after all, not 1.

haltman-at: Also this likely warrants a comment.

packages/abi-utils/lib/normalize.ts
```diff
@@ -0,0 +1,47 @@
...
+
+//NOTE: SchemaAbi is kind of loose and a pain to use.
+//So we'll generally coerce things to Abi before use.
+//(we combine this with adding a "function" tag for
+//entries that lack it)
+
+export const normalize = (looseAbi: SchemaAbi | Abi): Abi =>
+  (looseAbi as SchemaAbi).map(normalizeEntry);
+
+type Item<A> = A extends (infer I)[] ? I : never;
```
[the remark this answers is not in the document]
gnidan: I'm worried about adding this to that package because those types are autogenerated and I don't want to break them accidentally

packages/codec/lib/abi-data/utils.ts
```diff
@@ -56,43 +41,23 @@ export function abiHasPayableFallback(
...
-  if (abiEntry.type === "function" && abiEntry.constant) {
-    return "view";
-  }
-  return "nonpayable";
-}
-
 //NOTE: this function returns the written out SIGNATURE, not the SELECTOR
 export function abiSignature(
-  abiEntry: Abi.FunctionAbiEntry | Abi.EventAbiEntry
+  abiEntry: Abi.FunctionEntry | Abi.EventEntry
```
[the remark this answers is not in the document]
gnidan: I wasn't planning on it, but I could. What do you think?

packages/abi-utils/lib/normalize.ts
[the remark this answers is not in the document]
haltman-at: To be clear, what I mean by "adding that type" is factoring that type out in the schema, giving it its own definition there, not manually editing the Typescript or something.

packages/codec/lib/abi-data/utils.ts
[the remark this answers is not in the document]
haltman-at: I mean it makes sense to me!  Maybe not some of the more specialized ones, IDK.

packages/abi-utils/lib/normalize.ts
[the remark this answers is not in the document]
gnidan: Just tried it - it doesn't work. `json2ts` seems to only respect things that are defined at the top-level; it doesn't emit separate types for individual definitions.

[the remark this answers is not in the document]
haltman-at: Yes, by "add that type" what I'm saying is you should factor out that type as a top-level definition, like we have for so many other types!

packages/abi-utils/lib/arbitrary.test.ts
```diff
@@ -0,0 +1,55 @@
+import { testProp } from "jest-fast-check";
```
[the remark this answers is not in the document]
gnidan: This seems to be the new way of doing things, esp. with Jest. I'm kind of keen to move in this direction; are you opposed? I figure, since this is a new package, it's reasonably safe enough to break convention to try it out here.

packages/abi-utils/lib/normalize.ts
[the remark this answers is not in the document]
gnidan: That's what I did - it doesn't work. Notice how you can't do `import { NormalFunction } from "@truffle/contract-schema/spec"`. json2ts just doesn't create dedicated types for individual definitions.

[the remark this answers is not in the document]
haltman-at: I'm confused.  You can import `ImmutableReferences`, as an example.  What's different about this?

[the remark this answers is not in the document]
gnidan: no idea, but I tried it and it doesn't show up.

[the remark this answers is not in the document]
gnidan: can we call that out of scope for this PR?

packages/abi-utils/lib/arbitrary.ts
haltman-at: Oops you are correct, I misremembered, sorry.  18 is the default.  They sound too similar in English I guess. :P  Um, IDK, do what you want with this I guess?

packages/abi-utils/lib/arbitrary.ts
haltman-at: Nevermind, I think I misunderstood here.

packages/abi-utils/lib/normalize.ts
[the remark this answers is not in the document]
haltman-at: Eh, I guess?  I think it would be good to do (it's something I'd been thinking about doing myself for some time, because it's always been an inconvenience) but if it doesn't work for some reason then I guess it's not essential.

packages/abi-utils/lib/arbitrary.ts
gnidan: Our ABI schema says it's optional... I'm not sure, maybe it used to be?

gnidan: Also note: the TS types are normalized, so they wouldn't have optionality. Maybe the thing to do here is make sure that `normalize` includes `outputs`

packages/abi-utils/lib/arbitrary.ts
gnidan: it probably doesn't make things too much faster, I will change that

packages/abi-utils/lib/arbitrary.ts
gnidan: well, `Type` differentiates only on `n === 0`, so this covers the case where `n === 1`...

I dunno, this was kind of tricky to get working, due to fast-check performance problems and general confusery

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,315 @@
...
+
+  return item.map(type => `${type}[]`);
+});
+
+const reservedWords = new Set([
```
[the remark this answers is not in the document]
haltman-at: OK, here's my big list of things that should probably not be used as names, in addition to the ones already included (don't know if all of these are a problem but I'm erring on the side of safety).  Note there is some overlap due to Solidity not updating their list.

They are: `var`, `bool`, `string`, `int` and `uint` [and other int and uint types, if relevant], `byte`, `bytes` [and other bytesN types, if relevant], `fixed` and `ufixed` [and other fixed and ufixed types, if relevant], `enum`, `struct`, `mapping`, `address`, `new`, `delete`, `if`, `else`, `for`, `while`, `continue`, `break`, `return`, `throw`, `emit`, `try`, `catch`, `_` [maybe?], `function`, `modifier`, `event`, `constructor`, `fallback`, `receive`, `virtual`, `override`, `constant`, `immutable`, `anonymous`, `indexed`, `storage`, `memory`, `calldata`, `external`, `public`, `internal`, `payable`, `pure`, `view`, `private`, `returns`, `import`, `from`, `as`, `using`, `pragma`, `contract`, `interface`, `library`, `is`, `abstract`, `assembly`, `true`, `false`, `wei`, `gwei`, `szabo`, `finney`, `ether`, `seconds`, `minutes`, `hours`, `days`, `weeks`, `years`, `this`, `super`, `selfdestruct`, `suicide`, `now`, `msg`, `block`, `tx`, `abi`, `type`, `blockhash`, `gasleft`, `assert`, `revert`, `require`, `Error`, `sha3`, `sha256`, `keccak256`, `ripemd160`, `ecrecover`, `addmod`, `mulmod`, and `log0` through `log4`.

Don't know how relevant all of these are, obviously the parameterized ones would be troublesome to include, but yeah...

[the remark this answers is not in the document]
gnidan: Want to format that as a suggestion, so you can get commit credit?
Screen Shot 2020-10-09 at 18 37 15

packages/abi-utils/lib/normalize.ts
```diff
@@ -0,0 +1,47 @@
+import { Abi as SchemaAbi } from "@truffle/contract-schema/spec";
+import { Abi, Entry, StateMutability } from "./types";
+
+//NOTE: SchemaAbi is kind of loose and a pain to use.
```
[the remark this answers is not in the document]
gnidan: Maybe I'll just delete the comment? I think it's covered by the readme...

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -0,0 +1,326 @@
...
+  "relocatable",
+  "sealed",
+  "sizeof",
+  "static",
+  "supports",
+  "switch",
+  "typedef",
+  "typeof",
+  "unchecked"
+]);
```
gnidan, line 282: @haltman-at you can add a suggestion in reply to this comment, for your revised list

packages/abi-utils/lib/normalize.ts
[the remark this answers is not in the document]
gnidan: https://github.com/trufflesuite/truffle/issues/3431

packages/codec/lib/abi-data/utils.ts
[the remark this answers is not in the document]
gnidan: Hm. Let's leave it out of scope for now; it'd require updating the README and such, and this PR is already growing a bit long.

I can open another issue for that if you like :)

packages/abi-utils/lib/arbitrary.ts
haltman-at: Hm.  Should we change the ABI schema to have it mandatory?

packages/abi-utils/lib/arbitrary.ts
haltman-at: OK but like, the recursive call to `Type()` will automatically decrement `n`, right, and then the `n === 0` check will get tripped, right?  I think it's conceptually clearer to only check the condition once.

packages/abi-utils/lib/arbitrary.ts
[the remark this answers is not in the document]
haltman-at: It looks like you just replaced your list with mine, rather than merging them?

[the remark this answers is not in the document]
gnidan: Ah, misinterpreted your message! Will fix

packages/abi-utils/lib/arbitrary.ts
gnidan: Ah, apparently not: Solidity v0.5.0 [specifies](https://solidity.readthedocs.io/en/v0.5.0/abi-spec.html#json) that `outputs` "can be omitted if function doesn’t return anything"

So seems like this should get added to `normalize`

gnidan: (which I've now done)

packages/abi-utils/lib/arbitrary.ts
gnidan: Alright, got rid of this and it seems to work fine

packages/abi-utils/lib/arbitrary.ts
haltman-at: Oh wow!  I didn't realize that.  Thanks for correcting that!

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -231,124 +231,147 @@ const Type: fc.Memo<string> = fc.memo(n =>
...
-      : fc.tuple(Type(), fc.integer(1, 256));
-
-  return tuple.map(([type, length]) => `${type}[${length}]`);
-});
-
-const ArrayDynamic = fc.memo(n => {
-  const item = n <= 1 ? Primitive() : Type();
+const ArrayFixed = fc.memo(n =>
+  fc
+    .tuple(Type(n - 1), fc.integer(1, 256))
```
haltman-at, line 236: Note, my understanding is that the explicit `n-1` is unnecessary, but I haven't tried it and don't quite understand how this works, so IDK.  Anyway being more explicit isn't a bad thing, so, whatever.

haltman-at (approved):
OK!  Hitting approve on this, but I do think others (@eggplantzzz, @fainashalts, @cds-amal) should possibly weigh in on the question of where to put test files.

packages/abi-utils/lib/arbitrary.ts
gnidan: Yeah I decided to be more explicit because it seemed appropriate here.

gnidan: Cause otherwise `n` would be unused and such, and then someone might ask "why do you need `fc.memo`?"... and yeah

gnidan:
Let's plan to discuss test locations at the next strategy meeting.

As for now, I'm merging this! Thanks for helping me through this @haltman-at!

PR #3428, opened 2020-10-09
Db/remove artifacts loader subschemas

fainashalts:
This PR removes unnecessary subschemas (`artifacts` and `loaders`) from the db package. This is the first step toward making the `workspace` schema actually be the root schema for the db. The loader mutation test has been removed since it no longer exists, and the `loaders/schema/test` tests have been moved up into `loaders/test`.


packages/db/src/loaders/test/index.ts
```diff
@@ -55,7 +55,8 @@ afterAll(() => {
   tempDir.removeCallback();
 });
 
-it("loads artifacts and returns true ", async () => {
+//skipping because loadersSchema subschema no longer exists
+it.skip("loads artifacts and returns true ", async () => {
```
gnidan, line 59: You can just delete this test I think

gnidan (approved):

PR #3429, opened 2020-10-09
Require EventParameters include indexed

gnidan:
- Update codec/ast/utils to define parametersToAbi as generic


haltman-at (approved):
I guess this makes sense?

gnidan:
Cool. I'll merge and let the upstream CI deal with it

PR #3430, opened 2020-10-09
Db/load on compile

fainashalts:
This PR runs `db.loadCompilations` when WorkflowCompile.compile provides compilation output. Loading should only happen when a user runs the `truffle compile` command (should not happen upon `truffle test` command). 

To test, go to any Truffle project and add the following to your truffle config file:

```
db {
  enabled: true
}
```

Then run `truffle compile` from the project. An `ls -la` should show that the `.db` directory has been created. From Truffle, navigate to the `db` package and run `yarn start <path-to-project>`. You should be able to access a graphQL playground wherein you can query information about the project. A simple sample query: 

```
query GetContractNames {
	workspace {
    projects {
      id
    }
    contractNames
    networks {
    	name
    }
	}
}
```

The schema is accessible from the playground, which can be used to formulate further queries. 

Of note, this PR checks whether a compiled contract's bytecode is missing, and does not add the bytecode or contract in that case. This means we do not add abstract contracts or interfaces to Truffle DB's store of bytecodes/contracts for a project.


packages/workflow-compile/index.js
```diff
@@ -59,7 +61,7 @@ async function compile(config) {
 }
 
 const WorkflowCompile = {
-  async compile(options) {
+  async compile(options, loadDb = false) {
```
gnidan, line 64: We should avoid adding any arguments to this function - it should be based only on `config.db.enabled`

packages/workflow-compile/index.js
```diff
@@ -82,6 +84,17 @@ const WorkflowCompile = {
...
+
+    if (
+      config.db &&
+      config.db.enabled === true &&
+      contracts.length > 0 &&
+      loadDb === true
+    ) {
+      const db = new TruffleDB(config);
+      await db.loadCompilations({ contracts, compilations });
+    }
```
gnidan, line 96: This logic should get moved to `save`, since that's distinct from compilation.

packages/db/src/loaders/resources/contracts/index.ts
```diff
@@ -23,29 +23,31 @@ export function* generateContractsLoad(
...
 > {
-  const contracts = loadableContracts.map(loadableContract => {
-    const {
-      contract: { contractName: name, abi: abiObject },
-      path: { sourceIndex, contractIndex },
-      bytecodes,
-      compilation
-    } = loadableContract;
+  const contracts = loadableContracts
+    .filter(({ contract }) => contract.bytecode.bytes !== "")
```
gnidan, line 27: probably worth a comment here about why this gets filtered out

gnidan (changes requested):
This excludes the work to conditionally load names, right? Since that was added in `move-name-logic`?

I'm just asking - happy to call that part out of scope for this PR, but let's not lose track of it as a task to do

Requesting changes here just to avoid changing the workflow-compile interface.

packages/workflow-compile/index.js
fainashalts: I added this because we wanted to distinguish between actual compilation and compilation from test. So I pass the true flag from the compile command. Was there a better way to do that? `config.db.enabled` only tells us that the user has enabled using the db, but doesn't tell us whether they're running `truffle test`.

packages/db/src/loaders/resources/contracts/index.ts
fainashalts: added!

packages/workflow-compile/index.js
fainashalts: moved

packages/workflow-compile/index.js
fainashalts: I reworked this but don't love it. There's config.loadDb that I added in the compile command file, and then there's { names: true} being passed to `loadCompilations`. I added the names flag in the name logic PR, but looking at it now I feel like couldn't we drop it? Because if we are running `truffle test` we won't get to `loadCompilations` anyway. What do you think?

packages/workflow-compile/index.js
```diff
@@ -14,6 +14,8 @@ const SUPPORTED_COMPILERS = {
   external: require("@truffle/external-compile").Compile
 };
 
+const { TruffleDB } = require("@truffle/db");
```
gnidan, line 17: Did this get added to workflow-compile/package.json? We'll need to make sure it's there

(And, noticing the webpack.config.js change, we'll need to add it as an optionalDependency to truffle, if it's not already there)

gnidan (commented):
Looks good apart from verifying package.json stuff.

I haven't checked, but if those package.json changes are still outstanding, it might be worth adding a Truffle scenario test to make sure it's all hooked up (do `truffle compile` and test that the DB has stuff)

gnidan (approved):
it's all coming together!

Issue #3271, opened 2020-08-16
Truffle 5.x compiler option 'contracts_build_directory'


barakman:
## Issue

The `contracts_build_directory` compiler option appears to not being handled properly on Windows.

Running `truffle test` always ends up converting my contracts build directory (where Truffle creates all JSON artifacts) from `'./<my_project_folder>/build/contracts'` (which is also the default BTW) to some temporary Windows directory, for example, `'C:\\Users\\LOCALC~1\\AppData\\Local\\Temp\\test--12372-H5iUC1JGbek1'` (the suffix changes on every execution).

This, in turn, causes a full rebuild upon every execution (i.e., `truffle compile --all` is always executed automatically before the actual test begins).

I have added some printouts in your code (file `cli.bundled.js`), and found out that the problem may be stemming from when function `prepareConfigAndRunTests` is called with this temporary directory.

I believe that the problem might be here:
```
      Environment.detect(config)
        .then(() => copyArtifactsToTempDir(config))
        .then(({ config, temporaryDirectory }) => {
          return prepareConfigAndRunTests({
            config,
            files,
            temporaryDirectory
          });
        })
```
And since I know that this problem does not occur on neither Linux nor MacOS, I believe that it possibly stems from `Environment.detect`.

What I don't get is, when printing all Truffle configurable directories (`contracts_directory`, `build_directory`, `test_directory`, `contracts_build_directory`, etc) in your code, why the only one which is replaced with a temporary directory is `contracts_build_directory`.

Any help would be very much appreciated, as I am currently left to wait on compilation before every test.

Thanks

## Environment

* Operating System: Windows 10
* Truffle version (`truffle version`): 5.1.36
* node version (`node --version`): 10.16.0
* npm version (`npm --version`): 6.9.0

gnidan:
Hm I just started noticing that behavior, also! Thanks for reporting this issue, we'll look into it promptly!

gnidan:
(For the record, I'm able to reproduce this on macOS - I do `truffle compile` followed by another `truffle compile`; all contracts compile both times)

elenadimitrova:
Confirming this behaviour on macOS as well. https://github.com/trufflesuite/truffle/issues/3277 could be related.

gnidan:
Thanks for the added context @elenadimitrova, we will look into this!

barakman:
@gnidan : Any progress on this, or an ETA on a fix?

fainashalts:
[...]

PR #3438, opened 2020-10-15
Bug fix: Fix compilation during testing

eggplantzzz:
Compile contracts before creating and moving the built files to a temporary directory during testing.

Currently Truffle copies over the built artifacts to a temporary directory while testing. It then examines them to see whether they are in sync with the Solidity source files; if not, it compiles and saves them to the temporary directory. If you continuously run test, it will always compile the same files over and over again since they are never saved to the project's build directory and instead get deleted after the test run.

This PR checks to see if a compilation is necessary BEFORE copying the files to the temporary directory to ensure that the compilation gets saved properly. Thus on the second run, the files will be updated and it will no longer be necessary to compile those artifacts again. Tests written in Solidity, however, still need to be compiled each time.

Fixes https://github.com/trufflesuite/truffle/issues/3271

### DIFF
--- a/packages/core/lib/commands/test/compileContractsIfNecessary.js
+++ b/packages/core/lib/commands/test/compileContractsIfNecessary.js
@@ -17,7 +17,6 @@ const compileContractsIfNecessary = async config => {
     quiet: config.runnerOutputOnly || config.quiet
   });
 
-  // Compile project contracts and test contracts
   await WorkflowCompile.compileAndSave(compileConfig);
 };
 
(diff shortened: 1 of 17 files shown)


eggplantzzz:
Eventually we can change `compileContractsWithTestFilesIfNecessary` to just `compileSolidityTestFiles` but I didn't want to break the interface since this is exposed on `truffle.test.compileContractsWithTestFilesIfNecessary`.

davidmurdoch:
I have some thoughts on this.

Since `truffle test` doesn't currently have any side effects, this behavior might considered part of the public interface. We might be able to get the performance speed up AND keep the old behavior by always compiling to the same temporary location for each project. Bonus points if `truffle compile` itself uses this directory to speed things up if `truffle test` had previously been run.

Let me know if that makes sense or not, or if the concern just isn't valid.

eggplantzzz:
It was decided to leave the current behavior as it is so as not to mutate the user's artifacts. If they wish to stop constant compilation of their artifacts, they will have to make sure their artifacts are in sync with the built ones by compiling outside of testing. Either that or they can pass the `--compile-none` flag.

PR #3450, opened 2020-10-19
Move core resource type definitions to @truffle/db/meta

gnidan:
This stuff lived in a pouch-specific module, but it's more widely useful than that, so this PR moves this boilerplate logic into a more cohesive place, along with some additional changes along the way.

In particular:

- Define types for filtering resources based on mutability or namedness (`MutableResource`, etc.)
- Include in this new module the `IdObject` and `NamedResource` types from loaders
- Redefine each resource's input type as the actual resource input, not the input to the mutation.
- Define additional types for mutation input and mutation payload
- Tighten up the base collection type to use less `any`
- Make actual workspace Databases implementation more typesafe

### DIFF
--- a/packages/db/src/workspace/definitions.ts
+++ b/packages/db/src/workspace/definitions.ts
@@ -1,6 +1,6 @@
-import { Definitions } from "./pouch";
+import * as Pouch from "./pouch";
 
-export type WorkspaceCollections = {
+export type Collections = {
   bytecodes: {
     resource: DataModel.IBytecode;
     input: DataModel.IBytecodesAddInput;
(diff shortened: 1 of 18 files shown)


packages/db/src/workspace/definitions.ts
```diff
@@ -1,37 +1,40 @@
-import { Definitions } from "./pouch";
+import * as Meta from "@truffle/db/meta";
+import * as Pouch from "./pouch";
 
-export type WorkspaceCollections = {
+export type Collections = {
```
fainashalts, line 4: I'm a little iffy on the renaming here, because the `meta` file also has a `Collections` type.  I guess `WorkspaceCollections` won't work anymore once it is the root schema, but wondering if there is a better naming option here. Or am I misunderstanding how this works?

packages/db/src/workspace/pouch/databases.ts
```diff
@@ -209,11 +204,7 @@ export abstract class Databases<C extends Collections> {
         const id = this.generateId(collectionName, resourceInput);
 
         const resource = await this.get(collectionName, id);
-        const {
-          _rev
-        }: {
-          _rev: PouchDB.Core.RevisionId;
-        } = resource ? resource : {};
+        const { _rev = undefined } = resource ? resource : {};
```
fainashalts, line 207: Nice. I hadn't noticed that {} obviously wouldn't have a `_rev` value. Surprised we haven't seen an error due to this.

fainashalts (approved):
LGTM!

packages/db/src/workspace/definitions.ts
gnidan: I'm not so worried about the name collision? Maybe that's just my style. Nothing will ever import both - if necessary, I'm thinking we do `import * as Meta from "@truffle/db/meta"`

packages/db/src/workspace/pouch/databases.ts
gnidan: we didn't need it because the types were looser. now they're tighter, so it's complaining now.

gnidan:
Thanks for the review!

PR #3451, opened 2020-10-19
Remove distinct workspace concept from Truffle DB

gnidan:
- Remove `workspace { }` as a wrapper around all queries / mutations

- Split up workspace logic into:
  - `src/pouch/` for pouch-related stuff
  - `src/graphql/` for graphql-related stuff
  - `src/definitions/` for individual resource definitions

- Switch to housing resource-specific typeDefs and resolvers alongside the definitions of those resources

- Autogenerate the actual Truffle DB GraphQL schema based on these definitions

- Upgrade all the various graphql packages and stop using my outdated forks

- Do a bunch of code cleanup, including removing duplicate input types in favor of new `input ResourceReferenceInput { id: ID! }`


packages/db/src/pouch/index.ts
```diff
@@ -0,0 +1,72 @@
...
+import { DatabasesOptions } from "./databases";
+
+import { FSDatabases } from "./fs";
+import { MemoryDatabases } from "./memory";
+import { SqliteDatabases } from "./sqlite";
+
+import { Collections } from "@truffle/db/meta";
+import { Databases, Definitions } from "./types";
+
+export interface WorkspaceConfig {
```
fainashalts, line 14: Do we still want to call this `WorkspaceConfig`?

packages/db/src/pouch/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  adapter?: {
+    name: string;
+    settings?: any; // to allow adapters to define any options type
+  };
+}
+
+export const forDefinitions = <C extends Collections>(
+  definitions: Definitions<C>
+) => (config: WorkspaceConfig): Databases<C> => {
+  const { WorkspaceDatabases, settings } = concretize<C>(config);
```
fainashalts, line 25: Ditto `WorkspaceDatabases`

packages/db/src/schema.ts
```diff
@@ -1,26 +1,4 @@
...
-            return "Contract";
-          }
-        }
-      }
-    }
-  });
-}
-
-export const schema = readSchema();
+export const schema = forDefinitions(definitions);
```
fainashalts, line 4: Isn't this the exact same pattern from `db/src/connect`? Perhaps could be combined into one file somehow?

packages/db/src/test/nameRecord.graphql.ts
```diff
@@ -31,26 +31,11 @@ export const GetAllNameRecords = gql`
...
-    resource: ResourceInput!
-    previous: PreviousNameRecordInput
-  }
-
   mutation AddNameRecord(
     $name: String!
     $type: String!
-    $resource: ResourceInput!
-    $previous: PreviousNameRecordInput
+    $resource: ResourceReferenceInput!
```
fainashalts, line 37: Maybe I'm just missing it but where is this type imported?

packages/db/src/test/utils.ts
```diff
@@ -21,10 +22,10 @@ const tempDir = tmp.dirSync({ unsafeCleanup: true });
 tmp.setGracefulCleanup();
 
 export class WorkspaceClient {
-  private workspace: Workspace;
+  private workspace: Databases;
```
fainashalts, line 25: In keeping with removing workspace, should we rename this? Fine either way of course since it doesn't affect the root goal of removing the subschema.

fainashalts (approved):
I left a couple of comments but nothing blocking merge. LGTM!

packages/db/src/test/utils.ts
gnidan: Ah perhaps. I'm not totally happy with `Databases` as a name, so I liked the idea of keeping `workspace` as the context variable. Do you prefer calling it `databases`?

packages/db/src/test/nameRecord.graphql.ts
gnidan: No need to import these references, since queries get all the identifiers defined by the schema. So the schema defines `ResourceReferenceInput`, thus making it available to anyone who wants to query Truffle DB, without the need to define something like that themselves. You'll notice that's how I got away with deleting all those various `input` definitions everywhere.

packages/db/src/schema.ts
gnidan: It's the same pattern, but it comes from two different places. I prefer having them be separate modules - `connect` gets you the pouch stuff, `schema` gets you the graphql stuff

packages/db/src/pouch/index.ts
gnidan: Yeah, I don't like the name, but I couldn't think of anything better. Got any ideas?

packages/db/src/pouch/index.ts
gnidan: Nope, probably not! I'll change this; shouldn't be hard to come up with a new name.

packages/db/src/schema.ts
fainashalts: That makes sense!

packages/db/src/test/utils.ts
fainashalts: Hm. I mean what is it really that we're working with here? I suppose `databases` is accurate. `rootDatabases`? I'm terrible at coming up with names. It isn't a big deal, I guess I just thought someone new to the code would be like "what is a workspace in this context?"

packages/db/src/pouch/index.ts
gnidan: I just called it `constructor` :)

packages/db/src/test/utils.ts
gnidan: Hm, I'm torn. Your argument is compelling and you may have won me over, but let me ask you what you prefer, for writing resolvers:

```typescript
resolve: ({ id }, _, { databases }) => 
  databases.get("sources", id)
```

```typescript
resolve: ({ id }, _, { workspace }) => 
  workspace.get("sources", id)
```

gnidan: How bout this alternative: we rename `interface Databases` to `interface Workspace`? We can keep all the classes having the same name; we'd just do `class Databases<C extends Collections> implements Workspace<C extends Collections>`

gnidan: Alright I went the other direction like I proposed. Let me know what you think!

fainashalts: Oops sorry I missed your comments! `interface Workspace` seems fine!

fainashalts: And this way there's no need to change the resolvers, right?

gnidan: Exactly!

PR #3454, opened 2020-10-20
Replace module-alias with ttypescript in Truffle DB

gnidan:
Currently, @truffle/db is set up to require that users import `@truffle/db` by itself first, which alters Node's runtime dependency resolution to know about `@truffle/db/*` paths.

This is really gross, and we've now been favoring use of `ttypescript`, which allows us to update all generated (`dist/`) `require` statements to use the `paths` specified by the tsconfig. This is how we do things in @truffle/codec already.

In addition, this PR sets up [madge](https://github.com/pahen/madge), because madge is cool. In order to get this to work, I had to move `./configs/base.json` to `tsconfig.base.json`.

<details>
<summary>Preview module dependencies</summary>

</details>

You can now recreate this locally with `yarn madge --image deps.pdf` (and `yarn madge --circular` to verify we haven't introduced circular dependencies)

### DIFF
--- a/packages/db/.madgerc
+++ b/packages/db/.madgerc
@@ -1,0 +1,5 @@
+{
+  "excludeRegExp": ["\\.\\.", "test"],
+  "fileExtensions": ["ts"],
+  "tsConfig": "./tsconfig.base.json"
+}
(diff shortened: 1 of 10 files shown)


fainashalts (approved):
Nice! 🎉

gnidan:
Thanks for the review!

PR #3457, opened 2020-10-21
Bug fix: Separate out internal sources by context and fix assembly breakpoints

haltman-at:
**Note: This PR is a breaking change to `@truffle/debugger`.**  (Edit: And now **also to `@truffle/debug-utils`**.  And maybe also `@truffle/codec`?)

So, it turns out that generated sources don't have unique source indices per compilation like other sources do.

Instead, they have unique source indices per *compilation and context*.  "Context", here, recall, basically means bytecode -- i.e., a contract normally gets two contexts, one for its constructor bytecode and one for its runtime bytecode.  So, uh, that's a lot of potential collisions.  I, um, didn't realize things worked this way, and @chriseth confirms that this is not a bug, so, uh, that means it's time to reorganize a whole bunch of things inside the debugger in order to allow this to work.

And not just the debugger!  Let's go package-by-package describing the changes.  (Debugger will come last.)

## codec

The `shimContracts` method needs to change how it handles generated sources.  Previously, it inserted these into the list of sources it would output.  But now it can't do that, so, well, it no longer does.  The list of sources you get back from it will now only include user sources, not generated sources.  Instead, the `Contract` objects that it outputs now have a `generatedSources` field and a `deployedGeneratedSources` field, which contain what you would expect (`shimContracts` doesn't process them at all).  (As you can see, the type has been modified appropriately.)

**Note**: I could change this to have generated sources more resemble other sources.  Opinions on this, @gnidan?

However, `shimContracts` *does* check for any collisions between generated source IDs and user source IDs, and still sets the `unreliableSourceOrder` flag if any are found.  So there's that.

Arguably this could be called a breaking change to `codec`, but seeing as the previous way wasn't exactly functional, I'm not sure that's warranted.

## decoder

Surprisingly, `decoder` ended up not needing any changes!  Generated sources, being written in Yul, don't contain anything `decoder` needs to know about.  Contract declarations?  Nope.  Struct or enum declarations?  Nope.  Functions that could be pointed to by an internal function pointer?  Nope.  I originally thought I might have to make some changes here to prevent crashes (or errors resulting from caught crashes), but it seems the relevant error cases are already handled.  So, no changes here!

## solidity-utils

Surprisingly, this also ended up not needing any changes.  I did make one change, however, where I noticed some overcomplicated code that could be simplified.  But that's not a functional change.  The error case I thought would be relevant to `decoder` (attempting to look up a source ID beyond the end of the array of sources) turns out to be already handled.

## debugger, core, and debug-utils

And, here we get to the bulk of the changes.  Unlike `decoder`, `debugger` can't just ignore generated sources.  This means a bunch of reorganization is needed to incorporate them (or rather, the way they actually work, not the way I thought they worked).

Let's break this down submodule-by-submodule.

### trace and evm

These submodules, unsurprisingly, had no changes.  Obviously neither did `web3`.  The `stacktrace` submodule also had no changes, because it just piggybacks off of `solidity`, so I won't bother listing it separately later.

### solidity

So, the big change here is to the `sources` state.  Originally sources were simply stored by ID (index).  Once multi-compilation support was introduced, they were stored by compilation and ID.  Now... it's a bit more complicated.

It works like this.  You have the sources state, `sources`.  It's grouped by compilation, so you have `sources.byCompilation[compilationId]`.  This now has two fields: `userSources` and `internalSources`.  User sources are grouped by ID (index), so ultimately you'd get one as `sources.byCompilation[compilationId].userSources.byId[index]`.  Internal sources, however, are grouped by *context* and then index, so ultimatel you'd get one as `sources.byCompilation[compilationId].internalSources.byContext[contextHash].byId[index]`.  Yes, this is a pain in the ass, but I don't think there's any getting around it.  (Note that this is quite definitely a breaking change; the debugger CLI uses this selector a fair bit.)

In addition, the source objects themselves have gained a new addition.  There's still the `internal` flag indicating whether a source is generated, but now, when it's set to true, there'll be a new field, `internalFor`, indicating to what context the source corresponds.  This will be `undefined` for user sources.

With all that in mind, the corresponding selectors have changed a bit.  Let's take a look at them.

Naturally, the `solidity.views.findOverlappingRange` had to be changed to be grouped similarly.  So yeah.  I also renamed it to the less confusing `solidity.views.overlapFunctions`.

This also meant that `solidity.current.sources` had to be changed; now, instead of simply picking out the correct list of sources based on compilation, it has to use the compilation to get the user sources, the compilation and context to get the internal sources, and then combine these two together.  For convenience, I also added `solidity.current.overlapFunctions`, which does the same thing, but with the overlap functions.  So `solidity.current.findOverlappingRange` now depends just on `solidity.current.overlapFunctions`, rather than directly on `solidity.views.overlapFunctions` (or `solidity.views.findOverlappingRange`, as it was called before).

Also, for convenience, I added a `solidity.views.flattenedSources` selector, which gives the sources all flattened into a single array (each still has the information on where it came from within itself, of course, what with `compilationId` and `id` and `internal` and `internalFor`).  You'll see this again shortly.

That's kind of a pain, but it's not the worst thing we'll see.  But now, we move on to...

### session

This submodule basically just needed to be modified to accomodate the changes to `codec` and `solidity`.  Mostly what's different is the `normalize` function.  There are several differences here, but basically, since `shimContracts` no longer processes generated sources, that's now up to `normalize`.  And since the format of `sources` is now different, well, it now has to set up that different state.  Note btw how now `normalize` computes the context hashes, so that it can mark internal sources with the appropriate `internalFor`, rather than leaving that up to `evm`.

In addition, the `affectedInstances` selector was changed to accommodate the changed `solidity.info.sources`.

But OK, that was pretty straightforward.  Now we get to the real mess, which is...

### data and ast

Hoo boy.  Once again, as with `solidity`, the problem is how to present the debugger with a unified view of `scopes` showing only what it currently cares about (because showing it everything at once would require a bunch of rewriting and would make the code really verbose), while also having everything stored there in the background.

So, just like the `sources` state, the `scopes` state had to change.  The change is exactly the same (except that we now have `userNodes` vs `internalNodes` rather than `userSources` vs `internalSources`, and `byAstRef` rather than `byId`), so I won't detail that any further.

There are now two problems: 1. How do we set up the `scopes` state, and 2. how do we make use of it?

Setting up the `scopes` state is largely the job of the `ast` saga.  So, `visitAll` now uses `solidity.views.flattenedSources` to get all the sources, and now passes in `internalFor` to the `walk` function (and like every other function).  The `data.scope` and `data.declare` sagas don't take this input (because generated sources are never Solidity), but their Yul equivalents do.

Then of course the `scopes` reducer has to account for all this, which is tedious but straightforward, so I don't really need to describe that.

Now we get to the question of actually reading this state -- i.e., the data selectors.  This is where it truly gets messy.  These have been substantially reorganized.

So, to recall: Previously we had `data.info.scopes` (the raw scopes state), `data.views.scopes.raw` (reformatted a tiny bit for convenience), `data.views.scopes.inlined.raw` (adds inlined definitions), `data.views.scopes` (without inlined definitions but with inheritance processing and similar), `data.views.scopes.inlined` (inlined definitions *and* inheritance processing), and finally `data.current.scopes` and `data.current.scopes.inlined` (which differ from the `views` version in that they only contain what's relevant to the current compilation, indexed purely by `astRef`).

Maintaining all this the new way is basically just a big pain.  So basically what I did was that I kept the `views` versions of things where necessary, but mostly moved processing to `current`, where it's easier.

So in `info` and `views` we now have `data.info.scopes` (of course); `data.views.scopes` (corresponds to the old `data.views.scopes.raw`, as all inheritance processing is done in `current` now); `data.views.scopes.inlined` (corresponds to the old `data.views.scopes.inlined.raw`; I wasn't going to include this but it turns out we need it).  Meanwhile in `current` we have `data.current.scopes.raw`, `data.current.scopes`, `data.current.scopes.inlined.raw`, and `data.current.scopes.inlined`.  You can infer from above what these do.  Like I said, inheritance processing was moved here because it's just easier when you don't need to deal with all the extra crap.  Also like I said, I was going to move inlining here as well, but, well, turns out we do in fact need this in `views`, so, yeah.

But, whew, with all that done, `data` is saved.  So let's get to the last part...

### controller, core, debug-utils, and the debugger tests

Core and debug-utils are of course separate packages rather than submodules, and obviously the debugger tests are again another matter, but I'm grouping these all together.

You know what else adding more non-uniqueness breaks?  That's right, breakpoints!  Now, breakpoints set in internal sources including a `context` field with the context hash, because, well, there isn't enough information otherwise.  Note that if it's *not* in an internal source, this field should be `undefined`.

So, now everything involving breakpoints has to check or set this field.  I'm not going to list all the changes here, it's mostly straightforward.  However, an important point worth noting here is that I decided not to allow setting breakpoints in internal sources with `<source>:<line>` syntax; only user sources will be searched when setting those.  Since internal sources are right now all named `"#utility.yul"`, well... yeah.  But I could go back and change this, note.

Also note that right now, when printed out, breakpoints in internal sources will just have their source listed as `"an internal source"` rather than using the name of said internal source, rather than using the name of said internal source, which right now is always `"#utility.yul"` anyway, but I could change this if you think it's warranted.  (The reason it works like this is... mostly due to a mistake I made which I didn't think it was worth it to go back and change right now... though I certainly could...)

But breakpoint changes aren't the only changes to `core`; there's also the problem of colorization.  Specifically, storing the colorized sources.  Because sources can no longer just be uniquely identified by compilation ID and index, the colorized sources have been rearranged to be stored in a format similar to the `sources` state.  It's annoying, but straightforward.

Also, the debugger tests have in one place been changed to make use of `solidity.views.flattenedSources` rather than `solidity.info.sources`.

Anyway I think that sums it up!  Hopefully we'll never need to add yet *another* disambiguator of sources... @_@


haltman-at:
Oops, looks like I forgot to also fix up the assignments system!  So, let's talk about...

### data (again)

I forgot I had to also redo the assignments system for `data`, but now I have and it works fine.  In the process, I decided to also make a number of simplifications; the assignments system is frankly a mess due to having largely been written by me back when I had no clue what I was doing.

So, what's changed?

Well, first off, assignments now contain an `internalFor` field.  This field will be `null` if it's not from an internal source.

See, I decided that including or excluding fields conditionally like I was doing everywhere was just too complicated.  So, the `internalFor` field appears on all assignments, not just ones from internal sources; it'll just be `null` if it's from an internal source.

But, why stop there?  While I was at it, I changed things so that now, all assignments for local variables will include the `modifierDepth` field.  Just, if it's not a variable from a modifier, that field will be `null`.  That's a lot simpler than having to conditionally exclude it.  So, that's a nice simplification right there.

But, hell, we can simplify more, can't we?  In `data.current.identifiers.refs`, why are we futzing around with this long-hand way of looking up contract variables?  Let's just look them up by ID.  Same with builtins; why are we doing it this other way?  Do everything by ID!  So, that notoriously messy selector is now a whole lot simpler.

And why stop there -- let's get rid of the part where contract variables include the address in the ID.  Yes, we might want that eventually, there's a reason I added it, but the fact is that reason *still* has yet to come up, so to hell with it, let's just tear that out for now, we can add it back in should we ever finally get around to that stuff.

But the biggest simplification of all is to the `assignments` state itself -- most of the state is gone!  There's now just `byId`.  What with the simplification of `data.current.identifiers.refs`, there's simply no need for all this state where we store things by compilation, astRef, etc.  Just do everything by ID.

...it occurs to me now that I likely could have done *other* things in this PR this way too, had I thought of it... oh well.  It works, no need to go back and change it and come up with some new ID system instead of doing things the straightforward-but-more-tedious way.

haltman-at:
After talking to @gnidan, I've changed generated sources in `Codec` to use the `Source` type, as well as adding a `language` field to `Source`.  However, we still mostly don't respect the `language` field; I intend to push another commit to make it so that we make more use of it.  (Right now we basically just assume that user sources are Solidity and generated sources are Yul.)

haltman-at:
OK, I've now done that too.  Concretely this means:

1. In `core/lib/debug/printer`, we now rely on the `language` field to determine how to do source highlighting, rather than going based on whether it's an internal or user source.  Arguably this should be moved in side `debug-utils`, with a `language` argument rather than a `yul` flag?  Oh well.

2. In `data` and `ast`, we now use the `internalFor` variable even when dealing with Solidity, as we no longer assume that generated sources can't be Solidity.  However, for the sake of simplicity, I continued to assume that generated sources will never define new types (structs, enums, contracts, interfaces).  I really don't want to have to go back and make even more things account for the possibility of internal sources...

(3. Also, you may have noticed in the previous commit I changed two checks in `codec` and `decoder` so that instead of checking whether `ast.nodeType === "SourceUnit"`, they now check whether `language === "Solidity"`.)

haltman-at:
Yeah, why the hell not, I moved the switch on the `language` field to debug-utils for cleanliness.  **Note that this makes this PR a breaking change to debug-utils**.

haltman-at:
One more thing before review -- I've removed the `internal` flag from the `Source` type in Codec.  It was redundant, not really an intrinsic property of the source, and nothing used it (the debugger got the info the other way, i.e., where does the source appear).  So it's gone now.

packages/codec/lib/compilations/types.ts
```diff
@@ -135,4 +135,14 @@ export interface Contract {
    * The ID of the contract's primary source.
    */
   primarySourceId?: string;
+  /**
+   * The contract's generated sources object as output by Solidity 0.7.2 or later.
+   * Note that this will be a sparse array.
+   */
+  generatedSources?: Source[];
```
gnidan, line 142: Nice!

packages/core/lib/debug/printer.js
```diff
@@ -138,12 +159,14 @@ class DebugPrinter {
...
-    //missing the soure text
-    const colorizedSource = this.colorizedSources[compilationId][sourceId];
+    const { source, internal, internalFor } = this.session.view(
+      solidity.current.sources
+    )[sourceId];
+    //we don't just get extract the source text from the location because passed-in location may be
+    //missing the source text
+    const colorizedSource = internal
+      ? this.colorizedSources[compilationId].internal[internalFor][sourceId]
+      : this.colorizedSources[compilationId].user[sourceId];
```
gnidan, line 169: Is there a way to factor this so we can avoid distinguishing here about internalness?

packages/debug-utils/index.js
```diff
@@ -372,10 +373,12 @@ var DebugUtils = {
     }
     if (
       breakpoint.compilationId !== currentCompilationId ||
+      breakpoint.context !== currentInternalContext ||
```
gnidan, line 376: It kind of strikes me as a code smell that we expose the context here like this. Is there a way to provide an interface to the debugger to get the source more opaquely? I guess this goes back to your question of whether or not we should ensure source ID uniqueness... there might be value in doing that.

packages/debugger/lib/ast/sagas/index.js
```diff
@@ -7,31 +7,72 @@ import * as data from "lib/data/sagas";
 
 import ast from "../selectors";
 
-import flatten from "lodash.flatten";
 import jsonpointer from "json-pointer";
 
-function* walk(compilationId, sourceId, node, pointer = "", parentId = null) {
+function* walk(
+  compilationId,
+  internalFor,
```
gnidan, line 14: I think this identifier could be more clear - presumably it refers to a context?

packages/debugger/lib/solidity/selectors/index.js
```diff
@@ -361,17 +410,57 @@ let solidity = createSelectorTree({
...
+        )
+      ).filter(source => source)
+    ),
+
+    /**
+     * solidity.views.overlapFunctions
+     * grouped by compilation [then by context for internal ones]
      */
-    findOverlappingRange: createLeaf(["/info/sources"], compilations =>
+    overlapFunctions: createLeaf(["/info/sources"], compilations =>
```
gnidan, line 438: This name's a bit weird - what is an "overlap function"?

packages/debugger/lib/data/sagas/index.js
```diff
@@ -1007,14 +1043,13 @@ function literalAssignments(
...
-      ? {
-          compilationId,
-          astRef: node.id,
-          stackframe: currentDepth,
-          modifierDepth
-        }
-      : { compilationId, astRef: node.id, stackframe: currentDepth },
+    {
+      compilationId,
+      internalFor,
```
gnidan, line 1048: This pattern seems like it's getting out of hand. Is there a way we could make this kind of referencing more cohesive? Can we define some type that houses all relevant fields for such a lookup?

gnidan (commented):
Generally I think this is alright, but it's kind of increasing the mess in the same way in a number of places. I'm wondering if we can discuss options to clean this up before it goes out (if it's feasible to do that sort of thing)

haltman-at:
OK, at @gnidan's suggestion, I went back and simplified the structure here a fair bit!

In order to do this, we had to finally bite the bullet and give sources actual IDs, and relabel source indices to, y'know, indices, not IDs.  This actually turned out to be not that bad.  Note that, for various reasons, we actually ignore the passed in ID and assign our own, based on the compilationId, context (if internal), and source index; these are treated as totally opaque though, it's not like the variable IDs.

So, now, the `id` property of a source contains its ID; if you want its index, that's under `index`.  (Note the selector `data.current.sourceId` was renamed to `data.current.index`.)

The `sources` state under `solidity` has been reorganized again, but this time for the better.  It has three fields: `byId`, which contains the sources by ID; `byCompilationId`, which contains the user source IDs (just the IDs, to keep things normalized!), first by compilation ID and then by index (under `byIndex`, obviously); and `byContext`, which is similar but for internal sources, first by context and then by index.

With this, we can simplify things a lot:

1. Just about everything that's indexed by source is stored by source ID now; no more compilation-and-index, let alone compilation-and-context-and-index.  (Compilation *and* context was redundant anyway, context is sufficient.)  All you need is the source ID.
2. Previously, various places in the code had to perform the logic of combining user sources and internal sources.  Now this is done in one place: The new selector `solidity.current.sourceIds`, which gives you an array of the currently active source IDs (the indices into the array are the source indices).

Basically, so many places in the code that were previously passing around `compilationId` and `internalFor`, it's now all by source ID.  The `scopes` state in data is by source ID.  Etc.  I also replaced `solidity.views.flattenedSources` with `solidity.views.sources`, which is just the sources by their IDs.  I'm not going to list it all, but basically, it's been quite cleaned up!

In particular, breakpoints have been cleaned up; no more `compilationId` field in those (let alone `context`), the `sourceId` field can now just contain the source ID rather than the index and that's sufficient.  I'm considering this as a non-breaking change to breakpoints, since after all you'd presumably get the ID out of `source.id` anyway, which now contains the new source ID.  (Of course, this is all still a breaking change to `debugger` generally.)

Also, while I was at it, I fixed Yul breakpoints; the `node` field (I didn't rename it, for compatibility) can now contain an astRef instead of just a node ID, and it'll work fine.

Also, I got rid of the restriction I added earlier in this PR against putting breakpoints in internal sources from outside said source.  The real reason I did this after all was due to allowing it being messy, but, well, it's not messy anymore!

Now, while it is quite a bit cleaner, I couldn't clean up *everything* (not practically, anyway).  There are still a few warts.  Let's list those:

1. We still assume in several places that no types will be defined in generated sources, nor will internal function pointers ever point there.  Changing that seemed like too much work for something that's simply not happening at the moment.  Would likely require changes to Codec.
2. As a particular case of the above, because it's used by Codec, which expects it to be a certain way, `data.views.userDefinedTypes` is still organized by compilation ID rather than source ID.  Again, I didn't want to go messing with that stuff in Codec right now.
3. The `primarySource` field in a context object still has the source index, not the ID.  Changing that just seemed like too much trouble.  We only use it in one place, namely, computing affected instances, so I just had that selector use `solidity.info.sources` directly even though I avoid that elsewhere.  Although, due to the cleanup, even using `solidity.info.sources` directly isn't that bad anymore.
4. The biggest wart: Source index redundancy.  A bunch of things are by `sourceId` and `astRef`... but that means we're kind of using the source index twice.  Ideally it would factor into one of those, not both.  However, changing this just seemed like too much work.

Anyway yeah, here's a much better version than what you previously reviewed! :)

...now I need to see about your other comments...

packages/core/lib/debug/printer.js
haltman-at: There sure is :)

packages/debug-utils/index.js
haltman-at: Done! :)

packages/debugger/lib/ast/sagas/index.js
haltman-at: Well, it's gone now; `compilationId` and `internalFor` have been replaced by the lone `sourceId`. :)

packages/debugger/lib/data/sagas/index.js
haltman-at: Handled now by `sourceId`!

packages/debugger/lib/solidity/selectors/index.js
haltman-at: It's a function for a given source that takes a source range and returns the AST nodes that overlap it.  Not sure what a better name would be.

packages/debugger/test/data/helpers.js
```diff
@@ -4,7 +4,6 @@ const debug = debugModule("test:data:decode");
 import Ganache from "ganache-core";
 import { assert } from "chai";
 import changeCase from "change-case";
-import flatten from "lodash.flatten";
```
gnidan, line 7: Have we stopped using this now? Does this mean it's no longer necessary as a dependency? (I haven't looked)

gnidan (approved):
Alright @haltman-at, this looks good!

packages/debugger/test/data/helpers.js
haltman-at: No, it's still used elsewhere.  We could get rid of it and just use `[].concat`, but, whatever.  (I also *added* the use of `lodash.merge`, which is harder to replace...)

PR #3448, opened 2020-10-19
Dependency update: Update other web3 deps

eggplantzzz:
Update web3 ancillary packages to 1.2.9

PR #3459, opened 2020-10-22
Dependency update: Bump ancillary web3 packages to 1.2.9

eggplantzzz:
This work used to be in #3448 but was moved to this branch due to problems merging/rebasing that couldn't be resolved.

Update web3 ancillary packages to 1.2.9


haltman-at (approved):
Tests pass and we're holding off on getting rid of `@types/web3`, so this is good I guess!

PR #3460, opened 2020-10-23
Remove I prefix from autogenerated @truffle/db DataModel types

gnidan:
This would result in `interface Resource { id }` as well as `type Resource = Source | ...`, but I figure we don't need those; I'm just ignoring those from the output.

### DIFF
--- a/packages/db/types/stub.d.ts
+++ b/packages/db/types/stub.d.ts
@@ -2,59 +2,47 @@
 // graphql typescript definitions
 
 declare namespace DataModel {
-  type IResourceReferenceInput = any;
-  type IBytecode = any;
-  type IBytecodeInput = any;
-  type IBytecodesAddInput = any;
-  type IBytecodesAddPayload = any;
-  type ICompilation = any;
-  type ICompilationInput = any;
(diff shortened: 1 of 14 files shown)


packages/db/types/stub.d.ts
```diff
@@ -2,59 +2,47 @@
...
-  type IProjectName = any;
-  type IProjectNameInput = any;
-  type IProjectNamesAssignPayload = any;
-  type ISource = any;
-  type ISourceInput = any;
-  type ISourceRange = any;
-  type ISourcesAddInput = any;
-  type ISourcesAddPayload = any;
-
-  interface IWorkspaceQuery {
```
fainashalts, line 47: Did you remove these interfaces on purpose? Just checking!

fainashalts (approved):
Love it! I never liked the I prefix but thought there must have been a reason for it.

packages/db/types/stub.d.ts
gnidan: Yep - they were unused. Probably should've done that separately but meh

PR #3461, opened 2020-10-23
Scenario Test for DB load on compile

fainashalts:


packages/truffle/test/scenarios/compile/compile.js
```diff
@@ -385,3 +385,38 @@ describe("Repeated compilation of contracts with inheritance [ @standalone ]", f
...
+    try {
+      await CommandRunner.run("compile", config);
+    } catch (error) {
+      output = logger.contents();
+      log(output);
+      throw new Error(error);
+    }
+
+    const dbExists = checkForDb(config);
+    assert(dbExists === true);
```
gnidan, line 420: This should also test that a contract made it into the db, no?

fainashalts: Added! I just test that the contract names exist in the db, since I think that getting deeper into the contents of the db should be a task for the db tests, right?

gnidan: Yeah, agreed!

gnidan (approved):
Looks good!

PR #3462, opened 2020-10-23
Replace custom config type in @truffle/db with @truffle/config

gnidan:
... importing as type only, to keep it as a devDep

Note: this causes a problem with our git hooks, so I had to do `--no-verify`. Looks like this is a problem with us being very behind for our `husky` dependency.

### DIFF
--- a/packages/db/src/db.ts
+++ b/packages/db/src/db.ts
@@ -1,4 +1,5 @@
 import { GraphQLSchema, DocumentNode, parse, execute } from "graphql";
+import type TruffleConfig from "@truffle/config";
 import { generateCompileLoad } from "@truffle/db/loaders/commands";
 import { WorkspaceRequest, WorkspaceResponse } from "@truffle/db/loaders/types";
 import { WorkflowCompileResult } from "@truffle/compile-common";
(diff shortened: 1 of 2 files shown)


packages/db/src/db.ts
```diff
@@ -11,18 +12,6 @@ import {
 } from "@truffle/db/loaders/commands";
 import { toIdObject, NamedResource } from "@truffle/db/meta";
 
-interface IConfig {
-  contracts_build_directory: string;
-  contracts_directory: string;
-  working_directory?: string;
-  db?: {
```
fainashalts, line 18: What happens to this db field though? I don't think the `@truffle/config` has this field, are we adding it there? Well, it has `db: { enabled: bool }` so I guess adapter could be added there if needed?

gnidan: Hm, what do you think? Worth making this change? I kind of figured we could let it slide for now.

fainashalts: I like not having our own custom config and this is the only thing that stands out as different. I think it would be nice for the adapter field to be set in the user config if they don't want to use the default. So we can leave this out until we want to add that functionality?

fainashalts (approved):

packages/db/src/db.ts
gnidan: Maybe I'll call this a later task :)

PR #3463, opened 2020-10-23
Split @truffle/db build into multiple yarn scripts

gnidan:
So that it's possible to do _some_ of the steps but not _all_ of the steps

- Remove shell script that does building
- Define separate scripts for building with stubs, autogenerating the schema types, and then building the project for real


fainashalts (approved):
LGTM!

gnidan:
Thanks!

PR #3466, opened 2020-10-23
Unify @truffle/db loaders infrastructure

gnidan:
This PR improves internals for @truffle/db loaders system by collecting existing patterns into first-class citizens. Mostly, this is two-part:

## Move `runLoader` to a module inside `@truffle/db/loaders/run`

Pretty straightforward - this PR defines a new type `LoaderRunner`, and provides a constructor for obtaining such an value given a Truffle DB instance.

## Define a `Load` type for use in loaders

... to serve in place of the existing `Generator</* ... ugly params ... */>`s everywhere.

This type is generic, accepting between 0..2 type parameters as arguments:
```typescript
type Load<T = any, N extends RequestName | string = string>
```

These arguments function as follows:
- `T` represents value ultimately returned by the loader. Think: `const result: T = yield *generateResourceLoad()`
- `N` represents a specific kind of query or mutation defined by the data model. For loaders that only issue one kind of GraphQL request, this can be specified to add additional type safety; e.g. `function* generateSourcesLoad(...): Load<LoadedSources, "sourcesAdd">` doesn't make any kind of request besides the `sourcesAdd` mutation.

To set up the type safety, this PR defines a bunch of helper types that make `Load` possible, to represent the allowable request names as well as the corresponding data that comes back for a given request.

### DIFF
--- a/packages/db/src/loaders/run.ts
+++ b/packages/db/src/loaders/run.ts
@@ -1,0 +1,34 @@
+import { WorkspaceRequest, WorkspaceResponse } from "./types";
+
+export type LoaderRunner = <
+  Request extends WorkspaceRequest,
+  Response extends WorkspaceResponse,
+  Args extends unknown[],
+  Return
+>(
+  loader: (...args: Args) => Generator<Request, Return, Response>,
+  ...args: Args
+) => Promise<Return>;
+
+export const forDb = (db): LoaderRunner => async <
+  Request extends WorkspaceRequest,
+  Response extends WorkspaceResponse,
+  Args extends unknown[],
+  Return
+>(
+  loader: (...args: Args) => Generator<Request, Return, Response>,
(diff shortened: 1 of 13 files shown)


packages/db/src/loaders/run.ts
```diff
@@ -0,0 +1,32 @@
...
+export type LoaderRunner = <
+  A extends unknown[],
+  T = any,
+  N extends RequestName | string = string
+>(
+  loader: Loader<A, T, N>,
+  ...args: A
+) => Promise<T>;
+
+export const forDb = (db): LoaderRunner => async <
```
fainashalts, line 12: Neat!

fainashalts (approved):
LGTM!

gnidan:
Thanks @fainashalts!

Issue #3464, opened 2020-10-23
truffle test on v5.1.50 does not work when using a Websocket provider


dstarke:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

After upgrading to 5.1.50, Truffle does not work when configured to use a Websocket provider. 

## Steps to Reproduce

I have a project that was working correctly in 5.1.49. It is configured to use a Websocket provider. After upgrading to 5.1.50, it does not work. After changing the configuration to not use a Websocket provider, it works.

The failure occurs immediately when running `truffle test`.

## Expected Behavior

I expected the tests to run normally.

## Actual Results
(I've redacted my local path information with `...`)
```> Something went wrong while attempting to connect to the network. Check your network configuration.
ProviderError: 
    at .../node_modules/truffle/build/webpack:/packages/provider/wrapper.js:75:1
    at  .../node_modules/truffle/build/webpack:/packages/provider/wrapper.js:107:1
    at Object.callback ( .../node_modules/truffle/build/webpack:/packages/provider/wrapper.js:112:1)
    at  .../node_modules/truffle/build/webpack:/node_modules/web3/node_modules/web3-providers-ws/src/index.js:133:1
    at Array.forEach (<anonymous>)
    at WebsocketProvider._onMessage  (.../node_modules/truffle/build/webpack:/node_modules/web3/node_modules/web3-providers-ws/src/index.js:118:1)
    at W3CWebSocket._dispatchEvent [as dispatchEvent] ( .../node_modules/truffle/build/webpack:/node_modules/yaeti/lib/EventTarget.js:115:1)
    at W3CWebSocket.onMessage ( .../node_modules/truffle/build/webpack:/node_modules/websocket/lib/W3CWebSocket.js:234:1)
    at WebSocketConnection.<anonymous> ( .../node_modules/truffle/build/webpack:/node_modules/websocket/lib/W3CWebSocket.js:205:1)
    at WebSocketConnection.emit (events.js:315:20)
    at WebSocketConnection.processFrame ( .../node_modules/truffle/build/webpack:/node_modules/websocket/lib/WebSocketConnection.js:554:1)
    at  .../node_modules/truffle/build/webpack:/node_modules/websocket/lib/WebSocketConnection.js:323:34
    at processTicksAndRejections (internal/process/task_queues.js:79:11)
Truffle v5.1.50 (core: 5.1.50)
Node v12.18.4
[...]

PR #3472, opened 2020-10-26
Bug fix: Allow for `false` value in provider callback

eggplantzzz:
On web3 version 1.2.9, a websocket provider returns `false` as a value for `error` as noted in the issue below. On the previous version we used it returned `null`. This causes it to always throw in the post hook [here](https://github.com/trufflesuite/truffle/blob/47686d9dedc3174e3133b89b3da9f4cd1d2eae82/packages/provider/wrapper.js#L73). This allows for `error` to be `false` as well as `null`/`undefined`.

Fixes https://github.com/trufflesuite/truffle/issues/3464

### DIFF
--- a/packages/provider/wrapper.js
+++ b/packages/provider/wrapper.js
@@ -8,7 +8,7 @@ module.exports = {
    * Wraps an underlying web3 provider's RPC transport methods (send/sendAsync)
    * for Truffle-specific purposes, mainly for logging / request verbosity.
    */
-  wrap: function(provider, options) {
+  wrap: function (provider, options) {
     /* wrapping should be idempotent */
     if (provider._alreadyWrapped) return provider;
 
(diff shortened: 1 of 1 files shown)


packages/provider/wrapper.js
```diff
@@ -68,20 +65,18 @@ module.exports = {
...
   // after send/sendAsync
-  postHook: function(options) {
-    return function(payload, error, result) {
-      if (error != null) {
-        // wrap errors in internal error class
+  postHook: function (options) {
+    return function (payload, error, result) {
+      // web3 websocket providers return false and web3 http providers
+      // return null when no error has occurred...kind of obnoxious
+      if (error != null && error !== false) {
```
gnidan, line 72: Why not just `if (!error)`?

gnidan (approved):

packages/provider/wrapper.js
eggplantzzz: That sounds very reasonable.

eggplantzzz:
You know, it seems like we should consider getting rid of this wrapper around the provider. I don't think we currently use the hooks for anything right now and there have been other problems with wrapping things like `web3` in the past. I think at this point it is a net liability.

PR #3474, opened 2020-10-26
Internal improvement: Remove git add from husky steps

haltman-at:
This new version of Husky warns us that we have `git add` in our Husky steps, which apparently is unnecessary (or at least unnecessary now).  Removing it to get rid of the warning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,13 +39,11 @@
   "lint-staged": {
     "*.js": [
       "prettier --write",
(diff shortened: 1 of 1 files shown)


gnidan (approved):
Oh cool

haltman-at:
Merging this despite yellow because Travis is stuck or something.

PR #3475, opened 2020-10-27
Instrument @truffle/db with some initial logging

gnidan:
This PR adds [debug](https://www.npmjs.com/package/debug) in an internal `@truffle/db/logger` module, for use throughout the @truffle/db codebase.

This defines logger instances in all existing TS modules, as well as adds some initial logging in key areas.

### DIFF
--- a/packages/db/src/connect.ts
+++ b/packages/db/src/connect.ts
@@ -1,3 +1,6 @@
+import { logger } from "@truffle/db/logger";
+const debug = logger("db:connect");
+
 import { definitions } from "./definitions";
 import { forDefinitions } from "./pouch";
 
--- a/packages/db/src/db.ts
+++ b/packages/db/src/db.ts
@@ -1,3 +1,6 @@
+import { logger } from "@truffle/db/logger";
+const debug = logger("db:db");
+
 import { GraphQLSchema, DocumentNode, parse, execute } from "graphql";
 import type TruffleConfig from "@truffle/config";
 import { generateCompileLoad } from "@truffle/db/loaders/commands";
--- a/packages/db/src/definitions/bytecodes.ts
+++ b/packages/db/src/definitions/bytecodes.ts
@@ -1,3 +1,6 @@
+import { logger } from "@truffle/db/logger";
+const debug = logger("db:definitions:bytecodes");
+
 import gql from "graphql-tag";
 
 import { Definition } from "./types";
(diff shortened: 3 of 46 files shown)


gnidan:
There's more I want to do here... e.g. I am tempted to group loggers more by resource/collectionName where appropriate, but I figure it's probably more important to get _something_ in place as a start

gnidan:
This is what this looks like when I run `DEBUG=db* yarn test`:

Screen Shot 2020-10-26 at 22 20 20

packages/db/src/connect.ts
```diff
@@ -1,3 +1,6 @@
+import { logger } from "@truffle/db/logger";
+const debug = logger("db:connect");
```
fainashalts, line 2: Are you adding this here for later?

packages/db/src/db.ts
```diff
@@ -1,3 +1,6 @@
+import { logger } from "@truffle/db/logger";
```
fainashalts, line 1: Ditto, imported but not being used.

packages/db/src/definitions/bytecodes.ts
```diff
@@ -1,3 +1,6 @@
+import { logger } from "@truffle/db/logger";
```
fainashalts, line 1: Here too.

packages/db/src/graphql/types.ts
```diff
@@ -1,3 +1,6 @@
+import { logger } from "@truffle/db/logger";
+const debug = logger("db:graphql:types");
```
fainashalts, line 2: imported but not used.

packages/db/src/helpers/generateId.ts
```diff
@@ -1,3 +1,6 @@
+import { logger } from "@truffle/db/logger";
```
fainashalts, line 1: imported but not used.

fainashalts (approved):
This is so great! Will definitely make development easier by letting us see where things go wrong. I wonder if we could add some logging of the actual result somewhere, if the result is an error? Would be more informative, but I supposed while developing when you see where something is breaking you can add a log statement. 

I left a couple of comments about a few places where you imported the package but aren't using it, but since it looks like you did it a lot there was probably a good reason so I stopped adding those comments. :)

packages/db/src/connect.ts
gnidan: Yep - I added it to every module I could, in case we need it. We have a dedicated eslint rule to allow this as an unused variable. This is how we do things in other packages.

gnidan:
Re: logging actual results - that's another thing that I wanted to do but figured it could wait. My concern with logging actual results is that there needs to be a mechanism to say how verbose you want the logs - otherwise the output would just be entirely unwieldy.

PR #3477, opened 2020-10-27
Allow querying @truffle/db for resources given a list of IDs

gnidan:
This PR extends the existing query for all resources of a given type to allow looking up multiple resources at once given a list of IDs.

Query syntax looks something like:
```graphql
query {
  contracts(filter: { ids: ["0x..."] }) {
    id
  }
}
```


fainashalts (approved):
Neat!

gnidan:
Thanks!

PR #3482, opened 2020-10-29
Dependency update: Upgrade @trufflesuite/chromafi version to 2.2.1

haltman-at:
This should prevent `husky` from being installed as a dependency rather than a dev dependency.  Thanks to @interfect for noticing the problem.


gnidan (approved):
nice!

haltman-at:
Merging despite yellow because Travis is screwing up again, and those are allowed failures anyway.

PR #3483, opened 2020-10-29
Define project-based abstraction for loading Truffle stuff to @truffle/db

gnidan:
To bridge the gap between @truffle/db and Truffle at large.

This adds a top-level `const { Project } = require("@truffle/db")` that exposes methods corresponding to existing Truffle workflows.

### DIFF
--- a/packages/db/src/loaders/commands/migrate.ts
+++ b/packages/db/src/loaders/commands/migrate.ts
@@ -1,0 +1,131 @@
+import { logger } from "@truffle/db/logger";
+const debug = logger("db:loaders:commands:migrate");
+
+import { ContractObject } from "@truffle/contract-schema/spec";
+import { toIdObject, IdObject } from "@truffle/db/meta";
+import { Load } from "@truffle/db/loaders/types";
+
+import { generateContractGet } from "@truffle/db/loaders/resources/contracts";
+import {
+  generateTranasctionNetworkLoad,
+  generateNetworkIdFetch,
+  generateNetworkGet
+} from "@truffle/db/loaders/resources/networks";
+import {
+  LoadableContractInstance,
+  generateContractInstancesLoad
+} from "@truffle/db/loaders/resources/contractInstances";
+
+export interface GenerateMigrateLoadOptions {
+  // we'll need everything for this input other than what's calculated here
+  network: Omit<DataModel.NetworkInput, "networkId" | "historicBlock">;
+  artifacts: ContractObject[];
+}
+
+export function* generateMigrateLoad(
+  options: GenerateMigrateLoadOptions
+): Load<{
+  network: IdObject<DataModel.Network>;
+  contractInstances: IdObject<DataModel.ContractInstance>[];
+}> {
+  // for each artifact, first add a Network resource for the historic block
+  // when the contract was deployed
+  //
+  // one of these networks will be the latest, so that's what we'll return
+  // (for purposes of namekeeping, e.g.)
+
+  // first, get the networkId from the blockchain
+  const networkId = yield* generateNetworkIdFetch();
+
+  // start enumerating over the artifacts
+  let latestNetwork: DataModel.Network | undefined;
+  const contractNetworks: ContractNetwork[] = [];
+  for (const artifact of options.artifacts) {
+    if (!artifact.networks[networkId]) {
+      // skip over artifacts that don't contain this network
+      continue;
+    }
+
+    const { transactionHash } = artifact.networks[networkId];
+
+    // load the historical network
+    const network = yield* generateTranasctionNetworkLoad({
+      transactionHash,
+      network: {
+        name: options.network.name,
+        networkId
+      }
+    });
+
+    if (
+      latestNetwork &&
+      latestNetwork.historicBlock.height < network.historicBlock.height
+    ) {
+      latestNetwork = network;
+    }
+
+    // keep track of this new network alongside the artifact
+    contractNetworks.push({
+      network: toIdObject(network),
+      artifact
+    });
+  }
+
+  // now, let's get ready to add the contract instances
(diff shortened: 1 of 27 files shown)


gnidan (commented):
(round of comments as notes to myself)

packages/db/src/loaders/commands/migrate.ts
```diff
@@ -0,0 +1,131 @@
+import { logger } from "@truffle/db/logger";
+const debug = logger("db:loaders:commands:migrate");
+
+import { ContractObject } from "@truffle/contract-schema/spec";
+import { toIdObject, IdObject } from "@truffle/db/meta";
+import { Load } from "@truffle/db/loaders/types";
+
+import { generateContractGet } from "@truffle/db/loaders/resources/contracts";
+import {
+  generateTranasctionNetworkLoad,
```
fainashalts, line 10:
```suggestion
  generateTransactionNetworkLoad,
```

packages/db/src/loaders/commands/migrate.ts
```diff
@@ -0,0 +1,131 @@
...
+  generateNetworkGet
+} from "@truffle/db/loaders/resources/networks";
+import {
+  LoadableContractInstance,
+  generateContractInstancesLoad
+} from "@truffle/db/loaders/resources/contractInstances";
+
+export interface GenerateMigrateLoadOptions {
+  // we'll need everything for this input other than what's calculated here
+  network: Omit<DataModel.NetworkInput, "networkId" | "historicBlock">;
```
fainashalts, line 21: I'm a little confused on this one -- you're creating a network type that takes everything from `NetworkInput` and then removes the `networkId` or the `historicBlock`?

packages/db/src/loaders/commands/migrate.ts
```diff
@@ -0,0 +1,131 @@
...
+  for (const artifact of options.artifacts) {
+    if (!artifact.networks[networkId]) {
+      // skip over artifacts that don't contain this network
+      continue;
+    }
+
+    const { transactionHash } = artifact.networks[networkId];
+
+    // load the historical network
+    const network = yield* generateTranasctionNetworkLoad({
```
fainashalts, line 52:
```suggestion
    const network = yield* generateTransactionNetworkLoad({
```

packages/db/src/loaders/commands/migrate.ts
```diff
@@ -0,0 +1,131 @@
...
+          linkReferences: callBytecode.linkReferences
+        },
+        create: {
+          bytecode: toIdObject(createBytecode),
+          linkReferences: createBytecode.linkReferences
+        }
+      }
+    });
+  }
+
```
fainashalts, line 129: I imagine we would add `NetworkGenealogies` here?

packages/db/src/loaders/resources/compilations/find.graphql.ts
```diff
@@ -0,0 +1,11 @@
+import gql from "graphql-tag";
+
+export const FindCompilationContracts = gql`
```
fainashalts, line 3: Nice!

packages/db/src/loaders/resources/networks/index.ts
```diff
@@ -1,4 +1,116 @@
...
+  debug("Generated network get.");
+  return network;
+}
+
+export interface GenerateTransactionNetworkLoadOptions {
+  transactionHash: TransactionHash;
+  network: Pick<DataModel.NetworkInput, "name" | "networkId">;
+}
+
+export function* generateTranasctionNetworkLoad({
```
fainashalts, line 37:
```suggestion
export function* generateTransactionNetworkLoad({
```

packages/db/src/loaders/run.ts
```diff
@@ -2,26 +2,25 @@ import { logger } from "@truffle/db/logger";
...
   R extends RequestType | undefined = undefined
 >(
   loader: Loader<A, T, R>,
   ...args: A
 ) => Promise<T>;
 
-export const forDb = (db): {
-  forProvider(provider: Web3["currentProvider"]): {
+export interface Db {
+  query: (query: DocumentNode | string, variables: any) => Promise<any>;
```
fainashalts, line 19: Do these have to be `any`? I guess we don't know what they'll be ahead of time per se, but could we use `unknown` instead?

fainashalts (changes requested):
Found a typo, looks great otherwise!

packages/db/src/loaders/commands/migrate.ts
gnidan: This says that the `network` option must have every field in `DataModel.NetworkInput`, except `networkId` and `historicBlock` (because those two fields are the ones that `generateMigrateLoad` itself supplies)

packages/db/src/loaders/commands/migrate.ts
gnidan: Yep, hereabouts

packages/db/src/loaders/run.ts
gnidan: Hm, we _could_ know it ahead of time, but I put it as `any` out of laziness

fainashalts: I don't feel super strongly about it but I have been trying to avoid using `any` where I can.

fainashalts (approved):
LGTM!

packages/db/src/loaders/run.ts
gnidan: Definitely. `unknown` isn't much better, unfortunately. I'm inclined to let this slide, though, because this is kind of just passed to GraphQL sight-unseen, so it's whatever they say is allowable.

Also, GraphQL validates all requests coming in at runtime, so I'm less concerned about type-safety at this boundary point.

gnidan:
Thanks @fainashalts !

PR #3484, opened 2020-10-30
Bug fix: Check for missing AST in orderABI to prevent crash

haltman-at:
A crash of this form was pointed out by @interfect on Gitter.  Dunno why there's no AST, but looking at that can come later, I think; first I figure I'll just prevent the crash.


gnidan (approved):
(Why does it get to this point without an AST?)

haltman-at:
Yeah, that's the question, isn't it?  I figured I'd just submit this PR first and puzzle over that later...

PR #3485, opened 2020-10-30
Bug fix: Make forTx and session.load throw on failure (breaking change!)

haltman-at:
**This PR is a breaking change to `@truffle/debugger`.**

As requested by @seesemichaelj, this PR makes it so that when a debugger method fails to load a transaction, it will throw an error, rather than succeeding and then exposing that error via another mechanism.  (That other mechanism is still there, though, and we still use it in the CLI to avoid having to change too much.)

So, now the debugger's `load()` method will throw an error if it fails to load a transaction, rather than having no effect (other than setting the error), and `Debugger.forTx()` will throw if it fails to load the transaction, rather than returning a debugger with no transaction loaded.


gnidan (approved):
Sweet!

PR #3486, opened 2020-10-30
Add @truffle/db contract correlation to artifacts

gnidan:
Define property `db` on the ContractObject schema and the NetworkObject schema, allowing artifacts to contain relevant information about @truffle/db resources.

For example, artifacts will be expected to contain information something like:

```javascript
{
  // ... rest of artifact ...
  "db": {
    "source": {
      "id": "0xb63608deefb65ab93e8beebe5fb0397417c14dc54cb6cee5748e590d16ca5986"
    },
    "createBytecode": {
      "id": "0x5ad957bd618157fd28fac3ffa61dd5275ff6de8494e32323029a31daef7ac56a"
    },
    "callBytecode": {
      "id": "0x5ec2ead4cd51f88805b33da49c5ec4c57388992ae1f6492f8b17c38c5045d3d6"
    },
    "compilation": {
      "id": "0x8534281c01751b9701e53e91fd97733f092ab2a1048f215fb4ced35f67b2644d"
    },
    "contract": {
      "id": "0xa3d6cdfe66ead5c6843c5297248b79ab7090e8230247b2f8a9637ea7fd04b8c6"
    }
  }
}
```

### DIFF
--- a/packages/contract-schema/index.js
+++ b/packages/contract-schema/index.js
@@ -176,7 +176,8 @@ var properties = {
   },
   networkType: {},
   devdoc: {},
-  userdoc: {}
+  userdoc: {},
+  "db:contract": {}
 };
 
 /**
--- a/packages/contract-schema/index.js
+++ b/packages/contract-schema/index.js
@@ -177,7 +177,7 @@ var properties = {
   networkType: {},
   devdoc: {},
   userdoc: {},
-  "db:contract": {}
+  db: {}
 };
 
 /**
@@ -245,7 +245,6 @@ var TruffleContractSchema = {
             params,
             message,
             data,
-            schema, // eslint-disable-line no-unused-vars
             parentSchema
           }) =>
             util.format(
--- a/packages/compile-common/src/shims/LegacyToNew.ts
+++ b/packages/compile-common/src/shims/LegacyToNew.ts
@@ -24,7 +24,8 @@ export function forContract(contract: any): CompiledContract {
     userdoc,
     immutableReferences,
     generatedSources,
-    deployedGeneratedSources
+    deployedGeneratedSources,
+    db
   } = contract;
 
   return {
@@ -44,7 +45,8 @@ export function forContract(contract: any): CompiledContract {
     userdoc,
     immutableReferences,
     generatedSources,
-    deployedGeneratedSources
+    deployedGeneratedSources,
+    db
   };
 }
 
--- a/packages/compile-common/src/shims/NewToLegacy.ts
+++ b/packages/compile-common/src/shims/NewToLegacy.ts
@@ -18,7 +18,8 @@ export function forContract(contract: CompiledContract): any {
     userdoc,
     immutableReferences,
     generatedSources,
-    deployedGeneratedSources
+    deployedGeneratedSources,
+    db
   } = contract;
 
   return {
@@ -39,7 +40,8 @@ export function forContract(contract: CompiledContract): any {
     userdoc,
     immutableReferences,
     generatedSources,
-    deployedGeneratedSources
+    deployedGeneratedSources,
+    db
   };
 }
 
--- a/packages/compile-common/src/types.ts
+++ b/packages/compile-common/src/types.ts
@@ -45,6 +45,7 @@ export type CompiledContract = {
   immutableReferences: ImmutableReferences;
   generatedSources: any;
   deployedGeneratedSources: any;
+  db?: any;
 };
 
 export interface WorkflowCompileResult {
--- a/packages/contract-schema/spec/contract-object.spec.json
+++ b/packages/contract-schema/spec/contract-object.spec.json
@@ -90,6 +90,15 @@
     },
     "deployedGeneratedSources": {
       "$ref": "#/definitions/GeneratedSources"
+    },
+    "db:contract": {
+      "type": "object",
+      "description": "Reference to @truffle/db canonical ID object for correlation purposes",
+      "properties": {
+        "id": {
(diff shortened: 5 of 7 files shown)


gnidan:
@seesemichaelj you came up as a reviewer suggestion but I figure it might just be good for you to be aware that I'm trying to get this change in

packages/contract-schema/index.js
```diff
@@ -176,7 +176,8 @@ var properties = {
   },
   networkType: {},
   devdoc: {},
-  userdoc: {}
+  userdoc: {},
+  "db:contract": {}
```
seesemichaelj, line 180: Why not have a new schema for `db` where `contract` is defined? Seems like that's how it's going to be implemented, so I'm not sure why we should bundle it up like this. Not picky, just food for thought

gnidan: My reasoning for adding it here is so that, if you have `truffle compile` hooked up to save to @truffle/db, then it would also include this correlation information for future use, like with a subsequent `truffle migrate`.

I'm not sure what you mean by "new schema for `db`", could you expand on that? A new schema alongside `contract-schema.spec.json`?

seesemichaelj: > My reasoning for adding it here is so that, if you have truffle compile hooked up to save to @truffle/db, then it would also include this correlation information for future use, like with a subsequent truffle migrate.

Ya I'm cool with `db` living in this schema, but I'm mainly concerned about the use of `:` and see how there may be future `db:*` fields. So why not make it `db` which points to some other type/schema which then defines `contract`?

gnidan: I first considered a `db` field for this, containing `contract`, but my reasoning against it is that I can see it looking like this:
```javascript
{
  // ...

  db: {
   contract: { id: "0x..." }
  }

  // ...

  networks: {
    // ...

    "1": {
      // ...

      db: {
        contractInstance: { id: "0x..." }
      }
    }
  }
}
```

Which, I dunno, I guess it's fine, but it just kind of irks me? I will say: looking at it now, it irks me less than when I first considered it.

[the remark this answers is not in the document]
gnidan: I've considered this a bunch... we don't use the term "DB" for anything else right now. @truffle/db doesn't even call anything internally a DB. I'm happy if this never changes; I don't see why Truffle would be forced to call something a DB alone, rather than naming such a component in relation to its purpose (e.g., if there becomes some external persistence system for some other unrelated thing, then just calling that thing a "DB" is probably insufficient).

Ultimately, my perspective is that it benefits us to align identifiers with the brand name, even if it's not technically correct.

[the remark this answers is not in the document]
gnidan: Basically, to Truffle, @truffle/db is the database. It's not just an ORM because an ORM carries the notion of generality. Instead: there's one data model, and it's Truffle's data model. @truffle/db is what stores that data model for Truffle.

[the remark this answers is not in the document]
seesemichaelj: Fair arguments. Any response I have to this is really just down to being pedantic. I'm fine with what you have. It's just confusing for people that don't know the architecture of truffle that `db` is Truffle DB rather than some generic reference to another database. It's just ambiguous because you're using the ambiguous term, not the branded term

seesemichaelj: frankly, the only thing that irks me in your example is the fact we have two things that are so close in name and imply two different things (`contract` and `contractInstance`). I'm kind of fine with the multiple usages of `db` which have different types/schemas depending on their context (i.e. `networks["1"].db` shouldn't have a `contract` and `db` should not have a `contractInstance`)

But it seems like you understand my point. I'm not picky either way here, but do please consider how it may change. I don't want to see this becoming

``` json5
{
  // ...
  "db:contract": { id: "0x..." }
  "db:<somefuturething>": { /* .... */ }
}
```

when it's quite clear the above would be sensible to implement as 

``` json5
{
  // ...
  db: {
    contract: { id: "0x..." },
    somefuturething: { /* .... */ }
  }
}
```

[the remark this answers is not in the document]
gnidan: I hear that. My hope is that it's no worse than our other generic terms that we use everywhere that you just kind of have to realize "this is a truffle thing"

gnidan: The thing is: I don't think we'll add another `db:*` alongside an existing one. At the top level, it's a Contract resource. At the network level, it's a ContractInstance resource. With those two IDs you have access to everything else that would be represented by the artifact. This seems very unlikely to change.



> we have two things that are so close in name and imply two different things (`contract` and `contractInstance`).

They are indeed very different things. We had a naming discussion long ago, where the competing argument was to call it simply "Instance" rather than "ContractInstance". But clarity won out.

gnidan: (Oh, just in case it's not clear: I'm not at all trying to argue with you. I recognize all your points, for sure. I just want to provide context about my thinking so you know where I'm coming from)

gnidan: So it turned out to be better in implementation to do `db` instead of `"db:contract"`. My original thinking that "it wasn't necessary to include more than the contract" has turned into "actually we can get a bunch of other stuff in here for free that might come in handy"

fainashalts: I am only seeing this discussion now but am glad for the decision to do `db` instead. I think it makes sense to include the other resources as it seems you have now!

fainashalts (approved):
LGTM!

gnidan:
Thanks!

PR #3491, opened 2020-11-02
Ensure isomorphism in @truffle/db input/output

gnidan:
So that queries and mutations that accept a list of inputs thus return a list of outputs that one-to-one correspond to the inputs.

This allows filtering for lists of IDs, some of which may be unknown resources, and it allows mutating multiple resources, some of which may be repeated in the list or cause errors


fainashalts (approved):
LGTM!

gnidan:
Thanks!

PR #3494, opened 2020-11-03
Enhancement: Add debugger commands to turn internal sources on/off

haltman-at:
This PR adds two new commands to the debugger CLI, `g` and `G`.  (Not sure if these are the best letters to use but I couldn't think of anything better; they can be changed, anyhow.)

The `g` command makes it so that all debugger commands can now step into generated sources, even if not starting from inside one, instead of just `;` and `c`.  The `G` command restores the default behavior, where you can only step into a generated source if you're already inside one or use `;` (or `c` and happen to hit a breakpoint inside one).

In order to implement this, the controller now has an extra bit of state, `stepIntoInternalSources`, and there's a new session method to set this state, `setInternalStepping`.  The `stepNext` method checks this state and, if it's on, it ignores its conditions about not stopping inside an internal source unless it started one.  Since the other commands like `stepOver` and `stepInto` fundamentally rely on `stepNext`, only `stepNext` needed modifying.

And, that's really it.  Pretty simple.

I didn't add any tests in this PR; I just tested it manually.

### DIFF
--- a/packages/core/lib/debug/interpreter.js
+++ b/packages/core/lib/debug/interpreter.js
@@ -408,6 +408,18 @@ class DebugInterpreter {
         );
       }
     }
+    if (cmd === "g") {
+      this.session.setInternalStepping(true);
+      this.printer.print(
+        "All debugger commands can now step into generated sources."
+      );
+    }
+    if (cmd === "G") {
+      this.session.setInternalStepping(false);
+      this.printer.print(
+        "Commands other than (;) and (c) will now skip over generated sources."
+      );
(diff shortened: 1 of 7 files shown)


haltman-at:
Oh, one additional comment.  For the interface help, I decided to leave the actual help messages short, and put more detail in the messages you get when you actually enter the command.  So if you try it out, you'll find out what it does. :)

(Of course the website docs will need to be updated as well but that will come later.)

gnidan (approved):

PR #3496, opened 2020-11-03
Bug fix: Cause debugger CLI to error on invalid tx with --fetch-external

haltman-at:
This one's pretty simple.  Debugger CLI wouldn't error out if you tried to load an invalid transaction with `--fetch-external`.  Redid the startup so now it does.  Note I got rid of the reliance on the `session.status.isError` selector, although we still use that elsewhere.  Oh well.  Maybe we can redo that more thoroughly later.

Also, unrelated, but while I was at it I added a second commit changing `selectors.session` to just `session` everywhere in `interpreter.js` since that longhand version hasn't been necessary for a while.


gnidan (approved):
nice

PR #3507, opened 2020-11-06
Use license from config/package for new contracts

agonzalezro:
Even when you have a license listed in your project configuration it wasn't
being used for new contracts, with this small change the license listed in
config will be used.

How to see it working?

1. `truffle init someproject`
2. `cd someproject`
3. Edit `truffle-config.json` and add a key `license: "GPL"` or whatever
   license you want to use.
4. `truffle create contract Contract`
5. `head -n1 contracts/Contract.sol` should show the new license.

In case the license is not present in the `truffle-config.json` file it will
check for it in the `package.json` file. The process to check this is the same
than the previously mentioned but writting the license in `package.json`.

### DIFF
--- a/packages/core/lib/commands/create/helpers.js
+++ b/packages/core/lib/commands/create/helpers.js
@@ -10,6 +10,7 @@ const templates = {
   contract: {
     filename: path.join(__dirname, "templates", "Example.sol"),
     name: "Example",
+    license: "MIT",
     variable: "example"
   },
   migration: {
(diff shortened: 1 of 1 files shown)


gnidan:
Hey @agonzalezro - could you make this fallback to a local `package.json` license instead?

agonzalezro:
@gnidan good idea. I've done it, however, I am not sure it's the best way (but I didn't find a better one).

Please take a look & let me know. Thanks!

agonzalezro:
I don't know why it could be failing only for certain versions of Ubuntu. Ubuntu 14 with `PACKAGES=true` works, but Ubuntu 12 with `PACKAGES=true` seems to be the one failing before the others are cancelled. Any idea what could be going on? I will investigate what `PACKAGES=true` means.

haltman-at:
@agonzalezro I think it's just random unrelated failures, sorry; our CI still has some problems.  I went and restarted it just now... hopefully it works this time?  I don't think it's related to the actual contents of this PR.

packages/core/lib/commands/create/helpers.js
```diff
@@ -69,14 +70,28 @@ var Create = {
...
+
+      let license;
+      const Config = require("@truffle/config");
+      try {
+        license = Config.detect(options).license;
+      } catch {}
+
+      if (!license) {
+        const packageJSONPath = path.join(process.cwd(), "package.json");
+        license = require(packageJSONPath).license;
```
eggplantzzz, line 86: Don't we also need this in a `try/catch` in case the `package.json` doesn't exist? Then it will throw an error right?

agonzalezro: Yes, you are completely right. I've changed it. Thanks!

agonzalezro:
I will try to investigate and open a fix in other PR but I realised that if you use the function `replaceContents` more than once, for example in this PR we are doing it to add the license, sometimes it doesn't work. Probably something related with the flush of the open file, but I didn't want to mess more in this PR.

gnidan (approved):
Looks good now!

eggplantzzz (approved):
Looks good to me.

PR #3508, opened 2020-11-06
Allow `truffle create all [name]`

agonzalezro:
At the moment you need to create contracts, migrations and tests one by one,
with this option you will be able to create all at once.

Please, note that I didn't add any test for this. Happy to do it if you think they are needed and you can point me in the right direction.


eggplantzzz (approved):
I like it!

PR #3512, opened 2020-11-09
Db/network genealogy

fainashalts:
This PR add definitions/schema/resolvers for the `NetworkGenealogy` resource, as defined in the [data model](https://trufflesuite.github.io/artifact-updates/data-model.html).

The groundwork for adding network genealogies and searching for them is laid. The ability to look for potential ancestors and descendants is also laid out. Tests have been added in the `packages/db/src/` directory to ensure that basic operations can be performed with the `NetworkGenealogy` resource.

Namely, the tests check that: 
- NetworkGenealogy can be added
- It is possible to query for ancestors
- it is possible to query for descendants
- limits on these queries are respected
- networks in the  `alreadyTried` array are not returned as possible ancestors and descendants

Work remains to be done to hook this resource up to the loading logic of Truffle DB via the `truffle migrate` command.


packages/db/src/definitions/networkGenealogies.ts
```diff
@@ -0,0 +1,124 @@
...
+    },
+    Network: {
+      possibleAncestors: {
+        resolve: async ({id}, {limit, alreadyTried}, {workspace}) => {
+          const network = await workspace.get("networks", id);
+          const queryLimit = limit ? limit : 5;
+          const result = await workspace.find("networks", {
+            selector: {
+              "historicBlock.height": {
+                $gte: null,
```
fainashalts, line 57: Oddly, to get the sort to work I had to add the `$gte: null` line and also explicitly exclude the network I'm checking with. I'm a bit concerned about the `$ne` because that is done in-memory according to the docs which may be a hit to performance.

From the [pouchdb docs](https://pouchdb.com/guides/mango-queries.html): "For instance, $regex, $ne, and $not cannot use on-disk indexes, and must use in-memory filtering instead"

PR #3513, opened 2020-11-10
Redo @truffle/db's Project abstraction to work more naturally in batches

gnidan:
The purpose of this PR is to enable Project clients to link their inputs (WorkflowCompileResult and/or Artifacts) back with the created @truffle/db resources. This alters the method signatures of each Project method to return the same input data along with additional @truffle/db ID objects. **This is necessary because otherwise there's no way for `truffle migrate` to know what contract IDs to use**

For `project.loadCompile()`, this populates each `CompiledContract` with `db: { contract, source, callBytecode, createBytecode }`. For `project.loadMigrate()`, this populates the `NetworkObject` with `db: { contractInstance, network }`.

To make this possible, this adds a `src/process` module, similar to `src/graphql` and `src/pouch` in that it operates on any collection of resources. This module encompasses the existing `run` functionality along with two other components:
- A `resources` processor, which removes the need for custom queries per-resource, and replaces that with, e.g. `resources.load("sources, ids)`
- A batching system that operates on structured inputs and returns outputs in that same structure

Using those two components, this PR updates Project to define batch processes for compilation, migration, and name assignment.

In addition, this PR makes a few other changes:
- Remove `src/loaders`, since much of that functionality is no longer necessary (`src/loaders/resources/` is unneeded, since there's no `resources.load`, etc.), and whatever is needed fits better into `src/project`
- Add Bytecode to SourceMap, since that link is necessary and missing (and change the `json` field to the more accurate `data`)

I've also included an initial ARCHITECTURE.md file, since things are starting to come together there.

Here's what the deps look like now: [deps.pdf](https://github.com/trufflesuite/truffle/files/5518948/deps.pdf)


packages/db/src/process/batch.ts
```diff
@@ -0,0 +1,183 @@
+import { $, _ } from "hkts/src";
+export { _ };
+
+import { Collections } from "@truffle/db/meta";
+import { Process } from "./types";
+
+/**
+ * Configure a batch process to query and/or mutate resources.
+ *
+ * Batch processes take structured inputs, split those up and extract a list
```
fainashalts, line 10: This is so helpful!

packages/db/src/test/compilation.graphql.ts
```diff
@@ -13,7 +14,7 @@ export const AddCompilation = gql`
         compilations: [
           {
             compiler: { name: $compilerName, version: $compilerVersion }
-            sourceMaps: [{ json: $sourceMap }]
+            sourceMaps: [{ bytecode: { id: $bytecodeId }, data: $sourceMap }]
```
fainashalts, line 17: Just want to make a note here that we should update the data model to reflect this change.

fainashalts (approved):
LGTM! Thanks for the run-down on this before I reviewed it, that really helped! It looks like there are some conflicts that need to be resolved before this can be merged, fyi.

gnidan:
Thanks @fainashalts! I'll make that data model change tomorrow.

PR #3514, opened 2020-11-10
Internal improvement: Remove callbacks from cli.js interface with the commands

eggplantzzz:
This PR aims to get rid of the callback interface used by `packages/core/cli.js`. Currently it takes a callback as the last argument. The idea is that it would be much easier to reason about all this stuff if instead this returned a Promise. All of the command files' `run` methods are updated to return a Promise.

### DIFF
--- a/packages/core/test/lib/create.js
+++ b/packages/core/test/lib/create.js
@@ -16,136 +16,107 @@ describe("create", function () {
     config.artifactor = new Artifactor(config.contracts_build_directory);
   });
 
-  after("Cleanup tmp files", function (done) {
-    glob("tmp-*", (err, files) => {
-      if (err) done(err);
-      files.forEach(file => fse.removeSync(file));
-      done();
-    });
+  after("Cleanup tmp files", async () => {
+    const files = glob.sync("tmp-*");
+    files.forEach(file => fse.removeSync(file));
   });
 
-  it("creates a new contract", function (done) {
-    Create.contract(config.contracts_directory, "MyNewContract", function (
-      err
-    ) {
-      if (err) return done(err);
+  it("creates a new contract", async () => {
+    await Create.contract(config.contracts_directory, "MyNewContract", {});
 
-      const expectedFile = path.join(
-        config.contracts_directory,
-        "MyNewContract.sol"
-      );
-      assert.isTrue(
-        fse.existsSync(expectedFile),
-        `Contract to be created doesns't exist, ${expectedFile}`
-      );
+    const expectedFile = path.join(
+      config.contracts_directory,
+      "MyNewContract.sol"
+    );
+    assert.isTrue(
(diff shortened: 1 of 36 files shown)


packages/core/test/lib/create.js
```diff
@@ -16,136 +16,107 @@ describe("create", function () {
...
-          file.length - expectedSuffix.length
-        ) {
-          const fileData = fse.readFileSync(file, { encoding: "utf8" });
-          assert.isNotNull(fileData, "File's data is null");
-          assert.notEqual(fileData, "", "File's data is blank");
-
-          return done();
-        }
-      }
+  it("creates a new test", async () => {
```
gnidan, line 85: You even decallbacked internals :)

packages/truffle/scripts/test.sh
```diff
@@ -7,9 +7,9 @@ if [ "$GETH" == true ]; then
...
   mocha --timeout 50000 --grep @fabric-evm --colors $@
 elif [ "$COVERAGE" == true ]; then
-  NO_BUILD=true mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
+  NO_BUILD=true mocha --no-warnings --timeout 20000 --grep @geth --invert --colors $@
 elif [ "$INTEGRATION" == true ]; then
-  mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
+  mocha --no-warnings --timeout 20000 --grep @geth --invert --colors $@
 else
-  yarn build && mocha --no-warnings --timeout 7000 --grep @geth --invert --colors $@
+  yarn build && mocha --no-warnings --timeout 20000 --grep @geth --invert --colors $@
```
gnidan, line 14: these are unfortunate :/ promises slowwww. we should do further testing... we really can't afford to make truffle slower.

gnidan (changes requested):
this is amazing. but I got the end and saw the timeout bumps :(

is this really that much slower?

packages/truffle/scripts/test.sh
eggplantzzz: We will need to try and test the speed some other way. Raising the timeouts was not necessarily because they were slow. I got rid of a lot of the timeouts specified in the individual tests as I think a lot of them were not working.

eggplantzzz:
I ran some benchmarks against the currently released Truffle and it seems they are about the same in speed. I ran three commands three times against a toy project and got the following results:

|command           | 5.1.44                        | this PR |
|----------------|---------------------|--------|
| migrate --reset | 11.91, 12.68, 12.08   | 12.06, 12.12, 12.22

gnidan (approved):
I forgot the only concern I had left was timing, and didn't realize you proved that wrong! Sorry about the unnecessary delay.

This is great! Let's get this into `next` pronto!

PR #3516, opened 2020-11-11
Db/network genealogy develop

fainashalts:
This PR add definitions/schema/resolvers for the NetworkGenealogy resource, as defined in the data model.

The groundwork for adding network genealogies and searching for them is laid. The ability to look for potential ancestors and descendants is also laid out. Tests have been added in the packages/db/src/ directory to ensure that basic operations can be performed with the NetworkGenealogy resource.

Namely, the tests check that:

NetworkGenealogy can be added
It is possible to query for ancestors
it is possible to query for descendants
limits on these queries are respected
networks in the alreadyTried array are not returned as possible ancestors and descendants
Work remains to be done to hook this resource up to the loading logic of Truffle DB via the truffle migrate command.


packages/db/src/loaders/resources/networkGenealogies/index.ts
```diff
@@ -0,0 +1,2 @@
+export { AddNetworkGenealogies } from "./add.graphql";
+export { FindAncestors, FindDescendants } from "./search.graphql";
```
gnidan, line 2: I propose we just remove all the loaders changes, since you're not using them and I'm deleting them elsewhere :)

packages/db/src/definitions/networkGenealogies.ts
```diff
@@ -0,0 +1,124 @@
...
+import {Definition} from "./types";
+
+export const networkGenealogies: Definition<"networkGenealogies"> = {
+  createIndexes: [],
+  idFields: ["ancestor", "descendant"],
+  typeDefs: gql`
+    type CandidateSearchResult {
+      network: Network!
+      alreadyTried: [ID]! #will include all networks returned
+    }
```
gnidan, line 12: Stuff that relates to Networks more directly than NetworkGenealogies, like this, would probably fit better in `definitions/networks/`.

That way, this file is just the bare minimum about NetworkGenealogy, rather than including logic about what the existence of NetworkGenealogy does to Network

packages/db/src/definitions/networkGenealogies.ts
```diff
@@ -0,0 +1,124 @@
...
+    extend type Network {
+      possibleAncestors(
+        alreadyTried: [ID]!
+        limit: Int # will default to 5
+      ): [CandidateSearchResult]!
+      possibleDescendants(
+        alreadyTried: [ID]!
+        limit: Int # will default to 5
+      ): [CandidateSearchResult]!
+    }
```
gnidan, line 34: In the same reasoning as above, this could just be part of `type Network {}` in the first place (in `definitions/networks`)

packages/db/src/definitions/networkGenealogies.ts
```diff
@@ -0,0 +1,124 @@
...
+          return parent.network;
+        }
+      },
+      alreadyTried: {
+        resolve: (parent, __, {}) => {
+          return parent.alreadyTried;
+        }
+      }
+    }
+  }
```
gnidan, line 123: Same reasoning as above, this can go in `definitions/networks`

packages/db/src/test/networkGenealogy.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+import {
+  FindAncestors,
+  FindDescendants,
+  AddNetworkGenealogies
+} from "./networkGenealogy.graphql";
+import {AddNetworks} from "./network.graphql";
+
+// returnBlockHeader mocks the bare-bones return of a call to web3,
+// to check whether a network resource is a part of the network we're connected to
+const returnBlockHeader = blockNumber => {
```
gnidan, line 11: What do you think of the name `findBlockByNumber`? `returnBlockHeader` sounds like a noun, and functions should be verbs :)

packages/db/src/test/networkGenealogy.spec.ts
```diff
@@ -0,0 +1,284 @@
...
+    genealogyCheck = async (
+      id: string,
+      alreadyTried: [string],
+      type: string,
+      limit?: number
+    ) => {
+      let result;
+      let possibleMatches;
+
+      const queryLimit = limit ? limit : 5;
```
gnidan, line 61: JS has some nice syntax for default values you can use

```suggestion
      queryLimit: number = 5
    ) => {
      let result;
      let possibleMatches;
```

packages/db/src/definitions/networkGenealogies.ts
```diff
@@ -0,0 +1,124 @@
...
+                $ne: network.historicBlock.height
+              },
+              "networkId": network.networkId,
+              "id": {
+                $nin: alreadyTried
+              }
+            },
+            sort: [{"historicBlock.height": "asc"}],
+            limit: queryLimit
+          });
```
gnidan, line 98: You can use an object property default here

```suggestion
        resolve: async ({id}, {limit = 5, alreadyTried}, {workspace}) => {
          const network = await workspace.get("networks", id);
          const result = await workspace.find("networks", {
            selector: {
              "historicBlock.height": {
                $gte: null,
                $gt: network.historicBlock.height,
                $ne: network.historicBlock.height
              },
              "networkId": network.networkId,
              "id": {
                $nin: alreadyTried
              }
            },
            sort: [{"historicBlock.height": "asc"}],
            limit
          });
```

packages/db/src/definitions/networkGenealogies.ts
```diff
@@ -0,0 +1,124 @@
...
+                $ne: network.historicBlock.height
+              },
+              "networkId": network.networkId,
+              "id": {
+                $nin: alreadyTried
+              }
+            },
+            sort: [{"historicBlock.height": "desc"}],
+            limit: queryLimit
+          });
```
gnidan, line 68:
```suggestion
        resolve: async ({id}, {limit = 5, alreadyTried}, {workspace}) => {
          const network = await workspace.get("networks", id);
          const result = await workspace.find("networks", {
            selector: {
              "historicBlock.height": {
                $gte: null,
                $lt: network.historicBlock.height,
                $ne: network.historicBlock.height
              },
              "networkId": network.networkId,
              "id": {
                $nin: alreadyTried
              }
            },
            sort: [{"historicBlock.height": "desc"}],
            limit
          });
```

packages/db/src/definitions/networkGenealogies.ts
```diff
@@ -0,0 +1,124 @@
...
+
+          let untriedNetworks = result.map(network => {
```
gnidan, line 70:
```suggestion
          const untriedNetworks = result.map(network => {
```

packages/db/src/definitions/networkGenealogies.ts
```diff
@@ -0,0 +1,124 @@
...
+              "networkId": network.networkId,
+              "id": {
+                $nin: alreadyTried
+              }
+            },
+            sort: [{"historicBlock.height": "asc"}],
+            limit: queryLimit
+          });
+
+          let untriedNetworks = result.map(network => {
```
gnidan, line 100:
```suggestion
          const untriedNetworks = result.map(network => {
```

gnidan (changes requested):
Some organization changes, and a style tip

Let me know if these makes sense!

packages/db/src/loaders/resources/networkGenealogies/index.ts
fainashalts: You got it!

gnidan (approved):
This looks good. Couple comments for stuff that we shouldn't forget, but I'd say this is good to merge.

packages/db/src/resources/networks.ts
```diff
@@ -34,5 +47,79 @@ export const networks: Definition<"networks"> = {
...
+    Network: {
+      possibleAncestors: {
+        resolve: async ({id}, {limit = 5, alreadyTried}, {workspace}) => {
+          const network = await workspace.get("networks", id);
+          const result = await workspace.find("networks", {
+            selector: {
+              "historicBlock.height": {
+                $gte: null,
+                $lt: network.historicBlock.height,
+                $ne: network.historicBlock.height
```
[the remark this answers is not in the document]
fainashalts: Yes it should imply that! For some reason I was unable to get this to work otherwise. Happy to revisit it as a separate task but didn't want to spend more time on it as part of this PR.

packages/db/src/resources/networks.ts
```diff
@@ -34,5 +47,79 @@ export const networks: Definition<"networks"> = {
...
+  `,
+  resolvers: {
+    Network: {
+      possibleAncestors: {
+        resolve: async ({id}, {limit = 5, alreadyTried}, {workspace}) => {
+          const network = await workspace.get("networks", id);
+          const result = await workspace.find("networks", {
+            selector: {
+              "historicBlock.height": {
+                $gte: null,
```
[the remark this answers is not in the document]
fainashalts: Yep, unfortunately. This has to do with the order in which the selector selects things that are indexed. Without this check it doesn't work. I found this workaround/hack deep in the github issues of pouchdb if I remember correctly.

packages/db/src/resources/networks.ts
```diff
@@ -34,5 +47,79 @@ export const networks: Definition<"networks"> = {
...
+          return untriedNetworks;
+        }
+      },
+      possibleDescendants: {
+        resolve: async ({id}, {limit = 5, alreadyTried}, {workspace}) => {
+          const network = await workspace.get("networks", id);
+          const result = await workspace.find("networks", {
+            selector: {
+              "historicBlock.height": {
+                $gte: null,
```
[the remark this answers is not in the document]
fainashalts: No, you would think so, right? Basically this says it is greater than or equal to null AND it is greater than the height we're checking against. I believe these are checked in order.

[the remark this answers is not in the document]
gnidan: So strange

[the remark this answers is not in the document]
gnidan: Oh well. Let's just get this merged and cleanup later :)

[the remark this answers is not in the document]
fainashalts: Was able to remove the `$gte: null` check successfully! Still need the `$ne` check though, which I find annoying...

PR #3523, opened 2020-11-12
Allow configuring host for `truffle db serve`

gnidan:

PR #3524, opened 2020-11-12
Ensure artifacts are saved with db info

gnidan:
So that @truffle/db resources are referenced in the resulting `truffle compile` artifacts


fainashalts (changes requested):
This totally makes sense! LGTM but failing CI.

gnidan:
Yep, sorry, I got jumpy with requesting review, thinking this and #3523 were going to pass 🤭

fainashalts (approved):
LGTM!

PR #3525, opened 2020-11-12
Hook up `truffle migrate` to @truffle/db

gnidan:
So that networks and contract instances get saved in DB


fainashalts (approved):
Woohoo! LGTM! Except prettier seems to still be taking all the spaces away...

PR #3526, opened 2020-11-12
Allow querying @truffle/db bytecodes for parsed instructions

gnidan:
- Define `Bytecode.instructions` resolver to invoke @truffle/code-utils
- Expose optional parameter `count` to limit the number of instructions
- Remove source-related types and relationships, to be added later as part of compilations

Screen Shot 2020-11-12 at 13 27 42


packages/db/src/resources/bytecodes.ts
```diff
@@ -13,7 +14,7 @@ export const bytecodes: Definition<"bytecodes"> = {
       id: ID!
       bytes: Bytes!
       linkReferences: [LinkReference]
-      instructions: [Instruction!]
+      instructions(count: Int): [Instruction!]
```
haltman-at, line 17: I'm not clear on why this is a function.  (Am I reading this right?  It's a function?)  I assume that has something to do with `parseCode` currently taking a count, but, like, this still shouldn't be a function, right?  It should be, like, the right thing.

Perhaps it would make sense to hold off on this PR until we change `parseCode`?  I could do that pretty quickly assuming we agree on what course of action to take.

packages/db/src/resources/bytecodes.ts
```diff
@@ -29,36 +30,9 @@ export const bytecodes: Definition<"bytecodes"> = {
     type Instruction {
       opcode: String!
       programCounter: Int!
-      meta: InstructionMeta
-      sourceRange: SourceRange
       pushData: Bytes
     }
 
-    type InstructionMeta {
```
haltman-at, line 37: Oh man what old version of code-utils were you on when you wrote this :P

packages/db/src/resources/bytecodes.ts
```diff
@@ -29,36 +30,9 @@ export const bytecodes: Definition<"bytecodes"> = {
     type Instruction {
       opcode: String!
       programCounter: Int!
-      meta: InstructionMeta
-      sourceRange: SourceRange
```
haltman-at, line 33: I guess this is being removed due to not fitting into the structure?  I gather there is now some other mechanism for handling this?

haltman-at (commented):
OK, I think the walkthrough did indeed give me a bunch of the needed context for understanding this. :)

packages/db/src/resources/bytecodes.ts
gnidan: This is just GraphQL syntax - since it's an optional argument (as opposed to `count: Int!`), you would query as:
```graphql
{
  bytecodes {
    instructions {
      programCounter
    }
  }
}
```

or if wanted to specify a count, then you'd use parens: 

```graphql
{
  bytecodes {
    instructions(count: 5) {
      programCounter
    }
  }
}
```

GraphQL just doesn't make the distinction like you'd see in JS

packages/db/src/resources/bytecodes.ts
gnidan: Previously I thought the source range information belonged in bytecodes, but now I know that not to be the case. In the future, I figure we'll grab this sorta information and put it in `src/resources/compilations` underneath `SourceMap`

packages/db/src/resources/bytecodes.ts
gnidan: This is just being removed for the time being - it doesn't belong as part of `bytecodes`, but rather it makes more sense underneath `compilations` (as I said in comment above)

packages/db/src/resources/bytecodes.ts
haltman-at: I see!  OK, that seems fine then, but in that case I think we should make the change in #3528 and take approach (3) so that we don't accidentally strip metadata on bytecode that lacks it.  Do you want me to do that?

packages/db/src/resources/bytecodes.ts
haltman-at: Oh this comment was just a joke about `InstructionMeta` (not the source range stuff). :P  I notice that it includes information that `code-utils` used to return but doesn't anymore.  I've actually thought about putting back in the pushes and pops information, but not the rest, no way... those fields never really made a lot of sense...

packages/db/src/resources/bytecodes.ts
gnidan: sure, sounds good!

packages/db/src/resources/bytecodes.ts
haltman-at: Wait, if it's not associated to the bytecode, it would be associated to the contract, right?  Not the compilation?

packages/db/src/resources/bytecodes.ts
haltman-at: Oh, wait, sorry, further question about this.  I'm pretty sure that this is correct, but, just to be certain: This is bytecode as output by the compiler, right?  So e.g. for constructor bytecode it will never have arguments attached, right?

gnidan: correct

packages/db/src/resources/bytecodes.ts
gnidan: Yes, but SourceMap is associated to Compilation, so that's where it'd go, I think

haltman-at: I'm confused.  Why would SourceMap be associated to Compilation?  I would expect each Contract to have two SourceMaps, I don't see why it would go all the way back to Compilation...

packages/db/src/resources/bytecodes.ts
haltman-at: Ah crap actually, I just realized -- stripping the metadata the way I was hoping doesn't actually work because there can be more metadata than just that.  You know what, this is fine, let's ignore it.  We can worry about this at a later time.

packages/db/src/resources/bytecodes.ts
gnidan: that might be a better way to do it, but that's not where it is now :)

haltman-at: Ugh.  I guess I'll approve this, since this is unrelated??  But that seems seriously messed up.  I'll need to bug you and @fainashalts about this later, I guess? :-/

haltman-at (approved):
Approving this, but I really need to bug you and @fainashalts about some of the data model I think...

packages/db/src/resources/bytecodes.ts
gnidan: I'm still not convinced that compilation is wrong, so I anticipate your bugging :)

gnidan:
Thanks @haltman-at !

PR #3529, opened 2020-11-13
Allow Compilations to have missing sources in @truffle/db

gnidan:
This removes the non-nullability requirement for source and processed source inputs, as well as fixes the resolvers to handle the case where these are null.

This addresses the situation where @truffle/workflow-compile reports `null` paths for particular source indexes

### DIFF
--- a/packages/db/src/project/compile/compilations.ts
+++ b/packages/db/src/project/compile/compilations.ts
@@ -41,6 +41,7 @@ export const generateCompilationsLoad = Batch.Compilations.generate<{
   },
 
   *process({ entries }) {
+    debug("entries %o", entries);
     return yield* resources.load("compilations", entries);
   },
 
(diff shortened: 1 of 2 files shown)


fainashalts (changes requested):
Can you add a test with an empty source? Or is that overkill?

I also wonder what scenario you're imagining this is needed for - factory patterns? Something else?

gnidan:
> I also wonder what scenario you're imagining this is needed for - factory patterns? Something else?

This arises due to the way compile-solidity sets things up. Let's say you have sources `A.sol`, `B.sol`, and `C.sol`, none of which import each other. If you compile them all, then make a change to `B.sol` and run `truffle compile` again, you'll get the resulting sourceIndexes `[null, "B.sol", null]`. So we have to handle that in @truffle/db.

> Can you add a test with an empty source? Or is that overkill?

I'm not sure how best to add a unit test for this... it seems like this might make a better scenario test, but we'd have to wait for `truffle-db-mvp` for that. What do you think?

fainashalts:
That seems reasonable!

fainashalts (approved):
LGTM!

gnidan:
Thanks!

PR #3534, opened 2020-11-13
Db/add network genealogy

fainashalts:
This PR add definitions/schema/resolvers for the NetworkGenealogy resource, as defined in the data model.

The groundwork for adding network genealogies and searching for them is laid. The ability to look for potential ancestors and descendants is also laid out. Tests have been added in the packages/db/src/ directory to ensure that basic operations can be performed with the NetworkGenealogy resource.

Namely, the tests check that:

NetworkGenealogy can be added
It is possible to query for ancestors
it is possible to query for descendants
limits on these queries are respected
networks in the alreadyTried array are not returned as possible ancestors and descendants
Work remains to be done to hook this resource up to the loading logic of Truffle DB via the truffle migrate command.


gnidan (approved):
take 2!

PR #3535, opened 2020-11-13
Improve CLI printouts regarding generated sources

haltman-at:
Just some simple improvements to the debugger CLI regarding the `g` and `G` commands.  These commands will now print a different message if you try to set the debugger to a state it's already in.  Also, the `?` command will now also print out whether generated sources are on or off.

...also while I was at it I fixed a typo in a test description. :P


gnidan (approved):
Convenient!

PR #3536, opened 2020-11-13
Internal improvement: Remove unused selector

haltman-at:
This selector is completely unused and refers to a piece of state that no longer exists; it will only ever return `undefined`.  It's from back when one had to wait for decoding to be ready.  That's no longer the case, and this should have been tossed long ago.


gnidan (approved):
what decodingKeys selector

Issue #3528, opened 2020-11-13
Add options to yield better control of metadata stripping in code-utils's parseCode function


haltman-at:
## Issue

The `parseCode` function in `@truffle/code-utils` can be used in two ways, affecting the question of where it should stop disassembling.  One is that you can specify the number of instructions to disassemble, if you know it.  The other, if you don't know the number, is that it can strip the metadata off the end.

But really, there should be a way of invoking `parseCode`, without knowing the number of instructions in advance, that *won't* strip metadata off the end.  This is because languages other than Solidity and Yul don't put this metadata on the end, so stripping it off the end would be incorrect.

## Expected Behavior

Some possible solutions:

1. Add a way of invoking it that doesn't strip metadata off.
2. Add a way of invoking it where one passes in the compiler being used, and let it determine whether to strip metadata based on that.
3. Make it so that the strip-metadata mode actually checks whether the metadata it's stripping is valid CBOR, like we do in the `normalizeContexts` function in `@truffle/codec`.  Then there's no need for a separate option not to strip the metadata.

Of these, (1) seems overall simplest, but would probably be breaking.  (3) is a bit more complex, but is non-breaking.  (2) seems the worst solution.

Note that even if we do (1) we might also want to do (3).  Possibly as a separate option, if we're going that route...

haltman-at:
Oh, crap -- I forgot that there can be more that may need stripping beyond just the metadata.  Oh well... it's probably OK to have extra stuff, I guess?

haltman-at:
This is released now, we went with option (3).  (So, not a new option, just an adjustment to the behavior.)

PR #3537, opened 2020-11-14
Bug fix: Only strip metadata in parseCode if it's actually valid CBOR

haltman-at:
Addresses #3528.

Now, if you pass `null` as the count, or don't pass a count, `parseCode` will only attempt to strip metadata if it finds valid CBOR on the end.  Otherwise it will leave it alone and not strip anything.  Note it doesn't actually check the content of the CBOR, just if it's valid.

Note this still doesn't really achieve the intent of metadata stripping in `parseCode`, which is to strip out all data on the end to leave only code.  (Because e.g. contracts that might get created also go on the end...)  However I have no idea how one might reasonably achieve that, so, whatever.  Better to leave too much in than to remove too much.

I also added a test of this, and also fixed a test that wasn't actually checking what it was supposed to!


haltman-at:
Oh, I forgot to mention -- while I was at it, I also changed `parseCode` to use a `Uint8Array` internally instead of a `number[]`.  I did convert it back to `number[]` at the end for output, since I didn't want to go changing any externally-visible types or anything.

gnidan (approved):

PR #3538, opened 2020-11-14
Upgrade @truffle/db's typescript to ^4.1.1-rc

gnidan:
... and use trailing slash in `baseUrl` to fix weird issue with @zerollup/ts-transform-paths


gnidan:
I'll just do this again later now that 4.1 is actually out

PR #3541, opened 2020-11-18
Bug fix: Prevent confusing breakpoint behavior with internal sources (and fix node-based breakpoints)

haltman-at:
This PR fixes an issue where breakpoints in user sources could behave in a confusing way when generated sources are present.  It also fixes a second issue where node-based breakpoints could be hit multiple times without leaving the node.

Let's say you place a breakpoint on a line; hit `c` to get to that line; and then hit `c` again.  You would expect that the debugger doesn't stop on that line again until it's left it at least once.  However, if generated sources are present, then the debugger might appear to do just that -- the reason is that it *has* left it, by entering a generated source; but from the user's point of view, this looks wrong.

So, I added an additional check, to both node-based and line-based breakpoints.  If you started in a user source, and generated sources are turned off, then line-based breakpoints won't only stop on a change of line, they'll also only stop on a change of line from a user perspective.  And similarly with node-based breakpoints.

Actually, with node-based breakpoints I also added a check for if the location has changed at all.  This wasn't relevant back when `continueUntilBreakpoint` used `stepNext`, but now that it uses `advance`, this is relevant and I'm kind of horrified I missed this earlier!  Oh well, I assume people mostly use line-based breakpoints...

Note that I made it so that this new check (the one about user-source-POV position changes) is disabled if generated sources are turned on.  Because if they're turned on, presumably the user does not want to ignore them!

This change is particularly relevant for the upcoming Solidity 0.8.x, which is going to make generated sources a whole lot more common!  (Because they're used for checked arithmetic and all arithmetic will be checked by default.)

...also, while I was at it, I removed some unused imports from a debugger test that Husky didn't catch while it was broken.

### DIFF
--- a/packages/debugger/lib/controller/sagas/index.js
+++ b/packages/debugger/lib/controller/sagas/index.js
@@ -246,6 +246,7 @@ function* continueUntilBreakpoint(action) {
     //few lines down.  but at this point these are still the previous
     //values.
     let previousLine = currentLine;
+    let previousNode = currentNode;
     let previousSourceId = currentSourceId;
     if (!currentLocation.source.internal) {
       lastUserSourceId = currentSourceId;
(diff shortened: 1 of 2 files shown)


gnidan (approved):
Nice catch!

haltman-at:
> Nice catch!

Well, it was causing test failures when I tried out the 0.8.0 preview, so!

PR #3544, opened 2020-11-19
Internal improvement: Add GHA check that yarn.lock is not modified on bootstrapping

haltman-at:
OK, I have very little idea if I've done this correctly.  People please look this over.

The idea -- as suggested by @gnidan a bit ago -- is that we should have a GHA job that checks that, if `yarn bootstrap` is run on the branch you're PRing, `yarn.lock` is not modified.  The idea being that if running `yarn bootstrap` modifies `yarn.lock`, you should have done that on your branch and committed the change.

That's what this PR attempts to do.  I have very little idea if I've done this correctly, as mentioned.  Hopefully at least the new job won't fail this very PR? :P  (Edit: It finally didn't! :) )

You can also see that I made the Slack notification depend on this new job in addition to the existing jobs.

Also, while I was at it, I removed the `push` trigger for the `truffle-db` branch, because that branch doesn't exist anymore.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -13,6 +13,13 @@ on:
       - "*"
 
 jobs:
+  yarncheck:
+    runs-on: ubuntu-latest
+    steps:
+      - run: yarn bootstrap
+      - run: test -z "$(git diff)"
+
+
   build:
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


gnidan (approved):

haltman-at:
OK, if @gnidan thinks it's good I guess I'll go ahead and merge this... @eggplantzzz, @fainashalts, @cds-amal, be warned, this check will now be active! :)

Issue #3310, opened 2020-08-25
Truffle Decoder tests shouldn't use Truffle Test


haltman-at:
## Issue

Truffle Decoder currently uses Truffle Test to do its tests.  This is done via a peer dependency, but that still could potentially screw up Lerna, even if it hasn't been a problem thus far.

PR #3548, opened 2020-11-20
Internal improvement: Migrate Travis-based tests to GitHub Actions

haltman-at:
This is me attempting to migrate our remaining Travis-based tests to GitHub actions.

We'll, uh, see whether it works?


haltman-at:
OK, I'm going to try removing the `truffle obtain` step, on the basis that it's probably not necessary?

haltman-at:
Hm, so:

1. The coverage build is failing due to that `indexOf` problem.  I'll put in the hacky workaround for that, I guess, seeing as we don't currently know how to fix that.
2. Those two failing builds are supposed to be optional, but GitHub is counting the build as failing anyway.  I'll, uh, have to figure out how to fix that...?  I thought I it was just `continue-on-error`, but...

haltman-at:
Ugh, the fricking decoder tests, with their stupid `npx`s!  Of course those are ignoring the version of Truffle present here... we really need to do something about those (#3310).

Well, I guess I can make the debugger thing a separate PR...?

haltman-at:
I'm gonna convert this to draft until after next release, I think.

haltman-at:
Ugh, turns out GHA actually doesn't yet have a notion of allowed failures: https://github.com/actions/toolkit/issues/399 :-/

I guess I'll just close this PR for now.

PR #3552, opened 2020-11-21
Enhancement: Add enum objects to contract objects

haltman-at:
This PR (suggested by @gnidan) adds enums to contract objects.  Note that top-level enums are not covered; only enums defined in a contract.

So, let's say you define a contract `C` which defines `enum E { e0, e1, e2 }`.

Then now, the contract object for `C` will have a field `C.E` which is an object `{e0: 0, e1: 1, e2: 2}`.  So, you could write `C.E.e2` for `2`.  This allows one to pass in enums to functions that take them without needing to worry about their numerical value.

Also, all the enums will be under `C.enums`, so you could also do `C.enums.E.e2` for `2`.  (I made it so that `C.E` will only be created if it doesn't overwrite anything, so `C.enums.E` exists as a fallback way to get at `E`.)

Note, I set all this up in `bootstrap.js`.  This is... not really the right for place for this.  The right place is probably `constructorMethods.js`.  But, well, this was easier.  Sorry.  Truffle Contract is a gigantic mess of overly clever dynamic code, and finding the right place to put things is hard.  Well, we'll redo all this in the rewrite.

Anyway, yeah, despite the ugliness, now there are enums!

Also, prettier made some changes.

Hm, maybe I should go back and add a test...

### DIFF
--- a/packages/contract-tests/test/enums.js
+++ b/packages/contract-tests/test/enums.js
@@ -1,0 +1,22 @@
+var assert = require("chai").assert;
+var util = require("./util");
+
+describe("Enumerations", function () {
+  var Example;
+
+  before(async function () {
+    this.timeout(10000);
+
+    Example = await util.createExample();
+  });
+
+  it("Sets up enumeration objects on contracts", async function () {
+    const expected = {
+      ExampleZero: 0,
+      ExampleOne: 1,
+      ExampleTwo: 2
+    };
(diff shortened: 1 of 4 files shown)


haltman-at:
OK, went back and added a test of this!

gnidan (approved):
Looks good! This will be handy!

PR #3554, opened 2020-11-22
Change default @truffle/db adapter to `sqlite`

gnidan:
Since this will be vastly more performant than `fs`


fainashalts (approved):
LGTM!

PR #3556, opened 2020-11-22
Enable querying @truffle/db ABIs for individual entries

gnidan:
Use @truffle/abi-utils to normalize the ABI JSON and define corresponding types for querying in @truffle/db

### DIFF
--- a/packages/db/src/resources/contracts.ts
+++ b/packages/db/src/resources/contracts.ts
@@ -2,6 +2,8 @@ import { logger } from "@truffle/db/logger";
 const debug = logger("db:definitions:contracts");
 
 import gql from "graphql-tag";
+import { pascalCase } from "change-case";
+import { normalize } from "@truffle/abi-utils";
 
 import { Definition } from "./types";
 
@@ -28,6 +30,7 @@ export const contracts: Definition<"contracts"> = {
 
     type ABI {
       json: String!
+      entries: [Entry]
     }
 
     input ContractInput {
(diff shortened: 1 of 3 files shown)


packages/db/src/resources/contracts.ts
```diff
@@ -106,6 +166,18 @@ export const contracts: Definition<"contracts"> = {
...
+    ABI: {
+      entries: {
+        resolve({ json }) {
+          return normalize(JSON.parse(json));
+        }
+      }
+    },
+    Entry: {
+      __resolveType(obj) {
+        return `${pascalCase(obj.type)}Entry`;
```
haltman-at, line 179: What's going on here? I'm not sure I understand.

haltman-at (commented):
Seems basically OK to me.  Notes/questions:

1. Why does there not seem to be anything matching the possible values of the `type` field to the various types?
2. (Minor) Should these be called "entries" or "items"?  I've always called them "entries", but it seems everyone else calls them "items".

Also note obviously this will have to be updated once errors are added to the ABI. :)

packages/db/src/resources/contracts.ts
gnidan: Since `Entry` is polymorphic, GraphQL needs a discriminator, so we use `type` to construct the name for the correct Entry type (`FunctionEntry`, etc.)

gnidan:
> (Minor) Should these be called "entries" or "items"? I've always called them "entries", but it seems everyone else calls them "items".

well, now @truffle/abi-utils calls them "entries", so...

packages/db/src/resources/contracts.ts
haltman-at: I see! Kind of an odd way of doing it, but I guess that works!  So internally, the discriminator will be `"FunctionEntry"` rather than `"function"`, huh?  I guess that works.

haltman-at (approved):
Approving this assuming I've understood it correctly. :)

gnidan:
Thanks @haltman-at!

PR #3559, opened 2020-11-24
Internal improvement: Make source-fetcher throw on invalid network

haltman-at:
**Note: This PR is a breaking change to `@truffle/source-fetcher`**.

I realized that source-fetcher probably ought to throw if you give it an invalid network ID.  So now it does.  The `isNetworkValid` method is gone, because it's now unnecessary.  Note obviously this required appropriate modifications to `external.js`.

### DIFF
--- a/packages/core/lib/debug/external.js
+++ b/packages/core/lib/debug/external.js
@@ -5,6 +5,7 @@ const Web3 = require("web3");
 
 const Codec = require("@truffle/codec");
 const Fetchers = require("@truffle/source-fetcher").default;
+const {InvalidNetworkError} = require("@truffle/source-fetcher");
 
 const {DebugCompiler} = require("./compiler");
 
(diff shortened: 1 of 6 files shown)


packages/core/lib/debug/external.js
```diff
@@ -45,27 +33,32 @@ class DebugExternalHandler {
         }
       }
     } else {
-      sortedFetchers = allFetchers;
+      sortedFetchers = Fetchers;
     }
-    //to get the final list, we'll filter out ones that don't support this
+    const networkId = await new Web3(this.config.provider).eth.net.getId(); //note: this is a number
```
gnidan, line 38: BTW you don't need to do this. This is available on the config:
```suggestion
    const networkId = this.config.network_id; //note: this is a number
```

gnidan (changes requested):
Looks good apart from this one thing

packages/core/lib/debug/external.js
haltman-at: Oh whoa. We've always been able to do this...?

gnidan: Yeah, found it yesterday!

haltman-at:
Actually wait I want to make another change to this too, you'll see it in a moment...

haltman-at:
OK, I fixed that, and also I exported the error so I can do a proper error check with `instanceof` instead of relying on `error.name`.  For some reason I got the mistaken idea that would be somehow messy so didn't do that the first time around, but it's not.

gnidan (approved):
Looks good!

PR #3560, opened 2020-11-24
Enhancement: Add rudimentary Etherscan Vyper support to source-fetcher

haltman-at:
This PR adds rudimentary support for Vyper contracts on Etherscan to source-fetcher.  Because these have no associated settings, it's pretty straightforward; the source code is just the given source code (no need to process that), and for the compiler version, we just strip off the initial `vyper:`, which I'm taking to be a prefix added by Etherscan, rather than a legitimate part of the version string.

### DIFF
--- a/packages/source-fetcher/lib/etherscan.ts
+++ b/packages/source-fetcher/lib/etherscan.ts
@@ -209,7 +209,8 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
       },
       options: {
         language: "Vyper",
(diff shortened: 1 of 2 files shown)


gnidan (approved):

haltman-at:
Note: Vyper does currently support an EVM version option... but it looks like we currently don't support using it?  So I guess we'll ignore the EVM version info for now? :-/

haltman-at:
OK, you know what, I decided to go and add that in anyway, even if we don't currently support using it...

PR #3561, opened 2020-11-24
Internal improvement: Fix debugger test debug module names

haltman-at:
The debug module names in the debugger test didn't begin with `debugger`, and it was bugging me, so I changed it.

...also the one in `yul.js` was misnamed.


gnidan (approved):
Nice

PR #3564, opened 2020-11-24
Internal improvement: Add fetcher name to InvalidNetworkError

haltman-at:
Real minor thing, quick addendum to #3559, I realized this info should probably also go in the error.


gnidan (approved):
totes

PR #3565, opened 2020-11-24
Enhancement: Have debugger/decoder make use of sources info in compilations

haltman-at:
Now that #3550 is merged, this is possible!  **Edit: This PR is now a breaking change to `@truffle/codec`.**

This PR does two things.  The first is that it makes it so that `shimCompilations` (and the new `shimCompilation`) make use of the `sources` information on a compilation, if it is present.  Note that on `shimContracts`, I've added this as an optional argument on the *end* in order to prevent a breaking change.  If `inputSources` is passed to that, it will use that for the sources information, instead of trying to dig it out of `contracts` and `files`.

The second thing is that in `debug-utils`, it adds one more check for whether the debugger CLI should recompile: If there's a gap between the index of the last user source and the first generated source, that will now trigger a recompile.  (If there are any generated sources, that is, obviously.  Otherwise this has no effect.)

Oh also I changed language tags in codec and debugger to lowercase so that they're compatible with the ones from compile, initially this PR accidentally broke syntax highlighting. :P

### DIFF
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -9,22 +9,36 @@ import {
 } from "@truffle/contract-schema/spec";
 import {
   CompiledContract,
-  Compilation as CompilerCompilation
+  Compilation as CompilerCompilation,
+  Source as CompilerSource
 } from "@truffle/compile-common";
-import { Compilation, Contract, Source } from "./types";
+import {Compilation, Contract, Source} from "./types";
 
 export function shimCompilations(
   inputCompilations: CompilerCompilation[],
   shimmedCompilationIdPrefix = "shimmedcompilation"
 ): Compilation[] {
-  return inputCompilations.map(
-    ({ contracts, sourceIndexes }, compilationIndex) =>
-      shimContracts(
-        contracts,
-        sourceIndexes,
-        `${shimmedCompilationIdPrefix}Number(${compilationIndex})`
-      )
+  return inputCompilations.map((compilation, compilationIndex) =>
+    shimCompilation(
+      compilation,
+      `${shimmedCompilationIdPrefix}Number(${compilationIndex})`
+    )
   );
+}
+
+export function shimCompilation(
+  inputCompilation: CompilerCompilation,
+  shimmedCompilationId = "shimmedcompilation"
+): Compilation {
+  return {
+    ...shimContracts(
+      inputCompilation.contracts,
+      inputCompilation.sourceIndexes,
+      shimmedCompilationId,
+      inputCompilation.sources
+    ),
+    compiler: inputCompilation.compiler
+  };
 }
 
 /**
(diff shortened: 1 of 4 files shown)


haltman-at:
Oops, seems I messed something up, I'll have to fix this I guess.

haltman-at:
OK, should be fixed now.

packages/codec/lib/compilations/utils.ts
```diff
@@ -42,12 +56,16 @@ export function shimArtifacts(
...
+ * Note: if you pass in inputSources, sources will not have
+ * compiler set, so you should set that up separately, as in
+ * shimCompilation().
  */
 export function shimContracts(
   artifacts: (Artifact | CompiledContract)[],
   files?: string[],
-  shimmedCompilationId: string = "shimmedcompilation"
+  shimmedCompilationId: string = "shimmedcompilation",
+  inputSources?: CompilerSource[]
```
gnidan, line 68: So is ≤one of `inputSources`/`files` used? Should we make the interface reflect this?

gnidan (commented):
Couple quick comments/questions before I approve :)

packages/codec/lib/compilations/utils.ts
```diff
@@ -9,24 +9,38 @@ import {
 } from "@truffle/contract-schema/spec";
 import {
   CompiledContract,
-  Compilation as CompilerCompilation
+  Compilation as CompilerCompilation,
+  Source as CompilerSource
```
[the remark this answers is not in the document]
haltman-at: Oh, good idea!

packages/codec/lib/compilations/utils.ts
haltman-at: Yeah, if `inputSources` is given then `files` is ignored.  Which now that you mention it does make for an awkward interface, yes.  Hm.  Not sure what to do here.  Any ideas?

In the worst case I can just add a comment, but that's obviously not ideal...

gnidan: Well, you could make it `options: { files, etc } | { inputSources, etc }`

haltman-at: Yeah, I was hoping to avoid a breaking change.  Although IDK, this method, while exported, is pretty internal... maybe it's OK if we break it?  Or even unexport it? :P  If that's OK then I could totally do it that way.  Or we can just accept another breaking change to codec. :P

haltman-at: OK, seems we're breaking it!

packages/codec/lib/compilations/utils.ts
```diff
@@ -138,17 +159,34 @@ export function shimContracts(
...
+      for (let index in generatedSources) {
+        if (index in sources) {
+          unreliableSourceOrder = true;
+        }
       }
-    }
-    for (let index in deployedGeneratedSources) {
-      if (index in sources) {
-        unreliableSourceOrder = true;
+      for (let index in deployedGeneratedSources) {
```
gnidan, line 186: Oh btw I keep forgetting to let you know that `for (const` works just fine here

gnidan (approved):
Looks good!

packages/codec/lib/compilations/utils.ts
haltman-at: Oh yeah I've been a bit inconsistent with that; I have `for (const` in a lot of places, but not everywhere.  Well, I won't change that here right now, but can try to remember for the future...

I think maybe what confused me is that when you do a C-style for loop, then you do need to use `let`?  I think?

Issue #3562, opened 2020-11-24
Can't set EVM version (hardfork) version to use with Vyper


haltman-at:
## Issue

The Vyper compiler [has an option to pass in the version of the EVM to use](https://vyper.readthedocs.io/en/stable/compiling-a-contract.html#setting-the-target-evm-version), but as best I can tell we don't have any way to pass in that option like we do with solc, where you can put it in the solc settings in the config.

Note, although this is also a separate thing: We may want to start using the Vyper JSON interface?  That may not work with older versions of Vyper; as best I can tell it was only added in v0.1.0-beta.13.

## Environment

* Truffle version (`truffle version`): 5.1.54

haltman-at:
This is now done in #3568.

PR #3568, opened 2020-11-25
Enhancement: Add evmVersion option to compile-vyper

haltman-at:
Addresses #3562.  Doesn't switch to using the JSON interface as uhhh that is not actually so simple.  But just adding this is. :)

~~Warning: I have not tested this PR **at all**!  I am adding this purely based on the documentation, so, uh, yeah...~~ I have now tested it, it works. :)


gnidan (changes requested):
This is worth testing :) Preferably with an automated test, if there's a good way to do it, but at least I'd say this warrants verifying manually before we merge!

haltman-at:
Hm, do we have a way for automated testing of Vyper at the moment?  I'm not familiar with our compiler tests...

gnidan:
> Hm, do we have a way for automated testing of Vyper at the moment? I'm not familiar with our compiler tests...

There is this [test_compiler.js](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-vyper/test/test_compiler.js). I'd think it'd be pretty easy to test this stuff if you're familiar with an example bytecode difference between EVM versions

haltman-at:
I see, thanks!  OK, I can see about adding a test there.

haltman-at:
OK, I added a test!  I was never able to get Vyper working locally (at least not previously, didn't try right now) so I haven't run this test locally or done any sort of manual test -- hopefully it passes on CI!

haltman-at:
OK, attempt number 2...

haltman-at:
Hm, this could use another test though.  Let me add one more...

haltman-at:
Welp, OK, additional test is failing... :-/

haltman-at:
Oops, the second test just failed because of a copypaste error...

haltman-at:
Note: Now that I've added this single-quote check, wondering if we should use a `TruffleError` here, but I just used a basic `Error` for simplicity... `TruffleError` s are annoying...

gnidan (approved):

PR #3571, opened 2020-12-01
Bug fix: Better handling of Vyper source maps (including putting them in the right place in the artifact!)

haltman-at:
This might be a breaking change to `compile-vyper`?  Or to something? (Edit: We've agreed it's not since it never worked properly in the first place!)

So this has several parts to it:

1. Biggest change -- Vyper source maps will now go under `deployedSourceMap` in the artifact, not `sourceMap`, because they're source maps for the deployed bytecode!  (As best I can tell, Vyper doesn't support source maps for constructors, for whatever reason.)  Note that Vyper source maps are a different format from Solidity's, so I'm unsure this is *really* the right place, but, um, it's an improvement at least I guess...?

2. The decoder now checks that any source maps it's attempting to use are Solidity source maps, not Vyper source maps.

3. The debugger now attempts to convert Vyper source maps to Solidity ones.  Note that I'm doing this the lazy way -- just using the field which contains the Solidity-equivalent source map, which is missing on older versions; I'm not doing any actual conversion.

4. Finally, Vyper seems to indicate unmapped instructions with `-1:-1:0` instead of Solidity's `-1:-1:-1`, so I had `solidity-utils` do a conversion here, where if start and length are both `-1` then it'll put `-1` for the file, instead of whatever's present.

Also `prettier` made a bunch of changes which is annoying. :-/

### DIFF
--- a/packages/compile-vyper/index.js
+++ b/packages/compile-vyper/index.js
@@ -38,7 +38,7 @@ function execVyper(options, sourcePath, callback) {
   }
   const command = `vyper -f ${formats.join(",")} ${sourcePath}`;
 
-  exec(command, {maxBuffer: 600 * 1024}, function (err, stdout, stderr) {
+  exec(command, { maxBuffer: 600 * 1024 }, function (err, stdout, stderr) {
     if (err)
       return callback(
         `${stderr}\n${colors.red(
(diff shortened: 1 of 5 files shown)


gnidan (approved):

fainashalts (approved):
Cool! I wonder what else we'll find once we're all able to use vyper regularly on our machines. :)

haltman-at:
OK, I'm merging this despite the red X, because for some reason GitHub is counting cancelled jobs as a reason to fail this, even though on the rerun everything passed.

PR #3572, opened 2020-12-01
Turn on source maps in Vyper w/o an option

haltman-at:
As requested by @gnidan.  Gets rid of the sourceMap option for compiling Vyper contracts, in favor of just always getting the source map.  Tests have also been adjusted appropriately.

### DIFF
--- a/packages/compile-vyper/test/test_compiler.js
+++ b/packages/compile-vyper/test/test_compiler.js
@@ -85,7 +85,7 @@ describe("vyper compiler", function () {
     contracts.forEach((contract, index) => {
       assert(
         contract.deployedSourceMap,
-        `source map have to not be empty. ${index + 1}`
+        `source map has to not be empty. ${index + 1}`
       );
     });
   });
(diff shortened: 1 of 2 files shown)


packages/compile-vyper/test/test_compiler.js
```diff
@@ -79,17 +79,19 @@ describe("vyper compiler", function () {
...
-          }
-        }
-      }
+  it("outputs source maps", async () => {
+    const { compilations } = await Compile.all(config);
+    const { contracts } = compilations[0];
+    contracts.forEach((contract, index) => {
+      assert(
+        contract.deployedSourceMap,
+        `source map have to not be empty. ${index + 1}`
```
fainashalts, line 88:
```suggestion
        `source map has to not be empty. ${index + 1}`
```

haltman-at: Heh, yeah, may as well fix that!

gnidan (approved):
thanks for doing this!

haltman-at:
> thanks for doing this!

Of course! :)

PR #3582, opened 2020-12-04
Add specializations to source-fetcher

haltman-at:
Adds a `specializations` field to the output of source-fetcher.  This contains the libraries information and (when retrieving from Etherscan only) the constructor arguments information.  When retrieving from Sourcify, constructor arguments will not be present.  Also obviously libraries info is omitted for Vyper.

Constructor arguments is given as a hex string, of course, but note that I did *not* put a `"0x"` in front, for consistency with how we do it in the debugger (and also easy concatenation).

Libraries are given in a form understandable to solc.  When dealing with JSON (from Etherscan or Sourcify), well, it's right in the JSON.  When dealing with non-JSON Etherscan, we have to parse the `Library` field.  Fortunately that's just a semicolon-delimited list, with the individual entries being `name:address` pairs (although oddly the addresses lack a `"0x"` in front and are not checksummed... we handle that though).  Then, having constructed a `libraries` object containing this information, we make our libraries settings `{"": libraries}`, because using the empty string as a key tells Solc to use these libraries settings for *all* contracts.

Note I didn't really test this PR.

### DIFF
--- a/packages/source-fetcher/lib/etherscan.ts
+++ b/packages/source-fetcher/lib/etherscan.ts
@@ -166,7 +166,11 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
       options: {
         language: "Solidity",
         version: result.CompilerVersion,
-        settings: this.extractSettings(result)
+        settings: this.extractSettings(result),
+        specializations: {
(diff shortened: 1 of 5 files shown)


gnidan (approved):
Looks good

PR #3583, opened 2020-12-04
Enhancement: Add rudimentary support for debugging Vyper (experimental); add support for vyper-json

haltman-at:
(**A bunch has been added, see later comments**)

This PR adds rudimentary, experimental support for Vyper in Truffle Debugger.  It allows stepping through Vyper contracts and placing breakpoints.  Stacktraces may roughly work -- I haven't tested them too extensively -- but may be missing lots of information.  We apply Python syntax highlighting to Vyper sources, as there's no Vyper syntax highlighting module for `highlight.js`.

This PR does have one annoying side effect, I'll get to that below.

Vyper debugging has a number of limitations:

1. Only Vyper runtime bytecode may be debugged, not Vyper constructor bytecode.  This is a limitation of Vyper; the Vyper compiler [currently does not provide sourcemaps for constructor bytecode](https://github.com/vyperlang/vyper/issues/1998).

2. There is absolutely no support for tracking of variables or decoding of their values.  The point of this PR was to support what I could quickly and easily; adding that would be a whole project.  Note that doing this would also require making use of Vyper ASTs, which we currently don't store (we'd have to switch to the JSON interface to get those).

3. Even decoding of return values is unsupported; you might think *that* at least should work, since that can be done via the ABI, but Truffle Debugger currently uses the AST to find some of this information.  This could be fixed, but I didn't want to rework that part of the debugger right now; again, the point of this PR was to add what could be done easily.  Note, however, that decoding of revert messages *does* work.

4. Debugging Vyper versions prior to `0.1.0-beta.12` is not supported.  While Vyper source maps were added in `0.1.0-beta.7`, they did not include Solidity-style sourcemaps as a field until `0.1.0-beta.12`, and, again, the idea of this PR was to do things the easy way.

5. Line-based breakpoints work, but placing node-based breakpoints with `b` does not.  This is because currently Truffle Debugger relies on the AST to track these; I could have reworked it to do it based on source ranges instead, which would have allowed these to work, but that would be a breaking change and again the idea of this PR is to keep things simple.

6. Downloading and compiling Vyper contracts with `--fetch-external` is not supported.  Since we currently have no mechanism to compile with a specified version of Vyper, this just isn't really possible at the moment.

7. Stepping does not skip over variable declarations or anything like that.  Similarly there's no moving-down of breakpoints or anything like that.

Still, all that said -- it's rudimentary, experimental, Vyper debugging!  That's certainly something, don't you think? :)

I want to break down in a little more detail what was necessary for this to work.  I've helpfully split this PR up into a number of individual commits which you can also see below. :)  I'll go through these in order:

1. Allow CLI debugger to compile non-Solidity sources -- this is a pretty straightforward matter of using `workflow-compile` instead of `compile-solidity`.  When a specific list of sources is given, however, we do use `compile-solidity`, because A. I didn't see how to do that with `workflow-compile`, and B. only `external.js` uses that option and it's not going to be compiling any Vyper for the reasons stated above.

2. Parse ABI JSON in `compile-vyper` -- so, uh, `compile-vyper` was returning ABIs as strings rather than objects.  Now we parse them there as appropriate.  (This bug didn't affect artifacts due to the contract schema normalization.)

3. Allow stopping on no-node source ranges of positive length -- previously, the debugger wouldn't stop on a source range if there was no associated AST node.  Now instead of checking for a node, it checks whether the source range has positive length.

4. Add syntax highlighting for Vyper -- self explanatory; as mentioned above, we highlight it as if it were Python.

5. Split compile-vyper return into one compilation per file -- currently, since we're not using the JSON interface, `compile-vyper` compiles each Vyper file separately.  This means each one should be treated as a separate compilation to work with the debugger.  This commit does that.  Of course, once we start using the JSON interface, then it will be one big compilation again.

This commit does have one annoying side-effect: Compiling a large Vyper project will result in a *lot* of `vyper: <version>` printed to the screen.  Uh, I guess we can always just recommend that Vyper users use `--quiet` when compiling? :-/  Or I could start work on making use of the JSON interface... (although note, we'll still need the old way available to handle Vyper versions prior to `0.1.0-beta.13`).  (...or I could hack something together to make it work without doing that, where I put it all back in one big compilation but I hack up the source maps to make it work anyway, but, uh, that's going to be real hacky....)

6. Allow passing compiler to shimContracts -- OK, we don't really need to discuss this one.

7. Add vyper language inference to shimCompilations() -- since we don't have an AST at the moment, we just assume the language is Vyper if the compiler is vyper.

8. Set Vyper contracts to have primary source index 0 for now -- yeah, this one's quite a hack, but what else was I supposed to do...?  OK, notionally I could go to the extra effort of going by the source map, but c'mon, it's always going to be 0 for now.  I'd rather wait on that sort of thing until we have AST.

9. Allow placing breakpoints when no AST -- Previously, if there was no AST, the breakpoint resolver would say, hey, there's nowhere to *put* this breakpoint!  Now, if there's no AST, it says, OK, I just won't attempt to modify this one.

10. Restrict Solidity builtins to Solidity -- We already disabled them in Yul, now we properly restrict them to just Solidity.  It looked pretty odd having them in Vyper.  This does raise the question of whether we should find some other way to display this info, though, since not all of it has other displays in the debugger.

11. Skip decoding attempts when in Vyper -- Just puts some messaging in `interpreter.js` so that users understand that decoding isn't supported.  It also skips the attempt to decode return values (without a corresponding message) since, uh, yeah, that wasn't working as I mentioned above.

So that's that!  Let me know what you think, and hopefully the annoying side effect doesn't sink this...?  I can also try to solve it by one of the means above, as mentioned...


haltman-at:
Oops, I didn't think to update the `compile-vyper` tests.  OK, updated now.  I had to update them for two things:

1. Multiple compilations.  Um, for the purpose of simplicity I handled this in some cases by glomming together the compilations in the test, yes the very thing I complained about elsewhere, but these tests are weird...

2. Parsed ABI -- the tests actually assume the ABIs are strings!  Now I assume they're parsed.

haltman-at:
OK, I'll have to figure out why this is still failing...

haltman-at:
Note: I just redid a bunch of this PR.  I will write an explanation Monday.  Anyway hopefully now the tests pass.

haltman-at:
ALSO: I just realized I'm going to need to make some changes to `debugger` and such to account for the fact that Vyper sources may now have ASTs!  So, uh, yeah this isn't ready quite yet.

haltman-at:
OK, so I made several changes to this PR; most of them are fairly minor stuff to improve compatibility.  The big one though is that I've switched `compile-vyper` to use `vyper-json` when it's available.

Note that this means that Vyper contracts are now *not* always compiled alone, and I've changed `codec/lib/compilations/utils.ts` to reflect this.  Interesting note: Vyper ASTs have `src` properties the same as Solidity ASTs, so the bit that makes use of that doesn't need to change!  Note that if there is no AST it assumes it *is* compiled alone and says the source index is `0`.  Similarly with source maps -- `vyper-json` returns Solidity-style source maps, so if there's an old Vyper-style source map without the Solidity-equivalent as a field, well, we can say the source index is `0`.  (Vyper old-style source maps all assume that there's only a single source file anyway; it's not like they have this information encoded somewhere else.)

Anyway, let's talk about getting `vyper-json` working!

This is, um, a lot of copypaste from `compile-solidity` at the moment.  Will likely want to factor this into `compile-common` in a later commit, but for now I just did what it took to get it working.

So now, `compile-vyper` uses `vyper-json` if available; this is mostly handled in a separate file, `vyper-json.js`, as opposed to in `index.js`.  It uses basically the same flow as `compile-solidity`'s `run.js`, but some of the details are different.

Like, going through... first we read the sources.  (That's not in `run.js`, because there the sources are read beforehand.)  Then we do `collectSources`... I'm not clear on how necessary this step is, but I left it in.

Actually, a note here -- I left in all the `compilationTargets` stuff, and even added it to the non-JSON case in `index.js`.  Only afterward did I learn that `options.compilationTargets` won't exist unless you invoke the profiler and set it up.  Oops.  Well -- I can either add profiler stuff, tear that stuff out, or maybe just leave it in with a comment that this needs to be set up later. :)

Anyway, then we've got `prepareCompilerInput`.  This step is substantially different, so let's come back to it later.

Then we invoke the compiler.  That's the same.  Then we check for errors.  That's mostly the same.  (Additional copypaste warning here: I even copied in the file `compileerror.js` from `compile-solidity`... that'll have to get factored into `compile-common`, right?)

Then -- here's a new step -- we fixup the paths.  Turns out that Vyper returns file paths with the initial `/` stripped off, so we add that back on.

Then finally we do `processSources`, `processContracts`, and `processAllSources`, which is largely the same, though I made a fix or two; if we factor this I may make some further corrections/simplifications here.

OK, so what's different, other than the path fixup?  Let's start with the small things:

* `detectErrors` -- Old versions of Vyper don't have `formattedMessage`, so we fall back to `message` when it doesn't exist.  Note this change can be applied to Solidity as well so that wouldn't stop a factorization.
* `processAllSources` -- There's no `legacyAST`, obviously.  Could be factored but I'd rather leave it separate.
* `processContracts` -- A lot less fields!  And the compiler name is `"vyper"`, not `"solc"`. :P  Again, could be factored, but whatever.
* A parameter renaming -- the variable `solcVersion` has been renamed to just `version` throughout.  That should probably be applied in general. :)

OK, so now let's get to the big thing: `prepareCompilerInput`.  This is substantially different!  I don't expect *this* to be factored...

So, breaking it down:

1. The language is `"Vyper"` now, obviously.
2. `sources` and `prepareSources` are exactly the same
3. `settings` is a lot simpler, obviously -- I'll get to to `outputSelection` in a bit
4. `interfaces` is added; Solidity doesn't have this.  I've made it so that the user can set this in their config.
5. `outputSelection` is added *outside* of `settings`.  See, older versions of Vyper to put it outside; newer ones put it inside.  Fortunately, it seems that it's OK to put it in the wrong place in addition to the right one; Vyper (surprisingly, given its otherwise delicate stomach) doesn't vomit on this!

OK, that's not so different, but how do `outputSelection` and `prepareOutputSelection` differ?

1. There's a lot less output to select, obviously.  Note ASTs are in (we finally have Vyper ASTs!  They're pretty different from the Solidity ones!), but constructor sourcemaps are out
2. The overall format is simpler; Vyper also accepts the more complex Solidity format, but I figured I'd keep it simple
3. What is selected is version-sensitive!  Unlike Solidity, Vyper will complain if you ask for output it doesn't support, so we only include `userdoc` and `devdoc` on suffiicently recent version.

Actually, I made this change to the non-JSON `index.js` as well, so it will only ask for source maps on sufficiently recent versions, because those haven't always existed either!  But every version that supports JSON supports them, so the JSON version doesn't need to check for that.

Anyway, yeah, that's done!  We have `vyper-json` support now!

So why is the Geth build still failing?  Uh, I have no idea.  I'm going to have to investigate that...

haltman-at:
OK, found the problem that was making CI fail -- it was that we can't always assume the existence the `compiler` field, apparently...

haltman-at:
OK, fixed up some stuff, hooked up the profiler (sort of), and factored out some stuff into `compile-common` after discussing with @gnidan!

haltman-at:
OK, I added a commit to make it so that `truffle compile` won't print out multiple of the same version, so the annoying side-effect is no longer a problem! :)

haltman-at:
OK, I've done even more here!  Let me explain both of these new commits.

Commit 1: This isn't *directly* related to the point of this PR, but I ended up doing it anyway.  This commit is an internal improvement.  Basically, the default pattern in `contract-sources` looked for `.vy` and `.sol` files, and then `compile-solidity` would exclude things that were `.vy`.  I've changed it to a braoder array of file types in `contract-sources`, and to have `compile-solidity` restrict specifically to `.sol`.

This doesn't actually affect `compile-vyper` at all, since it doesn't use the default pattern; it passes its own.  Still, there you go.  I broadened the file types to add `.v.py` and `.vyper.py` (since `compile-vyper` already accepts those), to add `.json` (for JSON interfaces), and to add `.yul` (because maybe we'll support it in the future :) ).

You may ask: Why not just get rid of the file extension requirement entirely?  Well, the problem is I couldn't find a way to do that while *also* excluding directories.  At least, not with the library we're using.  Since I didn't want to rewrite that whole file, I settled for just broadening the set of file types.

Commit 2: This is the more important one -- it gets interfaces working!

If you were using non-JSON Vyper, interfaces would already work, as non-JSON Vyper will search the file system.  But JSON Vyper won't do that; any interfaces imported need to be explicitly passed in.

So, I expanded the search for files to include `.json` files.  What we do with this broader list of files differs in the JSON and non-JSON cases:

1. In the non-JSON case, we just exclude these from the list of things to compile.  We don't need to do anything else; the compiler will handle the rest by searching the file system for them.
2. In the JSON case, we split our `sources` into `properSources` and `interfaces`.  We then pass `properSources` as the sources and `interfaces` as the `interfaces`.  (We're no longer reading `interfaces` out of the config... that's not how that works, oy!)

Note that, while we're not doing this here, `interfaces` can also be used to pass in files that are being imported but not compiled.  So, if we ever have Vyper import parsing in the future, we may be able to take advantage of that when `necessary` is called, by separating our sources into the ones passed by the user, which would go under `sources`, and the ones flagged as necessary by the import parser, which would go under `interfaces`.  That'd be pretty cool!  But it's certainly not happening now...

haltman-at:
(Note that the test I added is an implicit one; I didn't add any tests to the test file, the test is that the existing tests pass with this new contract added!)

gnidan (approved):
Looks good, Harry!

haltman-at:
Woo, OK!  Had to resolve a merge conflict but will merge as soon as CI passes. :D

Issue #2843, opened 2020-02-21
deployer.deploy() not working with arguments and options


pizzarob:
I have a migration file as follows. I am trying to deploy a new instance of an already deployed contract using the `overwrite` parameter, however truffle gives me the error:

`Error: Invalid number of parameters for "undefined". Got 2 expected 1! `

Apparently trying to use the options object as an argument.

```
module.exports = async function (deployer, network, accounts) {
  const contractAInstance = await A.deployed();
  const contractBInstance = await B.deployed();

  await contractBInstance.setOperator(contractAInstance.address, false);
  await deployer.deploy(A, contractBInstance.address, { overwrite: true });
};

**Comment by pizza-r0b** [comment]
**Update**
Adding `from` to the options object makes the error go away

`  await deployer.deploy(A, contractBInstance.address, { from: accounts[0] overwrite: true });`

However I'd expect a new address to be given when asking for the deployed contract, but the new instance gives the same address as the old.

```js
await deployer.deploy(A, contractBInstance.address, { from: accounts[0] overwrite: true });
const newInstance = await A.deployed();
// newInstance.address is the same as the instance before using overwrite: true
```

CruzMolina:
Hey @pizza-r0b , it looks like you're trying to re-migrate a contract. What are you attempting to do with the `overwrite` property? I don't believe it's a valid truffle-related option. Have you tried using `truffle migrate --reset`? https://www.trufflesuite.com/docs/truffle/getting-started/running-migrations#command.

Can you share a repo with reproduction steps?

pizza-r0b:
@CruzMolina I am trying to re-deploy a new instance of the contract.

I think I may be using overwrite incorrectly. The docs don't do a great job explaining this.

>Additionally, the last argument is an optional object that can include the key named overwrite as well as other transaction parameters such as gas and from. If overwrite is set to false, the deployer won't deploy this contract if one has already been deployed.

But I imagine I can use `--reset` and specify `overwrite: false` on all the contracts in which I do not want to be overwritten.

CruzMolina:
Ah, I see. Yeah I believe there's two bugs here, one when using the `overwrite` (it should be acceptable to only pass that as an extra param, not needing other args like `from`), and another with the `address` not being properly reset on a new deployed contract instance.

eggplantzzz:
What is the `overwrite` option? Can anyone link to some docs?

CruzMolina:
[...]

PR #3586, opened 2020-12-07
Bug fix: Correct `is_tx_params` method

eggplantzzz:
Currently the `is_tx_params` method is used to determine whether an argument to a function is the tx params. This function checks the keys of the object provided and returns once it finds that one of them matches the whitelisted fields. This PR adds the `overwrite` key to the whitelist. ~and will also accept an empty object. Accepting an empty object does seem a bit strange to me. It seems like it would make user errors more likely or cause misunderstandings but I suppose it could be better UX. This fix causes `@truffle/contract` to check all of the keys and make sure they match some term from the whitelist.~

related to https://github.com/trufflesuite/truffle/issues/2843

### DIFF
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -37,15 +37,13 @@ const Utils = {
       "overwrite"
     ]);
 
-    // if an unrecognized parameter is present, then assume
-    // it is the last arg and not the tx params
     for (let fieldName of Object.keys(val)) {
-      if (!allowedFields.has(fieldName)) {
-        return false;
+      if (allowedFields.has(fieldName)) {
+        return true;
       }
     }
 
-    return true;
+    return false;
   },
 
   decodeLogs(_logs, isSingle) {
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -25,22 +25,27 @@ const Utils = {
     if (!Utils.is_object(val)) return false;
     if (Utils.is_big_number(val)) return false;
 
-    const allowedFields = {
-      from: true,
-      to: true,
-      gas: true,
-      gasPrice: true,
-      value: true,
-      data: true,
-      nonce: true,
-      privateFor: true
-    };
+    const allowedFields = new Set([
+      "from",
+      "to",
+      "gas",
+      "gasPrice",
+      "value",
+      "data",
+      "nonce",
+      "privateFor",
+      "overwrite"
+    ]);
 
+    // if an unrecognized parameter is present, then assume
+    // it is the last arg and not the tx params
     for (let fieldName of Object.keys(val)) {
-      if (allowedFields[fieldName]) return true;
+      if (!allowedFields.has(fieldName)) {
+        return false;
+      }
     }
 
-    return false;
+    return true;
   },
 
   decodeLogs(_logs, isSingle) {
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -36,14 +36,7 @@ const Utils = {
   isTxParams(val) {
     if (!Utils.is_object(val)) return false;
     if (Utils.is_big_number(val)) return false;
-
-    for (let fieldName of Object.keys(val)) {
-      if (allowedTxParams.has(fieldName)) {
-        return true;
-      }
-    }
-
-    return false;
+    return Object.keys(val).some(fieldName => allowedTxParams.has(fieldName));
   },
 
   decodeLogs(_logs, isSingle) {
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -21,11 +21,11 @@ const Utils = {
     return web3Utils.isBN(val) || web3Utils.isBigNumber(val);
   },
 
-  is_tx_params(val) {
+  isTxParams(val) {
     if (!Utils.is_object(val)) return false;
     if (Utils.is_big_number(val)) return false;
 
-    const allowed_fields = {
+    const allowedFields = {
       from: true,
       to: true,
       gas: true,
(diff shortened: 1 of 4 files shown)


eggplantzzz:
I think I also will write a couple of tests for this fix.

frangio:
@eggplantzzz I'm not sure this is the right approach because it's potentially a breaking change. There may be users forwarding params objects with additional keys and after upgrading it will stop working in those cases.

I think the empty object can just be a special case added on top of the previous approach.

eggplantzzz:
Yes I agree, after making the initial PR I thought the same thing. It does seem sufficient to just add `overwrite` to the whitelist. I'll go ahead and add support for an empty object as well.

frangio:
What kind of misunderstanding do you have in mind? I was thinking that adding the empty object was harmless because it would never appear in argument position.

eggplantzzz:
Well there is really no purpose for it as it does nothing. I guess I had a vague, small fear that it could possibly be something leading to confusion. I'm not sure if it is warranted.

packages/contract/lib/utils/index.js
```diff
@@ -21,23 +21,26 @@ const Utils = {
...
+      "gas",
+      "gasPrice",
+      "value",
+      "data",
+      "nonce",
+      "privateFor",
+      "overwrite"
+    ]);
+
+    for (let fieldName of Object.keys(val)) {
```
haltman-at, line 40: This `for` loop could be done more succinctly with `Object.keys(val).some(...)`.

haltman-at (commented):
I'm confused about one thing... the PR description says it will now accept empty objects, but the code looks like that's not the case?

eggplantzzz:
@haltman-at Yeah I was sort of discussing whether we should accept this above. I was kind of waiting to see what shakes out of that.

packages/contract/lib/utils/index.js
eggplantzzz: Yeah that sounds like a good edit!

packages/contract/lib/utils/index.js
```diff
@@ -21,23 +21,26 @@ const Utils = {
...
+    const allowedFields = new Set([
+      "from",
+      "to",
+      "gas",
+      "gasPrice",
+      "value",
+      "data",
+      "nonce",
+      "privateFor",
+      "overwrite"
```
davidmurdoch, line 37: I'm lacking some context here, so please bear with me :-)

What are `privateFor` and `overwrite` used for (since these aren't valid RPC properties)?

And should aliases like `input` (`data`) and `gasLimit` (`gas`) be considered? Also, `r`, `s`, and ,`v` are present in _raw_ transactions (though over JSON-RPC they would be serialized via RLP encoding, not as raw JSON)

davidmurdoch (changes requested):

packages/contract/lib/utils/index.js
haltman-at: I'm not @eggplantzzz, but I know `privateFor` is a quorum thing; no idea about `overwrite`.

Aliases of course would require not just allowing them but programming them to work, so that would be a bigger change...

haltman-at:
> @haltman-at Yeah I was sort of discussing whether we should accept an empty object above. I was kind of waiting to see what shakes out of that.
> 
> Do you have an opinion @haltman-at? I'll remove it from the description until it is added.

Nah, I have no real opinion on the matter.

packages/contract/lib/utils/index.js
haltman-at: Huh, looks like `overwrite` is specific to the deployer and signals it to do a reset?

eggplantzzz: Yeah, `overwrite` is Truffle-specific. You can find info [here](https://www.trufflesuite.com/docs/truffle/getting-started/running-migrations#deployer-deploy-contract-args-options-).

packages/contract/lib/utils/index.js
```diff
@@ -21,23 +21,26 @@ const Utils = {
...
-      gasPrice: true,
-      value: true,
-      data: true,
-      nonce: true,
-      privateFor: true
-    };
-
-    for (let field_name of Object.keys(val)) {
-      if (allowed_fields[field_name]) return true;
+    const allowedFields = new Set([
```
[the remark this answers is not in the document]
eggplantzzz: I have no problem with this, I'll do it.

eggplantzzz:
Ok, I'll go ahead and say that we allow an empty object. @haltman-at, can you imagine any any problems this may cause? Like if for some reason an empty object would be valid input to a Solidity function and Truffle would mark it as "transaction parameters" instead of processing it like it should?

packages/contract/lib/utils/index.js
davidmurdoch: @haltman-at i'm not sure where this stuff ends up, but the aliases _do_ work in ganache and other ethereum nodes.

haltman-at:
There are technically cases when an empty object would be a valid input to a Solidity function, but:

1. Only if you're still using Solidity 0.4.x
2. Even then, while technically legal back then, it wasn't actually *useful* and I don't know why anyone would do it,
3. Our overload handling is kind of crappy at the moment anyway so I'm not too worried about getting it perfect.

packages/contract/lib/utils/index.js
haltman-at: @davidmurdoch Interesting, I had no idea! :)

davidmurdoch:
Are there any related tests you can piggyback on for the new `overwrite` property?

eggplantzzz:
I could maybe write a couple of unit tests for this. This property is not really new, but in the past Truffle would fumble if it had none of the other tx params in the object provided since it wouldn't match.

davidmurdoch (approved):
LGTM!

PR #3588, opened 2020-12-08
Bug fix: Fix overwriting of legacyAST by AST

haltman-at:
So `contract-schema` had a mechanism for copying AST to legacyAST when the latter doesn't exist.  Problem is, the check didn't work -- it would check the `schemaVersion` on the dirty object, despite dirty objects typically not having this set -- and so AST was getting copied over legacyAST even when the latter existed.  I've replaced it with a simpler check: Now, AST is copied to legacyAST precisely when there is no legacyAST; `schemaVersion` is not checked.

### DIFF
--- a/packages/contract-schema/index.js
+++ b/packages/contract-schema/index.js
@@ -121,13 +121,10 @@ var properties = {
   ast: {},
   legacyAST: {
     transform: function (value, obj) {
-      var schemaVersion = obj.schemaVersion || "0.0.0";
-
-      // legacyAST introduced in v2.0.0
-      if (schemaVersion[0] < 2) {
-        return obj.ast;
-      } else {
+      if (value) {
         return value;
+      } else {
+        return obj.ast;
       }
     }
   },
--- a/packages/contract-schema/test/solc.js
+++ b/packages/contract-schema/test/solc.js
@@ -1,10 +1,10 @@
-var assert = require("assert");
-var solc = require("solc");
-var Schema = require("../");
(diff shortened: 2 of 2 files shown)


cds-amal (commented):
Nice find @haltman-at! I'm not sure, but do you think this merits a test?

haltman-at:
Oh, sure, I could add a test.  I think the easiest way would be to modify `solc.js` in the `contract-schema` tests.  But that one's a bit odd, in that it invokes `solc` directly, rather than going through `compile-solidity`, so it doesn't set up the source information like `ast` and `legacyAST`.  I guess I'll change it to use `compile-solidity`?  Kind of odd that it uses `solc` directly...

haltman-at:
Ugh.  Trying to do this ran into a circular dependency: This would mean that `contract-schema` would need to have a devdependency of `compile-common`, but `compile-common` has `contract-schema` as a dependency!  (Actually, it should be a devdependency, but I'll fix that later.)  OK, I'm going to need to find some other way to test this...

haltman-at:
OK @cds-amal, I added a test.  For the reasons stated above, I left most of that test file alone -- I didn't change to using `compile-solidity` -- but there is now a test and I did modernize the test file a tiny bit. :)

cds-amal (approved):
Nicely done @haltman-at ! LGTM

fainashalts (approved):
Awesome Harry! LGTM too.

PR #3589, opened 2020-12-08
Internal improvement: Avoid combining distinct compilations in debugger tests

haltman-at:
Distinct compilations were being combined.  I got rid of this, and also simplified the setup.  Debugger tests will now run directly off of compiler output.

### DIFF
--- a/packages/debugger/test/helpers.js
+++ b/packages/debugger/test/helpers.js
@@ -10,6 +10,7 @@ import Migrate from "@truffle/migrate";
 import Box from "@truffle/box";
 import Resolver from "@truffle/resolver";
 import * as Codec from "@truffle/codec";
+import flatten from "lodash.flatten";
 
 export async function prepareContracts(provider, sources = {}, migrations) {
   let config = await createSandbox();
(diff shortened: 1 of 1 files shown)


packages/debugger/test/helpers.js
```diff
@@ -27,14 +28,14 @@ export async function prepareContracts(provider, sources = {}, migrations) {
...
-        optimizer: {enabled: false, runs: 200},
+        optimizer: { enabled: false, runs: 200 },
         evmVersion: "constantinople"
       }
     }
   };
 
   await addContracts(config, sources);
-  let {contractNames, files} = await compile(config);
+  let { contractNames, compilations: rawCompilations } = await compile(config);
```
eggplantzzz, line 38: Oh interesting, I didn't know you could reference `compilations` this way.

packages/debugger/test/helpers.js
```diff
@@ -134,38 +128,18 @@ export async function defaultMigrations(contractNames) {
...
-  const contractNames = collectedCompilationOutput.contractNames.filter(
-    (item, index) => {
-      return collectedCompilationOutput.contractNames.indexOf(item) === index;
-    }
-  );
-  return {
-    contractNames,
-    files: sourceIndexes
-  };
+  return { compilations, contractNames };
```
eggplantzzz, line 142: Do you not need to adjust whatever code was previously calling this method? Or maybe it currently only uses `contractNames`?

eggplantzzz: Oh nevermind, I see that you already did adjust it in this file!

eggplantzzz (approved):
Seems good to me.

packages/debugger/test/helpers.js
haltman-at: See your other comment. :)

packages/debugger/test/helpers.js
haltman-at: Yup, your two comments answer one another. :)

PR #3590, opened 2020-12-09
Internal improvement: Remove unused reports.js

haltman-at:
You'd think that this is where the compilation report gets printed, right?  Nope!  That code is in `events/defaultSubscribers/compile.js`.  This is an unused near-duplicate (which is actually also slightly worse than the real one, not that it matters).


gnidan (approved):
Sweet

haltman-at:
Uh-oh, removing this causes [this test](https://github.com/trufflesuite/truffle/blob/develop/packages/truffle/test/scenarios/library/api.js) to fail.  @gnidan, does that make this a breaking change to `workflow-compile`?  I certainly hope it doesn't make it a breaking change to `truffle`...

(If somehow it *is* a breaking change to `truffle` and has to wait for v6, then in that case I'll just add a comment in the file telling people where to find the real code, so they don't get confused like I did, but obviously that's not the preferred solution!)

eggplantzzz:
I vote that this is not part of the official api for `workflow-compile` and therefore is not breaking.

eggplantzzz (approved):
I like deleting code.

Issue #3584, opened 2020-12-05
Cannot link Library and deploy just using "@truffle/contract"


elranu:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Cannot link Library just using "@truffle/contract". If I use the JS truffle test suit works.
## Steps to Reproduce

The contract is deployed correctly with the following code on the JS truffle test:
```javascript
        const lib = await ProxyLib.deployed();
        ProxyAccountGuardedV2.link(ProxyLib, lib.address);
        const proxyContract = await ProxyAccountGuardedV2.new(owner, guardian);
```

But it does not work the same contracts with the following code just using "@truffle/contract",
        
```javascript
        var contract = require("@truffle/contract");

        const proxyLib = contract(proxyLibJson);
        proxyLib.setProvider(this.web3.currentProvider);
        
        await proxyLib.detectNetwork();
        const libInstance = await proxyLib.at(this.proxyLibAddress);
        
        this.proxyContract = contract(this.jsonProxyAccount);

        this.proxyContract.setProvider(this.web3.currentProvider);
        
        await this.proxyContract.detectNetwork();
        await this.proxyContract.link(libInstance);
        
        this.instanceProxy = await this.proxyContract.new(this.addressOwner, this.addressGuardian); <--- I get the error here

        console.log('Proxy address (owner): ' + this.instanceProxy.address);
```

## Expected Behavior

Link the contract and deploy.

## Actual Results
I get the following error:

```
TypeError: library_address.replace is not a function

 at processTicksAndRejections (internal/process/task_queues.js:93:5) {hijackedStack: 'TypeError: library_address.replace is not…ns (internal/process/task_queues.js:93:5)', stack: 'TypeError: library_address.replace is not a f…ctions (internal/process/task_queues.js:93:5)', message: 'library_address.replace is not a function'}
src/services/eth-did-service.ts:72
arg0:TypeError: library_address.replace is not a function
    at ProxyAccountGuarded.<anonymous> (C:\Dev\repos\did-tools\src\tools\proxy-account.ts:62:58)
    at step (C:\Dev\repos\did-tools\src\tools\proxy-account.ts:33:23)
    at Object.next (C:\Dev\repos\did-tools\src\tools\proxy-account.ts:14:53)
    at fulfilled (C:\Dev\repos\did-tools\src\tools\proxy-account.ts:5:58)
[...]

PR #3592, opened 2020-12-10
Bug fix: Allow contract instances as an argument to the link method

eggplantzzz:
Currently on a `@truffle/contract` object (the type, i.e. `MetaCoin`) the `link` method only accepts another `@truffle/contract` object as an argument and will incorrectly link if you pass an instance. This PR adds support for passing a `@truffle/contract` instance (not the type, i.e. `await MetaCoin.deployed()`).

It also refactors the method into a big `switch` statement which more clearly conveys what is happening there.

~It is a little bit wonky to test for `contract._json.contractName` but I'm not sure if there is a better way to test that it is an instance. Ideas?~

related to https://github.com/trufflesuite/truffle/issues/3584

### DIFF
--- a/packages/contract/lib/contract/constructorMethods.js
+++ b/packages/contract/lib/contract/constructorMethods.js
@@ -6,6 +6,7 @@ const utils = require("../utils");
 const execute = require("../execute");
 const bootstrap = require("./bootstrap");
 const debug = require("debug")("contract:contract:constructorMethods");
+const OS = require("os");
 
 module.exports = Contract => ({
   configureNetwork({ networkType, provider } = {}) {
@@ -192,20 +193,30 @@ module.exports = Contract => ({
         //   - Contract.link({<libraryName>: <address>, ... })
         //   - Contract.link(<instance>)
         const obj = name;
-        if (obj.constructor && obj.constructor.contractName && obj.address) {
+        if (
+          obj.constructor &&
+          typeof obj.constructor.contractName === "string" &&
+          obj.address
+        ) {
           // obj is a Truffle contract instance
           this.link(obj.constructor.contractName, obj.address);
         } else {
           // obj is of the form { <libraryName>: <address>, ... }
-          Object.keys(obj).forEach(name => {
-            const a = obj[name];
-            this.link(name, a);
-          });
+          Object.keys(obj).forEach(name => this.link(name, obj[name]));
         }
         return;
       default:
-        const message = "Input to the link method is in the incorrect format.";
-        throw new Error(message);
+        const invalidInput = `Input to the link method is in the incorrect` +
+          ` format. Input must be one of the following:${OS.EOL}` +
+          `    - a library name and address                 > ("MyLibrary", ` +
+            `"0x123456789...")${OS.EOL}` +
+          `    - a contract type                            > ` +
+            `(MyContract)${OS.EOL}` +
+          `    - a contract instance                        > ` +
+            `(myContract)${OS.EOL}` +
+          `    - an object with library names and addresses > ({ <libName>: ` +
+            `<address>, <libName2>: <address2>, ... })${OS.EOL}`;
+        throw new Error(invalidInput);
     }
   },
 
--- a/packages/contract/lib/contract/constructorMethods.js
+++ b/packages/contract/lib/contract/constructorMethods.js
@@ -154,51 +154,60 @@ module.exports = Contract => ({
     delete this.network.address;
   },
 
+  // accepts 4 input formats
+  //  - (<name>, <address>)
+  //  - (<contractType>) - must have a deployed instance with an address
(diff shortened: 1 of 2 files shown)


haltman-at:
Repeating things from Slack, but: I like the idea of just letting anything with both a `contractName` and an `address` work.  That way, it works for instances, but also just works for any other object with those properties, which, like, why not, right? :)

haltman-at (approved):
Seems good to me!  Thanks @eggplantzzz!

packages/contract/lib/contract/constructorMethods.js
```diff
@@ -154,44 +154,59 @@ module.exports = Contract => ({
...
   link(name, address) {
-    // Case: Contract.link(instance)
-    if (typeof name === "function") {
-      const contract = name;
-
-      if (contract.isDeployed() === false) {
-        throw new Error("Cannot link contract without an address.");
-      }
+    switch (typeof name) {
+      case "string":
```
davidmurdoch, line 164: Previously the fall-through case allowed anything that _wasn't_ a `"function"` or `"object"` in this case.

Object properties are automatically cast to strings when used to get/set the property. So if someone was using a number, `"true"`, `"false"`, etc as a network name, they previously could use link(1, "address") and it would have worked (i.e., `Object.keys(this.network.links) === ["1"]`).

We probably _shouldn't_ allow this auto-casting behavior, and I actually prefer the way you are doing it in this PR. I just wanted to point out that this _could_ introduce an issue for some users; we should probably have a conversation about it juuussttt in case.

packages/contract/lib/contract/constructorMethods.js
```diff
@@ -154,44 +154,59 @@ module.exports = Contract => ({
...
+        const obj = name;
+        if (obj.constructor && obj.constructor.contractName && obj.address) {
+          // obj is a Truffle contract instance
+          this.link(obj.constructor.contractName, obj.address);
+        } else {
+          // obj is of the form { <libraryName>: <address>, ... }
+          Object.keys(obj).forEach(name => {
+            const a = obj[name];
+            this.link(name, a);
+          });
```
davidmurdoch, line 203:
```
const obj = {};
obj.constructor = {contractName: obj};
link(obj, "0xgoodbyestack");
```

:-)

If we care about users blowing the stack we should type-check these recursive calls with something like `typeof obj.constructor.contractName === "string"` (line 195) and `typeof name === "string"` (around line 202).

packages/contract/lib/contract/constructorMethods.js
```diff
@@ -154,44 +154,59 @@ module.exports = Contract => ({
...
+        }
+        return;
+      default:
+        const message = "Input to the link method is in the incorrect format.";
```
davidmurdoch, line 207: Can we specify the expected formats in the error message?

davidmurdoch (changes requested):

davidmurdoch:
Getting pedantic over the PR title here (sorry!)... are we considering this a bug fix or an enhancement/feature?

eggplantzzz:
I called it a bug fix because the docs seem to indicate that you currently can use the `link` method in the way I just enabled. See [this](https://www.trufflesuite.com/docs/truffle/reference/contract-abstractions#-code-mycontract-link-instance-code-). Although it is a bit difficult to tell because it seems that the docs use "instance" a little loosely which could refer to what I call the "type".

packages/contract/lib/contract/constructorMethods.js
eggplantzzz: I did consider this but it might be a little bit long-winded. However it is probably better to err on the side of being more helpful! I'll go add more detail.

packages/contract/lib/contract/constructorMethods.js
eggplantzzz: Yep, I think that should be considered a bug as well and I agree we shouldn't allow that input.

packages/contract/lib/contract/constructorMethods.js
eggplantzzz: Haha, that's pretty funny. I'll add in the check you suggested.

eggplantzzz:
So I updated the invalid input error to print the following:

packages/contract/lib/contract/constructorMethods.js
davidmurdoch: Ah, yup, you are right!

packages/contract/lib/contract/constructorMethods.js
davidmurdoch: @gnidan, any thoughts on this? :point_up:

gnidan: I think it's fine to get rid of the autocasting

davidmurdoch (approved):

PR #3593, opened 2020-12-10
Internal improvement: Add failure message to yarncheck job to increase clarity

haltman-at:
Two people have asked recently what the yarncheck job does, because, well, the yarncheck job doesn't say anywhere what it does!  This PR adds a message saying `"Please run yarn and commit all changes to yarn.lock"` when it fails so people will be able to see what it is they need to do. :)


davidmurdoch (approved):
:heart:

Issue #3387, opened 2020-09-23
Problem decoding events in delegatecalled, non-library contracts


eggplantzzz:
Referenced in issue #3320.

The decoder assumes only libraries get delegatecalled. If a delegatecalled, non-library contract emits events, they are not able to be decoded.

ffarhanaamin:
I am facing exactly the same issue! Any plan to give it a bit of time?

haltman-at:
OK, a fix for this is now out.

PR #3594, opened 2020-12-11
Allow decoder/codec to decode "extra" events and use this in tests (allows decoding events from non-library delegatecalls)

haltman-at:
Addresses #3387.  **Breaking change to `@truffle/codec`.**

This PR adds the `extras` option to codec's and decoder's log-decoding functionality.  In codec this is a breaking change because I replaced the `name` argument with an `options` argument containing both `name` and `extras`... and there are more options coming in the future, but not in this PR. :)  (So, let's be clear, this breaking change is one I would have had to make eventually anyway.)

In decoder this option is now available in `events()`, but also `decodeLog()` *itself* now takes options, although this is the only available option there.

Anyway, the `extras` option has three values.  The default is `"off"`, which preserves the decoder's existing behavior.  If you set it to `"on"`, then the decoder will no longer make the assumption that contracts only emit events that appear in their ABIs or a library ABIs -- it will act as if events can come from anywhere.  These extra decodings are tacked on to the end of the list.  That makes them hard to tell apart from the non-extras, so, uh, yeah, using this option is a bit icky.  Still, for those who want it, it's there.  Ideally we'd have a more structured way of dealing with this, but, uh, didn't want to make any breaking changes there.

Finally the third possibility is `"necessary"`.  In this case, extra decodings will only be returned if there were no normal decodings.

In addition to all this, this PR makes it so that Truffle Test passes the `"necessary"` option.  This is how we address #3387; if you make a delegatecall to a non-library contract, and it emits events, then Truffle Test will now be able to decode it... but it won't print such decodings unless it needs to.

(Note: @gnidan has suggested I find a way to reduce the redundant printouts of ambiguous events in Truffle Test.  I have something of an idea for this, but I'm saving that for a later PR.)

Anyway, how does this PR work?

Well, now, in `decodeEvent` in codec, we set up the extra allocations at the start, just like we set up the contract and library allocations.  We then put this on the end of the list of possible allocations... but inbetween the normal allocations and the extras, we also add `null`, which serves as a sentinel value.  Yeah, this is kind of hacky maybe, IDK, but it was the best way I could figure out.

Then, in the main loop, our first check is if our allocation is `null`.  If it is, we then either break out of the loop (if `extras === "off"`, or if `extras === "necessary"` but there's already at least one decoding), or we ignore the `null` and skip to the next iteration (if `extras === "on"`, or if `extras === "necessary"` but there are still no decodings).

And, well, that's it.  I'm not entirely happy with this, but, well, it does the job, and on the default settings the behavior is unchanged, at least.

I also added a test of this new functionality.


haltman-at:
Ugh, not clear why this is failing post-merge.  Will have to figure this out...

gnidan (approved):

PR #3598, opened 2020-12-14
Internal improvement: Reclassify some dependencies as devDependencies

haltman-at:
I noticed that `abi-utils` and `compile-common` both have `contract-schema` listed as a dependency.  However, both of these packages only use *types* from `contract-schema`, so this should be devDependency.  I've reclassified things appropriately.

Also, I updated an out-of-date comment in source-fetcher (as source-fetcher now supports Vyper sources on Etherscan, even though the CLI debugger won't actually make use of these).


eggplantzzz (approved):
Seems good

PR #3599, opened 2020-12-14
Internal improvement: Capitalize names of languages (Solidity, Yul, Vyper)

haltman-at:
Per earlier discussion, language names will now be capitalized.  Compiler names will not be -- so `"Vyper"` is the language, `"vyper"` is the compiler. :)  Anyway this PR does that.

This PR might be a breaking change to some packages?  Not sure.  I'll call it a **breaking change to debug-utils**, but we can probably get away with not calling it a breaking change to any others...

(Also as long as I was touching the syntax-highlighting code in `debug-utils` I also cleaned it up some, even though that's not the main point of this PR.)

### DIFF
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -262,7 +262,7 @@ export function getContractNode(
     if (foundNode || !source) {
       return foundNode;
     }
-    if (!source.ast || source.language !== "solidity") {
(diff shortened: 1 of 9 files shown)


haltman-at:
Ugh, I think this PR is failing because of the screwy way the decoder tests are written.  (It passes locally...)  Maybe I should finally fix that...?

gnidan (approved):
👌

PR #3601, opened 2020-12-14
Internal improvement: Rename solidity-utils package to source-map-utils

haltman-at:
**This PR deprecates the solidity-utils package. 🗒️ N.B. whoever releases this should do the deprecate business with npm**

It's no longer just for Solidity, but it is basically about source maps and things related to them, so that's its new name!  I left the version number the same?  Is that what you do when you do this?

Also, uh, wow that readme was out of date... the one thing in it wasn't even in solidity-utils anymore, and I think wasn't even in there when I started working here...


gnidan (approved):
Looks good!

PR #3602, opened 2020-12-14
Bug fix: Better handle deployed addresses library compilation

eggplantzzz:
When testing while using "pragma" as the version, Truffle would crash when attempting to create the `DeployedAddresses` library as there was no infrastructure to handle this. This PR updates the `rangeUtils` to return the default solc version in the case where "pragma" is specified.

### DIFF
--- a/packages/compile-solidity/compilerSupplier/rangeUtils.js
+++ b/packages/compile-solidity/compilerSupplier/rangeUtils.js
@@ -9,7 +9,7 @@ const RangeUtils = {
   //takes a version string which may be native or local, and resolves
   //it to one which is (presumably) either a version or a version range
   resolveToRange: function (version) {
-    if (!version) {
+    if (!version || version === "pragma") {
       return CompilerSupplier.getDefaultVersion();
     }
     //if version was native or local, must determine what version that
--- a/packages/compile-solidity/compilerSupplier/rangeUtils.js
+++ b/packages/compile-solidity/compilerSupplier/rangeUtils.js
@@ -9,6 +9,8 @@ const RangeUtils = {
   //takes a version string which may be native or local, and resolves
   //it to one which is (presumably) either a version or a version range
   resolveToRange: function (version) {
+    // if version is "pragma", we can just use the solc default since
+    // there is no explicit version specified
     if (!version || version === "pragma") {
       return CompilerSupplier.getDefaultVersion();
     }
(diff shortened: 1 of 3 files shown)


davidmurdoch:
Can we add a test for this change?

haltman-at (commented):
So, let me see if I have this straight.  This is a Truffle Test thing?  And it's OK to put whatever version we want for the contract if it's `"pragma"`, since then afterwards `"pragma"` will cause the version to be read back out of where we wrote it?  Do I have that right?

eggplantzzz:
Yes it is, so `rangeUtils` accepts a version which normally comes from the `config`. Perhaps it is better if this is done in `Deployed.ts` [here](https://github.com/trufflesuite/truffle/blob/develop/packages/resolver/lib/sources/truffle/Deployed.ts#L39). It should probably just send `undefined` or `null` so that `rangeUtils` sends back the default and doesn't need to know about this "pragma" thing. I like this better.

And yes! The Solidity source files can provide their own pragma expressions which will then be parsed and "dealt with" by `compile-solidity`.

haltman-at:
> Yes it is a test thing. So `DeployedAddresses` gets created on the fly and uses the config's version specification. It uses`rangeUtils` which accepts a version which normally comes from the `config`. Perhaps it is better if this is done in `Deployed.ts` [here](https://github.com/trufflesuite/truffle/blob/develop/packages/resolver/lib/sources/truffle/Deployed.ts#L39). It should probably just send `undefined` or `null` so that `rangeUtils` sends back the default and doesn't need to know about this "pragma" thing. Maybe this is better. Where do you think we should deal with this? In the resolver or the range utils function?

Yeah I think I agree that resolver makes more sense than where you currently have it.

> And yes! The Solidity source files can provide their own pragma expressions which will then be parsed and "dealt with" by `compile-solidity`.

Right, my point is that I'm checking that this is only being used in that context.  Which it sounds like it is!

eggplantzzz:
Although `compile-solidity` is also the only packages that knows about "pragma" so maybe it makes more sense there.

haltman-at:
> Although `compile-solidity` is also the only package that knows about "pragma" so maybe it makes more sense there.

Oh, true.  Good point.  Yeah, I agree; better to leave it where it is.  My one note might be to add a comment explaining why this is OK; like, add a comment saying, if it's pragma we can use an arbitrary version so we use the default.

haltman-at (approved):

davidmurdoch (approved):

gnidan (approved):
Neat!

PR #3569, opened 2020-11-26
Upgrade Ganache to v7

davidmurdoch:

PR #3604, opened 2020-12-14
Bug fix: Have codec rethrow unexpected errors in decoding

haltman-at:
So, I think Codec's error handling was causing some problems.  The decoding functions would just catch errors and *assume* they were expected errors we knew what to do with.  Seems they weren't always.  Now I've added a check for this, and the error will be rethrown if it's not a `DecodingError` (i.e., an expected error that we threw).

Kind of wondering if I should factor this, but not clear where the function would live?  This is probably OK for now.

(In more detail, the problem is that unexpected errors, instead of getting rethrown, were having their `error` field extracted -- which would be `undefined` since it wasn't a `DecodingError` -- and returned inside an error result.  This would then cause problems *later* when other code attempted to *use* that `undefined`, and we would have no useful stacktrace to the error's origin.  The solution, of course, is that unexpected errors should be rethrown, which means explicitly checking whether the error we got was expected or not.)

### DIFF
--- a/packages/codec/lib/errors.ts
+++ b/packages/codec/lib/errors.ts
@@ -36,3 +36,31 @@ export class StopDecodingError extends Error {
     this.allowRetry = Boolean(allowRetry);
   }
 }
+
+/**
+ * @hidden
+ */
+export function handleDecodingError(
+  dataType: Format.Types.Type,
+  error: any,
+  strict: boolean = false
+): Format.Errors.ErrorResult {
+  if (error instanceof DecodingError) {
+    //expected error
+    if (strict) {
+      //strict mode -- stop decoding on errors
+      throw new StopDecodingError(error.error);
+    } else {
+      //nonstrict mode -- return an error result
+      return <Format.Errors.ErrorResult>{
+        //I don't know why TS's inference is failing here and needs the coercion
+        type: dataType,
+        kind: "error" as const,
+        error: error.error
+      };
+    }
+  } else {
+    //if it's *not* an expected error, we better not swallow it -- rethrow!
+    throw error;
+  }
+}
--- a/packages/codec/lib/abi-data/decode/index.ts
+++ b/packages/codec/lib/abi-data/decode/index.ts
@@ -31,15 +31,19 @@ export function* decodeAbi(
     try {
       dynamic = abiSizeInfo(dataType, info.allocations.abi).dynamic;
     } catch (error) {
-      if (options.strictAbiMode) {
-        throw new StopDecodingError((<DecodingError>error).error);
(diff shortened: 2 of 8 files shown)


melyalex (approved):

haltman-at:
Good news: @davidmurdoch has used this PR to find out what was going wrong with #3569. :)

gnidan (changes requested):
(Discussed on Slack, but putting here for completeness: it seems worth factoring this now)

haltman-at:
OK, factored this.  Apologies about the generic name `handleDecodingError` -- I couldn't think of anything better, given that its job is to sometimes return and sometimes throw.

haltman-at:
Oh, one more note -- while I was at it, I deleted a bunch of dummy `_` variables in catch clauses that weren't using them.  I wasn't previously aware that you could have a catch clause without a variable, so I would do `catch (_) { ... }`.  Now that I've learned that you can just do `catch { ... }`, I've changed those catch clauses to do the latter.

packages/codec/lib/errors.ts
```diff
@@ -36,3 +36,31 @@ export class StopDecodingError extends Error {
...
 }
+
+/**
+ * @hidden
+ */
+export function handleDecodingError(
+  dataType: Format.Types.Type,
+  error: any,
+  strict: boolean = false
+): Format.Errors.ErrorResult {
```
gnidan, line 47: I don't think all of these should be positional args: `strict` definitely seems like the sort of thing that's worth making named; I'd even say that `dataType` shouldn't be positional. What do you think of:

```typescript
handleDecodingError(error, { type, strict })
```

(I wouldn't be opposed to `error` being named, also, but there's a good argument for it being positional :)

I know this PR isn't really worth debating minor semantics over, but wondering your thoughts if you considered this.

haltman-at: Hm, no, I hadn't considered it; thanks for pointing it out.  I'd rather just leave it as-is, though?  The `dataType` will always be necessary; can't have a return value without it.  I could see `strict` being named if there were also an `allowRetry` option, so that those two could exist on an equal footing, but there isn't, because errors with `allowRetry` set are always thrown manually, not via this pattern (now function).  This function is purely internal -- it's not exported from `Codec` -- so I think it's OK to leave this as-is?  If we want to make `strict`  a keyword argument it wouldn't be breaking or anything.

gnidan (approved):
:+1: since this is internal

Thanks for reminding me about `allowRetry`; I was trying to think of hypothetical options similar to `strict`. We can always revisit positional vs. keyword for if/when we add that.

PR #3609, opened 2020-12-15
When searching @truffle/db network relationships, return one CandidateSearchResult instead of list

gnidan:
- Move list to `networks`, renamed from `network`
- Return only a single `alreadyTried`, rather than a separate `alreadyTried` for each candidate

### DIFF
--- a/packages/db/src/resources/networks.ts
+++ b/packages/db/src/resources/networks.ts
@@ -1,5 +1,5 @@
 import { logger } from "@truffle/db/logger";
-const debug = logger("db:definitions:networks");
+const debug = logger("db:resources:networks");
 
 import gql from "graphql-tag";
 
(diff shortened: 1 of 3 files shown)


packages/db/src/resources/networks.ts
```diff
@@ -75,20 +75,18 @@ export const networks: Definition<"networks"> = {
...
-              network,
-              alreadyTried: alreadyTried
-            };
-          });
-
-          return untriedNetworks;
+          return {
+            networks,
+            alreadyTried: [
+              ...new Set([...alreadyTried, ...networks.map(({ id }) => id)])
```
fainashalts, line 81: Nice!

fainashalts (approved):
Awesome!

gnidan:
Thanks!

PR #3610, opened 2020-12-16
Internal improvement: Remove use of Truffle Test in decoder tests

haltman-at:
Addresses #3310.  This PR gets rid of the use of Truffle Test in the decoder tests, hopefully finally ending the headache they've posed and defusing any potential future problems they might pose with `lerna`.

To do this I reconstituted them along the lines of the debugger tests.  However I didn't want to remove the existing structure, so I kept the existing structure as two truffle projects in place and added `helpers.js`, much like exists in the debugger, to set things up.  However unlike in the debugger, we already have the structure of a Truffle project here, so we don't need to set up a sandbox, but rather just use `Config.detect` and `Environment.detect`.  We do use a temporary directory for the contracts build directory, but that's it.

The tests are a fair bit slower now due to their use of `Ganache.provider()`, but whatever.  Also I used `web3` and `abstractions` as global variables in one test, but, oh well, I think it's fine.

### DIFF
--- a/packages/debugger/test/helpers.js
+++ b/packages/debugger/test/helpers.js
@@ -15,7 +15,7 @@ import flatten from "lodash.flatten";
 export async function prepareContracts(provider, sources = {}, migrations) {
   let config = await createSandbox();
 
-  let accounts = await getAccounts(provider);
+  const accounts = await new Web3(provider).eth.getAccounts();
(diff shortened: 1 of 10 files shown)


gnidan (approved):
Sweet!

PR #3611, opened 2020-12-16
Consistently use Config.default() when compiling

haltman-at:
Sometimes when we compile, we take `Config.defaults()` and merge the given options into it.  Sometimes we don't.  I made it so we always do.  Note I didn't do it in each individual function, I just made sure each path eventually hits it.  I did this for all three compile packages.

Also, in the defaults, I added `settings` under `vyper`, and added a comment explaining where the default solc version is found (since you might expect it to be there, but it's not).  I originally also changed `vyper-json.js` in `compile-vyper` to not do `settings || {}`, since I figured that would be unnecessary now, but uh that didn't end up working, so, uh, I left it alone...?  Not sure if that means I did someting wrong elsewhere...

### DIFF
--- a/packages/compile-solidity/index.js
+++ b/packages/compile-solidity/index.js
@@ -1,11 +1,11 @@
-const debug = require("debug")("compile"); // eslint-disable-line no-unused-vars
+const debug = require("debug")("compile");
 const path = require("path");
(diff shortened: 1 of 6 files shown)


gnidan (approved):
Nice!

PR #3612, opened 2020-12-16
Internal improvement: Update compile-solidity tests for 0.8.0

haltman-at:
Updates the compile-solidity tests that always use the latest version so they'll work with 0.8.0 and the build won't break!

### DIFF
--- a/packages/compile-solidity/test/sources/v0.8.x/Version7Pragma.sol
+++ b/packages/compile-solidity/test/sources/v0.8.x/Version7Pragma.sol
@@ -1,0 +1,9 @@
+pragma solidity ^0.8.0;
+
+contract Version7Pragma {
+  uint x;
+
+  constructor() {
+    x = 5;
+  }
+}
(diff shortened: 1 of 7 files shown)


packages/compile-solidity/test/sources/v0.8.x/Version7Pragma.sol
```diff
@@ -0,0 +1,9 @@
+pragma solidity ^0.8.0;
+
+contract Version7Pragma {
```
eggplantzzz, line 3: Should this be `Version8Pragma` along with a file rename?

haltman-at: Oops! Thank you.

eggplantzzz (approved):
I like it!

haltman-at:
Thanks to @eggplantzzz for figuring out the correct solution here!  (Upgrading Ubuntu rather than doing manual package management like I was trying.)

haltman-at:
(Apologies about the messy commits here -- I'd squash it all, but I want to get this merged as soon as possible.)

PR #3614, opened 2020-12-17
Add support for settings.modelChecker

haltman-at:
Solidity 0.7.6 and up use `settings.modelChecker` rather than a separate `modelCheckerSettings`, so I've added support for that.

NB @gnidan and @fainashalts: Despite being moved under `settings` now, the model checker settings still do not go in the metadata.


gnidan (approved):
Easy!

PR #3615, opened 2020-12-17
Update Solidity sources (e.g. Truffle Init Migrations file) for 0.8.0

haltman-at:
Updates the Migrations file for 0.8.0.  Because we got rid of the explicit constructor earlier, all we should need to do is bump the pragma here. :)

Note @cds-amal I'm guessing you may want to do the same thing in your Truffle Boxes...

### DIFF
--- a/packages/core/lib/commands/create/templates/Example.sol
+++ b/packages/core/lib/commands/create/templates/Example.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity >=0.4.22 <0.8.0;
+pragma solidity >=0.4.22 <0.9.0;
 
 contract Example {
   constructor() public {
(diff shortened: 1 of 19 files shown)


cds-amal (commented):
Thanks for the heads up @haltman-at !  Do you think the [Example and Test templates](https://github.com/trufflesuite/truffle/tree/develop/packages/core/lib/commands/create/templates) for `truffle create` belongs in this PR as well?

haltman-at:
> Thanks for the heads up @haltman-at ! Do you think the [Example and Test templates](https://github.com/trufflesuite/truffle/tree/develop/packages/core/lib/commands/create/templates) for `truffle create` belongs in this PR as well?

Oh, thanks, I forgot about those!  Yes, they do, I'll go get those.  Also I'm going to put up a PR for `bare-box` because uhhh the debugger tests use those...

haltman-at:
OK @cds-amal I got those ones that you mentioned, and also a whole bunch of others I forgot about.  Pretty sure this is all of them.

haltman-at:
Also, here's my PR on bare-box: https://github.com/truffle-box/bare-box/pull/25

cds-amal (approved):
Nice @haltman-at ! LGTM.

PR #3618, opened 2020-12-17
Upgrade @truffle/db typescript to ^4.1.3

gnidan:
- Replace @zerollup/ts-transform-paths with typescript-transform-paths, since the former does not work anymore (for unknown reason)

- Fix an export that seems to go missing from compiler output, possibly due to unsupported syntax with typescript-transform-paths


haltman-at (approved):
I wasn't clear on what `afterDeclarations` meant but @gnidan explained it and so this seems good!

gnidan:
Thanks!

For reference, that usage came from the [typescript-transform-paths README](https://github.com/LeDDGroup/typescript-transform-paths)

PR #3619, opened 2020-12-17
Account for new-style invalid functions (uninitialized internal function pointers); update debugger tests to 0.8.0; fix various problems with generated sources

haltman-at:
OK, so, this PR does a few things.

The original purpose of this PR was to update how we check for designated invalid functions in `source-map-utils`.  Prior to 0.8.0 these were very simple: Just a `JUMPDEST` followed by an `INVALID`.  But as of 0.8.0, they're a bit more complicated, as they need to throw a `Panic(0x51)`.  Fortunately, it seems to always be the same, short, set series of instructions -- and I think this will hold with optimization turned on, too, because it's a pretty hard sequence to optimize!  So, now we check for either the old sequence (`JUMPDEST INVALID`), or the new sequence (which is ten instructions long and I'm not going to write it all out here, particularly as it includes a `PUSH32`).

Also, in Codec, I changed it so that such pointers now display as `<uninitialized>` rather than `<assert(false)>`, since an uninitialized internal function pointer no longer does the same thing as `assert(false)` (the former throws a `Panic(0x51)`, the latter a `Panic(0x01)`).

However, testing this necessitated updating the debugger tests to 0.8.0.  And doing this revealed another problem that needed fixing.

See, 0.8.0 makes much more use of generated sources than previous versions, and this was screwing some things up.  See the debugger often makes use of the idea that a given AST node is ready for processing when we're just about to leave it.  But if we're just about to leave it, from a user source into a generated source... yeah.  Sometimes we're jumping away in what used to be the middle of it, you know?  So now, we no longer count a source range as final if we're currently in a user source and about to jump into a generated source (unless we're at a context change).  I don't *think* this should screw anything up -- all tests are currently passing, after all?  But we'll see.  Definitely *not* doing this screwed things up for sure...

In addition to the above change to the `solidity` submodule, I also had to make a similar change to the `stacktrace` submodule.  The stacktrace submodule has to try to tell where a revert originated, see, which means that when an inner transaction reverts, it has to somehow identify that the revert from the outer transaction was caused by the inner one.  It does this by checking whether the current position changes inbetween receving the inner revert and throwing the outer revert.  If it doesn't, it infers that the inner one caused the outer one.  Except, once again, generated sources screw this up.  So we have to say that generated sources don't count as moving, and also not to process jumps until motion has occurred (otherwise we'll add generated sources that shouldn't appear -- there's nothing wrong with generated sources appearing in stacktraces, but these are the *wrong* generated sources!).

Funnily enough I kind of accounted for this when writing `stacktrace` -- I said that unmapped code doesn't count as having moved.  But when generated sources were added, I never updated this to also say that generated sources don't count as having moved.  Well, now I have.  Hooray.

### DIFF
--- a/packages/codec/lib/format/utils/inspect.ts
+++ b/packages/codec/lib/format/utils/inspect.ts
@@ -266,7 +266,7 @@ export class ResultInspector {
                   case "exception":
                     return coercedResult.value.deployedProgramCounter === 0
                       ? options.stylize(`[Function: <zero>]`, "special")
-                      : options.stylize(`[Function: assert(false)]`, "special");
+                      : options.stylize(`[Function: <uninitialized>]`, "special");
                   case "unknown":
                     let firstLine = `[Function: decoding not supported (raw info:`;
                     let secondLine = `deployed PC=${coercedResult.value.deployedProgramCounter}, constructor PC=${coercedResult.value.constructorProgramCounter})]`;
@@ -618,7 +618,7 @@ function nativizeWithTable(
             case "exception":
               return coercedResult.value.deployedProgramCounter === 0
                 ? `<zero>`
-                : `assert(false)`;
+                : `<uninitialized>`;
             case "unknown":
               return `<decoding not supported>`;
           }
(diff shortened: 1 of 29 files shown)


gnidan (approved):
Not bad!

PR #3620, opened 2020-12-17
Bug fix: Ensure range is a single version before using it in semver.gte

eggplantzzz:
This causes a crash when a range is used in this method as `gte` takes a single version and not a range.


eggplantzzz:
Ah looks like this was already fixed :)

PR #3621, opened 2020-12-17
Bug fix: Fix pragma version replacement for DeployedAddresses

eggplantzzz:
Replace pragma versions differently depending on whether we have a single version or a range


haltman-at (approved):
Makes sense to me!  Sorry for causing all these problems!

fainashalts:
Good catch! 🎉

PR #3623, opened 2020-12-18
Enhancement: Update globally-available variables for 0.8.0; also set _chainId in exposed Ganache instances

haltman-at:
The main purpose of this PR is to update the debugger's globally-available support for Solidity 0.8.0.  This means two things:

1. In Solidity 0.8.0, `msg.sender` and `tx.origin` are now of type `address`, rather than `address payable`.  That doesn't really matter that much, but we keep track of it, so I updated it.  This purely required changes in `codec`.

2. In Solidity 0.8.0, the `block` special variable has a new property `block.chainid`, which contains the chain ID as a `uint`.  I added support for this.  Note that there is a version check; it won't appear when debugging an earlier version.

Doing this second part required both changes in `codec/lib/special/decode`, but also in the debugger itself, to actually get the chain ID and pass it to `codec`.  Fortunately very little needed to be altered here, as it can really piggyback off of the existing globally-available variables code.  The only real work that needed to be done here was actually getting the chain ID in `web3/sagas` and putting it in the block info; the rest is basically taken care of automatically.

However, there was a bit of a snag here; while testing, I ran into Ganache's chain ID bug, where the `eth_chainId` method reports the ID as 1337, but the `CHAINID` opcode reports it as 1.  Since the debugger relies on `eth_chainId` to report the value of `block.chainid`, that means it would misreport it.  I wanted to minimize the cases where that would happen, so I found (what I think is) everywhere in Truffle that we spin up a user-exposed Ganache instance (not ones for internal testing) and added `_chainId: 1337` to the Ganache options, to make it consistent.  I don't think this should be a breaking change.  Again, I left our own test instances of Ganache alone, but I did add it to one such test instance, namely, the one in the debugger's globally-available variable tests, because I added tests of `block.chainid` to that, so that one would need to have it consistent.

Also, while it isn't the main point of the PR, I realized that `decoder` was using a different `ganache-core` version from all the other packages, so I made it consistent.  Possibly I should have made it consistent by updating the others, but, eh, we can do that later.  With any luck we'll be on v3 soon enough anyway. :)

### DIFF
--- a/packages/decoder/package.json
+++ b/packages/decoder/package.json
@@ -41,7 +41,7 @@
     "@types/debug": "^0.0.31",
     "@types/web3": "1.0.19",
     "chai": "^4.2.0",
-    "ganache-core": "^2.13.1",
+    "ganache-core": "2.13.0",
     "lodash.clonedeep": "^4.5.0",
     "lodash.flatten": "^4.4.0",
     "tmp": "^0.2.1",
--- a/packages/codec/lib/ast/import/index.ts
+++ b/packages/codec/lib/ast/import/index.ts
@@ -34,6 +34,7 @@ export function definitionToType(
       };
     case "address": {
       switch (Compiler.Utils.solidityFamily(compiler)) {
+        case "unknown": //I guess?
         case "pre-0.5.0":
           return {
             typeClass,
@@ -41,6 +42,7 @@ export function definitionToType(
             typeHint
           };
         case "0.5.x":
+        case "0.8.x":
           return {
             typeClass,
             kind: "specific",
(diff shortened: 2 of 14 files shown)


packages/decoder/package.json
```diff
@@ -41,7 +41,7 @@
     "@types/debug": "^0.0.31",
     "@types/web3": "1.0.19",
     "chai": "^4.2.0",
-    "ganache-core": "^2.13.1",
+    "ganache-core": "2.13.0",
```
gnidan, line 44: Why'd you pin this?

haltman-at: I mentioned it at the end -- just consistency, everything else was pinned to 2.13.0.  But instead I could go and pin everything to 2.13.1 so they're all upgraded. :)

gnidan: Ah sorry, missed that. Probably upgrade and unpin, but whatever, this makes sense for this PR.

gnidan (approved):

PR #3625, opened 2020-12-18
Db/language on processed source

fainashalts:
This PR does two things: 

1) Consumes the new `sources` array returned in the `workflowCompile` result, and uses that to load Sources rather than the contract sources array. This helps ensure that we don't miss sources that are not contracts.

2) Adds a `language` property to compilations' `processedSources` array.


fainashalts:
Closing this PR in favor of https://github.com/trufflesuite/truffle/pull/3652 because rebase was causing some strange issues with old commits

PR #3626, opened 2020-12-19
Merge resourceInput with existing resource

fainashalts:
This PR add a merge of `resourceInput` with an existing `resource` in `@truffle/db`. This is needed to handle the case where the underlying resource's contents change but the id does not. This is not likely to happen regularly but this should handle the edge case.

I did not write tests for this because I didn't see a straightforward way to do it, but one can test this by loading the db in a project (by running `truffle compile`), then adding a new value to any resource. Recompile the project, and check that the resource now contains the new value.

### DIFF
--- a/packages/db/src/pouch/databases.ts
+++ b/packages/db/src/pouch/databases.ts
@@ -5,8 +5,6 @@ import PouchDB from "pouchdb";
 import PouchDBDebug from "pouchdb-debug";
 import PouchDBFind from "pouchdb-find";
 
-import * as lodash from "lodash";
-
 import {
   CollectionName,
   Collections,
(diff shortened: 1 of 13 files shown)


packages/db/src/pouch/databases.ts
```diff
@@ -152,6 +153,24 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
     }
   }
 
+  public async merge<N extends CollectionName<C>>(
```
gnidan, line 156: This can be `private` I think - we don't need to call it elsewhere, right?

packages/db/src/pouch/databases.ts
```diff
@@ -176,7 +195,7 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
         // check for existing
         const resource = await this.get(collectionName, id);
         if (resource) {
-          return resource;
+          return this.merge(collectionName, resource, resourceInput);
```
gnidan, line 198: Just realized that you're not saving this - you're just returning it. Probably want to save the merged version!

gnidan (changes requested):
Looks good apart from a couple things!

Also seems like you need to commit `yarn.lock` :)

packages/db/src/pouch/databases.ts
fainashalts: Hm, I thought about that but I can't add the `private` modifier on the interface in `pouch/types`. Any thoughts?

fainashalts:
Closing this as not necessary right now. Something like this may be needed in the future but some additional thinking needs to go into proper implementation in the context of trying to keep resources immutable to the greatest extent possible.

PR #3627, opened 2020-12-19
Enhancement: Add support for decoding of panics and fallback return values (and also some other stacktracing improvements)

haltman-at:
OK, so, this PR does a few things.  I'll split it into three main groups: Panics; fallback return values; and other stacktrace-related improvements.

Part one: Panics.  This PR adds support for decoding of panics.  This affects Codec, the debugger, `debug-utils`, the debugger CLI, and `abi-utils`.

You may be wondering -- abi-utils?  It affects abi-utils because revert decodings now contain the corresponding ABI.  So, I've added error ABI entries to the types there.  Of course, these don't really exist yet, but acting as if they do seemed the best way to do this decoding in a future-proof manner.  In addition to simply adding the type, I also added support for the type to various ABI-related functions, both here and in Codec.  I did not add arbitraries for this type.

The changes to Codec here are quite small.  I just added spoofed ABIs to revert decoding output, and added an allocation for `Panic` alongside the allocation for `Error`.  Codec just gets the numeric panic code; it doesn't attempt to determine what this code means.  That's a job for `debug-utils`!  The new function `panicString()` takes a panic code (given as a number or a BN) and returns a message describing what it means.  There's also a second optional argument, `verbose`, for when a longer version is appropriate. :)

As for the debugger, it had to have changes made to the `stacktrace` submodule.  Unfortunately (unlike with `txlog` which received no changes), I wrote that submodule assuming revert messages would always be strings.  To keep the change from being breaking, the `message` field still holds a revert string when one is present, but the `panic` field will hold a panic code when one is present.

Speaking of stacktraces, that's another change to `debug-utils`!  The `formatStacktrace` function, used for formatting stacktraces both in the debugger CLI and in Truffle Test, accounts for these changes to `stacktrace`, and uses the aforementioned `panicString` to provide an explanation of the panic.

As for the debugger CLI, it needs to know how to print out these panics.  It'll print both a verbose version on transaction end, and a shorter version on pressing `v` after the transaction is done.  There's one thing here I'm not entirely happy with -- that last one prints it out in decimal, because that's what was convenient, even though everywhere else I'm printing panic codes, I'm doing it in hex (because that's how Solidity documentation describes them).  I think this is probably OK?  They are at least accompanied by a short explanation regardless. :)

So that's panics!  You'll note one thing I *didn't* do: I didn't add support for panic decoding in `contract/lib/reason.js`.  That thing is just such a hacked-up mess I didn't want to touch it, and I don't really a lot of familiarity with how it's used.  So I decided to just leave it alone.  And, of course, I obviously didn't touch revert string decoding in `ganache-core`, as that's a totally separate repo. :P

Part two: Fallback return values.  This was a little bit trickier.  Whereas panics are basically just like revert strings but with an integer instead of a string, these are new.  Now, the decoding here is pretty trivial; it's just raw binary data, there's not really anything to decode.  But it had to be fit into the existing structure.

So, first off, that means a new type of returndata decoding!  I gave it the tag `"returnessage"`, and it has `data` as just a hex string.  It also means a corresponding new type of allocation!  Not that this new type really contains any information beyond "hey, this is the type of allocation that means just treat the data as raw binary".  I created one instance of this allocation as an exported constant in `codec/abi-data/allocate`, for when it's necessary.  And, of course, Codec's `decodeReturndata` had to be updated to account for this new type of allocation and new type of return value.  Note this is not a breaking change to `codec`, as it's only possible to get this new type of output if you pass the new allocation in.

There were no changes to `decoder`; its return value decoding functionality already only works for functions, not constructors or anything else, so I saw no need to add any new support there.

Where there was a bit of work was `debugger` and the debugger CLI.  See, we don't want to "decode" and print fallback output unless the fallback is supposed to *have* output.  (Also, if there is output, we'll want to know its name.)  And the `fallback` ABI entry type hasn't changed, even if declared fallback functions can now include input and output.  So we can't rely on that.

So, instead, the debugger checks the AST for this information (via some new selectors).  And then the CLI can print it appropriately.  There is one little awkwardness here -- since I had codec return the data as just a hex string, we can't use `ResultInspector`, so, uh, instead we just fake things a little with some hackery.  It's OK I think?

Part three: Other stacktracing improvements I made while I was at it.

There's two of these.  First, I made it so that stacktraces only include the addresses for external calls, not internal calls.  That required going into `stacktrace` and adding the external/internal distinction to the stacktrace report object, which wasn't there previously, although since we had that information anyway that wasn't a big alteration.

The second isn't actually a change to stacktraces but rather to the debugger CLI's error location printout.  (Although, it required a change to stacktraces.)  See, when the debugger CLI prints out a stacktrace, it also prints out a selection of the location where the error happened.  Problem: Since 0.8.0 expands the use of generated sources, oftentimes the error-location is inside a generated sources.  This is not that helpful.  So, I added information to the stacktrace report object to distinguish generated sources from user ones.  This information isn't used by `formatStacktrace` -- actualy stacktrace printouts will still include generated sources -- but it is used by the error location printout, so that if the top of the stacktrace is a generated source, it'll look below that until it finds a user source, and it'll print out *that* location so you see something helpful.  (I think the mismatch is OK?)

Also, I factored this functionality into its own function in the debug printer, whereas before it was done twice manually in `interpreter.js`.

(Also, I added some tests.)

And that should be the end of the 0.7.6 / 0.8.0 compatibility work! :D

### DIFF
--- a/packages/abi-utils/lib/normalize.ts
+++ b/packages/abi-utils/lib/normalize.ts
@@ -7,8 +7,8 @@ export const normalize = (looseAbi: SchemaAbi | Abi): Abi =>
 type Item<A> = A extends (infer I)[] ? I : never;
 
 export const normalizeEntry = (looseEntry: Item<SchemaAbi> | Entry): Entry => {
-  if (looseEntry.type === "event") {
-    // nothing gets normalized for events right now
+  if (looseEntry.type === "event" || looseEntry.type === "error") {
+    // nothing gets normalized for events or errorsright now
     return looseEntry as Entry;
   }
 
--- a/packages/abi-utils/lib/types.ts
+++ b/packages/abi-utils/lib/types.ts
@@ -5,7 +5,8 @@ export type Entry =
   | ConstructorEntry
   | FallbackEntry
   | ReceiveEntry
-  | EventEntry;
+  | EventEntry
+  | ErrorEntry;
 
 export type StateMutability = "pure" | "view" | "nonpayable" | "payable";
 
@@ -38,6 +39,12 @@ export interface EventEntry {
   name: string;
   inputs: EventParameter[];
   anonymous: boolean;
+}
+
+export interface ErrorEntry {
+  type: "error";
+  name: string;
+  inputs: Parameter[];
 }
 
 export interface Parameter {
--- a/packages/codec/lib/abi-data/allocate/index.ts
+++ b/packages/codec/lib/abi-data/allocate/index.ts
@@ -21,6 +21,7 @@ import {
   ReturndataAllocation,
   FunctionReturndataAllocation,
   ConstructorReturndataAllocation,
+  AdditionalReturndataAllocation,
   ReturnImmutableAllocation,
   CalldataAllocations,
   CalldataAllocationTemporary,
@@ -58,6 +59,12 @@ interface EventParameterInfo {
   name: string;
   indexed: boolean;
 }
+
+export const FallbackOutputAllocation: AdditionalReturndataAllocation = {
+  kind: "returnmessage",
+  selector: new Uint8Array(), //empty
+  allocationMode: "full"
+};
 
 export function getAbiAllocations(
   userDefinedTypes: Format.Types.TypesById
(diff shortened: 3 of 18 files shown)


haltman-at:
Note: This last commit moves the priority of `"returnmessage"` above that of `"selfdestruct"`, because if we know the fallback might return data, and we get empty data back (and the transaction was marked successful), the idea that it just returned empty data is a better assumption than the idea that it self-destructed.

gnidan (approved):
Looks good, seems to make sense!

PR #3628, opened 2020-12-19
Further work on @truffle/plugins

rkalis:
This builds on the initial work done by @gnidan to pull out the plugin functionality into a separate package.
Contains these changes from his original code:

- Further split up `plugin.ts` into `Plugin.ts`, `Plugins.ts`, `utils.ts` and `interfaces.ts`
- Add `findPluginsForCommand()` function
- Fix `pluginPath` being used instead of `modulePath` in `Plugin:loadModule()`
- Add tests to `@truffle/plugins` package
- Add `@truffle/plugins` as dependency to `@truffle/core`
- Update `run.js` to use the new interface
- Update `run.js` tests to fit the new interface
- Update tests to check for specific error strings
- Update tests to assert that throwing tests actually throw
- Remove `plugin.js` tests since this functionality moved to `@truffle/plugins`
- Add `checkPluginConfig.js` tests (this is a subset of the old `plugin.js` tests)
- Do not catch loadModule() errors so misconfigurations get caught, rather than ignored


rkalis:
Also rebased this branch as well as the target branch on top of the current origin/develop.

haltman-at:
@rkalis, the yarncheck job is failing on this PR.  This means that you need to run `yarn` and commit all resulting changes to `yarn.lock`.  Thank you!

rkalis:
Thanks @haltman-at. Looks like the CI is now failing because it runs into a segmentation fault in @truffle/artifactor 😅

haltman-at:
Ah, yeah, those happen at the moment and we don't know why. :-/  I've restarted it, hopefully that doesn't happen this time.  Or hit any timeouts (there's one I upped recently because it was happening so often, if you want to pull that in...).  Sorry, our CI is kind of a mess at the moment...

rkalis:
Nice, now we have an actual test failure :D 

Looks like I forgot to update the relevant tests in the main `truffle` package.

rkalis:
Looks like the tests are fixed. 

One of the changes in this commit was to no longer catch and ignore certain types of misconfigurations. E.g. if two plugins are defined `plugins: ['plugin1', 'plugin2']`, where they both define the same run command, but plugin1 is misconfigured, it used to ignore that issue and use plugin2 instead. After this change, it tells you hey man plugin1 isn't working. 

It makes sense to me to make it explicit if something is misconfigured even if there's another plugin to use, but since this is a slight breaking change, I can also see why you wouldn't want to change that. So can also change that back.

packages/plugins/lib/interfaces.ts
```diff
@@ -0,0 +1,31 @@
...
+export type RawPluginConfig = PluginConfig | string;
+
+// internal representation of a particular plugin configuration
+export interface PluginConfig {
+  module: string;
+}
+
+// Partial representation of the truffle-config object
+// TODO: Properly define the full interface for this object in the appropriate module and import it here
+export interface TruffleConfig {
```
gnidan, line 12: So @truffle/config does export types - does it make sense to use that type directly?

gnidan (approved):
This looks good @rkalis!

packages/plugins/lib/interfaces.ts
rkalis: It looks like the `TruffleConfig` exported in `@truffle/config` doesn't have very strong typings:

```ts
declare class TruffleConfig {
    [key: string]: any;
    private _deepCopy;
    private _values;
    constructor(truffleDirectory?: string, workingDirectory?: string, network?: any);
    eventManagerOptions(config: TruffleConfig): any;
    addProp(propertyName: string, descriptor: any): void;
    normalize(obj: any): any;
    with(obj: any): TruffleConfig;
    merge(obj: any): TruffleConfig;
    static default(): TruffleConfig;
    static search(options?: any, filename?: string): string | null;
    static detect(options?: any, filename?: string): TruffleConfig;
    static load(file: string, options?: any): TruffleConfig;
    static getUserConfig(): Configstore;
    static getTruffleDataDirectory(): string;
}
```

So it doesn't tell you that e.g. `working_directory` is an existing key with a `string` value type. It might just be fine to go with this less strict type, but it might also be good to add stricter typings to the `@truffle/config` package at some point.

gnidan: Sounds good - let's keep it lenient for now!

rkalis:
I just rebased this on top of the latest `develop` since there were some changes made to the command interface (callbacks were replaced with promises). I made some slight changes to accommodate this.

PR #3633, opened 2020-12-21
Internal improvement: Reorganize code-utils to be more TypeScripty

haltman-at:
Code-utils, while it was converted to TypeScript, was done so in a not-very-typescripty manner, and it was getting annoying, so I decided to reorganize it.  Specifically:

* The source is now in an `src` directory rather than the project root
* Types are listed in `types.ts` rather than `typings/index.d.ts`
* The `types` field in the package.json now points to `dist/index.d.ts`
* The types `opcodeObject` and `opcodes` have been renamed `Instruction` and `OpcodeTable`, respectively, although the old names are also still exported for compatibility, in case anyone was using those
* The `parseCode` function now has an explicit return type
* `export = { ... }` has just been replaced with `export ...` -- except, OK, I *also* did `export default { parseCode }` for compatibility, so it should work no matter what import style you're using :P
* The `Instruction` type has been cleaned up -- `pc` and `name` are now mandatory, while `pushData` is now always a string
* Tests now import from the new locations

Also, computing pushData now uses just `Uint8Array`s, rather than `number[]`s; actual arrays only get involved in the rare case where pushData needs to be padded.


haltman-at:
Hm, OK, evidently I didn't do this quite right as `@truffle/db` is running into problems importing the types.  Will have to see about that...

gnidan (approved):

PR #3637, opened 2020-12-23
Enhancement: Allow debugger to debug Solidity test contracts (also, some other new options)

haltman-at:
As requested by @cds-amal.

This PR adds three new options to the debugger CLI: `--compile-tests`, `--force-recompile`, and `--force-no-recompile`.

The interesting new one is `--compile-tests`.  This causes the debugger CLI to compile your Solidity test contracts so you can debug them.  Hooray!  It also forces a recompilation more generally.

There's also `--force-recompile`.  It's not called `--compile-all` because it doesn't save the results.  Anyway, this forces the debugger to do a recompile on startup, even if its heuristics would say that it doesn't need to.  You know, in case there's a problem they miss.

Finally there's `--force-no-recompile`.  This forces the debugger *not* to recompile, in case the debugger erroneously thinks there's a problem, or you're pretty sure any problems are irrelevant, or you just think the check takes too long and you're sure there's no problems.  This is marked dangerous, for obvious reasons.  Combining it will either of the previous two options will error out.

Basically the `compile` method on `DebugCompiler` now takes a `withTests` option; if passed, it'll get the Solidity tests (in basically the same way that `truffle test` does) and set up a test resolver (for importing Truffle sources -- yup, you can debug these! :) )  Then it does the compilation!  And, well, the rest of this PR is pretty straightforward, I think.

Note that I changed the `compile` method on `DebugCompiler` to take named arguments instead of positional, and updated `external.js` to pass `sources` appropriately.  Since I don't think this can be called exposed in any reasonable sense, I wouldn't call this any sort of breaking change.

Of course the debugger help text has been updated appropriately; I'll put up a docs PR once this is merged.

No tests added; I just tested it manually.

### DIFF
--- a/packages/core/lib/commands/debug.js
+++ b/packages/core/lib/commands/debug.js
@@ -10,11 +10,26 @@ const command = {
       alias: "x",
       type: "boolean",
       default: false
+    },
+    "compile-tests": {
+      describe: "Allow debugging of Solidity test contracts",
+      type: "boolean",
+      default: false
+    },
+    "force-recompile": {
+      describe: "Force debugger to compile all contracts for extra safety",
+      type: "boolean",
+      default: false
+    },
+    "force-no-recompile": {
+      describe: "Force debugger to skip compilation (dangerous!)",
+      type: "boolean",
+      default: false
     }
   },
   help: {
     usage:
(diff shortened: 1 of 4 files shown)


gnidan (approved):
Looks good!

Issue #3647, opened 2020-12-28
Vyper compiler needed if json file(s) are present.


aarlt:
## Issue

Truffle tries to execute the vyper compiler, even if there are no vyper sources present. If truffle finds any `.json` file, it assumes that the project contains vyper code and tries to run the vyper compiler. 

@cameel did an analyses here: https://github.com/ethereum/solidity/pull/10672#issuecomment-749699412.

eggplantzzz:
Hey @aarlt thanks for reporting this! Looks like `compile-vyper` shouldn't care about JSON files. I'll try and put in a fix for this this week!

eggplantzzz:
Hmm actually it might be a bit more nuanced than that because Truffle currently has support for `vyper-json` which allows you to have input in JSON format. We'll have to figure out a different solution than just removing the detection of the "json" file extension.

eggplantzzz:
A fix was created in https://github.com/trufflesuite/truffle/pull/3650

cameel:
Thanks! [Seems to work fine with OpenZeppelin in our external tests](https://github.com/ethereum/solidity/pull/10672#issuecomment-757777256).

PR #3650, opened 2021-01-04
Bug fix: Prevent Vyper compilations when JSON but no Vyper

haltman-at:
Addresses #3647.

This PR makes it so that `compile-vyper` won't attempt to do anything if there are JSON files present in the contracts directory but no actual Vyper files.  I had JSON count as Vyper to support Vyper's native ability to import JSON, but we don't always want that, as there might be JSON present for other reasons and no actual Vyper.  So, I changed the "are there any Vyper files present?" check to restrict to actual Vyper files, not including JSON.


eggplantzzz (approved):
Thanks @haltman-at! I like it!

PR #3652, opened 2021-01-05
Db/processed sources language

fainashalts:
This PR does two things:

Consumes the new sources array returned in the workflowCompile result, and uses that to load Sources rather than the contract sources array. This helps ensure that we don't miss sources that are not contracts.

Adds a language property to compilations' processedSources array.


gnidan (approved):
Looks good!

Issue #3651, opened 2021-01-05
truffle debug: Compile your contracts...Error: ENOENT


cds-amal:
Reproduction steps
  1. clone this [project](https://github.com/cds-amal/psychic-octo-parakeet)
  2. start `ganache-cli` in a separate terminal
  3. run the tests: `truffle test`
  4. grab a transaction from ganache's output
  5. run the debugger with the compile-tests flag: `truffle debug --compile-tests <txn>`

I was able to reproduce this on both WSL and Arch. See arch session below

```

$ use-truffle-core
$ lsb_release -a
LSB Version:    1.4
Distributor ID: Arch
Description:    Arch Linux
Release:        rolling
Codename:       n/a

$ truffle version
Truffle (unbundled) (core: 5.1.60)
Solidity v0.5.16 (solc-js)
Node v10.22.1
Web3.js v1.2.9


$ truffle test

Compiling your contracts...
===========================
> Compiling ./contracts/Dummy.sol
> Compiling ./contracts/Migrations.sol
> Compiling ./test/TestDummy.sol
> Artifacts written to /tmp/test--7948-6Pksjlvts5pY
> Compiled successfully using:
   - solc: 0.5.16+commit.9c3226ce.Emscripten.clang



  TestDummy
    ✓ testTargetIsInitialized (54ms)

  Contract: Dummy
    ✓ should assert true


  2 passing (4s)

$ truffle debug --compile-tests 0x38e676885fb0e18f2191d4bb3ae256ecd77d6cf827d7617369d94d59db03384f
Starting Truffle Debugger...
⠹ Compiling your contracts...Error: ENOENT: no such file or directory, scandir '/home/amal/dev/txlog/truffle-test-template/build/contracts'
    at Object.readdirSync (fs.js:790:3)
    at Truffle.<anonymous> (/home/amal/work/truffle/packages/resolver/lib/sources/truffle/index.ts:20:53)
    at Generator.next (<anonymous>)
    at fulfilled (/home/amal/work/truffle/packages/resolver/dist/lib/sources/truffle/index.js:5:58)
Truffle (unbundled) (core: 5.1.60)
Node v10.22.1

```

haltman-at:
Thanks, I can now reproduce the problem!

PR #3654, opened 2021-01-05
Bug fix: Prevent crash from missing build directory in Truffle resolver source

haltman-at:
Addresses #3651.

This PR fixes a problem that if you tried to use a `truffle` resolver source, but the contracts build directory was missing, the `resolve` function would crash.  It would call `fs.readdirSync` on the build directory, but if the build directory was absent, this would crash.  So, I added a check to see first if it exists, and if not, it just sets the files found there to `[]`.  This seems to solve the problem.

### DIFF
--- a/packages/resolver/lib/sources/truffle/index.ts
+++ b/packages/resolver/lib/sources/truffle/index.ts
@@ -16,12 +16,11 @@ export class Truffle implements ResolverSource {
     if (importPath === "truffle/DeployedAddresses.sol") {
       const sourceFiles = await findContracts(this.options.contracts_directory);
 
-      let abstractionFiles: string[] =
+      const buildDirFiles: string[] =
         fse.existsSync(this.options.contracts_build_directory)
           ? fse.readdirSync(this.options.contracts_build_directory)
           : [];
-      const buildDirFiles = abstractionFiles;
-      abstractionFiles = buildDirFiles.filter(file => file.match(/^.*.json$/));
+      const abstractionFiles = buildDirFiles.filter(file => file.match(/^.*.json$/));
 
       const mapping: { [key: string]: string | false } = {};
 
--- a/packages/resolver/lib/sources/truffle/index.ts
+++ b/packages/resolver/lib/sources/truffle/index.ts
@@ -18,7 +18,7 @@ export class Truffle implements ResolverSource {
 
       let abstractionFiles: string[] =
         fse.existsSync(this.options.contracts_build_directory)
-          ? fse.readdirSync(this.options.contract_build_directory)
+          ? fse.readdirSync(this.options.contracts_build_directory)
           : [];
       const buildDirFiles = abstractionFiles;
       abstractionFiles = buildDirFiles.filter(file => file.match(/^.*.json$/));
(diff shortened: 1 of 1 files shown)


cds-amal:
I got an error while testing this branch against the same truffle project on WSL. 

Truffle branch: compile-tests-fix

```
$ truffle test

Compiling your contracts...
===========================
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be one of type string, Buffer, or URL. Received type undefined
    at Object.readdirSync (fs.js:785:3)
    at Truffle.<anonymous> (/home/amal/cds-dev/truffle/packages/resolver/lib/sources/truffle/index.ts:21:17)
    at fulfilled (/home/amal/cds-dev/truffle/packages/resolver/dist/lib/sources/truffle/index.js:5:58)
Truffle (unbundled) (core: 5.1.60)
Node v10.23.0

$ truffle version                                                                                                                                        1 ↵
Truffle (unbundled) (core: 5.1.60)
Solidity v0.5.16 (solc-js)
Node v10.23.0
Web3.js v1.2.9

```

cds-amal (changes requested):
Looks good @haltman-at. just one typo.

packages/resolver/lib/sources/truffle/index.ts
```diff
@@ -16,10 +16,11 @@ export class Truffle implements ResolverSource {
     if (importPath === "truffle/DeployedAddresses.sol") {
       const sourceFiles = await findContracts(this.options.contracts_directory);
 
-      let abstractionFiles: string[];
-      const buildDirFiles = (abstractionFiles = fse.readdirSync(
-        this.options.contracts_build_directory
-      ));
+      let abstractionFiles: string[] =
+        fse.existsSync(this.options.contracts_build_directory)
+          ? fse.readdirSync(this.options.contract_build_directory)
```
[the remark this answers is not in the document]
haltman-at: Oops!  No wonder this isn't passing CI.  Thanks!

packages/resolver/lib/sources/truffle/index.ts
```diff
@@ -16,10 +16,11 @@ export class Truffle implements ResolverSource {
...
 
-      let abstractionFiles: string[];
-      const buildDirFiles = (abstractionFiles = fse.readdirSync(
-        this.options.contracts_build_directory
-      ));
+      let abstractionFiles: string[] =
+        fse.existsSync(this.options.contracts_build_directory)
+          ? fse.readdirSync(this.options.contracts_build_directory)
+          : [];
+      const buildDirFiles = abstractionFiles;
```
cds-amal, line 23: The handoff between `abstractionFiles` and `buildDirFiles` is seems odd. Would you consider this suggestion?

```
      const buildDirFiles: string[] =
        fse.existsSync(this.options.contracts_build_directory)
          ? fse.readdirSync(this.options.contracts_build_directory)
          : [];
      const abstractionFiles = buildDirFiles.filter(file => file.match(/^.*.json$/));
```

haltman-at: Sure, that makes sense.  I should note that that's preexisting; I left the existing structure in place there.  But I agree that would make sense.

(Also, is this still changes requested, or should that be gone now?)

haltman-at: In any case I went ahead and did this.

cds-amal (approved):
LGTM! Thanks for resolving this so quickly @haltman-at !

PR #3658, opened 2021-01-05
Prevent config.merge() from modifying merge(); have config.merge() merge more fields; add config.clone() method; make sure cloned configs have getters & setters

haltman-at:
This PR does three things:

1. First, it adds a `clone` method to config, since that didn't exist.  Naturally, it just calls `this.with({})`. :)

2. Secondly, it has `config.merge()` call `this.clone()` at the start so that we don't modify the original.  Note, I didn't change any of the code relating to events because uh I don't really understand that stuff but it seems that `with` uses `this` even though it returns a copy so uh hoping this is alright...?

3. Finally, it has `config.merge()` actually merge a few more fields; previously it only merged `compilers`, now it also merges `solc` (for compatibility), `db`, `ens`, and `ethpm`.  Merging the `networks` field was discussed but not done as doing so may be unsafe if there are provider objects in there.  @gnidan had an idea for how to make that more safe, but that won't fit in this quick PR. :P

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -1,6 +1,5 @@
 import "source-map-support/register";
 import path from "path";
-import assignIn from "lodash.assignin";
 import merge from "lodash.merge";
 import Module from "module";
 import findUp from "find-up";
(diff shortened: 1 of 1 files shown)


haltman-at:
OK, fourth change: I changed `.with` to use `TruffleConfig.default()` as the base, rather than `Object.create(TruffleConfig.prototype)`, as the latter was causing problems due to not having its getters and setters set up.  That, uh, seems like a bug... have cloned configs not had getters and setters all this time??

haltman-at:
Alright, another change to this one.  This one changes `with` to use a manual loop to do the copying rather than using `lodash.assignIn`.

The reason for this change is so that we can put a try/catch on the individual properties, and not copy ones that can't be set directly.  Because now that I've made it so that cloned configs do in fact have their setters, well, we better not set the ones that can't be.

Now you may notice there is one concerning thing here -- isn't this more like an `assign` than an `assignIn`?  Well, yes.  But, since `TruffleConfig` doesn't inherit from anything, there shouldn't be any difference.  So, I'm hoping this works.

haltman-at:
Closing this because I couldn't get this to work and I think it's just too much of a mess.  We can revisit this at a later time; for now, I suppose we can just use `config.with({}).merge(...)` if we want to make sure we don't modify anything?

PR #3659, opened 2021-01-05
Allow `truffle debug` without a mnemonic

gnidan:
For people using Infura or another node that doesn't have unlocked accounts, the `eth_accounts` RPC method could result in an exception. Since one of the first things that Truffle does is get a list of unlocked accounts, this means that you have to set up HDWalletProvider or such just to do things like debug a transaction.

This PR catches these exceptions to delay the point of failure to situations that actually require being able to sign messages.

Note: this PR doesn't actually address those use cases at all, figuring that the existing opaque error in this situation is just as bad as any future opaque error that happens as a result. Assuming this is a low-impact downside, the **upside of this** is that you can debug mainnet transactions with an archive node without needing to specify a mnemonic!

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -58,24 +58,34 @@ class Console extends EventEmitter {
     });
   }
 
-  start() {
+  async start() {
     try {
       this.repl = repl.start({
         prompt: "truffle(" + this.options.network + ")> ",
         eval: this.interpret.bind(this)
       });
 
-      this.interfaceAdapter.getAccounts().then(fetchedAccounts => {
-        this.repl.context.web3 = this.web3;
-        this.repl.context.interfaceAdapter = this.interfaceAdapter;
-        this.repl.context.accounts = fetchedAccounts;
-      });
+      let accounts;
+      try {
+        accounts = await this.interfaceAdapter.getAccounts();
+      } catch {
+        // don't prevent Truffle from working if user doesn't provide some way
+        // to sign transactions (e.g. no reason to disallow debugging)
+        accounts = [];
+      }
+
+      this.repl.context.web3 = this.web3;
+      this.repl.context.interfaceAdapter = this.interfaceAdapter;
+      this.repl.context.accounts = accounts;
       this.provision();
 
       //want repl to exit when it receives an exit command
       this.repl.on("exit", () => {
         process.exit();
       });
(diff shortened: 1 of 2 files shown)


haltman-at:
That's weird... where on earth does the debugger get accounts?  I can't find anywhere where it does it explicitly -- because, I mean, duh, why would it.  So I guess it's doing it as part of some other operation that implicitly includes this.  I wonder what that might be.  Does `Evironment.detect` do this, or something?

gnidan:
Yep @haltman-at - rather than try to alter the behavior of `Environment.detect()` in a potentially risky fashion, I figured this would be an acceptable hack. Your thoughts?

haltman-at:
Wouldn't altering `Environment.detect` in this way be *less* risky?  That way you'd still get an exception when used outside of `Environment.detect`, whereas here, the exception is suppressed in all cases.  (I'm assuming you'd make the same modification, just there instead of here; I didn't have anything more elaborate in mind.)  I don't know, do you think there should be more to it than that?

gnidan:
Modified the behavior as per discussion above - @haltman-at I did this in environment and also console, figuring that also makes sense to allow.

packages/core/lib/console.js
```diff
@@ -58,18 +58,25 @@ class Console extends EventEmitter {
     });
   }
 
-  start() {
+  async start() {
```
haltman-at, line 61: Yay, getting rid of callbacks :)

Looks like changing this to `async` won't mess anything up, as it's already `await`ed where used, so that's nice!  However I'd suggest changing `return new Promise(() => {});` to just `return;` now that it's `async`.

haltman-at (commented):
Seems good to me, although I see one thing might be good to change.

packages/core/lib/console.js
gnidan: Ah, but that line is supposed to prevent the `await`-ability! See the hidden first `accept` argument there, which is never called ;)

haltman-at: Oh, so the idea is that awaiting it never returns until it gets the exit signal?  Interesting.  Might be worth a comment, I think, because that's pretty non-obvious.  Anyway I'll go hit the approve button, that's up to you. :)

haltman-at (approved):

PR #3661, opened 2021-01-06
Allow querying collections of project names

gnidan:
- Add `contracts` field to Project, to query for all current, name-assigned contracts
- Add `networks` field to Project, to query for all current, name-assigned networks
- Factor out a `resolve` function to resolve a given name and/or type for a project; update the existing `network` and `contract` resolvers to use this new function.

### DIFF
--- a/packages/db/src/resources/projects.ts
+++ b/packages/db/src/resources/projects.ts
@@ -3,7 +3,7 @@ const debug = logger("db:resources:projects");
 
 import gql from "graphql-tag";
 
-import { Definition } from "./types";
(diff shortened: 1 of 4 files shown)


fainashalts (approved):
LGTM but should we add some lines to our tests to ensure these queries are working as expected?

gnidan:
@fainashalts let me know if you think this test is sufficient! thanks!

fainashalts (approved):
LGTM! Thanks for adding a test. 🎉

PR #3664, opened 2021-01-06
Bug fix: Fix RangeUtils method

eggplantzzz:
There was an error in the `rangeContainsAtLeast` method and here it is fixed. Some tests have also been added.

### DIFF
--- a/packages/compile-solidity/compilerSupplier/rangeUtils.js
+++ b/packages/compile-solidity/compilerSupplier/rangeUtils.js
@@ -34,10 +34,10 @@ const RangeUtils = {
     //the following line doesn't, despite the flag, but does work with version ranges
     const rangeAtLeast =
       semver.validRange(range, { loose: true }) &&
-      !semver.ltr(comparisonVersion, range, {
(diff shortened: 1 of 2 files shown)


haltman-at (commented):
I think the code is correct as-is?  I'm confused as to what the point of this PR is.  Could you explain in more detail?

haltman-at:
Oh oops there is a problem here.  I think the right fix is to use `!gtr` instead of `!ltr`. >_>  Sorry about the mess...

haltman-at (changes requested):
OK, I think the correct fix is just replacing the use of `ltr` with the use of `gtr`.  Ugh, sorry I got mixed up here, the documentation is written in a confusing way.  Would you mind just making that change instead?  At least, assuming I'm right and that doing so doesn't cause any problems. >_>  Thanks!

haltman-at:
(Also thanks for adding tests, those should do a bit to make sure we've finally got this right! >_> )

haltman-at (approved):
Thanks Tyler!

PR #3665, opened 2021-01-07
Bug fix: Have compile-vyper's necessary() call sourcesWithDependencies(), remove compilationTargets logic

haltman-at:
Right now I occasionally get errors when compiling Vyper without `--all` because `compile-vyper`'s `necessary()` method calls `sources()` instead of `sourcesWithDependencies()`, so sometimes the compiler it can't locate imported sources.  So, this PR changes that.  I also removed the `compilationTargets` logic because -- now that I understand a bit better how that's supposed to work -- I've realized that's not the appropriate place to handle that.

Note that right now the `sourcesWithDependencies()` method in `compile-vyper` just calls `all()`, because we don't currently have an import parser for Vyper like we do with Solidity.  Obviously that's not great for performance, but it's better than failing to compile at all!

Anyway, now Vyper should always compile at least.  We might want to see about writing an import parser for Vyper like we have for Solidity?  I'm having trouble locating where the code for that even is...

### DIFF
--- a/packages/compile-vyper/index.js
+++ b/packages/compile-vyper/index.js
@@ -251,7 +251,7 @@ const Compile = {
     const updatedVyperPaths = updated.filter(path => {
       return path.match(/\.vy$|\.v.py$|\.vyper.py$/);
     });
-    return await Compile.sources({
+    return await Compile.sourcesWithDependencies({
       sources: files,
       options: options.with({
         compilationTargets: updatedVyperPaths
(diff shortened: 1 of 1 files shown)


gnidan (approved):
Makes sense to me.

Here's the Solidity imports parser: https://github.com/trufflesuite/truffle/blob/develop/packages/compile-solidity/parser.js

haltman-at:
Ah, thanks!  The `compile-solidity` logic is so tangled I really got lost looking for that...

PR #3667, opened 2021-01-07
Bug fix: Update profiler method to allow json files for compilation

eggplantzzz:
JSON files were not being included in compilation because the `shouldIncludePath` in the profiler was not updated to allow this. With the additions of support for abi JSON files, we now have to allow these sources to be whitelisted here so they can be passed to the Solidity compiler (after resolution and the generation of the relevant Solidity of course).

**Update:** There are a couple of additional changes added to this PR:
 - when the abi resolver has a problem resolving a JSON file (likely because it is Vyper-related JSON) it will return a blank Solidity file (with just a license line and pragma expression)
 - after `requiredSources` we exit compilation when we only find JSON files to be compiled (in this case no Solidity must import the JSON and thus must not be appropriate for `compile-solidity`)

### DIFF
--- a/packages/compile-solidity/index.js
+++ b/packages/compile-solidity/index.js
@@ -86,7 +86,7 @@ const Compile = {
 
     // we need to remove all JSON files so Vyper-related stuff doesn't get
     // sent to the compiler - the Profiler will find JSON imports
-    const filteredPaths = paths.filter(path => path.extname !== ".json");
+    const filteredPaths = paths.filter(file => path.extname(file) !== ".json");
 
     options = Config.default().merge(options);
     const { allSources, compilationTargets } = await Profiler.requiredSources(
--- a/packages/compile-solidity/index.js
+++ b/packages/compile-solidity/index.js
@@ -84,18 +84,15 @@ const Compile = {
       "resolver"
     ]);
 
-    // we need to remove all JSON files so Vyper-related stuff doesn't get
-    // sent to the compiler - the Profiler will find JSON imports
-    const filteredPaths = paths.filter(file => path.extname(file) !== ".json");
-
     options = Config.default().merge(options);
     const { allSources, compilationTargets } = await Profiler.requiredSources(
       options.with({
-        paths: filteredPaths,
+        paths,
         base_path: options.contracts_directory,
         resolver: options.resolver
       })
     );
+
 
     const hasTargets = compilationTargets.length;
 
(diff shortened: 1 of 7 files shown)


gnidan (approved):

haltman-at (commented):
Hm, so, obviously this needs to be fixed, but there's something I'm worried about here.   (This is kind of related to some of the stuff I was trying to do with Vyper, maybe we should discuss this.)

So, we need this so that `.json` files can work in imports.  But, if I'm not mistaken, I think this will *also* cause them to get compiled by themselves, right?  I think that would be a bad thing, mostly due to how it could interact with Vyper.  So I think you should also go into `compile-solidity` and add a filter there so that it initially only retrieves `.sol` files.  And then afterward, when computing dependencies, the profiler can find both `.sol` and `.json` sources and add those to the list.

Do I have right how that would work?  This does all get a bit confusing...

packages/compile-solidity/index.js
```diff
@@ -84,10 +84,14 @@ const Compile = {
       "resolver"
     ]);
 
+    // we need to remove all JSON files so Vyper-related stuff doesn't get
+    // sent to the compiler - the Profiler will find JSON imports
+    const filteredPaths = paths.filter(path => path.extname !== ".json");
```
haltman-at, line 89: This line doesn't look right.  Did you mean to do `path.extname(path)`?  Although, um, I guess that wouldn't work due to one name meaning two things.

Alternatively, you could just check that `path.endsWith(".sol")`.

haltman-at (commented):
Huh, CI is failing?  I'm hoping that's spurious, I'm gonna go restart it if you don't mind...

packages/compile-solidity/index.js
eggplantzzz: Doh, that is a mistake

haltman-at (approved):
Seems good to me, as long as it passes CI!

packages/compile-solidity/index.js
```diff
@@ -142,7 +152,9 @@ const Compile = {
               path.sep +
               path.relative(options.working_directory, contract);
           }
-          if (contract.match(blacklistRegex)) return;
+          if (contract.match(blacklistRegex) || contract.endsWith(".json")) {
```
haltman-at, line 155: Ooh, nice catch

haltman-at (approved):
Thanks Tyler!

PR #3691, opened 2021-01-10
Add missing id to @truffle/db jsonrpc requests

gnidan:
Since this field is required according to spec (even though Ganache doesn't require it)


fainashalts (approved):
LGTM!

gnidan:
Thanks!

PR #3693, opened 2021-01-11
Add missing indexes to @truffle/db resources

gnidan:
This PR adds indexes to NetworkGenealogy for `ancestor` and `descendant` (since that's how we look those up)


fainashalts (approved):
Nice!

gnidan:
Thanks!

PR #3695, opened 2021-01-11
Hook up @truffle/db name assignments in compile and migrate

gnidan:
This PR updates the logic for `truffle compile` and `truffle migrate` to assign contract names. This works by exposing a new `Compile.assignNames` method in workflow-compile and calling that from the places where it's relevant.

Note that doing it this way means that we get name assignments for those commands, but not for `truffle test`.

### DIFF
--- a/packages/core/lib/commands/compile.js
+++ b/packages/core/lib/commands/compile.js
@@ -93,7 +93,9 @@ const command = {
         );
       }
 
-      return await WorkflowCompile.save(config, compilationOutput);
+      const result = await WorkflowCompile.save(config, compilationOutput);
+      await WorkflowCompile.assignNames(config, result);
+      return result;
     }
   },
 
--- a/packages/core/lib/commands/migrate.js
+++ b/packages/core/lib/commands/migrate.js
@@ -201,7 +201,8 @@ const command = {
       conf.compiler = "none";
     }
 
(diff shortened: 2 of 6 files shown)


packages/core/lib/commands/compile.js
```diff
@@ -93,7 +93,9 @@ const command = {
         );
       }
 
-      return await WorkflowCompile.save(config, compilationOutput);
+      const result = await WorkflowCompile.save(config, compilationOutput);
+      await WorkflowCompile.assignNames(config, result);
```
fainashalts, line 97: This probably unlikely to be an issue, but should we do something to catch errors here? If not, is there a way to add a test?

fainashalts (changes requested):
Looks great except for the one comment

packages/core/lib/commands/compile.js
gnidan: Hm, not sure how to test it... Maybe expand the scenario test? Is that merged?

I think we want errors because the user will have explicitly enabled @truffle/db - if something fails, it seems to make sense to bubble it up to the user.

gnidan: OK updated this to have a real error message and corresponding test

fainashalts (approved):
LGTM!

gnidan:
Thanks a bunch!

PR #3698, opened 2021-01-11
Internal improvement: Increase some timeouts and fix some tests that were arrow functions

haltman-at:
So, there are these three tests that keep timing out.  I know that @eggplantzzz is working on getting us better CI, but in the meantime, I'm upping the timeouts on two of them.  The third one actually already had a longer timeout set, but, the call to set it did nothing because the test was defined as an arrow function!  For those who don't know, you don't want to write your tests to be arrow functions because then you can't to do things like `this.timeout(10000)` or `this.skip()`.  Anyway so in addition to upping those two timeouts I also made a bunch of tests no longer arrow functions, including the one that was causing the problem.

### DIFF
--- a/packages/core/test/config.js
+++ b/packages/core/test/config.js
@@ -16,7 +16,8 @@ describe("config", function () {
     from: "0x1234567890123456789012345678901234567890"
   };
 
-  before("Create a sandbox with extra config values", async () => {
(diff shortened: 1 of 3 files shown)


gnidan (approved):
Thanks @haltman-at!

PR #3700, opened 2021-01-12
Make DeployedAddresses compatible with .abi.json files

eggplantzzz:
Strip off .json and .abi.json from source file extensions when creating `DeployedAddresses`.

`DeployedAddresses` creates invalid Solidity when you have .json and .abi.json files as you get something like the following created in that contract:
```
function IConversion.abi.json() ...
```
This messes up compilation. In order to account for this we strip off the file extensions like it does for Solidity files.


haltman-at (approved):
Looks good to me!

PR #3701, opened 2021-01-12
Add license to Console.sol

eggplantzzz:
`Console.sol` currently has no license identifier and causes the compiler to print warnings which is ugly. Get rid of them!


haltman-at (approved):
Thanks Tyler! :)  Might also want to update that pragma while you're at it...

gnidan (approved):

PR #3704, opened 2021-01-13
Experimental feature: Compilation with pragma analysis

eggplantzzz:
This is an experimental feature where Truffle analyzes pragma expressions in a user's Solidity files, then organizes compilation such that each source is compiled with the version of the Solidity compiler that most suits it.


packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
...
+      const message = `Invalid semver expression (${expression}) found in` +
+        `one of your contract's imports.`;
+      throw new Error(message);
+    }
+  }
+};
+
+// takes an array of versions and an array of semver expressions
+// returns a version of the compiler or undefined if none can be found
+const findNewestSatisfyingVersion = ({ solcReleases, semverExpressions }) => {
```
haltman-at, line 31: Haha, you didn't end up using my complicated intersection function, I see XD

Also: If you're using an arrow function, but the function body is just a single return statement, no need to bother with braces and `return`, right? :)

Also: I think this function can be simplified a little by use of the semver function `maxSatisfying`:

```
const findNewestSatisfyingVersion = ({ solcReleases, semverExpressions }) =>
  maxSatisfying(
    semverExpressions.map(range => maxSatisfying(solcReleases, range)),
    ""
  )
```

The inner `maxSatisfying` does what it says it does -- finds the maximum solc version among the given releases that satisfies the given range.  The outer one is maybe a bit of a hack... I'm just using it to take the maximum of the results, passing `""` as the range to ensure they all satisfy it. :)

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
+const CompilerSupplier = require("./compilerSupplier");
+const Config = require("@truffle/config");
+const semver = require("semver");
+const Profiler = require("./profiler");
+const fse = require("fs-extra");
+const { run } = require("./run");
+const OS = require("os");
+const cloneDeep = require("lodash.clonedeep");
+
+const getSemverExpression = source => {
```
haltman-at, line 10: Do we want to throw an explicit error message here if one of the files has no version pragma?  I feel like that's worth an explicit error message, saying like, you have to put pragmas in every file if you want to use this feature.  Or is that not a restriction, and files with no pragmas just don't restrict the version used?  (If nothing has any pragmas, do we use latest Solidity?)

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
...
+  };
+  const compilerSupplier = new CompilerSupplier(supplierOptions);
+  const { releases } = await compilerSupplier.getReleases();
+
+  // collect sources by the version of the Solidity compiler that they require
+  const versionsAndSources = {};
+  for (const path of paths) {
+    const source = (await options.resolver.resolve(path)).body;
+
+    const parserVersion = findNewestSatisfyingVersion({
```
haltman-at, line 62: Hm, this means that we won't find a suitable compiler for parsing unless there's a single version they all have in common, right?  So that won't allow mixing of e.g. `pragma solidity ^0.5.0` and `pragma solidity ^0.6.0`? :-/  That's troublesome...

I guess in order for it to be otherwise... I guess the parser itself would have to inspect the pragma and load an appropriate Solidity compiler according to it.  Yeah I suppose that is a bit troublesome...

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
...
+
+const getSemverExpressions = sources => {
+  return sources.map(source => getSemverExpression(source));
+};
+
+const validateSemverExpressions = semverExpressions => {
+  const { validRange } = semver;
+  for (const expression of semverExpressions) {
+    if (semver.validRange(expression) === null) {
+      const message = `Invalid semver expression (${expression}) found in` +
```
haltman-at, line 22: Again, do we want to have an error here for no pragma present?  Or is that not an error?  Etc.

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
...
+      semverExpressions
+    });
+    if (!newestSatisfyingVersion) {
+      throwCompilerVersionNotFound({
+        path,
+        semverExpressions
+      });
+    }
+
+    if (versionsAndSources[newestSatisfyingVersion]) {
```
haltman-at, line 107: You actually don't need a conditional here!  Inside an object literal (note: *not* in any other context!), it's actually legal to do `...undefined`!  It acts the same as `...{}`.

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
...
+        ...allSources
+      };
+    } else {
+      versionsAndSources[newestSatisfyingVersion] = allSources;
+    }
+  }
+
+  const compilations = [];
+  for (const compilerVersion of Object.keys(versionsAndSources)) {
+    const compilationOptions = Object.assign(
```
haltman-at, line 119: This `Object.assign` doesn't appear to be doing anything?

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
...
+        ...versionsAndSources[newestSatisfyingVersion],
+        ...allSources
+      };
+    } else {
+      versionsAndSources[newestSatisfyingVersion] = allSources;
+    }
+  }
+
+  const compilations = [];
+  for (const compilerVersion of Object.keys(versionsAndSources)) {
```
haltman-at, line 118: Maybe just a `for ... in` here rather than `Object.keys()`?  IDK, I guess they're equivalent.

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
...
+const throwCompilerVersionNotFound = ({ path, semverExpressions }) => {
+  const message =
+    `Could not find a single version of the Solidity compiler that ` +
+    `satisfies the following semver expressions obtained from your source ` +
+    `files' pragma statements: ${semverExpressions.join(" - ")}. ` +
+    `${OS.EOL}Please check the pragma statements for ${path} and its imports.`;
+  throw new Error(message);
+};
+
+const compileWithPragmaAnalysis = async ({ paths, options }) => {
```
haltman-at, line 49: I just wanted to say, this ended up being simpler than I expected!  I was worried this was going to be terribly nasty, but fortunately looks like it isn't! :)

packages/compile-solidity/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,139 @@
...
+  const supplierOptions = {
+    events: options.events,
+    solcConfig: options.compilers.solc
+  };
+  const compilerSupplier = new CompilerSupplier(supplierOptions);
+  const { releases } = await compilerSupplier.getReleases();
+
+  // collect sources by the version of the Solidity compiler that they require
+  const versionsAndSources = {};
+  for (const path of paths) {
```
haltman-at, line 59: Sorry, one more comment here I just realized: We should filter `paths` by `shouldIncludePath` or something similar, right?  Otherwise we risk attempting to run this pragma analyis on Vyper files.

packages/resolver/lib/sources/truffle/Deployed.ts
```diff
@@ -46,14 +46,6 @@ export class Deployed {
       //remove "payable"s in conversions if we're before 0.6.0
       source = source.replace(/payable\((.*)\)/g, "$1");
     }
-    //regardless of version, replace all pragmas with the new version
```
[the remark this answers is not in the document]
haltman-at: Oh! Wait! Crap! I just realized, if we're going to get rid of this, we should also change the pragma above so it supports Solidity 0.4.x.  So, um, please make that change.  We can just make it `pragma solidity >= 0.4.15 < 0.9.0`, I figure.  (It won't work before 0.4.15 because of the `pure` keyword.)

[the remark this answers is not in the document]
eggplantzzz: Cool, I updated it!

packages/compile-solidity/normalizeOptions.js
```diff
@@ -0,0 +1,35 @@
...
+    options.compilers.solc.optimizer ||
+    {};
+
+  // Grandfather in old solc config
+  if (options.solc) {
+    options.compilers.solc.settings.evmVersion = options.solc.evmVersion;
+    options.compilers.solc.settings.optimizer = options.solc.optimizer;
+  }
+
+  // Certain situations result in `{}` as a value for compilationTargets
```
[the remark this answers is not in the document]
eggplantzzz: Thanks for the heads up!

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: Oh I guess that is a good point. I missed that, good catch!

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: Are pragma expressions optional in Solidity files? I had no idea you could omit it...

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: I don't think this is what we want though. We need to find the maximum version that satisfies every range. If you take the max for every range and then pick the highest then it is likely that you'll get a range that only satisfies a subset of the ranges input and therefore won't be compatible with some range.

For example, say you have `["0.6.0", ">0.5.0 <=0.7.2"]`. Won't this method return "0.7.2" which won't satisfy `"0.6.0"`?

packages/compile-solidity/test/compileWithPragmaAnalysis.js
```diff
@@ -0,0 +1,186 @@
...
+    solc: {
+      settings: {},
+      version: "analyzePragmas"
+    }
+  }
+});
+
+const releases = {
+  prereleases: [],
+  releases: [
```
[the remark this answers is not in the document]
eggplantzzz: It can't hurt.

packages/compile-common/src/profiler/profiler.ts
```diff
@@ -69,4 +69,24 @@ export class Profiler {
       allPaths
     });
   }
+
+  async requiredSourcesForSingleFile(options: TruffleConfig) {
```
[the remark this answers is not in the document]
eggplantzzz: I would say that sounds like a good idea. It is kind of obnoxious to pin on random things.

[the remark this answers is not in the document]
haltman-at: Hm, yeah, I guess it uses `paths` and `basePath` which is more of an established thing.  So the only really new thing here is `path` (as opposed to `paths`).  Maybe this is fine.

packages/compile-solidity/compileWithPragmaAnalysis.js
haltman-at: Yes, they're optional.  It's a compiler warning, that's all.

packages/compile-solidity/compileWithPragmaAnalysis.js
haltman-at: Oh, sily me, you're absolutely right.  Blargh.  Don't know why I keep getting this mixed up.

Yeah, nevermind, you have it right.  We could also use my ridiculous intersection function, but, uh, this way is probably better.

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: Oh cool

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: `for ... in` seems a little nicer.

packages/compile-common/src/profiler/profiler.ts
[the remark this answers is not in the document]
eggplantzzz: So I left this the way it is for now. In the future maybe we want to consider breaking these apart to keep the config object cleaner.

packages/compile-solidity/compileWithPragmaAnalysis.js
eggplantzzz: Note that this happens for each "path". This means that every source file will be analyzed this way to find an appropriate version of the parser.

haltman-at: Ah, I see, thanks!

But, hm, there's still a problem here, isn't there?  Because it's trying to determine what parser to use for a whole tree based on a single file -- the newest one that works with that file had better also work with all its dependencies.

OTOH, maybe that isn't a problem -- if the newest one that works with that file doesn't work with its dependencies, it should have specified its pragma better!  Also, we're just talking about getting a version for *parsing* here, and that will likely be fairly robust; parser changes happen but, like, relevant ones shouldn't happen that often, right?

So I guess this is probably fine, at least for an initial version.  Whew, I was worried there!

eggplantzzz: Yeah so this is also "experimental" which makes me feel better about the cracks people might find in this functionality haha.

eggplantzzz:
So at this point, it looks like the requirement is that all of a user's "parent" contracts must contain pragma expressions. We allow for the imports to not contain them and they will not be used to determine which compiler version to use. Only the pragma expressions found will be used.

haltman-at (approved):
Cool, I think this is finally good to go! :D

PR #3705, opened 2021-01-13
Force-exit @truffle/db's jest when it finishes

gnidan:
This hanging process was bothering me


gnidan:
It seems like jest doesn't want to quit because of open handles. Doing some investigation, my hunch is that `Environment.detect()` is opening some socket and somehow that's not getting cleaned up. Not sure what that's about, but this should get around the problem for now.

fainashalts (approved):
This makes me so happy! 🎉

PR #3707, opened 2021-01-14
Add immutable references to @truffle/db

fainashalts:
This PR adds `immutableReferences` to the data in `compilations`. Immutable references contain an `astNode`, as well as information about the reference -- where it starts (the offset) and what it's length is. The immutableReferences data is displayed as an array from graphql when queried. 

For additional information on what immutable references are and how Solidity treats them, see the docs here: https://docs.soliditylang.org/en/v0.6.5/contracts.html#immutable

Immutable references have been added to the relevant tests, but it is also possible to test via graphql playground. Simply compile a contract that contains an immutable reference, and from the `truffle/db` repo run `yarn start <path-to-your-project`. Then you can see the playground at localhost://4444 and run a query like this: 

```
query getAllCompilations {
  compilations {
    id
    immutableReferences {
      astNode
      length
      offsets
    }
  }
}
```


packages/db/src/resources/compilations.ts
```diff
@@ -23,6 +23,21 @@ export const compilations: Definition<"compilations"> = {
       processedSources: [ProcessedSource]!
       sourceMaps: [SourceMap]
       contracts: [Contract]!
+      immutableReferences: [ImmutableReference]!
+    }
+
+    type ImmutableReference {
+      astNode: String!
+      bytecode: Bytecode!
```
haltman-at, line 31: Just to be clear, that's a reference to a bytecode, right?  Like an ID or something?  I mean I have to assume so from context, but I don't know this code....

packages/db/src/resources/compilations.ts
```diff
@@ -54,6 +69,7 @@ export const compilations: Definition<"compilations"> = {
       processedSources: [ProcessedSourceInput]
       sources: [ResourceReferenceInput]!
       sourceMaps: [SourceMapInput]
+      immutableReferences: [ImmutableReferenceInput]
```
haltman-at, line 72: OK, I'm pretty confused what's going on here.  Why is immutable references here, when e.g. I don't see link references here?  Like, I don't know exactly how you have it organized, but in my mind, immutable references and link references are part of bytecode, which is part of a contract.  But I don't see contract, bytecode, or link references here.  So while I'll admit I don't really understand what this is, this certainly doesn't *look* like the right place, at least?  I would expect to see it alongside link references underneath bytecode, or something at least resembling that...

packages/db/src/resources/compilations.ts
fainashalts: Yes, correct!

packages/db/src/resources/compilations.ts
fainashalts: @gnidan and I discussed this and ultimately decided to put it in the compilation instead. The bytecode id is used to link it to the contract/bytecode at issue. Happy to discuss further, as I wasn't sure where it should go either initially...

haltman-at: Huh, interesting.  Is it because of the use of an AST ID?  That would make sense.

OK, so, the ownership structure's a bit weird.  But given a bytecode you can still get all its link references and immutable references, I'm assuming?  (Is there some sort of backlink there or is that just automatic?)

haltman-at: Actually, wait, now I'm wondering -- how *do* link references work, then?  Since those have to refer to the library they're linking... is the way those work that they're owned by the bytecode and they contain links to the contract object for the library?  Or what?

fainashalts: Hi @haltman-at ok sorry for the delay in responding on this. We don't have bytecode lookup right now afaik, but Bytecode is a resource so if you have its ID you can get the bytecode's info. That currently includes bytes, linkReferences, ad instructions but the resource definition does not include immutableReferences. I do think the reason us because we have the `astNode` there.

Link references are treated as part of the bytecode though. Does that make sense?

gnidan (approved):

packages/db/src/resources/compilations.ts
haltman-at: I guess that makes sense.  Based on what you said in the meeting earlier though it sounds like currently based on the bytecode ID you can't immediately get all link references and immutable references -- you can get the former but not the latter, you'd also need to know the compilation for the latter.  Do I have that right?  If so that'll need to be fixed I think (though probably that can be a separate PR from this one that just adds immutable references in the first place), you should be able to get immutable references for bytecode without having to know any other information.

So tangential but out of curiosity, do link references linking to a given library link to that library's contract object in the DB, or what?

gnidan: > So tangential but out of curiosity, do link references linking to a given library link to that library's contract object in the DB, or what?

So there's LinkReference and LinkValue - Bytecode defines LinkReference and ContractInstance defines LinkedBytecode which includes LinkValue. Those don't directly link to the corresponding ContractInstance (LinkValues just have the raw bytes), so you'd have to do another lookup, assuming those value bytes are an address, for the ContractInstance with that address

Note we could always make all these lookups more friendly once you get a chance to play with it and see what queries you want.

haltman-at: Oh, sorry, that's not what I'm asking.  I'm not asking about the link "value" (the address), but rather the link "key" (the library class for which an address is being linked).

gnidan: Ah that's not part of @truffle/db's data model right now. Link references define `name`, but that's a meaningless string.

If we want to have semantic relationships there, we'll have to move LinkReference to Compilation.

haltman-at: Oh, hm.  So maybe that should be done and then immutable references and link references would work the same way...?  I might bug you about this more on Slack.

gnidan: Yeah so my conclusion here really is that we need a ProcessedBytecode like we have ProcessedSource

PR #3709, opened 2021-01-15
Refactor @truffle/preserves

rkalis:
I went over all the code in @truffle/preserves, discussed several things with @gnidan and made several changes. Most importantly I refactored most of the "control/processes" module to reduce the amount of generics. I also changes some terms, most notable "Unknowns" are now "Value resolutions", "labels" are "results" or "resolutions" depending on context, and "thunk" is now "stringify" (this latter is still a bit of a misnomer, but I think it comes across better than "thunk").

- Complete refactor the control/processes/controllers module
  - Remove some layers of generics
  - Move all event types to a single file
  - Define interfaces for all different Controllers
  - Rename the "Unkowns" controller to ValueResolutionController
  - Remove succeed/fail from the initial "Controls" object
  - Change `public state` to `protected _state` + a `get()`
- Move loaders out of the targets folder into its own module
- Rename label(s) to result(s) if they refer to the result of a plugin
- Rename label(s) to resolution(s) if they refer to resolution of a value using the ValueResolutionController
- Move around the `import * as X; export { X }` statements so they are always inside the index.ts files
- Rename thunk to stringify
- Move test folder out of lib
- Some other smaller changes


rkalis:
Checks are still failing because the other packages still depend on the previous iteration.

rkalis:
Converted to draft since this depends on refactors of the other related preserve packages before it can be merged.

Issue #3702, opened 2021-01-13
Profiler.updated() shouldn't rely on artifacts


haltman-at:
## Issue

`Profiler.updated()` currently works by checking the timestamp on artifacts and comparing them to the timestamp on the source.  The problem with this approach is that artifacts correspond to contracts, not sources.  So this would create a problem for e.g. any source with no contracts, with possible further problems depending on the exact implementation (I haven't looked into it).  `Profiler.updated()` should use something that's source-based, not contract-based.

There are various ways we could do this, but the right way, I guess, is hooking it up to Truffle DB and having it use that.  This could either use compilation timestamps, or simple comparison of contents.  Either way, that's what should be used.  (There are other things that could be done that wouldn't be reliant on Truffle DB, but they'd basically be duplicative of our efforts there, so I don't think it makes sense to do that...)

## Environment

* Truffle version (`truffle version`): 5.1.61

Issue #3668, opened 2021-01-07
Imprecise error message "Error: Could not find\n import from any sources;" 


ivan-homoliak-sutd:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

When compiling a file containing syntactic error - a missing semicolon in other than the last import, one gets an unclear error message. 

## Steps to Reproduce

Compile the following contract with truffle using solc 0.7.x.

$> npx truffle compile

// SPDX-License-Identifier: MIT
pragma solidity ^0.7.0;

import "@openzeppelin/contracts/utils/EnumerableSet.sol"
import "@openzeppelin/contracts/math/SafeMath.sol";

contract MyContract {

}

## Expected Behavior

The error message should inform about the line number with the error and a more precise type of error.

## Actual Results
The following message is displayed:

Error: Error: Could not find
import from any sources; imported from /home/user/...

## Environment

* Operating System:  
Ubuntu 18.04 
* Truffle version (`truffle version`):
Truffle v5.1.60 (core: 5.1.60)
Solidity v0.5.16 (solc-js)                                                                                                                                                                                                              
Node v15.5.1                                                                                                                                                                                                                            
Web3.js v1.2.9  
* npm version (`npm --version`): 
7.3.0

fainashalts:
Hi @ivan-homoliak-sutd thanks for bringing this to our attention! We'll see what we can do to make this error more clear.

gnidan:
Fix for this is released in [v5.1.64](https://github.com/trufflesuite/truffle/releases/tag/v5.1.64)!

Issue #3666, opened 2021-01-07
No Vyper import parser, so Vyper must recompile everything


haltman-at:
## Issue

Because we don't have a Vyper import parser, `compile-vyper` is forced to assume that everything could be a dependency.  So, the issue here is to write a Vyper import parser, and then to hook it up in `compile-vyper`'s `soucesWithDependencies` method (similar to how `compile-solidity` does things) so that Vyper only compiles what it needs to.  Note that hooking it up may require providing a few other function parameters?  I'm not entirely clear on how the profiler works but I gather it needs various functions passed in...

## Environment

* Truffle version (`truffle version`): 5.1.60

gnidan:
Fixed in [v5.1.64](https://github.com/trufflesuite/truffle/releases/tag/v5.1.64)

PR #3720, opened 2021-01-20
Enhancement: Add Vyper import analysis and ability to import from non-FS resolver sources (...and various other changes to how compilation works)

haltman-at:
Addresses #3666.   **This PR contains breaking changes to `@truffle/resolver`, `@truffle/compile-common`, and `@truffle/compile-vyper`.**  Also, @eggplantzzz, it means that the `resolveDependencyPath` on a `ResolverSource` has to be written slightly differently, which I assume will affect your work on EthPMv3.  Details below; skip to the bit where I say "here's the part relevant to you, @eggplantzzz" if that's all you want to know about.

Anyway!  This PR adds import analysis for Vyper, so that `compile-vyper` won't have to always compile every Vyper contract.  It also makes it possible for Vyper users to import from non-FS resolver sources, e.g., doing `from example.contracts import ExampleContract`, where `example` is an NPM module for a Vyper Truffle project.  (Although this latter is not possible if you're using an old non-JSON version of Vyper.)

This sounds simple enough, right?  Just write an import parser for Vyper -- which can work the same way as the Solidity one -- and then hook it into the same profiler and resolver machinery that Solidity uses, right?

Well, turns out no.  There turn out to be two big problems with this:

1. The Vyper compiler, unlike the Solidity compiler, will only return a single `FileNotFoundError` if imports fail, rather than all of them.
2. Our import analysis and resolution is based around the idea that imports are described by file paths, or, at least, something like file paths.  But in Vyper you import modules, without a `.json` or `.vy` ending specified, and the compiler is responsible for deciding which one to use.

In addition, we encounter a number of smaller problems:

3. The Vyper compiler will throw an error if you give it a source which is not also a compilation target.
4. The way that absolute imports are supposed to work in Vyper doesn't combine well with the JSON input format (I'll describe this problem in more detail below).
5. The `sources()` function in `compile-vyper` didn't actually do the same thing as the `sources()` method in `compile-solidity`??
6. We don't want `abi-to-sol` getting in the way of Vyper's native ability to import JSON; we want to be able to resolve JSON as JSON, not as Solidity

Let's start with problem (1).  Because we can't use the Solidity method, we'll have to write an import parser ourselves.  Fortunately, Vyper has a much more rigid syntax than Solidity, so this is possible (without having to do all the work of writing a proper Vyper parser).  Yes, that's right, we're parsing with regular expressions.  But I think it should be good enough in this case.

Really, the only tricky thing is the problem of line continuations.  This would be straightforward -- merge adjacent lines with a space when a line ends in a `\` -- except that line continuation doesn't work at the end of a comment line... and whether a given `#` starts a comment depends on whether it's inside of a string literal.

Thankfully, a fairly naive approach is good enough here.  First we strip away any backslashes that occur on a line anywhere after `#`, then we perform line continuation.  Sure, this will be wrong if the `#` was inside of a string literal, but it won't matter!  Valid imports don't contain string literals, so this process won't break any valid imports.  Nor will it create any invalid ones from other pieces, as we never join two lines that shouldn't be, only fail to join some ones that should be.  Since the imports are all we care about, this is good enough.

(@gnidan, you may notice this process is a little different from what I described to you earlier, and which could be tripped up by backslashes at the end of comments; as you can see, I've made some refinements since then. :) )

Note that at the moment malformed imports may generate some weird error messages, but I'm planning to submit another PR soon to address #3668, so that problem will go away once that's merged.  (I wanted to do that as part of this PR, but @gnidan preferred it separate. :P )

Anyway, parsing imports is a pretty simple matter.  Note that what's ultimately returned is a module name, not a file path; translation to a file path is handled later.

Some additional notes on this step:
* If the `from a import b` syntax is used, the module name returned will be `a.b.`.  If `a` here is instead just a sequence of dots, then no dot is added inbetween (`from . import b` should parse to `.b`, not `..b`).
* No import is returned for statements of the form `from vyper.interfaces import ...`.  In Vyper, `vyper.interfaces` acts as something of a standard library of interfaces.  Obviously, we don't want to go trying to resolve that, so we filter such imports out.

This leaves big problem number (2).  But first, let's talk about the smaller problems (3) and (4).

Problem (3) initially prompted me to change `requiredSources()` so that it truly only returned the required sources, rather than all sources.  This could be considered a breaking change, but it's worth noting that it doesn't actually break anything; we've always been using `requiredSources` to, y'know, get the required sources, and the fact that it would return all the sources instead was something of a misfeature.  This change actually turned out not to solve problem (3), but I've left it in, rather than taking it out and turning it into a separate PR.

While doing this, I noticed that `requiredSources()` can actually resolve or import-parse the same source multiple times, which isn't great, and would be even worse with my addition.  So I added a caching mechanism to prevent such re-resolution or re-parsing.

Anyway, the real solution to problem (3) was to alter `vyper-json.js` so that it filtered out all sources that aren't targets and put them under `interfaces` rather than `sources`.  This also allowed me to simplify the output selection; it now *always* sets the only target to be `"*"`, because any source that's not a target no longer appears in `sources` in the first place!

What about problem (4)?  So, first I need to describe the problem.  Say you use an import in a Vyper file that isn't explicitly relative, like, say, a simple `import ImportedInterface as ImportedInterface`.  Where will Vyper look for this?  It will try interpreting it as a relative import, but it will also try interpreting it as relative to the project root (which I am interpreting in our case as the contracts directory).  So we need to support that behavior.

Now, with `vyper` (not `vyper-json`), this just means passing the `-p` option to it with the contracts directory.  But with `vyper-json`, it's not that simple.  When looking for an absolute import -- say, `ImportedInterface` above -- it'll truly treat it as an absolute import, and look in `/ImportedInterface.json` and `/ImportedInterface.vy`.  And it doesn't have a real equivalent of the `-p` option that we can use to inform it of the project root.  Oh, sure, it has a `-p` option, but it doesn't do what we want -- rather than changing how the compiler looks for sources inside the JSON, instead it causes the compiler to actually search the file system.  We don't want that!

Fortunately, @gnidan came up with a solution to this problem.  I've added an optional `contractsDirectory` argument to the `collectSources()` function in `compile-common`, which `vyper-json.js` passes.  If this is passed, any sources that begin with the contracts directory will have this prefix stripped off -- that is to say, in the input to the Vyper compiler, we actually pretend that the contracts directory is the root directory.

(Hm, it just occurred to me to wonder if that will screw up error messages.  Oh well, we can fix that later...)

This does mean that if you try to import, say, a Vyper NPM module, and it uses absolute imports in this way, it may not work here.  But I think that's an acceptable caveat.  If you're writing something to be imported by other people, use explicitly relative imports!

As for problem 5, um, oy, man, I did not realize that was different previously!  Well, I fixed it now... worth noting though that if you're using a non-JSON version of Vyper, the source texts passed in will be ignored, in favor of the Vyper compiler just reading things from the file system.

But now let's get back to the big problem: Vyper imports don't work like Solidity imports, because a single module might be found in any of several possible file paths.  This really screws with the assumptions made in `compile-common`.  In order to handle this, I decided to expand the options to the resolver.

I also had to do this to handle problem (6)!  So, previously, the `Resolver` constructor took only one option beyond the config, a boolean saying whether to include Truffle sources.  Now instead of a boolean, it takes an object of named options.  There's the option to include the Truffle sources, but there's now also an option to turn off the ABI resolver (so that JSON can be handled by FS instead).  (Actually, really, there's an option to turn *on* the ABI resolver, but it defaults to `true`, so...)  And then there's the third option, the one added to handle problem (2): The one to turn on the Vyper resolver source.

Before I describe that, though, I want to point out that this option-based design is not really ideal.  The problem, essentially, is that there's no way to build resolvers from other resolvers.  Consider what happens if you write Solidity tests.  Then `truffle test` creates a special resolver and puts this in the config.  Now, `compile-solidity` just uses whatever resolver is in the config, so this works.  But imagine if we supported writing tests in Vyper.  Then `truffle test` would create this special resolver, and then `compile-vyper` would... throw it away and create its own resolver so that it can turn on the special Vyper options.  That's no good!  What we need is a way to make resolvers from other resolvers.  So, this design works for the moment, but something better may have to be figured out in the future.

Anyway, let's talk about that option to turn on the Vyper resolver source, because it is a doozy!

The Vyper resolver source... well, so, my original plan for it was that it would act as a wrapper around another resolver source.  I'm actually hoping to do the same thing for the ABI resolver in a future PR, so that it's not limited to just the file system, although I don't want to do that until 5.2 is out since I know that @eggplantzzz is currently actively working on that stuff and I don't want to step on his toes.

Unfortunately, due to problems with `resolveDependencyPath`, this design didn't work, and so now instead the Vyper resolver source acts as a wrapper around *all* the resolver sources.  If you turn on this resolver source, the `Vyper` resolver source will swallow the other resolver sources and replace them:
```typescript
if (resolveVyperModules) {
  this.sources = [new Vyper(this.sources, options.contracts_directory)];
}
```
Yup.  Just like that.

So how does the Vyper resolver source work?

Well, for `require` it just returns `null`; it's never going to be used for `require`, after all, so whatever.  Although, I could instead alter it to just defer to the resolver sources it ~~ate~~ wraps if that's for some reason be deemed necessary?

Anyway, for `resolve`... well, you're basically going to want to see the code for this one!  But basically:
1. First it asks, is this actually just a file path, and tries deferring to the wrapped resolvers.
2. If that fails, it checks whether it could plausibly be a Vyper module, and if not, give up and exits.
3. Otherwise, if it could be, it translates the Vyper module name to a file path, but without extension.
4. It then attempts to resolve the resulting file path as either JSON or Vyper, and via one of the three methods of treating it as a relative import, treating it as relative to the contracts directory (if it wasn't explciitly relative), and just taking it as-is (if it wasn't explicitly relative).  JSON is prioritized over Vyper.  Each possibility is tried with each wrapped resolver to see if one of them will work.

(Note that the just taking it as-is option is what allows for it to work with, e.g., the NPM resolver source.)

So why does it work this way, rather than wrapping an individual resolver source, and just using that one?  In earlier versions of this PR it did, and the code I quoted above was instead
```typescript
if (resolveVyperModules) {
  this.sources = this.sources.map(source => new Vyper(source, options.contracts_directory));
}
```
So why not that?

Well, to understand that, we have to talk about `resolveDependencyPath`.  The problem is that `resolveDependencyPath` is a problem for the `Vyper` resolver source.  In fact originally I was planning on getting rid of this mechanism altogether but that caused even more problems.

For other resolver sources, `resolveDependencyPath` is simple; the source doesn't need to actually do a resolution to determine how to join two paths together.  But for `Vyper`, well, the only way to determine a canonical path for a dependency is to attempt to actually resolve it!  So that's what it does.

Note that this means that its `resolveDependencyPath` is actually `async`; I modified the types and code appropriately to allow this.  So, uh, yeah, `resolveDependencyPath` can be `async` now.

Also, to prevent re-resolution (once from `resolveDependencyPath` and once from `resolve`), I added a cache to the `Vyper` resolver source...

But that alone doesn't solve our problems!  Because, as mentioned above, in Vyper, unlike in Solidity, an "absolute" import can in fact be relative!  This means that I had to remove the mechanism that would keep `resolveDependencyPath` from being invoked unless it was dealing with an explicitly relative path, because now there can be *implicitly* relative paths too.

This means that (and here's the part relevant to you, @eggplantzzz) this now has to be handled by `resolveDependencyPath` itself.  So, **every ordinary resolver source is now responsible for checking, in its `resolveDependencyPath` method, whether the path passed in is explicitly relative, and returning it unchanged if it's not**.  Be aware!

So, why does the `Vyper` resolver source have to ~~eat~~ wrap the full array of sources?  Because since `resolveDependencyPath` has to attempt a resolution, if it didn't, it would have no way for `Vyper(FS(...))` to defer to `Vyper(NPM(...))` or etc.  So... yeah.

Now you might say, why not just keep an unswallowed `FS` resolver source -- or all of the original resolver sources -- and put those ahead of the Vyper resolver sources?  Well, that wouldn't work either, because their `resolveDependencyPath` methods would have no idea how to handle Vyper modules in place of file paths.  So... yeah.  We ended up with a single `Vyper` resolver source that ~~eats~~ wraps all the others.

Anyway I think that about sums it up??  Also I copied the relevant Solidity tests and modified them for Vyper.  Note that I didn't test modifying the JSON file, as that actually won't work -- if you compile JSON with Vyper you won't get an artifact for it, so (due to #3702) we don't have a way to tell when such files were updated.  Oh well.

But uh yeah, here it is finally...

**EDIT**: I forgot to mention!  I changed `requiredSources` so that, if told to compile everything, it will return everything as the compilation targets, not `[]` with a special meaning!  Sorry, I don't remember why I made this change, just that the old way was causing problems... I think it was that it made it too difficult to tell whether to skip compilation.  However, I left this special meaning of `[]` present in `compile-solidity/run.js` and `compile-vyper/vyper-json.js` because having it there didn't cause a problem, and changing it would have been more trouble than it's worth.  I forget the details though.


haltman-at:
OK, I updated the failing test in `core`.  Since it had a note on saying "TODO: move this to compile-solidity", I did that as well. :)

packages/compile-common/src/profiler/isExplicitlyRelative.ts
```diff
@@ -1,3 +1,3 @@
 export function isExplicitlyRelative(importPath: string): boolean {
-  return importPath.startsWith(".");
+  return importPath.startsWith("./") || importPath.startsWith("../");
```
eggplantzzz, line 2: Will there ever be a valid relative path that looks like `"folder/to/thing"`?

haltman-at: The Solidity documentation says:

> All path names are treated as absolute paths unless they start with the current `.` or the parent directory `..`.

So, in Solidity, this is not possible.  In Vyper, however, it is possible, which is why I had to change where this is checked.

Note that the change to this line of code doesn't relate to that though -- this change was just me fixing a bug I caught (since e.g. `".hiddenfolder/to/thing"` shouldn't be treated as a relative path).

packages/compile-vyper/index.js
```diff
@@ -178,53 +171,101 @@ async function compileAllNoJson({ sources, options, version }) {
...
+    }
+
+    //having gotten the sources from the resolver, we invoke compileJson
+    //ourselves, rather than going through Compile.sources()
+    Compile.display(compilationTargets, options);
+
+    const { version, json: useJson } = await checkVyper();
+
+    if (useJson) {
+      return compileJson({
```
eggplantzzz, line 234: What do you think about these two methods `compileJson` and `compileNoJson` having a different signature?

packages/compile-vyper/index.js
```diff
@@ -178,53 +171,101 @@ async function compileAllNoJson({ sources, options, version }) {
...
-    const files = await findContracts(fileSearchPattern);
-    debug("files: %O", files);
+    // no vyper targets found, no need to check Vyper
+    if (vyperFilesStrict.length === 0) {
+      return { compilations: [] };
+    }
 
-    return await Compile.sources({
-      sources: files,
+    return await Compile.sourcesWithDependencies({
```
eggplantzzz, line 267: We change this to account for JSON imports, right?

packages/compile-vyper/parser.js
```diff
@@ -0,0 +1,72 @@
+const debug = require("debug")("compile-vyper:parser");
+const OS = require("os");
+
+function parseImports(body) {
+  // WARNING: We're going to do this with crudely with regexes!!
```
eggplantzzz, line 5: Not that it matters too much but your comment is grammatically incorrect

packages/compile-vyper/profiler.js
```diff
@@ -0,0 +1,33 @@
+const debug = require("debug")("compile-vyper:profiler");
+const Common = require("@truffle/compile-common");
+const Resolver = require("@truffle/resolver");
+const { parseImports } = require("./parser");
+
+// Returns the minimal set of sources to pass to solc as compilations targets,
```
eggplantzzz, line 6: "pass to vyper"

packages/resolver/lib/sources/vyper.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    //only attempt this if what we have looks like a Vyper module
+    if (!importModule.match(/^[\w.]+$/)) {
+      debug("clearly not a Vyper module");
+      return { body: undefined, filePath: undefined };
+    }
+
+    const importPath = moduleToPath(importModule); //note: no file extension yet
+    debug("importPath: %s", importPath);
+    const explicitlyRelative = importModule[0] === ".";
```
eggplantzzz, line 52: Is there any reason to add the same check as to whether it starts with "./" or "../" like in the other utility method mentioned?

packages/resolver/lib/sources/npm.ts
```diff
@@ -74,6 +74,12 @@ export class NPM implements ResolverSource {
   // we're going to resolve it to some_module/contracts/AnotherContract.sol, ensuring
   // that when this path is evaluated this source is used again.
   resolveDependencyPath(import_path: string, dependency_path: string) {
```
eggplantzzz, line 76: This is not a necessary one but if you feel like updating `dependency_path` to `dependencyPath` it would make me feel ever so slightly better :)

packages/compile-vyper/index.js
haltman-at: It's a bit awkward, to be sure, but I'm not sure how to do better here.  In the no-JSON case, we're basically stuck passing paths to the compiler rather than full source texts.  So, I think it makes sense in context that they have different signatures?  It's not like either will be called directly by the outside.

packages/compile-vyper/index.js
haltman-at: Not just JSON imports, but any imports that aren't in the contracts folder, same as with Solidity.  E.g. if you want to import Vyper from an NPM module -- with this PR, you can now do that!  But of course that requires getting the dependencies...

packages/compile-vyper/parser.js
haltman-at: Even the comments are done with crudely, I guess XD

(But yeah I can go back and fix this)

packages/compile-vyper/profiler.js
haltman-at: Oops thanks, will have to fix that

packages/resolver/lib/sources/vyper.ts
haltman-at: Good question! :)  The answer is that this is actually deliberate, because we're checking here the unprocessed `importModule`, rather than the converted `importPath`.  Whereas `importPath` starting with a `.` could just mean a hidden file or directory, `importModule` starting with a `.` definitely indicates a relative path (see the code below for `moduleToPath`).

Maybe I should add a comment pointing this out.

packages/resolver/lib/sources/npm.ts
haltman-at: Oh ha yes, easy change :)

packages/compile-vyper/index.js
haltman-at: Actually, somewhat tangential thought -- I'm wondering if `sources` should just throw if JSON isn't available.  Because `sources` isn't called by `WorkflowCompile`, and, like, it's not going to do what the person using it likely expects.  I'm thinking maybe it should just throw... thoughts, @gnidan?

haltman-at:
Note: Since Tyler has marked that conversation as resolved, let me re-ask here the question: Should `sources` throw if `useJson` is false? (@gnidan?)

packages/resolver/lib/resolver.ts
```diff
@@ -1,28 +1,64 @@
...
+              options.contracts_build_directory
+            )
+          ]
+        : []),
       new FS(options.working_directory, options.contracts_build_directory)
     ];
 
-    if (includeTruffleSource) {
-      this.sources.unshift(new Truffle(options));
+    if (resolveVyperModules) {
```
gnidan, line 60: Have you gotten any ideas on how we could address this weird one-off behavior? I worry that code smells are starting to compound with https://github.com/trufflesuite/truffle/issues/2599.

gnidan (changes requested):
Looks pretty good @haltman-at for the amount of complexity it juggles!

One open question in case it was just a gotcha, but I'm good to approve apart from the one open question:

Yes, seems like it does make sense to have `.sources()` throw if JSON isn't supported.

packages/compile-common/src/sources.ts
```diff
@@ -10,13 +11,16 @@ import { Sources, CollectedSources } from "./types";
  */
 export function collectSources(
   originalSources: Sources,
-  originalTargets: string[] = []
+  originalTargets: string[] = [],
+  contractsDirectory: string = "" //only used by Vyper atm
 ): CollectedSources {
```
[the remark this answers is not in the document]
haltman-at: Ha!  We could if you think it's worth it?  I'm inclined to leave it as-is, but it could certainly be changed...

packages/compile-solidity/test/profiler/index.js
```diff
@@ -2,16 +2,14 @@ var assert = require("chai").assert;
 var fs = require("fs-extra");
 var glob = require("glob");
 var Box = require("@truffle/box");
-var Profiler = require("@truffle/compile-solidity/profiler");
+var Profiler = require("../../profiler");
 var Resolver = require("@truffle/resolver");
 var Artifactor = require("@truffle/artifactor");
 
-// TOOD: Move this to @truffle/compile-solidity!
```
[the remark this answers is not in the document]
haltman-at: I dunno.  Dependencies maybe?  Not fitting into how `compile-solidity` does its tests?  I just added the necessary devDependencies, disregarded uniformity, and moved it. :shrug:

packages/resolver/lib/resolver.ts
haltman-at: I mean the real answer is some sort of mechanism for making resolvers from existing ones, but I haven't seriously thought about how that would work.  The simplest way, I suppose, would be for resolvers to just expose the options they were created with... I mean, should I add that here?  If I do I'll want to go changing other parts of the code to actually make use of this new approach.

packages/compile-solidity/index.js
```diff
@@ -86,27 +90,27 @@ const Compile = {
...
+    );
+    if (solidityTargets.length === 0) {
       return { compilations: [] };
     }
-    const hasTargets = compilationTargets.length;
 
-    hasTargets
-      ? Compile.display(compilationTargets, options)
-      : Compile.display(allSources, options);
+    Compile.display(compilationTargets, options);
```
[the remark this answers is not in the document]
haltman-at: I made it so that the profiler will no longer return `[]` for the targets to indicate compile everything; instead it will just return everything as the target.  So, no more need for a case split here.

...ah, crap, I just realized I forgot to mention that at all in the PR writeup.  Guess I'll go back and edit that in...

packages/compile-common/src/sources.ts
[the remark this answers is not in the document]
gnidan: Sure, let's do it! (if only so that no one ever needs to remember whether sources or targets go first)

packages/compile-solidity/index.js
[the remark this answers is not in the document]
gnidan: I went back and re-read the whole writeup to see if I could find the explanation :)

packages/compile-solidity/test/profiler/index.js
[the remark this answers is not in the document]
gnidan: And no circular deps?

packages/resolver/lib/resolver.ts
gnidan: No definitely any further change is out of scope.

packages/compile-solidity/index.js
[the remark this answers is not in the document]
haltman-at: Ha, I just stuck it at the end.  This PR, and its writeup, was a mess, and I had a hard time trying to explain it coherently.  So no way was I going to try to edit this in somewhere good in the body!

packages/compile-solidity/test/profiler/index.js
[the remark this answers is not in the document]
haltman-at: There were not!

packages/resolver/lib/sources/vyper.ts
haltman-at: (Comment has since been added. :) )

gnidan (approved):
bingo bango

packages/compile-common/src/sources.ts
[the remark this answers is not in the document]
haltman-at: OK changing this turned out to be slightly nontrivial so we have deemed it not worth it and are leaving it as-is after all.

haltman-at:
Note: Merging despite yellow because the remaining ones are allowed failures!

PR #3724, opened 2021-01-20
Change "websockets" -> "websocket"

araskachoi:
the key `websockets` throws an error and needs to be changed to `websocket` to work. The commented out key should be corrected.


gnidan (approved):
Thanks for this!

PR #3725, opened 2021-01-20
Internal improvement: Add dummy CircleCI job to prevent failure from no config

haltman-at:
Adds a dummy CircleCI job so that CircleCI doesn't fail out due to no config being found.

We could probably just have a config file with no jobs or workflows in it, but, eh, erring on the safe side I guess.


gnidan (approved):
ya ok

PR #3726, opened 2021-01-21
Organize and document @truffle/db's API

gnidan:
This PR has a whole bunch. Most of the re-organization has been squashed into one large first commit because it was all related.

ℹ️ See [rendered docs](https://ipfs.io/ipfs/QmPbB4HdkVvatMCAWhNvT1mrX2sTerxEyqxFZjKyv1c2WM/)

This includes:

- Create a unified "meta system" in src/meta

- Move Workspace (et al.) to src/meta/data
  ... since this is used by more than just pouch (graphql needs it)

- Move graphql into meta system
  - Start defining common Definition/Definitions types
  - Start making unified `forDefinitions` constructor
  - Update src/schema to call this unified function, so that it's easy to remove src/schema later
  - Rename this module `graph` to conform to how we have `pouch` (and to eliminate casing debate)

- Move pouch into meta system proper
  - Expand meta Definitions type to include Pouch.Definitions
  - Update primary `forDefinitions` constructor to include pouch
  - Update src/workspace to passthru to this `forDefinitions`
  - Stop hand-rolling Definitions type in src/resources/types
  - Cleanup proper home for external Workspace type to remove from src/workspace

- Define submodule for pouch adapters and move that logic into it

- Move connect/serve logic to meta system
  - Define `forAttachAndSchema` in src/meta/interface
  - Use this function in unified `forDefinitions`
  - Make `src/connect` and `src/server` into simple passthroughs

- Move process into meta system

- Redo non-meta modules (e.g. src/db is replaced with src/system)

- Convert DataModel to explicit import (no longer available as ambient)

- Update DataModel code generation to include doc comments and to use consistent casing (e.g. `Abi` not `ABI`)

- Favor `Resource<"..."> instead of `DataModel.(...)` (e.g. `Resource<"sources">`)

- Rework IdObject to be generic as `IdObject<N extends CollectionName>` instead of `IdObject<R extends Resource<N extends CollectionName>>`

- Set up TypeDoc
  - Initialize config for default theme
  - Leave some configuration for google/typedoc-neo-theme in case we want to use that later

- Document things broadly from top to bottom


gnidan:
Calling this a draft until I write some usage examples and make another attempt at improving that introduction. In the meantime, there's a lot here, so requesting reviewers in case it'd help to get started sooner. (This is ready for review apart from those two things)

packages/db/src/index.ts
```diff
@@ -1,11 +1,264 @@
...
+ *
+ * At a high-level, this package provides a
+ * [GraphQL](https://graphql.org/) interface and stores data via one of
+ * several persistence backends thanks to [PouchDB](https://pouchdb.com/).
+ * Use of this package directly is intended mostly for other tools – end users
+ * can find interfaces to @truffle/db by way of Truffle itself, e.g. with the
+ * `truffle db serve` command that starts a GraphQL Playground HTTP server.
+ *
+ * This documentation serves to organize the modules and namespaces included
+ * in the package, both for tool-developer reference and for continued work
```
fainashalts, line 27:
```suggestion
 * in the package, both for tool developer reference and for continued work
```

packages/db/src/index.ts
```diff
@@ -1,11 +1,264 @@
...
+ *
+ * \@truffle/db makes its GraphQL schema available as the exported
+ * [[Graph.schema | `Graph.schema`]] variable, or view the SDL
+ * details in the [[Graph]] namespace description.
+ *
+ * ## Data model
+ *
+ * ### Structure
+ *
+ * Data is organized as a number of collections of representations of key
```
fainashalts, line 95:
```suggestion
 * Data is organized as collections of representations of key
```

packages/db/src/index.ts
```diff
@@ -1,11 +1,264 @@
...
+ * In addition, please see the [[Resources]] module for handy helper types for
+ * dealing with @truffle/db entities.
+ *
+ * ### Truffle-specific interface
+ *
+ * This package also provides an abstraction to interface with other
+ * Truffle data formats, namely `WorkflowCompileResult`, returned by
+ * \@truffle/workflow-compile, and the Truffle contract artifacts format,
+ * defined by @truffle/contract-schema. This abstraction covers two classes:
+ *   - [[Project.Project]] for operations that **do not** require a network
```
fainashalts, line 170: Really important distinction. I like how this turned out!

packages/db/src/index.ts
```diff
@@ -1,11 +1,264 @@
...
+ *     connection. Use function [[Project.initialize | `Project.initialize()`]]
+ *     to create.
+ *   - [[Project.ConnectedProject]] for operations that **do** require a
+ *     blockchain network. Use existing project abstraction's
+ *     [[Project.connect | `project.connect()`]] method to create.
+ *
+ * ### HTTP interface
+ *
+ * This package exposes the [[serve | `serve()`]] function, which returns an
```
fainashalts, line 179: This is a bit unclear. I suggest clarifying that `serve` is part of a truffle command that can be used to access this interface.

packages/db/src/project/index.ts
```diff
@@ -5,80 +5,147 @@ import type { Provider } from "web3/providers";
...
-   * is not supported directly)
+   * [[DataModel.Network | Network]] and [[DataModel.Contract | Contract]]
+   * resources (e.g., naming [[DataModel.ContractInstance | ContractInstance]]
+   * resources is not supported directly)
+   *
+   * This saves [[DataModel.NameRecord | NameRecord]] and
+   * [[DataModel.ProjectName | ProjectName]] resources to @truffle/db.
+   *
+   * Returns a list of NameRecord resources for completeness, although these
+   * may be regarded as an internal concern. ProjectName resources are not
```
fainashalts, line 110:
```suggestion
   * may be regarded as an internal concern. `ProjectName` resources are not
```

fainashalts (approved):
This is so great, really enjoyed the thorough documentation and the way you explained all the parts of this code. You mentioned on slack you might have more changes, happy to re-review if you feel it's necessary!

packages/db/src/index.ts
gnidan: Thanks!

packages/db/src/index.ts
gnidan: From the perspective of the @truffle/db package, `serve` is just another interface. If people want to use @truffle/db without Truffle, they'll likely want to use this function on its own.

Any ideas on making this more clear?

packages/db/src/project/index.ts
gnidan: I haven't been putting those in backticks... roughly my style has been to add a hyperlink for the first usage in a section, then just style it as normal text subsequently. Think we should use a different style?

packages/db/src/index.ts
```diff
@@ -1,11 +1,264 @@
+/**
```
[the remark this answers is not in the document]
gnidan: Normally this might go into a README, but I want the README to remain to distinct from the API documentation.

I'm looking into options for how we can include arbitrary markdown files, but existing plugins don't seem to have compatibility with latest TypeDoc yet.

packages/db/src/index.ts
gnidan: Hm having trouble putting my finger on the distinction I was trying to make by framing it this way... I want to indicate that "collections" is a well-defined thing. I think you're right, though, that it's evident enough thanks to context.

Good suggestion!

packages/db/src/index.ts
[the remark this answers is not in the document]
gnidan: Hm. I tried doing this via the [`includes`](https://typedoc.org/guides/options/#includes) mechanism, but it turns out that you can't do `[[include:...]]` from insight a file that's already being `[[included:...]]`-ed.

I think keeping this here for now seems like the least bad option? I could move it to the end of the file; would that be better?

gnidan:
Merging this now to help the backlog of PRs targeting this branch :)

Thanks again @fainashalts!

PR #3731, opened 2021-01-26
Ensure `truffle db serve` does not exit

gnidan:
Since there's no longer any such thing as "just don't call the callback"


eggplantzzz (approved):
I think this is cool? Just leave the Promise hanging until the user forces it closed?

gnidan:
Exactly!

PR #3732, opened 2021-01-26
Use Pouch's `allDocs` method for workspace.all()

gnidan:
Instead of delegating to `workspace.find()` with no selector, since this existing approach doesn't seem to return complete information for all adapters. (I observed missing data against a running CouchDB)

### DIFF
--- a/packages/db/src/meta/pouch/adapters/base.ts
+++ b/packages/db/src/meta/pouch/adapters/base.ts
@@ -81,10 +81,23 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
     const log = debug.extend(`${collectionName}:all`);
     log("Fetching all...");
 
-    const result = await this.find<N>(collectionName, { selector: {} });
+    try {
+      const { rows }: any = await this.collections[collectionName].allDocs({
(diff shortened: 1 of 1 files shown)


packages/db/src/meta/pouch/adapters/base.ts
```diff
@@ -81,10 +81,23 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
     const log = debug.extend(`${collectionName}:all`);
     log("Fetching all...");
 
-    const result = await this.find<N>(collectionName, { selector: {} });
+    try {
+      const { rows }: any = await this.collections[collectionName].allDocs({
+        include_docs: true
```
fainashalts, line 86: I am sure this is necessary, but seems so redundant, haha.

fainashalts (approved):
LGTM!

packages/db/src/meta/pouch/adapters/base.ts
gnidan: By default, `allDocs` just returns `_id` and `_rev`, apparently

PR #3733, opened 2021-01-26
Remove Network.fork field from @truffle/db

gnidan:
Since this has long been replaced by the network genealogy system


fainashalts (approved):
Oops! Good find!

gnidan:
Thanks!

PR #3734, opened 2021-01-26
Clean up @truffle/db dependencies and imports

gnidan:
- Fix imports (remove unused, make sure things that can be devDeps are, etc.)
- Ensure we use `import type ...` instead of `import ...` everywhere this makes sense

### DIFF
--- a/packages/db/.madgerc
+++ b/packages/db/.madgerc
@@ -1,9 +1,15 @@
 {
   "excludeRegExp": [
     "\\.\\.",
+    "resources/data.ts",
     "test",
     "logger.ts"
   ],
   "fileExtensions": ["ts"],
+  "detectiveOptions": {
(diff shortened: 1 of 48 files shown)


fainashalts:
Why `import type`? Am I correct that this also makes the import a type declaration? Just curious!

fainashalts (approved):
You were right, this one was pretty straightforward!

gnidan:
> Why `import type`?

This tells TS that the import is only needed at compile time - it won't be used at runtime, so there's no need to create a corresponding `require()` statement in the output JS. This means that things like `@truffle/contract-schema` can be `devDependencies`.

Seems like this is good practice to maintain - TS will complain if your code uses an imported value, so now I'm thinking to just use `import type` by default and remove `type` as-needed.

PR #3735, opened 2021-01-26
Extract dedicated @truffle/db Network abstraction

gnidan:
This PR adds a new class-based "Network" abstraction, similar to the existing Truffle-specific Project abstraction.

Broadly speaking, this Network abstraction provides an interface for loading networks and their genealogies. See [rendered docs](https://ipfs.io/ipfs/QmVDzdSG1DA1oRKvdeonMdqpVEUpnb9QgxfCNXYiXCZfgb/#network-abstraction) for an overview of the new interface.

This consequently removes this logic from the Project abstraction (instead, having the Project abstraction simply defer to the Network abstraction).

Specifically, however, this makes a few changes to the underlying logic:
- Fix formatting of JSON-RPC requests to use proper hex string encoding
- Add `minimumHeight` and `maximumHeight` field to `Network.ancestors` and `Network.descendants`, respectively.
- Use this new `minimumHeight` filter to reduce the number of ancestor/descendant queries we make for _n_ input networks, meaning O(1) queries where we used to do O(n).
- Have the Network abstraction _always_ load the genesis block. This is new behavior, but we will want the genesis block. So make sure we load that straight away.


packages/db/src/resources/networks/index.ts
```diff
@@ -38,13 +38,15 @@ export const networks: Definition<"networks"> = {
         minimumHeight: Int # default any height
         includeSelf: Boolean # default false
         onlyEarliest: Boolean # default false
+        batchSize: Int # default 10
```
fainashalts, line 41: So `batchSize` is different from `limit`? I see where both are used but my spidey sense tells me to ask about this...

packages/db/src/network/test/index.spec.ts
```diff
@@ -40,7 +40,7 @@ describe("Network", () => {
         fc.boolean()
       ],
       async ({ model, batches }, disableIndex) => {
-        debug("run #%o", run++);
+        debug.extend("run")("run #%o", run++);
```
fainashalts, line 43: that's handy!

packages/db/src/network/test/index.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+        debug("ids %O", ids);
+
+        expect(ids).toEqual(
+          new Set(expectedLatestDescendants.map(({ id }) => id))
+        );
+      },
+
+      {
+        numRuns: testConfig.numRuns,
+        interruptAfterTimeLimit: testConfig.timeout * 0.8 // leave padding
```
fainashalts, line 115: Is this used anywhere?

packages/db/src/network/test/mockProvider.ts
```diff
@@ -0,0 +1,56 @@
+import { logger } from "@truffle/db/logger";
```
fainashalts, line 1: This is great! I wonder if we could/should mock the provider for where we test contractInstances? I think we actually use web3 right now...

fainashalts (approved):
Looks great! I made a couple of comments, but no show-stoppers!

packages/db/src/resources/networks/index.ts
gnidan: `limit` is "how many genealogy iterations", `batchSize` is "how many networks to load from persistence at one time". This is probably worth documenting, but since it's mostly internal, I figure we can ignore for now.

packages/db/src/network/test/index.spec.ts
gnidan: yeah!

packages/db/src/network/test/index.spec.ts
gnidan: yep, this is necessary for fast-check not to cause jest to timeout

packages/db/src/network/test/mockProvider.ts
gnidan: yeah that'd be nice follow-on work!

PR #3738, opened 2021-01-26
Add Solidity-generated internal sources to @truffle/db

fainashalts:
This PR adds `generatedSources` to Truffle DB's schema and tests. More information on `generatedSources` and their role in Source Mappings can be found here: https://docs.soliditylang.org/en/v0.8.0/internals/source_mappings.html


packages/db/src/project/compile/contracts.ts
```diff
@@ -50,13 +50,16 @@ export const generateContractsLoad = Batch.Contracts.generate<{
...
+
     return {
       name,
       abi,
       compilation,
       processedSource,
       createBytecode,
-      callBytecode
+      callBytecode,
+      generatedSources: generatedSources
```
haltman-at, line 62: Can be shortened :)

haltman-at (approved):
I don't quite understand it, this being Truffle DB stuff, but this all seems to essentially make sense to me?  May want to wait for another approval though. :P

packages/db/src/project/loadCompile/contracts.ts
```diff
@@ -79,3 +88,19 @@ export const process = Batch.Contracts.configure<{
     };
   }
 });
+
+function toGeneratedSourceInput({ generatedSources }) {
```
gnidan, line 92: Should be plural:

```suggestion
function toGeneratedSourcesInput({ generatedSources }) {
```

packages/db/src/resources/contracts.ts
```diff
@@ -33,6 +33,16 @@ export const contracts: Definition<"contracts"> = {
       processedSource: ProcessedSource
       createBytecode: Bytecode
       callBytecode: Bytecode
+      callBytecodeGeneratedSources: [GeneratedSource]
+      createBytecodeGeneratedSources: [GeneratedSource]
+    }
+
+    type GeneratedSource {
+      ast: AST!
+      id: Int!
```
gnidan, line 42: Can we call this `index` instead?

gnidan (commented):
Couple minor things but looks good otherwise!

packages/db/src/resources/contracts.ts
fainashalts: I don't think this is an index, is it? If so, can you point me to the docs? I wasn't able to find much about this except for what Solidity calls it (which is why I called it `id`) and what data type it is.

haltman-at: I can say for a fact that it's a source index; but as for documentation, I don't think there is any.  But when Solidity talks about source IDs, it means source indices.  Same as in the compiler output, you know?  In any case, it's definitely a source index and that's how the debugger uses it.

fainashalts: Thanks for the clarification Harry!

gnidan (approved):

PR #3740, opened 2021-01-27
Make .db global

eggplantzzz:
~Currently the `connect` and `serve` methods expect a working directory to be passed in to specify where to save db data. This PR makes `@truffle/db` default to the global Truffle data directory when neither a `@truffle/config` object nor something that conforms to `ConnectOptions` is supplied.~

~Followup work to this is to disallow a `@truffle/config` object as input to these two methods. So Truffle eventually should default to using the global Truffle data directory unless the user provides a different directory as input. That work will be completed in a different PR.~

This PR makes `ConnectOptions` the only valid input (optional) for `connect` and `serve`.


packages/core/lib/commands/db/commands/serve.js
```diff
@@ -16,13 +16,17 @@ const command = {
...
     const Config = require("@truffle/config");
-    const {serve} = require("@truffle/db");
+    const { serve } = require("@truffle/db");
 
     const config = Config.detect(argv);
     const port = (config.db && config.db.port) || 4444;
     const host = (config.db && config.db.host) || "127.0.0.1";
+    const serveOptions =
+      config.db && config.db.directory
+        ? { directory: config.db.directory }
```
gnidan, line 26: Let's see if we can try to remove the need for this, favoring instead (e.g.):

```javascript
module.exports = {
  // rest of truffle-config.js
  db: {
    adapter: {
      name: "sqlite",
      settings: {
        directory: "..."
      }
    }
  }
}
```

This way, there's no need to pass a directory if the adapter won't need it (like couch/memory). It'd mean there's the loss of a convenient way to just say "use this base directory", but I'm not sure the need to make that convenient? It's not that much extra work to add a couple layers of wrapping for something for something that I'm assuming won't be used much.

Anyway not sure if there's a clean way to get rid of this but I'll write thoughts about it as I scroll down.

packages/db/src/meta/interface.ts
```diff
@@ -32,12 +34,29 @@ export const forAttachAndSchema = <C extends Collections>(options: {
...
+      const configStore = new Configstore(
+        "truffle",
+        {},
+        { globalConfigPath: true }
+      );
+      const truffleDataDirectory = path.dirname(configStore.path);
+      options = {
+        directory: truffleDataDirectory,
+        adapter: (config || {}).adapter
+      };
```
gnidan, line 58: Might make sense to move this to its own module just to keep it separate/contained. How about:

```typescript
import { getDefaultDataDirectory } from "@truffle/db/meta/defaults";
```

This sounds like the sort of thing that shouldn't be defined in the meta system, and would instead be something we'd string through the `forDefinitions` functions (renamed like `forDefinitionsAndConfig`), but also seems fine to save that for later.

packages/db/src/meta/interface.ts
```diff
@@ -69,12 +88,27 @@ export const forAttachAndSchema = <C extends Collections>(options: {
...
+    }
```
gnidan, line 111: On the topic of getting rid of `directory` from `ConnectOptions`, that would mean we won't be able to have this logic here. We'll have to find another place to do this.

packages/db/src/meta/pouch/adapters/fs.ts
```diff
@@ -15,10 +15,8 @@ export interface DatabasesSettings {
   directory: string;
 }
 
-export const getDefaultSettings: GetDefaultSettings = ({
-  workingDirectory
-}) => ({
-  directory: path.join(workingDirectory, ".db", "json")
+export const getDefaultSettings: GetDefaultSettings = ({ directory }) => ({
+  directory: path.join(directory, ".db", "json")
```
gnidan, line 19: Going off the wrapper idea, [it's a bit hacky, but] we could just make this not take any options and instead just call the wrapper.

(This is why I was getting at the business about making this something that gets passed to `forDefinitions`)

packages/db/src/meta/pouch/adapters/types.ts
```diff
@@ -57,7 +57,7 @@ export namespace Generic {
     A extends Adapters,
     N extends AdapterName<A> = AdapterName<A>
   > {
-    workingDirectory?: string;
+    directory?: string;
```
gnidan, line 60: Besides removing this field here **and** in `ConnectOptions`, we could remove it from `ConnectOptions` but have `AttachOptions` require it;  we could have `Pouch.Adapters.concretize` do the the default data directory look up and pass that to `getDefaultAdapters`.

Don't have a strong opinion on which way is better, but it seems like low harm and less code just to have each adapter be responsible for the lookup... so I think I slightly prefer that? Curious your thoughts if you have any.

packages/db/src/system.ts
```diff
@@ -47,10 +47,12 @@ export type ConnectOptions = Meta.ConnectOptions<Collections>;
 
 const system = Meta.forDefinitions(definitions);
 
-export const connect: (options: TruffleConfig | ConnectOptions) => Db =
-  system.connect;
+export const connect: (
+  options: TruffleConfig | ConnectOptions | undefined
```
gnidan, line 51: More normally, @truffle/db code does this sort of thing this way:

```suggestion
  options?: TruffleConfig | ConnectOptions
```

packages/db/src/system.ts
```diff
@@ -47,10 +47,12 @@ export type ConnectOptions = Meta.ConnectOptions<Collections>;
...
+) => Db = system.connect;
 
-export const serve: (options: TruffleConfig | ConnectOptions) => ApolloServer =
-  system.serve;
+export const serve: (
+  options: TruffleConfig | ConnectOptions | undefined
```
gnidan, line 55: (Another suggestion for batch :)

```suggestion
  options?: TruffleConfig | ConnectOptions
```

packages/db/src/meta/interface.ts
```diff
@@ -22,7 +24,7 @@ export interface Db<_C extends Collections> {
 }
 
 export interface ConnectOptions<_C extends Collections> {
-  workingDirectory: string;
+  directory?: string;
```
gnidan, line 27: Regarding my plan stated above, I'm thinking that we should try to remove this field entirely.

gnidan (commented):
Thanks @eggplantzzz; I'm excited that you've offered to get your feet wet with this package!

Here's some thoughts I have to run by you for this. Hopefully these comments aren't already covered by #3741 (I took a quick look there and didn't seem like it)

packages/core/lib/commands/db/commands/serve.js
eggplantzzz: This doesn't require that the user pass a directory, it will just use the directory that is supplied if it is present. However, if you think it is better to namespace it like this then I have no problem with that. Although I suppose the directory is not adapter-specific so maybe it doesn't belong in `adapter`. What do you think?

packages/db/src/meta/interface.ts
eggplantzzz: Why would we get rid of `directory` from `ConnectOptions`? We'll still need to allow the user to input a directory if they want to.

packages/db/src/system.ts
eggplantzzz: Ah yes, this is what I wanted.

packages/core/lib/commands/db/commands/serve.js
gnidan: My thinking that it's adapter-specific is because that field only ends up getting used by two out of the four adapters

eggplantzzz: Ok, that makes sense

packages/db/src/meta/interface.ts
eggplantzzz: resolved

packages/db/bin/server.js
```diff
@@ -1,14 +1,8 @@
...
 const { serve } = require("@truffle/db");
 
 const port = 4444;
 
-const config = Config.detect({
-  workingDirectory: process.argv[2] || process.cwd()
-});
-
-const server = serve(config);
+const server = serve();
```
gnidan, line 5: Should we still pass the config info here? @fainashalts might be using that functionality.

gnidan (approved):
Looks great!

packages/db/src/network/test/setup.ts
```diff
@@ -3,18 +3,13 @@ const debug = logger("db:network:test:setup");
 
 import { connect } from "@truffle/db";
 
-export const setup = async (options: {
-  identifier: string;
-}) => {
+export const setup = async (options: { identifier: string }) => {
   // @ts-ignore
```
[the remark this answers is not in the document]
eggplantzzz: okie doke!

packages/db/bin/server.js
eggplantzzz: So what was the result of our short conversation last week about this? I'm not sure we're using anything from `config`?

gnidan: We might want to still allow passing in that path, for the use case of running `yarn start <dir>` with a truffle-config that overrides the default.

eggplantzzz: Keeping the `Config.detect`!

packages/db/bin/server.js
```diff
@@ -3,7 +3,9 @@ const { serve } = require("@truffle/db");
 
 const port = 4444;
 
-const server = serve(Config.detect());
+const server = serve(Config.detect({
+  workingDirectory: process.argv[2] || process.cwd()
```
gnidan, line 7: this works? doesn't need to be `working_directory`? that'd be swell!

eggplantzzz: Yep! I actually just reverted it back to what it was before I edited it but just confirmed myself that `@truffle/config` handles this. See https://github.com/trufflesuite/truffle/blob/develop/packages/config/src/index.ts#L147-L149

PR #3741, opened 2021-01-27
Remove config from db

eggplantzzz:


packages/db/package.json
```diff
@@ -59,7 +59,6 @@
   "devDependencies": {
     "@gql2ts/from-schema": "^2.0.0-4",
     "@truffle/compile-common": "^0.5.1",
-    "@truffle/config": "^1.2.34",
```
gnidan, line 62: Nice!

packages/db/src/meta/interface.ts
```diff
@@ -34,15 +33,9 @@ export const forAttachAndSchema = <C extends Collections>(options: {
 }) => {
   const { attach, schema } = options;
 
-  const connect = (
-    config: TruffleConfig | ConnectOptions<C> | undefined
-  ): Db<C> => {
-    // TODO: get rid of @truffle/config stuff and rename field to "directory"
+  const connect = (config: ConnectOptions<C> | undefined): Db<C> => {
```
gnidan, line 36: I believe everywhere else uses `?` instead of `| undefined`:

```suggestion
  const connect = (config?: ConnectOptions<C>): Db<C> => {
```

gnidan: Oh oops I guess technically this should've gone on the other PR. Well whatever, pick whichever way you want the merge conflict ;)

packages/db/src/meta/interface.ts
```diff
@@ -88,13 +81,9 @@ export const forAttachAndSchema = <C extends Collections>(options: {
     };
   };
 
-  const serve = (config: TruffleConfig | ConnectOptions<C> | undefined) => {
-    // TODO: get rid of @truffle/config stuff and rename field to "directory"
+  const serve = (config: ConnectOptions<C> | undefined) => {
```
gnidan, line 84: (Same as above)

```suggestion
  const serve = (config?: ConnectOptions<C>) => {
```

packages/db/src/meta/interface.ts
gnidan, line 36: Oh also, let's call this `options` since there's no longer anything named config here.

```suggestion
  const connect = (options: ConnectOptions<C> | undefined): Db<C> => {
```

packages/db/src/meta/interface.ts
gnidan: (Same as above for the other above)

```suggestion
  const serve = (options: ConnectOptions<C> | undefined) => {
```

packages/db/src/system.ts
```diff
@@ -47,12 +46,10 @@ export type ConnectOptions = Meta.ConnectOptions<Collections>;
 
 const system = Meta.forDefinitions(definitions);
 
-export const connect: (
-  options: TruffleConfig | ConnectOptions | undefined
-) => Db = system.connect;
+export const connect: (options: ConnectOptions | undefined) => Db =
```
gnidan, line 49: (Same as above, re: optionals)

```suggestion
export const connect: (options?: ConnectOptions) => Db =
```

packages/db/src/system.ts
```diff
@@ -47,12 +46,10 @@ export type ConnectOptions = Meta.ConnectOptions<Collections>;
...
+  system.connect;
 
-export const serve: (
-  options: TruffleConfig | ConnectOptions | undefined
-) => ApolloServer = system.serve;
+export const serve: (options: ConnectOptions | undefined) => ApolloServer =
```
gnidan, line 52: (Same as above, re: optionals)

```suggestion
export const serve: (options?: ConnectOptions) => ApolloServer =
```

packages/truffle/test/scenarios/compile/compile.js
```diff
@@ -328,8 +329,14 @@ describe("Compilation with db enabled", async () => {
...
-  function checkForDb(config) {
-    const dbPath = path.join(config.working_directory, ".db");
+  function checkForDb() {
+    const configStore = new Configstore(
+      "truffle",
+      {},
+      { globalConfigPath: true }
+    );
+    const truffleDataDirectory = path.dirname(configStore.path);
+    const dbPath = path.join(truffleDataDirectory, ".db");
```
gnidan, line 339: We could use `TruffleConfig.getDefaultDataDirectory()` here instead, in order to express more directly the idea that the behavior should match Truffle's user-level config.

packages/truffle/test/scenarios/compile/compile.js
```diff
@@ -361,7 +368,7 @@ describe("Compilation with db enabled", async () => {
   it("creates a populated .db directory when db is enabled", async function () {
     this.timeout(12000);
 
-    const dbExists = checkForDb(config);
```
gnidan, line 364: Later gater!

packages/truffle/test/scenarios/compile/compile.js
```diff
@@ -378,12 +385,15 @@ describe("Compilation with db enabled", async () => {
...
+    const db = connect();
     const results = await db.execute(GetAllContracts, {});
 
     // number of contracts matches number of contracts in the project directory
     // (plus one library in this one)
-    assert(results.data.contracts.length === 4);
+    const names = ["Contract", "InnerLibrary", "Migrations", "RelativeImport"];
+    for (const name of names) {
+      assert(results.data.contracts.some(contract => contract.name === name));
+    }
```
gnidan, line 396: Ooh a free test improvement :D

gnidan (changes requested):
Looks great! Just a couple minor things.

packages/truffle/test/scenarios/compile/compile.js
eggplantzzz: Well this also needed to be changed because it looked like other contracts were now showing up for this query since everything is kept here now.

packages/truffle/test/scenarios/compile/compile.js
eggplantzzz: This makes sense, so `@truffle/config` will become the `devDependency` and not `configstore`.

eggplantzzz: Hmm, maybe we should just get rid of the special `configstore` logic and use `TruffleConfig` in `@truffle/db` as well. It is either `@truffle/config` or `configstore`.

packages/db/src/meta/interface.ts
eggplantzzz: Ooh, I committed this but then realized that `options` is the argument passed to the parent function so no-can-do.

PR #3743, opened 2021-01-27
Add "Panic" to forbidden names in ABI arbitraries

haltman-at:
Since Solidity 0.8.1 is out now, which adds `Panic` to the syntax as something you can catch, I figured I'd add `Panic` to the forbidden words in the ABI arbitraries.  It's probably not actually reserved as such, but I figure it's good to err on the safe side and avoid any potential naming collision with any sort of Solidity-meaningful word.


gnidan (approved):
Easy!

PR #3744, opened 2021-01-28
Bug fix: Fix Truffle's Assert libraries for Solidity 0.8.1

haltman-at:
Solidity 0.8.1 wasn't supposed to be a breaking release, sure, but, well, there are certain conversions they meant to disallow in 0.8.0 but forgot to, so now they're disallowed in 0.8.1.  So, I've updated our Assert libraries once again to ensure that they still compile with Solidity 0.8.1.


gnidan (approved):
Sweet!

PR #3747, opened 2021-01-28
Dependency update: Update highlightjs-solidity to 1.0.21 (update syntax highlighting for Solidity 0.8.1)

haltman-at:
Updated highlightjs-solidity to highlight the word `Panic` now that you can catch those, this is just the PR here to now use this. :)


haltman-at:
(Forgot to run yarn at first, oops :P )

PR #3749, opened 2021-01-29
Bug fix: Rework SafeSend for increased compatibility in Solidity testing (fix Solidity testing for versions <0.4.22, remove warnings on >=0.7.0)

haltman-at:
Note: This PR changes how `SafeSend` works, which is arguably **a breaking change to `@truffle/resolver`**.  Dunno if you want to make sure this gets released together with #3720, since that also has breaking changes to `@truffle/resolver`.

This PR gets rid of `OldSafeSend` and `NewSafeSend` and goes back to having just a single `SafeSend`.  This new `SafeSend` should work, without warnings, on all versions of Solidity back to 0.4.3.  Previously, `OldSafeSend` only worked back to 0.4.22, meaning that `truffle test` would not work with older Solidity versions.  We don't actually support Solidity all the way back to 0.4.3 at the moment, only to 0.4.9, but still. :) 

So, why the change?  Well:
1. It was annoying having two versions of `SafeSend`,
2. `NewSafeSend` would cause warnings on versions >= 0.7.0,
3. As mentioned above, `OldSafeSend` didn't work for versions < 0.4.22.

However, the new `SafeSend` works differently from the older ones, in that the recipient is passed to, and the ether paid to, the `deliver` function, rather than to the constructor.  This is to avoid the need for an explicit constructor, because it's that explicit constructor that caused a bunch of the need for different versions.  Specifically:
1. On versions <0.4.22, there was no `constructor` keyword, and constructors were specified as a function whose name matched the contract.
2. On versions >=0.5.0, constructors must be specified with `constructor`.
3. On versions 0.5.x and 0.6.x, constructors require an explicit visibility.
4. On versions 0.7.x and later, putting an explicit visibility on a constructor causes a warning.

What a pain!  Better to get rid of the explicit constructor and have the recipient passed to, and ether paid to, `deliver` instead.  (Note that an explicit constructor is needed both for having the constructor taking an argument, and for having it be payable.)

But this leaves another problem, which is the original reason that `SafeSend` was split in two: The problem of `address payable`.  In versions >=0.5.0, the argument to `selfdestruct` must be of type `address payable`.  Whereas in versions <0.5.0, that type doesn't exist.  And if we have `recipient` merely of be of type `address` in version >=0.5.0 as well, then we'd need to somehow convert it to `address payable`, and that's its own can of worms.  In versions < 0.8.0, you'd do that by doing `address(uint160(recipient))`, which is 0.4.x-kosher (it's just a no-op there), but in 0.8.x, you'd have to do `payable(recipient)` instead, which is not 0.4.x-legal, so, uh, yeah.

Luckily @gnidan figured out a workaround for this: Use inline assembly!  By doing `assembly { selfdestruct(recipient) }`, you get something that works not only with 0.8.x but works all the way back to 0.4.3, which is further back than we currently support.  Any further back than that, and I think we'd need multiple versions again (or resort to dynamic generation). :P

But hey!  It's a new `SafeSend`, that handles all versions we support with no warnings!  So that's nice!


haltman-at:
Note: I just realized that the assertion libraries currently have pragmas specifying they're only for 0.4.15 and up.  I'll have to see if I can improve that... :)

haltman-at:
Btw, for the curious, this new singular `SafeSend`:

1. Won't work with versions <0.4.3 because before that `selfdestruct` had to be called `suicide`;
2. Moreover won't work with versions <0.4.0 because before that there was no `payable` keyword;
3. And moreover won't work with versions <0.3.1 because before that there was no inline assembly.

And perhaps there are other obscure reasons why it might not work with other older versions, but I don't know of them!

haltman-at:
Note: I just realized that the assertion libraries currently have pragmas specifying they're only for 0.4.15 and up.  I may have to see if I can improve that!  But I think that'll probably be a separate PR...

eggplantzzz:
This is great!

eggplantzzz (approved):

haltman-at:
Note: Holding off on merging this as @gnidan agrees this should be grouped with the Vyper imports work.

PR #3750, opened 2021-01-30
Bug fix: Allow resolution failure in resolveAllSources(); allow resolution of empty sources

haltman-at:
Addresses #3668.

This PR does two things:

Firstly, it allows resolution to fail in `resolveAllSources`.  This is so that resolution errors will show up to the user as compile errors rather than often-unhelpful Truffle errors.  To do this, the `resolve()` wrapper in `profiler.ts` is modified to include a try/catch, and a definedness check is added in `resolveAllSources`.  Note that I wanted to make sure unexpected errors get rethrown, but resolver doesn't currently throw structured errors; and, well, I didn't want to bother making it do so, this is just a quick fix, so I just checked the start of the message instead.

Secondly, it allows resolution of empty sources.  Currently we just checked whether `body` is truthy, which would mean that empty sources could not be resolved.  So I changed this everywhere to `body !== undefined`.  (This includes inside the Vyper resolver source, since that one acts as a wrapper.)  Note that the Truffle resolver source, unlike the others, would return `body: null` if nothing were found, rather than `body: undefined`, so I changed this resolver source to use `undefined` to signify failure like the others.

### DIFF
--- a/packages/compile-common/src/profiler/profiler.ts
+++ b/packages/compile-common/src/profiler/profiler.ts
@@ -55,8 +55,24 @@ export class Profiler {
 
     debug("paths: %O", paths);
 
-    const resolve = ({ filePath, importedFrom }: UnresolvedSource) =>
-      resolver.resolve(filePath, importedFrom);
+    const resolve = async ({ filePath, importedFrom }: UnresolvedSource) => {
+      //we want to allow resolution failure here.  so, if a source can't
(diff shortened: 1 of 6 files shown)


gnidan (approved):
Seems reasonable!

haltman-at:
Merging despite yellow because it's just Travis that hasn't finished yet!

PR #3751, opened 2021-01-30
Bug fix: Only compile Truffle Tests with required test libraries (and SafeSend)

haltman-at:
Currently, Solidity tests with Truffle Test always attempt to compile all the test libraries, even if they're not all used.  This is a problem because some of the test libraries might not work with really old versions of Solidity.  This means that, with really old versions of Solidity, you can't use Solidity tests, even if you're not using our test libraries.

This PR changes it so that only the ones the user actually imports get compiled (well, OK, and `SafeSend`, since we need that one for our own purposes).  Doing this is really simple: We just don't tell it to compile all of them!  OK, but what if they do import a test library, don't we need it then?  No!  The profiler and resolver handle that and ensure that any necessary test libraries get passed to the compiler.  Aside from `SafeSend`, there simply isn't a need to specify them specially!  I have tried Solidity tests that use our assertion libraries and they work fine on this branch.

So, now Solidity tests should work regardless of your Solidity version, since `SafeSend` works back to 0.4.3, and we only support back to 0.4.9 in the first place.  And we can possibly see later about the versions on the test libraries, but, at least they can now only cause a compilation failure when actually imported.

### DIFF
--- a/packages/core/lib/testing/SolidityTest.js
+++ b/packages/core/lib/testing/SolidityTest.js
@@ -110,26 +110,12 @@ const SolidityTest = {
     let solcVersion = config.compilers.solc.version;
     solcVersion = RangeUtils.resolveToRange(solcVersion);
 
-    const truffleLibraries = [
-      "truffle/Assert.sol",
-      "truffle/AssertAddress.sol",
-      "truffle/AssertAddressArray.sol",
-      "truffle/AssertBalance.sol",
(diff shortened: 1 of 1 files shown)


haltman-at:
OK, this turned out not to be as simple as I thought.  This PR is going to need some more work.

haltman-at:
I'm going to close this for now.  Will try this again at some other point.

PR #3754, opened 2021-01-31
Expose ID generation logic as first-class interface in @truffle/db

gnidan:
Apart from internals (see commits), this PR adds two interfaces:

- `generateId()` exported by the root package, which takes a collection name and an input for that collection, and returns the expected ID.
- GraphQL `<resource>Id` queries for each collection (i.e., `sourceId`, etc.), to accept a corresponding input as argument and return the expected ID for that input.

This is to facilitate the difficulty around using @truffle/db externally when you know everything about a resource but don't know what to call it.

One internal change probably worth noting here: the existing `generateId()` function used by tests (which took just one argument and didn't filter for specified ID fields) has been replaced with the new function that does this collection-aware filtering.


fainashalts (approved):
This is great, makes sense to rework this!

gnidan:
I thought so too :)

Thanks @fainashalts!

Issue #3755, opened 2021-01-31
Vyper compile fails with "SyntaxError: Unexpected end of JSON input"


acryptosx:
Truffle 5.1.57 is last version where vyper compiles, 5.1.58 to 5.1.64 fails with:

SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/compile-vyper/index.js:151:1
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/compile-vyper/index.js:79:1
    at ChildProcess.exithandler (child_process.js:299:7)
    at ChildProcess.emit (events.js:315:20)
    at maybeClose (internal/child_process.js:1048:16)
    at Socket.<anonymous> (internal/child_process.js:439:11)
    at Socket.emit (events.js:315:20)
    at Pipe.<anonymous> (net.js:673:12)

gnidan:
Uh oh! What version of Vyper? Any other project info you can share?

(@haltman-at, you'll want to see this!)

acryptosx:
vyper --version
0.2.7+commit.0b3f3b3

haltman-at:
Hm, I'm not encountering any problems when I try it... do you have a reproduction example I could check out?  Thank you!

(Actually it's quite curious that it's erroring at at `JSON.parse` in `index.js`, since that would mean that it's attempting to compile with `vyper` rather than `vyper-json`, even though you're using Vyper 0.2.7.  I guess you somehow only installed `vyper` and not `vyper-json`?  Interesting.  Well, it ought to work with or without `vyper-json`, so this is still a bug either way.)

Anyway like I said, I think I'm going to need a reproduction example here to find out just what's going wrong.  Would you be able to provide one?

haltman-at:
Oh, silly me -- I forgot to disable my own `vyper-json` before attempting to reproduce this.  I've reproduced it just fine now.  OK, I'll have to see what's causing this.  Thanks for the report!

haltman-at:
OK, I've put up #3757 to fix this!

acryptosx:
Is it better to install vyper-json? The docs are a bit sparse on using vyper with truffle.

haltman-at:
[...]

PR #3757, opened 2021-02-02
Bug fix: Fix several Vyper compilation bugs (specific to non-JSON Vyper and prerelease Vyper)

haltman-at:
Addresses #3755.

This PR fixes three bugs in Vyper compilation.  Two of these are specific to non-JSON compilation, and one is specific to Vyper prerelease versions.

1. In non-JSON compilation, the `source_map` format was included even if the version was too low (apparently it's only `vyper-json` that complains about such things); if the version was high enough, it would be included twice!  Now I've removed it from the starting array of formats, so if it doesn't get pushed on, it's not there, and if it does get pushed on, it's the only copy.  (It's this one that caused #3755; including `source_map` twice caused the source map to get read as the empty string, and then `JSON.parse` would fail on that.)
2. Also in non-JSON compilation, later code below didn't account for the possibility that the source map might not exist, and would crash on attempting to parse `undefined` as JSON.  Now we check for that.
3. Finally, our version checks against Vyper prerelease versions didn't actually work, as Vyper prerelease versions report the version as (e.g.) `0.1.0b17` rather than `0.1.0-beta.17`.  Semver doesn't recognize and can't work with the former format (so no, changing our comparison versions to that format doesn't help!), so instead I wrote a quick regex-replace to normalize the reported versions.  Now the version checks actually work.  Note that this bug only affected Vyper prerelease versions, as release versions would compare correctly here.

### DIFF
--- a/packages/compile-vyper/index.js
+++ b/packages/compile-vyper/index.js
@@ -37,7 +37,8 @@ function checkVyper() {
 
 // Execute vyper for single source file
 function execVyper(options, sourcePath, version, callback) {
-  const formats = ["abi", "bytecode", "bytecode_runtime", "source_map"];
+  const formats = ["abi", "bytecode", "bytecode_runtime"];
+  debug("version: %s", version);
   if (
     semver.satisfies(version, ">=0.1.0-beta.7", {
       loose: true,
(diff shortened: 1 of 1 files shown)


gnidan (approved):
Nice! Glad to get this sorted out!

PR #3759, opened 2021-02-02
Use `vyper --standard-json` if vyper-json not available

haltman-at:
This PR makes it so that if `vyper-json` is not installed, we can still compile Vyper using JSON if `vyper` 0.2.5 or later is installed, using `vyper --standard-json`.  This feature, to compile via JSON but with the `vyper` executable instead of a separate `vyper-json` executable, was only added in Vyper 0.2.5, so we can't do it any earlier than that.

Ideally there would be a better way to check for this than checking the version, but I don't see one that wouldn't involve invoking `vyper` a third time, and I want to avoid that.  So, I think this is fine.

So now `checkVyper`, when `json` is set to true, will also return `jsonCommand`, which will either be `"vyper-json"` or `"vyper --standard-json"`.  This will be passed to `compileJson`, which will use the specified command when invoking the compiler.  Pretty simple.

Anyway, this way people won't need `vyper-json` installed to get the benefits of JSON-based Vyper compilation (if they're using Vyper 0.2.5 or later, anyway).

### DIFF
--- a/packages/compile-vyper/index.js
+++ b/packages/compile-vyper/index.js
@@ -20,16 +20,35 @@ function checkVyper() {
   return new Promise((resolve, reject) => {
     exec("vyper-json --version", function (err, stdout, _stderr) {
       if (err) {
+        //vyper-json not available, check vyper
         exec("vyper --version", function (err, stdout, stderr) {
           if (err) {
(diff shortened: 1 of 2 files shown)


gnidan (approved):
Sweet!

PR #3761, opened 2021-02-03
Allow querying @truffle/db for a project's contract instances

gnidan:
This PR adds the `Project.contractInstance` and `Project.contractInstances` queries for easy deployed instance lookups.

Since contract instances _themselves_ are not named, these queries filter based on **contract name** and/or **network name**. (`Project.contractInstance` requires both, `Project.contractInstances` allows either/both/neither). These queries use these names to determine a "best fit" match for contract instances.

The process is a bit tricky, and different for contract names vs. for network names.

For filtering by network name:
- The name resolution part is easy: there's no need to look for historical name records for a given network. Assume that the currently assigned network resource for that name represents the actual blockchain we want, and that the contract instances we're looking for are deployed to some ancestor of this network resource.
- So, we just need to find all ancestors and filter contract instances by those IDs.

For filtering by contract name (or omitting this filter, thus filtering for "any known contract name"):
- The name resolution is the more important one here: search through each contract name history until we find that it has a corresponding contract instance (possibly filtered by network)
- Stop searching for each particular contract name once we have a matching contract instance for that name (possibly continuing for other names not found yet).

Putting these together, the process primarily searches through the history for a number of contract names simultaneously, keeping a sparse list of discovered contract instances (matching index with list of contract names) on a _per contract name_ basis. At each contract name revision, query for contract instances corresponding to that contract and match against the network filter. Return all contract instances found this way after we've exhausted _every contract name_: for each, either find a matching instance for that name or exhaust the known history of that name.

This PR does a couple other things that are arguably related:
- Remove `Network.fork` again. Guess I missed this is in #3733.
- Add a friendly overload to `workspace.find()` that lets you query for a list of IdObjects, preserving order between queried list and returned results. Since we do this in a lot of places.
- Fix an internal return type.

I've included a pretty barebones test, just to exercise the essential mechanism here: what happens if contracts are deployed to one network, then deployed to another network after some revisions?

### DIFF
--- a/packages/db/src/resources/projects/resolveContractInstances.ts
+++ b/packages/db/src/resources/projects/resolveContractInstances.ts
@@ -1,0 +1,214 @@
+import { logger } from "@truffle/db/logger";
+const debug = logger("db:resources:projects:resolveContractInstances");
+
+import gql from "graphql-tag";
+import { delegateToSchema } from "graphql-tools";
+import type * as graphql from "graphql";
+import type {
+  DataModel,
+  NamedCollectionName,
+  IdObject,
+  SavedInput,
+  Workspace
+} from "@truffle/db/resources/types";
+import { resolveNameRecords } from "./resolveNameRecords";
+
+/**
+ * For given contract and/or network names, find all contract instances
+ * matching either/both of those filters.
+ *
+ * Returns contract instances for the most current contract according to that
+ * contract's name record history - i.e., if a contract has been revised since
+ * being deployed to mainnet, this function will return the contract instance
+ * for that past revision.
+ */
+export async function resolveContractInstances(
+  project: IdObject<"projects">,
+  inputs: {
+    contract?: DataModel.ResourceNameInput;
+    network?: DataModel.ResourceNameInput;
+  },
+  context: {
+    workspace: Workspace;
+  },
+  info: graphql.GraphQLResolveInfo
+): Promise<SavedInput<"contractInstances">[]> {
+  const { workspace } = context;
+
+  const contractNameRecords = await resolveNameRecords(
+    project,
+    { ...inputs.contract, type: "Contract" },
+    { workspace }
+  );
+
+  const contractInstances: SavedInput<"contractInstances">[] = [];
+  debug("inputs %O", inputs);
+
+  for await (const { skip, contracts } of findResourcesHistories<"contracts">({
+    collectionName: "contracts",
+    nameRecords: contractNameRecords,
+    workspace
+  })) {
+    let stepContractInstances = await workspace.find("contractInstances", {
+      selector: {
+        "contract.id": { $in: contracts.map(({ id }) => id) }
+      }
+    });
+
+    if (inputs.network) {
+      const ancestors = await filterProjectNetworkAncestors({
+        project,
+        network: inputs.network,
+        candidates: stepContractInstances.map(({ network }) => network),
+        workspace,
+        info
+      });
+
+      const ancestorIds = new Set([...ancestors.map(({ id }) => id)]);
+      stepContractInstances = stepContractInstances.filter(({ network }) =>
+        ancestorIds.has(network.id)
+      );
+    }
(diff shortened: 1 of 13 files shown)


packages/db/src/resources/projects/resolveContractInstances.ts
```diff
@@ -0,0 +1,214 @@
...
+  do {
+    const skip = (...indexes: (number | undefined)[]) => {
+      for (const index of indexes) {
+        if (typeof index === "number") {
+          nameRecords[index] = undefined;
+        }
+      }
+    };
+
+    // @ts-ignore
```
fainashalts, line 126: We're racking up `@ts-ignore`'s, haha. Slightly worried about technical debt.

packages/db/src/resources/projects/resolveContractInstances.ts
```diff
@@ -0,0 +1,214 @@
...
+  project: IdObject<"projects">;
+  network: DataModel.ResourceNameInput;
+  candidates: IdObject<"networks">[];
+  workspace: Workspace;
+  info: graphql.GraphQLResolveInfo;
+}): Promise<IdObject<"networks">[]> {
+  const { project, network, workspace, info } = options;
+
+  // short-circuit early to avoid extra queries and to avoid index lookup guard
+  if (options.candidates.length === 0) {
```
fainashalts, line 160: Can we destructure `options.candidates` here? Not super important but it would be a little easier to read.

packages/db/src/resources/projects/resolveContractInstances.ts
```diff
@@ -0,0 +1,214 @@
...
+    return [];
+  }
+
+  const candidates = await workspace.find("networks", options.candidates);
+
+  // find earliest among candidates to specify minimumHeight to ancestors query
+  const earliestCandidate = candidates
+    .slice(1)
+    .reduce(
```
fainashalts, line 169: This is one of my favorite use cases for `.reduce` 🎉

packages/db/src/resources/projects/resolveContractInstances.ts
```diff
@@ -0,0 +1,214 @@
...
+          : network,
+      candidates[0]
+    );
+
+  // query ancestors for project network name
+  const { network: { ancestors = [] } = {} } = await delegateToSchema({
+    schema: info.schema,
+    operation: "query",
+    fieldName: "project",
+    returnType: info.schema.getType("Project") as graphql.GraphQLOutputType,
```
fainashalts, line 182: A little confused about this -- why do you need to cast this type?

fainashalts (approved):
LGTM! Nice! 🎉

packages/db/src/resources/projects/resolveContractInstances.ts
gnidan: Heh this one was just laziness in a place where there's low risk. I can comment to that effect

packages/db/src/resources/projects/resolveContractInstances.ts
gnidan: I wanted to use `candidates` later so I avoided destructuring it here. Can change!

packages/db/src/resources/projects/resolveContractInstances.ts
gnidan: Cause graphql's types aren't friendly, it seems

gnidan: (The reason is that `getType` doesn't know if it's returning an output type or an input type, but `returnType` demands only output types. But _we_ know that `"Project"` will be an output type)

packages/db/src/resources/projects/resolveContractInstances.ts
gnidan: Turned out to be easy enough to be able to remove this.

gnidan:
Thanks @fainashalts!

PR #3762, opened 2021-02-03
Use ethers when sending transactions for stacktracing, instead of full manual (fix stacktracing w/o instamining)

haltman-at:
This is mostly an internal improvement, but it's also a bug fix, since previously, if you ran `truffle test -t` and used a node that doesn't instamine like Ganache does, you'd get problems.  I think there are also some other cases it fixes that I previously hadn't tested properly before.

(Note: I didn't add any tests in this PR since I wanted to get it up quickly, but I did manually test all 8 cases with the code pushed here!  The 8 cases being: VM errors or not; instamining or not; and `eth_sendTransaction` or sending signed transactions).

Anyway.  This PR redoes `sendTransactionManual` so that it no longer actually does any manual provider calls; instead it uses `ethers`.  (Together with web3!  Yes, both of them together!)

The reason for this is that my manual method didn't work if there wasn't instamining, because I assumed I could get back a transaction hash from `eth_sendTransaction` (or `eth_sendRawTransaction`), and in realistic cases without instamining, you can't.  I don't want to write polling or subscribing logic myself, obviously.  So instead I used `ethers`!  Web3 already includes ethers, so this shouldn't make the `@truffle/contract` bundle bigger.  (Note: Not entirely certain of this -- I may need to consolidate versions here?)

Now you may ask, why did I do this, rather than just using web3?  Well, the reason is the same as back when I introduced this: We need to get the transaction hash even when the transaction fails, but web3 throws an error in that case, and, more to the point, when VM errors are on, it throws a *useless* error that we can't do much with.  Now, `ethers` will also throw when a transaction reverts (except in certain contexts it seems), but unlike `web3` it always throws useful errors that we can process, which is much better.

So the new logic for how to send the transaction, and get the hash and receipt, is as follows.  I'll break this down into cases.  Note that we have separate cases for whether the `from` address is in the web3 wallet, because web3 treats that case specially and signs the transaction itself rather than using `eth_sendTransaction`, and I wanted to replicate that behavior.

In all cases, we start by setting up `ethersProvider = new ethers.providers.Web3Provider(web3.currentProvider)`.  Also, since we're going to be using ethers rather than web3, we need to call the gas parameter `gasLimit` instead, so we do that translation in advance.  Anyway, cases!

1. Unsigned transaction, it succeeds.  In this case we send the transaction with `ethersProvider.getSigner().sendUncheckedTransaction(transaction)` and get the transaction hash.  We then use `ethersProvider.waitForTransaction(txHash)` to wait for it to be mined, and then use web3 to get the receipt.  Btw, you may wonder why I used `sendUncheckedTransaction` rather than `sendTransaction`; the reason is because the latter doesn't seem to work in this case.  IDK.  But this does.
2. Unsigned transaction, it reverts, no VM errors. This works exactly like case (1); `sendUncheckedTransaction` doesn't throw on revert, so there's no problem here.
3. Unsigned transaction, it reverts, VM errors are on.  In this case, the `sendUncheckedTransaction` will throw, but it'll throw a useful error we can get the transaction hash out of.  So, we get it from there, and then get the receipt with web3 as before.
4. Signed transaction (with a web3 account `account`), it succeeds.  For signed transactions we set up `ethersWallet = new ethers.Wallet(account.privateKey, ethersProvider)`.  We then send it and get the receipt with `(await ethersWallet.sendTransaction(transaction)).wait()`.  We then extract the transaction hash from the receipt; however, we also have to do some translation on the receipt, as it will have some `BigNumber`s (that's the ethers class of that name, not the other one) that web3 returns as numbers, so we have to translate those.
5. Signed transaction, it reverts, VM errors are on.  In this case sending the transaction will throw, but it throws in the same way as case (3), so we handle it the same way.
6. Signed transaction, it reverts, no VM errors.  In this case sending the transaction will throw, but it throws in a different way from case (4).  In this case the error includes both the transaction hash and the receipt, so we just get them both from there.

Note that if anything throws an error that isn't of either of the expected forms, we rethrow it so that we don't accidentally swallow unexpected errors.

So that's the logic!  The errors that get thrown at the end in the revert cases may have accidentally changed somewhat?  IDK.  I'm not sure that really matters so much.  Anyway yeah, this should be better than what was there before!

### DIFF
--- a/packages/contract/lib/execute.js
+++ b/packages/contract/lib/execute.js
@@ -8,6 +8,7 @@ const handlers = require("./handlers");
 const override = require("./override");
 const reformat = require("./reformat");
 const { formatters } = require("web3-core-helpers"); //used for reproducing web3's behavior
+const ethers = require("ethers");
 
 const execute = {
   // -----------------------------------  Helpers --------------------------------------------------
@@ -535,17 +536,14 @@ const execute = {
     return execute.sendTransactionManual(web3, params, promiEvent);
   },
 
+  //this is less manual now, it uses ethers, whew
+  //(it's still more manual than using web3)
   sendTransactionManual: async function (web3, params, promiEvent) {
-    //note: to head off any potential problems with Webpack (contract *has* to
-    //work on web!), I'm going to resort to manual promise creation rather than
-    //using util.promisify :-/
     debug("executing manually!");
-    const send = rpc =>
-      new Promise((accept, reject) =>
-        web3.currentProvider.send(rpc, (err, result) =>
-          err ? reject(err) : accept(result)
-        )
-      );
+    //set up ethers provider
+    const ethersProvider = new ethers.providers.Web3Provider(
+      web3.currentProvider
+    );
     //let's clone params
     let transaction = {};
     for (let key in params) {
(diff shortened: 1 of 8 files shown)


haltman-at:
OK, I merged in `develop` and set the `ethers` version to `^5.0.7`, because 5.0.7 is what web3 1.3.4 uses.  (I put the caret in case we upgrade web3 later but forget to correspondingly upgrade ethers.)  @gnidan, please let me know whether this was the correct thing to do?

Although I should note that with web3 1.3.4 it's no longer true that requiring ethers no longer increases the bundle size, as (unlike 1.2.9) web3 1.3.4 only seems to require one particular part of ethers, not the whole thing... so uh let me know to what extent this is a problem.  We could probably import things more piecemeal too, but we'd still be importing things web3 doesn't use.

haltman-at:
Welp, reverted that!  @gnidan, again, please let me know -- should I bother matching the version on ethers?  Or what?

haltman-at:
OK, after talking with @gnidan, instead of doing that, I upgraded the `ethers` version in interface-adapter!  Matching the one in web3 is difficult as 1.2.9 uses 5.0.0-beta.153.  But at least it uses the abi-coding parts?

haltman-at:
Btw, if people want, while I'm at it I could change `interface-adapter` to only import `@ethersproject/abi`, since that's all it actually uses, and then the import could presumably work typed...

gnidan (commented):
@haltman-at generally this looks good, but I think it's worth moving this logic to its own module / splitting into separate functions. Maybe `contract/lib/transaction/`. I think it'd make the flow more transparent and the effort won't go to waste.

haltman-at:
> @haltman-at generally this looks good, but I think it's worth moving this logic to its own module / splitting into separate functions. Maybe `contract/lib/transaction/`. I think it'd make the flow more transparent and the effort won't go to waste.

OK, did that!

gnidan (approved):
Looks good!

haltman-at:
Note: Holding off on merging this for now as per discussion on Slack with @gnidan due to the strange bundling errors.  Will have ot figure out what to do about this.

cds-amal (approved):
LGTM @haltman-at ! Good idea to refactor to `manual-send.js`.

haltman-at:
OK after downgrading to ethers v4 I had to change the code, but the good news is that the new code is simpler.  Also, I (finally!) added a test of stacktracing!  It's pretty rudimentary, but at least we finally have *something*!  Gonna re-request review.

gnidan (approved):
Looks good!

PR #3763, opened 2021-02-03
Update web3 to 1.3.4

haltman-at:
Finally!

Note that this PR updates all web3 packages to 1.3.4, but leaves `@types/web3` in place.  Ideally we eventually tear that out, but I think that can be done later on a package-by-package basis.


haltman-at:
Hm, looks like there's a real failure here, will have to go see what's causing that and fix that.

eggplantzzz (approved):
yeehaw!

PR #3769, opened 2021-02-04
Add Network.genesis query to @truffle/db

gnidan:
... as a minor convenience instead of Network.ancestors


fainashalts (approved):
Nice! When would you get the "no known genesis" error, as a user?

gnidan:
> When would you get the "no known genesis" error, as a user?

Should never happen @fainashalts, unless the user is using @truffle/db as a library and neglecting to add the genesis block. But the Network abstraction handles this so it really shouldn't happen ever.

gnidan:
Thanks for looking!

PR #3771, opened 2021-02-05
Run lerna-update-wizard, and various other dependency improvements

haltman-at:
As sugested in #3767, I have run lerna-update-wizard (so this PR resolves that issue).  But that's not all I did; I also did some other dependency cleanup while I was at it.  Specifically:

1. I noticed some `@types` packages were dependencies, when clearly they should have been devDependencies;
2. I noticed `compile-vyper` had `eslint` as a dependency, clearly unnecessary;
3. I noticed `compile-solidity` had `solc` as a dependency, which it wasn't using.

Other than that, yeah, updated a lot of versions.  Sometimes I updated to latest, sometimes I just consolidated on the latest one we had installed; it depended on the particular package and how confident I was that upgrading would be OK!

However one particular devDependency took some additional care, namely, `@types/node` to work.  In this case I *downgraded* it in some packages to consolidate on 12.12.21.  Why?  Because, um, for some reason I just couldn't get newer versions to work.  IDK.

Note this package doesn't upgrade ethers beyond what was already there, as in #3762.  I figure #3762 can handle that.

So, uh, hopefully I didn't screw anything up and this passes CI?


haltman-at:
Hm, seems my Typescript upgrade didn't go so well.  I guess I'll try it again with a different version... but then again, it is just a dev dependency, so if I can't unify it, no big deal really probably.

haltman-at:
...OK, seems that one wasn't the only problem.  I guess I'll drop a few more of these commits as well.

haltman-at:
OK, I reverted the giant commit that updated 23 different things, since, without the Typescript upgrade, it was causing compile errors.  I guess I'll have to redo that one with a little more discretion...

haltman-at:
OK, I'm giving up on the idea of consolidating `typedoc`, too.  I'm very confused about what's up with that one.  And I think I'm going to have to leave `mocha` alone as well.

haltman-at:
...OK, apparently changeCase is also a problem.  That's a surprise...

gnidan (approved):

Issue #2152, opened 2019-07-03
Truffle Console: Print inidicator when auto-resolving promise


haltman-at:
## Issue

If an expression entered at the Truffle console evaluates to a promise, the console will automatically resolve it and print the result rather than printing the promise itself.  While helpful by itself, this can be confusing on context.  For instance, if one, forgetting an `await`, writes:

`result = (await MyContract.deployed()).method()`

Then `result` will hold a promise and, and the console will show the result of this promise: an object with such fields as `tx`, `receipt`, and `logs`.  So it appears that `result` has these fields.  And yet, because `result` actually holds a promise, attempting to print `result.tx`, `result.receipt`, or `receipt.logs` will confusingly yield `undefined`.  You can see the fields right there, yet you can't print them individually!

In order to mitigate this problem, it might help if, when auto-resolving and printing a promise, some indicator was printed to show that this was happening and make it clear that the actual expression that was entered evaluates to a promise rather than what it appears to evaluate to.

For instance, maybe one could print `Entered expression evaluates to promise; auto-resolving:` on a separate line before printing the resolution.

Another alternative would be to directly display the evaluated result as something like `Promise resolving to ...`, but for various reasons I like this solution less.

In any case, any indicator of this sort would make it clearer when you accidentally forgot an `await`.

## Environment

* Truffle version (`truffle version`): 5.0.24

gnidan:
(Requirements being coming up with the specific wording / output modification to make this more clear)

PR #3775, opened 2021-02-07
Print indicator when auto-resolving promise

Xuefeng-Zhu:
When auto-resolving and printing a promise, some indicator was printed to show that this was happening and make it clear that the actual expression that was entered evaluates to a promise rather than what it appears to evaluate to.

Fix https://github.com/trufflesuite/truffle/issues/2152

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -267,20 +267,31 @@ class Console extends EventEmitter {
       return callback(error);
     }
 
-    // Ensure our script returns a promise whether we're using an
-    // async function or not. If our script is an async function,
-    // this will ensure the console waits until that await is finished.
-    Promise.resolve(runScript(script))
-      .then(value => {
-        // If there's an assignment to run, run that.
-        if (assignment) return runScript(vm.createScript(assignment));
-        return value;
-      })
-      .then(value => {
-        // All good? Return the value (e.g., eval'd script or assignment)
-        callback(null, value);
-      })
-      .catch(callback);
+    const runScriptResult = runScript(script);
+
+    if (runScriptResult instanceof Promise) {
+      // Ensure our script returns a promise whether we're using an
+      // async function or not. If our script is an async function,
+      // this will ensure the console waits until that await is finished.
+      Promise.resolve(runScriptResult)
+        .then(value => {
+          // If there's an assignment to run, run that.
+          if (assignment) return runScript(vm.createScript(assignment));
+          return value;
+        })
+        .then(value => {
+          // All good? Return the value (e.g., eval'd script or assignment)
+          runScript(
+            vm.createScript(
+              'console.log("Entered expression evaluates to promise; auto-resolving:")'
+            )
+          );
+          callback(null, value);
+        })
+        .catch(callback);
+    } else {
+      callback(null, runScriptResult);
+    }
   }
 }
 


eggplantzzz:
Thanks for submitting the PR @Xuefeng-Zhu! I'll get this reviewed.

packages/core/lib/console.js
```diff
@@ -267,20 +267,31 @@ class Console extends EventEmitter {
...
-        return value;
-      })
-      .then(value => {
-        // All good? Return the value (e.g., eval'd script or assignment)
-        callback(null, value);
-      })
-      .catch(callback);
+    const runScriptResult = runScript(script);
+
+    if (runScriptResult instanceof Promise) {
```
eggplantzzz, line 272: So adding this `if` statement actually breaks some functionality here. Can you remove it?

You can confirm this by running `truffle develop` in a project directory, running migrations, and then doing something like 
```
const instance = await MyContract.deployed();
```

I think the script is run so quickly that the code takes the `else` and not the `if`. This causes `instance` to not get an assignment.

packages/core/lib/console.js
```diff
@@ -267,20 +267,31 @@ class Console extends EventEmitter {
...
+        .then(value => {
+          // If there's an assignment to run, run that.
+          if (assignment) return runScript(vm.createScript(assignment));
+          return value;
+        })
+        .then(value => {
+          // All good? Return the value (e.g., eval'd script or assignment)
+          runScript(
+            vm.createScript(
+              'console.log("Entered expression evaluates to promise; auto-resolving:")'
```
eggplantzzz, line 286: Can we change the text a bit here? How about "The expression evaluates to a Promise; ..."?

packages/core/lib/console.js
eggplantzzz: Oh maybe it is not that it runs so quickly but rather Truffle supports `await` in the console and it finishes and resolves before checking the conditional. We'll need to implement this differently.

Xuefeng-Zhu: yeah. It seems to be caused by await special handling. I will try to fix it

packages/core/lib/console.js
```diff
@@ -274,22 +274,21 @@ class Console extends EventEmitter {
...
-          // If there's an assignment to run, run that.
-          if (assignment) return runScript(vm.createScript(assignment));
-          return value;
-        })
         .then(value => {
           // All good? Return the value (e.g., eval'd script or assignment)
           runScript(
             vm.createScript(
-              'console.log("Entered expression evaluates to promise; auto-resolving:")'
+              'console.log("The expression evaluates to a Promise; ...")'
```
eggplantzzz, line 281: Sorry, I didn't literally mean for the "..." to make it in there.

The text I was trying to suggest was "The expression evaluates to a Promise; auto-resolving:"

Sorry for not being more clear.

Xuefeng-Zhu: updated. thanks

eggplantzzz:
Thanks @Xuefeng-Zhu, any progress on figuring out a solution to the issue related to `await`?

Xuefeng-Zhu:
@eggplantzzz Need to change the `const` to `let`. sorry, I will get it fixed today.  Thanks

Xuefeng-Zhu:
@eggplantzzz I have fixed and tested the change. It should work properly now.

eggplantzzz:
Thanks again @Xuefeng-Zhu, looks like this PR still has some broken behavior.

Currently when you do
```
contractInstance.myMethod(...);
```
or
```
await contractInstance.myMethod(...);
```
where those are both calls to send a transaction via a contract method, you get the receipt printed. With this PR it breaks the one where we `await`. In that case, we get a `Promise { pending }` printed. We can't really break existing behavior.

eggplantzzz:
@Xuefeng-Zhu any progress on maintaining the same behavior in the console?

Xuefeng-Zhu:
@eggplantzzz Sorry, I was quite busy these days. Will loop back to this issue when I get time

eggplantzzz:
No problem for this one as well, just keep us posted!

eggplantzzz:
Hey @Xuefeng-Zhu, do you have any progress to report or do you plan on continuing these PRs? Let us know what your status is.

PR #3776, opened 2021-02-07
Implement @truffle/preserve

rkalis:
This PR replaces #3709 and extracts only the @truffle/preserve changes so it can be merged into develop without completing the other packages.

- Complete refactor the control/processes/controllers module
  - Remove some layers of generics
  - Move all event types to a single file
  - Define interfaces for all different Controllers
  - Rename the "Unkowns" controller to ValueResolutionController
  - Add update() function to ValueResolution for intermediate values
  - Remove succeed/fail from the initial "Controls" object
  - Change `public state` to `protected _state` + a `get()`
- Move loaders out of the targets folder into its own module
- Rename label(s) to result(s) if they refer to the result of a plugin
- Rename label(s) to resolution(s) if they refer to resolution of a value using the ValueResolutionController
- Move around the `import * as X; export { X }` statements so they are always inside the index.ts files
- Rename thunk to stringify
- Move test folder out of lib
- Some other smaller changes

### DIFF
--- a/packages/preserve/.gitignore
+++ b/packages/preserve/.gitignore
@@ -1,0 +1,4 @@
+node_modules/*
+yarn.lock
+package-lock.json
+dist
--- a/packages/preserve/.gitignore
+++ b/packages/preserve/.gitignore
@@ -2,3 +2,4 @@ node_modules/*
 yarn.lock
 package-lock.json
 dist
+coverage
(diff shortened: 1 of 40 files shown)


gnidan (approved):
Looks good, love the changes you made to my original implementation!

PR #3781, opened 2021-02-08
Enable `strictNullChecks` in @truffle/db

gnidan:
This PR turns on TS's `strictNullChecks` option to ensure we avoid the landmine of "cannot destructure property of undefined". Good riddance, I say!

Mostly this means checking for undefined / null in various places where we hadn't done so previously, and that's the bulk of this PR. Besides that, though, there are a few other substantive changes:

- Simplify build process:
  - Use `ts-node --transpile-only` rather than having `yarn prepare` perform a dedicated separate compilation just for generating the DataModel.
  - Remove `yarn build:stubbed`, since ts-node handles compilation internally.
  - Remove `types/stubs.d.ts`! We don't need it! Replace it with new `bin/register`, just so that ts-node knows that the `_DataModel` identifier exists (even though `--transpile-only` means it won't use it for anything)

- Explicitly add nullable values to the domain and range of a number of functions/methods:
  - Align on the strategy of accepting undefined as possible input when getting or mutating resources. This includes accepting nully values for IDs/inputs in queries/mutations, as well as with the process interface (e.g. `resources.load("networks", [undefined])`), and all the way down to the `Workspace` interface.
  - Related to this, allow `toIdObject()` and `generateId()` to accept undefined values; this is to make it easier to just pass return values around (if something returns `Resource | undefined`, this PR now lets that return value be passed directly to `toIdObject`).
  - Allow undefined values for the Network abstraction's `includeBlocks` and `includeTransactions` methods, since the Project abstraction may naturally pass undefineds there, and it's easier to add it to the Network abstraction domain than to filter inside Project.
  - Cleanup the Project abstraction's internal batch logic and make sure each process module acknowledges possible undefineds for input/output.

- Improve type robustness in a number of places:
  - Besides accepting nully values for `toIdObject()`, also make that function polymorphic to `Resource | SavedInput`, since both have IDs and can adequately serve as input for generating an IdObject. What's more, go further and expand this polymorphism to accept any subset of either of those types that picks at least the `"id"` field (`Pick<Resource, "id">` is fine for generating IdObjects!)
  - Improve the Query / Mutation types defined by the process subsystem. This wasn't being used much, probably because it didn't work right, but this PR fixes that so you can do `Query<"networks">` or `Query<"network">` or `Mutation<"networksAdd">` and get the type for the returned `data` for that corresponding request. (_Aside: in the future, it'd be nice to type `db.execute()` as returning `Query | Mutation`, possibly even generically!_)
  - Identify incorrect nullability requirements in typeDefs.
  - Remove unnecessary ts-ignores in the code!

- Staple a layer of duct tape over the tests:
  - Apply `@ts-ignore` liberally in places where jest expectations clearly provide the guarantees for us (to save the trouble of a bunch of type gymnastics for no gain)
  - Try not to rely on that strategy though: where appropriate, cast `db.execute()` results as `{ data: Query<...> }` or `{ data: Mutation<...> }`, so that TS has a vague idea about what's coming back.
 
That's roughly the extent of it, but see commits for more detail.


packages/db/src/project/loadMigrate/contracts.ts
```diff
@@ -19,29 +19,41 @@ export const process = Batch.configure<{
...
-  result: {
-    callBytecode?: {
-      linkReferences: { name: string }[];
-    };
-    createBytecode?: {
-      linkReferences: { name: string }[];
-    };
-  };
+  result:
+    | {
```
fainashalts, line 31: What is this `|` about? Looks like a bitwise or but that can't be it, and wrong syntax for pipeline operator. Is it a TypeScript thing?

packages/db/src/resources/bytecodes.ts
```diff
@@ -20,7 +20,7 @@ export const bytecodes: Definition<"bytecodes"> = {
   typeDefs: gql`
     type Bytecode implements Resource {
       bytes: Bytes!
-      linkReferences: [LinkReference]
+      linkReferences: [LinkReference!]
```
fainashalts, line 23: This should be allowed to be empty, shouldn't it? Empty array if the bytecode has no link references? Maybe I missed something...

packages/db/src/resources/bytecodes.ts
```diff
@@ -42,7 +42,7 @@ export const bytecodes: Definition<"bytecodes"> = {
 
     input BytecodeInput {
       bytes: Bytes!
-      linkReferences: [LinkReferenceInput]
+      linkReferences: [LinkReferenceInput!]
```
fainashalts, line 45: Ditto here, thought this could be empty.

fainashalts (approved):
Overall LGTM! I had some questions/comments that I wouldn't mind answers to before you merge but wanted to give the approve in case you're working when I'm sleeping!

packages/db/src/meta/process/resources.ts
```diff
@@ -83,7 +83,7 @@ export const resourceProcessorsForDefinitions = <C extends Collections>(
         type: "graphql",
         request: gql`
           mutation Load${Resources}(
-            $inputs: [${Resource}Input!]!
+            $inputs: [${Resource}Input]!
```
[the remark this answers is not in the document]
gnidan: I found out we were wrong about this! The inner `!` doesn't imply "at least one item", it just means "every item will be non-null".

Since this PR broadly says "alright, lots of things inside arrays can be null", result is: remove this constraint.

packages/db/src/meta/process/types.ts
```diff
@@ -122,19 +122,37 @@ export type QueryName<
...
-      C,
-      N
-    >["resource"]
-      ? Resource<C, N> | null
-      : Q extends CollectionProperty<"names", C, N>["resources"]
-      ? (Resource<C, N> | null)[] | null
-      : never;
-  };
-}[CollectionName<C>];
+type UnionToIntersection<T> = (T extends any ? (x: T) => any : never) extends (
```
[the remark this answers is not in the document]
gnidan: Oh you're right I should reference where I borrowed it from!

packages/db/src/process.ts
```diff
@@ -39,6 +39,35 @@ export type Processor<
  */
 export type RequestType = Meta.Process.RequestType<Collections>;
 
+/**
+ * @hidden
```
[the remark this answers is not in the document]
gnidan: Tells TypeDoc not to render it. I might want to add this in, but I was just following the convention for a couple other hidden exports in this module.

packages/db/src/project/loadCompile/compilations.ts
```diff
@@ -168,19 +168,41 @@ function toSourceMapInputs(options: {
...
-          offsets: reference[1].map(({ start }) => start)
-        };
-      });
+      return Object.entries(contract.immutableReferences)
+        .map(([astNode, slices]):
+          | DataModel.ImmutableReferenceInput
+          | undefined => {
+          // HACK start/length might actually be required, but contract-schema
+          // is wrong?
+          const definedSlices = slices.filter(
```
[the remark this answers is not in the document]
gnidan: In `contract-schema`, there's no `"required": ["start", "length"]`, so that slice definition gets converted to the TS type `{ start?: number; length?: number }`.

I'm assuming this is wrong, but I'm handling it here just to avoid the distraction (maybe this is worth opening an issue to track? let me know!)

packages/db/src/project/loadMigrate/contracts.ts
gnidan: Yeah - called a "type union". The normal syntax is `type X = A | B`, which means every X is either an A or a B (more akin to a logical-or vs. bitwise-or).

TS allows this nice syntax where you can put `| ` at the beginning so you can make a bunch of options line up all nice-like:
```typescript
type X =
  | A
  | B
  | C
  | D
  | E;
```

packages/db/src/resources/bytecodes.ts
gnidan: See above for how we were wrong about what this means; it doesn't mean at least one item, it means the array can't have any null values in it. Which is what we want, I'd say - we won't get missing link references ever (like we will for `Compilation.sources`, e.g.)

packages/db/src/resources/bytecodes.ts
gnidan: Ditto above :)

packages/db/src/meta/process/types.ts
gnidan, line 125:
```suggestion
// borrowed from https://fettblog.eu/typescript-union-to-intersection
type UnionToIntersection<T> = (T extends any ? (x: T) => any : never) extends (
```

packages/db/src/network/fetch/knownLatest.ts
```diff
@@ -51,6 +51,10 @@ export function* process<
     }
   `
   );
+  if (!resource) {
+    return;
```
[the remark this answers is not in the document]
gnidan: Maybe? I handle it elsewhere. I think it's safe enough.

Issue #3756, opened 2021-01-31
Bug: Truffle async function test failure should cause a shell exit with error code


zhouhanseng:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Hello, everyone. I find that Truffle async function without done() test failure does not cause a shell exit with error code which leads to wrong CI report

## Steps to Reproduce

```ts
// exit with error code.
it('should return -1 when the value is not present', (done) => {
  assert.equal([1, 2, 3].indexOf(2), -1);
  done()
});

// does not exit with error code.
it('should return -1 when the value is not present', async () => {
  assert.equal([1, 2, 3].indexOf(2), -1);
});
```

Please provide the shortest amount of steps to reproduce your issue.

## Expected Behavior

`it('', async () => {//...})` failure should exit with error code.

![image](https://user-images.githubusercontent.com/63733714/106385502-91e42d00-640b-11eb-9148-86a65436c6ef.png)

## Actual Results

`it('', async () => {//...})` failure does not exit with error code.

![image](https://user-images.githubusercontent.com/63733714/106385480-83961100-640b-11eb-9bf7-6873a2b53400.png)

## Environment

* Operating System: macOS 10.15.7 and ubuntu-latest
* Ethereum client: ganache
* Truffle version (`truffle version`): ^5.1.64
* node version (`node --version`): 14.13.0
* npm version (`npm --version`): 7.1.0

davidmurdoch:
Thanks @zhouhanseng for opening this issue! I've moved it to our backlog to be worked ASAP.

haltman-at:
Hi, we've just released Truffle 5.1.66 which fixes the problem of exit statuses not being propagated.  However it seems like your problem may be more specific.  Does Truffle 5.1.66 fix your problem, or is there still more to do here?  Thank you!

MicaiahReid:
We haven't heard anything back on this, so we're calling it fixed! Please open another issue if you're finding that this is still happening.

PR #3782, opened 2021-02-08
Bug fix: Propagate exit statuses properly

haltman-at:
This might address #3756, I'm not sure.  We'll have to ask them.

This PR fixes a bug where exit status codes weren't being set correctly; this is particularly relevant to Truffle Test, where the exit code should be equal to the number of failed tests.  Truffle Test was returning this number correctly, but `cli.js` was ignoring it and returning zero regardless!  Now this is fixed.


gnidan (approved):
Wow, easy!

PR #3783, opened 2021-02-09
Internal improvement: Add compatibleNativize function (and some missing codec/decoder functions)

haltman-at:
This PR adds the `compatibleNativize` function to Truffle Codec, intended for use as part of the Truffle Contract rewrite.  It, uh, also adds some other stuff that in the process turned out to be missing.

The purpose of `compatibleNativize` is to act like `nativize`, except that instead of yielding a JS equivalent for use in the debugger, it yields the sort of thing Truffle Contract (by means of ethers) currently outputs.  There are some slight differences, but, as I'll describe below, I don't think they're a problem.

There are three `compatibleNativize` functions: `compatibleNativize`, which takes an individual value; `compatibleNativizeReturn`, which takes a `ReturndataDecoding` (assumed to be of type `"return"`, it returns `undefined` otherwise); and `compatibleNativizeEventArgs`, which takes a `LogDecoding`.  (Note that this last one returns *only* the `args` part of the current-style event decoding, not anything else.)

Each of these functions takes three arguments.  First, the thing to be nativized.  Second, `userDefinedTypes`, since it may have to handle user defined types; I'll discuss below how this could possibly be eliminated.  Thirdly and optionally, `numberFormatter`.  Now, ethers decodes numbers as its own `BigNumber` class (not to be confused with the more common class of that name), but we will by default will decode numbers as `BigInt`s, and you can pass a `numberFormatter` that takes `BigInt`s and outputs someting else if you want something else.  Unfortunately I couldn't get the typing on this to work like it should, but I'm not sure how much it matters seeing as how the output of all this stuff is all `any` anyway.

Let's start with `compatibleNativize`.  It's a lot like `nativize`.  However, it begins by calling `abifyValue`, to make sure that we're doing things the ABI way.  It then has a `switch` like `nativize`, but, because we're dealing with ABI types only, and no circularities, it's much simpler.  The cases of `"fixed"`, `"ufixed"`, and `"function"` all simply return `undefined`.  Currently in Truffle Contract these throw, but we do not want this throwing!  Since we don't want, like, a decoding error screwing up your ability to get a transaction receipt or whatever.  So they simply return `undefined`.

The one case that has some seriously different behavior from `nativize` is the case of `"tuple"`.  In this case, rather than returning an array, or an object keyed with the component names, we return an array which is *also* keyed by the component names (when they exist).  That's because this is what `ethers` does, and so what Truffle Contract does currently.

There's one other quirk here: If we're nativizing a value of reference type that couldn't be decoded because it was indexed, we nativize it to the undecoded hash.  I'll discuss this case more below.

Then we have `compatibleNativizeReturn`.  This checks how many return values there are; if there's only one, it just calls `compatibleNativize` and that's it; it doesn't wrap the result in anything else.  If the count of return values is different from 1, though, it returns the results wrapped in an object, keyed both by index and by name (when names exist).

Finaly we have `compatibleNativizeEventArgs`.  This one *always* wraps the results in an object, whether there's 0, 1, or more; and it also includes a `__length__` field to indicate how many of them there are.

And that's basically it!  Although there are some slight differences from how ethers does things, which I'll discuss below.  I don't think they'll be problems, though.

...except, that's not quite it, because while writing this I realized something was missing.  It doesn't affect the rest of the PR, but -- the `abifyReturndata` function never got properly exported from `Codec` like the other abification functions.  Moreover, it didn't get corresponding functions in `@truffle/decoder`.  So, uh, I added those.

Which brings up the point -- having to pass in `userDefinedTypes` is kind of inconvenient, isn't it?  Should I add `decoder` versions of the `compatibleNativize` functions, that wouldn't need that as an argument, like how the `decoder` versions of the abify functions don't need it?  Not sure there's much of a point, but it's something I could do.

OK, so what are the differences from current behavior, and why don't I think they're a problem?

1. Numbers in tuples actually get decoded properly, instead of decoded as strings -- I'd call that a bug fix!
2. Indexed reference types -- currently, sometimes events with these are decoded by leaving the undecoded hash in place undecoded, but sometimes they're not decoded at all.  I decided it was simpler to just always do the former more informative thing.  Like, this shouldn't cause problems, right?
3. To avoid importing ethers in `codec`, we don't use their `Result` class, instead we just use objects.  I think this is fine -- in Ethers v5 I'm pretty sure `Result` isn't even a class anymore, just an interface.
4. Things that aren't encoded right -- `ethers` will handle these fine, whereas we'll fail to decode them.  This shouldn't really happen as long as we're only decoding things output by Solidity (return values and event arguments), so I think it's fine, and there's not really an way we can change this without fundamentally changing how our decoding works.
5. What happens if you have an event with `__length__` as one of the argument names?  Currently, the way this acts is... weird and not very useful.  I replaced it with having it still be the argument count, overwriting any argument by that name, because that seems more expected.  I seriously doubt that anyone was relying on the old behavior.

Also: I put these functions in a new `export.js`, and also reexported `nativize` and `ResultInspector` from there for consistency.  So yeah.

Also there's a test added.

### DIFF
--- a/packages/codec/lib/export.ts
+++ b/packages/codec/lib/export.ts
@@ -1,0 +1,182 @@
+import debugModule from "debug";
+const debug = debugModule("codec:export");
+
+import * as Abify from "./abify";
+import * as Format from "@truffle/codec/format";
+import {
+  LogDecoding,
+  ReturndataDecoding
+} from "@truffle/codec/types";
+
+import { ResultInspector, nativize } from "@truffle/codec/format/utils/inspect";
+export { ResultInspector, nativize };
+
+type NumberFormatter = (n: BigInt) => any //not parameterized since we output any anyway
+
+/**
+ * This function is similar to [[Format.Utils.Inspect.nativize|nativize]], but
+ * is intended to match the way that Truffle Contract currently returns values
+ * (based on the Ethers decoder).  As such, it only handles ABI types, and in
+ * addition does not handle the types fixed, ufixed, or function.  Note that in
+ * these cases it returns `undefined` rather than throwing, as we want this
+ * function to be used in contexts where it had better not throw.  It also does
+ * not handle circularities, for similar reasons.
+ *
+ * To handle numeric types, this function takes an optional second argument,
+ * numberFormatter, that tells it how to handle numbers; this function should
+ * take a BigInt as input.  By default, this function will be the identity,
+ * and so numbers will be represented as BigInts.
+ *
+ * Note that this function begins by calling abify, so out-of-range enums (that
+ * aren't so out-of-range as to be padding errors) will not return `undefined`.
+ * Out-of-range booleans similarly will return true rather than `undefined`.
+ * However, other range errors may return `undefined`; this may technically be a
+ * slight incompatibility with existing behavior, but should not be relevant
+ * except in quite unusual cases.
+ *
+ * In order to match the behavior for tuples, tuples will be transformed into
+ * arrays, but named entries will additionally be keyed by name.  Moreover,
+ * indexed variables of reference type will be nativized to an undecoded hex
+ * string.
+ */
+export function compatibleNativize(
+  result: Format.Values.Result,
+  numberFormatter: NumberFormatter = x => x
+): any {
+  return compatibleNativizeAbified(
+    Abify.abifyResult(result),
+    numberFormatter
+  );
+}
+
+//HACK; this was going to be parameterized
+//but TypeScript didn't like that, so, whatever
+interface MixedArray extends Array<any> {
+  [key: string]: any
+}
+
+function compatibleNativizeAbified(
+  result: Format.Values.Result,
+  numberFormatter: NumberFormatter
+): any {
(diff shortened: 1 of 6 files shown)


gnidan:
We should probably rename `nativize` to `unsafeNativize` and just call `compatibleNativize` `nativize`

gnidan:
Also, to avoid losing explicitness, how about a `nativize(..., { format: "ethers" })` as an optional+default?

haltman-at:
> Which brings up the point -- having to pass in userDefinedTypes is kind of inconvenient, isn't it? Should I add decoder versions of the compatibleNativize functions, that wouldn't need that as an argument, like how the decoder versions of the abify functions don't need it? Not sure there's much of a point, but it's something I could do.

Seems like I should indeed do this!

haltman-at:
Actually, I just realized -- while currently `compatibleNativize` requires extra info, this is actually inessential; I could just rewrite it to not need that extra info!  That's probably a better approach.

gnidan (approved):

PR #3784, opened 2021-02-09
Allow users to enable @truffle/db as a system of record for their Truffle projects

gnidan:
With this PR, adding `db: { enabled: true }` will turn on @truffle/db for use with `truffle compile`, `truffle migrate`, etc.

This only adds data to @truffle/db right now; nothing inside Truffle is currently configured to read from @truffle/db.


packages/core/lib/commands/db/index.js
```diff
@@ -0,0 +1,46 @@
+const OS = require("os");
+const serveCommand = require("./commands/serve");
+
+const usage =
```
eggplantzzz, line 4: The style of these; does it really match with the rest of the Truffle help messages? Maybe we should indent "Available sub-commands" and "serve \tStart..." a bit more...or we could list the subcommands like we list the options? So they are indented far in, what do you think?

fainashalts (approved):
Drumroll please... LGTM!

gnidan:
It's in :)

PR #3785, opened 2021-02-09
Convert GeneratedSource to ProcessedSource in @truffle/db

gnidan:
This PR re-uses the ProcessedSource entity for contract generated sources, removing the GeneratedSource types in the process.

In addition, this ensures that the input `id` property from the compile-common type is used for the index of a sparse array, as opposed to storing these sources in a tightly-packed array whose items encode their index. This also replaces the `any` in those types with actual interfaces (cc @haltman-at).

To get this to work, since each ProcessedSource references a Source resource, this PR reworks the sources-loading inside the Project abstraction to include information from generated sources (on top of the `sources` field already used). To do this, I've gotten rid of the hokey `Batch.Sources` abstraction and just used `Meta.Batch` directly.

Also, while I was it it, I made some moves to simplify the Meta.Batch abstraction: removing a bunch of internal generics. This was creating noise without benefit, since we've had to `@ts-ignore` everything anyway. Hopefully this is a step in the right direction there.

### DIFF
--- a/packages/db/src/project/loadCompile/contracts.ts
+++ b/packages/db/src/project/loadCompile/contracts.ts
@@ -40,7 +40,7 @@ export const process = Batch.Contracts.configure<{
     };
   };
   resources: {
-    contract: IdObject<"contracts">;
+    contract: IdObject<"contracts"> | undefined;
   };
   entry: Input<"contracts">;
   result: IdObject<"contracts"> | undefined;
@@ -94,7 +94,7 @@ export const process = Batch.Contracts.configure<{
     return yield* resources.load("contracts", entries);
   },
 
-  convert<_I, _O>({ result, input: contract }) {
+  convert({ result, input: contract }) {
     return {
       ...contract,
       db: {
--- a/packages/db/src/project/loadCompile/contracts.ts
+++ b/packages/db/src/project/loadCompile/contracts.ts
@@ -5,7 +5,7 @@
 import { logger } from "@truffle/db/logger";
 const debug = logger("db:project:loadCompile:contracts");
 
-import type { Input, IdObject } from "@truffle/db/resources";
+import type { DataModel, Input, IdObject } from "@truffle/db/resources";
 import { resources } from "@truffle/db/process";
 import * as Batch from "./batch";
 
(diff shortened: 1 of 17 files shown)


packages/db/src/project/loadCompile/sources.ts
```diff
@@ -5,43 +5,194 @@
...
+      return [
+        ...compilationsBefore,
+        {
+          ...compilation,
+          contracts: [
+            ...contractsBefore,
+            {
+              ...contract,
+              [property]: [
+                ...generatedSourcesBefore,
```
fainashalts, line 202: this merge is getting a little deep! seems a reasonable way to do this though.

packages/db/src/project/loadMigrate/batch.ts
```diff
@@ -116,47 +116,40 @@ export const configure = <C extends Config>(
...
       },
       breadcrumb
     }) {
       const { artifactIndex } = breadcrumb;
-      return artifacts[artifactIndex].networks[networkId];
+      return (artifacts[artifactIndex].networks || {})[networkId];
     },
 
-    initialize<I, O>({ inputs: { artifacts, network } }) {
+    initialize({ inputs: { artifacts, network } }) {
```
fainashalts, line 130: Oh wow I like this a lot better. Much cleaner and easier to read for me! ⭐

fainashalts (approved):
LGTM!

packages/db/src/project/loadCompile/contracts.ts
```diff
@@ -20,6 +20,20 @@ export const process = Batch.Contracts.configure<{
     contractName: string;
     abi: any;
     sourcePath: string;
+    generatedSources?: {
+      id: number;
```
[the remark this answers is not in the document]
gnidan: Yeah but this is the type from the WorkflowCompileResult input

packages/db/src/project/loadCompile/sources.ts
gnidan: Yeah :/

gnidan:
Thanks @fainashalts!

PR #3786, opened 2021-02-09
Fix possible undefined for Contract queries in @truffle/db

gnidan:
- Fix for Contract.compilation, since it's optional
- Fix for Contract.processedSource, since compilation is optional


fainashalts (approved):
LGTM!

gnidan:
Thanks!

PR #3790, opened 2021-02-10
Bug fix: Have txlog handle constructor code that should be unmapped but isn't

haltman-at:
This PR makes it so that txlog will, when in a constructor context, treat code mapped to a contract definition node as if it were unmapped code.  This is due to [this Solidity bug](https://github.com/ethereum/solidity/issues/10935).

I'm a little worried this PR will screw things up for optimized code.  We'll see, I guess? :-/


cds-amal (approved):
Looks good!

gnidan (approved):

PR #3792, opened 2021-02-10
Implement @truffle/preserve-fs

rkalis:
This PR replaces #3722 and extracts only the @truffle/preserve-fs changes so it can be merged into develop without completing the other packages.

- Extract common types/interfaces into separate types.ts file
- Slightly reorder fs.ts to accommodate reading the file top-down
- Slight changes due to refactoring of @truffle/preserve
- Split up test file into fs.fixture.ts and fs.test.ts

### DIFF
--- a/packages/preserve-fs/.gitignore
+++ b/packages/preserve-fs/.gitignore
@@ -1,0 +1,4 @@
+node_modules/*
+yarn.lock
+package-lock.json
+dist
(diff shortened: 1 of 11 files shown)


gnidan (approved):
Looks good! Guess this didn't need to change much. I like your separating the fixtures.

PR #3793, opened 2021-02-10
Implement @truffle/preserve-to-ipfs

rkalis:
This PR replaces #3760 and extracts only the @truffle/preserve-to-ipfs changes so it can be merged into develop without completing the other packages.

- Update to the latest IPFS libraries
- Update other dependencies
- Extract IPFS interfaces from updated TS IPFS library where possible (not possible everywhere unfortunately)
- Update code to use refactored @truffle/preserves
- Separate fixture data from tests
- Test that the expected events are emitted


rkalis:
So regarding the CI failure, it looks like the IPFS libraries don't support Node 10 (https://github.com/ipfs/js-ipfs/issues/3558). Seems like we woouldn't want to update our minimum Node requirement to v12 because of that, but perhaps it could be possible to require Node 12 in order to use truffle preserve?

rkalis:
So the CI is failing because of the Node v10 issue. Would you say the way to go here is to disable preserve-to-ipfs tests for v10 (e.g. by checking `process.version`) @gnidan ?

gnidan (approved):
This looks good @rkalis!

0racl3z (commented):
2thumbs!^^

0racl3z (commented):
Grateful!

0racl3z (commented):
Thanks!

0racl3z (commented):
Thanks a ton!

PR #3794, opened 2021-02-10
Upgrade to webpack 5: significantly speed up runtime and reduce bundle size

tcoulter:
This PR upgrades `webpack` for the `truffle` package to webpack 5, as well as associated packages and plugins (`webpack-cli`, `copy-webpack-plugin`, etc.). It also includes configuration changes to shave off 62% of the overhead required to run the bundled `truffle` command, as well as includes performance optimizations to make some commands faster than the unbundled version! In my quick test of `truffle compile`, for example, the bundle was ~25% faster than the non-bundled CLI. As a side effect of all this, the total size of the build directory as been reduced to 46M from 211M. Big win. 

### Split Chunks + smaller source maps

The speedup and bundle size optimizations come from two major configuration changes to the `webpack.config.js` file: 

- `devtool: "nosources-source-map"`: At its core, this removes the source code from the source map files. (Did you know: In the `source-map` option, the full code gets copied? I didn't.). This of course reduces bundle size because less data is copied. It also reduces runtime because there's less data node will have to preprocess before running the bundle. 
- `splitChunks: "all"`. This is mostly magic, if I'm honest. Here, webpack will analyze which pieces of code are shared between all entry points (the pieces of code are called "chunks") and then will split out those pieces into separate bundles and source maps. This amplifies the speedup of the `nosources-source-map` option above: Not only does node not have to process code that'll be unused by the command that's run, but it doesn't have to process the source maps for that code either.

### Speed up screenshot

This results in some big gains. Check out the results with `truffle compile`. The first result is the new bundle created by webpack 5; the second is the existing bundle; and the third is the existing CLI, unbundled. Note that the new bundle is *faster* than the unbundled version. This is likely because less code is being loaded by node in the bundle versus the unbundled version.  

image (6)

### No need for `commands.bundled.js`? 

You'll notice that `cli.js` and `console-child.js` no longer use the bundled `commands.bundled.js` file. This is because webpack is doing the chunking for us, and will separate out shared chunks automatically, which means we no longer need to separate out this code ourselves. From my 1000 ft view, it doesn't look like we need it. We should remove that configuration from the bundle if so. 

### New faces in the `build` directory

Go check out the `build` directory after bundle. You should see some new faces, namely a bunch of numbered chunks. You should also see that the entry point files are teeny tiny (in kilobytes!). This is because chunks have been separated out into numbered files that only webpack understands, and the entry point javascript files no longer need to contain all of the code. 

I made an executive decision to limit chunk size to about ~5M. This enforces sizes are small and ensures we won't load code we don't need. I did notice a speed up when doing this, but only in the one to two-tenths of a second. Is it worth it? removing the `maxSize` key from the `splitChunks` configuration will reduce the amount of files created if we like less files better.

Here's some screenshots of the new stuff: 

image
image

### Less to download! 

I already mentioned this, but the new bundle size is wayyyyy smaller than the previous bundle size. 46M vs. 211M respectively. Gives users much less to download and should make installation easier. 

New bundle: 
image

Existing bundle: 
image

### How do you like me now? 

image


packages/core/lib/console.js
```diff
@@ -163,12 +163,19 @@ class Console extends EventEmitter {
     const spawnOptions = { stdio: ["inherit", "inherit", "pipe"] };
 
     const spawnInput = "--network " + options.network + " -- " + inputStrings;
-    spawnSync(
+    let spawnResult = spawnSync(
```
gnidan, line 166:
```suggestion
    const spawnResult = spawnSync(
```

packages/truffle/scripts/webpack5hack.js
```diff
@@ -0,0 +1,32 @@
+const fs = require("fs");
```
gnidan, line 1: Can we change the name of this file to something semantic, like `patchHighlightJsSolidity.js`?

tcoulter: Absolutely

tcoulter: Fixed and committed. Marked as resolved.

packages/truffle/scripts/patchHighlightJsSolidity.js
```diff
@@ -0,0 +1,32 @@
+const fs = require("fs");
```
fainashalts, line 1: Can we use `fse`/`fs-extra` here instead? Just for consistency I think we've been using `fse` elsewhere in the codebase.

tcoulter: Sure, good catch. Will add.

eggplantzzz (approved):
I like it!

gnidan (approved):
Nicely done!

Issue #3799, opened 2021-02-15
Having "-" in import filename break compilation in weird way (5.1.66)


hellwolf:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Since "5.1.66", there are certain import stops working in weird way. And it's been observed that it happens when the file name contains "-" symbol.

## Steps to Reproduce

Using truffle version: 5.1.66
Using solidity version: 0.7.6

- truffle init
- install `@openzeppelin/contracts`
- create a test contract, starting:

```
pragma solidity ^0.7.0;
import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```

truffle compile, it should pass.

But change to the following deliberately:

```
import "@open-zeppelin/contracts/token/ERC20/IERC20.sol";
```

Noticing a "-" in between, it fails with a mysterious error:

```
CompileError: /home/hellwolf/tmp/truffle-test/contracts/ERC20.sol:5:1: ParserError: Source "Ȣbn-zeppelin/contracts/utils/Context.sol" not found: Callback not supported.
import "@open-zeppelin/contracts/utils/Context.sol";
^--------------------------------------------------^
```

Note that, the project I am working on unfortunately has a "-" in the file name, namely https://www.npmjs.com/package/@superfluid-finance/ethereum-contracts. It really made a head scratchier!

## Expected Behavior

Should import file with its name containing "-" and compile correctly.

## Actual Results

Truffle 5.1.66 failed with mysterious error.

## Environment

* Operating System:  ubuntu
* Ethereum client: n/a
* Truffle version (`truffle version`): 5.1.66
* node version (`node --version`): 12.20.2
* npm version (`npm --version`): 6.14.11

haltman-at:
Hi, I'm a little confused by this report.  Of course importing from `@open-zeppelin/contracts` doesn't work because that's not the name of the project; it's `@openzeppelin` rather than `@open-zeppelin`.  As for the mysterious error you're seeing, that's simply solc's error message when it can't locate an imported source; we recently decided to start exposing those error messages rather than replacing them with our own as we had been.

However, you say that imports with dashes in them don't work more generaly?  Do you have a reproduction example for that we could check out?  (One that doesn't rely on trying to import a non-existent source, that is.)  Meanwhile I'll try to reproduce this myself, I suppose, and see if I can figure out what's going on.

hellwolf:
[...]

PR #3802, opened 2021-02-16
Bug fix: Re-add missing solc dependency in compile-solidity

haltman-at:
Addresses #3799.

In PR #3771, when attempting to clean up dependencies, I deleted the `solc` dependency from `compile-solidity`, because, hey, we don't invoke solc that way, right?  I searched for `require("solc")` to be sure.  Big mistake -- it's being used for `solc/wrap`, and my search didn't catch that.  Don't have that, and you can't compile!  Have the wrong version, and your compilation will be full of errors due to malformed import strings!  (I don't know why this happens, but it does.)  Anyway, re-added that dependency to fix the problem.  Sorry about that.

### DIFF
--- a/packages/compile-solidity/package.json
+++ b/packages/compile-solidity/package.json
@@ -26,7 +26,8 @@
     "request": "^2.85.0",
     "request-promise": "^4.2.2",
     "require-from-string": "^2.0.2",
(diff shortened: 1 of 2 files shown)


eggplantzzz (approved):
I like it!

PR #3803, opened 2021-02-17
Update the minimum version of Node Truffle supports

eggplantzzz:


eggplantzzz:
Do we want to roll back the versions supported a little bit more?

haltman-at (commented):
My one note here is that whie you're at it you might also want to change the use of `semver.satisfies` to a `semver.gte`, so that we're not relying on pasting strings together.

Otherwise, seems fine to me I guess, but I'm going to hold off on hitting approve on account of I feel like someone else ought to comment on what the lowest version should be.  Like, do we really want the most *recent* version of Node 10?  I know, you've already asked that question, but my point is I don't know the answer either!  (@gnidan?)

haltman-at:
Actually, you know what, I'll go ahead and say that it probably should be rolled back a bit more.  According to https://node.green/ , node 10.9.0 supports the same stuff that 10.23.0 does, so maybe 10.9.0?

eggplantzzz:
This all seems very sensible.

haltman-at (approved):
Seems good to me!

PR #3818, opened 2021-02-19
Integrate preserve functionality into @truffle/core

rkalis:
- Add preserve plugins to @truffle/plugins
- Change from @gnidan's code: plugin json contains `tag` rather than `defaultTag`
- Integrate preserve functionality into @truffle/core
- Allow command help to be a thunk instead of an object
- Add tests in plugins, core, and truffle packages


gnidan (approved):
Lookin' good!

PR #3859, opened 2021-02-21
Show progress when webpacking truffle

gnidan:
@tcoulter's excellent work in #3794 made Truffle _so much faster_, but the tradeoff there is that the build itself is slower now. Easy choice, but I couldn't stand to see nothing happening in the terminal.

This PR enables [webpack's ProgressPlugin](https://webpack.js.org/plugins/progress-plugin/) to output percentage complete to the console.

### DIFF
--- a/packages/truffle/package.json
+++ b/packages/truffle/package.json
@@ -15,8 +15,9 @@
   },
   "scripts": {
     "analyze": "./scripts/analyze.sh",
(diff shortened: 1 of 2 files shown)


haltman-at (approved):

gnidan:
Thanks @haltman-at !

PR #3861, opened 2021-02-22
Use correct bytecode for immutableReferences in @truffle/db

gnidan:
Since only runtime bytecodes can have immutables.


haltman-at (approved):

gnidan:
Thanks @haltman-at!

PR #3868, opened 2021-02-22
Allow ContractInstances in @truffle/db without transactionHash information

gnidan:
This PR removes the non-nullability requirement for ContractInstance.creation.transactionHash to accommodate situations where it may be infeasible to obtain this information. As a result, ContractInstance.network may now represent any Network on which this instance is available at the end of that Network's historicBlock.

This means a couple things:
- `network` can no longer be an ID field, since this looser constraint means `network` is no longer unique; use `contract` and `creation` when computing ContractInstance IDs (in addition to the unchanged `address`).
- When doing `project.loadMigrate()`, if any input artifacts are missing a transactionHash, load the latest block as a Network resource and use that for any such ContractInstances.

### DIFF
--- a/packages/db/src/resources/types.ts
+++ b/packages/db/src/resources/types.ts
@@ -52,7 +52,7 @@ export type Collections = {
   contractInstances: {
     resource: DataModel.ContractInstance;
     input: DataModel.ContractInstanceInput;
-    idFields: ["address", "network"];
+    idFields: ["contract", "address", "creation"];
     names: {
       resource: "contractInstance";
       Resource: "ContractInstance";
--- a/packages/db/src/network/index.ts
+++ b/packages/db/src/network/index.ts
@@ -307,7 +307,10 @@ export class Network {
 
     debug("latest %O", latest);
 
-    if (latest.historicBlock.height > this._knownLatest.historicBlock.height) {
+    if (
+      latest &&
+      latest.historicBlock.height > this._knownLatest.historicBlock.height
+    ) {
       this._knownLatest = latest;
     }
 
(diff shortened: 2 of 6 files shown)


packages/db/src/resources/types.ts
```diff
@@ -52,7 +52,7 @@ export type Collections = {
   contractInstances: {
     resource: DataModel.ContractInstance;
     input: DataModel.ContractInstanceInput;
-    idFields: ["address", "network"];
+    idFields: ["contract", "address", "creation"];
```
fainashalts, line 55: This is fine but we really should consider/plan a migration system post-haste. What if you hadn't realized the need for this until after we released DB into the wild?

fainashalts (approved):
This all LGTM but just wanna be sure I understand one mechanic: if we are missing the txHash and can't set up the Network resource like we normally would, we just fetch the latest that the DB has? But like, that could be wrong, couldn't it? Just wanna be sure we've considered some sort of way to convey that "this network wasn't computed the normal way" maybe? Maybe not, just a thought as I was reading this!

gnidan:
> if we are missing the txHash [...] we just fetch the latest that the DB has?

In this situation, we fetch the latest from the chain and add that as its own Network resource (btw, I use "fetch" to mean "from the blockchain" and "query" to mean "from @truffle/db")

> Just wanna be sure we've considered some sort of way to convey that "this network wasn't computed the normal way" maybe?

I have a change in mind for this (define a ContractInstanceNetwork resource, to make ContractInstance <---> Network many-to-many), but that will have to wait. In the meantime, I think we just have to take `ContractInstance.network` as this semi-canonical record.

packages/db/src/resources/types.ts
gnidan: Yeah, I know :o 

See below... I already have another change that I foresee here, for after release :(

gnidan:
Well, scary issues aside, gonna merge this! Thanks @fainashalts!

PR #3869, opened 2021-02-22
Allow Project.contractInstance lookup by address+network

gnidan:
This PR adds the ability to look up a Project's single contract instance by address+network, rather than by contract+network.

This is a bit of an expedient hack; a better solution might be to generalize the filtering here, but I've avoided doing that now since this approach won't incur much of a performance penalty and minimizes the risk of breaking the existing mechanisms.


fainashalts (approved):
This seems fine to me!

gnidan:
Thanks!

PR #3870, opened 2021-02-22
Report contractName in source-fetcher if possible

gnidan:
@truffle/source-fetcher is great for obtaining all the sources that went into compiling the contract instance at a given address!

Unfortunately, it does not make it easy to figure out which resulting compiled contract it was that ended up getting to deployed to that address.

This PR adds that information as an optional field to the result type, since Etherscan provides us with this information but Sourcify does not. (In the future, perhaps there's a solution when fetching from Sourcify to do an `eth_getCode` and bytecode match?)


haltman-at (approved):
I should note that using this in place of matching bytecode is dangerous, but, sure, this is a reasonable thing to add.

gnidan:
Thanks!

PR #3871, opened 2021-02-23
Add initial @truffle/db-kit demo package

gnidan:
This PR defines a new package that provides:
- a few utilities for interacting between @truffle/db and a few other Truffle libraries
- the `db-kit` bin to provide a CLI interface to these utilities

For this package's library interface, this PR provides the following interfaces:
- `DbKit.Utils.queryCompilation()`, to read compilation info from @truffle/db for a given network+address
- `DbKit.Utils.prepareProjectInfo()`, to convert that compilation info into a @truffle/decoder-compatible [ProjectInfo](https://www.trufflesuite.com/docs/truffle/codec/interfaces/_truffle_decoder.projectinfo.html) (to make it easy to initialize a decoder)
- `DbKit.Utils.fetchExternal()`, to fetch verified Solidity contracts from Etherscan, compile those, and populate @truffle/db with the resulting compilation/contract/contract instance information.

As for the CLI interface (**the exciting part**), this PR includes the following capabilities:
- Decode function information and events for a given transaction hash
- Decode storage variables for a contract at a given address

This CLI currently accepts only the `--network=<network-name>` parameter (similar to Truffle commands), and simply prompts the user for information via a terminal wizard UI.

<details>
<summary><strong>See example screenshot of <code>npx @truffle/db-kit --network=mainnet</code></strong></summary>
Screen Shot 2021-02-22 at 20 23 58
</details>

For both capabilities above, this CLI will determine all related addresses, query @truffle/db for any information about those addresses, and fetch from etherscan / compile / load to @truffle/db for any missing data. For storage decodings, the only supported "related address" is the given address itself. For transaction decodings, "related addresses" comprise (for each that exists): `transaction.to`, `receipt.contractAddress`, and all addresses in `receipt.logs`.

This PR implements this CLI using the [ink](https://github.com/vadimdemedes/ink) library, which renders React components inside the terminal.

**This package is intended to be experimental, to demonstrate @truffle/db, not as a permanent place for this behavior.** The goal of this PR is to expose a low-risk separate software component, to serve as a testbed for development before we start making all of Truffle lean on @truffle/db. Eventually, I figure, we'll want to turn this into `truffle decode` instead. But for now, the hope is just to release something cool.

As a result, this doesn't include tests (I don't know how to do React testing), and error-handling is minimal. My hope is that these omissions make sense given the intended purpose here.


packages/db-kit/package.json
```diff
@@ -0,0 +1,75 @@
...
+  "files": [
+    "dist"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/trufflesuite/truffle.git"
+  },
```
davidmurdoch, line 28: If you add a `directory` then npm will link directly to this package's directory instead of the root.

```suggestion
  "repository": {
    "type": "git",
    "url": "https://github.com/trufflesuite/truffle.git",
    "directory": "packages/db-kit"
  },
```

packages/db-kit/src/utils/prepareProjectInfo.ts
```diff
@@ -0,0 +1,75 @@
...
+function findImmutableReferences(immutableReferences, bytecode) {
+  const forBytecode = immutableReferences.filter(
+    ({ bytecode: { id } }) => id === bytecode.id
+  );
+
+  const mapped = forBytecode.map(({ astNode, length, offsets }) => ({
+    [astNode]: offsets.map(start => ({ start: start, length }))
+  }));
+
+  return mapped.reduce((a, b) => ({ ...a, ...b }), {});
```
davidmurdoch, line 74: I'm not sure of the context, but for large datasets this could be a very expensive function, compared to an optimal implementation that doesn't iterate over the array multiple times, and over each arrays properties multiple times. If this is in a hot path I'd consider optimizing.

packages/db-kit/src/utils/fetchExternal.ts
```diff
@@ -0,0 +1,192 @@
...
+}> {
+  const { config, address } = options;
+  const fetchers = await forTruffleConfig({ config });
+
+  for (const fetcher of fetchers) {
+    //now comes all the hard parts!
+    //get our sources
+    let result;
+    try {
+      result = await fetcher.fetchSourcesForAddress(address);
```
davidmurdoch, line 129: Can/should this be parallelized?

packages/db-kit/src/cli/components/Header.tsx
```diff
@@ -0,0 +1,23 @@
...
+      </Text>
+      <Newline />
+      <Text>
+        This tool provides a few utilities to demo @truffle/db and other Truffle
+        libraries. Since this is for demo purposes, please consider this package
+        experimental and subject to future deprecation: we plan to move these
+        utilities to be part of Truffle itself.
+      </Text>
+      <Newline />
+      <Text>Thanks for trying this out! We hope you enjoy.</Text>
```
davidmurdoch, line 20: This would be nice to copy into this project's README.

packages/db-kit/src/cli/components/ActivationInstructions.tsx
```diff
@@ -0,0 +1,55 @@
...
+          truffle-config.js:{" "}
+        </Text>
+        <Text italic>Normal configuration</Text>
+      </Text>
+      <SyntaxHighlight language="javascript" code={simpleInstructions} />
+      <Text color="gray">```</Text>
+    </Box>
+    <Text>
+      By default, @truffle/db stores information for all your Truffle projects
+      in one place. For your current user, this will be directory:
```
davidmurdoch, line 46: I feel like this reads a little bit better:

```suggestion
      in one place. For your current user, this will be this directory:
```

davidmurdoch (changes requested):
This is such a cool utility tool! Really great work and turnaround on getting this to this point! 👏👏👏👏👏👏👏👏


This review is a high level cursory review. I didn't look at any of the React component logic, for example or follow the logic.

I did made a few comments and suggestions, nothing too crazy though.

packages/db-kit/package.json
gnidan: Neat!

packages/db-kit/src/cli/components/ActivationInstructions.tsx
gnidan: Agreed

packages/db-kit/src/utils/fetchExternal.ts
gnidan: Right now, the existence of this loop is just unnecessary boilerplate, since we only support one fetcher in db-kit right now (Etherscan).

In general though, this can't be parallelized, since the way truffle-config's `sourceFetchers` property works is that it specifies a priority-ordered list of fetchers. Not sure if this is a requirement that's worth changing (cc @haltman-at), but I posit that any such change now would be out of scope of this PR. Let me know if you concur @davidmurdoch! Thanks!

packages/db-kit/src/utils/prepareProjectInfo.ts
gnidan: Hm, this is likely a hot path. I'll switch to use a single loop. Good catch!

packages/db-kit/src/cli/decodeTransaction/Inputs.tsx
```diff
@@ -0,0 +1,81 @@
...
+      onSubmit({
+        transactionHash,
+        transaction,
+        receipt,
+        addresses
+      });
+    }
+  }, [transaction, addresses]);
+
+  const spinners = [
```
honestbonsai, line 38: I found this section a bit difficult to follow. Looks like you have 0 up to a max of 2 elements in this array?

packages/db-kit/src/cli/decodeTransaction/Splash.tsx
```diff
@@ -0,0 +1,163 @@
...
+  const body = decoder ? (
+    <Box flexDirection="column">
+      <Box justifyContent="center" marginBottom={1}>
+        <Divider width={10} />
+      </Box>
+      <Result decoder={decoder} transaction={transaction} receipt={receipt} />
+    </Box>
+  ) : (
+    <></>
+  );
```
honestbonsai, line 137:
```suggestion
  const body = decoder && (
    <Box flexDirection="column">
      <Box justifyContent="center" marginBottom={1}>
        <Divider width={10} />
      </Box>
      <Result decoder={decoder} transaction={transaction} receipt={receipt} />
    </Box>
  );
```

packages/db-kit/src/cli/decodeTransaction/Splash.tsx
```diff
@@ -0,0 +1,163 @@
...
+          <Box paddingX={2} flexDirection="column">
+            {...spinners}
+          </Box>
+        </Box>
+      ) : (
+        <></>
+      )}
+      {body}
+    </Box>
+  );
```
honestbonsai, line 162:
```suggestion
  return (
    <Box flexDirection="column">
      {showLoaders && (
        <Box
          key="loaders"
          flexDirection="column"
          borderStyle="round"
          borderColor="dim"
        >
          <Box paddingX={1}>
            <Text dimColor bold>
              Related addresses:
            </Text>
          </Box>
          <Box paddingX={2} flexDirection="column">
            {...spinners}
          </Box>
        </Box>
      )}
      {body}
    </Box>
  );
```

packages/db-kit/src/cli/decodeTransaction/useDecodedTransaction.ts
```diff
@@ -0,0 +1,70 @@
...
+    setState: (state: EventState) => void;
+  }[] = [];
+  for (const log of receipt.logs) {
+    const [state, setState] = useState<EventState>({
+      log,
+      decoding: undefined,
+      complete: false
+    });
+    eventStates.push({ state, setState });
+  }
```
honestbonsai, line 47: Interesting, don't see this structure too often.

I feel like a single state var like `const [receiptLogs, setReceiptLogs] = useState()` is more conventional.

honestbonsai: Then on line 57 you construct the new array and `setState` just once

packages/db-kit/src/cli/hooks/useConfig.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface UseConfigOptions {
+  network: Pick<Resources.Resource<"networks">, "name">;
+}
+
+export function useConfig({ network: { name } }) {
+  const [config, setConfig] = useState<TruffleConfig | undefined>(undefined);
+
+  useEffect(() => {
+    setImmediate(() => {
```
honestbonsai, line 15: TIL what `setImmediate` is

packages/db-kit/src/cli/hooks/useDecoder.ts
```diff
@@ -0,0 +1,37 @@
...
+  const [decoder, setDecoder] = useState<WireDecoder | undefined>();
+  const { done, compilations, statusByAddress } = useCompilations(options);
+
+  useEffect(() => {
+    if (compilations) {
+      prepareProjectInfo({ compilations })
+        .then(projectInfo => forProject(options.config.provider, projectInfo))
+        .then(setDecoder);
+    }
+  }, [done]);
```
honestbonsai, line 34: Another possible missing dependency here, the eslint rules will help.

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+    props: Props;
+  };
+};
+
+export const definitions: Definitions<MenuModes> = {
+  "decode-transaction": {
+    label: "Decode transaction",
+    propsInputComponent: props => {
+      return <DecodeTransaction.Inputs {...props} />;
+    },
```
honestbonsai, line 52: If you're just forwarding the props it should be the same as

```
propsInputComponent: DecodeTransaction.Inputs,
```

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+    screenComponent: props => {
```
honestbonsai, line 53: Same as above

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+      return <DecodeTransaction.Splash {...props} />;
+    }
+  },
+  "decode-address": {
+    label: "Decode contract address",
+    propsInputComponent: props => {
```
honestbonsai, line 59: Same as above

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+      return <DecodeAddress.Inputs {...props} />;
+    },
+    screenComponent: props => {
```
honestbonsai, line 62: Same as above

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+export const Menu = (props: Props) => {
+  const { config, db, project, onDone } = props;
+  const [mode, setMode] = useState<"wait" | ModeName<MenuModes>>("wait");
+
+  const handleSelect = ({ value }) => {
+    setMode(value);
+  };
+
+  const [element, setElement] = useState(<></>);
```
honestbonsai, line 82: `null` is usually the standard to represent "nothing" being rendered. 

```suggestion
  const [element, setElement] = useState(null);
```

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const [inputProps, setInputProps] = useState<
+    ModeInputProps<MenuModes, ModeName<MenuModes>> | undefined
+  >(undefined);
+
+  useEffect(() => {
+    setInputProps(undefined);
+  }, [mode]);
```
honestbonsai, line 89: I don't really get why `mode` is in the dep array here, when it's not being used in the effect. The eslint rules will help here.

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  useEffect(() => {
+    const definition = definitions[mode];
+
+    if (mode === "wait") {
+      setElement(<></>);
+      return;
+    }
```
honestbonsai, line 97: Seems redundant to set an element into state, when we can just use this `mode === "wait"` condition directly in the render.

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+    if (definition.effect) {
+      definition.effect({ config, db, project, onDone });
+    }
+
+    if (definition.screenComponent) {
+      setElement(
```
honestbonsai, line 104: Same as above

packages/db-kit/src/cli/menu/Menu.tsx
```diff
@@ -0,0 +1,147 @@
...
+            onSubmit: setInputProps
+          })}
+          {inputProps ? (
+            definition.screenComponent({
+              ...props,
+              ...inputProps
+            })
+          ) : (
+            <></>
+          )}
```
honestbonsai, line 118:
```suggestion
          {inputProps && (
            definition.screenComponent({
              ...props,
              ...inputProps
            })
          )}
```

honestbonsai (changes requested):

honestbonsai:
Cool tool! Thanks for showing it off at the demo the other day.

honestbonsai:
Cool tool! Looks like a lot of hard work went into this.

packages/db-kit/src/cli/components/DefinitionList.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  entryComponent?: (props: {
+    nameComponent: (props: React.PropsWithChildren<{}>) => ReactNode;
+    width: number;
+    maxNameWidth: number;
+    name: string;
+    node: ReactNode;
+  }) => ReactNode;
+
+  nameComponent?: (props: React.PropsWithChildren<{}>) => ReactNode;
```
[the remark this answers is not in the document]
gnidan: Oh the capitalization was the problem! Is this not weird? `<DefinitionList NameComponent={...} />`?

packages/db-kit/src/cli/components/codec/Constructor.tsx
```diff
@@ -0,0 +1,51 @@
+import React from "react";
+import { Box, Text, Newline } from "ink";
+
+import type { Transaction, TransactionReceipt } from "web3-core";
+import { ConstructorDecoding } from "@truffle/codec";
+
+import { DefinitionList } from "@truffle/db-kit/cli/components/DefinitionList";
```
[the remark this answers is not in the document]
gnidan: Yeah I dunno why, relative imports make me cringe. This follows the pattern I established in @truffle/db for importing outside a semantic boundary. This works because I'm using tsonfig's `paths` compiler setting along with some other wizardry that I'll spare the details of unless you're curious.

packages/db-kit/src/cli/components/codec/Value.tsx
```diff
@@ -0,0 +1,53 @@
...
+          {Format.Types.typeStringWithoutLocation(result.type)}
+        </Text>
+      </Text>
+      {indexed ? (
+        <Text dimColor italic>
+          {"<indexed>"}
+        </Text>
+      ) : (
+        <></>
+      )}
```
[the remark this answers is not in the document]
gnidan: Neat! I'll want make sure this doesn't cause trouble with `strictNullChecks`... but definitely didn't realize React just played nice with nully values! Thanks for the education!

packages/db-kit/src/cli/decodeAddress/useDecodedAddress.ts
```diff
@@ -0,0 +1,63 @@
...
+          });
+        })
+        .catch(() => {
+          setVariablesState({
+            variables: undefined,
+            complete: true
+          });
+        });
+    });
+  }, []);
```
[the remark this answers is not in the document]
gnidan: Oh neat! I ran into so many problems with getting these dependency arrays right. I'll try the eslint plugin; thank you!

packages/db-kit/src/cli/decodeTransaction/Inputs.tsx
```diff
@@ -0,0 +1,81 @@
...
+  useEffect(() => {
+    if (transactionHash && transaction && receipt && addresses) {
+      onSubmit({
+        transactionHash,
+        transaction,
+        receipt,
+        addresses
+      });
+    }
+  }, [transaction, addresses]);
```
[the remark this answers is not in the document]
gnidan: I had all of them there and React complained about "too many re-renders". In this case, `addresses` is guaranteed to exist only if the others exist, so I was hoping to skirt around the code smell for now. I'll try again though, maybe it was a different underlying issue that's now fixed.

packages/db-kit/src/cli/decodeTransaction/Inputs.tsx
gnidan: That's right. Probably worth making a `<Spinners />` component or somesuch, eh?

packages/db-kit/src/cli/decodeTransaction/useDecodedTransaction.ts
gnidan: I had it your way and it was giving me a lot of trouble. If I get to it, I'll try switching back to that. How big a deal do you think it is, having all these extra states?

packages/db-kit/src/cli/hooks/useConfig.ts
gnidan: It's handy!

packages/db-kit/src/cli/menu/Menu.tsx
gnidan: oh, duh!

packages/db-kit/src/cli/menu/Menu.tsx
gnidan: I probably don't understand how that dependency array works. What I intend here is: every time the mode changes, reset the input props to be empty. Is this not correct?

packages/db-kit/src/cli/App.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  useEffect(() => {
+    if (shouldQuit) {
+      exit();
+    }
+  });
+
+  let element = <></>;
+
+  if (!error && project && db && config) {
```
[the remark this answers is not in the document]
gnidan: It's a bit ugly with the alternate cases, but I wanted to avoid nested logic boolean short-circuiting. Definitely better than before; thanks!

packages/db-kit/src/utils/fetchExternal.ts
```diff
@@ -0,0 +1,192 @@
...
+          return await Fetcher.forNetworkId(networkId, options);
+        } catch (error) {
+          if (!(error instanceof InvalidNetworkError)) {
+            throw error;
+          }
+        }
+      })
+  );
+
+  return fetchers.filter((fetcher): fetcher is Fetcher => !!fetcher);
```
haltman-at, line 111: Sorry, just a syntax question here, since I'm unfamiliar with this syntax.  Is `(x): x is T => e` the same thing as just `(x: T) => e`?  I'm more familiar with the latter, which I think is also clearer/less redundant.

Also as long as I'm nitpicking I will always suggest using `Boolean(x)` rather than `!!x` since converting something to boolean really shouldn't be thought of as double negation... (also alternatives: just `x`, or, in a context like this one, `x !== undefined`...)

packages/db-kit/src/utils/fetchExternal.ts
```diff
@@ -0,0 +1,192 @@
...
+  });
+}
+
+/**
+ * Accepts a TruffleConfig and returns a priority-sorted list of fetchers based
+ * on provided configuration and whether available fetchers support the current
+ * blockchain network.
+ * @dev HACK this currently only supports EtherscanFetcher
+ */
+export async function forTruffleConfig(options: {
```
haltman-at, line 66: Oh, factoring this out into its own function is neat.  Maybe this should go somewhere more accessible and then the debugger could use this?  (Of course, the restriction to Etherscan would have to live outside of this function.)

packages/db-kit/src/utils/fetchExternal.ts
```diff
@@ -0,0 +1,192 @@
...
+            throw error;
+          }
+        }
+      })
+  );
+
+  return fetchers.filter((fetcher): fetcher is Fetcher => !!fetcher);
+}
+
+async function fetchAndCompile(options: {
```
haltman-at, line 114: Similar comment!

haltman-at (commented):
OK, took a look at the parts related to me, just some minor comments.

packages/db-kit/src/utils/fetchExternal.ts
gnidan: In this code, I've omitted the argument type because it can be inferred; think of it like this (remember `Fetcher` here is a local variable):
```typescript
(fetcher: BaseFetcher): fetcher is Fetcher => !!fetcher
```

This tells TypeScript that this function will accept any kind of fetcher, and the runtime value will inform the type system that the argument is a _specific_ extension of that fetcher type.

Re: `!!`, this is the style I'm using in @truffle/db, so I'd prefer not to diverge from that here. Good point though, maybe worth changing!

packages/db-kit/src/utils/fetchExternal.ts
gnidan: Oh yeah all this stuff should get pulled out of here :)

packages/db-kit/src/cli/decodeAddress/useDecodedAddress.ts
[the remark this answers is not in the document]
gnidan: This one saw no complaints!

packages/db-kit/src/cli/decodeTransaction/Inputs.tsx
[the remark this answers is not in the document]
gnidan: Works like a charm :)

packages/db-kit/src/cli/hooks/useDecoder.ts
gnidan: No complaints from eslint!

packages/db-kit/src/cli/menu/Menu.tsx
gnidan: Anyway eslint seems fine with this? Let me know if you're still incredulous!

packages/db-kit/src/cli/menu/Menu.tsx
gnidan: I remember: I originally tried this, but I didn't know the bit about capitalization! Well, that's fixed now.

packages/db-kit/src/cli/menu/Menu.tsx
gnidan: Oh the reason for this is to handle the `definition.effect` business. Reworking this exceeds my React-fu. Later problem, maybe?

packages/db-kit/src/cli/menu/Menu.tsx
gnidan: (Same response as above)

packages/db-kit/src/cli/decodeTransaction/Inputs.tsx
gnidan: Can this be a later problem?

packages/db-kit/src/cli/decodeAddress/Result.tsx
```diff
@@ -0,0 +1,135 @@
...
+    // @ts-ignore
+    setWidth(measureElement(ref.current).width);
+  }, []);
+
+  const { summary, variables } = useDecodedAddress({
+    decoder,
+    address
+  });
+
+  let summaryElement = (
```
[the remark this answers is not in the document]
gnidan: This one didn't clean up quite as nicely. Can this be a later problem? I figure we will want to give this whole "multiple sections of output" business a once-over anyway.

gnidan:
@honestbonsai I believe I got all your comments; the remaining unresolved conversations are all things where I'm unsure of something and would like your advice, or changes I'd like to petition wait for later (I really don't want to invalidate the many many hours of manual testing I've already done)

> Cool tool! Looks like a lot of hard work went into this.

Thanks :) It was fun!

packages/db-kit/src/utils/fetchExternal.ts
haltman-at: Oh -- this is a type predicate.  I see.  I didn't get that; OK, that is pretty different from what I thought it was.  Nevermind.  You should have just said that. :P  (My confusion here is that I thought this was some sort of notation for the argument type, but it's actually the return type.  I'm not used to seeing type predicates as arrow functions...)

OK, nevermind, I get what this is doing now.

packages/db-kit/src/cli/decodeTransaction/useDecodedTransaction.ts
fainashalts: I kinda like the `state` and `setState`, personally. Wasn't aware that the other way was the more conventional approach!

packages/db-kit/src/utils/prepareProjectInfo.ts
```diff
@@ -0,0 +1,75 @@
+import { Resources } from "@truffle/db";
+import type { ProjectInfo } from "@truffle/decoder";
+
+export interface PrepareProjectInfoOptions {
+  compilations: Resources.Resource<"compilations">[];
+}
+
+export async function prepareProjectInfo({
```
fainashalts, line 8: This is neat! So the code here makes the compilation output decoder-friendly? What is the plan for this eventually? When DB replaces artifacts will the decoder be rewritten to expect things in the DB structure?

packages/db-kit/src/utils/queryCompilation.ts
```diff
@@ -0,0 +1,113 @@
...
+              compiler {
+                name
+                version
+              }
+
+              sourceMaps {
+                bytecode {
+                  id
+                }
+                data
```
fainashalts, line 40: For a minute I thought this was wrong. Which reminded me we should maybe update the data model at https://trufflesuite.github.io/artifact-updates/data-model.html. Or maybe we aren't using it anymore? I can't remember but want to put this here in case we should update it...

fainashalts:
+1 This is so great! Everything looks fine to me, looks like you have some changes requested from others but this is all very exciting! I skimmed the react stuff since we have Heyse's expertise for it. Thanks Heyse!

packages/db-kit/src/utils/prepareProjectInfo.ts
gnidan: The decoder defines its own format, partially because that's just how we did it, but also because it has its own requirements. We might be able to update parts of the ProjectInfo type to align with DB, but it's not really decided yet. I've also got it in mind that maybe the better alternative is to have the decoder query @truffle/db directly (as an alternative to passing in this information upfront)

packages/db-kit/src/utils/queryCompilation.ts
gnidan: Ahhh yeah. It'd be nice to update the data model, but maybe not before release? I barely use it anymore; the typeDefs have really become more canonical. What do you think?

gnidan:
Thanks @fainashalts!

packages/db-kit/src/utils/fetchExternal.ts
gnidan: heh my default is to explain the idea I guess. but I should know that you know terms :)

davidmurdoch (approved):

packages/db-kit/src/cli/components/DefinitionList.tsx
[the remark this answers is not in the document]
honestbonsai: Yea its a little weird looking, but it's technically correct, as components are always capitalized, vs React nodes which are not capitalized.

packages/db-kit/src/cli/components/codec/Constructor.tsx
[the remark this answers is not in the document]
honestbonsai: Thats fine with me

packages/db-kit/src/cli/decodeAddress/Result.tsx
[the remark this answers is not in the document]
honestbonsai: Yea later is fine

packages/db-kit/src/cli/decodeAddress/useDecodedAddress.ts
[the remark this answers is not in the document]
honestbonsai: Cool, good with me then

packages/db-kit/src/cli/decodeTransaction/Inputs.tsx
honestbonsai: Yea works for me

packages/db-kit/src/cli/decodeTransaction/useDecodedTransaction.ts
honestbonsai: React Hooks always have to run in the same order. So you're letting something from the outside (the `receipt.logs`) control the ordering that the hooks are run in.

packages/db-kit/src/cli/menu/Menu.tsx
honestbonsai: If the linter is fine, I'm fine haha

packages/db-kit/src/cli/menu/Menu.tsx
honestbonsai: Sure later works

packages/db-kit/src/cli/menu/Menu.tsx
honestbonsai: Later is fine

honestbonsai (approved):
ok rereviewed, I think everything got pushed to a later, which I’m fine with if issues are made so we don’t forget

gnidan:
Thanks everyone! Merging this now.

PR #3883, opened 2021-02-26
Bug fix (internal improvement?): Allow txlog to process internal calls happening at end of tx

haltman-at:
So txlog had a problem where internal calls at the very end of a transaction wouldn't have their info properly recorded.  (I'm a little unclear on why this happens, or if the info we get is really meaningful, but we should at least include it.)  The problem was twofold:

1. Because it occurred at the very end, and rather than checking the jump location we check what's after the function definition node, this would get processed as a jump into internal sources (which it's not), so I had to special-case that;
2. Because it happend on the very last step, it would never get processed, as we don't update things happening after the last step!  So I had to special case that.

This is a bit of a hack, but the info is there now at least?

### DIFF
--- a/packages/debugger/lib/txlog/selectors/index.js
+++ b/packages/debugger/lib/txlog/selectors/index.js
@@ -202,10 +202,8 @@ let txlog = createSelectorTree({
         (ready || stepsRemaining <= 2) && //HACK: see below
         node &&
         node.nodeType === "FunctionDefinition" &&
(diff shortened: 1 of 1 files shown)


haltman-at:
Note: I should really change both the upper bounds to `2` later before I merge this...

gnidan (approved):

haltman-at:
Actually no, instead I removed the second check and just left the first one in.

PR #3885, opened 2021-02-26
Make minor edit to db help output

eggplantzzz:
This edits the help output for the db command. This is an attempt to format it more similarly to the output for the other commands. I'm not attached to the changes made in this PR so we can close it if people don't like this as much. I'll post some screenshots to show the comparison. Here is the output with the changes made in this PR 
Screen Shot 2021-02-26 at 10 45 12 AM


eggplantzzz:
Here is the logging output without the changes introduced in this PR. 
Screen Shot 2021-02-26 at 10 49 06 AM

gnidan (approved):
Looks so much better!

PR #3897, opened 2021-03-01
Internal improvement: Add new, compatible nativize function and some missing codec/decoder functions

haltman-at:
**This PR is a breaking change to `@truffle/codec`.**

Redo of #3783 because it accidentally got merged early.  Please see description and all comments there.


gnidan (approved):

haltman-at:
Note: I also added web3-style nativization of external functions, even though ethers per se doesn't do that, just web3's hacked version.

Issue #3896, opened 2021-03-01
Error: Cannot find module '@truffle/db'


Sptz:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Installing Truffle yields errors in requesting Python as well as Visual Studio. This is both on a year old Windows 10 installation as well as a fresh one. Post installation, running any truffle command apart from _init_ results in the error message in the title. 

## Actual Results

Error message from typing _truffle test_ for example:

```
Error: Cannot find module '@truffle/db'
Require stack:
- C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\cli.bundled.js
at Function.Module._resolveFilename (internal/modules/cjs/loader.js:880:15)
at Function.Module._load (internal/modules/cjs/loader.js:725:27)
at Module.require (internal/modules/cjs/loader.js:952:19)
at require (internal/modules/cjs/helpers.js:88:18)
at Object.168060 (C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\webpack:\external "@truffle\db":1:1)
at __webpack_require__ (C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\webpack:\webpack\bootstrap:18:1)
at Object.577135 (C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\workflow-compile\index.js:12:28)
at __webpack_require__ (C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\webpack:\webpack\bootstrap:18:1)
at Object.952422 (C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\core\lib\testing\Test.js:9:25)
otstrap:18:1)
at prepareConfigAndRunTests (C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\core\lib\commands\test\prepareConfigAndRunTests.js:3:16)
at Object.run (C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\core\lib\c at processTicksAndRejections (internal/process/task_queues.js:93:5)
at Command.run (C:\Users\Pedro\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\core\lib\command.js:136:1)
Truffle v5.2.1 (core: 5.2.0)
[...]

PR #3898, opened 2021-03-01
Handle situations where @truffle/db could not install

gnidan:
@truffle/db is an optional dependency of Truffle because it uses sqlite3, which requires natively compiled components. If you can't install those, you won't be able to install @truffle/db. But of course Truffle shouldn't fail!

Problem is, we weren't actually `try`/`catch`-ing the corresponding `require()` statements!

This PR should fix that.

Ref: #3896

### DIFF
--- a/packages/migrate/Migration.js
+++ b/packages/migrate/Migration.js
@@ -7,7 +7,11 @@ const {
   Web3Shim,
   createInterfaceAdapter
 } = require("@truffle/interface-adapter");
-const {connect, Project} = require("@truffle/db");
+
+let Db;
+try {
+  Db = require("@truffle/db");
+} catch {}
 
 const ResolverIntercept = require("./ResolverIntercept");
 
@@ -91,7 +95,7 @@ class Migration {
         const message = `Saving migration to chain.`;
 
         if (!this.dryRun) {
-          const data = {message: message};
+          const data = { message: message };
           await this.emitter.emit("startTransaction", data);
         }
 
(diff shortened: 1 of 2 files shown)


cds-amal (approved):
LGTM!

packages/migrate/Migration.js
```diff
@@ -109,17 +113,22 @@ class Migration {
...
         .contracts()
         .map(abstraction => abstraction._json);
-      if (this.config.db && this.config.db.enabled && artifacts.length > 0) {
-        const db = connect(this.config.db);
-        const project = await Project.initialize({
+      if (
+        Db &&
+        this.config.db &&
+        this.config.db.enabled &&
+        artifacts.length > 0
```
cds-amal, line 120: Not anything to hold back this PR, but this could be refactored out.

packages/workflow-compile/index.js
```diff
@@ -92,27 +95,32 @@ const WorkflowCompile = {
...
 
     await fse.ensureDir(config.contracts_build_directory);
 
-    if (options.db && options.db.enabled === true && contracts.length > 0) {
+    if (
+      Db &&
+      options.db &&
+      options.db.enabled === true &&
+      contracts.length > 0
+    ) {
```
cds-amal, line 113: DRY refactor opportunity

cds-amal (commented):
Nothing to hold back this PR, but something to consider.

packages/migrate/Migration.js
gnidan: Let's figure out where to put this :)

gnidan:
Hooray we have green! Thanks @cds-amal; made a separate issue for your concerns!

PR #3900, opened 2021-03-02
Bug fix: Fix logging during 'pragma' compilation

eggplantzzz:
When using the experimental "pragma" setting for compilation, the sources to be compiled were not being logged. This PR extracts the logic for logging into its own file and is used in this flow.


packages/compile-solidity/index.js
```diff
@@ -84,7 +84,7 @@ const Compile = {
       return { compilations: [] };
     }
 
-    Compile.display(compilationTargets, options);
+    displaySources({ paths: compilationTargets, options });
```
haltman-at, line 87: Incorrect function name here, please fix!

haltman-at (changes requested):
Looks good except for one thing that needs fixing!

packages/compile-solidity/index.js
eggplantzzz: Ah missed that one. Good catch!

haltman-at (approved):

Issue #3903, opened 2021-03-02
Add library file path to link references in compiler output


haltman-at:
## Issue

This would be an additional field alongside the library name.  Solc gives us this, but we discard it.  It's important though for disambiguation when the same project contains two libraries with the same name.

## Environment

* Truffle version (`truffle version`): 5.2.2

Issue #3838, opened 2021-02-20
Truffle codec's bytecode decodings should include link references


haltman-at:
## Issue

Currently, when decoding returned bytecode, Truffle Codec will tell you what the bytecode is and the value of its immutables, but not the addresses of its linked libraries.  It should have the capability to do that.

Note that doing this will almost certainly require changing Truffle Codec over to the new (now pretty old) bytecode format finally.  Although I'm unclear on some of the details here; depending on just what information we need, this may end up being a Truffle-DB-only feature.  Not sure.

PR #3905, opened 2021-03-03
Internal improvement: Add file path info for link references to compilation output

haltman-at:
Addresses #3903; does not address #3838, that'll come later.  Figured I'd do this quickly first though, and I can get #3838 later.

We've been discarding the library file path info that `solc` gives us in link references.  However, this can be important for distinguishing just which library is meant if there are multiple libraries with the same name (since you can't define two contracts with the same name in the same file; and this is how `solc` distinguishes after all).  We probably don't want to store it this way internally in Truffle DB or in the decoder or anything, but it should at least be an unambiguous distinguisher which can be used on intake to convert into our own unambiguous format.

Note that this info will necessarily be `null` when shimmed, so, y'know, be prepared for the possibility that this info isn't there.  But it's nice to be able to use it when present.


haltman-at:
OK, it seems this is failing CI because Truffle DB isn't prepared to handle the extra output.  I'll have to look into this.

haltman-at:
I'm going to go ahead and close this, I don't think there's any need to leave this open...

PR #3907, opened 2021-03-03
Collect work toward developing support for ethpm v3

eggplantzzz:
For reference https://github.com/trufflesuite/truffle/pull/3504


haltman-at:
Note that this needs to take account of the changes to `resolveDependencyPath` in #3720.  Specifically, whereas before `resolveDependencyPath` could assume the path passed in is relative (because that was checked elsewhere), now it's responsible for checking if it's relative, and returning it unchanged if it's not.  I changed this in all the other resolver sources, but since this PR adds a new one, it'll have to take account of that.

eggplantzzz:
closing this until this is ready to be more actively worked on - presumably after @truffle/db is more developed?

PR #3911, opened 2021-03-08
Update version of Geth used

eggplantzzz:
Do not merge. This is just to see how CI behaves with this version of Geth.


haltman-at:
Confirmed by @eggplantzzz to be OK to merge now! :)

Issue #3849, opened 2021-02-20
Rename constants from snake case to camel case


haltman-at:
## Issue

We have a lot of constants in a number of places that are named in snake case, these should be renamed to camel case.

PR #3916, opened 2021-03-10
Renamed snake_case to camelCase at desired files

gitpurva:
#3849 I have renamed at appropiate files to camelCase. If any changes then please let me know.

### DIFF
--- a/packages/artifactor/test/customoptions.js
+++ b/packages/artifactor/test/customoptions.js
@@ -14,7 +14,7 @@ describe("Custom options", () => {
       prefix: "tmp-test-contract-",
     });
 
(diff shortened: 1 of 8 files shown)


haltman-at (approved):
Looks good to me, thanks for doing this!  There's more to do for #3849, but this is a useful start.

Two more I spotted while reviewing this, that you may want to go back and add if you have time (but it's fine if you don't): In `core/test/ethpm.js` there's a variable `expected_contract_name` that you missed, and in `resolver/test/globalnpm.ts` there's a variable `read_file_sync_stub` that you missed (a different one from the ones with that name that you got!).

Thanks for helping out!

Issue #3831, opened 2021-02-20
Support decoding return values from Vyper transactions (via ABI)


haltman-at:
## Issue

The debugger will decode the return value at the end of a transaction, but it doesn't support this for Vyper.  Because return value decoding is ABI-based and Vyper uses the same ABI as Solidity, doing this shouldn't be too hard; it just isn't possible at the moment due to some awkwardness with the debugger internals.  Still, it shouldn't be too hard and it's something we should do.

haltman-at:
Note: This problem will also apply to Yul once we support Yul (#1906).  (Yul has no ABI, of course, but the problem still applies to return values from constructors.)  Again, this is purely a problem with debugger internals.  More specifically the issue is that it relies on there being a `ContractDefinition` AST node for the contract.  This shouldn't be necessary; it's just arranged that way at present.  So the problem is to rearrange it so that's no longer a requirement for decoding.

Issue #3827, opened 2021-02-20
Make sub-line breakpoints work when debugging Vyper or other languages


haltman-at:
## Issue

Currently these work by AST node ID, and so only work for Solidity and Yul.  These should be changed to work by source range instead, so they can work for anything.  Note that this would be a breaking change to the debugger, but not one that should really affect anyone else (the Teams debugger (@kevinbluer) doesn't support sub-line breakpoints yet, although who knows if it might in the future).

Issue #1906, opened 2019-04-12
Support Yul as a language


axic:
With [Solidity 0.5.7](https://github.com/ethereum/solidity/releases/tag/v0.5.7) it is possible to compile Yul contracts via "Standard JSON".

It follows the very same interface as compiling Solidity contracts and as a result it should be simple enough to support it in Truffle.

Please let me know about any questions here, I'm happy to answer any.

CruzMolina:
Hey @axic! Been taking a look at getting this setup. I noticed a couple of things that _appear_ to be off but I'm not sure if I'm doing something wrong.

If I have a `sample.json` that includes:
```
`{
    "language": "Yul",
    "sources": { "input.yul": { "content": "{ sstore(0, 1) }" } },
    "settings": {
        "outputSelection": { "*": { "*": ["*"], "": [ "*" ] } },
        "optimizer": { "enabled": true, "details": { "yul": true } }
    }
}`
```

and I run `solc --standard-json sample.json` using `0.5.7+commit.6da8b019.Darwin.appleclang`, `solc` hangs indefinitely.

Also, if I have a `sample.yul` that includes:
```
{ sstore(0:u256, 1:u256) }
```

and I run `solc --yul sample.yul`, `solc` outputs:
```
Warning: Yul and its optimizer are still experimental. Please use the output with care.
input.yul:1:3: Error: Function not found.
{ sstore(0:u256, 1:u256) }
  ^----^
```

(the above output also occurs if I try `solc --evm-version <version> <yulFile>` or `solc --machine <target> <yulFile>`)

haltman-at:
OK @axic (I'll also ping @chriseth here :) ) so I took a stab at implementing this the other day.  Unfortunately it turns out that there are some issues with supporting Yul and I don't think we're going to do this right now, but perhaps in the future once some of this is resolved.  Note I tried this with solc 0.8.1.

The biggest problem is that the Yul compiler only accepts a single file at a time right now.  It sounds like you're intending on supporting multiple file compilation at some future point, and I think at the least it would make sense for us to wait until that exists.  I'm wary of supporting the single-file version if the multi-file version is going to happen later, because... well, I'll skip the details, but supporting both wouldn't play nicely with how we have `compile-solidity` laid out, and we'd have to reorganize some things that I'd hope to avoid.

[...]

PR #3920, opened 2021-03-12
Enhancement: Add experimental support for Yul compilation

haltman-at:
This PR addresses #1906, as requested by @axic, adding Yul support, or at least initial experimental Yul support, into Truffle.

To do this I did *not* add a new package, but rather altered `compile-solidity` so that it would compile Yul as well as Solidity.  This is a bit of a hack, but it works.

Specifically, compile-solidity will now separate out Solidity (and JSON) files from Yul files.  Solidity (and JSON) files will be compiled as normal.  Yul files will be run one-by-one (since Yul only supports single-source compilation) through the compiler.  (This change is made both to `sources()` and to `sourcesWithDependencies()`.)  So, each Yul file will result in a separate compilation in the output.

To make this work, the `run` function now accepts an optional `language` parameter (defaulting to `"Solidity"`); you can set it to `"Yul"` to compile Yul.  Also, since Yul does not support imports, I simply skipped import processing for Yul files, rather than attempt to ensure that import processing would work with them.  (I also loaded sources via `fs` rather than the resolver... let me know if that's not OK!)  Also, using Yul is disallowed with pragma-based compilation, as Yul has no version pragmas.

One complication is that Yul [does not currently](https://github.com/ethereum/solidity/issues/11043) provide `sources` output.  So I had to alter `processContracts` so that if, if `sources` is absent, it simply won't populate the fields that rely on `sources`.  But also, I had to make sure *we* still emitted `sources` output somehow.  Fortunately, since Yul is single-source, I just added a check: If `sources` is absent and there was exactly one source, use that to populate our `sources` array.

There was another issue: Currently, any attempt to compile Yul produces a compiler warning, warning you that Yul is still experimental.  Since Yul must be invoked one file at a time, this meant that compiling a project with multiple Yul files would produce multiple copies of this compiler warning.  So I made it so that we suppress this repetitive compiler warning, and instead emit our own warning if any Yul files are compiled.  It's not clear how we'd handle this in the future, once Yul is less experimental, but @gnidan has discussed creating a more general system for compiler warning consolidation (that could e.g. also be applied to license warnings), so if we do that then I guess that would cover it.

I also had to add workarounds for [this issue with Yul sourcemaps](https://github.com/ethereum/solidity/issues/11068).

While this makes it possible to compile, migrate, and debug Yul, using Yul definitely comes with a number of caveats; I'm not going to claim this is full Yul support or anything.   Still, it's a start, at least.  Here are some caveats I can think of:

1. Solc emits no ABI when compiling Yul, because it has no way to.  Since we require an ABI, we set it equal to `[]`.  But this basically means all contract interactions must be performed manually.
2. Moreover, because there's no ABI, if your Yul contract's constructor is supposed to take arguments, you won't be able to deploy it with Truffle.  The lack of a `constructor` ABI entry will cause Truffle to assume that the constructor takes no arguments.  I don't believe we offer any way to deploy a contract while manually giving it arguments in hexadecimal.
3. When using Yul, there is no guarantee that you will get a deployed bytecode for your contract; or, if you do, that it will be correct.  This would make the deployed contract impossible to recognize, so transactions against the contract could not meaningfully be debugged.
4. If you use Yul's immutables mechanism, the compiler [does not emit](https://github.com/ethereum/solidity/issues/11044) `immutableReferences`, potentially again leading to the deployed bytecode being unrecognizable, with no ability to debug transactions against the contract.  (Then again, maybe this doesn't matter that much, because why would you use immutables when you can't deploy contracts that require arguments...?)
5. Solc currently [does not emit ASTs](https://github.com/ethereum/solidity/issues/9590) when compiling Yul, so the debugger won't be able to track variables when debugging it.  (And certain breakpoints won't work due to debugger bug #3827.)
6. Similarly, debugger bug #3831 also applies, although this is only relevant when debugging a deployment.
7. Our Yul syntax highlighting is currently only intended to work for Solidity-assembly Yul, not standalone Yul, so standalone Yul keywords and builtins like `object` or `linkersymbol` won't be highlighted.

...honestly, I'm not even sure that's all of them.  I'm calling this "experimental" for a reason.  But, here it is, for those who want it...


haltman-at:
Hm. The test I added seems to be running out of memory on Node 10...?  No idea why, it runs fine locally using Node 10.

haltman-at:
So, um, I don't know what to do about this. It's only failing on Node 10 CI (although it wasn't failing locally), but now I can't even bootstrap locally on Node 10, so I'm kind of stuck here.  I mean notionally I could try just think up a less memory-intensive version by examining the code for where lots of memory is used (maybe I could run some sort of memory profiling under Node 12 under the assumption the high-memory-use spots are the same), but that sounds fairly dicey.  I could just remove the test, but, uh, I don't really want to do that for obvious reasons.  Or we could just wait until we drop Node 10 support in August...?  I don't like leaving PRs outstsanding that long, that's a good way to get nasty merge conflicts.  We can't merge it as-is; it'll break all our builds.  @gnidan, any ideas? :-/

haltman-at:
OK, I got this passing by downgrading the solc used in the test to 0.5.17, removing the source map part of the test (no Yul sourcemaps before 0.6.4), and also stripping it down and turning up the timeout -- those last two parts may have been unnecessary given that the real solution was the downgrade, oh well.  But I can beef the test back up if people want?

packages/compile-solidity/index.js
```diff
@@ -63,54 +87,76 @@ const Compile = {
...
-              contracts,
-              compiler: { name, version }
-            }
-          ]
-        }
-      : { compilations: [] };
+    return {
+      compilations: [...solidityCompilations, ...yulCompilations]
+    };
   },
```
gnidan, line 160: OK hack is probably fine. Could we at least split this up into separate private methods or functions, one for Solidity specifics and one for Yul specifics?

gnidan (changes requested):
This generally seems reasonable, but we can't leave `sourcesWithDependencies` so trawling

packages/compile-solidity/index.js
haltman-at: OK, I've factored out the Yul parts into their own function. I left the Solidity parts in place because that's both the original functionality and also much simpler -- it didn't seem worth factoring.  But obviously if you really think that should be pulled out as well it could be.

gnidan (approved):
Cool, thanks for splitting out the Yul stuff!

Issue #3913, opened 2021-03-09
"truffle migrate —reset" failed from "geth 1.10.0"


daenamkim:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

I have been working with truffle and experienced `truffle migrate —reset` is not working from geth 1.10.0, it showed `only replay-protected (EIP-155) transactions allowed over RPC`. I found out workaround from https://github.com/ethereum/go-ethereum/releases/tag/v1.10.0, adding `--rpc.allow-unprotected-txs` on local geth running. For local geth will be fine but I am not sure after all remote nodes (of using geth) are updated to v1.10.x.

## Steps to Reproduce

- Upgrade your local geth node to v1.10.x
- Setup your truffle and config: provider and network ID points to your local geth
- run `truffle migrate --reset` multiple times (I am not sure on the first because I have updated geth while I was working)

## Expected Behavior

`truffle migrate --reset` should work without any problem (in my opinion) or show a guide what a has to do at least.

## Actual Results

```
1_initial_migration.js
======================

   Replacing 'Migrations'
   ----------------------

Error:  *** Deployment Failed ***

"Migrations" -- only replay-protected (EIP-155) transactions allowed over RPC.
```

## Environment

* Operating System: macOS 11.2.1 (20D74)
* Ethereum client: geth v1.10.1
* Truffle version (`truffle version`): Truffle v5.2.3 (core: 5.2.3)
* node version (`node --version`): v14.15.5
* npm version (`npm --version`): 6.14.11 (FYI, I am not using npm but yarn 1.22.10)

Jovonni:
+1 I thought I was going crazy @daenamkim! 

Everything was working until randomly started getting this error, I just started seeing this error about 3-4 hours ago.  Did you by any chance change your secrets you were using? That was the last thing I remember doing before seeing this, but I am unaware if it was related. I am also on a mac btw

```bash
npx truffle migrate --reset --network ropsten

Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.



Starting migrations...
======================
> Network name:    'ropsten'
> Network id:      3
> Block gas limit: 8000000 (0x7a1200)


1_initial_migration.js
======================

   Replacing 'Migrations'
   ----------------------

Error:  *** Deployment Failed ***

"Migrations" -- only replay-protected (EIP-155) transactions allowed over RPC.

```

# Results of "networks --clean"
[...]

PR #3923, opened 2021-03-16
Bug fix: Include the chain id with tx params to be eip-155 compliant

eggplantzzz:
This PR includes the chain id in the tx params so that sending transactions through web3 will be EIP-155 compliant.

There are probably some issues to be resolved still for chains with custom chain ids; this will most likely be the case for Geth 1.10.0 and newer.

**Update**: It seems that this works for Geth without any extra configuration regarding `chainId`.

### DIFF
--- a/packages/artifactor/package.json
+++ b/packages/artifactor/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "author": "Tim Coulter",
   "repository": "https://github.com/trufflesuite/truffle/tree/master/packages/artifactor",
-  "version": "4.0.97",
+  "version": "4.0.98",
   "main": "dist/index.js",
   "scripts": {
     "build": "tsc",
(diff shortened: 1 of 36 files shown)


gnidan (approved):

gnidan:
So much digging and experimenting for a change this small :)

fainashalts (approved):
Sweet! Can confirm this addresses an EIP-1155 issue I have recently seen!

haltman-at:
This should probably be cached to avoid excessive network requests.  However, this will do for now, and we can change that later.

eggplantzzz:
@haltman-at @fainashalts @gnidan If you get a chance, please test this PR on your own to ensure that this change works for you.

haltman-at:
Note this PR should of course also fix #3913.

haltman-at:
Oh @eggplantzzz I just realized, you should really unpin the Geth version on this PR so that CI can properly test it!  You should do that before merging.  (Also possibly the caching change if you want but otherwise I'll get that later.)

eggplantzzz:
@haltman-at I'll go ahead and unpin geth but leave the cacheing change for later.

eggplantzzz:
I kept the version of Geth pinned to 1.9.25 as the newer versions are still breaking CI. My strategy is to most likely investigate in the next week.

shoenseiwaso:
Note that this fix did not address #3913.

PR #3926, opened 2021-03-18
Implement @truffle/preserve-to-buckets

rkalis:
- Connect with buckets using auth keys
- Clear existing bucket contents
- Upload files to the bucket
- Add a skipped test with production test
- Improve preserve error logging for thrown non-error objects


gnidan (approved):
Looks good!

Issue #3924, opened 2021-03-17
Remove source-map-support from packages that shouldn't have it?


haltman-at:
## Issue

We currently have `source-map-support` in every package that gets built.  According to @davidmurdoch, it should only go in top-level CLI packages, not library packages.  Someone should determine whether this is true and remove it from the others if so.

## Environment

* Truffle version (`truffle version`): 5.2.4

haltman-at:
Oops, this is done now.

PR #3929, opened 2021-03-19
Internal improvement: Remove source-map-support from most packages, add it to db-kit

haltman-at:
Addresses #3924.  This removes `source-map-support` from everywhere that isn't a top-level CLI application/script.  So it's removed from all packages from `core` (where it's used in `cli.js`) and `environment` (where it's used in `chain.js`).  In addition, I added it to `db-kit`, adding the registration in `bin/cli.ts`.

### DIFF
--- a/packages/abi-utils/package.json
+++ b/packages/abi-utils/package.json
@@ -45,7 +45,6 @@
   "dependencies": {
     "change-case": "3.0.2",
     "faker": "^5.3.1",
-    "fast-check": "^2.12.1",
-    "source-map-support": "^0.5.19"
+    "fast-check": "^2.12.1"
   }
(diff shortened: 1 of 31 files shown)


eggplantzzz:
Don't we need source map support in any built package? Especially if they are going to be used on their own.

haltman-at:
So I was a little uncertain of this to be honest.  But @gnidan and @davidmurdoch say it should only go in top-level CLI stuff, not just any built package.  I don't really understand why this is; hopefully they can clarify.  They also mentioned that other people putting`source-map-support` in library packages is causing problems for us with Jest...

gnidan (approved):

PR #3931, opened 2021-03-19
Refactor recipe dependency system

rkalis:
- Consolidate loaders and recipes
- Recipes have input and output labels
- If a recipe specifies an input label they need another recipe that
  has the same label as an output label
- @truffle/core will provide the "initial input label" (path)
- Refactor all existing recipes/loaders to match this new system


gnidan (approved):
All in all, this looks just like what I'd expect (and then some! Getting rid of loaders is awesome)
 
I'm a little unsure about the namespace for labels (e.g. `filecoin-deal-cid` seems a bit ugly, but I think maybe it's not worth bikeshedding). Did you have any alternative ideas?

rkalis:
We could use camelcase for the namespace so they can be accessed as `inputs.filecoinDealCid` rather than `inputs["filecoin-deal-cid"]`.

PR #3943, opened 2021-03-23
Implement parser for mapping key lookup (e.g. `m[0][1]`)

gnidan:
Had this lying around from when I discovered [parjs](https://github.com/GregRos/parjs) a few weeks ago. Decided to clean it up as a draft for review, in case it's something we can utilize in @truffle/decoder.

As it stands, this PR adds a new package @truffle/parse-mapping-lookup, whose name is terrible and needs to change. This new package exposes a function `parse()`, to accept a struct/mapping lookup expression (`m[0]`, `ledger.balances[0]`, etc.) and return an AST representation of that expression.

This handles literal floating point numbers and strings, but it doesn't do anything special with those values. It does handle escape sequences (thanks to [parjs's own example](https://github.com/GregRos/parjs/blob/master/src/examples/json.ts)). It **currently loses information**, in that `m["0x"]` will parse the same as `m["\"0x\""]`. @haltman-at I figure this is something that would be handled by @truffle/encoder, if we want it, so I didn't get fancy.

This also squashes all possibly useful error handling; that will likely be something to improve even slightly before merging.

See [test cases](https://github.com/trufflesuite/truffle/blob/decoder/parjs/packages/parse-mapping-lookup/src/parser.spec.ts) for an overview of supported syntax.

Possibly minor note: I borrowed the original tsconfig for this new package from @truffle/db, so it uses `src/`, not `lib/`). This was because it seemed to be the fastest way to get jest working for tests, but we might want to update this to borrow from one of the codec packages instead.


packages/parse-mapping-lookup/src/ast.ts
```diff
@@ -0,0 +1,111 @@
...
+    kind: "pointer",
+    path
+  };
+};
+
+/*
+ * Lookup
+ */
+
+export interface Lookup {
```
haltman-at, line 66: These names -- `Lookup` vs `Access` -- are not very intuitive.  Can we replace them with something more explanatory?  Say `MemberAccess` and `IndexAccess`, matching Solidity?

packages/parse-mapping-lookup/src/parser.ts
```diff
@@ -0,0 +1,139 @@
...
+
+  // Repeat the char/escape to get a sequence, and then put between quotes to get a string
+  export const stringP = stringEntriesP.pipe(many(), stringify(), between('"'));
+}
+
+const numberP = int().pipe(or(float()));
+
+const stringP = Strings.stringP;
+
+const literalP = numberP.pipe(
```
haltman-at, line 88: Solidity has other sorts of literals that I think need to be accounted for.  Specifically:
1. Boolean literals (`true` and `false`)
2. Hex string literals -- I assume static bytestring literals `0x1234` (and address literals) fall under numeric literals, but there's also dynamic bytestring / hex string literals `hex"deadbeef"`.
3. It also might be worth supporting Solidity units here, as those are effectively part of numeric literals?

packages/parse-mapping-lookup/src/ast.ts
```diff
@@ -0,0 +1,111 @@
...
+    kind: "access",
+    index
+  };
+};
+
+/*
+ * Literal
+ */
+
+export interface Literal {
```
haltman-at, line 100: So, I don't like the way literals work here -- especially the way that `0` and `"0"` are not distinguished from one another.  I'd suggest having multiple node types for literals, so that we can distinguish numeric literals and other literals from string literals.  So we could have, IDK, `text: string` for non-string literals (the actual text of the literal), and `content: string` for string literals (the content of the string, not the text of the literal), perhaps?

packages/parse-mapping-lookup/src/ast.ts
```diff
@@ -0,0 +1,111 @@
...
+  };
+};
+
+/*
+ * Access
+ */
+
+export interface Access {
+  kind: "access";
+  index: Literal;
```
haltman-at, line 85: So here's one more thing -- what about enums?  Those aren't *literals* per se.  One possibility would be to allow not just literals here, but I think the more appropriate route (although admittedly hackier) would be to allow "enum literals" as a type of literal as well (in addition to the other types I suggested above).

haltman-at (commented):
OK, I think my biggest comment is that I think we should distinguish string literals from other literals.  But see below for more detailed comments.

I guess some of these comments are like "is this supposed to be Solidity or JS", and I guess the answer is Solidity, so, uh, I guess it's the Solidity branch of those conditionals that you want to take. :P

packages/parse-mapping-lookup/package.json
```diff
@@ -0,0 +1,84 @@
...
+  "directories": {
+    "dist": "dist"
+  },
+  "files": [
+    "dist"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "repository": {
```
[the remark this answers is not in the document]
gnidan: Yes!

packages/parse-mapping-lookup/src/parser.ts
haltman-at: Oops, one more I forgot about -- string literals are normally restricted to ASCII, but you can also do unicode string literals as `unicode"..."`; better account for that too.

haltman-at: Oops, one more I forgot about -- string literals are normally restricted to ASCII, but you can also do unicode string literals as `unicode"..."`; better account for that too.

packages/parse-mapping-lookup/src/ast.ts
gnidan: Updated to `MemberLookup` and `IndexAccess` if you're good with that.

haltman-at: Sure, that's sufficiently unambiguous.

packages/parse-mapping-lookup/src/ast.ts
gnidan: Updated this to add `type` field with multiple distinct extension interfaces (`BooleanLiteral`, etc.)

packages/parse-mapping-lookup/package.json
```diff
@@ -0,0 +1,84 @@
...
+  "scripts": {
+    "build": "yarn build:dist && yarn build:docs",
+    "clean": "rm -rf ./dist",
+    "prepare": "yarn build",
+    "build:dist": "ttsc",
+    "build:docs": "typedoc",
+    "madge": "madge ./src/index.ts",
+    "test": "jest --verbose --detectOpenHandles --forceExit --passWithNoTests"
+  },
+  "devDependencies": {
```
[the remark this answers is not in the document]
gnidan: Was this just `typescript`, do you know?

packages/parse-mapping-lookup/package.json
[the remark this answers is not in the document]
gnidan: Issue created: https://github.com/trufflesuite/truffle/issues/3998

packages/parse-mapping-lookup/package.json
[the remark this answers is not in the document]
gnidan: This looks like it's just `typescript`, right? We'll have to remember this if another upgrade happens to get merged before this.

packages/parse-mapping-lookup/src/parser.ts
gnidan: - [ ] Boolean literals
- [ ] `hex"deadbeef"` literals
- [ ] `unicode"..."` literals
- [ ] Solidity units (` gwei`, etc.)

gnidan: Hm so unicode literals are a bit annoying, it seems, because Parjs seems to just assume its input is in the BMP (See their README section [Parsing Unicode](https://github.com/GregRos/parjs#parsing-unicode)).

It should be doable to restrict `""` literals to ASCII, but it might be a bit annoying to write the parser for that?

What do you think of this:
- Exclude `unicode"..."` as unsupported for now (since BMP restriction suggests that we shouldn't make the promise of full-unicode support)
- Don't restrict `""` to ASCII in the meantime, to afford _some way_ of specifying literals that include BMP characters.

Alternatively, if you want to be a sticker about `""` meaning ASCII, we could just define our own literal `bmp""` to be more explicit.

Thoughts?

packages/parse-mapping-lookup/src/ast.ts
gnidan: What forms do enum literals take? Is it always `<identifier>.<identifier>`? Or you can do `<identifier>.<identifier>.<identifier>` also, right?

packages/parse-mapping-lookup/src/parser.ts
haltman-at: I don't think we need to actually enforce the ASCII restriction.  Especially as that's relatively recent, older Solidity doesn't make a distinction.

I think it's also OK to exclude `unicode"..."` for now.  Proper Unicode handling will be a pain regardless of how we do it, so maybe it's OK if we just don't support it all properly for now.  (Especially the `\xNN` escape codes...)

packages/parse-mapping-lookup/src/ast.ts
haltman-at: Yup, you can do both!

packages/parse-mapping-lookup/package.json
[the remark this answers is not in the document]
haltman-at: Yup, I think so.

packages/parse-mapping-lookup/src/parser.ts
gnidan: > (Especially the \xNN escape codes...)

I don't want to know

haltman-at: Too bad I'm going to explain anyway! :P  In Solidity, you can do "\xNN" to insert a particular byte.  Note, not a particular *character* -- a particular *byte*, regardless of whether or not this makes valid UTF-8.  However, if it's invalid UTF-8, then you won't be able to use it as a `string`, only as a `bytes` or `bytesN` (since string literals can be used as either).  So, e.g., `"unicode\xA1"` is not the same as `unicode"\u00A1"` or `unicode"¡"`, but rather is an error if used as a `string` rather than a `bytes`/`bytesN`.

packages/parse-mapping-lookup/src/ast.ts
gnidan: (Ended up implementing this one tonight also.)

packages/parse-mapping-lookup/src/parser.ts
gnidan: Ha! So does Solidity handle this and error if you try to assign such a literal to a `string`?

> Too bad I'm going to explain anyway! 

Good cause I was lying.

haltman-at: The Solidity compiler detects it upfront.  It *is* possible to put bad UTF-8 inside a string at runtime; there are no runtime checks on whether your UTF-8 is good.  But attempting to do so directly from a literal (whether `unicode` or `hex` or whatever) will trip the compile-time check and cause a compile error.  If you really want to put bad UTF-8 inside a `string`, you'll need to be sneakier about it.  (Putting it in a `bytes` and the converting that `bytes` to a `string` will do the trick. :) )

PR #3945, opened 2021-03-23
Repair geth CI jobs

eggplantzzz:


packages/truffle/test/sources/migrations/error/contracts/Loops.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.5.0;
 contract Loops {
   uint public id;
   constructor() public {
-    for(uint i = 0; i < 10000; i++){
+    for(uint i = 0; i < 10000000; i++){
```
gnidan, line 7: Why'd this get bumped up?

gnidan (approved):
Swiggity-sweet

packages/truffle/test/sources/migrations/error/contracts/Loops.sol
eggplantzzz: Previously `--targetgaslimit` was set to something like 7000000. I think the default for Geth is 8000000. So it wasn't running out of gas when I removed the option.

PR #3947, opened 2021-03-24
Internal improvement: Cache chain ID to avoid extra network requests

haltman-at:
This PR is a follow-on to #3923.  That PR added an extra network request with each transaction sent, and I wanted to fix that.

So, this PR adds a chain ID field to contract constructor objects; I called it `_chainId` because I was slightly worried about naming conflicts, but we can change it if that's not what you think it should be named.

Anyway, now when we send transactions, we check that field, and use the chain ID from there if possible; if not we fetch the chain ID from the network and then store it there.

Note, the way I wanted to do this actually would be to get it once in, say, `setProvider`, but we can't do that because it's not `async`.  So, I did the caching where it's used instead.  However I did have `setProvider` (or more generally, `configureNetwork`) set `this._chainId` to `undefined` so we don't have this stale cache mucking things up should the network change.

### DIFF
--- a/packages/contract/lib/contract/constructorMethods.js
+++ b/packages/contract/lib/contract/constructorMethods.js
@@ -29,6 +29,9 @@ module.exports = Contract => ({
     // save properties
     this.currentProvider = provider;
     this.networkType = networkType;
+
+    //invalidate cached chain ID
+    this._chainId = undefined;
   },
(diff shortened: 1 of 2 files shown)


gnidan (approved):
Hm seems reasonable!

PR #3948, opened 2021-03-25
Pass config file value to console if it exists

fainashalts:
This PR passes along the config to `truffle console` when the `--config` flag is used. This is useful for using the console with a custom config file.

### DIFF
--- a/packages/core/lib/console-child.js
+++ b/packages/core/lib/console-child.js
@@ -9,7 +9,11 @@ const inputStrings = input[1];
 
 //detect config so we can get the provider and resolver without having to serialize
 //and deserialize them
-const detectedConfig = Config.detect({network: yargs(input[0]).argv.network});
+const args = yargs(input[0]).argv;
+const detectedConfig = Config.detect({
+  network: args.network,
+  config: args.config
+});
 const customConfig = detectedConfig.networks.develop || {};
 
 //need host and port for provider url
@@ -25,7 +29,7 @@ detectedConfig.networks.develop = {
   port: customConfig.port || 9545,
   network_id: customConfig.network_id || 5777,
   provider: function () {
-    return new Web3.providers.HttpProvider(url, {keepAlive: false});
+    return new Web3.providers.HttpProvider(url, { keepAlive: false });
   }
 };
 
--- a/packages/core/lib/console-child.js
+++ b/packages/core/lib/console-child.js
@@ -9,11 +9,8 @@ const inputStrings = input[1];
 
 //detect config so we can get the provider and resolver without having to serialize
 //and deserialize them
-const args = yargs(input[0]).argv;
-const detectedConfig = Config.detect({
-  network: args.network,
-  config: args.config
-});
+const { network, config } = yargs(input[0]).argv;
+const detectedConfig = Config.detect({ network, config });
 const customConfig = detectedConfig.networks.develop || {};
 
 //need host and port for provider url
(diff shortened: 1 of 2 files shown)


packages/core/lib/console-child.js
```diff
@@ -9,7 +9,11 @@ const inputStrings = input[1];
 
 //detect config so we can get the provider and resolver without having to serialize
 //and deserialize them
-const detectedConfig = Config.detect({network: yargs(input[0]).argv.network});
+const args = yargs(input[0]).argv;
+const detectedConfig = Config.detect({
+  network: args.network,
+  config: args.config
+});
```
gnidan, line 16:
```suggestion
const { network, config } = yargs(input[0]).argv;
const detectedConfig = Config.detect({ network, config });
```

packages/core/lib/console.js
```diff
@@ -161,18 +161,29 @@ class Console extends EventEmitter {
...
+      spawnInput =
+        "--config " +
+        options.config +
+        " --network " +
+        options.network +
+        " -- " +
+        inputStrings;
+    } else {
+      spawnInput = "--network " + options.network + " -- " + inputStrings;
+    }
```
gnidan, line 175: What about cleaning this up to be something like:

```suggestion
    const settings = ["config", "network"]
      .filter(setting => options[setting])
      .map(setting => `--${setting}=${options[setting]}`);
    const spawnInput = `${settings} -- ${inputStrings}`;
```

packages/core/lib/console.js
```diff
@@ -161,18 +161,29 @@ class Console extends EventEmitter {
     // errors/warnings in child process - specifically the error re: having
     // multiple config files
     const spawnOptions = { stdio: ["inherit", "inherit", "pipe"] };
+    let spawnInput;
+    if (options.config) {
```
gnidan, line 165: Do different kinds of falsiness matter here? What happens if you do `"config" in options` instead?

Just want to make sure we're covering our bases.

fainashalts: Using `config` without the hashes doesn't do anything! So I think we're good unless you have any other scenarios in mind?

packages/core/lib/console.js
fainashalts: This won't work, the spawnInput is incorrect. But I'd love to clean it up this way, will push a change in a couple of hours!

gnidan: Oh oops, forgot a join! Edited.

packages/core/lib/console.js
gnidan: Nah should be alright? Thanks!

packages/core/lib/console.js
fainashalts: Oops! Sorry I just pushed this without realizing you'd done it here! Same solution, adding the join. :) I also removed the `=` because it didn't seem necessary.

gnidan: It's not necessary, but using `=` seems to me like best practice; whenever writing examples, I always include it (there's no risk of misinterpreting what's a key/value pair vs. a boolean flag + separate positional arg)

Not worth changing now, though.

gnidan (approved):

0racl3z (approved):

0racl3z (approved):

0racl3z (approved):

0racl3z (commented):
thx!

PR #3954, opened 2021-03-30
Bug fix: Downgrade @ethereumjs/tx

eggplantzzz:
It looks like upgrading from `ethereumjs-tx` without extra configuration still left in some "Invalid sender"/EIP-155-related  problems. This PR downgrades this library until we want to spend time more carefully looking at this upgrade.

### DIFF
--- a/packages/hdwallet-provider/package.json
+++ b/packages/hdwallet-provider/package.json
@@ -17,11 +17,11 @@
   },
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@ethereumjs/tx": "^3.0.2",
     "@trufflesuite/web3-provider-engine": "15.0.13-1",
     "any-promise": "^1.3.0",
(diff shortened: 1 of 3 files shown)


haltman-at (approved):
Seems straightforward enough!  Thanks a bunch Tyler!

bguiz (approved):
Yes please, do include this in the next release, switching from hdwallet-provider 1.2.3 to 1.2.5 introduced the following error (network is RSK):
```json
{"jsonrpc":"2.0","id":6704181714997867,"error":{"code":-32602,"message":"Missing parameter, gasPrice, gas or value"}}
```

... and reversing this bit is probably going to help:
https://github.com/trufflesuite/truffle/compare/v5.2.4...v5.2.6#diff-6e87f121eed734c9794dcdf3e6d69c1e39561f75b43f07babffd5862e25eeeb1

PR #3966, opened 2021-04-06
Add initial `truffle db query <query>` command

gnidan:
Screen Shot 2021-04-06 at 2 09 05 PM

Note: this is pretty barebones, e.g. this does not yet support requests with variables (should it?)

### DIFF
--- a/packages/core/lib/commands/db/commands/query.js
+++ b/packages/core/lib/commands/db/commands/query.js
@@ -1,0 +1,37 @@
+const command = {
+  command: "query",
+  description: "Query @truffle/db",
+  builder: {},
+  help: {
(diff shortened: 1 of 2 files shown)


fainashalts (commented):
So I think this is good, and I am excited about it! But I would like variables to be supported from the get-go unless it is too much work. Would make it more robust and usable right away I think. Giving my approval still as this could definitely be done later if that's better!

gnidan:
Probably won't get the chance to add variables this week, and I know @JoranHonig is looking for this feature, so I'll merge this now and open an issue for adding support for variables later.

Thanks @fainashalts!

PR #3967, opened 2021-04-06
Fix problem when adding non-null resources alongside null resources in @truffle/db

gnidan:


packages/db/src/meta/pouch/adapters/base.ts
```diff
@@ -211,30 +211,31 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
     );
 
     const resourceInputById = input[collectionName]
-      .filter((_, index) => resourceInputIds[index])
       .map((resourceInput, index) => ({
```
gnidan, line 214: The filter meant that the `index` passed to the function passed to `map` no longer lined up properly.

The solution is to move the filter to later.

fainashalts (approved):
LGTM!

gnidan:
Thanks!

Issue #3957, opened 2021-03-31
Cannot migrate to custom blockchain since truffle@5.2.5


thecodingshrimp:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

When trying to migrate via truffle to a custom blockchain (using openethereum) I get an error saying: 
```
Error:  *** Deployment Failed ***

"Migrations" received: Returned error: Invalid params: unknown field `chainId`, expected one of `type`, `from`, `to`, `gasPrice`, `gas`, `value`, `data`, `nonce`, `condition`, `accessList`..
   * This error is common when Infura is under heavy network load.
   * Try: setting the 'confirmations' key in your network config
          to wait for several block confirmations between each deployment.

    at /home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/deployer/src/deployment.js:365:1
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at Migration._deploy (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/migrate/Migration.js:74:1)
    at Migration._load (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/migrate/Migration.js:61:1)
    at Migration.run (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/migrate/Migration.js:212:1)
    at Object.runMigrations (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/migrate/index.js:150:1)
    at Object.runFrom (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/migrate/index.js:110:1)
    at Object.run (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/migrate/index.js:87:1)
    at runMigrations (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/core/lib/commands/migrate.js:263:1)
    at Object.run (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/core/lib/commands/migrate.js:228:1)
    at Command.run (/home/leo/Desktop/test_folder/test_folder/node_modules/truffle/build/webpack:/packages/core/lib/command.js:136:1)
```

[...]

PR #3969, opened 2021-04-07
Bug fix: Update hdwallet-provider to be EIP-155 compliant on main chains

eggplantzzz:
So currently `chainId` is incorrectly passed to `web3.eth.sendTransaction` (as a functional hack). As far as EIP-155 compliance goes, this work should really be done in @truffle/hdwallet-provider as the provider is the thing responsible for signing transactions.

This PR removes the `chainId` argument from this call (which is causing problems with OpenEthereum as you can see in [this issue](https://github.com/trufflesuite/truffle/issues/3957) and adds `chainId` as a constructor argument for hdwallet-provider. In the process, ethereumjs-tx has been upgraded. Eventually this should be upgraded to @ethereumjs/tx, but that will have to wait until Truffle stops supporting Node 10 as it requires Node >12.

We will still have a problem with people using hdwallet-provider with  chains with custom chain ids when also needing to be EIP-155 compliant as ethereumjs-tx cannot autodetect the settings for these chains and will require a significantly greater amount of setup. For now this should be ok; perhaps in the future we can do some work to also allow users to hook into this custom setup and make custom chains functional for the use described above.


haltman-at:
> As far as EIP-155 compliance goes, this work should really be done in @truffle/hdwallet-provider as the provider is the thing responsible for signing transactions.

This isn't always true, though?  Like, yes, you *can* do an `eth_sendTransaction` to a provider, and have it sign it and translate it into an `eth_sendRawTransaction` to the ultimate provider; but you can also just do an `eth_sendRawTransaction` yourself.  In fact, Web3's `web3.eth.sendTransaction` method will check whether the sending account appears in `web3.eth.accounts.wallet`; and if so it will sign it itself, and do an `eth_sendRawTransaction`, rather than doing an `eth_sendTransaction`.  I don't think we should assume that everyone is using a translation layer such as `hdwallet-provider` as their provider, and that nobody is using a direct provider?  (Man, I really do not like the provider abstraction, it really is doing too much...)

I think that inserting the parameter at the origin would -- were it not causing the problems -- be the correct thing.  Since it *is* causing problems, this PR might be the appropriate solution.  But I'd consider this -- adding it in translation -- to be more of a hack than just adding it at the origin!  (Of course, if it weren't causing problems, the best thing to do might be to add it to *both*, but I guess we can't do that.)

I mean, really, going by the web3 documentation, we ought to be using the `common` (or `defaultCommon`) properties to set this, rather than sticking in `chainId` directly, but I gather this was tried and didn't work for some reason?  And I imagine it would cause the same problems with OpenEthereum regardless. :-/

Anyway sorry I guess that's basically a rant that has no real bearing on whether the actual PR should be merged or not rather than an actual review... uh, I'll do that later I guess...

packages/contract/lib/execute.js
```diff
@@ -525,20 +525,6 @@ const execute = {
   //(well, OK, it's lacking some things there too, but again, good enough
   //for our purposes)
   sendTransaction: async function (web3, params, promiEvent, context) {
-    // get the chainId to be compliant with EIP-155
-    // Geth wants this value in base 16 prefixed by "0x"
-    let chainId;
-    if (context.contract._chainId) {
-      chainId = context.contract._chainId;
```
haltman-at, line 532: If you're removing this, you should also remove the line I added in `constructorMethods.js` to invalidate the cache.  (See #3947.)

packages/hdwallet-provider/src/index.ts
```diff
@@ -47,6 +48,7 @@ class HDWalletProvider {
       shareNonce = true,
       derivationPath = `m/44'/60'/0'/0/`,
       pollingInterval = 4000,
+      chainId,
```
haltman-at, line 51: So this requires the `chainId` to be passed to the constructor?  Will that require any additional work on the part of the user or do we have things set up so that should happen automatically?

haltman-at (commented):
Mostly seems to make sense?  Two questions/notes.

(I'm still kind of grumbly that we have to remove it from contract because of OpenEthereum, but meanwhile you're just doing what you have to in removing it...)

packages/hdwallet-provider/src/index.ts
eggplantzzz: They will need to enter it into the config. If they don't, they will just sign transactions in an EIP-155 non-compliant manner. This won't matter when they are using something like Ganache with unlocked accounts because it doesn't check the signature.

haltman-at: Oh, that's kind of annoying.  Shouldn't HDWalletProvider be able to determine it itself, rather than requiring some new extra step for people to do?  I don't know, maybe we're OK with this... @gnidan, you have an opinion?

eggplantzzz: I guess it is plausible that we could include an http lib in hdwallet-provider and make the call for the id manually when the `signTransaction` method gets executed? I'm not sure about that...

haltman-at: Wait, I'm confused -- HDWalletProvider wraps another provider, right?  That it forwards requests to, and sometimes translates?  E.g., it translates `eth_sendTransaction` into `eth_sendRawTransaction`?  So it can already interact with that provider, and do an `eth_chainId`?  Why would you need to add a new library?

eggplantzzz: This is a bit confusing to me because the `signMessage` method is defined in the constructor. I think the provider gets added later and `signMessage` won't have access to the context it seems. I'll have to do a little more research and see if there is a way we could possibly rearrange it or if we even might somehow have access.

haltman-at: Well, I think `signMessage` is supposed to just be for *signing* the message.  Presumably you'd add `chainId` somewhere outside of it.  Remember, adding the chain ID and signing are two separate things!

eggplantzzz: Yes of course, but the problem is I don't know where we can fetch the chain id because we can't do it in the constructor.

haltman-at: How would that happen?  And is that even a real problem?  Like, again, we can think of adding the chain ID as something entirely separate from signing.  Adding the chain ID is something we need to do if we receive an RPC request that requires signing, right, such as `eth_signTransaction` or `eth_sendTransaction`, right?  If HDWalletProvider receives one of those, we can add the chain ID, and then sign.  If someone truly wants to manually invoke signing, rather than sending an `eth_signTransaction` or `eth_sendTransaction`, well, maybe it's OK if the chain ID isn't added?

eggplantzzz: But where will we add the fetching? HDWalletProvider is instantiated and then passed to web3. Then web3 uses it to sign when it needs to. There is no room in between the constructor and when `signMessage` gets called to add the chain id. Hypothetically we could check the provider for some kind of custom method to fetch the chain id in Truffle (like if it has the `init` method or something like that) but that feels pretty janky.

haltman-at: OK, I went and looked at the code some more and I think I see now what I was misunderstanding.

I thought the room to add this extra logic was, y'know, in the request-response loop.  What I did not understand was that we do not manually provide that loop, but rather rely on provider-engine for it.  Since we don't actually control the request-response loop per se, it looks to me like you're right; there's nowhere to add it.

So, I guess this is OK.  It's kind of disappointing that we're relying on the user to provide this, but, oh well, so it is.  I don't know whether @gnidan has any opinions on this, but I don't see any way to avoid this for now.

Hopefully eventually OpenEthereum fixes their stuff and we can put this back in `@truffle/contract`, so we don't need to rely on user input for it.

haltman-at: Btw, since we're adding a new field to the network config, don't forget to update the docs on trufflesuite.com!

haltman-at (approved):
OK, I guess this is fine. It's kind of disappointing that we're relying on the user to provide this, but, oh well, so it is; I don't see any way to avoid it at the moment. I don't know whether @gnidan has any opinions on this, but like I said, I don't see any way to avoid this for now.

Hopefully eventually OpenEthereum fixes their stuff and we can put this back in @truffle/contract, so we don't need to rely on user input for it.

(Btw, since this is effectively adding a new field to the network config, don't forget to update the config reference docs on trufflesuite.com!)

PR #3970, opened 2021-04-07
Stop populating network for undeployed artifacts in @truffle/db

gnidan:
- Extract `populateNetworks` helper method
- Short-circuit when address does not exist


fainashalts (approved):
Nice! Thanks for reasoning through this with me today!

gnidan:
> Thanks for reasoning through this with me today!

Thanks for finding the bug :)

Issue #3951, opened 2021-03-29
"Error: Could not find artifacts for" when deploying contracts from node_modules


haydenyoung:
When compiling and deploying a contract from a node modules source, I'm seeing the following:
```
        throw new Error("Could not find artifacts for " + import_path + " from any sources");
        ^

Error: Could not find artifacts for /@uniswap/v2-core/UniswapV2Pair from any sources
```
which seems to be exactly related to https://github.com/trufflesuite/truffle/issues/1791.

I'm running v5.2.3.

Could it have been reintroduced?

Thanks

eggplantzzz:
Thanks for opening this, could you provide us with some code and steps to recreate this error?

haydenyoung:
thanks for the response.

To replicate:

Set up a simple project:

```
mkdir TestProject
cd TestProject
truffle init
npm init
```
Install the uniswap core:

```
npm i --save @uniswap/v2-core
```
Create a migration script, something like 2_testproject.js.

In 2_testproject.js, add:

```
const UniswapV2Pair = artifacts.require("@uniswap/v2-core/UniswapV2Pair");

module.exports = function (deployer) {

};
```
With a local blockchain running (e.g. ganache) and the development config uncommented in truffle-config.js, run:

```
truffle migrate
```

This should result in:

```
        throw new Error("Could not find artifacts for " + import_path + " from any sources");
```

It may be I have the require referenced incorrectly.

Let me know if you need any more information.

eggplantzzz:
I see what the problem is, right now Truffle expects npm packages to be Truffle projects while that uniswap library has a different structure. Truffle can't seem to locate the artifact because of this difference in structure. As a workaround for now you could just copy the artifact file (the JSON file from `./node_modules/build` manually into your project's `build/contracts` directory and Truffle should find it.

Perhaps we should consider making Truffle look in the `build` directory for npm packages? Seems like it could be a common place that artifacts are kept.

eggplantzzz:
I put in [this PR](https://github.com/trufflesuite/truffle/pull/3973) to check the `build` folder of packages installed by npm.

haydenyoung:
Thanks @eggplantzzz looks great. Yeah at the moment I am using the "copy over to my build/contracts" fix as you stated and that is working for me.

Thanks again.

PR #3973, opened 2021-04-13
Enhancement: Look inside "build" folder when resolving artifacts

eggplantzzz:
When checking npm packages for artifacts, Truffle currently expects that package to have the structure of a Truffle project (i.e. it looks in `build/contracts` for artifacts). This PR also checks the `build` folder in those packages. This was prompted by https://github.com/trufflesuite/truffle/issues/3951 where a Uniswap package had artifacts located in a `build` directory.

### DIFF
--- a/packages/resolver/lib/sources/globalnpm.ts
+++ b/packages/resolver/lib/sources/globalnpm.ts
@@ -1,7 +1,7 @@
 import path from "path";
 import fs from "fs";
 const detectInstalled: any = require("detect-installed");
-const get_installed_path: any = require("get-installed-path");
+const getInstalledPath: any = require("get-installed-path");
 
 import { ResolverSource } from "../source";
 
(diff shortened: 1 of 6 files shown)


haltman-at (commented):
Makes sense.  One question: Should this apply to GlobalNPM as well?

eggplantzzz:
I think that follows!

packages/resolver/lib/sources/npm.ts
```diff
@@ -68,6 +60,25 @@ export class NPM implements ResolverSource {
     return { body, filePath: import_path };
   }
 
+  resolveAndParse(basePath: string, packageName: string, contractName: string) {
```
eggplantzzz, line 63: Do you think this should be a utility method outside of this class since it is really not standard in the same way `require` and `resolve` are?

haltman-at (approved):

packages/resolver/lib/sources/npm.ts
haltman-at: I think it's OK as it is -- yeah I agree it's not great, but it's not like this is going to spread any further, at least for now.  Rule of three, not rule of two, right? :)  IDK, @gnidan may have stronger opinions on this. :P

packages/resolver/lib/sources/globalnpm.ts
```diff
@@ -10,41 +10,58 @@ export class GlobalNPM implements ResolverSource {
...
+    const subDirs = [`build${path.sep}contracts`, "build"];
+    for (const subDir of subDirs) {
+      const possiblePath = path.join(
+        packagePath,
+        subDir,
+        `${contractName}.json`
       );
       try {
-        const result = fs.readFileSync(expected_path, "utf8");
+        const result = fs.readFileSync(possiblePath, "utf8");
```
cds-amal, line 45: Is the intention to determine if the possible path exists and then parse it? It seems weird to use the exception handler to continue the loop.

Also, what happens if there are multiple paths exists? That is a consumable file in build, as well as contracts folder?

eggplantzzz: Basically the npm and global npm resolvers have two locations that they search in the package: `/build/contracts` and `/build`. It will look in those locations in that order and use the first artifact it finds.

I would kind of agree with you about using the error handling but this is in fact what Node recommends. Check out the docs for [`stat`](https://nodejs.org/api/fs.html#fs_fs_stat_path_options_callback).

cds-amal: TIL node's weird try/catch recommendation. Thanks @eggplantzzz! Is there's documentation to explain the path ordering/precedence?

eggplantzzz: Are you talking about documentation about the resolver? Absolutely not :)

PR #3974, opened 2021-04-15
Enhancement: Create @truffle/encoder and use it in @truffle/decoder for mapping keys

haltman-at:
**This PR contains breaking changes to `@truffle/decoder`.**  (Also, @bergarces, I don't expect you can review this, but you may want to be aware of it.)

This PR creates the `@truffle/encoder` package, as well as the functionality in `@truffle/codec` to support it.  The main purpose of this package is primarily to be used in the `@truffle/contract` rewrite, so this is mostly an internal improvement; however, it does also make use of `@truffle/encoder` inside `@truffle/decoder` for handling user input of mapping keys (to the `watchMappingKey` function and similar).  As such I've removed the warning from such functions that they do not check user input and should not be directly exposed, because now they (mostly) do!

(Note that one of the functions of `@truffle/encoder` is ENS resolution, and I've made it so that `@truffle/decoder` will, by default, have ENS enabled; I'm not sure if we want to change this, but I thought I should point it out.  Note that this is also laying some groundwork for putting reverse-ENS resolution in the decoder in the future.)

Despite its name, the new functionality in `@truffle/encoder` is primarily not related to encoding, but rather providing a usable interface on top of existing encoding functionality.  All of the encoding functionality already existed in `@truffle/codec`; the trouble is, they required input in the form of the decoder output format.  So the primary function of `@truffle/encoder` is to recognize user input and attempt to translate it into a `Format.Values.Value`.  (Not a `Format.Values.ErrorResult`!  If the input is bad we throw an exception!)  It can then also performs encoding, of course.

Note that in addition to recognizing user input for Solidity values, it also has to recognize user input for transaction options.  As such, the output of the encoder does not only include encoded data, but also other transaction options.  This does include the Truffle-specific transaction option `overwrite`, and the Quorum-specific transaction option `privateFor` (that last one gets some special handling, as I'll discuss later).  As part of this, I've added `Format.Types.OptionsType` and `Format.Values.OptionsValue` (and etc); these do not reflect an actual decoded Solidity value, but rather wrapped user input for transaction options.

In addition to the addition of `@truffle/encoder`, the `@truffle/decoder` package has had its interface reworked in order to make the two more parallel.  They're not totally parallel, but they're as parallel as I could reasonably manage.  Some of the functionality in `@truffle/decoder` has been moved into `@truffle/codec` so that `@truffle/encoder` can use it as well.  Note that @gnidan and I decided to remove the Decoder functionality to be initialized with a `config` object.

So, there's a lot to discuss.

## The core wrapping functionality

The core wrapping functionality lives in the new `codec/lib/wrap` folder.  It didn't really fit anywhere else.  The main interface, which covers overloard resolution as well as just wrapping, lives in `index.ts`; the actual core wrapping functionality lives in `wrap.ts` for non-elementary types and `elementary.ts` for elementary types.  There's also `priority.ts`, which is used for resolving overloads when more than one overload matches.

There's some stuff in `wrap.ts` and `elementary.ts` that probably really belongs in `utils.ts` or `errors.ts` (like, all the regexes and error messages).   Oh well, we can change that if need be.  (Also, it's possible that some of the stuff in `Codec.Conversion` really belongs here, or vice versa?)

In order to support ENS resolution, the wrapping functionality is generator-based; it can yield requests for more information, such as a request to resolve an ENS name.  However, that's not all it's used for.  See, we want to recognize user input given in the form of a big number class from any of several common big number libraries.  However, we don't want to add any new big number libraries to `@truffle/codec` itself -- so the only big number libraries that are recognized by `codec/lib/wrap` itself are `bn.js` and `big.js`.  (It does recognize BigInts, of course, but that's not a library. :) )  For recognizing others -- `bignumber.js`, Ethers's `BigNumber` class (which is not the same), and Ethers's `FixedNumber` class -- it farms out the task to `@truffle/encoder` using the request system.  It reports back to `codec/lib/wrap` whether the input is a valid integer or decimal, if so its value (as a `bigint` or `Big`, respectively), and if not why not.

(This system is also used for processing ICAP addresses. :P )

Note that because of the heterogeneity of possible types of responses, `codec/lib/wrap` uses typed responses, unlike the decoding functionality which just uses plain `Uint8Array`, with no wrapper, as the response type.

### Wrapping and encoding

OK, so how does wrapping and encoding work?  Well, you pass in a type to wrap as, an input to wrap as that type, and it tries to wrap it.  If it can't, it throws a `TypeMismatchError`.  (Note: This error is not further subtyped; I relied on messages for providing more specific information about what went wrong.)  I'm mostly not going to go over the details here; rather, you can see the documentation if you want a list of all possible forms of input.  I am going to note some particular ones that are worth discussing, though:

1. For each type, you can pass an object of the form `{ type: ..., value: ... }`, where `type` is a string corresponding to the type you want to wrap as.  The purpose of doing this isn't really for straight wrapping but rather for overload resolution; you do this so that the wrong type *won't* accept.  However, this system does have some warts and oddities.  For instance, for array types, you just use `type: "array"`, not `type: uint8[]` or whatever.  So you're limited in what you can specify.  If you tag an actual element, you can then, say, distinguish `uint8[]` from `bytes1[]`, but not `uint8[]` from `uint8[2]`.  (And if there are no elements to tag, then, well, you can't do that.)

This system has some further warts when it comes to structs, because, um, hopefully you don't have a struct whose fields are `type` and `value`, with `type` being a string, because then you may have some trouble passing it as an object... there are ways around this, but it's worth noting that this edge case exists.  (You might similarly run into problems if your struct somehow resembles a `Format.Values.Value`...)

Also, you can use the type aliases `"uint"` for `"uint256"`, `"int"` for `"int256"`, `"byte"` for `"bytes1"`, `"fixed"` for `"fixed128x18"`, and `"ufixed"` for `"ufixed128x18"`.  However I did not add type aliases for Vyper types, so you can't e.g. use `"decimal"` for `"fixed168x10"`.  I could add that though.

2. To maintain compatibility with ethers, *anything* is accepted as a boolean... except not literally anything, because, say, `{ type: "uint", value: 1 }` will fail.  Also, the string `"false"` (with any capitalization) will also be encoded as `false` rather than `true`.  (As will `new Boolean(false)`, and various similar stuff.)

3. You can pass in a `Format.Values.Value`, and, well, it'll work.  In a few very particular contexts, you can also pass in a `Format.Values.ErrorResult`, but normally those will be rejected.  (Specifically: You can use an enum out-of-range error as if it were a number; and certain boolean errors can be used for `true`.)

4. There's no way to pass in a string with bad UTF-8.  Well, there's one way: You can pass it in as a `Format.Values.Value`.  But otherwise you can only pass in strings with valid UTF-8.

5. For decimal types, input with more precision than the allowed number of decimal places will be rejected.  Also, input in the form of a `number` will be rejected if it's large enough that it represents possible loss of precision for that type... yes, this means that `1` is not valid input to something expecting a `fixed128x18`!  Users are encouraged to use safer forms of input.  Also, even though I allowed units on numeric strings for integer types, I didn't allow it for decimals, because it's not entirely clear what it would mean.  (Should `"1 gwei"` *really* be interpreted as `1e9` when the type is `fixed128x18`?  Wouldn't it perhaps be more appropriate to treat it as `1e-9` in that case?  To head off these sorts of questions, I decided the best thing to do is just not support it.)

6. If the encoder knows that a given unsigned integer type is actually supposed to be an enum, it will reject inputs that are out of range for that enum.  Obviously, however, it will accept them if it does not know this.

7. Some forms of input I thought about accepting but didn't add support for, but which notionally we could: (note that some of these are maybe obviously bad ideas, but I'm listing them here anyway)
* `Uint8Array` (&similar) for strings (would allow invalid UTF-8)
* More units for integer types
* More type aliases, as mentioned above, or the array thing (that turned out to be troublesome to implement unfortunately)
* Base-64-encoded strings for `bytes` and `bytesN` (I don't like this idea for multiple reasons, but I did think of it so I'll list it)
* The `decimal.js` big number library (I think this is an obvious no :P )
* Allowing underscores in numeric strings (whether decimal or hex) since Solidity does that (kind of leaning towards putting that in if we want to use it with your parser)

### Overload resolution

This is where things start to really get messy.  We need to resolve overloads -- and not just by length.

Now, if we can resolve overload just by length, we do that, and in that case we just defer to the unique function matching that length, because that way, if it fails, we'll get a `TypeMismatchError`, which is fairly informative, as opposed to a pretty uninformative `NoOverloadsMatchedError`.  But sometimes there's more than one function matching that length, and in that case we have to invoke the actual overload resolution system.  (Again, we try to avoid invoking this because we want better error messaging, but sometimes it can't be avoided.)

Now if only one function matches the input, great, we're done.  And if none do, obviously we throw a `NoOverloadsMatchedError`.  If multiple do though, we're going to decide between them on the basis of which has input types that we consider to be "more specific".  Specificity is only a partial ordering, so there might not be a unique best overload, in which case we throw a `NoUniqueBestOverloadError`.

Specificity is only a partial order when comparing types that are fairly similar though; for totally different types, it's a total order.  "More specific" doesn't mean that one type is more specific than another in terms of its *values*, but rather in terms of what input it will accept.  Thus, `bytes` is considered more specific than `string`, because if you input `"0xdeadbeef"`, you probably wanted `bytes`, not `string`.

Naturally the whole specificity system is... not always guaranteed to give you what you want, but, well, there's no way to guarantee that, so yeah.  It's better than what we currently have, which is that overload resolution only really works by length, and apparently sometimes not even then.  If you want to guarantee you get the right overload, you can always use `{ type: ..., value: ...}` input, or you can just pick out the method you want by signature or selector rather than by name, you know?

I expect the more questionable parts of the specificity system won't come up that much, but, well, we'll see I guess.  (One annoying aspect of the specificity system is that it doesn't allow for rearranging specificities based on the particular input; doing that is hard, unfortunately, unless we want to switch to a fitness-score system, but I don't want to do that.)

Note that the `wrap` function in `codec/lib/wrap` takes a `loose` flag as one of its options -- this turned on when overload resolution is *not* being used.  But for overload resolution, it's turned off, which means that, among other things, extra fields won't be allowed in struct input.  Will this break anything?  Well, as mentioned, overload resolution never really worked properly in the first place, so arguably it only breaks things that were never properly supported in the first place.

### Dealing with transaction options

OK, but how do transaction options affect the above?

The functions for wrapping transactions, and for resolving overloads, take an optional `allowOptions` flag.  (I decided I didn't want it on by default, and you even have to pass it explicitly in `@truffle/encoder`.  Obviously, when used in `@truffle/contract`, we will always pass this flag.)

This implies that Codec now needs to know about the transaction options we support!  So there's now an `Options` type in Codec.  I guess once we switch over to using `@truffle/encoder`, this would become the source of truth for supported transaction options...

Transaction options can be given in any form accepted for the type of that transaction option -- `gas`, `gasPrice`, `value`, `nonce` take integer input; `from` and `to` take address input; `data` takes bytestring input; `overwrite` takes boolean input; and, um, `privateFor` is a special case.

See, `privateFor` is supposed be given as an array of length-32 bytestrings... but they're not supposed to be given as hex strings, but rather as base-64.  Now as mentioned we could add base-64 as an accepted type of input for bytestrings, but I think that would be a bad idea, so instead this field is just treated as a special case.

Two more options worth singling out (doubling out?) are `to` and `data`.  What's noteworthy about these is that while these are *checked* (e.g., you'll get a rejection if you pass `"0xf"` rather than `"0xf0"` for `data`), their values are ultimately not used.  Well, OK, `to` is used in some cases, but it's often not used.  So, uh, be aware of that.

But, OK, the real messy thing I want talk about is, how does this affect overload resolution?

Well, it definitely makes it messier.  Let's say you have two overloads of length `n` and `n+1`, and you pass `n+1` arguments... do you want to have to use the overload resolution system here?  That would suck!  So instead we have some special handling here so that you can still get your nice `TypeMismatchError` with their detailed error messages.  It's possible some things might slip through the cracks here and fall back on the specificity-based overload resolution system, but I think it should generally work pretty well.

That said, in order to cover those cracks, `options` has been given highest-specificity in the overload resolver.  I don't really like that, but I've done it for compatibility.

## The encoder interface

OK, let's talk about the encoder interface itself.  We can break this down into two parts: Initialization and use.  We'll discuss use first.

### Use

So, we have three classes: The `ProjectEncoder`, the `ContractEncoder`, and the `ContractInstanceEncoder`, parallelling the decoder.  (`WireDecoder` has been renamed to `ProjectDecoder`. :) )

`ProjectEncoder` is mostly useful for `wrap` and `wrapElementaryValue`.  (The latter is just `wrap` restricted to `wrapElementaryValue`, which is useful for type reasons.  It's `ProjectEncoder` that's used inside the decoder, since it just uses `wrapElementaryValue`.)  It can also do encoding and overload resolution, but the interface is kind of awkward, and you have to put in lots of stuff manually; if you really want to do encoding and overload resolution, you probably want one of the others.  (Note that there's no `encodeMappingKey` method, because, well, I didn't see that as useful; the only encoding methods are transaction encoding.)

The `ContractEncoder` is where most of the usable encoding functionality lives.  There's `encodeTransaction` and `resolveAndEncode`.  The `encodeTransaction` function takes the ABI for the function (or constructor) to be encoded (although it really just uses this to locate the allocation it set up earlier, rather than encoding directly based on it!), an array of the inputs, an an object with flags you want to pass (which right now is just `allowOptions`).  It returns an object with transaction options, with the encoded data in the `data` field.  The `resolveAndEncode` function works similarly, except that it takes an array of ABIs (which must be function ABIs, no constructors here), and it returns an object `{ abi: ..., tx: ... }`, where `abi` is the ABI for the overload that got picked and `tx` is the transaction options (including `data`) as above.  Note integers in these are given as BNs.

The wrapping-but-not-encoding analogue of the above two are also exposed, since they might be useful.  Also, note that trying to encode a constructor will throw an exception if there are unlinked libraries.

The `ContractInstanceEncoder` is basically the same; it doesn't really add much.  Its only real addition is that the `to` field in the transaction options will be automatically populated with the contract instance's address.  So yeah.

There is one big caveat on using the above: It's all async!  It's all async because of the possibility of ENS resolution.  For a while I had sync versions too that didn't do that, but, oy, the result was just too confusing to keep track of.  So, it's all async!

### Initialization (and the decoder interface changes)

Initialization works a lot like `@truffle/decoder`, but how `@truffle/decoder` is initialized has also been changed to make the two more parallel.

I'm going to skip the more granular changes, and just talk about the major ones:

1. `provider` and `projectInfo` are no longer separate arguments, but rather options on a single argument; and passing a plain array of artifacts instead of a `ProjectInfo` is no longer allowed

This is a little inconvenient, but, oh well.

2. `{ config }` is no longer a valid `projectInfo`

It was causing too many problems.

3. `{ commonCompilations }` is now a valid `projectInfo`

Compilations in `@truffle/compile-common` form.  Ideally it would just be called `compilations`, but, well, that name was already taken.

4. `encoder` can take `networkId` instead of `provider` (since that's all that's needed for linking)

5. There's now the `ens` field for passing ENS information!

You can pass `provider` (which can be `null` to disable EN) and `registryAddress` (it will use the default one if this isn't passed).

Note: I decided to make ENS on by default if a non-ENS provider is given, so you have to explicitly disable ENS with `{ ens: { provider: null } }`.  **I don't know if this is a good idea, and we might want to change this.**  But right now it's on by default in the encoder if you pass a provider for any reason; and it's on by the default in the decoder in general (because the decoder *always* needs a provider!).

#### Moving things into `@truffle/codec`

Note that a bunch of the decoder initialization code, and related types, has been moved into `@truffle/codec`.  So yeah.

## Documentation

The documentation should fit in with the existing system for Codec and Decoder and such, but uh I just realized I never actually tested it...?  We should probably make sure that actually works.  (Actually, I did give it something of a test while writing this, and... I'm not sure it works due to `@gnd/typedoc` needing a newer version of Typescript??  Well, we can fix that...)

## Testing

I wrote 547 tests of this thing!  Also there are some fastcheck tests.  That said the tests don't cover everything.  Currently there are no tests of library linking (or throwing errors if not everything is linked); I just tested that manually.  The fast-check tests are pretty skimpy.  And, uh, the manual tests are currently all in one giant file.  Maybe that should be broken up.

Other things worth noting:
* The `big.js` package has been upgraded from version 4 to version 6.

OK there's probably way more I could say, this is a huge PR, but I'm going to hope that's a sufficient summary!  As always I've left questions throughout about stuff I could have done differently...


haltman-at:
Oh, another change to input format that might be worth making is that in some places we only accept `0x`, and not `0X`, and we should likely accept both.

haltman-at:
Oh!  Here's another thing that falls under "forms of input I thought about accepting but didn't do".  This is another one I think I should probably change.

So you'll notice that objects with an `address` field are only accepted as a contract object if they *don't* have a `selector` field.  I'm uneasy with this -- what if someone makes a contract and for some reason names one of their methods `selector`?  Then it would fail to work as an address.

I think the right way to do this would be to get rid of that requirement and just move `function` to be higher-priority than `address`, so that if an object with a selector field *can* be interpreted as either, but will prefer to be interpreted as `function` *if* the `selector` field has the correct form (which for a contract instance abstraction it wouldn't, as it wouldn't be a hex string).

But, the problem with that is, if `function` goes higher than `address`, and `address` goes higher than `bytes` (which I do *not* want to change), then that means that `function` goes higher than `bytes` and `bytes24`, which I don't like (you can input a function as a hex string of address + selector, but I think that if someone does that then `bytes` or `bytes24` should be prioritized over `function`).

Maybe we just make that change and live with it because, y'know, overload resolution <shrug>.  I do wish I could have written this in a way where priority could have depended on the specific input, but, well, as you know, that turned out to be a problem, so... yeah.

haltman-at:
Note: We may have to see whether we need to downgrade `@ensdomains/ensjs` to the older package that won't print warnings.

haltman-at:
Or, it might just need the fixes that were made to #4089!

haltman-at:
OK, it seems like I don't need any of the changes that went into #4089, as apparently the only needed changes were in `core`.

haltman-at:
OK, I'm going to close this rather than continue to keep this up to date.  Go see #4142 instead.

PR #3975, opened 2021-04-15
Dependency update: Bump solc to 0.6.0 in contract-schema

eggplantzzz:
So that Truffle uses the same version of Solidity that compile-solidity uses, this PR bumps the version in contract-schema and updates a test.

I hope this makes CI just a little bit faster.

### DIFF
--- a/packages/compile-solidity/package.json
+++ b/packages/compile-solidity/package.json
@@ -28,7 +28,7 @@
     "request-promise": "^4.2.2",
     "require-from-string": "^2.0.2",
     "semver": "^7.3.4",
-    "solc": "^0.6.0"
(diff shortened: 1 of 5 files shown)


haltman-at:
So, that test failure you're getting?  I ran into the same problem on `wrap`.  (Oh boy, this is going to lead to a merge conflict... :-/ )  The reason is that the test of local Solidity is now getting 0.6.0 instead of 0.5.0, so you have to update that test as well.

eggplantzzz:
I figured I would get others :) I'll chase

haltman-at:
Oh hey you got it passing :)

gnidan (approved):

PR #3983, opened 2021-04-19
Allow different blockchain network architectures in @truffle/reporters

bergarces:
This PR moves gas calculations and units from the reporter to the interface-adapter. That way we can bring a new architecture without branching code in the reporter.

In summary:
- Unit print values are no longer hardcoded in the message and have been added as part of the data passed.
- In order to operate with gas values to calculate totals, interface adapter returns BN values, as well as a method to convert the value to the final unit that we want to display. Conversion needs to occur at the end as BN does not support decimals.
- Added a couple of assertions to existing tests to verify that units are being added to the data payload.

### DIFF
--- a/packages/interface-adapter/lib/adapter/types.ts
+++ b/packages/interface-adapter/lib/adapter/types.ts
@@ -1,3 +1,4 @@
+import BN from "bn.js";
 import {
   Block as EvmBlock,
   BlockType as EvmBlockType,
@@ -18,6 +19,18 @@ export type BlockType = EvmBlockType | any;
 export type Transaction = EvmTransaction | any;
 export type TransactionReceipt = EvmTransactionReceipt | any;
 export type TxHash = string;
+export type TransactionReportData = {
+  timestamp: number;
+  from: string;
+  balance: string;
+  gasUnit: string;
+  gasPrice: string;
+  gas: BN;
+  mainUnit: string;
+  value: string;
+  cost: BN;
+  convertCost: (val: BN) => string;
+};
 
 export interface InterfaceAdapter {
   getNetworkId(): Promise<NetworkId>;
@@ -29,4 +42,5 @@ export interface InterfaceAdapter {
   getCode(address: string): Promise<string>;
   getAccounts(): Promise<string[]>;
   estimateGas(transactionConfig: Transaction): Promise<number>;
+  getTransactionReportData(receipt: TransactionReceipt): Promise<TransactionReportData>;
 }
--- a/packages/interface-adapter/lib/adapter/types.ts
+++ b/packages/interface-adapter/lib/adapter/types.ts
@@ -1,4 +1,4 @@
-import BN from "bn.js";
+import type BN from "bn.js";
 import {
   Block as EvmBlock,
   BlockType as EvmBlockType,
--- a/packages/interface-adapter/lib/adapter/types.ts
+++ b/packages/interface-adapter/lib/adapter/types.ts
@@ -19,17 +19,16 @@ export type BlockType = EvmBlockType | any;
 export type Transaction = EvmTransaction | any;
 export type TransactionReceipt = EvmTransactionReceipt | any;
 export type TxHash = string;
-export type TransactionReportData = {
+export type TransactionCostReport = {
   timestamp: number;
   from: string;
   balance: string;
   gasUnit: string;
   gasPrice: string;
   gas: BN;
-  mainUnit: string;
+  valueUnit: string;
   value: string;
   cost: BN;
-  convertCost: (val: BN) => string;
 };
 
 export interface InterfaceAdapter {
@@ -42,5 +41,6 @@ export interface InterfaceAdapter {
   getCode(address: string): Promise<string>;
   getAccounts(): Promise<string[]>;
   estimateGas(transactionConfig: Transaction): Promise<number>;
-  getTransactionReportData(receipt: TransactionReceipt): Promise<TransactionReportData>;
+  getTransactionCostReport(receipt: TransactionReceipt): Promise<TransactionCostReport>;
+  displayCost(value: BN): string;
 }
--- a/packages/interface-adapter/lib/adapter/web3/index.ts
+++ b/packages/interface-adapter/lib/adapter/web3/index.ts
@@ -1,9 +1,12 @@
+import BN from "bn.js";
 import { Web3Shim } from "../../shim";
 import {
   InterfaceAdapter,
   EvmBlockType,
   Provider,
-  Transaction
+  Transaction,
+  TransactionReceipt,
+  TransactionReportData
 } from "../types";
 
 export interface Web3InterfaceAdapterOptions {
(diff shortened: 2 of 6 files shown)


packages/interface-adapter/lib/adapter/web3/index.ts
```diff
@@ -53,4 +56,30 @@ export class Web3InterfaceAdapter implements InterfaceAdapter {
...
+    return {
+      timestamp: block.timestamp,
+      from: tx.from,
+      balance: Web3Shim.utils.fromWei(balance, "ether"),
+      gasPrice: Web3Shim.utils.fromWei(gasPrice, "gwei"),
+      gas,
+      gasUnit: "gwei",
+      value: Web3Shim.utils.fromWei(value, "ether"),
+      cost,
+      totalUnit: "ETH",
```
eggplantzzz, line 81: Should this just be named `unit`? Or maybe `unitType`? Why "total"?

eggplantzzz: Oh I see how it is used, should there be separate unit types sent for "total cost" and "value sent"? Is it likely they will ever be different?

eggplantzzz: Or we could consider something like the following pattern:
```
cost: {
  value: 2444,
  unit: "ether"
}
```

II'm not sure if that is overkill though.

bergarces: I did it like that on the first commit, but then I changed my mind as balance, value and cost were using the same unit (which should be the same for tezos, with TEZ as the main unit and mutez for the fee).

In retrospective, `totalUnit` is not a good name, `mainUnit` or `displayUnit` would probably be better.

bergarces: There's something else I could do similar to what you suggest, but it might be a bit overkill.

Since some of the values I return are already strings (as they don't need to be added up), and some are BNs, I could just always return
```
type BNValue = {
  value: BN,
  displayUnit: string,
  convertMethod: (val: BN) => string
}
```

That way I don't make assumptions of what the reporter needs and which values will need to be added up. It might look a bit cumbersome though.

packages/interface-adapter/lib/adapter/types.ts
```diff
@@ -18,6 +19,18 @@ export type BlockType = EvmBlockType | any;
 export type Transaction = EvmTransaction | any;
 export type TransactionReceipt = EvmTransactionReceipt | any;
 export type TxHash = string;
+export type TransactionReportData = {
```
gnidan, line 22: What if we got rid of `Data`? How's just `TransactionReport`? or maybe `TransactionCostReport` if we wanted to be more explicit that way.

packages/interface-adapter/lib/adapter/types.ts
```diff
@@ -18,6 +19,18 @@ export type BlockType = EvmBlockType | any;
...
+  timestamp: number;
+  from: string;
+  balance: string;
+  gasUnit: string;
+  gasPrice: string;
+  gas: BN;
+  mainUnit: string;
```
gnidan, line 29:
```suggestion
  valueUnit: string;
```

packages/interface-adapter/lib/adapter/types.ts
```diff
@@ -18,6 +19,18 @@ export type BlockType = EvmBlockType | any;
...
+  value: string;
+  cost: BN;
+  convertCost: (val: BN) => string;
```
gnidan, line 32: Naming-wise, `convertCost()` is a bit implicit... convert to what? How about `displayCost()` instead?

Separately, what about moving this function to the interface adapter itself? e.g. `interfaceAdapter.displayCost()`. Maybe the name could be better, but this function won't change per-transaction report, right?

packages/reporters/reporters/migrations-V5/reporter.js
```diff
@@ -367,35 +369,21 @@ class Reporter {
...
-        data.receipt.blockNumber
-      );
-
-      // if geth returns null, try again!
-      if (!block) return this.postDeploy(data);
-
-      data.timestamp = block.timestamp;
-
-      const balance = await data.contract.interfaceAdapter.getBalance(tx.from);
+      const { convertCost, ...txData } = await data.contract.interfaceAdapter.getTransactionReportData(data.receipt);
```
gnidan, line 372: This makes me think that we should just move `interfaceAdapter` to the reporter itself. Maybe that can be separate.

gnidan (changes requested):
This looks good! Mostly had comments about naming, but wondering if your `convertCase` should get moved to the interface adapter proper.

packages/interface-adapter/lib/adapter/types.ts
bergarces: The reason I did it this way is because `convertCost` is not only called on `postDeploy`, but also on `postMigrate`. `postDeploy` has the contract as part of the data to call the interface adapter, but `postMigrate` doesn't.

It can be refactor so that `postMigrate` takes an object with the `interfaceAdapter` as well as `isLast` parameter (which I suppose it's possible since that occurs in `@truffle/migrate`).

packages/reporters/reporters/migrations-V5/reporter.js
bergarces: Not sure I'm following you here. Do you mean that reporter should have its own `interfaceAdapter` when it's instantiated instead of using the one from the contract?

It would limit the ability to use different `interfaceAdapter` per contract, but that's something that wouldn't work anyway as they need to have the same cost unit to add them up.

bergarces: It would also sort the issue I mentioned on the previous comment, about `postMigrate` not having access to `interfaceAdapter` right now.

gnidan: Oh I see. I did not consider that different contracts can have different interface adapters.

gnidan: Thanks for the insight!

packages/interface-adapter/lib/adapter/types.ts
gnidan: Let's just pass the same data to `postMigrate` as we do to `postDeploy`. I prefer that instead of this function here.

packages/interface-adapter/lib/adapter/types.ts
bergarces: Renamed to `TransactionCostReport`.

packages/interface-adapter/lib/adapter/types.ts
bergarces: Renamed.

packages/interface-adapter/lib/adapter/types.ts
bergarces: I've refactored `postMigrate` to take a data object like other events. The payload is not the same as `postDeploy` as only the `interfaceAdapter` is needed and we don't have access to everything else, but I've been able to get rid of `convertCost` from the report and now `displayCost` is an `interfaceAdapter` method.

packages/interface-adapter/lib/adapter/types.ts
```diff
@@ -1,3 +1,4 @@
+import BN from "bn.js";
```
gnidan, line 1:
```suggestion
import type BN from "bn.js";
```

gnidan (approved):
This looks good! One minor note about `import type`.

**Note** before releasing this, we need to make sure we know if this is major/minor for any packages it touches. It's probably breaking something.

bergarces:
> **Note** before releasing this, we need to make sure we know if this is major/minor for any packages it touches. It's probably breaking something.

Regarding breaking changes:
- `@truffle/interface-adapter`: Only adds methods and add fields to types.
- `@truffle/migrate`: `_deploy` method has been changed, but I've checked that it only gets called internally and no tests call it directly.
- `@truffle/reporters`: `postDeploy` interface and data payload is not changed. All changes are internal. And `postMigrate` interface is changed (breaking change), but the only place where it is called (`@trufffle/migrate`) has been updated accordingly.

PR #3997, opened 2021-04-22
Replace request library with axios

eggplantzzz:
The request library has been long deprecated. See https://www.npmjs.com/package/request.  This PR removes it (and related libraries) and uses axios instead.


haltman-at:
It starts removing it, I guess?  I know there are other places we use `request`.

eggplantzzz:
Why is the coveralls build suddenly working?

haltman-at:
> Why is the coveralls build suddenly working?

Beats me, but it seems to be a sporadic thing.  Like it also produced a report on `compile-yul`, but not on any of my other recent PRs.

packages/source-fetcher/lib/sourcify.ts
```diff
@@ -111,7 +111,8 @@ const SourcifyFetcher: FetcherConstructor = class SourcifyFetcher
     sourcePath: string
   ): Promise<string> {
     return await this.requestWithRetries<string>({
-      uri: `https://${this.domain}/contract/${this.networkId}/${address}/sources/${sourcePath}`
+      url: `https://${this.domain}/contract/${this.networkId}/${address}/sources/${sourcePath}`,
```
haltman-at, line 114: In this case, the response is not JSON, so I believe you'll have to set `responseType: text`.

(IMO, it also might be a good idea to explicitly set `responseType: json` where you rely on that default, just for explicitness.)

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -89,22 +89,20 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
     throw lastError;
   }
 
-  private async makeRequest(address: string): Promise<EtherscanSuccess> {
```
haltman-at, line 92: Could we not change how this works, please?  Let's keep all the method signatures the same if we can.  In this case, you've changed a bunch of functions to return different types than they used to, and the result isn't even correct.

I think you've gotten mixed up between whether the *request* succeeds, at a protocol level; and whether the returned result (*on* success) represents a success or a failure at an *API* level.  If the original request fails, the call to `request` will throw!  That's *not* what `response.status === "0"` is checking for!  That check only occurs if the request *succeeds*, and is checking to see if the returned request or response represents a success or a failure at an *API* level!  It can't be cut out!

Similarly, above, where I have `response.result`, you just have `data`, instead of `data.result`!  No, `result` is an actual field here!  The variable `response` here does *not* represent something like the axios response, with all its info; it represents the equivalent `response.data`.  Again, you've gotten the two levels confused.

Please redo this correctly so that the functionality and method signatures remain the same.

haltman-at (changes requested):
The changes to `etherscan.ts` are incorrect, and it's possible some modifications to `sourcify.ts` may be necessary, too.

packages/source-fetcher/lib/etherscan.ts
haltman-at: To clarify -- a failure at an API level would mean, e.g., that the requested contract isn't verified.  Request succeeded, but there was nothing to get.

eggplantzzz: Yes to be honest this is all very hard for me to reason about, especially since I can't poke at it somehow. I'll try again and see if I can fix it up.

haltman-at: As best I can tell I think you handled the 404 stuff correctly!  The difference in the success cases is that axios returns an object with the body in the `data` field, while request by default just returns the body.

haltman-at: Also sorry again about the lack of tests on `source-fetcher`... that's been a problem I've needed to fix for a long time. :-/

Actually, if you want to test things out yourself, I just realized, it's not actually that hard, I was just overcomplicating things.  Here's some Goerli transactions you can test it with:

1. `0x6533c2fe79231ef26982b9bec58efe8d9f79e53857b2ed2fab7d750a8e1bce77` (tests etherscan, single-source case)
2. `0x5cb5a3e533e4095233dddc1940a7c4bc586aca1f98995e0a5d5795bb049ff466` (tests etherscan, multi-source case)
3. `0x377a18f15eef123ed9f63d0725591e9b5b36854c7fa0c4c474c5933cb081c1a5` (tests etherscan, json case)
4. `0x1b3c57ba0d6ececbfeadcc12dfdb8e7663cfbf58b73fe6edb2a50ae11a6c7da3` (tests sourcify)

So, try debugging those transactions on Goerli with `--fetch-external` in an otherwise empty project, and see whether the source code gets recognized or not!

Although, right now, the sourcify one isn't working for me even on develop, hm... will need to see what's up with that...

haltman-at: Well uh I figured out what's going wrong with `sourcify` and uh sorry @eggplantzzz but I'm about to cause a merge conflict >_>

eggplantzzz: No problem, I resolved it and re-implemented this. Thanks for the testing steps!

eggplantzzz:
Thanks @haltman-at! I still need to test this using what you provided above.

haltman-at:
I actually tested those transactions myself, but if you want to test them in addition to be sure, well, can't hurt. :)

eggplantzzz:
Oh great!

PR #3999, opened 2021-04-24
contract: bump browserify and exorcist versions

jwendell:
browserify current version depends on an old version of `util`,
which does not have the function `callbackify` defined. This leads
the browser to emit this error:

```
Uncaught (in promise) "TypeError: callbackify is not a function"
```

Also, bump exorcist to match their dependency on browserify 17.0.

### DIFF
--- a/packages/contract/package.json
+++ b/packages/contract/package.json
@@ -33,9 +33,9 @@
     "web3-utils": "1.2.9"
   },
   "devDependencies": {
-    "browserify": "^14.0.0",
+    "browserify": "^17.0.0",
     "debug": "^4.1.0",
-    "exorcist": "^1.0.1",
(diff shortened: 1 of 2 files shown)


haltman-at:
Hey, could you redo this against `develop` rather than `next`?  (I tried changing the base, but, um, the two were too different for that to be workable.)

We should really get rid of the thing where we suggest making PRs against `next`; that's out of date.  Sorry about that.  I'll put in a PR to change that.  (I think something this small could be done on `develop` regardless, though! :) )

jwendell:
Done: https://github.com/trufflesuite/truffle/pull/4003

haltman-at:
Thank you!

PR #4004, opened 2021-04-27
Bug fix: Fix sourcify support in source-fetcher

haltman-at:
This PR fixes the support for sourcify in the source fetcher, due to the changes to their API.

This PR does not add support for constructor arguments with sourcify, because I haven't figured out how that works yet; I was going to include it but then it turned out to be harder than expected, and I figured it was more important to just get this fix out fast.  I'll mark that down as an issue for the future though.


haltman-at:
Oops, opened this prematurely.

Issue #3998, opened 2021-04-24
Not all packages' `package.json` files include `repository` field + `directory`


gnidan:
NPM respects the `repository` field, and it allows us to specify subdirectory, i.e. for monorepo purposes!

All packages in this monorepo should probably have this populated, e.g.:
```javascript
{
  "name": "@truffle/core",
  /* ... */
  "repository": {
    "type": "git",
    "url": "https://github.com/trufflesuite/truffle.git",
    "directory": "packages/core"
  },
  /* ... */
}
```

(Raised by @haltman-at [here](https://github.com/trufflesuite/truffle/pull/3943#discussion_r617047761))

Rishabhraghwendra18:
Hi . First timer this side . Is this issue open ?

haltman-at:
Oops, no, this should be closed now!  Thanks for catching this!

PR #4018, opened 2021-05-05
Internal improvement: Add/update repository field and other missing fields

haltman-at:
Addresses #3998.

This PR updates all package's `package.json` files to all use the same format for the `repository` field, with explicit `directory` specified.  I also added the `homepage` and `bugs` fields wherever I noticed them missing.

### DIFF
--- a/packages/truffle/package.json
+++ b/packages/truffle/package.json
@@ -6,7 +6,8 @@
   "homepage": "https://github.com/trufflesuite/truffle/",
   "repository": {
     "type": "git",
-    "url": "https://github.com/trufflesuite/truffle.git"
+    "url": "https://github.com/trufflesuite/truffle.git",
+    "directory": "packages/truffle"
   },
   "bugs": {
     "url": "https://github.com/trufflesuite/truffle/issues"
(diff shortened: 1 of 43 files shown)


packages/truffle/package.json
```diff
@@ -4,7 +4,10 @@
   "description": "Truffle - Simple development framework for Ethereum",
   "author": "consensys.net",
   "homepage": "https://github.com/trufflesuite/truffle/",
-  "repository": "https://github.com/trufflesuite/truffle/tree/master/packages/truffle",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/trufflesuite/truffle.git"
```
gnidan, line 9: Does this one not need a `directory`?

gnidan (approved):
This is probably fine but I think `truffle` probably could use a `directory` field (the package itself is not at the root, after all!)

packages/truffle/package.json
haltman-at: Hm, I guess it should have one.  I'll go change that.

Issue #4000, opened 2021-04-26
hdwallet-provider@1.3.0 not letting deploy


aster2709:
- [✅] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue
@truffle/hdwallet-provider@1.3.0 not letting deploy contracts, however works just as fine for 1.2.6
// with 1.3.0
![image](https://user-images.githubusercontent.com/39130640/116079745-cf2c2400-a6b5-11eb-9484-7ed5c1215a9c.png)
// with earlier versions
![image](https://user-images.githubusercontent.com/39130640/116079794-df440380-a6b5-11eb-818f-debf0e151fcb.png)

## Steps to Reproduce
try to deploy with @truffle/hdwallet-provider@1.3.0
```const provider = new HDWalletProvider(mnemonic, provider);```

## Expected Behavior
expected to deploy 

## Actual Results
```
node deploy.js 
deploying from 0x31Dac9cD8B4fDc7B1d4F139e47a5703512460240
node:internal/process/promises:245
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<Object>".] {
  code: 'ERR_UNHANDLED_REJECTION'
}

```

## Environment

* Operating System: linux ubuntu
* Ethereum client:infura api
* Truffle version (`truffle version`): @truffle/hdwallet-provider@1.3.0
* node version (`node --version`):@15.14.0
* npm version (`npm --version`): @7.7.6

**Comment by christoph2806** [comment]
Hi, I experienced a similar problem today while deploying to xDai (chainId = 100).
```
$ truffle migrate --reset --network xdai

Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.



Starting migrations...
======================
> Network name:    'xdai'
> Network id:      100
> Block gas limit: 12500000 (0xbebc20)


01_initial_migration.js
=======================

   Replacing 'Migrations'
   ----------------------

Error:  *** Deployment Failed ***

"Migrations" -- Invalid chain id..

    at /home/christoph/.nvm/versions/node/v10.15.3/lib/node_modules/truffle/build/webpack:/packages/deployer/src/deployment.js:365:1
    at process._tickCallback (internal/process/next_tick.js:68:7)
[...]

PR #4019, opened 2021-05-06
Fetch chain id for more robust support for EIP155 compliant transactions

eggplantzzz:
See https://github.com/trufflesuite/truffle/issues/4000

In order to better support EIP155, we now fetch the chain id for the user if none is supplied. We also imlement the ethereumjs-common module which allows up to support chains with custom chain ids.


packages/hdwallet-provider/package.json
```diff
@@ -22,14 +22,14 @@
...
     "ethereum-protocol": "^1.0.1",
+    "ethereumjs-common": "^1.5.0",
     "ethereumjs-tx": "^2.1.2",
     "ethereumjs-util": "^6.1.0",
     "ethereumjs-wallet": "^1.0.1"
   },
   "devDependencies": {
     "@types/bip39": "^2.4.2",
     "@types/ethereum-protocol": "^1.0.0",
-    "@types/ethereumjs-tx": "^1.0.1",
```
gnidan, line 32: Was this unused?

packages/hdwallet-provider/src/index.ts
```diff
@@ -36,7 +44,10 @@ class HDWalletProvider {
   private walletHdpath: string;
   private wallets: { [address: string]: ethJSWallet };
   private addresses: string[];
-  private chainId?: number | string;
+  private chainId?: number;
+  private chainSettings: any;
```
gnidan, line 48: Can't we reuse the type from constructor? Or does this need to be `any`?

packages/hdwallet-provider/src/index.ts
```diff
@@ -36,7 +44,10 @@ class HDWalletProvider {
...
+  private hardfork: string;
+  private initialized: Promise<any>;
```
gnidan, line 50: Since this won't actually have a value, we can use `void`

```suggestion
  private initialized: Promise<void>;
```

packages/hdwallet-provider/src/index.ts
```diff
@@ -143,7 +135,8 @@ class HDWalletProvider {
             cb(null, tmpWallets[address].getPrivateKey().toString("hex"));
           }
         },
-        signTransaction(txParams: any, cb: any) {
+        async signTransaction(txParams: any, cb: any) {
```
gnidan, line 138: @davidmurdoch you said this was cool?

gnidan (commented):
Looks good! Couple comments about types, and wondering about `async` functions that take a callback (I think @davidmurdoch you told me that this was fine here?)

packages/hdwallet-provider/src/index.ts
```diff
@@ -131,6 +108,21 @@ class HDWalletProvider {
     const tmpAccounts = this.addresses;
     const tmpWallets = this.wallets;
 
+    // if user supplied the chain id, use that - otherwise fetch it
+    if (chainId || (chainSettings && chainSettings.chainId)) {
+      this.chainId = chainId || chainSettings.chainId;
```
davidmurdoch, line 113: `chainId` could be `0`, so you'll need to check for null/undefined explicitly

packages/hdwallet-provider/src/index.ts
```diff
@@ -155,10 +148,27 @@ class HDWalletProvider {
           } else {
             cb("Account not found");
           }
-          const chain = (chainId && typeof chainId === "string") ?
-            parseInt(chainId) :
-            chainId;
-          const tx = new Transaction(txParams, { chain });
+          const chain = (self.chainId && typeof self.chainId === "string") ?
+          // @ts-ignore
+            parseInt(self.chainId) :
```
davidmurdoch, line 153: I think you should parse the `chainId` to a number up front, where it is set.

packages/hdwallet-provider/src/index.ts
```diff
@@ -155,10 +148,27 @@ class HDWalletProvider {
...
+            self.chainId;
+          const KNOWN_CHAIN_IDS = new Set([1, 3, 4, 5, 42]);
+          let txOptions;
+          // @ts-ignore
+          if (chain && KNOWN_CHAIN_IDS.has(chain)) {
+            txOptions = { chain };
+          } else if (chain) {
```
davidmurdoch, line 160: `chain` may be `0` here, so you should check for null/undefined instead.

packages/hdwallet-provider/src/index.ts
```diff
@@ -215,18 +225,94 @@ class HDWalletProvider {
...
 
-  public send(
+  private initialize(): Promise<void> {
+    return new Promise(resolve => {
+      this.sendAsync({
+        jsonrpc: '2.0',
+        id: Date.now(),
+        method: 'eth_chainId',
+        params: []
+      }, (_error, response) => {
```
davidmurdoch, line 235: We should `reject` if there is an error, or if the response doesn't contain a valid QUANTITY in the `result`.

packages/hdwallet-provider/src/index.ts
```diff
@@ -215,18 +225,94 @@ class HDWalletProvider {
     });
   }
 
-  public send(
+  private initialize(): Promise<void> {
+    return new Promise(resolve => {
+      this.sendAsync({
```
davidmurdoch, line 230: If you call `this.engine.sendAsync` directly here will that obviate the need for the `this.initialized` check in `this.sendAsync`?

packages/hdwallet-provider/src/index.ts
```diff
@@ -215,18 +225,94 @@ class HDWalletProvider {
...
 
   public sendAsync(
     payload: JSONRPCRequestPayload,
     callback: JSONRPCErrorCallback | Callback<JsonRPCResponse>
   ): void {
-    this.engine.sendAsync.call(this.engine, payload, callback);
+    // @ts-ignore
+    if (this.initialized) {
+      this.initialized.then(() => {
+        this.engine.sendAsync.call(this.engine, payload, callback);
```
davidmurdoch, line 311: note: I don't think the `.call(this.engine` is necessary here, as the "call-site" is already `this.engine`.

packages/hdwallet-provider/package.json
davidmurdoch: It's already listed in `dependencies`.

davidmurdoch (changes requested):

packages/hdwallet-provider/package.json
eggplantzzz: I guess, so - looks like the types are included. If it builds, it should be good, no?

packages/hdwallet-provider/src/index.ts
eggplantzzz: I don't see why not. Is it worth the trouble of allowing `chainId` as a string?

packages/hdwallet-provider/src/index.ts
eggplantzzz: Ah yes, I didn't think of that.

packages/hdwallet-provider/src/index.ts
eggplantzzz: Yeah this is kind of a leftover from earlier. Currently I have this typed as a number, is it worth allowing it as a string as well?

packages/hdwallet-provider/src/index.ts
eggplantzzz: I agree, though we don't have a concept of `QUANTITY` in Truffle so maybe just for the error and check that it's a number.

packages/hdwallet-provider/src/index.ts
eggplantzzz: Why? We still need to wait for the chain id to be fetched.

eggplantzzz: Oh, I see what you mean. This is only for the call to get the chain id. I think that allows us to get rid of the if/then? TypeScript seems to think so!

packages/hdwallet-provider/src/index.ts
davidmurdoch: Yup. This should be fine, as Web3 ProviderEngine is 100% callback style and ignores the returned Promise, and I can't foresee them updating the library to support Promises.

packages/hdwallet-provider/src/index.ts
davidmurdoch: As long as it's a hex string it's probably fine.

Teccchnically the result could be greater than `Number.MAX_SAFE_INTEGER`, but I'm not sure this is an edge case worth handling.

packages/hdwallet-provider/src/index.ts
davidmurdoch: Maybe someone may have been passing the `sendAsync` function around (unattached from this.engine)? It might be nice to clean it up since I noticed it (verify nothing weird is happening first, of course).

packages/hdwallet-provider/src/index.ts
```diff
@@ -131,6 +104,21 @@ class HDWalletProvider {
     const tmpAccounts = this.addresses;
     const tmpWallets = this.wallets;
 
+    // if user supplied the chain id, use that - otherwise fetch it
+    if (typeof chainId !== "undefined" || (chainSettings && typeof chainSettings.chainId !== "undefined")) {
```
davidmurdoch, line 108: If the user passed in `null` for the `chainId` or `chainSettings.chainId` we won't fetch the `chainId` from the network, we'll just use `null`. Is that the intent here?

packages/hdwallet-provider/src/index.ts
```diff
@@ -131,6 +104,21 @@ class HDWalletProvider {
...
+      this.chainId = chainId || chainSettings.chainId;
+      this.initialized = Promise.resolve();
+    } else {
+      this.initialized = this.initialize();
+    }
+
+    // EIP155 compliant transactions are enabled for hardforks later
+    // than "spurious dragon"
```
davidmurdoch, line 116:
```suggestion
    // than or equal to "spurious dragon"
```

packages/hdwallet-provider/src/index.ts
davidmurdoch: @eggplantzzz

packages/hdwallet-provider/src/index.ts
```diff
@@ -215,18 +217,97 @@ class HDWalletProvider {
...
+        if (error) {
+          reject(error);
+        }
+        if (response && response.result) {
+          if (isNaN(parseInt(response.result, 16))) {
+            const message = "When requesting the chain id from the node, it" +
+              `returned the malformed result ${response.result}.`;
+            throw new Error(message);
+          }
+          this.chainId = response.result;
```
davidmurdoch, line 237:
```suggestion
          this.chainId = parseInt(response.result, 16);
```

packages/hdwallet-provider/src/index.ts
```diff
@@ -215,18 +217,97 @@ class HDWalletProvider {
...
+      this.engine.sendAsync({
+        jsonrpc: '2.0',
+        id: Date.now(),
+        method: 'eth_chainId',
+        params: []
+    }, (error: any, response: JSONRPCResponsePayload) => {
+        if (error) {
+          reject(error);
+        }
+        if (response && response.result) {
```
davidmurdoch, line 231: The node may respond with `response.error`, in which case we should handle that by rejecting.

packages/hdwallet-provider/src/index.ts
```diff
@@ -215,18 +217,97 @@ class HDWalletProvider {
...
+        const wallet = ethJSWallet.fromPrivateKey(privateKey);
+        const address = wallet.getAddressString();
+        this.addresses.push(address);
+        this.wallets[address] = wallet;
+      }
+    }
+  };
+
+
+  public async send(
```
davidmurdoch, line 296: My gut is telling me you'll want to keep this as `non-async` and just use the `.then` pattern you used for `sendAsync` here as well.

packages/hdwallet-provider/src/index.ts
eggplantzzz: Yeah I talked to @gnidan and proposed that at that point it's on them. In the future if people really want to be able to pass in `null` and make a good argument, then I suppose we can allow it. For now I'll claim that this is fine unless someone else wants to argue with me.

packages/hdwallet-provider/src/index.ts
eggplantzzz: good catch!

packages/hdwallet-provider/src/index.ts
eggplantzzz: No, currently it will probably break if they put `null` in there. That is what `chainId` will be set to: that or `undefined`. I was just saying that I don't really want to care about that. Just like if they put some random object as a value for `chainId`, I don't want to care about that either. Unless we want to implement a much more robust input validation utility here.

packages/hdwallet-provider/src/index.ts
eggplantzzz: So I did this because we were previously returning a value. Do we just want to fire this one off async and return the Promise? And in that case, would it be any different than what I've done here?

davidmurdoch: Ah. Looks like Web3ProviderEngine throws on send anyway? https://github.com/MetaMask/web3-provider-engine/blob/2189a1a74552cf88cff6b1ec2b61aac6769fc92e/index.js#L114

Have you tried this?

packages/hdwallet-provider/src/index.ts
```diff
@@ -215,18 +217,97 @@ class HDWalletProvider {
...
+  private initialize(): Promise<void> {
+    return new Promise((resolve, reject) => {
+      this.engine.sendAsync({
+        jsonrpc: '2.0',
+        id: Date.now(),
+        method: 'eth_chainId',
+        params: []
+    }, (error: any, response: JSONRPCResponsePayload) => {
+        if (error) {
+          reject(error);
```
davidmurdoch, line 229:
```suggestion
          reject(error);
          return;
```

packages/hdwallet-provider/src/index.ts
```diff
@@ -155,10 +144,23 @@ class HDWalletProvider {
...
-          const chain = (chainId && typeof chainId === "string") ?
-            parseInt(chainId) :
-            chainId;
-          const tx = new Transaction(txParams, { chain });
+          const chain = self.chainId;
+          const KNOWN_CHAIN_IDS = new Set([1, 3, 4, 5, 42]);
+          let txOptions;
+          if (typeof chain !== "undefined" && KNOWN_CHAIN_IDS.has(chain)) {
+            txOptions = { chain };
+          } else if (typeof chain !== "undefined") {
```
slhodak, line 152: 👍

packages/hdwallet-provider/src/index.ts
eggplantzzz: another good catch

packages/hdwallet-provider/src/index.ts
eggplantzzz: Answered in Slack but leaving here for reference; Truffle uses an old, forked version of web3-provider-engine which calls `sendAsync` from `send`. https://github.com/trufflesuite/provider-engine/blob/master/index.js#L116-L118

packages/hdwallet-provider/src/index.ts
eggplantzzz: So I got rid of this check for `response` and `response.result` since the spec implies one or the other will be there. I also created a utility type to account for the `error` field and reject if it is there.

davidmurdoch (approved):

HankHoneymoon:
This also fixed the issue for me.

ferittuncer:
Can't migrate to Sokol, getting "Error: Chain with ID 77 not supported". 

[ethereumjs-common](https://github.com/ethereumjs/ethereumjs-monorepo/tree/master/packages/common/src/chains) has no definition for Sokol.

Issue #4006, opened 2021-04-27
Add support for constructor arguments in sourcify source-fetcher


haltman-at:
## Issue

[Sourcify](https://github.com/ethereum/sourcify) recently added support for constructor arguments; our sourcify source-fetcher should be updated to match.  Note I haven't yet figured out exactly how this works -- the sourcify docs gives the address to access constructor arguments at, but no further details; so I'd like to make sure we have an actual example to test this with before adding this.

## Environment

* Truffle version (`truffle version`): 5.3.3

PR #4020, opened 2021-05-07
Enhancement: Add partial match, contract name, and constructor arguments support to Sourcify source-fetcher

haltman-at:
Addresses #4006, among other things.  (**Note**: This PR has been updated since this was originally written!)

This PR makes a few improvements to the Sourcify source fetcher; I'll list them in decreasing order of importance.

1. This PR makes it so that rather than just checking for a full match, it'll check for a partial match as well.  This means that it will be possible to debug more contracts with `truffle debug --fetch-external`, since some contracts may only be verified as a partial match.  It checks for a full match first, and if that fails, tries for a partial match.  Arguably these could be two separate fetchers, but that seems like it would be confusing, so I did it this way.
2. This PR adds support for getting contract names, which will be useful for Truffle DB.  I had thought Sourcify didn't supply contract names; turns out it can be extracted from the metadata, oops.  Note that `contractName` is still *not* a required field in a returned `SourceInfo` though, because of Vyper contracts on Etherscan.  We still have no way to get a contract name for those.
3. **Added**: This PR adds support for getting constructor arguments from Sourcify.  Note that these aren't always listed, so this remains an optional field.  (Sourcify only stores the constructor arguments when they were necessary for verification; otherwise it throws them away.)
4. The Sourcify source fetcher had this useless language check?  It's gone now.  It was supposed to guard against any problems that might result from, like, getting metadata for a Yul contract, if that existed?  But it turns out it in fact does not exist -- solc does not produce metadata when compiling Yul -- so there's no need to guard against it.  Also, we support compiling Yul now, so if such metadata *did* exist there would no longer be a need to guard against it.
5. I shortened some code in one or two places. :P


packages/source-fetcher/lib/etherscan.ts
```diff
@@ -91,7 +91,7 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
 
   private async makeRequest(address: string): Promise<EtherscanSuccess> {
     //not putting a try/catch around this; if it throws, we throw
-    const response: AxiosResponse<EtherscanResponse> = await axios.get(
+    const response = (await axios.get(
```
eggplantzzz, line 94: A small nitpick but should we rename this to something that indicates that it is not the response and rather the payload? Either way I'm fine with it.

packages/source-fetcher/lib/sourcify.ts
```diff
@@ -52,50 +52,62 @@ const SourcifyFetcher: FetcherConstructor = class SourcifyFetcher
...
-                  : await this.getSource(address, sourcePath)
-            })
-          )
-        ))
-      );
-    } else {
-      //don't bother attempting to fetch sources if it's Yul
-      sources = {};
-    }
+    sources = Object.assign(
```
eggplantzzz, line 75: So what does the removal of the check for language mean? So the sources can be Solidity or Yul...can they be Vyper?

haltman-at: This is explained under #4.  The language check is removed because it's unnecessary, because the sources can only ever be Solidity in the first place.  Yul and Vyper don't have metadata, so there's not some way to use them with Sourcify.

packages/source-fetcher/lib/etherscan.ts
haltman-at: I think the name's fine.  It is a response, just not in the sense you're talking about.  I should probably add the type annotations back in though; I'll go do that.

eggplantzzz (approved):

PR #4022, opened 2021-05-07
Disable warning for `--config` by adding system for global command options

fainashalts:
This PR seeks to ensure that the warning about unsupported command options does not happen in situations where the command does, in fact, support that option. The target is specifically the `--config` option, but in doing this work I pulled out
`--network` as well since it was common to many of the commands Truffle offers.

The idea here was to create a way to address command options that are common across multiple commands, since the alternative would have been to add the `--config` option in 7 places. A special case was set up for the `develop` command because it really shouldn't support a `--network` option, but should be able to support a `--config` option, conceivably.

Expected behavior: 

1) The commands identified in the new `command-options.js` file should not throw a warning when accompanied by the `--config` option. 
2) The commands should show the `config` and (when appropriate) `network` usage and options when a user uses the `truffle help <command>` command. 
3) All commands and options should work as before.


packages/core/lib/command.js
```diff
@@ -107,13 +108,20 @@ class Command {
...
+
+
+    const validOptions = allValidOptions
       .map(item => {
         let opt = item.option.split(" ")[0];
         return opt.startsWith("--") ? opt : null;
       })
       .filter(item => item != null);
 
+
```
fainashalts, line 124:
```suggestion
```

packages/core/lib/commands/compile.js
```diff
@@ -58,7 +52,7 @@ const command = {
         internal: true,
         description:
           "Save the raw compiler results into <output-file>, overwriting any existing content."
-      }
+      },
```
fainashalts, line 55:
```suggestion
      }
```

packages/core/lib/commands/debug.js
```diff
@@ -61,7 +57,7 @@ const command = {
         option: "--compile-none",
         description:
           "Forces the debugger to use artifacts even if it detects a problem.  Dangerous; may cause errors."
-      }
+      },
```
fainashalts, line 60:
```suggestion
      }
```

packages/core/lib/commands/migrate.js
```diff
@@ -108,7 +102,7 @@ const command = {
         option: "--describe-json",
         description:
           "Adds extra verbosity to the status of an ongoing migration"
-      }
+      },
```
fainashalts, line 105:
```suggestion
      }
```

packages/core/lib/commands/command-options.js
```diff
@@ -0,0 +1,13 @@
...
+      description:
+        "Specify the network to use. Network name must exist in the configuration."
+    },
+    {
+      option: "--config <file>",
+      description:
+        "Specify configuration file to be used. The default is truffle-config.js"
+    }
+];
+ module.exports = options;
```
gnidan, line 13: let's move this file outside of `lib/commands`, since that directory is only actual commands right now

packages/core/lib/commands/migrate.js
```diff
@@ -47,7 +47,7 @@ const command = {
   },
   help: {
     usage:
-      "truffle migrate [--reset] [--f <number>] [--to <number>] [--network <name>]\n" +
```
gnidan, line 50: We want this to appear in the usage string, no?

packages/core/lib/commands/compile.js
```diff
@@ -58,8 +52,9 @@ const command = {
         internal: true,
         description:
           "Save the raw compiler results into <output-file>, overwriting any existing content."
-      }
-    ]
+      },
+    ],
+    allowedGlobalOptions: ["--network <name>", "--config <file>"]
```
gnidan, line 57: not necessary, but could we just make this:
```suggestion
    allowedGlobalOptions: ["network", "config"]
```

packages/core/lib/commands/console.js
```diff
@@ -4,7 +4,7 @@ const command = {
     "Run a console with contract abstractions and commands available",
   builder: {},
   help: {
-    usage: "truffle console [--network <name>] [--verbose-rpc]",
```
gnidan, line 7: I think we want to keep `--network` everywhere it appears in the usage (because it will be so common)

packages/core/lib/commands/command-options.js
fainashalts: moved!

packages/core/lib/commands/migrate.js
gnidan: Nevermind, didn't realize it was tacked on

packages/core/lib/commands/console.js
fainashalts: We discussed this offline. Network is added later in the help function! :)

packages/core/lib/command-options.js
```diff
@@ -0,0 +1,13 @@
...
+      option: "--network <name>",
+      description:
+        "Specify the network to use. Network name must exist in the configuration."
+    },
+    {
+      option: "--config <file>",
+      description:
+        "Specify configuration file to be used. The default is truffle-config.js"
+    }
+];
```
gnidan, line 12: We discussed turning this into an object, something:

```javascript
const options = {
  network: {
    option: "--network <name>",
    description: "..."
  },
  /* ... */
}
```

Then, down below, rather than doing a filter, just use each command's specified `allowedGlobalOptions` (e.g. `allowedGlobalOptions: ["network"]`) to retrieve from that object mapping.

packages/core/lib/command.js
```diff
@@ -107,7 +108,14 @@ class Command {
     if (!Array.isArray(inputStrings)) inputStrings = inputStrings.split(" ");
     // Method `extractFlags(args)` : Extracts the `--option` flags from arguments
     const inputOptions = extractFlags(inputStrings);
-    const validOptions = result.command.help.options
+
+    //adding allowed global options as enumerated in each command
+    const allowedGlobalOptions = commandOptions.filter((option) => result.command.help.allowedGlobalOptions.includes(option.option));
```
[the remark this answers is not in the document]
gnidan: Note you may want some kind of error-handling here? In case someone mistypes `network` or something.

packages/core/lib/commands/compile.js
eggplantzzz: Do we need to retain this info somehow though? It is pretty nice to have the descriptive `<name>` as a descriptor with each option. Any ideas on how to do this and keep that?

packages/core/lib/command.js
```diff
@@ -107,7 +108,12 @@ class Command {
     if (!Array.isArray(inputStrings)) inputStrings = inputStrings.split(" ");
     // Method `extractFlags(args)` : Extracts the `--option` flags from arguments
     const inputOptions = extractFlags(inputStrings);
-    const validOptions = result.command.help.options
+
+    //adding allowed global options as enumerated in each command
+    const allowedGlobalOptions = result.command.help.allowedGlobalOptions.map(tag => commandOptions[tag].option ? commandOptions[tag] : "");
```
gnidan, line 113: Better to avoid use of `""` where it's not necessary:
```suggestion
    const allowedGlobalOptions = result.command.help.allowedGlobalOptions
      .filter(tag => tag in commandOptions)
      .map(tag => commandOptions[tag].option);
```

This will mean that Truffle ignores malformed commands and falls back on the existing warning mechanism for unknown flags, which seems to be what you want.

packages/core/lib/commands/help.js
```diff
@@ -52,12 +55,19 @@ const command = {
       commandHelp = await commandHelp(options);
     }
 
-    console.log(`\n  Usage:        ${commandHelp.usage}`);
+    const allowedGlobalOptions = commandHelp.allowedGlobalOptions.map(tag => commonOptions[tag].option ? commonOptions[tag] : "");
```
gnidan, line 58: Same as above:

```suggestion
    const allowedGlobalOptions = commandHelp.allowedGlobalOptions
      .filter(tag => tag in commonOptions)
      .map(tag => commonOptions[tag].option);
```

gnidan: also you're calling it `commonOptions` here but `commandOptions` above - which is it?

packages/core/lib/commands/help.js
```diff
@@ -52,12 +55,19 @@ const command = {
...
+    const validOptionsUsage = allowedGlobalOptions.map(option => "[" + option.option + "]" ).join(" ");
```
gnidan, line 59: style:
```suggestion
    const validOptionsUsage = allowedGlobalOptions
      .map(({ option }) => `[${option}]`)
      .join(" ");
```

gnidan (changes requested):
Few small things left

packages/core/lib/commands/compile.js
gnidan: Ah yeah, I wasn't clear in my comments - see how Faina has it now!

packages/core/lib/commands/help.js
fainashalts: Renamed!

packages/core/lib/command.js
fainashalts: I just pushed a commit switching this out for `filter` as you suggested. What you have here doesn't work because we don't want to return the short option, but the object attached to it. Let me know if that doesn't make sense!

PR #4024, opened 2021-05-08
Replace old name for `unsafeNativizeVariables`

gnidan:


haltman-at (approved):
Oh, yikes, thanks!  The dangers of not using Typescript...

Issue #4016, opened 2021-05-05
Bad Requests behind a proxy since version 5.3.4


ityreh:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Using a truffle operation **on a Windows System** with truffle version `5.3.4`, that sends a HTTP request, you get a **Bad Request** Error. My Windows System is **behind a proxy**.

Maybe the [switch to axios](packages/compile-solidity/compilerSupplier/loadingStrategies/VersionRange.js) in `5.3.4` is causing this issue.

## Steps to Reproduce

Upgrade your truffle version to `5.3.4` and try to use an operation that sends a HTTP request, e.g. `truffle unbox metacoin` or in an existing project `truffle compile --list`.

## Expected Behavior

Truffle does not send bad requests.

## Actual Results

One of the operations that sends a bad request ist `truffle compile --version`.

Output with version `5.3.4` (not working):

```
C:\Temp>truffle compile --list
× Fetching solc version list from solc-bin. Attempt #1
× Fetching solc version list from solc-bin. Attempt #2
× Fetching solc version list from solc-bin. Attempt #3
Error: Failed to complete request to: version URLs. Are you connected to the internet?

Error: Request failed with status code 400
    at VersionRange.errors (C:\Users\xn06112\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\compile-solidity\compilerSupplier\loadingStrategies\LoadingStrategy.js:66:1)
    at C:\Users\xn06112\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\compile-solidity\compilerSupplier\loadingStrategies\VersionRange.js:158:1
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Object.listVersions (C:\Users\xn06112\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\core\lib\commands\compile.js:122:1)
    at Object.run (C:\Users\xn06112\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\core\lib\commands\compile.js:71:1)
    at Command.run (C:\Users\xn06112\AppData\Roaming\npm\node_modules\truffle\build\webpack:\packages\core\lib\command.js:140:1)
Truffle v5.3.4 (core: 5.3.4)
Node v16.1.0
```

Output with version `5.3.3` (working):

```
C:\Temp>truffle compile --list
[...]

PR #4025, opened 2021-05-10
Bug fix: Add maxRedirects field to axios requests

eggplantzzz:
A user reported [here](https://github.com/trufflesuite/truffle/issues/4016) that they were experiencing issues with making http requests while using Truffle from behind a proxy. It seemed reasonable to assume this had something to do with switching axios in for the deprecated request library (see https://github.com/trufflesuite/truffle/pull/3997). After a small amount of investigation I discovered that axios has an option for `maxRedirects` that defaults to 5. In an attempt to fix this http request issue, this PR increases `maxRedirects` to 50.


eggplantzzz:
In source fetcher we could make that method that accepts the request object take a couple of parameters that get passed in instead?

Issue #4026, opened 2021-05-11
Better display of error context in debugger


haltman-at:
## Issue

Right now, the debugger doesn't do a great job of displaying the context when an error occurs.  It will show you a stacktrace and the location where it occurred, and you can press `s` again to repeat this with more source context, but that's it.

This is a little confusing if you expect that stepping ought to end *on* the error.  But in fact it goes past it, to the end of the trace; indeed, until we reach the end of the trace, we can't *know* that we had the right error earlier, so it's not really possible to identify the correct error ahead of time and stop there.

It's also not great for getting further information about the error, like, what were the values of the variables at the time?

Ideally, we could have the debugger reset to where the error occurred.  However, right now, the debugger isn't rewindable (#3854).  So, instead, we could have the debugger remember relevant information from the error context (such as variable values) and print it at the end.

A stopgap solution for now might be to simply have the debugger remember on what step the error occurred, and offer to return to that step.

## Environment

* Truffle version (`truffle version`): 5.3.5

haltman-at:
Note: #4029 has now been released as a stopgap solution to this, but I'm going to leave this open for now.

Issue #3854, opened 2021-02-20
Add some form of backward stepping to Truffle Debugger


haltman-at:
As stated.  This would need a lot of figuring out.

mnaei:
I was hoping to find something like the inverse of "n" command. 

"y" and "Y" are still useful.

haltman-at:
Yes, to be clear, the intent of the issue is that there should be an inverse of `n`; that's why it's still open.

PR #4029, opened 2021-05-13
Enhancement: Add `y` and `Y` commands to debugger to return to last error

haltman-at:
This is sort of a primitive hack to address #4026; not sure it should be considered a full resolution of it.

This PR adds two commands to the debugger CLI, and adds selectors to the debugger library to support this.  The `Y` command returns to the last place an error was through.  The `y` command, only usable at the end of the trace, returns to the final error, and does nothing if the trace ends on the final error.

Since the debugger doesn't currently step backwards (#3854), this is done inefficiently by resetting and then stepping to the required step.  But, it'll do for now?

In order to do this, I've added `innerReturnIndex` to the `stacktrace` state.  It keeps track of the index of the last error.  Note that unlike `innerReturnPosition` and `innerReturnStatus`, it's only set to `null` when there is no previous error; it tracks the index of the *last* error, not the index of the *current* error.  So we need to know when to update it; it's only updated when we're not already processing an error (`returnCounter === 0`).  Otherwise, it's left alone.

Note that the `y` command has to manually check if we're at a final revert, because if there is one, it won't affect `innerReturnIndex`.

### DIFF
--- a/packages/core/lib/debug/interpreter.js
+++ b/packages/core/lib/debug/interpreter.js
@@ -7,7 +7,7 @@ const ora = require("ora");
 
 const DebugUtils = require("@truffle/debug-utils");
 const selectors = require("@truffle/debugger").selectors;
-const { session, solidity, trace, evm, controller } = selectors;
+const { session, solidity, stacktrace, trace, evm, controller } = selectors;
 
 const analytics = require("../services/analytics");
 const repl = require("repl");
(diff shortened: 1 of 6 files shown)


packages/core/lib/debug/interpreter.js
```diff
@@ -365,6 +365,48 @@ class DebugInterpreter {
         this.printer.print("");
       }
     }
+    if (cmd === "y") {
+      if (this.session.view(session.status.loaded)) {
+        if (this.session.view(trace.finished)) {
+          if (!this.session.view(evm.current.step.isExceptionalHalting)) {
```
gnidan, line 371: These nested `if`s are killin' me

gnidan (approved):
Let's ship this :)

PR #4031, opened 2021-05-14
Separate PouchDB adapter interface from existing Workspace interface in @truffle/db

gnidan:
Currently, @truffle/db defines the [`Workspace<C>`](https://www.trufflesuite.com/docs/truffle/db/interfaces/meta.workspace.html) interface for use across many modules, implementing that workspace directly via the class hierarchy of PouchDB adapters (i.e., the base [`Databases<C>`](https://www.trufflesuite.com/docs/truffle/db/classes/meta.pouch.adapters.base.databases.html) class).

In order to support future work of allowing schema migrations, we can't abide the constraints defined by `Workspace<C>` - this interface accepts/returns data in _the current schema_. We need an interface that can read/write to our PouchDB adapter via _past versions of the schema_ as well.

This PR removes `implements Workspace<C>` from the PouchDB adapter classes, implementing instead the new `Adapter<C>` interface that provides weaker typing. It adds a new concrete class `AdapterWorkspace<C>`, to then translate an `Adapter<C>` into a `Workspace<C>`. Thus, existing code will continue to use `Workspace<C>`, but whatever future schema migration system can be built to use `Adapter<C>` directly.

To support this new Adapter interface, this PR also introduces some new naming: adapters now operate on Records, which have IDs (PouchDB's `_id` field), and SavedRecords, which have all the other Pouch fields (like `_rev`). The purpose of this new terminology is to distinguish these low-level types from the existing high-level types (e.g. `Input` and `SavedInput`), which are exposed by the Workspace interface and part of the current data model.

### DIFF
--- a/packages/db/src/meta/pouch/adapters/base.ts
+++ b/packages/db/src/meta/pouch/adapters/base.ts
@@ -18,6 +18,11 @@ import type { Workspace, Historical } from "@truffle/db/meta/data";
 
 import type { Definition, Definitions } from "@truffle/db/meta/pouch/types";
 
+export interface DatabasesOptions<C extends Collections> {
+  definitions: Definitions<C>;
+  settings: any;
+}
+
 /**
  * Aggegrates logic for interacting wth a set of PouchDB databases identified
  * by resource collection name.
@@ -29,7 +34,7 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
 
   private ready: Promise<void>;
 
-  constructor(options) {
+  constructor(options: DatabasesOptions<C>) {
     this.setup(options.settings);
 
     this.definitions = options.definitions;
--- a/packages/db/src/meta/pouch/adapters/base.ts
+++ b/packages/db/src/meta/pouch/adapters/base.ts
@@ -11,10 +11,11 @@ import type {
   MutationInput,
   MutationPayload,
   MutableCollectionName,
+  Input,
   SavedInput
 } from "@truffle/db/meta/collections";
 import * as Id from "@truffle/db/meta/id";
-import type { Workspace, Historical } from "@truffle/db/meta/data";
+import type { Historical } from "@truffle/db/meta/data";
 
 import type { Definition, Definitions } from "@truffle/db/meta/pouch/types";
 
@@ -23,11 +24,39 @@ export interface DatabasesOptions<C extends Collections> {
   settings: any;
 }
 
+export interface Adapter<C extends Collections> {
+  every<N extends CollectionName<C>, I extends { id: string }>(
+    collectionName: N
+  ): Promise<Historical<I>[]>;
+
+  retrieve<N extends CollectionName<C>, I extends { id: string }>(
+    collectionName: N,
+    references: (Pick<I, "id"> | undefined)[]
+  ): Promise<(Historical<I> | undefined)[]>;
+
+  search<N extends CollectionName<C>, I extends { id: string }>(
+    collectionName: N,
+    options: PouchDB.Find.FindRequest<{}>
+  ): Promise<Historical<I>[]>;
+
+  record<N extends CollectionName<C>, I extends { id: string }>(
+    collectionName: N,
+    inputs: (I | undefined)[],
+    options: { overwrite?: boolean }
+  ): Promise<(Historical<I> | undefined)[]>;
+
+  forget<N extends CollectionName<C>, I extends { id: string }>(
+    collectionName: N,
+    references: (Pick<I, "id"> | undefined)[]
+  ): Promise<void>;
+}
+
+
 /**
  * Aggegrates logic for interacting wth a set of PouchDB databases identified
  * by resource collection name.
  */
-export abstract class Databases<C extends Collections> implements Workspace<C> {
+export abstract class Databases<C extends Collections> implements Adapter<C> {
   private collections: CollectionDatabases<C>;
   private definitions: Definitions<C>;
   private generateId: Id.GenerateId<C>;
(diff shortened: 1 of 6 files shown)


gnidan:
Note: I've added two comments to clarify the boundary between Workspace and Adapter, including to define explicit marshaling/unmarshaling and to refine naming.

packages/db/src/meta/pouch/adapters/base.ts
```diff
@@ -153,192 +146,115 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
...
   }
 
-  public async remove<M extends MutableCollectionName<C>>(
-    collectionName: M,
-    input: MutationInput<C, M>
-  ): Promise<void> {
+  public async forget<N extends CollectionName<C>, T>(
+    collectionName: N,
+    references: (RecordReference<T> | undefined)[]
+  ) {
```
fainashalts, line 234: Did you mean to drop the return type annotation here?

packages/db/src/meta/pouch/types.ts
```diff
@@ -6,6 +6,81 @@ import PouchDB from "pouchdb";
...
+
+/**
+ * Low-level interface for PouchDB adapters - allows basic CRUD operations
+ * for arbitrary (weakly-typed) data types.
+ */
+export interface Adapter<C extends Collections> {
+  /**
+   * Retrieve and return every saved record for a given collection name
+   */
+  every<N extends CollectionName<C>, T>(
```
fainashalts, line 36: What are these function names based on? Is there guidance somewhere on how these CRUD-y functions should be named that you based this on? I know you can't use the `all` because you need it in the workspace file, just wondering how you arrived at this naming given that limitation.

packages/db/src/meta/pouch/types.ts
```diff
@@ -6,6 +6,81 @@ import PouchDB from "pouchdb";
...
+    options: PouchDB.Find.FindRequest<{}>
+  ): Promise<SavedRecord<T>[]>;
+
+  /**
+   * Create or update saved records of a given collection name.
+   *
+   * @param records - Can be sparse
+   * @return - Items in list correspond to `records` by index (also sparse)
+   */
+  record<N extends CollectionName<C>, T>(
```
fainashalts, line 67: I especially don't love `record` -- confusing when one of the types inside is `Record`

packages/db/src/meta/pouch/workspace.ts
```diff
@@ -0,0 +1,227 @@
...
+      throw new Error(`Unable to generate ID for "${collectionName}" resource`);
+    }
+
+    return {
+      ...input,
+      _id: id
+    };
+  }
+
+  private unmarshal<N extends CollectionName<C>>(
```
fainashalts, line 207: This is a cool way to do this, I feel like!

fainashalts (approved):
LGTM! Minor comments/questions.

packages/db/src/meta/pouch/adapters/base.ts
```diff
@@ -153,192 +146,115 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
...
-  public async update<M extends MutableCollectionName<C>>(
-    collectionName: M,
-    input: MutationInput<C, M>
-  ): Promise<MutationPayload<C, M>> {
-    await this.ready;
+    const savedRecordById: {
+      [id: string]: SavedRecord<I>;
+    } = (
+      await Promise.all(
+        Object.entries(recordsById).map(async ([_id, record]) => {
```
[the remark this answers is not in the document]
gnidan: Ah, that's probably a point worth driving home: Adapter now uses `_id` exclusively, so everything inside `src/meta/pouch/adapters`, and the new AdapterWorkspace converts those to `id`. Workspace now never outputs anything with `_id`, or any other Pouch-internal records (like `_rev`)

There's one exception, which might be worth changing: `fixIdSelector` is still used (it moved from Workspace.find to Adapter.search, now on L141 of this file). That function allows this Adapter to handle `id` when it's used in a low-level PouchDB find query.

packages/db/src/meta/pouch/adapters/base.ts
```diff
@@ -153,192 +146,115 @@ export abstract class Databases<C extends Collections> implements Workspace<C> {
...
-      [collectionName]: resourceInputIds.map(id => resourcesById[id])
-    } as unknown) as MutationPayload<C, M>;
+    return records.map(record => record && savedRecordById[record._id]);
   }
 
-  public async remove<M extends MutableCollectionName<C>>(
-    collectionName: M,
-    input: MutationInput<C, M>
-  ): Promise<void> {
+  public async forget<N extends CollectionName<C>, T>(
```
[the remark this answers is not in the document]
gnidan: `delete` might be better, but I was trying to avoid verb collisions with Workspace, and that defines `remove`

packages/db/src/meta/pouch/adapters/base.ts
gnidan: I did think to myself, "ah yes, TS knows the return type from the interface" and proceeded to be lazy and not add it.

But this sorta thing is generally better listed explicitly :) I'll add them all.

packages/db/src/meta/pouch/types.ts
gnidan: Gosh Faina I was just trying to be cute with naming cause I needed a bunch of verbs. We can use standard ones I suppose :)

Probably also doesn't matter much if some of the verbs collide with some of the verbs that Workspace uses? I had implemented them all in the same file to start, so I couldn't have collisions at that point and never changed it.

packages/db/src/meta/pouch/types.ts
gnidan: ugh it's so bad I know. is there another verb that means "create or update"? `save`?

packages/db/src/meta/pouch/workspace.ts
gnidan: I'm proud of this pair! 😄

packages/db/src/meta/pouch/adapters/base.ts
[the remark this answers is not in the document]
fainashalts: Oh! I see. Ok, that totally makes sense in the context of what you're doing here.

packages/db/src/meta/pouch/types.ts
fainashalts: I think `save` is fine! Tough to account for creating and updating in one word.

packages/db/src/meta/pouch/adapters/base.ts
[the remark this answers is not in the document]
gnidan: Changed this to `delete`

gnidan:
Fixed all the suggested name changes, and also moved the `fixIdSelector` logic to AdapterWorkspace as mentioned [here](https://github.com/trufflesuite/truffle/pull/4031#discussion_r636625338).

packages/db/src/meta/pouch/adapters/base.ts
```diff
@@ -18,6 +18,11 @@ import type { Workspace, Historical } from "@truffle/db/meta/data";
 
 import type { Definition, Definitions } from "@truffle/db/meta/pouch/types";
 
+export interface DatabasesOptions<C extends Collections> {
+  definitions: Definitions<C>;
+  settings: any;
```
cds-amal, line 22: Does this have to be an `any` type?

packages/db/src/meta/pouch/types.ts
```diff
@@ -75,7 +75,7 @@ export interface Adapter<C extends Collections> {
    *
    * @param references - Can be sparse
    */
-  forget<N extends CollectionName<C>, T>(
+  delete<N extends CollectionName<C>, T>(
```
cds-amal, line 78: like this name better!

packages/db/src/meta/pouch/types.ts
```diff
@@ -64,7 +64,7 @@ export interface Adapter<C extends Collections> {
    * @param records - Can be sparse
    * @return - Items in list correspond to `records` by index (also sparse)
    */
-  record<N extends CollectionName<C>, T>(
+  save<N extends CollectionName<C>, T>(
```
cds-amal, line 67: :)

cds-amal (commented):
Any tests for the marshalling?

I think this is OK for approving the PR but would still like to see the type magic commented as well as the functional logic is.

packages/db/src/meta/pouch/adapters/base.ts
gnidan: It could be `{}` if you prefer, but it's `any` right now because it allows subclasses to define whatever settings they want.

packages/db/src/meta/pouch/types.ts
```diff
@@ -4,8 +4,36 @@ const debug = logger("db:meta:pouch:types");
...
+    collectionName: N,
+    references: (Pick<I, "id"> | undefined)[]
+  ): Promise<(Historical<I> | undefined)[]>;
+
+  search<N extends CollectionName<C>, I extends { id: string }>(
+    collectionName: N,
+    options: PouchDB.Find.FindRequest<{}>
+  ): Promise<Historical<I>[]>;
+
+  record<N extends CollectionName<C>, I extends { id: string }>(
```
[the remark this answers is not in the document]
gnidan: This got changed!

packages/db/src/meta/pouch/adapters/base.ts
cds-amal: That makes sense! A comment to let subclass authors know the intent is better.

cds-amal (approved):
LTGM! Thanks @gnidan

gnidan:
Thanks @fainashalts and @cds-amal!

PR #4034, opened 2021-05-14
Internal improvement: Implement reset for txlog

haltman-at:
Somehow when writing `txlog` I forgot to implement resetting it.  This has been a pain when trying to debug problems with txlog.

Unfortunately txlog is complicated enough that directly resetting it wasn't practical, so I reused a hack from `evm`: We store the initial call, and in order to reset, rather than attempting to directly reset to after the initial call, we reset to just *before* the initial call and then replay that call.  (At the beginning, there are normally two `txlog` actions; one to set the origin, the other to do the initial call.  We reset to inbetween these two, so we do keep the origin.)

Note this doesn't really alter the end result of `txlog`, but it will make debugging it substantially easier in the future.

I also added a test of this functionality.  Also as you can see I also made some style changes. :P  (Husky still isn't running, but...)

### DIFF
--- a/packages/debugger/lib/evm/sagas/index.js
+++ b/packages/debugger/lib/evm/sagas/index.js
@@ -215,7 +215,7 @@ export function* callstackAndCodexSaga() {
 }
 
 export function* reset() {
-  let initialCall = yield select(evm.transaction.initialCall);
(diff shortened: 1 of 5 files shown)


cds-amal (approved):
This is pretty cool. Looks good!

PR #4035, opened 2021-05-16
WIP: @truffle/browser-provider

rkalis:


Hbutlercapone (approved):

gnidan:
(Note since I just remembered it: we'll need to make host configurable)

rkalis:
Just pushed the latest commit, I'd say that the infrastructure part is ready for a first round of review.

Manual testing can be done in two ways:
#### Manually starting the dashboard (useful when doing frontend development since it gives you hot reloading)
- Run `yarn` in Truffle root
- Run `yarn start` in `packages/dashboard`
- Run `yarn test` in `packages/browser-provider`
- Open a browser on `localhost:3000` (should auto-open) and click "Connect Web3"
- You should see a request come in and click "process"
#### Having BrowserProvider start the dashboard (closer to real world usage)
- Run `yarn` in Truffle root
- Run `yarn test` in `packages/browser-provider`
- Open a browser on `localhost:5000` and click "Connect Web3"
- You should see a request come in and click "process"

In both cases you need to close the dashboard page at the end to stop the process since the message bus keeps running until no more clients or listeners are connected.

If you make any changes to the message bus code you need to run `yarn build` for the other packages to pick it up, if you're having the BrowserProvider start the dashboard you also need to run `yarn build` after making any changes to the dashboard frontend code.

#### Notable things that we still need to do before calling it an MVP

- Configurable host for dashboard + message bus
- WalletConnect support?
- Polish on the frontend side
- Testing
	  - Automated testing (currently none)
	  - Testing with different browsers / platforms
	  - Testing usage within ethers.js / web3.js
	  - Testing usage with truffle migrations
	  - Testing usage with other non-truffle deployment tools
	  - Testing usage with hardware wallets
- Other things?

rkalis:
Just pushed some new fixes and manual testing examples. Also updated the README with some information on the testinng process:

## Automated testing

`test/` contains a few very simple tests that test basic functioning of the BrowserProvider + Message Bus + Dashboard infrastructure. It uses a mocked "dashboard" that rather than opening a browser window just forwards all requests to Ganache. These tests still need to be expanded.

## Manual testing

`manual-test/` contains some more "real-world" usage of the BrowserProvider. The `manual-test.ts` script contains some requests to the BrowserProvider, including `eth_sendTransaction` and `eth_signTypedData_v4`. `metacoin-truffle/` contains the source code for a simple MetaCoin project using Truffle. `metacoin-hardhat/` contains the same contract source code but using Hardhat.

To run these manual tests use the following commands:

```
yarn test:manual:basic
yarn test:manual:metacoin-truffle
yarn test:manual:metacoin-hardhat
```

Note that the dashboard does not automatically start so you'll have to open `http:localhost:5000` in your browser to use the browser provider dashboard.

---

The configurable host stuff hasn't been added yet, so @gnidan testing with your setup is probably still difficult. But for @eggplantzzz and @kevinbluer it should be possible to run the metacoin examples for truffle + hardhat, and you can also use those examples to try out  some other things if you like.

Let me know of any issues that you find. In the mean time I'll work on expanding the automated test cases.

rkalis:
About a month ago I wrote up a list of things that still needed to be done before we have an MVP. Since then I've finished the following:
* Create + style an actual frontend
* Add REJECT functionality
* Add automated testing for the message bus and browser provider
* Streamline manual testing + integration into truffle migrate / hardhat deploy
* Add `keepAlive` config to enable manual connection management

What still needs to be done from the original list:
* Configurable host for dashboard + message bus
* Go over all code to verify edge cases + clean up code

What we additionally want to do as MVP functionality:
* Display connected network
* Add some warning when connected to mainnet

Some nice to haves that we want to add before or after MVP:
* WalletConnect support
* Add prompt when current Metamask network is different from the "expected" network
* Dropdown selection for network
* Sequence diagrams of the BrowserProvider / Dashboard / Message Bus setup

packages/dashboard/src/utils/utils.ts
```diff
@@ -0,0 +1,79 @@
...
+  const buffer = Buffer.from(base64, "base64");
+  const stringifiedJson = buffer.toString("utf8");
+  const json = JSON.parse(stringifiedJson);
+
+  return json;
+};
+
+// TODO: Replace these functions with those in the Dashboard Message Bus package
+// Only difference rn is that this uses browser WebSocket, while the other uses Node.js WebSocket
+export const connectToMessageBusWithRetries = async (port: number, retries = 1, tryCount = 1): Promise<WebSocket> => {
```
gnidan, line 26: This needs to be updated to support `host`; it looks like you did that for the dashboard-message-bus version, but missed here.

Hbutlercapone: Will be doing asap. Thank you so much

rkalis:
We went over the dashboard + browser-provider today. The configurable hostname is working. But we did identify some issues when running on Nick's machine. 

One was an issue when running `truffle develop` and just trying to sign a message, it wouldn't propagate the response.

The other issue was when running migrations and switching networks in Metamask while the migrations were already running would stall the migrations forever.

From here we have a few more TODOs left:
- Debug the issues mentioned above
- Add actual logging (using the "debug" package - see truffle develop workflow)
- See if we can add a way to make sure the user is connected to the right network *before* allowing connections
- Add docs for adding new message types
- Add option to "invalidate" messages so that the dashboard can clean up any open requests that have already been fulfilled by other dashboards

rkalis:
All of the remaining TODOs fom above have been implemented.

rkalis:
I tried to implement WalletConnect today, but couldn't get it to work (yet). Instead I set the default host for the dashboard to "0.0.0.0" meaning that the dashboard should also be accessible from mobile devices on the same network if users want to use the dashboard using mobile wallets (which would be the main use case for WalletConnect).

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
```diff
@@ -0,0 +1,178 @@
...
+    await promisify(this.listenServer.close.bind(this.listenServer))();
+    this.emit("terminate");
+  }
+
+  private async processRequest(
+    socket: WebSocket,
+    data: WebSocket.Data,
+    listeners: WebSocket[]
+  ) {
+    if (typeof data !== "string") return;
```
kevinweaver, line 103: ❓ I'm assuming we expect Data to be some json object?  What case does this guard against?

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
```diff
@@ -0,0 +1,178 @@
...
+        `Sending response for message ${message.id}`,
+        "responses"
+      );
+      this.logToClients(response, "responses");
+
+      const encodedResponse = jsonToBase64(response);
+      socket.send(encodedResponse);
+      this.unfulfilledRequests.delete(data);
+
+      this.invalidateMessage(message.id);
```
kevinweaver, line 128: ❓ Why do we send "invalidate" messages once requests have been processed?  Is this to prevent another instance from processing it or something?

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
rkalis: All messages (log, invalidate, dashboard-provider) are JSON.stringified and then base64 encoded before they're sent over the wire, and then they're decoded at the other side of the wire. 

So we expect all of the messages to come in as a `"string"` type. But it's technically possible to send `Buffer` data through the websocket connection. Essentially this line just *ignores* that buffer data.

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
rkalis: Yeah that's it. If you have two dashboard windows open then all requests get sent to both dashboards. After processing the request in one dashboard, they would persist in the other dashboard. This "invalidate" step tells that other dashboard that the request no longer needs to be fulfilled.

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
```diff
@@ -0,0 +1,184 @@
...
+import { promisify } from "util";
+
+interface UnfulfilledRequest {
+  socket: WebSocket;
+  data: WebSocket.Data;
+}
+
+export class DashboardMessageBus extends EventEmitter {
+  private requestsServer: WebSocket.Server;
+  private listenServer: WebSocket.Server;
```
gnidan, line 21: The naming here is a bit asymmetric - are there other names we could use?

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
gnidan: Why not just convert buffer to string?

packages/dashboard-message-bus/lib/message/types.ts
```diff
@@ -0,0 +1,44 @@
+export interface Message {
+  id: number;
+  type: string;
+  payload: any;
+}
+
+export interface DashboardProviderMessage extends Message {
+  type: "dashboard-provider";
```
gnidan, line 8: can probably just call this 
```suggestion
  type: "provider";
```

packages/dashboard-message-bus/lib/message/types.ts
```diff
@@ -0,0 +1,44 @@
+export interface Message {
+  id: number;
+  type: string;
+  payload: any;
+}
+
+export interface DashboardProviderMessage extends Message {
```
gnidan, line 7: Each of these classes should probably have a docstring indicating what kind of message they're for

packages/dashboard-provider/lib/DashboardProvider.ts
```diff
@@ -0,0 +1,149 @@
...
+  getMessageBusPorts,
+  base64ToJson,
+  LogMessage
+} from "@truffle/dashboard-message-bus";
+import { startDashboardInBackground } from "@truffle/dashboard";
+import { timeout } from "promise-timeout";
+import { DashboardProviderOptions } from "./types";
+import debugModule from "debug";
+
+export class DashboardProvider {
```
gnidan, line 23: Should we make this support EIP-1193 while we're at it?

packages/dashboard/src/utils/constants.ts
```diff
@@ -0,0 +1,12 @@
...
+  "eth_sendTransaction",
+  "eth_decrypt",
+  "eth_signTypedData",
+  "eth_signTypedData_v1",
+  "eth_signTypedData_v3",
+  "eth_signTypedData_v4",
+  "personal_sign"
+];
+
+export const UNSUPPORTED_REQUESTS = ["eth_sign"];
```
gnidan, line 12: what's wrong with this one?

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
rkalis: We can do that as well. Assume it's utf8?

packages/dashboard/src/utils/constants.ts
rkalis: eth_sign silently fails with metamask, also when using it without the dashboard-provider. personal_sign works though. I can definitely document this better though. What do you think the best place is for that?

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
rkalis: Yeah most of the property names in that entire class are a bit off, but I'm not sure what the way to go is.

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
gnidan: yeah, that seems reasonable if it's not too much effort.

packages/dashboard/src/utils/constants.ts
gnidan: probably just give a more specific error message for this case, rather than a single bucket for everything unsupported.

unless we're being vague to conform with existing JSON-RPC vagueness

packages/dashboard-message-bus/README.md
```diff
@@ -0,0 +1,39 @@
+# @truffle/dashboard-message-bus
+
+:warning: **This is a middleware package and should only be used when developing integrations for the truffle dashboard** :warning:
+
+The `@truffle/dashboard-message-bus` manages the communication between the Truffle Dashboard and any other Truffle components, such as the browser provider. It works by running two separate WebSocket servers. One server is for clients to connect to and send requests, while the other server is for listeners (dashboards) to connect to and receive requests.
```
eggplantzzz, line 5: Did we decide whether to call this "browser provider" or "dashboard provider"?

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
```diff
@@ -0,0 +1,184 @@
...
+    super();
+  }
+
+  async start() {
+    this.listenServer = await startWebSocketServer({
+      host: this.host,
+      port: this.listenPort
+    });
+
+    this.listenServer.on("connection", (newListener: WebSocket) => {
```
eggplantzzz, line 41: I wish I knew more about managing connections so I could comment on this :)

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
```diff
@@ -0,0 +1,184 @@
...
+        this.terminateIfNoConnections();
+      });
+
+      this.logToClients("Client connected", "connections");
+
+      this.clientSockets.push(newClient);
+    });
+  }
+
+  async ready() {
```
eggplantzzz, line 81: Would we ever want this to time out if something is not working properly?

packages/dashboard-message-bus/README.md
rkalis: Woops, that should be "dashboard provider"

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
rkalis: Sure, will do!

packages/dashboard-message-bus/lib/utils.ts
```diff
@@ -0,0 +1,141 @@
+import WebSocket, { ServerOptions } from "isomorphic-ws";
+import { Message, PortsConfig } from "./types";
+import any from "promise.any";
+import delay from "delay";
+import axios from "axios";
+
+any.shim();
```
eggplantzzz, line 7: Is `Promise.any` unnavailable?

rkalis: `Promise.any` is a Node 15 feature so we can't rely on it directly since we have to support Node 12.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any#browser_compatibility

gnidan (commented):
Had a few more thoughts after playing with this again just now.

Firstly, should go on the record saying that I **really** like the `truffle dashboard` flow. It's totally won me over. The @truffle/dashboard-provider way of doing things is just unnecessary clunkiness. Seriously good thinking @rkalis.

But playing with this, it occurred to me that there's a couple easy things we could do to further remove clunkiness – namely allowing dashboard configuration inside `truffle-config.js` and having syntactic sugar like `use: "dashboard"` or `dashboard: true` inside a network config instead of manual URL.

These two conveniences beg the question, though, "what about `truffle develop`?" Which got me thinking... seems like there's something _really_ slick there - see my comment describing a potential user flow. 

Maybe in our next sync, we could discuss all these? Or @rkalis if you're planning on thinking about this stuff while most of the rest of us are off doing our American holiday traditions, would love to hear your thoughts asynchronously so we waste less airtime when we're all back :) No rush though if you were planning on not doing anything while we're off for Thanksgiving.

Anyway, very cool stuff. I'm excited to demo @truffle/dashboard during Tuesday's Infura Developer Community call!

packages/dashboard-provider/manual-test/metacoin-truffle/truffle-config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  networks: {
+    browser: {
+      url: "http://localhost:5000/rpc",
+      network_id: "*",
+      skipDryRun: true,
+      networkCheckTimeout: 120000
+    }
+  }
+};
```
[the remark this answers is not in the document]
cds-amal: Maybe the provider could alert the user to add/switch MM to the truffle develop network along with further needed instructions.

gnidan:
Just noticed that there's an ugly error message if you have `truffle dashboard` running but the browser tab is closed... is there a way for us to detect this and give a friendlier message, like, "Hey! You closed the Truffle Dashboard page. Please open this URL in your browser: https://..."?

rkalis:
Interesting, what error did you get? We should definitely be able to make it better.

gnidan:
> Interesting, what error did you get? We should definitely be able to make it better.

Just the usual timeout, which makes sense. But the message-bus definitely knows when the browser isn't connected, so should be possible to pass this information along back to Truffle!

rkalis:
Right now we had it set up like this because we wanted to allow the user to open their dashboard tab after a request is already in flight. 

If we return an error instead, then we enforce that the dashboard needs to be connected before sending requests. And it also means that if you accidentally close the tab, you can't quickly re-open it and pick up where you left off.

That might be fine, but I'm not sure if it's necessarily a UX improvement in that sense.

gnidan:
Everything has to be complicated, doesn't it? OK I guess we'll have to think this through. Thanks!

packages/dashboard-provider/manual-test/metacoin-truffle/truffle-config.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  networks: {
+    browser: {
+      url: "http://localhost:5000/rpc",
+      network_id: "*",
+      skipDryRun: true,
+      networkCheckTimeout: 120000
+    }
+  }
```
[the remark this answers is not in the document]
rkalis: We could default to `"*"` but allow someone to specify another network ID if they want to have the added safety?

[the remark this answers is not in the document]
gnidan: yeah definitely we must support override, just thinking that this new flow means that `"*"` is probably a safe default!

.eslintrc.json
```diff
@@ -2,7 +2,9 @@
   "root": true,
   "env": {
     "es6": true,
-    "node": true
+    "node": true,
+    "jest": true,
+    "browser": true
```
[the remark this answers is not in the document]
rkalis: Done.

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
rkalis: Updated to pub-sub terminology.

packages/dashboard-provider/manual-test/metacoin-truffle/truffle-config.js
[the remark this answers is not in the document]
gnidan: Watching the reaction from my demo yesterday, I think `truffle dashboard` is going to make quite a big splash by itself.

Still want to see this, but I think we can wait until after initial release and prioritize alongside other enhancements.

packages/dashboard-provider/lib/DashboardProvider.ts
rkalis: We definitely could, but given that we don't want to promote the `DashboardProvider` as the "main" way to interact with the dashboard (and instead want to promote using `truffle dashboard` + RPC URL), I'm not sure if it adds too much. Wouldn't hurt to have of course, but it's maybe not worth the effort at this point.

packages/dashboard-provider/manual-test/metacoin-truffle/truffle-config.js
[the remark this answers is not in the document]
rkalis: Would definitely be cool to have this. At the same time, when using `truffle develop` it's much less interesting to use the Truffle Dashboard in its current state. Right now the main use of the dashboard is to deploy your contracts with metamask so you don't have to copy-paste your mnemonic. With `truffle develop` that is much less relevant since you ca just use the `truffle develop` mnemonic for deployments.

In later stages of the dashboard, like with the visual debugger I think it makes a lot more sense to have this `truffle develop` integration. So I agree that we should hold off on this for now, and revisit at that point.

packages/dashboard-provider/lib/DashboardProvider.ts
gnidan: Agreed

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
rkalis: For now we'll leave it like this.

packages/dashboard-provider/manual-test/metacoin-truffle/truffle-config.js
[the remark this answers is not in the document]
rkalis: We decided to take this one step further and add a builtin "dashboard" network that connects to the dashboard at its default location (or a different one as configured in the truffle-config, but not taking into account any CLI options). The network can be overridden in the `truffle-config.js` file if other options need to be set (e.g. networkChecktimeout).

rkalis:
To come back to Nick's comment about the network timeout error. We can update the error message on the migrations side to say something like "the request timed out, do you have your dashboard tab open?".

Hbutlercapone (approved):
Approve merge @truffle/dashboard #4035

packages/config/src/index.ts
```diff
@@ -209,6 +230,7 @@ class TruffleConfig {
 
     config.merge(staticConfig);
     config.merge(options);
+    config.addDefaultNetworks();
```
eggplantzzz, line 233: I think this should probably go in the constructor. Ugh! Unfortunately Truffle is not consistent about how it creates Truffle Config objects. Mostly I think it gets loaded using `Config.default()` or `Config.detect()`.

packages/config/src/index.ts
```diff
@@ -48,6 +48,27 @@ class TruffleConfig {
...
   }
 
+  private addDefaultNetworks(): TruffleConfig {
+    const dashboardHost = this.dashboard.host === "0.0.0.0"
+      ? "localhost"
+      : this.dashboard.host;
+
+    const dashboardNetwork = {
+      url: `http://${dashboardHost}:${this.dashboard.port}/rpc`,
+      network_id: "*",
```
eggplantzzz, line 58: Should some of these settings just go in `configDefaults.js`? Like the ones that aren't calculated? That is why we need this method, right?

packages/config/src/index.ts
eggplantzzz: So actually we might need it in `detect` as well.

packages/config/src/configDefaults.ts
```diff
@@ -33,6 +33,11 @@ export const getInitialConfig = ({
     build: null,
     resolver: null,
     artifactor: null,
+    dashboard: {
+      host: '0.0.0.0',
```
cds-amal, line 37: Just a drive by observation that `0.00.0` would bind to all resolved network addresses on a multi homed computer if used

packages/config/src/index.ts
rkalis: Yeah exactly, I don't think we can add the dynamic ones there easily. So since I couldn't have it *all* in the default values I figured it'd be cleaner to move it all out of that file, so at least it's in the same place

Would you suggest that we put some values in the defaults, like:

```js
  networks: {
    dashboard: {
      network_id: "*"
      ...
    }
  }
```

And then add "networks" to the deepCopy list so that they get merged with the user overrides?

packages/config/src/index.ts
rkalis: As far as `detect` goes, it calls `load`, so that case is already covered. But you're right that we should also include it in the default. Let me try moving it to the constructor.

packages/config/src/configDefaults.ts
rkalis: Hmm right. Do you think that would be a problem, would it be better to just default to `localhost` then and allow someone to specify 0.0.0.0 if they want?

cds-amal: I think defaulting to `localhost` and documenting the options for users to make an educated choice is better all around.

rkalis: Either way is fine by me. Anyone else have a different opinion? Otherwise we'll go with your suggestion Amal.

Hbutlercapone (approved):
@truffle/dashboard #4035

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    const message = createMessage("provider", req.body);
+    sendAndAwait(this.socket, message)
+      .then(response => res.json(response.payload))
+      .catch(next);
+  }
+
+  private async startMessageBus() {
+    const messageBusPublishPort = await getPort({ host: this.host });
```
eggplantzzz, line 112: This is not something to hold this PR back but I'm curious, what does passing it the `host` do? I read the docs and I'm still not sure I understand what this does.

rkalis: If there are multiple local addresses, this allows you to specify for which address you want to get an available port.

packages/dashboard-provider/test/MockDashboard.ts
```diff
@@ -0,0 +1,71 @@
...
+  Message
+} from "@truffle/dashboard-message-bus";
+import { JSONRPCRequestPayload } from "ethereum-protocol";
+import { promisify } from "util";
+import WebSocket from "ws";
+import Ganache from "ganache-core";
+
+// TODO: This mock dashboard was copy-pasted from the dashboard-provider tests
+// We should figure out whether we want to make this DRYer
+export default class MockDashboard {
```
gnidan, line 15: But... these are the `dashboard-provider` tests! 😉 

See my analogous comment over in `packages/dashboard/test/MockDashboard.ts` for more explanation, but could you replace the `TODO` with maybe this:

```suggestion
// NOTE This mock dashboard was copy-pasted from the dashboard tests
export default class MockDashboard {
```

packages/dashboard-provider/manual-test/metacoin-hardhat/contracts/MetaCoin.sol
```diff
@@ -0,0 +1,43 @@
...
+pragma solidity >=0.4.25 <0.9.0;
+
+import "./ConvertLib.sol";
+
+// This is just a simple example of a coin-like contract.
+// It is not standards compatible and cannot be expected to talk to other
+// coin/token contracts. If you want to create a standards-compliant
+// token, see: https://github.com/ConsenSys/Tokens. Cheers!
+
+contract MetaCoin {
```
gnidan, line 12: the truffle repo is a hungry beast that demands we continue to define `MetaCoin.sol` in new places all the time.

praise be the demands of the code

(sorry it's getting late here)

packages/dashboard-provider/manual-test/manual-test.ts
```diff
@@ -0,0 +1,165 @@
...
+    method: "net_version",
+    params: [],
+    id: 1
+  });
+
+  const chainId = chainIdResult?.result ?? 1;
+
+  // Code taken from https://docs.metamask.io/guide/signing-data.html#sign-typed-data-v4
+
+  const msgParams_v4 = JSON.stringify({
```
gnidan, line 82: cc @kevinweaver @kevinbluer @MicaiahReid I entrust with you three the knowledge that there exists this example of doing EIP-712 inside the Truffle tests. 

(okay, and anyone else that sees this... let's remember that we have this example of how to do EIP-712! like, we may want to reference this later!)

packages/dashboard-provider/lib/types.ts
```diff
@@ -0,0 +1,19 @@
+export interface DashboardProviderOptions {
```
gnidan, line 1: Could we move this interface to be inside `./lib/DashboardProvider.ts`? No reason to move the useful documentation info further away from the implementation ;)

gnidan (changes requested):
Have I told you that I like your code style @rkalis? Thanks for all your work on this. I think I'm only requesting changes for a couple things, and I've tried to phrase such cases explicitly. The rest of the comments are generally questions or to encourage a bit of discussion (apart from snark and quick suggestions, of course ;)

Since this is a lot to review, I've tried to call out things that specific people on the team might want to be aware of, to try to get the right eyes on the right things. Additionally, should note that this review also only looks at implementation, not any kind of acceptance/blackbox review. (I've done a bunch of acceptance testing previously, and I've no doubt that I'll be doing it again soon)

Overall, I'm quite comfortable merging this without significant revision... even just reading the code gets me amped about this! Really, thank you for this very significant contribution to this project.

packages/config/src/configDefaults.ts
```diff
@@ -155,6 +168,22 @@ export const configProps = ({
...
+
+          config = {
+            network_id: "*",
+            networkCheckTimeout: 120000,
+            ...userOverrides,
+            url: `http://${host}:${port}/rpc`,
+            skipDryRun: true
+          };
+        }
+
```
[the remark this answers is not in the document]
rkalis: Yeah after having implemented it in a few different spots we settled on this being the "least bad" place to do it.

packages/dashboard-provider/test/MockDashboard.ts
rkalis: Good point 😆

packages/dashboard-message-bus/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "@types/jest": "^27.0.3",
+    "@types/node": "^12.0.0",
+    "@types/promise.any": "^2.0.0",
+    "@types/web3": "^1.2.2",
+    "jest": "^27.4.5",
+    "ts-jest": "^27.1.2",
+    "typescript": "^4.5.4"
+  },
+  "dependencies": {
+    "@types/ws": "^7.2.0",
```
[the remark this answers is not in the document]
gnidan: @rkalis has confirmed that this is a regular dep. on purpose (actual best practice aside, which remains nebulous). Good enough for me!

packages/dashboard/README.md
```diff
@@ -0,0 +1,98 @@
...
+### Logging
+
+When setting the dashboard server's `verbose` option to `true`, debug output is logged in the following "debug" namespaces
+
+- `dashboard-message-bus:connections` - logs connections and disconnections of message bus publishers and subscribers
+- `dashboard-message-bus:requests` - logs requests that get sent from publishers to subscribers
+- `dashboard-message-bus:responses` - logs responses sent back from subscribers to publishers
+- `dashboard-message-bus:errors` - logs errors that occur in the message bus
+
+## Development
```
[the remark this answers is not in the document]
eggplantzzz: I would say leave it here for now. It is not that big of a file.

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
```diff
@@ -0,0 +1,194 @@
...
+
+      this.publishers.push(newPublisher);
+    });
+  }
+
+  async ready() {
+    if (this.subscribers.length > 0) return;
+    await delay(1000);
+    await this.ready();
+  }
```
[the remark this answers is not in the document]
tenthirtyone: `while` blocks the event loop. `clearInterval` can offer a way to poll without blocking or affecting the callstack in this way.

If it has already been explained wave me off - I'm late to the review. But what is the need to do the `delay(1000)`? I would expect it to imply there is a race condition?

tenthirtyone:
@rkalis I like what you did. I think your message bus is a really smart way to take the STUN/TURN concept/pattern and apply it here with multiple connections. I took some time to poke at my favorite nits and I think you've done a nice job making it robust. I didn't spend much time on the react code because a quick once over looked organized and well structured. 

I did find one error when the user cancels the transaction in metamask for `yarn test` in dashboard-provider. I just rejected the first transaction and the console showed an unhandled error. It looks like you caught this everywhere else I tested it. 

Nice work. I definitely understand why you will be missed! 

All that said, I'm still new to dashboard. I don't have the full context on the implementation details. I will happily second the PR if the unit tests are filled out with more integration examples. Otherwise I'll defer to someone more senior that's been a larger part of the dashboard initiative and can :+1:  from just the code review.

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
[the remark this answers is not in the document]
rkalis: So the message bus is "ready" when it has at least 1 subsciber to send things to. Essentially this function polls every second to see if it's "ready" now.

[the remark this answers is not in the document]
rkalis: As for the event loop blocking. I *think* that blocking the event loop is what we want to do here. This `ready()` promise is used to block the sending of requests until it can send them somewhere.

packages/core/lib/utils/utils.js
```diff
@@ -16,4 +16,19 @@ const extractFlags = inputArguments => {
...
+    return Config.detect(options);
+  } catch (error) {
+    // Suppress error when truffle can't find a config
+    if (error.message === "Could not find suitable configuration file.") {
+      return Config.default();
+    } else {
+      throw error;
+    }
+  }
+};
```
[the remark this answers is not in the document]
rkalis: I got this code from `truffle version` but in the `run` / `meta` refactoring, my change to use it there was reverted. I added it back so that this function is now used in both `dashboard` and `version`.

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
```diff
@@ -0,0 +1,194 @@
...
+    this.subscribeServer.on("connection", (newSubscriber: WebSocket) => {
+      newSubscriber.on("close", () => {
+        this.logToPublishers("Subscriber disconnected", "connections");
+
+        this.subscribers = this.subscribers.filter(
+          subscriber => subscriber !== newSubscriber
+        );
+
+        this.terminateIfNoConnections();
+      });
```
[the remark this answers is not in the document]
rkalis: As my proposed alternative I refactored the add/remove logic into separate methods so that it's clearer what is happening in this callback without having to add extra comments. I *did* add a jsdoc comment to these new functions with some notes.

packages/dashboard-message-bus/lib/utils.ts
```diff
@@ -0,0 +1,157 @@
...
+  const buffer = Buffer.from(base64, "base64");
+  const stringifiedJson = buffer.toString("utf8");
+  const json = JSON.parse(stringifiedJson);
+
+  return json;
+};
+
+/**
+ * Starts a websocket server and waits for it to be opened
+ * @dev If you want to attach a listener *before* opening, so not use this
```
eggplantzzz, line 34: What does this comment mean to say?

rkalis: Oh, so > do*

rkalis: Expanded the comment a bit further.

packages/dashboard-message-bus/lib/DashboardMessageBus.ts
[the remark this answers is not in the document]
gnidan: I wonder if this timeout is something that should be configurable. What's the downside of blocking here? I can't really think of one, other than the initial one second might be slow? But I doubt it.

Anyway, "able to actually send a request somewhere" is a decent definition of "ready" if you ask me :)

packages/dashboard-message-bus/lib/utils.ts
```diff
@@ -0,0 +1,144 @@
...
+  sockets: WebSocket[],
+  message: Message
+) => {
+  const encodedMessage = jsonToBase64(message);
+  sockets.forEach(socket => {
+    socket.send(encodedMessage);
+  });
+};
+
+export const broadcastAndAwaitFirst = async (
```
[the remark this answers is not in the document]
gnidan: Chance to apply pedantry has expired

packages/dashboard/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "postcss": "^7",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
+    "react-json-view": "^1.21.3",
+    "react-scripts": "^4.0.0",
+    "tailwindcss": "npm:@tailwindcss/postcss7-compat",
+    "ts-jest": "^27.1.2",
+    "ts-node": "^10.4.0",
+    "typescript": "^4.1.4"
+  },
```
gnidan, line 76: we'll need one of these (i learned the hard way)

```suggestion
  },
  "publishConfig": {
    "access": "public"
  },
```

packages/config/src/configDefaults.ts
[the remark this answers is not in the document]
kevinweaver: ❓ Do we already have documentation anywhere about the implications of the "test" and "develop" special networks?  It's unlikely that someone would happen to name their network "dashboard" and then hit issues with this, but if we do have any docs about special networks, it may be worth surfacing this new one there. (not blocking)

packages/dashboard-message-bus/lib/utils.ts
```diff
@@ -0,0 +1,158 @@
...
+ * Convert the base64 representation of a JS object or value to its JS representation
+ * @dev This is the reverse of `jsonToBase64` and is not expected to work with other base64 formats
+ */
+export const base64ToJson = (base64: string) => {
+  const buffer = Buffer.from(base64, "base64");
+  const stringifiedJson = buffer.toString("utf8");
+  const json = JSON.parse(stringifiedJson);
+
+  return json;
+};
```
kevinweaver, line 30: If we do `jsonToBase64` and `base64ToJson` elsewhere in truffle, it might make sense to push them up into a higher level truffle util. From a quick cursory glance, I found [one here](https://github.com/trufflesuite/truffle/blob/f83191a654d29750c1c3d84e354085a3b33b646c/packages/environment/develop.js#L34) but if that's it, we could also wait for a rule of 3. (not blocking)

packages/dashboard/.gitignore
```diff
@@ -0,0 +1,24 @@
...
+# production
+/build
+/dist
+
+# misc
+.DS_Store
+.env.local
+.env.development.local
+.env.test.local
+.env.production.local
```
kevinweaver, line 20: ⚠️ Do you also mind adding `.env` to this?

packages/dashboard/.gitignore
```diff
@@ -0,0 +1,24 @@
...
+# testing
+/coverage
+
+# production
+/build
+/dist
+
+# misc
+.DS_Store
+.env.local
```
kevinweaver, line 17: ⚠️ Do you also mind adding `.env` to this? (blocking unless there's a reason to allow it)
```suggestion
.env.production.local
.env
```

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -0,0 +1,170 @@
...
+  host?: string;
+
+  /** Boolean indicating whether the POST /rpc endpoint should be exposed (default: true) */
+  rpc?: boolean;
+
+  /** Boolean indicating whether debug output should be logged (default: false) */
+  verbose?: boolean;
+
+  /** Boolean indicating whether whether starting the DashboardServer should automatically open the dashboard (default: true) */
+  autoOpen?: boolean;
```
kevinweaver, line 32: 😍 Thanks for the comments!

packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
```diff
@@ -0,0 +1,121 @@
...
+  };
+
+  const header = request.payload.method;
+
+  const body = <div>{formatDashboardProviderRequestParameters(request)}</div>;
+
+  const footer = (
+    <div className="flex justify-start items-center gap-2">
+      <Button onClick={process} text="PROCESS" />
+      <Button onClick={reject} text="REJECT" />
```
kevinweaver, line 110: Non-blocking nit to open discussion: I don't typically love when UIs over-use all caps as it's generally a heavy handed way of attention grabbing that is often better addressed with more nuance using varying font weights.  ALL CAPS FEELS LIKE I'M BEING YELLED AT

kevinweaver (commented):
Some non-blocking comments. I don't want to keep pushing this back with random changes, so it might be worth gathering a list of any final suggestions, documenting them even just in a comment for whoever picks this up next after merge

packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
```diff
@@ -0,0 +1,121 @@
...
+          </div>
+        );
+      }
+      case "personal_sign": {
+        const [message, from] = request.payload.params;
+        return (
+          <div>
+            <div>ACCOUNT: {from}</div>
+            <div className="flex gap-2">
+              <div>MESSAGE:</div>
```
kevinweaver, line 90: Non-blocking nit to open discussion: I don't typically love when UIs over-use all caps as it's generally a heavy handed way of attention grabbing that is often better addressed with more nuance using varying font weights.  ALL CAPS FEELS LIKE I'M BEING YELLED AT

rkalis: We'll update the "body text" to regular case. And we'll leave the buttons as allcaps, but we'll use CSS to make it allcaps rather than actually writing an allcaps string.

Hbutlercapone (approved):
Merged @truffle/dashboard #4035

gnidan:
@Hbutlercapone if you're in a hurry, you can install `npm install -g truffle@dashboard` :)

Hbutlercapone:
I am not just coding

Hbutlercapone:
dabbling and catching up on stuffs

packages/config/src/configDefaults.ts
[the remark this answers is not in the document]
gnidan: I believe there's no documentation about these special networks. We definitely should address this in our user-facing docs, and we may even want to try and collect knowledge into developer-facing docs (like how+where each of these networks is implemented, known gotchas, etc.)

packages/core/lib/utils/utils.js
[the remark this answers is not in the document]
gnidan: Opened issue here: https://github.com/trufflesuite/truffle/issues/4714

packages/dashboard/test/DashboardServer.test.ts
```diff
@@ -0,0 +1,144 @@
...
+import Web3 from "web3";
+import { getMessageBusPorts } from "@truffle/dashboard-message-bus";
+import MockDashboard from "./MockDashboard";
+import { DashboardServer } from "../lib";
+
+jest.setTimeout(200000);
+
+// TODO: These tests were copy-pasted from the browser-provider tests
+// We should figure out whether we want to make this DRYer
+describe("DashboardServer", () => {
```
gnidan, line 12: Is there a good way to add automated tests for reverted transactions and/or rejected requests? Maybe I'm missing these, but I didn't see them when I took a look just now.

gnidan (approved):
I've just gone over this again, and everything I had outstanding is resolved. I've opened a couple issues for a couple outstanding enhancements mentioned in comments (like WalletConnect and web UI network switching).

I've got one last open question, but here's a +1  from me! One down, one to go 😄 

Thanks again @rkalis 🙏

packages/dashboard/test/DashboardServer.test.ts
rkalis: We don't have any reverted tx tests. Probably what we'd want to do is add some contract deployments / interactions to the automated testing. And then with those contract interactions it should be easy to include a couple of interactions that we expect to revert.

packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
rkalis: Updated.

gnidan:
Better hit merge before we get more conflicts :)

PR #4037, opened 2021-05-17
Dependency update: Update to the newer @ensdomains/ensjs

eggplantzzz:
The package ethereum-ens was rewritten and renamed @ensdomains/ensjs. This PR switches out the old for the new. Thank you.


packages/contract-tests/test/lib/utils/ens.js
```diff
@@ -48,15 +52,31 @@ describe("convertENSNames", () => {
   });
 
   it("converts ens names in address fields to addresses", async () => {
-    let result = await ens.convertENSNames(args, methodABI, Web3);
-    assert(result[0] === "0x123");
+    let result = await ens.convertENSNames({
```
haltman-at, line 55: Should these tests set `networkId`?

packages/contract-tests/test/lib/utils/ens.js
```diff
@@ -26,18 +26,22 @@ const methodABI = {
   type: "function"
 };
 const args = ["my.ens.name", 555, { from: "the.other.name" }];
+const ensSettings = {
+  enabled: true,
+  networkId: 1
+};
 
 describe("convertENSNames", () => {
```
haltman-at, line 34: This is likely too much work for this PR and should be saved for another time, but I'm wondering, would it be possible to change these tests to not use mocks but rather to use `@truffle/deployer` to set up a real ENS instance on a testnet?  (This is how I'm testing the ENS support in `@truffle/encoder` in #3974, and to my mind it seems like a better test than using mocks.)

packages/deployer/ens.js
```diff
@@ -1,23 +1,22 @@
...
 
   determineENSRegistryAddress() {
-    if (this.ensSettings.registryAddress) {
-      return this.ensSettings.registryAddress;
-    } else if (
-      this.ensjs &&
-      this.ensjs.registryPromise._rejectionHandler0._address
-    ) {
-      return this.ensjs.registryPromise._rejectionHandler0._address;
+    if (this.ens.registryAddress) {
```
haltman-at, line 16: I notice this function doesn't check for a `networkId` and use that.  I'm assuming that's deliberate, and that doing things by network ID is done elsewhere?   Or what?

Edit: Oh, I see, it's only used for determining the address of a registry address to deploy to (if I'm understanding correctly?).  Dunno if it's worth renaming this to make that clearer.

packages/deployer/ens.js
```diff
@@ -32,63 +31,51 @@ class ENS {
...
+      const resolvedAddress = await this.ensjs.name(name).getAddress("ETH");
       return { resolvedAddress };
-    } catch (error) {
-      if (error.message !== "ENS name not found") throw error;
-      const PublicResolverArtifact = require("@ensdomains/resolver")
-        .PublicResolver;
-      const PublicResolver = contract(PublicResolverArtifact);
-      PublicResolver.setProvider(this.provider);
+    }
+    // deploy a resolver if
```
haltman-at, line 48: Comment seems incomplete?

packages/contract-tests/test/lib/utils/ens.js
```diff
@@ -26,18 +26,22 @@ const methodABI = {
   type: "function"
 };
 const args = ["my.ens.name", 555, { from: "the.other.name" }];
+const ensSettings = {
```
haltman-at, line 29: So having looked through the rest of the code I'm a bit confused.  This `ensSettings` has a `networkId` field, but everywhere else, it seems like `networkId` is separate from `ensSettings` or `ens`, and isn't pulled from it, and any potential `ensSettings.networkId` or `ens.networkId` is ignored.  Meanwhile things do look inside `ensSettings` or `ens` for `registryAddress`, which isn't here.  I'm confused as to what's going on here and why this doesn't seem to match the rest of the code.  Am I misunderstanding?

haltman-at (commented):
Seems mostly OK, a few things I'm confused about though.

packages/contract-tests/test/lib/utils/ens.js
eggplantzzz: In this case I think "why not?". This basically will test the case where a user is connected to mainnet. Right now, if the user doesn't explicitly specify a registry address, it will use the network id to determine the address. It will fail if the address is not specified AND there is no network id. In the normal flow there will always be a network id set by Truffle. Does that make sense?

packages/contract-tests/test/lib/utils/ens.js
eggplantzzz: Good catch, I think this was accidentally left in from a previous iteration of this. `ensSettings` is really just a reference to `config.options.ens` which should only contain `registryAddress` and `enabled`. I'll go ahead and update these.

packages/contract-tests/test/lib/utils/ens.js
eggplantzzz: I figured those kinds of tests would be more appropriate in deployer rather than in contract-tests since they don't use a "real ENS instance". But that sounds like it might be good over there.

haltman-at: Yeah I mean probably don't do this on this PR, would be better to just get this merged I think, just noting it for later.

My thought is that the tests make sense here, because -- having tested the deployer elsewhere -- here you *use* the deployer to test this.  (Same as I'm doing in `encoder`.)  I don't mean an instance of the deployer *or* ensjs; I mean an instance of the ENS contracts, the ENS registry.  You use the deployer's functionality where it can set up an ENS registry on a testnet.  Then, instead of mocking, you can test using an actual ENS registry that you set up. :)

That seems like a better test to me, but as noted that probably shouldn't actually go in this PR.

packages/contract-tests/test/lib/utils/ens.js
haltman-at: But these tests don't seem to set either a network ID *or* a registry address.  I'm confused.  There's no `networkId` argument passed, and the `ensSettings` set above don't include a registry address either.  (They *do* include a `networkId`, but as best I can tell from the code that's ignored.)

Are these tests supposed to use network ID, or are they supposed to use registry address?  Because it seems like in neither case is the information getting passed in.

eggplantzzz: Oh sorry, I got confused about which tests we were speaking of here. These tests stub out the `getNewENSJS` method. When you use the constructor (`require("@ensdomains/ensjs").default`) you need to feed it either a network id or address; and address takes precedence. In this case we won't instantiate and so don't need it.

packages/deployer/ens.js
eggplantzzz: Which name specifically are you saying is a candidate for a rename?

haltman-at: `determineENSRegistryAddress`

packages/contract-tests/test/lib/utils/ens.js
haltman-at: I think it would be better if it matched the real interface as closely as possible.  But I suppose it's not essential if you think it's OK.

packages/deployer/ens.js
haltman-at: Oh, @eggplantzzz has explained, I misunderstood: it doesn't need to check for the network ID because that was already done when setting up `ensjs`.  So in that case you'll get it from the `ensjs`.

packages/contract-tests/test/lib/utils/ens.js
haltman-at: Oh, OK, seems this is fixed now.

haltman-at (approved):
OK, @eggplantzzz says there's another thing that needs fixing here, but I'll leave that to him!

Issue #3822, opened 2021-02-20
Txlog `callexternal` actions of `message` kind should have returndata field 


haltman-at:
## Issue

Txlog `callexternal` actions of `message` kind should have a field `returndata` with the hex-encoded data that was returned.  This is important because as of 0.7.6, Solidity makes it much easier to return such data.

PR #4038, opened 2021-05-17
Internal improvement: Add `returnData` to txlog

haltman-at:
Addresses #3822.  However I called the field `returnData` rather than `returndata`.

In more detail:

The original version of txlog would track return values for functions, as well as revert messages, but it wouldn't track raw binary returned from a fallback function.  That's because the ability to do this in Solidity was only introduced in 0.7.6!  So we didn't track it as it wasn't something that should ordinarily happen.

So, now we do that.  This didn't require any big code changes, because for fallback functions that can return data, `codec` already includes the raw data as a possible return decoding.  So we do it the same way we do `returnValues`, just instead of looking for the `function` call type and the `return` decoding type, we look for the `message` call type and the `returnmessage` decoding type.

The one thing I worry about here is that this approach will *only* work for fallback functions that are supposed to be able for return data, not for anything else.  We might want to make sure this works more broadly, which would require doing things slightly differently.  Not that different though; this still wouldn't be that hard.  Let me know if you think that's worth doing.

**Edit**: Actually I did go back and redo it this more robust way, see comment below.

### DIFF
--- a/packages/debugger/lib/txlog/actions/index.js
+++ b/packages/debugger/lib/txlog/actions/index.js
@@ -134,12 +134,13 @@ export function instantCreate(
 }
 
 export const EXTERNAL_RETURN = "TXLOG_EXTERNAL_RETURN";
-export function externalReturn(pointer, newPointer, decodings) {
+export function externalReturn(pointer, newPointer, decodings, returnData) {
   return {
     type: EXTERNAL_RETURN,
     pointer,
     newPointer,
-    decodings
+    decodings,
+    returnData
   };
 }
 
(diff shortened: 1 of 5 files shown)


haltman-at:
Yeah actually I decided to go back and redo it in this more robust way.  Now there's a new field on the `EXTERNAL_RETURN` action, namely, `returnData`, that holds the raw data, and we get the raw data from there rather than from the decoding.  We populate this field with a new selector, `txlog.current.returnData`, which just gets it from `evm.current.step.returnValue`.  So still pretty easy.

gnidan (approved):
Looks good!

PR #4066, opened 2021-05-25
Convert @truffle/expect to TypeScript

gnidan:
**:information_source: This is a minor release for @truffle/expect**

This PR converts @truffle/expect in order to offer type-safe guarantees of the post-conditions of using `expect.options`, etc.

This also adds `expect.has(options: {}, key: string)`, as a more simple form of `expect.options(options: {}, key: string[])`.

**Note**: for `expect.one()`, which is designed for either-or cases, the type guard asserts _exactly one_ key so as to facilitate likely usage (`expect.one(options, ["a", "b"]); const x = "a" in options ? options.a : options.b`), and also because taking the combination of all allowable expected values is just kind of excessive work for the type-checker. We may want to change the runtime behavior here, assuming it won't cause hard-to-fix breakages.

Post conditions for the asserted types here are tested using [tsd](https://github.com/SamVerschueren/tsd)

### DIFF
--- a/packages/expect/src/index.ts
+++ b/packages/expect/src/index.ts
@@ -1,35 +1,72 @@
-const Expect = {
-  options(options, expected_keys) {
-    expected_keys.forEach(key => {
-      if (options[key] == null) {
-        throw new Error(`Expected parameter '${key}' not passed to function.`);
-      }
-    });
-  },
+/**
+ * Object type O that includes non-nullable keys K
+ */
+export type Has<O extends {}, K extends string> = O &
+  Required<NonNullable<Pick<O, K & keyof O>>>;
 
-  one(options, expected_keys) {
-    const found: number[] = [];
+/**
+ * Object type O that includes exactly one non-nullable value among keys K
+ */
+export type HasOne<O extends {}, K extends string> = O &
+  {
+    [N in K]: Has<O, N>;
+  }[K];
 
-    expected_keys.forEach(key => {
-      if (options[key] != null) {
-        found.push(1);
-      } else {
-        found.push(0);
-      }
-    });
+/**
+ * Asserts at runtime that `options` contains `key`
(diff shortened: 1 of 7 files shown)


packages/expect/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+}
+
+/**
+ * Asserts at runtime that `options` contains at least one of `expectedKeys`
+ *
+ * Post-condition: this narrows type of `options` to include _exactly one_ of
+ * `expectedKeys`, even though at runtime this accepts more than one key.
+ */
+export function one<O extends {}, K extends string>(
```
eggplantzzz, line 46: Why do we need `K extends string`? Couldn't you just type `expectedKeys` as something like `string[]`?

eggplantzzz: I now expect the answer to this is similar to what we were talking about earlier. `K` provides more information to the TS compiler for later. Tell me if I'm wrong.

gnidan: It's necessary because we concern ourselves with actual key values, not just the arbitrary `string` type. So `one(options, ["size"])` becomes `one<O, "size">`, which means that later we can do `options.size` precisely.

gnidan: You're exactly right @eggplantzzz

packages/expect/src/index.ts
```diff
@@ -0,0 +1,72 @@
...
+  options: O,
+  expectedKeys: K[]
+): asserts options is HasOne<O, K> {
+  let found = false;
+
+  for (const key of expectedKeys) {
+    try {
+      has(options, key);
```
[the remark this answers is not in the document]
gnidan: I like the latter! Certainly because it seems wise to account for the specific error only.

PR #4068, opened 2021-05-27
Tezos compiler

bergarces:
Summary:
- Adds support for LIGO and Michelson compilations.
- Sets framework to add SmartPy compilation when finished.

Changes:
- @truffle/artifactor: Minor change to stop relying on `contract_name` being present for debugging information. It will work as well if `contractName` is present.
- @truffle/codec: This seems to be called only by debugger, so no apparent need for Tezos for now.
- @truffle/compile-common: Add a compiler that can be shared amongst all Tezos compilers to avoid code repetition. Tezos does not have imports, so it's simpler.
- @truffle/compile-common: ProfilerConfig properties should be optional, but since existing compilers are not using Typescript it wasn't apparent.
- @truffle/compile-common: `CompiledContract` type now can be `EvmCompiledContract` or `TezosCompiledContract`.
- @truffle/compile-common: `NewToLegacy` shim is only needed for evm, so a method that discriminates between architectures is used instead.
- @truffle/compile-common: `LegacyToNew` shim can safely assume evm is the architecture.
- @truffle/compile-ligo and @truffle/compile-michelson: Added packages, which implement a CompileStrategy and make use of the new compiler in @truffle/compile-common for Tezos.
- @truffle/contract-schema: Added `architecture`, `michelson` and `initialStorage` fields.


haltman-at:
Looks like there might some timeouts that need upping?

eggplantzzz:
Closing this for PR list maintenance.

PR #4069, opened 2021-05-27
Make spinners for 'unbox' repect the `--quiet` option

eggplantzzz:
**Note**: This is a breaking change for @truffle/events

This PR updates the EventManager to accept and pass a `quiet` parameter instead of `muteLogging`. `quiet` is used in the rest of Truffle and it will be up to the handlers themselves to decide what `quiet === true` means. In the case of unboxing, if this option is passed then it will not display any of the spinners and show no output.

### DIFF
--- a/packages/events/defaultSubscribers/unbox.js
+++ b/packages/events/defaultSubscribers/unbox.js
@@ -22,52 +22,62 @@ module.exports = {
   handlers: {
     "unbox:start": [
       function () {
+        if (this.quiet) return;
         this.logger.log(`${OS.EOL}Starting unbox...`);
         this.logger.log(`=================${OS.EOL}`);
       },
     ],
     "unbox:preparingToDownload:start": [
       function () {
+        if (this.quiet) return;
         this.spinner = this.ora("Preparing to download box").start();
       },
     ],
     "unbox:preparingToDownload:succeed": [
       function () {
+        if (this.quiet) return;
         this.spinner.succeed();
       },
     ],
     "unbox:downloadingBox:start": [
       function () {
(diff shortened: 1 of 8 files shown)


packages/events/defaultSubscribers/unbox.js
```diff
@@ -22,52 +22,62 @@ module.exports = {
...
     "unbox:preparingToDownload:succeed": [
       function () {
+        if (this.quiet) return;
         this.spinner.succeed();
       },
     ],
     "unbox:downloadingBox:start": [
       function () {
+        if (this.quiet) return;
         this.spinner = this.ora("Downloading").start();
```
gnidan, line 45: hey, why don't you just wrap `this.ora(...).start()`? so that way you don't need to `if()` statement all over the place

gnidan (approved):
But I'll just give you the approve so we can get this into this week's release. You can decide if you wanna refactor the `if()` statements :)

packages/events/defaultSubscribers/unbox.js
eggplantzzz: we'll have to wrap everything I think - then pass the action like "start", "fail", and "succeed" I think

gnidan: can just make a class, no? with `start`/`fail`/etc. methods

eggplantzzz: At that point we're just pushing the problem somewhere else - I was imagining a function that had a switch statement. Then we only have to have one guard at the beginning of it

Issue #4071, opened 2021-05-27
Have etherscan source-fetcher include etherscan header for line numbering purposes


haltman-at:
## Issue

The Etherscan website, when displaying sources, prepends a header comment; however it doesn't do this when fetching the actual sources via the API.  But this means that when we use our downloaded sources in the debugger, the line numbers are off compared to the website.  (Thanks to @0xfoobar for noticing this.)  So @gnidan has suggested that the Etherscan source-fetcher should add this header back in so that things match.

## Environment

* Truffle version (`truffle version`): 5.3.7

PR #4072, opened 2021-05-28
Bug fix: Add header to single sources downloaded from Etherscan

haltman-at:
Addresses #4071.  See that issue for why I'm doing this.

The header is only added in the single-source Solidity case, because that's the only case where Etherscan adds a header on the website.

I tested this manually and with this change the line numbers now match up.

Note that the header we're using contains a generic placeholder in place of the date, since there doesn't seem to be any way for us to get the submission date from the API (and even if we could I'm not sure it would be so important to do!).

### DIFF
--- a/packages/source-fetcher/lib/etherscan.ts
+++ b/packages/source-fetcher/lib/etherscan.ts
@@ -14,6 +14,12 @@ import {
 } from "./common";
 import axios from "axios";
 
+const etherscanCommentHeader = `/**
+ *Submitted for verification at Etherscan.io on 20XX-XX-XX
+*/
+
+`; //note we include that final newline
+
 //this looks awkward but the TS docs actually suggest this :P
 const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
   implements Fetcher {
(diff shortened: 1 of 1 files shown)


packages/source-fetcher/lib/etherscan.ts
```diff
@@ -167,7 +173,10 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
     return {
       contractName: result.ContractName,
       sources: {
-        [filename]: result.SourceCode
+        //we prepend this header comment so that line numbers in the debugger
+        //will match up with what's displayed on the website; note that other
+        //cases don't display a similar header on the website
+        [filename]: etherscanCommentHeader + result.SourceCode
```
gnidan, line 179: Note that this is probably better style:
```suggestion
        [filename]: `${etherscanCommentHeader}${result.SourceCode}`
```

gnidan (approved):

packages/source-fetcher/lib/etherscan.ts
haltman-at: Eh, for a simple concatenation like this I'm not going to bother unless you really think it's important... which evidently you don't as you approved. :)

Issue #3992, opened 2021-04-21
Truffle should not be stupid about gas + gas price


gnidan:
Right now, Truffle assumes per-network that there will be one constant gas price (with some default, IIRC), and then may or may not compute gas limit on a per-transaction basis (although a common pattern is to specify a constant gas limit for every transaction it makes to that network).

This sort of approach is absolutely terrible. Truffle should figure out the gas using the normal gas estimation techniques (maybe this simply means discouraging configuring `gas`), and also Truffle should determine reasonable gas price **for a transaction of the estimated size**, given current network conditions.

Raising this issue as a follow-up from various issues related to mainnet deployment, most recently #3888.

So, this issue represents two problems:
- [x] Truffle should not attempt to provide any default gas price. (**Implemented in #4073**)
- [x] Truffle should use the result of `eth_estimateGas` (times some buffer) as the default gas limit for any transaction it makes. (**In progress via #4296**)

gnidan:
So in doing #4296, @fainashalts discovered that the naïve strategy of doing an `eth_estimateGas` is insufficient for transactions that revert: in such situations, `eth_estimateGas` does not provide a quantity in response (it instead errors). This means that Truffle won't be able to proceed with sending the transaction, since it won't have all the information it needs to be able to do so. We need Truffle to be able to send transactions, even if it knows those transactions will revert, because users must be able to inspect such reverted transactions (e.g. via stacktrace inspection, etc.). As a consequence of this, we'll need some additional logic to produce a gas limit value when the estimate cannot be determined via `eth_estimateGas`.

Best solution for this remains undetermined, but we've uncovered a few options (all specifically relating to the case where `eth_estimateGas` fails due to revert):
a. Use the block limit (or some value based on the block limit, e.g. `blockLimit / 2`) as the default transaction gas limit
b. Do a binary search to find the boundary for gas limit between the transaction's failing for "out of gas" and the transaction's failing for "revert", and use that value when sending the transaction to the network.
c. Petition to modify the JSON-RPC and all client implementations so as to ensure `eth_estimateGas` reports a sensible gas limit value even for transactions that revert

These all seem not-ideal; currently looking for an option (d)

eggplantzzz:
Can we close this since #4296 was merged and released?

gnidan:
Yep, we went with option (a) above!

PR #4073, opened 2021-05-28
Remove default gasPrice

bergarces:
Partially addresses https://github.com/trufflesuite/truffle/issues/3992

Removes default gasPrice when none is specified as part of the network config.

It will use the last few blocks median gas price instead, as per web3js documentation: https://web3js.readthedocs.io/en/v1.2.0/web3-eth.html#getgasprice


gnidan (approved):
Let's do it (well maybe get a third opinion here first :)

eggplantzzz (approved):

PR #4078, opened 2021-05-30
Add truffle init next steps

Shih-Yu:
@cds-amal and @kevinbluer is this what you were looking for?


kevinbluer:
Thanks so much for the contribution @Shih-Yu! We're actually going to go with https://github.com/trufflesuite/truffle/pull/4082 as it's a little more verbose in terms of output and thus makes it easier for users to copy / paste the commands to try them out. 

That said, we super appreciate the contribution and let me know if you'd like assistance with finding some additional issues to work on 👍

Issue #4041, opened 2021-05-18
Truffle Init - No Tests?


JimLynchCodes:
Hi, just wondering why there are no tests in the scaffolded project when running `truffle init`...

Can I make a PR adding this? Where in the main truffle source code should I put this file?

thanks!

cds-amal:
Hi @jimLynchCodes,

Thanks for bringing this issue up. By design there are no tests for truffle init but we could possibly include verbiage to guide developers on next steps, which would include tests.

cc: @kevinbluer , ideas?

kevinbluer:
Hey @JimLynchCodes, 

One idea we had after a quick discussion would be to extend the output following a successful `init`...currently it's as follows:
```
Starting init...
================

> Copying project files to /Users/bluer/Developer/temp

Init successful, sweet!
```
This could potentially be elaborated upon to say something like: "_Why not try `truffle create contract HelloWorld` and `truffle create test HelloWorld` to create a new contract and its associated test?_"

PR very much welcome for this if you'd be interested. Location of the console output as it current stands is at the following: https://github.com/trufflesuite/truffle/blob/develop/packages/events/defaultSubscribers/init.js#L23

Thanks again 👍

JimLynchCodes:
I like the "why not try `create contract` idea, but to me it seems weird to have to run an extra command to have it scaffold tests.

To me there isn't really a situation where you should want to scaffold a contract with no test (since, if you don't want it to work properly, why are you creating it in the first place? 😅)

eggplantzzz:
Yes but after `truffle init` there are really no contracts to test yet.

JimLynchCodes:
Hmm, good point. Maybe it makes sense to have an "Example" contract?

I have found that having only the Migrations.sol file that you are not supposed to touch is very confusing for beginners...

Rishabhraghwendra18:
Hi @JimLynchCodes . First timer this side . is this issue open ?

gnidan:
@Rishabhraghwendra18 looks like this issue should've been closed, assuming #4082 (now released) is sufficient in addressing this issue.

Closing this for issue maintenance, then, unless either of you report that there's more we could do. Thanks all!

PR #4082, opened 2021-05-30
add next steps for init command

suchris:
This PR adds instruction to help a developer after they created a truffle project with `truffle init` as discussed in #4041

### DIFF
--- a/packages/events/defaultSubscribers/init.js
+++ b/packages/events/defaultSubscribers/init.js
@@ -23,10 +23,10 @@ module.exports = {
         this.logger.log(`${OS.EOL}Init successful, sweet!${OS.EOL}`);
         this.logger.log(`Try our scaffold commands to get started:`);
         this.logger.log(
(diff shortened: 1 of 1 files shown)


kevinbluer:
This looks great @suchris, thanks for the contribution! Would you mind switching out the two `*`'s in the output to `$`'s as we briefly discussed this internally and felt it was a little more representative of a shell command? E.g.

```
  $ truffle create contract YourContractName # scaffold a contract
  $ truffle create test YourTestName         # scaffold a test
```

Thanks again and other than that it looks great!

kevinbluer (changes requested):
As per the comment if you could switch the two `*` references to `$` that would be awesome!

kevinbluer (approved):
LGTM @suchris 👌

Portalssecurity (commented):
$ truffle create contract YourContractName # scaffold a contract
  $ truffle create test YourTestName         # scaffold a test

Issue #4049, opened 2021-05-20
Truffle breakpoints fail if two contracts have the same name


0xfoobar:
## Issue

I'm debugging an external contract, and there are two contracts with the same name which both get called in the course of the transaction.

## Steps to Reproduce

Run a transaction that involves two contracts with the same name.

Then run `truffle debug -x $TXHASH` and enter `b ContractName:100` or whatever line number is valid.

## Expected Behavior

To be able to specify which contract I want to set a breakpoint in.

## Actual Results

![m4C6POmo](https://user-images.githubusercontent.com/83792390/118901474-39f10980-b8c8-11eb-898c-a523b4e5888f.png)

## Environment

* Operating System: macOS 10.15.7
* Ethereum client: ganache 6.12.1
* Truffle version (`truffle version`): 5.3.4
* node version (`node --version`): 15.5.1
* npm version (`npm --version`): 7.8.0

fainashalts:
Hi @0xfoobar thank you for bringing this up! We are aware of this issue and in the process of implementing a fix that involves setting breakpoints in all matching contracts. We'll keep you posted!

gnidan:
Closing, released in [Truffle v5.3.10](https://github.com/trufflesuite/truffle/releases/tag/v5.3.10).

PR #4087, opened 2021-06-02
Enhancement: Allow setting breakpoints when two sources have the same name

haltman-at:
Addresses #4049.  Or at least partly does?  It's something of a stopgap, but I don't see how we're going to do better for now.

Currently, when you try in the CLI debugger to add a breakpoint in a specified file, if there are multiple matching files, the CLI asks for disambiguation.  However, as per #4049, sometimes some sources might be named identically.  So this PR makes it so that if there's one matching file whose source path is a substring of all other matching source paths -- including, say, if there are a bunch of source paths which are identical -- then instead of asking for disambiguation, it'll print a warning and then add (or remove) them all.

In order to do this, I factored the breakpoint-setting function a bit, but be warned, it's still a big mess of imperative code that prints to the console a bunch!  Changing that is perhaps something that can be undertaken at a later time.  Anyway, now there's one function that determines the breakpoint objects to add/remove, and another to perform resolution and do the adding.  (They both print to the console...).  The former function returns the list of breakpoints, which will usually have length 1, but it can also return multiple breakpoints in the case discussed above, can return `[]` if something goes wrong and it doesn't know where to add anything, and will return `null` in the special case of `B all` (or `b all` :P ).  So yeah I used `[]` to mean "no breakpoints" and `null` to mean "all breakpoints"... this is called out in the comments so hopefully that isn't confusing.

Note I didn't really test the actual case this PR is made for; but uh I did test it on existing cases to make sure it doesn't break things?

Also, while I was at it, I made two other changes:
1. I fixed a bug I accidentally introduced in #4030, where typing `b` could get you an error if the debugger can't determine your location
2. I made it so that typing `b` followed by a large negative number won't hang the debugger!

### DIFF
--- a/packages/core/lib/debug/interpreter.js
+++ b/packages/core/lib/debug/interpreter.js
@@ -45,9 +45,30 @@ class DebugInterpreter {
   }
 
   async setOrClearBreakpoint(args, setOrClear) {
+    const breakpoints = this.determineBreakpoints(args, setOrClear); //note: not pure, can print
+    if (breakpoints !== null) {
+      for (const breakpoint of breakpoints) {
+        await this.setOrClearBreakpointObject(breakpoint, setOrClear);
+      }
+    } else {
+      //null is a special value representing all, we'll handle it separately
+      if (setOrClear) {
+        // only "B all" is legal, not "b all"
+        this.printer.print("Cannot add breakpoint everywhere.");
+      } else {
+        await this.session.removeAllBreakpoints();
+        this.printer.print("Removed all breakpoints.");
+      }
+    }
+  }
+
+  //NOTE: not pure, also prints!
+  //returns an array of the breakpoints, unless it's remove all breakpoints,
+  //in which case it returns null
+  //(if something goes wrong it will return [] to indicate do nothing)
+  determineBreakpoints(args, setOrClear) {
     //setOrClear: true for set, false for clear
     const currentLocation = this.session.view(controller.current.location);
-    const breakpoints = this.session.view(controller.breakpoints);
 
     const currentStart = currentLocation.sourceRange.start;
     const currentLength = currentLocation.sourceRange.length;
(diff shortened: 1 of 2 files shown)


packages/core/lib/debug/interpreter.js
```diff
@@ -132,42 +152,76 @@ class DebugInterpreter {
...
+            shortSource => matchingSources.every(
+              source =>
+                typeof source.sourcePath !== "string" //just ignore these I guess?
+                || source.sourcePath.includes(shortSource.sourcePath)
+            )
+          )
+        ) {
+          //exceptional case
+          this.printer.print(
+            `WARNING: Multiple source files matched, but one is a proper substring of all others, so disambiguation is not possible.  Acting on all of them.`
```
gnidan, line 172: I think this can be simplified:
```suggestion
            `WARNING: Multiple source files matched; could not disambiguate.  Setting breakpoints for:`
```

haltman-at: Well, we can't go with this exactly.  It can't be "setting breakpoints for", because this also applies to *removing* breakpoints.  We can have it switch as appropriate (I didn't bother as you can see, but it wouldn't be hard), but it can't just be "setting".

Also, I don't like "setting breakpoints for:" because the normal messages will still print for each.  If it were "setting breakpoints for:", you would expect to just see the *location* printed for each, not the full message.

Also, I think we do need a "but" in there.  "Multiple source files matched; could not disambiguate" sounds like we don't perform disambiguation.  We need to say "Multiple source files matced, but could not disambiguate" to make it clear that we would disambiguate if we could.

...although, really, *we* don't disambiguate, we ask the *user* to disambiguate.  Hence why I said "disambiguation is not possible", not "could not disambiguate".  IDK, maybe that's getting too nitpicky.

cds-amal: Does the user need to know the internals? Can we inform the user we encountered ambiguity because of multiple source name collision and set expectations that way?

cds-amal: Thanks for the clarification, it makes sense. How about rewording to present course of action followed by reasoning? 

> WARNING: acting on all sources because disambiguation is not possible due to one filename is a proper substring of all others.

haltman-at: So, rewording a little, if we take @cds-amal's approach, but also shorten things a bit as @gnidan wanted, I might write something like...

> WARNING: Acting on all matching sources because disambiguation between them is not possible.

How does that sound? :)

cds-amal: I'm ok with that!

haltman-at: OK, I've made this change.

cds-amal (approved):
Looks good!

Issue #1676, opened 2019-01-29
Unboxing from a branch fails when a slash is in the branch name


adrianmcli:
## Issue

This is the syntax for unboxing from a branch:

```
truffle unbox owner-name/repo-name#branch-name
```

However, when `branch-name` has a forward slash in it, unboxing fails with an error:

```
Error: Box specified in invalid format
```

This is where the problem is: https://github.com/trufflesuite/truffle/blob/e913be03c564cd967d58a67ddc16360188db3ead/packages/truffle-core/lib/commands/unbox.js#L7-L34

## Steps to Reproduce

Try to unbox with the following command:

```
truffle unbox truffle-box/webpack-box#fix/fail-to-unbox
```

## Expected Behavior

It should just unbox normally.

## Actual Results

Error stack:

```
Error: Box specified in invalid format
    at normalizeURL (/Users/adrianli/n/lib/node_modules/truffle/build/webpack:/packages/truffle-core/lib/commands/unbox.js:33:1)
    at Object.run (/Users/adrianli/n/lib/node_modules/truffle/build/webpack:/packages/truffle-core/lib/commands/unbox.js:123:1)
    at Command.run (/Users/adrianli/n/lib/node_modules/truffle/build/webpack:/packages/truffle-core/lib/command.js:113:1)
    at Object.<anonymous> (/Users/adrianli/n/lib/node_modules/truffle/build/webpack:/packages/truffle-core/cli.js:47:1)
    at __webpack_require__ (/Users/adrianli/n/lib/node_modules/truffle/build/webpack:/webpack/bootstrap 10c8e5c48cca88d8ca36:19:1)
    at /Users/adrianli/n/lib/node_modules/truffle/build/webpack:/webpack/bootstrap 10c8e5c48cca88d8ca36:65:1
    at Object.<anonymous> (/Users/adrianli/n/lib/node_modules/truffle/build/cli.bundled.js:71:10)
    at Module._compile (internal/modules/cjs/loader.js:688:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:699:10)
    at Module.load (internal/modules/cjs/loader.js:598:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:537:12)
    at Function.Module._load (internal/modules/cjs/loader.js:529:3)
    at Function.Module.runMain (internal/modules/cjs/loader.js:741:12)
    at startup (internal/bootstrap/node.js:285:19)
[...]

PR #4088, opened 2021-06-03
Handle unbox with slashes in branch name Fix/unbox 1676

cds-amal:
This PR refactors unbox's normalize function and now handles unbox ing branches with slashes (#1676). Tests have been added to maintain backward compatibility. All current unit and integration tests work, so hopefully...🤞  

Also added debug output to log input/output state for easier debugging just in case.

 
```console
$ DEBUG=unbox truffle unbox pet-shop#for/testing ps-dir
  unbox {
  unbox   in: 'pet-shop#for/testing',
  unbox   out: 'https://github.com:truffle-box/pet-shop-box#for/testing'
  unbox } +0ms
```


packages/box/box.ts
```diff
@@ -19,50 +21,75 @@ const defaultPath = "https://github.com:trufflesuite/truffle-init-default";
...
-  }
 
-  // rewrite https://github.com/truffle-box/metacoin format in
-  //         https://github.com:truffle-box/metacoin format
-  if (url.match(/.com\//)) {
-    url = url.replace(/.com\//, ".com:");
-  }
+  // process https? or git prefixed input
+  //
+  if (/^(https?|git)/i.test(url) ){
```
eggplantzzz, line 35: Stupid nitpick but irregular spacing near the end of the line.

packages/box/box.ts
```diff
@@ -19,50 +21,75 @@ const defaultPath = "https://github.com:trufflesuite/truffle-init-default";
...
+    const protocolRex = new RegExp(
+      "(?<protocol>(https://|git@))"  + 
+      "(?<service>[^@/]+)"            + // service
+      "(/|:)"                         + // separator. will be ignored in ouput
+      "(?<org>[^/#]+)"                + // org
+      "/"                             +
+      "(?<repo>[^/#]+)"               + // repo
+      "#?"                            + // optional branch separator
+      "(?<branch>[^#]+)?"             + // capture branch (undefined if unmatched)
```
eggplantzzz, line 44: That's cool, I didn't know you could name these matches

packages/box/test/box.js
```diff
@@ -4,7 +4,7 @@ const assert = require("assert");
 const inquirer = require("inquirer");
 const sinon = require("sinon");
 const Config = require("@truffle/config");
-const Box = require("../");
+const { default: Box, normalizeSourcePath  }= require("../");
 const TRUFFLE_BOX_DEFAULT =
   "https://github.com:trufflesuite/truffle-init-default";
```
eggplantzzz, line 9: I wonder if we want to eventually move to using the bare-box in the truffle-box repo instead of this one...

packages/box/test/normalizeSourcePath.js
```diff
@@ -0,0 +1,172 @@
...
+      input: "git@github.com:truffle-box/a-box#some/complex/branch",
+      expected: "https://github.com:truffle-box/a-box#some/complex/branch",
+    }, {
+      input: "git@github.com/truffle-box/a-box#some/complex/branch",
+      expected: "https://github.com:truffle-box/a-box#some/complex/branch",
+    }, {
+      input: "GIT@GITHUB.COM/TRUFFLE-BOX/A-BOX#SOME/COMPLEX/BRANCH",
+      expected: "https://GITHUB.COM:TRUFFLE-BOX/A-BOX#SOME/COMPLEX/BRANCH",
+    }].forEach(({ input, expected }) => {
+      it(`pass \`${input}\``, () => {
```
eggplantzzz, line 22: should all these be "passes"?

packages/events/defaultSubscribers/unbox.js
```diff
@@ -94,7 +94,7 @@ module.exports = {
     "unbox:fail": [
       function () {
         if (this.quiet) return;
-        this.spinner.fail();
+        if (this.spinner) this.spinner.fail();
```
eggplantzzz, line 97: wahoo

packages/box/test/box.js
```diff
@@ -29,22 +29,40 @@ describe("@truffle/box Box", () => {
...
     afterEach(() => {
       Box.checkDir.restore();
     });
 
-    it("unboxes truffle box from github", done => {
-      Box.unbox(TRUFFLE_BOX_DEFAULT, destination, {}, config).then(
-        truffleConfig => {
-          assert.ok(truffleConfig);
+    describe("unboxes truffle box", () => {
+      it("from GitHub master branch", done => {
```
eggplantzzz, line 38: not necessary at all but you could rewrite these using `async` :)

eggplantzzz:
I only found some piddly stuff that doesn't much matter. I didn't nitpick the regexes very much though and kind of relied on the comments. Someone with better regex skills should probably look haha

packages/box/box.ts
cds-amal: I'll apply prettier to handle this. I held off while trying to come up with a technique to document the regexes.

packages/box/box.ts
cds-amal: named capture groups are really useful!

packages/box/test/box.js
eggplantzzz: I think it is probably mainly historically significant. Actually we don't even use this any more. `truffle init` copies files and no longer requires an internet connection. Maybe we can update some tests to use that infrastructure instead of unbox.

packages/box/box.ts
```diff
@@ -19,50 +21,132 @@ const defaultPath = "https://github.com:trufflesuite/truffle-init-default";
...
+      "(?<branch>#[\\w./-]+)?",
+
+      // the input string must be consumed fully at this point to match
+      "$",
+    ].join(""));
+
+    const match = url.match(protocolRex);
+    if (match) {
+      const { groups: G } = match;
+      const branch = G["branch"] ? `${G["branch"]}` : "";
```
eggplantzzz, line 77: Does `${G["branch"]}` need the \`${..}\`? Why not just `G["branch"]`?

packages/box/box.ts
```diff
@@ -19,50 +21,132 @@ const defaultPath = "https://github.com:trufflesuite/truffle-init-default";
...
+      // branch is 1 or many (word, dot or dash) and can be optional
+      "(?<branch>#[\\w./-]+)?",
+
+      // the input string must be consumed fully at this point to match
+      "$",
+    ].join(""));
+
+    const match = url.match(protocolRex);
+    if (match) {
+      const { groups: G } = match;
```
gnidan, line 76: Don't abbreviate variables! Ever!

packages/box/box.ts
cds-amal: yes, that's better!

packages/box/box.ts
cds-amal: haha, done.

PR #4089, opened 2021-06-03
Revert "Revert "Dependency update: Update to the newer @ensdomains/ensjs""

eggplantzzz:
Reverts trufflesuite/truffle#4075


haltman-at (commented):
Looks like you forgot to add `crypto` to the package.json (unless there's some reason it doesn't need to be).

Other than that, looks good to me!  Thanks Tyler!

haltman-at:
Actually, wait, sorry, one other thing -- this fix happens in `core`.  But we're using ENS in `contract`.  That means that anyone using `contract` outside of its use in `@truffle/core` will still have this problem.  That sounds like a problem to me.  Is there a way to avoid that?

eggplantzzz:
crypto is native to Node and doesn't need to be included there...it just gets stripped out during bundling by webpack.

Ach, you're right. It probably will error when using the two packages standalone. I'll try out some other solutions in contract and deployer.

eggplantzzz:
Ok so I verified that those packages don't cause the warning to be logged when required and used independent of the Truffle bundle.

haltman-at (approved):
Nice, sounds good to me then! Thanks a bunch Tyler. :)

Issue #4090, opened 2021-06-03
Debugger exits violently when evaluating expressions with variables that don't exist


tcoulter:
When using the debugger, if you evaluate an expression that contains a variable name that doesn't exist, the debugger will output hundreds of lines of garbage output then throw an error. The example below shows me deliberately entering the name of a variable that doesn't exist; however, the error happens just as easily with typos, which ruins the debugging experience given that the debugger halts.

![TruffleDebuggerError](https://user-images.githubusercontent.com/92629/120714652-55d3de00-c478-11eb-8d01-388ac04e6e30.gif)

### To reproduce:

1. Start debugger with a transaction you have sources for, against Ganache
2. When the debugger starts up, enter the text `:thisdoesnotexist`, or literally anything that starts with a `:`, so long as there isn't a variable name that matches what you entered
3. Watch as the debugger outputs a lot of code (on my machine, it fills up my console buffer), finally halting and displaying an error.

### Versions:

```
$ npx truffle version
Truffle v5.3.6 (core: 5.3.6)
Solidity - 0.6.4 (solc-js)
Node v12.18.3
Web3.js v1.3.5
```

haltman-at:
Hm, I can't reproduce this error.  Funny -- this is not the first time someone's reported one of these code-dump errors, but we've never been able to reproduce them consistently.  Or, I think someone on the team managed to reproduce one once, but we never got back to it and found a way to do it consistently?  Anyway, for the moment I can't reproduce this, so could you confirm whether this still happens on latest Truffle?  Thanks!  (Not that anything relevant should have changed in that time, but just to confirm...)

haltman-at:
Ah, it was @cds-amal who managed to earlier reproduce an error like this.  Will need to bug him about how he managed to do that...

gnidan:
No more exiting violently, as of [Truffle v5.3.10](https://github.com/trufflesuite/truffle/releases/tag/v5.3.10)!

Thanks for the bug report @tcoulter 🙇

PR #4091, opened 2021-06-05
Remove listeners for `unhandledRejection` after loading soljson, rather than `uncaughtException`

cds-amal:
Some time in the past soljson registered for `uncaughtException` and we adapted it. the most recent listens for `unhandledRejection` and aborts. 

It may be related to #4090


cds-amal:
This may not be the proper fix because we should account for the appropriate listener based on soljson's version.

cds-amal:
Closing this for the more general solution (#4092) that would work with multiple soljson versions

PR #4092, opened 2021-06-05
Only remove listeners left around by soljson

gnidan:
So, here's a hypothesis for the `uncaughtException` business in compile-solidity.

Solidity upgraded Emscripten from v1.35.4 to v1.37.21 in solc v0.5.1 (in [this PR](https://github.com/ethereum/solidity/pull/4486)). Inspecting the Emscripten changelog, [this issue](https://github.com/emscripten-core/emscripten/issues/4392) was closed as part of v1.36.6. Looks like the fix here may have changed buggy behavior that left `uncaughtException`s to linger.

Since we support old Solidity and need to handle pre-v0.5.1 behavior, this PR proposes that we alter the listener removal by first capturing the listeners that existed before loading solc and then removing only those listeners that are new after loading solc.

This is a different proposed solution than #4091... I haven't done any digging re: that PR's hypothesis that we must remove listeners for `unhandledRejection`, but I suspect that may be a red herring. This PR should address the "exits violently" part of #4090 (as well as many other issues over the past however-long)

### DIFF
--- a/packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
+++ b/packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
@@ -77,16 +77,22 @@ class LoadingStrategy {
     );
   }
 
-  /**
-   * Cleans up error listeners set (by solc?) when requiring it. (This code inherited from
-   * previous implementation, note to self - ask Tim about this)
-   */
-  removeListener() {
-    const listeners = process.listeners("uncaughtException");
-    const execeptionHandler = listeners[listeners.length - 1];
+  markListeners() {
+    return new Set(process.listeners("uncaughtException"));
+  }
 
-    if (execeptionHandler) {
-      process.removeListener("uncaughtException", execeptionHandler);
+  /**
+   * Cleans up error listeners left by older versions of soljson (pre-0.5.1).
+   * Use with `markListeners()`
+   */
+  removeListener(markedListeners) {
+    const listeners = process.listeners("uncaughtException");
+    const newListeners = listeners.filter(
+      listener => !markedListeners.has(listener)
+    );
+
+    for (const listener of newListeners) {
+      process.removeListener("uncaughtException", listener);
     }
   }
 
--- a/packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
+++ b/packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
@@ -78,21 +78,26 @@ class LoadingStrategy {
   }
 
   markListeners() {
-    return new Set(process.listeners("uncaughtException"));
+    return {
+      uncaughtException: new Set(process.listeners("uncaughtException")),
+      unhandledRejection: new Set(process.listeners("unhandledRejection")),
+    };
   }
 
   /**
-   * Cleans up error listeners left by older versions of soljson (pre-0.5.1).
+   * Cleans up error listeners left by soljson
    * Use with `markListeners()`
    */
   removeListener(markedListeners) {
-    const listeners = process.listeners("uncaughtException");
-    const newListeners = listeners.filter(
-      listener => !markedListeners.has(listener)
-    );
+    for (const eventName of ["uncaughtException", "unhandledRejection"]) {
+      const listeners = process.listeners(eventName);
+      const newListeners = listeners.filter(
+        (listener) => !markedListeners[eventName].has(listener),
+      );
 
-    for (const listener of newListeners) {
-      process.removeListener("uncaughtException", listener);
+      for (const listener of newListeners) {
+        process.removeListener(eventName, listener);
+      }
     }
   }
 
--- a/packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
+++ b/packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
@@ -89,14 +89,12 @@ class LoadingStrategy {
    * Use with `markListeners()`
    */
   removeListener(markedListeners) {
-    for (const eventName of ["uncaughtException", "unhandledRejection"]) {
-      const listeners = process.listeners(eventName);
-      const newListeners = listeners.filter(
-        (listener) => !markedListeners[eventName].has(listener),
-      );
-
-      for (const listener of newListeners) {
-        process.removeListener(eventName, listener);
+    for (const eventName in markedListeners) {
+      const marked = markedListeners[eventName];
+      for (const listener of process.listeners(eventName)) {
+        if (!marked.has(listener)) {
+          process.removeListener(eventName, listener);
+        }
       }
     }
   }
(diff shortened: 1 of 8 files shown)


gnidan:
Another approach might be to look at the actual version of solc, but I went with this because the explicit set comparison will handle the problem more generally... and cover nightlies and such, so we won't need to identify safe/unsafe versions so rigorously.

cds-amal:
Maybe both `unhandledRejection` and `uncaughtException` listeners need to be removed. Here's a quick way to tickle solcjson's `unhandledRejection`. 

```console
reprod tangerine
use-truffle-core

truffle init

cat << EOF > ./migrations/1_initial_migration.js
const Migrations = artifacts.require("Migrations");

module.exports = function (deployer) {
  Promise.reject(4242);
  deployer.deploy(Migrations);
};
EOF

truffle test
```

packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -77,16 +77,22 @@ class LoadingStrategy {
     );
   }
 
+  markListeners() {
+    return new Set(process.listeners("uncaughtException"));
```
cds-amal, line 81: Note, the current version of soljson listens for `unhandledRejection`. The logic in the rest of this PR can be extended to consider both handlers.

```suggestion
    return {
      uncaughtException: new Set(process.listeners("uncaughtException")),
      unhandledRejection: new Set(process.listeners("unhandledRejection""))
    };
```

cds-amal (commented):
I like this approach and think it should be extended. I tested it with a rejected promise in migrations and noted soljson's `unhandledRejection` listener results in the EMSCRITEN screen dump.

The repro steps for the bug is in a previous comment.

packages/compile-solidity/compilerSupplier/loadingStrategies/Local.js
```diff
@@ -21,7 +23,8 @@ class Local extends LoadingStrategy {
     }
     //HACK: if it has a compile function, assume it's already wrapped
     wrapped = soljson.compile ? soljson : solcWrap(soljson);
-    this.removeListener();
+
+    this.removeListener(markedListeners);
```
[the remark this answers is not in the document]
gnidan: Pushed these; @cds-amal would you mind verifying that the `try`/`finally` blocks don't cause any wonkiness?

packages/compile-solidity/compilerSupplier/loadingStrategies/Local.js
```diff
@@ -10,22 +10,23 @@ class Local extends LoadingStrategy {
...
-    this.removeListener(markedListeners);
-    return wrapped;
+      try {
+        soljson = originalRequire(compilerPath);
+      } catch (error) {
+        throw this.errors("noPath", localPath, error);
+      }
+      //HACK: if it has a compile function, assume it's already wrapped
+      wrapped = soljson.compile ? soljson : solcWrap(soljson);
+      return wrapped;
```
cds-amal, line 26: Is the temp variable needed if it's returned immediately in the next statement? 

```suggestion
      return soljson.compile ? soljson : solcWrap(soljson);
```

packages/compile-solidity/compilerSupplier/loadingStrategies/VersionRange.js
```diff
@@ -10,10 +10,13 @@ const LoadingStrategy = require("./LoadingStrategy");
...
   compilerFromString(code) {
     const markedListeners = this.markListeners();
-    const soljson = requireFromString(code);
-    const wrapped = solcWrap(soljson);
-    this.removeListener(markedListeners);
-    return wrapped;
+    try {
+      const soljson = requireFromString(code);
+      const wrapped = solcWrap(soljson);
+      return wrapped;
```
cds-amal, line 16:
```suggestion
      return solcWrap(soljson);
```

packages/compile-solidity/compilerSupplier/loadingStrategies/VersionRange.js
```diff
@@ -34,12 +37,15 @@ class VersionRange extends LoadingStrategy {
...
-    debug("soljson %o", soljson);
-    const wrapped = solcWrap(soljson);
-    this.removeListener(markedListeners);
-    return wrapped;
+    try {
+      const filePath = this.resolveCache(fileName);
+      const soljson = originalRequire(filePath);
+      debug("soljson %o", soljson);
+      const wrapped = solcWrap(soljson);
+      return wrapped;
```
cds-amal, line 45:
```suggestion
      return solcWrap(soljson);
```

cds-amal (changes requested):
some minor nits.

packages/compile-solidity/compilerSupplier/loadingStrategies/Local.js
gnidan: Nah probably not, unless we wanna debug it later :)

cds-amal (approved):

cds-amal:
Nice one @gnidan !

packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -77,16 +77,27 @@ class LoadingStrategy {
...
+      const listeners = process.listeners(eventName);
+      const newListeners = listeners.filter(
+        (listener) => !markedListeners[eventName].has(listener),
+      );
 
-    if (execeptionHandler) {
-      process.removeListener("uncaughtException", execeptionHandler);
+      for (const listener of newListeners) {
+        process.removeListener(eventName, listener);
+      }
```
davidmurdoch, line 100: This could be condensed to:

```javascript
for (const eventName in markedListeners) {
  const marked= markedListeners[eventName];
  process.listeners(eventName).forEach((listener) => {
    if (!marked.has(listener)) {
      process.removeListener(eventName, listener);
    }
  });
}
```

And if we change `markedListeners` to a `Map<string, Array<AnyFunction>` it could be:

```javascript
for (const [eventName, marked] of markedListeners) {
  process.listeners(eventName).forEach((listener) => {
    if (!marked.has(listener)) {
      process.removeListener(eventName, listener);
    }
  });
}
```

davidmurdoch (approved):

packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
gnidan: Borrowed the first one :) Thanks @davidmurdoch!

gnidan:
(I feel like there's a joke here that begins "how many engineers does it take to add a mutex lock?" :)

Thanks @cds-amal and @davidmurdoch! Will merge when green.

packages/compile-solidity/compilerSupplier/loadingStrategies/LoadingStrategy.js
```diff
@@ -89,14 +89,12 @@ class LoadingStrategy {
...
-      );
-
-      for (const listener of newListeners) {
-        process.removeListener(eventName, listener);
+    for (const eventName in markedListeners) {
+      const marked = markedListeners[eventName];
+      for (const listener of process.listeners(eventName)) {
+        if (!marked.has(listener)) {
+          process.removeListener(eventName, listener);
+        }
```
cds-amal, line 97: nice! Thank you!

cds-amal (approved):

Issue #4093, opened 2021-06-07
Codec & debugger should read compiler version from CBOR when not present in artifact


haltman-at:
## Issue

Currently, codec & debugger, when they need to determine the Solidity version for a particular contract, rely on it being present in the input (artifacts or compilations).  For compilations of course we always have this, but it can be missing from artifacts.

However, in Solidity 0.5.9 and later, this information is included in the contract's CBOR, and we should use that if the information is missing from the artifact.  (We probably don't want to use it in preference to the one in the artifact, since the information in the CBOR is lower-resolution.)

One thing to beware of here: The information is stored differently depending on whether the contract was compiled with a release version of Solidity or a prerelease version.  Make sure you account for both cases.

(Further note: This will only work for Solidity contracts, of course, not Yul contracts, and certainly not other languages such as Vyper.  And as mentioned, this information was only added in 0.5.9.  The CBOR info itself was added in 0.4.7, but we currently don't support Solidity older than 0.4.9.)

## Environment

* Truffle version (`truffle version`): 5.3.9

gnidan:
Closing, released in [Truffle v5.3.10](https://github.com/trufflesuite/truffle/releases/tag/v5.3.10).

PR #4094, opened 2021-06-08
Enhancement: Detect solc version from CBOR if absent in artifact

haltman-at:
Addresses #4093.

This PR makes it so that, if a given contract is lacking compiler information, the decoder/debugger (this is handled purely in codec so it applies to both) will read it out of the bytecode if possible.

Starting in Solidity 0.5.9, the CBOR section of Solidity bytecode includes the version in the `solc` field.  For prerelease versions the version is given as a string; for release versions it's given as a 3-byte bytestring, with the bytes being the major, minor, and patch components of the version.

This PR makes it so that `normalizeContexts` will read this information from the bytecode and put it in the context if the context doesn't have it.  Note that it only affects contexts, not sources.  The `normalizeContexts` function might not seem like the most natural place to put it, but it was the easiest; that's where we're already doing CBOR processing, after all, and it's a single location where we can stick this and have it instantly effective everywhere.  Because we're now using the CBOR for two different things there, I decoded it all first in a separate pass so we could just refer to it later, rather than going through all the trouble of extracting and decoding it twice.

I didn't add any tests (neither debugger nor decoder is very well set up at the moment to test this easily), but I did some quick manual tests.

### DIFF
--- a/packages/codec/lib/contexts/utils.ts
+++ b/packages/codec/lib/contexts/utils.ts
@@ -2,6 +2,7 @@ import debugModule from "debug";
 const debug = debugModule("codec:contexts:utils");
 
 import * as Evm from "@truffle/codec/evm";
+import { CompilerVersion } from "@truffle/codec/compiler";
 import { Context, Contexts } from "./types";
 import escapeRegExp from "lodash.escaperegexp";
 import * as cbor from "cbor";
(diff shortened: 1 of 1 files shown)


cds-amal (commented):
@haltman-at can you explain how to test the code manually? I'd like to see the different scenarios get triggered.

gnidan:
I guess solc won't have 257 patch releases for a given minor :)

haltman-at:
Heh, I hope not!

cds-amal (approved):
LGTM!

Thanks for the detailed steps @haltman-at! I really enjoyed reviewing this PR and stepping through the code. I was able to test with different solidity versions and will take your word for Yul and Vyper contracts. Seems like a good time for me to start working with these two!

PR #4097, opened 2021-06-09
Use provider.send directly in @truffle/decoder

gnidan:
Tearing out web3.js for this purpose

### DIFF
--- a/packages/decoder/lib/decoders.ts
+++ b/packages/decoder/lib/decoders.ts
@@ -1,6 +1,7 @@
 import debugModule from "debug";
 const debug = debugModule("decoder:decoders");
 
+import { promisify } from "util";
 import * as Abi from "@truffle/abi-utils";
 import * as Codec from "@truffle/codec";
 import {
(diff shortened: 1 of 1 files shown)


packages/decoder/lib/decoders.ts
```diff
@@ -742,15 +763,23 @@ export class WireDecoder {
...
   }
 
   /**
    * @protected
    */
-  public getDeployedContexts(): Contexts.Contexts {
-    return this.deployedContexts;
+  public async send(options: { method: string; params: any[] }): Promise<any> {
+    const { method, params } = options;
+    const { result } = await promisify(this.provider.send.bind(this.provider))({
```
danfinlay, line 775: I would recommend you check if the provider has a `request` method and use it first. The `send` method [is currently deprecated](https://github.com/MetaMask/providers/blob/e3353443fa4b967ff768d666be1f8937b8d1bfde/src/MetaMaskInpageProvider.ts#L245-L254).

gnidan:
Closing this because build's failing and there's more work to do, and because the immediate need for this is now gone.

PR #4100, opened 2021-06-09
Bug fix: Prevent (extremely unlikely) crash from cbor null

haltman-at:
Quick followup to #4094, I just realized that I forgot to account for the case of `null` (which is an object, but which you better not try to access the properties of).  Mind you, this should never actually happen, barring an extremely unlikely coincidence, but just to be safe, I'm putting in checks for it here.  (I mean, a number of the cases I check for here should never happen...)


gnidan (approved):
Sure yup

PR #4106, opened 2021-06-11
Add optimistic to potential etherscan networks

fainashalts:
This PR adds the "optimistic" etherscan to allowed networks in the `source-fetcher` package so we can fetch external information from the mainnet Optimism network.


haltman-at (approved):
Looks good to me!  I'll suggest adding `optimistic-kovan` as well, but if you want to just get this merged then that can be added later. :)

gnidan (approved):

PR #4112, opened 2021-06-15
Upgrade @truffle/codec's typedoc to v0.20.36

gnidan:
This PR removes `@gnd/typedoc` and replaces it with the latest stable typedoc release.

Few things were broken, which this PR accommodates:
- [typedoc-plugin-external-module-name](https://github.com/christopherthielen/typedoc-plugin-external-module-name) seems to get broken, which we were using for `@module`. Fortunately it seems like this tag works without that, though!
- Decoder re-exports a couple things from codec... to ensure these don't show up, this PR adds [typedoc-plugin-remove-references](https://github.com/eyworldwide/typedoc-plugin-remove-references).
- All the documentation comments for each module/namespace, defined _on the export identifier_, needed to be moved into those modules themselves, and marked with `@packageDocumentation`. This PR hopefully got them all!
- Inside docs/tsconfig and the typedoc config, paths no longer worked as specified relative to the tsconfig's `rootDir`; instead, this PR updates those to be relative to the `docs/` directory itself.
- Decoder stopped being recognized, so this PR adds that to the `include`.

### DIFF
--- a/packages/codec/docs/typedoc.json
+++ b/packages/codec/docs/typedoc.json
@@ -1,12 +1,8 @@
 {
   "name": "Truffle Decoding and Encoding",
-  "mode": "modules",
-  "module": "commonjs",
-  "theme": "../../../node_modules/@trufflesuite/typedoc-default-themes/bin/default/",
-  "includeDeclarations": false,
-  "ignoreCompilerErrors": true,
-  "preserveConstEnums": true,
+  "theme": "../../node_modules/@trufflesuite/typedoc-default-themes/bin/default/",
   "excludeExternals": true,
+  "plugin": ["typedoc-plugin-remove-references"],
   "categoryOrder": [
     "Data",
     "ABI data location",
@@ -29,6 +25,5 @@
   ],
   "exclude": [
     "**/test/*"
-  ],
-  "target": "ES6"
+  ]
 }
--- a/packages/codec/docs/tsconfig.json
+++ b/packages/codec/docs/tsconfig.json
@@ -32,7 +32,8 @@
 
   },
   "include": [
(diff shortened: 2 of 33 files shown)


packages/codec/docs/typedoc.json
```diff
@@ -1,12 +1,8 @@
 {
   "name": "Truffle Decoding and Encoding",
-  "mode": "modules",
-  "module": "commonjs",
-  "theme": "../../../node_modules/@trufflesuite/typedoc-default-themes/bin/default/",
-  "includeDeclarations": false,
```
haltman-at, line 6: Hm, why these deleted lines?

haltman-at (approved):
Seems good to me.  One additional thing you might want to do -- right now we have a bunch of
```typescript
import * as Name from "path";
export { Name };
```
We should, if I'm not mistaken, be able to replace this with
```typescript
export * as Name from "path";
```
I *think* the reason we didn't do this previously was due to it not playing nicely with typedoc.  But with these changes I think it should work now?  I could be wrong, it could blow up or something, but if you have the time you might want to try it?

Otherwise this seems good.  Thanks for doing this!

packages/codec/docs/typedoc.json
gnidan: Those options don't exist anymore!

gnidan:
> I think the reason we didn't do this previously was due to it not playing nicely with typedoc.

I'm not sure - I tried this `export * as` syntax in another project recently, and it was still broken. Didn't think to try it here; maybe that can be a later change and we'll see?

gnidan:
Thanks Harry!

PR #4126, opened 2021-06-21
Minor corrections for Tezos compiler upgrade

bergarces:
- @truffle/artifactor: Minor refactor to remove code duplication and show warning if `contract_name` is missing but `contractName` is present.
- @truffle/compile-common: Make properties from `ProfilerConfig` type optional, as they are currently not being passed in some instances of @truffle/compile-vyper.

None of these are breaking changes.

### DIFF
--- a/packages/artifactor/tsconfig.json
+++ b/packages/artifactor/tsconfig.json
@@ -9,7 +9,7 @@
     "sourceMap": true,
     "outDir": "dist",
     "baseUrl": ".",
-    "lib": ["es2017"],
+    "lib": ["es2017","dom"],
     "paths": {
       "web3": [
         "../../node_modules/@types/web3/index",
--- a/packages/artifactor/tsconfig.json
+++ b/packages/artifactor/tsconfig.json
@@ -9,7 +9,7 @@
     "sourceMap": true,
     "outDir": "dist",
     "baseUrl": ".",
-    "lib": ["es2017","dom"],
+    "lib": ["es2017"],
     "paths": {
       "web3": [
         "../../node_modules/@types/web3/index",
--- a/packages/compile-common/src/profiler/profiler.ts
+++ b/packages/compile-common/src/profiler/profiler.ts
@@ -10,8 +10,8 @@ import { requiredSources, RequiredSourcesOptions } from "./requiredSources";
 import { convertToAbsolutePaths } from "./convertToAbsolutePaths";
 
 export interface ProfilerConfig {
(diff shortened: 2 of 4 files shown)


packages/compile-common/src/profiler/profiler.ts
```diff
@@ -10,8 +10,8 @@ import { requiredSources, RequiredSourcesOptions } from "./requiredSources";
 import { convertToAbsolutePaths } from "./convertToAbsolutePaths";
 
 export interface ProfilerConfig {
-  parseImports: RequiredSourcesOptions["parseImports"];
-  shouldIncludePath: RequiredSourcesOptions["shouldIncludePath"];
+  parseImports?: RequiredSourcesOptions["parseImports"];
+  shouldIncludePath?: RequiredSourcesOptions["shouldIncludePath"];
```
bergarces, line 14: Here's `compile-vyper` creating an instance of profiler Profiler without them. For `compile-ligo` and `compile-michelson` I need to create it with `shouldIncludePath` but without `parseImports`.

packages/artifactor/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "sourceMap": true,
     "outDir": "dist",
     "baseUrl": ".",
-    "lib": ["es2017"],
+    "lib": ["es2017","dom"],
```
haltman-at, line 12: I don't understand the point of this, since obviously there's no actual DOM here... you say it's about getting rid of some warning?  What warning is that, exactly?

haltman-at (approved):
Seems good, although I don't entirely understand the `"dom"` thing.

packages/artifactor/tsconfig.json
haltman-at: Oh, I see -- when using `console`.  I failed to parse that correctly at first.  Is there some way to tell it about Node instead of telling it about the DOM?

bergarces: I have removed that change, as "dom" is for browser apps mainly.

When I was working with it, a warning was showing, but it shouldn't have since @types/node package was already there. I have removed it... and the warning is gone anyway.

bergarces: I think I was getting the issue before bootstrapping the project and @types/node not being installed yet. I blindly followed vscode suggestion, but the issue is not there with a bootstrapped project.

haltman-at:
Great, seems good then!

Issue #2472, opened 2019-10-18
Environment script to be ran by Truffle when starting a console.


hickscorp:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

Note that this is not an issue, but rather a feature request. I asked a question on the Truffle's Gitter and I was pointed at opening an issue here.

The original question was:
> Is there a way to have a script ran in the same scope at the truffle console every time it loads? I would like to have some functions and stuff ready automatically whenever I enter truffle console.

Let me expand on that - when in a truffle console, it'd be great to have a way to have some "environmental" setup done per-project. For instance, having a set of functions automatically loaded from a module. From this standpoint, the module would expose a function that would receive important things (Like a regular Truffle script) - one of the important one being `artifacts`.

gnidan:
Hm we can probably hook up the `truffle exec` plumbing to this. What would this be... a config option?

```javascript
module.exports = {
  /* ... rest of config ... */
  console: {
    exec: "<path/to/script.js>"
  }
}
```

something like that maybe?

hickscorp:
@gnidan I think your suggestion is pretty much what would be needed for a first iteration - as long as the script receives proper environment (particularly `artifacts`).

hickscorp:
Bump - was there any progress on this - or similar?

cds-amal:
This would be a feature I would use! @gnidan would love to discuss progressing this.

hickscorp:
For example, every single time I open my truffle console, I have to copy-paste this in it:

```
const BN = require('bn.js');
const generateUtils = require('./scripts/utils.js')
const netId = process.env.NETWORK_ID
const zosFile = `.openzeppelin/dev-${netId}.json`
const zosAbi = JSON.parse(fs.readFileSync(zosFile))
const people = Object.assign(require(`./conf/addresses.${netId}`), require(`./conf/addresses.private.${netId}`))
const governance = { from: people.pk2m }
const issuance = { from: people.issuer }
const registry = await Registry.at(zosAbi.proxies[`TTContracts/Registry`][0].address)
const access = await Access.at(zosAbi.proxies[`TTContracts/Access`][0].address)
const register = await Register.at(zosAbi.proxies[`TTContracts/Register`][0].address)
const transact = await Transact.at(zosAbi.proxies[`TTContracts/Transact`][0].address)
const token = await Token.at(zosAbi.proxies[`TTContracts/Token`][0].address)
```

[...]

PR #4127, opened 2021-06-21
Feature: Run user-supplied scripts to populate the console

eggplantzzz:
In https://github.com/trufflesuite/truffle/issues/2472 it was suggested to add a way of providing Truffle with scripts to be run when starting the console environment. This PR provides the ability so specify paths to scripts in `truffle-config.js`. 

Users can now add a `console.require` property which will either be a string or an array. If the user supplies a string, it must be a path to a JS file that will be `require`d. The `module.exports` must be a JS object with named keys. The keys will be variables set in the console environment and they will be set to the value of their value. So for example, consider the case where the user supplies the following JS:
```
// user-specified JS file
module.exports = {
  breakfast: "yes please"
};
```
When the user starts up the console via `truffle console` or `truffle develop`, the environment should have the variable "breakfast" which will be equal to the string "yes please".

The other possible value for `console.require` is an array. Each member of the array must either be a string or an object that contains at least a `path` property. Strings will be treated exactly as described in the preceding paragraph. Objects must contain a `path` property (that will have a string value) that will once again specify a path to a file to be `require`d. It will be treated exactly like strings.

The only other member allowed for objects is a `as` property. The variables specified by the JS will then be namespaced under whatever string value is provided for `as`. The following is an example of specifying the `console.require` format when not using just a plain string.
```
// truffle-config.js
module.exports = {
  console: {
    require: [
      "../../food.js",
      { path: "./myStuff.js" },
      { path: "../someOtherStuff.js", as: "myVariables" }
    ]
  }
};
```

That configuration would load the variables found in "../../food.js", "./myStuff.js", and "../someOtherStuff.js". Furthermore, the variables found in "../someOtherStuff.js" would be available namespaced under `myVariables`.

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -142,12 +142,12 @@ class Console extends EventEmitter {
       // to sign transactions (e.g. no reason to disallow debugging)
       accounts = [];
     }
+    // we load user variables first so as to not clobber ours
+    this.hydrateUserDefinedVariables();
 
     this.repl.context.web3 = this.web3;
     this.repl.context.interfaceAdapter = this.interfaceAdapter;
     this.repl.context.accounts = accounts;
-
-    this.hydrateUserDefinedVariables();
   }
 
   provision() {
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -86,12 +86,14 @@ class Console extends EventEmitter {
   }
 
   hydrateUserDefinedVariables() {
-    if (
-      (!this.options.console || !this.options.console.require) &&
-      !this.options.require &&
-      !this.options.r ||
-      this.options["require-none"]
-    ) return;
+  // exit if feature should be disabled
+  if (this.options["require-none"]) return;
+
+  // exit if no hydrate options are set
+  if (!this.options.console || !this.options.console.require) &&
+    !this.options.require &&
+    !this.options.r 
+  ) return;
 
     const requireFromPath = target => {
       return path.isAbsolute(target) ?
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -114,7 +114,7 @@ class Console extends EventEmitter {
       "either a string or an array. If you specify an array, its members " +
       "must be paths or objects containing at least a `path` property.";
 
-    const requireValue = this.options.require || this.options.console.require;
+    const requireValue = this.options.r || this.options.require || this.options.console.require;
 
     if (typeof requireValue === "string") {
       addToContext(requireFromPath(requireValue));
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -65,18 +65,7 @@ class Console extends EventEmitter {
         eval: this.interpret.bind(this)
       });
 
-      let accounts;
-      try {
-        accounts = await this.interfaceAdapter.getAccounts();
-      } catch {
-        // don't prevent Truffle from working if user doesn't provide some way
-        // to sign transactions (e.g. no reason to disallow debugging)
-        accounts = [];
-      }
-
-      this.repl.context.web3 = this.web3;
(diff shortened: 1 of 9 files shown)


eggplantzzz:
I want to write some tests for this. I may just instantiate an instance of it and unit test `setUpEnvironment`. Then we can check to make sure `this.repl.context` has what we expect it to have. Any other ideas?

packages/core/lib/console.js
```diff
@@ -89,12 +78,71 @@ class Console extends EventEmitter {
...
+      // don't prevent Truffle from working if user doesn't provide some way
+      // to sign transactions (e.g. no reason to disallow debugging)
+      accounts = [];
+    }
+
+    this.repl.context.web3 = this.web3;
+    this.repl.context.interfaceAdapter = this.interfaceAdapter;
+    this.repl.context.accounts = accounts;
+
+    this.hydrateUserDefinedVariables();
```
cds-amal, line 143: The current ordering would allow our globals to be over written. 

```suggestion
    this.hydrateUserDefinedVariables();
    
    this.repl.context.web3 = this.web3;
    this.repl.context.interfaceAdapter = this.interfaceAdapter;
    this.repl.context.accounts = accounts;

```

cds-amal (changes requested):
I looooooove this PR @eggplantzzz . TY TY

cds-amal:
Discussion: does it make sense to have a `--ignore-user-imports` to disable this behavior from the command line?

It occurred to me while testing and I wanted to try the console without editing `truffle-config`. Would certainly have been quicker.

gnidan:
> Discussion: does it make sense to have a --ignore-user-imports to disable this behavior from the command line?

I feel more strongly about the other way (we should allow `truffle console --require="<...>"`), but we might want this direction also. For consistency with compile, I propose `--require-none`.

packages/core/lib/console.js
gnidan: Agreed that we probably want to make sure our globals are safe. We're giving a namespacing mechanism, anyway!

eggplantzzz: I would say just allow it. Maybe they want their own version of web3 or something. These globals are, after all, for the user's convenience.

eggplantzzz:
I still would like to implement the `--require` flag for the command line.

packages/core/lib/console.js
cds-amal: We could document that Truffle will set the context of `web3`, `interfaceAdapter` and `accounts` after user scripts are loaded and recommend a pattern for overriding. For example, creating an `overrideTruffleGlobals(global)` snippet they invoke at the console prompt.

eggplantzzz: I don't understand why we should care about this? Truffle itself doesn't use them. Let them make their own custom web3 or accounts if they want. Why is this a problem?

cds-amal: This ordering gives us flexibility to insert more globals in the future without worrying about users snippets and breaking behavior.

eggplantzzz:
I added `--require-none` to the help for `truffle develop` and `truffle console`. Additionally, only one of `-r`, `--require`, and values for `console.require` (in the `truffle-config.js`) is accepted and processed and in that order of precedence. It will use the first one it finds values for.

cds-amal (approved):
Nice tests! @eggplantzzz

packages/core/lib/console.js
```diff
@@ -89,12 +78,78 @@ class Console extends EventEmitter {
...
     }
   }
 
+  hydrateUserDefinedVariables() {
+    if (
+      (!this.options.console || !this.options.console.require) &&
+      !this.options.require &&
+      !this.options.r ||
+      this.options["require-none"]
+    ) return;
```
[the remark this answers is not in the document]
eggplantzzz: I like it better!

PR #4128, opened 2021-06-21
Bug fix: Add crypto lib and method to silence warning

eggplantzzz:
To silence a warning that occurs whenever a command is run in the console, this PR includes the crypto library with the get-random-values package. This warning comes from a dependency of an ENS package used by Truffle which warns if there is not a good source for randomness.
Screen Shot 2021-06-21 at 5 04 44 PM

This PR prevents the above warning from displaying in these cases. This is identical to what is done in `cli.js` [here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/cli.js#L6-L10).

### DIFF
--- a/packages/core/lib/console-child.js
+++ b/packages/core/lib/console-child.js
@@ -4,6 +4,12 @@ const Config = require("@truffle/config");
 const Web3 = require("web3");
 const yargs = require("yargs");
 
+const crypto = require("crypto");
+global.crypto = crypto;
+// we need to make sure this function exists so ensjs doesn't complain
+// it requires getRandomValues for some functionalities
+global.crypto.getRandomValues = require("get-random-values");
+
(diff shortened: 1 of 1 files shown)


haltman-at (approved):
Oh, thanks for getting this!

packages/core/lib/console-child.js
```diff
@@ -4,6 +4,12 @@ const Config = require("@truffle/config");
 const Web3 = require("web3");
 const yargs = require("yargs");
 
+const crypto = require("crypto");
+global.crypto = crypto;
+// we need to make sure this function exists so ensjs doesn't complain
+// it requires getRandomValues for some functionalities
```
cds-amal, line 10: Nice hint

cds-amal (approved):
LGTM

Issue #4120, opened 2021-06-17
Debugger - TypeError: Cannot read property 'definition' of undefined


regata:
## Issue

Debugger is crashing when trying to step through the code. Any help would be appreciated.

```
UniswapV2Pair.sol:

471:         address _token1 = token1;
472:         require(to != _token0 && to != _token1, 'UniswapV2: INVALID_TO');
473:         if (amount0Out > 0) _safeTransfer(_token0, to, amount0Out); // optimistically transfer tokens
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                   
debug(development:0xed24c113...)> o
⠋ Stepping...TypeError: Cannot read property 'definition' of undefined
    at /Users/gnidan/src/work/truffle/packages/debugger/lib/data/selectors/index.js:1045:53
    at /home/arbitrageur/arbitrage/node_modules/reselect/lib/index.js:76:25
    at /home/arbitrageur/arbitrage/node_modules/reselect/lib/index.js:36:25
    at /home/arbitrageur/arbitrage/node_modules/reselect/lib/index.js:90:33
    at Function.<anonymous> (/home/arbitrageur/arbitrage/node_modules/reselect/lib/index.js:36:25)
    at /home/arbitrageur/arbitrage/node_modules/reselect-tree/lib/index.js:85:1
    at /home/arbitrageur/arbitrage/node_modules/reselect/lib/index.js:86:37
    at /home/arbitrageur/arbitrage/node_modules/reselect/lib/index.js:36:25
    at runSelectEffect (/home/arbitrageur/arbitrage/node_modules/@redux-saga/core/dist/redux-saga-core.dev.cjs.js:734:26)
    at runEffect (/home/arbitrageur/arbitrage/node_modules/@redux-saga/core/dist/redux-saga-core.dev.cjs.js:1208:7)
    at digestEffect (/home/arbitrageur/arbitrage/node_modules/@redux-saga/core/dist/redux-saga-core.dev.cjs.js:1275:5)
    at next (/home/arbitrageur/arbitrage/node_modules/@redux-saga/core/dist/redux-saga-core.dev.cjs.js:1165:9)
    at currCb (/home/arbitrageur/arbitrage/node_modules/@redux-saga/core/dist/redux-saga-core.dev.cjs.js:1255:7)
[...]

PR #4131, opened 2021-06-22
Bug fix: Fix debugger crash from cross-compilation revert

haltman-at:
Addresses #4120.

This PR fixes a bug in how I implemented #4015.  In it, when decoding an error, we attempt to get the node where the error occurred from the stacktrace.  However, I attempted to look up the node, by its ID, in `data.current.scopes.inlined`.  This doesn't work, because `data.current.scopes.inlined` only contains information about the current compilation, and the error might have occurred in a different compilation.  So, instead, we have to look it up in `data.views.scopes.inlined`, which contains information about all compilations.  Looking it up in there requires knowing the source ID in addition to just the node ID, so we now get both the source ID and the node ID from the stacktrace, and then use those with `data.views.scopes.inlined` to get the actual node.  We then proceed as before.

### DIFF
--- a/packages/debugger/lib/data/selectors/index.js
+++ b/packages/debugger/lib/data/selectors/index.js
@@ -1026,13 +1026,13 @@ const data = createSelectorTree({
     ),
 
     /**
-     * data.current.errorNodeId
+     * data.current.errorLocation
      * note: we can't get the actual node from stacktrace,
-     * it only stores the ID
+     * it doesn't store that
      */
-    errorNodeId: createLeaf(
+    errorLocation: createLeaf(
       [stacktrace.current.innerReturnPosition, stacktrace.current.lastPosition],
-      (innerLocation, lastLocation) => ((innerLocation || lastLocation || {}).node || {}).id
+      (innerLocation, lastLocation) => innerLocation || lastLocation || {}
     ),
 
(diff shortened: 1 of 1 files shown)


packages/debugger/lib/data/selectors/index.js
```diff
@@ -1041,8 +1041,16 @@ const data = createSelectorTree({
...
-      (id, scopes) => id !== undefined ? scopes[id].definition : null
+      ["./errorLocation", "/views/scopes/inlined"],
+      (errorLocation, scopes) => {
+        const sourceId = (errorLocation.source || {}).id;
+        const astId = (errorLocation.node || {}).id;
+        if (sourceId !== undefined && astId !== undefined) {
+          return scopes[sourceId][astId].definition;
+        } else {
+          return null;
+        }
```
cds-amal, line 1052: This could be refactored as a ternary expression, but certainly not a blocker.

cds-amal (approved):
LGTM @haltman-at

packages/debugger/lib/data/selectors/index.js
haltman-at: Hm, yeah, not going to bother I don't think.

Issue #4134, opened 2021-06-23
`compile-solidity` package depends on an outdated solc-js version


cameel:
## Issue
`compile-solidity` depends on solc-js 0.6.0. This is an old version that had a bug that is the cause of https://github.com/ethereum/solidity/issues/11489 / #4039. The bug has been fixed a year ago (https://github.com/ethereum/solc-js/pull/470) in 0.6.9 so an update to at least that version is needed. Preferably Truffle should always use the latest one.

Note that Truffle used to depend on the latest solc-js from 0.6.x series but looks like it was changed to strictly 0.6.0 in #3975.

fainashalts:
Hi @cameel thanks for bringing this up! We have deliberately kept this version of solc-js to maintain our support for Node 10 (our tests failed when we tried to make the upgrade in the past). When we drop support for Node 10 in the near future, we will revisit this.

cameel:
Maybe it would be a good idea to at least bump it back up to 0.6.9 to make it work with 0.8.2? Truffle used to use that (or higher) version due to `^0.6.0` up to 5.3.2.

By the way, what was the exact issue with node 10? We did have some yet unresolved issues with that version (https://github.com/ethereum/solidity/issues/9297) so I'm curious what that might be about.

haltman-at:
I don't remember the exact issue, so I'm making a branch where we bump it back up to 0.6.9 and we can see!

haltman-at:
OK, I put up #4136 and... it's actually passing just fine?  Huh.  I guess maybe there was no problem after all.

haltman-at:
OK, closing this for now, now that the immediate problem has been addressed.  We likely should update further, though.  Feel free to let us know if you want us to reopen this, or just file a new issue.

PR #4136, opened 2021-06-23
Dependency update: Update solc/wrapper to 0.6.9

haltman-at:
Attempts to address #4134... we'll see where exactly the tests fail.


haltman-at:
Um, hm.  Everything passed.  Maybe there's no problem after all?

haltman-at:
Note: We could try upgrading this even further; I just started with 0.6.9.  But of course we could go further as long as the tests still pass...

gnidan (approved):

Issue #4119, opened 2021-06-17
Absolute paths in contract metadata


cameel:
## Issue
When compiling a Solidity project, Truffle gives the compiler absolute paths of all source files. Relative paths are used only for source located in libraries found by the [resolver](https://github.com/trufflesuite/truffle/tree/develop/packages/resolver) (e.g. when importing from `node_modules/`) but the `fs` module of the resolver sometimes creates absolute paths too.

The use of absolute paths to refer to Solidity source files makes it harder to reproduce the same bytecode on different machines because they are included in contract metadata,the hash of which is always appended at the end of the bytecode. It also potentially leaks user's local information (the name of the home directory if the contract is stored there).

Truffle uses [Standard JSON](https://docs.soliditylang.org/en/latest/using-the-compiler.html#compiler-input-and-output-json-description), which [gives it complete control over paths in compiler's VFS](https://docs.soliditylang.org/en/latest/path-resolution.html#initial-content-of-the-virtual-filesystem). The only real constraint is that they should match imports. Using relative paths for files from main project directory should be completely doable according to my knowledge.

There are two potential problems:
1. Files located somewhere outside of the main project directory and not being a part of a library.
2. Absolute paths used in imports.

I see several ways to deal with them, depending on how backwards-compatible this change needs to be:
1. Keep using absolute paths in those cases but warn the user that his file structure may leak into metadata.
2. Add extra features to keep fully support such use cases. For example let the user have multiple project directories (declared in the config file) so that paths can be made relative to them.
3. Disallow them completely (that's for example what Hardhat does).

## Context
We're also currently working on solving this problem in the Solidity compiler (https://github.com/ethereum/solidity/issues/11408, https://github.com/ethereum/solidity/issues/11409, https://github.com/ethereum/solidity/issues/11410). Unfortunately this will only help people who use the CLI. Truffle's use of Standard JSON mode isolates it from such changes as it's the path resolver that determines paths here.

This problem is also a direct cause of #1621 (and https://github.com/trufflesuite/truffle-compile/issues/77). 

## Steps to Reproduce
```bash
mkdir /tmp/path-test/
cd /tmp/path-test/
truffle init
echo 'module.exports.compilers.solc.version = "0.8.5";' >> truffle-config.js

npm install @openzeppelin/contracts
truffle create contract C
echo 'import "../node_modules/@openzeppelin/contracts/utils/Arrays.sol";' >> contracts/C.sol
truffle compile
[...]

PR #4137, opened 2021-06-23
Replace project directory in metadata with "project://"

haltman-at:
Addresses #4119.

Per discussion there, this PR changes our input to the Solidity compiler so that the Truffle project directory will be replaced (in input to the Solidity compiler) with `"project://"`.  In addition to affecting the metadata, this also affects the `absolutePath` field on the AST, although we don't currently using that.  It does *not* affect the `sourcePath` field in the artifacts.

Note that this only affects sources that live in the project directory.  Sources imported from, e.g., NPM, will be unaffected.  However, since sources imported from NPM will be given by an NPM path rather than an absolute file path, this is fine.

What about FS sources living outside the project directory?  Well, those probably just won't work anymore, as the compiler won't be able to find them.

To implement this, I reused existing functionality from `compile-vyper`.  (Well, the functionality lives in `compile-common`, but it was added for Vyper purposes.)  We already do something similar for Vyper, although slightly differently and for different reasons.  I just expanded the functionality to cover replacing a prefix instead of just removing a prefix, and invoked it in `compile-solidity` as well.

I also added a test, and moved a comment I noticed was in the wrong place.

### DIFF
--- a/packages/compile-solidity/run.js
+++ b/packages/compile-solidity/run.js
@@ -14,11 +14,18 @@ async function run(rawSources, options, language = "Solidity") {
 
   // Ensure sources have operating system independent paths
   // i.e., convert backslashes to forward slashes; things like C: are left intact.
+  // we also strip the project root (to avoid it appearing in metadata)
+  // and replace it with "project://"
   const {
     sources,
     targets,
     originalSourcePaths
-  } = Common.Sources.collectSources(rawSources, options.compilationTargets);
+  } = Common.Sources.collectSources(
+    rawSources,
+    options.compilationTargets,
+    options.working_directory,
+    "project:/" //only one slash, other will already be present
+  );
 
   // construct solc compiler input
   const compilerInput = prepareCompilerInput({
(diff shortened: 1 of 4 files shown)


packages/compile-solidity/run.js
```diff
@@ -14,11 +14,18 @@ async function run(rawSources, options, language = "Solidity") {
...
   const {
     sources,
     targets,
     originalSourcePaths
-  } = Common.Sources.collectSources(rawSources, options.compilationTargets);
+  } = Common.Sources.collectSources(
+    rawSources,
+    options.compilationTargets,
+    options.working_directory,
+    "project:/" //only one slash, other will already be present
```
eggplantzzz, line 27: Maybe this is a little pedantic but is it technically more proper to have this argument be the whole string? Then we should strip off the relative path portion and replace it with "exactly" what we will be using? Like what if (and yeah, I think this is pretty pedantic) one day we decide to use "project::" instead of "project:/"?

eggplantzzz:
I find it pretty surprising that this is already supported by the Solidity compiler.

packages/compile-solidity/run.js
gnidan: agree about replacement style - better to replace the whole thing

cds-amal: +1 on that as well

haltman-at: OK, I've changed this.

cds-amal (approved):
Looks good @haltman-at

PR #4144, opened 2021-06-25
Compiler tezos upgrade

bergarces:
- @truffle/codec: Specifies that the contracts passed are `EvmCompiledContract`. And stops exporting methods that are not used by other packages.
- @truffle/compile-common: Modifies `CompiledContract` type so that it can be `EvmCompiledContract` or `TezosCompiledContract`.
- @truffle/compile-common: Adds `TezosCompiler` so that it can be used by invidual compilers (Michelson, LIGO and SmartPy).
- @truffle/compile-common: `NewToLegacy` is modified so that it's only applied to EvmCompiledContract.
- @truffle/compile-solidity: Updated to use new CompiledContract and artifact format.
- @truffle/compile-vyper: Updated to use new CompiledContract and artifact format.
- @truffle/external-compile: Updated to use new CompiledContract and artifact format.
- @truffle/compile-michelson: New compiler for Michelson files.
- @truffle/contract-schema: Add two new fields: `architecture` and `michelson`.
- @truffle/config: Add Michelson compiler to the defaults.
- @truffle/contract-sources: Add tz to the list of supported extensions.
- @truffle/workflow-compile: Add Michelson compiler to the list of compilers.


packages/compile-michelson/src/compileStrategy.ts
```diff
@@ -0,0 +1,61 @@
+import { CompilerResult, ICompileStrategy } from "@truffle/compile-common";
+import { Parser } from '@taquito/michel-codec';
+import { readFileSync } from "fs";
+import path from "path";
+
+import { version } from "../package.json";
+
+const compiler = {
```
haltman-at, line 8: Huh, is this really right?  You're setting the compiler name and version, not to that of the compiler being invoked, but to that of the Truffle package?  Could you explain the reasoning behind this?

packages/codec/lib/compilations/utils.ts
```diff
@@ -11,7 +11,7 @@ import * as Common from "@truffle/compile-common";
 import { Compilation, Contract, Source, VyperSourceMap } from "./types";
 
 export function shimCompilations(
-  inputCompilations: Common.Compilation[],
+  inputCompilations: (Common.Compilation & { contracts: Common.EvmCompiledContract[] })[],
```
haltman-at, line 14: Could we assign some type name to this so we're not writing this out longhand all the time?

packages/codec/lib/compilations/utils.ts
```diff
@@ -22,8 +22,8 @@ export function shimCompilations(
   );
 }
 
-export function shimCompilation(
-  inputCompilation: Common.Compilation,
+function shimCompilation(
```
haltman-at, line 25: I'd leave the `export`s on these functions; they may not currently be used externally but they were written with the intent that they might be.

haltman-at (commented):
I didn't look much at the new code (as opposed to the changes to existing code) so this is not a full review, but I did have a few comments.

packages/compile-michelson/src/compileStrategy.ts
bergarces: There is no compiler being invoked. Michelson contracts are just parsed from plain Michelson to JSON Michelson using Taquito parser.

I'm open to suggestions for the name and version, but I just followed what was done on the AlphaTez branch for Michelson. https://github.com/trufflesuite/truffle/blob/alphaTez/packages/compile-michelson/src/index.ts

The Ligo compiler will have the version from the compiler that is invoked though.

packages/codec/lib/compilations/utils.ts
bergarces: Fair enough, I'll leave them exported as they were.

packages/codec/lib/compilations/utils.ts
bergarces: This was the easier way I came across to represent it given that it only needs to be specified twice.

Otherwise it involves creating an EvmCompilation type that contains EvmCompiledContract. It's doable, but I'd argue it's more cumbersome for these two instances.

bergarces: In retrospective. I've extracted the type at the top of the file without exporting it. That way it can be reused without repeating it at all.

packages/compile-michelson/src/compileStrategy.ts
haltman-at: Oh, huh.  I wonder if there is some better way to indicate this though.  (Some fictional value for `name`, and maybe omitting the version entirely or something?)  @gnidan, do you have any thoughts on this...?

eggplantzzz:
Closing this for PR list maintenance.

PR #4149, opened 2021-06-29
Internal update: clean dependencies

cds-amal:
This PR removes unused dependencies from project workspaces .

| Workspace          | Flagged dependencies     |   Action                   |
|----------------------|-------------------------------|-------------------------|
| contract-schema      | crypto-js                     | removed                 |
| | | |
| source-fetcher       | @types/node                   | removed                 |
| | | |
| reporters            | node-emoji                    | removed                 |
| | | |
| hdwallet-provider    | any-promise                   | removed                 |
|                      | bindings                      | removed                 |
| | | |
| preserve-to-filecoin | node-fetch                    | removed                 |
| | | |
| require                      | web3                          | removed                 |
| | | |
| compile-common      | mocha                         | move to devDep |
|                                    | ts-node                       | move to devDep |
| | | |
| debug-utils                | highlight.js                 | removed                 |
| | | |
| resolver                      | supports-color            | removed                 |
| | | |
| db                               | pascal-case                 | removed                 |
| | | |
| migrate                       | web3                          | removed                 |
| | | |
| debugger                    | redux-cli-logger          | removed                 |
|                                    | web3-eth-abi               | removed                 |
| | | |
| core              | @truffle/contract-sources       | removed                 |
|                      | app-module-path                    | move to devDep                 |
|                      | node-ipc                                  | removed                 |
| | | |


packages/plugins/package.json
```diff
@@ -35,7 +36,6 @@
   },
   "dependencies": {
     "@truffle/error": "^0.0.8",
-    "app-module-path": "^2.2.0",
```
gnidan, line 38: This needs to stay a regular dependency (it's used by `lib/Plugins.ts`)

packages/source-fetcher/package.json
```diff
@@ -26,7 +26,6 @@
   },
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@types/node": "^15.0.1",
```
gnidan, line 29: This package's tsconfig [references](https://github.com/trufflesuite/truffle/blob/develop/packages/source-fetcher/tsconfig.json#L21) this. Make sure you can remove it there before saying it's not a dependency.

Although it probably should be a devDep.

packages/debug-utils/package.json
```diff
@@ -23,7 +23,6 @@
     "bn.js": "^5.1.3",
     "chalk": "^2.4.2",
     "debug": "^4.3.1",
-    "highlight.js": "^10.4.0",
```
gnidan, line 26: Maybe make sure that chromafi doesn't need this as a peerDependency?

packages/core/package.json
```diff
@@ -39,17 +38,12 @@
...
-    "@truffle/preserve-to-buckets": "^0.2.3",
-    "@truffle/preserve-to-filecoin": "^0.2.3",
-    "@truffle/preserve-to-ipfs": "^0.2.3",
     "@truffle/provider": "^0.2.33",
     "@truffle/provisioner": "^0.2.21",
     "@truffle/require": "^2.0.66",
     "@truffle/resolver": "^7.0.16",
     "@truffle/source-fetcher": "^0.5.1",
     "@truffle/workflow-compile": "^3.2.17",
-    "app-module-path": "^2.2.0",
```
gnidan, line 52: This needs to be made a devDep because it's [used](https://github.com/trufflesuite/truffle/blob/develop/packages/core/test/lib/run.js#L15) by tests

packages/core/package.json
```diff
@@ -39,17 +38,12 @@
     "@truffle/migrate": "^3.2.31",
     "@truffle/plugins": "^0.2.3",
     "@truffle/preserve": "^0.2.3",
-    "@truffle/preserve-fs": "^0.2.3",
-    "@truffle/preserve-to-buckets": "^0.2.3",
-    "@truffle/preserve-to-filecoin": "^0.2.3",
-    "@truffle/preserve-to-ipfs": "^0.2.3",
```
gnidan, line 45: These need to stick around because they are included as [`defaultPlugins`](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/commands/preserve/plugins.js#L3-L8)

gnidan (changes requested):
Noticed some errors here

packages/debug-utils/package.json
cds-amal: `highlight.js` is a [dependency of chromafi](https://github.com/F1LT3R/chromafi/blob/master/package.json#L29)

packages/source-fetcher/package.json
cds-amal: I'm not sure what the intention was here. According to the [docs](https://www.typescriptlang.org/tsconfig#typeRoots)

> If typeRoots is specified, **_only_** packages under typeRoots will be included.  

[This config](https://github.com/trufflesuite/truffle/blob/develop/packages/source-fetcher/tsconfig.json#L17-L21) implies that `node` is excluded because it wouldn't appear in `typeRoots` list of paths

```
    "typeRoots": [
      "./typings",
      "../../**/node_modules/@types/debug"
    ],
    "types": ["node"]
```
      
Have you used any tools to verify this behavior? I'm gonna spend an hr with [tsd](https://github.com/SamVerschueren/tsd) to see if it can help.

packages/core/package.json
cds-amal: moved to devDeps

packages/core/package.json
cds-amal: Reverted. I think we should have integration tests for these.

packages/source-fetcher/package.json
gnidan: Just try taking it out and see if it breaks

packages/compile-common/package.json
```diff
@@ -21,6 +21,8 @@
     "@truffle/contract-schema": "^3.4.1",
     "@types/fs-extra": "^8.1.0",
     "@types/mocha": "^5.2.7",
+    "mocha": "8.0.1",
```
eggplantzzz, line 24: There are tests in this package so I think we need mocha.

packages/migrate/package.json
```diff
@@ -26,8 +26,7 @@
     "@truffle/reporters": "^2.0.0",
     "@truffle/require": "^2.0.66",
     "emittery": "^0.4.1",
-    "glob": "^7.1.6",
-    "web3": "1.4.0"
+    "glob": "^7.1.6"
```
eggplantzzz, line 29: The almighty glob!

packages/resolver/package.json
```diff
@@ -33,8 +33,7 @@
     "debug": "^4.3.1",
     "detect-installed": "^2.0.4",
     "get-installed-path": "^4.0.8",
-    "glob": "^7.1.6",
-    "supports-color": "^7.1.0"
+    "glob": "^7.1.6"
```
eggplantzzz, line 36: !!

packages/source-fetcher/package.json
eggplantzzz: Yeah you should just see if this builds. This is a package that is not "required" and used really so you can't really tell if it is necessary until you clear out all `node_modules` and try to build. If TS doesn't complain afterwards, you are probably good.

eggplantzzz: Will people be trying to install this package into their projects and building with TS? :)

eggplantzzz:
I like it!

eggplantzzz (approved):

gnidan (approved):

PR #4153, opened 2021-06-30
remove dead infurakey reference

tansaku:
seems like this is dead code?  it's around in a few other places and could be a source of confusion

### DIFF
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -19,7 +19,6 @@
  */
 
 // const HDWalletProvider = require('@truffle/hdwallet-provider');
-// const infuraKey = "fj4jll3k.....";
 //
 // const fs = require('fs');
 // const mnemonic = fs.readFileSync(".secret").toString().trim();
--- a/packages/decoder/test/current/truffle-config.js
+++ b/packages/decoder/test/current/truffle-config.js
@@ -19,7 +19,6 @@
  */
 
 // const HDWallet = require("@truffle/hdwallet-provider");
-// const infuraKey = "fj4jll3k.....";
 //
 // const fs = require("fs");
 // const mnemonic = fs.readFileSync(".secret").toString().trim();
--- a/packages/decoder/test/legacy/truffle-config.js
+++ b/packages/decoder/test/legacy/truffle-config.js
@@ -19,7 +19,6 @@
  */
 
 // const HDWallet = require("@truffle/hdwallet-provider");
-// const infuraKey = "fj4jll3k.....";
 //
 // const fs = require("fs");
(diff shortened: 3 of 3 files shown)


haltman-at:
Hey, you seem to have opened this PR directly against `master`.  PRs should be be made against `develop`; could you change this?  Thank you!

As for the change itself... a better approach, rather than deleting that line, might be to update the examples to use the Infura key.  What do you think of that?

cds-amal:
Hi @tansaku !

This file is copied as `truffle-config.js` when a user initializes a new project (`truffle init [name]`). Those commented snippets are intended to help the user configure the different Truffle options. 

See [documentation](https://www.trufflesuite.com/docs/truffle/reference/configuration)

tansaku:
right @cds-amal - but shouldn't the infura secret key not be in a file committed to git?  

That infuraKey is not mentioned in the documentation you link to - shouldn't it be removed?

eggplantzzz:
Yeah we also have `mnemonic` that we don't reference anywhere else in the file though. I think they are just meant to be examples of how you might go about configuring your project. I don't know, I don't really know if we should delete them all - they might be helpful for people?

tansaku:
they might - but isn't putting the project secret in a plain text file that people might commit to a repo something we'd want to avoid people doing?

tansaku:
@haltman-at can re-open PR against develop once we're clear what we want  - as for updating the examples?  Sure but maybe let's start by deleting something not needed that's actively bad practice, i.e. encouraging folks to add project secrets to files they might check in ... then again, I'm not even sure which of the many files this thing is in are the right ones to edit

eggplantzzz:
Yeah, that's a good point that we shouldn't encourage people to record secrets this way. That is a convincing argument to me - what do you think @cds-amal and @haltman-at?

cds-amal:
Thanks @tansaku . I agree we shouldn't promote bad secrets practice. OTOH we should still offer helpful information to guide new users. Perhaps change the example to use the `dotfiles` module and warn not to upload their environment variables.

tansaku:
so which are the files that should the `const infuraKey` removed from?  everywhere it occurs in the repo?

davidmurdoch:
@tansaku, we chatted internally and agree that this reference to infuraKey should be removed, as well as any others. Can you update the PR to remove the other references as well? We've found two other references in our tests, which we figure would be good to update since people often use our tests as documentation.

https://github.com/trufflesuite/truffle/blob/abfb3646eea9d57e0a6f9186c12bb51680c181d8/packages/decoder/test/current/truffle-config.js#L22
and
https://github.com/trufflesuite/truffle/blob/abfb3646eea9d57e0a6f9186c12bb51680c181d8/packages/decoder/test/legacy/truffle-config.js#L22

tansaku:
hopefully this new PR does the trick https://github.com/trufflesuite/truffle/pull/4164

PR #4155, opened 2021-07-01
Dependency update: Update web3 to 1.4.0

haltman-at:
Updates web3 to 1.4.0.

This PR does *not* add support for access lists.  I think that's better saved for a separate later PR.


cds-amal (approved):
LGTM and the test are passing!

PR #4157, opened 2021-07-01
Bug fix: Change "project://" to "project:/"

haltman-at:
Alteration to #4137.  Double-slashes break the ability of contracts in `test` to import projects in `contract`.  Replacing with single slashes so that these imports can work.

### DIFF
--- a/packages/compile-solidity/run.js
+++ b/packages/compile-solidity/run.js
@@ -15,7 +15,7 @@ async function run(rawSources, options, language = "Solidity") {
   // Ensure sources have operating system independent paths
   // i.e., convert backslashes to forward slashes; things like C: are left intact.
   // we also strip the project root (to avoid it appearing in metadata)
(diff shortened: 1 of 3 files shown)


cds-amal:
I'm worried that it took a smoke test to discover this issue. How should this be tested?

haltman-at:
Well ideally we'd have a test where we set up a Truffle project that includes a Solidity test.  Personally I'd be fine merging this w/o a test (I'd like to get this released today), but if it's not too hard... wondering where we might have similar tests already, maybe we can do it w/ a bit of copypaste.

haltman-at:
Ah -- I think I found a good place to add a test.  We already have tests of Solidity testing, they just don't import any contracts.  We can make one that imports Migrations.

gnidan (approved):

cds-amal (approved):

PR #4168, opened 2021-07-08
fix inferLanguage error with undefined sourcePath

PierreJeanjacquot:
when sourcePath is `undefined`, `inferLanguage` throws `TypeError: Cannot read property 'endsWith' of undefined`.

### DIFF
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -355,6 +355,7 @@ function inferLanguage(
     if (compiler.name === "vyper") {
       return "Vyper";
     } else if (compiler.name === "solc") {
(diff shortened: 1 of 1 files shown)


haltman-at (commented):
Oh, thanks for catching this!  One request: Could you add a comment explicitly noting the assumption you're making, that sources compiled with solc but with absent `sourcePath` are assumed to be Solidity?  Thank you!

cds-amal (approved):
Thanks for this @PierreJeanjacquot !

PierreJeanjacquot:
@haltman-at sure, I will add it

haltman-at (approved):
Great, thanks for catching this!

Issue #4152, opened 2021-06-30
truffle unbox failed


RunHippo:
- The Problem
truffle unbox webpack

```
This directory is non-empty...
? Proceed anyway? (Y/n) 
Starting unbox...
=================

? Proceed anyway? Yes
✔ Preparing to download box
✖ Downloading
Unbox failed!
✖ Downloading
Unbox failed!
Error: Error connecting to github.com. Please check your internet connection and try again.
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/box/dist/lib/utils/unbox.js:45:1
    at Generator.throw (<anonymous>)
    at rejected (/usr/local/lib/node_modules/truffle/build/webpack:/packages/box/dist/lib/utils/unbox.js:6:41)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)

```
Local Environment

```
Truffle v5.3.13 (core: 5.3.13)
Node v14.16.0
```

yipu0v0:
Same
```
> truffle unbox pet-shop

Starting unbox...
=================
✔ Preparing to download box
✖ Downloading
Unbox failed!
✖ Downloading
Unbox failed!
Error: Error connecting to github.com. Please check your internet connection and try again.
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/box/dist/lib/utils/unbox.js:45:1
    at Generator.throw (<anonymous>)
    at rejected (/usr/local/lib/node_modules/truffle/build/webpack:/packages/box/dist/lib/utils/unbox.js:6:41)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
Truffle v5.3.13 (core: 5.3.13)
Node v14.16.1
```

And connection to github.com works well using `wget github.com`.

Using proxy or not does not effect the result.

cds-amal:
Hi @RunHippo,

We were unable to reproduce this. Though we have seen it when we simulate network connection issues. Could it be an internet issue on your side?

yipu0v0:
[](url)
![屏幕快照 2021-07-01 下午4 39 25](https://user-images.githubusercontent.com/5365604/124094114-0c80a980-da8b-11eb-9685-e1fd2e5b4454.png)
I think the output of error omit necessary key information. URL and error should be printed, not just `Error connecting to github.com`. Actually I can connect to github.com in my terminal.

cds-amal:
[...]

PR #4169, opened 2021-07-08
Preserve non 404 errors when unboxing

cds-amal:
This PR preserves `axios` errors that could occur during box validation. Previously all non 404 errors were assumed to be a network issue. Hopefully we'll have better information to support issues like #4152

### DIFF
--- a/packages/box/lib/utils/unbox.ts
+++ b/packages/box/lib/utils/unbox.ts
@@ -25,9 +25,11 @@ async function verifyVCSURL(url: string) {
       .replace("github.com", "raw.githubusercontent.com")
       .replace(/#.*/, "")}/master/truffle-box.json`
   );
+
+  const testUrl = `https://${configURL.host}${configURL.path}`;
   try {
     await axios.head(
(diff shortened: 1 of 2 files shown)


packages/box/lib/utils/unbox.ts
```diff
@@ -25,9 +25,11 @@ async function verifyVCSURL(url: string) {
       .replace("github.com", "raw.githubusercontent.com")
       .replace(/#.*/, "")}/master/truffle-box.json`
   );
+
+  const testUrl = `https://${configURL.host}${configURL.path}`;
   try {
     await axios.head(
-      `https://${configURL.host}${configURL.path}`,
+      testUrl,
```
eggplantzzz, line 32: Should we name this something more like `repoUrl`? Not necessary but just an idea, it seems a little more explicit.

eggplantzzz (approved):
I like it!

packages/box/lib/utils/unbox.ts
cds-amal: Yes, this is clearer.

PR #4176, opened 2021-07-12
Internal improvement: Replace always-true expression with true

haltman-at:
This expression, `this.config.fetchExternal`, is always true due to being in an `else` block for an `if (!this.config.fetchExternal)`.  So, I just replaced it with `true`.  Presumably this is here because there didn't used to be a branch there, but now there is, so, yeah, this can now be simplified.


cds-amal (approved):
Looks good!

PR #4177, opened 2021-07-13
Enhancement: Add syntax highlighting for operators and punctuation (in debugger)

haltman-at:
I recently published highlightjs-solidity 1.2.0, which adds syntax highlighting for operators and punctuation.

This PR updates us to that version of highlightjs-solidity, and adds colors in `trufflePalette` for operators and punctuation.  (And also the lack of a color for `property`, in case that gets added in the future.)

I went with blue for operators and purple for punctuation, and, well, it looked fine when I tried it so I didn't revise it any further.


cds-amal (approved):
Can't wait to see how it looks.

PR #4178, opened 2021-07-13
Run all integration tests

cds-amal:
This PR allows CI to run all integration tests and follows web3's lead to pin GETH to v1.10.3. It should be noted that web3 didn't document a reason to pin and we should probably follow up with them (@gnidan ?)


haltman-at (approved):
I figure we should merge this quickly and we can worry about unpinning later!

Issue #2080, opened 2019-06-06
`Truffle test` should allow user to run a specific test using `mocha -g`


onggunhao:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

Following up from Twitter conversation with @gnidan - ref https://twitter.com/gnidan/status/1136386257481863169

---------------------------

## Issue

We should have a way to run individual truffle tests, similar to how `mocha -g` works (greps for a test string).
* There is an alternative way to do this by specifying `.only` in Mocha tests
* Being able to pass in `-g` allows us to run specific tests, e.g. in CI

## Steps to Reproduce

```
truffle test -- -g "name of test"
```

## Follow-ups

I'll be working on this, which is similar to issue #7 opened by @oed.

onggunhao:
@gnidan, I splunked through the codebase. Can I ask for your feedback on developer UX and codebase intuition?

# Developer experience

```shell
usage:
      "truffle test [<test_file>] [--compile-all] [--network <name>] [--verbose-rpc] [--show-events] **[--grep]**",
    options: [
      {
        option: "--grep",
        description:
          "Only run tests matching this string or regexp, similar to `mocha --grep`"
      },
```

# Implementation

The flow of execution of `truffle test` is the following (all in `truffle-core`):
-> `cli.js:45` & `lib/command.js:110-113`: parses args into `options` object
-> `lib/commands/test.js:112`: 

) and sends it to `command.run` (:56)  
-> `lib/command.js:run` parses args into options object (:110), passes it to `lib/commands/test.js` (:113)
-> `lib/commands/test.js` passes it to 

`/truffle-core/lib/commands/test.js` parses arguments from `truffle-core/cli.js`:
- [ ] Isolate the `-g` flag from arguments 

`

onggunhao:
@gnidan, I splunked through the codebase. Can I ask for your feedback on developer UX and codebase intuition?

# Developer experience

```shell
usage:
      "truffle test [<test_file>] [--compile-all] [--network <name>] [--verbose-rpc] [--show-events] **[--grep]**",
    options: [
      {
        option: "--grep",
        description:
          "Only run tests matching this string or regexp, similar to `mocha --grep`"
      },
```

# Implementation

I'm fairly inexperienced with javascript + mocha - so writing the code splunk trail here:

The flow of execution of `truffle test` is the following (all in `truffle-core`):
[...]

PR #4179, opened 2021-07-13
Feature: Add `--grep` option to `truffle test`

eggplantzzz:
This PR adds a `--grep|-g` command line option to `truffle test` and it works exactly like the `--grep` option for mocha. [See the mocha docs](https://mochajs.org/api/mocha). Basically you can provide a regex for mocha to search the test descriptions. It will then only run the tests whose descriptions match.

One thing to note with this PR is that there is an order of precedence when it comes to setting parameters in the `truffle-config.js` and on the command line with flags (`--bail`). We have decided to let the command line flags take precedence over the settings in `truffle-config.js`. So in other words if you set `bail = false` in your `truffle-config.js` and pass it `--bail` on the command line, `bail` will be set to `true`.

One other note here, I am trying to make it a convention in the command files to write options and their aliases in the form `--<option>|-o`. I have altered the algorithm that parses the options to determine whether the user is possibly passing in an unsupported option to deal with separating them.

This feature was suggested in [this issue](https://github.com/trufflesuite/truffle/issues/2080).

### DIFF
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -43,7 +43,10 @@ export const getInitialConfig = ({
       enabled: false,
       registryAddress: null
     },
-    mocha: {},
+    mocha: {
+      bail: false,
+      grep: null
+    },
     compilers: {
       solc: {
         settings: {
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -43,6 +43,7 @@ export const getInitialConfig = ({
       enabled: false,
       registryAddress: null
     },
+    mocha: {},
     compilers: {
       solc: {
         settings: {
@@ -94,6 +95,7 @@ export const configProps = ({
     compilers() {},
     ens() {},
     console() {},
+    mocha() {},
 
     build_directory: {
       default: () => path.join(configObject.working_directory, "build"),
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -121,7 +121,6 @@ class TruffleConfig {
     propertyNames.forEach(key => {
       try {
         if (typeof clone[key] === "object" && this._deepCopy.includes(key)) {
-
           this[key] = merge(this[key], clone[key]);
         } else {
           this[key] = clone[key];
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -23,7 +23,7 @@ class TruffleConfig {
     workingDirectory?: string,
     network?: any
   ) {
-    this._deepCopy = ["compilers"];
+    this._deepCopy = ["compilers", "mocha"];
     this._values = getInitialConfig({
       truffleDirectory,
       workingDirectory,
@@ -121,6 +121,7 @@ class TruffleConfig {
     propertyNames.forEach(key => {
       try {
         if (typeof clone[key] === "object" && this._deepCopy.includes(key)) {
+
           this[key] = merge(this[key], clone[key]);
         } else {
           this[key] = clone[key];
--- a/packages/core/lib/commands/test/index.js
+++ b/packages/core/lib/commands/test/index.js
@@ -51,7 +51,7 @@ const command = {
       `[--network <name>]${OS.EOL}                             ` +
       `[--verbose-rpc] [--show-events] [--debug] ` +
       `[--debug-global <identifier>] [--bail]${OS.EOL}                      ` +
-      `       [--stacktrace[-extra]]`,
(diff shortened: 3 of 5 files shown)


packages/config/src/configDefaults.ts
```diff
@@ -43,6 +43,10 @@ export const getInitialConfig = ({
       enabled: false,
       registryAddress: null
     },
+    mocha: {
```
eggplantzzz, line 46: Currently we are relying on the fact that arbitrary flags that get passed get tacked on to the config object. I decided to make it official and create defaults for mocha.

packages/config/src/index.ts
```diff
@@ -23,7 +23,7 @@ class TruffleConfig {
     workingDirectory?: string,
     network?: any
   ) {
-    this._deepCopy = ["compilers"];
+    this._deepCopy = ["compilers", "mocha"];
```
eggplantzzz, line 26: I'm not sure if there are any unexpected ramifications of putting this here. It looks like it only influences how `merge` works. Any other problems I didn't see?

packages/core/lib/commands/test/index.js
```diff
@@ -100,8 +100,8 @@ const command = {
         description: "Suppress all output except for test runner output."
       },
       {
-        option: "--bail",
-        description: "Bail after first test failure.  Alias: -b"
+        option: "--bail|-b",
```
eggplantzzz, line 103: Are we ok with making this the standard way to document aliases in the help? Currently in some places they are listed at the end of the description. See [debug's description for `--fetch-external`](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/commands/debug.js#L44). I would propose we change all instances of that to this format. Changing these will also prevent Truffle from warning the user for these aliases now that the algorithm for determining valid options has been updated. See the changes in this PR for `packages/core/lib/command.js`.

cds-amal: I prefer this location but wouldn't mind replacing the pipe with a comma: `--bail, b`

eggplantzzz: I'm not dead set against the `,` but prefer the pipe. What do other command line tools use for this kind of a situation? Should we take a vote?

packages/core/lib/commands/test/index.js
```diff
@@ -113,18 +113,32 @@ const command = {
...
+
+    // parse out command line flags to merge in to the config
+    const grep = options.grep || options.g;
+    const bail = options.bail || options.b;
 
-    const config = Config.detect(options);
+    const config = Config.detect(options).merge({
+      mocha: {
+        grep,
+        bail
```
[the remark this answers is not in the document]
eggplantzzz: Really we can add any mocha options we want. It will just make the docs more junked up the more we add. Can we consider adding other options to be outside the scope of this particular PR?

packages/core/lib/commands/test/index.js
cds-amal: I've seen cli option aliases listed with both commas or spaces. I don't think it matters in this case. it looks odd and I'm willing to look past :)

packages/core/lib/commands/test/index.js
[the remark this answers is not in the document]
cds-amal: My reasoning is that anyone using mocha filters will have a better experience with the full set of filter options - specify a search by string, or by regex and the ability to invert the search match. Can we vote on this one? Either way I want get this PR approved sooner rather than later.

packages/core/lib/commands/test/index.js
eggplantzzz: Well let's implement whatever is common practice. I'll go ahead and change this to `,`.

packages/core/lib/commands/test/index.js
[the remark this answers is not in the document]
eggplantzzz: I'm absolutely not opposed to adding more mocha options here. Let's implement them elsewhere.

packages/core/lib/commands/test/index.js
eggplantzzz: Well git does space pipe like this 
Screen Shot 2021-07-16 at 11 38 48 AM

packages/config/src/index.ts
cds-amal: Couldn't see references beyond this file.

cds-amal (approved):
Looks good to me!

cds-amal:
@eggplantzzz , I love that you can now use filters like `truffle test /pattern/i`

PR #4009, opened 2021-04-27
Configure eslint for mocha globals

bergarces:
- Added `.eslintrc.json` files to stop commit hooks from wrongly flagging errors.
- Fixed some of the flagged errors in test files.
- Not all errors from all packages have been fixed. Truffle test files that make use of global variables such as `artifacts` are still affected.
- An alternative to adding a `.eslintrc.json` file to every root folder, would be to add the filepath to the mocha exception list in the root `.eslintrc.json`.

Issue #3530, opened 2020-11-13
Internal issue: Husky frequently does nothing


haltman-at:
## Issue

Husky keeps insisting that no files match its rules, and so is currently doing nothing on JS and TS files!

haltman-at:
OK, it seems to be working again for whatever reason now, so I'll close this I guess?

haltman-at:
Reopening this, this has gotten to be a real pain.  Sometimes it works, sometimes it doesn't.  It seems to work after doing a `yarn add`, but then stops working a while later?  IDK.

cds-amal:
I'll review the work on the stalled PR and document the state of Husky.

haltman-at:
Husky now works!  Thanks so much @cds-amal!

PR #4180, opened 2021-07-13
Internal improvement: Add .eslintrc.json to Truffle projects in our source tree

haltman-at:
This is something of a follow-on to #4009.  Right now Husky is still broken (#3530), but once it's fixed, ESLint could potentially cause problems for `.js` files in Truffle projects that use Truffle-defined globals.  Specifically this would apply to test files that use globals, and migration scripts that use globals.

Since we have a bunch of those in our source tree, I added `.eslintrc.json` files to them that should take care of the problem.

There's one case I wasn't sure about, which is, what do we do about `initSource`?  If we put a `.eslintrc.json` in there, that means `truffle init` will create that file for people!  I went ahead and put it there anyway, but I did it in a separate commit since I wasn't so sure about it.  We could always change `truffle init` to ignore dotfiles, or something?

### DIFF
--- a/packages/core/lib/commands/init/initSource/.eslintrc.json
+++ b/packages/core/lib/commands/init/initSource/.eslintrc.json
@@ -1,0 +1,28 @@
+{
+  "overrides": [
+    {
+      "files": ["./test/**/*.js"],
+      "env": {
+        "mocha": true
+      },
+      "globals": {
+        "interfaceAdapter": "readonly",
+        "web3": "readonly",
+        "assert": "readonly",
+        "expect": "readonly",
+        "artifacts": "readonly",
+        "config": "readonly",
+        "debug": "readonly",
+        "contract": "readonly"
+      }
+    },
+    {
+      "files": ["./migration/**/*.js"],
+      "globals": {
+        "artifacts": "readonly",
+        "config": "readonly"
+      }
+    }
+  ]
+}
+
--- a/packages/core/lib/commands/init/initSource/.eslintrc.json
+++ b/packages/core/lib/commands/init/initSource/.eslintrc.json
@@ -17,7 +17,7 @@
       }
     },
     {
-      "files": ["./migration/**/*.js"],
+      "files": ["./migrations/**/*.js"],
       "globals": {
         "artifacts": "readonly",
         "config": "readonly"
@@ -25,4 +25,3 @@
     }
   ]
 }
-
--- a/packages/core/lib/commands/init/initSource/.eslintrc.json
+++ b/packages/core/lib/commands/init/initSource/.eslintrc.json
@@ -1,27 +1,3 @@
 {
-  "overrides": [
-    {
-      "files": ["./test/**/*.js"],
-      "env": {
-        "mocha": true
-      },
-      "globals": {
-        "interfaceAdapter": "readonly",
-        "web3": "readonly",
-        "assert": "readonly",
-        "expect": "readonly",
-        "artifacts": "readonly",
-        "config": "readonly",
-        "debug": "readonly",
-        "contract": "readonly"
-      }
-    },
-    {
-      "files": ["./migrations/**/*.js"],
-      "globals": {
-        "artifacts": "readonly",
-        "config": "readonly"
-      }
-    }
-  ]
+  "extends": ["../../../../../../.eslintrc.truffle.json"]
 }
--- a/packages/db/src/project/test/compilationSources/.eslintrc.json
+++ b/packages/db/src/project/test/compilationSources/.eslintrc.json
@@ -1,0 +1,27 @@
+{
+  "overrides": [
+    {
+      "files": ["./test/**/*.js"],
+      "env": {
+        "mocha": true
+      },
+      "globals": {
+        "interfaceAdapter": "readonly",
+        "web3": "readonly",
+        "assert": "readonly",
+        "expect": "readonly",
+        "artifacts": "readonly",
+        "config": "readonly",
+        "debug": "readonly",
+        "contract": "readonly"
+      }
+    },
+    {
+      "files": ["./migration/**/*.js"],
+      "globals": {
+        "artifacts": "readonly",
+        "config": "readonly"
+      }
+    }
+  ]
+}
--- a/packages/db/src/project/test/compilationSources/.eslintrc.json
+++ b/packages/db/src/project/test/compilationSources/.eslintrc.json
@@ -17,7 +17,7 @@
       }
     },
     {
-      "files": ["./migration/**/*.js"],
+      "files": ["./migrations/**/*.js"],
       "globals": {
         "artifacts": "readonly",
         "config": "readonly"
--- a/packages/db/src/project/test/compilationSources/.eslintrc.json
+++ b/packages/db/src/project/test/compilationSources/.eslintrc.json
@@ -1,27 +1,3 @@
 {
-  "overrides": [
-    {
-      "files": ["./test/**/*.js"],
-      "env": {
-        "mocha": true
-      },
-      "globals": {
-        "interfaceAdapter": "readonly",
-        "web3": "readonly",
-        "assert": "readonly",
-        "expect": "readonly",
-        "artifacts": "readonly",
-        "config": "readonly",
-        "debug": "readonly",
-        "contract": "readonly"
-      }
-    },
-    {
-      "files": ["./migrations/**/*.js"],
-      "globals": {
-        "artifacts": "readonly",
-        "config": "readonly"
-      }
-    }
-  ]
+  "extends": ["../../../../../.eslintrc.truffle.json"]
 }
--- a/.eslintrc.package.json
+++ b/.eslintrc.package.json
@@ -1,0 +1,28 @@
+{
+  "overrides": [
+    {
+      "files": ["./test/**/*.js"],
+      "env": {
+        "mocha": true
+      },
+      "globals": {
+        "interfaceAdapter": "readonly",
+        "web3": "readonly",
+        "assert": "readonly",
+        "expect": "readonly",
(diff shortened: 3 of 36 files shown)


cds-amal:
Ooh interesting! There's quite a bit of code duplication. I don't know the answer but I wonder if there's a way to leverage lerna for code sharing

cds-amal (commented):
@haltman-at Could we use one of [eslint's ignore configuration](https://eslint.org/docs/user-guide/configuring/ignoring-code#ignorepatterns-in-config-files) to protect these files?

If not, in case you want to do some linting, there is the option of using the `.eslintrc.js` [configuration file](https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats) and utilize inheritance to minimize duplication. What do you think?

haltman-at:
How exactly are you suggesting we use `ignorePatterns`?  Like we would have at the root a list of all Truffle projects within the source tree?  (We can't just ignore things within `migrations` or `test` folders, for obvious reasons!)  And yes ideally I would not like to turn off linting entirely, so I don't really like that solution.

I don't understand your second suggestion; I am using inheritance.  Could you clarify?

cds-amal:
I'm didn't have an exact suggestion the ignorePatterns setup but wanted to consider the mechanic for when Husky comes back online. 

eslint can be configured via a JavaScript file, `.eslintrc.js`, which gives a more flexibility to share common configuration though it's not strictly inheritance 😅 

```
const base = require("./up/the/tree/base.eslintrc.js")
config = {...base, /* customizations */ }
module.exports = config
```

haltman-at:
> I'm didn't have an exact suggestion the ignorePatterns setup but wanted to consider the mechanic for when Husky comes back online.

I mean, here's the PR.  It's up right now; I'm not waiting until Husky comes back online.  As I said, I do not see how the mechanic can be usefully applied here, so the only way we will end up using it is if someone else has a way.
 
> eslint can be configured via a JavaScript file, `.eslintrc.js`, which gives a more flexibility to share common configuration though it's not strictly inheritance sweat_smile
> 
> ```
> const base = require("./up/the/tree/base.eslintrc.js")
> config = {...base, /* customizations */ }
> module.exports = config
> ```

Well, we don't need to do our own custom inheritance, because, again, ESLint already has an inheritance mechanism, that this PR already uses; we don't need to reinvent that.

But, yes, I suppose using `.eslintrc.js` and `require` would allow us to centralize things somewhat and reduce code duplication.  I'm not entirely certain it's worth it, but it could be a good idea.  If we do do that, possibly we should similarly deduplicate the other duplicated `.eslintrc.json`?  As mentioned above, right now we have a lot of those due to #4009.

Alternatively these could be done as two separate steps...

packages/core/lib/commands/init/initSource/.eslintrc.json
```diff
@@ -0,0 +1,28 @@
+{
```
eggplantzzz, line 1: We should consider ignoring this in the webpack config. I don't think we want to copy this over for users when they do a `truffle init`, or do we?

eggplantzzz:
@haltman-at I think you are misunderstanding @cds-amal. When he is talking about an inheritance mechanism he is referring to having a base config with all of the shared configuration. This would prevent all this duplication. This seems like a good idea to me. What do you mean by "eslint has an inheritance mechanism that this PR already uses"?

packages/db/src/project/test/compilationSources/.eslintrc.json
```diff
@@ -0,0 +1,27 @@
...
+        "assert": "readonly",
+        "expect": "readonly",
+        "artifacts": "readonly",
+        "config": "readonly",
+        "debug": "readonly",
+        "contract": "readonly"
+      }
+    },
+    {
+      "files": ["./migration/**/*.js"],
```
eggplantzzz, line 20: All of these entries should be `./migrations/**/*.js`, right? They are currently all the singular form.

haltman-at: Oh crap, missed that, thanks!

haltman-at:
We have a base `.eslintrc.json` in the project root.  That is our base config with all the shared configuration.  The `.eslintrc.json` files I am adding use the `overrides` mechanism, causing them to inherit from the `.esllintrc.json` in the project root, and just add their own things.  Again, this is exactly the same as to how things were done in #4009, when Bernardo added a bunch `.eslintrc.json` files for our test files.

Now, I recognize that this use of inheritance does not achieve as much code deduplication as might be desired, and Amal suggested an alternative mechanism that would achieve that.  (I'm assuming that you'd want to apply that to the ones Bernardo added as well, not just these new ones?)  But this PR *is* already using inheritance, and the additional deduplication would not come from inheritance, but rather from the use of `.eslintrc.js` and `require`.

packages/core/lib/commands/init/initSource/.eslintrc.json
haltman-at: Yes, I mentioned this potential problem in the PR description; I agree we likely want to exclude it somehow.  The webpack config would be one way, although not necessarily the only way.  (E.g., it could also be done by modifying how the `truffle init` command works so that it doesn't copy hidden files, or by moving the file one directory up and modifying it appropriately.)  Do you think the webpack config is how it should be done?

eggplantzzz:
Oh sorry, I didn't realize that it already somehow inherited automatically from that. Just read about it in the docs!

I think what Amal was proposing was having a shared base configuration (maybe at the project root) and extending from that...maybe something [like using this mechanism](https://eslint.org/docs/user-guide/configuring/configuration-files#using-a-configuration-file).  @cds-amal is that what you were getting at?  That way we wouldn't have so much boilerplate.

packages/core/lib/commands/init/initSource/.eslintrc.json
eggplantzzz: If we go the way of adding it to the webpack config, it looks like [this][https://eslint.org/docs/user-guide/configuring/configuration-files#using-a-configuration-file) is what we want.

cds-amal:
My concern was the boilerplate and made a mistake referring to codesharing as inheritance. :/

I do prefer using the `js` mechanism so that I can quickly scan an eslintrcJS module to get its gist without checking for text diffs. 

However it looks like the numbers prefer JSon and I can respect that :)

eggplantzzz:
@cds-amal I don't think we need to use JS in order to extend a base config. We can use the `extends` property [here](https://eslint.org/docs/user-guide/configuring/configuration-files#using-a-configuration-file) like I linked Harry to in a previous comment.

gnidan:
Just to chime in: let's json-based eslintrc files that extend from a root eslintrc. We should probably also update the eslintrc files that @bergarces worked on, to match this new approach.

haltman-at:
@eggplantzzz Hm, I guess we could use `extends`, but where would the base files live?  I guess we'd have to avoid calling them `.eslintrc.json`, so that they wouldn't be active in where they live?

Although I should note -- I wouldn't really call these *base* files!  Our existing base `.eslintrc.json` is a base file.  These are more like reference files.

Also, @cds-amal, @eggplantzzz, neither of you has addressed what should be done about the `.eslintrc.json` files that Bernardo added in #4009.  Should those converted to use this mechanism too, as long as we're doing this?

haltman-at:
OK after talking to @gnidan I think I'm going to use `.eslint.json` and `.eslint.test.json` and `.eslint.truffle.json`!

cds-amal:
> Also, @cds-amal, @eggplantzzz, neither of you has addressed what should be done about the .eslintrc.json files that Bernardo added in #4009. Should those converted to use this mechanism too, as long as we're doing this?

I vote for consistency. It would be nice to get it all done together, if that is what you're proposing.

haltman-at:
OK, I've redone this as requested, and moved the package eslintrc jsons over to this system too.  (I ended up going with `.eslintrc.package.json` rather than `.eslintrc.test.json` because that described where they go.)

Note I left the one in the `truffle` package alone, due to the fact that it's slightly different from the others.

Also, we now definitely have to figure out how to exclude the `.eslintrc.json` from Truffle Init, seeing as how the new one is actually meaningless!

Possibilities:
1. Tell webpack to exclude it, like Tyler suggested
2. Move it one level up, and make it custom (rather than having it rely on this new system)
3. Tell Truffle Init not to copy this file when running

It seems like people have been leaning towards (1)... do people think that's the right answer?

eggplantzzz:
What do other people think? I still think just omitting it in the webpack config is probably good. What do you mean by moving it one level up @haltman-at? Does that mean just moving it out of the source files and having the rules kind of "trickle down" into the project?

haltman-at:
No, simply moving it one level up with no alterations wouldn't work.  Note how it's specified what files it applies to; the rules cannot "trickle down".  I mean move the file one level up, and then *alter it appropriately*.  (As in, make it so that applying to, e.g., `./test/etc`, it would apply to `./initSources/test/etc`.  So it would become a special case.

gnidan:
Sure we can keep the eslintrc for `truffle init`

haltman-at:
I don't understand. It has to be there, right? So the question is what mechanism to use filter it out of people's Truffle projects (filtering it out in webpack, in Truffle init, moving it elsewhere...).

gnidan:
Oh sorry @haltman-at I misunderstood. I'm not sure it much matters how we filter it out. Whatever works.

haltman-at:
OK, I'll figure something out then.  I'll avoid solution (2) though because I don't want to make more special cases.

haltman-at:
OK I ended up going with (3) so that it works properly even when not bundled. :)

gnidan (approved):
Let's ship it and see if anything needs tweaking after we use the config for awhile.

Thanks for this @haltman-at!

PR #4184, opened 2021-07-14
Add BASEFEE opcode to code-utils

haltman-at:
Since London is already out on testnets.  This probably won't do very much at the moment, as I doubt many people are yet using this opcode (it's not in Yul yet), but, still, here it is.


cds-amal (approved):
Looks good.

Issue #4161, opened 2021-07-08
Direct imports from project directory do not have the `project:/` prefix


cameel:
## Issue
When using [direct imports](https://docs.soliditylang.org/en/latest/path-resolution.html#direct-imports) (i.e. ones not starting with `./` or `../`) that reference files in project directory Truffle finds the file and includes it in JSON input for the compiler but without the `project:/` prefix (#4137). This results in the file being present in the input twice, under different paths which may lead to errors that seem inexplicable to users.

Since dropping the `project:/` prefix is not an option (I know that Truffle needs some prefix for internal reasons), the only solution I can suggest is to disallow such imports.

## Steps to Reproduce
Here's an example situation where this causes a real issue:
```bash
mkdir truffle-test/
cd truffle-test/
npm install truffle
npx truffle init
echo 'module.exports.compilers.solc.version = "0.8.6";' >> truffle-config.js

echo 'interface I {}' > contracts/I.sol
echo 'import "contracts/I.sol"; contract C is I {}' > contracts/C.sol
echo 'import "./C.sol"; import "./I.sol"; contract D is I, C {}' > contracts/D.sol
npx truffle compile
```

## Actual Results
```solidity
DeclarationError: Identifier already declared.
 --> project:/contracts/D.sol:1:19:
  |
1 | import "./C.sol"; import "./I.sol"; contract D is I, C {}
  |                   ^^^^^^^^^^^^^^^^^
Note: The previous declaration is here:
 --> contracts/I.sol:1:1:
  |
1 | interface I {}
  | ^^^^^^^^^^^^^^

Compilation failed. See above.
```

The `sources` dict in the JSON input contains these entries:
```json
"project:/contracts/C.sol": { "content": "import \"contracts/I.sol\"; contract C is I {}\n" },
"project:/contracts/D.sol": { "content": "import \"./C.sol\"; import \"./I.sol\"; contract D is I, C {}\n" },
"project:/contracts/I.sol": { "content": "interface I {}\n" },
[...]

PR #4190, opened 2021-07-17
Bug fix: Disallow non-absolute non-relative FS imports

haltman-at:
This PR addresses #4161, sort of, by just disallowing such imports.

So, the problem in #4161 regards what happens when you use FS-resolved direct imports that are neither absolute nor explicitly relative.  The problem is, there's not really any reasonable way to make these work.

Truffle tries to give each source a canonical path -- not necessarily a file path, but a path it's going to pass with it to solc.  Normally when making file-system imports, you use a relative path.  This works with Truffle and with Solidity; Truffle can find the source you're looking for to pass to solc, and Solidity can resolve the relative path among the sources you pass to it.  This also works with non-FS imports for the same reasons.

But what if you use a direct import?  In this case, Solidity performs no resolution of relative paths.

Now, if the direct import is absolute, or isn't FS-resolved at all, then this isn't a problem, because the imported path, being given in full, will match what we pass to solc.  (Although, absolute file imports might have problems, if, say you're on Windows; but you're not really supposed to use those in the first place although I haven't explicitly disallowed them here.)  But what if it's neither absolute, nor explicitly relative?

Then we get a mismatch: The path where Solidity looks for the import will be a shortened path, whereas the path we pass to solc for it will be the full path.  It won't find it!  Alternatively, we may end up passing it under *both* paths, which causes other problems.

There's not really any reasonable way for us to resolve this (unless we're willing to drop the `project:/` prefix, and have just e.g. `contracts/C.sol` instead of `project:/contracts/C.sol`).  Because regardless of what we do, solc **will** look at the short path, not the path we want.  So either we have to make the short path the canonical path (and drop `project:/` as just mentioned), or we have to disallow such things.  (Or we have to go to substantially more effort and try actually rewriting import statements or something, which I don't think we're about to do.)

Note that the existing behavior for such imports was for them to be resolved **relative to the directory the command was running in**, so they wouldn't have worked consistently anyway.  So, hopefully, this change isn't breaking anything anyone was relying on, because it shouldn't be breaking anything that consistently worked?

(Now we'll see if the tests pass. :P )

### DIFF
--- a/packages/compile-solidity/test/test_nonrelative.js
+++ b/packages/compile-solidity/test/test_nonrelative.js
@@ -1,0 +1,61 @@
+const debug = require("debug")("compile:test:test_nonrelative");
+const path = require("path");
+const { Compile } = require("@truffle/compile-solidity");
+const assert = require("assert");
+const Resolver = require("@truffle/resolver");
+const process = require("process");
+
+describe("Non-relative non-absolute file paths", function () {
+  this.timeout(5000); // solc
+
+  const options = {
+    working_directory: __dirname,
+    contracts_directory: path.join(__dirname, "./sources/badSources"),
+    contracts_build_directory: path.join(__dirname, "./does/not/matter"), //nothing is actually written, but resolver demands it
+    compilers: {
+      solc: {
+        version: "0.8.6",
+        settings: {
+          optimizer: {
+            enabled: false,
+            runs: 200
+          }
+        }
+      }
+    },
+    quiet: true
+  };
+  options.resolver = new Resolver(options);
+
+  before("Set working directory", function () {
+    originalWorkingDirectory = process.cwd();
+    process.chdir(options.contracts_directory);
+  });
+
+  it("Refuses to compile non-relative non-absolute paths", async function () {
+    this.timeout(150000);
+    const paths = [
+      "Nonrelative.sol",
+    ].map(filePath => path.join(options.contracts_directory, filePath));
+
+    debug("current dir: %s", process.cwd());
+
+    try {
+      await Compile.sourcesWithDependencies({
+        paths,
+        options
+      });
+      assert.fail("Compilation should have failed");
+    } catch (error) {
+      debug("error: %O", error);
+      if(!error.message.includes('Source "Imported.sol" not found')) {
+        throw error; //rethrow errors that aren't the one we expect
+      }
+      //otherwise, we're good
+    }
+  });
+
+  after("Reset working directory", function () {
+    process.chdir(originalWorkingDirectory);
+  });
+});
--- a/packages/resolver/lib/sources/fs.ts
+++ b/packages/resolver/lib/sources/fs.ts
@@ -57,29 +57,31 @@ export class FS implements ResolverSource {
     return path.basename(sourcePath, ".sol");
   }
 
-  async resolve(importPath: string, importedFrom: string) {
-    importedFrom = importedFrom || "";
-    const possiblePaths = [
-      importPath,
-      path.join(path.dirname(importedFrom), importPath)
-    ];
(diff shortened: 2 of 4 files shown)


cds-amal:
> Note that the existing behavior for such imports was for them to be resolved *relative to the directory the command was running in* 
wow

packages/resolver/lib/sources/fs.ts
```diff
@@ -57,29 +57,31 @@ export class FS implements ResolverSource {
     return path.basename(sourcePath, ".sol");
   }
 
-  async resolve(importPath: string, importedFrom: string) {
-    importedFrom = importedFrom || "";
-    const possiblePaths = [
-      importPath,
-      path.join(path.dirname(importedFrom), importPath)
-    ];
+  async resolve(importPath: string, importedFrom?: string) {
```
cds-amal, line 60: Is `importedFrom` still necessary?

cds-amal (changes requested):
Wow @haltman-at. The old behavior is quite astonishing! I think we should have tests for this new behavior.

packages/resolver/lib/sources/fs.ts
haltman-at: Well, remember, resolver sources have to maintain a uniform interface.  You're right that `importedFrom` is ignored here though... probably it should get marked with a `_` (husky didn't remind me of this since its not working!).  It's possible you could drop it without causing compilation problems?  I didn't try it.

haltman-at:
So you're saying there should be a test of compilation failure upon doing this?  Well, I can try to find a place to add that...

haltman-at:
OK I added the test, and also manually confirmed that the test fails without the fix.

packages/compile-solidity/test/test_nonrelative.js
```diff
@@ -0,0 +1,61 @@
...
+        paths,
+        options
+      });
+      assert.fail("Compilation should have failed");
+    } catch (error) {
+      debug("error: %O", error);
+      if(!error.message.includes('Source "Imported.sol" not found')) {
+        throw error; //rethrow errors that aren't the one we expect
+      }
+      //otherwise, we're good
```
cds-amal, line 54: Not a nit but a question. Did you use this method for better error output instead of what's available with [assert.rejects](https://nodejs.org/api/assert.html#assert_assert_rejects_asyncfn_error_message)?

packages/resolver/lib/sources/fs.ts
cds-amal: Good point about the uniform interface. In that case the `_` param prefix  would be appropriate. However, lets defer lint issues until husky is working.

cds-amal (approved):
Looks good @haltman-at!

packages/compile-solidity/test/test_nonrelative.js
eggplantzzz: I suppose this is better anyways because you can assert _why_ it failed to make sure it rejects the way you expect it instead of for some other unexpected reason.

cds-amal: That's true @eggplantzzz , that `why` text would certainly be more helpful that `it failed` among a sea of CI output.

haltman-at: Honestly, I didn't know about `assert.rejects`.  But I prefer this way anyway for the reason Tyler said.

packages/resolver/lib/sources/fs.ts
haltman-at: Eh, I'd rather change this now rather than make some future person deal with it.  I'll go change this now and then merge once tests pass again.

PR #4194, opened 2021-07-20
Standardize option aliases

eggplantzzz:
This PR continues work from https://github.com/trufflesuite/truffle/pull/4179 and standardizes the format in which options are listed in the command files' help.

### DIFF
--- a/packages/core/lib/commands/debug.js
+++ b/packages/core/lib/commands/debug.js
@@ -30,8 +30,8 @@ const command = {
   },
   help: {
     usage:
-      "truffle debug [<transaction_hash>] [(--fetch-external|-x)]" + OS.EOL +
-      "                             [(--compile-tests|--compile-all|--compile-none)]",
+      "truffle debug [<transaction_hash>] [--fetch-external|-x]" + OS.EOL +
+      "                             [--compile-tests|--compile-all|--compile-none]",
     options: [
       {
         option: "<transaction_hash>",
@@ -39,7 +39,7 @@ const command = {
           "Transaction ID to use for debugging.  Mandatory if --fetch-external is passed."
       },
       {
-        option: "(--fetch-external|-x)",
+        option: "--fetch-external|-x",
         description:
           "Allows debugging of external contracts with verified sources."
       },
--- a/packages/core/lib/commands/debug.js
+++ b/packages/core/lib/commands/debug.js
@@ -30,8 +30,8 @@ const command = {
   },
   help: {
     usage:
-      "truffle debug [<transaction_hash>] [--fetch-external]" + OS.EOL +
-      "                             [--compile-tests|--compile-all|--compile-none]",
+      "truffle debug [<transaction_hash>] [(--fetch-external|-x)]" + OS.EOL +
+      "                             [(--compile-tests|--compile-all|--compile-none)]",
     options: [
       {
         option: "<transaction_hash>",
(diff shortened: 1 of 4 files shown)


packages/core/lib/commands/debug.js
```diff
@@ -30,16 +30,16 @@ const command = {
   },
   help: {
     usage:
-      "truffle debug [<transaction_hash>] [--fetch-external]" + OS.EOL +
-      "                             [--compile-tests|--compile-all|--compile-none]",
+      "truffle debug [<transaction_hash>] [(--fetch-external|-x)]" + OS.EOL +
+      "                             [(--compile-tests|--compile-all|--compile-none)]",
```
cds-amal, line 34: This seems a bit congested. The `--fetch-external`, `|` and the `-x` are squashed together. The `l|` is particularly distracting. Adding a space looks better? I think @gnidan has stronger typography opinions than me :)

```
  Usage:        truffle debug [<transaction_hash>] [(--fetch-external | -x)]
                             [(--compile-tests | --compile-all | --compile-none)] [--network <name>] [--config <file>]

```

eggplantzzz: I am ok with doing either one. Let's just make sure everyone is happy with whatever standard we pick. Here are two screenshots for comparison.
Screen Shot 2021-07-22 at 5 01 34 PM
Screen Shot 2021-07-22 at 5 02 36 PM

haltman-at: Do we really need parentheses inside brackets?  The brackets already act as a grouper; there's not a need for a redundant second grouper inside that.

packages/core/lib/commands/debug.js
```diff
@@ -30,18 +30,18 @@ const command = {
...
     options: [
       {
         option: "<transaction_hash>",
         description:
           "Transaction ID to use for debugging.  Mandatory if --fetch-external is passed."
       },
       {
-        option: "--fetch-external",
+        option: "(--fetch-external|-x)",
```
haltman-at, line 42: Why are there parentheses here?  These are top-level parentheses, it's not grouping these within anything!

eggplantzzz: Ok so maybe we don't need the parens here, I'm ok with that. I'm just trying to establish a standard. I think above that they are meant to group the option and its alias to say "hey, only one of the things inside the parens". You vote them gone completely? Also, spaces between options and the | or not? See the above screenshot.

gnidan: seems like no spaces is more common

haltman-at: > Ok so maybe we don't need the parens here, I'm ok with that. I'm just trying to establish a standard. I think above that they are meant to group the option and its alias to say "hey, only one of the things inside the parens". You vote them gone completely? Also, spaces between options and the | or not? See the above screenshot.

No grouping is needed because there's no precedence inversion or ambiguity.  You only group when it would be incorrect otherwise.

packages/core/lib/commands/debug.js
cds-amal: The consensus is no space around the pipes? I still think `--fetch-external|-x` is a bit wonky.

eggplantzzz: @cds-amal We are using no parens and no space around the pipe.

gnidan (approved):

Issue #4193, opened 2021-07-19
Duplicate import names incorrectly detected on windows 10


cds-amal:
When compiling unboxed metacoin project on windows it complains that `Duplicate Contract names found for ConvertLib` 

Seems that there is a path.sep issue here. When `import './ConvertLib.sol';` is processed it resolves to   `'C:\\Users\\amal-admin\\dev\\mc\\contracts\\ConvertLib.sol'` which uses a different path.sep from other compilationTargets. See trace below for context

```
  compile compilationTargets: [
  compile   'C:/Users/amal-admin/dev/mc/contracts/ConvertLib.sol',
  compile   'C:/Users/amal-admin/dev/mc/contracts/MetaCoin.sol',
  compile   'C:/Users/amal-admin/dev/mc/contracts/Migrations.sol',
  compile   'C:\\Users\\amal-admin\\dev\\mc\\contracts\\ConvertLib.sol'
  compile ] +1s
```

## Steps to Reproduce

On windows
1. truffle unbox metacoin mc
2. truffle compile

## Expected Behavior

There should be no duplicate arning

## Actual Results
```
C:\Users\amal-admin\dev\mc>truffle compile

Compiling your contracts...
===========================
> Compiling .\contracts\ConvertLib.sol
> Compiling .\contracts\MetaCoin.sol
> Compiling .\contracts\Migrations.sol
> Compiling .\contracts\ConvertLib.sol
> Artifacts written to C:\Users\amal-admin\dev\mc\build\contracts
> Compiled successfully using:
   - solc: 0.5.16+commit.9c3226ce.Emscripten.clang


> Duplicate contract names found for ConvertLib.
> This can cause errors and unknown behavior. Please rename one of your contracts.
```

## Edited run with DEBUG=*
```
Compiling your contracts...
===========================
  compile-vyper paths: [
  compile-vyper   'C:/Users/amal-admin/dev/mc/contracts/ConvertLib.sol',
  compile-vyper   'C:/Users/amal-admin/dev/mc/contracts/MetaCoin.sol',
  compile-vyper   'C:/Users/amal-admin/dev/mc/contracts/Migrations.sol'
  compile-vyper ] +0ms
  compile-vyper vyperFilesStrict: [] +6ms
  compile paths: [
  compile   'C:/Users/amal-admin/dev/mc/contracts/ConvertLib.sol',
  compile   'C:/Users/amal-admin/dev/mc/contracts/MetaCoin.sol',
[...]

PR #4195, opened 2021-07-20
Bug fix: Use backslashes on Windows while resolving imports

haltman-at:
This PR is intended to address #4193.  **This needs to be tested on Windows!**  This is a quick PR that I think should resolve the problem, but I am working without actual testing here.

As best I can tell -- remember, I'm working without testing here -- the problem is that resolution of relative paths is causing aliasing of sources.  Because some paths are being given with slashes, but import resolution leads to the use of backslashes instead.

~Except, hm, we might need to apply this imports too, not just at the beginning?  Uh, I guess I need to go make more changes to this...~ OK, made that change.

### DIFF
--- a/packages/compile-common/src/profiler/getImports.ts
+++ b/packages/compile-common/src/profiler/getImports.ts
@@ -1,4 +1,5 @@
 import debugModule from "debug";
+import path from "path";
 const debug = debugModule("compile-common:profiler:getImports");
 
 //HACK: do *not* import ResolvedSource from @truffle/resolver because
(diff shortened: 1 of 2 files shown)


cds-amal:
Gosh, we really need a windows test runner. I can't think of a way to test this without actually testing on windows :/

cds-amal (approved):
LGTM! Nice work @haltman-at !

haltman-at:
OK, you've tested it on Windows?

haltman-at:
@cds-amal confims that he has so I'm gonna merge this!

PR #4196, opened 2021-07-21
Fix Truffle build on Windows.

cds-amal:
Windows cannot reference paths with `../..`.   Reference the command directly as `typedoc` is a preserve dependency and should be available for a script.


haltman-at:
I take it you've tested this and this actually fixes the Windows build?

cds-amal:
Yep, needed to do this to test your other PR

haltman-at (approved):
Cool, let's do this then! :)

Issue #3981, opened 2021-04-19
Errors and failure when running --dry-run with truffle db enabled


haydenyoung:
## Issue

When I issue the command truffle deploy --dry-run, the migration.sol contract deploys without issue but any subsequent contract fails with various error messages:

`TypeError: Cannot read property 'contract' of undefined`

`GraphQLError: Syntax Error: Invalid number, expected digit but got: "f". `

When skipping dry run, the contracts are deployed successfully.

## Steps to Reproduce

Create a new project:

```
mkdir MetaCoin
cd MetaCoin
truffle unbox MetaCoin
truffle init
```

Configure Kovan testnet and enable Truffle db in truffle-config.js.

Enable testnet:

```
kovan: {
    provider: new HDWalletProvider(
        PRIVATE_KEY,
        KOVAN_URL
    ),
    network_id: "42",
    confirmations: 4,
    timeoutBlocks: 10,
    gasPrice: 10000000000,
    networkCheckTimeout: 10000
}
```

Enable truffle db:

```     
db: {
    enabled: true
}
```

Run:

`truffle deploy --network kovan --reset --dry-run`

The above will fail after the migrations deployment (it is also noticeably slow).

## Expected Behavior

Dry run reports gas costs and completes without error.

## Actual Results

Possible errors:

```
Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.



Migrations dry-run (simulation)
===============================
> Network name:    'kovan-fork'
> Network id:      42
> Block gas limit: 12500000 (0xbebc20)


1_initial_migration.js
======================

   Deploying 'Migrations'
   ----------------------
   > block number:        24325108
   > block timestamp:     1618811305
   > account:             0x119eC8C4200dFafB4E40e73f976b8C874BFf9752
   > balance:             10.20014689606288811
   > gas used:            149175 (0x246b7)
   > gas price:           10 gwei
   > value sent:          0 ETH
   > total cost:          0.00149175 ETH

   -------------------------------------
   > Total cost:          0.00149175 ETH

TypeError: Cannot read property 'contract' of undefined
    at extract (/home/User/.nvm/versions/node/v12.19.0/lib/node_modules/truffle/node_modules/@truffle/db/src/project/loadMigrate/contracts.ts:42:40)
    at Object.process (/home/User/.nvm/versions/node/v12.19.0/lib/node_modules/truffle/node_modules/@truffle/db/src/meta/batch.ts:68:31)
[...]

PR #4198, opened 2021-07-22
Bug fix: Sanitize network name for graphql query in db

eggplantzzz:
Inspired by https://github.com/trufflesuite/truffle/issues/3981

This PR fixes a bug caused by network names with characters that, when inserted into a GraphQL query, would break the query. For example, the network name "my-network" would break the query due to the "-". The PR sanitizes the name with a regex.


gnidan (approved):

eggplantzzz:
Thanks for the help on this @gnidan!

PR #4199, opened 2021-07-22
Add an PR template

cds-amal:
Please don't merge. 

We should probably discuss as a team if this would be useful


cds-amal:
Closing this in favor of alternative solution using labels suggested by @gnidan

Issue #400, opened 2017-05-10
On windows "truffle compile" fails on "import" statements


popra:
## Issue

With truffle 3.2.2 on windows, "truffle compile" fails on "import" statements, with error:

sol:1:1: : Source "XXX.sol" not found: File import callback not supported
import "./XXX.sol";
^----------------^

It works fine with truffle 3.2.1.

## Steps to Reproduce

Create any contract that imports another .sol file.
truffle compile

## Expected Behavior

contracts should compile

## Actual Results

Error:
sol:1:1: : Source "XXX.sol" not found: File import callback not supported
import "./XXX.sol";

## Environment

* Operating System: Windows 10
* Truffle version: 3.2.2
* node version: 6.10.3
* npm version: 3.10.10

hister:
I have the same issue just create a new project and on compile i get this:
` Source "ConvertLib.sol" not found: File import callback not supported`

KaTXi:
Same here, I was about to report the same error. Happens with old projects that were working on 3.2.1 and brand new ones created with "truffle init"

tcoulter:
Hi all, thanks for the reports. This was a bug specifically for Windows - my apologies. I just pushed a fix to one of Truffle's dependencies. To get the fix, simply uninstall then reinstall Truffle. It's important you _do_ uninstall Truffle first to ensure the latest dependencies are downloaded. Cheers!

popra:
Thank you! Works nicely.

hister:
@popra The same error appears when I do `truffle test` but no problem when running `truffle compile`

kowalski:
I see the same problem on Linux. I'm using docker for development.
Host OS: Ubuntu 14.04
Docker version: Docker version 1.11.2, build b9f10c9

Steps to reproduce.
1. Run machine:
```
docker run -it node:7.7.2 bash
```

2. Update stuff inside 
```
apt-get -y update &&\
    apt-get -y upgrade &&\
    apt-get install build-essential -y
```

3. Install truffle
```
npm install -g truffle@3.2.2
```

4. Make dummy project
```
root@27d5ee4ec94d:/# mkdir -p /tmp/truffle
root@27d5ee4ec94d:/# cd /tmp/truffle
root@27d5ee4ec94d:/tmp/truffle# truffle init
Downloading project...
Project initialized.

  Documentation: http://truffleframework.com/docs

Commands:

  Compile: truffle compile
  Migrate: truffle migrate
  Test:    truffle test

```

[...]

PR #4202, opened 2021-07-26
Fix truffle resolve on windows

cds-amal:
This PR fixes truffle solidity imports on Windows  (`import truffle/Assert.sol`). 

See: #400 , and [windows CI test results](https://github.com/trufflesuite/truffle/runs/3157919036#step:6:283). This branch is working towards getting a windows CI workflow.
 
Note: couldn't figure out a way to test resolving `DeployedAddress`

### DIFF
--- a/packages/resolver/lib/sources/truffle/index.ts
+++ b/packages/resolver/lib/sources/truffle/index.ts
@@ -13,7 +13,7 @@ export class Truffle implements ResolverSource {
   }
 
   async resolve(importPath: string) {
-    if (importPath === "truffle/DeployedAddresses.sol") {
+    if (importPath === `truffle${path.sep}DeployedAddresses.sol`) {
       const sourceFiles = await findContracts(this.options.contracts_directory);
 
       const buildDirFiles: string[] =
@@ -88,7 +88,7 @@ export class Truffle implements ResolverSource {
     ];
 
     for (const lib of truffleLibraries) {
-      if (importPath === `truffle/${lib}.sol`) {
+      if (importPath === `truffle${path.sep}${lib}.sol`) {
         const actualImportPath =
           // @ts-ignore
           typeof BUNDLE_VERSION !== "undefined"
--- a/packages/resolver/test/truffle.js
+++ b/packages/resolver/test/truffle.js
@@ -1,0 +1,43 @@
+const assert = require("assert");
+const path = require("path");
+
+// Use transpiled output to properly test the path manipulation
+// truffle resolver utilizes
+const { Truffle } = require("../dist/lib/sources/truffle");
+const resolver = new Truffle({});
+
+describe("truffle resolve [ @win ]", function () {
+
+  describe("assertion contracts", () => {
+    [
+      "Assert",
+      "AssertAddress",
+      "AssertAddressArray",
+      "AssertBalance",
+      "AssertBool",
+      "AssertBytes32",
+      "AssertBytes32Array",
+      "AssertGeneral",
+      "AssertInt",
+      "AssertIntArray",
+      "AssertString",
+      "AssertUint",
+      "AssertUintArray",
+      "SafeSend",
+    ].forEach(lib => {
+      it(`resolves truffle/${lib}.sol`, async () => {
+        const dependency = path.join("truffle", `${lib}.sol`);
+        let result = await resolver.resolve(dependency);
+        assert(
+          result.filePath.includes(dependency),
+          `should have resovled 'truffle${path.sep}${lib}.sol'`
+        );
+      })
+    })
+  });
+
+  describe.skip("DeployedAddresses Contracts", () => {
+    // "DeployedAddresses"
+    // TODO: Not sure how to set this up
+  })
+});
--- a/packages/resolver/test/truffle.js
+++ b/packages/resolver/test/truffle.js
@@ -6,7 +6,7 @@ const path = require("path");
 const { Truffle } = require("../dist/lib/sources/truffle");
 const resolver = new Truffle({});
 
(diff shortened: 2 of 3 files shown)


gnidan:
> Note: couldn't figure out a way to test resolving DeployedAddress

There's an example that shows this in [our Solidity testing docs](https://www.trufflesuite.com/docs/truffle/testing/writing-tests-in-solidity)

gnidan:
But I don't think we want this, right? Are you supposed to use `\` for imports in Solidity on Windows?

cds-amal:
> But I don't think we want this, right? Are you supposed to use `\` for imports in Solidity on Windows?
> 
> **Update**: Looks like Solidity says to use forward slashes in your code, exclusively:
> 

This PR is more related to consistent use of path.sep to make sure the mapping from virtual filesystem to local filesystem is consistent. Solidity imports still use '/', and files on disk are addressed with the OS specific `path.sep`

packages/resolver/lib/sources/truffle/index.ts
```diff
@@ -13,7 +13,7 @@ export class Truffle implements ResolverSource {
   }
 
   async resolve(importPath: string) {
-    if (importPath === "truffle/DeployedAddresses.sol") {
+    if (importPath === `truffle${path.sep}DeployedAddresses.sol`) {
```
gnidan, line 16: This variable comes to us with a backslash?

cds-amal: yes. https://github.com/trufflesuite/truffle/commit/bba3e555f710deb11c249e1dc967480f614ac0d7

gnidan: oh interesting! well, there ya go. good catching this!

eggplantzzz:
There was conversation about adding tests for the `DeployedAddresses` contract but we decided that the fix here is way more valuable and should be released sooner. The added tests can be follow-up work.

packages/resolver/lib/sources/truffle/index.ts
eggplantzzz: So Truffle now takes all these paths and converts them to `path.sep`? Will this happen when, for instance, someone does a `truffle unbox metacoin` and gets the imports included in the tests which contain "/" in all paths? That's interesting, so going forward Windows users will be able to use whichever one they want?

haltman-at: Since you two apparently missed that whole business -- yup, exactly, Windows users can use either.  What we now do is:

1. When resolving imports, we convert all slashes to `path.sep`;
2. When passing things to the Solidity compiler, we convert all `path.sep` to slashes.

This code is part of (1), so everything should be `path.sep`.

packages/resolver/test/truffle.js
```diff
@@ -0,0 +1,43 @@
+const assert = require("assert");
+const path = require("path");
+
+// Use transpiled output to properly test the path manipulation
+// truffle resolver utilizes
+const { Truffle } = require("../dist/lib/sources/truffle");
+const resolver = new Truffle({});
+
+describe("truffle resolve [ @win ]", function () {
```
haltman-at, line 9: This `"@win"` tag doesn't do anything at the moment, right?  It's for later?

packages/resolver/test/truffle.js
```diff
@@ -0,0 +1,43 @@
...
+          result.filePath.includes(dependency),
+          `should have resovled 'truffle${path.sep}${lib}.sol'`
+        );
+      })
+    })
+  });
+
+  describe.skip("DeployedAddresses Contracts", () => {
+    // "DeployedAddresses"
+    // TODO: Not sure how to set this up
```
haltman-at, line 41: My understanding (based on what @gnidan has said) is that we generally avoid including `TODO` code here... if you're not going to write the code, probably just leave out the whole `describe`.  (Although yes including a note about how it'd be good to add this test would likely be a good thing.)

packages/resolver/test/truffle.js
haltman-at, line 9: This `"@win"` tag doesn't do anything at the moment, right?  It's for later?

haltman-at (commented):
Thanks for catching this Amal!  Wanted to leave some comments on the tests.

packages/resolver/test/truffle.js
cds-amal: Thanks @haltman-at , I'll remove that and we can get to the test in follow up work

packages/resolver/test/truffle.js
cds-amal: Oh, good catch. This PR grabbed commits from the `windows/runner` branch and I missed that. The tag allows for filtering tests for specific runs in that branch.

packages/resolver/test/truffle.js
haltman-at: OK, again I'd suggest adding back in a comment -- something to the effect of, there should be a `DeployedAddresses` test here but that's more difficult -- but this seems good enough to approve at least. :)

haltman-at (approved):
Thanks for catching this Amal, seems good to me!  (Also I'd still suggest adding in a comment about `DeployedAddresses` if you have time.)

packages/resolver/lib/sources/truffle/index.ts
eggplantzzz: I had understood that it would be for relative paths and didn't realize they would get converted for Truffle libraries as well. That's why I asked.

cds-amal:
Thanks @haltman-at. I added an [issue to more easily track it](https://github.com/trufflesuite/truffle/issues/4204).

Issue #4192, opened 2021-07-19
db processes when not enabled


cds-amal:
When compiling a project with db disabled, it should not be in the execution trace. This was observed on both windows and linux OS.

## Versions

$ truffle version
Truffle v5.4.0 (core: 5.4.0)
Solidity v0.5.16 (solc-js)
Node v12.22.1
Web3.js v1.4.0

## truffle-config.js
```JavaScript
module.exports = {
  // Uncommenting the defaults below
  // provides for an easier quick-start with Ganache.
  // You can also follow this format for other networks;
  // see <http://truffleframework.com/docs/advanced/configuration>
  // for more details on how to specify configuration options!
  //
  //networks: {
  //  development: {
  //    host: "127.0.0.1",
  //    port: 7545,
  //    network_id: "*"
  //  },
  //  test: {
  //    host: "127.0.0.1",
  //    port: 7545,
  //    network_id: "*"
  //  }
  //}
  //
};
```

## Reproduction steps

```console
reprod meta-coin
truffle unbox metacoin mc
cd mc
DEBUG=db* truffle compile
```

## Expected 

There should be no `db*` logs

## Actual results

```console
─$ DEBUG=db* truffle compile
  db:meta:graph:schema Creating DefinitonSchema for sources... +0ms
  db:meta:graph:schema Created ImmutableDefinitionSchema for sources. +1ms
  db:meta:graph:schema Creating DefinitonSchema for bytecodes... +0ms
  db:meta:graph:schema Created ImmutableDefinitionSchema for bytecodes. +0ms
  db:meta:graph:schema Creating DefinitonSchema for compilations... +0ms
  db:meta:graph:schema Created ImmutableDefinitionSchema for compilations. +0ms
  db:meta:graph:schema Creating DefinitonSchema for contracts... +0ms
  db:meta:graph:schema Created ImmutableDefinitionSchema for contracts. +0ms
  db:meta:graph:schema Creating DefinitonSchema for contractInstances... +0ms
  db:meta:graph:schema Created ImmutableDefinitionSchema for contractInstances. +0ms
  db:meta:graph:schema Creating DefinitonSchema for networks... +0ms
  db:meta:graph:schema Created ImmutableDefinitionSchema for networks. +0ms
  db:meta:graph:schema Creating DefinitonSchema for nameRecords... +1ms
  db:meta:graph:schema Created ImmutableDefinitionSchema for nameRecords. +0ms
[...]

Issue #4141, opened 2021-06-24
Db.connect is not a function error when running truffle migrate without running truffle compile first


fainashalts:
## Issue

It appears that when DB is enabled and one runs `truffle migrate` without running `truffle compile` first, you get this error. 

## Steps to Reproduce

1) create a truffle project, and enable db in the config
2) run `truffle migrate` with the network of your choice
3) behold!

## Expected Behavior

Since `truffle migrate` runs a compile if you haven't compiled, I would expect that this would not occur. 

## Actual Results

`Db.connect is not a function` error. 

<img width="1071" alt="Screen Shot 2021-06-24 at 3 49 22 PM" src="https://user-images.githubusercontent.com/8952139/123342005-c912c200-d503-11eb-8c82-9398694c7b6b.png">

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.3.11
* node version (`node --version`): 12.14.0
* npm version (`npm --version`): 6.14.5

RealLau:
@fainashalts I got the same issue, did you resolve it?

scald:
> I tried to run `truffle compile`(no error occurs) before `truffle migrate`, still got the same error.

same here. 

Truffle v5.3.13 (core: 5.3.13)
Solidity v0.5.16 (solc-js)
Node v14.14.0
npm 6.14.8

eggplantzzz:
@fainashalts was a fix for this released? I cannot reproduce this error.

eggplantzzz:
Does anyone here have specific steps to reproduce this error?

cds-amal:
@RealLau , @scald are you still able to reproduce with the [latest version of truffle](https://github.com/trufflesuite/truffle/releases)?

scald:
@eggplantzzz @cds-amal cannot reproduce with latest version. thanks!

fainashalts:
I'm closing this issue as it appears to be resolved!

PR #4203, opened 2021-07-26
Internal improvement: Only run Db code when it is enabled

eggplantzzz:
This PR moves the require statements for @truffle/db into a scope that is run only when db is enabled. I did a few tests and this seems to save an average of 1 second during the compilation flow on my machine when db is not enabled.

There is one possible inefficiency here; in workflow-compile we now require Db 2 times as we are using it in `compileAndSave` and `assignNames`. This might slow the flow down when Db is enabled, or maybe it will be loading it from the cache the second time and there really is no issue here?

Inspired by https://github.com/trufflesuite/truffle/issues/4192


gnidan:
> This might slow the flow down when Db is enabled, or maybe it will be loading it from the cache the second time and there really is no issue here?

Should be no issue; Node caches modules it requires

eggplantzzz:
Ok so for this we decided to create a new package (db-utils probably) to put the `try/catch` stuff for loading db in and hide it away from our sensitive eyes.

packages/core/lib/commands/compile.js
```diff
@@ -91,7 +91,9 @@ const command = {
     }
 
     const result = await WorkflowCompile.save(config, compilationOutput);
-    await WorkflowCompile.assignNames(config, result);
+    if (config.db && config.db.enabled) {
```
fainashalts, line 94: ooh! I wonder if the error from #4141 was related to this line!

fainashalts (commented):
This LGTM, but I wonder if we should throw an error in db-kit and the db commands (serve and query) to tell the user that they have to enable DB and it must be installed in order to use these? For completeness' sake. What do you think?

packages/db-utils/src/index.ts
```diff
@@ -0,0 +1,7 @@
+export const getDb = () => {
```
cds-amal, line 1: What do you think about renaming the function to `getDbOrNull`? 

```suggestion
export const getDbOrNull = () => {
```

Then the use would be clear that the function could return `null` and eliminate the need for a redundant comment (currently this function could return null)

```
let Db = getDbOrNull();
if (Db) {...}
```

packages/migrate/Migration.js
```diff
@@ -120,35 +115,39 @@ class Migration {
...
-        });
-
-        const result = await project
-          .connect({ provider: this.config.provider })
-          .loadMigrate({
-            network: {
-              name: this.config.network
-            },
-            artifacts
+        // currently if Db fails to load, getDb returns `null`
```
cds-amal, line 122: This comment would not be needed if the function name were changed to `getDbOrNull()` or something similar.

packages/core/lib/commands/compile.js
cds-amal: @eggplantzzz (sorry, couldn't comment on this without unresolved this conversation)

Does it make sense to move this guard to the `getDb()` function? Of course we'd have to change the interface to accept a `ConfigType` or null, executing the guard only if a config were passed.

cds-amal (commented):
Nice work Tyler. I left some comments/questions for a discussion

packages/db-utils/src/index.ts
eggplantzzz: I don't really like `getDbOrNull`, I think it would confuse me if I wasn't familiar. Maybe the comments I wrote are unnecessary.

eggplantzzz:
I also don't think I like the idea of moving the guard. I feel like it makes more sense to put that logic as part of the main command flow and not hidden in the module as my hunch is that it will make what is happening more obvious when reading through the code. Do you disagree?

cds-amal:
I think DB should only be loaded if enabled in config, because it is currently optional and this is an indication the guard belongs in the new module. 

This way would make it more convenient to introduce DB into a new module and revert to a regular import if it stops being an optional dependency (and no longer behind a feature flag)

eggplantzzz:
I'm not really persuaded that it should be moved inside. As of right now I don't think this utility should know more about the config object. It should merely and dumbly load the Db if it can. The instructions for running the command should determine whether db should be loaded in my view. And if it becomes required then it won't really be a hassle at all to just remove the `config.db.enabled`-type checks. Any other opinions on this?

fainashalts:
@eggplantzzz I still think we should use this util folder everywhere DB is being required, which includes `db-kit` and `lib/commands/db`. And in those places, since they are specific-to-db packages, we should throw an error if the `getDb()` call returns null. I feel like this PR isn't complete without adding this functionality.

fainashalts:
Also I think it makes sense to keep the config check in the specific modules, otherwise we have to give the new `db-utils` package access to the config which seems sub-optimal. I don't feel super strongly either way, but that's my two cents!

eggplantzzz:
@fainashalts So I talked with @gnidan and he pointed out that we don't need to use this utility in db-kit because it is a required dependency and will need to be installed to use it. Everywhere else, where it is optional, should use this package.

packages/core/lib/commands/db/commands/serve.js
```diff
@@ -17,7 +17,15 @@ const command = {
    */
   run: async function (argv) {
     const Config = require("@truffle/config");
-    const { serve } = require("@truffle/db");
+    const { getTruffleDb } = require("@truffle/db-loader");
+    const Db = getTruffleDb();
```
cds-amal, line 21: This shouldn't block the PR, but we should probably guard `db serve` behind the `db: enabled` config/feature flag

cds-amal (approved):
Nice! Tests pass and looks good to me!

fainashalts (commented):
LGTM! Thanks for addressing my previous comments! :)

PR #4206, opened 2021-07-27
Fix: Update more windows anomalies

cds-amal:
This PR fixes some more issues related to the path.sep issues

1. use EOL for comparison to avoid \r \r\n between OS
1. use `path.sep` in globalnpm resolver
1. use  X-platform algo for detecting paths (utils/unbox)
1. update tests accordingly
1. update compile common to refer to filesystem location using path.sep appropriately


packages/resolver/test/globalnpm.ts
```diff
@@ -5,6 +5,7 @@ import * as sinon from "sinon";
 import path from "path";
 import fs from "fs";
 import { describe, it } from "mocha";
+import { EOL} from "os";
```
haltman-at, line 8: Style note since Husky's not running at the moment, `{ EOL}` should be `{ EOL }` :P

packages/resolver/test/globalnpm.ts
```diff
@@ -5,6 +5,7 @@ import * as sinon from "sinon";
 import path from "path";
 import fs from "fs";
 import { describe, it } from "mocha";
```
haltman-at, line 7: Oh, huh, what are these doing here?  These don't need to be imported, they're globals... (not that you changed that, sorry, I just noticed that and it seemed off)

packages/resolver/lib/sources/globalnpm.ts
```diff
@@ -44,7 +44,7 @@ export class GlobalNPM implements ResolverSource {
       try {
         const result = fs.readFileSync(possiblePath, "utf8");
         return JSON.parse(result);
-      } catch (e) {
+      } catch (_err) {
```
haltman-at, line 47: Style note: You can actually just do `catch { ... }` if you're not using the error object.

packages/resolver/lib/sources/globalnpm.ts
```diff
@@ -14,7 +14,7 @@ export class GlobalNPM implements ResolverSource {
 
     let [packageName] = importPath.split("/", 1);
     if (detectInstalled.sync(packageName)) {
-      const regex = new RegExp(`/${packageName}$`);
+      const regex = new RegExp(`${path.sep}${packageName}$`);
```
haltman-at, line 17: Careful here!  I don't think this does what you want.

Inputs to the `RegExp` constructor take as input a regexp string.  If it contains regexp-significant characters -- which a backslash certainly is! -- the result won't be what you want.

In order to solve this, you either need to escape the string before you pass it in (lodash has a function for this, that we use in some places), or -- and I think this is the better option when possible -- not use a regex.

Fortunately, it looks like in this case we can indeed get away with not using a regex.  It looks to me like this regex is only being used as the first argument to `replace`.  You don't need to pass a regex there, you can pass a plain string and it'll be interpreted as a plain string.  So, that's the thing to do here.

(Note that if it were a *global* regex, then you'd be forced to take the escaping route; we can't use `replaceAll` because that currently only exists on Node 16.  Fortunately it isn't.)

packages/core/test/lib/commands/compile.js
```diff
@@ -57,6 +57,7 @@ describe("compile", function () {
         quiet: true
       })
     );
+    console.log("length", Object.keys(contracts).length);
```
haltman-at, line 60: Should this be a `debug`?  Did you mean to delete this before PRing?  Or is this deliberate?

packages/box/test/box.js
```diff
@@ -67,7 +69,7 @@ describe("@truffle/box Box", () => {
       });
     });
 
-    it("unboxes truffle box from local folder", async () => {
+    it("unboxes truffle box from local folder [win]", async () => {
```
haltman-at, line 72: This tag doesn't actually do anything, right?  Is that why you didn't include the `@` sign?  I'm thinking it should probably be included for consistency even though it doesn't actually do anything atm, and you can just leave a comment about how it does nothing?  Or something?

packages/box/test/box.js
```diff
@@ -4,10 +4,12 @@ const assert = require("assert");
 const inquirer = require("inquirer");
 const sinon = require("sinon");
 const Config = require("@truffle/config");
+
+// TODO: Add tests for x-platform normalizing paths
```
haltman-at, line 8: Reminder not to PR `TODO` comments. :)

packages/box/box.ts
```diff
@@ -24,9 +24,20 @@ const defaultPath = "git@github.com:trufflesuite/truffle-init-default";
...
+
+  debug('---');
+  debug({ 
+    normalizedPath, 
+    resolvedPath,
+    pathExists
+  });
+
+  if(pathExists) {
+    debug({ in: url, out: normalizedPath});
```
haltman-at, line 39: Style note again about missing space before `}`. :P

packages/box/box.ts
```diff
@@ -24,9 +24,20 @@ const defaultPath = "git@github.com:trufflesuite/truffle-init-default";
...
+  const pathExists = fse.existsSync(resolvedPath);
+
+  debug('---');
+  debug({ 
+    normalizedPath, 
+    resolvedPath,
+    pathExists
+  });
+
+  if(pathExists) {
```
haltman-at, line 38: Can you explain why you added this check?

packages/compile-common/src/profiler/updated.ts
```diff
@@ -33,28 +33,21 @@ interface SourceFilesArtifactsUpdatedTimes {
   [filePath: string]: number; // ms since epoch
 }
 
+const getKeyFromPath = (fsPath: string): string => fsPath.split(path.sep).join("/");
```
haltman-at, line 36: Oh, heh, this is a nice way of doing the replace!  I should use that elsewhere...

packages/compile-common/src/profiler/updated.ts
```diff
@@ -108,50 +101,43 @@ function findUpdatedFiles(
...
       const artifactsUpdatedTime =
-        sourceFilesArtifactsUpdatedTimes[sourceFile] || 0;
+        sourceArtifactsUpdatedTimes[sourceFile] || 0;
       const sourceFileUpdatedTime = (
-        sourceFileStat.mtime || sourceFileStat.ctime
-      ).getTime();
+        stat.mtimeMs || stat.ctimeMs
+      ); 
 
       if (sourceFileUpdatedTime > artifactsUpdatedTime) return sourceFile;
```
haltman-at, line 116: Since you're changing this from a map to a filter, I'd like to suggest changing this line to match, even if the current line technically works.  As in, I think this should read `return sourceFileUpdatedTime > artifactsUpdatedTime;`.

haltman-at (changes requested):
Thanks for taking this on Amal!  I do have to hit request changes due to one incorrect use of regular expressions, and there's one other change that I think warrants explanation, but other than that I mostly just have style comments.  It'll be good to see this merged!

packages/compile-common/src/profiler/updated.ts
eggplantzzz: By the way, did we need to change it to `mtimeMs`? I think this change was prompted by an assumption that turned out not to be true, right?

packages/core/test/lib/commands/compile.js
cds-amal: Yea, I was going to make a draft PR to rebase before marking it ready. But since it picked up review, I'll be more careful in the future. Though I'll likely do a rebase at some point to drop some unrelated commits

packages/resolver/test/globalnpm.ts
cds-amal: Updated. Thanks.

packages/resolver/lib/sources/globalnpm.ts
cds-amal: I didn't know that! thanks @haltman-at . Updated.

packages/core/test/lib/commands/compile.js
cds-amal: Removed.

packages/box/test/box.js
cds-amal: It doesn't belong here. Removed!

packages/box/box.ts
cds-amal: I miss husky. Updated!

packages/compile-common/src/profiler/updated.ts
cds-amal: :)

packages/box/test/box.js
cds-amal: Removed! thanks.

packages/box/box.ts
cds-amal: Previously this function determined if the path was a file-system path to be used by `truffle unbox` by checking if if the path starts with `.`,`~` or '/'. I changed the check to be: does this path exist on the file-system to handle windows
paths. There should be more unit tests for this code path.

packages/compile-common/src/profiler/updated.ts
cds-amal: Oh gosh! For sure @haltman-at !

cds-amal: @eggplantzzz I changed the comparison to use `fs.Stat` `xtimeMs` variant because they're already numbers (number of milliseconds since the POSIX Epoch) instead of calling `mtime.getTime()` (mtime is a Date object)

packages/resolver/lib/sources/globalnpm.ts
cds-amal: Good catch! I missed the \ metachar here! The original used a regex to match the suffix and a non-regex could potentially match a non-suffix portion of the string. What do you think of  ```RegExp(`\\${path.sep}${packageName}$`)```?

haltman-at: Oh, I see, I missed the part that you want to match specifically a suffix.

Your version might work, but it's not obvious to me; can you check whether that sort of `RegExp(...)` does indeed work as you expect?  If it does then great!  Although you'll want to leave an explanatory comment for sure.

If it doesn't you might want to use a proper escaping package (again, lodash has that), or find a non-regex way of doing it (there are certainly ways... like use `endsWith` to see if it matches, and, if it does, slice off an appropriate number of characters and append the new thing -- obviously that's a bit more involved).

packages/box/test/box.js
haltman-at: OK, if you say so!

packages/box/box.ts
haltman-at: Hm, OK.  So it's not a problem that if the path takes the form of a file system path, but doesn't actually exist, it won't take that branch?  Because obviously that's different.

packages/resolver/lib/sources/globalnpm.ts
haltman-at: OK I just checked it myself and it seems that your approach does indeed work, regarding `path.sep`.  That said, I still think it might be a good idea to either use a proper escaping library, or to take the non-regex route, on account of the possibility of special characters in `packageName` -- unless there's some reason we can be sure that won't happen?  I think NPM packages might be allowed to include `.` in their names, which might cause problems.

cds-amal: For the `.`, it would be ok, a dot meta will match the dot target. But, I'll sleep better using `endsWith`. Thanks for getting me here @haltman-at

packages/box/box.ts
cds-amal: Yes, if it doesn't exist as a path, the next checks would be url like tests and finally a default error case.

haltman-at: OK, so under this PR it'll hit the default error case, whereas previously it would have done what exactly?

cds-amal: Ok, I'm going to move this to a draft because the unboxing piece needs more work. Previously it gave more meaningful and actionable errors (apart from network shenanigans). This PR breaks that. 

@haltman-at thanks for the push back on this. I think I have a better way forward that would keep meaningful errors. Most of the unboxing branches on isURLPathLike logic based on unix fs assumptions. I think we can more accurately determine if a url is a VCS like url (https:, git:)

chris-shyft:
excited about this patch - good job guys!

PR #4208, opened 2021-07-28
Dependency update: Update web3 to 1.5.0

haltman-at:
This updates web3 to 1.5.0, which adds London support.  Note that this PR does not add any sort of London support into Truffle; that will be done separately later.  This PR just updates web3 to a version of web3 that supports London.


cds-amal (approved):
Looks good and tests are happy!

PR #4213, opened 2021-07-29
Enhancement: Make Truffle variables available in `console.require` JS

eggplantzzz:
This PR further develops the feature allowing users to populate the console environment using `console.require`. This uses @truffle/require to inject Truffle variables (currently just `web3`, `interfaceAdapter`, and `accounts`) into the scripts provided by users setting the `console.require` property in `truffle-config.js`. This way users can use these variables for calculations etc.

One problem here is that I think this allows users to mutate these variables. Perhaps we should also consider just injecting clones of these values instead?


cds-amal (commented):
This will be very helpful @eggplantzzz! Can you add tests for this please?

cds-amal (commented):
Just a note that we should probably update the documentation for the truffle commands explaining the new capabilities.

packages/core/lib/console.js
```diff
@@ -61,13 +61,16 @@ class Console extends EventEmitter {
 
   async start() {
     try {
+      const context = {};
+      this.repl = { context };
+      await this.setUpEnvironment();
+      this.provision();
+
```
cds-amal, line 68: We should document this. This is necessary because we want the repl's initial context to be provisioned and set up appropriately because `repl.start()` will immediately start consuming `process.stdin`, which would makes testing difficult. This change ensures that repl's env is provisioned correctly in normal and test use cases.

Maybe it's worth factoring out these lines to a helper function `getInitialREPLContext()`?

packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -42,5 +43,36 @@ module.exports = {
...
       }
     });
-  }
+  },
+  runInDevelopEnvironment: (commands = [], config) => {
+    let child;
+    // open `truffle develop`
+    const cmdLine = "node " +
+      path.join(__dirname, "../../../", "core", "cli.js") +
+      " develop";
```
cds-amal, line 52: We should follow the same logic as the original commandrunner, using `process.env.NO_BUILD` to pick the correct entry point

packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -42,5 +43,36 @@ module.exports = {
...
+
+    return new Promise((resolve, reject) => {
+      child = exec(cmdLine, { cwd: config.working_directory });
+
+      if (child.error) return reject(child.error);
+
+      child.stderr.on("data", (data) => {
+        config.logger.log("ERR: ", data);
```
cds-amal, line 60: This is a good idea to include these prefixes. They'll be useful when debugging!

packages/truffle/test/scenarios/commands/develop.js
```diff
@@ -23,20 +25,38 @@ describe("develop", function () {
...
+    this.timeout(70000);
+
+    await CommandRunner.runInDevelopEnvironment(["breakfast"], config);
+    const output = logger.contents();
+
+    //prepare a helpful message to standout in CI log noise
+    const formatLines = (lines) =>
+      lines.split("\n").map((line) => `\t---truffle develop log---\t${line}`)
+        .join("\n");
+    const expectedValue = "eggs and sausage, and a side of toast";
```
cds-amal, line 36: what?! no coffee!

packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -42,5 +43,36 @@ module.exports = {
         reject(child.error);
       }
     });
-  }
+  },
+  runInDevelopEnvironment: (commands = [], config) => {
+    let child;
```
cds-amal, line 48: This declaration can move to the promise body.

packages/truffle/test/scenarios/commands/build.js
```diff
@@ -1,6 +1,6 @@
 const assert = require("assert");
-const CommandRunner = require("../commandrunner");
-const MemoryLogger = require("../memorylogger");
+const CommandRunner = require("../commandRunner");
+const MemoryLogger = require("../MemoryLogger");
```
cds-amal, line 3: I thought this renaming was a stylistic choice but notice one is camel case and the other pascal case. Which do you prefer?

cds-amal (changes requested):
Nice @eggplantzzz ! We now have a way to test both REPLs.

packages/truffle/test/scenarios/commands/build.js
eggplantzzz: I did it because `MemoryLogger` is basically a class and `CommandRunner` is more a utility. I'd be ok with renaming it `CommandRunner.js` if you prefer though.

cds-amal: Seems reasonable to me.

packages/core/lib/console.js
eggplantzzz: So what is really going on here is that we're shortening the time from when the repl starts and the context is added. This way all values are calculated ahead of time and injected directly after. I almost made a comment and decided not to, I'll add one!

cds-amal: It seems that whenever a node repl is created (`repl.start(options);` or `new repl.REPLServer(options);`) it automatically starts consuming its `stdin`. That is, there isn't a way to start a repl server without it automatically starting to consume `stdin`. My hypothesis is this the time between consuming stdin and resolving its context is not a problem when the repl is connected to a terminal (normal usage, invoking the command line) but when it starts as an exec'd child-process `stdin` is immediately consumed (before resolving the context).

The other issue here is the `context = {}; this.repl = { context }` business which looks odd but is necessary because of how repl creation works. The functions to update the contexts mutate `this.repl.context` and this is the current solution. :/ 

I'm still not happy with it, but I'd like a comment to remind the next person of the reason for the `context = {}; this.repl = { context }` business

cds-amal (changes requested):
I only have a request to document the reasoning for the repl.context oddity.

packages/core/lib/console.js
eggplantzzz: Why wouldn't they just let you input the context as an argument to `start`?

eggplantzzz: I'm going to just edit `setUpEnvironment` to return the context instead of doing the mutating - why not?

cds-amal: It really is odd! The only pattern I see in the docs is to use a fluent style to set the context immediately after invoking `repl.start()`
```
const repl = require('repl');
const msg = 'message';

repl.start('> ').context.m = msg;
```

packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -1,22 +1,15 @@
...
-      : (execString =
-          "node " +
-          path.join(__dirname, "../", "../", "build", "cli.bundled.js") +
-          " " +
-          command);
+  getExecString: function () {
+    return process.env.NO_BUILD
+      ? `node ${path.join(__dirname, "../", "../", "../", "core", "cli.js")}`
+      : `node ${path.join(__dirname, "../", "../", "build", "cli.bundled.js")}`;
+  },
```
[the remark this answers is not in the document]
cds-amal: your call :)

packages/core/lib/console.js
```diff
@@ -104,55 +102,60 @@ class Console extends EventEmitter {
...
         }
       });
     } else {
       throw new Error(errorMessage);
     }
+    return userGlobals;
   }
 
-  async setUpEnvironment() {
+  async calculateTruffleAndUserGlobals() {
```
cds-amal, line 158: Might have to update tests for this API change?

cds-amal (approved):
Nice, the console start up reads much cleaner!

cds-amal:
I just noticed the CI is failing :/ I suspect it might be the tests not keeping up with the console api change

eggplantzzz:
Yep looks like I need to update the unit tests.

cds-amal (approved):
LGTM! 🎉

PR #4214, opened 2021-07-29
Add London support (minus access lists) to Truffle Contract

haltman-at:
This PR adds support for London transactions, minus access lists, to Truffle Contract.

This PR is a draft because it needs to be tested.  Unfortunately, it can't be tested with Ganache at the moment, because Ganache doesn't support London yet.

Obviously, prefer to wait until Ganache supports London before doing this, but London is coming soon and after discussing things with @gnidan we decided this is important enough not to wait.  Access lists are less important, so for now we're still going to wait on that one.

### DIFF
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -11,6 +11,8 @@ const allowedTxParams = new Set([
   "to",
   "gas",
   "gasPrice",
+  "maxFeePerGas",
+  "maxPriorityFeePerGas",
   "value",
   "data",
   "nonce",


haltman-at:
Hm, so it seems to work, sort of... but, like, it demands that maxFeePerGas and maxPriorityFeePerGas be given as hex strings.  Uh, we can work around that if we have to, but in the meantime I'm going to go file an issue with web3...

haltman-at:
Oh, geez, it's worse than that!  London transactions don't work at all because web3 adds a default `gasPrice` field!  So web3 1.5.0 is basically just broken... well, I'll go file an issue...

haltman-at:
Filed issues:
https://github.com/ChainSafe/web3.js/issues/4210
https://github.com/ChainSafe/web3.js/issues/4211

cds-amal:
How can I test this with the latest geth @haltman-at ?

haltman-at:
> How can I test this with the latest geth @haltman-at ?

Is there a particular thing you're confused about, or are you looking for some sort of walkthrough?  In the latter case Slack might be more appropriate here than GitHub comments...

Hbutlercapone (approved):

eggplantzzz (approved):
I like it!

Issue #4165, opened 2021-07-08
Difference in behavior between absolute import paths starting with multiple slashes and ones starting with a single one


cameel:
## Issue
As far as I can tell, Truffle no longer resolves absolute paths in imports (i.e. `/a/b/c.sol`) - it simply does not add the file to compiler's JSON input, which results in a compilation error. This is perhaps not the best way to report it (an explicit error saying that absolute paths are not supported would be better) but overall dropping support for them is a good thing in the light of #4119.

The problem is that a single leading slash is not the only way to specify an absolute path. `//a/b/c.sol` or `///a/b/c.sol` still results in an absolute path in compiler's input.

I haven't checked this on Windows but `\\` or `C:\\` might be producing similar results. Note that paths starting with `//` (and `\\` on Windows) are actually [UNC paths](https://en.wikipedia.org/wiki/Path_(computing)#UNC). On Windows they are used for network shares while on Unix-like systems they often work like absolute paths, though they're still considered different - `//`, unlike `///` or more slashes is not squashed into a single `/` by shell.

## Steps to Reproduce
```bash
cd /tmp/
mkdir truffle-test/
cd truffle-test/
npm install truffle
npx truffle init
echo 'module.exports.compilers.solc.version = "0.8.6";' >> truffle-config.js

echo 'contract C {}' > contracts/C.sol
echo 'import "//tmp/truffle-test/contracts/C.sol";' > contracts/D.sol
echo 'import "///tmp/truffle-test/contracts/C.sol";' > contracts/E.sol
npx truffle compile
```
## Actual Results
```
> Artifacts written to /tmp/truffle-test/build/contracts
> Compiled successfully using:
   - solc: 0.8.6+commit.11564f7e.Emscripten.clang


> Duplicate contract names found for C.
> This can cause errors and unknown behavior. Please rename one of your contracts.

> Duplicate contract names found for C.
> This can cause errors and unknown behavior. Please rename one of your contracts.
```

The `sources` dict in the JSON input contains these entries:
```json
"project:/contracts/C.sol": { "content": "contract C {}\n" },
"project:/contracts/D.sol": { "content": "import \"//tmp/truffle-test/contracts/C.sol\";\n" },
"project:/contracts/E.sol": { "content": "import \"///tmp/truffle-test/contracts/C.sol\";\n" },
"//tmp/truffle-test/contracts/C.sol": { "content": "contract C {}\n" },
[...]

PR #4216, opened 2021-08-02
Bug fix: Disallow absolute import paths not beginning with project root (for real)

haltman-at:
Addresses #4165.

So, the actual effect of this PR isn't actually to require things to be in the project directory (that was already typically the case);
rather it's to bar getting around it by starting your import path with `"//"` or the like.  Issue #4165 was followed because you could import absolute paths if you just started them with `"//"`.  This PR ends that loophole -- all paths passed to the Solidity compiler have to be either relative, or begin with the project directory (as a string; no starting with `"//"`).  So, now such imports will fail, because their source got filtered out.

Note: @cameel raises the point that we might want to test this with Windows.  It's possible that due to the normalization, we should actually be checking for an an initial `"/"` rather than using `path.isAbsolute`.  @cds-amal would you possibly be willing to test that...?

### DIFF
--- a/packages/compile-common/src/sources.ts
+++ b/packages/compile-common/src/sources.ts
@@ -18,6 +18,10 @@ export function collectSources(
   replacement: string = "/"
 ): CollectedSources {
   const mappedResults = Object.entries(originalSources)
+    .filter(([originalSourcePath, _]) =>
+      !path.isAbsolute(originalSourcePath) ||
+      originalSourcePath.startsWith(baseDirectory)
+    )
     .map(([originalSourcePath, contents]) => ({
       originalSourcePath,
       contents,
--- a/packages/compile-solidity/test/test_nonrelative.js
+++ b/packages/compile-solidity/test/test_nonrelative.js
@@ -4,6 +4,9 @@ const { Compile } = require("@truffle/compile-solidity");
 const assert = require("assert");
 const Resolver = require("@truffle/resolver");
 const process = require("process");
+const tmp = require("tmp");
+tmp.setGracefulCleanup();
+const fs = require("fs");
 
 describe("Non-relative non-absolute file paths", function () {
   this.timeout(5000); // solc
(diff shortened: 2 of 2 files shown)


cameel:
>  It's possible that due to the normalization, we should actually be checking for an an initial `"/"` rather than using `path.isAbsolute`.

Fun fact: I'm not sure how it's in the library you're using but in Boost (C++) a path like `/a/b/c.sol` (or `\a\b\c.sol`)  is **not** considered absolute on Windows. Only `C:/a/b/c.sol` (or `C:\a\b\c.sol`) is. That's because a path without the drive letter is not sufficient to unambiguously locate a file (it's relative to the drive that contains the working dir).

That really gave me a pause when doing normalization in the compiler since we'd generally prefer the `/a/b/c.sol` form whenever possible because it at least makes it possible to have an identical path on Linux, macOS and Windows. Yet another reason to disallow absolute paths in the future.

cameel:
Also, please make sure you test paths starting with `///` too. Two slashes represent UNC paths and might be treated specially by the library.

packages/compile-common/src/sources.ts
```diff
@@ -18,6 +18,10 @@ export function collectSources(
   replacement: string = "/"
 ): CollectedSources {
   const mappedResults = Object.entries(originalSources)
+    .filter(([originalSourcePath, _]) =>
```
cds-amal, line 21: Is `_` necessary here?

cds-amal (commented):
Thanks @haltman-at. Can you add a test for this please? I'll check on Windows tomorrow.

packages/compile-common/src/sources.ts
haltman-at: Technically not, I suppose, but I feel like it would be confusing without it.  We've got a pair `[key, value]` and we're filtering on just the key, so I put named the value `_`; technically one could omit the value and just write `[key]`, but that seems more confusing to me; it makes it look like only the key is present.

cds-amal: sounds good!

haltman-at:
OK @cds-amal, I've added a test as requested.  However this still needs Windows testing I guess.

packages/compile-solidity/test/test_nonrelative.js
```diff
@@ -59,3 +62,58 @@ describe("Non-relative non-absolute file paths", function () {
...
+    options.resolver = new Resolver(options);
+    const importedPath = path.join(options.contracts_directory, "Imported.sol");
+    const importerPath = path.join(options.contracts_directory, "DoubleSlash.sol");
+    await fs.promises.copyFile(
+      path.join(__dirname, "./sources/badSources/Imported.sol"),
+      importedPath
+    );
+    await fs.promises.writeFile(
+      importerPath,
+      `import "/${importedPath}";` //note the extra slash
```
cds-amal, line 101: This is certainly not a blocker for this PR. 

I just want to mention we should have a categorization method that selects tests for the current platform. In this test, `/drive:\Users\... is not a valid Windows path`. Perhaps tags can work, or we can inject platform specific guards in the tests. Something to discuss with the team I guess.

cds-amal (approved):
Looks good to me @haltman-at . Sorry for the confusion and delay in getting this reviewed.

packages/compile-solidity/test/test_nonrelative.js
haltman-at: Yeah I think that might be better addressed on #4206?

cds-amal: Yes, of course.

PR #4221, opened 2021-08-04
Remove web3 types from hdwallet-provider

eggplantzzz:
While working on hdwallet-provider I notices that the web3 types are not necessary in this package. This PR removes them from hdwallet-provider.


haltman-at (approved):
Good catch, thanks Tyler!

PR #4223, opened 2021-08-05
Add TypeScript to @truffle/compile-solidity and convert compilerSupplier module

gnidan:
(This follows #4222)

This PR adds TypeScript to @truffle/compile-solidity and sets it up as the `prepare` step. TS is configured to `allowJs` here, so as to leave most of the code in JS, but this PR does convert the `compilerSupplier/` module into TS.

Couple notes on this:

- There's odd behavior when doing a combination of `allowJs: true, declaration: true`, where if the `outDir` is inside the `rootDir`, then TS will generate declaration files for JS sources, then subsequently fail to compile with a TS5055 error ("Cannot write ... because it will overwrite input file"). Best I can figure, TS attempts to read `dist/<file>.d.ts` as input and also write it as output. **As a result**, I've set `rootDir` to `src/` rather than our conventional `./`. We'll want to change this in the future when we switch the entire package to TS and can turn off `allowJs`.
- A bunch of other Truffle packages do deep imports into some part of compile-solidity. These paths will need to include `dist/` from now on. This breaks the interface, naturally.
- All the TS files in `compilerSupplier/` now provide named exports only; this also is breaking.


.github/workflows/nodejs.yml
```diff
@@ -7,6 +7,7 @@ on:
     - master
     - develop
     - next
+    - supplier/*
```
gnidan, line 10: (Will need to take this out before merge)

packages/compile-solidity/package.json
```diff
@@ -47,7 +47,10 @@
     "glob": "^7.1.6",
     "mocha": "8.1.2",
     "sinon": "^9.0.2",
-    "tmp": "^0.2.1"
+    "tmp": "^0.2.1",
+    "ttypescript": "^1.5.12",
```
cds-amal, line 51: Note, [ttypescript](https://github.com/cevek/ttypescript/issues/122) maintenance status is a concern

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -72,7 +81,7 @@ class VersionRange {
       .map(solcVersion => {
         if (semver.satisfies(solcVersion, version)) return solcVersion;
       })
-      .filter(solcVersion => solcVersion);
+      .filter((solcVersion): solcVersion is string => !!solcVersion);
```
cds-amal, line 84: What does `solcVersion is string` do here? How is it different from `.filter((solcVersion: string) => !!solcVersion)`

cds-amal (approved):
Looks good to me. Though it's probably safer to get more eyes on the TS

packages/compile-solidity/package.json
gnidan: Understood, but I'm hoping we can use in this PR and figure out a solution across the board.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
gnidan: It's a [User-defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards); TS knows that the `boolean` returned by that function informs the type of `solcVersion`

gnidan: (Since the `.map()` operation will possibly return `undefined`, returning a type predicate like that means that TS won't think the type of `satisfyingVersions` is `string | undefined`)

cds-amal: TIL. Interesting syntax. Thanks!

packages/compile-solidity/src/compilerSupplier/normalizeSolcVersion.ts
```diff
@@ -0,0 +1,4 @@
+export const normalizeSolcVersion = (input) => {
```
haltman-at, line 5: Exported functions should really have type annotations, I think.

packages/compile-solidity/src/compilerSupplier/observeListeners.ts
```diff
@@ -18,6 +24,7 @@ class ObservedListeners {
   cleanup() {
     for (const eventName in this.listeners) {
       const marked = this.listeners[eventName];
+      // @ts-expect-error since eventName: string
```
haltman-at, line 27: OK, I pretty much ignored the listener stuff on the last PR, but this `@ts-expect-error` has made me curious.  I'm a little confused here because:

1. How are we using `process` without importing it?
2. What's this `process.listeners()` function, given that I can't find it documented in the Node API docs?
3. What type does it actually take (that apparently includes strings), what type is it typed as taking (that apparently does not include strings), and why is `@ts-expect-error` better here than some sort of conversion or coercion?

packages/compile-solidity/src/compilerSupplier/rangeUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+import fs from "fs-extra";
+import semver from "semver";
+import { Native, Local } from "./loadingStrategies";
+import { CompilerSupplier } from "./index";
+
+/**
+ * takes a version string which may be native or local, and resolves
+ * it to one which is (presumably) either a version or a version range
+ */
+export function resolveToRange(version) {
```
haltman-at, line 11: I think these functions should get type annotations, since they're exported.

packages/core/lib/testing/SolidityTest.js
```diff
@@ -3,7 +3,7 @@ const Suite = require("mocha/lib/suite.js");
 const Deployer = require("@truffle/deployer");
 const { Compile } = require("@truffle/compile-solidity");
 const { Shims } = require("@truffle/compile-common");
-const RangeUtils = require("@truffle/compile-solidity/compilerSupplier/rangeUtils");
+const RangeUtils = require("@truffle/compile-solidity/dist/compilerSupplier/rangeUtils");
```
haltman-at, line 6: Geez, I hope that goes away in an upcoming PR... maybe mark this?

packages/core/lib/testing/Test.js
```diff
@@ -10,10 +10,10 @@ const WorkflowCompile = require("@truffle/workflow-compile");
 const Resolver = require("@truffle/resolver");
 const TestRunner = require("./TestRunner");
 const SolidityTest = require("./SolidityTest");
-const RangeUtils = require("@truffle/compile-solidity/compilerSupplier/rangeUtils");
+const RangeUtils = require("@truffle/compile-solidity/dist/compilerSupplier/rangeUtils");
```
haltman-at, line 13: Similarly with these two.

packages/resolver/lib/sources/truffle/Deployed.ts
```diff
@@ -1,6 +1,6 @@
 const web3Utils = require("web3-utils");
 import semver from "semver";
-import RangeUtils from "@truffle/compile-solidity/compilerSupplier/rangeUtils";
+import * as RangeUtils from "@truffle/compile-solidity/dist/compilerSupplier/rangeUtils";
```
haltman-at, line 3: ...isn't there some way to make this work without `/dist/`?  Is that coming in one of the later PRs?

haltman-at (commented):
Mostly seems fine, a few things I'm confused about.

packages/compile-solidity/src/compilerSupplier/index.ts
```diff
@@ -1,12 +1,24 @@
...
+import semver from "semver";
 
-const { Docker, Local, Native, VersionRange } = require("./loadingStrategies");
+import { Docker, Local, Native, VersionRange } from "./loadingStrategies";
 
 const defaultSolcVersion = "0.5.16";
 
-class CompilerSupplier {
+export class CompilerSupplier {
+  private events: unknown;
```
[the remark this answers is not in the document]
gnidan: I'm fine switching to `any` if you think it's better; I generally default to using `unknown`, but I don't have an argument for it here.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/Docker.ts
```diff
@@ -67,8 +76,9 @@ class Docker {
 
     // grab the latest release from the forked releases stream;
     // coerce semver to remove possible `-alpine` suffix used by this repo
-    const latestRelease = semver.coerce(await asyncFirst(forkedReleases))
-      .version;
+    const latestRelease = semver.coerce(
+      await asyncFirst(forkedReleases)
+    )?.version;
```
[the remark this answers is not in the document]
gnidan: I'm pretty sure Husky ran for me at some point while doing this work?

Guess we'll see...

packages/compile-solidity/src/compilerSupplier/index.ts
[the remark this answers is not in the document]
haltman-at: OK, but what about the rest?

packages/compile-solidity/src/compilerSupplier/loadingStrategies/Docker.ts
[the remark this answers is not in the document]
haltman-at: I guess?

packages/compile-solidity/src/compilerSupplier/observeListeners.ts
gnidan: > 1. How are we using `process` without importing it?

It's a Node.js global.

> 2. What's this `process.listeners()` function, given that I can't find it documented in the Node API docs?

Oh jeez, good question. I don't know where I found this; I added it originally back in #4092. TS seems to know it? Weird.

> 3. What type does it actually take (that apparently includes strings), what type is it typed as taking (that apparently does not include strings), and why is `@ts-expect-error` better here than some sort of conversion or coercion?

Hm, it looks like I can cast this `eventName as NodeJS.Signals` and it compiles... lately I've taken to using `@ts-ignore` or `@ts-expect-error` instead of casting, since I don't (as a habit) want to muck with the actual types of things when I can just tell the compiler not to care about something.

Here it doesn't matter either way, but in some situations, `as` can remove all kinds of type information that you might want elsewhere.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/Docker.ts
```diff
@@ -151,7 +161,7 @@ class Docker {
         const tooManyRequests =
           error && error.response && error.response.status === 429;
 
-        return (
+        return !!(
```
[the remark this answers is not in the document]
gnidan: Did so

packages/compile-solidity/src/compilerSupplier/index.ts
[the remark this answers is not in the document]
gnidan: Fixed those. Thanks for catching them; I meant to go back and add real types.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/Docker.ts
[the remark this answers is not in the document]
gnidan: Lemme know if you want me to play it safe and change it :)

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -1,14 +1,23 @@
...
+import axios from "axios";
+import semver from "semver";
+import solcWrap from "solc/wrapper";
+import { Cache } from "../Cache";
+import { observeListeners } from "../observeListeners";
+import { NoVersionError, NoRequestError } from "../errors";
+
+export class VersionRange {
+  private config: {
+    events: any;
```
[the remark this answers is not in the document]
gnidan: (Added a comment here)

packages/compile-solidity/src/compilerSupplier/observeListeners.ts
haltman-at: Crazy!  Yeah let's go with that rather than using `@ts-expect-error`.

packages/core/lib/testing/SolidityTest.js
gnidan: It doesn't go away.

I don't have a good answer to this: in db and codec, etc., the approach we've settled on is to export everything in one module, subdivided into namespaces. Unfortunately, this means you can't do any tree-shaking; all the modules everywhere end up imported all at once, resulting in a large module size in memory and risk of incompatibility with things like `fs`.

I've left these here for now because I didn't want to go ahead and follow this prior approach here, knowing now the pretty severe downsides. My conclusion was that this deep-package import could really be considered internal to Truffle (users of the package very likely won't need to do this), and so it seemed acceptable for the time being.

Would love to discuss this though.

packages/compile-solidity/src/compilerSupplier/observeListeners.ts
haltman-at: (Although I still don't understand your claim that casting/coercing loses type info, since it only loses info in that particular expression... it doesn't lose info when you use that variable without a cast or coercion...)

packages/core/lib/testing/Test.js
gnidan: (Resolving this to move discussion to [other comment thread](https://github.com/trufflesuite/truffle/pull/4223#discussion_r688190965))

packages/resolver/lib/sources/truffle/Deployed.ts
gnidan: 😢 

(Resolving this to move discussion to [other comment thread](https://github.com/trufflesuite/truffle/pull/4223#discussion_r688190965))

packages/compile-solidity/src/compilerSupplier/loadingStrategies/Docker.ts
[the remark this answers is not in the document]
haltman-at: Eh let's leave it in for now and we can change it if it causes problems I guess.

packages/compile-solidity/src/compilerSupplier/observeListeners.ts
gnidan: > (Although I still don't understand your claim that casting/coercing loses type info, since it only loses info in that particular expression... it doesn't lose info when you use that variable without a cast or coercion...)

Well, if your expression gets passed to some function that doesn't have an explicit type annotation, that function will receive the casted type. This has bitten me enough times that I now avoid `as` like the plague.

gnidan: But here, have an `as` :)

packages/core/lib/testing/SolidityTest.js
haltman-at: Blech. :-/  Well, if you think this is the approach to take, I'll leave it to you.

packages/compile-solidity/src/compilerSupplier/index.ts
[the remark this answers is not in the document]
haltman-at: Hey, seems you missed a few here?  You got the ones inside `strategyOptions` but not outside.

haltman-at (commented):
[My review is that you should fix those few types that you missed that I mentioned above. :) ]

packages/compile-solidity/src/compilerSupplier/observeListeners.ts
haltman-at: > Well, if your expression gets passed to some function that doesn't have an explicit type annotation, that function will receive the casted type.

I am so confused as to why you would ever do this!  If the function doesn't have an explicit type annotation (and why doesn't it...?) then why are you even casting what you pass in?  But I think if anything the lesson here is to put type annotations on your functions...

packages/compile-solidity/src/compilerSupplier/index.ts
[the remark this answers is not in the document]
gnidan: Fixed those!

haltman-at (approved):
Alright, looks good!

gnidan:
Thanks @haltman-at and @cds-amal!

PR #4224, opened 2021-08-05
Add common @truffle/supplier

gnidan:
To build a supplier for a given software component (e.g. the solc CompilerSupplier), @truffle/supplier now exposes `forDefinition({ determineStrategy, strategyConstructors })`, which returns a `createSupplier` function as a result. `createSupplier` functions take a language-specific set of options, determine which of the defined strategies to use based on those options, then instantiate the determined strategy with those options.

This new package provides type-safety support by way of `forDefinition`'s being generic to a particular Supplier _Specification_. Defining an explicit Supplier specification (and corresponding Specifications for each Strategy) will ensure that `supplier.load(version?: string)` and `supplier.list()` throws errors unless the type-checker can guarantee the strategy implements those methods, using type guards, e.g.:
```typescript
const supplier = createCompilerSupplier(options);

// some strategies (e.g. Native) can't accept a specific version.
// without this check, TS will throw a type error
if (supplier.allowsLoadingSpecificVersion()) {
  const { solc } = await supplier.load("0.5.0");
  // ...
}

// some strategies don't allow you to list all versions
// another type check :)
if (supplier.allowsListingVersions()) {
  const { releases } = await supplier.list();
  // ...
}
```

~~**Question**: should this be moved to a new package (maybe @truffle/supplier)? There's nothing specific to compilers about it, really, and if we want to move common code for multiple languages here later, we might find ourselves adding dependencies to @truffle/compile-common that we don't want.~~ (doing this)


gnidan:
> **Question**: should this be moved to a new package (maybe @truffle/supplier)? There's nothing specific to compilers about it, really, and if we want to move common code for multiple languages here later, we might find ourselves adding dependencies to @truffle/compile-common that we don't want.

(I opened #4248 in case the answer to this question is "yes")

packages/compile-common/src/compilerSupplier/constructor.ts
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Output object created by specified constructor
+     */
+    result: any;
+  };
+
+  /**
+   * Getter type for constructor arg
+   */
+  export type Options<S extends Specification> = S["options"];
```
haltman-at, line 20: So basically you're making it look like there's something nice and generic going on here when actually it's all manual. :P

packages/compile-common/src/compilerSupplier/constructor.ts
```diff
@@ -0,0 +1,35 @@
...
+   */
+  export type Result<S extends Specification> = S["result"];
+}
+
+/**
+ * An object of type Constructor<S> is a JS constructor (or "class") that
+ * takes a sole argument (of the specified "options" type), and instantiates
+ * a new object in memory (of the specified "result" type).
+ */
+export type Constructor<S extends Constructor.Specification> = new (
```
haltman-at, line 33: Whoa, I had no idea about the use of `new` in types...

haltman-at (commented):
OK honestly my initial thought on reading this is, this is all so abstruse that I have no idea what it's doing.  This PR needs a more detailed description, basically.  Or perhaps more it should be combined with #4225... I shouldn't have to read a *later* PR to understand *this* one!  That indicates it should be a single PR that can be reviewed as a whole!  After all these types aren't doing anything by themselves.

(Edit: Or possibly it needs more inline comments explaining things on a more line-by-line basis... which it might need even *if* it's combined with #4225.)

I will also just generally question whether the types really need to be this complex, and ask whether perhaps there's a more comprehensible way of doing this.  I think this is going to be write-only code to everyone on the team but Nick.

Apologies for not doing a detailed review, but like... geez.  What a headache that would be!

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
+import type { Constructor as FreeConstructor } from "./constructor";
+import type { Results as FreeResults } from "./results";
+
+export namespace Strategy {
+  /**
+   * Type-level description of a compiler supplier strategy
+   */
+  export type Specification = {
```
haltman-at, line 8: All of these fields need to have their meaning documented in comments.  IMO, this should include which ones are "real" types, which ones are just flags for manually-programmed behavior, and which ones are a mix, but Nick denies that this is a meaningful distinction. :P  Regardless, they all should have their meaning documented in comments.

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+    allowsListingVersions: boolean;
+  };
+
+  /**
+   * Scoped re-exports of Constructor, using Strategy.Specification as
+   * generic param
+   */
+  export namespace Constructor {
+    /**
+     * Getter for the specified constructor, intersected with fixed result
```
haltman-at, line 21: Calling this a "getter" seems confusing, there are no getters in the usual sense here, these are all types!

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+     * type of Strategy<S>
+     */
+    export type Specification<S extends Strategy.Specification> =
+      S["constructor"] & { result: Strategy<S>; };
+
+    /**
+     * Constructor options argument for a specified strategy
```
haltman-at, line 28: Probably worth explicitly noting that this is just a re-export (so people don't go wasting time trying to figure out what it's doing).

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+   * Scoped re-exports of Constructor, using Strategy.Specification as
+   * generic param
+   */
+  export namespace Constructor {
+    /**
+     * Getter for the specified constructor, intersected with fixed result
+     * type of Strategy<S>
+     */
+    export type Specification<S extends Strategy.Specification> =
+      S["constructor"] & { result: Strategy<S>; };
```
haltman-at, line 25: Why are you adding the `result` field?  I don't actually understand that, it seems to go against what you're mostly doing?

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+     */
+    export type Options<S extends Strategy.Specification> = FreeConstructor.Options<
+      Strategy.Constructor.Specification<S>
+    >;
+  }
+
+  /**
+   * Type representing a constructor for specified strategy
+   */
+  export type Constructor<S extends Strategy.Specification> = FreeConstructor<
```
haltman-at, line 38: Again, note re-export

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+    Strategy.Constructor.Specification<S>
+  >;
+
+  /**
+   * Results types for a particular strategy
+   */
+  export namespace Results {
```
haltman-at, line 45: More re-exports to note

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+      Strategy.Results.Specification<S>
+    >;
+
+    export type List<S extends Strategy.Specification> = FreeResults.List<
+      Strategy.Results.Specification<S>
+    >;
+  }
+
+  /**
+   * Getter type for whether specified strategy allows argument to load()
```
haltman-at, line 59: Another confusing use of "getter" :P

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+   */
+  export type AllowsLoadingSpecificVersion<
+    S extends Strategy.Specification
+  > = S["allowsLoadingSpecificVersion"];
+
+  /**
+   * Getter type for whether specified strategy provides version listing
```
haltman-at, line 66: Another confusing use of "getter" :P

packages/compile-common/src/compilerSupplier/supplier.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export namespace Supplier {
+  /**
+   * Type-level description of a supplier
+   */
+  export type Specification = {
+    options: any;
+    results: FreeResults.Specification;
+    strategies: {
+      [strategyName: string]: Omit<FreeStrategy.Specification, "results">;
```
haltman-at, line 12: This should probably be changed so that the strategies' results types have to agree with that of the supplier, rather than being omitted.

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+        load(version?: string): Strategy.Results.Load<S>;
+      }
+    : {
+        load(): Strategy.Results.Load<S>;
+      }) &
+  (true extends Strategy.AllowsListingVersions<S>
+    ? {
+        list(): Strategy.Results.List<S>;
+      }
+    : {});
```
haltman-at, line 90: There's some copy-and-paste here... should these component types get individual names to reduce that?

packages/compile-common/src/compilerSupplier/strategy.ts
```diff
@@ -0,0 +1,106 @@
...
+  > = S["allowsListingVersions"];
+
+}
+
+/**
+ * An object of type Strategy<S> provides version loading and possibly version
+ * listing functionality
+ */
+export type Strategy<S extends Strategy.Specification> = BaseStrategy<S> &
+  (true extends Strategy.AllowsLoadingSpecificVersion<S>
```
haltman-at, line 79: Probably worth explicitly calling out the type-as-flag trick being used here.

haltman-at (commented):
OK, second attempt at a review.  I'm still going to skip over `supplier.ts`, which is the really confusing one, for now; Nick says I should probably wait on that until he's worked on this some more to make it clearer.

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: I'm not sure what copy-and-paste you're talking about?

packages/compile-common/src/compilerSupplier/supplier.ts
gnidan: done!

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: (reworded)

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: (reworded)

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: noted

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: noted

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: In the abstract, a constructor specification must include types for the options as well as types for the instantiated object result... but for a strategy, the constructor's `result` is always going to be that strategy.

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: explicitly noted.

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: I stopped using that word :)

packages/compile-common/src/compilerSupplier/strategy.ts
gnidan: Included docstrings for all fields in all specifications.

packages/compile-common/src/compilerSupplier/constructor.ts
gnidan: sure... we'll go with that?

packages/compile-common/src/compilerSupplier/strategy.ts
haltman-at: Oh, I see.  You're restricting that existing field, not adding a new one.

gnidan: Yup

packages/compile-common/src/compilerSupplier/strategy.ts
haltman-at: Like, comparing the definition of `Strategy` and the definition of `BaseStrategy`, where in both cases you give these conditions for them to also have the types
`{ load(version?: string): Strategy.Results.Load<S>; }` and `{ list(): Strategy.Results.List<S>; }`.  Should those types have names?

gnidan: Hm, yeah ok! How do you like this: https://github.com/trufflesuite/truffle/pull/4224/commits/bdd85cc7271f9af290f121ad0b32bb8bec753cdc

packages/supplier/README.md
```diff
@@ -0,0 +1,95 @@
...
+     };
+   }
+   ```
+
+2. Define one or more strategies.
+
+   e.g.:
+
+   ```typescript
+   import { Mixin } from "ts-mixer";
```
haltman-at, line 32: Could you make a point of also an including an example of how you'd do things without `ts-mixer`?

haltman-at (commented):
OK, further comments on this.  I remain really wary of the whole thing due to the complexity of the types; I'm still not sure that this is necessary, and that you couldn't get most of the effect a simpler (if looser) way.

Also, reviewing these would have been a lot easier had I read from the bottom up!  Each thing is defined in terms of things *below* it, and those components below it aren't easily-blackboxed abstractions, but rather preliminaries that need to be understood first for the whole thing to make sense.

packages/supplier/src/supplier.ts
```diff
@@ -0,0 +1,181 @@
...
+  strategyConstructors
+}: Supplier.Definition<S>): CreateSupplier<S> => <
+  N extends Supplier.StrategyName<S>
+>(
+  options: Supplier.Strategy.Constructor.Options<S, N>
+): Supplier<S, N> => {
+  const strategyName = determineStrategy(options);
+  const Strategy = strategyConstructors[strategyName];
+
+  // @ts-ignore since we can't figure out N from inside
```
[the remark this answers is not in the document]
gnidan: I'm of the mind now that `@ts-ignore` is safer than coercions

packages/supplier/src/supplier.ts
```diff
@@ -0,0 +1,181 @@
...
+ * @dev for known strategy names, this computes the method signature of
+ *      `load()` according to the corresponding strategy specification;
+ *
+ *      for the default of "any strategy name", this type will resolve to an
+ *      interface that may require the use of defined type guards
+ *      (e.g. `if(supplier.allowsListingVersions()) { supplier.list(); }`)
+ */
+export type Supplier<
+  S extends Supplier.Specification,
+  N extends Supplier.StrategyName<S> = Supplier.StrategyName<S>
```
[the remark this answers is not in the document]
gnidan: The former allows the parameter to be omitted, so you can do `Supplier<S>` instead of `Supplier<S, N>`

packages/supplier/src/supplier.ts
```diff
@@ -0,0 +1,181 @@
...
+> = Supplier.StrategyName<S> extends N
```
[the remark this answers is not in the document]
gnidan: If `N extends T` (enforced by the generic param) and `T extends N`, then `N` equals `T`. So this distinguishes supplier types for any strategy name vs. a specific strategy name.

I'll add a comment for this.

packages/supplier/src/supplier.ts
```diff
@@ -0,0 +1,181 @@
...
+  };
+
+  /**
+   * An object of type Definition<S> comprises constructors for each of the
+   * specified strategies, as well as a method to determine the name of which
+   * strategy to use for a given set of input options
+   */
+  export type Definition<S extends Supplier.Specification> = {
+    determineStrategy(
+      options: Options<S, Supplier.StrategyName<S>>
```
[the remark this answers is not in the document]
gnidan: Yep. I also think maybe it'd be clearer if the second param here were optional for `Options`

packages/supplier/src/supplier.ts
```diff
@@ -0,0 +1,181 @@
...
+    /**
+     * Type-level specification of method results for specified supplier;
+     * used in conjunction with each strategy individually
+     */
+    export type Specification<S extends Supplier.Specification> = S["results"];
+  }
+
+  export namespace Strategies {
+    /**
+     * Type-level specification of all strategies for a specified supplier
```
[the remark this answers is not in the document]
gnidan: Mapping from names to strategy specifications, yep.

packages/supplier/src/supplier.ts
```diff
@@ -0,0 +1,181 @@
...
+     * for all strategies.
+     */
+    results: FreeResults.Specification;
+
+    /**
+     * A supplier specification includes specifications for all strategies by
+     * name.
+     */
+    strategies: {
+      [strategyName: string]: FreeStrategy.Specification;
```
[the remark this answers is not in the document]
gnidan: Unfortunately, there's not a good way to do this enforcement here; since this is the base specification type, there's not a good way to enforce the cross-linkage.

The enforcement happens elsewhere: `forDefinition` will fail to type-check if given an `S` whose strategies' `results` types do not agree with each other.

packages/supplier/src/strategy.ts
```diff
@@ -0,0 +1,179 @@
...
+   * Type-level description of a supplier strategy
+   */
+  export type Specification = {
+    /**
+     * A strategy specification includes a constructor specification, minus
+     * the constructor specification's `result` property, since this module
+     * will inherently deal in constructors that result in an instance of
+     * the specified Strategy.
+     */
+    constructor: Omit<FreeConstructor.Specification, "result">;
```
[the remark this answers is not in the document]
gnidan: Yeah; the idea for this `Omit` is that the strategy specification **should not** include `result` for `constructor`.

packages/supplier/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "@types/node": "12.12.21",
+    "mocha": "8.0.1",
+    "ts-node": "^9.0.0",
+    "tsd": "^0.17.0",
+    "ttypescript": "^1.5.12",
+    "typescript": "^4.2.0",
+    "typescript-transform-paths": "^3.2.1"
+  },
+  "dependencies": {
+    "colors": "^1.4.0",
```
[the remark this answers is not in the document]
gnidan: Oops!

packages/supplier/src/supplier.ts
```diff
@@ -0,0 +1,181 @@
...
+
+    /**
+     * A supplier specification includes a unified `results` specification,
+     * which defines the `supplier.load()` and `supplier.list()` return types
+     * for all strategies.
+     */
+    results: FreeResults.Specification;
+
+    /**
+     * A supplier specification includes specifications for all strategies by
```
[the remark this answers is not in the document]
gnidan: I'm not sure what you're suggesting? I'm probably missing a more charitable interpretation of your statement, but I don't think it's correct to say that it's the _set of keys_ that define this; rather, _a given specification_ is the definition, itself... the keys are a part of that definition.

packages/supplier/src/supplier.ts
```diff
@@ -0,0 +1,181 @@
...
+ */
+export type CreateSupplier<S extends Supplier.Specification> = <
+  N extends Supplier.StrategyName<S>
+>(
+  options: Supplier.Strategy.Constructor.Options<S, N>
+) => Supplier<S, N>;
+
+/**
+ * An object of type Supplier<S, N> provides an interface for loading and
+ * possibly listing versions of the supplied component
```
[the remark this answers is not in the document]
gnidan: Does the signature for the type not count as documentation here? :)

I'll expand on this in natural language.

gnidan:
> Also, reviewing these would have been a lot easier had I read from the bottom up! Each thing is defined in terms of things below it, and those components below it aren't easily-blackboxed abstractions, but rather preliminaries that need to be understood first for the whole thing to make sense.

If you recall our discussion about this, it seems that the question of this directionality hinges on the open question of "which exports are truly external?" My stance is that the exports towards the bottom of the file are inherently more internal, and thus should not be at the top. Do you disagree and think that these types are intrinsically an externally-facing concern?

> OK, further comments on this. I remain really wary of the whole thing due to the complexity of the types; I'm still not sure that this is necessary, and that you couldn't get most of the effect a simpler (if looser) way.

So, the question is: is the additional type safety here worth the cognitive load? My thinking (obv.) is yes: I want the compiler to tell people who are building supplier strategies that they must know upfront (e.g.) if their strategy allows listing versions. I want new suppliers and new strategies for existing suppliers to require as little duplication as possible, while still making sure that all resultant `createSupplier(options)` functions are well-typed. I want to prevent mismatched return types for different strategies' `load()` methods. In my experience, these sorts guarantees are tremendously helpful.

I posit, thus, that this is a good trade-off: the cognitive load when implementing a new supplier is that you must figure out how to write a specification type; in return, you won't have to worry about figuring out what type something is _anywhere else_; the compiler will tell you when you mess up, and in the future, changes to that supplier won't run the risk of being incomplete.

(In case I'm not getting my full argument across: by counterexample, think of all the places in the code that one needs to touch just to do something as simple as add a field to @truffle/contract-schema. That's the sort of thing I'm trying to avoid with these specification types.)

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
haltman-at: Sorry, I should be more explicit.  I think the comment should state that.

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
haltman-at: Oh!  It's a default value.  I see, I didn't get that.  Perhaps a comment indicating that would be useful?

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
haltman-at: Oh, hm.  Well, if you can't enforce it in the type system proper, I'd suggest at least adding a comment to indicate that that's how you should use it.

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
haltman-at: I still don't understand this, sorry.  Admittedly you haven't written your comment yet.  But what you've said only explains the *condition* of the ternary.  That doesn't explain why you would use `BaseStrategy` in the one case and `FreeStrategy` in the other.  The thing I don't understand about the ternary is that the effect of the condition seems, as best I can tell, to be unrelated to the condition.

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
haltman-at: Yes, that's a different issue from what I'm talking about.

Your comment currently states, "a supplier specification includes specifications for all strategies by name".  OK.  But what defines what constitutes "all strategies"?  Someone attempting to use this type might misunderstand this, thinking that what constitutes "all strategies" is defined elsewhere -- perhaps as a fixed list defind somewhere -- and go on a wild goose chase trying to find where so they can find out what they are so they can be sure to include them all.

In fact, there is no such external list; it is the set of keys to this field that *defines* what constitutes "all strategies" in this context.  The comment should be clear that this field is *defining* what constitutes "all strategies", rather than the alternate interpretation that it *must include* all strategies, the set of which is defined elsewhere.

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
haltman-at: I strongly disagree with this point of view.  I do not think we should be using `ts-ignore`... basically at all.  We used it in that one place in `encoder` only because the configuration inconsistency (where the same code had to compile with `strictNullChecks` both on and off) left us with basically no other choice.

In general, we should try to work within TS, not ignore it.  Obviously coercions are not fully "working within TS", but they're closer.  TS does for instance perform rudimentary checks on coercions to make you're not sure you're doing anything too stupid with them, for instance; it obviously doesn't do that with ts-ignore.

You have elsewhere, although not here, made arguments as to why coercion is unsafe.  I do not see that any of these arguments apply here.  Indeed, I'm not convinced those arguments apply much of anywhere, as you've never provided any concrete examples, and the ones you've vaguely described seem to me to rely on what I would consider to be other bad practices; I would say those are the problems, not the coercion.

If you want to convince me on this point you are going to need to provide much more in the way of argument.

I would strongly discourage the use of ts-ignore both here and everywhere.

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
gnidan: I revised the docstring, hopefully it's clear enough from that?

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
gnidan: I added the comment

haltman-at:
> If you recall our discussion about this, it seems that the question of this directionality hinges on the open question of "which exports are truly external?" My stance is that the exports towards the bottom of the file are inherently more internal, and thus should not be at the top. Do you disagree and think that these types are intrinsically an externally-facing concern?

The idea of some things as internal and some things as external relies on the idea that interface can be separated from implementation, that the user can focus on certain things and ignore their internals.  Like, if you define a function, the user shouldn't need to look at what other functions that function calls to understand it; they shouldn't need to read the body at all.  You black-box the implementation and just focus on the interface.

The problem is, with types, there is very little distinction between a type's "interface" and it's "implementation".  Frankly I have little idea how I'd separate the two.  To understand a type, you generally need to have some understanding of what each of its components means.  This is how you come to understand a type -- by reading each of the components and the comments on them.  The whole thing is both interface and implementation; there's no separate body you can ignore!

Now often you don't need to understand all the components immediately and fully, so you can sensibly work top-down... but this relies on the components making some sort of a priori intuitive sense, so that you can have some reasonable provisional idea about what they are.  That doesn't apply here, because none of these types are intuitive.  My reaction on reading these isn't "oh OK I guess that's a blah, I can check the details of it later", it's, "Huh? what the hell is a blah? Guess I'd better stop what I'm reading and go jump to that and see".  And now I'm trying to figure out *two* confusing types.

Basically I don't think much of any of this is internal, because, well... it's types.  I don't know how one could be confident that they could black-box much of any of this.

...OK I got kind of pulled away before replying to the rest of this, I think to make sure this at least goes up I'm going to just hit post and can reply to the rest after in a separate comment...

haltman-at:
> So, the question is: is the additional type safety here worth the cognitive load? My thinking (obv.) is yes: I want the compiler to tell people who are building supplier strategies that they must know upfront (e.g.) if their strategy allows listing versions. I want new suppliers and new strategies for existing suppliers to require as little duplication as possible, while still making sure that all resultant createSupplier(options) functions are well-typed. I want to prevent mismatched return types for different strategies' load() methods. In my experience, these sorts guarantees are tremendously helpful.

Again, I think you could still get most of the type safety without such complication.

But I think there's another problem here: *What* new suppliers and new strategies?  How many of these are actually going to get written?

My understanding was that the point of this was to allow the writing of a Vyper CompilerSupplier.  Note, that's specifically a second *complier* supplier; it doesn't require a more general "supplier" notion.  And, um, do we expect to be writing any more suppliers than that...?

I guess what I'm realizing now is that part of my objection is that this is a lot of abstraction that is hardly ever going to get used.  You're trying to abstract things that we simply have no plan for.  Better to just make an abstraction that covers the things we actually plan to do.

Actually, this makes me realize there's some bits of the design I hadn't properly processed and criticized before, but I'll post that in a separate comment.

> I posit, thus, that this is a good trade-off: the cognitive load when implementing a new supplier is that you must figure out how to write a specification type; in return, you won't have to worry about figuring out what type something is anywhere else; the compiler will tell you when you mess up, and in the future, changes to that supplier won't run the risk of being incomplete.

I don't think "you won't have to worry about figuring out what type something is anywhere else" is a good goal.  Like, you should always have types on your functions.

> (In case I'm not getting my full argument across: by counterexample, think of all the places in the code that one needs to touch just to do something as simple as add a field to @truffle/contract-schema. That's the sort of thing I'm trying to avoid with these specification types.)

What makes you think that anything like that situation would arise without this mechanism?

haltman-at:
OK, so, here's something I only just realized; Why do strategies have names?

In the existing way, strategies don't have names.  This whole strategy-name thing is invented here.

Like, imagine we stuck closer to the original code.  Then instead of having `determineStrategy` which returns a name, and having strategies indexed by those names, we would instead simply have a list of strategies, and each strategy instead of having a name would have an `isApplicable` (or something) predicate, and `determineStrategy`, rather than being something user-supplied, would just consist of checking the list of strategies in order to find the first one that's applicable.

The whole strategy-name thing seems like a needless complication.  What if that were cut out?  (Although, perhaps it's impossible to do all the type things you want to do if that were cut out?)

packages/supplier/src/supplier.ts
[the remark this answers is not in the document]
haltman-at: Oh yeah, revising the docstring is a better idea.  What I was confused about was the syntax, so I was thinking of just adding a comment to clear up the syntax, but revising the docstring is actually probably the better idea.

Unfortunately the new material that you added to the docstring is not entirely clear.  When you write "When N is omitted, Supplier<S> represents a supplier that uses any specified strategy"... "any specified strategy" is a terribly unclear phrase.  No strategy was specified!  I eventually figured out that you meant "any of the strategies appearing in the specification `S`".

gnidan:
Closing this for cleanup. If we want to continue on this front, I think it might be worth starting from tabula rasa

PR #4225, opened 2021-08-05
Use @truffle/supplier to rewrite @truffle/compile-solidity's CompilerSupplier

gnidan:
Follows from #4223 (for TS support in @truffle/compile-solidity) and #4224 (for @truffle/supplier)

This PR removes the `CompilerSupplier` class defined in @truffle/compile-solidity and replaces it with a `createCompilerSupplier` function, built via the mechanisms provided by #4224. This new function behaves largely the same as the prior class constructor.

Some notable changes here:
- Docker no longer requires a version upfront; if you don't specify a version but you specify `docker: true`, then it'll use the default solc verson (same as how VersionRange works)
- RangeUtils relies on the common interface for `list()`, which returns a promise, so `resolveToRange()` now is `async`.
- I've added `allowsLoadingSpecificVersion()` and `allowsListingVersions()` type guards to each of the loading strategies
- Each of the loading strategies now returns `{ solc }` instead of `solc`. **(Should we change this to remove the extra wrapping?)**

### DIFF
--- a/packages/compile-solidity/package.json
+++ b/packages/compile-solidity/package.json
@@ -23,6 +23,7 @@
     "@truffle/config": "^1.3.4",
     "@truffle/contract-sources": "^0.1.12",
     "@truffle/expect": "^0.0.18",
+    "@truffle/supplier": "^0.1.0-0",
     "axios": "^0.21.1",
     "axios-retry": "^3.1.9",
     "debug": "^4.3.1",
@@ -34,7 +35,8 @@
     "original-require": "^1.0.1",
     "require-from-string": "^2.0.2",
     "semver": "^7.3.4",
-    "solc": "0.6.9"
+    "solc": "0.6.9",
+    "ts-mixer": "^6.0.0"
   },
   "devDependencies": {
     "@truffle/artifactor": "^4.0.118",
--- a/packages/compile-solidity/src/compilerSupplier/index.ts
+++ b/packages/compile-solidity/src/compilerSupplier/index.ts
@@ -32,18 +32,28 @@ export const createCompilerSupplier = forDefinition<
       solcConfig: { docker = false, version }
     } = options;
 
+    // docker case: simple flag
     if (docker) {
       return "docker";
     }
 
+    // native case: version passed as string literal `"native"`
     if (version === "native") {
       return "native";
     }
 
-    if (version && fileExists(version)) {
+    // local soljson case: version is arguably a path
+    if (version && (
+      // an absolute path is definitely a path
+      path.isAbsolute(version) ||
+      // interpret version string as path if its point to real file on disk
+      fs.existsSync(version)
+    )) {
       return "local";
     }
 
+    // remote soljson case: version is unspecified or conforms to semver
+    // (default case)
     if (!version || semver.validRange(version)) {
       return "version-range";
     }
@@ -65,7 +75,3 @@ export const createCompilerSupplier = forDefinition<
     "version-range": VersionRange
   }
 });
-
-function fileExists(localPath) {
-  return fs.existsSync(localPath) || path.isAbsolute(localPath);
-}
--- a/packages/compile-solidity/src/compilerSupplier/index.ts
+++ b/packages/compile-solidity/src/compilerSupplier/index.ts
@@ -3,7 +3,7 @@ import fs from "fs";
 import semver from "semver";
 
 import type { Results } from "./types";
-import { forDefinition } from "@truffle/supplier";
+import { Supplier, forDefinition } from "@truffle/supplier";
 import { Docker, Local, Native, VersionRange } from "./loadingStrategies";
 
 export namespace CompilerSupplier {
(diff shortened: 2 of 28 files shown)


packages/compile-solidity/src/compilerSupplier/index.ts
```diff
@@ -2,114 +2,70 @@ import path from "path";
...
+    "local": Local,
+    "native": Native,
+    "version-range": VersionRange
   }
+});
 
-  fileExists(localPath) {
-    return fs.existsSync(localPath) || path.isAbsolute(localPath);
-  }
+function fileExists(localPath) {
```
gnidan, line 69: As @haltman-at [mentioned](https://github.com/trufflesuite/truffle/pull/4222#discussion_r687277390), this should be renamed or inlined.

packages/compile-solidity/src/compilerSupplier/index.ts
```diff
@@ -2,114 +2,70 @@ import path from "path";
...
     const message =
-      `Could not find a compiler version matching ${userSpecification}. ` +
+      `Could not find a compiler version matching ${version}. ` +
       `compilers.solc.version option must be a string specifying:\n` +
       `   - a path to a locally installed solcjs\n` +
       `   - a solc version or range (ex: '0.4.22' or '^0.5.0')\n` +
       `   - a docker image name (ex: 'stable')\n` +
       `   - 'native' to use natively installed solc\n`;
-    return new Error(message);
+    throw new Error(message);
```
gnidan, line 58: As @haltman-at [mentions](https://github.com/trufflesuite/truffle/pull/4222#discussion_r687276676), this likely warrants its own error class to conform with the rest of this module.

packages/core/lib/commands/compile.js
```diff
@@ -97,10 +97,10 @@ const command = {
...
 
   listVersions: async function (options) {
-    const { CompilerSupplier } = require("@truffle/compile-solidity");
+    const { createCompilerSupplier } = require("@truffle/compile-solidity");
     const { asyncTake } = require("iter-tools");
 
-    const supplier = new CompilerSupplier({
+    const supplier = createCompilerSupplier({
       solcConfig: {
         ...options.compilers.solc,
```
gnidan, line 105: As @haltman-at [mentions](https://github.com/trufflesuite/truffle/pull/4222#discussion_r687283884), this will cause problems if `version` is set to `"native"` or to a local soljson path.

This should probably explicitly set `version: undefined` to ensure VersionRange or Docker get used.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -9,39 +9,77 @@ import solcWrap from "solc/wrapper";
...
+    results: Results.Specification;
+    allowsLoadingSpecificVersion: true;
+    allowsListingVersions: true;
+  };
+}
 
-export class VersionRange {
+export class VersionRange implements CompilerSupplier.Strategy<
+  VersionRange.Specification
+> {
```
gnidan, line 35: As @haltman-at mentions in [his review of #4222](https://github.com/trufflesuite/truffle/pull/4222#pullrequestreview-728040585), we should remove the method `findNewestValidVersion()` and replace it with `semver.maxSatisfying()`.

packages/compile-solidity/package.json
```diff
@@ -34,7 +35,8 @@
     "original-require": "^1.0.1",
     "require-from-string": "^2.0.2",
     "semver": "^7.3.4",
-    "solc": "0.6.9"
+    "solc": "0.6.9",
```
axic, line 38: Why is this using such an old version?

haltman-at: We're only using it for `solc/wrapper`, so it mostly doesn't matter.  It's an old version because there may have been problems with newer versions on Node 10, which we haven't yet officially dropped support for?  I may be wrong about this and there may not be a problem after all.  Once we drop support for Node 10 we should definitely be able to update.

axic: > We're only using it for solc/wrapper, so it mostly doesn't matter.

It matters exactly for that reason, as newer wrapper versions have backwards compatibility functionality, so as long as you use the latest wrapper, it is "guaranteed" to work with every previous version. However it does not offer forwards compatibility, so we can create a breaking release binary which won't work with an old wrapper anymore.

axic: In this particular case I think you are fine, as the last such change happened for 0.6.9 in the wrapper, but keep in mind that you may encounter breakage in the future.

axic: >  It's an old version because there may have been problems with newer versions on Node 10, which we haven't yet officially dropped support for? I may be wrong about this and there may not be a problem after all. Once we drop support for Node 10 we should definitely be able to update.

Hmm, can you report the issue? Circleci runs with node 8 / 10 / 12 on solc-js, and we do not see any problem.

axic: And added CI for node 14/16 just to be sure: https://github.com/ethereum/solc-js/pull/548

haltman-at: OK, you might want to open a separate issue for this since this is basically unrelated to the contents of this PR.

gnidan:
Closing this for cleanup. If we want to continue on this front, I think it might be worth starting from tabula rasa

PR #4227, opened 2021-08-05
Add commented storage options for @truffle/db

cds-amal:
This PR adds commented db settings to `truffle-config.js` so that a developer can quickly enable and configure local db storage.

### DIFF
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -94,7 +94,7 @@ module.exports = {
   },
 
   // Truffle DB is currently disabled by default; to enable it, change enabled:
-  // false to enabled: true, the default storage location can also be
+  // false to enabled: true. The default storage location can also be
   // overridden by specifying the adapter settings.
   //
   // Note: your previously migrated contracts will not be available in
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -99,7 +99,7 @@ module.exports = {
   //
   // Note: your previously migrated contracts will not be available in
   // truffle-db. We recommnd you backup your artifacts before utilizing Truffle
-  // DB the first time with:
+  // DB for the first time with:
   // $ truffle migrate --reset --compile-all
 
   db: {
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -104,7 +104,7 @@ module.exports = {
 
   db: {
     enabled: false
-    //, adapter: {
+    // adapter: {
     //   name: "sqlite",
     //   settings: {
     //     directory: ".truffle-db"
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -95,7 +95,7 @@ module.exports = {
 
   // Truffle DB is currently disabled by default; to enable it, change enabled:
   // false to enabled: true. The default storage location can also be
-  // overridden by specifying the adapter settings.
+  // overridden by specifying the adapter settings, as shown in the commented code below.
   //
   // Note: your previously migrated contracts will not be available in
   // truffle-db. We recommnd you backup your artifacts before utilizing Truffle
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -97,8 +97,9 @@ module.exports = {
   // false to enabled: true, the default storage location can also be
   // overridden by specifying the adapter settings.
   //
-  // Note: if you migrated your contracts prior to enabling this field in your Truffle project and want
-  // those previously migrated contracts available in the .db directory, you will need to run the following:
+  // Note: your previously migrated contracts will not be available in
+  // truffle-db. We recommnd you backup your artifacts before utilizing Truffle
+  // DB the first time with:
   // $ truffle migrate --reset --compile-all
 
   db: {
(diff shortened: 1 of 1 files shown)


packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -93,13 +93,22 @@ module.exports = {
     }
   },
 
-  // Truffle DB is currently disabled by default; to enable it, change enabled: false to enabled: true
+  // Truffle DB is currently disabled by default; to enable it, change enabled:
+  // false to enabled: true, the default storage location can also be
```
fainashalts, line 97:
```suggestion
  // false to enabled: true. The default storage location can also be
```

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -93,13 +93,22 @@ module.exports = {
...
+  // overridden by specifying the adapter settings.
```
fainashalts, line 98:
```suggestion
  // overridden by specifying the adapter settings, as shown in the commented code below.
```

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -93,13 +93,22 @@ module.exports = {
...
   //
-  // Note: if you migrated your contracts prior to enabling this field in your Truffle project and want
-  // those previously migrated contracts available in the .db directory, you will need to run the following:
+  // Note: your previously migrated contracts will not be available in
+  // truffle-db. We recommnd you backup your artifacts before utilizing Truffle
+  // DB the first time with:
```
fainashalts, line 102:
```suggestion
  // DB for the first time with:
```

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -93,13 +93,22 @@ module.exports = {
...
   // $ truffle migrate --reset --compile-all
 
   db: {
     enabled: false
```
fainashalts, line 106:
```suggestion
    enabled: false,
```

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -93,13 +93,22 @@ module.exports = {
...
+    //, adapter: {
```
fainashalts, line 107:
```suggestion
    // adapter: {
```

fainashalts (changes requested):
Looks good, a couple of small edits!

packages/core/lib/commands/init/initSource/truffle-config.js
cds-amal: The comma is needed to separate `enabled: false` which  would be needed if the adapter setting is un-commented

cds-amal:
Thanks @fainashalts, I love suggested commits!

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -93,13 +93,22 @@ module.exports = {
...
 
-  // Truffle DB is currently disabled by default; to enable it, change enabled: false to enabled: true
+  // Truffle DB is currently disabled by default; to enable it, change enabled:
+  // false to enabled: true. The default storage location can also be
+  // overridden by specifying the adapter settings, as shown in the commented code below.
   //
-  // Note: if you migrated your contracts prior to enabling this field in your Truffle project and want
-  // those previously migrated contracts available in the .db directory, you will need to run the following:
+  // Note: your previously migrated contracts will not be available in
+  // truffle-db. We recommnd you backup your artifacts before utilizing Truffle
```
fainashalts, line 101:
```suggestion
  // truffle-db. We recommend you backup your artifacts before utilizing Truffle
```

fainashalts (changes requested):
Oops, missed one typo on my first pass!

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -93,13 +93,22 @@ module.exports = {
...
+  // DB for the first time with:
   // $ truffle migrate --reset --compile-all
```
gnidan, line 103: Note that this won't "back up" anything! This will destroy the backups.

fainashalts: Oh hahaha I can't believe I missed that! Come to think of it, the previously migrated contracts *would* still be available via artifacts, right? Just not in `db`? What should we suggest to people in that situation? This may be a bit more nuanced than I originally thought.

cds-amal: I read that as back up your artifacts --stop-- before moving on to using truffle db. I see the ambiguity now.

packages/core/lib/commands/init/initSource/truffle-config.js
cds-amal: Changed the text to remove ambiguity. See latest changes

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -93,13 +93,23 @@ module.exports = {
...
-  // those previously migrated contracts available in the .db directory, you will need to run the following:
+  // NOTE: It is not possible to migrate your contracts to truffle DB and you should
+  // make a backup of your artifacts to a safe location before enabling this feature.
+  //
+  // After you backed up your artifacts you can utilize db by running migrate as follows: 
   // $ truffle migrate --reset --compile-all
 
   db: {
-    enabled: false
+    enabled: false,
```
fainashalts, line 107: do we need to comment this out to be consistent with the metacoin box?

fainashalts: Or alternatively uncomment it there I suppose. I don't have a preference, just want consistency.

fainashalts:
I think this looks good (one more comment added!). My only remaining concern is that I think you should probably copy the text in the config here to be exactly what you see in the metacoin-box. And in general we should probably add it to all of our boxes in this way so its consistent (would be nice if there was an automated way to update them...)? Though of course that could be a task for another time.

fainashalts (approved):

PR #4232, opened 2021-08-06
Use mixin-based architecture for building compile-solidity loading strategies

gnidan:
Use [ts-mixer](https://github.com/tannerntannern/ts-mixer) which provides the `extends Mixin(...)` construct.

Define mixins `AllowsLoadingSpecificVersion`, `ForbidsListingVersions`, et al., as well as a `HasCache` mixin, since that's common to VersionRange+Docker.

**Question**: should the allows/forbids mixins be moved to compile-common? They might belong there.

(Just a little bonus PR on top of #4225, in case the mixin approach looks nice. :)


gnidan:
Closing this and redoing this in #4225

Issue #4235, opened 2021-08-06
Truffle debug's code highlighting incorrectly highlights some code


davidmurdoch:
## Issue

`truffle debug 0x4e67f687fbec1989df366e3a4b6a25e8b62bca8db80afe90918dcc8fbe425ec7 -x` results in some "pretty" colors in the debug output:

![image](https://user-images.githubusercontent.com/187813/128570390-4afa6e83-35bb-4c95-8966-a566bf045ad6.png)

## Steps to Reproduce

I used an unreleased version of ganache v7 for this (will release in a `@canary` release today). Repro:

1. `npm uninstall ganache-cli -g && npm install ganache@canary -g`
2. `ganache --fork URL_HERE`
3. set up your truffle config, or however you want to get truffle debug to listen to your ganache instance
4. `truffle debug 0x4e67f687fbec1989df366e3a4b6a25e8b62bca8db80afe90918dcc8fbe425ec7 -x`
5. Cycle through the contract to get the pretty colors!

## Expected Behavior

Normal colors, not pretty ones.

## Actual Results

![image](https://user-images.githubusercontent.com/187813/128570390-4afa6e83-35bb-4c95-8966-a566bf045ad6.png)

## Environment

* Operating System: Kubuntu 20.04
* Ethereum client: ganache@canary
* Truffle version (`truffle version`): v5.4.5 (core: 5.4.5)
* node version (`node --version`): v12.22.1
* npm version (`npm --version`): 6.14.12

haltman-at:
Note: This is apparently a mainnet transaction.

haltman-at:
OK, merged a PR to fix this.

PR #4236, opened 2021-08-07
Update highlightjs-solidity to 1.2.1

haltman-at:
Addresses #4235.

Updates highlightjs-solidity to a version that includes [this](https://github.com/highlightjs/highlightjs-solidity/pull/51) PR.  So, we'll fix highlighting problems, at the expense of no longer having punctuation highlighting.


cds-amal (approved):
Looks good @haltman-at.

Issue #4239, opened 2021-08-08
Truffle console context is incomplete


cds-amal:
Truffle console is missing some definitions in its context.

| Truffle v5.3.9                       | Truffle v5.4.x                              |
|-------------------------------------|------------------------------------|
| truffle(develop)> Object.keys(this) | truffle(develop)> Object.keys(this)|
|  global                             |                                    |
|  clearInterval                      |                                    |
|  clearTimeout                       |                                    |
|  setInterval                        |                                    |
|  setTimeout                         |                                    |
|  queueMicrotask                     |                                    |
|  clearImmediate                     |                                    |
|  setImmediate                       |                                    |
|  __core-js_shared__                     |                                    |
|  @@any-promise/REGISTRATION         |                                    |
|  web3                               | web3                               |
|  interfaceAdapter                   | interfaceAdapter                   |
|  accounts                           | accounts                           |
|  ConvertLib                         | ConvertLib                         |
|  MetaCoin                           | MetaCoin                           |
|  Migrations                         | Migrations                         |

eggplantzzz:
Fixed with https://github.com/trufflesuite/truffle/releases/tag/v5.4.6.

Issue #4238, opened 2021-08-07
replace global with globalThis


segersb:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue
The "global" variable should be replaced with the standardized "globalThis"
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

## Steps to Reproduce

run `truffle develop`

truffle(develop)> let accounts = await web3.eth.getAccounts()
Uncaught ReferenceError: global is not defined
    at evalmachine.<anonymous>:2:19

global = globalThis

truffle(develop)> let accounts = await web3.eth.getAccounts()
undefined
truffle(develop)> accounts
[
  '0xc30567b10027c99C6b22d578B9dEB85990b954bE',
  '0xDCF9e2547212Ad204db456Af4B99a868e25293b0',
  '0xaAC71A7B59C83A67bF1d616F411C1570fC58029D',
  '0x7Aa3211e33D24eF45B42016d6c5B748657f477f2',
  '0x3B2D7Cc5bA1f7B9d2a711C0A1cBab28Ef0cB7B46'
]

## Expected Behavior

Truffle develop should work without using the "global = globalThis" workaround

## Actual Results

Truffle develop returns the error "global is not defined"

## Environment

* Operating System: Windows
* Ethereum client: Truffle develop

Truffle v5.4.5 (core: 5.4.5)
Solidity v0.5.16 (solc-js)
Node v14.17.4
Web3.js v1.5.1

cds-amal:
Thanks for raising this issue @segersb . We're aware of it and have a proposed fix awaiting review. 
A temporary workaround is to explicitly define `global` in the REPL: 
```
global = this 
```
Note: `globalThis` is not available in node 10

segersb:
Yeah, I found the workaround and I'm using the LTS version of Node.
The stable node version is 14 and I think the latest is 16, so I guess most people will be ok with a min version of 10.
I'll close the issue since it's already being handled.
Thanks

PR #4242, opened 2021-08-09
Fix/console globals

cds-amal:
This PR
  - changes how the REPL context is instantiated and. 
  - removes annoying `undefined` from REPL output. (See [docs](https://nodejs.org/dist/latest-v14.x/docs/api/repl.html#repl_repl_start_options))
  - adds tests using the new framework @eggplantzzz introduced.
  - should resolve 
    - #4238 
    - #4239 

This is necessary to capture the globals that the REPL includes in the context, which needed for a good user experience. Being able to `console.log` contributes to good experience.

The motivation for the previous change was to allow the truffle context variables to be hydrated before processing user input which is needed for our tests.  This PR still allows for that by coordinating when to send the input from tests. The sequence of events is as follows:

Child Process:
  1. REPL starts with nodes context and prompt = ''
  1. REPL loads truffle specific variables into context
  1. REPL sets the prompt to '(develop)>'

Test runner:
  1. Scans child's sdtout looking for '(develop)>' before sending input

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -61,22 +61,31 @@ class Console extends EventEmitter {
 
   async start() {
     try {
-      // we calculate the context before calling `start`  so that the context
-      // variables are available as early as possible after the repl starts
-      const context = { ...(await this.calculateTruffleAndUserGlobals()) };
-      // we set up this.repl here because `this.provision` sets values on
-      // this.repl.context - this will finish setting up the environment
-      this.repl = { context };
-      // this hydrates the environment with the user's contracts
-      this.provision();
-
+      // start the repl with an empty prompt and show a proper one when
+      // the repl has set up its context and is ready to accept input
       this.repl = repl.start({
-        prompt: "truffle(" + this.options.network + ")> ",
+        prompt: "",
+        ignoreUndefined: true,
         eval: this.interpret.bind(this)
       });
-      this.repl.context = context;
 
-      //want repl to exit when it receives an exit command
+      // Get and set Truffle and User Globals
+      const truffleAndUserGlobals = await this.calculateTruffleAndUserGlobals() ;
+      Object.entries(truffleAndUserGlobals).forEach(([key, value]) => {
+        Object.defineProperty(this.repl.context, key, {
+          configurable: false,
+          enumerable: true,
+          value,
+        });
+      });
+
+      // repl is ready - set and display prompt
+      this.repl.setPrompt("truffle(" + this.options.network + ")> ");
+      this.repl.displayPrompt();
+
+      // hydrate the environment with the user's contracts
+      this.provision();
+
       this.repl.on("exit", () => {
         process.exit();
       });
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -71,11 +71,7 @@ class Console extends EventEmitter {
       // Get and set Truffle and User Globals
       const truffleAndUserGlobals = await this.calculateTruffleAndUserGlobals() ;
       Object.entries(truffleAndUserGlobals).forEach(([key, value]) => {
-        Object.defineProperty(this.repl.context, key, {
-          configurable: false,
-          enumerable: true,
-          value,
-        });
+        this.repl.context[key] = value;
       });
 
       // repl is ready - set and display prompt
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -65,7 +65,6 @@ class Console extends EventEmitter {
       // the repl has set up its context and is ready to accept input
       this.repl = repl.start({
         prompt: "",
-        ignoreUndefined: true,
         eval: this.interpret.bind(this)
       });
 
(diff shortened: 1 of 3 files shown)


packages/core/lib/console.js
```diff
@@ -61,22 +61,31 @@ class Console extends EventEmitter {
...
-      this.repl = { context };
-      // this hydrates the environment with the user's contracts
-      this.provision();
-
+      // start the repl with an empty prompt and show a proper one when
+      // the repl has set up its context and is ready to accept input
       this.repl = repl.start({
-        prompt: "truffle(" + this.options.network + ")> ",
+        prompt: "",
+        ignoreUndefined: true,
```
davidmurdoch, line 68: Is this change considered breaking?

cds-amal: I'm not sure, but leaning towards removing it.

cds-amal: [Adding documentation for reviewers]
[ignoreUndefined](https://nodejs.org/dist/latest-v14.x/docs/api/repl.html#repl_repl_start_options) :- <boolean> If true, specifies that the default writer will not output the return value of a command if it evaluates to undefined. **Default**: false.

packages/core/lib/console.js
```diff
@@ -61,22 +61,31 @@ class Console extends EventEmitter {
...
 
-      //want repl to exit when it receives an exit command
+      // Get and set Truffle and User Globals
+      const truffleAndUserGlobals = await this.calculateTruffleAndUserGlobals() ;
+      Object.entries(truffleAndUserGlobals).forEach(([key, value]) => {
+        Object.defineProperty(this.repl.context, key, {
+          configurable: false,
+          enumerable: true,
+          value,
+        });
```
cds-amal, line 79: This is a breaking change that we may not want. Users will no longer be able to delete truffleAndUserGlobals from the global context. This means, accounts for example:
```
delete this.accounts // will not delete or reassign accounts
accounts.splice(1,5) // will still change the contents, along with push, pop
```

cds-amal (commented):
wanted to point out a breaking change for discussion/review. I don't have a strong opinion on the behavior.

packages/core/lib/console.js
eggplantzzz: Do we want this? It should behave like a normal Node repl, right?

packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -51,6 +54,20 @@ module.exports = {
       });
 
       child.stdout.on("data", data => {
+        // accumulate buffer from chunks
```
eggplantzzz, line 57: Is this not currently working the way we thought it was?

packages/core/lib/console.js
```diff
@@ -61,22 +61,31 @@ class Console extends EventEmitter {
...
+        ignoreUndefined: true,
         eval: this.interpret.bind(this)
       });
-      this.repl.context = context;
 
-      //want repl to exit when it receives an exit command
+      // Get and set Truffle and User Globals
+      const truffleAndUserGlobals = await this.calculateTruffleAndUserGlobals() ;
+      Object.entries(truffleAndUserGlobals).forEach(([key, value]) => {
+        Object.defineProperty(this.repl.context, key, {
```
eggplantzzz, line 75: Is it important that we define these using `defineProperty`? Why couldn't we just push them more naïvely into the context like we were doing before?

packages/core/lib/console.js
cds-amal: Removing it from this PR to be revisited in follow on work

packages/core/lib/console.js
cds-amal: After discussing with @eggplantzzz we feel users should have unfettered access to the context.

packages/truffle/test/scenarios/commandRunner.js
cds-amal: The child process will only print the repl prompt when it is ready (has its context loaded with truffle variables). The command runner monitors the child's output, looking for that repl prompt (flag) before sending data to it

packages/core/lib/console.js
cds-amal: Reverted to prior behavior.

packages/core/lib/console.js
```diff
@@ -61,22 +61,26 @@ class Console extends EventEmitter {
...
       this.repl = repl.start({
-        prompt: "truffle(" + this.options.network + ")> ",
+        prompt: "",
         eval: this.interpret.bind(this)
       });
-      this.repl.context = context;
 
-      //want repl to exit when it receives an exit command
+      // Get and set Truffle and User Globals
+      const truffleAndUserGlobals = await this.calculateTruffleAndUserGlobals() ;
```
eggplantzzz, line 72: There is an extra, unnecessary space at the end of the line

eggplantzzz (approved):
I think it looks good, I did a simple test and all the globals are there.

PR #4245, opened 2021-08-10
Bug fix: Fix incorrect variable name

haltman-at:
This file refers to `boxContents`, but no such variable exists.  It looks like `projectFiles` was meant.


cds-amal (approved):
Good catch Harry!

PR #4248, opened 2021-08-11
Make dedicated @truffle/supplier package

gnidan:
This moves the compiler supplier stuff out of compile-common and into a new package of its own. 

(This is follow-on work for #4224; if we move forward with this new package, I'll need to rebase #4225 and #4232)


gnidan:
Merging this to avoid juggling two branches

PR #4255, opened 2021-08-12
Adds note in README to use nvm when installing truffle

JimLynchCodes:


gnidan (approved):
Thanks for this!

haltman-at (approved):
Seems good to me!

Issue #4237, opened 2021-08-07
Events ignore quiet setting (so e.g. compilation warnings are displayed at inappropriate times)


haltman-at:
## Issue

The events system largely ignores the `quiet` setting.  This causes things to be printed to the console when they shouldn't be.  For instance, if you do `truffle debug --fetch-external`, and the downloaded contract has compilation warnings, you'll see those compilation warnings, even though the debugger makes sure to set `quiet: true` when performing compilation.

Now, if you look at the code in `events`, you'll notice that there are a number of checks of `this.quiet`.  These checks do work.  But the problem is, `this.quiet` is only set on events system startup!  And often `quiet` is set to true *after* the events system is initialized.  So in order for `quiet` to work, somehow the event system would have to know that `quiet` was now true!

Basically, the `quiet` option in the events system doesn't currently achieve the goal of letting us use `quiet: true` in the way we normally would.

A possible stopgap would be to add manual `quiet` checks before emitting such events, but obviously that's not a great solution.  (Or maybe it's the right solution after all?  IDK.)

Another thing we could do for now is to re-initialize the events system everywhere that we set `quiet: true` after the events system has been initialized. Whee. :-/

A third thing we could do for now would be to, everywhere we set `quiet: true` after the events system has been initialized, reach into the events system and manually set *its* `quiet` to true, assuming that's possible... :-/

So uh yeah none of this seems great but we ought to do *something* because right now starting up the debugger can show you compiler warnings.

## Steps to Reproduce

As an example, try debugging mainnet transaction 0x4e67f687fbec1989df366e3a4b6a25e8b62bca8db80afe90918dcc8fbe425ec7 (the same one from #4235) with `--fetch-external`.  You'll see compilation warnings displayed as the debugger starts up.  Those shouldn't be shown during debugger startup.

## Environment

* Truffle version (`truffle version`): 5.4.5

eggplantzzz:
Thanks @haltman-at. Yeah, so the events system gets re-initialized every time a new config is loaded but, like you mentioned, it looks like the value of `quiet` must change without re-initializing a new config. The events system has a method for hooking into the settings called [`updateSubscriberOptions`](https://github.com/trufflesuite/truffle/blob/develop/packages/events/EventManager.js#L45-L49) that we can use to update the system. Fix incoming!

eggplantzzz:
[...]

PR #4265, opened 2021-08-13
Fix/quiet option

eggplantzzz:
After looking into https://github.com/trufflesuite/truffle/issues/4237, I discovered that Truffle Config's `with` method wasn't updating the events system properly if an option pertaining to it was passed. This PR fixes the way Truffle Config merges in properties and updates the event system when doing so. 

In addition this PR adds guards into the obtain subscriber so that it won't log (mostly via ora) when `--quiet` is `true`.

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -42,7 +42,7 @@ class TruffleConfig {
   }
 
   private eventManagerOptions(
-    options: TruffleConfig | { quiet?: boolean, logger?: any, subscribers?: any[] }
+    options: TruffleConfig | Pick<TruffleConfig, "quiet" | "logger" | "subscribers">
   ): any {
     const optionsWhitelist = ["quiet", "logger", "subscribers"];
     return pick(options, optionsWhitelist);
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -43,7 +43,7 @@ class TruffleConfig {
 
   private eventManagerOptions(
     options: TruffleConfig | Partial<Pick<TruffleConfig, "quiet" | "logger" | "subscribers">>
-  ): any {
+  ): Partial<TruffleConfig> {
     const optionsWhitelist = ["quiet", "logger", "subscribers"];
     return pick(options, optionsWhitelist);
   }
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -40,9 +40,8 @@ class TruffleConfig {
     );
   }
 
-  public eventManagerOptions(config: TruffleConfig): any {
-    let muteLogging;
-    const { quiet, logger, subscribers } = config;
+  public eventManagerOptions(options: any): any {
+    const { quiet, logger, subscribers } = options;
     return { logger, quiet, subscribers };
   }
 
@@ -102,8 +101,12 @@ class TruffleConfig {
     const current = this.normalize(this);
     const normalized = this.normalize(obj);
 
-    let eventsOptions = this.eventManagerOptions(this);
-    this.events.updateSubscriberOptions(eventsOptions);
+    const currentEventsOptions = this.eventManagerOptions(this);
+    const optionsToMerge = this.eventManagerOptions(obj);
+    this.events.updateSubscriberOptions({
+      ...currentEventsOptions,
+      ...optionsToMerge
+    });
 
     return assignIn(
       Object.create(TruffleConfig.prototype),
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -42,8 +42,8 @@ class TruffleConfig {
   }
 
   private eventManagerOptions(
-    options: TruffleConfig | Partial<Pick<TruffleConfig, "quiet" | "logger" | "subscribers">>
-  ): Partial<TruffleConfig> {
+    options: Partial<TruffleConfig>
+  ): Partial<Pick<TruffleConfig, "quiet" | "logger" | "subscribers">> {
     const optionsWhitelist = ["quiet", "logger", "subscribers"];
     return pick(options, optionsWhitelist);
   }
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -42,7 +42,15 @@ class TruffleConfig {
 
   public eventManagerOptions(options: any): any {
     const { quiet, logger, subscribers } = options;
-    return { logger, quiet, subscribers };
+    const eventManagerOptions: any = { quiet, logger, subscribers };
+    // filter undefined values
+    return Object.keys(eventManagerOptions)
+      .reduce((a: any, key: string) => {
+        if (typeof eventManagerOptions[key] !== "undefined") {
+          a[key] = eventManagerOptions[key];
+        }
+        return a;
+      }, {});
   }
 
   public addProp(propertyName: string, descriptor: any): void {
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -42,7 +42,7 @@ class TruffleConfig {
   }
 
   private eventManagerOptions(
-    options: TruffleConfig | Pick<TruffleConfig, "quiet" | "logger" | "subscribers">
+    options: TruffleConfig | Partial<Pick<TruffleConfig, "quiet" | "logger" | "subscribers">>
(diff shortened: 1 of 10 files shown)


packages/config/src/index.ts
```diff
@@ -40,10 +40,17 @@ class TruffleConfig {
     );
   }
 
-  public eventManagerOptions(config: TruffleConfig): any {
-    let muteLogging;
-    const { quiet, logger, subscribers } = config;
-    return { logger, quiet, subscribers };
+  public eventManagerOptions(options: any): any {
+    const { quiet, logger, subscribers } = options;
+    const eventManagerOptions: any = { quiet, logger, subscribers };
```
haltman-at, line 45: This should probably get a real type rather than being `any`.

packages/events/defaultSubscribers/obtain.js
```diff
@@ -9,12 +9,14 @@ module.exports = {
   handlers: {
     "obtain:start": [
       function() {
+        if (this.quiet) return;
```
haltman-at, line 12: Aha, good find!  (Same comment on the rest of these. :) )

packages/config/src/index.ts
```diff
@@ -40,10 +40,17 @@ class TruffleConfig {
...
+    // filter undefined values
+    return Object.keys(eventManagerOptions)
```
haltman-at, line 47: I see what this loop is doing, but the way it goes about it is fairly awkward.  It might be simpler to just mutate `eventManagerOptions`.  Like, for instance,

```javascript
for (let key in eventManagerOptions) {
  if (eventManagerOptions[key] === undefined) {
    delete eventManagerOptions[key];
  }
}
return eventManagerOptions;
```

I think something like that would be easier to read.  Also you don't need to use `typeof` just to check for undefined. :)

haltman-at (commented):
Thanks for catching this Tyler!  Just a few comments.

packages/events/defaultSubscribers/obtain.js
eggplantzzz: It would be nice to find a way to not have to write this guard every time. Maybe we could do something like create a base class that wraps all handler methods or something, any ideas?

packages/config/src/index.ts
eggplantzzz: Ok I'll go ahead and make this easier to read. I wanted to find a more functional way to do this.

eggplantzzz: So I ended up using `_.pick` for this method because it more cleanly does exactly what I'm trying to do here :)

packages/events/defaultSubscribers/obtain.js
haltman-at: Huh, maybe?  I don't really have a good idea.

packages/config/src/index.ts
haltman-at: Oh, wonderful!

packages/config/src/index.ts
```diff
@@ -40,10 +41,11 @@ class TruffleConfig {
     );
   }
 
-  public eventManagerOptions(config: TruffleConfig): any {
-    let muteLogging;
-    const { quiet, logger, subscribers } = config;
-    return { logger, quiet, subscribers };
+  private eventManagerOptions(
+    options: TruffleConfig | { quiet?: boolean, logger?: any, subscribers?: any[] }
+  ): any {
```
haltman-at, line 46: This return type should I think get a real type rather than being `any`, I think.  The uses of `any` for `logger` and `subscribers` I think are fine.  But the return type shouldn't just be `any`.  (You'll likely want to make an actual type definition here.)

(There's actually probably a better way to type this that @gnidan might know, but I don't care enough to be that picky. :P )

haltman-at (commented):
Cool, just one remaining `any` I think needs to change.

packages/events/defaultSubscribers/obtain.js
cds-amal: That's an interesting approach @eggplantzzz!

cds-amal (changes requested):
Can a test be added to verify that quiet works?

packages/config/src/index.ts
haltman-at: I guess the fancy way to do this would be with the `Pick` type: https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

But as mentioned I won't insist on that :P

eggplantzzz: Oh yeah that is probably the way to do it!

packages/events/defaultSubscribers/obtain.js
eggplantzzz: It's a bit tricky because it is not the case that you want to prevent *every* handler from executing when `quiet` is on. Maybe there is a way to designate them to be silent when `quiet` is on...

packages/config/src/index.ts
eggplantzzz: Although, is there a way to pick where all properties are optional? Would that be
```
Partial<Pick<TruffleConfig, "quiet" | "logger" | "subscribers">>
```?

**Comment by eggplantzzz** [comment]
So I added a couple of tests for `truffle obtain` and also added quiet to the list of "global options".

File: packages/events/defaultSubscribers/obtain.js (hunk 2 above)
Comment (cds-amal) [reply packages/events/defaultSubscribers/obtain.js:12]: Depending on how it's set up, there could be a base implementation that short circuits under certain default conditions that is overridable for those that want to control the short circuit logic

File: packages/truffle/test/scenarios/commands/obtain.js [hunk 5]
```diff
@@ -0,0 +1,62 @@
...
+const Config = require("@truffle/config");
+const fse = require("fs-extra");
+const path = require("path");
+const assert = require("assert");
+const sandbox = require("../sandbox");
+
+let logger, config, project, expectedPath;
+
+describe("truffle obtain", function () {
+  project = path.join(__dirname, "../../sources/develop");
```
cds-amal, line 12: Obtain should use its own test fixture, I think.

cds-amal (changes requested):
This looks great. I have one small nit to break the dependency on another develop's fixture.

packages/truffle/test/scenarios/commands/obtain.js
eggplantzzz: yep, I forgot to add one for it

haltman-at (approved):
Cool, seeems good to me!  Thanks Tyler!

packages/config/src/index.ts
eggplantzzz: I thought that would be the return type as well but TS didn't like it for some reason. So I just left it as `Partial<TruffleConfig>`.

packages/config/src/index.ts
```diff
@@ -43,7 +43,7 @@ class TruffleConfig {
 
   private eventManagerOptions(
     options: TruffleConfig | Partial<Pick<TruffleConfig, "quiet" | "logger" | "subscribers">>
-  ): any {
+  ): Partial<TruffleConfig> {
```
haltman-at, line 46: Sorry, I know I already hit approve, but something seems to have gotten mixed up here, and apparently I missed it.

When I was talking about using `Pick`, I was talking about the *return* type (that being the type that needed to change, since previously it was `any`).  Using `Partial<TruffleConfig>` seems pretty underspecified.  I don't know why you would use `Pick` for the input type; that seems needlessly overcomplicated.  Using `TruffleConfig` seems fine for the input type; I had no objection to that.  It's the return type that was a problem, because it was `any`, and that's where I think the `Pick` belongs.

haltman-at (commented):
Sorry, I know I already hit approve, but it seems I skimmed a bit too quick and missed something...

packages/config/src/index.ts
eggplantzzz: Yeah, I wrestled with the TS compiler a while but couldn't get it to like the return type. Maybe it will like the input type being `Partial<TruffleConfig>` and the return type to be the one specified for the arguments.

eggplantzzz: Updated!

cds-amal (approved):
I likes! Thanks @eggplantzzz

Issue #4260, opened 2021-08-12
Dependency cycle detected on bootstrap (db-loader / db / workflow-compile)


haltman-at:
## Issue

Lerna is picking up a dependency cycle when bootstrapping.

```
lerna WARN ECYCLE Dependency cycles detected, you should fix these!
lerna WARN ECYCLE @truffle/db-loader -> @truffle/db -> @truffle/workflow-compile -> @truffle/db-loader
```

## Environment

* Truffle version (`truffle version`): 5.4.6

gnidan:
Tested that this doesn't show up anymore.

PR #4269, opened 2021-08-16
Use @truffle/resolver as devDep for @truffle/db, instead of @truffle/workflow-compile

gnidan:
Since @truffle/db uses @truffle/resolver but does not use @truffle/workflow-compile, and the latter causes a dependency cycle.

May address #4260


haltman-at (approved):
A quick search of the code confirms the situation.  Thanks Nick!

PR #4270, opened 2021-08-17
Internal improvement: Use import type in codec wherever possible

haltman-at:
This PR changes all `import`s in codec to `import type` wherever possible.  Completes what was started in #4110.

Note I didn't actually try breaking down imports into smaller parts; I only changed imports that can be changed as a whole.  Breaking down imports into smaller pieces is perhaps something that can be done later.

#4142 may have to be updated for this once this is merged.


cds-amal (commented):
This looks good to me, though I might be missing something. What criteria did you use for "wherever possible" Harry?

haltman-at:
I mean, for each import, I checked whether it was only used for types, or if it was also used for values.  If I couldn't find any values it was used for, I changed it to `import type`.  Sometimes I mistakenly marked something `import type` that wasn't, and then I got a compile error, so I changed it back.

cds-amal:
Thanks for the explanation!

cds-amal (approved):
Thanks @haltman-at, let's get this in!

PR #4273, opened 2021-08-19
Revert "Revert "Upgrade @ethereumjs/common to ^2.4.0 and @ethereumjs/tx to ^3.3.0""

eggplantzzz:
Do not merge this PR until October when support for Node.js version 10 is discontinued. 

This PR updates @ethereumjs/common and @ethereumjs/tx in @truffle/hdwallet-provider.

Reverts trufflesuite/truffle#4271

PR #4276, opened 2021-08-23
support eip-1559

XWJACK:


eggplantzzz:
Thanks for the PR @XWJACK! We probably have to wait until Ganache fully supports 1559 before releasing this, but we also have to wait until October as well. That is because the updated versions of the @ethereumjs libraries (made in https://github.com/trufflesuite/truffle/pull/4273) aren't compatible with Node 10 as they require at least 12-something. Truffle will stop supporting Node 10 in October.

packages/hdwallet-provider/src/index.ts
```diff
@@ -170,7 +170,13 @@ class HDWalletProvider {
               )
             };
           }
-          const tx = Transaction.fromTxData(txParams, txOptions);
+
+          // Taken from https://github.com/ethers-io/ethers.js/blob/2a7ce0e72a1e0c9469e10392b0329e75e341cf18/packages/abstract-signer/src.ts/index.ts#L215
+          const hasEip1559 = (txParams.maxFeePerGas !== undefined || txParams.maxPriorityFeePerGas !== undefined);
```
eggplantzzz, line 175: Maybe we should also ensure that `type: "0x02"` is in the tx params as well if we detect that it is 1559. In other words maybe we should merge that param in after we find out it is a 1559 tx.

XWJACK: Because new transactions must have `maxFeePerGas` and `maxPriorityFeePerGas` parameters.  And the new transaction type will be automatically set(replaced) by `FeeMarketEIP1559Transaction`. So we don’t need to set `type: "0x02"`

https://github.com/ethereumjs/ethereumjs-monorepo/blob/ae9989c6efd5d7916938351ae0f065b9fc9adb42/packages/tx/src/eip1559Transaction.ts#L181-L182

eggplantzzz: Ah ok cool, in the README examples they set the type manually. Looks good!

eggplantzzz:
I just published @truffle/hdwallet-provider@eip1559-beta. This dist tag release includes the work in this PR as well as the work in https://github.com/trufflesuite/truffle/pull/4273.

If you get a chance to try this and have issues then let us know!

eggplantzzz:
So I rebased #4273 and we're going to merge it soon. @XWJACK any chance you can resolve the conflict? I'm going to try and make sure hdwallet-provider is 1559 compliant over the next few weeks. I think there may have been some more work to do on this front?

cds-amal (approved):
LGTM! Thanks @XWJACK, @eggplantzzz !

eggplantzzz:
This will need a version bump once we get the changes to @trufflesuite/we3-provider-engine released.

eggplantzzz:
Thanks again for this @XWJACK!

PR #4282, opened 2021-08-26
Dependency update: Bump mocha to ^9.1.0

eggplantzzz:


eggplantzzz:
Closing this. mocha-webpack (which looks no longer maintained) doesn't seem to be compatible with mocha 9. We eventually need to figure out a solution for testing in debugger which doesn't involve mocha-webpack.

PR #4296, opened 2021-09-01
Use gas estimate as default transaction gas limit

fainashalts:
This addresses the remaining part of #3992 

Prior to this PR, Truffle's `getGasEstimate()` function never actually estimated gas. There was a hardcoded default gas value in the configuration defaults that was added to the params before we hit the gas estimation function. The gas estimation function would check for gas in the params before estimating, and that value was never empty. 

A workaround was to add `gas: 0` to the config, which would force the estimation function to do the estimation. This was not ideal as it is not clear what is actually happening and why. This fix removes the default gas value to ensure that gas is estimated if the user does not provide their own gas in the transaction params.

*Note: user-provided gas params are still respected, and gas estimation will not occur if a non-zero integer is provided as the gas from the user's code*

*Note: gas estimation for transactions that will revert does not error gracefully, which in general Truffle's code is aware of, but we want to be able to continue the user flow in the case of debugging, so in that situation we provide a default gas of 1/2 the blockLimit*

### DIFF
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -78,11 +78,6 @@ export const configProps = ({
   const resolveDirectory = (value: string): string =>
     path.resolve(configObject.working_directory, value);
 
-  const defaultTXValues = {
-    gas: 6721975,
-    from: null
-  };
-
   return {
     // These are already set.
     truffle_directory() {},
@@ -158,7 +153,7 @@ export const configProps = ({
           config = {};
         }
 
-        config = assignIn({}, defaultTXValues, config);
+        config = assignIn({}, config);
 
         return config;
       },
@@ -173,7 +168,7 @@ export const configProps = ({
         try {
           return configObject.network_config.from;
         } catch (e) {
-          return defaultTXValues.from;
+          return null;
         }
       },
       set() {
@@ -187,7 +182,7 @@ export const configProps = ({
         try {
           return configObject.network_config.gas;
         } catch (e) {
-          return defaultTXValues.gas;
+          return null;
         }
       },
       set() {
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -1,4 +1,3 @@
-import assignIn from "lodash.assignin";
 import * as path from "path";
 import Provider from "@truffle/provider";
 import TruffleConfig from "./";
@@ -152,8 +151,6 @@ export const configProps = ({
         if (config === null || config === undefined) {
           config = {};
         }
-
-        config = assignIn({}, config);
 
         return config;
       },
--- a/packages/contract/lib/execute.js
+++ b/packages/contract/lib/execute.js
@@ -13,6 +13,7 @@ const execute = {
   // -----------------------------------  Helpers --------------------------------------------------
   /**
    * Retrieves gas estimate multiplied by the set gas multiplier for a `sendTransaction` call.
+   * Lacking an estimate, sets gas to have of latest blockLimit
    * @param  {Object} params     `sendTransaction` parameters
    * @param  {Number} blockLimit  most recent network block.blockLimit
    * @return {Number}             gas estimate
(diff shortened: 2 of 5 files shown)


packages/interface-adapter/lib/adapter/web3/index.ts
```diff
@@ -49,8 +49,13 @@ export class Web3InterfaceAdapter implements InterfaceAdapter {
     return this.web3.eth.getAccounts();
   }
 
-  public estimateGas(transactionConfig: Transaction) {
-    return this.web3.eth.estimateGas(transactionConfig);
+  public async estimateGas(transactionConfig: Transaction) {
+    try {
```
fainashalts, line 53: `try/catch` is designed for handling exceptions really, so I'm a little wary of using it for control flow. is there a better way to do this? unfortunately because there's no error handling for this case in the gas estimation call I can't just check for the promise resolution.

packages/contract/lib/execute.js
```diff
@@ -22,20 +23,34 @@ const execute = {
...
-          );
-
-          // Don't go over blockLimit
+          // there are situations where the web3 gas estimation function in interfaceAdapter
+          // fails, specifically when a transaction will revert; we still want to continue
+          // the user flow for debugging purposes; so we provide a default gas for
+          // that situation, equal to half of the blockLimit for the latest block
           const limit = utils.bigNumberify(blockLimit);
+          let bestEstimate;
+          if(gas === null) {
```
eggplantzzz, line 40: stupid nit but you are missing a space after `if`

packages/interface-adapter/lib/adapter/web3/index.ts
```diff
@@ -49,8 +49,13 @@ export class Web3InterfaceAdapter implements InterfaceAdapter {
...
+      const gasEstimate = await this.web3.eth.estimateGas(transactionConfig);
+      return gasEstimate;
+    } catch {
```
eggplantzzz, line 56: this `catch` should probably test to ensure that the `try` failed in the way your are expecting it to fail and only if it does return `null` - maybe the question is, "can `web3.eth.estimateGas` fail in a different way in which we would want to throw here?

fainashalts: I made some changes to this because `contract-tests` do expect the VM error that throws here. So now we only catch the error if the user is debugging, so they can get past this point and debug to see why their code is reverting.

packages/contract/lib/execute.js
fainashalts: Fixed! :)

packages/interface-adapter/lib/adapter/web3/index.ts
```diff
@@ -49,8 +49,20 @@ export class Web3InterfaceAdapter implements InterfaceAdapter {
     return this.web3.eth.getAccounts();
   }
 
-  public estimateGas(transactionConfig: Transaction) {
-    return this.web3.eth.estimateGas(transactionConfig);
+  public async estimateGas(transactionConfig: Transaction, stacktrace = false) {
+    // web3 does not error gracefully when gas estimation fails due to a revert,
+    // so in cases where we want to get past this (debugging/stacktracing), we must
+    // catch the error and return null instead
+    if(stacktrace === true) {
```
eggplantzzz, line 56: haha sorry, same nit about lacking a space

packages/interface-adapter/lib/adapter/web3/index.ts
```diff
@@ -49,8 +49,20 @@ export class Web3InterfaceAdapter implements InterfaceAdapter {
...
+      try {
+        const gasEstimate = await this.web3.eth.estimateGas(transactionConfig);
+        return gasEstimate;
+      } catch {
```
eggplantzzz, line 60: I have the same comment here, we should probably ensure that this is failing in whatever way `estimateGas` fails when it the tx always reverts. If it fails for another reason we should throw.

packages/interface-adapter/lib/adapter/web3/index.ts
fainashalts: no prob! thanks for pointing it out! fixed.

packages/interface-adapter/lib/adapter/web3/index.ts
fainashalts: In general I agree with this sentiment, but this is a very odd situation as far as how gas estimation errors are handled. It does not appear that geth (and thus web3) provides information that can properly allow us to distinguish between different error cases when calling `eth_estimateGas`. That is the reason for the try/catch, because we want to be able to skip the error here and get to a more informative error for someone attempting to debug. Does that make sense?

I think it would be great to have better error handling on the geth side, that we could then consume. I will be writing up an issue in that repo soon to that effect in hopes a change can be made (though a similar problem was brought up 4 years ago to no avail -- https://github.com/ethereum/go-ethereum/issues/15671).

eggplantzzz: So I'm understanding that you are saying that when `estimateGas` fails for any reason, we cannot know right away why it is failing? It might be failing because the tx always reverts or it might be failing because the input is improperly formatted, we have no way of telling for sure why it is failing?

Because if, for example, the input is improper - then we want to throw an error here for sure and not tell the user something like "your tx always reverts". That would be very unhelpful. But if we can't tell for sure why and we just have to display a more general error message, then I guess that's what we have to do. :)

packages/config/src/configDefaults.ts
```diff
@@ -158,7 +153,7 @@ export const configProps = ({
           config = {};
         }
 
-        config = assignIn({}, defaultTXValues, config);
+        config = assignIn({}, config);
```
eggplantzzz, line 156: Do we even need this `assignIn` now that we're not really merging anything?

packages/interface-adapter/lib/adapter/web3/index.ts
```diff
@@ -49,8 +49,20 @@ export class Web3InterfaceAdapter implements InterfaceAdapter {
...
+    // catch the error and return null instead
+    if (stacktrace === true) {
+      try {
+        const gasEstimate = await this.web3.eth.estimateGas(transactionConfig);
+        return gasEstimate;
+      } catch {
+        return null;
+      }
+    } else {
+      return this.web3.eth.estimateGas(transactionConfig);
```
eggplantzzz, line 64: So when estimate gas fails, we only want to throw in the case that `stacktrace === false` and return `null` otherwise?

packages/interface-adapter/lib/adapter/web3/index.ts
fainashalts: Yeah unfortunately we won't know. I am more ok with this because we will still error here in almost all situations, basically unless someone is debugging.

packages/interface-adapter/lib/adapter/web3/index.ts
fainashalts: Yep!

packages/config/src/configDefaults.ts
fainashalts: hm. I guess we don't actually need this line anymore. I can remove it.

packages/interface-adapter/lib/adapter/web3/index.ts
fainashalts: There is even a PR that got closed without being merged that doesn't directly address this but does mention that error handling for this call isn't great: https://github.com/ethereum/EIPs/pull/2724

I will write up an issue in the geth repo, but my sense is that the best way to advocate there for this is to ask for a separate endpoint, `eth_estimateGasVerbose` or something. I'll be trying to advocate there but in the meantime I'm afraid this is what we have to work with.

PR #4299, opened 2021-09-07
Replace configstore with conf

eggplantzzz:
This PR replaces configstore with the conf package. Alongside is also a migration script that will be used to migrate data from the old configstore location to the new one. The new package uses 3 different (OS-appropriate) locations depending on which OS the user has. configstore used only 1 location: `~/.config`.

After a discussion with @cds-amal we decided it would be good to make a lot of noise about this change so that users will be very aware that this directory has changed. It is possible that people are using this directory if they use @truffle/config which has the utility method `getTruffleDataDirectory` which returns the location used by Truffle for configuration, the compiler cache, and the TruffleDB `.db` directory.

~TODO: The migration utility methods still need to be added to each command flow. This will mean that every time a user runs a command, we will check whether a migration is necessary and if so, perform it.~


packages/core/lib/command.js
```diff
@@ -76,6 +76,12 @@ class Command {
   async run(inputStrings, options) {
     const result = this.getCommand(inputStrings, options.noAliases);
 
+    try {
```
eggplantzzz, line 79: I built this to fail silently but perhaps we should consider adding some sort of a warning message here so that if there are issues people can report them?

fainashalts: I think it would be nice to console out a message to the effect of "unable to enable user-level config file. please let the Truffle team know" or something?

fainashalts: definitely not a dealbreaker either way though

cds-amal: Can you add a `debug` module output at least? It would be useful to turn on to track issues?

packages/core/lib/config-migration.js
```diff
@@ -0,0 +1,47 @@
...
+const copy = require("./copy");
+const TruffleConfig = require("@truffle/config");
+
+module.exports = {
+  oldTruffleDataDirectory: path.join(OS.homedir(), ".config", "truffle"),
+
+  needsMigrated: function () {
+    const oldConfig = path.join(this.oldTruffleDataDirectory, "config.json");
+    const conf = new Conf({ projectName: "truffle" });
+    return fse.existsSync(oldConfig) && oldConfig !== conf.path;
```
cds-amal, line 15: if adding the debug module, would be nice to log if it needs migration

cds-amal: Should there be another component to this logic? Stuff gets copied the first time, and there's not delete (or rename) of the original so the second time a command runs,
`fse.existsSync(oldConfig) && oldConfig !== conf.path` will be true again? Am I missing something?

gnidan (approved):
looks good to me

packages/core/lib/config-migration.js
```diff
@@ -0,0 +1,51 @@
...
+    const folders = ["compilers", ".db"];
+    for (const folder of folders) {
+      await this.migrateFolder(folder);
+    }
+  },
+
+  migrateGlobalConfig: function () {
+    const conf = new Conf({ projectName: "truffle" });
+    // set this to true so we know not to migrate next time
+    conf.set("migrated", true);
```
cds-amal, line 32: I'm worried that `migrated` can be set and subsequent migration can fail which would leave "config" in an unknown state. Can `migrated` be set iff the migration is successful?

cds-amal (changes requested):

packages/core/lib/config-migration.js
gnidan: hm, so thinking about this more - `migrated` isn't necessary, even, right? cause if it's in the new location, it's migrated; if it's in the old location, it's not. do we need more than that?

eggplantzzz: Ok, so this is true! I thought that the new config was created on instantiation of the new config (`new Conf({ ...`) but it actually doesn't create the file until a field is set. So I think the logic for `needsMigrated` can be simplified too.

gnidan: I think "migration finished" is synonymous with "the file in the new location is a valid JSON file"

cds-amal: @gnidan the operation isn't currently atomic, so there might be an issue if migration doesn't run to successful completion.

cds-amal: @eggplantzzz Perhaps **migrate** is confirmation the migration finished successfully.

gnidan: > @gnidan the operation isn't currently atomic, so there might be an issue if migration doesn't run to successful completion.

Sure, but even if it's not atomic, what we care about is whether it's completed or not - if the file is valid and in a new location, that's enough information to say that migration is complete? Or am I missing something?

cds-amal: My concern is if [this logic fails](https://github.com/trufflesuite/truffle/blob/4930591364f1126a9f9773178e2111d981a76024/packages/core/lib/config-migration.js#L24-L26) . `.db` and or `compilers` could be incomplete, and I'm not sure what a partially copied `.db` means.

eggplantzzz: Yeah so we should figure out what the criteria for this is. And is it guaranteed that you'll even have `config.json`? You may just have `/compilers`. Maybe this is overkill and we should just check if the directory exists (the new directory) and if it is assume everything is migrated.

eggplantzzz:
It was decided that instead of creating a field called "migrated" in the config, we would create a "version" field to allow us to determine if we have migrated or not.

packages/core/lib/config-migration.js
```diff
@@ -8,20 +8,22 @@ const TruffleConfig = require("@truffle/config");
 const debugModule = require("debug");
 const debug = debugModule("core:config-migration");
 
+const CURRENT_CONFIG_VERSION = 1;
```
cds-amal, line 11: Since we have a config version, should we add this output to `truffle version` ?

cds-amal (approved):
Nice @eggplantzzz, I likes.

packages/core/lib/config-migration.js
gnidan: I don't think this is worth doing. we don't have output for contract-schema version, e.g.

PR #4300, opened 2021-09-11
Add support for block.basefee in debugger, and fix a bug with internal function pointers

haltman-at:
This PR does two things, which are basically unrelated, except that (for reasons you'll see) I had to do the second to get the tests passing for the first.

The main thing this PR does is to add support in the debugger for Solidity's `block.basefee` builtin, introduced in Solidity 0.8.7, because I forgot to do this when Solidity 0.8.7 came out.  Note that this builtin will only be displayed on Solidity versions 0.8.7 or later; on older versions it won't show up.  This involved adding a new value to the `SolidityFamily` type in `codec`.

Speaking of which, I got tired of updating all the `case` statements there, so I updated most of them to only list out earlier versions of Solidity that behaved differently, and have all versions with the current behavior just be `default`.  That way most of these just won't need to be updated in the future anymore.  I did leave one of them explicit because I expect it might still change in the future.

Note that if you have a transaction in a pre-London block, and its code was nonetheless compiled with Solidity 0.8.7 or later, then we'll show `block.basefee` as being zero.  Its "actual" value is, well, not defined; I figure 0 is an OK substitute.  Better perhaps would be to exclude it to display, but that would require a new mechanism to do that, and I didn't want to bother with that.

Testing this change was a bit of a pain -- since stable Ganache doesn't support London yet, I had to do my real tests manually.  But I did update the globally-available tests to keep them from breaking, even if "updating" them here just meant checking that basefee comes out as 0 in a pre-London block, rather than checking that it comes out as the correct value in a post-London block.

However, in doing this update, I encountered a problem: Our tests were on Solidity version 0.8.4... and upgrading them to 0.8.5 or later broke the tests that involved uninitialized internal function pointers.  Turns out, they changed the way those work *again* in Solidity 0.8.5.  (I'm going to have to update "Data Representation in Solidity" again!  I didn't know about that!)

With one of the broken tests, the debugger wasn't actually doing anything wrong, I just had to update the tests.  But for the other two, I had to update `source-map-utils` so that the new style (or, I guess, the *new* new style) of uninitialized internal function pointers would be recognized as such by the debugger.

Basically, the Solidity 0.8.4 way was to have them point to the `panic_error_0x51` function (although at the time it wasn't named).  Now instead they point to a function that just jumps *to* the `panic_error_0x51` function.  So we have to detect that, and I updated `source-map-utils` to do so.

Actually, hm, wait, I just realized... currently we detect whether a function is `panic_error_0x51` by checking its bytecode.  However, I could also add a check to check its AST `nodeType` and name too, since we have that information.  We wouldn't want to rely solely on that, of course, but...?  Well, I dunno.  I'm going to put this up for review, and if you think that should be changed, let me know!  (Or maybe I'll just go back and make the change anyway. :P )

### DIFF
--- a/packages/codec/lib/compiler/types.ts
+++ b/packages/codec/lib/compiler/types.ts
@@ -7,5 +7,5 @@ export interface CompilerVersion {
 
 //NOTE: Families 0.5.0 and up will be named by the lowest version that
 //fits in the given family.  So e.g. 0.5.x covers 0.5.x-0.7.x;
-//0.8.x covers 0.8.x-current (as I write this)
-export type SolidityFamily = "unknown" | "pre-0.5.0" | "0.5.x" | "0.8.x";
+//0.8.x covers 0.8.0-0.8.6; 0.8.7+ covers 0.8.7 to current (as I write this)
+export type SolidityFamily = "unknown" | "pre-0.5.0" | "0.5.x" | "0.8.x" | "0.8.7+";
--- a/packages/codec/lib/ast/import/index.ts
+++ b/packages/codec/lib/ast/import/index.ts
@@ -47,8 +47,7 @@ export function definitionToType(
             kind: "general",
             typeHint
           };
-        case "0.5.x":
-        case "0.8.x":
+        default:
           return {
             typeClass,
             kind: "specific",
--- a/packages/codec/lib/compiler/utils.ts
+++ b/packages/codec/lib/compiler/utils.ts
@@ -9,11 +9,17 @@ export function solidityFamily(compiler: CompilerVersion): SolidityFamily {
     return "unknown";
   }
   if (
+    semver.satisfies(compiler.version, ">=0.8.7", {
+      includePrerelease: true
+    })
+  ) {
+    return "0.8.7+";
+  } else if (
+    //see comment below about the weird-looking condition
     semver.satisfies(compiler.version, "~0.8 || >=0.8.0", {
       includePrerelease: true
     })
   ) {
-    //see comment below about the weird-looking condition
     return "0.8.x";
   } else if (
     semver.satisfies(compiler.version, "~0.5 || >=0.5.0", {
--- a/packages/codec/lib/special/decode/index.ts
+++ b/packages/codec/lib/special/decode/index.ts
@@ -109,6 +109,9 @@ export function* decodeMagic(
       if (solidityVersionHasChainId(info.currentContext.compiler)) {
         variables.push("chainid");
       }
+      if (solidityVersionHasBaseFee(info.currentContext.compiler)) {
+        variables.push("basefee");
+      }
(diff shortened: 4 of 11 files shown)


haltman-at:
Also, a decoder test kept timing out for no apparent reason, so I've upped the timeout on that.

gnidan (approved):
This looks good. I think it's fine to leave it as basefee of `0` for pre-london; not worth the effort to filter it out, I agree.

As far as the bytecode match goes, sure if you'd like, update it to check the `nodeType`!

packages/codec/lib/compiler/types.ts
```diff
@@ -7,5 +7,5 @@ export interface CompilerVersion {
 
 //NOTE: Families 0.5.0 and up will be named by the lowest version that
 //fits in the given family.  So e.g. 0.5.x covers 0.5.x-0.7.x;
-//0.8.x covers 0.8.x-current (as I write this)
-export type SolidityFamily = "unknown" | "pre-0.5.0" | "0.5.x" | "0.8.x";
+//0.8.x covers 0.8.0-0.8.6; 0.8.7+ covers 0.8.7 to current (as I write this)
+export type SolidityFamily = "unknown" | "pre-0.5.0" | "0.5.x" | "0.8.x" | "0.8.7+";
```
cds-amal, line 11: `0.8.7+` seems like an odd semver, unless it's our convention?

cds-amal (approved):
LGTM. Can't think of a negative scenario using 0 as default.

haltman-at:
Btw, I filed an [issue with web3](https://github.com/ChainSafe/web3.js/issues/4326) about the fact that base fees are coming back as hex strings.  Once that's fixed, and we upgrade to a web3 that fixes it, we can get rid of the `parseInt` there.

packages/codec/lib/compiler/types.ts
haltman-at: It's not any sort of convention; it's not even really semver.  If you look at this list of "solidity families", you'll see that the whole thing is pretty ad-hoc. :shrug:  I can change it if you think that would be helpful, since obviously this is all prety arbitrary.

cds-amal: No, it's fine, there's no need to change it. The code comment now makes sense! It was surprising to see that's all.

haltman-at:
I've added another commit to this PR with the following changes:

1. I fixed two bugs (I had `return false;` rather than `return null;` at one point, and I didn't check that there were *any* pushes in the push sequence.)
2. (the main one) I've made the change to check for the name `panic_error_0x51` that I mentioned above
3. I changed some `let`s to `const`s because it was bothering me :P

There is a little bit of copypaste in this commit, but I think it's fine.

gnidan (approved):
Looks good!

PR #4309, opened 2021-09-16
Internal improvement: Pull Profiler out into its own package

eggplantzzz:
This PR strips the Profiler out of @truffle/compile-common and puts it into its own package. 

Welcome to the newest Truffle package @truffle/profiler!


haltman-at (approved):
Looks good.  I noticed some irregularities in the `package.json` files, but on inspection, I see that those were already present in the existing one.  So probably better to just handle those in a separate PR; I'm going to go ahead and hit approve on this one.

(If you're wondering what they are and want to get them anyway, it's that I noticed that the `homepage`, `bugs`, and `publishConfig` fields are missing.  Adding a `files` field would probably also be a good idea?)

haltman-at:
Actually, one other question: Should profiler be a prerelease version in this PR, so it can be bumped up to a release version when it's published?  Like right now it's 0.1.0, which I assume means it would be bumped up to 0.1.1 when it's released? (@gnidan?)

haltman-at (approved):
Seems good!

gnidan:
(Yeah, for new packages I'll do `-0` just to avoid the `0.1.1` start)

PR #4312, opened 2021-09-17
Bug fix: Fix ethers-compatible nativization of empty bytestring

haltman-at:
So, the new `nativize` function in codec (as opposed to the old unsafe one) is supposed to match ethers's behavior, right?  Well, um, turns out ethers has a special behavior for the empty bytestring that I didn't realize... instead of being nativized as `"0x"`, it's rendered as `null`!  This is seems to me to be pretty dumb, but, well, the point is to be compatible, so, uh, yeah, I've changed it. :-/  (Fortunately this isn't yet actually used anywhere...)

I didn't bother adding a test of this, but I could if people think it's necessary.

### DIFF
--- a/packages/codec/lib/export.ts
+++ b/packages/codec/lib/export.ts
@@ -119,7 +119,8 @@ function ethersCompatibleNativize(
         case "bool":
           return (<Format.Values.BoolValue>result).value.asBoolean;
         case "bytes":
-          return (<Format.Values.BytesValue>result).value.asHex;
+          const asHex = (<Format.Values.BytesValue>result).value.asHex;
+          return asHex !== "0x" ? asHex : null;
         case "address":
           return (<Format.Values.AddressValue>result).value.asAddress;
         case "contract":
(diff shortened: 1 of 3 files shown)


cds-amal (commented):
I'd like to a test for this if we have to maintain Ethers compatibility. Do we have tests that could answer that question?

gnidan (approved):

haltman-at:
> I'd like to a test for this if we have to maintain Ethers compatibility. Do we have tests that could answer that question?

Good question.  Currently, the tests for this functionality are [here](https://github.com/trufflesuite/truffle/blob/develop/packages/decoder/test/current/test/compatible-nativize.js).  However you'll notice they don't actually test against Ethers, but rather against fixed outputs; actually testing against Ethers would be a bit more trouble.

Since @gnidan has approved this, I think I'm just going to go ahead and merge this, and we can maybe handle the testing thing separately later?

haltman-at:
...OK, I guess I'm not going ahead and just merging this, then.  @cds-amal, do you think it suffices to extend the existing tests in a straightforward way?  Like I said, actually testing against Ethers would be more trouble, and perhaps is better saved for later?

cds-amal:
Sure thing @haltman-at, that suffices to cover this change and revisit with follow up work.

haltman-at:
OK @cds-amal, I've added that test.

cds-amal (approved):
Thanks @haltman-at, looks good!

PR #4330, opened 2021-09-27
Dependency update: Update highlightjs-solidity to 2.0.1

haltman-at:
A pretty small part of the 0.8.8 updates, but an easy one. :)  Updates our syntax highlighting for 0.8.8. :)  (Of course, it won't do much good until I actually make the debugger *support* those new parts of 0.8.8 that need highlighting, but, whatever, I did this first anyway.)


cds-amal (approved):
LGTM!

PR #4331, opened 2021-09-27
Internal improvement: Decoder surgery

eggplantzzz:
This PR aims to simplify the dependencies of the decoder by pulling out most of web3 and just using web3-eth.

### DIFF
--- a/packages/decoder/lib/Web3Adapter.ts
+++ b/packages/decoder/lib/Web3Adapter.ts
@@ -1,0 +1,21 @@
+import eth from "web3-eth";
+
+export class Web3Adapter {
+  public provider: any;
+
+  constructor (provider: any) {
+    this.provider = provider;
+  }
+
+  async getCode (address: string, block: string) {
+    return await eth.getCode(address, block);
+  }
+
+  async getBlock (block: string) {
+    return await eth.getBlock(block);
+  }
+
+  async getPastLogs (address: string, from: string, to: string) {
+    return await eth.getPastLogs(address, from, to);
+  }
+}
--- a/packages/decoder/lib/Web3Adapter.ts
+++ b/packages/decoder/lib/Web3Adapter.ts
@@ -25,4 +25,16 @@ export class Web3Adapter {
   public async getPastLogs ({ address, fromBlock, toBlock }: PastLogsOptions): Promise<any> {
     return await this.Eth.getPastLogs({ address, fromBlock, toBlock });
   }
+
+  public async getNetworkId (): Promise<string> {
+    return (await this.Eth.net.getId()).toString();
+  }
+
+  public async getBlockNumber (): Promise<number> {
+    return await this.Eth.getBlockNumber();
+  }
+
+  public async getBalance (address: string, blockNumber: number) {
+    return await this.Eth.getBalance(address, blockNumber);
+  }
 }
--- a/packages/decoder/lib/Web3Adapter.ts
+++ b/packages/decoder/lib/Web3Adapter.ts
@@ -1,21 +1,28 @@
-import eth from "web3-eth";
+import { Eth } from "web3-eth";
+type PastLogsOptions = {
+  toBlock?: string | number;
+  fromBlock?: string | number;
+  address?: string | string[];
+}
 
 export class Web3Adapter {
   public provider: any;
+  public Eth: Eth;
 
   constructor (provider: any) {
     this.provider = provider;
+    this.Eth = new Eth(provider);
   }
 
-  async getCode (address: string, block: string) {
-    return await eth.getCode(address, block);
+  public async getCode (address: string, block: string | number) {
+    return await this.Eth.getCode(address, block);
   }
 
-  async getBlock (block: string) {
-    return await eth.getBlock(block);
+  public async getBlock (block: string | number) {
+    return await this.Eth.getBlock(block);
   }
 
-  async getPastLogs (address: string, from: string, to: string) {
-    return await eth.getPastLogs(address, from, to);
+  public async getPastLogs ({ address, fromBlock, toBlock }: PastLogsOptions): Promise<any> {
+    return await this.Eth.getPastLogs({ address, fromBlock, toBlock });
   }
 }
--- a/packages/decoder/lib/Web3Adapter.ts
+++ b/packages/decoder/lib/Web3Adapter.ts
@@ -4,6 +4,7 @@ type PastLogsOptions = {
   fromBlock?: string | number;
   address?: string | string[];
 }
+import type BN from "bn.js";
 
 export class Web3Adapter {
   public provider: any;
@@ -40,5 +41,9 @@ export class Web3Adapter {
 
   public async getTransactionCount (contractAddress: string, blockNumber: any) {
     return this.Eth.getTransactionCount(contractAddress, blockNumber);
+  }
+
+  public async getStorageAt (address: string, position: BN, block: string | number) {
+    return this.Eth.getStorageAt(address, position, block);
   }
 }
--- a/packages/decoder/lib/Web3Adapter.ts
+++ b/packages/decoder/lib/Web3Adapter.ts
@@ -34,7 +34,11 @@ export class Web3Adapter {
     return await this.Eth.getBlockNumber();
   }
 
-  public async getBalance (address: string, blockNumber: number) {
+  public async getBalance (address: string, blockNumber: any) {
     return await this.Eth.getBalance(address, blockNumber);
+  }
+
+  public async getTransactionCount (contractAddress: string, blockNumber: any) {
+    return this.Eth.getTransactionCount(contractAddress, blockNumber);
   }
 }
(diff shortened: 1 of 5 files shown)


packages/decoder/lib/decoders.ts
```diff
@@ -45,7 +46,7 @@ const SourceMapUtils = require("@truffle/source-map-utils");
  * @category Decoder
  */
 export class ProjectDecoder {
-  private web3: Web3;
+  private web3Adapter: Web3Adapter;
```
eggplantzzz, line 49: Does this need to be public?

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
+import Web3Eth from "web3-eth";
+type PastLogsOptions = {
```
eggplantzzz, line 2: I lifted most of this type from the web3 types, I don't know if we can get rid of the optional arguments as I don't know whether the decoder always includes these or not.

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+  toBlock?: string | number;
+  fromBlock?: string | number;
+  address?: string | string[];
+}
```
gnidan, line 6: This should probably not live amongst the imports; if it's not getting exported, maybe move it to the end of the file?

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+import type BN from "bn.js";
+
+export class Web3Adapter {
+  public provider: any;
+  private eth: any;
```
gnidan, line 11: Surely we can do better than `any` here

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  constructor (provider: any) {
```
gnidan, line 13: (and here)

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+    this.provider = provider;
+    // @ts-ignore - seems like the types are out of sync with the implementation
```
gnidan, line 15: you have an `any` **and** a `ts-ignore`!? happy to take a look with you if you had trouble getting better types working

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
+import Web3Eth from "web3-eth";
+type PastLogsOptions = {
+  toBlock?: string | number;
+  fromBlock?: string | number;
+  address?: string | string[];
+}
+import type BN from "bn.js";
+
+export class Web3Adapter {
+  public provider: any;
```
gnidan, line 10: Something I just realized - earlier today in Slack, @eggplantzzz, you said that direct use of `provider` requires that we separately handle websockets. Why is that? Wouldn't `.send()`, et al. work the same, regardless of http vs. ws? @davidmurdoch I guess you were also part of the discussion? I feel like I missed an important detail there!

packages/decoder/lib/decoders.ts
gnidan: Doesn't seem like it - why might it need to be public?

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  public async getNetworkId () {
+    return (await this.eth.net.getId()).toString();
+  }
+
+  public async getBlockNumber (): Promise<number> {
+    return await this.eth.getBlockNumber();
+  }
+
+  public async getBalance (address: string, blockNumber: any) {
```
gnidan, line 39: I get why this `any` might be useful, but I don't like it here in this internal adapter. It's pretty bad if the decoder and this adapter can't agree on a type more well-defined than `any`!

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+    return await this.eth.getBalance(address, blockNumber);
+  }
+
+  public async getTransactionCount (contractAddress: string, blockNumber: any) {
```
gnidan, line 43: Same with this `any`

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+    return this.eth.getTransactionCount(contractAddress, blockNumber);
+  }
+
+  public async getStorageAt (address: string, position: BN, block: string | number) {
```
gnidan, line 47: What's this return?

packages/decoder/lib/Web3Adapter.ts
gnidan: Also, what's the return value?

packages/decoder/lib/Web3Adapter.ts
gnidan: (also what's the return value?)

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  public async getCode (address: string, block: string | number) {
+    return await this.eth.getCode(address, block);
+  }
+
+  public async getBlock (block: string | number) {
+    return await this.eth.getBlock(block);
+  }
+
+  public async getPastLogs ({ address, fromBlock, toBlock }: PastLogsOptions): Promise<any[]> {
```
gnidan, line 27: Can we ditch this `any`? :)

gnidan (changes requested):
Mostly just here to request the elimination of `any`s.

Also, in some places I'm suggesting explicit return values - @haltman-at might come in and take a hard line to require them for all methods, but I think we need good return types at least for the methods that have complex return values. (i.e. it's probably fine to let it slide for `Promise<string>`, etc., since text editors take the sting out of the unknown there)

packages/decoder/lib/decoders.ts
gnidan: Doesn't seem like it - why might it need to be public?

packages/decoder/lib/Web3Adapter.ts
davidmurdoch: I was referring to the case of implementing your own internal provider (like using `axios` to send rewuests) you'd need to also handle the websocket case.

packages/decoder/lib/decoders.ts
haltman-at: Yeah there's already a method to get it...

packages/decoder/lib/Web3Adapter.ts
```diff
@@ -0,0 +1,50 @@
+import Web3Eth from "web3-eth";
```
haltman-at, line 1: So, wait, we're not using web3, but we *are* using web3-eth?  Does that mean the problematic parts of web3 are the *rest* of web3, and web3-eth is fine, and we don't have to do anything manually?

If we're OK with using web3-eth, then I'd like to make a suggestion: Ditch the adapter, and just do everything with web3-eth directly in `decoders.ts`.  (Arguably we should be doing that everywhere.)  This adapter doesn't really add anything if that's the case.

(Yeah, this kind of contradicts my previous comment.  I would synthesize them as follows: My primary recommendation would be to ditch the adapter entirely, but *if* you want to keep it, it should live in a separate package, like being part of `InterfaceAdapter`, rather than going here.)

haltman-at (commented):
Some comments about how this is structured.  Going to also go through and maybe make some notes on Nick's comments.

packages/decoder/lib/Web3Adapter.ts
haltman-at: Worth noting, the web3 types are kind of unreliable atm; that's why we use `@types/web3` rather than trusting the types it comes with.  (When I tried switching to that, it went badly.)  They've been planning to fix this but I don't think they've done so yet?

haltman-at:
> Mostly just here to request the elimination of `any`s.
> 
> Also, in some places I'm suggesting explicit return values - @haltman-at might come in and take a hard line to require them for all methods, but I think we need good return types at least for the methods that have complex return values. (i.e. it's probably fine to let it slide for `Promise<string>`, etc., since text editors take the sting out of the unknown there)

I do in fact want to take a hard line on that, anything exported should be typed as explicitly as possible!

packages/decoder/lib/Web3Adapter.ts
haltman-at: As I mentioned above, with web3, if you're not using `@types/web3`, then the types are indeed often out of sync with the implementation, unfortunately. :-/  (Unless that's fixed now, but I don't think it is?)  I'm assuming there's not a `@types/web3-eth` we can use?  Otherwise, uh... maybe you can manually type things...?

haltman-at:
(Also obviously I agree with Nick about eliminating `any`, unless there's really some reason that it would be impractical to do so.)

PR #4333, opened 2021-09-28
Add support for Solidity 0.8.8's user-defined value types, and other changes to support 0.8.8 and 0.8.9

haltman-at:
This PR adds support for Solidity 0.8.8's user-defined value types.  Most of this is fairly straightforward, there's just a lot of places in the code that needed to be touched.

The biggest change is that of course a whole new typeclass, `"userDefinedValueType"`, had to be added to the type system, oh boy!  These work mostly like you expect, but I have a few notes on them.

Firstly, I had to have these work a bit differently from arrays/structs/mappings in terms of how they handle errors in the underlying thing.  See, with an array (for instance), the array can decode to a `Value`, but have individual elements that decode to `ErrorResult`s instead.  By contrast, with UDVTs, I made it so that if the underlying value decodes to an `ErrorResult`, then so does the UDVT value itself.

Why?  Well, uh, to avoid circular imports.  See, UDVTs can be used as mapping keys.  This means they have to be defined in `elementary.ts`.  But in order for a `Value` to be able to hold an `ErrorResult`, they'd have to be defined in `value.ts`.  So, I just made it so that a `Value` cannot hold an `ErrorResult`.  This is inconsistent with how we handle container types, but, oh well, whatever.  (Also it means that we don't have to worry about seemingly valid mapping keys that aren't actually valid mapping keys...)

Also, I added a few more collections of types (`BuiltInValueType`, `AbiType`) to the type system in order to make a few things work.  These aren't used very extensively, and they don't necessarily work completely well, but, oh well, at least they're there?

Another thing that has to be noted is the storage allocation of UDVTs.  Solidity 0.8.8 has a bug where these always take up a full word regardless of the size of the underlying type!  So, uh, that's how we allocate them.  We'll have to change this later when Solidity 0.8.9 comes out; this will unfortunately require an explicit version switch, as much as we try to avoid those.  I could add that to this PR if you think that's warranted, to be ready for it, but for now I've left it out.  (There's also a Solidity 0.8.8 bug where UDVTs are missing the `canonicalName` field in the AST, but that's easy to work around.)

One unexpected complication came from handling mapping keys!  Remember the no-op problem, that mapping keys given by expressions whose outermost layer was a no-op at the EVM level wouldn't be picked up?  Well, if `T` is a UDVT, then `T.wrap` and `T.unwrap` are both no-ops!  So, special handling had to be added for these in the debugger, alongside the other no-ops.

It's also worth noting how I chose to display such types in `ResultInspector`.  I made it so that, e.g., if you wrap `-1` in a type called `MyInt` defined in a contract `C`, it will display in the debugger (or anything using `ResultInspector`) as `C.MyInt.wrap(-1)` (and the `-1` will of course be colorized).  I think that conveys what's going on pretty well. :)

I also added support for UDVTs in `wrapElementaryValue` (and `constructSlot`), even though that'll be going away once #4142 is merged.  (And obviously once this PR is merged I'll have to update that one!)

One caveat to all this: Due to the way we handle `constant`s, it's unlikely that any UDVTs declared as `constant` will be decodeable.  But I see that as a minor problem, and something we can go back and do something about later if we really want.  We're already pretty limited in what `constant`s we can decode.

I also fixed up a few other things while I was at it, including factoring some repetitive code in `ast/import/index.ts`, and a bug in `isContractDefinedType`; I won't detail all these changes though, a number of them are just like changing some `let`s to `const`s. :P

Finally, of course, I added tests.  I decided not to add any specific new tests of UDVTs, but rather just sprinkled them into some existing tests.  I think it should be sufficient.  Some of the stuff not covered by these tests I checked manually.

And that's it!  Until Solidity 0.8.9, anyway.  (Or until I go back and do that work now in preparation for it. :P )

### DIFF
--- a/packages/codec/lib/basic/decode/index.ts
+++ b/packages/codec/lib/basic/decode/index.ts
@@ -37,7 +37,7 @@ export function* decodeBasic(
 
   switch (dataType.typeClass) {
     case "userDefinedValueType": {
-      const fullType = <Format.Types.UdvtType>(
+      const fullType = <Format.Types.UserDefinedValueTypeType>(
         Format.Types.fullType(dataType, info.userDefinedTypes)
       );
       if (!fullType.underlyingType) {
@@ -64,7 +64,7 @@ export function* decodeBasic(
       switch (underlyingResult.kind) { //yes this switch is a little unnecessary :P
         case "value":
           //wrap the value and return
-          return <Format.Values.UdvtValue>{ //no idea why need coercion here
+          return <Format.Values.UserDefinedValueTypeValue>{ //no idea why need coercion here
             type: fullType,
             kind: "value" as const,
             value: underlyingResult
@@ -76,7 +76,7 @@ export function* decodeBasic(
           //does not cause an error in the whole thing, but to do that here
           //would cause problems for the type system :-/
           //so we'll just be inconsistent
-          return <Format.Errors.UdvtErrorResult>{ //TS is being bad again :-/
+          return <Format.Errors.UserDefinedValueTypeErrorResult>{ //TS is being bad again :-/
             type: fullType,
             kind: "error" as const,
             error: {
--- a/packages/codec/lib/basic/decode/index.ts
+++ b/packages/codec/lib/basic/decode/index.ts
@@ -36,6 +36,57 @@ export function* decodeBasic(
   debug("pointer %o", pointer);
 
   switch (dataType.typeClass) {
+    case "userDefinedValueType": {
+      const fullType = <Format.Types.UDVTType>(
+        Format.Types.fullType(dataType, info.userDefinedTypes)
+      );
+      if (!fullType.underlyingType) {
+        const error = {
+          kind: "UserDefinedTypeNotFoundError" as const,
+          type: fullType,
+        };
+        if (strict || options.allowRetry) {
+          throw new StopDecodingError(error, true);
+          //note that we allow a retry if we couldn't locate the underlying type!
+        }
+        return {
+          type: fullType,
+          kind: "error" as const,
+          error
+        };
+      }
+      const underlyingResult = yield* decodeBasic(
+        fullType.underlyingType,
+        pointer,
+        info,
+        options
+      );
+      switch (underlyingResult.kind) { //yes this switch is a little unnecessary :P
+        case "value":
+          //wrap the value and return
+          return <Format.Values.UDVTValue>{ //no idea why need coercion here
+            type: fullType,
+            kind: "value" as const,
+            value: underlyingResult
+          };
+        case "error":
+          //wrap the error and return an error result!
+          //this is inconsistent with how we handle other container types
+          //(structs, arrays, mappings), where having an error in one element
+          //does not cause an error in the whole thing, but to do that here
+          //would cause problems for the type system :-/
+          //so we'll just be inconsistent
+          return <Format.Errors.UDVTErrorResult>{ //TS is being bad again :-/
+            type: fullType,
+            kind: "error" as const,
+            error: {
+              kind: "WrappedError",
+              error: underlyingResult
+            }
+          };
+      }
+      break; //to satisfy TS :P
+    }
     case "bool": {
       if (!checkPadding(bytes, dataType, paddingMode)) {
         let error = {
--- a/packages/codec/lib/basic/decode/index.ts
+++ b/packages/codec/lib/basic/decode/index.ts
@@ -704,7 +704,7 @@ function checkPadding(
   userDefinedTypes?: Format.Types.TypesById
 ): boolean {
   const length = byteLength(dataType, userDefinedTypes);
-  let paddingType = getPaddingType(dataType, paddingMode);
+  const paddingType = getPaddingType(dataType, paddingMode);
   if (paddingMode === "permissive") {
     switch (dataType.typeClass) {
       case "bool":
@@ -737,11 +737,10 @@ function removePaddingDirect(
   paddingType: PaddingType
 ) {
   switch (paddingType) {
-    case "left":
-    case "signed":
-      return bytes.slice(-length);
     case "right":
       return bytes.slice(0, length);
+    default:
+      return bytes.slice(-length);
   }
 }
 
@@ -757,6 +756,8 @@ function checkPaddingDirect(
       return checkPaddingRight(bytes, length);
     case "signed":
       return checkPaddingSigned(bytes, length);
+    case "signedOrLeft":
+      return checkPaddingSigned(bytes, length) || checkPaddingLeft(bytes, length);
   }
 }
 
@@ -770,9 +771,14 @@ function getPaddingType(
     case "default":
     case "permissive":
       return defaultPaddingType(dataType);
-    case "zero":
-      let defaultType = defaultPaddingType(dataType);
+    case "zero": {
+      const defaultType = defaultPaddingType(dataType);
       return defaultType === "signed" ? "left" : defaultType;
+    }
+    case "defaultOrZero": {
+      const defaultType = defaultPaddingType(dataType);
+      return defaultType === "signed" ? "signedOrLeft" : defaultType;
+    }
   }
 }
 
--- a/packages/codec/lib/basic/decode/index.ts
+++ b/packages/codec/lib/basic/decode/index.ts
@@ -37,7 +37,7 @@ export function* decodeBasic(
 
   switch (dataType.typeClass) {
     case "userDefinedValueType": {
-      const fullType = <Format.Types.UDVTType>(
+      const fullType = <Format.Types.UdvtType>(
         Format.Types.fullType(dataType, info.userDefinedTypes)
       );
       if (!fullType.underlyingType) {
@@ -64,7 +64,7 @@ export function* decodeBasic(
       switch (underlyingResult.kind) { //yes this switch is a little unnecessary :P
         case "value":
           //wrap the value and return
-          return <Format.Values.UDVTValue>{ //no idea why need coercion here
+          return <Format.Values.UdvtValue>{ //no idea why need coercion here
             type: fullType,
             kind: "value" as const,
             value: underlyingResult
@@ -76,7 +76,7 @@ export function* decodeBasic(
           //does not cause an error in the whole thing, but to do that here
           //would cause problems for the type system :-/
           //so we'll just be inconsistent
-          return <Format.Errors.UDVTErrorResult>{ //TS is being bad again :-/
+          return <Format.Errors.UdvtErrorResult>{ //TS is being bad again :-/
             type: fullType,
             kind: "error" as const,
             error: {
--- a/packages/codec/lib/core.ts
+++ b/packages/codec/lib/core.ts
@@ -4,6 +4,8 @@ const debug = debugModule("codec:core");
 import type * as Abi from "@truffle/abi-utils";
 import * as Ast from "@truffle/codec/ast";
 import * as AbiData from "@truffle/codec/abi-data";
+import * as Compiler from "@truffle/codec/compiler";
+import type { PaddingMode } from "@truffle/codec/common";
 import * as Topic from "@truffle/codec/topic";
 import type * as Pointer from "@truffle/codec/pointer";
 import type {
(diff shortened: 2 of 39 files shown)


haltman-at:
OK, since 0.8.9 has come out, I've added additional changes to deal with the changes in 0.8.9.  Let's go over those.

First of all, the storage allocation bug is fixed... but this means that we have to be able to allocate one way for 0.8.8, and another way for 0.8.9.  To handle this, when gathering up user-defined types (either in the debugger or in the decoder), we now keep track of the compiler that was used for them as well.  Right now I've made it so that this is only passed in to the allocator for storage allocations, but now that we're keeping track, it should be easy to have it affect other allocators too in the future if necessary.  (It could also be useful in the future if we ever want to go decoding Vyper variables or something.)

I don't like explicitly switching on the Solidity version like this, but, I didn't see any good alternative here.  If we still did things based on AST nodes, we could have relied on the fact that `canonicalName` is missing in 0.8.8; but we now do things based on `Type`s, so that's not really an option, unless we're going to smuggle that information into the `Type`, which we don't want to do.

But, I'm afraid I've actually added more explicit version switches in these changes... an entire *three* of them, in fact. D:  Sorry.  Actually, I think two of these are eliminable; this is how I did it, but if you want me to, I could redo it to remove those two.  Anyway, I should explain what this is all about.

See, Solidity 0.8.9 changed how immutables are encoded -- both in the constructor context and in the deployed context.  In the deployed context, of course, immutables live in the bytecode itself; they take up a full 32 bytes there, regardless of the size of the type.  This of course raises the issue of padding.  Prior to Solidity 0.8.9, immutables were always zero-padded, regardless of the type; now, they're padded the normal way (i.e., signed types are sign-padded).  In the constructor context, immutables live in memory.  Prior to Solidity 0.8.9, while in memory, they were always right-padded, regardless of the type; now, they're padded the normal way (i.e., `bytesN` is right-padded, other types are left-padded, and signed types are sign-padded).

So, how do we handle this?  Well, for the case of immutables in memory, I didn't see any other option but to add an explicit version switch.  So, I did that.

For the case of immutables in bytecode... I also added an explicit version switch.  Two of them, in fact, since that code gets called in two places.  But here I'm not confident it was necessary, and I'm thinking of redoing it.  Of course, I want to see what your opinion is; possibly I could redo it in a separate PR afterward, if we want to get this one out quickly.

So, what's the alternative?  Well, there's a few.  We could use the `"permissive"` padding mode, which just doesn't check the padding.  But I don't want to do that, because permissive padding mode is intended for cases when the padding might possibly be anything at all, and so we shouldn't perform any check on it.  Here, we know it's one of two things, just not which; so I'd like to run some sort of check on it, to make sure it's not total garbage.

We could add a new padding mode -- this was my original approach, to add a new padding mode I called `"consistent"`.  Now currently, a given padding mode, and a given data type, determines a unique pading type; to make `"consistent"` work, I changed that, so that a given padding mode with a given data type could have multiple padding types.  Unfortunately that ran into problems with error cases, because in error cases we report the padding type, but, oops, it's not unique anymore.

At this point I gave up and put in the explicit version switches I mentioned.  It was only afterward I thought of a third possibility: I could add `"consistent"` as a new padding mode *and* as a new padding *type* as well.  Oh well.  Like I said, I can redo it this way if you think that makes sense; or I could save it for a separate PR; or we could just leave it as-is.  Regardless, that's the state of things.

Of course I also updated the tests to 0.8.9; but I made sure to also run the tests myself on 0.8.8 to make sure that the new code works with both!

haltman-at:
Sorry, one other thing I should mention: We may want to start using the Solidity compiler's storage layout output?  We can't rely solely on that, because it only exists in sufficiently recent versions, but it might potentiall allow us to get away with not having a version switch here?  That would be more work though so I didn't do it that way in this PR; we could possibly go back and do that later.

haltman-at:
One additional note, I moved the compiler version check for UDVTs from `bytesLength` (where I don't think it really belonged) to `storageSizeAndAllocate`.  I can explain this move in more detail if people want but otherwise I think I'll skip it. :P

haltman-at:
OK, I made the padding change, thereby removing two explicit version checks.  I didn't use the name `"consistent"`, that's too confusing, instead I went with `"defaultOrZero"` and `"signedOrLeft"`; rather more explicit.

Btw, there might be a better way to handle the memory thing, but
1. It would still presumably require an explicit version check somewhere, and
2. I don't want to do that in this PR, that's too much work!

As always, I made sure to run the tests locally with both 0.8.8 and 0.8.9.

packages/codec/lib/format/elementary.ts
```diff
@@ -208,6 +218,17 @@ export interface EnumValue {
   };
 }
 
+/**
+ * A UDVT value
+ *
+ * @Category User-defined elementary types
+ */
+export interface UDVTValue {
```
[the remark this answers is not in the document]
gnidan: (and of course for type and result)

packages/codec/lib/basic/decode/index.ts
```diff
@@ -36,6 +36,57 @@ export function* decodeBasic(
...
+    case "userDefinedValueType": {
+      const fullType = <Format.Types.UDVTType>(
+        Format.Types.fullType(dataType, info.userDefinedTypes)
+      );
+      if (!fullType.underlyingType) {
+        const error = {
+          kind: "UserDefinedTypeNotFoundError" as const,
+          type: fullType,
+        };
+        if (strict || options.allowRetry) {
```
cds-amal, line 48: How does strict effect retry?

packages/codec/lib/core.ts
```diff
@@ -850,7 +851,7 @@ function* decodeBytecode(
         value = yield* decode(dataType, variable.pointer, info, {
           allowRetry: true, //we know we're in full mode
           strictAbiMode: true,
-          paddingMode: "zero" //force zero-padding!
+          paddingMode: "defaultOrZero"
```
cds-amal, line 854: Why is this necessary?

packages/decoder/test/current/test/downgrade-test.js
```diff
@@ -411,11 +411,11 @@ function verifyAbiDecoding(decoding, address) {
   assert.lengthOf(decoding.arguments, 4);
   //we'll skip verifying the names as well
 
-  //first argument: {{x: 7, y: 5}, z: 3}
+  //first argument: {{x: 7, y: -5}, z: 3}
   assert.strictEqual(decoding.arguments[0].value.type.typeClass, "tuple");
   assert.deepEqual(
     Codec.Format.Utils.Inspect.unsafeNativize(decoding.arguments[0].value),
-    [[7, 5], 3]
+    [[7, -5], 3]
```
cds-amal, line 418: what does the -5 mean?

packages/decoder/lib/decoders.ts
```diff
@@ -1707,7 +1709,10 @@ export class ContractInstanceDecoder {
         break;
       case "mapping":
         let keyType = parentType.keyType;
-        if (keyType.typeClass === "enum") {
+        if (
+          keyType.typeClass === "enum" ||
+          keyType.typeClass === "userDefinedValueType"
+        ) {
```
cds-amal, line 1715: Is this correct? enum and UDVT are different.

packages/codec/lib/format/elementary.ts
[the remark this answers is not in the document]
haltman-at: Sure thing, that seems like a straightforward change to make.  I think I'd like to just go with `UdvtValue`?  The non-abbreviated version is really long...

packages/decoder/test/current/test/downgrade-test.js
haltman-at: This is just a test fixture.  Since I changed that particular member from an unsigned type to a signed type, I figured I'd also change the test fixture to a negative number (as those are more likely to cause problems).

packages/codec/lib/core.ts
haltman-at: So, this code is for decoding immutables.  Prior to Solidity 0.8.9, immutables, rather than using the normal padding, were always zero-padded regardless of type.  However, starting in 0.8.9, they use normal padding.  So, previously this code was set to assume zero-padding.  However, now I've changed it to tolerate either zero-padding or normal padding.  Notionally, I could have it specifically check for one or the other based on the compiler version, but I really prefer to avoid version checks wherever possible, so instead I just had it tolerate either.

packages/codec/lib/storage/allocate/index.ts
```diff
@@ -541,10 +561,24 @@ function storageSizeAndAllocate(
...
+      }
+      //otherwise, treat them normally
+      //DELIBERATE FALL-TRHOUGH
     default:
       //otherwise, it's a direct type
       return {
-        size: { bytes: Basic.Allocate.byteLength(dataType, userDefinedTypes) },
+        size: {
+          bytes: Basic.Allocate.byteLength(dataType, userDefinedTypes)
+        },
```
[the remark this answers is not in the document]
haltman-at: :)

packages/codec/lib/basic/decode/index.ts
haltman-at: Neither affects the other; they act fairly independently.  I'm guessing you meant to ask a more general question about why this condition?  I'll answer that instead if you don't mind. :)

In non-strict mode, when something goes wrong, we return an `ErrorResult`, whereas in strict mode, when something goes wrong, we throw.

Or rather, that's what we do for non-retriable errors; for retriable errors, if `allowRetry` is set, then what we do instead is to trigger a retry.  Of course, the way we trigger a retry is to throw an error, but set its `allowRetry` bit so that a retry will occur.

So, putting it together, on a retriable error, we throw if either `strict` *or* `allowRetry` is set, and we set its `allowRetry` flag based on the value of `options.allowRetry`.  Retries are for falling back from full mode to ABI mode, so this error, being unable to locate a user-defined type in full mode, is an example of a retriable error.  Thus, the condition.

packages/decoder/lib/decoders.ts
haltman-at: Yes, but both need to be covered here.  Notice what this if block does -- it fetches the full type information for the type.  We need that for both enums and for UDVTs, because we need the full type information to perform wrapping.  For enums we need to know the enum's overall range, and we need to know the names of its options to allow people to use them as input.  For UDVTs we need to know the underlying type because, well, everything is based on that!

Of course, there are two other types of user-defined types; however for contracts we don't need the full type information because we already know everything relevant for wrapping purposes, and as for structs, well, those aren't elementary types and so aren't relevant here.

Now you might say, isn't handling this here a bit awkward?  To which I say, yes!  Arguably, this code belongs inside `wrapElementaryValue` itself.  However, I didn't need to see a need to change it in this PR; especially as, once the encoder work is merged, this code will be eliminated.  So I'm fine with leaving it this way for now and just adding the extra typeclass to the condition.

packages/codec/lib/format/elementary.ts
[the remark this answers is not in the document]
haltman-at: OK, made that change.

packages/decoder/lib/decoders.ts
haltman-at: (Alternately, we could eliminate the check and just retrieve the full type information no matter what!  I just did it the same way it already was because like I said I didn't see a need to redo this.)

cds-amal: Thanks for the explanation!

packages/codec/lib/core.ts
cds-amal: Ah, that context helps! Thank you!

packages/codec/lib/basic/decode/index.ts
cds-amal: This indeed was the question I couldn't articulate. Thanks @haltman-at !

cds-amal (approved):
Thanks for the explanations @haltman-at. LGTM, pending @gnidan's asks.

packages/codec/lib/basic/decode/index.ts
haltman-at: Actually I just realized I made a slight mistake in the above -- for retriable errors, we don't set `allowRetry` on the error based on the `allowRetry` option; instead we just always set it to `true` regardless.  It's up to the caller to know whether it wants to retry or not.

packages/codec/lib/format/elementary.ts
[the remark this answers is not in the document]
gnidan: I prefer spelling it out, reasoning being that "UVDT" probably won't be easily recognized by users of this package. I fear that the spelled-out form is also unrecognizable, but at least it's more clear that it's something to be googled that way.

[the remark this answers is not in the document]
haltman-at: Hm. Can we rely on the docs, and the fact that the `typeClass` discriminator is spelled out long form as `"userDefinedValueType"`, to assist here? `UserDefinedValueType` is just so long!  And then there's the question of what to call the type objects; `UserDefinedValueTypeType` is hella awkward, but `UserDefinedValueType` alone is kind of inconsistent.  The latter is probably the better option, if we have to, but I like how the acronym obviates the problem. :P

Also anyone using this package's output directly had better be prepared to know all sorts of uncommon things about the Solidity type system, I kind of want to say. :P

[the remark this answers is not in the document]
cds-amal: I agree that it should be spelled out.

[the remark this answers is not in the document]
haltman-at: Hm, OK, I guess I can spell it out.  But, does that mean we do `UserDefinedValueTypeType` (ick), or `UserDefinedValueType` together with `UserDefinedValueTypeValue` (inconsistent)?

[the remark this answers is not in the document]
haltman-at: It seems @cds-amal also agrees, so that's what I'm going with!

PR #4339, opened 2021-09-30
Internal improvement: only check permissions for monorepo packages

cds-amal:
- This PR will only check permissions for packages in the monorepo against **truffle** org. 
- It drops checking against the **trufflesuite** org because we're not publishing there on npmjs.

### DIFF
--- a/scripts/npm-access.js
+++ b/scripts/npm-access.js
@@ -6,6 +6,14 @@ const getPkgPermissions = userOrOrg => {
   return JSON.parse(stringResponse);
 };
 
+const getMonorepoPackages = () => {
+  // get list of monorepo packages and remove lerna branding from output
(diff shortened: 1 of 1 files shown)


scripts/npm-access.js
```diff
@@ -6,6 +6,14 @@ const getPkgPermissions = userOrOrg => {
   return JSON.parse(stringResponse);
 };
 
+const getMonorepoPackages = () => {
+  // get list of monorepo packages and remove lerna branding from output
+  const pkgs = execSync('$(yarn bin)/lerna ls | grep -v "^lerna"')
```
cds-amal, line 11: should probably remove the `$(yarn bin)` since the script will have the correct path to resolve lerna

gnidan: will this work on windows?

cds-amal: Nope, the filtering should be done in node space.

Issue #3887, opened 2021-02-26
allow truffle test to be ran without forcing deployment PLEASE


eyooooo:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

I have royally fucked up all of my migrations - i want to  just run tests against local ganache WITHOUT my deploy scripts being called. I wish there was a way to force truffle test to NEVER run deploy scripts. 

## Steps to Reproduce

fuck up a bunch of deployments then try to run some tests

## Expected Behavior

if i am only calling .new() in my test scripts, truffle should NOT try to run my migrations. 

## Actual Results

truffle wants to run my migrations 

## Environment

* Operating System: 
* Ethereum client: any
* Truffle version (`truffle version`): latest
* node version (`node --version`): 14
* npm version (`npm --version`):  any

eggplantzzz:
@eyooooo Thanks for the suggestion! This could be a handy feature, we'll stick in the backlog!

gnidan:
Action here seems to be to add a `truffle test --migrate-none` flag, similar to how we have `truffle test --compile-none`

PR #4340, opened 2021-10-01
Add support of truffle test --migrate-none and event message #3887

dongmingh:


packages/core/lib/testing/Test.js
```diff
@@ -244,7 +244,7 @@ const Test = {
     const migrateConfig = config.with({
       reset: true,
       resolver: resolver,
-      quiet: true
+      quiet: false
```
gnidan, line 247: Should we set this to `false` if-and-only-if `--migrate-none` is turned on? `truffle test` silences migrate output on purpose right now, and I don't think we want to change that most of the time.

I do like your use of this mechanism, though! Let the migrate system do its own logging!

packages/migrate/Migration.js
```diff
@@ -185,6 +185,14 @@ class Migration {
    * @param  {Object}   options  config and command-line
    */
   async run(options) {
+    // check --migrate-none
+    if (config.migrateNone || config['migrate-none']) {
+        if (config.events) {
+            config.events.emit("migrate:nothingToMigrate");
```
gnidan, line 191: I know you copied this from `truffle test --compile-none`, but I think we should use a better message here, e.g. `migrate:skipped`, which would say something like "> Migrations have been skipped, not migrating."

(Maybe there should be an issue to fix this for the compile case)

gnidan (commented):
Here's a few small comments; would be good for others to review also!

packages/core/lib/testing/Test.js
haltman-at: Yeah I'm also confused by this; I don't think we should be setting it to `false` at all?  Presumably we want to suppress the printout regardless.

haltman-at:
A note on the git usage here -- you seem to have forked the repo and are making a PR from that fork?  That's a bit troublesome for other people to get at.  (Like, I wanted to check out your branch to inspect it, but I couldn't, because it's on another repo.)  The thing to do instead is to simply make a branch, on the main repo, and PR that branch, not create a whole fork of the repo.  Feel free to ask about this if you're not sure how that works.

packages/migrate/Migration.js
```diff
@@ -185,6 +185,14 @@ class Migration {
    * @param  {Object}   options  config and command-line
    */
   async run(options) {
+    // check --migrate-none
+    if (config.migrateNone || config['migrate-none']) {
```
haltman-at, line 189: Style note: We use 2-space indenting here, not 4. :)  We used to have `prettier` handle this sort of thing for you, but, uh, husky hasn't been working in quite a while (#3530), so we're stuck doing it ourselves for now. :-/  I believe @cds-amal was working on fixing this, but I'm not sure what the current status of that is.

packages/core/lib/testing/Test.js
dongmingh: will change it to false if and only if --migrate-none

packages/events/defaultSubscribers/migrate.js
```diff
@@ -0,0 +1,24 @@
+const OS = require("os");
+
+module.exports = {
+  initialization: function () {
+    this.logger = console;
+  },
+  handlers: {
+    "migrate:start": [
```
haltman-at, line 8: OK, something really does not make sense here.  You've added a handler for `"migrate:start"`... but you haven't added anything to actually *emit* this event, nor have you removed any existing code that would perform this printout manually.  What's going on here?

packages/migrate/Migration.js
dongmingh: will change accordingly

packages/migrate/Migration.js
```diff
@@ -185,6 +185,14 @@ class Migration {
...
+        if (config.events) {
```
haltman-at, line 190: Is this check necessary?  I don't think we check this in other places, do we?  We can assume that `config.events` exists, right?

haltman-at (commented):
Reviewing this a bit early, normally I'd wait until it's passing CI. :P

haltman-at:
Btw, if your PR is meant to address a specific issue (#3887), you should mention that issue in the text, not just the title, because if you only put it in the title, it doesn't become a clickable link.

...however, you want to avoid using keywords that will trigger GitHub's auto-close thing, because we don't use that here. :)  (We close issues when the fix is released, rather than when it's merged.)  So, uh, be careful of that... I like to say things like "Addresses #3887" because that won't trip GitHub's system. :)

packages/events/defaultSubscribers/migrate.js
dongmingh: remove migrate:start

packages/migrate/Migration.js
dongmingh: yes, config.events exists, remove the if-statement

Issue #3837, opened 2021-02-20
Add tests of `truffle debug --fetch-external`, `@truffle/fetch-and-compile`, or `@truffle/source-fetcher`


haltman-at:
## Issue

The `--fetch-external` feature, and the `fetch-and-compile` and `source-fetcher` packages, have no tests.  Some tests should be added for at least one of these.

haltman-at:
There are now tests of this due to #4424!

PR #4341, opened 2021-10-01
Bug fix: Fix `fetchAndCompile()`

haltman-at:
This function didn't actually work, because we never actually set `this.recognized` to true anywhere in `SingleRecognizer`.  Now we do!

...and yeah, this wasn't caught till now because the fetch-and-compile package has no tests (#3837)... :-/


gnidan (approved):
oops!

PR #4342, opened 2021-10-04
Revert "Update web3 to 1.6.0"

haltman-at:
ChainSafe suggested on Slack we revert the web3 update as an experiment, see if CI is any better as a result.  (Again, inferring causality may be hard, but...).  So, here's the PR for that.


cds-amal (approved):
Approved. Lets see what happens.

haltman-at:
Well, this one passed first try, so that's a good sign!

PR #4343, opened 2021-10-04
Internal improvement: Unpin Ubuntu version in GitHub Actions

haltman-at:
It seems that, at some point, `ubuntu-latest` in GitHub Actions finally updated to 20.04, meaning we can finally unpin this.  (The "latest" version of OSes in GHA often lags behind the actual latest that they make available, so if for some reason you need the actual latest -- which we did for a while -- you have to pin it.)


gnidan (approved):
Cool!

PR #4346, opened 2021-10-05
Internal improvement: Extract web3 eth

eggplantzzz:
This PR is a continuation of the work in https://github.com/trufflesuite/truffle/pull/4331.

This removes web3-eth as a dependency from @truffle/decoder. A ProviderAdapter class is created to make rpc calls and those are now implemented manually as opposed to using web3 for this purpose.


packages/decoder/lib/ProviderAdapter.ts
```diff
@@ -0,0 +1,167 @@
...
+  } else if (typeof block === "string" && !isNaN(parseInt(block))) {
+    // block is a string representation of a number
+    if (block.startsWith("0x")) return block;
+    // convert to hex and add '0x' prefix in case block is decimal
+    return `0x${parseInt(block).toString(16)}`;
+  } else if (typeof block === "number") {
+    return `0x${block.toString(16)}`;
+  } else {
+    throw new Error(
+      "The block specified must be a number or one of the strings 'latest'," +
```
eggplantzzz, line 52: Here I decided to neglect mentioning "genesis", should it be mentioned here? I figured it would encourage people to use the rpc spec terms.

packages/decoder/lib/ProviderAdapter.ts
```diff
@@ -0,0 +1,167 @@
...
+    return await this.sendRequest({
+      method: "eth_getCode",
+      params: [
+        address,
+        blockToFetch
+      ]
+    });
+  }
+
+  public async getBlock (block: BlockSpecifier) {
```
eggplantzzz, line 102: Should this maybe be renamed to `getBlockByNumber`? Looks like web3's `getBlock` method is used for both "eth_getBlockByNumber" and "eth_getBlockByHash"

gnidan: Yeah, I think that rename is a good idea

packages/decoder/lib/types.ts
```diff
@@ -1,4 +1,5 @@
 import type BN from "bn.js";
+import type Web3 from "web3";
```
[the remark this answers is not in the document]
eggplantzzz: I think so? It is used as a type on line 339, unless we can change that.

packages/decoder/lib/ProviderAdapter.ts
eggplantzzz: It does also accept "latest", "earliest"....etc.

Issue #4348, opened 2021-10-05
update event message of truffle test --compile-none


dongmingh:
- [ ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

The event message of `truffle test --compile-none` shows
> Everything is up to date, there is nothing to compile.

Indeed, the compilations are skipped upon the request of the test.  To reflect the actual behavior, the event message can be updated to
> Compilations have been skipped, not compiling.

## Steps to Reproduce

1. issue command `truffle test --compile-none`
2. check the event message

## Expected Behavior

An event message:
> Compilations have been skipped, not compiling.

## Actual Results

> Everything is up to date, there is nothing to compile.

## Environment

* Operating System: Mac M1
* Ethereum client:
* Truffle version (`truffle version`): v5.4.11
* node version (`node --version`): v16.10.0
* npm version (`npm --version`): v6.14.15

PR #4349, opened 2021-10-05
Update `truffle test --compile-none` event message

dongmingh:
This PR is to update `truffle test --compile-none` event message to
> Compilations have been skipped, not compiling.
to reflect the actual test.
This PR addresses #4348.


packages/events/defaultSubscribers/compile.js
```diff
@@ -62,11 +62,11 @@ module.exports = {
         this.logger.log(`${OS.EOL}    ${infos.join()}`);
       }
     ],
-    "compile:nothingToCompile": [
```
gnidan, line 65: We can't just get rid of this! With compilation, we might have "nothing to compile", or we might be skipping. There are two separate situations here.

gnidan (changes requested):

eggplantzzz:
Yeah so you will need to check whether the user provided `--compile-none` and if so, create a new event for something like  "compile: compilationSkipped". Otherwise, if they did not pass that option and nothing was compiled, then nothing was out-of-date (needed compilation).

haltman-at (approved):
Redundantly approving this. :)

Issue #4335, opened 2021-09-29
Environment.develop overwrites config options instead of adding to them


haltman-at:
## Issue

(Split off from #4327.)

In the [`Environment.develop` function](https://github.com/trufflesuite/truffle/blob/develop/packages/environment/environment.js#L73), you can see that `config.networks[network]` is overwritten with a custom, barebones, network config.  Presumably, the new network config should keep other existing config options in place, and only overwrite `network_id`, `provider`, and other ways of specifying a provider; leaving in place such things as `gas`, `disableConfirmationListener`, etc.

This prevents configuring the network used in `truffle develop`.  It also prevents configuring the network used in `truffle test`, although solving that will also require addressing #4327.

## Environment

* Truffle version (`truffle version`): 5.4.12

haltman-at:
Oh hey looks like this was never closed.  Closing!

PR #4356, opened 2021-10-08
environment.develop overwrite config options

dongmingh:
This PR fixes issue #4335 that config network is overwritten
with a custom, barebones network config.

### DIFF
--- a/packages/environment/test/develop.js
+++ b/packages/environment/test/develop.js
@@ -1,0 +1,60 @@
+const assert = require("chai").assert;
+const { default: Box } = require("@truffle/box");
+const { Environment } = require("@truffle/environment");
+const sinon = require("sinon");
+
+describe("Environment develop", function () {
+  let config;
+
+  const expectedNetwork = {
+    port: 8545,
+    network_id: 1342,
+    gas: 90000
+  };
+
+  const testOptions = {
+    port: 5678,
+    network_id: expectedNetwork.network_id,
+    gas: 7000
+  };
+
+  beforeEach("Create a sandbox", async function () {
+    this.timeout(5000);
+    config = await Box.sandbox("default");
+    config.network = "network";
+    config.networks = {
+      network: {
+        port: expectedNetwork.port,
+        network_id: 4321,
+        gas: expectedNetwork.gas
+      }
+    };
+    sinon.stub(Environment, "detect");
+  });
+
+  afterEach("Restore Environment detect", async function () {
+    Environment.detect.restore();
+  });
+
+  it("Environment.develop overwrites the network_id of the network", async function() {
+    await Environment.develop(config, testOptions);
+    const mutatedNetwork = config.networks[config.network];
+
+    assert.equal(
+      mutatedNetwork.port,
+      expectedNetwork.port,
+      "The port of the network should be unchanged."
+    );
+    assert.equal(
+      mutatedNetwork.network_id,
+      expectedNetwork.network_id,
+      "The network_id of the network should be overwritten."
+    );
+    assert.equal(
+      mutatedNetwork.gas,
+      expectedNetwork.gas,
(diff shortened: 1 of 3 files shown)


packages/environment/test/develop.js
```diff
@@ -0,0 +1,67 @@
...
+    gas: 90000
+  };
+
+  const testOptions = {
+    port: 5678,
+    network_id: expectedNetwork.network_id,
+    gas: 7000
+  };
+
+  before("Create a sandbox", async function () {
```
eggplantzzz, line 23: You probably want to use `beforeEach` and `afterEach` here. `before` and `after` only get run once while the others run before and after each `it` block.

eggplantzzz: Even though in this case you only have one test so it doesn't matter.

packages/environment/test/develop.js
```diff
@@ -0,0 +1,67 @@
...
+    config = await Box.sandbox("default");
+    config.network = "network";
+    config.networks = {
+      network: {
+        port: expectedNetwork.port,
+        network_id: 4321,
+        gas: expectedNetwork.gas
+      }
+    };
+    sinon.stub(Environment, "detect").returns();
```
eggplantzzz, line 34: If you are not concerned with having a return value, just delete the chained `.returns` here.

packages/environment/test/develop.js
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  after("Cleanup tmp files", function (done) {
```
eggplantzzz, line 37: I'm not going to push too hard on this but I would prefer to use `async`/`await` syntax in place of the `done` callback.

packages/environment/test/develop.js
```diff
@@ -0,0 +1,67 @@
...
+    Environment.detect.restore();
+    glob("tmp-*", (err, files) => {
```
eggplantzzz, line 39: I don't think you need this tmp file cleanup stage as I don't think any are created in this test.

packages/environment/test/develop.js
```diff
@@ -0,0 +1,67 @@
...
+      if (err) done(err);
+      files.forEach(file => fs.removeSync(file));
+      done();
+    });
+  });
+
+  it("Environment.develop overwrites the network_id of the network", async function() {
+    await Environment.develop(config, testOptions);
+    const returnedNetwork = config.networks[config.network];
```
eggplantzzz, line 48: A nit, but this isn't really a "returned network". This test works because the config is mutated by `Environment.develop`, right?

packages/environment/test/develop.js
haltman-at: I think we should definitely be avoiding callbacks in new code if there's not a reason we need them...

packages/environment/test/develop.js
haltman-at: I think that's more than a nit, that's a significant misname.  It's a mutated network or an altered network, not a returned network!

packages/environment/test/develop.js
dongmingh: change to beforeEach and afterEach

packages/environment/test/develop.js
dongmingh: ok, will remove the chained .returns

packages/environment/test/develop.js
dongmingh: agree, change to async/await

packages/environment/test/develop.js
dongmingh: will remove tmp file cleanup

eggplantzzz (approved):
I like it!

cds-amal (approved):
Thanks @dongmingh ! Looks good to me!

PR #4365, opened 2021-10-13
Respect solc version when importing ABI JSON files inside Solidity code

gnidan:
This PR upgrades abi-to-sol to ^0.5.1, taking advantage of the new functionality in that package to generate version-specific Solidity output.

In order to accommodate this new functionality, we need to know what Solidity version to generate, and so this PR builds on the work of #4420 by hooking the compiler information into the `ABI` ResolverSource. 

Mostly straightforward, but I noticed one gotcha: when `resolver.resolve()` receives compiler info, it likely receives a version string containing the full commit and target info. This PR ensures the `ABI` ResolverSource strips that off because otherwise abi-to-sol throws an exception.

### DIFF
--- a/packages/resolver/lib/sources/abi.ts
+++ b/packages/resolver/lib/sources/abi.ts
@@ -3,7 +3,7 @@ import type { ContractObject } from "@truffle/contract-schema/spec";
 import { generateSolidity } from "abi-to-sol";
 import type { ResolverSource } from "../source";
 
-export class ABI {
+export class ABI implements ResolverSource {
 
   wrappedSource: ResolverSource;
 
@@ -17,7 +17,11 @@ export class ABI {
     return null;
   }
 
-  async resolve(importPath: string, importedFrom: string = "") {
+  async resolve(
+    importPath: string,
+    importedFrom: string = "",
+    options: { compiler?: { name: string; version: string } } = {}
+  ) {
     let filePath: string | undefined;
     let body: string | undefined;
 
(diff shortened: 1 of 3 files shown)


cds-amal (dismissed):
Looks good @gnidan . We'll know if it breaks soon enough :)

haltman-at:
This no longer needs to be merge-after-target, right? :)

haltman-at:
OK, *now* this no longer needs to be merge-after-target, right? :)

packages/resolver/lib/sources/abi.ts
```diff
@@ -68,3 +80,14 @@ export class ABI {
...
+
+function determineSolidityVersion(
+  compiler?: { name: string; version: string }
+): string | undefined {
+  if (!compiler) {
+    return;
+  }
+
+  const { version } = compiler;
+  return version.split("+")[0];
```
haltman-at, line 92: This line could really use a comment.

packages/resolver/lib/sources/abi.ts
```diff
@@ -17,19 +17,30 @@ export class ABI {
...
+      importPath,
+      importedFrom,
+      options
+    );
     if (resolution.body === undefined) {
       return { filePath, body };
     }
 
+    const { compiler } = options;
+    const solidityVersion = determineSolidityVersion(compiler);
```
haltman-at, line 42: OK, this is actually a pre-existing thing rather than a change you made, but I think it's worth adding a comment in here somewhere that if we're in this code we can assume we're compiling Solidity (because this resolver source is turned off for Vyper).

haltman-at (commented):
Seems good, just two quick comments.

packages/resolver/lib/sources/abi.ts
gnidan: Sure; agreed!

gnidan: Added this as a doc comment

packages/resolver/lib/sources/abi.ts
gnidan: Added!

haltman-at (approved):
Cool, seems good to me!

gnidan:
Thanks! Will merge when green.

gnidan:
(Also reworded this to be an enhancement, rather than a dependency update, to be more explicit about the reasoning behind this upgrade.)

PR #4375, opened 2021-10-27
Internal improvement: Eth adapter

eggplantzzz:
This PR performs some formatting for the provider adapter in decoder. This adds an argument to the method responsible for sending so that each method responsible for an rpc call can specify its own formatter to use before returning the value. This will help maintain parity with the types that web3 returns (web3 transforms/formats a lot of the rpc response values) and hopefully make it easier to switch out in favor of this utility eventually.


haltman-at (approved):
Seems good to me!

Issue #3834, opened 2021-02-20
Add runToEnd method to debugger


haltman-at:
## Issue

Currently you can run the debugger to the end by using `continueUntilBreakpoint` and either not having any breakpoints set or specifying that you want no breakpoints.  A number of people find this unintuitive, so we should add an explicit `runToEnd` method.

Also, the debugger tests that use `continueUntilBreakpoint` as a way of running to the end could be updated to use this instead.

haltman-at:
This is now done.  Closing!

PR #4378, opened 2021-10-29
Added an explicit debugger method to run to the end

sukanyaparashar:
### ISSUE
The  debugger runs to the end with continueUntilBreakpoint method without specifying the breakpoints. See issue #3834 

### SOLUTION
Added an explicit runToEnd method and use that method in the debugger tests.

### DIFF
--- a/packages/debugger/lib/controller/sagas/index.js
+++ b/packages/debugger/lib/controller/sagas/index.js
@@ -216,15 +216,11 @@ function* stepOver() {
  * runToEnd - run the debugger till the end
  */
 function* runToEnd() {
-  var finished;
+  let finished;
 
   do {
     yield* advance();
-
     finished = yield select(controller.current.trace.finished);
-    if (finished) {
-      break; //can break immediately if finished
-    }
   } while (!finished);
 }
 
--- a/packages/debugger/lib/controller/sagas/index.js
+++ b/packages/debugger/lib/controller/sagas/index.js
@@ -22,7 +22,8 @@ const STEP_SAGAS = {
   [actions.STEP_OVER]: stepOver,
   [actions.STEP_INTO]: stepInto,
   [actions.STEP_OUT]: stepOut,
-  [actions.CONTINUE]: continueUntilBreakpoint
+  [actions.CONTINUE]: continueUntilBreakpoint,
+  [actions.RUN_TO_END]: runToEnd
 };
 
 export function* saga() {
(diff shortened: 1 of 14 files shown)


packages/debugger/lib/controller/sagas/index.js
```diff
@@ -211,6 +212,22 @@ function* stepOver() {
   );
 }
 
+/**
+ * runToEnd - run the debugger till the end
+ */
+function* runToEnd() {
+  var finished;
```
gnidan, line 219: Never use `var`!
```suggestion
  let finished;
```

packages/debugger/lib/controller/sagas/index.js
```diff
@@ -211,6 +212,22 @@ function* stepOver() {
...
+
+  do {
+    yield* advance();
+
+    finished = yield select(controller.current.trace.finished);
+    if (finished) {
+      break; //can break immediately if finished
+    }
```
gnidan, line 227: This is redundant, since we check the `while(!finished)` immediately following this.
```suggestion
```

gnidan (changes requested):

packages/debugger/lib/controller/sagas/index.js
sukanyaparashar: Oh yes, I will remove that part and make the changes.

packages/debugger/lib/controller/sagas/index.js
cds-amal: We prefer `let` and `const` for variable declaration because it makes the code easier to reason about. With `var` you have to be aware of scoping and other implications of how the interpreter splits variable declaration from initialization. `let` and `const` define blocked scoped local variables and are easier to reason about. There's even annoying global pollution with `var` that thankfully `let` doesn't allow. See: [Hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting) and [let statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let) for more information.


Truffle's codebase has lots of `var` declarations because it was written before `let` and `const` came into the language. Now that they are available we should use them.

sukanyaparashar: Thanks @cds-amal . I will keep this thing in mind while writing codes.

haltman-at (approved):
Seems good to me!  Thanks for picking this up Sukanya.

PR #4381, opened 2021-10-29
Create a web3 http provider that can handle massive JSON blobs without crashing due to string or buffer length issues

davidmurdoch:


packages/stream-provider/test/stream.ts
```diff
@@ -0,0 +1,64 @@
...
+          if (err) return void reject(err);
+          resolve(result);
+        }
+      );
+    });
+    assert.strictEqual(result.structLogs.length, 4);
+    result.structLogs.forEach((log: Buffer) => {
+      assert.strictEqual(log.length, SIZE);
+    });
+  }) as any).timeout(0);
```
davidmurdoch, line 63: I don't know what is up with the mocha types, but it thinks `it()` returns `void` but it doesn't.

packages/stream-provider/lib/index.ts
```diff
@@ -0,0 +1,102 @@
...
+
+export default class StreamingWeb3HttpProvider extends Web3HttpProvider {
+  /**
+   * Should be used to make async request
+   *
+   * @method send
+   * @param {Object} payload
+   * @param {Function} callback triggered on end with (err, result)
+   */
+  send(
```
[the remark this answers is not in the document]
davidmurdoch: No. This is a drop-in replacement for `Web3HttpProvider`, which doesn't implement `request`.

packages/core/lib/command-utils.js
```diff
@@ -324,9 +325,11 @@ const displayGeneralHelp = options => {
...
         "Usage: truffle <command> [options]"
     )
-    .epilog("See more at https://trufflesuite.com/docs/" + 
-      OS.EOL +
-      "For Ethereum JSON-RPC documentation see https://ganache.dev")
+    .epilog(
+      "See more at https://trufflesuite.com/docs/" +
+        OS.EOL +
+        "For Ethereum JSON-RPC documentation see https://ganache.dev"
+    )
```
davidmurdoch, line 332:
```suggestion
    .epilog("See more at https://trufflesuite.com/docs/" + 
      OS.EOL +
      "For Ethereum JSON-RPC documentation see https://ganache.dev")
```

PR #4388, opened 2021-11-03
Internal improvement: Fix yarncheck

haltman-at:
Turns out I accidentally broke yarncheck, and the lockfile got out of sync!  This PR fixes it.

I thought that `--frozen-lockfile` meant that `yarn` would error out if the lockfile needed to be updated.  No, turns out it means it just doesn't update it.  Oops.

So, this PR changes yarncheck back to working the old way.  Of course, I left in the `--ignore-scripts` flag so it'll still be fast.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -19,7 +19,8 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
       - run: npm install -g yarn
-      - run: yarn install --ignore-scripts --frozen-lockfile --ignore-engines
+      - run: yarn install --ignore-scripts --ignore-engines
+      - run: test -z "$(git diff)" || (echo 'Please run yarn and commit all changes to yarn.lock'; false)
 
 
   build:


gnidan (approved):

cds-amal (approved):
Nice catch! Let's get it in!

.github/workflows/nodejs.yml
```diff
@@ -19,7 +19,8 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
       - run: npm install -g yarn
-      - run: yarn install --ignore-scripts --frozen-lockfile --ignore-engines
+      - run: yarn install --ignore-scripts --ignore-engines
+      - run: test -z "$(git diff)" || (echo 'Please run yarn and commit all changes to yarn.lock'; false)
```
cds-amal, line 23: This will not work with windows. But shouldn't stop this from merging.

I'll make a note of it in the [windows/runner](https://github.com/trufflesuite/truffle/tree/windows/runner) branch

haltman-at: Ooh, thanks, I wasn't thinking about that.  Good catch.  But yes we should definitely merge this anyway for now.

PR #4390, opened 2021-11-04
Bug fix: Allow fetch-and-compile to compile sources with absolute paths

haltman-at:
This PR fixes a problem where fetch-and-compile would fail to compile sources that used absolute paths in their source paths.  This can't happen for sources in the single-file etherscan format, but it can happen for the other formats.  The problem was that, due to #4216, these sources would be filtered out by `collectSources`.

To address this, I added a `noTransform` option to `Compile.run()` (I also changed `language` to be a named rather than positional argument), which makes it so that `collectSources` will not filter the sources or rewrite their prefix (it'll still do its other usual transformations, of course).  I then made it so that `Compile.sources()` passes this option.  Note that I did *not* make any changes to compile-vyper; I'll discuss that more in a bit.

Now this may seem like a significant change, but actually, outside of tests, `Compile.sources()` is currently only used in fetch-and-compile, so it's not really a problem.  As for those tests, well, it turns out our test *of* that path transformation was using `Compile.sources()`, so I rewrote it to use `Compile.sourcesWithDependencies` instead.

So, compile-solidity will no longer apply the prefix transformation when using `Compile.sources()`.  However I left Vyper alone.  (Note that compile-vyper performs a different transformation than compile-solidity.)  This is basically because it's irrelevant at the moment.  First of all, compile-vyper's `sources()` isn't used anywhere.  But not only that, even once we add Vyper support to fetch-and-compile, it still won't immediately matter, because currently Sourcify doesn't support Vyper, and Etherscan only supports Vyper in single-source form, meaning there are no imports to worry about, so this sort of path-rewriting shouldn't have any effect.

Also, the transformation in compile-vyper (which predates the one in compile-solidity!) is mostly intended to make certain imports work (see #3720, sorry for the poor quality of the writing there); so while not changing it risks eventually breaking things (in the distant future :P ), changing it *also* risks breaking things in that distant future.  So I feel like this is a problem for then rather than now.

Anyway fetch-and-compile, including `truffle debug -x`, should now be fixed for these Solidity contracts!

### DIFF
--- a/packages/compile-solidity/index.js
+++ b/packages/compile-solidity/index.js
@@ -20,7 +20,11 @@ async function compileYulPaths(yulPaths, options) {
     //for simplicity, since there are no imports to worry about)
     const yulSource = fs.readFileSync(path, { encoding: "utf8" });
     debug("Compiling Yul");
-    const compilation = await run({ [path]: yulSource }, yulOptions, "Yul");
+    const compilation = await run(
+      { [path]: yulSource },
+      yulOptions,
+      { language: "Yul" }
+    );
     debug("Yul compiled successfully");
 
     // returns CompilerResult - see @truffle/compile-common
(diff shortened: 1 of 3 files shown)


gnidan (approved):
Looks good! Thanks @haltman-at!

cds-amal:
This is a really nice solution @haltman-at !

PR #4393, opened 2021-11-04
Update link for Ganache to anticipate v7 release

eggplantzzz:
~This should be merged after the release of Ganache v7. That release will no longer be called ganache-cli but rather just ganache.~ These repos have already been changed so this can be merged.


davidmurdoch:
The UI is actually at `ganache-ui` already.

davidmurdoch (approved):

PR #4397, opened 2021-11-04
[RFR] Enhancement prevent get block call for read only data method

msoni89:
Hello Team,

According to documents, it clearly mentioned that the **Calls** type method is free and no gas cost is needed.

image

We can see in the code **prepareCall** method of truffle/contract/execute.js is calling **detectNetwork** method which is calling getBlock for gas-limit which is an unnecessary extra call for **Calls** type method.

Maybe this is not the best solution but we can prevent those network calls but passing extra arguments which detect its **Calls** type or **Send** type methods.

Please let me know, is there any other way to suppress those calls because in our application we are initializing more than 30 contracts in parallel, and due to net_version and get_block adding extra overhead on the network.


gnidan:
Oooh, thanks for doing this @msoni89! Sorry to hear about all the unnecessary network requests in your project.

I don't think there's a way to prevent this right now, so you're right to suggest a code change. We'll review this in the next few business days; with luck maybe we can get it released late next week for you.

packages/contract/lib/execute.js
```diff
@@ -140,7 +149,10 @@ const execute = {
       }
 
       execute
-        .prepareCall(constructor, methodABI, args)
+        .prepareCall(constructor, methodABI, args, {
+          id: constructor.network_id,
+          blockLimit: 0
```
eggplantzzz, line 154: Is `0` really a sensible default for this?

msoni89: @eggplantzzz Yes you are correct, it doesn't matter during calls types calls.

hellwolf: @eggplantzzz do you have a suggestion how should it look like?

hellwolf:
Looking forward to having this merged and released!

packages/contract/lib/execute.js
msoni89: @eggplantzzz Please review it again, I removed blockLimit parameter from **_detectNetworkArguments** which is not required during calls.

msoni89: @gnidan @eggplantzzz Please have a look.

hellwolf: ping

eggplantzzz:
Sorry for the delay on getting back to you. I think there may be a slightly better way to do this: perhaps by just making that last argument something like `isCall` (boolean) to decide whether to make the call to `detectNetwork`. In that case you could just add a `true` to the code flow for making a call. Let me think on it for a bit and get back to you in a couple of hours.

eggplantzzz:
So I think I like the solution I proposed in my earlier comment better. You can add a fourth argument `isCall` and pass `true` from the call to `prepareCall` in `call`. Not necessary but you could also make all of these arguments named in an object as well and that would make me happy. Are you up for it?

haltman-at (commented):
I definitely agree with @eggplantzzz that the current solution, using an object, is awkward; I'm not seeing any reason for the argument to be an object.

One possible solution is @eggplantzzz's idea where you pass a boolean, is one way; another way is to simply pass the number as it is, rather than passing the number inside an object.  With either way, when you're preparing something other than a call, you can simply omit the argument, rather than passing `{}`.

packages/contract/lib/execute.js
```diff
@@ -87,7 +93,10 @@ const execute = {
       args = processedValues.args;
       params = processedValues.params;
     }
-
+    const { id } = _detectNetworkArguments;
+    if (id) {
```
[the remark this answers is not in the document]
msoni89: @haltman-at I would like to go with @eggplantzzz solution.

@haltman-at and @eggplantzzz thank you for the suggestions.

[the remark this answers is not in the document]
msoni89: @haltman-at @eggplantzzz pushed changes, please review it again.

packages/contract/lib/execute.js
```diff
@@ -183,7 +187,7 @@ const execute = {
       const promiEvent = new PromiEvent(false, constructor.debugger);
 
       execute
-        .prepareCall(constructor, methodABI, arguments)
+        .prepareCall(constructor, methodABI, arguments, false)
```
eggplantzzz, line 190: Don't even pass `false`, just leave the fourth argument off. Then in `prepareCall` you can do something 
```
if (isCall) {
  return { args, params };
}
const network = await constructor...
```
You know what I mean?

packages/contract/lib/execute.js
```diff
@@ -87,7 +88,9 @@ const execute = {
       args = processedValues.args;
       params = processedValues.params;
     }
-
+    if (isCall) {
```
eggplantzzz, line 91: Sorry, one more nitpick; would you mind leaving a comment here about not needing to make a call for the network info in the case of making a "call"? Add that and then I am good to go on this! Thanks for doing this work @msoni89!

haltman-at: Alternative suggestion: Perhaps instead of `isCall`, the variable could be called something like `useCachedNetworkId`?  In that case, the comment would not go here, but rather where `true` is passed for it.  (Either way works I think.)  But yes, thanks for doing this!

msoni89:
Hello,
Thank you @eggplantzzz @haltman-at  for the review and If possible could you please also check .at() method implementation. Maybe I am not correct or understanding it wrongly. 

From .at() method, we are calling detectNetwork() which can be reduced or ignored by setting or flag then it will be really helpful and will reduce more network calls.

For the single contract, it won't a problem but for multiple contracts, it's an overhead on the network. I think gas/code will not change for first contact and n contracts.

Thanks in advance.

eggplantzzz (approved):
I like it! We'll try and get this merged so it can be released by the end of the week.

haltman-at (approved):
Thanks for this!

haltman-at:
> Hello, Thank you @eggplantzzz @haltman-at for the review and If possible could you please also check .at() method implementation. Maybe I am not correct or understanding it wrongly.
> 
> From .at() method, we are calling detectNetwork() which can be reduced or ignored by setting or flag then it will be really helpful and will reduce more network calls.
> 
> For the single contract, it won't a problem but for multiple contracts, it's an overhead on the network. I think gas/code will not change for first contact and n contracts.

Hm, maybe you could open an issue for that?  And possibly submit a PR for it too, if you want!

eggplantzzz:
Thanks again @msoni89!

PR #4399, opened 2021-11-05
Bug fix: Fix decoding of constant `bytesN` initialized as string literal

haltman-at:
When writing the code for decoding constants of type `bytesN`, I assumed that these would be initialized via hex numeric literals.  I didn't learn until later that string/hexstring literals can also be used as `bytesN`, and it turns out the code I wrote doesn't work for those.

Note that this problem didn't only apply to constants, but also applies to mapping keys of type `bytesN`, if you used a string/hexstring literal as a mapping key to a mapping which has `bytesN` for its key type.  Those wouldn't work properly, and this PR fixes that.  (It also affected constants used as mapping keys.)

So, I fixed that decoding to be a little more general.  I also expanded our testing of constants, of literals used as mapping keys, and constants used as mapping keys, to make sure this all works properly.


gnidan (approved):
Nice!

cds-amal (approved):
Nice @haltman-at !

PR #4400, opened 2021-11-05
Accept url and provider arguments in hdwallet-provider constructor

eggplantzzz:
In order to make hdwallet-provider a bit more slick, accept the arguments `provider` and `url` in the constructor.

### DIFF
--- a/packages/hdwallet-provider/test/provider.test.ts
+++ b/packages/hdwallet-provider/test/provider.test.ts
@@ -208,20 +208,6 @@ describe("HD Wallet Provider", function () {
       assert(number === 0);
     });
 
-    it("throws on invalid mnemonic", () => {
-      try {
-        provider = new HDWalletProvider({
-          mnemonic: {
-            phrase: "I am not a crook"
-          },
-          providerOrUrl: "http://localhost:8545"
-        });
-        assert.fail("Should throw on invalid mnemonic");
-      } catch (e) {
-        assert(e.message.includes("Mnemonic invalid or undefined"));
-      }
-    });
-
     it("provides for a default polling interval", () => {
       const mnemonicPhrase =
         "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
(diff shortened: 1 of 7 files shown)


cds-amal (commented):
Introducing separate options to capture the individual components of `providerOrUrl` seems redundant. I wonder what functionality/use case is added that isn't possible right now.

gnidan:
> I wonder what functionality/use case is added that isn't possible right now.

I'd argue that it's not about functionality here; it's about building interfaces that are nice to use.

I guess that use case is: "As a user, I should be able to construct an instance of HDWalletProvider by specifying either `provider` or by specifying `url`".

eggplantzzz:
I think @gnidan wanted to make `provider` and `url` the "primary" way of providing these options as it is nicer. Additionally we can leave `providerOrUrl` for backwards compatibility and remove it from the docs.

cds-amal:
Oh, I see. It is a nicer interface. Is it worth starting the deprecation process for **providerOrUrl**?

eggplantzzz:
Do we really even need to do anything about that @cds-amal? We could just remove it from the docs.

cds-amal:
It feels weird that we have a richer interface for `provider` and `url` but still maintain logic for `providerOrUrl`. Deprecating would allow us to remove `providerOrUrl` config support. 

Of course we'll have to update our tests, boxes and other scaffolding to reflect the newer config either way

packages/hdwallet-provider/test/provider.test.ts
```diff
@@ -304,5 +290,50 @@ describe("HD Wallet Provider", function () {
       const number = await web3.eth.getBlockNumber();
       assert(number === 0);
     });
+
```
cds-amal, line 293: I think the [provider tests setups](https://github.com/trufflesuite/truffle/blob/45e3dbe3738699997d6f0985e352ccdec6477560/packages/hdwallet-provider/test/provider.test.ts#L143) should use `url` instead of `providerOrUrl` to reflect the newer interface as tests are used to document usage.

cds-amal: There are a few other spots where `providerOrUrl` is used when `url` would suffice.

cds-amal:
Based on @gnidan's feedback depreciation shouldn't be considered on this PR. Didn't know positional arguments!

packages/hdwallet-provider/test/provider.test.ts
eggplantzzz: good idea, changed them in the test!

cds-amal (approved):
Looks good! Let's get it in.

PR #4402, opened 2021-11-05
Removed workingDirectory as it no longer exists

kevinweaver:
## Problem
The docs for truffle db ConnectOptions include a `workingDirectory: string;` attribute which is no longer valid.

## Solution
Remove it.


gnidan (approved):
thanks!

PR #4403, opened 2021-11-07
Run tests against Node.js v16 (LTS) in CI

chenrui333:
👋  Since nodejs 16 just became the latest LTS version, let's also add it into the CI checks. Thanks!

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
-        node-version: [12.x, 14.x]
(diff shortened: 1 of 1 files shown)


gnidan:
Thanks for this @chenrui333!

I'll bet the Node v16 tests are hanging because NPM v7 is known to cause issues with Truffle. @cds-amal anything we can do about this? I guess for this PR, we could figure out how to tell GHA to use NPM v6 explicitly.

cds-amal:
We can adjust GHA to fallback to **npm v6** for **Node 16** matrix targets. I'll take a look at this tomorrow.

eggplantzzz:
Closing this, we will open another PR for this when we figure out more info concerning Node 16 and Truffle.

Issue #4353, opened 2021-10-07
Debugger: Request for displaying Opcodes as a group


cds-amal:
Currently the debugger shows the current opcode when stepping with `;`.  I would like to see the opcodes grouped for more insights. I'm not sure how this could be displayed, perhaps `p +instructions` ?

haltman-at:
I don't know about `p instructions` (`p ins`?).  If we really want it to be a special thing it could be `p cod` or something.  But I think I'd just change the default opcode printout to show a display of, say, 3 behind and 3 ahead, in addition to the current one, much like our default view of the source code doesn't just show the particular line we're on.

If that's not enough we could maybe add something for seeing the full disassembly, but I'd start by just expanding what we show by default.

gnidan:
This is done and now released in [Truffle v5.4.20](https://github.com/trufflesuite/truffle/releases/tag/v5.4.20). Thanks!

PR #4412, opened 2021-11-09
Allow Truffle debugger to display opcodes as a group

dongmingh:
This PR is to address the request in the issue #4353 to allow
Truffle debugger to display opcodes as a group.  The debugger
will now display 3 previous instructions and 3 upcoming instructions
in addition to current instruction.

### DIFF
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -184,6 +184,7 @@ class DebugPrinter {
 
   printInstruction(locations = this.locationPrintouts) {
     const instruction = this.session.view(solidity.current.instruction);
+    const instructions = this.session.view(solidity.current.instructions);
     const step = this.session.view(trace.step);
     const traceIndex = this.session.view(trace.index);
     const totalSteps = this.session.view(trace.steps).length;
@@ -209,10 +210,33 @@ class DebugPrinter {
       this.config.logger.log(DebugUtils.formatStack(step.stack));
       this.config.logger.log("");
     }
-    this.config.logger.log(
-      DebugUtils.formatInstruction(traceIndex + 1, totalSteps, instruction)
-    );
-    this.config.logger.log(DebugUtils.formatPC(step.pc));
+
+    // printout instructions
+    const previousInstructions = 3;
+    const upcomingInstructions = 3;
+    const currentIndex = instruction.index;
+    // printout 3 previous instructions
+    if (currentIndex - previousInstructions + 1 > 0) {
+      this.config.logger.log("...");
+    }
+    for (let i = currentIndex - previousInstructions; i < currentIndex; i++) {
+      if (i >= 0) {
+        this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
+      }
+    }
+
+    // printout current instruction
+    this.config.logger.log(DebugUtils.formatCurrentInstruction(instruction));
+
+    // printout 3 upcoming instructions
+    for (let i = currentIndex + 1; i <= currentIndex + upcomingInstructions; i++) {
+      if (i >= instructions.length) break;
+      this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
+    }
+    if (currentIndex + upcomingInstructions < instructions.length) {
+      this.config.logger.log("...");
+    }
+
     this.config.logger.log("");
     this.config.logger.log(step.gas + " gas remaining");
   }
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -212,14 +212,13 @@ class DebugPrinter {
     }
 
     this.config.logger.log("Instructions:");
-    this.config.logger.log("");
 
     // printout instructions
-    // add an ellipse if there exist additional instructions before
     const previousInstructions = 3;
     const upcomingInstructions = 3;
     const currentIndex = instruction.index;
 
+    // add an ellipse if there exist additional instructions before
     if (currentIndex - previousInstructions >= 0) {
       this.config.logger.log("...");
     }
@@ -246,7 +245,12 @@ class DebugPrinter {
     }
 
     this.config.logger.log("");
-    this.config.logger.log("Step ", + traceIndex + "/" + totalSteps);
+    this.config.logger.log(
+      "Step " +
+      (traceIndex + 1).toString() +
+      "/" +
+      totalSteps.toString()
+    );
     this.config.logger.log(step.gas + " gas remaining");
   }
 
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -215,7 +215,7 @@ class DebugPrinter {
     this.config.logger.log("");
 
     // printout instructions
-    // add an ellipse if the first instruction of the stack will not be included in the printout
+    // add an ellipse if there exist additional instructions before
     const previousInstructions = 3;
     const upcomingInstructions = 3;
     const currentIndex = instruction.index;
@@ -240,7 +240,7 @@ class DebugPrinter {
       this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
     }
 
-    // add an ellipse if the last instruction of the stack is not included in the printout
+    // add an ellipse if there exist additional instructions after
     if (currentIndex + upcomingInstructions < instructions.length) {
       this.config.logger.log("...");
     }
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -211,33 +211,42 @@ class DebugPrinter {
       this.config.logger.log("");
     }
 
+    this.config.logger.log("Instructions:");
+    this.config.logger.log("");
+
     // printout instructions
+    // add an ellipse if the first instruction of the stack will not be included in the printout
     const previousInstructions = 3;
     const upcomingInstructions = 3;
     const currentIndex = instruction.index;
-    // printout 3 previous instructions
-    if (currentIndex - previousInstructions + 1 > 0) {
+
+    if (currentIndex - previousInstructions >= 0) {
       this.config.logger.log("...");
     }
-    for (let i = currentIndex - previousInstructions; i < currentIndex; i++) {
-      if (i >= 0) {
-        this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
-      }
+    // printout 3 previous instructions
+    for (let i = Math.max(currentIndex - previousInstructions, 0);
+      i < currentIndex;
+      i++) {
+      this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
     }
 
     // printout current instruction
     this.config.logger.log(DebugUtils.formatCurrentInstruction(instruction));
 
     // printout 3 upcoming instructions
-    for (let i = currentIndex + 1; i <= currentIndex + upcomingInstructions; i++) {
-      if (i >= instructions.length) break;
+    for (let i = Math.min(currentIndex + 1, instructions.length);
+      i <= currentIndex + upcomingInstructions;
+      i++) {
       this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
     }
+
+    // add an ellipse if the last instruction of the stack is not included in the printout
     if (currentIndex + upcomingInstructions < instructions.length) {
       this.config.logger.log("...");
     }
 
     this.config.logger.log("");
+    this.config.logger.log("Step ", + traceIndex + "/" + totalSteps);
     this.config.logger.log(step.gas + " gas remaining");
   }
 
--- a/packages/debug-utils/index.js
+++ b/packages/debug-utils/index.js
@@ -530,13 +530,16 @@ var DebugUtils = {
     }
   },
 
-  formatInstruction: function (traceIndex, traceLength, instruction) {
+  formatCurrentInstruction: function (instruction) {
+    const pc = this.formatPC(instruction.pc);
+    const formattedInstruction = this.formatInstruction(instruction);
     return (
-      "(" +
-      traceIndex +
-      "/" +
-      traceLength +
-      ") " +
+      truffleColors.mint("-> " + formattedInstruction + pc )
+    );
+  },
+
+  formatInstruction: function (instruction) {
+    return (
       truffleColors.mint(instruction.name + " " + (instruction.pushData || ""))
     );
   },
@@ -546,7 +549,7 @@ var DebugUtils = {
     if (hex.length % 2 !== 0) {
       hex = "0" + hex; //ensure even length
     }
-    return "  PC = " + pc.toString() + " = 0x" + hex;
+    return " (PC=" + pc.toString() + ", 0x" + hex+")";
   },
 
   formatStack: function (stack) {
--- a/packages/debug-utils/index.js
+++ b/packages/debug-utils/index.js
@@ -534,7 +534,7 @@ var DebugUtils = {
     const pc = this.formatPC(instruction.pc);
     const formattedInstruction = this.formatInstruction(instruction);
     return (
-      truffleColors.mint("-> " + formattedInstruction + pc )
+      "-> " + truffleColors.mint(formattedInstruction) + pc
     );
   },
 


packages/core/lib/debug/printer.js
```diff
@@ -209,10 +209,33 @@ class DebugPrinter {
...
     }
-    this.config.logger.log(
-      DebugUtils.formatInstruction(traceIndex + 1, totalSteps, instruction)
-    );
-    this.config.logger.log(DebugUtils.formatPC(step.pc));
+
+    // printout instructions
+    const previousInstructions = 3;
+    const upcomingInstructions = 3;
+    const currentInstructionIndex = instructions.findIndex(instruction => instruction.pc === step.pc);
```
haltman-at, line 216: You don't need to iterate over `instructions` to find the current index!  It's already stored in the `index` field in `solidity.current.instruction`.  This is why I pointed that out to you earlier that you want to use both selectors; you can do this with a constant-time access, you don't need a linear-time search.

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +209,33 @@ class DebugPrinter {
...
+    // printout 3 previous instructions
+    if (currentInstructionIndex-previousInstructions+1 > 0) {
```
haltman-at, line 218: Company style nitpick: Style here is to put spaces around operators. This doesn't really matter, obviuosly; it's purely a consistency thing.  But, that is the style here.

Sorry, I realize this is another thing that you shouldn't really have to worry about. :-/  We used to have `prettier` to handle this for us...  hopefully that's fixed soon.  You'll have to ask @cds-amal what the status of that is.

Btw, I know I said we try to avoid abbreviating variable names here, but if you're doing this a bunch of arithmetic with them, it might help to find *some* suitable way of abbreviating them.  (Like, `currentInstructionIndex` could maybe just be `currentIndex`?  Or something like that?)  Idk, it's a balance, I guess.  Sorry, I realize that's not exactly clear unambiguous instructions, but yeah.  Generally you should err on the side of longer names, but for cases like this it might make sense to pull back on that a little.

packages/debug-utils/index.js
```diff
@@ -530,13 +530,16 @@ var DebugUtils = {
     }
   },
 
-  formatInstruction: function (traceIndex, traceLength, instruction) {
+  formatCurrentInstruction: function (instruction) {
+    const pc = this.formatPC(instruction.pc);
+    const fmtInstr = this.formatInstruction(instruction);
```
haltman-at, line 535: This, by contrast, is the sort of overly-abbreviated variable name I would warn against.  I'd suggest expanding this to `formattedInstruction`; `fmtInstr` isn't really clear.

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +209,33 @@ class DebugPrinter {
...
+      this.config.logger.log("...");
+    }
+    for (let i = currentInstructionIndex-previousInstructions; i<currentInstructionIndex; i++) {
```
haltman-at, line 221: Spaces again.

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +209,33 @@ class DebugPrinter {
...
+      if (i >= 0) {
+        this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
+      }
+    }
+
+    // printout current instruction
+    this.config.logger.log(DebugUtils.formatCurrentInstruction(instructions[currentInstructionIndex]));
+
+    // printout 3 upcoming instructions
+    for (let i = currentInstructionIndex+1; i<currentInstructionIndex+upcomingInstructions+1; i++) {
```
haltman-at, line 231: Spaces again.

haltman-at (changes requested):
Hi @dongmingh, this is just a partial review for now, hope you don't mind; I'll come back and do a full review later after trying it out.

Most of this review is style nitpicks that aren't really important but which you should change for consistency.  Sorry.  (OK, one is a style nitpick that I think you should change for clarity, and which is therefore a bit more important.)

There is one actually important thing that I think needs to be changed, however, which is why I've hit "request changes" on this... you should fix the way you get the current instruction index.  I think that needs to be fixed before this can be merged, so I've marked this one as "request changes".  But that's the only truly important one here.

Anyway sorry this is just a partial review, but like I said I'll come back and write more after I've had a chance to actually try it out.  Thanks for picking this up!

packages/core/lib/debug/printer.js
dongmingh: Thanks for pointing this out.  Change to use the index field of solidity.current.instruction.

packages/core/lib/debug/printer.js
dongmingh: add spaces around operators, and rename currentInstructionIndex to currentIndex, thx

packages/core/lib/debug/printer.js
dongmingh: added spaces

packages/core/lib/debug/printer.js
dongmingh: added spaces, thx

packages/debug-utils/index.js
dongmingh: rename to formattedInstruction, thx

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +210,33 @@ class DebugPrinter {
...
-    this.config.logger.log(
-      DebugUtils.formatInstruction(traceIndex + 1, totalSteps, instruction)
-    );
-    this.config.logger.log(DebugUtils.formatPC(step.pc));
+
+    // printout instructions
+    const previousInstructions = 3;
+    const upcomingInstructions = 3;
+    const currentIndex = instruction.index;
+    // printout 3 previous instructions
```
cds-amal, line 218: A very minor nit, I think this comment should explain the the logic for printing the ellipse.

haltman-at:
Thanks for changing this!  This looks basically good, but I still need to try it out before approving it.

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +210,33 @@ class DebugPrinter {
...
+
+    // printout current instruction
+    this.config.logger.log(DebugUtils.formatCurrentInstruction(instruction));
+
+    // printout 3 upcoming instructions
+    for (let i = currentIndex + 1; i <= currentIndex + upcomingInstructions; i++) {
+      if (i >= instructions.length) break;
+      this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
+    }
+    if (currentIndex + upcomingInstructions < instructions.length) {
```
cds-amal, line 236: same minor nit.

cds-amal (approved):
Thanks for adding this functionality @dongmingh ! I'm already using it :) I left a minor minor nit, feel free to ignore.

haltman-at:
By the way @dongmingh, I notice you seem to always rebase and force-push.  This makes it harder to keep track of.  I would suggest that more often you should simply add new commits and do a normal push.  Obviously there are some times where that's not tenable, but when it is tenable, it keeps the history clearer.

packages/core/lib/debug/printer.js
haltman-at: Btw, note that if you did need to look it up by `pc` rather than by `index`, you could use the `solidity.current.instructionAtProgramCounter` selector.  But obviously there isn't actually a need for that here. :)

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +210,33 @@ class DebugPrinter {
       this.config.logger.log(DebugUtils.formatStack(step.stack));
       this.config.logger.log("");
     }
-    this.config.logger.log(
-      DebugUtils.formatInstruction(traceIndex + 1, totalSteps, instruction)
-    );
-    this.config.logger.log(DebugUtils.formatPC(step.pc));
+
+    // printout instructions
+    const previousInstructions = 3;
```
haltman-at, line 215: I was worried this would be too much, but having tried it out, I think this is actually good.  Whew!

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +210,33 @@ class DebugPrinter {
...
+    const upcomingInstructions = 3;
+    const currentIndex = instruction.index;
+    // printout 3 previous instructions
+    if (currentIndex - previousInstructions + 1 > 0) {
```
haltman-at, line 219: Style comment: I think this inequality is a little unclear.  I'd suggest rewriting it to `if (currentIndex - previousInstructions >= 0`.)  I also agre with Amal that it's worth adding a comment explaining what this inequality is doing.

packages/core/lib/debug/printer.js
haltman-at: Agreed, this could use a comment explaining what it's doing.

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +210,33 @@ class DebugPrinter {
...
+        this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
+      }
+    }
+
+    // printout current instruction
+    this.config.logger.log(DebugUtils.formatCurrentInstruction(instruction));
+
+    // printout 3 upcoming instructions
+    for (let i = currentIndex + 1; i <= currentIndex + upcomingInstructions; i++) {
+      if (i >= instructions.length) break;
```
haltman-at, line 233: Really minor thing, but instead of using `break`, you can use `Math.min`. :)  Feel free to ignore this, the current way is also good; just pointing out the opportunity if you want to make this cleaner. :)

packages/debug-utils/index.js
```diff
@@ -530,13 +530,16 @@ var DebugUtils = {
     }
   },
 
-  formatInstruction: function (traceIndex, traceLength, instruction) {
+  formatCurrentInstruction: function (instruction) {
+    const pc = this.formatPC(instruction.pc);
+    const formattedInstruction = this.formatInstruction(instruction);
+    return (
+      truffleColors.mint("-> " + formattedInstruction + pc )
```
haltman-at, line 537: So, having tried this... I don't think we want to apply the mint color to the whole thing.  I think the PC value should be left uncolored.

I'm not sure what to do about the arrow.  Probably make it a different color than the mint to make it stand out?

packages/debug-utils/index.js
```diff
@@ -530,13 +530,16 @@ var DebugUtils = {
...
+    );
+  },
+
+  formatInstruction: function (instruction) {
```
haltman-at, line 541: Note: This might be technically a breaking change to `debug-utils`... however that's fine, it's purely an internal package, so feel free to break it as much as you want. :)  It only matters when it comes time to do a release and you have to bump the version.

haltman-at (changes requested):
OK, I have one thing I think definitely needs to change, one or two things I think should probably change, and a few style comments.

The thing that definitely needs to change is that the step counter has vanished.  That needs to be put back in.  That said it doesn't need to be grouped with the instructions view... maybe it would make more sense to group it with the gas display?

Also, when I look at this display in the debugger, I think that with the changes made, this section would be clearer with some sort of header; I'd suggest adding that.  Not sure what it would be called though.  Anyway yeah, that and the color thing I mention below should probably change.

And then I have a few style nitpicks as always, although they're basically just things Amal has already said!

packages/core/lib/debug/printer.js
dongmingh: add a comment:
```
// add an ellipse if the first instruction of the stack will not be included in the printout
```

packages/core/lib/debug/printer.js
dongmingh: add a comment:
```
// add an ellipse if the last instruction of the stack is not included in the printout
```

packages/core/lib/debug/printer.js
```diff
@@ -209,10 +210,33 @@ class DebugPrinter {
...
+    // printout instructions
+    const previousInstructions = 3;
+    const upcomingInstructions = 3;
+    const currentIndex = instruction.index;
+    // printout 3 previous instructions
+    if (currentIndex - previousInstructions + 1 > 0) {
+      this.config.logger.log("...");
+    }
+    for (let i = currentIndex - previousInstructions; i < currentIndex; i++) {
+      if (i >= 0) {
```
haltman-at, line 223: Sorry, one more comment -- similar to the `Math.min` thing, I would suggest replacing this `if` with a `Math.max` in the bound, so you don't have to use a `for/if` structure.

packages/core/lib/debug/printer.js
dongmingh: changes made

packages/core/lib/debug/printer.js
dongmingh: thx for trying it out

cds-amal:
> By the way @dongmingh, I notice you seem to always rebase and force-push. This makes it harder to keep track of. I would suggest that more often you should simply add new commits and do a normal push. Obviously there are some times where that's not tenable, but when it is tenable, it keeps the history clearer.

I also had a hard time following the conversation over time because of the force pushes. We should explain this in our dev-handbook. I'll post a PR.

packages/core/lib/debug/printer.js
haltman-at: That's not the clearest, but it'll do, I suppose.  (Instructions don't go on the stack, and it would make more sense to emphasize that there *exist* additional instructions before, rather than that the first one *isn't* present, but whatever, ultimately it gets the point across.)

packages/core/lib/debug/printer.js
haltman-at: That's not the clearest, but it'll do, I suppose. (Same comment as above.)

packages/core/lib/debug/printer.js
haltman-at: (Sorry, normally I wouldn't nitpick comments like this!  Just it seems like you could use some help with your prose writing, so I thought I'd point out how it could be written clearer. :) )

packages/debug-utils/index.js
dongmingh: change to
```
"-> " + truffleColors.mint(formattedInstruction) + pc
```

packages/core/lib/debug/printer.js
dongmingh: changes made

packages/core/lib/debug/printer.js
dongmingh: > > By the way @dongmingh, I notice you seem to always rebase and force-push. This makes it harder to keep track of. I would suggest that more often you should simply add new commits and do a normal push. Obviously there are some times where that's not tenable, but when it is tenable, it keeps the history clearer.
> 
> I also had a hard time following the conversation over time because of the force pushes. We should explain this in our dev-handbook. I'll post a PR.

I just git push the PR with no force.  Thanks.

packages/core/lib/debug/printer.js
dongmingh: changes made

packages/core/lib/debug/printer.js
dongmingh: changes made

packages/core/lib/debug/printer.js
```diff
@@ -209,11 +210,43 @@ class DebugPrinter {
...
-    this.config.logger.log(
-      DebugUtils.formatInstruction(traceIndex + 1, totalSteps, instruction)
-    );
-    this.config.logger.log(DebugUtils.formatPC(step.pc));
+
+    this.config.logger.log("Instructions:");
+    this.config.logger.log("");
+
+    // printout instructions
+    // add an ellipse if there exist additional instructions before
```
cds-amal, line 218: This comment should be moved to before the conditional that displays the eclipse

packages/core/lib/debug/printer.js
```diff
@@ -209,11 +210,43 @@ class DebugPrinter {
...
+      i < currentIndex;
+      i++) {
+      this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
+    }
+
+    // printout current instruction
+    this.config.logger.log(DebugUtils.formatCurrentInstruction(instruction));
+
+    // printout 3 upcoming instructions
+    for (let i = Math.min(currentIndex + 1, instructions.length);
```
cds-amal, line 237: This is neat!

cds-amal (commented):
Minor comment placement nit.

packages/core/lib/debug/printer.js
dongmingh: good catch, thx

packages/core/lib/debug/printer.js
dongmingh: Thanks Harry for this suggestion

packages/core/lib/debug/printer.js
```diff
@@ -209,11 +210,43 @@ class DebugPrinter {
...
+      i <= currentIndex + upcomingInstructions;
+      i++) {
```
haltman-at, line 239: This is definitely not the style `prettier` would use, but, to hell with it, I'm not going to hold things up over that!  It shouldn't be your job to deal with that anyway; we need to get Husky fixed and then you won't ever have to worry about this again! (@cds-amal)

(This applies to the other `for` loop as well, obviously.)

packages/debug-utils/index.js
```diff
@@ -530,13 +530,16 @@ var DebugUtils = {
     }
   },
 
-  formatInstruction: function (traceIndex, traceLength, instruction) {
+  formatCurrentInstruction: function (instruction) {
+    const pc = this.formatPC(instruction.pc);
+    const formattedInstruction = this.formatInstruction(instruction);
+    return (
+      "-> " + truffleColors.mint(formattedInstruction) + pc
```
haltman-at, line 537: Ah, I see you went with no color on the arrow.

haltman-at (approved):
OK, seems good to me!

packages/core/lib/debug/printer.js
```diff
@@ -209,11 +210,43 @@ class DebugPrinter {
       this.config.logger.log(DebugUtils.formatStack(step.stack));
       this.config.logger.log("");
     }
-    this.config.logger.log(
-      DebugUtils.formatInstruction(traceIndex + 1, totalSteps, instruction)
-    );
-    this.config.logger.log(DebugUtils.formatPC(step.pc));
+
+    this.config.logger.log("Instructions:");
+    this.config.logger.log("");
```
haltman-at, line 215: Ah, sorry, one last thing before merging -- could you get rid of this extra blank line?  I don't think we need this, I think we want it more compact.  Thanks!

packages/core/lib/debug/printer.js
```diff
@@ -209,11 +210,43 @@ class DebugPrinter {
...
+      this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
+    }
+
+    // add an ellipse if there exist additional instructions after
+    if (currentIndex + upcomingInstructions < instructions.length) {
+      this.config.logger.log("...");
+    }
+
     this.config.logger.log("");
+    this.config.logger.log("Step ", + traceIndex + "/" + totalSteps);
```
haltman-at, line 249: Sorry, just noticed two problems here!

1. Bigger problem: You forgot to add 1 to the traceIndex before printing.  The traceIndex is 0-indexed, but we print it out 1-indexed.  You'll notice the old code adds 1 to it; the new code should do the same.
2. Smaller problem: You have a stray comma in here that's causing an extra space to appear in your output.

Btw, in new code we generally use template strings rather than manual string concatenation.  I won't insist on that though... there's probably a lot of manual string concatenation around here already so it's hardly a big deal. :P

haltman-at (changes requested):
Blargh, sorry, one final thing!  You seem to have accidentally altered the index by 1, that needs to be fixed!

packages/core/lib/debug/printer.js
dongmingh: changes made

packages/core/lib/debug/printer.js
dongmingh: thanks for catching these.  Changes made.

packages/debug-utils/index.js
dongmingh: I tried different colors such as red, purple etc using truffleColors().  At the end, I chose no color.  Please feel free to suggest another color.

packages/core/lib/debug/printer.js
dongmingh: Ok, let's leave this for now and wait for Husky to work and fix all.

packages/core/lib/debug/printer.js
gnidan: Yeah, for reference, you can do this:
```javascript
console.log(`This is a ${"template"} string`);
```

haltman-at: For a more illustrative example, in this case one could do `Step ${traceIndex + 1}/${totalSteps}`.

haltman-at (approved):
OK, I think this is finally good to go!  Thanks Dongming!

PR #4417, opened 2021-11-10
Rename event more appropriately after its respective command

eggplantzzz:
This event was accidentally named after the incorrect command flow. While it was named "migrate:skipped", it really should have been named "test:**" as it is part of the flow for `truffle test` and not `truffle migrate`.

### DIFF
--- a/packages/core/lib/testing/Test.js
+++ b/packages/core/lib/testing/Test.js
@@ -108,7 +108,7 @@ const Test = {
 
     if (config.migrateNone || config['migrate-none']) {
       if (config.events) {
-        config.events.emit("migrate:skipped");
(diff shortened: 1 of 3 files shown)


dongmingh (approved):
look good to me

cds-amal (commented):
[defaultSubscriber/index.js](https://github.com/trufflesuite/truffle/blob/e3feb095628ef2d0c6f41655f3e8d759090e3a24/packages/events/defaultSubscribers/index.js#L3) needs to be updated to reflect file rename

haltman-at (approved):

cds-amal (approved):
Looks good!

Issue #4384, opened 2021-11-02
Resolver.resolve should have access to compiler/version information for current compilation


gnidan:
## Background

Basics:
- [abi-to-sol](https://www.npmjs.com/package/abi-to-sol) is a package that takes ABI JSON as input and produces Solidity code that compiles to that ABI JSON. It allows you to specify a Solidity version range, and will produce output that is compatible with that specified range.
- Truffle provides "experimental" support for making this super easy: you can have `.abi.json` files in your `contracts/` directory, import them in Solidity like `import "./Uniswap.abi.json";`, and it will be as-if you imported `interface Uniswap` with all the functions, etc., defined by that ABI.

Now, my [recent attempt to upgrade abi-to-sol](https://github.com/trufflesuite/truffle/pull/4365) fails CI because later versions of this dependency produce output that is compatible with only a narrow range of Solidity versions (default `>=0.7.0 <0.9.0`). Previously, this wasn't a problem our current version of that dependency does not change its output syntax at all, and the pragma it produced was overly broad.

abi-to-sol has gotten a lot more robust, in that it now produces valid Solidity output much, much more reliably, and so we could probably make this Truffle feature not-so-experimental, if only we could upgrade.

For the Truffle feature, we invoke abi-to-sol during `resolver.resolve()`, which in turn happens via the profiler. If the user imports a `.abi.json` file in their Solidity code, `resolver.resolve()` will recognize the file extension, read the corresponding ABI JSON from disk, and then call abi-to-sol to produce a Solidity source instead of this raw JSON

Ideally, when the user does this, the produced Solidity source should match whatever pragma it was imported by, and abi-to-sol output should be compatible with this pragma. So if you're in a file that has `pragma solidity ^0.4.24 || 0.8.9;`, and you import an ABI JSON file, then Truffle should tell abi-to-sol that it needs to produce output that's also got `pragma solidity ^0.4.24 || 0.8.9;`.

## Issue

The issue is: `resolver.resolve()` doesn't know anything about the language, compiler version, settings, etc. for which it's invoked! So there's no way for this functionality to specify the correct version to abi-to-sol (as per the background above), since it doesn't have this info at all.

And since this codepath is run via the profiler, we also need the profiler to know this information. Right now, the profiler also does not get information of this kind.

So maybe there's a clean way for compile-solidity, etc. to invoke the profiler with this information? And then this information could get passed along to `resolver.resolve()`? Hopefully this is not just asking for a giant mess.

[...]

PR #4420, opened 2021-11-10
Pass compiler info to resolver sources

gnidan:
Seeks to address #4384 

This PR roughly does two things:
1. Ensure we pass the compiler info from compile-solidity all the way through the Profiler and ultimately to the ResolverSources, which may choose to handle this information.
2. Invoke the CompilerSupplier before we run the Profiler, since otherwise we won't have the information yet (currently, compile-solidity invokes the CompilerSupplier _after_ invoking the Profiler, so this PR changes that behavior)

I've tried to be very safe about this (i.e., by making all new arguments optional), but maybe this is overly cautious and we can afford to be looser about it.

### DIFF
--- a/packages/compile-solidity/src/compilerSupplier/index.ts
+++ b/packages/compile-solidity/src/compilerSupplier/index.ts
@@ -1,6 +1,3 @@
-import debugModule from "debug";
-const debug = debugModule("compile-solidity:compilerSupplier");
-
 import path from "path";
 import fs from "fs";
 import semver from "semver";
@@ -70,11 +67,6 @@ export class CompilerSupplier {
 
     if (strategy) {
       const solc = await strategy.load(userSpecification);
-      debug(
-        "Loaded solc v%s using %s strategy",
-        solc.version(),
-        strategy.constructor.name
-      );
       return { solc };
     } else {
       throw new BadInputError(userSpecification);
--- a/packages/compile-solidity/src/compilerSupplier/index.ts
+++ b/packages/compile-solidity/src/compilerSupplier/index.ts
@@ -1,3 +1,6 @@
+import debugModule from "debug";
+const debug = debugModule("compile-solidity:compilerSupplier");
+
 import path from "path";
 import fs from "fs";
 import semver from "semver";
@@ -67,6 +70,11 @@ export class CompilerSupplier {
 
     if (strategy) {
       const solc = await strategy.load(userSpecification);
+      debug(
+        "Loaded solc v%s using %s strategy",
+        solc.version(),
+        strategy.constructor.name
+      );
       return { solc };
     } else {
       throw new BadInputError(userSpecification);
--- a/packages/compile-solidity/src/index.js
+++ b/packages/compile-solidity/src/index.js
@@ -146,7 +146,11 @@ const Compile = {
       options.with({
         paths: solidityPaths,
         base_path: options.contracts_directory,
-        resolver: options.resolver
+        resolver: options.resolver,
+        compiler: {
+          name: "solc",
+          version: solc.version()
+        }
       })
     );
     debug("compilationTargets: %O", compilationTargets);
--- a/packages/compile-solidity/src/index.js
+++ b/packages/compile-solidity/src/index.js
@@ -129,6 +129,12 @@ const Compile = {
     options = Config.default().merge(options);
     options = normalizeOptions(options);
 
+    const supplier = new CompilerSupplier({
+      events: options.events,
+      solcConfig: options.compilers.solc
+    });
+    const { solc } = await supplier.load();
+
     //note: solidityPaths here still includes JSON as well!
     const [yulPaths, solidityPaths] = partition(paths, path =>
       path.endsWith(".yul")
@@ -161,7 +167,7 @@ const Compile = {
     if (Object.keys(allSources).length > 0) {
       const solidityOptions = options.with({ compilationTargets });
       debug("Compiling Solidity");
-      const compilation = await run(allSources, solidityOptions);
+      const compilation = await run(allSources, solidityOptions, { solc });
       debug("Solidity compiled successfully");
 
       // returns CompilerResult - see @truffle/compile-common
--- a/packages/compile-solidity/src/run.js
+++ b/packages/compile-solidity/src/run.js
@@ -12,8 +12,11 @@ async function run(rawSources, options, internalOptions = {}) {
     return null;
   }
 
-  const language = internalOptions.language || "Solidity"; //could also be "Yul"
-  const noTransform = internalOptions.noTransform; //turns off project root transform
+  const {
+    language = "Solidity", // could also be "Yul"
+    noTransform = false, // turns off project root transform
+    solc, // passing this skips compilerSupplier.load()
+  } = internalOptions;
 
   // Ensure sources have operating system independent paths
   // i.e., convert backslashes to forward slashes; things like C: are left intact.
@@ -42,6 +45,7 @@ async function run(rawSources, options, internalOptions = {}) {
   // perform compilation
   const { compilerOutput, solcVersion } = await invokeCompiler({
     compilerInput,
+    solc,
     options
   });
   debug("compilerOutput: %O", compilerOutput);
@@ -214,14 +218,18 @@ function prepareOutputSelection({ targets = [] }) {
 /**
  * Load solc and perform compilation
  */
-async function invokeCompiler({ compilerInput, options }) {
+async function invokeCompiler({ compilerInput, options, solc }) {
   const supplierOptions = {
     parser: options.parser,
     events: options.events,
     solcConfig: options.compilers.solc
   };
-  const supplier = new CompilerSupplier(supplierOptions);
-  const { solc } = await supplier.load();
+
+  if (!solc) {
+    const supplier = new CompilerSupplier(supplierOptions);
+    ({ solc } = await supplier.load());
+  }
+
   const solcVersion = solc.version();
 
   // perform compilation
--- a/packages/profiler/src/index.ts
+++ b/packages/profiler/src/index.ts
@@ -51,7 +51,8 @@ export class Profiler {
       resolver,
       paths,
       base_path: basePath,
-      contracts_directory: contractsDirectory
+      contracts_directory: contractsDirectory,
+      compiler // { name, version }
     } = options;
 
     debug("paths: %O", paths);
@@ -61,7 +62,7 @@ export class Profiler {
       //be resolved, it will show up as a compile error rather than a Truffle
       //error.
       try {
-        return await resolver.resolve(filePath, importedFrom);
+        return await resolver.resolve(filePath, importedFrom, { compiler });
       } catch (error) {
         //resolver doesn't throw structured errors at the moment,
         //so we'll check the messag to see whether this is an expected error
--- a/packages/profiler/src/index.ts
+++ b/packages/profiler/src/index.ts
@@ -95,10 +95,15 @@ export class Profiler {
   async requiredSourcesForSingleFile(options: TruffleConfig) {
     expect.options(options, ["path", "base_path", "resolver"]);
 
-    const { resolver, path, base_path: basePath } = options;
+    const {
+      resolver,
+      path,
+      base_path: basePath,
+      compiler // { name, version }
+    } = options;
 
     const resolve = ({ filePath, importedFrom }: UnresolvedSource) =>
-      resolver.resolve(filePath, importedFrom);
+      resolver.resolve(filePath, importedFrom, { compiler });
 
     const allPaths = convertToAbsolutePaths([path], basePath);
     const updatedPaths = allPaths;
--- a/packages/profiler/src/index.ts
+++ b/packages/profiler/src/index.ts
@@ -70,6 +70,11 @@ export class Profiler {
         if(error.message.startsWith("Could not find ")) {
           return undefined;
         } else {
+          debug(
+            "Error resolving filePath: %o via importedFrom: %s",
+            filePath,
+            importedFrom
+          );
           //rethrow unexpected errors
           throw error;
         }
--- a/packages/profiler/src/index.ts
+++ b/packages/profiler/src/index.ts
@@ -70,11 +70,6 @@ export class Profiler {
         if(error.message.startsWith("Could not find ")) {
           return undefined;
         } else {
-          debug(
-            "Error resolving filePath: %o via importedFrom: %s",
-            filePath,
-            importedFrom
-          );
           //rethrow unexpected errors
           throw error;
         }
--- a/packages/resolver/lib/resolver.ts
+++ b/packages/resolver/lib/resolver.ts
@@ -81,14 +81,23 @@ export class Resolver {
 
   async resolve(
     importPath: string,
-    importedFrom: string
+    importedFrom: string,
+    options: {
+      compiler?: {
+        name: string;
+        version: string;
+      }
+    } = {}
   ): Promise<ResolvedSource> {
     let body: string | null = null;
     let filePath: string | null = null;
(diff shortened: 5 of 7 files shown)


packages/profiler/src/index.ts
```diff
@@ -51,7 +51,8 @@ export class Profiler {
       resolver,
       paths,
       base_path: basePath,
-      contracts_directory: contractsDirectory
+      contracts_directory: contractsDirectory,
+      compiler // { name, version }
```
gnidan, line 55: So I'm not sure we want to call it this, since we already have `truffle compile --compiler=<solc|vyper|external>`. This shouldn't cause problems, but it might be weird? Thoughts?

packages/compile-solidity/src/run.js
```diff
@@ -214,14 +218,18 @@ function prepareOutputSelection({ targets = [] }) {
...
+async function invokeCompiler({ compilerInput, options, solc }) {
   const supplierOptions = {
     parser: options.parser,
     events: options.events,
     solcConfig: options.compilers.solc
   };
-  const supplier = new CompilerSupplier(supplierOptions);
-  const { solc } = await supplier.load();
+
+  if (!solc) {
```
gnidan, line 228: You'll see here that I've made this `solc` param optional, but maybe we shouldn't worry about breaking the interface to `run()`, since it's entirely internal. I did it this way just to be safe.

haltman-at: Yeah, there's no need to worry about breaking the interface to `run`... you may recall I also did so just recently!

packages/profiler/src/index.ts
haltman-at: Hm, tricky.  I agree it's not ideal to have overlap there. `currentCompiler`? `compilerSpecifier`?  IDK.

packages/compile-solidity/src/run.js
```diff
@@ -12,8 +12,11 @@ async function run(rawSources, options, internalOptions = {}) {
     return null;
   }
 
-  const language = internalOptions.language || "Solidity"; //could also be "Yul"
-  const noTransform = internalOptions.noTransform; //turns off project root transform
+  const {
```
haltman-at, line 15: Oh, I forgot you can do this.  Thanks for cleaning this up!

haltman-at (commented):
Hm, OK, I have a few comments on this.  I'm not going to tag these by line number because they're kind of general, and some of them relate to changes that were *not* made, so that's difficult.

1. Not a comment on the code, but -- I got the point of the new `compiler` field after reading the code, but this would have been easier had it been explained beforehand. :P  Actually, I'm still not entirely clear on why it was made an options field rather than something else; I have guesses as to why but it would be nice to know explicitly the reason.
2. I notice you didn't pass `solc` in the pragma case, when compiling Yul, or in `sources()`, leaving things the old way in those cases.  I guess it's not necessary to alter things there, but my inclination is that consistency is better.  Then again, maybe not -- it would be more consistent, but it would also be doing the load in more places, so maybe this way is better.
3. Speaking of that `compiler` field again -- I notice in some places you passed it as `{ compiler }`, rather than just `compiler`, even though there doesn't seem to be any need for this (there's no other options it goes along with in those contexts).  Is there a reason for this?

Finally, not a review comment, but it's worth noting that -- if you do change (2) to work one way consistently rather than leaving that as-is -- then this would potentially allow `run` to make choices based on the version.  (Although I guess actually there are ways to set that up w/o making that change, just by moving where in `run` the load happens.)  Now to be clear I don't *want* `run` to do that, because that adds substantial maintenance burden!  But it does *allow* us to do it if it ever becomes necessary, as was discussed for instance in [this Solidity issue](https://github.com/ethereum/solidity/issues/12153)...

packages/compile-solidity/src/index.js
```diff
@@ -129,6 +129,12 @@ const Compile = {
     options = Config.default().merge(options);
     options = normalizeOptions(options);
 
+    const supplier = new CompilerSupplier({
+      events: options.events,
+      solcConfig: options.compilers.solc
+    });
+    const { solc } = await supplier.load();
+
```
cds-amal, line 137: would be nice to log some metadata about the loaded solc

packages/profiler/src/index.ts
cds-amal: `compilerSpecification`? definitely something other than compiler

packages/profiler/src/index.ts
```diff
@@ -61,7 +62,7 @@ export class Profiler {
       //be resolved, it will show up as a compile error rather than a Truffle
       //error.
       try {
-        return await resolver.resolve(filePath, importedFrom);
+        return await resolver.resolve(filePath, importedFrom, { compiler });
```
cds-amal, line 65: would like to see a debug-log in the exception path. Also, are the comments wrt structured errors up to date?

packages/resolver/lib/resolver.ts
```diff
@@ -81,14 +81,23 @@ export class Resolver {
...
   async resolve(
     importPath: string,
-    importedFrom: string
+    importedFrom: string,
+    options: {
+      compiler?: {
+        name: string;
+        version: string;
+      }
+    } = {}
```
cds-amal, line 90: Should this be a typed and shared?

packages/resolver/lib/source.ts
```diff
@@ -2,7 +2,16 @@ import type { ContractObject } from "@truffle/contract-schema/spec";
...
-  resolve(importPath: string, importedFrom: string): Promise<SourceResolution>;
+  resolve(
+    importPath: string,
+    importedFrom: string,
+    options?: {
+      compiler?: {
+        name: string;
+        version: string;
+      }
+    }
```
cds-amal, line 13: type reused here

cds-amal (commented):
Looks good to me. I don't have much feedback other than I would like to see some more debug instrumentation, which would have been helpful earlier this week :/ This, is of course, your choice.

I did notice that debugModule for `packages/profiler/src/index.ts` is `compile-common:profiler` and should probably be updated.

haltman-at:
> Looks good to me. I don't have much feedback other than I would like to see some more debug instrumentation, which would have been helpful earlier this week :/ This, is of course, your choice.

Huh, the whole idea of adding debug instrumentation for use by other people is pretty weird to me; if I'm debugging a problem and I want debug output somewhere, I just add it temporarily!  And then sometimes I leave it in and actually check it in to CI. :P

packages/profiler/src/index.ts
gnidan: note that `compiler` is used in this form exactly in @truffle/db, so that's why I went with this

packages/profiler/src/index.ts
gnidan: (I believe the comments are up-to-date, but which comment are you asking about?)

packages/resolver/lib/resolver.ts
gnidan: My thinking was that this type was within a reasonable threshold of simplicity, since it's only defined here and in the `ResolverSource` interface. Seems a bit unnecessary for only 2 duplications?

packages/compile-solidity/src/run.js
gnidan: you bet!

gnidan:
> 1. Not a comment on the code, but -- I got the point of the new `compiler` field after reading the code, but this would have been easier had it been explained beforehand. :P  Actually, I'm still not entirely clear on why it was made an options field rather than something else; I have guesses as to why but it would be nice to know explicitly the reason.

I'm adding it as an `options` field because it seems like a reasonable good practice in this situation, where it's this new third parameter that's added as optional to a method whose signature has largely remained unchanged. Definitely does not seem like the sort of surprise positional argument you'd want to see here!

If you're asking about why I didn't add it to, say, the constructor? Well, I would've preferred that! Except that the resolver gets constructed long before we have any notion of compiler versions; remember that compile-solidity requires a `resolver` instance passed upfront!

> 2\. I notice you didn't pass solc in the pragma case, when compiling Yul, or in sources(), leaving things the old way in those cases. I guess it's not necessary to alter things there, but my inclination is that consistency is better. Then again, maybe not -- it would be more consistent, but it would also be doing the load in more places, so maybe this way is better.

Unfortunately the way the pragma stuff is set up, it'd be a bit more work to make that consistent (I tried it)... basically, I didn't want too in the weeds reworking the architecture of `compileWithPragmaAnalysis`, which is built on the assumption that `run()` will be doing its own `supplier.load()`. I can rework this if you'd like, but it seemed to me like a follow-on task (esp. since I'm hoping we'll get to #4439 soon and have a chance to refactor for that).

As far as Yul goes, well, the way it's working is fine, and there's no benefit to doing the compiler pre-load. I can make this consistent with normal `sourcesWithDependencies` behavior if you prefer; I just didn't want to risk breaking code I don't understand :)

> 3\. Speaking of that `compiler` field again -- I notice in some places you passed it as `{ compiler }`, rather than just `compiler`, even though there doesn't seem to be any need for this (there's no other options it goes along with in those contexts).  Is there a reason for this?

I pass it as `{ compiler }` because it's an options-style argument, that's why! I feel very strongly that it's better style to prefer named arguments in these kinds of situations because the value of the argument never strays far from the argument name (thus improving readability).

> ... if you do change (2) to work one way consistently rather than leaving that as-is -- then this would potentially allow `run` to make choices based on the version.

Think this makes sense as follow-on work? I think there's a lot we could do to make this all more consistent in conjunction with #4439. Let me know if you'd rather not take the risk to delay it, though. Like I said; I just sought to take the approach here to be safe with my changes!

packages/compile-solidity/src/index.js
gnidan: Pushed a change to add this in `compile-solidity:compilerSupplier`, rather than here directly

packages/profiler/src/index.ts
gnidan: added debug statement

packages/resolver/lib/resolver.ts
cds-amal: fair enough

packages/compile-solidity/src/compilerSupplier/index.ts
```diff
@@ -67,6 +70,11 @@ export class CompilerSupplier {
 
     if (strategy) {
       const solc = await strategy.load(userSpecification);
+      debug(
+        "Loaded solc v%s using %s strategy",
+        solc.version(),
```
cds-amal, line 75: oh, this is convenient!

gnidan: Yeah but I suspect it's causing test timeouts? On Docker, this will shell out (maybe that's something to cache)

gnidan: Wait I was wrong; it doesn't shell out. Huh I wonder why the addition of some debug statements would cause worse timeouts. Maybe it's environmental? I'll try again in the morning.

gnidan: (Reverting the two commits to add instrumentation to test that those are what's causing the timeouts)

gnidan:
So, @cds-amal, seems like adding `debug` statements slows things down enough to cause timeouts. Do you still think it's worth it?

packages/compile-solidity/src/compilerSupplier/index.ts
cds-amal: I thought it was convenient that all loaded strategies would return a `version` function.  bummer.

cds-amal:
> So, @cds-amal, seems like adding `debug` statements slows things down enough to cause timeouts. Do you still think it's worth it?

I'm not sure how debug statements would slow things down. Aren't they conditional? Unless the expression passed to debug is evaluated and throws?

packages/profiler/src/index.ts
haltman-at: Like, it goes in `options` like this there?

haltman-at:
I mean, the expression passed to `debug` is evaluated no matter what, although I don't know if that's relevant here.

haltman-at:
> Think this makes sense as follow-on work? I think there's a lot we could do to make this all more consistent in conjunction with #4439. Let me know if you'd rather not take the risk to delay it, though. Like I said; I just sought to take the approach here to be safe with my changes!

I mean to be clear I'm hoping we *don't* do this, just noting that it now *allows* us to if we need to.

haltman-at:
> I'm adding it as an options field because it seems like a reasonable good practice in this situation, where it's this new third parameter that's added as optional to a method whose signature has largely remained unchanged. Definitely does not seem like the sort of surprise positional argument you'd want to see here!

> If you're asking about why I didn't add it to, say, the constructor? Well, I would've preferred that! Except that the resolver gets constructed long before we have any notion of compiler versions; remember that compile-solidity requires a resolver instance passed upfront!

So as best I can tell, you're talking specifically about why it was added as an options field in *resolver* (I'm inferring that from your talk about "the constructor"... it took me a while to realize what constructor you meant!)  But, I'm not asking about how things are passed just in resolver, I'm asking about how they're passed everywhere you changed things.

That said, if you're putting it in `options` in resolver, then yeah, it makes sense to also put it in `options` elsewhere.  (And am I correct in understanding that it already is in there in `db`?)

haltman-at:
> Unfortunately the way the pragma stuff is set up, it'd be a bit more work to make that consistent (I tried it)... basically, I didn't want too in the weeds reworking the architecture of compileWithPragmaAnalysis, which is built on the assumption that run() will be doing its own supplier.load(). I can rework this if you'd like, but it seemed to me like a follow-on task (esp. since I'm hoping we'll get to #4439 soon and have a chance to refactor for that).

Yeah I think it's fine to leave this as-is for now.  As you point out, there isn't really a need for this mechanism in other cases.

haltman-at:
> I pass it as { compiler } because it's an options-style argument, that's why! I feel very strongly that it's better style to prefer named arguments in these kinds of situations because the value of the argument never strays far from the argument name (thus improving readability).

I don't follow.  It's an options-style argument, but why?  I see how this makes sense for `requiredSourcesForSingleFile`, but I'm not seeing it for the other cases.  When you say "it's better style to prefer named arguments in these kinds of situations", what kind of situations are those?

IDK, named arguments for just a single thing bother me because like... it's just needless indirection.  I guess my thinking is, every step ought to mean something; meaningless steps are confusing, they make the reader look for some significance that isn't there.  In this case, the wrapping and unwrapping are meaningless steps.  It makes sense if you anticipate more named arguments in the future, but otherwise...?

packages/profiler/src/index.ts
gnidan: Yep!

gnidan:
> IDK, named arguments for just a single thing bother me

I know they bother you :)

You ask "what situations?" for using named-style arguments, and my answer would unequivocally be "all situations", but certainly "situations where there's more than two arguments". I see two benefits:
- It affords future additions to the namespace, like you mention
- But primarily, the benefit is that using named options in this way means that you always get the name of the field, so you don't end up with caller code like `functionWithLotsOfArgs(1,2,3,true,true,true,true,true)`, where you're forced to go to the definition to see the meaning of each. Using named options provides clear benefits even with only a single such option.

(This is a hill I will die on)

packages/profiler/src/index.ts
haltman-at: Then I guess this is no worse than what already exists! :)

haltman-at:
I mean yeah for inputs that can be fairly opaque I buy it, but in this case `compiler` is *already* an object whose form makes its meaning apparent.

...in any case, I may as well go ahead and approve this, since at this point we're just bikeshedding. :P

haltman-at (approved):

gnidan:
Thanks @haltman-at and @cds-amal!

I'll merge this then open a follow-up to look into that weirdness with `solc.version()`

PR #4424, opened 2021-11-11
Add Arbitrum and Polygon to source-fetcher

leeftk:
This PR is created to add the `arbitrum` and `poylgon` networks to the `source-fetcher` package. This PR also includes tests to asses the functionality of the `fetch-and-compile` as well as the changes made to `source-fetcher`. 
### Notes
The tests are located inside the `fetch-and-compile` package. The tests use sinon stubs to test that the contract address being passed as an argument to `fetch-and-compile` is verified on etherescan. The networks being tested are `mainnet` `arbitrum` and `polygon`. The changes to `source-fetcher` itself include adding the `arbitrum` and `poylgon`  networks to `etherscan.ts` and adding the corresponding network ids to `networksById` in `common.ts`. As well as creating a test file located in `test/fetch.test.ts`.


packages/fetch-and-compile/package-lock.json
```diff
@@ -0,0 +1,876 @@
+{
```
cds-amal, line 1: This file doesn't belong in a yarn managed project.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,130 @@
...
+        module: "contract",
+        action: "getsourcecode",
+        address: '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D',
+        apikey: ''
+      },
+      responseType: "json",
+      maxRedirects: 50
+    }).returns(Promise.resolve({ data: fixture.mainnetData }))
+    //@ts-ignore
+    axios.get.callThrough();
```
cds-amal, line 33: This code is duplicated twice and could be factored

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,130 @@
...
+      let contractName = result.sourceInfo.contractName
+      //assert contract name is correct 
+      assert.equal(contractName, "UniswapV2Router02")
+      //assert contract name is not undefined
+      assert.notEqual(contractName, undefined)
+    })
+  })
+});
+
+describe("fetches contract on arbitrum and checks for verification", () => {
```
cds-amal, line 51: I think this test module could be cleaned up if all the tests resided under one describe. I don't feel strongly about it, but there is an opportunity to refactor.

cds-amal (changes requested):
Nice work @leeftk. Passing tests make me happy!

haltman-at:
Btw @leeftk I just reran the tests because there was a timeout; if this keeps happening you may want to up the timeout on the failing test.

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -62,7 +62,7 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
     if (networkName === undefined || !supportedNetworks.includes(networkName)) {
       throw new InvalidNetworkError(networkId, "etherscan");
     }
-    this.suffix = networkName === "" ? "mainnet": `${networkName}`;
+    this.networkName = "" ? "mainnet": networkName;
```
cds-amal, line 65:
```suggestion
    this.networkName = networkName;
```

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+          network_id: 1
+        }
+      },
+      network: "mainnet",
+    });
+    //asserting that mainnet url and contract address is passed as args
+    axiosMock("https://api.etherscan.io/api", '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', fixture.mainnetData)
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile('0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', config)
```
cds-amal, line 40: The contract address can be a variable.
```
const address = "0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D";
axiosMock(..., address, ...)
fetchAndCompile(address, ...)
```

```

File: packages/fetch-and-compile/test/fetch.test.ts [hunk 6]
```diff
@@ -0,0 +1,81 @@
...
+    //asserting that arbitrum url and contract address is passed as args
+    axiosMock("https://api.arbiscan.io/api", '0xBf00759D7E329d7A7fa1D4DCdC914C53d1d2db86', fixture.arbitrumData)
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile('0xBf00759D7E329d7A7fa1D4DCdC914C53d1d2db86', config)
+    let contractName = result.sourceInfo.contractName
+    assert.equal(contractName, "stARBIS")
+    assert.notEqual(contractName, undefined)
+  });
+  it('verfies contract from polygon', async () => {
```
haltman-at, line 63: Repeat same comments as for above two blocks, not listing it all out again

packages/fetch-and-compile/test/fixture.js
```diff
@@ -0,0 +1,2825 @@
+const mainnetData = {
```
haltman-at, line 1: These test fixtures are huge!  Wouldn't it make more sense to use something smaller?

haltman-at (commented):
This looks pretty good on the whole, thanks @leeftk!  I have a number of comments but they're mostly pretty nitpicky, basic structure seems pretty good.

I'd prefer it if there were test fixtures for the other cases that source-fetcher handles, rather than just the single-source Solidity Etherscan case; but the fact is, putting any tests in this package at all is a huge improvement, and given everything else it probably makes more sense to save that for follow-on work, rather than putting that in this PR.

packages/fetch-and-compile/test/fixture.js
eggplantzzz: Yeah it would be nice to have something smaller. I think we just used the actual response from the call it makes in the test.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      apikey: ''
+    },
+    responseType: "json",
+    maxRedirects: 50
+  }).returns(Promise.resolve({ data: data }))
+}
+afterEach(()=>
+    //@ts-ignore
+    //restoring stub
+    axios.get.restore()
```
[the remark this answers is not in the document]
eggplantzzz: Can you indent only 2 spaces? In Truffle we never use anything other than 2 space indentation. Also in general I think we use "function" notation for functions rather than arrow functions in Mocha tests because that's what Mocha recommends. In other words,
`afterEach(function() {...` for the callbacks in Mocha's `it`s, `after`s, `beforeEach`s, etc. It is mostly to have a useful `this` value. Arrow functions don't bind `this` the same way that "function" syntax does.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
+import assert from "assert";
+import { describe, it } from "mocha";
+import Config from "@truffle/config";
+import { fetchAndCompile } from "../lib/index";
+import axios from "axios";
+import sinon from "sinon";
+const fixture: any = require("./fixture.js");
+
+function axiosMock(url: string, address: string, data: object) {
```
eggplantzzz, line 9: How would you feel about changing this method name? You are not mocking axios, you are stubbing the axios get method. Maybe `stubAxiosGetMethod` or something like that.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      networks: {
+        mainnet: {
+          network_id: 1
+        }
+      },
+      network: "mainnet",
+    });
+    //asserting that mainnet url and contract address is passed as args
+    axiosMock("https://api.etherscan.io/api", '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', fixture.mainnetData)
+    //@ts-ignore
```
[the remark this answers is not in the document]
eggplantzzz: TS doesn't seem to be smart enough to know that `axios.get` here is a Sinon stub. It thinks it is an Axios reponse object. Is there a better way to do this?

packages/fetch-and-compile/test/fixture.js
haltman-at: Actually, on looking at this some more, what bugs me here is not so much the *size* but rather the lack of *structure*.  I think this would be a lot more swallowable if, instead of using double-quoted string literals with `\n`s everywhere for line breaks, it used backticked-string literals with actual linebreaks.  (We do this a bunch in the debugger tests for including Solidity sources.)  Likely, you should make that change even if you change it to something smaller.

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
haltman-at: Oh, this is true, in Mocha you should use `function() { ... }` rather than arrow functions, I'd forgotten that.

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
haltman-at: No idea, I just think it should be noted in a code comment.

packages/fetch-and-compile/test/fixture.js
leeftk: I was actually able to decrease the size of the fixture without failing the tests. It seems like just the first line of the contract and the contract name will suffice to pass the tests.

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
leeftk: Sweet I'll change these and get ride of the arrow functions. Thanks.

packages/fetch-and-compile/test/fixture.js
haltman-at: Yeah I wouldn't recommend going that far!  We do want an actual fully-working fixture; we want to be able to add more tests in the future.  Please don't optimize the fixture to only pass the particular tests that you happened to add!

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
haltman-at: Note to be clear of course I don't mean that you shouldn't use arrow functions at *anywhere* in Mocha, just like, they shouldn't be used as the test functions used in an `it`, `describe`, `afterEach`, etc.

packages/fetch-and-compile/test/fixture.js
haltman-at: Oh wait I just realize I misunderstood what you meant, you just meant stripping the contract down to the bare minimum, not going even further than that; for some reason I misunderstood and thought you meant the latter.  Yeah that would be fine, although I might put in at least some token function that can be checked for in the future or something if we want to expand the tests.

packages/fetch-and-compile/package.json
```diff
@@ -5,7 +5,8 @@
   "main": "dist/index.js",
   "scripts": {
     "build": "tsc",
-    "prepare": "yarn build"
+    "prepare": "yarn build",
+    "test": "yarn build && mocha --exit -r ts-node/register test/**/*.test.ts --timeout 100000"
```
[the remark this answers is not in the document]
leeftk: Thanks for the heads up! Yea I decreased it down to 20000 to make it a bit more reasonable. 👍

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    //restoring stub
+    axios.get.restore()
+  })
+  it('resolves with verified contract', () => {
+    return fetchAndCompile('0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', config).then(result => {
+      let contractName = result.sourceInfo.contractName
+      //assert contract name is correct 
+      assert.equal(contractName, "UniswapV2Router02")
+      //assert contract name is not undefined
+      assert.notEqual(contractName, undefined)
```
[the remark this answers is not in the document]
leeftk: Awesome thanks for bringing this up. Yea I changed these to async/await syntax.

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -54,12 +55,14 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
...
-      "kovan-optimistic"
+      "kovan-optimistic",
+      "arbitrum",
+      "polygon"
     ];
     if (networkName === undefined || !supportedNetworks.includes(networkName)) {
       throw new InvalidNetworkError(networkId, "etherscan");
     }
-    this.suffix = networkName === "mainnet" ? "" : `-${networkName}`;
+    this.suffix = networkName === "" ? "mainnet": `${networkName}`;
```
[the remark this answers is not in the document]
leeftk: Yeaaaa this makes a lot more sense. I updated this as well thanks for the suggestion.

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -90,8 +93,26 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
   private async makeRequest(address: string): Promise<EtherscanSuccess> {
     //not putting a try/catch around this; if it throws, we throw
     await this.ready;
+    let url = '';
+    switch (this.suffix){
```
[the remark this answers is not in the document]
leeftk: I moved this to its on function as Harry recommended below so I actually got rid of all this.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,130 @@
...
+    }).returns(Promise.resolve({ data: fixture.arbitrumData }))
+     //@ts-ignore
+     axios.get.callThrough();
+  })
+  afterEach(() => {
+    //@ts-ignore
+    axios.get.restore()
+  })
+  it('resolves with verified contract', () => {
+    return fetchAndCompile('0xBf00759D7E329d7A7fa1D4DCdC914C53d1d2db86', config).then(result => {
```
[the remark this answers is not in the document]
leeftk: Sweat thanks!

packages/fetch-and-compile/package-lock.json
leeftk: Makes sense, I deleted it. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
leeftk: Yup yup, makes sense I refactored this to make it a bit cleaner.

packages/fetch-and-compile/test/fetch.test.ts
leeftk: Definitely I made this change when I refactored above. Thanks!

packages/source-fetcher/lib/etherscan.ts
leeftk: Awesome thanks for the suggestions. Made this change!

packages/fetch-and-compile/test/fetch.test.ts
leeftk: Yup yup a variable makes a ton of more sense. Made this change! Thanks.

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -90,8 +93,26 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
   private async makeRequest(address: string): Promise<EtherscanSuccess> {
     //not putting a try/catch around this; if it throws, we throw
     await this.ready;
+    let url = '';
+    switch (this.networkName){
```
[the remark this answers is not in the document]
leeftk: Okay cool I actually put this into its own function like you suggested. Makes it a lot cleaner like you stated. :)

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -90,8 +93,26 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
   private async makeRequest(address: string): Promise<EtherscanSuccess> {
     //not putting a try/catch around this; if it throws, we throw
     await this.ready;
+    let url = '';
```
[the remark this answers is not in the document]
leeftk: Yup yup that makes sense thanks.

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -90,8 +93,26 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
...
+      case "arbitrum" :
+        url = "https://api.arbiscan.io/api";
+        break;
+      case "polygon" :
+        url = "https://api.polygonscan.com/api";
+        break;
+      case "mainnet":
+        url = "https://api.etherscan.io/api";
+        break;
+      case "kovan":
```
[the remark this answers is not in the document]
leeftk: Yea I used this for testing and forgot it in there. Nice catch. :)

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
leeftk: Okay I added a comment here to briefly explain why we used ts-ignore.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
+import assert from "assert";
```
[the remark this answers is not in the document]
leeftk: Removed this, thanks for catching that. I originally had planned on using it but never did. Thanks!

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+import { describe, it } from "mocha";
```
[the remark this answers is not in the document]
leeftk: Okay sweet I added the `.eslintrc.json` file and just imported `mocha`.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      network: "mainnet",
+    });
+    //asserting that mainnet url and contract address is passed as args
+    axiosMock("https://api.etherscan.io/api", '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', fixture.mainnetData)
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile('0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', config)
+    let contractName = result.sourceInfo.contractName
+    assert.equal(contractName, "UniswapV2Router02")
+    assert.notEqual(contractName, undefined)
```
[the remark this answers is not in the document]
leeftk: Yup yup, removed this. Makes sense to me!

packages/fetch-and-compile/test/fetch.test.ts
leeftk: Yea great idea thanks.

packages/source-fetcher/lib/etherscan.ts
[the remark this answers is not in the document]
haltman-at: Great, glad that worked out!

packages/fetch-and-compile/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+    "extends": ["../../.eslintrc.package.json"]
+  }
```
haltman-at, line 3: Sorry for more stuff that should be taken care of by `prettier`, but -- oh, hang on, this is a JSON file, it wouldn't be taken care of by `prettier` even if it were working! :P  Anyway, sorry for more of these tiny style nitpicks, but something's gone wrong with your indentation here.

Btw, you seem to be using an editor that doesn't insert newline at end of file (as indicated in the GitHub display with that big red minus-in-a-circle), so you may want to add those manually (until `prettier` is working again... although for this file, as it's JSON, you'd have to do it manually anyway. :P )

packages/fetch-and-compile/package.json
```diff
@@ -46,6 +47,14 @@
   "devDependencies": {
     "@truffle/compile-common": "^0.7.22",
     "@truffle/config": "^1.3.12",
+    "@types/chai": "^4.2.22",
+    "@types/debug": "^4.1.7",
+    "@types/mocha": "^9.0.0",
+    "@types/node": "^16.11.6",
+    "@types/semver": "^7.3.9",
```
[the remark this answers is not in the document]
haltman-at: Hey @leeftk, still looking for an answer to your question.  It looks like you're going with just leaving them mismatched for now, which isn't a problem, but it would be good to explicitly note that you are making that decision, and, ideally, why.

packages/fetch-and-compile/package.json
```diff
@@ -41,11 +42,19 @@
...
     "@truffle/compile-solidity": "^6.0.0",
     "@truffle/source-fetcher": "^0.5.9",
-    "semver": "^7.3.4"
+    "debug": "^4.3.2",
+    "semver": "^7.3.5"
   },
   "devDependencies": {
     "@truffle/compile-common": "^0.7.22",
     "@truffle/config": "^1.3.13",
+    "@types/chai": "^4.2.22",
```
haltman-at, line 51: You removed chai, but you left in `@types/chai`.

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -86,12 +89,24 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
       { retries: 3, minTimeout: this.delay * initialTimeoutFactor }
     );
   }
-
+  
+  private urlSwitcher(address: string){
```
haltman-at, line 93: This function needs to be renamed.  The name `urlSwitcher` does not describe what it does.  Usually, functions are named as verb phrases.  I would suggest a name like `determineUrl`.

Also, you've given this function a parameter `address`, but it's never used.  That should be removed.

Also I notice you've gone the route of making this an instance method, and having it switch based on `this.networkName`...  that works, but maybe it would be better to have `networkName` passed in explicitly as a parameter, and have this function be a static method (or outside the class entirely) rather than an instance method?  I feel like that would make the data flow more explicit.  Pure functions are generally preferable when feasible.

(In fact, down below, where you actually *invoke* it, you invoked it as if the network name was a parameter to be passed in!  Whichever route you go, better make it consistent.)

Also, this function is missing a return signature; you should add that.

Also -- sorry for more of these tiny things that should be handled by `prettier`, hopefully @cds-amal will have that fixed soon -- but there should be a space between `)` and `{`.  (You have this as well on the `switch` line.)

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -86,12 +89,24 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
...
+      default:
+        return `https://api-${this.networkName}.etherscan.io/api`;
+    }
+  }
   private async makeRequest(address: string): Promise<EtherscanSuccess> {
     //not putting a try/catch around this; if it throws, we throw
     await this.ready;
     const responsePromise = axios.get(
-      `https://api${this.suffix}.etherscan.io/api`,
+      this.urlSwitcher(this.networkName),
```
haltman-at, line 109: See previous comment about this!

packages/fetch-and-compile/test/fixture.js
```diff
@@ -0,0 +1,64 @@
...
+        "EVMVersion": "Default",
+        "Library": "",
+        "LicenseType": "MIT",
+        "Proxy": "0",
+        "Implementation": "",
+        "SwarmSource": "ipfs://4b803eff2b53da590615ae9571dfcb7a4fa564c5f1d196b7e24bac3dcf52f96d"
+      }
+    ]
+  }
+  module.exports = {mainnetData,arbitrumData,polygonData}
```
haltman-at, line 64: Sorry for more of this nipitcking, but... there should be spaces here. `{ mainnetData, arbitrrumData, polygonData }`.

packages/fetch-and-compile/test/fixture.js
```diff
@@ -0,0 +1,64 @@
...
+        "ConstructorArguments": "000000000000000000000000000000000000000000000000000000000000014000000000000000000000000000000000000000000000000000000000000001800000000000000000000000003324af8417844e70b81555a6d1568d78f4d4bf1f000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000010000000000000000000000001948abc5400aa1d72223882958da3bec643fb4e50000000000000000000000001b02da8cb0d097eb8d57a175b88c7d8b47997506000000000000000000000000c2132d05d31c914a87c6611c10748aeb04b58e8f000000000000000000000000d3b71117e6c1558c1553305b44988cd944e97300000000000000000000000000a5e0829caced8ffdd4de3c43696c57f7d7a678ff0000000000000000000000000000000000000000000000000000000000000007536861726520560000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000025356000000000000000000000000000000000000000000000000000000000000",
+        "EVMVersion": "Default",
+        "Library": "",
+        "LicenseType": "MIT",
+        "Proxy": "0",
+        "Implementation": "",
+        "SwarmSource": "ipfs://4b803eff2b53da590615ae9571dfcb7a4fa564c5f1d196b7e24bac3dcf52f96d"
+      }
+    ]
+  }
```
haltman-at, line 63: Missing semicolon here, and also in other similar spots.

packages/fetch-and-compile/test/fixture.js
haltman-at: So, uh, @leeftk, any action on this?  What's going on here?

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      network: "mainnet",
+    });
+    const address = '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D';
+    //asserting that mainnet url and contract address is passed as args
+    stubAxiosGetMethod("https://api.etherscan.io/api", address, fixture.mainnetData);
+    //TS can't detect that is a sinon stub so we have to use ts-ignore
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile(address, config);
+    let contractName = result.sourceInfo.contractName;
```
haltman-at, line 46: Btw, I only just noticed you're getting `contractName` out of `sourceInfo`.  That tests the fetching part, but not so much the compiling part (other than that the compiling part doesn't crash).  I think this test (and the others) would be better if, in addition to testing the contract name found in `sourceInfo`, it also tested the contract name found in `compilations`.

haltman-at: Btw @leeftk I don't know if you've merged in or rebased against `develop` since #4447 was merged, but if so you might want to add tests of that as well. :)  Or if you don't want to do that then I can get it as follow-on work.

haltman-at (commented):
Reviewed this a second time and I still think there are places the style needs to be improved.  (And, yes, more nitpicks that ought to be handled by `prettier`, sorry, but more substantial style comments as well.)  There's also one functional change to the tests that I think should be made so that they're actually testing more of the functionality.

Still waiting for responses on some of the stuff from the earlier review... really you should have done that, and resolved everything that came up in the previous review, before requesting this next round!  Anyway, I did this second round, but please do still respond on the unresolved stuff from the previous round.

packages/fetch-and-compile/test/fixture.js
leeftk: I spoke with @amal and he suggested removing the `\n` and the '+' as you also did above. So I went that route instead. I didn't edit the ABI nor the Source text. That's what came back from the response. The only thing I changed was removing the line breaks as I figured it leave the data as is would give us the option to expand our tests in the future if need be. But if we want to go with the a different option then I can remove everything besides the first line of each Source code and it will still compile. But I can't leave just a function in there as you suggested above as it won't compile and the tests fail. So besides what I submitted the only other option is leaving just the `Pragma solidity <version>` line.

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -90,8 +93,26 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
...
+        url = "https://api.polygonscan.com/api";
+        break;
+      case "mainnet":
+        url = "https://api.etherscan.io/api";
+        break;
+      case "kovan":
+        url = "https://api-kovan.etherscan.io/api";
+        break;
+      default:
+       url = `https://api-${this.networkName}.etherscan.io/api`;
```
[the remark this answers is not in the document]
leeftk: I fixed this. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
+import assert from "assert";
+import { describe, it } from "mocha";
+import Config from "@truffle/config";
+import { fetchAndCompile } from "../lib/index";
```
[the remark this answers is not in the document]
leeftk: Didn't make any changes here cause @gnidan hasn't responded but I can reach out in Slack if need be.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    params: {
+      module: "contract",
+      action: "getsourcecode",
+      address: address,
+      apikey: ''
+    },
+    responseType: "json",
+    maxRedirects: 50
+  }).returns(Promise.resolve({ data: data }))
+}
```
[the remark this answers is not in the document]
leeftk: Fixed this. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+afterEach(()=>
+    //@ts-ignore
+    //restoring stub
+    axios.get.restore()
+)
```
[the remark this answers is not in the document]
leeftk: Fixed this. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+    //asserting that mainnet url and contract address is passed as args
+    axiosMock("https://api.etherscan.io/api", '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', fixture.mainnetData)
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile('0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', config)
+    let contractName = result.sourceInfo.contractName
+    assert.equal(contractName, "UniswapV2Router02")
+    assert.notEqual(contractName, undefined)
+  })
```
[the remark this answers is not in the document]
leeftk: Fixed this. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+  sinon.stub(axios, 'get').withArgs(url, {
+    params: {
+      module: "contract",
+      action: "getsourcecode",
+      address: address,
+      apikey: ''
+    },
+    responseType: "json",
+    maxRedirects: 50
+  }).returns(Promise.resolve({ data: data }))
```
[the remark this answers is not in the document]
leeftk: Fixed. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    const config = Config.default().merge({
+      networks: {
+        arbitrum: {
+          network_id: 42161
+        }
+      },
+      network: "arbitrum",
+    });
+    //asserting that arbitrum url and contract address is passed as args
+    axiosMock("https://api.arbiscan.io/api", '0xBf00759D7E329d7A7fa1D4DCdC914C53d1d2db86', fixture.arbitrumData)
```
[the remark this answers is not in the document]
leeftk: Yup yup, I did this for all three of them. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+      },
+      network: "mainnet",
+    });
+    //asserting that mainnet url and contract address is passed as args
+    axiosMock("https://api.etherscan.io/api", '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', fixture.mainnetData)
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile('0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', config)
+    let contractName = result.sourceInfo.contractName
+    assert.equal(contractName, "UniswapV2Router02")
```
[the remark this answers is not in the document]
leeftk: Okay thanks for the heads up. Fixed.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+      },
+      network: "mainnet",
+    });
+    //asserting that mainnet url and contract address is passed as args
+    axiosMock("https://api.etherscan.io/api", '0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', fixture.mainnetData)
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile('0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D', config)
+    let contractName = result.sourceInfo.contractName
```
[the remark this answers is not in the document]
leeftk: Fixed this. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
leeftk: Yup yup, I changed these as per the mocha docs. Thanks for the heads up.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+          network_id: 42161
+        }
+      },
+      network: "arbitrum",
+    });
+    //asserting that arbitrum url and contract address is passed as args
+    axiosMock("https://api.arbiscan.io/api", '0xBf00759D7E329d7A7fa1D4DCdC914C53d1d2db86', fixture.arbitrumData)
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile('0xBf00759D7E329d7A7fa1D4DCdC914C53d1d2db86', config)
```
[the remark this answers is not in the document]
leeftk: Okay sounds good, I went through and added all the ones I saw that were missing. Thanks.

packages/fetch-and-compile/test/fetch.test.ts
leeftk: I removed these, thanks for the heads up.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    let contractName = result.sourceInfo.contractName
+    assert.equal(contractName, "stARBIS")
+    assert.notEqual(contractName, undefined)
```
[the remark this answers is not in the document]
leeftk: Okay thanks, removed it.

packages/fetch-and-compile/package.json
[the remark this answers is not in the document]
leeftk: Hey yea sorry I didn't answer here I guess I spoke with Amal about it and just forgot. It seems like there's different versions for these packages on NPM so for `@types/debug'  the latest version is 4.1.7. While the latest version of `debug' is 4.3.2. And this is the case for the other mismatches as well. So I could downgrade `debug` I guess but since it builds like you said above I figured I would just leave them. I'm open to any suggestion though so let me know what you think is best. Thanks :).

packages/fetch-and-compile/.eslintrc.json
leeftk: Okay I'll take a look at this and comment once it's fixed. Also my apologies for not replying to the previous comments before requesting the next round. I fixed all the things you mentioned beside the issue relating to importing modules as I was waiting to hear back on that one before fixing. I just didn't comment on them here before pushing the changes cause I didn't know that was part of the workflow but now that I know I will be sure to do that in the future. Thanks for bringing that up and thanks for reviewing this again!

haltman-at: To be clear, the idea isn't that you have to comment on *everything*; just that, you know, for each thing it should be made clear what's going on.  If it's a simple thing you can just directly implement, well, you can just do that; not much to say.  (Although it's still useful to hit the thumbs-up button to acknowledge the comment.)  But if you're leaving something unaltered for now, or if the code changes are complex or wouldn't be obvious ot the reviewer... you should use this space to make sure the reviewer understands the current status of things!

packages/fetch-and-compile/package.json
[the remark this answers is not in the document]
haltman-at: OK, but I was asking about `semver`, not `debug`?

packages/fetch-and-compile/test/fixture.js
haltman-at: > I spoke with @amal and he suggested removing the \n and the '+' as you also did above. So I went that route instead.

I don't understand.  There's no concatenation, sure, but I'm still seeing a lot of `\r\n` everywhere.  I suggested using template strings (backticks) so that you can use actual linebreaks, so that it looks more structured.  (You could do the same with the ABI field, so that it's not full of `\"`s.  Is that something you plan to do?

> I didn't edit the ABI nor the Source text. That's what came back from the response.

Oh huh!  My misunderstanding.  They looked mismatched to me.  But I guess not?

> But if we want to go with the a different option then I can remove everything besides the first line of each Source code and it will still compile.

No, please do not do that!

OK, let me explain in more detail what I think are good vs bad ways to strip down the tests.

Fundamentally, the tests should be *coherent*.  They don't have to be a real contract instance off of Etherscan, but they should be what the output *would* be if you wrote a real contract, deployed it, and then verified it on Etherscan.  Remember: We want to be able to write more detailed tests later!

When I talked about stripping down the fixtures, I didn't mean just deleting parts of it without regard for the rest.  I meant rewriting the whole fixture as a new, minimal contract, with all fields adjusted appropriately.

At the very least, the source text ought to include an actual contract, not just a pragma line!  This whole package is based around the idea of *contracts* with verified sources, not just verified sources by themselves.  Note, by the way, that removing the contract entirely is incompatible with the suggestion I made about how the contract name should be tested.

It seems there was a bit of a misunderstanding here -- you talked about stripping down the source file:
> I was actually able to decrease the size of the fixture without failing the tests. It seems like just the first line of the contract and the contract name will suffice to pass the tests.
and I was like, no, don't do that!  But then on rereading I concluded you meant the first line of the *contract*, i.e. the contract declaration, rather than the first line of the *source file*.  After all, you were checking the contract name in the tests, so you must have included the contract, right?  And a minimal contract declararation like that seemed fine, although I suggested including a function too.

(Btw, you say you got compile errors... what compile errors?  Could you go into more detail there?  I ask because when you say you "left in a function", I'm not clear on what exactly you did.  What I had meant was to just put in like a little stub function that does nothing, but you may have taken a different approach...)

But I hadn't yet noticed that, in your tests, you were getting the contract name out of `sourceInfo` rather than `compilations` -- which would allow you to get the contract name *without* having an actual contract.  And now with this comment you've made it explicit that, no, you really were suggesting stripping out the entire contract.

So, please don't do that.  There should, in fact, be a contract.  It can be a minimal one, but it must be a contract.  Moreover, if you go this route, the rest of the response has to be adjusted to match; you can't just alter one component of it and leave the rest the same.

That said, as I've pointed out above, if you just switch to using backticked literals (template strings), so that the whole thing looks more structured, there may just not be much of a need to strip it down in the first place.

Another approach, which you might want to consider, would be to pull the source text out into its own file, `Whatever.sol`, and read that text in here, rather than having it inline, if it's so long.  You could also possibly have `Whatever.abi.json`, etc.  Basically, breaking it up might make it easier to process.  Then having it be so large would be less of a concern.

I'm hoping that makes it clear what I'm asking for?  As always, if it's not, please ask me about any parts that are unclear.

packages/fetch-and-compile/package.json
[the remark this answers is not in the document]
haltman-at: Oh, OK, that makes sense then.  Thanks for clearing that up!

packages/fetch-and-compile/test/fixture.js
leeftk: Awesome thanks for the clarity in this comment. So basically I decided to go with your last suggestion of splitting up the `fixture.js` file into a few different files to make it a bit neater and also easier to consume. I created individual files like `arbitrum.abi.js` for the abi's  and did the same for the source info (labeled `arbitrum.sol.js` in this example). Before pushing this up I just wanted to confirm that this is what you were getting at in that last part where you mentioned. 

> Another approach, which you might want to consider, would be to pull the source text out into its own file, Whatever.sol, and read that text in here, rather than having it inline, if it's so long. You could also possibly have Whatever.abi.json, etc. Basically, breaking it up might make it easier to process. Then having it be so large would be less of a concern.

If so then we should be good!

packages/fetch-and-compile/.eslintrc.json
leeftk: Okay sweet thanks for the clarity on that :)

packages/fetch-and-compile/test/fetch.test.ts
leeftk: Great catch! I didn't even think that about this until you mentioned it. I added an assertion that also tests the `compilations` part of the function. Let me know if what I added will suffice.

packages/source-fetcher/lib/etherscan.ts
haltman-at: This seems to have gotten marked resolved, but I'm marking it unresolved because not all of the above has been addressed.

packages/fetch-and-compile/test/fixture.js
haltman-at: > I created individual files like arbitrum.abi.js for the abi's and did the same for the source info (labeled arbitrum.sol.js in this example). Before pushing this up I just wanted to confirm that this is what you were getting at in that last part where you mentioned.

Well, partly, but also if you are splitting them up like that they should probably be the appropriate types of files -- like, `.sol` and `.json` -- rather than forced into being JS files.

I'm going to go ahead and take a guess... did you make them JS files so that you could read them with `require`?  I'd suggest just using the appropriate file types and then reading those the appropriate way (like, using `fs.readFileSync`, and then for the JSON potentially doing `JSON.parse` if that's appropriate in this context).

leeftk: I'm going to resolve this but if we need to try a different method just let me know.

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
leeftk: Still haven't heard back but I'm guess since this works it should be fine? Going to resolve and if we need to keep discussing it we can reopen.

packages/source-fetcher/lib/etherscan.ts
leeftk: Cool we updated this function and changed the name to `determineUrl`. Thanks for the catch.

packages/fetch-and-compile/.eslintrc.json
```diff
@@ -1,3 +1,4 @@
 {
     "extends": ["../../.eslintrc.package.json"]
-  }
\ No newline at end of file
+
```
eggplantzzz, line 3: small nit, can you get rid of this space?

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
haltman-at: Going to have to unresolve this... I just noticed that you actually put in the line `import "mocha"`.  That line doesn't actually do anything; you should delete it.

You may be wondering, then, what do you have to do to get `describe` and `it`?  Nothing!  You don't have to do anything!  The test framework sets them up as globals; in order to get them working, there is not anything you have to import in any form.  You can just straight-up use them with no setup.

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
haltman-at: Sure thing.  You may want to bug him to be sure, he often doesn't seem to notice GitHub pings, but unless he complains I'm not going to make a fuss over it.

packages/source-fetcher/lib/etherscan.ts
haltman-at: Cool, I see also you took the approach of making it an instance method rather than a pure function.  That works -- as long as you're consistent. :)

packages/fetch-and-compile/test/fetch.test.ts
haltman-at: Cool, I see you've done this!  I have some comments on that but those will be separate. :)

packages/fetch-and-compile/test/fixture.js
haltman-at: Yes, I think I'm going to have to insist on doing this properly.  Please use the appropriate file types and read them in as appropriate; don't stuff them in `.js` files and use `require` like this.

(Moreover, if you *were* to put them in `.js` files, you should -- as I've mentioned above -- be using backticks, i.e. template literals, to make them more readable!  But in actuality these shouldn't be in `.js` files in the first place.)

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    stubAxiosGetMethod(
+      "https://api.etherscan.io/api",
+      address,
+      fixture.mainnetData
+    );
+    //TS can't detect that is a sinon stub so we have to use ts-ignore
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile(address, config);
+    let contractNameSourceInfo = result.sourceInfo.contractName;
```
haltman-at, line 52: This `let` can be `const`, I should note.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    let contractNameCompliations =
```
haltman-at, line 53: A few notes here.  In order of least important to most important:

1. This `let` can be a `const`.
2. This variable name is misspelled; you have "compliations" instead of "compilations".
3. But most importantly, this variable name is misleading; you need to rename this.

Above, you have a variable named `contractNameSourceInfo`.  That's a little bit confusing, but context makes it clear that you mean the contract name coming from the source info.

But then you have a second variable called `contractNameCompilations`.  So that's the contract name coming from the compilations, right?  Nope!  It's not a contract name at all.  That is therefore a misleading variable name; both it's misleading on its own, and it's made more misleading by the false parallelism.

This variable needs to be given a name that more accurately reflects what it is.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      result.compileResult.compilations[0].contracts;
+    assert(
```
haltman-at, line 55: Thanks for adding this check. :)

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+afterEach(function () {
+  //TS can't detect that is a sinon stub so we have to use ts-ignore
+  //@ts-ignore
+  //restoring stub
+  axios.get.restore();
+});
+
+describe("fetchAndCompile", function () {
+  it("verifes contract from mainnet", async function () {
```
haltman-at, line 33: Nitpicking, but just noticed a spelling error here; "verifes" should be "verifies".

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    let contractNameCompliations =
+      result.compileResult.compilations[0].contracts;
+    assert(
+      contractNameCompliations.some(
+        item => item.contractName === "UniswapV2Router02"
+      )
+    );
+    assert.equal(contractNameSourceInfo, "UniswapV2Router02");
+  });
+  it("verifes contract from arbitrum", async function () {
```
haltman-at, line 62: "Verifes" again

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      "https://api.arbiscan.io/api",
+      address,
+      fixture.arbitrumData
+    );
+    //TS can't detect that is a sinon stub so we have to use ts-ignore
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile(address, config);
+    let contractNameSourceInfo = result.sourceInfo.contractName;
+    let contractNameCompliations =
```
haltman-at, line 83: Same comments as before, not going to repeat it all.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      result.compileResult.compilations[0].contracts;
+    assert(
+      contractNameCompliations.some(item => item.contractName === "stARBIS")
+    );
+    assert.equal(contractNameSourceInfo, "stARBIS");
+  });
+  it("verfies contract from polygon", async function () {
```
haltman-at, line 90: Different spelling error here, you have "verfies" for "verifies".

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+      "https://api.polygonscan.com/api",
+      address,
+      fixture.polygonData
+    );
+    //TS can't detect that is a sinon stub so we have to use ts-ignore
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile(address, config);
+    let contractNameSourceInfo = result.sourceInfo.contractName;
+    let contractNameCompliations =
```
haltman-at, line 111: Same comment here as above, not going to repeat it all again.

haltman-at (commented):
Whew, OK, I'd say the big problems are mostly fixed!  I do have a few more comments though, as you can see, although they're mostly on style matters.  (Separately, I've also gone back and commented on some of the other threads above... I've unresolved those so you should be able to find them.)

One other question: Weren't you going to use the test fixtures to use smaller contracts?  You don't appear to have done that.  I don't think that's a necessity, but you should make clear whether you still intend to do that or not.

packages/fetch-and-compile/test/fixture.js
leeftk: So when we fetch a contract from Etherscan you're saying that in the response in `SourceCode` that's actually Solidity and not JS? I was under the assumption that the response was a JS object which is why I put them in a JS file.

haltman-at: Your question seems to indicate a confusion of levels.  I could just say "Yes, that's Solidity", but I'm not sure that would properly address your confusion.

(Also, remember here, I'm not *just* talking about the Solidity sources, which ought to be in `.sol` files, but also the ABI JSONs, which ought to be in `.json` files, and treated as JSON, not as Javascript.  JSON and Javascript may often be identical in representation, but they are conceptually different and should not be treated as interchangeable!)

...I originally started writing a much longer explanation, but I couldn't get it into a form I was happy with, so I'm just going to give a really short answer and hope it suffices: The overall response that we get back from Etherscan is JSON.  The `SourceCode` field of that JSON is a string, which is represented in the response as a JSON-encoded string.  The contents of the underlying string, that is represented by this JSON-encoded form, are the Solidity source.

(Remember: Everything that gets sent over the wire is ultimately just a bunch of bytes.  It's up to us to impose an appropriate interpretation on it!)

This test fixture you are creating, meanwhile, is not JSON, but rather is a Javascript object that will be serialized to JSON.  So, let's ignore the JSON layer, since here we are not actually dealing with JSON.  Then, looking at this Javascript object, it has a `SourceCode` field, which is a string, and that string consists of the Solidity source.  So, the `SourceCode` field of the object you are setting up, is a member of a Javascript object, and as such is a Javascript string, and the contents of that string are Solidity source.

You want to have that string, the Solidity source, represented by a file.  The appropriate way to represent a string by a file is for the file's contents to consist of the contents of that string.  (And since the contents of that string are Solidity source, it would be appropriate to give that file a name ending in `.sol`.)  To instead represent a string in a file by making that file a JS source file, and encoding the string as an JS string literal inside that JS source file, is needless complication.

...I'm not sure if that is sufficient to clear up your confusion; I realize that this is perhaps not the most in-depth, and I don't know where you're starting from.  Well -- if you want to go deeper into questions of data representation, I think Slack or Zoom might be a better place to ask me about them rather than here... we can go over examples or whatever if you want!

packages/fetch-and-compile/test/fetch.test.ts
leeftk: Okay thanks, I paired with Tyler and we came up with some names for these that may be a bit less confusing. Let me know what you think.

packages/fetch-and-compile/test/fetch.test.ts
leeftk: Commented on this below. Hope the new names are a bit less confusing :)

packages/fetch-and-compile/test/fixture.js
leeftk: Okay thanks for the explanation here. I paired with Tyler and he went into a bit more detail as well as to why this is important. I wasn't understanding at first from the comments but after getting a visual representation I was able to understand what you meant here so thanks for the detailed response, it really helped. I created `.sol` files for the Solidity files and also changed the ABI's to `.json` files. Also to answer an earlier question that you asked, I had trouble finding smaller contracts and it ended up taking more time than I expected. The only smaller contract I was able to find was one for the Arbitrum network so I switched that one out and left the others as is. I guess most contracts these days use the OpenZepplin libraries so even if the main contract is small it comes with a bunch of extra stuff that I don't want to have to deal with. So I chose to just use the contracts I had originally and go with one of the of methods you suggested of how to decrease the fixture. Hope that helps cover everything. Thanks Harry!

cds-amal:
Hey @leeftk it looks like develop was merged into your branch. Can you fallback and do a rebase against develop instead?

packages/fetch-and-compile/test/abi/mainnet.abi.json
```diff
@@ -0,0 +1,2 @@
..."},{"internalType":"uint256","name":"deadline","type":"uint256"}],"name":"swapExactTokensForTokensSupportingFeeOnTransferTokens","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"uint256","name":"amountOut","type":"uint256"},{"internalType":"uint256","name":"amountInMax","type":"uint256"},{"internalType":"address[]","name":"path","type":"address[]"},{"internalType":"address","name":"to","type":"address"},{"internalType":"uint256","name":"deadline","type":"uint256"}],"name":"swapTokensForExactETH","outputs":[{"internalType":"uint256[]","name":"amounts","type":"uint256[]"}],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"uint256","name":"amountOut","type":"uint256"},{"internalType":"uint256","name":"amountInMax","type":"uint256"},{"internalType":"address[]","name":"path","type":"address[]"},{"internalType":"address","name":"to","type":"address"},{"internalType":"uint256","name":"deadline","type":"uint256"}],"name":"swapTokensForExactTokens","outputs":[{"internalType":"uint256[]","name":"amounts","type":"uint256[]"}],"stateMutability":"nonpayable","type":"function"},{"stateMutability":"payable","type":"receive"}]';
```
eggplantzzz, line 1: what a fixture!

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+
+afterEach(function () {
+  //TS can't detect that is a sinon stub so we have to use ts-ignore
+  //@ts-ignore
+  //restoring stub
+  axios.get.restore();
+});
+
+describe("fetchAndCompile", function () {
+  it.only("verifies contract from mainnet", async function () {
```
eggplantzzz, line 33: get rid of this `only`

packages/fetch-and-compile/test/fixture.js
```diff
@@ -0,0 +1,84 @@
...
+        .readFileSync(path.resolve(__dirname, "./sol/polygon.sol"))
+        .toString(),
+      ABI: fs
+        .readFileSync(path.resolve(__dirname, "./abi/polygon.abi.json"))
+        .toString(),
+      ContractName: "GrowthVault",
+      CompilerVersion: "v0.8.9+commit.e5eed63a",
+      OptimizationUsed: "1",
+      Runs: "200",
+      ConstructorArguments:
```
eggplantzzz, line 72: you should probably get rid of the newline here

packages/fetch-and-compile/test/abi/mainnet.abi.json
leeftk: 😂 thanks!

packages/fetch-and-compile/test/fixture.js
leeftk: Is there a way to stop husky from doing this? Every time I create a new commit it goes back lol

eggplantzzz: oh if it is a `prettier` thing then I suppose it is fine

packages/fetch-and-compile/test/sol/arbitrum.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity >=0.7.0 <0.9.0;
  * @title Storage
  * @dev Store & retrieve value in a variable
  */
-contract Storagelsdjflsa {
+contract Storage {
```
eggplantzzz, line 7: a much better contract name ;)

kevinweaver (commented):
Looks like this has already been combed over pretty thoroughly! 👍  Going to leave the seal of approval for @haltman-at

packages/source-fetcher/lib/etherscan.ts
```diff
@@ -54,12 +56,14 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
       "rinkeby",
       "goerli",
       "optimistic",
-      "kovan-optimistic"
+      "kovan-optimistic",
+      "arbitrum",
+      "polygon"
```
kevinweaver, line 61: We should probably add the testnet versions as well.  Arbitrum's testnet etherscan is https://testnet.arbiscan.io/ and polygon's is  https://mumbai.polygonscan.com/

leeftk: Okay I wrote to you in Slack about this. Let me know if you think this should be included in this PR or if we should merge this so you can work on the bridge box and then we can create a separate PR to add those two.

haltman-at: FWIW, my opinion is that it would be OK to save that for later.

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
...
+    stubAxiosGetMethod(
+      "https://api.etherscan.io/api",
+      address,
+      fixture.mainnetData
+    );
+    //TS can't detect that is a sinon stub so we have to use ts-ignore
+    //@ts-ignore
+    axios.get.callThrough();
+    const result = await fetchAndCompile(address, config);
+    const contractNameFromResult = result.sourceInfo.contractName;
```
haltman-at, line 52: Minor suggestion here: Everything comes from `result`, maybethis would be better called `contractNameFromSourceInfo`?  (Same comment for the other two instances, not going to repeat it.)

packages/fetch-and-compile/test/fetch.test.ts
```diff
@@ -0,0 +1,118 @@
+import assert from "assert";
+import "mocha";
```
haltman-at, line 2: I pointed this out elsewhere but given how big the conversation on this PR has gotten you might have missed it, so, I wanted to point out again that this line can be deleted.  Rather than explain in more detail here I'll just link to [where I talked about this before](https://github.com/trufflesuite/truffle/pull/4424#discussion_r783526144).

packages/fetch-and-compile/test/abi/polygon.abi.json
```diff
@@ -0,0 +1 @@
...ype":"address","name":"","type":"address"},{"internalType":"address","name":"","type":"address"},{"internalType":"uint256","name":"","type":"uint256"}],"name":"transferFrom","outputs":[{"internalType":"bool","name":"","type":"bool"}],"stateMutability":"pure","type":"function"},{"inputs":[{"internalType":"address","name":"newOwner","type":"address"}],"name":"transferOwnership","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[],"name":"updateTotalCapital","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"nonpayable","type":"function"},{"inputs":[{"internalType":"uint256","name":"_shares","type":"uint256"}],"name":"withdraw","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[],"name":"withdrawFee","outputs":[{"internalType":"uint256","name":"","type":"uint256"}],"stateMutability":"view","type":"function"},{"inputs":[],"name":"withdrawTokensFee","outputs":[],"stateMutability":"nonpayable","type":"function"},{"inputs":[],"name":"yelLiquidityRouter","outputs":[{"internalType":"address","name":"","type":"address"}],"stateMutability":"view","type":"function"},{"stateMutability":"payable","type":"receive"}]';
```
haltman-at, line 1: You have this and the other ABIs quoted and with a semicolon after.  You should just have the text of the ABI JSON itself; there shouldn't be quotes around them or a semicolon after.

packages/fetch-and-compile/test/fixture.js
```diff
@@ -0,0 +1,84 @@
+const fs = require("fs");
+const path = require("path");
+
+const mainnetData = {
+  status: "1",
+  message: "OK-Missing/Invalid API Key, rate limit of 1/5sec applied",
+  result: [
+    {
+      SourceCode: fs
+        .readFileSync(path.resolve(__dirname, "./sol/mainnet.sol"))
```
haltman-at, line 10: A note on use of `readFileSync`: Rather than first doing `fs.readFileSync(path)` to read the file into a `Buffer` and then doing `buffer.toString()` to render it as a string, you can just do `fs.readFileSync(path, "utf8")` to directly read the file in as a string, which would hopefully avoid two passes over the file.

haltman-at (commented):
Thanks for this @leeftk!  Still a few minor things but I think this should probably *finally* be it.

One other thing in addition to the comments below: Why do the sol files go in a `sol` directory and the ABI files in an `abi` directory?  To my mind it would make more sense to group them by what contract they're for; or alternatively to not group them at all and just throw them all in a single `sources` directory.  I don't know, this probably isn't important, but I thought I'd comment on it.

packages/source-fetcher/lib/etherscan.ts
kevinweaver: Agreed, this is not blocking and fine for later.

packages/fetch-and-compile/test/fetch.test.ts
leeftk: Hmmmm, so I tried removing this line and the tests fail without it. I modeled `fetch.test.ts` after these these tests that @eggplantzzz shared with me https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/test/urlValidation.test.ts and they seem to import `assert` and `{ describe, it }`. Could this be because they're `ts` and not `js`? Also the error I'm getting are `TSError: ⨯ Unable to compile TypeScript:
test/fetch.test.ts:24:1 - error TS2304: Cannot find name 'afterEach'` in case it helps.

haltman-at: Oh, weird!  Yeah I guess I don't know much of anything about using mocha with TS.  I guess maybe I'm just wrong about this one!

haltman-at (approved):
OK, I think this is finally good to go!

packages/fetch-and-compile/test/fetch.test.ts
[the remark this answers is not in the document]
leeftk: We resolved this in a different comment below so closing here.

Issue #4422, opened 2021-11-11
@truffle/contract - Verify json artifact before instantiation


tgbv:
Does `@truffle/contract` do any sort of verification amid instantiating JSON contract artifacts? I mistakenly imported another variables into constructor and it didn't crash. However it did throw an expected exception on `deployed()` method because naturally the artifact was invalid, however I did not know that. 

Have the following code:
```js
import truffleContract from "@truffle/contract";

(async function(){

  const res = await get("/contracts");

  /**
  * @type { Array<Json Contract Artifact Object> }
  */
  const contractsObj = res.data;

  /**
  * @type { {string: Truffle Contract} }
  */
  const contracts = {};
  
  for (const contractObj of contractsObj) {
    const contractTruffle = truffleContract(contractsObj); // mistake here
  
    contractTruffle.setProvider(window.ethereum);
  
    contracts[contractObj.contractName] = contractTruffle;
  }

  await contracts.SomeContract.deployed(); // throws famous exception here
})();
```

There is a mistake, on line 18: instead of using _contractObj_, `truffleContract()` uses _contract**s**Obj_ as parameter. Even thought the artifact is invalid it doesn't throw an error. Execution carries on until line 25, in which the famous one is thrown:
> Uncaught (in promise) Error: Contract has not been deployed to detected network (network/artifact mismatch)

The exception message itself **is valid**, because **I have imported invalid artifacts**, however **I don't know** I imported invalid artifacts because I got **no exception** on line 18. Furthermore if you `console.log(contracts.SomeContract)` it will appear as a normal truffle contract object, which would give no clue for newbies related to debugging. 

I believe we should add artifact checker amid contract instantiation. Is there a reason it shouldn't be?

MicaiahReid:
@tgbv, thanks for the issue! We'll look into adding this validation upfront. Thanks!

tgbv:
@MicaiahReid is truffle used for contracts other than ethereum?

MicaiahReid:
@tgbv Truffle only supports EVM-compatible contracts.

PR #4432, opened 2021-11-15
Add artifact verification before instantiation (issue #4422)

tgbv:
Link to issue: https://github.com/trufflesuite/truffle/issues/4422

Given that functionality for checking contract artifacts is present in `@truffle/contract-schema`,
I integrated that functionality in `@truffle/contract`.

Given that no tests for checking that functionality are written in @truffle/contract-schema/test/schema.js
Or, I do not see them,
I wrote 2 tests for checking contract artifacts amid `Schema.normalize(...)`.

----------------

After this commit, an exception should be thrown if contract artifact passed to `@truffle/contract` constructor is invalid. Unfortunately **I did not write tests** for this behaviour because I don't know where. Should they be in `@truffle/core` ? 🤔

### DIFF
--- a/packages/contract/index.js
+++ b/packages/contract/index.js
@@ -2,10 +2,8 @@ const Schema = require("@truffle/contract-schema");
 const Contract = require("./lib/contract");
 const truffleContractVersion = require("./package.json").version;
 
-const contract = (json = {}) => {
-  const normalizedArtifactObject = Schema.normalize(json, {
-    validate: true
-  });
+const contract = (json = {}, opts = {}) => {
+  const normalizedArtifactObject = Schema.normalize(json, opts);
 
   // Note we don't use `new` here at all. This will cause the class to
   // "mutate" instead of instantiate an instance.
--- a/packages/contract/index.js
+++ b/packages/contract/index.js
@@ -3,7 +3,9 @@ const Contract = require("./lib/contract");
 const truffleContractVersion = require("./package.json").version;
 
 const contract = (json = {}) => {
-  const normalizedArtifactObject = Schema.normalize(json);
+  const normalizedArtifactObject = Schema.normalize(json, {
+    validate: true
+  });
 
   // Note we don't use `new` here at all. This will cause the class to
   // "mutate" instead of instantiate an instance.
--- a/packages/contract-schema/test/schema.js
+++ b/packages/contract-schema/test/schema.js
@@ -22,4 +22,25 @@ describe("Schema", function() {
       assert(abiErrors);
     }
   });
+
+  it("normalizes a correct input", function () {
+    Schema.normalize(MetaCoin);
+  });
+
+  it("throws exception when attempting to normalize an invalid schema", function () {
+    var invalid = {
+      abi: -1
+    };
+
+    try {
+      Schema.normalize(invalid, {
+        validate: true
+      });
+    } catch (err) {
+      var abiErrors = err.errors.filter(function (error) {
+        return error.dataPath === ".abi";
+      });
+      assert(abiErrors);
+    }
+  });
 });
--- a/packages/contract-schema/test/schema.js
+++ b/packages/contract-schema/test/schema.js
@@ -24,7 +24,9 @@ describe("Schema", function() {
   });
 
   it("normalizes a correct input", function () {
-    Schema.normalize(MetaCoin);
+    Schema.normalize(MetaCoin, {
+      validate: true
+    });
   });
 
   it("throws exception when attempting to normalize an invalid schema", function () {


gnidan:
The existing tests should cover this sufficiently. Let's see if they pass!

(Thanks for opening this!)

fainashalts:
@tgbv thank you for opening this PR! It looks like the tests aren't passing, can you take a look? Please reach out here if you need guidance!

tgbv:
@fainashalts those tests are failing on `develop` as well. I don't think they are related to the change I made.

Tests are failing on package `@truffle/resolver`, which I don't know what it's about 🤔 I could use some help from you guys

tgbv:
The way I see it problem is tricky. Tests are failing because JSON artifact inputs from db migrations tests are invalid, or do not match the existing schemas --- an instantiation is attempted inside `@truffle/resolver/lib/resolver.ts` @ line 72. 

That raises some concerns like: 

1. Are the validation schemas from `@truffle/contract-schema/spec` reliable? I can see they are tested inside `@truffle/contract-schema/test` but does that cover literally all Solidity versions?
2. If they are valid, does artifact schema modify with each Solidity version upgrade? Should we cover all those cases?

Ultimately, can we rely on `validate()` from `@truffle/contract-schema`?

tgbv:
I think latest commit will not break the tests. Can you please re-run them?

haltman-at (commented):
So, I don't see how this PR achieves its stated goals.  It doesn't seem to actually add any verification step; the only change to the `contract` code it makes is to allow passing a second `opts` argument (default value `{}`) to the `contract` function.

By itself, this will have no effect.  This will *allow* people to validate their artifacts when they run `contract` if they so desire, but it doesn't cause it to happen by default.  *Allowing* validation doesn't seem to me to accomplish much, as they could already run `Schema.validate` if they really wanted; and anyway, `contract` is I think more likely to be invoked from elsewhere in Truffle (such locations not having been updated) rather than manually.

(Also, giving `contract` a second argument kind of bothers me...)

Also, while the tests added are helpful (because indeed we did not have tests of `Schema.normalize` with the `validate` option set, so thanks for adding those), they don't seem to directly test the functionality added in this PR?  For that, a test of `contract` should be added to the `contract-tests` package.

But overall it seems to me that this PR doesn't achieve its stated purpose?  And I'm wary of adding a second argument to `contract`.

tgbv:
> So, I don't see how this PR achieves its stated goals. 

You are right, this PR doesn't achieve what it was intended. 

Adding a default artifact verification causes [tests](https://github.com/trufflesuite/truffle/runs/4242428826?check_suite_focus=true#step:6:1019) related to db package to fail. That discouraged me to carry on with coding, as now I see implementing this feature 'out of the box' may require a lot of modifications across different packages, which may lead to drawback compatibility issues --- I don't know how @trufflesuite is versioned, which changes get to be productive and which don't.

> Allowing validation doesn't seem to me to accomplish much, as they could already run Schema.validate if they really wanted

You are right as well. Because of what I stated earlier, I chose to add the validation as an optional parameter, to keep package backward-compatible. I also have some concerns with the validation [stated here](https://github.com/trufflesuite/truffle/pull/4432#issuecomment-973805580).

> Also, while the tests added are helpful (because indeed we did not have tests of Schema.normalize with the validate option set, so thanks for adding those), they don't seem to directly test the functionality added in this PR? For that, a test of contract should be added to the contract-tests package.

I did not know where to add the actual contract instantiation tests, now you gave me a clue, thanks! 
Regarding the other tests with normalization+validation, should I create another issue with a separate PR for it?

> And I'm wary of adding a second argument to contract.

It's my first contribution, certainly I do not seek to reinvent the wheel 😄  Since `Schema.validate` exists and as you said can be called upfront, I believe we can drop this PR unless you have other suggestions.

haltman-at:
> Adding a default artifact verification causes [tests](https://github.com/trufflesuite/truffle/runs/4242428826?check_suite_focus=true#step:6:1019) related to db package to fail. That discouraged me to carry on with coding, as now I see implementing this feature 'out of the box' may require a lot of modifications across different packages, which may lead to drawback compatibility issues --- I don't know how @trufflesuite is versioned, which changes get to be productive and which don't.

Ah, yeah.  @gnidan, @fainashalts, any idea what to do about this?

> > Also, while the tests added are helpful (because indeed we did not have tests of Schema.normalize with the validate option set, so thanks for adding those), they don't seem to directly test the functionality added in this PR? For that, a test of contract should be added to the contract-tests package.

Ah, yeah, the package structure can be a bit opaque at times.  Sorry about that!  Yes for technical reasons (that I don't recall offhand, I wasn't involved with this) the tests of `@truffle/contract` live in the `contract-tests` package.
 
> I did not know where to add the actual contract instantiation tests, now you gave me a clue, thanks! Regarding the other tests with normalization+validation, should I create another issue with a separate PR for it?

If you like!

> > And I'm wary of adding a second argument to contract.
> 
> It's my first contribution, certainly I do not seek to reinvent the wheel smile Since `Schema.validate` exists and as you said can be called upfront, I believe we can drop this PR unless you have other suggestions.

Hm, that might be the thing to do... @gnidan, @eggplantzzz, any opinion on this?

eggplantzzz:
Personally @haltman-at the second argument to contract doesn't really bother me. It seems it could be useful in the future if we want to have other options available when creating contract objects; but I admit, it is kind of abstract and I don't have concrete ideas offhand.

I see this PR (in its current form) as maybe a nice convenience to offer to people who use this package. I don't really have strong feelings about all this though so if you dislike these changes, I'll be happy to stand aside.

haltman-at:
Hm, I guess we should figure out what we're doing with this... @gnidan, any opinion on this?

packages/contract/index.js
```diff
@@ -2,8 +2,8 @@ const Schema = require("@truffle/contract-schema");
 const Contract = require("./lib/contract");
 const truffleContractVersion = require("./package.json").version;
 
-const contract = (json = {}) => {
-  const normalizedArtifactObject = Schema.normalize(json);
+const contract = (json = {}, opts = {}) => {
+  const normalizedArtifactObject = Schema.normalize(json, opts);
```
gnidan, line 6: If we're adding an argument to this function, I think we should namespace things (and don't abbreviate `opts`)

```suggestion
const contract = (json = {}, options = {}) => {
  const normalizedArtifactObject = Schema.normalize(json, options.normalize || {});
```

gnidan (changes requested):
I think adding an option is fine, but let's make sure it's the last option we'll need to add (via namespacing)

fainashalts:
I'm not sure offhand why the DB tests would fail as a result of something like this, but my hunch is that there is something going on with the type when you add a new argument. 🤷‍♀️

eggplantzzz:
Hey @tgbv, do you feel like pursuing this further? Let us know if you want to keep working on this! Thanks!

tgbv:
> Hey @tgbv, do you feel like pursuing this further? Let us know if you want to keep working on this! Thanks!

Hello, I lost track of it. Given how app is structured, I will not continue with it. Feel free to do what you wish with the PR 😃

eggplantzzz:
Ok, thanks for letting us know!

eggplantzzz:
Closing this for issue maintenance, thanks!

PR #4436, opened 2021-11-15
Internal improvement: Remove reporters

eggplantzzz:
This PR gets rid of the reporters package. All of the reporter's duties are now handled by the event system. Much of the code responsible for logging has just been ported over to @truffle/events.

**A note**: One other interesting thing to note about this PR is that it changes how the event system receives and updates configuration. Currently the event system only receives the limited amount of config that it uses; something like a logger, the quiet option, and something else. This PR makes it accept the whole config object and creates a reference to it on the Subscriber. It then updates this reference every time the options are updated. I thought this was a better way to do this since the implementation in the reporters package required several new options and it would start to get a bit unwieldy to manually accept each one of those options. One can imagine this getting out of hand.


eggplantzzz:
Oy @gnidan, can I get a review on this? I figure you are the person to ask.

packages/events/package.json
```diff
@@ -23,6 +23,7 @@
   },
   "dependencies": {
     "emittery": "^0.4.1",
-    "ora": "^3.4.0"
+    "ora": "^3.4.0",
+    "web3-utils": "^1.6.0"
```
gnidan, line 27: Should maybe make this match the version range we have elsewhere (since I noticed it shows up as new in `yarn.lock`)

packages/migrate/Migration.js
```diff
@@ -11,11 +10,9 @@ const ResolverIntercept = require("./ResolverIntercept");
 const { getTruffleDb } = require("@truffle/db-loader");
 
 class Migration {
-  constructor(file, reporter, config) {
+  constructor(file, config) {
```
gnidan, line 13: Do we need an `expect.options(config, ["events"])` here?

gnidan: Alternatively we could define an `emit` method inside `Migration` here, like I've suggested for `Deployment`... and then we could put the `if` guard there and have them both conform to each other. (I don't think we want to have Migration call `deployer.emit`)

packages/migrate/Migration.js
```diff
@@ -89,18 +84,23 @@ class Migration {
 
       if (Migrations && Migrations.isDeployed()) {
         const message = `Saving migration to chain.`;
-
         if (!this.dryRun) {
           const data = { message: message };
-          await this.emitter.emit("startTransaction", data);
+          await options.events.emit(
```
gnidan, line 89: This isn't inside an `if` statement (should it be?)

gnidan (changes requested):
This is awesome! Just have some minor stylistic suggestions and a question or two.

yarn.lock
```diff
@@ -28699,6 +28699,19 @@ web3-utils@^1.0.0-beta.31:
     underscore "1.9.1"
     utf8 "3.0.0"
 
+web3-utils@^1.6.0:
```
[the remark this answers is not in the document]
eggplantzzz: no idea!

packages/events/package.json
eggplantzzz: ah I see, I must have just used yarn to install it without checking the version

packages/migrate/Migration.js
eggplantzzz: yeah in general I don't oppose the idea of having emit methods in each of these classes/utilities - I'll go ahead and implement them

packages/deployer/src/deployment.js
```diff
@@ -161,64 +163,36 @@ class Deployment {
...
-   * @private
-   * @param  {Object} parent Deployment instance. Local `this` belongs to promievent
-   * @param  {String} hash   tranactionHash
-   */
-  async _hashCb(parent, state, hash) {
-    const eventArgs = {
-      contractName: state.contractName,
-      transactionHash: hash
-    };
-    state.transactionHash = hash;
```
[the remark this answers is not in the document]
eggplantzzz: I think this was removed because as far as I can tell it is not used anywhere in the code. Please double-check me on this but I couldn't find a reference to this method anywhere.

packages/deployer/src/deployment.js
```diff
@@ -161,64 +163,36 @@ class Deployment {
...
-    const eventArgs = {
-      contractName: state.contractName,
-      receipt: receipt
-    };
-
-    // We want this receipt available for the post-deploy event
-    // so gas reporting is at hand there.
-    state.receipt = receipt;
-    await parent.emitter.emit("receipt", eventArgs);
-    this.removeListener("receipt", parent._receiptCb);
```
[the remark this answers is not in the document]
eggplantzzz: Same comment about this method as about `_hashCb`

packages/deployer/src/deployment.js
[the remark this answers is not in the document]
gnidan: Hm, maybe keep it in and remove it as a separate PR, to make it easy to revert later if it breaks something?

I don't feel too strongly about it. Would be curious what was using it in the past (if anything)

packages/events/package.json
gnidan: That's what I'm here for

packages/deployer/src/deployment.js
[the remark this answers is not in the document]
gnidan: Hm, am I missing something? I don't see a response to [this question](https://github.com/trufflesuite/truffle/pull/4436#discussion_r779176172), if that's what you're talking about.

packages/migrate/Migration.js
gnidan: Cool, then the original comment here about needing `expect.options` should really only apply if we want to keep the missing `if` statement that I call out [here](https://github.com/trufflesuite/truffle/pull/4436#discussion_r779196154).

packages/deployer/src/deployment.js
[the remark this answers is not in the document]
eggplantzzz: I re-added the unused things that got removed. I think that comment got deleted or something since it was "un-deleted".

packages/deployer/test/helpers/eventSystem.js
```diff
@@ -0,0 +1,72 @@
...
+    }
+  }
+};
+
+module.exports = {
+  mockEventsSystem,
+  getAllEventsByName,
+  preDeployOccurredForNames,
+  postDeployOccurredForNames,
+  linkingOccurredForName
```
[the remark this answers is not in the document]
eggplantzzz: But would that mean `getAllEventsByName` would take `events` as well as an argument?

packages/migrate/index.js
```diff
@@ -4,6 +4,7 @@ const glob = require("glob");
 const expect = require("@truffle/expect");
 const Config = require("@truffle/config");
 const Migration = require("./Migration");
+const emitEvent = require("./emitEvent");
```
gnidan, line 7: Could we change this to like `makeEmit`? Something like:
```typescript
const emit = makeEmit(options);
emit("migrate:dryRun:notAccepted");
```

What do you think?

eggplantzzz: Is the purpose of this that you don't want to pass `options` each time?

eggplantzzz: Hmm but this is kind of annoying because on the one hand, `Migration` is a class and we can set this new thing `this.emit`. `Migrate` is just an object with utilities in; there is no instantiation so we'd have to create a new one in each context in which it is used, which sucks. Did you have a different kind of idea?

packages/deployer/test/helpers/eventSystem.js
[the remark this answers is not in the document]
gnidan: We can probably skip this for now.

packages/migrate/index.js
gnidan: This is probably fine (and not worth blocking merge over)

gnidan (approved):
Taking another look at this, I think this is good to go :)

Thanks @eggplantzzz!

PR #4440, opened 2021-11-17
Remove usage of promisify in decoder

gnidan:
Since this was causing problems for tx-insight when trying to run decoder inside a browser extension


ramonbrugman:
****

haltman-at (commented):
Oh boy, manual promises!  My one note here is I'd suggest adding a comment in the code saying why you're doing this, so that someone else doesn't come along and undo it later, because my inclination is always to replace manual promises when I see them. :P

gnidan:
makes sense @haltman-at - lemme know how the comment looks

haltman-at (approved):
Seems good to me!

gnidan:
Thanks @haltman-at!

Issue #4379, opened 2021-10-29
`truffle config ...` does not report anything after it is run


eggplantzzz:
After you run `truffle config <something>`, it returns you to the prompt without any printout. This is very confusing and leaves the user not knowing whether running the command was successful or if it failed.

Try running `truffle config set analytics true` and you can see the process just ends without any confirmation.

cds-amal:
It would be nice to add a `truffle config list` command to see current settings similar to npm's behavior.

gnidan:
This is done and now released in [Truffle v5.4.20](https://github.com/trufflesuite/truffle/releases/tag/v5.4.20). Thanks!

PR #4443, opened 2021-11-18
Add command `truffle config list`

MrGrizz:
This PR add command 'truffle config list' mentioned in #4379

### DIFF
--- a/packages/core/lib/commands/config.js
+++ b/packages/core/lib/commands/config.js
@@ -24,7 +24,7 @@ const command = {
       },
       {
         option: "list",
-        description: "List a Truffle config option values.",
+        description: "List all Truffle config values",
       }
     ],
     allowedGlobalOptions: []
@@ -78,7 +78,7 @@ const command = {
 
       return;
     } else if (command.list) {
-      options.logger.log("; Truffle configs");
+      options.logger.log("; Truffle config values");
       options.logger.log(`analytics = ${googleAnalytics.getAnalytics()}`);
     } else {
       const config = Config.detect(options);
--- a/packages/core/lib/commands/config.js
+++ b/packages/core/lib/commands/config.js
@@ -167,7 +167,6 @@ const parse = function (args) {
       break;
     }
     case "list": {
-      set = false;
       list = true;
       break;
     }
--- a/packages/core/lib/commands/config.js
+++ b/packages/core/lib/commands/config.js
@@ -3,7 +3,7 @@ const command = {
   description: "Set user-level configuration options",
   help: {
     usage:
-      "truffle config [--enable-analytics|--disable-analytics] [[<get|set> <key>] [<value-for-set>]]",
+      "truffle config [--enable-analytics|--disable-analytics] [<list>] [[<get|set> <key>] [<value-for-set>]]",
     options: [
       {
         option: "--enable-analytics",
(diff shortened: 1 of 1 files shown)


gnidan:
Love this attention to `truffle config`! Thank you!

cds-amal (commented):
Thanks for this @MrGrizz! I have one issue with the PR around git practice. Can you rebase the last commit against upstream develop instead of merging?

MrGrizz:
@cds-amal Done. Sorry for that mistake.

packages/core/lib/commands/config.js
```diff
@@ -158,6 +166,11 @@ const parse = function (args) {
 
       break;
     }
+    case "list": {
+      set = false;
```
eggplantzzz, line 170: You shouldn't need to set the value for `set` as it is already `false`. Actually I have no idea why this happens on line 120 either.

packages/core/lib/commands/config.js
```diff
@@ -21,6 +21,10 @@ const command = {
       {
         option: "set",
         description: "Set a Truffle config option value."
+      },
+      {
+        option: "list",
+        description: "List a Truffle config option values.",
```
eggplantzzz, line 27: So I think this should be "List all Truffle config values." That's what this feature is meant to do, right? List all the stored Truffle config values?

packages/core/lib/commands/config.js
```diff
@@ -73,6 +77,9 @@ const command = {
       }
 
       return;
+    } else if (command.list) {
+      options.logger.log("; Truffle configs");
```
eggplantzzz, line 81: And this should probably read "Truffle config values"

packages/core/lib/commands/config.js
MrGrizz: Done.

packages/core/lib/commands/config.js
MrGrizz: You're right. Done.

eggplantzzz:
I have one or two more nitpicks and then this should be good to go as far as I'm concerned. Thanks!

packages/core/lib/commands/config.js
```diff
@@ -73,6 +77,9 @@ const command = {
       }
 
       return;
+    } else if (command.list) {
+      options.logger.log("; Truffle config values");
```
eggplantzzz, line 81: Could you remove the `; ` from the beginning of this string? I changed it locally to look like this at the command line.

Screen Shot 2021-12-02 at 10 36 53 AM

packages/core/lib/commands/config.js
```diff
@@ -21,6 +21,10 @@ const command = {
       {
         option: "set",
         description: "Set a Truffle config option value."
+      },
+      {
+        option: "list",
+        description: "List all Truffle config values",
```
eggplantzzz, line 27: Can you add a `.` at the end of this sentence. That seems to be convention for all descriptions.

packages/core/lib/commands/config.js
MrGrizz: Removed. I just add `;` because I followed `npm config list` :D

eggplantzzz (approved):
Thanks @MrGrizz, I like it!

cds-amal (approved):
LGTM! Thanks @MrGrizz

Issue #4446, opened 2021-11-19
@truffle/fetch-and-compile doesn't report which fetcher it used


gnidan:
## Issue

This package supports fetching from both Etherscan and from Sourcify, but it does not provide any output to say which of these was used. This information is a **must** for certain use cases, where you want to be dang-sure that you can verify a contract instance.

## Steps to Reproduce

1. Run this:
    ```typescript
    import { fetchAndCompile } from "@truffle/fetch-and-compile";
    /* ... */
    const { 
      fetchedFrom // (or another name)
    } = await fetchAndCompile(/* ... */);
    ``` 
2. Observe that there is no such output.

## Expected Behavior

I'd expect to find this information.

## Actual Results

There is no such output.

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.4.21
* node version (`node --version`):
* npm version (`npm --version`):

haltman-at:
To be clear, this issue is only about the `fetchAndCompile` function in specific, rather than the package as a whole, right?  Like I don't think we need to make the debugger version do this, right?  Assuming it's just the former, I think this should be pretty straightforward.

gnidan:
Yep!

PR #4447, opened 2021-11-19
Enhancement: Have `fetchAndCompile` return name of fetcher used

haltman-at:
Addresses #4446.

The `SingleRecognizer` now records and returns the name of the fetcher used, in addition to the other information.  Note that to do this I put `fetchedVia` into the `FetchAndCompileResult` type; so I removed the third argument from `addCompiledInfo` as it was now redundant.  I think we can reasonably say that's not really a breaking change, as it doesn't break either `fetchAndCompile` or `fetchAndCompileForDebugger`; it only breaks `fetchAndCompileForRecognizer`, which... c'mon, that's pretty internal (it's not even used elsewhere in Truffle although I did export it).

The functionality of `fetchAndCompileForDebugger` remains unchanged.

If we really are worried about that sort of thing, I can redo it in a truly non-breaking way, but this was the most straightforward way.

Note this PR has no tests, because fetch-and-compile has no tests; but once #4424 is merged I can go back and add tests of this.  (Or if @leeftk merges this in then he can add the tests on his PR, but I don't know that he wants that extra work. XD )

### DIFF
--- a/packages/fetch-and-compile/lib/debug.ts
+++ b/packages/fetch-and-compile/lib/debug.ts
@@ -79,14 +79,13 @@ export class DebugRecognizer implements Recognizer {
 
   async addCompiledInfo(
     info: FetchAndCompileResult,
-    address: string,
-    fetcherName: string
+    address: string
   ): Promise<void> {
     debug("compileResult: %O", info.compileResult);
     const compilations = info.compileResult.compilations;
     const shimmedCompilations = Codec.Compilations.Utils.shimCompilations(
       compilations,
-      `externalFor(${address})Via(${fetcherName})`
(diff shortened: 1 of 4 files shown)


gnidan:
@haltman-at it might be a bit of extra work for you to maintain the branch this way, but if you rebase this against #4424 then you and @leeftk could collaborate to get tests here in this PR!

gnidan (approved):
Approving this without tests because this might end up being a blocker on the tx-insight side otherwise

cds-amal (commented):
I'd prefer if we had tests for this but @gnidan approved! So it's your call @haltman-at . Perhaps the tests could be add on work after #4424 is merged?

haltman-at:
Yes, that's my intention.

PR #4448, opened 2021-11-19
Bug: view function can cause insufficient funds error

hellwolf:
In addition to #4397

We also noticed that sometimes view function can cause the following errors (on polygon, could be specific to their node?):

```
err: insufficient funds for gas * price + value: address 0xd66E40b0c30595bEc72153B502aC1E0c4785991B have 13763260682756217913 want 50000000000000000000000 (supplied gas 500000)
```

This occurs when using truffle, and set a default gas price in the network config, for example:

```
        matic: {
            gasPrice: +process.env.MATIC_GAS_PRICE || 30e9,
        },
```

The proposed solution is to remove "gasPrice" parameters for view calls.

## How to reproduce

https://github.com/hellwolf/truffle_contract_staticcall_gasprice_issue

        // This will trigger the bug, by using high default gasPrice
        // `GAS_PRICE=1e12 npx truffle --network matic exec scripts/test.js`
        // but without GAS_PRICE setting, it works
        // `npx truffle --network matic exec scripts/test.js`

```
$ npx truffle --network matic exec scripts/test.js 
Using network 'matic'.

usdcx name:  Super USDC (PoS)
```

```
$ GAS_PRICE=1e12 npx truffle --network matic exec scripts/test.js 
Using network 'matic'.

err: insufficient funds for gas * price + value: address 0x627306090abaB3A6e1400e9345bC60c78a8BEf57 have 62406659 want 600000000000000000000000000 (supplied gas 50000000)
Truffle v5.4.21 (core: 5.4.21)
Node v14.18.1
```

### DIFF
--- a/packages/contract/lib/execute.js
+++ b/packages/contract/lib/execute.js
@@ -99,7 +99,7 @@ const execute = {
       // view function fail with "insufficient funds for gas" errors"
       // For calls intending to skip network check, it should also skip this check to.
       // A workaround would be simply having the gasPrice field delted
-      params.gasPrice;
+      delete params.gasPrice;
       return { args, params };
     }
     const network = await constructor.detectNetwork();
--- a/packages/contract/lib/execute.js
+++ b/packages/contract/lib/execute.js
@@ -95,6 +95,11 @@ const execute = {
     }
     //skipNetworkCheck flag used to skip network call for read data (calls type) methods invocation
     if (skipNetworkCheck) {
+      // Having gasPrice set (often as default gas price in truffle-config) can also cause
+      // view function fail with "insufficient funds for gas" errors"
+      // For calls intending to skip network check, it should also skip this check to.
+      // A workaround would be simply having the gasPrice field delted
+      params.gasPrice;
       return { args, params };
     }
     const network = await constructor.detectNetwork();


haltman-at:
Hi, could you open an issue for this with reproduction steps?  It's not entirely clear to me if the underlying problem is actually as you describe, and I think this requires some investigation.  I think the rest of us need to be clear on exactly what's going on here before we go ahead with this.  Thanks!  (I could review and suggest changes to what you did do, but I think that would be premature if we're not yet certain that this is indeed the underlying problem.)

hellwolf:
@haltman-at 

How to reproduce

https://github.com/hellwolf/truffle_contract_staticcall_gasprice_issue

        // This will trigger the bug, by using high default gasPrice
        // `GAS_PRICE=1e12 npx truffle --network matic exec scripts/test.js`
        // but without GAS_PRICE setting, it works
        // `npx truffle --network matic exec scripts/test.js`

haltman-at:
Thanks!  Would you mind opening this as an issue though?

hellwolf:
> Thanks! Would you mind opening this as an issue though?

Sure! https://github.com/trufflesuite/truffle/issues/4457

haltman-at:
Cool, thank you!

haltman-at (changes requested):
OK, now that I understand the issue better the idea makes sense.  I don't think this is quite the right implementation, though.

haltman-at:
Argh!  My actual detailed comment got lost.  Rewriting it...

packages/contract/lib/execute.js
```diff
@@ -95,6 +95,11 @@ const execute = {
     }
     //skipNetworkCheck flag used to skip network call for read data (calls type) methods invocation
     if (skipNetworkCheck) {
+      // Having gasPrice set (often as default gas price in truffle-config) can also cause
+      // view function fail with "insufficient funds for gas" errors"
+      // For calls intending to skip network check, it should also skip this check to.
+      // A workaround would be simply having the gasPrice field delted
+      delete params.gasPrice;
```
haltman-at, line 102: I have a few comments here:

1. If the `skipNetworkCheck` variable is going to now also perform this function, it should probably be renamed to `isCall` or something.  (Although, if you see my comment (4) below, you'll see that I think that actually this is the wrong place for this code, so there wouldn't be a need to rename the variable after all.)
2. I'm wary of using the `delete` operator here; it's probably fine, but just to be safe, you might be better off writing something like `params = { ...params, gasPrice: 0 }` or similar.  (Or perhaps `undefined` rather than `0` if you want to be even safer.)
3. The `maxFeePerGas` and `maxPriorityFeePerGas` fields can also cause this problem, so they should be treated similarly.
4. I think this isn't the right place for this code, as per [this comment](https://github.com/trufflesuite/truffle/issues/4457#issuecomment-977276979) on the issue.  With the code as it is, it will not only delete any defaults, but also any ones that are manually set, which I don't think we should be doing.  I think instead this code should live in the `getTxParams` function in `utils/index.js`.  Then, you could have an `isCall` flag, and if it were set you could have it ignore certain fields from `constructor.class_defaults`.  So defaults would be ignored, but manually set ones would be respected.

So, I think this is a good start, but not quite the right solution.  But thanks for kicking this off and bringing the issue to our attention!

hellwolf: Thank you, please help to find the right place to fix. This pull request can be seen as just a way to showing the problem!

haltman-at: OK!  I'm a little unclear here, does that mean you are intending to turn this over to us rather than complete it yourself?  We can do that if you don't have time, but you should say so explicitly if that's your intention... we all already have quite a bit to do, after all, and will take the approach that's less work for us if possible. :)

wxy:
@hellwolf
Perhaps this problem can be solved by using type 2 transactions, i.e. instead of gas/gasprice, use maxFeePerGas/maxPriorityFeePerGas. I had a similar problem on the main web, also see issue #4482 and PR #4516.

haltman-at:
> Perhaps this problem can be solved by using type 2 transactions, i.e. instead of gas/gasprice, use maxFeePerGas/maxPriorityFeePerGas. I had a similar problem on the mainnet, also see issue #4482 and PR #4516.

No, I've tested it and those options have the same issues; see my review comment above.

packages/contract/lib/execute.js
haltman-at: OK, since it's been 29 days, I'm going to take this over if you don't mind... I think I'm going to put up a new PR that supersedes this one, but let me know if you still want to finish this yourself!

haltman-at: OK, I've put up PR #4606 to finish this.

hellwolf: Hi @haltman-at , thanks for taking it forward. I didn't have time to follow up further since then.

haltman-at: No problem!

haltman-at:
Closing this as superseded by #4606.

Issue #4437, opened 2021-11-16
Allow the numbers of instructions displayed in Truffle debugger to be configurable


dongmingh:
## Issue

Truffle debugger p command displays a group of instructions: the current instruction, 3 previous instructions and 3 upcoming instructions, see PR #4412.  This is to request that the debugger will make the numbers of previous and upcoming instructions configurable.

## Environment

* Truffle version (`truffle version`): v5.4.11 does not support the group instructions display yet

haltman-at:
A note, if someone is going to do this, I'd suggest making the `l` command similarly configurable.  (That might be easier to start with, honestly...)

kevinbluer:
Closing as this is addressed in https://github.com/trufflesuite/truffle/releases/tag/v5.4.24

PR #4455, opened 2021-11-22
Allow the display of the number of instruction to be configurable

dongmingh:
This PR addresses #4437 by making the display of the number of
instructions and source file configurable.

Use `p -<num>|+<num>` to configure number of instructions before and
after the current instruction. The default is to display 3 instructions
before and 3 instructions after the current instruction.

Use `l -<num>|+<num>` to configure the number of line of source file
before and after the current line of source file. The default is to
display 5 lines before and 3 lines after the current line.


kevinweaver (approved):
👍 Makes sense, I appreciate the comments!  I'm not familiar if or how we test changes at this level, so I would defer to others if they think it's worth the lift to add testing.

packages/core/lib/debug/interpreter.js
```diff
@@ -738,6 +750,11 @@ class DebugInterpreter {
   updatePrintouts(userArgs, selections, printOuts) {
     let tempPrintouts = new Set();
     for (let argument of userArgs) {
+      // skip line number options
+      if (((argument[0] === "-") || (argument[0] === "+")) &&
```
haltman-at, line 754: OK, this part I think has some style problems.

Firstly, I would strongly suggest that you not use multi-line conditionals (or other blocks) without braces.  These can be something of a hazard.  If you have a conditional, either it should use braces, or it should fit on a single line.

Secondly, there's some repetition here with this check on the following lines.  I'd like to suggest that this be consolidated somehow.

packages/core/lib/debug/interpreter.js
```diff
@@ -738,6 +750,11 @@ class DebugInterpreter {
...
+        (!isNaN(argument.slice(1))))
```
haltman-at, line 755: Hm, using `isNaN` like this is a bit opaque.  At least to me; maybe it's common elsewhere, I dunno.  I'd suggest changing it to `isNaN(Number(...))` for clarity, and maybe adding a comment.  Maybe that's overkill, I dunno, but I'd suggest clarifying it somehow at least.

packages/core/lib/debug/interpreter.js
haltman-at: Actually, hold on -- on looking at this some more... do we even need to explicitly skip these?  It seems to me that the remaining code here will ignore any numbers put in just fine.  I'm not sure we need an explicit skip here; I suspect this addition can just be deleted.  Am I missing something?  (Of course if you do delete it, it's probably worth putting in a comment about this. :) )

packages/core/lib/debug/interpreter.js
```diff
@@ -761,6 +778,21 @@ class DebugInterpreter {
     }
     return tempPrintouts;
   }
+
+  // update the numbers of lines to be printed from options -<num>|+<num> from user args
+  updatePrintoutLines(userArgs, currentLines) {
```
haltman-at, line 783: I think this function could use a better name... after all, it doesn't actually *update* anything, does it?

packages/core/lib/debug/interpreter.js
```diff
@@ -761,6 +778,21 @@ class DebugInterpreter {
...
+    let beforeLines = currentLines[0], afterLines = currentLines[1];
```
haltman-at, line 784: Note that you can use destructuring to write this as just `let [beforeLines, afterLines] = currentLines;`.  (Of course, really you should be using objects rather than arrays, but you can destructure those too. :) )

packages/core/lib/debug/interpreter.js
```diff
@@ -761,6 +778,21 @@ class DebugInterpreter {
...
+    for (let argument of userArgs) {
```
haltman-at, line 785: This `let` (and the following one) can both be a `const`, btw.  You should generally use `const` when possible.  This area of the code hasn't been touched in a while so it has a lot of `let`s that don't need to be, but when writing new code you should try to use `const` when possible.

packages/core/lib/debug/interpreter.js
```diff
@@ -761,6 +778,21 @@ class DebugInterpreter {
...
+      let newLines = Number(argument.slice(1));
+      if (isNaN(newLines)) continue;
+      if (argument[0] === "-") {
+        beforeLines = newLines;
+      } else if (argument[0] === "+") {
+        afterLines = newLines;
+      }
```
haltman-at, line 792: So, any arguments that aren't of the correct form will just be ignored.  I think that's a reasonable decision for the debugger (it's not something I'd suggest in most contexts, but in the debugger I think it's OK), but I think it's worth making explicit with a comment.

packages/core/lib/debug/interpreter.js
```diff
@@ -628,7 +634,13 @@ class DebugInterpreter {
       case "l":
         if (this.session.view(session.status.loaded)) {
           this.printer.printFile();
-          this.printer.printState(LINES_BEFORE_LONG, LINES_AFTER_LONG);
+          // update the numbers of lines to be printed
+          const {beforeLines, afterLines} = this.updatePrintoutLines(
+            splitArgs,
+            this.printer.stateLines
+          );
+          this.printer.stateLines = [beforeLines, afterLines];
```
haltman-at, line 642: Hm, this is different from the functionality we discussed on Slack, and this is the other reason I'm hitting "request changes".  This line of code means that commands like `l +5` will have a persistent effect.  Did you decide to do it differently?  Can you explain what your intention here was?

This sort of thing is why I really recommend using your initial PR writeup to describe in more detail the decisions you made when writing the PR.  You can use it to highlight decisions and tradeoffs, making it clear what downsides are deliberate tradeoffs, vs which are simply accidents or bugs.  It can be very helpful both to reviewers, and to later people who need to investigate the changes!

packages/core/lib/debug/printer.js
```diff
@@ -72,6 +72,14 @@ class DebugPrinter {
...
     this.sectionPrintouts = new Set(["bui", "glo", "con", "loc"]);
     this.sections = ["bui", "glo", "con", "loc"]; //should remain constant
+
+    // numbers of instructions before and after the current instruction to be printed
+    // used by command (p)
+    this.instructionLines = [3, 3];
+
+    // numbers of lines before and after the current line to be printed
+    // used by command (l) and others
+    this.stateLines = [5, 3];
```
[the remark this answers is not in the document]
haltman-at: (Actually, it's possible this variable might not need to exist at all; see other comments regarding that...)

packages/core/lib/debug/printer.js
```diff
@@ -72,6 +72,14 @@ class DebugPrinter {
     //   loc: Local variables
     this.sectionPrintouts = new Set(["bui", "glo", "con", "loc"]);
     this.sections = ["bui", "glo", "con", "loc"]; //should remain constant
+
+    // numbers of instructions before and after the current instruction to be printed
+    // used by command (p)
+    this.instructionLines = [3, 3];
```
haltman-at, line 78: Just noting the array thing here too, for reference

haltman-at (changes requested):
This seems mostly pretty good functionality-wise.  There are one or two things that need to change, but I'll get to those below.

I also want to say @dongmingh that this seems like a big improvement in style over the previous few PRs, this one was definitely easier to read and review, particularly due to the use of helpful code comments.  Sorry about taking so long to get to it!  I do still have a number of comments on style that I think should be addressed, but they're not that large, I don't think.

One general comments that don't apply to any specific lines: In addition to the changes here so far, the help text in `debug-utils` will need updating.

packages/core/lib/debug/interpreter.js
```diff
@@ -628,7 +634,13 @@ class DebugInterpreter {
       case "l":
         if (this.session.view(session.status.loaded)) {
           this.printer.printFile();
-          this.printer.printState(LINES_BEFORE_LONG, LINES_AFTER_LONG);
```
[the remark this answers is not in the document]
dongmingh: These two constants are used in the printErrorLocation() call in case s.

packages/core/lib/debug/printer.js
```diff
@@ -135,8 +143,8 @@ class DebugPrinter {
   }
 
   printState(
-    contextBefore = 2,
-    contextAfter = 0,
+    contextBefore = this.stateLines[0],
```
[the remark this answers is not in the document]
dongmingh: Thanks for sharing the design concepts.

My thinking was that debugger always printouts the user configured numbers.  Function printState() is called by commands `p`, `l`, `;`, `Y`, `r`, and `t`, hence I make the changes in the printState() to have them all printout the user configured numbers and randomly choose 5 before and 3 after as default.

Based on the design concepts, may we do the following?
1. For `l` command, set 2 lines before and no lines after as default for source printout.  For `p` command, set 3 before and 3 after as default for instruction printout. Once the numbers are configured, they will be saved and used for printout respectively.

2. For source printout, many commands access printState(), only have commands `p`, `l`, and `;` to print out the user configured numbers (default 2-and-0) and the rest commands remain unchanged, i.e., print 2 before and 0 after always.

packages/core/lib/debug/interpreter.js
dongmingh: You are right, no need to skip these since the remaining code will ignore any numbers. The skip line code will be deleted.

packages/core/lib/debug/interpreter.js
dongmingh: The code is deleted, see the comment above.

packages/core/lib/debug/interpreter.js
dongmingh: how about parsePrintoutLines()?

packages/core/lib/debug/interpreter.js
dongmingh: No, I do not mean to do it differently.  Actually, parsePrintoutLines() need to do `if -/+ follows by space, continue`.

packages/core/lib/debug/interpreter.js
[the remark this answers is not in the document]
haltman-at: Oh, I see!  I missed that, thanks.  That raises the question: Should that case, or `printErrorLocation`, make use of the settings you're introducing?  I think, if they're going to be saved, it probably should be.  (And if they're not going to be saved, maybe `s` should also get these options?  It could share the behavior with`l`, since they're both about printing out source code.)

packages/core/lib/debug/interpreter.js
haltman-at: Makes sense!

packages/core/lib/debug/interpreter.js
haltman-at: OK!

packages/core/lib/debug/interpreter.js
```diff
@@ -603,7 +603,13 @@ class DebugInterpreter {
         await this.setOrClearBreakpoint(splitArgs, false);
         break;
       case "p":
-        //first: process which locations we should print out
+        // update the numbers of instructions to be printed
+        const {beforeLines, afterLines} = this.updatePrintoutLines(
+          splitArgs,
+          this.printer.instructionLines
+        );
+        this.printer.instructionLines = [beforeLines, afterLines];
```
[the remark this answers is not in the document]
haltman-at: Yeah, I think it's probably fine to leave this as-is for now and worry about it in the future.  I don't have any great idea about how to do this better in a way that's nice.  Maybe @gnidan has an opinion, but if he doesn't, I think this is fine.

I guess maybe we could just for now stick comments on the variables, where they're declared in `printer.js`, and note that they're public variables and may be modified from the outside?  Would be good to do that on the other ones that are like this too.  Man, somehow all of us missed this issue before.

packages/core/lib/debug/printer.js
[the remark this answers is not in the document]
haltman-at: Hm, OK, yeah, when you describe it that way, I think I see better what you're doing here... yeah, OK, I agree that saving it, in both cases, makes sense, and we should ignore what we discussed on Slack earlier.

But, I think it should just be `l` (as well as `s`, see other comment about `s`) that uses the custom numbers... I think `p` and `;` probably shouldn't, because they're more about low-level display than about source display, and I wouldn't want the source display taking up too much of the space there.

But, I guess that's up to you; test it out and see how it looks.  My thinking though is it would be too much clutter outside of `l` and `s`.

packages/core/lib/debug/interpreter.js
haltman-at: OK I guess you've explained in the other comment why you changed it, so, I don't think we need to worry about this.

packages/core/lib/debug/interpreter.js
```diff
@@ -603,7 +600,9 @@ class DebugInterpreter {
         await this.setOrClearBreakpoint(splitArgs, false);
         break;
       case "p":
-        //first: process which locations we should print out
+        // determine the numbers of instructions to be printed
+        this.parsePrintoutLines(splitArgs, this.printer.instructionLines);
```
haltman-at, line 604: This doesn't make sense.  I suggested renaming the function from `updatePrintoutLines` because it was a pure function and didn't update anything.  You changed the name to `parsePrintoutLines`... but you *also* changed it to be a function that *does*, in fact, perform the update!  Those two changes don't go together; if you were going to do the latter, you should have skipped the former.

Personally, I would suggest keeping the new name and changing this back to a pure function; it's generally better to use pure functions when reasonable.  You're more or less using `this.printer.instructionLines` as an output parameter here; those are confusing and I would say should be avoided.

packages/core/lib/debug/interpreter.js
```diff
@@ -761,6 +765,21 @@ class DebugInterpreter {
     }
     return tempPrintouts;
   }
+
+  // parse the numbers of lines options -<num>|+<num> from user args
+  parsePrintoutLines(userArgs, currentLines) {
+    for (const argument of userArgs) {
+      if (argument.length < 2) continue;
```
haltman-at, line 772: It's a little unclear what this line is doing.  I mean, I think I understand -- it's there so that a bare `+` or `-` isn't interpreted as `+0` or `-0` -- but I only picked up on that because I've been dealing a lot with details of number parsing lately.  I think you should add a comment here to clarify that; other readers won't get that.  Or, if I've misunderstood what it's doing, then it *definitely* needs a comment! :P

packages/core/lib/debug/printer.js
```diff
@@ -62,6 +62,7 @@ class DebugPrinter {
     //   cal: Calldata
     //   mem: Memory
     //   sta: Stack
+    // Note that this is a public variable and can be modified from outside.
```
haltman-at, line 65: Thanks for adding this!

packages/core/lib/debug/printer.js
```diff
@@ -72,6 +72,14 @@ class DebugPrinter {
     //   loc: Local variables
     this.sectionPrintouts = new Set(["bui", "glo", "con", "loc"]);
     this.sections = ["bui", "glo", "con", "loc"]; //should remain constant
+
+    // numbers of instructions before and after the current instruction to be printed
+    // used by command (p)
```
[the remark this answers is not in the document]
haltman-at: Are you updating this or are you leaving this alone?

packages/debug-utils/index.js
```diff
@@ -39,8 +39,8 @@ const commandReference = {
   "u": "step out",
   "n": "step next",
   ";": "step instruction (include number to step multiple)",
-  "p": "print instruction & state (`p [mem|cal|sto]*`; see docs for more)",
-  "l": "print additional source context",
+  "p": "print instruction & state (`p [mem|cal|sto|line-number]*`; see docs for more)",
+  "l": "print additional source context (`l [line-number]`; see docs for more)",
```
haltman-at, line 43: Thanks for updating this!  I don't think this quite conveys what we want, though.  Firstly, this doesn't include the necessary `+/-`.  Secondly, "line number" means the ordinal number of a given line of code, not a cardinal count of lines of count.  How about `l [+<lines-ahead>] [-<lines-back>]`?

packages/debug-utils/index.js
```diff
@@ -39,8 +39,8 @@ const commandReference = {
   "u": "step out",
   "n": "step next",
   ";": "step instruction (include number to step multiple)",
-  "p": "print instruction & state (`p [mem|cal|sto]*`; see docs for more)",
-  "l": "print additional source context",
+  "p": "print instruction & state (`p [mem|cal|sto|line-number]*`; see docs for more)",
```
haltman-at, line 42: Again, I don't think this conveys what we want.  Possibly we can just leave this out of the quick help -- I feel like this is a somewhat niche use case.  Is the quick help even long enough to accommodate this?  Like I think this sould be something more like `p [mem|cal|sto]* [+<instructions-ahead>] [-<instructions-back>]`, but that's getting fairly long... we could abbreviate the word "instructions" to "instrs" or something, not sure offhand how much that helps...

packages/core/lib/debug/interpreter.js
haltman-at: Hm, this seems to have been renamed in an inconsistent fashion.  Well -- I have a longer comment about that once the rest of the review is done, but noting it here also for now.

packages/core/lib/debug/interpreter.js
```diff
@@ -603,7 +603,13 @@ class DebugInterpreter {
         await this.setOrClearBreakpoint(splitArgs, false);
         break;
       case "p":
-        //first: process which locations we should print out
+        // update the numbers of instructions to be printed
```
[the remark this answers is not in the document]
haltman-at: Hm, as with the function name, this seems to have been changed in an inconsistent manner.

packages/core/lib/debug/interpreter.js
```diff
@@ -628,7 +634,13 @@ class DebugInterpreter {
       case "l":
         if (this.session.view(session.status.loaded)) {
           this.printer.printFile();
-          this.printer.printState(LINES_BEFORE_LONG, LINES_AFTER_LONG);
+          // update the numbers of lines to be printed
```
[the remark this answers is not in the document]
haltman-at: Hm, as with the function name, this seems to have been changed in an inconsistent manner.

packages/core/lib/debug/interpreter.js
```diff
@@ -761,6 +765,21 @@ class DebugInterpreter {
...
+
+  // parse the numbers of lines options -<num>|+<num> from user args
+  parsePrintoutLines(userArgs, currentLines) {
+    for (const argument of userArgs) {
+      if (argument.length < 2) continue;
+      const newLines = Number(argument.slice(1));
+      // ignore the arguments that are not of the correct form, number
+      if (isNaN(newLines)) continue;
+      if (argument[0] === "-") {
+        currentLines.beforeLines = newLines;
```
haltman-at, line 777: As mentioned elsewhere, I think this function should be pure, and should return a value, like it was before, rather than altering its input.

packages/core/lib/debug/printer.js
```diff
@@ -70,8 +71,19 @@ class DebugPrinter {
...
+    // Note that this is a public variable and can be modified from outside.
+    this.instructionLines = { beforeLines: 3, afterLines: 3 };
+
+    // numbers of lines before and after the current line to be printed
+    // used by commands (l) and (s)
+    // Note that this is a public variable and can be modified from outside.
+    this.sourceLines = { beforeLines: 2, afterLines: 0 };
```
haltman-at, line 86: Shouldn't these be 5 and 3?

haltman-at (commented):
OK, this now seems good functionality-wise!  Just a few style and doc matters to clean up.

packages/core/lib/debug/interpreter.js
dongmingh: I will ad a comment here for clarity, thx.

packages/core/lib/debug/interpreter.js
dongmingh: I will keep the new name and change this function back to pure.

packages/core/lib/debug/interpreter.js
haltman-at: Thank you!

packages/debug-utils/index.js
haltman-at: Oops, previous comment got a bit garbled in places, fixed it now.

packages/debug-utils/index.js
dongmingh: The help looks better with `instrs`:
```
(g) turn on generated sources, (G) turn off generated sources except via `;`
(p) print instruction & state (`p [mem|cal|sto|]* [+<instrs-ahead>] [-<instrs-back>]`; see docs for more)
(l) print additional source context (`l [+<lines-ahead>] [-<lines-back>]`; see docs for more)
```

haltman-at: Nice!

packages/core/lib/debug/printer.js
```diff
@@ -70,8 +71,17 @@ class DebugPrinter {
...
     //   loc: Local variables
+    // Note that this is a public variable and can be modified from outside.
     this.sectionPrintouts = new Set(["bui", "glo", "con", "loc"]);
     this.sections = ["bui", "glo", "con", "loc"]; //should remain constant
+
+    // numbers of instructions before and after the current instruction to be printed
+    // Note that this is a public variable and can be modified from outside.
+    this.instructionLines = { beforeLines: 3, afterLines: 3 };
+
+    // numbers of lines before and after the current line to be printed
```
haltman-at, line 82: Hm, you previously had a comment here clarifying that this was just for `l` and `s`... not sure why you deleted that, that seems helpful.

packages/debug-utils/index.js
haltman-at: OK trying it out these are still too long I'm afraid... want these to fit in an 80 character width I think.

For `l` this is easy to fix -- we don't need the "see docs for more" note because that's the whole syntax!  That note only exists on `p` because we couldn't fit the whole syntax on one line.

For `p`... I think we're just going to have to drop the bit about +/-, and just leave that in the online documentation; I don't think there's any way to make that fit.

haltman-at (commented):
OK, just two final notes!  One on a comment you deleted, one on the help printout.

(I forgot to do this as a proper review, sorry.  Anyway you can see my comments above.)

packages/debug-utils/index.js
haltman-at: Come to think of it, the "see docs for more" note on `v` doesn't need to be there either, does it?  Dunno if getting rid of that needs to go in this PR, but it should be removed at some point.  (I can pick that up later if you don't have time to get it.)

packages/core/lib/debug/printer.js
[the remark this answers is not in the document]
haltman-at: So, I noticed you put this back in, but you put it back in as just `p`...?  I'm confused here, you hit thumbs-up on my comment about `;`, which I took as stating that you intended to mention `;` as well, but you haven't done that?

haltman-at (approved):
OK, I think this is good enough to merge.  I think you should go back and fix that comment about `p` and `;`, but I'm not going to hold it up over that!

packages/debug-utils/index.js
haltman-at: Thanks!

PR #4463, opened 2021-11-23
Account for the oldest kind of provider by favoring `provider.sendAsync()` over `provider.send()`

gnidan:
... to account for a problem with MetaMask's injected provider, which assumes the very old semantics that `provider.send()` returns a synchronous response (not even via a callback!)

### DIFF
--- a/packages/decoder/lib/ProviderAdapter.ts
+++ b/packages/decoder/lib/ProviderAdapter.ts
@@ -104,11 +104,20 @@ export class ProviderAdapter {
     if (isEip1193Provider(this.provider)) {
       return await this.provider.request({ method, params });
     } else {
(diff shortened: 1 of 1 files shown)


haltman-at (approved):
I have to wonder if there's a better way to do this -- like, you'd think there'd be a third type of provider rather than this hack -- but I'm just going to assume that there is a reason you didn't do it that way and hit approve. :P

gnidan:
Yeah, would be better to improve the types, but that requires care and thought... and who knows if this is the last such incompatibility or if there will be more to follow!

Anyway, we've talked a bunch about moving this adapter to a dedicated place. Seems like we can clean this up then.

PR #4465, opened 2021-11-23
Bug fix: Remove filename with colon from repo to prevent Windows problems

haltman-at:
It seems that having a filename with a colon in it (that I added in #4363) was making it impossible to clone the repo on Windows.  This file was a Solidity source used in a test.  Fortunately, that test used `Compile.sources()`, rather than `Compile.sourcesWithDependencies()` or similar, and thus there was no actual need to have the Solidity source live in a real file with a real file name; I replaced with a bunch of constants in the test file instead.

Note that I removed the file with new commits; I did *not* purge it from the git history or anything, so you might still run into problems on Windows trying to check out versions inbetween when #4363 was merged and when this gets merged.  But at least it should be possible to clone the repo again.  (I hope; if not, then, uh, someone will need to go rewrite history...)


packages/compile-solidity/test/test_oldversions.js
```diff
@@ -21,16 +19,41 @@ const compileOptions = {
...
+  }
+}
+`;
+
+//I've put a colon in the filename to be extra-sure
+//that this is indeed testing what it's supposed to
+//(that colons in file paths don't screw things up;
+//"project:/" should suffice to trigger the issue but
+//I want to be extra sure)
+const sourcePath = `${workingDirectory}/contracts/FilenameWith:Colon.sol`;
```
cds-amal, line 51: I wonder if we should use `path.join`, but it shouldn't block this. I mean, that's what the `windows/runner` branch is for.

cds-amal (approved):
Looks good. Thanks @haltman-at !

packages/compile-solidity/test/test_oldversions.js
haltman-at: Oh, yeah, probably.  Oh well, like you say, that can be handled later...

farisei77 (approved):

PR #4467, opened 2021-11-24
Typo fix

JiechengZhao:
There is a typo in the previous README as below.
| Parameter | Type | Default | Required | Description |
| ------ | ---- | ------- | ----------- | ----------- |
| `chainId` | `number/|string` | `undefined` | [ ] | Specify to enable signed transactions that are EIP-155 compliant for major chains. |


gnidan (approved):
Looks good! Thanks for catching this and fixing!

PR #4483, opened 2021-12-02
Update Ganache to v7.x

eggplantzzz:
This PR salvages work done in https://github.com/trufflesuite/truffle/pull/3569 which updates Truffle to use Ganache v7.x. Thanks to @davidmurdoch! 👍


haltman-at:
So are we closing #3569, then?

eggplantzzz:
Yes I believe we can, @davidmurdoch are you okay with this?

THRIVEbluegooglecheck:
0=0xDb06322ceA4D7e56f538eB54E907ab469d7878E2

THRIVEbluegooglecheck:
=0

packages/debugger/test/data/returnvalue.js
```diff
@@ -279,8 +287,9 @@ describe("Return value decoding", function () {
     try {
       await instance.fail(); //web3 throws on failure
     } catch (error) {
-      txHash = error.hashes[0]; //it's the only hash involved
+      txHash = error.receipt.transactionHash;
     }
+    if (!txHash) assert.fail("should have errored and set txHash");
```
haltman-at, line 292: Given that we're using Chai, I think it would make more sense to just write this line as
```assert.isDefined(txHash, "should have errored and set txHash")```
(That also applies throughout this file, I won't tag each individual one.)

File: packages/debugger/test/solidity.js [hunk 2]
```diff
@@ -327,8 +335,9 @@ describe("Solidity Debugging", function () {
       try {
         await instance.run(); //this will throw because of the revert
       } catch (error) {
-        txHash = error.hashes[0]; //it's the only hash involved
+        txHash = error.receipt.transactionHash;
       }
+      if (!txHash) assert.fail("should have errored and set txHash");
```
haltman-at, line 340: Another assertion to rewrite.

packages/debugger/test/stacktrace.js
```diff
@@ -126,8 +134,9 @@ describe("Stack tracing", function () {
     try {
       await instance.run(0); //this will throw because of the revert
     } catch (error) {
-      txHash = error.hashes[0]; //it's the only hash involved
+      txHash = error.receipt.transactionHash;
     }
+    if (!txHash) assert.fail("should have thrown and set txHash");
```
haltman-at, line 139: And another one of these; there's more in this file of course but I won't mark them each individually.

packages/debugger/test/txlog.js
```diff
@@ -376,8 +384,9 @@ describe("Transaction log (visualizer)", function () {
     try {
       await instance.testRevert(); //this will throw because of the revert
     } catch (error) {
-      txHash = error.hashes[0]; //it's the only hash involved
+      txHash = error.receipt.transactionHash;
     }
+    if (!txHash) assert.fail("should have errored and set txHash");
```
haltman-at, line 389: And another one of these.

packages/decoder/test/current/test/downgrade-test.js
```diff
@@ -579,12 +579,15 @@ async function runErrorTestBody(mangledCompilations) {
...
   //we need the raw return data, and contract.call() does not exist yet,
   //so we're going to have to use web3.eth.call()
-
-  let data = await web3.eth.call({
-    to: deployedContract.address,
-    data: selector
-  });
-
+  let data;
+  try {
```
haltman-at, line 583: Huh... I'm confused why this change is necessary even though VM errors are off.  What's going on here...? (@davidmurdoch?)

packages/decoder/test/current/test/wire-test.js
```diff
@@ -1072,26 +1070,30 @@ describe("Over-the-wire decoding", function () {
...
+
       //we need the raw return data, and contract.call() does not exist yet,
       //so we're going to have to use web3.eth.call()
-  
-      let data = await web3.eth.call({
-        to: deployedContract.address,
-        data: selector
-      });
+      let data;
+      try {
```
haltman-at, line 1086: Same question here as previous comment, naturally.  (And also later in this file but I won't mark them all.)

packages/interface-adapter/test/fabric-evm-getId.test.ts
```diff
@@ -29,18 +34,24 @@ async function prepareGanache(
...
 
-describe("fabric-evm getId Overload", function() {
-  it("returns networkID as valid string instead of number w/ fabric-evm=true", async function() {
-    const preparedGanache = await prepareGanache(true);
-    const networkID = await preparedGanache.interfaceAdapter.getNetworkId();
-    assert(typeof networkID === "string");
-    await preparedGanache.server.close();
+describe("fabric-evm getId Overload", function () {
+  it("returns networkID as valid string instead of number w/ fabric-evm=true", async function () {
+    const preparedGanache = (await prepareGanache(true)) as any;
```
haltman-at, line 39: Why is `as any` necessary here?  (And also below, of course.)  Especially because I notice you didn't use it in the other file...

packages/interface-adapter/test/quorum-getBlock.test.ts
```diff
@@ -33,23 +37,29 @@ async function prepareGanache(
...
   });
 
-  it("recovers block timestamp as number w/ quorum=false", async function() {
-    const preparedGanache = await prepareGanache(false);
-    const block = await preparedGanache.interfaceAdapter.getBlock(0);
-    const expectedBlockTime = new BN(genesisBlockTime.getTime()).divn(1000);
-    assert.strictEqual(block.timestamp, expectedBlockTime.toNumber());
-    await preparedGanache.server.close();
+  it("recovers block timestamp as number w/ quorum=false", async function () {
+    const preparedGanache = (await prepareGanache(false)) as any;
```
haltman-at, line 56: Another unnecessary-seeming `as any`.

packages/truffle/test/scenarios/migrations/dryrun.js
```diff
@@ -5,35 +5,27 @@ const assert = require("assert");
...
+  before(async function () {
     this.timeout(10000);
     config = await sandbox.create(project);
     config.network = "development";
     config.logger = logger;
     config.mocha = {
       reporter: new Reporter(logger)
     };
-
-    const provider = new Web3.providers.HttpProvider("http://localhost:8545", {
```
haltman-at, line 28: So was all this stuff just not used, I'm guessing?

packages/truffle/test/scenarios/migrations/errors.js
```diff
@@ -25,12 +23,6 @@ describe("migration errors", function () {
     config.mocha = {
       reporter: new Reporter(logger)
     };
-
-    const provider = new Web3.providers.HttpProvider("http://localhost:8545", {
```
haltman-at, line 29: Same question here, this is all unused?

packages/contract/lib/manual-send.js
```diff
@@ -44,7 +44,11 @@ async function sendTransactionManual(web3, params, promiEvent) {
 
 function handleError(error) {
   debug("error: %O", error);
-  if (error.data && Object.keys(error.data).length === 3) {
+  if (error.data && error.data.hash) {
```
haltman-at, line 47: Thanks for getting this!

packages/contract/lib/reason.js
```diff
@@ -13,7 +13,13 @@ const reason = {
   _extract: function(res, web3, _interfaceAdapter) {
     //I'm not sure why interfaceAdapter is here if it's not used,
     //so I just put an underscore in front of its name for now...
-    if (!res || (!res.error && !res.result)) return;
+    if (!res) return;
+    if (!res.error && (
```
haltman-at, line 17: This logic doesn't look correct to me; and if it somehow is correct, then I'm pretty sure it's been written in a way that's overcomplicated.

This conditional is trying to detect if there's *no* error, right?  So it ought to be checking that there is *both* no 2.x error *and* no 7.x error.  But instead it uses an `||`.  Moreover, it first checks `!res.error`, and then checks `res.error`... that can't be right.  (And if it *is* right, it probably ought to be able to use some Boolean algebra to simplify that.)

packages/contract/lib/reason.js
```diff
@@ -23,14 +29,28 @@ const reason = {
       res && typeof res === "object" && typeof res.result === "string";
 
     if (isObject) {
+      // TODO: ganache 2.0 returns the reason string already when
```
haltman-at, line 32: You've left a `TODO` comment in your code, @gnidan isn't going to like that. :P  So, uh, you should probably figure out whether you want to include that in this PR, or just convert this to a non-TODO comment about how this would be a good thing to do in the future.

packages/contract/lib/reason.js
```diff
@@ -23,14 +29,28 @@ const reason = {
       res && typeof res === "object" && typeof res.result === "string";
 
     if (isObject) {
```
haltman-at, line 31: Does this condition -- which is perhaps named a little misleadingly -- still apply in the Ganache 7 case?

packages/contract/lib/reason.js
```diff
@@ -23,14 +29,28 @@ const reason = {
...
+      // vmErrorsOnRPCResponse === true. Why not use it?
       const data = res.error.data;
-      const hash = Object.keys(data)[0];
+      let resData;
+      if ("result" in data) {
+        // there is a single result (only in ganache 3.0+)
+        resData = data.result;
+      } else {
+        // handle `evm_mine`, `miner_start`, batch payloads, and ganache 2.0
+        // TODO: there are 1 or more failed transactions, handle multiple?
```
haltman-at, line 41: Another `TODO` comment. :P  Either handle multiple or just make a note to handle them later!

haltman-at (commented):
OK, most of this seems good, but I have a bunch of questions on a few parts.

packages/debugger/test/data/global.js
```diff
@@ -160,7 +160,13 @@ describe("Globally-available variables", function () {
     provider = Ganache.provider({
       seed: "debugger",
       gasLimit: 7000000,
-      _chainId: 1337 //temporary until Ganache v3!
+      hardfork: "istanbul",
```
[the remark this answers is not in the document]
haltman-at: (Btw, my preferred approach for this would be to actually update this test to remove the hack, but uh I think it's fine to leave that in for now and I can remove it later. :P )

packages/contract-tests/test/deploy.js
```diff
@@ -1,11 +1,17 @@
...
-  var Example;
-  var web3;
-  var providerOptions = { vmErrorsOnRPCResponse: false };
-
+  let Example;
+  let web3;
+  // ganache v2 had a default blockGasLimit of 0x6691b7, which these tests rely on
+  const providerOptions = {
+    gasLimit: 0x6691b7,
+    hardfork: "istanbul",
```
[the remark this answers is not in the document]
eggplantzzz: @davidmurdoch do you know why Istanbul? I think this might have been something you added?

packages/contract-tests/test/errors.js
```diff
@@ -1,9 +1,13 @@
...
-var util = require("./util");
+const assert = require("chai").assert;
+const util = require("./util");
 
 describe("Client appends errors (vmErrorsOnRPCResponse)", function () {
-  var Example;
-  var providerOptions = { vmErrorsOnRPCResponse: true }; // <--- TRUE
+  let Example;
+  const providerOptions = {
+    hardfork: "istanbul",
```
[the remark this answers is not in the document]
eggplantzzz: @davidmurdoch

packages/contract-tests/test/errors.js
```diff
@@ -83,34 +87,26 @@ describe("Client appends errors (vmErrorsOnRPCResponse)", function () {
...
         assert(
-          e.stack.includes("Error: base fee exceeds gas limit"),
+          e.stack.includes("Error: intrinsic gas too low"),
           "Should keep hijacked error description"
         );
         assert(
           e.stack.includes("/test/errors.js:"),
           "Should include original stack details"
         );
-        assert(
```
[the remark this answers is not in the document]
eggplantzzz: To be honest I don't really know why these were here in the first place but after discussion it seemed like something we didn't care about (maybe we did at one point). I probably should remove the other ones as well, these were just the ones that caused tests to fail.

packages/core/lib/commands/develop/run.js
```diff
@@ -51,9 +52,9 @@ module.exports = async options => {
     mnemonic,
     gasLimit: customConfig.gas || 0x6691b7,
     gasPrice: customConfig.gasPrice || 0x77359400,
-    noVMErrorsOnRPCResponse: true,
     time: config.genesis_time,
-    _chainId: 1337 //temporary until Ganache v3!
+    vmErrorsOnRPCResponse: customConfig.vmErrorsOnRPCResponse || false,
```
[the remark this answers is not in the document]
eggplantzzz: Ah you are probably right that this should be left off until later.

packages/debugger/test/context-vyper.js
```diff
@@ -42,19 +42,28 @@ const sources = {
...
+      miner: {
+        instamine: "strict"
+      }
+    });
   });
 
   before("Prepare contracts and artifacts", async function () {
     this.timeout(30000);
-
+    console.log("in prepare ----");
```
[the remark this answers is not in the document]
eggplantzzz: Indeed :)

packages/contract-tests/test/errors.js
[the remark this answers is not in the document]
eggplantzzz: I removed a couple of assertions involving directly checking `e.hijackedStack`. I left in the tests whose descriptions reference hijacked stack because some of them appear like they could still be relevant.

packages/contract-tests/test/errors.js
```diff
@@ -236,14 +235,16 @@ describe("Client appends errors (vmErrorsOnRPCResponse)", function () {
       const example = await Example.new(1);
       try {
         const accounts = await Example.web3.eth.getAccounts();
-        await example.runsOutOfGas({ from: accounts[0], gas: "0x6691b7" });
+        // specifying a gasLimit as ganache @2.7.0 no longer limits gas
```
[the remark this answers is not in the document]
eggplantzzz: It seems like there used to be no gas limit and now in order to ensure it runs out of gas in this way we specifcy it manually?

packages/debugger/test/data/global.js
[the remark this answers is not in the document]
eggplantzzz: David suggested that it is likely here because these tests might rely on `gasPrice = 0`.

packages/debugger/test/data/returnvalue.js
eggplantzzz: Sure!

packages/contract-tests/test/errors.js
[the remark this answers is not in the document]
davidmurdoch: istanbul allows running with a `0` gasPrice

packages/contract-tests/test/deploy.js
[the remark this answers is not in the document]
davidmurdoch: istanbul allows running with a `0` gasPrice

packages/truffle/test/scenarios/migrations/dryrun.js
eggplantzzz: yep there was a bunch of extra stuff

packages/truffle/test/scenarios/migrations/errors.js
eggplantzzz: yup

packages/debugger/test/endstate.js
```diff
@@ -33,42 +33,51 @@ uint x;
...
     //does not presently work)
     let txHash;
     try {
-      await instance.run(); //this will throw because of the revert
+      await instance.run();
     } catch (error) {
-      txHash = error.hashes[0]; //it's the only hash involved
+      txHash = error.receipt.transactionHash;
     }
+    if (!txHash) assert.fail("it should have set the txHash");
```
lsqproduction, line 78: here's another reformat to `assert.isDefined(txHash, "should have errored and set txHash");` ?

packages/deployer/test/ens.js
```diff
@@ -20,7 +20,7 @@ let ganacheOptions,
 describe("ENS class", () => {
   before(() => {
     ganacheOptions = {
-      // note that when vmErrorsOnRPCResponse is true, eager instamine must be enabled (default)
+      // note that when vmErrorsOnRPCResponse is true, strict instamine must be enabled (default)
```
davidmurdoch, line 23: `vmErrorsOnRPCResponse: true` only works with `instamine: "eager"` (which is the default)

packages/contract-tests/test/deploy.js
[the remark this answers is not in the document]
haltman-at: Huh, no idea why that's something we want here, but I'll take your word for it.  Is it worth adding a comment here saying that in the future this should be updated not to use zero `gasPrice` so this can be moved off Istanbul in the future?

packages/contract-tests/test/errors.js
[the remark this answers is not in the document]
haltman-at: OK but then you mean Ganache **7** no longer limits gas, right?

packages/debugger/test/data/global.js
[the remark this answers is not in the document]
haltman-at: Oh, these tests don't rely on gas price being zero; they do, however, currently rely on base fee being zero (until the hack is removed and the tests are properly updated :) ).  So, like, if you're going to pin a specific hardfork, you could at least use Berlin rather than Instabul. :)

packages/contract-tests/test/errors.js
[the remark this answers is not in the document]
gnidan: (Should we also put a comment here?)

packages/contract-tests/test/errors.js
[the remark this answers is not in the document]
gnidan: (The other ones have also been removed)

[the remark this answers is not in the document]
haltman-at: OK, seems these have been removed now.

packages/contract-tests/test/errors.js
```diff
@@ -236,14 +235,16 @@ describe("Client appends errors (vmErrorsOnRPCResponse)", function () {
...
+        // estimations to the default block gas limit.
```
gnidan, line 239: Maybe either remove this comment or reword it to be independent of history (we can say _why_ we specify `gasLimit`... seems better than commenting on the past)

packages/debugger/test/data/global.js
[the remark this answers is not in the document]
gnidan: Let's make sure we don't forget the follow-on here... maybe open an issue?

packages/contract-tests/test/deploy.js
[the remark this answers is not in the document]
gnidan: Putting this catch-all issue here https://github.com/trufflesuite/truffle/issues/4633

packages/contract-tests/test/errors.js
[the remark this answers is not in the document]
gnidan: Putting this catch-all issue here https://github.com/trufflesuite/truffle/issues/4633

packages/debugger/test/data/global.js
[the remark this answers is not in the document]
gnidan: Issue made here https://github.com/trufflesuite/truffle/issues/4633

packages/decoder/test/current/test/downgrade-test.js
gnidan: See papertrail: https://github.com/trufflesuite/ganache/issues/1496

gnidan: We probably want this papertrail as a comment, actually! Leaving this unresolved until we get a comment :)

gnidan: @cds-amal please resolve when you feel this is sufficiently documented

packages/decoder/test/current/test/wire-test.js
gnidan: @cds-amal same here :)

packages/contract/lib/reason.js
gnidan: Let's undo this change and instead borrow @haltman-at's change in https://github.com/trufflesuite/truffle/pull/4506.

packages/contract/lib/reason.js
```diff
@@ -13,7 +13,13 @@ const reason = {
...
     //I'm not sure why interfaceAdapter is here if it's not used,
     //so I just put an underscore in front of its name for now...
-    if (!res || (!res.error && !res.result)) return;
+    if (!res) return;
+    if (!res.error && (
+        !res.result /* ganache 2.0 */
+        || (res.error && res.error.data && !res.error.data.result /* ganache 7.x */)
+      )
+    ) return;
+
```
gnidan, line 22:
```suggestion
    if (!res || (!res.error && !res.result)) return;
```

packages/contract/lib/reason.js
```diff
@@ -23,14 +23,28 @@ const reason = {
       res && typeof res === "object" && typeof res.result === "string";
 
     if (isObject) {
+      // TODO: ganache 2.0 returns the reason string already when
+      // vmErrorsOnRPCResponse === true. Why not use it?
```
gnidan, line 27:
```suggestion
      // NOTE that Ganache >=2 returns the reason string when
      // vmErrorsOnRPCResponse === true, which this code could
      // be updated to respect (instead of computing here)
```

packages/contract/lib/reason.js
```diff
@@ -23,14 +23,28 @@ const reason = {
...
       const data = res.error.data;
-      const hash = Object.keys(data)[0];
+      let resData;
+      if ("result" in data) {
+        // there is a single result (only in ganache 3.0+)
+        resData = data.result;
+      } else {
+        // handle `evm_mine`, `miner_start`, batch payloads, and ganache 2.0
+        // TODO: there are 1 or more failed transactions, handle multiple?
```
gnidan, line 35:
```suggestion
        // handle `evm_mine`, `miner_start`, batch payloads, and ganache 2.0
        // NOTE this only works for a single failed transaction at a time.
```

packages/debugger/test/endstate.js
gnidan:
```suggestion
    assert.isDefined(txHash, "should have errored and set txHash");
```

packages/contract/lib/reason.js
gnidan: Or maybe we don't need to borrow #4506 (assuming tests pass :)

packages/deployer/test/ens.js
```diff
@@ -20,25 +20,26 @@ let ganacheOptions,
 describe("ENS class", () => {
   before(() => {
     ganacheOptions = {
+      // note that when vmErrorsOnRPCResponse is true, strict instamine must be enabled (default)
```
gnidan, line 23:
```suggestion
      // note that when vmErrorsOnRPCResponse is true, `"eager"` instamine must be enabled (default)
```

packages/decoder/test/current/test/downgrade-test.js
```diff
@@ -613,12 +613,15 @@ async function runErrorTestBody(mangledCompilations) {
...
```
gnidan, line 617:
```suggestion
  // NOTE we wrap this in `try`/`catch` to accommodate new eth_call behavior
  // see https://github.com/trufflesuite/ganache/issues/1496
  try {
```

packages/decoder/test/current/test/wire-test.js
gnidan, line 1086:
```suggestion
      // NOTE we wrap this in `try`/`catch` to accommodate new eth_call behavior
      // see https://github.com/trufflesuite/ganache/issues/1496
      try {
```

gnidan (approved):

packages/debugger/test/data/global.js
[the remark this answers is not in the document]
haltman-at: OK I went and updated this particular one to London; the rest are still on Istanbul though.

PR #4484, opened 2021-12-03
Allow recognition of deployed Vyper bytecode with immutables

haltman-at:
This PR allows the debugger and decoder to recognize bytecode from a deployed Vyper contract that uses immutables.

Now, unlike Solidity, the Vyper compiler doesn't given us any `immutableReferences` output.  But fortunately, unlike Solidity, Vyper puts its immutables at the end of the bytecode -- like constructor arguments -- rather than inline.  This means that we can simply allow "constructor arguments" at the end of deployed bytecode, so long as the context we're trying to match is a Vyper one.

(I thought about not restricting it to Vyper, but decided to leave that restriction in for now.)

So, with this change, it is now possible to debug (i.e. step through) a transaction made against a Vyper contract that uses immutables.  (It also affects things like `decodeTransaction` and `decodeEvent` in the decoder.)

I also added a test of this functionality.  Because this test requires having Vyper installed, I put it in a separate file from the other context tests, with separate setup, so that not having Vyper installed doesn't cause the others to fail.  (This is the first actual automated test to be added of Vyper in the debugger!)

I also updated relevant comments, and some minor style matters.

### DIFF
--- a/packages/codec/lib/contexts/utils.ts
+++ b/packages/codec/lib/contexts/utils.ts
@@ -42,13 +42,17 @@ export function findContext(
 }
 
 export function matchContext(context: Context, givenBinary: string): boolean {
-  let { binary, isConstructor } = context;
-  let lengthDifference = givenBinary.length - binary.length;
-  //first: if it's not a constructor, they'd better be equal in length.
-  //if it is a constructor, the given binary must be at least as long,
-  //and the difference must be a multiple of 64
+  const { binary, compiler, isConstructor } = context;
+  const lengthDifference = givenBinary.length - binary.length;
+  //first: if it's not a constructor, and it's not Vyper,
+  //they'd better be equal in length.
+  //if it is a constructor, or is Vyper,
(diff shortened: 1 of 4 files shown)


haltman-at:
Blargh, this is failing CI because CI isn't getting Vyper 0.3.1.  @gnidan, is it OK if I just add this test as skipped for now?

cds-amal:
oh wow, vyper is at v0.3.1 in the [cheeseshop](https://pypi.org/project/vyper/) and should be picked up by pip. I wonder what's happening here.

cds-amal:
aha! The vyper team dropped support for python3.6 at [vyper-0.3.0](https://github.com/vyperlang/vyper/releases/tag/v0.3.0)

haltman-at:
Oh! I didn't think of that at all!

cds-amal (approved):
Looks good @haltman-at ! Let's get it in :)

Issue #4156, opened 2021-07-01
Add support for access lists


haltman-at:
## Issue

Web3 1.4.0 is out, and add supports for access lists.  I've already put up #4155 to update web3 to 1.4.0, but once that's done, we then need to add support for access lists.

Note that #3974 & #4142 will also presumably need considerable updating to handle access lists.

## Environment

* Truffle version (`truffle version`): 5.3.13

haltman-at:
Hm, wondering if actually we want to wait on Ganache 7 for this.

kevinbluer:
Closing as this is addressed in https://github.com/trufflesuite/truffle/releases/tag/v5.4.24

PR #4491, opened 2021-12-03
Allow users to specify transaction type explicitly, and allow use of access lists, including ENS support

haltman-at:
Addresses #4156.

Both web3 and ethers allow setting the transaction type expicitly with the `type` option, but we weren't allowing this.  This PR whitelists the `type` option in Truffle Contract, and then in addition I added it to the list of options that users can explicitly set in their config, because it seems like something that users might want a default for.  Note that the default default is set to `undefined`, not `null`; web3 will error if you specify `type: null` (no Pokemon jokes please :P ).

Note that web3 doesn't *fully* respect the type you specify -- if you specify type 1, but don't specify an access list (and with Truffle currently, you still *can't* specify an access list!), it'll use type 0 instead -- but I don't think that's any big reason not to do this.

**Edit**: I've also now updated this PR to finally also add support for access lists into Truffle Contract.  We were holding off on this until Ganache 7 is out, but now that the beta is out, @gnidan said that's good enough. :)

**Edit again**: I also now added ENS support for those access lists. :)

~~I didn't include any tests in this PR, be aware.~~  No longer true!  I added a test of ENS support for access lists. :)

### DIFF
--- a/packages/contract-tests/test/lib/utils/ens.js
+++ b/packages/contract-tests/test/lib/utils/ens.js
@@ -25,7 +25,18 @@ const methodABI = {
   stateMutability: "pure",
   type: "function"
 };
-const args = ["my.ens.name", 555, { from: "the.other.name" }];
+const txParams = {
+  from: "the.other.name",
+  accessList: [{
+    address: "third.party",
+    storageKeys: []
+  }]
+};
+const args = [
+  "my.ens.name",
+  555,
+  txParams
+];
 const ensSettings = { enabled: true };
 const networkId = 1;
 
@@ -33,6 +44,7 @@ describe("convertENSNames", () => {
   beforeEach(() => {
     const expectedInput1 = "my.ens.name";
     const expectedInput2 = "the.other.name";
+    const expectedInput3 = "third.party";
     sinon.stub(ens, "getNewENSJS").returns({
       name: input => {
         if (input === expectedInput1) {
@@ -40,6 +52,9 @@ describe("convertENSNames", () => {
         }
         if (input === expectedInput2) {
           return { getAddress: () => Promise.resolve("0x987") };
+        }
+        if (input === expectedInput3) {
+          return { getAddress: () => Promise.resolve("0xabc") };
         }
         return Promise.reject("The input was not what was expected");
       }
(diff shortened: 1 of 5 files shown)


packages/contract-tests/test/lib/utils/ens.js
```diff
@@ -25,14 +25,26 @@ const methodABI = {
   stateMutability: "pure",
   type: "function"
 };
-const args = ["my.ens.name", 555, { from: "the.other.name" }];
+const txParams = {
+  from: "the.other.name",
+  accessList: [{
```
eggplantzzz, line 30: Is this how you specify this? Do you have links to any docs? I looked at the web3 docs and they kind of ambiguously say to spcify "a list of addresses and storage keys that the transaction plans to access".

eggplantzzz (approved):
Seems good to me.

packages/contract-tests/test/lib/utils/ens.js
haltman-at: Yeah, unfortunately, it's not really documented anywhere; I basically just had to consult the code of `@ethereumjs/tx` to see how it's supposed to work.  There are two allowed forms:

1. An array of `{ address, storageKeys }` objects.  In this case, `address` is a string (being, y'know, an address)` and `storageKeys` is an array of strings (each being a hex string describing 32 bytes).
2. An array of `[address, storageKeys]` tuples.  In this case, `address` is a Uint8Array, and `storageKeys` is an array of Uint8Arrays.

As the comment I wrote mentions, I decided that there's not much reason to handle the latter. :P

haltman-at: Here's where the type is actually defined: https://github.com/ethereumjs/ethereumjs-monorepo/blob/c74035e12765ab4bce8cb2bead634e2887af0d7c/packages/tx/src/types.ts#L80

haltman-at: OK, it turns out I was wrong about the second form being allowed.  I'll go delete that comment, I guess!

eggplantzzz: Thanks @haltman-at!

Issue #4477, opened 2021-11-30
truffle migrate failed in box v5.1-example


dongmingh:
## Issue

The `truffle migrate` failed in box [v5.1-example](https://github.com/truffle-box/v5.1-example-box)

## Steps to Reproduce

Create a new directory
cd the new directory
truffle unbox v5.1-example
cd v5.1-example
uncomment the development network in truffle-config.js
```
     development: {
      host: "127.0.0.1",     // Localhost (default: none)
      port: 8545,            // Standard Ethereum port (default: none)
      network_id: "*",       // Any network (default: none)
     },
```
truffle compile
open another terminal and cd to v5.1-example
ganache-cli
cd back to the first terminal
truffle migrate
Then see the error messages given in section Actual Results.

## Expected Behavior

The `truffle migrate` is executed successfully. 

## Actual Results

The error messages are:
```
╭─dongminghwang@dongmings-mbp  ~/sandbox/v5.1-example ‹node-16.10.0›  ‹› 
╰─$ truffle migrate

Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.



Starting migrations...
======================
> Network name:    'development'
> Network id:      1638202305352
> Block gas limit: 6721975 (0x6691b7)


1_initial_migration.js
======================

   Deploying 'Migrations'
   ----------------------

Error:  *** Deployment Failed ***

"Migrations" -- Cannot destructure property 'registryAddress' of 'ens' as it is undefined..

    at /Users/dongminghwang/truffle/packages/deployer/src/deployment.js:365:17
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Migration._deploy (/Users/dongminghwang/truffle/packages/migrate/Migration.js:70:7)
    at Migration._load (/Users/dongminghwang/truffle/packages/migrate/Migration.js:56:5)
    at Migration.run (/Users/dongminghwang/truffle/packages/migrate/Migration.js:217:5)
    at Object.runMigrations (/Users/dongminghwang/truffle/packages/migrate/index.js:150:9)
    at Object.runFrom (/Users/dongminghwang/truffle/packages/migrate/index.js:110:12)
    at Object.run (/Users/dongminghwang/truffle/packages/migrate/index.js:87:7)
    at runMigrations (/Users/dongminghwang/truffle/packages/core/lib/commands/migrate.js:258:18)
    at Object.run (/Users/dongminghwang/truffle/packages/core/lib/commands/migrate.js:223:7)
[...]

PR #4496, opened 2021-12-06
Fix incorrect key name in migrate prepareCall

dongmingh:
This PR fixed an error in the key name ens when calling convertENSNames in the migrate prepareCall. The error causes truffle-box v5.1-example-box fails to migrate, see issue #4477.


eggplantzzz (approved):
Thanks @dongmingh!

PR #4497, opened 2021-12-07
Internal improvement: Fix confusing comment on `toHexString` function

haltman-at:
Was working on codec when I noticed this comment made no sense.  It was clearly supposed to say "left" instead of "length".  So I fixed it, and also changed it to be an `@param` comment, FWIW.


cds-amal (approved):
Looks good.

PR #4503, opened 2021-12-08
@truffle/db Option to save .db directory to project directory

### LINKED ISSUE #4495: Users should be able to configure the default location for @truffle/db persistence
## Issue

Right now, @truffle/db by default stores everything in the global Truffle config directory (`~/.config/truffle` or such on Linux, other directories on other platforms). This can cause trouble because @truffle/db remains a bit experimental, and so users might not want to risk collision or breakages and may instead prefer to store their persistent records on a per-project basis.

This issue proposes that we add a user-level config flag to favor per-project persistence over the current default of global persistence.

tenthirtyone:
#### Summary 
Adds the `saveLocally` property to the `db` section of the `truffle-config.js`

```
db: {
      enabled: true,
      saveLocally: true,      
    },
```

When enabled, the db adapters will return the working directory as the location to save the `.db` folder. 

#### To Test
1. Enable the db flag above in your `truffle-config.js`. 
2. I tested using db-kit, it's enough to just start the application. 

ER: `.db` directory exists in `~/.config/truffle-nodejs/` (@truffle/config `getTruffleDataDirectory()`)

3. Bomb out of db-kit (ctrl-c)
4. Enable the `saveLocally` config flag above with the already enabled db flag. 
5. Start db-kit

ER: `.db` directory exists in your truffle project directory (@truffle/config `detect().working_directory`)

Closes #4495

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,6 @@ packages/debugger/.tmp
 
 # @truffle/contract
 packages/contract/dist
+
+# @truffle/db
+**/*.db
--- a/packages/db/README.md
+++ b/packages/db/README.md
@@ -22,6 +22,7 @@ system that can keep track of multiple contracts in this situation.
 ## Built With
 
 Truffle DB is built with:
+
 - TypeScript
 - GraphQL
 - Apollo
@@ -36,9 +37,17 @@ db: {
   enabled: true
 }
 ```
+
 Note: Enabling Truffle DB does not affect artifacts, but will produce a new `.db`
 directory when you compile or migrate your project.
 
+Add the following to your user truffle config (`~/.config/truffle-nodejs/config.json`) to enable saving the `.db` directory to your project directory
+
+```
+db: {
+  saveToProjectRoot: true
+}
+```
 
 It will soon be possible to load and access Truffle DB data via `truffle compile` and `truffle migrate`.
 Stay tuned!
--- a/packages/db/README.md
+++ b/packages/db/README.md
@@ -33,7 +33,8 @@ Add the following to your `truffle-config.js` file in order to enable Truffle DB
 
 ```
 db: {
-  enabled: true
+  enabled: true,
+  saveLocally: true // Optional - save the .db directory to the project directory. Default: false.
 }
 ```
 Note: Enabling Truffle DB does not affect artifacts, but will produce a new `.db`
--- a/packages/db/README.md
+++ b/packages/db/README.md
@@ -33,8 +33,7 @@ Add the following to your `truffle-config.js` file in order to enable Truffle DB
 
 ```
 db: {
-  enabled: true,
-  saveLocally: true // Optional - save the .db directory to the project directory. Default: false.
+  enabled: true
 }
 ```
 Note: Enabling Truffle DB does not affect artifacts, but will produce a new `.db`
--- a/packages/db/src/meta/pouch/adapters/fs.ts
+++ b/packages/db/src/meta/pouch/adapters/fs.ts
@@ -29,7 +29,7 @@ export const getDefaultSettings: GetDefaultSettings = () => {
   return {
     directory: path.join(
       userConfig.get("db")?.saveLocally 
-        ? new Config().working_directory  
+        ? Config.default().working_directory 
         : Config.getTruffleDataDirectory()
         , ".db", "json")
   };
--- a/packages/db/src/meta/pouch/adapters/fs.ts
+++ b/packages/db/src/meta/pouch/adapters/fs.ts
@@ -17,7 +17,7 @@ export interface DatabasesSettings {
 }
 
 type UserConfigDbSettings = {
-  saveToProjectRoot: boolean | undefined,
+  saveToProjectRoot?: boolean,
 };
 
 type UserConfig = {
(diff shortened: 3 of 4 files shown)


gnidan:
It looks like this PR modifies the project-level config, where we ideally want to save users the trouble of configuring this on a per-project basis... instead I think we want this to be a user-level config option. Happy to explain more in our next call.

packages/db/README.md
```diff
@@ -33,7 +33,8 @@ Add the following to your `truffle-config.js` file in order to enable Truffle DB
 
 ```
 db: {
-  enabled: true
+  enabled: true,
+  saveLocally: true // Optional - save the .db directory to the project directory. Default: false.
```
gnidan, line 37: This shouldn't be in this example anymore, since we've moved this option from truffle-config over to the user-level config

packages/db/src/meta/pouch/adapters/fs.ts
```diff
@@ -15,10 +15,25 @@ import Config from "@truffle/config";
 export interface DatabasesSettings {
   directory: string;
 }
+type UserConfigDbSettings = {
```
gnidan, line 18: missing blank line between top-level identifiers:

```suggestion
}

type UserConfigDbSettings = {
```

packages/db/src/meta/pouch/adapters/fs.ts
```diff
@@ -15,10 +15,25 @@ import Config from "@truffle/config";
...
+  get: (key: "db") => UserConfigDbSettings,
+};
+
+export const getDefaultSettings: GetDefaultSettings = () => {
+  const userConfig: UserConfig = Config.getUserConfig();
+  
+  return {
+    directory: path.join(
+      userConfig.get("db")?.saveLocally 
+        ? new Config().working_directory
```
gnidan, line 32: `new Config()` is generally considered private, and you should do `Config.detect()` or `Config.default().merge({ ... })` instead.

Earlier I told you that we should just do `Config.default()`, but I think that's wrong: I think we do need a proper `Config.detect()` here

packages/db/src/meta/pouch/adapters/sqlite.ts
```diff
@@ -15,9 +15,25 @@ export interface DatabasesSettings {
...
+  get: (key: "db") => UserConfigDbSettings,
+};
+
+export const getDefaultSettings: GetDefaultSettings = () => {
+  const userConfig: UserConfig = Config.getUserConfig();
+  
+  return {
+    directory: path.join(
+      userConfig.get("db")?.saveLocally 
+        ? new Config().working_directory
```
gnidan, line 32: (same comment as above)

gnidan (changes requested):
few minor things. thanks for taking this on!

cds-amal:
`saveLocally` is ambiguous. Would you consider `db.saveToProjectRoot`?

packages/db/src/meta/pouch/adapters/fs.ts
tenthirtyone: :+1:  Switched back to `.detect()`

tenthirtyone:
@cds-amal You got it, thanks for taking a look.

.gitignore
```diff
@@ -17,3 +17,6 @@ packages/debugger/.tmp
 
 # @truffle/contract
 packages/contract/dist
+
+# @truffle/db
+**/*.db
```
gnidan, line 22: Heh, it's kind of unfortunate that this is necessary :)

packages/db/README.md
```diff
@@ -36,9 +37,17 @@ db: {
   enabled: true
 }
 ```
+
 Note: Enabling Truffle DB does not affect artifacts, but will produce a new `.db`
 directory when you compile or migrate your project.
 
+Add the following to your user truffle config (`~/.config/truffle-nodejs/config.json`) to enable saving the `.db` directory to your project directory
```
gnidan, line 44: So there's some infrastructure for doing this automatically via `truffle config set <...>`. I forget the exact syntax; @eggplantzzz or @cds-amal do you remember offhand?

In any event, we probably want to recommend that approach rather than forcing users to find this file and edit it manually - since the path differs based on OS (Mac will use `~/Library/Application\ Support/truffle-nodejs`, for instance)

gnidan (commented):
Looks good! Just one remaining comment about the README. Thanks @tenthirtyone!

packages/db/README.md
tenthirtyone: Oh fantastic. 

I wasn't aware of the convenience command. I can probably drill down - `packages/core/lib/commands/config.js` (but wouldn't mind the exact syntax).

cds-amal: Unfortunately [the infrastructure isn't quite there yet](https://github.com/trufflesuite/truffle/blob/644c4dea08c2313144642b10d4de100a4c2bd613/packages/core/lib/commands/config.js#L90-L91)

.gitignore
cds-amal: This will be interesting. We'd want to encourage backups and such, but git repositories don't seem like the correct mechanism to use.

packages/db/README.md
gnidan: Maybe we need to do a separate PR to get this TODO addressed, and then rebase this once that's merged?

tenthirtyone: I can jump over to tackle it since it's all connected.

tenthirtyone:
Closing and making it part of https://github.com/trufflesuite/truffle/pull/4680

PR #4506, opened 2021-12-08
Allow more permissive format in reason determination (make `reason.js` work with Ganache 7)

haltman-at:
I just noticed that `reason.js` causes crashes when used with Ganache 7, due to it returning its errors in a new format.  So on a revert, you get a useless error message, instead of the revert message.

I've fixed this up to work with Ganache 7 while still working the old way.

Um, I only realized just now that this might be duplicative of work already going on on #4483.  @eggplantzzz, let me know if so and I can close this. :P

### DIFF
--- a/packages/contract/lib/reason.js
+++ b/packages/contract/lib/reason.js
@@ -24,13 +24,16 @@ const reason = {
 
     if (isObject) {
       const data = res.error.data;
-      const hash = Object.keys(data)[0];
+      //new format: encoded message is in data.result
+      //old format: there is no data.result, but the encoded message can
+      //be found in data[hash].return, where hash is the first key
+      const encodedMessage = data.result || Object.values(data)[0].return;
 
-      if (data[hash].return && data[hash].return.includes(errorStringHash)) {
+      if (encodedMessage && encodedMessage.includes(errorStringHash)) {
         try {
           return web3.eth.abi.decodeParameter(
             "string",
-            data[hash].return.slice(10)
+            encodedMessage.slice(10)
           );
         } catch (_) {
           return undefined;


packages/contract/lib/reason.js
```diff
@@ -24,13 +24,16 @@ const reason = {
...
     if (isObject) {
       const data = res.error.data;
-      const hash = Object.keys(data)[0];
+      //new format: encoded message is in data.result
+      //old format: there is no data.result, but the encoded message can
+      //be found in data[hash].return, where hash is the first key
+      const encodedMessage = data.result || Object.values(data)[0].return;
 
-      if (data[hash].return && data[hash].return.includes(errorStringHash)) {
+      if (encodedMessage && encodedMessage.includes(errorStringHash)) {
```
cds-amal, line 32: This is a really nice solution!

cds-amal (approved):
LGTM! merging is up to y'all

eggplantzzz:
Yes this is duplicate work. Take a look at the edit made to `reason.js` in the other PR https://github.com/trufflesuite/truffle/pull/4483/files. @davidmurdoch was the one that made that edit and it seems to handle a bit more than this PR.

But I admit that you and David might be in a better place to judge which solution is better. So take a look and let me know whether we should accept this solution or close it in favor of the v7 PR. Thanks!

THRIVEbluegooglecheck:
Video game controller.
buttona=Android volume up
buttonb=Androidvolume down
buttonc=Power Button

davidmurdoch:
@eggplantzzz I like harry's solution better, but it could use some comments explaining why we have two ways of handling this.

THRIVEbluegooglecheck:
G-11:11

haltman-at:
I'll just close this and I can look at / comment on #4483 when it's ready.

PR #4507, opened 2021-12-08
Amending the build system for debugger to webpack5

michaeljohnbennett:
This helps, hopefully, with the sourcemap resolution and build config.

Also during this we spotted a typo in the reducers.js that when fixed
causes tests to fail. Have added a FIXME on this line.

Will comment in the PR.


packages/debugger/lib/stacktrace/reducers.js
```diff
@@ -81,13 +81,14 @@ function lastPosition(state = null, action) {
   switch (action.type) {
     case actions.JUMP_IN:
     case actions.JUMP_OUT:
-    case actions.ETERNAL_CALL:
+    case actions.EXTERNAL_CALL:
```
haltman-at, line 84: Btw, reminder for the future, this sort of thing should really be a separate PR.  But it's already in here so, whatever!

packages/debugger/package.json
```diff
@@ -19,9 +19,9 @@
     "docs": "esdoc",
     "prepare": "yarn build",
     "start": "node ./webpack/dev-server.js",
-    "test": "mocha-webpack --webpack-config webpack/webpack.config-test.js --recursive; STATUS=$?; rm -r dist; yarn build; (exit $STATUS)",
-    "test:coverage": "nyc mocha-webpack --webpack-config webpack/webpack.config-test.js --recursive",
-    "test:debug": "NODE_ENV=testing node --inspect ./node_modules/.bin/mocha-webpack --webpack-config webpack/webpack.config-test.js --recursive"
+    "test": "mochapack --webpack-config webpack/webpack.config-test.js --recursive && yarn build; (exit $STATUS)",
```
haltman-at, line 22: So now that we're using mochapack, we should be able to get rid of the junk on this line, right?  We should be able to just do `mochapack --webpack-config webpack/webpack.config-test.js --recursive`?  The reason for all the rest was due to handling some problems caused by mocha-webpack, and the `(exit $STATUS)` doesn't even make sense anymore.

(OK, I just checked, the problem is indeed gone and we can get rid of all this.)

packages/debugger/package.json
```diff
@@ -49,11 +49,10 @@
     "@truffle/migrate": "^3.2.55",
     "@truffle/resolver": "^7.0.39",
     "@truffle/workflow-compile": "^3.2.41",
-    "babel-core": "^6.26.0",
-    "babel-loader": "^7.1.2",
-    "babel-plugin-transform-object-rest-spread": "^6.26.0",
-    "babel-plugin-transform-runtime": "^6.23.0",
-    "babel-preset-env": "^1.6.1",
+    "@babel/core": "^7.16.0",
```
haltman-at, line 52: Whoa, you managed to upgrade to Babel 7 and get things working, amazing!

Of course, that leaves the question of if we even still *need* babel; maybe that can be torn out?  Or maybe that'd be better saved for a later PR.

packages/debugger/package.json
```diff
@@ -64,17 +63,16 @@
     "faker": "^5.3.1",
     "fs-extra": "^9.1.0",
     "ganache-core": "2.13.0",
-    "istanbul-instrumenter-loader": "^3.0.1",
+    "@jsdevtools/coverage-istanbul-loader": "^3.0.5",
     "mocha": "5.2.0",
-    "mocha-webpack": "2.0.0-beta.0",
+    "mochapack": "^2.1.2",
```
haltman-at, line 68: :tada:

packages/debugger/webpack/webpack.config.js
```diff
@@ -1,4 +1,4 @@
-const merge = require("webpack-merge");
+const {merge} = require("webpack-merge");
```
haltman-at, line 1: Incredibly nitpicky style comment: Should be `{ merge }` instead of `{merge}`. :P  (We're supposed to have `prettier` to handle such things, but it's broken at the moment... @cds-amal is working on getting it working again, though.)

haltman-at (commented):
So, my main reaction on looking at this is that I'm not really sure I'm qualified to review this... I don't understand most of what's going on here!  Then again, I'm not really sure any of us do, so, uh, yeah.  Anyway, I've commented on what I can, but mostly I'm just trusting you to know what you're doing here and the tests to catch any problems.

One thing that doesn't fit on any particular line:  When I run the debugger tests now, the check marks on the individual tests don't show up green.  Is there any way to fix that?  Screenshot to illustrate:

nocolor

packages/debugger/package.json
michaeljohnbennett: I think a later PR, there are some things its doing to the code that would need a bit more of cleanup I think.

packages/debugger/webpack/webpack.config.js
michaeljohnbennett: I shall fix.

packages/debugger/package.json
michaeljohnbennett: I'll remove this.

michaeljohnbennett:
on the issue of the CLI colors for mocha, yes, I'm not sure, I think this is from mochapack and I did see this come up as some kind of deprecation issue. I'm going to bump the mocha version up and recommit, see if this fixes it. I imagine mochapack expects a more recent mocha version and is probably using the newer command line switches for colors.

packages/debugger/lib/stacktrace/reducers.js
michaeljohnbennett: I know, it was the one line that started all this 🤷🏻‍♂️

michaeljohnbennett:
upping mocha has fixed the colors, hopefully the merge fixes and yarn files will get the pre-checks all green 🤞🏻

haltman-at (approved):
Seems good to me, I guess!  I mostly just have to trust you know what you're doing here, but, well, the tests are passing, so presumably it's correct!

PR #4508, opened 2021-12-08
Bug fix: Fix incorrect action name and account for null location

haltman-at:
This is an odd one.  The `EXTERNAL_CALL` case in the `lastPosition` reducer in `stacktrace` was typo'd to `ETERNAL_CALL`.  So external calls were getting ignored!

Why didn't this cause test failures?  Well, presumably because this would only affect the source positions given in the stacktrace, which we weren't testing.  (Notionally I could add tests for those, but due to this PR's status as an add-on to another PR, that would be impractical here; we can go back and do that later if we think it's necessary.)

Note that fixing the typo caused failures though... because I didn't account for the fact that the initial `EXTERNAL_CALL` has `location` equal to `null`.  So, I just checked for that case and ignored it.

### DIFF
--- a/packages/debugger/lib/stacktrace/reducers.js
+++ b/packages/debugger/lib/stacktrace/reducers.js
@@ -81,14 +81,14 @@ function lastPosition(state = null, action) {
   switch (action.type) {
     case actions.JUMP_IN:
     case actions.JUMP_OUT:
-    // FIXME: WHEN THIS IS ADDED IN CORRECTLY THE TESTS FAIL.
-    // case actions.EXTERNAL_CALL:
+    case actions.EXTERNAL_CALL:
     case actions.EXTERNAL_RETURN:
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
Makes sense @haltman-at . Can we open an issue for adding the tests?

michaeljohnbennett (approved):

michaeljohnbennett:
this all looks good, It seems when merging in develop its caused some regression with vyper command line tests (single test fails). 

will need to investigate why.

cds-amal:
Can you try rebasing against develop instead of merging? the latest develop uses python3.8 in order to satisfy vyper v0.3.1.

haltman-at:
> Makes sense @haltman-at . Can we open an issue for adding the tests?

You can make an issue if you want, but otherwise I could just add tests later?

PR #4510, opened 2021-12-08
fix typo

MehdizadeMilad:


eggplantzzz:
Thanks for suggesting this change to the docs! I changed the PR to be against the `develop` branch as that is the branch we use to merge new work.

packages/contract/README.md
```diff
@@ -308,7 +308,7 @@ instance.getValue.call().then(function(val) {
 });
 ```
 
-Even more helpful, however is we *don't even need* to use `.call` when a function is marked as `view` or `pure`, because `@truffle/contract` will automatically know that that function can only be interacted with via a call:
+Even more helpful, however we *don't even need* to use `.call` when a function is marked as `view` or `pure`, because `@truffle/contract` will automatically know that that function can only be interacted with via a call:
```
eggplantzzz, line 311: I think we want to leave in the "is" in this case as this part of the sentence requires a verb to be grammatically correct and to say exactly *what* is more helpful.

cds-amal (commented):
I agree with @eggplantzzz though a comma after "however" may be helpful.

packages/contract/README.md
haltman-at: The `is` isn't necessarily here; the verb phrase is that we don't even need etc.  But there should be a comma after the however.

eggplantzzz: If you write "Even more helpful, ..." it seems that you are referencing something that was just mentioned. That is not the case here. This text means to say, "hey, here is something else that is helpful".

haltman-at: ? This is referencing the `.call` functionality that was described immediately preceding.

eggplantzzz: Yeah, it is referencing `.call` which is goes on to describe afterwards.

eggplantzzz:
So I think we should close this, any opposed?

cds-amal:
no opposition 👍

Issue #4482, opened 2021-12-02
Provides default configuration of the main network supporting EIP 1559


wxy:
- [X] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue. (Ticket: 521615)

---------------------------

## Issue

The default recommended configuration for mainnet is not provided in `truffle-config`, so it leads to misconfiguration.

I used the old style configuration parameter, gas/gasprice, which works fine for both the local development environment and the Rinkeby network. However, it now does not work on Mainnet with EIP 1559 support.

Using the gas/gasprice parameter, contract deployment can be done, but strange errors occur during console interactions, even for read-only function calls.

The only way to interact properly is to replace the parameters with maxFeePerGas/maxPriorityFeePerGas for Type 2 transactions.

So, I suggest to give the default mainnet configuration parameters in `truffle-config` , rather than just mentioning them in your online documentation.

Finally, one more suggestion: Due to the unstable gas state of the ethereum network, predefining a fixed amount of gas may cause network confirmation timeouts, so I suggest providing a fastest/most economical parameter so that Truffle can prioritize the fastest or most economical way to complete contracts when deploying them.

fainashalts:
Hi @wxy just to clarify, you mean the default configuration in the `truffle-config.js` file that you get when you run `truffle init`, right? I think this is a great suggestion and we will add it to our backlog. This is also a great first issue, so feel free to jump in with a PR if you're interested or hoping this gets resolved sooner!

wxy:
Thank you for your reply. I can indeed try to raise a PR rather than just a issue. I'll give it a try.

PR #4516, opened 2021-12-09
Adding a default configuration to the Mainnet

wxy:
As a solution to the #4482 issue, I have provided this patch.

The recommended configuration for the Mainnet should be provided in the truffle-conf.js. After upgrading to EIP-1559, the Mainnet reports strange errors when deploying and interacting with contracts via Truffle/Infura that do not use type 2 transactions.  For example:

- -32000 err: max fee per gas less than block base fee
- Error: cannot estimate gas; transaction may fail or may require manual gas limit

So, I recommend to provide the type 2 transaction parameter in the default configuration.

I also provided a copy of the Rinkeby configuration, since it seems that Ropsten is not popular and Opensea's test network is based on Rinkeby.

In addition, I added the HDWalletProvider's wallet serial number to the configuration. This allows for separate deployments to Mainnet and Rinkeby/Ropsten with different addresses in same mnemonic.

If there is anything wrong with this patch, please let me know and I will improve it. I am very happy to contribute to the Truffle project.

### DIFF
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -48,43 +48,31 @@ module.exports = {
     // },
     // Another network with more advanced options...
     // advanced: {
-    //  port: 8777,             // Custom port
-    //  network_id: 1342,       // Custom network
-    //  gas: 8500000,           // Gas sent with each transaction (default: ~6700000)
-    //  gasPrice: 20000000000,  // 20 gwei (in wei) (default: 100 gwei)
-    //  from: <address>,        // Account to send txs from (default: accounts[0])
+    //  port: 8777,            // Custom port
+    //  network_id: 1342,      // Custom network
+    //  gas: 8500000,          // Gas sent with each transaction (default: ~6700000)
+    //  gasPrice: 20000000000, // 20 gwei (in wei) (default: 100 gwei)
+    //  from: <address>,       // Account to send txs from (default: accounts[0])
     //  websocket: true        // Enable EventEmitter interface for web3 (default: false)
     // },
     // Useful for deploying to a public network.
+    // You can create separate configurations for multiple public networks using different names, e.g. mainnet, ropsten, rinkeby.
+    // public: {
     // NB: It's important to wrap the provider as a function.
-    // mainnet: {
-    //  provider: () => new HDWalletProvider(mnemonic, `https://mainnet.infura.io/v3/YOUR-PROJECT-ID`, 0),
-    //  network_id: 1,       // Mainnet's id
-    // NB: You need to use type 2 transactions on the main network, otherwise you may get odd errors when interacting.
-    // Type 2 transactions use the maxFeePerGas/maxPriorityFeePerGas instead of gas/gasprice parameter (https://eips.ethereum.org/EIPS/eip-1559).
-    // Please set it according to the current network condition, 20% higher than the minimum value is recommended.
-    //  maxFeePerGas: 100000000000,       // 100 gwei 
-    //  maxPriorityFeePerGas: 1500000000, // 1.5 gwei
-    //  confirmations: 2,    // # of confs to wait between deployments. (default: 0)
-    //  timeoutBlocks: 200,  // # of blocks before a deployment times out  (minimum/default: 50)
-    //  skipDryRun: true     // Skip dry run before migrations? (default: false for public nets )
-    // },
-    // rinkeby: {
-    //  provider: () => new HDWalletProvider(mnemonic, `https://rinkeby.infura.io/v3/YOUR-PROJECT-ID`, 1),
-    //  network_id: 4,       // Rinkeby's id
-    //  gas: 6000000,        // Rinkeby has a lower block limit than mainnet
-    //  gasPrice: 1000000000,// 1 gwei
-    //  confirmations: 2,    // # of confs to wait between deployments. (default: 0)
-    //  timeoutBlocks: 200,  // # of blocks before a deployment times out  (minimum/default: 50)
-    //  skipDryRun: true     // Skip dry run before migrations? (default: false for public nets )
-    // },
-    // ropsten: {
-    //  provider: () => new HDWalletProvider(mnemonic, `https://ropsten.infura.io/v3/YOUR-PROJECT-ID`, 2),
-    //  network_id: 3,       // Ropsten's id
-    //  gas: 5500000,        // Ropsten has a lower block limit than mainnet
-    //  confirmations: 2,    // # of confs to wait between deployments. (default: 0)
-    //  timeoutBlocks: 200,  // # of blocks before a deployment times out  (minimum/default: 50)
-    //  skipDryRun: true     // Skip dry run before migrations? (default: false for public nets )
+    // NB: Please use a different provider URL for different public networks.For examples, 
+    // NB:   mainnet : `https://mainnet.infura.io/v3/YOUR-PROJECT-ID`
+    // NB:   rinkeby : `https://rinkeby.infura.io/v3/YOUR-PROJECT-ID`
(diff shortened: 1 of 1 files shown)


haltman-at (commented):
Hm, do we really want a whole *separate* mainnet configuration here?  (And rinkeby, and ropsten...)  Maybe it'd be better to just add these to the example `advanced` network.

Thanks for adjusting the indentation, btw; that's been bugging me. :)

wxy:
> Hm, do we really want a whole _separate_ mainnet configuration here? (And rinkeby, and ropsten...) Maybe it'd be better to just add these to the example `advanced` network.

Yes, we have advanced network configurations, but not for EIP-1559. And, it might be easier for developers if there is an official reference for the mainnet configuration?

I'm raising this PR because I'm really troubled by this situation and hope it can help more developers. Make it easy for them to add the mainnet configuration instead of digging into the documentation (I found this out by reading the documentation again.)

Thanks for your reply.

cds-amal:
I find these boilerplate comments very helpful in other projects and would absolutely take advantage of them when setting up truffle projects instead of hunting down documentation :) If there's an way to reduce the boilerplate without losing information I'd support that too.

cds-amal:
This seems like a good compromise @eggplantzzz,  what do you think @wxy ?

wxy:
@cds-amal @eggplantzzz , thank you for your suggestions.

This is an acceptable solution.  I have just tested type 2 transactions on rinkeby/ropsten and it also works.So it is possible to provide suggested parameters for type 2 transactions for both mainnet/rinkeby/ropsten.

If everyone agrees, then I'll go and change the configuration?

wxy:
Also, I have a best practice for the configuration here, which could probably be put in the documentation.When using HDWalletProvider, you can create multiple wallets under one helper and use different wallets for testing and deployment, e.g. wallet 0 for ropsten, wallet 1 for Rinkeby, wallet 2 for Mainnet, etc. This avoids being able to track behaviour between networks based on the same wallet address.

cds-amal:
> Also, I have a best practice for the configuration here, which could probably be put in the documentation.When using HDWalletProvider, you can create multiple wallets under one helper and use different wallets for testing and deployment, e.g. wallet 0 for ropsten, wallet 1 for Rinkeby, wallet 2 for Mainnet, etc. This avoids being able to track behaviour between networks based on the same wallet address.

Can you share an example please? I'm not sure what you mean

wxy:
> Can you share an example please? I'm not sure what you mean

for examples:

```
// mainnet
new HDWalletProvider(mnemonic, `https://mainnet.infura.io/v3/YOUR-PROJECT-ID`, 0),
// ropsten
new HDWalletProvider(mnemonic, `https://mainnet.infura.io/v3/YOUR-PROJECT-ID`, 1),
// rinkeby
new HDWalletProvider(mnemonic, `https://mainnet.infura.io/v3/YOUR-PROJECT-ID`, 2),
```

wxy:
Please review it or leave some comments.
@gnidan @cds-amal @eggplantzzz

cds-amal:
Hi @wxy, the team is offline and away on holiday for two weeks. I'm sure we'll get back to you then. I'm an advocate for improving our boilerplate and appreciate the thoughtfulness your raised in this PR. Sorry about the timing :/.

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -48,28 +48,37 @@ module.exports = {
...
-    // gas: 5500000,        // Ropsten has a lower block limit than mainnet
-    // confirmations: 2,    // # of confs to wait between deployments. (default: 0)
-    // timeoutBlocks: 200,  // # of blocks before a deployment times out  (minimum/default: 50)
-    // skipDryRun: true     // Skip dry run before migrations? (default: false for public nets )
+    // NB: Please use a different provider URL for different public networks.For examples, 
+    // NB:   mainnet : `https://mainnet.infura.io/v3/YOUR-PROJECT-ID`
+    // NB:   rinkeby : `https://rinkeby.infura.io/v3/YOUR-PROJECT-ID`
+    // NB: Also, it is recommended to use the same mnemonic and use different indexes (0, 1, 2 ...) for different public networks.
+    //  provider: () => new HDWalletProvider(mnemonic, `https://network.io`, 0),
+    //  network_id: 1,                      // Network ID: mainnet => 1; ropsteb => 3; rinkeby => 4
```
eggplantzzz, line 67: misspelled "ropsten" here (ropsteb)

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -48,28 +48,37 @@ module.exports = {
...
+    // NB: You need to use type 2 transactions on the public network, otherwise you may get odd errors when interacting.
```
eggplantzzz, line 68: I don't really think it is necessary yet to use type 2 transactions - I think you just won't benefit from these newer features, right?

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -48,28 +48,37 @@ module.exports = {
...
+    // NB: Type 2 transactions use the maxFeePerGas/maxPriorityFeePerGas instead of gas/gasprice parameter (https://eips.ethereum.org/EIPS/eip-1559).
```
eggplantzzz, line 69: `gasprice` should be `gasPrice`

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -48,28 +48,37 @@ module.exports = {
...
+    //  timeoutBlocks: 200,                 // # of blocks before a deployment times out  (minimum/default: 50)
+    //  skipDryRun: true                    // Skip dry run before migrations? (default: false for public nets )
     // },
     // Useful for private networks
     // private: {
-    // provider: () => new HDWalletProvider(mnemonic, `https://network.io`),
-    // network_id: 2111,   // This network is yours, in the cloud.
-    // production: true    // Treats this network as if it was a public net. (default: false)
+    //  provider: () => new HDWalletProvider(mnemonic, `https://network.io`),
+    //  network_id: 2111,   // This network is yours, in the cloud.
```
eggplantzzz, line 80: why do you write "in the cloud" here?

packages/core/lib/commands/init/initSource/truffle-config.js
haltman-at: That's my understanding also...

packages/core/lib/commands/init/initSource/truffle-config.js
wxy: Original text

packages/core/lib/commands/init/initSource/truffle-config.js
cds-amal: @wxy can you share an example of odd errors you encountered?

wxy: @cds-amal @haltman-at @eggplantzzz 
I've mentioned it in the commit comments for this PR, but I don't have the original full log and error message anymore.

This is just a few suggestions and perhaps just my own personal experience of the problem. I thought it might be helpful to more people as well, hence submit this PR. Also, the trend is to use type 2 transactions anyway, so I would suggest changing to type 2. however, if this is officially deemed unnecessary, then I will close this PR.

wxy: Also, for type 2 transactions, in addition to fixing the error I encountered, I think it has the extra benefit of using the gas/gasPrice parameter, where you may be paying more gas, and using maxFeePerGas/maxPriorityFeePerGas, which will use no more gas than the limit based on current network conditions.

cds-amal: Thanks @wxy , I will discuss with the team. Preparing for the Type 2 transactions was a perspective I didn't consider. Thanks for raising it!

wxy:
Since an official decision is required, I will close this PR.

Issue #2124, opened 2019-06-22
feat: allow setting mocha reporter via CLI


yxliang01:
mocha has a cool feature that allows the change of reporter for different wanted output formats. However, currently Truffle doesn't support the change of reporter via CLI. I don't think setting the frequently-changing reporter via config is a good idea either. This issue is a feature request that adds a new command line argument to dynamically change the mocha reporter.

Thanks

eggplantzzz:
Thanks for the suggestion, we'll look into it!

haltman-at:
This is now released!

PR #4517, opened 2021-12-09
Added command line argument for changing mocha reporter

sukanyaparashar:
### ISSUE
To allow setting mocha reporter as a command line argument rather than changing it in the truffle-config.js file while using $truffle test command. See issue #2124.

### SOLUTION
A command line argument has been added as an enhancement feature in the $truffle test command for changing the mocha reporter type. The command can be used as - $truffle test --reporter reporter-name.

### DIFF
--- a/packages/core/lib/commands/test/index.js
+++ b/packages/core/lib/commands/test/index.js
@@ -132,7 +132,7 @@ const command = {
       },
       {
         option: "--reporter|-r <name>",
-        description: "Specify the reporter type for mocha test. Default: 'spec'"
+        description: "Specify the type of mocha reporter to use during testing. Default: 'spec'"
       }
     ],
     allowedGlobalOptions: ["network", "config"]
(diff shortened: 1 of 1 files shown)


packages/core/lib/commands/test/index.js
```diff
@@ -123,6 +129,10 @@ const command = {
         option: "--grep|-g",
         description: "Use mocha's \"grep\" option while running tests. This " +
           "option only runs tests that match the supplied regex/string."
+      },
+      {
+        option: "--reporter|-r <name>",
+        description: "Specify the reporter type for mocha test. Default: 'spec'"
```
eggplantzzz, line 135: This sentence sounds a bit odd. Maybe it would be better if it was something like, "Specify the type of mocha reporter to use during testing."?

sukanyaparashar: I think your sentence is sounding more proper @eggplantzzz . I will change it as you suggested.

eggplantzzz (approved):
I like it!

PR #4523, opened 2021-12-11
Dependency update: Update lerna to 4.0.0

haltman-at:
Note: I ran into a problem doing this where updating lerna to 4.0.0 caused webpack to fail on `rx-lite-aggregates` (a transitive dependency).  I had to look up some [internet black magic](https://stackoverflow.com/a/51054589) to fix this.  Frankly I don't really understand this fix, but it seems to do the job.


cds-amal (approved):
Looks good to me.

gnidan (changes requested):
Just putting "request changes" here so we can pump the brakes.

At this point it seems like we should wait til the new year to merge a lerna update like this? I know that I had said we could just do a Wednesday release this week, but we won't have the team at full capacity to resolve issues until January, and it just seems prudent to wait.

Happy to revisit this call if people have strong feelings!

haltman-at:
Cool, holding off on this makes sense to me.

eggplantzzz (approved):

Issue #4521, opened 2021-12-09
Internal issue: Most `resolverOptions` in resolver are now unnecessary and should be replaced by use of compiler info


haltman-at:
## Issue

Now that PRs #4420 and #4365 have been merged, the `translateJsonToSolidity` and `resolveVyperModules` resolver options are a redundant mechanism.

These options should be removed, and the checks that switch on them should instead look at the compiler version that was passed (`translateJsonToSolidity` should become, are we using Solidity, and `resolveVyperModules` should become, are we using Vyper).

Moreover, any code that passes these options in should changed back so that it no longer does so.

## Environment

* Truffle version (`truffle version`): 5.4.23

leeftk:
This was resolved in the new release 5.4.27. Closing!

Issue #3855, opened 2021-02-20
Support writing tests in Vyper


haltman-at:
## Issue

Don't know if this is worth doing, but it's a thing we could do.  Note that there's some trickiness here because we wouldn't be able to easily supply an assertion library in the same way we do with Solidity, because Vyper has no notion of linking.  The best idea @gnidan and I had was that test contracts in Vyper would optionally accept an address parameter to its constructor, and when deploying them we'd give them the address of the assertion library as the argument to the constructor (assuming the constructor takes an argument).

Note that we wouldn't necessarily need to write separate Vyper assertion libraries; we could just add JSON files with the ABIs for our existing assertion libraries and rely on Vyper's native ability to import JSON.

PR #4529, opened 2021-12-11
Internal improvement: Remove unnecessary resolver options, check compiler instead

haltman-at:
Addresses #4521.  **Is a breaking change to `@truffle/resolver`**.

Most of what there is to say about this one is already discussed in #4521.  The awkward option that defaulted to `true` is now gone; I imagine Nick will be happy about that.

Only one resolver invocation needed to be altered, and that was the one in `compile-vyper`, because that's the one all these auxiliary options were created for!  Now that `includeTruffleSources` is once again the only resolver option, I could notionally change the signature back to the old way, where instead of passing `{ includeTruffleSources: true }` you just passed `true`, but, um, I think it's fairly obvious why we don't want to do that.

(Getting rid of those extra options is also quite helpful if we ever want to implement #3855. :)  The way the options worked previously wouldn't have worked very well with that.)

Note that passing the full version to the profiler meant that I had to move the call to `checkVyper` a bit earlier than it previously was.  Previously it would only be called if there were actual Vyper files to compile; now it can be called as long as there is Vyper in your project, even if no Vyper needs to be compiled at the moment.  This is relevant because the call to `checkVyper` can be a bit slow.  I'm guessing this is probably OK?  But figured it's worth asking.

~~I considered moving all the compiler-checking logic into the individual resolver sources, so that we could just always have all resolver sources active, rather than the current logic where the language controls which ones are active; however, that seemed like unnecessary work, and is a bit awkward with the wrapper thing, so I didn't do that.~~  Edit: I went back and redid it this way!  See later comments.

### DIFF
--- a/packages/resolver/lib/resolver.ts
+++ b/packages/resolver/lib/resolver.ts
@@ -25,26 +25,27 @@ export class Resolver {
     const { includeTruffleSources } = resolverOptions;
 
     this.options = options;
-    this.sources = [
+
+    let basicSources: ResolverSource[] = [
       new EthPMv1(options.working_directory),
       new NPM(options.working_directory),
       new GlobalNPM(),
       new FS(options.working_directory, options.contracts_build_directory)
     ];
-
     if (includeTruffleSources) {
-      this.sources.unshift(new Truffle(options));
+      basicSources.unshift(new Truffle(options));
     }
 
-    if (options.compiler && options.compiler.name === "solc") {
-      this.sources = [].concat(
-        ...this.sources.map(source => [new ABI(source), source])
-      );
-    }
+    //set up abi-to-sol resolution
+    this.sources = [].concat(
+      ...basicSources.map(source => [new ABI(source), source])
+    );
 
-    if (options.compiler && options.compiler.name === "vyper") {
-      this.sources = [new Vyper(this.sources, options.contracts_directory)];
-    }
+    //set up vyper resolution rules
+    this.sources = [
+      new Vyper(basicSources, options.contracts_directory),
+      ...this.sources //for Vyper this is redundant
+    ];
   }
 
   // This function might be doing too much. If so, too bad (for now).
--- a/packages/resolver/lib/resolver.ts
+++ b/packages/resolver/lib/resolver.ts
@@ -10,15 +10,7 @@ import { EthPMv1, NPM, GlobalNPM, FS, Truffle, ABI, Vyper } from "./sources";
 
 export interface ResolverOptions {
   includeTruffleSources?: boolean;
-  translateJsonToSolidity?: boolean;
-  resolveVyperModules?: boolean;
 }
-
-const defaultResolverOptions = {
-  includeTruffleSources: false,
-  translateJsonToSolidity: true,
-  resolveVyperMoudles: false,
-};
 
 export class Resolver {
   options: any;
@@ -30,15 +22,7 @@ export class Resolver {
       ["working_directory", "contracts_build_directory", "contracts_directory"]
     );
 
-    resolverOptions = {
-      ...defaultResolverOptions,
-      ...resolverOptions
-    };
-    const {
-      includeTruffleSources,
-      translateJsonToSolidity,
-      resolveVyperModules
-    } = resolverOptions;
+    const { includeTruffleSources } = resolverOptions;
 
     this.options = options;
     this.sources = [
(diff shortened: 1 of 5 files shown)


gnidan:
> **Arguably counts as a breaking change to `@truffle/resolver`**; @gnidan, can you comment on that?

Yep, I'd say it's breaking! Internal-only packages we can break for whatever the dang heck we want :)

> ... but, um, I think it's fairly obvious why we don't want to do that.

:D

> Previously it would only be called if there were actual Vyper files to compile; now it can be called as long as there is Vyper in your project, even if no Vyper needs to be compiled at the moment. This is relevant because the call to `checkVyper` can be a bit slow. I'm guessing this is probably OK? But figured it's worth asking.

Definitely worth asking! I'm inclined to disagree—the whole point of having this profiler is to reduce extra compiler work :). Do we really want to call out to Vyper if we're only compiling Solidity files?

> I considered moving all the compiler-checking logic into the individual resolver sources, so that we could just always have all resolver sources active, rather than the current logic where the language controls which ones are active; however, that seemed like unnecessary work, and is a bit awkward with the wrapper thing, so I didn't do that.

👍 but maybe we consider this in the future

haltman-at:
> Definitely worth asking! I'm inclined to disagree—the whole point of having this profiler is to reduce extra compiler work :). Do we really want to call out to Vyper if we're only compiling Solidity files?

Hm, true.  How do we handle it then -- just leave out the version and put in a `//HACK` comment?

gnidan:
> > Definitely worth asking! I'm inclined to disagree—the whole point of having this profiler is to reduce extra compiler work :). Do we really want to call out to Vyper if we're only compiling Solidity files?
> 
> Hm, true. How do we handle it then -- just leave out the version and put in a `//HACK` comment?

I suppose :/

Let's see how it looks? It's gotta be better than those options you're removing, at least

haltman-at:
Oh hm, looks like this is failing CI because there's a case that relies on abi-to-sol, but the compiler isn't getting passed in as `"solc"` as required, so abi-to-sol isn't activated and the compilation fails.  Will have to figure out what to do about that.  (The answer may have to be, have abi-to-sol turned on when we can't identify the compiler, but I'll see if there's a better solution before resorting to that.)

gnidan:
> have abi-to-sol turned on when we can't identify the compiler

What version will you tell abi-to-sol to output?

haltman-at:
> What version will you tell abi-to-sol to output?

Let's see first *if* I have to resort to that before asking such questions!  But, um, apparently it must have *some* default behavior I can fall back on if it comes to that, because it was previously running in these failing cases where the version info isn't present...

gnidan:
> But, um, apparently it must have _some_ default behavior I can fall back on if it comes to that

Default is `>=0.7.0 <0.9.0`, but we could possibly extend that. Problem is, when you start to broaden the range to include 0.5.x or worse, you start running into scenarios where output Solidity would be forced to choose between conflicting syntax forms, and abi-to-sol throws an exception in such a situation.

It'd be good to know why it's not getting the info, whether this is only a problem for the test or something we will have to worry about.

haltman-at:
Well, yes, determining that is the next step here!  But I'm not doing that tonight. :P

haltman-at:
OK, so the way I did things had the big problem that I was looking for compiler on resolver *construction*, but it's passed on resolver *invocation*!

So, I redid it so that all resolver sources are always active, and checking the compiler is the job of the individual resolver sources.

This does have the annoying side effect that, when Vyper fails to resolve a source, that source gets checked twice.  But, oh well... the only ways I saw around that were kind of horrific.

packages/resolver/lib/resolver.ts
```diff
@@ -30,37 +22,30 @@ export class Resolver {
...
+      basicSources.unshift(new Truffle(options));
     }
 
-    if (translateJsonToSolidity) {
-      this.sources = [].concat(
-        ...this.sources.map(source => [new ABI(source), source])
-      );
-    }
+    //set up abi-to-sol resolution
+    this.sources = [].concat(
```
eggplantzzz, line 40: So now all sources will be "abi-to-sol"-ified whereas before it was optional? What cases were there before where we wouldn't want this to be the case?

haltman-at: Well, before it was on by default, but could be turned off; specifically, `compile-vyper` would turn it off.  We didn't want it on for Vyper.  Now instead, the `ABI` resolver source will check that you're using `solc`.  So the source will be *there* when using Vyper, it just won't *do* anything.  Same as how now, the `Vyper` resolver source will be there when you use Solidity, but it won't do anything.

eggplantzzz (approved):
I like it!

PR #4537, opened 2021-12-13
Set default solc to version range

kevinweaver:
## Problem
When a user runs `truffle init`, we set the default solc version to a specific version.  This is generally fine, though we could be more flexible.

## Solution
As per [the point](https://github.com/trufflesuite/truffle/pull/4371#discussion_r734055821) raised by @haltman-at, Truffle accepts version ranges and will use the latest matching version.  Let's add the `^` to account for supported versions in the range.

### DIFF
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -81,7 +81,7 @@ module.exports = {
   // Configure your compilers
   compilers: {
     solc: {
-      version: "0.8.10",    // Fetch exact version from solc-bin (default: truffle's version)
+      version: "^0.8.10",    // Fetch exact version from solc-bin (default: truffle's version)
       // docker: true,        // Use "0.5.1" you've installed locally with docker (default: false)
       // settings: {          // See the solidity docs for advice about optimization and evmVersion
       //  optimizer: {


gnidan:
Note that this means that Truffle will make extra network requests every time you compile, to see if there's a new version.

We have this pinned specifically to avoid this extra chattiness.

fainashalts:
> Note that this means that Truffle will make extra network requests every time you compile, to see if there's a new version.
> 
> We have this pinned specifically to avoid this extra chattiness.

 But if you use the wrong version don't you get that not so helpful error about the solc version?

> 

How big a deal is that? Isn't doing so better than getting an error at compilation? And couldn't the user bypass the network requests by specifying their version? I could be convinced either way but I sort of think this is an improvement...

gnidan:
We often have users who use Truffle with no internet connection, and they have reported this as a source of confusion (back when we had it set to caret)

As a result, we made dang-sure that the default compilation story works offline as much as possible.

Now, not opposed to changing, but I'm not sure I'm in favor either. Just trying to give the full story.

Not sure about compilation errors - aren't we usually pretty good about keeping the default set to latest these days?

fainashalts:
Hm, you make a good point. I hadn't really considered the offline user's needs, and our default version is updated now. I don't feel too strongly about either outcome here.

gnidan:
It's a tricky problem!

cds-amal:
I'm against raising network requirements and prefer we revisit bumping the version pin periodically.

kevinweaver:
Thanks for the context, this all makes sense!  Closing in favor of leaving it and bumping periodically.

Issue #4538, opened 2021-12-14
Exceptions that @truffle/fetch-and-compile throws suppress underlying cause


gnidan:
## Issue

@truffle/fetch-and-compile throws exceptions that are either "fetch" errors or "compile" errors, without providing detail about what the underlying failure condition was. This is problematic, and users of this package have asked for error information to be more detailed.

(See [relevant code](https://github.com/trufflesuite/truffle/blob/6d45d847cc1dc6d119becfc6b8e7ea69d7589754/packages/fetch-and-compile/lib/recognizer.ts#L47))

haltman-at:
Assuming you just want this for `fetchAndCompile`, and not more detailed errors from the package as a whole ( :P ) I think this should be doable.

PR #4539, opened 2021-12-14
Enhancement: Have fetchAndCompile rethrow particular errors

haltman-at:
Addresses #4538.

This PR makes it so that when `fetchAndCompile` hits an error when fetching or compiling, it will actually rethrow that error, rather than throwing a generic unhelpful error.  This required a slight change to the recognizer interface, but not even a breaking one.

(Note that `fetchAndCompileForDebugger` still swallows particular errors.)

This PR has no tests because tests for this package are waiting on #4424, but tests can be added once that PR is merged.

### DIFF
--- a/packages/fetch-and-compile/lib/fetch.ts
+++ b/packages/fetch-and-compile/lib/fetch.ts
@@ -83,6 +83,7 @@ async function tryFetchAndCompileAddress(
 ): Promise<void> {
   let found: boolean = false;
   let failureReason: FailureType | undefined; //undefined if no failure
+  let failureError: Error | undefined;
   //(this includes if no source is found)
   for (const fetcher of fetchers) {
     //now comes all the hard parts!
@@ -94,6 +95,7 @@ async function tryFetchAndCompileAddress(
     } catch (error) {
       debug("error in getting sources! %o", error);
       failureReason = "fetch";
+      failureError = error;
       continue;
     }
     if (result === null) {
@@ -129,6 +131,7 @@ async function tryFetchAndCompileAddress(
     } catch (error) {
       debug("compile error: %O", error);
       failureReason = "compile";
+      failureError = error;
       continue; //try again with a different fetcher, I guess?
     }
     //add it!
@@ -141,6 +144,7 @@ async function tryFetchAndCompileAddress(
       address
     );
     failureReason = undefined; //mark as *not* failed in case a previous fetcher failed
+    failureError = undefined;
     //check: did this actually help?
     debug("checking result");
     if (!recognizer.isAddressUnrecognized(address)) {
@@ -157,7 +161,7 @@ async function tryFetchAndCompileAddress(
   }
   if (found === false) {
     //if we couldn't find it, add it to the list of addresses to skip
-    recognizer.markUnrecognizable(address, failureReason);
+    recognizer.markUnrecognizable(address, failureReason, failureError);
   }
 }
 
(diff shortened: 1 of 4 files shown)


packages/fetch-and-compile/lib/fetch.ts
```diff
@@ -94,6 +95,7 @@ async function tryFetchAndCompileAddress(
     } catch (error) {
       debug("error in getting sources! %o", error);
       failureReason = "fetch";
+      failureError = error;
```
cds-amal, line 98: I might be missing something but It seems odd to introduce `failureError` and keep `failureReason`.

packages/fetch-and-compile/lib/recognizer.ts
```diff
@@ -39,9 +39,11 @@ export class SingleRecognizer implements Recognizer {
     return this.recognized ? undefined : this.address;
   }
 
-  markUnrecognizable(address: string, reason?: FailureType): never {
+  markUnrecognizable(address: string, reason?: FailureType, error?: Error): never {
     //just throw...
-    if (reason) {
+    if (error) {
+      throw error;
+    } else if (reason) {
```
cds-amal, line 46: This seems ok if `reason` can occur without `error`, but I may be misunderstanding the logic.

cds-amal (commented):
@haltman-at reason and error don't seem to be coupled to the same state. Is it possible to have a `reason` that doesn't include an `error`?

packages/fetch-and-compile/lib/recognizer.ts
haltman-at: There is one way at this can occur at the moment: If you try to fetch-and-compile a Vyper source, which this package doesn't support at the moment, then you'll end up here with `reason` equal to `"language"` but `error` not set (since nothing threw an actual error to rethrow).

cds-amal: Thanks for clarifying that.

davidmurdoch: would this be a good time to use an AggregateError? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError (polyfill for node < 15: https://www.npmjs.com/package/aggregate-error)

packages/fetch-and-compile/lib/fetch.ts
cds-amal: That is good to know! I'll have to remember the different clients

packages/fetch-and-compile/lib/recognizer.ts
davidmurdoch: Ganache uses one if multiple things go wrong on start up: https://github.com/trufflesuite/ganache/blob/956df4af117e3baa255abba4f4a38dee78eb6e64/src/packages/core/src/server.ts#L256

cds-amal (approved):
Thanks for the explanations @haltman-at . Looks good to me!

packages/fetch-and-compile/lib/recognizer.ts
haltman-at: Fascinating!  I didn't know about that.  Yes, I think it might!  But do you mind if I save that for a later PR?  I'm not entirely sure offhand how to integrate that into this well.

haltman-at: Oh, hm, looks like `AggregateError` is only available on Node 15 and later.  Seems to me like that kind of rules it out for now... (how did you get away with using it in Ganache?)

haltman-at: Oh duh the polyfill you linked.  Silly me.

cds-amal: That is really nice @davidmurdoch

davidmurdoch: Harry, yeah, I don't think it's super important, it's just kinda nice.

PR #4541, opened 2021-12-14
Internal improvement: Up timeout on debugger tests

haltman-at:
Some debugger tests have been timing out a bunch lately.  Getting these when I see them.


cds-amal (approved):
LGTM.

PR #4543, opened 2021-12-15
Internal improvement: Increase timeout on another debugger test

haltman-at:
Found another one of these tests that was timing out, upped it.

(Also, `prettier` is working!  Yay!)


cds-amal (approved):
Looks good to me! Oh boy, we're gonna get a lot of these prettier updates in commits for a while.

PR #4544, opened 2021-12-15
Internal improvement: Allow ES2021 globals in ESLint

haltman-at:
We need this so it doesn't complain about `BigInt` and such!

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -10,7 +10,6 @@
   "parser": "@typescript-eslint/parser",
   "plugins": ["@typescript-eslint"],
   "parserOptions": {
-    "ecmaVersion": 2021,
     "sourceType": "module"
   },
   "ignorePatterns": ["dist"],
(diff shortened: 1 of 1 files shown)


.eslintrc.json
```diff
@@ -10,7 +10,7 @@
   "parser": "@typescript-eslint/parser",
   "plugins": ["@typescript-eslint"],
   "parserOptions": {
-    "ecmaVersion": 9,
+    "ecmaVersion": 2021,
```
cds-amal, line 13: according to the docs, this entry may not be necessary. We could drop it, or set it to `12`.

haltman-at: That's true; I just changed it for consistency, even though it's not necessary.

haltman-at: OK after talkign to @cds-amal more I'm just going to delete this option for simplicity.

cds-amal (approved):
Thanks @haltman-at. let!s get it in!

PR #4545, opened 2021-12-15
Internal improvement: Expand pattern for tests in .eslintrc.json

haltman-at:
I was getting some problems because ESLint didn't recognize some files as test files.  So I expanded the pattern to cover any `ts` or `js` file in a test directory; and I also turned on both `mocha` and `jest` globals.

### DIFF
--- a/.eslintrc.package.json
+++ b/.eslintrc.package.json
@@ -1,9 +1,10 @@
 {
   "overrides": [
     {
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
Thanks Harry. I expect there will be a couple more of these config ones :/

PR #4547, opened 2021-12-16
Internal improvement: Remove debugger's remote dev server

haltman-at:
Nobody uses this anymore, and according to @gnidan (the only one who ever really did use it) it wasn't really practical anyway.  Let's toss it!

(...unless someone wants to keep it...?)

Doing this meant that there were no longer different environments for `store`, so I got rid of that whole mechanism too.  There are still two different environments for webpack, though (production and test).

I made sure to test this manually before PRing it, of course, to ensure I didn't totally break the debugger. :)


gnidan (approved):
burninate!

haltman-at:
Oh hm -- for some reason this PR caused `webpack` to be moved out of `truffle`'s node_modules and into the workspace node_modules.  So I had to adjust the paths used for building Truffle!  I did a quick `truffle init` smoke test w/the bundled version to be sure that hadn't completely screwed everything up, but, uh, that's certainly worth noting.

dongmingh (approved):
looks good to me

PR #4548, opened 2021-12-16
Add null guards to debugger code.

michaeljohnbennett:
PR to encapsulate any fixes needed to get TS style code talking to the debugger better.


packages/debugger/lib/session/index.js
```diff
@@ -250,15 +250,17 @@ export default class Session {
...
-              sources.internal[contextHash][index] = {
-                ...source,
-                compiler: source.compiler || compiler,
-                compilationId: compilation.id,
-                index,
-                id: makeSourceId(compilation.id, contextHash, index),
-                internal: true,
-                internalFor: contextHash
-              };
+              if (source) {
```
haltman-at, line 253: Maybe add a comment on this line like `//VSCode extension breaks w/o this check` or something?  Otherwise this seems like exactly the sort of thing a well-meaning simplifier might remove later.

packages/debugger/lib/session/index.js
```diff
@@ -290,15 +292,17 @@ export default class Session {
...
```
haltman-at, line 295: Same note here.

haltman-at (commented):
Seems good!  My one suggestion is to perhaps add some comments to prevent future mistakes.

haltman-at (approved):
Seems good to me, thanks!

PR #4550, opened 2021-12-16
Dependency update: Update coveralls to 3.1.1

haltman-at:
Just updating this.  We're not really using this at the moment, but per discussion with @gnidan we're not removing it (it would, after all, be quite nice to get code coverage working again!).

...I was kind of hoping this would help resolve some dependabot alerts, but now that I look more closely, it probably actually won't.  Oh well.  Still, seems good on net, right? :)


eggplantzzz (approved):
super

PR #4551, opened 2021-12-16
Dependency update: Update lerna-update-wizard to 1.1.0

haltman-at:
Updating this seems harmless (since it's just a command-line wizard that we have around for convenience), and I'm hoping it might help resolve some dependabot alerts, possibly?


eggplantzzz (approved):
groovy

PR #4552, opened 2021-12-16
Feature request: Add url param to debug command

### LINKED ISSUE #4295: Feature request: set network to a URL via the command line w/o truffle config file
When analyzing transactions and debugging coding problems, I often need to perform "one off" debugging, usually for projects that aren't mine. I find that the Truffle debugger interface is the best around, and Remix is the second best. Currently, I'm debugging assembly and would love to debug a transaction without a config file. In fact, having to create a config file/directory to analyze a single transaction is cumbersome enough that'll I'll choose to use Remix instead. I'd rather do something like this: 

```
$ truffle debug 0xabcd... --network http://localhost:8545
```

It would also be great if the debugger didn't try to compile anything during command. It would be alright if it attempts to compile and finds no code -- though again, there's no config file so the debugger would need to continue like normal. 

Thoughts? 

Bottom line: Creating a Truffle project is tedious and cumbersome when using the tool for analysis purposes. Would love a way to specify config options on the command line.

santanaluiz:
The goal is to be able to debug a project without having to compile or have to add a truffle-config file

Related to #4295


packages/config/test/methods.test.ts
```diff
@@ -51,6 +51,16 @@ describe("when it can't find a config file", () => {
       TruffleConfig.detect();
     }, "should have thrown!");
   });
+
+  it("doesn't throw when url is provided", () => {
+    const options = { url: "http://localhost:7545" };
+    const config = TruffleConfig.detect(options);
+    const networks = config.networks
```
haltman-at, line 58: This is me being really nitpicky, but, missing semicolon here.  Btw, we now have `prettier` and `eslint` on this repo (dunno when you forked it), so if you run `yarn bootstrap` after cloning it to set things up, `prettier` will automatically fix things like this for you. :)

packages/core/lib/commands/debug/meta.js
```diff
@@ -27,6 +27,10 @@ module.exports = {
       describe: "Force debugger to skip compilation (dangerous!)",
       type: "boolean",
       default: false
+    },
+    "url": {
```
haltman-at, line 31: Since you added this to debugger, I'm assuming you intended it as a debugger-specific thing.  If it's intended as a general feature then of course it'll have to go with the other general options.

packages/core/lib/commands/debug/meta.js
```diff
@@ -27,6 +27,10 @@ module.exports = {
...
+      describe: "Forces/set server URL",
```
haltman-at, line 32: I think this description could use some work; also we would normally talk about a "provider" in this context rather than a "server".  Perhaps "Use specified URL for provider"?

packages/core/lib/commands/debug/meta.js
```diff
@@ -59,7 +63,11 @@ module.exports = {
         option: "--compile-none",
         description:
           "Forces the debugger to use artifacts even if it detects a problem.  Dangerous; may cause errors."
-      }
+      },
+      {
```
haltman-at, line 67: Maybe move this one up so that it displays earlier?  It's more useful than a dangerous option like `--compile-none`; it probably shouldn't be all the way at the end.  Maybe it could go after `--fetch-external`?

packages/core/lib/commands/debug/meta.js
```diff
@@ -59,7 +63,11 @@ module.exports = {
...
+        option: "--url",
+        description: "Sets the server URL inline and ignore truffle config file"
```
haltman-at, line 69: Again, I'd reword this description.  I'll suggest something like "Connects to a specified provider given via URL, ignoring networks in config"

haltman-at (changes requested):
Thanks for this!  I think it needs a few changes.  The biggest thing is I'm unsure if `Config.detect` is the right place for the code you added; and using a default config as the base will cause some bugs.  But, this is a good starting point, so thank you for putting this in.

One other comment that doesn't attach to any particular line: I notice you didn't anywhere check that the user didn't set both `--network` and `--url`.  If possible, it might be good to detect that somewhere and error out, because it shouldn't be possible to use both of those simultaneously; they don't make sense together, and we wouldn't want people relying on whatever behavior results from combining them.

packages/config/src/index.ts
```diff
@@ -178,6 +178,21 @@ class TruffleConfig {
   public static detect(options: any = {}, filename?: string): TruffleConfig {
     let configFile;
     const configPath = options.config;
+    const url = options.url;
+
+    if (url) {
```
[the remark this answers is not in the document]
cds-amal: I don't understand the implications of this being a general feature option, but it absolutely makes sense to be scoped to the debugger CLI as that's the use case specified by #4295, which is an absolutely brilliant one.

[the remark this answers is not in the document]
santanaluiz: Makes sense. I agree that adding it to the debugger CLI might make more sense.

What about the possible issues with default configs you just mentioned? I am wondering what can cause problem and if we should combine `--url` always with `--compile-none` for example.
Any thoughts here?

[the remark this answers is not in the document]
haltman-at: I mean I'm pretty the solution here, is, as mentioned, to merge the new `networks` into the existing config, *not* into a default one.  (Although as part of the idea is that this should work without a config file present, I guess it would use a default one if none exists?)  As I also mentioned, this probably doesn't belong in `Config.detect`, but rather in `core/lib/commands/debug`, around the call to `Config.detect`.  (I guess you'd have to catch errors from the `detect` to handle the no-config case?)

> I am wondering what can cause problem

I mean, Solidity versions are one example; I don't know of others *offhand*, but like -- it's far safer to make specific modifications to the existing config, rather than creating a new one that ignores the existing one entirely.

> if we should combine --url always with --compile-none for example.

No, absolutely not.  The `--compile-none` option is explicitly marked as dangerous; it's not something you would ever want turned on by accident just as a result of using `--url`.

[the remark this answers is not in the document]
santanaluiz: I did quite some changes. It seems to look better now.

1. `url` param is scoped for the `debug` command only
2. `compileNone` is set to `true` by default
3. user can override `network` inline if they want
4. added tests to debug command

packages/core/lib/commands/debug/inlineConfigNetwork.js
```diff
@@ -0,0 +1,17 @@
...
+  const Config = require("@truffle/config");
+
+  const configObject = Config.default();
+  configObject.networks = {
+    inline_config: {
+      url: options.url,
+      network_id: "*"
+    }
+  };
+  configObject.network = options.network ? options.network : "inline_config";
```
eggplantzzz, line 11: Why are we doing this? In this flow, network will never be set before reaching this line. Will this not be the case some time in the future?

packages/core/lib/commands/debug/inlineConfigNetwork.js
```diff
@@ -0,0 +1,17 @@
+module.exports = function (options) {
+  const Config = require("@truffle/config");
+
+  const configObject = Config.default();
+  configObject.networks = {
```
eggplantzzz, line 5: This is kind of a thing of principle but instead of overwriting the entire networks property, add to it like
```
configObject.networks["inline_config"] = {...
```

packages/core/lib/commands/debug/inlineConfigNetwork.js
santanaluiz: fair point. I believe we can remove that

packages/core/lib/commands/debug/inlineConfigNetwork.js
santanaluiz: this makes sense, but if we are using config `default()`, unless it is planned to change in the future, it is an empty object, right?

eggplantzzz: You are correct, in this case it will make no functional difference. But in principle it is probably better to just add something to it rather than overwrite the entire thing.

packages/core/lib/commands/debug/run.js
```diff
@@ -1,14 +1,21 @@
...
+  const inlineConfigNetwork = require("./inlineConfigNetwork");
   const debug = debugModule("lib:commands:debug");
 
   const { Environment } = require("@truffle/environment");
   const Config = require("@truffle/config");
 
   const { CLIDebugger } = require("../../debug");
 
-  const config = Config.detect(options);
+  let config = null;
```
eggplantzzz, line 12: Assuming this logic is correct and we want this, a slick way to do this would be the following:
```
const config = options.url ? inlineConfigNetwork(options) : Config.detect(options);
```

packages/core/lib/commands/debug/inlineConfigNetwork.js
haltman-at: You can also use `Config.merge()` rather than writing to it directly if you want to be safe!  But writing to it directly should be fine here.

packages/core/lib/commands/debug/run.js
haltman-at: Also, minor style matter, but if you just want to create the `config` variable, it makes more sense to just write `let config;` rather than `let config = null;`.  Writing `let config = null;` suggests to the reader that the particular value `null` is somehow meaningful, when in fact it will never get used.

packages/core/lib/commands/debug/run.js
```diff
@@ -1,14 +1,21 @@
...
+  if (options.url) {
```
haltman-at, line 13: So, I don't think this is what we want to be doing.  Just because a `url` was supplied, does not mean that an existing config should be wholly ignored.

Rather, like I said before, I think that if `url` is passed and a config exists, we should use it and merge in the new network; and if one doesn't exist, only then should we create a new one from scratch (and set `compileNone` to `true`).

packages/core/lib/commands/debug/inlineConfigNetwork.js
```diff
@@ -0,0 +1,17 @@
...
+    inline_config: {
+      url: options.url,
+      network_id: "*"
+    }
+  };
+  configObject.network = options.network ? options.network : "inline_config";
+  configObject.compileNone = true;
+  configObject._ = [];
```
haltman-at, line 13: Why is this line here?  Is this necessary?

packages/core/lib/commands/debug/meta.js
```diff
@@ -59,6 +63,11 @@ module.exports = {
         option: "--compile-none",
         description:
           "Forces the debugger to use artifacts even if it detects a problem.  Dangerous; may cause errors."
+      },
+      {
+        option: "--url",
+        description:
+          "Connects to a specified provider given via URL, ignoring networks in config. NOTE: compileNone is set to true when url is specified"
```
haltman-at, line 70: I don't think we should be setting `compileNone` to `true` whenever `url` is specified.  I think we should only be doing it if `url` is specified and there's no config.

haltman-at (changes requested):
Thanks!  I still think the logic for determining whether to use an existing config or a new config should be changed, though.

Also, I still think something should be added to `run.js` to check that `network` and `url` weren't both passed in as options.

packages/core/lib/commands/debug/run.js
santanaluiz: makes sense. I misunderstood your previous coment

packages/core/lib/commands/debug/inlineConfigNetwork.js
santanaluiz: removed

packages/core/lib/commands/debug/meta.js
santanaluiz: fixed and updated description

packages/core/lib/commands/debug/meta.js
```diff
@@ -59,6 +63,11 @@ module.exports = {
         option: "--compile-none",
         description:
           "Forces the debugger to use artifacts even if it detects a problem.  Dangerous; may cause errors."
+      },
+      {
+        option: "--url",
+        description:
+          "Connects to a specified provider given via URL, ignoring networks in config. NOTE: compileNone is set to true when a config file is not found AND url is specified"
```
haltman-at, line 70: I think you can leave out the note; this is really a behind-the-scenes thing.  The `compileNone` option isn't dangerous in this context like it usually is.  Although hm... probably it should be made clearer that this allows using the debugger outside of a Truffle project.  I think that part is worth making explicit!  So maybe could just have a second sentence, "Allows using the debugger outside of a Truffle project."?

packages/core/lib/commands/debug/mergeConfigNetwork.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function (config, options) {
+  const inlineConfigNetwork = "inline_config";
+  config.networks[inlineConfigNetwork] = {
+    url: options.url,
+    network_id: "*"
+  };
+  config.network = inlineConfigNetwork;
+  config.merge(options);
```
haltman-at, line 8: So, this works, because `config.merge()` does in fact modify `config` (something I'd forgotten until just now... yikes), but for consistency with our other code I think should actually return the modified config.

packages/core/lib/commands/debug/run.js
```diff
@@ -1,14 +1,19 @@
...
   const { Environment } = require("@truffle/environment");
-  const Config = require("@truffle/config");
-
   const { CLIDebugger } = require("../../debug");
 
-  const config = Config.detect(options);
+  const config = loadConfig(options);
+
+  if (options.url) {
+    mergeConfigNetwork(config, options);
```
haltman-at, line 14: Follow-on to previous comment, for consistency I'd suggest `config = mergeConfigNetwork(config, options);`.

packages/core/lib/commands/debug/loadConfig.js
```diff
@@ -0,0 +1,17 @@
+module.exports = function (options) {
+  const Config = require("@truffle/config");
+
+  let config;
+  try {
+    config = Config.detect(options);
+  } catch (error) {
+    if (options.url) {
```
haltman-at, line 8: This is maybe being nitpicky, I think it would be good here to also check that the error is of the correct type (like, `error instanceof TruffleError`... maybe check the message too if you want to be really sure, although maybe that's overkill) before invoking this handling, and rethrowing otherwise.

packages/core/lib/commands/debug/meta.js
haltman-at: Hm, this was marked resolved, but it appears to be in the same place still?

packages/core/lib/commands/debug/run.js
```diff
@@ -1,14 +1,19 @@
...
+  const loadConfig = require("./loadConfig");
   const debug = debugModule("lib:commands:debug");
 
   const { Environment } = require("@truffle/environment");
-  const Config = require("@truffle/config");
-
   const { CLIDebugger } = require("../../debug");
 
-  const config = Config.detect(options);
+  const config = loadConfig(options);
```
haltman-at, line 11: I still think there should be something prior to this that checks that you didn't set both `--network` and `--url`.

haltman-at (commented):
OK, this is better, thanks again!  A few comments still but they're pretty small.  Some of this is stuff that I commented on earlier, so let me know what you're doing with those... if you intend to leave them alone, you should make a note of that so we all know what's going on.

packages/core/test/lib/commands/debug.js
```diff
@@ -0,0 +1,56 @@
...
+  describe("mergeConfigNetwork(config, options)", () => {
+    beforeEach(() => {
+      config = Config.default();
+      options = {
+        url: "http://urlhost:1234"
+      };
+    });
+
+    it("should create networks item in config", () => {
+      mergeConfigNetwork(config, options);
```
haltman-at, line 37: Oh also, for consistency (see previous comments), this should also be `config = ...`;

packages/core/lib/commands/debug/loadConfig.js
santanaluiz: I wouldn't add the check for the message. If the message is changed in the `Config`, we have a non-functional code here

packages/core/lib/commands/debug/meta.js
santanaluiz: Added there, let me know what you think

packages/core/lib/commands/debug/loadConfig.js
santanaluiz: added the `instanceof` check, but as I said, I don't think a check for the message is a good idea. Other 2 options would be: 
1. having a constant with the message in Config file - which doesnt look like a good idea
2. Having a more specific error `TruffleConfigFileNotFoundError`, which doesn't seem to be the direction the rest of the codebase is going

Anyhow, seems fine now

santanaluiz: added the `instanceof` check, but as I said, I don't think a check for the message is a good idea. Other 2 options would be: 
1. having a constant with the message in Config file - which doesnt look like a good idea
2. Having a more specific error `TruffleConfigFileNotFoundError`, which doesn't seem to be the direction the rest of the codebase is going

Anyhow, seems fine now

packages/core/lib/commands/debug/run.js
santanaluiz: added a check. Not very happy with the copy/wording though. Let me know if you have a better idea :)

packages/core/lib/commands/debug/loadConfig.js
haltman-at: Yeah, that makes sense.  Thanks!

packages/core/lib/commands/debug/meta.js
```diff
@@ -35,6 +39,11 @@ module.exports = {
       OS.EOL +
       "                             [--compile-tests|--compile-all|--compile-none]",
     options: [
+      {
+        option: "--url",
+        description:
+          "Connects to a specified provider given via URL, ignoring networks in config. This option allows using the debugger outside of a Truffle project. That's why compileNone is set to true when a config file is not found AND url is specified"
```
haltman-at, line 45: I think the the final sentence here needs to be cut.  That's implementation details that users don't need to know about and will just confuse them; also, it makes this description just too long.

packages/core/lib/commands/debug/meta.js
```diff
@@ -35,6 +39,11 @@ module.exports = {
       OS.EOL +
       "                             [--compile-tests|--compile-all|--compile-none]",
     options: [
+      {
```
haltman-at, line 42: The order of these controls the order of display, so I'd suggest moving this down... `--url` shouldn't be the very first thing, the transaction hash should be.  Somewhere after this but before the compile options, I think.

haltman-at (commented):
Great, thanks so much!  Just one/two minor final things that you can fix really quickly.

(Now could we get a second reviewer on this so we could get this in? :) )

packages/core/lib/commands/debug/loadConfig.js
```diff
@@ -0,0 +1,18 @@
...
+  const Config = require("@truffle/config");
+  const TruffleError = require("@truffle/error");
+
+  let config;
+  try {
+    config = Config.detect(options);
+  } catch (error) {
+    if (error instanceof TruffleError && options.url) {
+      config = Config.default();
+      config.compileNone = true;
```
cds-amal, line 11: Can a comment be added to document this error resolution/ code path please?

cds-amal (changes requested):
Very nice work @santanaluiz ! I left one minor nit about documenting an exception path

santanaluiz:
@haltman-at @cds-amal thanks for your reviews. I did the last changes required.
I hope it's fine now :)

cds-amal:
Hi @santanaluiz could you update the comments to explain why `compileNone` is set here? It will help future contributors reason about this awesome feature.

santanaluiz:
@cds-amal updated the comment, please, share any ideas if you think of a better explanation copy

cds-amal:
> @cds-amal updated the comment, please, share any ideas if you think of a better explanation copy

@santanaluiz This update is good. Thank you!

cds-amal (approved):
I'm excited for this feature @santanaluiz ! Thanks so much. :bow: Here's my [recorded smoke test](https://asciinema.org/a/rc5p0MCS0C07z7aH9QDN1al6R)

santanaluiz:
really happy to support 🙇 

Thanks for sharing the smoke test result

haltman-at (commented):
Sorry, one final thing!  On actually testing this out, I noticed a bug.  If you pass a URL and there's no config, then `compileNone` is supposed to be set.  In reality, this doesn't happen.  This was confusing, but here's why.

Suppose you're in this case. Then `loadConfig` will set `config.compileNone` to true.  However, since `options.compileNone` is still false, the later call to `mergeConfigNetwork` will set `config.compileNone` back to false.

So, uh, something needs to be altered to fix that.  Sorry about that!  But glad I actually tested this. :)

santanaluiz:
@haltman-at very good catch! I override this on the options, added a comment for clarity and updated tests

packages/core/lib/commands/debug/loadConfig.js
```diff
@@ -0,0 +1,24 @@
...
+  } catch (error) {
+    if (error instanceof TruffleError && options.url) {
+      // in case config file is not detected (exception thrown) AND url is provided in the options,
+      // We use default config and set compileNone to true. Since there are is no config files and url is provided,
+      // It is assumed that truffle debug is being used for analysis and debugging and that there is nothing to compile.
+      // E.g. analysing/debugging a single transaction of an external project
+      config = Config.default();
+      config.compileNone = true;
+      // forcing options to also set compileNone to true, otherwise it can be later override on Config.merge(...)
+      options.compileNone = true;
```
haltman-at, line 17: Oh, hm, you fixed it by mutating the `options` that was passed in... I don't think this is a good idea, it means that `loadConfig` has a hidden side effect.  (I know, some of these functions have the hidden side effect of altering `config`, but that we just have to accept.)  Can you fix it without mutating the `options` that was passed in?  A quick way if you want to keep the code this way would be to clone and have this return `{ config, options }` instead of just `config`, maybe not the best way, but just throwing things out...

santanaluiz: actually, I had done some changes meanwhile.
I agree it doesn't look good.
Also, I am not happing returning `config` and `options`

So I did a couple of changes
1. Remove the `mergeConfigNetwork` from `run.js`. Is there a use-case we want to manually add a provider but we can't set `--network` as an argument?
2. Added `mergeConfigNetwork` here to `loadConfig`. So we merge and `options` to defaul config and then set `compileNone` to `true`

wdyt?

haltman-at: Ah!  Yes, I agree that is a better structure.  However there's no `mergeConfigNetwork` in that first branch, there should be one there too, right?

santanaluiz: What do you mean in the first branch? do you mean on the `try` block? 
Why should we have it there? if `url` is provided and we have a config file, we basically ignore `url` anyway, right?

To be honest it might make sense to log a warning, saying the parameter is ignored. But other than that, I don't see a reason. Did I miss something?

haltman-at: Yes, I mean in the `try` block.  My understanding that the desired behavior was that `url` is not ignored in such case, just the network is overridden.

santanaluiz: I see what you mean. Updating...

santanaluiz: should be fine now. Please, let me know of any further point

haltman-at (approved):
Alright, let's finally get this in! :)  (Assuming nobody else spots any other problems real quick.)

haltman-at (changes requested):
Blargh, sorry, I'm afraid there is another problem after all.  On actually trying this out, I'm hitting a crash in `getCompilations` in `core/lib/debug/cli.js`.  The problem is that `artifacts` is `undefined` (because the build directory doesn't exist), but since `compileNone` is set, we try to apply `shimArtifacts` to it.  Note we can't just always replace undefined `artifacts` by `[]`, because this mean different things.  But we can do it in specific contexts.

So, we could, for instance, have `gatherArtifacts` return `[]` instead of `undefined` if `compileNone` is set; or we could have `getCompilations` specially handle the possibility of `artifacts` being undefined and `compileNone` being set.  Former seems probably easier to me?  But right now it crashes so I'm afraid we can't merge this with that outstanding. :-/

santanaluiz:
Good catch! But I we should check also if config file was skipped. 

Because in the current implementation we can run `$ truffle  debug --url http://host:port --compile-none` 
and it still works in case the user has compiled sources in the `build` directory.

Therefore... 
1. I have added a `skippedConfigFile` field to `config` in this specific use case.
2. Updated tests
3. Added a comment for clarity in there

haltman-at:
Oh, you're absolutely right!  This whole time I was conflating `compileNone` with not checking the artifacts... even though that's not what `compileNone` does.  You're right, that should be separate!  Silly me.  OK, will review this now quickly.

haltman-at (approved):
Great!  Thank you so much!  This all seems to be good now.  Will merge shortly unless anyone has any objections for some reason!

PR #4554, opened 2021-12-17
Internal improvement: Add missing dependency to @truffle/require

haltman-at:
This dependency wasn't listed, thanks to @eggplantzzz for pointing this out!


cds-amal (approved):
Looks good!

PR #4557, opened 2021-12-20
Dependency Update:  graphql-tools from v6-v8

tenthirtyone:
No problems encounter installing or running unit tests against the packages split out from the monorep


packages/db/src/meta/graph/types.ts
```diff
@@ -2,7 +2,7 @@ import { logger } from "@truffle/db/logger";
 const debug = logger("db:graphql:types");
 
 import type * as graphql from "graphql";
-import type { IResolvers } from "graphql-tools";
+import { IResolvers } from "@graphql-tools/utils";
```
gnidan, line 5: In general, we should `import type` wherever possible - normal imports will naturally generate JS code, but if you only import the type, then the JS gets thrown out! (which is nice if you don't need it.)

Also, if you only ever `import type` a package, it typically means you can reduce it to a devDependency, which is also nice.

```suggestion
import type { IResolvers } from "@graphql-tools/utils";
```

packages/db/package.json
```diff
@@ -36,6 +36,9 @@
     "test": "jest --verbose --detectOpenHandles --forceExit"
   },
   "dependencies": {
+    "@graphql-tools/delegate": "^8.4.3",
+    "@graphql-tools/schema": "^8.3.1",
+    "@graphql-tools/utils": "^8.5.5",
```
gnidan, line 41: Looks like this can be moved to `devDependencies` (assuming it's only used with `import type`)

gnidan (changes requested):
Nice! Glad this was easy; thanks for looking!

Couple notes about a TS feature maybe you didn't know about :)

packages/db/src/meta/graph/schema.ts
```diff
@@ -3,7 +3,8 @@ const debug = logger("db:meta:graph:schema");
 
 import gql from "graphql-tag";
 import type * as graphql from "graphql";
-import { makeExecutableSchema, IResolvers } from "graphql-tools";
+import { makeExecutableSchema } from "@graphql-tools/schema";
+import { IResolvers } from "@graphql-tools/utils";
```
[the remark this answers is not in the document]
tenthirtyone: :+1:

packages/db/src/meta/graph/types.ts
tenthirtyone: :+1:

packages/db/package.json
tenthirtyone: :+1:

gnidan:
Just noticed you've committed `package-lock.json`. Could you remove this and make sure you use `yarn`? Don't want to have conflicting lockfiles!

tenthirtyone:
::facepalm::

gnidan (approved):
Cool, looks like this is now good to go!

PR #4560, opened 2021-12-20
Dependency update: Update highlightjs-solidity to 2.0.3

haltman-at:
I updated highlightjs-solidity for Solidity 0.8.11; this PR just updates us to use the new version I just put out.


cds-amal (approved):
Nice @haltman-at . Let's get it in!

Issue #4565, opened 2021-12-22
Debugger crashes when used with Geth versions 1.10.9 and up


haltman-at:
## Issue

When the debugger is used with Geth, on versions 1.10.9 and up, it crashes.  This is due to [this Geth issue](https://github.com/ethereum/go-ethereum/issues/24146); there's no memory in the traces and this causes a crash.

It'll be easy enough to fix the crash, but the question then becomes, what do we actually do without memory?  One possibility would be to compute the memory ourselves, but I don't think that's really reasonable.

Another possibility is to act as if memory is just always empty (so memory variables will always display as 0, or empty string, or empty array, etc).  I think this is a reasonable interim workaround, but it'll be confusing to users.  I think that if we do this, we should also print a warning on debugger startup.  (This will mean that the necessary information, that memory is missing, will have to somehow make its way from the `web3` saga to the debugger CLI... presumably it'd get stored in the state somewhere.)

## Steps to Reproduce

Download Geth (version 1.10.9 or later).  Run `geth --dev --http --http.api eth,net,debug`.  (If you're using my repo of solidity test cases, you may want to instead run `get --dev --http --http.api eth,net,debug --http.port 7545`).  Run a transaction against it, then attempt to debug it.

## Environment

* Ethereum client: Geth, versions 1.10.9 and up (most recent at time of submission: 1.10.13)
* Truffle version (`truffle version`): 5.4.25

kasvtv:
Can confirm this as well for Remix. After many hours yesterday, I found that the Remix debugger does not work when connected through the Web3 Provider to a local geth instance ever since version 1.10.9, including the latest 1.10.13.

To any beginner trying to solve the issue, if you have Golang installed, you can rebuild geth 1.10.8 from source using:
```sh
go get github.com/ethereum/go-ethereum@v1.10.8
cd $GOPATH/pkg/mod/github.com/ethereum/go-ethereum@v1.10.8/cmd/geth
go build -o $GOPATH/bin .
```
Replace `$GOPATH` with `%GOPATH%` if using Windows command line.

Make sure the old installation of geth is removed. Then verify the version by running `geth version`.

haltman-at:
Fix for the immediate problem with memory is now released; however there's still a potential problem with storage that I'll have to write up an issue for later.

PR #4568, opened 2021-12-22
Bug fix: Pass `enableMemory: true` when requesting traces

haltman-at:
Addresses #4565.  It turns out you can pass `enableMemory: true` to get the memory in recent geth versions, so this PR does that.

I didn't do a comprehensive test of other clients (didn't check besu or erigon), but Ganache (both 6 and 7 at least) have no problem with this option being passed.

Note this still leaves the problem of us not getting storage.  I thought the debugger doesn't use that; I forgot that actually it still does in certain cases (namely: when debugging failed contract creations, unless they were performed with `CREATE2`).  So we'll want to get that too, but I figure that can be left for later; first let's fix the big obvious crash.

### DIFF
--- a/packages/debugger/lib/web3/adapter.js
+++ b/packages/debugger/lib/web3/adapter.js
@@ -16,7 +16,7 @@ export default class Web3Adapter {
       {
         jsonrpc: "2.0",
         method: "debug_traceTransaction",
-        params: [txHash, {}],
+        params: [txHash, { enableMemory: true }], //recent geth versions require this option
         id: new Date().getTime()
       }
     );


cds-amal (approved):
Looks good. Nice work sleuthing this change @haltman-at ! Hopefully the official GETH docs will soon reflect changes in [GETH PR 23558](https://github.com/ethereum/go-ethereum/pull/23558)

Lets see how tests come back :)

haltman-at:
Yeah, the tests of course are just testing it with Ganache 6... I also manually tested it with Ganache 7, and, obviously, with Geth itself, both newer versions that have the problem and older versions that don't.  I skipped erigon and besu testing, as mentioned.

PR #4578, opened 2022-01-01
refactor: improve number matching

martinmunillas:
Reviewing the code to look how could i implement typescript migrations, i saw that the number of the migration was being obtained with the parseInt function that returns the first matching number, even though this works i think is not the "correct"  way to do it as it is not declarative enough, so in this PR i did the following:

- Improve number matching by using regex instead of parseInt
- Change order of filters
- Add warning in case some of the matching files doesn't have a number

### DIFF
--- a/packages/migrate/Migration.js
+++ b/packages/migrate/Migration.js
@@ -14,13 +14,21 @@ class Migration {
   constructor(file, reporter, config) {
     this.file = path.resolve(file);
     this.reporter = reporter;
-    this.number = parseInt(path.basename(file));
+    this.number = Migration.getNumber(path.basename(file));
     this.emitter = new Emittery();
     this.isFirst = false;
     this.isLast = false;
     this.dryRun = config.dryRun;
     this.interactive = config.interactive;
     this.config = config || {};
+  }
+
+  static getNumber(fileName) {
+    const match = fileName.match(/\d+/);
+    if (!match.length) {
+      return null;
+    }
+    return Number(match[0]);
   }
 
   // ------------------------------------- Private -------------------------------------------------
@@ -114,11 +122,7 @@ class Migration {
       let artifacts = resolver
         .contracts()
         .map(abstraction => abstraction._json);
-      if (
-        this.config.db &&
-        this.config.db.enabled &&
-        artifacts.length > 0
-      ) {
+      if (this.config.db && this.config.db.enabled && artifacts.length > 0) {
         // currently if Truffle Db fails to load, getTruffleDb returns `null`
         const Db = getTruffleDb();
 
@@ -185,12 +189,8 @@ class Migration {
    * @param  {Object}   options  config and command-line
    */
   async run(options) {
-    const {
-      interfaceAdapter,
-      resolver,
-      context,
-      deployer
-    } = this.prepareForMigrations(options);
+    const { interfaceAdapter, resolver, context, deployer } =
+      this.prepareForMigrations(options);
 
     // Connect reporter to this migration
     if (this.reporter) {
--- a/packages/migrate/Migration.js
+++ b/packages/migrate/Migration.js
@@ -24,6 +24,9 @@ class Migration {
   }
 
   static getNumber(fileName) {
+    if (!fileName || typeof fileName !== "string") {
+      return null;
+    }
     const match = fileName.match(/^\d+/);
     if (!match) {
       return null;
--- a/packages/migrate/Migration.js
+++ b/packages/migrate/Migration.js
@@ -24,8 +24,8 @@ class Migration {
   }
 
   static getNumber(fileName) {
-    const match = fileName.match(/\d+/);
-    if (!match.length) {
+    const match = fileName.match(/^\d+/);
+    if (!match) {
       return null;
     }
     return Number(match[0]);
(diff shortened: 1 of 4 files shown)


packages/migrate/Migration.js
```diff
@@ -23,6 +23,14 @@ class Migration {
     this.config = config || {};
   }
 
+  static getNumber(fileName) {
+    const match = fileName.match(/\d+/);
```
cds-amal, line 27: This regex is not equivalent to parseInt which introduces subtle changes to the migrations behavior. `Regex.match` will match the first group of digits in a string *wherever* they occur while `parseInt` will only match sequential digits at the beginning of a string.

packages/migrate/Migration.js
```diff
@@ -23,6 +23,14 @@ class Migration {
...
+    if (!match.length) {
```
cds-amal, line 28: [String.prototype.match](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match#return_value) returns null if no match is found.

cds-amal (commented):
Thanks for taking a look at this @martinmunillas . I left some initial feedback and will get back to this during the week. I love the idea of giving feedback on unmatched files.

packages/migrate/Migration.js
martinmunillas: oh right, i'll fix it right away

packages/migrate/Migration.js
martinmunillas: right, thanks, i'll fix it

packages/migrate/Migration.js
martinmunillas: done :D

packages/migrate/Migration.js
martinmunillas: done :D

martinmunillas:
i'm not sure why tests didn't pass... it seems like is not related to what i changed, if it is please let me know why is it :D

eggplantzzz:
Those are failures in tests related to migrations so it seems very likely that they are related to what you did. Did you take a look at them yet?

packages/migrate/index.js
```diff
@@ -42,12 +42,21 @@ const Migrate = {
...
           null
       )
+      .filter(file => {
+        const hasNumber = Migration.getNumber(path.basename(file)) !== null;
+        if (!hasNumber) {
+          process.emitWarning(
+            `Migration file ${file} does not have a number at the beggining of the file, \
+            migration file names should match the schema \`\${\\d+}you_migration_file.js\``
+          );
+        }
```
cds-amal, line 57: Neat! Can you add a test for this please?

Also, I realize this was here from before, but there might be a way to combine the two filters `migrations_file_extension_regexp`. (No need to change this, just leaving it here as a note)

cds-amal (changes requested):
Thanks @martinmunillas 

It looks like there's a type error in the current commit when checking for completedMigrations. I have concerns about breaking the current state of migration and deployment and we'll have to figure out how to add tests to cover our bases.  

I love the improved developer experience by adding feedback for ignored files in the migration folder.

packages/migrate/index.js
```diff
@@ -213,7 +222,7 @@ const Migrate = {
     } else {
       completedMigration = 0;
     }
-    return parseInt(completedMigration);
+    return Migration.getNumber(completedMigration);
```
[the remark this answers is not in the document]
martinmunillas: yes my bad, for some reason i thought `completedMigration` was a filename but is not, i'll revert this change

packages/migrate/index.js
martinmunillas: i'll do it

martinmunillas:
is there any discord or something were i could get a little more feedback and help? i can't get all tests running on my local so i have to wait for the github action to see if tests are passing

eggplantzzz:
@martinmunillas can you give me more information about why you can't get tests running locally? How are you trying to run them? I can try to help you troubleshoot that if you'd like.

eggplantzzz:
Why do you jump to the conclusion that it has something to do with m1? What version of npm and node are you using?

martinmunillas:
> Why do you jump to the conclusion that it has something to do with m1? What version of npm and node are you using?

because it says "no native build was found for platform=darwin arch=arm64..."
node: 16.10.0
yarn: 1.22.17

eggplantzzz:
Can you try using Node 14 and let me know if that works? I also cannot run it with Node 16.

martinmunillas:
with node 14.18.3 some unrelated tests keep failing
Screen Shot 2022-01-11 at 15 22 28

eggplantzzz:
Ok so now you probably need to install the Vyper compiler. You can check out the Vyper docs [here](https://vyper.readthedocs.io/en/stable/installing-vyper.html).

martinmunillas:
> Hey @martinmunillas, did you ever get the test suite to run? Do you intend on continuing work on this feature? Just let us know! Thanks!

Hi yesss, sorry i totally forgot about it, this weekend i'll try installing the vyper compiler and will fix the conflicts

eggplantzzz:
Ok cool @martinmunillas! Just keep us posted. Thanks!

martinmunillas:
truffle team, sorry for wasting your time, but i won't keep forward with this pr, i'm just not with the time to set the environment and all, feel free to close or take the pr to you

eggplantzzz:
closing for issue maintenance

PR #4580, opened 2022-01-03
Bug fix: Avoid using url in CLI debugger when no url specified

haltman-at:
Oops, it seems that #4552 broke the debugger when `--url` is *not* specified; looks like nobody checked that.  Basically it tries to use `url` no matter what; the check for that got lost somewhere along the way.  This is a quick fix for that; it adds in a check so that `url` isn't used unless it was actually specified.


cds-amal (commented):
Nice catch @haltman-at ! Can you add a test for this please?

haltman-at:
Oh, good idea.  OK, I've added a test.  (I also changed the tests in that file to not be arrow functions, since it's mocha.)

cds-amal (approved):
Looks good to me!

Issue #4561, opened 2021-12-20
Debugger's `stepOut` function doesn't work well with multiline source ranges


haltman-at:
## Issue

Currently, the debugger's `stepOut` function has a special case where it checks if you're on a multiline source range and if so does `stepOver` instead.  It's not entirely clear what this was intended to accomplish (do you remember, @gnidan?), but it doesn't seem to work as `stepOver` should.

The intent may have been that when you're on a function declaration node or contract declaration node, you step over it rather than into it.  But due to the way `stepOver` works, you end up stepping into it anyway.  Moreover, this also means that if you happen to be on a multiline source range within a function body, you don't exit the function as you should.

So, it's likely that this special case should be removed.  However, removing this special case raises the possibility that issues like #4502 will arise, so care must be taken to ensure that doesn't happen.

Alternatively, perhaps there should be a special case here, and should happen instead is that `stepOver` should be altered to step over in such cases rather than in.  (Actually, probably that latter part should be done *regardless*!)  If so it's possible that such a special case might be better off checking the current node type, rather than if the source range is multiline.

## Environment

* Truffle version (`truffle version`): 5.4.25

haltman-at:
Fixed now.

PR #4586, opened 2022-01-04
Fix debugger stepover function for a multiline source range

dongmingh:
This PR fixed the bug reported in #4561 that debugger stepover
function does not work on a multiline source range within a
function body.

### DIFF
--- a/packages/debugger/lib/controller/sagas/index.js
+++ b/packages/debugger/lib/controller/sagas/index.js
@@ -207,8 +207,9 @@ function* stepOver() {
     // line (which may be in a new file)
     (currentDepth > startingDepth ||
       (currentLocation.source.id === startingLocation.source.id &&
-        currentLocation.sourceRange.lines.start.line ===
-          startingLocation.sourceRange.lines.start.line))
+        startingLocation.isMultiline ^
+          (currentLocation.sourceRange.lines.start.line ===
+            startingLocation.sourceRange.lines.start.line)))
   );
 }
 
@@ -297,15 +298,13 @@ function* continueUntilBreakpoint(action) {
             length === currentLength &&
             (currentSourceId !== previousSourceId ||
               currentStart !== previousStart ||
-              currentLength !== previousLength
-            ) &&
+              currentLength !== previousLength) &&
             //if we started in a user source (& allow internal is off),
             //we need to make sure we've moved from a user-source POV
             (!startedInUserSource ||
               currentSourceId !== lastUserSourceId ||
-              currentStart !== lastUserStart || 
-              currentLength !== lastUserLength
-            )
+              currentStart !== lastUserStart ||
+              currentLength !== lastUserLength)
           );
         }
         //otherwise, we have a line-style breakpoint; we want to stop at the


packages/debugger/lib/controller/sagas/index.js
```diff
@@ -207,8 +207,9 @@ function* stepOver() {
     // line (which may be in a new file)
     (currentDepth > startingDepth ||
       (currentLocation.source.id === startingLocation.source.id &&
-        currentLocation.sourceRange.lines.start.line ===
-          startingLocation.sourceRange.lines.start.line))
+        startingLocation.isMultiline ^
+          (currentLocation.sourceRange.lines.start.line ===
+            startingLocation.sourceRange.lines.start.line)))
```
cds-amal, line 212: whoa! Can you explain this logic @dongmingh? Is there a more readable way to express this?

haltman-at (commented):
Could you explain the logic here?  I'm not clear on how this change you made accomplishes the intended behavior.

packages/debugger/lib/controller/sagas/index.js
dongmingh: 1. this is the condition in a do-while loop

2. there are two cases: a multi line and a single line.  The existing code fails the multi line case.
  - For multi line case:
     - isMultiline is true
     - the starting lines and ending lines are the same at the starting and ending of the source range, so continue stepping until the starting lines are the same
    ```
       startingLocation.isMultiline && 
          (currentLocation.sourceRange.lines.start.line !==
            startingLocation.sourceRange.lines.start.line)
    ```
  - For single line case:
     - isMultiline is false
     - continue stepping until the starting lines are different
    ```
       !startingLocation.isMultiline && 
          (currentLocation.sourceRange.lines.start.line ===
            startingLocation.sourceRange.lines.start.line)
    ```

3. if combine the two logics above into one statement, see below, which is very `long` and hard to read.
   ```
        (startingLocation.isMultiline && 
          (currentLocation.sourceRange.lines.start.line !==
            startingLocation.sourceRange.lines.start.line))  ||
        (!startingLocation.isMultiline && 
          (currentLocation.sourceRange.lines.start.line ===
            startingLocation.sourceRange.lines.start.line))
   ```

4. the statement can be simplified as
   ```
        startingLocation.isMultiline ^
          (currentLocation.sourceRange.lines.start.line ===
            startingLocation.sourceRange.lines.start.line)
   ```

It may help users if adding detail comments here for explanation.

dongmingh:
@haltman-at Here is the data I captured in `stepOver` using StacktraceTest.sol.  Note that run3() calls run2() and run2() calls run1().
The data contents are
location   functionDepth   isMultiline   rangeStartingLine   rangeEndingLine   nodeID   nodeName   nodeType

```
starting:  0 true 3 128 483 StacktraceTest ContractDefinition
current:  0 true 33 38 235 run3 FunctionDefinition
current:  0 false 34 34 224 undefined FunctionCall
current:  0 false 34 34 223 undefined Literal
current:  0 false 34 34 224 undefined FunctionCall
current:  0 false 35 35 228 undefined FunctionCall
current:  0 false 35 35 227 succeed Identifier
current:  0 false 35 35 226 run2 Identifier
current:  0 false 35 35 228 undefined FunctionCall
current:  1 true 39 44 253 run2 FunctionDefinition
current:  1 false 40 40 242 undefined FunctionCall
current:  1 false 40 40 241 undefined Literal
current:  1 false 40 40 242 undefined FunctionCall
current:  1 false 41 41 246 undefined FunctionCall
current:  1 false 41 41 245 succeed Identifier
current:  1 false 41 41 244 run1 Identifier
current:  1 false 41 41 246 undefined FunctionCall
current:  2 true 45 51 263 run1 FunctionDefinition
current:  2 false 47 47 260 undefined FunctionCall
current:  2 false 47 47 259 undefined Literal
current:  2 false 47 47 260 undefined FunctionCall
current:  2 true 45 51 263 run1 FunctionDefinition
current:  1 false 41 41 246 undefined FunctionCall
current:  1 false 42 42 250 undefined FunctionCall
current:  1 false 42 42 249 undefined Literal
current:  1 false 42 42 250 undefined FunctionCall
current:  1 true 39 44 253 run2 FunctionDefinition
current:  0 false 35 35 228 undefined FunctionCall
current:  0 false 36 36 232 undefined FunctionCall
current:  0 false 36 36 231 undefined Literal
current:  0 false 36 36 232 undefined FunctionCall
current:  0 true 33 38 235 run3 FunctionDefinition
current:  0 true 33 38 235 run3 FunctionDefinition
```

For multi line case, the source range beginning line and ending line are the same (for example, both starting and ending line are 45 and 51 respectively for functionDepth=2, and 39 and 44 for functionDepth=1).  Hence continue stepping until the rangeStartingLine are the same.

For single line case, just need to step over to the next source line.

Does this make sense? or I miss something here?

packages/debugger/lib/controller/sagas/index.js
cds-amal: Thanks @dongmingh. The comments preceding the `while` block seem outdated with the new code and if we combining bitwise and logical operators it'd be helpful to be explicit in the description/comment.

cds-amal: @dongmingh `a ^ b` is equivalent to `a !== b` for booleans and avoids the casting and between boolean and number for the bitwise operator.

```
        startingLocation.isMultiline !==
          (currentLocation.sourceRange.lines.start.line ===
            startingLocation.sourceRange.lines.start.line)))
```

dongmingh: very nice, thanks

haltman-at:
OK, I read your explanation on @cds-amal's comment, but I still don't understand how this produces the desired behavior.  So if we start on a multiline source range, we continue stepping until we encounter that same source range again?  I see that this will produce the desired behavior in some circumstances -- such as when starting from a function declaration when entering that function -- but I think this can go very wrong in other cases.

For instance, you can also hit a function declaration node when *returning* from a function; this won't produce the behavior in that case.  Moreover, function declarations are not the only type of multiline statement, and for more general multiline statements, it's entirely possible that you'll never hit that same statement again.  I think with more general multiline statements, I think it's possible to get all sorts of potential behaviors here that we don't want.

In addition, this approach strikes me as being potentially brittle against compiler optimizations.  Now, it's true that we don't guarantee that the debugger will work properly when compiler optimizations are used -- the debugger can get very thrown off by those in a number of ways -- but we still want to try to be robust against them when possible, and I suspect it's possible to do better here.  (Although, you shouldn't focus on this, as there will always be *some* problems with optimizations; you should focus on fixing the other things, and hopefully a bit more robustness will follow automatically!)

I'm not convinced this approach is workable.  Maybe there's some easy fix that I'm missing, but my suspicion is that this may need a rethink in order to produce something that works correctly in every case (or every reasonable case).

haltman-at:
Actually, hold on -- this won't even result in the desired behavior in all cases where you *are* on a function definition while entering that function!  Imagine the function `f` calls into a function `g`, which lives in a different file from `f`.  Because of how you grouped the conditions, attempting to step over `f` in this way from its function definition node, would result in the debugger stopping as soon as you entered `g`.

Of course, that's easy enough to fix, but I don't think that's the fundamental problem here; I think the issues I raised in my previous comment are more fundamental and will not be so easy to fix.

dongmingh:
So, I missed many other cases.  Can you recommend any sol that covering other cases?

haltman-at:
I don't think you really need any specific, unusual test cases here to turn up these problems.  The reason I have that library of test cases is to test out unusual features of Solidity of the EVM, but the problems here will show up in pretty ordinary cases.  I think you can construct such cases yourself; indeed, some such cases (such as hitting the function definition node again at the end of a function) will necessarily have turned up in the transactions you're already running.  I am similarly doubtful that specific test cases are necessary for any problems that will come up with conditionals, loops, and other multiline statements.

If you really want specific examples, then `StacktraceTest` has an example of a multiline statement near the beginning that fits into none of these categories (it's a multi-line array literal), and Nick's magic square example has plenty of loops and conditionals.  But, again, I would encourage you to create your own if you need more; nothing exotic is the problem here.

There are some more specific Solidity features that may also cause some trouble -- modifiers may cause some trouble, as well as the way that base constructors are implemented -- and I can point out test cases for those if you like, but you might want to focus on getting the more basic cases correct first before tackling those.

By the way, I notice that this PR attempts to handle stepping over multiline statements in general, but the original problem was about stepping over functions, not arbitrary multiline statements; obviously, if you can do the latter, that's great, but it's OK to restrict the scope if that allows a more workable solution.

packages/debugger/lib/controller/sagas/index.js
```diff
@@ -202,13 +202,19 @@ function* stepOver() {
...
+    //   start: { line, column },
+    //   end: { line, column }
+    // }
     (currentDepth > startingDepth ||
       (currentLocation.source.id === startingLocation.source.id &&
-        currentLocation.sourceRange.lines.start.line ===
-          startingLocation.sourceRange.lines.start.line))
+        startingLocation.isMultiline &&
+        JSON.stringify(currentLocation.sourceRange.lines) !==
+          JSON.stringify(startingLocation.sourceRange.lines)))
```
haltman-at, line 217: Note: You should avoid comparing objects this way unless you really have to

haltman-at (commented):
This is a summary review because the main review ended up happening in a meeting, more or less, but I should write down the main points:

1. This causes `stepOver` to instead behave as `stepNext` when not on a multiline statement
2. This still has the same problem with if/while, etc
3. There's a problem with change of source
4. End of function actually technically works, but only accidentally; it needs to be commented if you're going to do that

Overall this probably needs a rework.

Also this PR should probably have a test but that's a minor thing.

cds-amal:
> This is a summary review because the main review ended up happening in a meeting, more or less, but I should write down the main points:
> 
> 1. This causes `stepOver` to instead behave as `stepNext` when not on a multiline statement
> 2. This still has the same problem with if/while, etc
> 3. There's a problem with change of source
> 4. End of function actually technically works, but only accidentally; it needs to be commented if you're going to do that
> 
> Overall this probably needs a rework.
> 
> Also this PR should probably have a test but that's a minor thing.

Thanks for including this!

dongmingh:
@haltman-at concerning your comments, please check the following if they make sense. Thanks!

the first three points can be fixed by bringing back the original logic:
```
    either: function depth is greater than starting (ignore function calls)
    or, if we're at the same depth, keep stepping until we're on a new line (which may be in a new file)
```
and add the new logic for function:
```
    if FunctionDefinition, keep stepping until we're back to the same line of the same file
```

For point 4 regarding the end of function, 
my understanding is that when we're at the end of a function where the nodeType is FunctionDefinition (I change to use nodeType instead of multiline), the function depth will decrease after leaving the function, hence the function depth will be less than starting and the stepping stops because the logic requires `currentDepth >= startingDepth`.

haltman-at:
> the first three points can be fixed by bringing back the original logic:

Yes, I think the addition about `isMultline` should be removed.  Note that the problem here is logically unrelated to anything being multiple lines; a multiline statement need not be a function definition, and a function definition can be contained on a single line.

> if FunctionDefinition, keep stepping until we're back to the same line of the same file

I think there are two problems with this suggestion.

The smaller problem is that I would think we would want to step until we are *past* the end of the function, not *at* the end of the function.  But that I think would be easily remedied.

The larger problem is that I think "step until we are back at the same point" is fundamentally the wrong approach.  Are you sure that will reliably get you to the end of the function?  If you think you can guarantee that, then maybe I'm wrong and this is approach is workable, but I suspect that this just isn't going to be something you'll want to rely on.  Note that this is also going to be unlikely to be robust against optimization.  Now, it's true that lots of things in the debugger aren't robust against optimization, but we try to be when we can; and I think it's probably possible to do this in a way that's more robust than that.

haltman-at:
Btw, thanks for actually spelling out the reasoning here... makes this much easier to evaluate. :)

As for my suggestion for how to approach this in a way that would be more robust... can you do this by looking at the function depth, instead of whether you've returned to the same point?  We know there are some irregularities with function depth at the beginning of a function, but in my experience it's always been reliable at the end (possibly with some complications with external function calls, which you may want to test out).  If you can find a way to handle the irregularities that occur at the beginning, you should be able to rely on the end (I think).  I think that would be more robust against optimization.

dongmingh:
This is very helpful. I will continue to dig into this. Thanks so much.

haltman-at:
Cool, let me know if you run into any obstacles there!

dongmingh:
@haltman-at, I tested both approaches, end of function vs function depth, with both StacktraceTest and inner-outer and both reach the end of function.  However, with consideration of robustness and optimization, I will use function depth.  I can submit PR for review if you agree with this.

haltman-at:
If you think you've got it right, you shouldn't be asking me, you should just be submitting it.  I mean, unless there are particular details you're uncertain of, or tradeoffs you're not sure I'll be OK with; then it might make sense to ask, but I'd suggest asking about that sort of thing on Slack rather than here for a quicker response.

That said, your notes do make me a bit suspicious that this might *not* be ready yet.  Sure, you used the function depth to detect the end... but how are you handling the irregularities that can occur at the beginning?  Are you sure that your implementation can handle those?  If you're sure it can, then go ahead and submit it; but if you're not, then this may need more time in the oven.  (And if you don't know what irregularities I'm talking about, then we should discuss this more!)

(Also, reminder that now that you've changed this PR's functionality, you should also retitle it.)

dongmingh:
Discard this and will open another one.

Issue #4595, opened 2022-01-06
Debugger mixes together mapping keys from different failed contract creations


haltman-at:
## Issue

I just realized we don't have an open issue for this (even though this has been a thing for a while)... I don't think it's particularly high-priority, but I figure it ought to at least be listed somewhere.

Suppose you have a transaction with multiple failed contract creations, each of which was initiated was the `CREATE` opcode (as opposed to `CREATE2`, and as opposed to being the initial transaction).  Then the debugger will not distinguish their mapping keys from one another; any mapping keys in one will also be applied to all the others.

## Steps to Reproduce

Use the `Tester` contract from the `creation-failure` test case in `solidity-test-cases`.  Run `mixup(2)`; be sure to give it a lot of gas, as otherwise it'll run out.

## Expected Behavior

In the inner most constructor invocation before the result, the value of `mixup` ought to be displayed as `Map { 0 => 81 }`.

## Actual Results

Instead, it's displayed as `Map { 0 => 81, 1 => 0, 2 => 0}`, the mapping keys from other failed attempts to create this contract having been inappropriately applied to this one.

## Environment

* Truffle version (`truffle version`): 5.4.26

Issue #4583, opened 2022-01-03
Debugger errors when doing `p sto` or `v` within certain failed contract creations on Geth 1.10.5 and up


haltman-at:
## Issue

Suppose you are debugging a transaction, and that transaction creates a contract with the `CREATE` opcode (as opposed to `CREATE2`, or the whole transaction being a contract creation), and that contract creation reverts.

Step into that contract creation, and type `p sto`, or `v`.  If you are using Geth 1.10.5 or later, the debugger will error.

This is because, as of 1.10.5, Geth no longer includes `storage` in its traces.  The debugger mostly doesn't use `storage` anymore... but when inside a reverted contract creation that was initiated with `CREATE`, it still does.  When it doesn't find it there, it crashes.

This is follow-on work to #4565, which dealt with Geth no longer supplying memory.  The good news is that unlike with memory, we only use `storage` in this particular case, so the error isn't very widespread.  The bad news is I can't find any way to turn storage back on.

The good news is that there is a way we can eliminate our reliance on this remaining use of `storage`; I guess that's just what we'll have to do!

## Steps to Reproduce

See above.  If you want a specific example, you can use `NestedException.run()` from the `creation-failure` from the `solidity-test-cases` repo.

## Expected Behavior

Debugger should correctly report storage (for `p sto`) and variables (for `v`).

## Actual Results

Debugger errors.

## Environment

* Ethereum client: Geth 1.10.5 (or later)
* Truffle version (`truffle version`): 5.4.26

leeftk:
This has been fixed in v5.4.27 release!

Issue #2894, opened 2020-03-17
Add system for debugger to track storage not accessed in transaction


haltman-at:
## Issue

Currently, the debugger can only see the contents of storage slots that have been accessed so far during the transaction being debugged; similarly it can only see mapping keys that have been accessed during the transaction being debugged.  This problem came up recently in for instance #2889.

The idea is to allow one to load a transaction in "reconstruct mode", in which all relevant storage will be reconstructed beforehand, so that storage variables will have accurate values even if they haven't been accessed so far during the transaction.

Also, reconstruct mode should allow manual watching of mapping keys.  However, note that in order for this to work, one would necessarily have to restart the transaction for watching a mapping key to take effect.

haltman-at:
Also, in addition to storage, this mode would track nonces.  And, of course, bytecode.

Later versions would probably also add balance tracking, but storage, bytecode, and nonce should suffice for an initial version; balance can come later.

haltman-at:
Oh, and the self-destruct list.  That would need to go in the initial version too (as part of tracking bytecode).

haltman-at:
The idea here has now changed -- instead of a separate mode, this will be the debugger's default operation, and it will be based on `debug_storageRangeAt` and `debug_accountAt`.  (@gnidan this seems worth re-prioritizing right?)

haltman-at:
Be aware though: The behavior of nonces changed back in Spurious Dragon!  Before Spurious Dragon, contracts started at 0; now they start at 1.

cliffoo:
[pull#5286](https://github.com/trufflesuite/truffle/pull/5286) is merged and made it to the [latest release v5.5.22](https://github.com/trufflesuite/truffle/releases/tag/v5.5.22). Not closing since feature is added to library, but not yet the CLI.

PR #4596, opened 2022-01-06
Bug fix: Remove remaining debugger reliance on traced storage

haltman-at:
Addresses #4583.

This PR eliminates the debugger's remaining uses of `storage` taken from the trace.  I had already mostly eliminated the use of `storage` in #3102; this gets rid of the remaining uses.

The debugger mostly doesn't use the `storage` field, because it uses the codex to determine storage, and the codex is built from the stack.  However, there's one case where the debugger bypasses the codex when determining storage: When inside a failed contract creation that was initiated with the `CREATE` opcode, the debugger would bypass the codex and read the `storage` field directly.

The reason for this is that, in order to meaningfully use the codex, we need to know the current address.  However, currently, the debugger has no way to determine the current address when inside a failed contract creation that was initiated with the `CREATE` opcode.  This will be remedied eventually (see issue #2894), but for now we're stuck with that limitation.  So, currently, in such cases, the debugger treats the address as zero.

As such, the debugger would bypass the codex when the current storage address was equal to zero; and it would not update the codex with a zero account.  But now, we can't bypass the codex anymore.  So how do we handle this?

Well, firstly, we now allow the codex to contain a zero account; in fact, we start it off with a zero account.  But the zero account isn't like other accounts; it gets special treatment.  It's treated as transient, basically; the zero account is always specific to the stackframe it's in.  That is to say, unlike with other accounts, the zero account of each stackframe is prevented from affecting the zero accounts of other stackframes.  Thus, each failed contract creation gets its own storage; if there are multiple on the stack at once, then they're in different stackframes, and their storages don't affect one another.

To accomplish this, whenever a new stackframe is added to the codex, its zero account is wiped.  Note that we don't make any changes to the code handling what happens when a stackframe returns; when a stackframe using the zero account returns, it necessarily does so by *failing*, so its changes already don't get saved, without further intervention needed.  So the wipe on stackframe creation is sufficient to achieve the desired effect that each zero account is isolated to the specific stackframe it's in.

Also, while I've removed the guard preventing `SSTORE`s from affecting the zero account, I've left in place the guard preventing `SLOAD`s from affecting the zero account.  This is for two reasons.  Firstly, the behavior of an `SLOAD` on the codex when a value was not already present is to update *all* stackframes, not just the top one (because if a value was not already present, then the read value is preexisting and thus exists in all current stackframes); obviously, for the zero account, which is supposed to be isolated to a specific stackframe, we don't want that.  But also, there's simply no *need* to account for such `SLOAD`s, because if we're inside a contract creation, an `SLOAD` that's loading a value not already present *must* be loading zero!  So why bother, given that we treat unknown storage as zero anyway?

(I mean, notionally I could have such `SLOAD`s affect the zero account, just I'd do it specially and have it only affect the current stack frame instead of all of them... but I figured this was simpler, if arguably slightly less correct.)

Note that this PR does *not* do anything to address #4595... I tried to do something about that and then I realized it was harder than I thought, so I left it alone; I don't think there's any particular urgency on that one.

As a side benefit, eliminating the reliance on storage means we can now pass the `disableStorage` flag when requesting traces; this ought to reduce the size of the traces we get back, at least with some clients (I don't know if they all respect it).

This PR also adds a test of decoding storage variables while inside a failed contract creation.  However, I must warn you that this PR was mostly tested manually.  A proper test of it requires multiple nested contract creations... we've historically avoided such tests in the debugger due to how slow they can be, so I've similarly avoided it here.  Also, obviously a key part of this PR is that it makes the debugger work with recent versions of Geth, but the debugger tests use Ganache, and I left that as-is; again, I tested that manually.  Possibly we should add some Geth debugger tests, but I'd prefer to save that for a different PR.

Anyway, with this, the debugger no longer relies on the `storage` field at all, and its Geth-compatibility problems should now be solved!

### DIFF
--- a/packages/debugger/lib/evm/reducers.js
+++ b/packages/debugger/lib/evm/reducers.js
@@ -277,9 +277,11 @@ function codex(state = DEFAULT_CODEX, action) {
   const safePop = array => (array.length > 2 ? array.slice(0, -1) : array);
 
   //later: may add "force" parameter
+  //note: we don't need to wipe zero account when saving, because we'll never
+  //attempt to save the zero account in the first place
   const safeSave = array =>
     array.length > 2
-      ? array.slice(0, -2).concat([wipeZeroAccount(array[array.length - 1])])
+      ? array.slice(0, -2).concat([array[array.length - 1]])
       : array;
 
   switch (action.type) {
--- a/packages/debugger/lib/evm/reducers.js
+++ b/packages/debugger/lib/evm/reducers.js
@@ -149,13 +149,8 @@ const DEFAULT_AFFECTED_INSTANCES = { byAddress: {} };
 function affectedInstances(state = DEFAULT_AFFECTED_INSTANCES, action) {
   switch (action.type) {
     case actions.ADD_AFFECTED_INSTANCE:
-      const {
-        address,
-        binary,
-        context,
-        creationBinary,
-        creationContext
-      } = action;
+      const { address, binary, context, creationBinary, creationContext } =
+        action;
       return {
         byAddress: {
           ...state.byAddress,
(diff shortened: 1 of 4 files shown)


packages/debugger/lib/evm/reducers.js
```diff
@@ -332,7 +345,13 @@ function codex(state = DEFAULT_CODEX, action) {
...
       const { address, slot, value } = action;
       if (address === Codec.Evm.Utils.ZERO_ADDRESS) {
-        //as always, we do not maintain a zero page
+        //even though we now have a zero page, we still don't allow SLOADs to
+        //affect it.  firstly, because there will never be preexsting data on
+        //the zero page (it's only used for contract creations), so any SLOAD
+        //should only ever be of data that we already know (or that is zero).
+        //secondly, because the zero page represents something that is specific
+        //to a single stackframe, we definitely do *not* want to update every
+        //stackframe with its storage!
```
cds-amal, line 354: thanks for this explanation!

cds-amal (approved):
Looks good!

Issue #4559, opened 2021-12-20
Truffle displays an incorrect cost deploying to Arbitrum Rinkeby


kevinweaver:
- [x] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue. (Ticket 617473)

---------------------------

## Issue

When deploying a contract to Arbitrum, truffle displays a total cost that is different from the actual cost.

## Steps to Reproduce

`truffle unbox arbitrum`
*in .env, set an infura key and a mnemonic* with Arbitrum Rinkeby ETH accessible [here](https://faucets.chain.link/arbitrum-rinkeby)
`npm run compile:arbitrum`
`npm run migrate:arbitrum --network=arbitrum_testnet`

You will see something like:
```
   Deploying 'SimpleStorage'
   -------------------------
   > transaction hash:    0xb3142a02eb78cc5f6f682100bc9a13cdc94c9cb3b45ddcae85739d3416266caf
      ...
   > total cost:          0.000013137251486502 ETH
```
Take this transaction ID to https://testnet.arbiscan.io.

## Expected Behavior

The cost that truffle reports should be the same as the cost arbiscan reports.

## Actual Results

truffle total cost: 0.000013137251486502 ETH
[arbiscan](https://testnet.arbiscan.io/tx/0xb3142a02eb78cc5f6f682100bc9a13cdc94c9cb3b45ddcae85739d3416266caf) total cost: 0.00001050980184 ETH

The support ticket example showed an even larger difference using a much larger contract:

truffle total cost: 2.69049662 ETH
[arbiscan](https://testnet.arbiscan.io/tx/0x016a9a31d47447104c1ef71b1f9ebe06ac4fa45c6e0dbd21cd0e6a192ca4de66) total cost:  0.00271965395244 ETH.

## Note
I confirmed that this is specific to Arbitrum and not other L2s like Optimism

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`):  v5.4.16
* node version (`node --version`): c14.18.0
* npm version (`npm --version`):  8.3.0

kevinweaver:
Ah! It looks like we aren't [accounting](https://github.com/trufflesuite/truffle/blob/a6418b7347139eaf2fa4ae0328a824a98d5434be/packages/interface-adapter/lib/adapter/web3/index.ts#L82) for the fact that Arbitrum fees are surfaced differently than mainnet fees:

Aribtrum receipt:
[...]

PR #4600, opened 2022-01-06
Use effectiveGasPrice and fall back to gasPrice when missing.

kevinweaver:
## Problem
In, #4559 we noticed that truffle was displaying incorrect transaction fees when using Arbitrum. This has to do with [complexity](https://developer.offchainlabs.com/docs/differences_overview#transaction-receipts) that exists for their L2 that makes `transaction.gasPrice` unreliable compared to `receipt.effectiveGasPrice`. However, this isn't a problem that we need Arbitrum-specific code to support because `gasPrice` has been officially [deprecated](https://github.com/ethereum/execution-specs/pull/251) in favor of `effectiveGasPrice` post-EIP-1559 anyway.

## Solution
When reporting transactions fees, set the gas price to `receipt.effectiveGasFee` if it exists, otherwise `transaction.gasPrice`

## To do
Check that this works on different versions of ganache, EIP-1559 & legacy transactions, Arbitrum & Ethereum.

### DIFF
--- a/packages/interface-adapter/lib/adapter/web3/index.ts
+++ b/packages/interface-adapter/lib/adapter/web3/index.ts
@@ -69,7 +69,9 @@ export class Web3InterfaceAdapter implements InterfaceAdapter {
     return this.web3.eth.getBlockNumber();
   }
 
-  public async getTransactionCostReport(receipt: TransactionReceipt): Promise<TransactionCostReport> {
+  public async getTransactionCostReport(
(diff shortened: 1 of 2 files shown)


fainashalts:
Hi @kevinweaver! Asking you as an open source contributor at large :) Do you know what happened to this PR? Should it be closed, or did this get resolved with the new version of Arbitrum?

Issue #4566, opened 2021-12-22
Debugger prints error message when using `p` in self-destruct tx


haltman-at:
## Issue

Suppose you're trying to debug a transaction where a contract successfully self-destructs... well, at the moment, for the most part, you can't, but that's a separate issue (and we're not going to fix that until Ganache 7.1 is out and we've done something about #2894).

But, it's not *completely* true that you can't.  You can still step through with `;` and see *some* of the assembly state, even if it's not all there.  However, now when stepping through with `;` or `p`, you'll get an error message.

We can't at the moment actually print out all the helpful information you would get normally -- that'll have to wait -- but we could at least do something better than erroring.

## Steps to Reproduce

Create a contract that can self-destruct.  Perform a transaction that makes it self-destruct.  Debug that transaction.  Type `p`.

## Expected Behavior

Not an error message.

## Actual Results

```
debug(development:0x912b205d...)> p

  No data on stack.

Instructions:
Uncaught TypeError: Cannot read property 'toString' of undefined
    at ReadStream.emit (events.js:314:20)
    at ReadStream.onkeypress (readline.js:205:10)
    at REPLServer.self._ttyWrite (repl.js:867:9)
    at REPLServer.Interface._ttyWrite (readline.js:1003:14)
    at REPLServer.Interface._line (readline.js:658:8)
    at REPLServer.Interface._onLine (readline.js:329:10)
    at REPLServer.EventEmitter.emit (domain.js:483:12)
    at REPLServer.emit (events.js:314:20)
    at REPLServer.onLine (repl.js:777:10)
    at REPLServer.runBound [as eval] (domain.js:440:12)
    at bound (domain.js:427:14)
    at REPLServer.bound interpreterCallbackified (util.js:210:5)
    at DebugInterpreter.interpreter (/home/sniffnoy/truffle/truffle/packages/core/lib/debug/interpreter.js:617:26)
    at DebugPrinter.printInstruction (/home/sniffnoy/truffle/truffle/packages/core/lib/debug/printer.js:247:39)
    at Object.formatCurrentInstruction (/home/sniffnoy/truffle/truffle/packages/debug-utils/index.js:535:21)
    at Object.formatPC (/home/sniffnoy/truffle/truffle/packages/debug-utils/index.js:549:18)
```

## Environment

* Truffle version (`truffle version`): 5.4.25

haltman-at:
[...]

PR #4602, opened 2022-01-06
Debugger print a warning message if it does not have code of the contract

dongmingh:
The debugger prints error message when it does not have the code of
the contract, see issue #4566. The issue is raised in self-destruc transactions.
This PR prints a warning message instead.

### DIFF
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -217,25 +217,6 @@ class DebugPrinter {
     }
 
     this.config.logger.log("Instructions:");
-
-    // printout instructions
-    const previousInstructions = this.instructionLines.beforeLines;
-    const upcomingInstructions = this.instructionLines.afterLines;
-    const currentIndex = instruction.index;
-
-    // add an ellipse if there exist additional instructions before
-    if (currentIndex - previousInstructions > 0) {
-      this.config.logger.log("...");
-    }
-    // printout 3 previous instructions
-    for (
-      let i = Math.max(currentIndex - previousInstructions, 0);
-      i < currentIndex;
-      i++
-    ) {
-      this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
-    }
-
     if (!instruction || instruction.pc === undefined) {
       // printout warning message if the debugger does not have the code for this contract
       this.config.logger.log(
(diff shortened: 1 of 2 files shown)


packages/debug-utils/index.js
```diff
@@ -532,16 +535,19 @@ var DebugUtils = {
   },
 
   formatCurrentInstruction: function (instruction) {
+    // return warning message if the debugger does not have the code for this contract
+    if (!instruction || !instruction.pc) {
```
haltman-at, line 539: The value `instruction.pc` is a number; it may be equal to 0.  You shouldn't use `!value` to test for the absence of a value when that value is a number, because this will treat numbers with a value of zero as if they are undefined.

With this code, if you type `p` at the beginning of a transaction, you will see this warning, regardless of whether it is appropriate.

packages/debug-utils/index.js
```diff
@@ -532,16 +535,19 @@ var DebugUtils = {
...
+      return "WARNING: The debugger does not have the code for this contract.";
```
haltman-at, line 540: Minor note: In other places that the debugger prints out warnings they get bolded, so you might want to apply some bold formatting here (`debug-utils` imports `chalk` so you can use that).

haltman-at (dismissed):
One little bug that needs fixing but otherwise seems good.

(Edit: Actually, see additional comment about whether this code is in the right place.)

packages/debug-utils/index.js
haltman-at: Actually, sorry, one other comment: Is this really the appropriate place for this check?  This is where the error occurs, yes; but, like, putting the check here is essentially making the assumption that nothing *else* in instruction printing will error.  It *doesn't*, but that's sort of, like, good luck, right?  I don't know if it's a good idea to rely on that; wouldn't it be better to perform the check before any of this stuff is printed?

dongmingh: I put the check in index.js since this is the final place so that we only need to check in one place.
Another consideration is to check `instruction` and `instruction.pc` before calling formatCurrentInstruction() and  prints warning message.  Then shall the gas information be shown?
I am OK with either.

haltman-at: My point isn't about whether it's called inside `formatCurrentInstruction` or outside `formatCurrentInstruction`, but rather `formatCurrentInstruction` is the correct code to guard in the first place.  Like, my suspicion is that the correct code to guard is the entire segment that prints out the instructions; this currently has no name but it's lines 228-261 in `printInstruction`.  It's true that `formatCurrentInstruction` is the only part in there that currently errors, I'm just not sure it's a good idea to rely on the rest not erroring.

Like, the code structure should ideally reflect the overall logical structure, right?  So if we have three operations `doA()`, `doB()`, and `doC()`, and these operations only make sense when `stuffIsValid` is true, then even if technically `doB()` is the only one that actually *errors* otherwise, surely it makes more sense to do `if (stuffIsValid) { doA(); doB(); doC(); }` rather than `do A(); if (stuffIsValid) do B(); doC(); }`, right?  Because the latter is doing things that are meaningless, even they happen to work out OK.  That's how I see it, at least.

dongmingh: Make sense, and calling `formatCurrentInstruction` is the first place to guard.  I look the entire `printInstruction` and do not see any thing else could cause error.  I will move the guard to printer.js.

packages/core/lib/debug/printer.js
```diff
@@ -243,13 +236,23 @@ class DebugPrinter {
       this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
     }
 
-    // printout current instruction
-    this.config.logger.log(DebugUtils.formatCurrentInstruction(instruction));
+    if (!instruction || instruction.pc === undefined) {
+      // printout warning message if the debugger does not have the code for this contract
+      this.config.logger.log(
+        `${colors.bold(
```
haltman-at, line 242: Aha, I see you realized you have to use `colors` here. :)

packages/core/lib/debug/printer.js
```diff
@@ -243,13 +236,23 @@ class DebugPrinter {
       this.config.logger.log(DebugUtils.formatInstruction(instructions[i]));
     }
 
-    // printout current instruction
-    this.config.logger.log(DebugUtils.formatCurrentInstruction(instruction));
+    if (!instruction || instruction.pc === undefined) {
```
haltman-at, line 239: ? The guard is still placed just around `formatCurrentInstruction`.  I'm confused; do you intend to make the change I suggested or not?  Again, I wasn't talking about putting it just around `formatCurrentInstruction`, but about the whole instruction-printing segment.

haltman-at (commented):
OK, this works, but there's still the issue of just where the guard should be placed.  I'm confused here because it seems like you agreed with me regarding where it should be placed, but then didn't place it there.

packages/core/lib/debug/printer.js
dongmingh: Will place the guard around the entire instruction printing segment

haltman-at (approved):
OK, I think this makes more sense, thanks!

dongmingh:
Thanks

PR #4603, opened 2022-01-07
Enhancement: Allow `ContractInstanceDecoder` to decode transactions sent to that instance even if no bytecode was supplied

haltman-at:
Oddly, it looks like we forgot to ever open an issue for this.  Oops.  Oh well.  (Also, sorry about all the changes caused by `prettier` and ESLint.)

(**Edit**: To make this more reviewable, let me point out here where the real changes are:
1. `allocate/index.ts` -- only the added export on line 61 is real, everything else is `prettier`... sorry (OK, and lines 1259-1267 were deleted due to ESLint noticing they were never used)
2. `decoders.ts` -- nearly all the changes are real, only a tiny bit is `prettier`
3. `downgrade-test.js` -- the added test is real, the rest is `prettier`
)

Anyway, this PR allows the `ContractInstanceDecoder` to decode transactions sent to that particular contract instance, even if no bytecode was supplied for that contract in the project information.  It also adds a test of this functionality.

Most of the pieces of this PR were actually already in place, they just needed to be hooked up.  Specifically:

1. The `additionalContexts` system in the `ContractInstanceDecoder` already allowed said decoder to recognize the address of its instance as belonging to its instance, and
2. The `noBytecodeAllocations` system in the `ContractDecoder` already set up backup allocations to use if there's no bytecode available.

The problem, then, is to connect these things together.  Suppose that `instanceDecoder.decodeTransaction()` is called, and no bytecode was ever supplied for the instance.  The `additionalContexts` mechanism, which was already in place, allows the decoder to correctly get a context for this instance anyway.  The problem is that this context has no calldata allocations associated to it.

But, we know what calldata allocations to use for it: `noBytecodeAllocations`!  So... what if we just patched the `noBytecodeAllocations`, into the calldata allocations, in a place where `@truffle/codec` can see it and make use of it?  That's basically all we do here.

There are some guards on this to prevent it from activating inappropriately and potentially messing things up.  First off, we only patch the allocations like this if the context we found is coming from the `additionalContexts`, and not the main contexts.  The main contexts have their allocations already; they don't need this.  Only the `additionalContexts` never had their allocations set up.  Secondly, we only patch in these additional allocations if there aren't already allocations for that context hash; we allow the pre-existing ones to override the new ones.  (These two guards are possibly redundant, but, oh well... that doesn't seem like a bad thing, really.)

But, um, that's mostly it; it was basically just hooking up components that already existed.  Hopefully the many changes due to `prettier` don't obscure the core of it too much...

### DIFF
--- a/packages/decoder/lib/decoders.ts
+++ b/packages/decoder/lib/decoders.ts
@@ -194,7 +194,12 @@ export class ProjectDecoder {
    */
   public async decodeTransactionWithAdditionalContexts(
     transaction: DecoderTypes.Transaction,
-    additionalContexts: Contexts.Contexts = {}
+    additionalContexts: Contexts.Contexts = {},
+    additionalAllocations?: {
+      [
+        selector: string
+      ]: AbiData.Allocate.FunctionCalldataAndReturndataAllocation;
+    }
   ): Promise<CalldataDecoding> {
     const block = transaction.blockNumber;
     const blockNumber = await this.regularizeBlock(block);
@@ -206,6 +211,24 @@ export class ProjectDecoder {
       additionalContexts
     );
 
+    let allocations = this.allocations;
+    if (context && !(context.context in this.contexts)) {
+      //if the context comes from additionalContexts,
+      //we'll add the additional allocations to the allocations;
+      //however, we'll allow other allocations to override it...
+      //it's only supposed to be used if necessary!
+      allocations = {
+        ...this.allocations,
+        calldata: {
+          ...this.allocations.calldata,
+          functionAllocations: {
+            [context.context]: additionalAllocations,
+            ...this.allocations.calldata.functionAllocations
+          }
+        }
+      };
+    }
+
     const data = Conversion.toBytes(transaction.input);
     const info: Evm.EvmInfo = {
       state: {
@@ -213,7 +236,7 @@ export class ProjectDecoder {
         calldata: data
       },
       userDefinedTypes: this.userDefinedTypes,
-      allocations: this.allocations,
+      allocations,
       contexts: { ...this.deployedContexts, ...additionalContexts },
       currentContext: context
     };
@@ -655,7 +678,9 @@ export class ContractDecoder {
 
   private allocations: Codec.Evm.AllocationInfo;
   private noBytecodeAllocations: {
-    [selector: string]: AbiData.Allocate.CalldataAndReturndataAllocation;
+    [
+      selector: string
+    ]: AbiData.Allocate.FunctionCalldataAndReturndataAllocation;
   };
   private userDefinedTypes: Format.Types.TypesById;
   private stateVariableReferences: Storage.Allocate.StateVariableAllocation[];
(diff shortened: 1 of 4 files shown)


packages/decoder/test/current/test/downgrade-test.js
```diff
@@ -290,6 +290,36 @@ describe("Graceful degradation when information is missing", function () {
...
   });
 
+  it("Decodes transactions sent to the instance via additionalContexts", async function () {
+    let mangledCompilations = clonedeep(compilations);
+    let downgradeTest = mangledCompilations[0].contracts.find(
+      contract => contract.contractName === "DowngradeTest"
+    );
+    downgradeTest.deployedBytecode = undefined;
+
+    let deployedContract = await abstractions.DowngradeTest.new();
```
eggplantzzz, line 300: Why does `deployedContract` not have a `deployedBytecode` field after deployment? And what is the point of setting `downgradeTest.deployedBytecode` to `undefined` because `downgradeTest`? It is not used in this test after that assignment.

packages/decoder/lib/decoders.ts
```diff
@@ -206,14 +211,32 @@ export class ProjectDecoder {
...
+        ...this.allocations,
+        calldata: {
+          ...this.allocations.calldata,
+          functionAllocations: {
+            [context.context]: additionalAllocations,
+            ...this.allocations.calldata.functionAllocations
+          }
+        }
+      };
+    }
```
eggplantzzz, line 230: Can the above assignment be done with
```
allocations.calldata.functionAllocations = {
  [context.context]: additionalAllocations,
  ...this.allocations.calldata.functionAllocations
};
```
Is that the same? It seems a bit simpler.

haltman-at: Well, `allocations` was set to `this.allocations`, so doing it that way would alter `this.allocations, which I don't want.

packages/decoder/test/current/test/downgrade-test.js
haltman-at: > Why does deployedContract not have a deployedBytecode field after deployment?

It does, but that's not where the decoder is going to look for it. :)  It's going to look in the project info, that it was deleted from.

> And what is the point of setting downgradeTest.deployedBytecode to undefined because downgradeTest? It is not used in this test after that assignment.

It is, just not explicitly!  It's part of `mangledCompilations`; `downgradeTest` is part of that, so the purpose of that line is to alter `mangledCompilations`.

packages/decoder/lib/decoders.ts
eggplantzzz: Ah right!

packages/decoder/test/current/test/downgrade-test.js
eggplantzzz: Oh ok I see. But when I print the value of `deployedContract.deployedBytecode` I get `undefined` in this test.

haltman-at: ...huh.  I guess there's some aliasing going on there that I forgot about...?

cds-amal (approved):
This is really cool @haltman-at . I'm glad you added the test!

Issue #4457, opened 2021-11-22
View function can cause insufficient funds error


hellwolf:
- [ ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

In addition to #4397

We also noticed that sometimes view function can cause the following errors (on polygon, could be specific to their node?):

```
err: insufficient funds for gas * price + value: address 0xd66E40b0c30595bEc72153B502aC1E0c4785991B have 13763260682756217913 want 50000000000000000000000 (supplied gas 500000)
```

This occurs when using truffle, and set a default gas price in the network config, for example:

```
        matic: {
            gasPrice: +process.env.MATIC_GAS_PRICE || 30e9,
        },
```

The proposed solution is documented in https://github.com/trufflesuite/truffle/pull/4448

## Steps to Reproduce

https://github.com/hellwolf/truffle_contract_staticcall_gasprice_issue

        // This will trigger the bug, by using high default gasPrice
        // `GAS_PRICE=1e12 npx truffle --network matic exec scripts/test.js`
        // but without GAS_PRICE setting, it works
        // `npx truffle --network matic exec scripts/test.js`

```
$ npx truffle --network matic exec scripts/test.js 
Using network 'matic'.

usdcx name:  Super USDC (PoS)
```

```
$ GAS_PRICE=1e12 npx truffle --network matic exec scripts/test.js 
Using network 'matic'.

err: insufficient funds for gas * price + value: address 0x627306090abaB3A6e1400e9345bC60c78a8BEf57 have 62406659 want 600000000000000000000000000 (supplied gas 50000000)
Truffle v5.4.21 (core: 5.4.21)
Node v14.18.1
```

## Environment

* Operating System: Ubuntu 20
* Ethereum client: ?
* Truffle version (`truffle version`): v5.4.21
* node version (`node --version`): v14.18.1
* npm version (`npm --version`): ?

haltman-at:
OK, I can reproduce the issue.  Thanks for bringing this up.  It definitely does seem that, in this instance, a call is failing with insufficient value.  Moreover, this is happening when using an external service, not Ganache, so this isn't a Ganache issue.  As such, it does look like there is something to be done here.

The approach taken in your PR #4397 is one possibility (although the PR itself will have to be fixed; as currently written it does nothing).  But I think we should make sure we understand the exact problem a bit better before taking any action.

[...]

PR #4606, opened 2022-01-07
Ignore config gas prices when making a call

haltman-at:
Addresses #4457.  Supersedes #4448 because it looks @hellwolf isn't coming back to finish that.

This PR makes it so that, when making a call (not a transaction!) with Truffle Contract, any defaults for `gasPrice`, `maxFeePerGas`, or `maxPriorityFeePerGas` coming from the config are ignored.  The reason for this is that including too-high values for these parameters can cause "insufficient funds" errors on certain networks, even though you're making a call, not a transaction.  This is annoying!

Note that any values for these parameters *explicitly specified when making the call* are still respected, because, I don't know, maybe the user wants to be able to set these.  But any defaults coming from the config will be ignored.

To do this, I added an extra parameter to `getTxParams`, which when passed will exclude those parameters from the defaults.  I renamed the `skipNetworkCheck` flag in `prepare` to `isCall`, since it now both skips network checks, and activates this behavior.

Also, `prettier` has some changes, as always.

I didn't add a test for this PR because, well... the contract tests are confusing and I wasn't sure how to add a test for this, sorry.  But I tested it manually and verified that it worked!

### DIFF
--- a/packages/contract/lib/utils/index.js
+++ b/packages/contract/lib/utils/index.js
@@ -133,22 +133,27 @@ const Utils = {
   },
 
   // Extracts optional tx params from a list of fn arguments
-  getTxParams(methodABI, args) {
+  getTxParams(methodABI, args, ignoreDefaultGasPriceParams = false) {
     const constructor = this;
 
     const expectedArgCount = methodABI ? methodABI.inputs.length : 0;
 
-    let tx_params = {};
+    let txParams = {};
     const lastArg = args[args.length - 1];
 
-    if (
-      args.length === expectedArgCount + 1 &&
-      Utils.isTxParams(lastArg)
-    ) {
-      tx_params = args.pop();
+    if (args.length === expectedArgCount + 1 && Utils.isTxParams(lastArg)) {
+      txParams = args.pop();
     }
 
-    return Utils.merge(constructor.class_defaults, tx_params);
+    let defaultParams = constructor.class_defaults;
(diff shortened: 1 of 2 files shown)


packages/contract/lib/utils/index.js
```diff
@@ -133,22 +133,27 @@ const Utils = {
   },
 
   // Extracts optional tx params from a list of fn arguments
-  getTxParams(methodABI, args) {
+  getTxParams(methodABI, args, ignoreDefaultGasPriceParams = false) {
```
eggplantzzz, line 136: Might it be worth making a comment describing why this last param might be useful? Specifically about why we implemented it?

haltman-at: Sure, that makes sense.  I've added a comment.

lsqproduction (approved):
Looks good

packages/contract/lib/utils/index.js
```diff
@@ -133,22 +133,31 @@ const Utils = {
...
     const constructor = this;
 
     const expectedArgCount = methodABI ? methodABI.inputs.length : 0;
 
-    let tx_params = {};
+    let txParams = {};
```
eggplantzzz, line 141: nice

haltman-at: Haha yeah figured I'd change that while I was at it. :)

eggplantzzz (approved):
I like it!

PR #4610, opened 2022-01-11
Bug fix: Replace special characters with underscores when fetching from Sourcify

haltman-at:
It turns out that Sourcify will replace colons in file paths with underscores.  Note that the metadata *itself* is unchanged, but the actual file paths are transformed.  So after we read the file paths out of the metadata, we then have to transform them before using them to fetch the corresponding file.  We don't transform them for any other purpose though; we still save it under the untransformed file path so that everything fits together properly.

I do not know if this is the only such transformation they make; it may be necessary to make others as well.  I have asked them on Gitter, hopefully they respond.

This PR has no tests because source-fetcher still has no tests but hopefully that will be rectified soon!

(Also, `prettier` made several changes.)

### DIFF
--- a/packages/source-fetcher/lib/sourcify.ts
+++ b/packages/source-fetcher/lib/sourcify.ts
@@ -9,7 +9,8 @@ import retry from "async-retry";
 
 //this looks awkward but the TS docs actually suggest this :P
 const SourcifyFetcher: FetcherConstructor = class SourcifyFetcher
-  implements Fetcher {
+  implements Fetcher
+{
   get fetcherName(): string {
     return "sourcify";
   }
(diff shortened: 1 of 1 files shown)


gnidan (approved):
Good catch!

cds-amal (approved):
Yikes! Looks good. I wonder if there are other mutations

haltman-at:
Yup, hopefully they get back to us about that!  Otherwise... I can try verifying a contract with lots of special characters in the file name and see what happens? :P

haltman-at:
OK, it seems like this actually applies to any characters other than ASCII alphanumerics, the hyphen, and the period, and the forward slash.  So I've updated the PR accordingly.

PR #4611, opened 2022-01-11
Fixing Debugger sourcemap issues.

### LINKED ISSUE #4594: Internal issue: Debugger tests no longer have source maps
## Issue

The debugger tests don't seem to have source maps anymore; as a result, when one fails, the stacktrace is useless, as it just points into `debugger.js`.  Presumably this is a side-effect of some of the changes made in #4507.

## Environment

* Truffle version (`truffle version`): 5.4.26

michaeljohnbennett:
Small PR fix to revert the sourcemap config using a relative sourceroot.

reverting back to default processor and source-map flag.﻿

Fixes #4594

### DIFF
--- a/packages/debugger/webpack/webpack.config-common.js
+++ b/packages/debugger/webpack/webpack.config-common.js
@@ -1,10 +1,9 @@
 const path = require("path");
-const webpack = require("webpack");
 const nodeExternals = require("webpack-node-externals");
 
 module.exports = {
   entry: "./debugger.js",
-  devtool: false,
+  devtool: "source-map",
   target: "node",
 
   module: {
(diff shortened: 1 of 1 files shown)


haltman-at (approved):
I'm just going to assume you know what you're doing here. :)  Thanks a bunch!

haltman-at:
Btw, just a note on the naming and commit messages: Commit messages and PR titles are supposed to be written in a particular style.  So instead of "Reverting the sourceMap config..." it should be "Revert the sourceMap config...", and the PR instead of "Fixing Debugger..." should be "Fix Debugger...".  (@gnidan likes to say they should be written in the imperative, although I would insist it's actually a bare infinitive. :P  Regardless, you use the plain form of the verb, not a participle.)

This doesn't really matter, obviously, but thought I'd point it out for the future, since we do like to stick to this style.

haltman-at:
Btw, you added #4594, but you didn't mention it in the text!  You should mention it in the text!

(We actually typically don't use linked issues per se on this repo, as that causes GitHub to close on merge, whereas on this repo we do close on release.  But mention it in the text!)

PR #4612, opened 2022-01-11
Bug fix: Remove compilationTarget field from metadata from Sourcify

haltman-at:
This PR fixes some compile errors that could occur when using `fetch-and-compile` to fetch sources off of Sourcify.  See, Sourcify gives us the *metadata*, which is not *quite* the same thing as actual compiler input.  For the most part we account for this, but in this case we hadn't.  The metadata settings include a setting called `compilationTarget`.  Actual compiler input doesn't include that setting, but we weren't removing it, and so sometimes the compiler would error out becasue of the presence of this setting.  So, now we delete it to prevent errors.

Ideally I would disunify the `SolcSettings` type into `SolcSettings` and `SolcMetadata`, but I didn't bother as that just seemed like needless complication (for instance, then I'd have to write a type predicate to distinguish them if I wanted functions to be able to operate on both without type errors).

I didn't add a test for the usual reasons but I tested it manually, and, if nothing else, it causes us to get more meaningful compile errors. :P

### DIFF
--- a/packages/source-fetcher/lib/common.ts
+++ b/packages/source-fetcher/lib/common.ts
@@ -24,15 +24,18 @@ export function makeFilename(name: string, extension: string = ".sol"): string {
   }
 }
 
-export const makeTimer: (
-  milliseconds: number
-) => Promise<void> = util.promisify(setTimeout);
+export const makeTimer: (milliseconds: number) => Promise<void> =
+  util.promisify(setTimeout);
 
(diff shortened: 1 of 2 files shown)


gnidan (approved):
I think it's fine to delay making a proper distinction between metadata and compiler input, but would you mind opening an issue for it? Would be nice to have the paper trail in the future.

Other than that - this looks good!

Issue #3867, opened 2021-02-22
Deployer.deploy should return the deployed contract


haltman-at:
## Issue

In migrations, it's a pain to have to write

```javascript
await deployer.deploy(Contract);
let instance = await Contract.deployed();
```

It would be much more convenient to be able to write

```javascript
let instance = await deployer.deploy(Contract);
```

So, `deployer.deploy` ought to return (a promise that resolves to) the deployed contract.  Right now it doesn't return anything and you have to get the deployed contract in a separate step.

## Environment

* Truffle version (`truffle version`): 5.1.67

sukanyaparashar:
> Hm, so I looked at this a bit and I'm uncertain whether it's possible to fix this without a breaking change. In Truffle 6 we can rip out all the deferred chain nonsense, but obviously we can't do that yet. It seems like right now in order for the deferred chain to work, `deployer.deploy` has to return the deferred chain instance, rather than the contract instance. I'm pretty doubtful there's any way to make the former await to the latter...

I have also looked into this issue but didn't find a proper way to return the contract instance. I have a draft PR opened for this but I think for now I might delete the draft as it seems we will have to wait for this to be resolved.

eggplantzzz:
@haltman-at would the breaking change be just for @truffle/deployer or would it be breaking for Truffle more generally?

haltman-at:
I'd call it breaking for Truffle, as it would break lots of existing migrations code.  I suppose that's a question for @gnidan if we want to be sure, though.

PR #4621, opened 2022-01-14
Changed the deploy method to return the contract instance

sukanyaparashar:
### ISSUE
The deployer.deploy() method doesn't return the contract instance and needs an additional line of code as Contract.deployed() to get it. See issue #3867 .

### SOLUTION
New async/await methods are added to get the contract instance directly from the deployer.deploy() method.

### DIFF
--- a/packages/deployer/index.js
+++ b/packages/deployer/index.js
@@ -49,11 +49,19 @@ class Deployer extends Deployment {
     return this.queueOrExec(link(library, destinations, this));
   }
 
-  deploy() {
+  deploy1() {
     const args = Array.prototype.slice.call(arguments);
     const contract = args.shift();
 
     return this.queueOrExec(this.executeDeployment(contract, args, this));
+  }
+
+  async deploy() {
+    const args = Array.prototype.slice.call(arguments);
+    const contract = args.shift();
+
+    let instance = await this.executeDeployment(contract, args, this);
+    return instance;
   }
 
   new() {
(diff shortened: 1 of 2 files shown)


sukanyaparashar:
I have written new methods to return the contract instance directly from the deployer.deploy() method. However, I didn't delete the previous methods and just renamed them.

packages/deployer/index.js
```diff
@@ -49,13 +49,21 @@ class Deployer extends Deployment {
...
+  deploy1() {
     const args = Array.prototype.slice.call(arguments);
     const contract = args.shift();
 
     return this.queueOrExec(this.executeDeployment(contract, args, this));
   }
 
+  async deploy() {
+    const args = Array.prototype.slice.call(arguments);
+    const contract = args.shift();
```
cds-amal, line 61:
``` suggestion
  async deploy(contract, ...args) {
```
[Rest parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters) didn't exist when this was originally written. Let's use it if we can now.

packages/deployer/index.js
```diff
@@ -49,13 +49,21 @@ class Deployer extends Deployment {
...
+
+    let instance = await this.executeDeployment(contract, args, this);
```
cds-amal, line 63: is `this` necessary as the final argument?

cds-amal (commented):
I had started reviewing this without realizing this was on hold. I'll leave my review as a comment and get back to it when it's ready.

PR #4630, opened 2022-01-19
Internal improvement: Remove unused package from debugger

haltman-at:
Looks like I missed removing this package when doing #4547.  Well, it's unused, so let's finish what I started and remove it from the manifest!  This should also address several of the warnings that people are getting on installation. :)


cds-amal (approved):
Looks good.

Issue #4627, opened 2022-01-18
Mocha reporters defined in truffle.config no longer used by default (v5.4.25)


cgewecke:
## Issue

Hello! I help maintain [eth-gas-reporter][1]. 

Am [cross-posting recent issues opened there][2] that v5.4.25 has changed expectations about how truffle loads 3rd party mocha reporters. 

Prior to that version, when someone defined a reporter in truffle.config.js's mocha options it was used by default.

With 5.4.25, the `--reporter` cli flag is the only way to select a reporter other than `spec`. 

A workaround for eth-gas-reporter users is to use the cli flag:
```
truffle test --reporter eth-gas-reporter
```

[1]: https://github.com/cgewecke/eth-gas-reporter
[2]: https://github.com/cgewecke/eth-gas-reporter/issues/268

## Steps to Reproduce
+ truffle unbox metacoin
+ yarn add eth-gas-reporter
+ configure `truffle.config.js` with:
  ```js 
  mocha: {
    reporter: "eth-gas-reporter"
  }
  ```
+ run `truffle test`

## Expected Behavior

There should be a table with gas readings for each deployment and method invocation at the end of the suite

## Actual Results

No table

## Environment

* Operating System: Mac OSX
* Ethereum client: ganache-cli latest
* Truffle version (`5.4.25`):
* node version (`14`):

giuseppeg:
@cgewecke thank you, you saved my day. I have been trying to figure out why the reporter wouldn't work for a while.

eggplantzzz:
Oh no, sorry for breaking this functionality. We'll try and get a fix out for this!

haltman-at:
This should be fixed now!

PR #4632, opened 2022-01-19
Fix mocha reporter issue to work with both command line and truffle-config.js

sukanyaparashar:
### ISSUE
Setting the mocha reporter in the truffle-config.js file was not working after the --reporter flag was added as the command line argument. See #4627 .

### SOLUTION
The mocha reporter can now be set in the truffle-config.js and it can be run using the truffle test command. And if it is not set in the truffle-config.js file, it can be specified as an command line argument like -
truffle test --reporter <reporter_name>
And if nothing is specified when running the truffle test command or in the truffle-config.js file, the default mocha reporter is displayed which is "spec".

### DIFF
--- a/packages/core/lib/commands/test/meta.js
+++ b/packages/core/lib/commands/test/meta.js
@@ -14,7 +14,7 @@ module.exports = {
       type: "boolean",
       default: false
     },
-    debug: {
+    "debug": {
       describe: "Enable in-test debugging",
       type: "boolean",
       default: false
@@ -28,13 +28,13 @@ module.exports = {
       type: "boolean",
       default: false
     },
-    bail: {
+    "bail": {
       alias: "b",
       describe: "Bail after first test failure",
       type: "boolean",
       default: false
     },
-    stacktrace: {
+    "stacktrace": {
       alias: "t",
       describe: "Produce Solidity stacktraces",
       type: "boolean",
@@ -45,10 +45,10 @@ module.exports = {
       type: "boolean",
       default: false
     },
-    reporter: {
+    "reporter": {
       alias: "r",
       describe: "Specify the type of mocha reporter",
-      default: "spec"
+      default: undefined
     }
   },
   help: {
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -5,10 +5,14 @@ const parseCommandLineFlags = options => {
   const reporter = options.reporter || options.r;
 
   /**
-   * If the reporter is returned as undefined, it will overlap the reporter specified in the
-   * config and will display the default mocha reporter "spec".
-   * This if-else condition is explicitly written to avoid this overlapping when user specifies a mocha reporter type
+   * This if-else condition is explicitly written to avoid the overlapping of
+   * the config by the default mocha reporter type when user specifies a mocha reporter type
    * in the config and doesn't specify it as the command line argument.
+   * If condition - The reporter is not returned when it is undefined by the user in the command line which in
+   * turn will check the reporter value specified in the config.
+   * This is required because if the reporter is returned as undefined, it will overlap the reporter specified in the
+   * config and will display the default mocha reporter "spec".
+   * Else condition - The reporter is returned only when the user specifies a reporter type from the command line
    */
   if (reporter === undefined) {
     return {
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -5,8 +5,10 @@ const parseCommandLineFlags = options => {
   const reporter = options.reporter || options.r;
 
   /**
-   * This if-else condition is explicitly defined so that when the reporter is not specified by the user
-   * as the command line argument, it returns the reporter specified in the config.
+   * If the reporter is returned as undefined, it will overlap the reporter specified in the
+   * config and will display the default mocha reporter "spec".
+   * This if-else condition is explicitly written to avoid this overlapping when user specifies a mocha reporter type
+   * in the config and doesn't specify it as the command line argument.
    */
   if (reporter === undefined) {
     return {
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -3,13 +3,22 @@ const parseCommandLineFlags = options => {
   const grep = options.grep || options.g;
   const bail = options.bail || options.b;
   const reporter = options.reporter || options.r;
-  return {
-    mocha: {
-      grep,
-      bail,
-      reporter
-    }
-  };
+  if (reporter === undefined) {
+    return {
+      mocha: {
+        grep,
+        bail
+      }
+    };
+  } else {
+    return {
+      mocha: {
+        grep,
+        bail,
+        reporter
+      }
+    };
+  }
 };
 
 module.exports = async function (options) {
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -3,6 +3,11 @@ const parseCommandLineFlags = options => {
   const grep = options.grep || options.g;
   const bail = options.bail || options.b;
   const reporter = options.reporter || options.r;
+
+  /**
+   * This if-else condition is explicitly defined so that when the reporter is not specified by the user
+   * as the command line argument, it returns the reporter specified in the config.
+   */
   if (reporter === undefined) {
     return {
       mocha: {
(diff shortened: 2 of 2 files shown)


packages/core/lib/commands/test/run.js
```diff
@@ -3,13 +3,22 @@ const parseCommandLineFlags = options => {
...
   const bail = options.bail || options.b;
   const reporter = options.reporter || options.r;
-  return {
-    mocha: {
-      grep,
-      bail,
-      reporter
-    }
-  };
+  if (reporter === undefined) {
```
haltman-at, line 6: Is there some reason this conditional is needed, to exclude `reporter` when it's undefined?  Does something go wrong if you go with the simpler `return { mocha: { grep, bail, reporter } }` regardless of whether `reporter` is defined or not?

packages/core/lib/commands/test/meta.js
```diff
@@ -45,10 +45,10 @@ module.exports = {
       type: "boolean",
       default: false
     },
-    reporter: {
+    "reporter": {
       alias: "r",
       describe: "Specify the type of mocha reporter",
-      default: "spec"
+      default: undefined
```
haltman-at, line 51: Could you explain the reason for this change?  (I'm pretty sure I know, having made exactly this sort of mistake before... but I imagine most people won't find this obvious, I think this could do with some explanation.)

haltman-at (commented):
Seems basically good, but two quick comments -- one on something that can be simplified, one on something that could be explained.

One other minor note, on writing commit messages / PR names: There's a specific style we write these in; they shouldn't be written in past tense ("Mocha reporter issue fixed...") but rather in plain-form present tense ("Fix Mocha reporter issue...").  (Nick would say the imperative, I would say the bare infinitive. :P )

packages/core/lib/commands/test/run.js
sukanyaparashar: @eggplantzzz and I found out that when the reporter is undefined, then it doesn't fall back to the truffle-config.js specification of the mocha reporter defined there. It just displays the default mocha reporter "spec". So, it needs to be excluded when the reporter is undefined.

haltman-at: Oh, weird!  It might be worth adding a code comment to note that, because this looks like the sort of thing a well-meaning simplifier might accidentally remove otherwise.  Thanks for explaining that!

sukanyaparashar: Sure, I will add a comment above the code for better understanding of the condition.

packages/core/lib/commands/test/meta.js
sukanyaparashar: This doesn't actually make much difference though. As undefined also will display the "spec" type of mocha reporter.

haltman-at: Uh, apologies, but that doesn't seem like an answer to what I asked...?

(I'm not saying this change is wrong -- I'm pretty sure it's right!  But it's not *obvious*, and you didn't explain it, so I want the explanation written down to ensure that everyone understands.)

packages/core/lib/commands/test/run.js
```diff
@@ -3,13 +3,22 @@ const parseCommandLineFlags = options => {
...
+    return {
+      mocha: {
+        grep,
+        bail
+      }
+    };
+  } else {
```
cds-amal, line 13: Just a small point that the else clause of this isn't necessary because both branches return.

cds-amal (changes requested):
I wasn't sure how to test this PR. Can a test be added for this change?

packages/core/lib/commands/test/run.js
haltman-at: @sukanyaparashar explained above that actually this is necessary?

cds-amal: I meant the second return doesn't need to be enclosed in a redundant else clause. Sorry I wasn't clear and it is a minor point.

haltman-at: Just speaking for myself here, I think the explicit `else` here is clearer...

sukanyaparashar: @cds-amal This explicit else is necessary because when the reporter is not specified as the command line argument (its is undefined), it should fall back to the truffle-config.js file specification and it happens only when the reporter is not returned. If we return the reporter as undefined, then it displays the default mocha reporter (which is "spec") and does not fall back to the truffle-config.js file.

packages/core/lib/commands/test/run.js
eggplantzzz: @haltman-at if you explicitly set `reporter` to the value `undefined`, it does a merge later that cycles through the keys overwriting other values that might have been set. In this case, it will find that `reporter` has a value, namely `undefined`, and overwrite the other value.

packages/core/lib/commands/test/run.js
sukanyaparashar: Oh! I misunderstood what you said @cds-amal . Yes, I also agree that both the styles work but else may make the code more clearer to understand.

packages/core/lib/commands/test/meta.js
sukanyaparashar: I have made the default as "undefined" because of the fact that the value of the reporter will be undefined if the user doesn't specify any specific reporter as the command line argument and also in the config. And in that case, the default mocha reporter "spec" will be used. However, both behaves in the same way. So, instead of specifying "spec" as the default, i preferred using "undefined" as the default.

packages/core/lib/commands/test/run.js
haltman-at: Yes, that would make sense given what @sukanyaparashar said!

packages/core/lib/commands/test/meta.js
```diff
@@ -28,13 +28,13 @@ module.exports = {
...
     },
-    bail: {
+    "bail": {
       alias: "b",
       describe: "Bail after first test failure",
       type: "boolean",
       default: false
     },
-    stacktrace: {
+    "stacktrace": {
```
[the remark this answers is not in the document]
haltman-at: Yeah, what prettier does is, if one key in the object needs to be quoted, they all get quoted.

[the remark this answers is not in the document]
eggplantzzz: Oh I see, thanks!

packages/core/lib/commands/test/run.js
cds-amal: It's your choice, stylistically.

packages/core/lib/commands/test/meta.js
haltman-at: Oh, so it's just a style thing?  Interesting, that's different from what I had thought -- I had assumed that you changed it because using `"spec"` would cause it to accidentally use `"spec"` in cases where it shouldn't.  I caused a problem like that a while back, and assumed this was similar.  Glad you explained it, because I evidently misunderstood!

cds-amal (approved):
I'm ok with this, and would defer to testing as followup work.

packages/core/lib/commands/test/run.js
```diff
@@ -3,13 +3,27 @@ const parseCommandLineFlags = options => {
...
-  return {
-    mocha: {
-      grep,
-      bail,
-      reporter
-    }
-  };
+
+  /**
+   * This if-else condition is explicitly defined so that when the reporter is not specified by the user
```
haltman-at, line 8: This comment doesn't really explain the reasoning, I'm afraid.  It's not the sort of comment that would stop a well-meaning simplifier from taking out the conditional.

The point of comments is to point out things that are non-obvious or counterintuitive, and what's counterintuitive about this code is the fact that `{ mocha, bail }` acts differently from `{ mocha, bail, reporter: undefined }`.  This is an uncommon situation in Javascript; typically, having a member equal to `undefined`, and having it not present, act the same.  Thus, a well-meaning simplifier who was not aware that the distinction was significant in this case, would likely remove this branch, re-introducing the problem.

I think the comment should make a point of expressing that the undefined-vs-absent distinction really is significant in this case.

haltman-at (commented):
My one comment is that I think your comment needs revision. :)  Otherwise seems good!

packages/core/lib/commands/test/run.js
sukanyaparashar: Thanks @haltman-at . I will make the changes.

haltman-at: Hm, I don't see how the rewritten comment accomplishes the goal.  It doesn't highlight the distinction between absence and undefined.  Again, most readers will think of these as being generally the same, so when you say "If the reporter is returned as undefined", most readers will think of that as being the same as it being absent.  In order to highlight the distinction, you should add a clause like "as opposed to completely absent".  More generally, if you want to highlight a distinction, you need to present both its sides, not just one of them.

haltman-at (approved):
Alright, let's get this in!

PR #4636, opened 2022-01-21
Remove deprecated field from README

kevinweaver:
The example Contract initialization in the README includes a deprecated field and alludes to it being a required field, so I removed it.


haltman-at (approved):
Oh, nice catch!

PR #4637, opened 2022-01-21
Configure Web3 to use WebsocketProvider when connecting to ganache

cds-amal:
This PR improves the performance of `truffle test` on the command line, and `test` within the console and develop REPL. The speed increase comes from bypassing the web3's polling mechanism, which kicks in after about 1sec. This added delay accumulates in an unpleasant user experience when there are many user tests.

Also note, this PR changes the instamine mode of `truffle develop` to  `strict` which is consistent with other truffle testing paths.

@davidmurdoch suggested configuring keep-alive and reconnect logic via options to web3, but I'm not sure if it's necessary in our use-cases here, and I still have to grok Web3's implementation and defaults. I propose this as follow-up work if we can gain a better developer experience now.

## Status

This PR is currently blocked by an issue with `@truffle/db`.

### DIFF
--- a/packages/contract-tests/test/util.js
+++ b/packages/contract-tests/test/util.js
@@ -75,14 +75,11 @@ const util = {
     options = options || {};
     Object.assign(options, { logger: log, ws: true });
 
-    let provider;
     const web3 = new Web3();
 
-    process.env.GETH
-      ? (provider = new Web3.providers.HttpProvider("http://127.0.0.1:8545", {
-          keepAlive: false
-        }))
-      : (provider = ganache.provider(options));
+    const provider = process.env.GETH
+      ? new Web3.providers.WebsocketProvider("ws://127.0.0.1:8545")
+      : ganache.provider(options);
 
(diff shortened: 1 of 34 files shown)


eggplantzzz (approved):
I like it!

cds-amal:
#5065 should be taken into consideration for this PR

eggplantzzz:
Currently there seems to be an issue on the Ganache side that is making CI hang during the tests. From @davidmurdoch in a private chat:
```
The issue is that if a request comes into ganache right after (or during) a disconnect ganache won't ever respond (it won't even reject/throw). so the receiving end keeps waiting for ganache to respond.
```

We'll have to wait until this is resolved and ensure we have the proper setup/teardown on the Truffle side.

eggplantzzz:
We should see how the test suite fares after rebasing against `develop`.

eggplantzzz:
closing this as we switched the testing framework to use "eager" instamine mode which is not technically correct behavior but speeds up tests significantly.

PR #4641, opened 2022-01-24
Bug fix: Account for Geth format in reason.js

haltman-at:
Added a case for `data` being a string directly, which @davidmurdoch says is how Geth does it, and how Ganache will do it starting in 7.0.1.

@davidmurdoch, can you confirm that I've got this correct?

~~(Also, is the current way `return` or `result`?  You said `return`, but our current code is `result` and it seems to pass tests...)~~

Edit: @davidmurdoch confirms it is indeed `result`.

### DIFF
--- a/packages/contract/lib/reason.js
+++ b/packages/contract/lib/reason.js
@@ -38,9 +38,7 @@ const reason = {
         // NOTE this only works for a single failed transaction at a time.
         const hash = Object.keys(data)[0];
         const errorDetails = data[hash];
-        resData =
-          errorDetails.result /* ganache 3.0 */ ||
-          errorDetails.return /* ganache 2.0 */;
+        resData = errorDetails.return /* ganache 2.0 */;
       }
 
       if (resData && resData.includes(errorStringHash)) {
--- a/packages/contract/lib/reason.js
+++ b/packages/contract/lib/reason.js
@@ -10,7 +10,7 @@ const reason = {
    * @param  {InterfaceAdapter}      interfaceAdapter a new helpful friend
    * @return {String|Undefined}      decoded reason string
    */
-  _extract: function(res, web3, _interfaceAdapter) {
+  _extract: function (res, web3, _interfaceAdapter) {
     //I'm not sure why interfaceAdapter is here if it's not used,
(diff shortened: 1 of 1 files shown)


packages/contract/lib/reason.js
```diff
@@ -28,25 +28,24 @@ const reason = {
...
         resData = data.result;
       } else {
         // handle `evm_mine`, `miner_start`, batch payloads, and ganache 2.0
         // NOTE this only works for a single failed transaction at a time.
         const hash = Object.keys(data)[0];
         const errorDetails = data[hash];
         resData =
-          errorDetails.result /* ganache 3.0 */
-          || errorDetails.return /* ganache 2.0 */;
+          errorDetails.result /* ganache 3.0 */ ||
```
eggplantzzz, line 42: while you are doing this would you mind updating this comment to v7? - feel free to ignore this if this is annoying

eggplantzzz (approved):
I like it!

packages/contract/lib/reason.js
davidmurdoch: this line was only true when ganache v7 was going to be called 3.0.0. It should be safe to remove this check.

packages/contract/lib/reason.js
```diff
@@ -28,25 +28,24 @@ const reason = {
       // be updated to respect (instead of computing here)
       const data = res.error.data;
       let resData;
-      if ("result" in data) {
-        // there is a single result (only in ganache 3.0+)
+      if (typeof data === "string") {
+        resData = data; // geth, Ganache >7.0.0
```
davidmurdoch, line 32: I'm guess we don't have any tests that hit this?

packages/contract/lib/reason.js
haltman-at: Oh, good to know!  I'll go change that quickly.

packages/contract/lib/reason.js
haltman-at: Not at present, it would seem.

gnidan (approved):

cds-amal (approved):
Looks good!

Issue #4486, opened 2021-12-03
`fetchAndCompile()` should accept multiple addresses at a time.


gnidan:
## Issue

@truffle/source-fetcher respects API request rate limits by introducing a delay between the requests it makes, keeping clock state on a per-invocation basis. @truffle/fetch-and-compile's `fetchAndCompile()` currently only provides a means of accepting one address at a time, and so it provides no means by which to take advantage of this underlying source-fetcher capability.

This effectively makes this package unsuitable for the use case of fetching-and-compiling in bulk. It's a real bummer cause sometimes transactions have lots of EVM logs from different addresses and you just want the compilations for all of them, like, y'know?

The interface for this function should be expanded to also accept lists of addresses, ideally including removing duplicates from lists of non-unique addresses, and the function should ensure that API request delays properly work between requests for addresses in that list.

haltman-at:
Oh interesting, you want to overload the function?  I was assuming it would be a separate function.  I think that would make more sense so that you don't have to break the return type.

Note that much of what you're asking for would be automatically handled by `fetch.ts` so long as you can write an appropriate recognizer. :)

gnidan:
Oh dang, you're right that it's hard to do this without breaking the interface. Maybe we should just do that now while the Codefi integration is fresh on everyone's mind. I'll ask internally if this is something they can be prepared to accommodate.

If we are okay with breaking the interface, then I think we should return what we have now but grouped by address in an object, as well as a convenient concatenation of all the `compilations` in a single array

gnidan:
This is addressed and released in [Truffle v5.4.31](https://github.com/trufflesuite/truffle/releases/tag/v5.4.31).

PR #4644, opened 2022-01-25
Enhancement: Create `fetchAndCompileMultiple` function

haltman-at:
This PR creates the `fetchAndCompileMultiple` function, addressing issue #4486.  The reason for this, remember, is that if you use `fetchAndCompile` multiple times in sequence, this can result in timeouts, as the multiple invocations won't be aware of one another; the point of using `fetchAndCompileMultiple` is so that all the fetching can happen on a common clock and backoff schedule, avoiding timeout errors.

Most of this is pretty straightforward, it just consists of implementing a new recognizer, the `MultipleRecognizer`, which works a lot like the `DebugRecognizer`.  However there are a few particular things I want to point out:

1. Other than removing duplicate addresses, this recognizer currently makes *no* attempt to avoid unnecessary fetches/compiles like the `DebugRecognizer` does!  It does not employ any sort of bytecode matching.  I figured it was better to just get out an initial version that doesn't bother with that, and then we can go back and improve this later (likely this should be an issue :P ).  Note that fixing this would require making use of a provider!  I guess that would be an optional argument, or something?

2. Since this attempts to get results for multiple addresses, it didn't make sense to me to have it throw if something goes wrong getting one of those addresses.  Instead I had it return `{ results, failures }`; it doesn't throw.  Note that `failures` does contain the error information (if any) for each address that failed, so that information isn't lost.

3. Now that fetch-and-compile has tests (thanks @leeftk!), I added a test for this. :)  While I was at it, I decided to clean up the existing tests a little (hope you don't mind @leeftk :P ).  I didn't go too far with this though since that wasn't my main goal here... however, as I'm intending to go back and add more tests to `fetch-and-compile` in a later PR, I might clean up the tests some more when I go back and do that.


packages/fetch-and-compile/lib/multiple.ts
```diff
@@ -0,0 +1,60 @@
...
+} from "./types";
+
+export class MultipleRecognizer implements Recognizer {
+  private unrecognizedAddresses: string[];
+  private addressesToSkip: Set<string> = new Set();
+  private results: { [address: string]: FetchAndCompileResult } = {};
+  private failureLog: { [address: string]: FetchAndCompileFailureRecord } = {};
+
+  constructor(addresses: string[]) {
+    this.unrecognizedAddresses = [...new Set(addresses)]; //remove duplicates and clone
```
cds-amal, line 18: Would the input include checksum'd addresses?

packages/fetch-and-compile/test/fixture.js
```diff
@@ -2,89 +2,128 @@ const fs = require("fs");
...
-      EVMVersion: "istanbul",
-      Library: "",
-      LicenseType: "GNU GPLv3",
-      Proxy: "0",
-      Implementation: "",
-      SwarmSource:
-        "ipfs://6dd6e03c4b2c0a8e55214926227ae9e2d6f9fec2ce74a6446d615afa355c84f3"
-    }
-  ]
+  "0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D": {
```
cds-amal, line 5: Nice!

cds-amal (approved):
Looks good @haltman-at! Good fixture reorg!

packages/fetch-and-compile/lib/multiple.ts
haltman-at: Oh, good question.  Yeah, I had been assuming all inputs would be checksummed, but I guess that's not necessarily the case, is it?  Perhaps I should either checksum everything or lowercase everything to better ensure that addresses don't occur twice?

cds-amal: Either would be fine, though it'd be faster and less logic to normalize by lowercase.

leeftk:
Loving the new file names here! Makes a lot more sense this way and it's way easier to track. Also I love this new feature, everything looks great! My only question is I'm guessing this works for `Arbitrum` and `Polygon` as well without the need for adding any tests for them? Overall, the changes look great. Can't wait to see how this package evolves!

haltman-at:
> My only question is I'm guessing this works for Arbitrum and Polygon as well without the need for adding any tests for them?

Yeah, I didn't see a need to test every combination of possible inputs.  There's nothing in the new logic that's network-specific, so I figured testing it on one network would suffice.  I might go back and add additional tests on more networks when I go back to expand the testing later, but I'll probably keep those tests on the basic `fetchAndCompile` for simplicity.

packages/fetch-and-compile/lib/multiple.ts
haltman-at: Cool, I'll do that then.

leeftk (approved):
> > My only question is I'm guessing this works for Arbitrum and Polygon as well without the need for adding any tests for them?
> 
> Yeah, I didn't see a need to test every combination of possible inputs. There's nothing in the new logic that's network-specific, so I figured testing it on one network would suffice. I might go back and add additional tests on more networks when I go back to expand the testing later, but I'll probably keep those tests on the basic `fetchAndCompile` for simplicity.

Makes sense. Looks good to me!

gnidan:
> 1. Other than removing duplicate addresses, this recognizer currently makes _no_ attempt to avoid unnecessary fetches/compiles like the `DebugRecognizer` does!  It does not employ any sort of bytecode matching.  I figured it was better to just get out an initial version that doesn't bother with that, and then we can go back and improve this later (likely this should be an issue :P )

Hm, when would this come up? Like if you were fetchAndCompiling a bunch of Uniswap pool addresses, which all have the same bytecode?

I think there's value in reducing to a minimal set of compilations as output, but maybe it can wait.

Definitely worth opening an issue for this, at least!

haltman-at:
> Hm, when would this come up? Like if you were fetchAndCompiling a bunch of Uniswap pool addresses, which all have the same bytecode?

Well, I don't have a great idea how often it would actually come up.  But yes the simplest way it could come up would be if you had a bunch of addresses which all had the same bytecode.  Another way is if you had two different contracts, `A` and `B`, and suppose that the source code of `A` includes a declaration for `B` as well (for instance, maybe `A` inherits from `B`).  Then if `A` is fetched and compiled first, it should be possible to skip getting `B`, as you already have it in the source code for `A`.

This functionality exists in `fetchAndCompileForDebugger`, but not in `fetchAndCompileMultiple`, because it requires 1. bytecode matching, and 2. *getting* the bytecode in the first place.

gnidan:
Hm, alright. Yeah, let's just make an issue for now and prioritize that separately.

haltman-at:
OK, it seems I messed up the tests and some of them were actually hitting real Etherscan instead of the mock!  So I removed the use of `callThrough()` to ensure that can't happen.  I also reworked how the test fixture was organized to make it simpler.

Most of that was unnecessary, though, because it turned out the real problem was inconsistencies arising from the use of lowercase.  So @cds-amal I changed everything to checksum-case instead.  Anyway now this should be good!

packages/fetch-and-compile/lib/multiple.ts
```diff
@@ -16,8 +17,8 @@ export class MultipleRecognizer implements Recognizer {
 
   constructor(addresses: string[]) {
     this.unrecognizedAddresses = [
-      ...new Set(addresses.map(address => address.toLowerCase()))
-    ]; //remove duplicates (case insensitive) and clone
+      ...new Set(addresses.map(Web3Utils.toChecksumAddress))
```
cds-amal, line 20: @haltman-at `toChecksumAddress` can throw and interrupt the flow. Of course, if the inputs are valid addresses, there's nothing to worry about.

PR #4647, opened 2022-01-25
Internal improvement: Remove @ensdomains/ens and include ENSRegistry artifact in deployer

eggplantzzz:
This PR removes the package @ensdomains/ens. The package is only used for the ENSRegistry artifact, which is now saved in @truffle/deployer. This will eliminate the baggage of including @ensdomains/ens in Truffle. In the future we might move to relocate the ENSRegistry artifact to @truffle/resolver but we should probably wait until other parts of Truffle want to access the artifact or we decide to include more artifacts in Truffle.

### DIFF
--- a/packages/deployer/ens.js
+++ b/packages/deployer/ens.js
@@ -27,7 +27,7 @@ class ENS {
   }
 
   async deployNewDevENSRegistry(from) {
-    const ENSRegistryArtifact = require("@ensdomains/ens").ENSRegistry;
+    const ENSRegistryArtifact = require("./ENSRegistry");
     const ENSRegistry = contract(ENSRegistryArtifact);
     ENSRegistry.setProvider(this.provider);
     const ensRegistry = await ENSRegistry.new({ from });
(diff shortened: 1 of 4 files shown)


haltman-at:
(Note, of course, for anyone who might get confused, that this is unrelated to the problems with `@ensdomains/ensjs`. :P )

packages/deployer/ens.js
```diff
@@ -46,8 +46,8 @@ class ENS {
       return { resolvedAddress };
     }
     // deploy a resolver if one isn't set
-    const PublicResolverArtifact = require("@ensdomains/resolver")
-      .PublicResolver;
+    const PublicResolverArtifact =
+      require("@ensdomains/resolver").PublicResolver;
```
haltman-at, line 50: Hm, so we're getting rid of the dependency on `@ensdomains/ens`, but not on `@ensdomains/resolver`?

(I'm actually not clear on *why* we're getting rid of the dependency on `@ensdomains/ens`, given that we seem to just instead be using its contents wholesale... what's the advantage of that?)

eggplantzzz: Eventually we should probably get rid of this as well. These ens packages carry extra baggage given their dependencies. I think the built artifacts in the package are also not current, they were built at a time when a different registry was being used. I don't really think this should make much of a difference, but better to be current anyways I suppose.

haltman-at: So what's the difference between the two that results in it making sense to do them one at a time rather than both at once?

haltman-at: Oh, huh, so you have a PR to do the other one that targets this one, so I guess ultimately they're both going to be in this one after all?

eggplantzzz: I like PRs that are small, it makes them very easy to review. :)

cds-amal (approved):
Very nice! Let's get it in.

PR #4648, opened 2022-01-25
Bug fix: Fix incorrect typing on fetchAndCompile return value

haltman-at:
It seems I was using the incorrect type here; `compile-solidity` (and `compile-vyper`, etc) return a `CompilerResult`, not a `WorkflowCompileResult` (only `workflow-compile` returns the latter).  This PR corrects that.


haltman-at:
Closing this due to problems.  Will make a new PR later that fixes those.

PR #4651, opened 2022-01-26
Have individual compile packages return WorkflowCompileResult (instead of CompilerResult)

haltman-at:
Follow-on to #4650.  This makes it so that `compile-solidity`, `compile-vyper`, and `external-compile` all return a `WorkflowCompileResult`, instead of just a `CompilerResult`.  To do this, the logic for attaching `contracts` was moved from `workflow-compile` to `compile-common` and put in a new function called `promoteCompileResult`.  (Maybe not the best name, but feel free to suggest a better one.)  Of course, `workflow-compile` as a result got logic for concatenating the `contracts`.

Also, I wanted to use `flatMap`, but TS complained because the lib setting was on es2017.  So I updated it to es2019 (in all packages that were on es2017); we no longer support Node 10 so anything from es2019 is fair game.  (@gnidan, you think I should update `target` as well? :) )

### DIFF
--- a/packages/compile-common/src/compilations.ts
+++ b/packages/compile-common/src/compilations.ts
@@ -1,0 +1,9 @@
+import type { CompilerResult, WorkflowCompileResult } from "./types";
+
+export function promoteCompileResult(
+  result: CompilerResult
+): WorkflowCompileResult {
+  const { compilations } = result;
+  const contracts = compilations.flatMap(compilation => compilation.contracts);
+  return { compilations, contracts };
+}
--- a/packages/compile-common/src/compilations.ts
+++ b/packages/compile-common/src/compilations.ts
@@ -7,3 +7,7 @@ export function promoteCompileResult(
   const contracts = compilations.flatMap(compilation => compilation.contracts);
   return { compilations, contracts };
 }
+
+export function emptyWorkflowCompileResult(): WorkflowCompileResult {
+  return { compilations: [], contracts: [] };
+}
--- a/packages/compile-vyper/index.js
+++ b/packages/compile-vyper/index.js
@@ -10,6 +10,7 @@ const findContracts = require("@truffle/contract-sources");
 const Config = require("@truffle/config");
 const { Profiler } = require("@truffle/profiler");
 const { requiredSources } = require("./profiler");
+const { Compilations } = require("@truffle/compile-common");
 
 const { compileJson } = require("./vyper-json");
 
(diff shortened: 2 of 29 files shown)


packages/compile-common/src/compilations.ts
```diff
@@ -0,0 +1,9 @@
+import type { CompilerResult, WorkflowCompileResult } from "./types";
+
+export function promoteCompileResult(
+  result: CompilerResult
+): WorkflowCompileResult {
+  const { compilations } = result;
+  const contracts = compilations.flatMap(compilation => compilation.contracts);
```
gnidan, line 7: I knew I shouldn't have mentioned category theory earlier 😅

packages/compile-vyper/index.js
```diff
@@ -216,7 +217,7 @@ const Compile = {
     // no vyper files found, no need to check vyper
     // (note that JSON-only will not activate vyper)
     if (vyperFiles.length === 0) {
-      return { compilations: [] };
+      return { compilations: [], contracts: [] };
```
gnidan, line 220: Not promoting this with the helper? 😁

(This may be worth a `Compilations.emptyWorkflowCompileResult()`, in seriousness)

gnidan (commented):
Hooray I managed to hold back the urge to mention monads in this review

gnidan:
(Side note: for sake of ease of review, might be worth doing those lib/target bumps in separate PRs. Esp. when the "files changed" includes prettier changes as well)

packages/compile-vyper/index.js
haltman-at: Oh, that's not a bad idea...

packages/compile-common/src/compilations.ts
haltman-at: I've been waiting to finally be able to use `flat()` and `flatMap()` for quite a while now :)

haltman-at:
Yeah, true, that might have been better.  This is what happens when you don't plan, I guess.  I don't think it's worth splitting up now, though, unless someone insists.

gnidan:
> Yeah, true, that might have been better.  This is what happens when you don't plan, I guess.  I don't think it's worth splitting up now, though, unless someone insists.

If you add my suggested function, I'll approve this so no one else need sift through the package.json changes ;)

packages/compile-vyper/index.js
haltman-at: OK, added this!

gnidan (approved):
Cool!

PR #4652, opened 2022-01-26
Internal improvement: Update TS target to es2018

haltman-at:
We no longer support Node 10, and Node 12 supports everything from ES2019, so this should be safe, right?

### DIFF
--- a/packages/abi-utils/tsconfig.json
+++ b/packages/abi-utils/tsconfig.json
@@ -4,7 +4,7 @@
 
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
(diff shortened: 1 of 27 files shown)


gnidan (approved):
> We no longer support Node 10, and Node 12 supports everything from ES2019, so this should be safe, right?

Looks like it! (https://node.green/#ES2019)

This is probably worth a broader process discussion... do we want to up these targets every Node release? (But for now, approving this because no one complains about Node v10 :)

gnidan:
Although seems like something's broken!

haltman-at:
Oh, goddammit.  It looks like Browserify can't handle `catch` without a variable, which was introduced in ES2019.  And of course there's been no updates to Browserify since 17.0.0.  So either we move `contract` off of Browserify and onto Webpack or something (not something I want to do in this PR), or we settle for only updating to ES2018 instead.

...I'll go file an issue for getting Browswerify out of Truffle Contract, and in the meantime I'll downgrade this PR to ES2018...

PR #4653, opened 2022-01-26
Internal improvement: Remove lodash.flatten dependency

haltman-at:
We're on Node 12 now, no need to be using `lodash.flatten` anymore, that's what `array.flat` and `array.flatMap` are for. :)  I've taken out `lodash.flatten` and replaced it with uses of those functions. :)

(Note there are still plenty of places in our code that are doing other ad-hoc things instead of using `flat` and `flatMap`, but, uh, I don't think there's any hurry to go replacing those all immediately...)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -18804,7 +18804,7 @@ lodash.find@4.6.0:
   resolved "https://registry.yarnpkg.com/lodash.find/-/lodash.find-4.6.0.tgz#cb0704d47ab71789ffa0de8b97dd926fb88b13b1"
   integrity sha1-ywcE1Hq3F4n/oN6Ll92Sb7iLE7E=
 
-lodash.flatten@^4.2.0, lodash.flatten@^4.4.0:
+lodash.flatten@^4.2.0:
   version "4.4.0"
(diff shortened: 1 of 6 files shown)


yarn.lock
```diff
@@ -18804,7 +18804,7 @@ lodash.find@4.6.0:
   resolved "https://registry.yarnpkg.com/lodash.find/-/lodash.find-4.6.0.tgz#cb0704d47ab71789ffa0de8b97dd926fb88b13b1"
   integrity sha1-ywcE1Hq3F4n/oN6Ll92Sb7iLE7E=
 
-lodash.flatten@^4.2.0, lodash.flatten@^4.4.0:
+lodash.flatten@^4.2.0:
```
eggplantzzz, line 18807: Is this still left in somewhere else?

haltman-at: I have to assume it's still a transitive dependency.  Checking `yarn why`, it's apparently used by `cheerio`, which seems to be used in various places.

I don't think it's particular essential root out this dependency entirely; it's not causing dependabot alerts or warnings or anything.  I just wanted to get rid of our direct usage now that we can just use `flat` and `flatMap`.

eggplantzzz (approved):
I like it!

Issue #4635, opened 2022-01-21
Debugger not working with web3, fails after Debugger.forTx call


davidcallanan:
- [ ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

I am trying to use the `@truffle/debugger` package with a web3 provider.

There is an error after calling `await Debugger.forTx(...)`

## Steps to Reproduce

```
import Web3 from "web3";
import Debugger from "@truffle/debugger";

let web3provider = new Web3.providers.HttpProvider("https://eth-mainnet.alchemyapi.io/v2/my_secret_key");
let web3 = new Web3(web3provider, null, {
	transactionConfirmationBlocks: 1,
});

let contracts = [... generated programatically ...];
let transactionHash = "0x03e784bfcb8f6175f7a321586f671dff08a87070f62f5c9ae928e8f2c31e17bd"; // randomly chosen from etherscan

// error happens during this function call:
let bugger = await Debugger.forTx(
	transactionHash,
	{
		provider: web3,
		contracts,
	},
);
```

## Expected Behavior

Debugging should not error.

## Actual Results

```
C:\Users\David\Documents\git\crypto-debug\backend\node_modules\.pnpm\web3-providers-http@1.7.0\node_modules\web3-providers-http\lib\index.js:86
    var request = this._prepareRequest();
                       ^
TypeError: this._prepareRequest is not a function
    at HttpProvider.send (C:\Users\David\Documents\git\crypto-debug\backend\node_modules\.pnpm\web3-providers-http@1.7.0\node_modules\web3-providers-http\lib\index.js:86:24)
    at node:internal/util:360:7
    at new Promise (<anonymous>)
    at promisify (node:internal/util:346:12)
    at xn.getTrace (C:\Users\leefaria\code\truffle\packages\debugger\lib\web3\adapter.js:13:24)
    at runCallEffect (C:\Users\David\Documents\git\crypto-debug\backend\node_modules\.pnpm\@redux-saga+core@1.1.3\node_modules\@redux-saga\core\dist\redux-saga-core.dev.cjs.js:528:21)
    at runEffect (C:\Users\David\Documents\git\crypto-debug\backend\node_modules\.pnpm\@redux-saga+core@1.1.3\node_modules\@redux-saga\core\dist\redux-saga-core.dev.cjs.js:1208:7)
[...]

PR #4661, opened 2022-01-27
Bug fix: Use bind with provider.send in debugger's adapter

haltman-at:
Addresses #4635.

I'll be honest, I don't fully understand this one.  For some reason, the way that `provider.send` was being invoked, somehow `this` was getting set to the wrong value (i.e., not the provider).  As such, when `provider.send()` attempted to call `this._prepareRequest()`, intending to call its own method `provider._prepareRequest()`, it instead called `somethingElse._prepareRequest()`, causing a crash.  (I didn't bother determining the value of `somethingElse`.)

I have no idea why this happened in this case, and, for that matter, why it *doesn't* happen normally... this doesn't happen when the debugger is used via the CLI, for instance!

Regardless, it was happening, so I just used `bind` to enforce that `this` was set correctly.  This is kind of the lazy approach -- ideally I should figure out exactly why this was happening and perhaps root out the problem earlier -- but I don't think I have a reasonable chance of actually figuring that out, and it seems unlikely to me that the solution would be different in that case, so I just don't think it's worth it to try to figure that out; I think the lazy solution is appropriate here.

### DIFF
--- a/packages/debugger/lib/web3/adapter.js
+++ b/packages/debugger/lib/web3/adapter.js
@@ -10,7 +10,8 @@ export default class Web3Adapter {
   }
 
   async getTrace(txHash) {
-    let result = await promisify(this.web3.currentProvider.send)(
+    const provider = this.web3.currentProvider;
+    const result = await promisify(provider.send.bind(provider))(
       //send *only* uses callbacks, so we use promsifiy to make things more
       //readable
       {
--- a/packages/debugger/lib/web3/adapter.js
+++ b/packages/debugger/lib/web3/adapter.js
@@ -11,22 +11,22 @@ export default class Web3Adapter {
 
   async getTrace(txHash) {
     const provider = this.web3.currentProvider;
-    const result = await promisify(provider.send.bind(provider))(
-      //send *only* uses callbacks, so we use promsifiy to make things more
-      //readable
-      {
-        jsonrpc: "2.0",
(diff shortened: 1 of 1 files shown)


packages/debugger/lib/web3/adapter.js
```diff
@@ -10,7 +10,8 @@ export default class Web3Adapter {
   }
 
   async getTrace(txHash) {
-    let result = await promisify(this.web3.currentProvider.send)(
+    const provider = this.web3.currentProvider;
+    const result = await promisify(provider.send.bind(provider))(
```
cds-amal, line 14: Might be worth leaving a comment explaining the need for the bind.

cds-amal (approved):
Wow, what an issue! It's a shame we couldn't get a local reproduction to add to our test suite. Nothing stands out in the change and our tests are passing, so let's get this in.

haltman-at:
Well, I should note that I *can* reproduce the issue locally, I just don't have any idea why it occurs in this case and not others.  So I don't know that I can reliably reproduce it in a test.  Which is why I didn't include one.

packages/debugger/lib/web3/adapter.js
haltman-at: Yeah, not a bad idea.

cds-amal: That works!

Issue #4660, opened 2022-01-26
Upgrade Ganache to v7.0.1 once it's released


gnidan:
## Issue

Just opening this issue to make sure we explicitly upgrade Ganache version to v7.0.1 to address the bug raised in #4639

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.4.30
* node version (`node --version`):
* npm version (`npm --version`):

gnidan:
This is addressed and released in [Truffle v5.4.31](https://github.com/trufflesuite/truffle/releases/tag/v5.4.31).

Issue #4639, opened 2022-01-22
Uncaught TypeError: Cannot read properties of null (reading 'return')


fabianorodrigo:
Hi,

I have some tests that run fine on ganache 2.5.4 but got this exception when I tried to run on ganache 7.

`1) Contract: BetToken
       "before each" hook for "Should Ether goes to ERC20 owner after destroy contract":
     Uncaught TypeError: Cannot read properties of null (reading 'return')
      at Object._extract (/home/fabianorodrigo/.nvm/versions/node/v16.13.2/lib/node_modules/truffle/build/webpack:/packages/contract/lib/reason.js:29:1)
      at /home/fabianorodrigo/.nvm/versions/node/v16.13.2/lib/node_modules/truffle/build/webpack:/packages/contract/lib/reason.js:64:1
      at /home/fabianorodrigo/.nvm/versions/node/v16.13.2/lib/node_modules/truffle/build/webpack:/packages/provider/wrapper.js:107:1
      at XMLHttpRequest.request.onreadystatechange (/home/fabianorodrigo/.nvm/versions/node/v16.13.2/lib/node_modules/truffle/build/webpack:/node_modules/web3/node_modules/web3-providers-http/lib/index.js:98:1)
      at XMLHttpRequestEventTarget.dispatchEvent (/home/fabianorodrigo/.nvm/versions/node/v16.13.2/lib/node_modules/truffle/build/webpack:/node_modules/xhr2-cookies/dist/xml-http-request-event-target.js:34:1)
      at XMLHttpRequest.exports.modules.996763.XMLHttpRequest._setReadyState (/home/fabianorodrigo/.nvm/versions/node/v16.13.2/lib/node_modules/truffle/build/webpack:/node_modules/xhr2-cookies/dist/xml-http-request.js:208:1)
      at XMLHttpRequest.exports.modules.996763.XMLHttpRequest._onHttpResponseEnd (/home/fabianorodrigo/.nvm/versions/node/v16.13.2/lib/node_modules/truffle/build/webpack:/node_modules/xhr2-cookies/dist/xml-http-request.js:318:1)
      at IncomingMessage.<anonymous> (/home/fabianorodrigo/.nvm/versions/node/v16.13.2/lib/node_modules/truffle/build/webpack:/node_modules/xhr2-cookies/dist/xml-http-request.js:289:47)
      at IncomingMessage.emit (node:events:402:35)
      at endReadableNT (node:internal/streams/readable:1343:12)
[...]

PR #4663, opened 2022-01-27
Dependency update: Update ganache to 7.0.1

haltman-at:
Addresses #4660, #4639.

Note I had to udpate some tests in `@truffle/decoder` to account for the changed `eth_call` error format.


eggplantzzz:
In a slack chat discussion with @davidmurdoch, v7.0.0 Ganache returns
```
{
   "error": {
     "message": "...",
     "code": ...
     "data": {
          ... bunch of props, like `hash` and `programCounter`...
         "result": "<raw revert hex string>" <---- this moves (See below)
     }
   } 
}
```
and now is changed to return
```
{
   "error": {
     "message": "...", // <- we'll change the message
     "code": ...
     "data": "<raw revert hex string>"  <---- new home
   } 
}
```
for reverts during `eth_call`.

In the case of the tests here, web3 will throw and the error will be the error object above.

PR #4664, opened 2022-01-27
Internal improvement: Add a bunch more testing to fetch-and-compile

haltman-at:
This PR adds a whole bunch more testing to fetch-and-compile (and also cleans up the existing tests a tiny bit while I'm at it).

Firstly, it adds more checks to the existing tests (these checks are of course carried over to the new tests as well).  Before, the tests only checked for the contract name in `result.sourceInfo` and in `result.compilations`; but since #4651 is merged, I had them check `result.contracts` for the name as well.  I also had each test check that `fetchedVia` is correct.

A lot more could still be done regarding adding checks, of course, but for simplicity I decided to stop there.

But also, this PR adds a bunch more tests to exercise code-paths in `source-fetcher` that the original tests didn't.  This PR adds tests of the Etherscan multi-source case and the Etherscan JSON case; it also adds tests of Sourcify, testing both full match and partial match, and being sure to test what happens if there are special characters in the filename.  I also added config to ensure that each test tries only the intended fetcher.

This PR does *not* add tests of the Etherscan Vyper case, since, even though that works in `source-fetcher`, `fetch-and-compile` doesn't support it yet.

Note that for the Sourcify fixtures, I decided it was a lot easier to just use the file system rather than try to fit them into some JS object like how the Etherscan fixtures work.  Also, sorry if `L1StandardBridge.json` is unreadable, but, well, I didn't really think there was any practical way to make it readable while still being sure to keep the same structure.

Also, I actually made one *non*-test change, although not a functional change: In the Sourcify source fetcher, where we invoke Axios, I changed it from `axios(...)` to the synonymous `axios.request(...)`.  This is because you can't easily use Sinon to stub a call to `axios()` itself, but you can totally stub `axios.request()`. :)


cds-amal (approved):
Looks good @haltman-at! 

These tests sure have a lot of fixtures, and with new networks to support in the future, we may have to revisit how we organize them. Not for this PR though!

Issue #4665, opened 2022-01-27
Dry-run prints a bunch of Ganache logging output


haltman-at:
## Issue

I deployed a contract to Goerli with `truffle migrate`; the dry-run that ran initially included a ton of Ganache logging output that should have been hidden (and previously was).

Apologies for lack of details, but I didn't save the output and I'm hoping the above is sufficient for reproduction.

## Environment

* Truffle version (`truffle version`): 5.4.30

PR #4668, opened 2022-01-27
Bug fix: Suppress Ganache logging when running the dry run

eggplantzzz:
This addresses https://github.com/trufflesuite/truffle/issues/4665.

Ganache logs by default, this PR suppresses logging by adding an extra argument to `Environment.fork` that allows passing options for starting Ganache. It then passes the appropriate options for suppressing logging when doing the dry run.

### DIFF
--- a/packages/environment/environment.js
+++ b/packages/environment/environment.js
@@ -26,7 +26,7 @@ const Environment = {
   },
 
   // Ensure you call Environment.detect() first.
-  fork: async function (config) {
+  fork: async function (config, ganacheOptions) {
     expect.options(config, ["from", "provider", "networks", "network"]);
 
     const interfaceAdapter = createInterfaceAdapter({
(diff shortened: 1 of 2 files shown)


packages/environment/environment.js
```diff
@@ -47,15 +47,16 @@ const Environment = {
     const upstreamNetwork = config.network;
     const upstreamConfig = config.networks[upstreamNetwork];
     const forkedNetwork = config.network + "-fork";
-    const ganacheOptions = {
+    const options = {
```
haltman-at, line 50: I think this variable naming is a little unclear; it's not clear from the names what the distinction is supposed to be between `options` and `ganacheOptions`.  It might have been better to just reassign `ganacheOptions`.

haltman-at (approved):
Looks basically good, thanks for getting this!

packages/environment/environment.js
eggplantzzz: Yeah maybe you're right, I'm just slightly allergic to re-assignments. I'll update this :)

PR #4670, opened 2022-01-28
Internal improvement: Organize sources for etherscan tests in fetch-and-compile

haltman-at:
As requested by @cds-amal on #4664.  I didn't do it there because I misunderstood what he was asking.  Anyway, now I've done it -- I took the sources and organized them by the network and address that they pertain to.  So, that should make that less of a mess.  (Note these all go under `etherscan`, since the sourcify ones were already in their own folder.)


cds-amal (approved):
Love this! Thanks @haltman-at

PR #4671, opened 2022-01-28
Internal improvement: refactor connectOrStart

cds-amal:
This PR 
  - updates the use of the Ganache's `server.close()` method, which uses a promise instead of node's callback idiom. We should tear down Ganache properly.
  - cleans up the logic for `connectOrStart`
  - adds 2 tests to verify `connectOfStart` try/catch logic path.

### DIFF
--- a/packages/environment/.eslintrc.json
+++ b/packages/environment/.eslintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": ["../../.eslintrc.package.json"]
+}
--- a/packages/environment/develop.js
+++ b/packages/environment/develop.js
@@ -129,7 +129,10 @@ const Develop = {
       disconnect = await this.connect(options);
       started = true;
     } finally {
-      return { disconnect, started };
+      return {
+        disconnect,
+        started
+      };
     }
   }
 };
--- a/packages/environment/develop.js
+++ b/packages/environment/develop.js
@@ -4,7 +4,7 @@ const { spawn } = require("child_process");
 const debug = require("debug");
 
 const Develop = {
-  start: async function(ipcNetwork, options = {}) {
+  start: async function (ipcNetwork, options = {}) {
     let chainPath;
 
     // The path to the dev env process depends on whether or not
@@ -39,7 +39,7 @@ const Develop = {
     });
   },
 
-  connect: function(options) {
+  connect: function (options) {
     const debugServer = debug("develop:ipc:server");
     const debugClient = debug("develop:ipc:client");
     const debugRPC = debug("develop:ganache");
@@ -69,7 +69,7 @@ const Develop = {
     if (options.log) {
       debugRPC.enabled = true;
 
-      loggers.ganache = function() {
+      loggers.ganache = function () {
         // HACK-y: replace `{}` that is getting logged instead of ""
         var args = Array.prototype.slice.call(arguments);
         if (
@@ -88,21 +88,21 @@ const Develop = {
       ipc.config.maxRetries = 0;
     }
 
-    var disconnect = function() {
+    var disconnect = function () {
       ipc.disconnect(ipcNetwork);
     };
 
     return new Promise((resolve, reject) => {
-      ipc.connectTo(ipcNetwork, connectPath, function() {
-        ipc.of[ipcNetwork].on("destroy", function() {
+      ipc.connectTo(ipcNetwork, connectPath, function () {
+        ipc.of[ipcNetwork].on("destroy", function () {
           reject(new Error("IPC connection destroyed"));
         });
 
-        ipc.of[ipcNetwork].on("truffle.ready", function() {
+        ipc.of[ipcNetwork].on("truffle.ready", function () {
           resolve(disconnect);
         });
 
-        Object.keys(loggers).forEach(function(key) {
+        Object.keys(loggers).forEach(function (key) {
           var log = loggers[key];
           if (log) {
             var message = `truffle.${key}.log`;
@@ -113,30 +113,23 @@ const Develop = {
     });
   },
 
-  connectOrStart: async function(options, ganacheOptions) {
+  connectOrStart: async function (options, ganacheOptions) {
     options.retry = false;
 
     const ipcNetwork = options.network || "develop";
 
-    let connectedAlready = false;
+    let started = false;
+    let disconnect;
 
     try {
-      const disconnect = await this.connect(options);
-      connectedAlready = true;
-      return {
-        started: false,
-        disconnect
-      };
+      disconnect = await this.connect(options);
     } catch (_error) {
       await this.start(ipcNetwork, ganacheOptions);
       options.retry = true;
-      const disconnect = await this.connect(options);
-      if (connectedAlready) return;
-      connectedAlready = true;
-      return {
-        started: true,
-        disconnect
-      };
+      disconnect = await this.connect(options);
+      started = true;
+    } finally {
+      return { disconnect, started };
     }
   }
 };
--- a/packages/environment/test/connectOrStartTest.js
+++ b/packages/environment/test/connectOrStartTest.js
@@ -1,0 +1,33 @@
+const assert = require("chai").assert;
+const Develop = require("../develop");
+
+describe("connectOrStart test network", async function () {
+  const ipcOptions = { network: "test" };
+  const ganacheOptions = {
+    host: "127.0.0.1",
+    network_id: 666,
+    port: 6969
+  };
+
+  it("starts Ganache when no Ganache instance is running", async function () {
+    const connection = await Develop.connectOrStart(ipcOptions, ganacheOptions);
+    assert.isTrue(connection.started);
+    assert.isFunction(connection.disconnect);
+    connection.disconnect();
+  });
+
+  it("connects to an established ganache instance", async function () {
+    //establish a connection
+    const { disconnect: connectionOneDisconnect } =
+      await Develop.connectOrStart(ipcOptions, ganacheOptions);
+
+    //invoke the method again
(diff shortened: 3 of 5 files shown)


packages/environment/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": ["../../.eslintrc.package.json"],
+    "globals": {
+        "BUNDLE_CHAIN_FILENAME": "readonly"
```
haltman-at, line 4: What's this for?

packages/environment/develop.js
```diff
@@ -113,29 +113,25 @@ const Develop = {
...
-      return {
-        started: false,
-        disconnect
-      };
+      disconnect = await this.connect(options);
     } catch (_error) {
       await this.start(ipcNetwork, ganacheOptions);
       options.retry = true;
-      const disconnect = await this.connect(options);
-      if (connectedAlready) return;
```
haltman-at, line 134: Huh, there's no way this could ever be hit, is there?  And then it does something nonsensical too.  Yikes.  Glad to see that gone.

haltman-at (commented):
Seems basically good to me, confused about something in `.eslintrc.json` though.

packages/environment/test/connectOrStartTest.js
```diff
@@ -0,0 +1,33 @@
...
+    host: "127.0.0.1",
+    network_id: 666,
+    port: 6969
+  };
+
+  it("starts Ganache when no Ganache instance is running", async function () {
+    const connection = await Develop.connectOrStart(ipcOptions, ganacheOptions);
+    assert.isTrue(connection.started);
+    assert.isFunction(connection.disconnect);
+    connection.disconnect();
```
davidmurdoch, line 16: You may want to try/finally these, to ensure things shuts down even if the tests fail.

cds-amal: Good idea, thanks @davidmurdoch.

packages/environment/.eslintrc.json
cds-amal: ESlint complained about a no-undef ruld. It's to resolve [this global name](https://github.com/trufflesuite/truffle/blob/connect-or-start/packages/environment/develop.js#L13) introduced by our webpack config. The setting configures ESLint to recognize BUNDLE_CHAIN_FILENAME as valid variable name and check its usage is read only

haltman-at: Oh, weird!  I had no idea about that.  Do you know, is this used anywhere else?  I'm wondering if it makes sense to put this in the global one.

haltman-at: OK I just checked and it looks like it's not used anywhere else, so I guess it's fine to just put it here.

cds-amal: Yea, I looked too. The ESLint config setup is very flexible now!

packages/environment/test/connectOrStartTest.js
```diff
@@ -0,0 +1,54 @@
...
+        await Develop.connectOrStart(ipcOptions, ganacheOptions)
+      ).disconnect;
+
+      //invoke the method again
+      const result = await Develop.connectOrStart(ipcOptions, ganacheOptions);
+      connectionTwoDisconnect = result.disconnect;
+      assert.isFalse(result.started);
+    } catch (error) {
+      console.log(error);
+      assert.fail("Develop.connectOrStart should not have thrown!");
```
davidmurdoch, line 43: You can omit the `catch` here, i think... the reason the test fails should be apparent in the error logs reported by the test runner. It seems strange to me to catch the actual error and then re-throw something else; it could make debugging the reason it failed in the first place more difficult.

haltman-at: Oh, yeah, I missed this, but I want to add my agreement here.  Catch-and-rethrow or catch-and-fail is generally a bad idea for exactly this reason, it obscures the error's origin.  If you want throwing to cause a failure, I would also say that the way to do that is by just not catching at all.

haltman-at: To be clear if you need to do some stuff before rethrowing the *original* error that makes sense, it's throwing a *new* one, as with `assert.fail()` here, that's the problem.  In this case you log the original error before failing?  Which does avoid losing that information, but ultimately you're not accomplishing that couldn't be accomplished more simply by not catching in the first place.

cds-amal: Thanks @haltman-at, @davidmurdoch. I agree that not catching is better in this case. Updated!

cds-amal:
This PR continues to be dogged by some race condition(s). It is odd that the CI does not print the completed tasks. What a PITA.

haltman-at:
So, wait, is this working now, or was that just luck?

cds-amal:
Hard to tell @haltman-at. I'm going to wait on the fix to [Ganache-2185](https://github.com/trufflesuite/ganache/issues/2185) as it seems to be relevant during my investigation.

cds-amal:
Yep, definitely lucky the last time. I re-ran CI and [it failed](https://github.com/trufflesuite/truffle/runs/5026047922?check_suite_focus=true#step:6:1028) unable to shut down Ganache, killing the test-runner that needed to be manually cancelled.

```
@truffle/environment: ✔ Setting up box
@truffle/environment:     ✓ Environment.develop overwrites the network_id of the network
@truffle/environment:   2 passing (3s)
@truffle/environment:   1 failing
@truffle/environment:   1) connectOrStart test network
@truffle/environment:        connects to an established Ganache instance:
@truffle/environment:      Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/home/runner/work/truffle/truffle/packages/environment/test/connectOrStartTest.js)
@truffle/environment:       at listOnTimeout (internal/timers.js:554:17)
@truffle/environment:       at processTimers (internal/timers.js:497:7)
Error: The operation was canceled.
```

packages/environment/test/connectOrStartTest.js
```diff
@@ -0,0 +1,51 @@
...
+        connection.disconnect();
+      }
+    }
+  });
+
+  it("connects to an established Ganache instance", async function () {
+    let connectionOneDisconnect, connectionTwoDisconnect;
+
+    try {
+      //establish a connection
```
lsqproduction, line 29: Does this starts the connection and disconnect it?

cds-amal: I'm not sure what you're asking. This test verifies subsequent calls to `Develop.connectOrStart` after a connection has been established, will connect instead of starting the underlying service. It does start a connection and eventually cleans it up in the `finally` block at the end of the test.

haltman-at:
Oh hey a quick note, the `.eslintrc.json` changes are no longer necessary as of #4698; I'd suggest rebasing if you haven't and removing that change because it's redundant.

packages/environment/test/connectOrStartTest.js
lsqproduction: ah ok. got it. thank you.

cds-amal:
> Oh hey a quick note, the `.eslintrc.json` changes are no longer necessary as of #4698; I'd suggest rebasing if you haven't and removing that change because it's redundant.

Good catch! rebasing.

packages/environment/test/develop.js
```diff
@@ -29,14 +29,10 @@ describe("Environment develop", function () {
         gas: expectedNetwork.gas
       }
     };
-    sinon.stub(Environment, "detect");
-  });
-
-  afterEach("Restore Environment detect", async function () {
```
haltman-at, line 35: Is there a reason for deleting this?  Do we not want to restore after each test?

packages/environment/test/develop.js
```diff
@@ -29,14 +29,10 @@ describe("Environment develop", function () {
...
-    Environment.detect.restore();
   });
 
-  it("Environment.develop overwrites the network_id of the network", async function() {
+  it("Environment.develop overwrites the network_id of the network", async function () {
+    sinon.stub(Environment, "detect");
```
haltman-at, line 35: I'm confused why this is being stubbed... what is this for?

haltman-at (commented):
Seems mostly good to me (as best I can understand anyway); just two questions about the tests.

packages/environment/test/develop.js
haltman-at: Oh I guess we don't want to restore if it's not a stub in the first place...

packages/environment/test/develop.js
cds-amal: `Environment.develop()` invokes `Environment.detect()` as its final step, which isn't needed to validate the test's assertion.  I'll add a clarifying comment and leave potential refactoring for followup work.

packages/environment/test/develop.js
cds-amal: I undid the commit that did this factor because it was unrelated to the race.

eggplantzzz (approved):
lookin good

packages/environment/test/develop.js
haltman-at: Oh, huh, OK.  Odd use and I'd suggest getting rid of it, but I guess it's OK to put that off? <shrug>

haltman-at (approved):
OK, looks good to me!

packages/environment/test/develop.js
cds-amal: This is one of those things that touches TruffleConfig's side-effect behavior. There are a bunch of methods in `Environment` that modify `TruffleConfig` :/ . A code smell to be addressed later.

PR #4674, opened 2022-01-30
use resolved promises

cds-amal:
Quick test to evaluate side-effects of using resolved promises


haltman-at:
My opinion is we should go with #4673 instead.  I don't think it makes sense for us to allow people to drop `await` here, but nowhere else; this is just going to cause confusion when people take code that works here and try to use it elsewhere.  And really I don't think we should allow dropping `await` in general!  Because that will also cause confusion, as it will encourage conflation between a promise and its result.

Issue #4368, opened 2021-10-15
There should be `DecodingInspector` analogous to `ResultInspector` (and test and db-kit should use it)


haltman-at:
## Issue

Currently we display decodings from Truffle Decoder / Codec in various ad-hoc ways in Truffle Test (events), db-kit (events and calldata), and the debugger (returndata).  There should be a `DecodingInspector` class that does for these what `ResultInspector` does for individual variables, and these should (at least mostly) be replaced by that.  Alternatively, the `ResultInpsector` class itself could be expanded to cover this case, but that seems more awkward.

(Also, I'm not *entirely* confident that the debugger's handling of return values makes sense to be unified with this; maybe it should, maybe it shouldn't, we'll have to see.)

## Environment

* Truffle version (`truffle version`): 5.4.15

lsqproduction:
Addressed in v5.5.2.

PR #4681, opened 2022-02-02
Enhancement: Create decoding inspectors (and use them in some places)

haltman-at:
Addresses #4368, or mostly does, anyway.

This PR creates three decoding inspector classes, `CalldataDecodingInspector`, `LogDecodingInspector`, and `ReturndataDecodingInspector`, analogous to `ResultInspector`.  I put these in `export.ts` rather than `format/utils/inspect.ts` as I think `export.ts` is basically the new location for this stuff.

In addition, I updated various code to use these inspectors rather than their own result-inspecting logic.  However I didn't update *all* code this way.  I'll discuss this more in a bit.

First off, I updated Truffle Test to use the `LogDecodingInspector` for event printing.  The logic for the `LogDecodingInspector` is basically copied from how Truffle Test already prints events.

Second, I updated the debugger CLI to use the `ReturndataDecodingInspector` for *some* things; specifically, custom errors and returned bytecode.  I left it alone for other things, as in many cases the debugger wants to do something a bit more specialized than what a generic `ReturndataDecodingInspector` can reasonably expected to do.  I mostly don't think this is a problem, but there's one case that bugs me; I'll get back to this in a moment.

I did *not* update db-kit to use the `CalldataDecodingInspector`.  That's because the relevant part of db-kit is a bunch of tsx that I don't know how to deal with, and maybe it's just fine for it to do it's own thing, I don't know.

So, let me tell you about the part of this PR that bothers me, and that I'm hoping to get some advice on.

The debugger actually has two separate functions for printing the return from a transaction; there's `printReturnValue` and there's `printRevertMessage`.  The former prints any return value, including unsuccesful ones such as revert messages; the latter is more specialized and only deals with unsuccessful returns.  They're invoked in different situations, to be clear, and in those cases that they both handle (unsuccessful returns) they print slightly different things.

So, let's talk about the `ReturndataDecodingInspector` and custom errors.

Because data returned could be any of a number of things, I decided that the `ReturndataDecodingInspector` would generally preface its result with an appropriate descriptor, such as `Returned values:` or `Error thrown:` or whatever.  For custom errors, I picked it to match up with what the debugger was already doing, so I could just use it in the debugger.  Except really I picked it to match up with what `printReturnValue` was already doing... leaving `printRevertMessage` in the lurch, because it printed that sort of header a bit differently.

The result is that `printRevertMessage` still uses `formatCustomError` from debug-utils, which is now basically duplicated code.  I don't really like this, but I'm a little unsure what to do about it.  Possibilities:

1. Just leave the duplicated code in.  I don't like this for obvious reasons.
2. Just use `ReturndataDecodingInspector` anyway; sure, the printouts don't agree at the moment, but who says they can't?  I don't really like this because I wrote them differently for a reason.
3. Use `ReturndataDecodingInspector`, and then do a search/replace.  This is hella hacky; it means that altering one message will break the other.
4. Alter `ReturndataDecodingInspector` to not put those headers.  Seems kind of unsatisfactory for anyone using `ReturndataDecoingInspector` on its own.
5. The `ReturndataDecodingInspector` relies on a function called `formatFunctionLike` for formatting custom errors (the other inspectors use this too for formatting things that look like functions).  Right now that function isn't exported because I meant it to be an internal thing.  I could export it, though, and then `printRevertMessage` could invoke it directly, and we could get rid of the duplicated code in `formatCustomError` (or we could alter `formatCustomError` to use it).  I'm not sure I like exposing that function, but, this is what I'm currently leaning towards as the cleanest solution.

Thoughts?

### DIFF
--- a/packages/codec/lib/export.ts
+++ b/packages/codec/lib/export.ts
@@ -1,22 +1,30 @@
 import debugModule from "debug";
 const debug = debugModule("codec:export");
 
-import type * as Format from "@truffle/codec/format";
+import OS from "os";
+import util from "util";
+import * as Format from "@truffle/codec/format";
 import type {
+  CalldataDecoding,
   LogDecoding,
-  ReturndataDecoding
+  ReturndataDecoding,
+  AbiArgument
 } from "@truffle/codec/types";
 import * as Conversion from "@truffle/codec/conversion";
 
-import { ResultInspector, unsafeNativize } from "@truffle/codec/format/utils/inspect";
+import {
+  ResultInspector,
+  unsafeNativize,
+  InspectOptions
+} from "@truffle/codec/format/utils/inspect";
 export { ResultInspector, unsafeNativize };
 
-type NumberFormatter = (n: BigInt) => any //not parameterized since we output any anyway
+type NumberFormatter = (n: BigInt) => any; //not parameterized since we output any anyway
 
 //HACK; this was going to be parameterized
 //but TypeScript didn't like that, so, whatever
 interface MixedArray extends Array<any> {
-  [key: string]: any
+  [key: string]: any;
 }
 
 /**
@@ -114,7 +122,8 @@ function ethersCompatibleNativize(
           )).value.asBN;
           return numberFormatter(Conversion.toBigInt(asBN));
         case "enum":
-          const numericAsBN = (<Format.Values.EnumValue>(result)).value.numericAsBN;
+          const numericAsBN = (<Format.Values.EnumValue>result).value
+            .numericAsBN;
           return numberFormatter(Conversion.toBigInt(numericAsBN));
         case "bool":
           return (<Format.Values.BoolValue>result).value.asBoolean;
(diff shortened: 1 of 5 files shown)


gnidan (approved):
Looks good!

haltman-at:
But what about the code duplication I pointed out...?  You really think I should leave that as-is...?

gnidan:
> But what about the code duplication I pointed out...? You really think I should leave that as-is...?

Oh weird; the GitHub app on my phone only showed me the first two paragraphs of your description. I'll retract my approve until I can get the chance to review with the full description in mind.

packages/codec/lib/export.ts
```diff
@@ -272,3 +282,212 @@ function ethersCompatibleNativizeEventArgs(
...
+              return prefix + formatted;
+            }
+          );
+          return `Immutable values:${OS.EOL}${formattedValues.join(OS.EOL)}`;
+        } else {
+          return firstLine;
+        }
+      case "unknownbytecode":
+        return "Bytecode was returned, but it could not be identified.";
+    }
```
eggplantzzz, line 446: Is there any reason to include a `default` case in these? It could act as a check, but then again will TS catch these errors anyways?

eggplantzzz: Well perhaps you could consider adding a `default` that throws?

dongmingh (approved):
looks good to me

dongmingh:
Regarding the code duplication, I like option 5 to make `formatFunctionLike` public so that `printRevertMessage` can invoke it directly and the duplicated code in `formatCustomError` can be removed.  This option seems to be cleaner.

haltman-at:
OK, I'll probably go with that then, thanks!  (Note: If you think this change is sufficiently important that you want to dismiss your approval until I've done that, you can click on the "1 approval" at the bottom, and click on your approval, and there is a button to dismiss your review.  This can be useful sometimes when a PR changes substantially after having been approved earlier.)

haltman-at:
OK, I've updated this PR with the deduplication.  Note that this makes this technically a breaking change to debug-utils... I guess I'll mark it as such?

haltman-at:
Oh wait, I can alter this to be nonbreaking, by adding an `indent` parameter to `formatFunctionLike`.  I'll go do that.

haltman-at:
@dongmingh Just want to check this is OK to go now?  Since you didn't rereview it.  (If you don't respond soon I'm just going to take your leaving it approved as saying that's OK and just go ahead and merge it.)

PR #4682, opened 2022-02-02
Bug fix: Fix decoding of bytecode with internal function pointers in immutables

haltman-at:
Here's a problem I noticed today: Codec will fail to decode bytecode returned from a constructor when it contains immutables of `function internal` type.  I was debugging a transaction that did this (a test transaction I made, obviously, not something on mainnet or something) and I noticed it was failing to decode the bytecode at the end.

Why did this occur?  Well, we use strict mode for decoding immutables; this is convenient for various reasons.  (Like, it allows for us to distinguish retryable errors and fall back into ABI mode if one occurs; this would be much more difficult without strict mode.)  But there's a problem -- strict mode was meant for ABI decoding; it's actually called `strictAbi` mode internally!  So, I made it so that codec will refuse to decode internal function pointers in strict mode, because those don't go in the ABI.

This presents a problem here.  We could turn off strict mode for immutables, but then we have to handle the retry problem a different way, which gets ugly quickly.  We could split strict mode in two, but, uh, that's a fair bit of work.  So, I figured the simplest solution was to just allow strict mode to decode internal function pointers.  It's a bit of an abuse of strict mode, but, I mean, the case my guard was *intended* to catch (unlike the one it actually caught!) is a can't-happen situation, so, I don't see the harm.

I did have to add a bit more boilerplate to internal function pointer as a result, as now it has to account for both strict mode and loose mode.  But that's no big deal.  (Possibly we could factor this boilerplate error-handling, like we factored some of the other boilerplate error-handling in codec, but it didn't seem worth it to me.)

That said, just changing this didn't fully fix the problem, because, oops, the debugger and decoder weren't passing in the internal functions table when decoding return values!  So I had the debugger do that.  I also added a test of this; or rather, expanded our existing test of decoding returned bytecode to include an immutable internal function pointer. :)

What about the decoder?  Well, the decoder doesn't currently support decoding returned bytecode at all, so, I made no changes there; if it did, though, then I'd have to move the setting up of the internal functions table from the instance decoder to the contract decoder.  (By which I mean, I actually *did* that and got it to compile before realizing it was pointless, so, uh, at least now I know that it's possible should it come up in the future.  I could go back and add that change in if you really want some future-proofing, but be aware that that would be adding untested code...)

Also, while I was at it, I changed some `let`s to `const`s.  Annoyingly, `prettier` also made a bunch of changes, so let me highlight the real changes for you:

1. In `basic/decode/index.ts`, the real changes are the ones from lines 312-685 (in the new version); everything prior is either `prettier` or me changing `let` to `const`.
2. `data/sagas/index.js`, the real changes are the ones on line 152 and earlier.
3. In the tests, all the changes are real.

### DIFF
--- a/packages/codec/lib/basic/decode/index.ts
+++ b/packages/codec/lib/basic/decode/index.ts
@@ -43,7 +43,7 @@ export function* decodeBasic(
       if (!fullType.underlyingType) {
         const error = {
           kind: "UserDefinedTypeNotFoundError" as const,
-          type: fullType,
+          type: fullType
         };
         if (strict || options.allowRetry) {
           throw new StopDecodingError(error, true);
@@ -61,10 +61,13 @@ export function* decodeBasic(
         info,
         options
       );
-      switch (underlyingResult.kind) { //yes this switch is a little unnecessary :P
+      switch (
+        underlyingResult.kind //yes this switch is a little unnecessary :P
+      ) {
         case "value":
           //wrap the value and return
-          return <Format.Values.UserDefinedValueTypeValue>{ //no idea why need coercion here
+          return <Format.Values.UserDefinedValueTypeValue>{
+            //no idea why need coercion here
             type: fullType,
             kind: "value" as const,
             value: underlyingResult
@@ -76,7 +79,8 @@ export function* decodeBasic(
           //does not cause an error in the whole thing, but to do that here
           //would cause problems for the type system :-/
           //so we'll just be inconsistent
-          return <Format.Errors.UserDefinedValueTypeErrorResult>{ //TS is being bad again :-/
+          return <Format.Errors.UserDefinedValueTypeErrorResult>{
+            //TS is being bad again :-/
             type: fullType,
             kind: "error" as const,
             error: {
@@ -280,7 +284,7 @@ export function* decodeBasic(
       switch (dataType.visibility) {
         case "external":
           if (!checkPadding(bytes, dataType, paddingMode)) {
-            let error = {
+            const error = {
               kind: "FunctionExternalNonStackPaddingError" as const,
               paddingType: getPaddingType(dataType, paddingMode),
               raw: Conversion.toHexString(bytes)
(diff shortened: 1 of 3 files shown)


packages/codec/lib/basic/decode/index.ts
```diff
@@ -306,22 +310,23 @@ export function* decodeBasic(
...
+          //strict mode, which was meant for ABI decoding, but oh well.
           if (!checkPadding(bytes, dataType, paddingMode)) {
+            const error = {
+              kind: "FunctionInternalPaddingError" as const,
+              paddingType: getPaddingType(dataType, paddingMode),
+              raw: Conversion.toHexString(bytes)
+            };
+            if (strict) {
+              throw new StopDecodingError(error);
+            }
```
cds-amal, line 325: Does this throw need to be guarded by strict? The preceding comment block and this code is confusing. It's also likely I need to re-read your PR explanation :/

cds-amal (commented):
Just one question @haltman-at . I'll give this another pass later today.

packages/codec/lib/basic/decode/index.ts
haltman-at: Yes, all the throws need to be guarded by strict, because that's how the decoder works; when something goes wrong, its response depends on whether it's in strict mode or loose mode.  In strict mode it throws an error, but in loose mode, it just returns an `ErrorResult` instead.

Strict mode is used for things that could possibly have multiple interpretations; it's used for decoding events and for decoding returned data.  Loose mode is used for things that should only have one interpretation; it's used for decoding transactions and state variables.

packages/codec/lib/basic/decode/index.ts
```diff
@@ -306,22 +310,23 @@ export function* decodeBasic(
...
-            //internal functions don't go in the ABI!
-            //this should never happen, but just to be sure...
-            throw new StopDecodingError({
-              kind: "InternalFunctionInABIError" as const
-            });
-          }
+          //note: we no longer error here if strict is true, because we
+          //want to be able to use strict mode to decode immutables, and
+          //these can be immutables.  yes, this is a bit of an abuse of
+          //strict mode, which was meant for ABI decoding, but oh well.
```
cds-amal, line 316: Oh, I see! No long *immediately* error here! It makes sense now.

packages/codec/lib/basic/decode/index.ts
cds-amal: Thanks!

Aivolt1 (approved):

packages/codec/lib/basic/decode/index.ts
haltman-at: Ah, yes, sorry, I meant we no longer error at that particular point!  Do you think that's worth clarifying?

cds-amal: Yes, please.

cds-amal (approved):
Looks good to me!

packages/codec/lib/basic/decode/index.ts
haltman-at: Cool, I'll go rewrite this comment a little and then merge.

Issue #4408, opened 2021-11-08
Empty migration results in total cost being printed as `0 undefined`


haltman-at:
## Issue

Suppose you have a migrations file, but no actual migrations occur in it.  When you run `truffle migrate`, the total cost is printed as `0 undefined`, rather than `0 wei` or `0 ether` or just `0`.

## Steps to Reproduce

Create a new Truffle project, edit `1_initial_migrations.js` by removing the `deployer.deploy` line, then try migrating your project.

## Expected Behavior

The 0 costs should not have "undefined" displayed as their unit.

## Actual Results

```
Starting migrations...
======================
> Network name:    'development'
> Network id:      1636404659203
> Block gas limit: 12000000 (0xb71b00)


1_initial_migration.js
======================
   -------------------------------------
   > Total cost:                   0 undefined


Summary
=======
> Total deployments:   0
> Final cost:          0 undefined
```

## Environment

* Truffle version (`truffle version`): 5.4.18

haltman-at:
Oh hey @lsqproduction, process note, typically we don't close issues on the Truffle repo until the fix is released, rather than as soon as it's merged.  If you don't mind, I'm going to reopen this one so it can be closed properly when the release is done. :)

gnidan:
This is half-done and now released in [Truffle v5.4.20](https://github.com/trufflesuite/truffle/releases/tag/v5.4.20).

(Leaving this open because it's fixed for individual migrations but not for the total at the bottom)

cc @lsqproduction

hughes-ch:
Should be fixed with pull request #4684. I've also added a test to reproduce (and prevent regressions)

cds-amal:
This should be resolved in the latest [release](https://github.com/trufflesuite/truffle/releases/tag/v5.4.32)

PR #4684, opened 2022-02-02
Fixing issue #4408 (Empty migration results in total cost being printed as 0 undefined)

hughes-ch:
I've also added a regression test so it won't happen again.


haltman-at:
Note that's #4408 (mentioning in the text so we'll actually have a link).

haltman-at (approved):
Looks good to me, thanks for this!  Could we get a second reviewer on this?

cds-amal (approved):
Thanks for this @hughes-ch, appreciate the tests!

Issue #4678, opened 2022-02-01
`truffle version` should include the managed version of Ganache


cds-amal:
It would be useful information to surface the version of (managed) Ganache included with Truffle.

haltman-at:
Now it does!

PR #4686, opened 2022-02-03
Core: add managed ganache version info to the output of truffle version

lsqproduction:
Resolved issue #4678 by reading from Truffle's package.json for Ganache's version number.
Once Ganache adds .version, we can make a change to use that instead. 

Also, BUNDLE_VERSION is causing an eslint no-undef rule complaint, this variable is now on the package level eslintrc.json, should this be moved to global eslintrc.json?

### DIFF
--- a/packages/core/.eslintrc.json
+++ b/packages/core/.eslintrc.json
@@ -1,3 +1,6 @@
 {
-  "extends": ["../../.eslintrc.package.json"]
+  "extends": ["../../.eslintrc.package.json"],
+  "globals": {
+    "BUNDLE_VERSION":"readonly"
+  }
 }
--- a/packages/core/.eslintrc.json
+++ b/packages/core/.eslintrc.json
@@ -1,6 +1,6 @@
 {
   "extends": ["../../.eslintrc.package.json"],
   "globals": {
-    "BUNDLE_VERSION":"readonly"
+    "BUNDLE_VERSION": "readonly"
   }
 }
--- a/packages/core/lib/version.js
+++ b/packages/core/lib/version.js
@@ -60,9 +60,15 @@ const logWeb3 = (logger = console) => {
   logger.log(`Web3.js v${web3Version}`);
 };
 
+const logGanache = (logger = console) => {
+  const ganacheVersion = pkg.dependencies.ganache;
+  logger.log(`Ganache V${ganacheVersion}`);
+};
+
 const logAll = (logger = console, config) => {
   const versionInformation = info(config);
   logTruffle(logger, versionInformation);
+  logGanache(logger);
   logSolidity(logger, versionInformation, config);
   logNode(logger);
   logWeb3(logger);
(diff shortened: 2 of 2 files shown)


haltman-at:
> Also, BUNDLE_VERSION is causing an eslint no-undef rule complaint, this variable is now on the package level eslintrc.json, should this be moved to global eslintrc.json?

I think either way makes sense, but let @cds-amal know if you're doing the latter, because he's also dealing with this on #4671.

haltman-at (approved):
Looks good to me!  Might want to check about what to do about `BUNDLE_VERSION` before merging though.

cds-amal:
> Looks good to me! Might want to check about what to do about `BUNDLE_VERSION` before merging though.

#4671 uses `BUNDLE_CHAIN_FILENAME` but `BUNDLE_VERSION` is used in three places, resolver, core, and truffle (for webpack). I don't think this warrants moving up to global, but also don't feel strongly about that position.

cds-amal (approved):
LGTM, thanks @lsqproduction !

haltman-at:
Oh, I missed the distinction there, thanks!  In that case my opinion is that it should probably be moved up to global.  I don't know that it makes a big difference though.

packages/core/.eslintrc.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": ["../../.eslintrc.package.json"]
+  "extends": ["../../.eslintrc.package.json"],
+  "globals": {
+    "BUNDLE_VERSION":"readonly"
```
eggplantzzz, line 4: just a nit, can you add a space after the `:`?

packages/core/lib/version.js
```diff
@@ -60,9 +60,15 @@ const logWeb3 = (logger = console) => {
   logger.log(`Web3.js v${web3Version}`);
 };
 
+const logGanache = (logger = console) => {
+  const ganacheVersion = pkg.dependencies.ganache;
+  logger.log(`Ganache V${ganacheVersion}`);
```
eggplantzzz, line 65: one more nit, can you make the `V` lowercase so it matches the format that we print web3's version in?

packages/core/.eslintrc.json
lsqproduction: okie thank you for catching it.

packages/core/lib/version.js
lsqproduction: okie I will make the change.

lsqproduction:
> 



> > Looks good to me! Might want to check about what to do about `BUNDLE_VERSION` before merging though.
> 
> #4671 uses `BUNDLE_CHAIN_FILENAME` but `BUNDLE_VERSION` is used in three places, resolver, core, and truffle (for webpack). I don't think this warrants moving up to global, but also don't feel strongly about that position.
I will keep it at the package level then.

PR #4694, opened 2022-02-04
Dedupe slack notification channel

cds-amal:
This PR updates the Slack notification target channel in order to consolidate our GitHub notifications. 

As we evaluate the Slack Integration, we may eventually need to turn this one off. We want to ensure all job runs are captured, but it is unclear if the Slack integration can provide that coverage.

Changes
- change notification channel to #truffle-ci-notifications
- update fields

N.B. secretes.SLACK_WEBHOOK_URL also has to be updated here: https://trufflesuite.slack.com/apps/A0F7XDUAZ-incoming-webhooks\?tab\=more_info

## how to test

Watch the #truffle-ci-integration channel and see if things are integrated.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -61,7 +61,7 @@ jobs:
         with:
           channel: '#truffle-ci-notifications'
           status: ${{ job.status }}
-          fields: workflow,job,commit,message,ref,author,pullRequest
+          fields: commit,author,pullRequest
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


haltman-at (approved):
Makes sense to me.  Why the removed fields, I'm wondering?

cds-amal:
> Makes sense to me. Why the removed fields, I'm wondering?

Mostly to reduce noise and remove redundancies
- `took` reported the duration run-time for the notification action which isn't meaningful in my opinion
- `message` and `commit` kinda resolved to the same information, but commit presents it better
- `repo` is always going to be truffle
- `eventName` will always be pull_request
- adding `pullRequest` gives us the link to the PR

PR #4698, opened 2022-02-05
Internal improvement: Put all BUNDLE globals in main .eslintrc.json

haltman-at:
What with @lsqproduction dealing with `BUNDLE_VERSION` in #4686, and @cds-amal dealing with `BUNDLE_CHAIN_FILENAME` in #4671, I thought it was probably best to just take the whole list of these variables and stick them in the main `.eslintrc.json`, so we don't have to keep worrying about this (unless we add new ones of these!).

(Hope you don't mind @cds-amal!)

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,6 +4,13 @@
     "es2021": true,
     "node": true
   },
+  "globals": {
+    "BUNDLE_VERSION": "readonly",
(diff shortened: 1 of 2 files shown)


cds-amal (approved):
LGTM @haltman-at, I don't mind at all.

PR #4703, opened 2022-02-07
Remove unused lodash.pick types dependency

cds-amal:


haltman-at (approved):
Thanks for getting this!

Issue #4643, opened 2022-01-24
@truffle/fetch-and-compile provides no interface for reporting supported networks


gnidan:
## Issue

How does one know what networks are supported? You can't! Would be nice to remedy this.

## Steps to Reproduce

1. Read @truffle/fetch-and-compile's interface
2. Look for a way to determine supported networks
3. Observe this doesn't exist

## Expected Behavior

I'd expect some kind of interface for listing supported networks.

## Actual Results

There exists no interface for listing supported networks.

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.4.30
* node version (`node --version`):
* npm version (`npm --version`):

haltman-at:
Now it does!

PR #4705, opened 2022-02-08
Enhancement: Add getSupportedNetworks() to source-fetcher and fetch-and-compile

haltman-at:
Addresses #4643.  This PR adds a `getSupportedNetworks()` function to `@truffle/fetch-and-compile`, and a static `getSupportedNetworks()` method to each source fetcher in `@truffle/source-fetcher`.  It also adds a quick test of this functionality.

This PR is mostly pretty straightforward.  For the source-fetcher part, each source fetcher now has a static `getSupportedNetworks()` function.  I could have added an instance version too (like we have for `fetcherName`) but I didn't see a need so I didn't.  I moved the networks information out of `common.ts` and into a new file `networks.ts`.  For now it's pretty bare-bones and continues to make the assumption that network ID equals chain ID.

The networks are returned indexed by name, and each is given as an object with fields `name`, `networkId`, and `chainId`.  Right now those latter two will always be equal, but we may need to allow them to be unequal at some point in the future.

As for the fetch-and-compile part, its `getSupportedNetworks()` function is mostly similar.  Note that this function is not for a specific fetcher, but draws from all of them and combines the results together.  Well, usually it will draw from all of them -- it takes an optional config; if a config is passed in, its `sourceFetchers` field will control which fetchers are pulled from.  In addition, the networks returned by this function all have a `fetchers` field, which gives the names of the fetchers that support that network.

I also added a test, and also did some minor cleanup of various things while I was at it.  And that's it!

I'm intending to follow this up by adding support for more networks to both fetchers, but @gnidan thought this ought to be done first. :)

### DIFF
--- a/packages/fetch-and-compile/lib/fetch.ts
+++ b/packages/fetch-and-compile/lib/fetch.ts
@@ -28,8 +28,12 @@ export async function fetchAndCompileForRecognizer(
 }
 
 //sort/filter fetchers by user's order, if given; otherwise use default order
-function getSortedFetcherConstructors(config: Config): FetcherConstructor[] {
-  const userFetcherNames: string[] | undefined = config.sourceFetchers;
+export function getSortedFetcherConstructors(
+  config?: Config
+): FetcherConstructor[] {
+  const userFetcherNames: string[] | undefined = (
+    config || { sourceFetchers: undefined }
+  ).sourceFetchers;
   let sortedFetchers: FetcherConstructor[] = [];
   if (userFetcherNames) {
     for (let name of userFetcherNames) {
--- a/packages/fetch-and-compile/lib/types.ts
+++ b/packages/fetch-and-compile/lib/types.ts
@@ -1,5 +1,16 @@
 import type { WorkflowCompileResult } from "@truffle/compile-common";
-import type { SourceInfo } from "@truffle/source-fetcher";
+import type {
+  SourceInfo,
+  NetworkInfo as FetcherNetworkInfo
+} from "@truffle/source-fetcher";
+
+export interface NetworkInfo extends FetcherNetworkInfo {
+  fetchers: string[];
+}
+
+export interface NetworkInfos {
+  [name: string]: NetworkInfo;
+}
 
 export type FailureType = "fetch" | "compile" | "language";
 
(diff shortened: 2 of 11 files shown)


packages/fetch-and-compile/lib/types.ts
```diff
@@ -1,5 +1,16 @@
...
+import type {
+  SourceInfo,
+  NetworkInfo as FetcherNetworkInfo
+} from "@truffle/source-fetcher";
+
+export interface NetworkInfo extends FetcherNetworkInfo {
+  fetchers: string[];
+}
+
+export interface NetworkInfos {
```
gnidan, line 11: Why don't we just call this `SupportedNetworks`? Since it's the return type of `getSupportedNetworks()`?

packages/source-fetcher/lib/types.ts
```diff
@@ -16,6 +17,16 @@ export interface Fetcher {
   fetchSourcesForAddress(address: string): Promise<SourceInfo | null>;
 }
 
+export interface NetworkInfo {
+  name: string;
+  networkId: number;
+  chainId: number;
+}
+
+export interface NetworkInfos {
```
gnidan, line 26: (Same comment about name)

cds-amal (commented):
Does it make sense to have a fetcher versions that relate to external API services?

packages/fetch-and-compile/lib/types.ts
haltman-at: Sure, that would make sense.

haltman-at:
> Does it make sense to have a fetcher versions that relate to external API services?

I don't understand the question.  Can you expand on this?

packages/fetch-and-compile/lib/fetch.ts
```diff
@@ -28,8 +28,12 @@ export async function fetchAndCompileForRecognizer(
...
 
 //sort/filter fetchers by user's order, if given; otherwise use default order
-function getSortedFetcherConstructors(config: Config): FetcherConstructor[] {
-  const userFetcherNames: string[] | undefined = config.sourceFetchers;
+export function getSortedFetcherConstructors(
+  config?: Config
+): FetcherConstructor[] {
+  const userFetcherNames: string[] | undefined = (
+    config || { sourceFetchers: undefined }
+  ).sourceFetchers;
```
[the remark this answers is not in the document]
cds-amal: oh! right! Thanks

gnidan (approved):

PR #4706, opened 2022-02-08
Bug fix: Prevent source-map-utils from splitting surrogate pairs

haltman-at:
This function in `source-map-utils` used `source.split("")` to split `source` into characters.  This splits surrogate pairs.  So I changed it to `Array.from(source)`, which doesn't.

...of course, doing this still doesn't *really* solve the problem, as what we really want here is something more like grapheme clusters, not code points.  But solving that problem is harder and would presumably require introducing a dependency, so I thought I'd just throw up this PR quickly to start.  I could open an issue for that though?

### DIFF
--- a/packages/source-map-utils/index.js
+++ b/packages/source-map-utils/index.js
@@ -9,7 +9,7 @@ var SourceMapUtils = {
   getCharacterOffsetToLineAndColumnMapping: function (source) {
     var mapping = [];
 
-    source = source.split("");
+    source = Array.from(source);
 
     var line = 0;
     var column = 0;
@@ -422,9 +422,7 @@ var SourceMapUtils = {
       for (let offset = 0; offset < template.length; offset++) {
         const instruction = instructions[index + offset];
         const comparison = template[offset];
-        if (
-          !instruction || instruction.name !== comparison.name
-        ) {
+        if (!instruction || instruction.name !== comparison.name) {
           return false;
         }
         if (
--- a/packages/source-map-utils/index.js
+++ b/packages/source-map-utils/index.js
@@ -9,7 +9,9 @@ var SourceMapUtils = {
   getCharacterOffsetToLineAndColumnMapping: function (source) {
     var mapping = [];
 
-    source = Array.from(source);
+    source = Array.from(source); //note: this will correctly handle
+    //surrogate pairs, but there's still the problem of grapheme
+    //clusters!  We should do something about that later.
 
     var line = 0;
     var column = 0;


packages/source-map-utils/index.js
```diff
@@ -9,7 +9,7 @@ var SourceMapUtils = {
   getCharacterOffsetToLineAndColumnMapping: function (source) {
     var mapping = [];
 
-    source = source.split("");
+    source = Array.from(source);
```
cds-amal, line 12: This should have a comment to capture the surrogate pair concern, as well as a test to validate the change.

cds-amal (changes requested):
I asked for a test because I'm unsure of the context and couldn't find the bug/issue it fixes. Does this have to do with Unicode surrogate pairs? I think a test would add context and documentation.

cds-amal:
> ...of course, doing this still doesn't really solve the problem, as what we really want here is something more like grapheme clusters, not code points. But solving that problem is harder and would presumably require introducing a dependency, so I thought I'd just throw up this PR quickly to start. I could open an issue for that though?

Please open the issue @haltman-at.  I''d love to learn more!

packages/source-map-utils/index.js
haltman-at: I'm not sure this really needs a comment saying that this *doesn't* split surrogate pairs when this is, like, just the right way to do it.  OTOH this could perhaps use a comment saying this still doesn't handle grapheme clusters?

haltman-at:
> I asked for a test because I'm unsure of the context and couldn't find the bug/issue it fixes.

Yes, there wasn't an open issue for this, I just noticed this while going through some code with @dongmingh and threw up a quick PR for it.

>  Does this have to do with Unicode surrogate pairs?

Yes.  So, let's say you have a string `s` and you want to split into characters.  Because Javascript was designed back in the days of 16-bit Unicode, before surrogate pairs were a consideration, Javascript uses UTF-16 and considers one "character" to be one 16-bit scalar value.  So, `s.split("")` will not split up `s` into individual Unicode code points, but rather into 16-bit scalar values, even if that means splitting a surrogate pair into two surrogates.

Obviously, that's not what we want.  Fortunately, `Array.from(s)` came later and so it got defined to split things into UTF-16 code points, rather than 16-bit scalar values!

Of course, as mentioned, that still isn't really what we want due to combining characters and grapheme clusters, but it's an improvement, at least.

> I think a test would add context and documentation.

I'm not sure this really needs a test, it's fairly minor, but I guess I could add one if you really want.  Frankly, it'd be a little annoying because `source-map-utils` doesn't have tests at the moment, so either I'd have to set up mocha for it, or I'd have to test this indirectly via the debugger (which is how most of the stuff in here is currently tested).  It seems kind of overkill for just a straightforwardly correct change, where we change the way we split a string into characters from the wrong way that MDN will warn you about using, to the way that is specified to work correctly!

haltman-at:
> Please open the issue @haltman-at. I''d love to learn more!

OK, I'll open an issue once this is merged; no sense in opening it before that, as I see it.

packages/source-map-utils/index.js
cds-amal: Ah, that makes sense, relating to grapheme clusters.

cds-amal:
> > Does this have to do with Unicode surrogate pairs?
> 
> Yes. So, let's say you have a string `s` and you want to split into characters. Because Javascript was designed back in the days of 16-bit Unicode, before surrogate pairs were a consideration, Javascript uses UTF-16 and considers one "character" to be one 16-bit scalar value. So, `s.split("")` will not split up `s` into individual Unicode code points, but rather into 16-bit scalar values, even if that means splitting a surrogate pair into two surrogates.
> 

Thanks for the explanation, I didn't know this about `split`

> Obviously, that's not what we want. Fortunately, `Array.from(s)` came later and so it got defined to split things into UTF-16 code points, rather than 16-bit scalar values!
> 

TIL!  thank you.

> I'm not sure this really needs a test, it's fairly minor, but I guess I could add one if you really want. Frankly, it'd be a little annoying because `source-map-utils` doesn't have tests at the moment, so either I'd have to set up mocha for it, or I'd have to test this indirectly via the debugger (which is how most of the stuff in here is currently tested). It seems kind of overkill for just a straightforwardly correct change, where we change the way we split a string into characters from the wrong way that MDN will warn you about using, to the way that is specified to work correctly!

Sadly, I couldn't reason about the code change without the context you provided here and am satisfied that you tested this in the debugger. I'll defer to you on whether to add a mocha test :)

haltman-at:
> > Obviously, that's not what we want. Fortunately, Array.from(s) came later and so it got defined to split things into UTF-16 code points, rather than 16-bit scalar values!

> TIL! thank you.

I should probably note that strictly speaking this is not actually a feature of `Array.from()`, but rather of the string iterator, since `Array.from()` works with the iterator for whatever you pass in.  Thus for instance `[...s]` is another valid and equivalent way of doing this.

packages/source-map-utils/index.js
haltman-at: OK, I've added this comment.

cds-amal (approved):
Thanks @haltman-at. I'm glad you started this, let's get it in!

PR #4709, opened 2022-02-09
Update truffle/contract README examples

kevinweaver:
The truffle contract README includes some misleading examples.  I updated these and linted the file.  Happy to remove the linting if it's offensive.

I also updated a lot of instances of ```MyContract.deployed().then(function(instance) {...})``` to ```const instance = await MyContract.deployed();``` since that syntax is more common these days.

### DIFF
--- a/packages/contract/README.md
+++ b/packages/contract/README.md
@@ -17,17 +17,28 @@ $ npm install @truffle/contract
 
 ### Usage
 
-First, set up a new web3 provider instance and initialize your contract, then `require("@truffle/contract")`. The input to the `contract` function is a JSON blob defined by [@truffle/contract-schema](https://github.com/trufflesuite/truffle/tree/master/packages/contract-schema). This JSON blob is structured in a way that can be passed to all truffle-related projects.
+First, set up a new web3 provider instance and initialize your contract, then `require("@truffle/contract")`.
+The input to the `contract` function is a JSON blob defined by [@truffle/contract-schema](https://github.com/trufflesuite/truffle/tree/master/packages/contract-schema). This JSON blob is structured in a way that can be passed to all truffle-related projects.
 
 ```javascript
 const provider = new Web3.providers.HttpProvider("http://localhost:8545");
 const contract = require("@truffle/contract");
 
-// Artifact json previosly saved by @truffle/artifactor
+const MyContract = contract({
+  abi: [...], // minimum required
+  address: "0x...", // optional
+  // many more
+});
+MyContract.setProvider(provider);
+```
+
+Instead of providing these values manually though, the contract-scheme matches the output of the @truffle/artifactor
+
+```javascript
+// json output provided by @truffle/artifactor
 const contractArtifact = require("./path/to/contractArtifact.json");
 
 const MyContract = contract(contractArtifact);
-MyContract.setProvider(provider);
 ```
 
 You now have access to the following functions on `MyContract`, as well as many others:
--- a/packages/contract/README.md
+++ b/packages/contract/README.md
@@ -22,14 +22,14 @@ The contract-schema matches the output of the [@truffle/artifactor](https://gith
 
 ```javascript
 const provider = new Web3.providers.HttpProvider("http://localhost:8545");
-const MyContract = require("./path/to/contractArtifact.json"); //produced by truffle compile
+const MyContract = require("./path/to/contractArtifact.json"); //produced by Truffle compile
 const contract = require("@truffle/contract");
 
 const instance = contract(MyContract);
 ```
 
-If you aren't using truffle artifacts, the input to the `contract` function is a JSON blob defined by [@truffle/contract-schema](https://github.com/trufflesuite/truffle/tree/master/packages/contract-schema). This JSON blob is structured in a way that can be passed to all Truffle-related projects.
-Minimally, it requires the contract ABI, but passing this alone may cause issues using certain features of truffle contract.
+If you aren't using Truffle artifacts, the input to the `contract` function is a JSON blob defined by [@truffle/contract-schema](https://github.com/trufflesuite/truffle/tree/master/packages/contract-schema). This JSON blob is structured in a way that can be passed to all Truffle-related projects.
+Minimally, it requires the contract ABI, but passing this alone may cause issues using certain features of Truffle contract.
 
 ```javascript
 const instance = contract({
--- a/packages/contract/README.md
+++ b/packages/contract/README.md
@@ -40,14 +40,10 @@ Each instance is tied to a specific address on the Ethereum network, and each in
 
 ```javascript
 let deployed;
-MyContract.deployed()
-  .then(function (instance) {
-    deployed = instance;
-    return instance.someFunction(5);
-  })
-  .then(function (result) {
-    // Do something with the result or continue with more transactions.
-  });
+const instance = await MyContract.deployed();
+deployed = instance;
+const result = await instance.someFunction(5);
+// Do something with the result or continue with more transactions.
 ```
 
 or equivalently in ES6 <sup>(node.js 8 or newer)</sup>:
@@ -250,10 +246,8 @@ From Javascript's point of view, this contract has three functions: `setValue`, 
 When we call `setValue()`, this creates a transaction. From Javascript:
 
 ```javascript
-instance.setValue(5).then(function (result) {
-  // result object contains import information about the transaction
-  console.log("Value was set to", result.logs[0].args.val);
-});
+const result = instance.setValue(5);
+console.log("Value was set to", result.logs[0].args.val);
 ```
 
 The result object that gets returned looks like this:
@@ -292,7 +286,7 @@ in the `logs` array.
 We can call `setValue()` without creating a transaction by explicitly using `.call`:
 
 ```javascript
-instance.setValue.call(5).then(...);
+const result = await instance.setValue.call(5);
 ```
 
 This isn't very useful in this case, since `setValue()` sets things, and the value we pass won't be saved since we're not creating a transaction.
@@ -302,19 +296,17 @@ This isn't very useful in this case, since `setValue()` sets things, and the val
 However, we can _get_ the value using `getValue()`, using `.call()`. Calls are always free and don't cost any Ether, so they're good for calling functions that read data off the blockchain:
 
 ```javascript
-instance.getValue.call().then(function (val) {
-  // val represents the `value` storage object in the solidity contract
-  // since the contract returns that value.
-});
+const val = instance.getValue.call();
+// val represents the `value` storage object in the solidity contract
+// since the contract returns that value.
 ```
 
 Even more helpful, however is we _don't even need_ to use `.call` when a function is marked as `view` or `pure`, because `@truffle/contract` will automatically know that that function can only be interacted with via a call:
 
 ```javascript
-instance.getValue().then(function (val) {
-  // val reprsents the `value` storage object in the solidity contract
-  // since the contract returns that value.
-});
+const val = instance.getValue();
+// val reprsents the `value` storage object in the solidity contract
+// since the contract returns that value.
 ```
 
 #### Processing transaction results
@@ -322,11 +314,10 @@ instance.getValue().then(function (val) {
 When you make a transaction, you're given a `result` object that gives you a wealth of information about the transaction. You're given the transaction hash (`result.tx`), the decoded events (also known as logs; `result.logs`), and a transaction receipt (`result.receipt`). In the below example, you'll recieve the `ValueSet()` event because you triggered the event using the `setValue()` function:
 
 ```javascript
-instance.setValue(5).then(function (result) {
-  // result.tx => transaction hash, string
-  // result.logs => array of trigger events (1 item in this case)
-  // result.receipt => receipt object
-});
+const result = await instance.setValue(5);
+// result.tx => transaction hash, string
+// result.logs => array of trigger events (1 item in this case)
+// result.receipt => receipt object
 ```
 
 #### Sending Ether / Triggering the fallback function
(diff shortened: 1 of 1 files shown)


packages/contract/README.md
```diff
@@ -10,51 +10,50 @@ $ npm install @truffle/contract
...
 
-var MyContract = contract({
-  abi: ...,
-  address: ..., // optional
-  // many more
-})
+// Artifact json previosly saved by @truffle/artifactor
+const contractArtifact = require("./path/to/contractArtifact.json");
+
+const MyContract = contract(contractArtifact);
```
kevinweaver, line 29: This json blob threw me off and made me think I could simply pass the ABI without running it through the artifactor.

packages/contract/README.md
```diff
@@ -80,51 +79,47 @@ v4.0.2 or higher.
 Let's use `@truffle/contract` with an example contract from [Dapps For Beginners](https://dappsforbeginners.wordpress.com/tutorials/your-first-dapp/). In this case, the abstraction has been saved to a `.sol` file by [@truffle/artifactor](https://github.com/trufflesuite/truffle/tree/master/packages/artifactor):
 
 ```javascript
+const contract = require("@truffle/contract");
 // Require the package that was previosly saved by @truffle/artifactor
-var MetaCoin = require("./path/to/MetaCoin.json");
+const metacoinArtifact = require("./path/to/MetaCoin.json");
+const metacoin = contract(metacoinArtifact)
```
kevinweaver, line 87: Happy to change these variable names if anyone has alternatives

packages/contract/README.md
```diff
@@ -80,51 +79,47 @@ v4.0.2 or higher.
...
-  alert("Balance of account two is " + balance_of_account_two + "!") // => 1.5
-}).catch(function(err) {
+  result = await coin.sendCoin(account_one, 1.5, { from: account_two });
+  balance_of_account_two = await coin.balances.call(account_two);
+  console.log("Balance of account two is " + balance_of_account_two + "!"); // => 1.5
+} catch(function (err) {
   // Easily catch all errors along the whole execution.
   alert("ERROR! " + err.message);
-});
+};
```
kevinweaver, line 124: Decided to wrap this in a try catch to keep this error handling.

packages/contract/README.md
```diff
@@ -80,51 +79,47 @@ v4.0.2 or higher.
...
-MetaCoin.at(contract_address).then(function(instance) {
-  coin = instance;
+const coin = await metacoin.at(contract_address)
 
+try {
   // Make a transaction that calls the function `sendCoin`, sending 3 MetaCoin
   // to the account listed as account_two.
-  return coin.sendCoin(account_two, 3, {from: account_one});
-}).then(function(result) {
+  let result = await coin.sendCoin(account_two, 3, { from: account_one });
```
kevinweaver, line 105: Decided to replace all of the `then()`s with individual `await`s.

packages/contract/README.md
haltman-at: Syntax is off here -- you're got `catch(function (err) {` instead of `catch (err) {`

packages/contract/README.md
haltman-at: I think using a capital would be more in line with convention here, since the artifact represents a class rather than an instance.

packages/contract/README.md
eggplantzzz: it would be nice if we could explicitly define what the minimum is that you'll need for an artfact - and the consequences of including less information than the entire artifactor output

packages/contract/README.md
eggplantzzz: Yeah, `MetaCoin` for the Truffle Contract object and something like `instance` for the deployed instance is convention I think.

packages/contract/README.md
cds-amal: Second Tyler's suggestion

cds-amal: If we're referring to the shape of the JSON `contract` needs I think it's worth noting in the documentation. I suspect it will at least match [what web3's contract abstraction needs](https://web3js.readthedocs.io/en/v1.7.0/glossary.html#glossary-json-interface)

haltman-at: My understanding is the absolute *minimum* you need is an ABI.  The problem with showing that, though, is that it might confuse people into passing `abi` instead of `{ abi }`, the exact problem we're trying to solve!

kevinweaver: This PR came about because I attempted to pass: 
```
{
  abi: ...,
  address: ..., 
}
```
as the old README suggested and it didn't work.  It was about a month ago at this point, and it's very possible I was doing something wrong, though.  I'll try it again and see if I hit the same issue.

kevinweaver: I tried passing only the abi and it worked, so it was definitely user error on my part.  Instead of removing the original example, I added a second on here to show that you can additionally use the output of the artifactor.  Let me know if any of you think of a more concise way of explaining this.

packages/contract/README.md
kevinweaver: 👍

kevinweaver:
I think this is ready for re-review when you get a moment @haltman-at @eggplantzzz @cds-amal, thanks!

packages/contract/README.md
```diff
@@ -10,51 +10,58 @@ $ npm install @truffle/contract
...
+- Synchronized transactions for better control flow (i.e., transactions won't finish until you're guaranteed they've been mined).
+- Promises. No more callback hell. Works well with `ES6` and `async/await`.
+- Default values for transactions, like `from` address or `gas`.
+- Returning logs, transaction receipt and transaction hash of every synchronized transaction.
 
 ### Usage
 
-First, set up a new web3 provider instance and initialize your contract, then `require("@truffle/contract")`. The input to the `contract` function is a JSON blob defined by [@truffle/contract-schema](https://github.com/trufflesuite/truffle/tree/master/packages/contract-schema). This JSON blob is structured in a way that can be passed to all truffle-related projects.
+First, set up a new web3 provider instance and initialize your contract, then `require("@truffle/contract")`.
+The contract-schema matches the output of the [@truffle/artifactor](https://github.com/trufflesuite/truffle/tree/master/packages/artifactor), so you can pass in the artifact file truffle produces when you run `truffle compile`.
```
eggplantzzz, line 21: This still seems kind of unclear to me. Is contract-schema a thing I don't know about? Shouldn't we be talking about the input to the @truffle/contract default export?

eggplantzzz: Ahh I see you moved the other stuff down below. But is contract-schema a thing?

eggplantzzz, line 21: Probably capitalize Truffle.

packages/contract/README.md
```diff
@@ -10,51 +10,58 @@ $ npm install @truffle/contract
...
+  abi: [...], // minimum required
+  address: "0x...", // optional
   // many more
-})
-MyContract.setProvider(provider);
+});
+instance.setProvider(provider);
 ```
 
 You now have access to the following functions on `MyContract`, as well as many others:
```
eggplantzzz, line 43: You'll need to change this to `instance` like you did above.

packages/contract/README.md
```diff
@@ -10,51 +10,58 @@ $ npm install @truffle/contract
...
 
-* `at()`: Create an instance of `MyContract` that represents your contract at a specific address.
-* `deployed()`: Create an instance of `MyContract` that represents the default address managed by `MyContract`.
-* `new()`: Deploy a new version of this contract to the network, getting an instance of `MyContract` that represents the newly deployed instance.
+- `at()`: Create an instance of `MyContract` that represents your contract at a specific address.
```
eggplantzzz, line 45: Same with changing `MyContract` to `instance`.

packages/contract/README.md
```diff
@@ -10,51 +10,58 @@ $ npm install @truffle/contract
...
+- `deployed()`: Create an instance of `MyContract` that represents the default address managed by `MyContract`.
```
eggplantzzz, line 46: same

packages/contract/README.md
```diff
@@ -10,51 +10,58 @@ $ npm install @truffle/contract
...
+- `new()`: Deploy a new version of this contract to the network, getting an instance of `MyContract` that represents the newly deployed instance.
```
eggplantzzz, line 47: same

packages/contract/README.md
```diff
@@ -80,51 +87,47 @@ v4.0.2 or higher.
...
 
 // Remember to set the Web3 provider (see above).
-MetaCoin.setProvider(provider);
+instance.setProvider(provider);
 
 // In this scenario, two users will send MetaCoin back and forth, showing
 // how @truffle/contract allows for easy control flow.
-var account_one = "5b42bd01ff...";
-var account_two = "e1fd0d4a52...";
+const account_one = "5b42bd01ff...";
```
eggplantzzz, line 100: Would you be willing to re-write all of these variable names in camel case? That is the convention followed now in Truffle.

packages/contract/README.md
kevinweaver: 👍 The sentences below no longer made sense, like "Create an instance of `instance` that represents the default address managed by `instance`", so I updated the example to:

```
const MyContract = require("./path/to/contractArtifact.json"); //produced by truffle compile
const contract = require("@truffle/contract");

const instance = contract(MyContract);
```
That way, we can leave "MyContract" here, and retain `instance` in the code and have them make sense.

packages/contract/README.md
kevinweaver: responded above

packages/contract/README.md
kevinweaver: responded above

packages/contract/README.md
kevinweaver: responded above

packages/contract/README.md
eggplantzzz: I don't see your response above but I'm assuming you decided to revert back to `MyContract`.

packages/contract/README.md
```diff
@@ -80,51 +80,47 @@ v4.0.2 or higher.
 Let's use `@truffle/contract` with an example contract from [Dapps For Beginners](https://dappsforbeginners.wordpress.com/tutorials/your-first-dapp/). In this case, the abstraction has been saved to a `.sol` file by [@truffle/artifactor](https://github.com/trufflesuite/truffle/tree/master/packages/artifactor):
 
 ```javascript
+const contract = require("@truffle/contract");
 // Require the package that was previosly saved by @truffle/artifactor
-var MetaCoin = require("./path/to/MetaCoin.json");
+const MetacoinArtifact = require("./path/to/MetaCoin.json");
+const instance = contract(MetacoinArtifact);
```
haltman-at, line 86: I might be mistaken, but won't using `contract` like this get you something for the class, not the instance?

packages/contract/README.md
```diff
@@ -10,51 +10,51 @@ $ npm install @truffle/contract
...
+const MyContract = require("./path/to/contractArtifact.json"); //produced by truffle compile
+const contract = require("@truffle/contract");
 
-var MyContract = contract({
-  abi: ...,
-  address: ..., // optional
-  // many more
-})
-MyContract.setProvider(provider);
+const instance = contract(MyContract);
```
haltman-at, line 28: Again the instance vs constructor question.

packages/contract/README.md
```diff
@@ -10,51 +10,51 @@ $ npm install @truffle/contract
...
 
-You now have access to the following functions on `MyContract`, as well as many others:
+If you aren't using truffle artifacts, the input to the `contract` function is a JSON blob defined by [@truffle/contract-schema](https://github.com/trufflesuite/truffle/tree/master/packages/contract-schema). This JSON blob is structured in a way that can be passed to all Truffle-related projects.
+Minimally, it requires the contract ABI, but passing this alone may cause issues using certain features of truffle contract.
 
-* `at()`: Create an instance of `MyContract` that represents your contract at a specific address.
-* `deployed()`: Create an instance of `MyContract` that represents the default address managed by `MyContract`.
-* `new()`: Deploy a new version of this contract to the network, getting an instance of `MyContract` that represents the newly deployed instance.
+```javascript
+const instance = contract({
```
haltman-at, line 35: Hm, do you get an instance if address is set?  Do we need to show it both with and without address to demonstrate both instance and class usage?

packages/contract/README.md
haltman-at: Yeah `at` and etc is on constructors, not instances, right?

packages/contract/README.md
```diff
@@ -10,51 +10,51 @@ $ npm install @truffle/contract
...
 
-or equivalently in ES6 <sup>(node.js 8 or newer)</sup>:  
+Each instance is tied to a specific address on the Ethereum network, and each instance has a 1-to-1 mapping from Javascript functions to contract functions. For instance, if your Solidity contract had a function defined `someFunction(uint value) {}` (solidity), then you could execute that function on the network like so:
 
-  ```javascript
-  const instance = await MyContract.deployed();
-  const result = await instance.someFunction(5);  
-  ```  
+```javascript
+const deployed;
```
haltman-at, line 52: Illegal JS here; you can't do `const a; a=5;`, you need to use `let` for that.  Although, we don't really need this `deployed` variable at all, do we?

packages/contract/README.md
```diff
@@ -10,51 +10,51 @@ $ npm install @truffle/contract
...
+const instance = await MyContract.deployed();
```
haltman-at, line 53: Oh dear, name reuse here!  Above you use `instance` for what appears to be the class, but here `MyContract` is the class and `instance` is in fact the instance... better make this consistent!

haltman-at (commented):
Confused about some instance vs class and variable naming stuff.

packages/contract/README.md
kevinweaver: ohh good catch, I'll fix, thanks!

packages/contract/README.md
kevinweaver: I'm not even sure why I added these... deleted.

packages/contract/README.md
kevinweaver: Oops, I misunderstood this point. I believe it should be fixed now.

packages/contract/README.md
kevinweaver: Fixed this naming

haltman-at (approved):
OK, this seems good to me now!  Thanks for getting this Kevin!

PR #4727, opened 2022-02-11
Reclassify @truffle/config as dependency of @truffle/fetch-and-compile

haltman-at:
Apparently this needs to be a dependency, not a devdependency, because `Config` is used directly in the signature of exported functions.

I didn't move over `compile-common` as well because `WorkflowCompileResult` is only used within other types, not directly on exported function signatures.  That's OK, right?  Well, uh, let me know if it isn't...

### DIFF
--- a/packages/fetch-and-compile/package.json
+++ b/packages/fetch-and-compile/package.json
@@ -41,6 +41,7 @@
   "dependencies": {
     "@truffle/codec": "^0.11.26",
     "@truffle/compile-solidity": "^6.0.12",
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
OK! This looks good to go. I did need to alter the `tsconfig.json` as you suggested, and we should update the README to capture that point

haltman-at:
> OK! This looks good to go. I did need to alter the `tsconfig.json` to set `skipLibCheck` as you suggested, and we should update the README to capture that point

You should post that comment on the README PR!

Issue #4654, opened 2022-01-26
`@truffle/contract` still uses Browserify


haltman-at:
## Issue

Truffle Contract still uses Browserify to build.  This is a problem because Browserify is incompatible with certain features of ES2019 and is no longer being updated.  This is why in #4652 I could only update the TS target to ES2018, not ES2019 as I had planned.  I expect that as time goes on this problem will only get worse.  We should remove Browserify and replace it with Webpack or something.

## Environment

* Truffle version (`truffle version`): 5.4.30

haltman-at:
Hooray, this is done!

PR #4728, opened 2022-02-13
Migrate contract to use webpack from browserify

cds-amal:
Quick timebox to investigate migrating the contracts package from browserify to webpack. (See #4654)

- [x] webpack'd version loads in browser
- [x] webpack'd version does not load in node; this is ok since it is intended for the browser


packages/contract/index.js
```diff
@@ -2,6 +2,17 @@ const Schema = require("@truffle/contract-schema");
 const Contract = require("./lib/contract");
 const truffleContractVersion = require("./package.json").version;
 
+// we need to make sure this function exists so ensjs doesn't complain as it requires
+// getRandomValues for some functionalities - webpack strips out the crypto lib
+// so we shim it here
+if (global.crypto && !global.crypto.getRandomValues) {
```
eggplantzzz, line 8: This shimming of the `getRandomValues` function is duplicated a couple of times in the codebase. It is caused by @ensdomains/ensjs and hopefully we can get off that lib soon and get rid of these. Maybe we should leave a note in the comments accompanying all these shims that we can get rid of it after that dep is gone.

eggplantzzz:
I tested the result of this new bundle by running the contract-tests package's tests using it. You can do this by building, switching the `main` field in @truffle/contract to point to `"./dist/truffle-contract.js"`, and then running `yarn test` in @truffle/contract-tests.

There is still a "Buffer has been deprecated" error message that occurs when requiring the bundled package but I'm unsure as to which dependency is causing this.

lsqproduction:
> contract-tests



> I tested the result of this new bundle by running the contract-tests package's tests using it. You can do this by building, switching the `main` field in @truffle/contract to point to `"./dist/truffle-contract.js"`, and then running `yarn test` in @truffle/contract-tests.
> 
> There is still a "Buffer has been deprecated" error message that occurs when requiring the bundled package but I'm unsure as to which dependency is causing this.

what are some ways to test this PR? I tried the steps mention above and got the below:
```
~/code/truffle/packages/contract-tests ‹node-16.16.0›  ‹› (biffy-to-wacker*)
╰─$ yarn test                                                     1 ↵
yarn run v1.22.19
$ ./scripts/test.sh
bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
WARNING: Missing strong random number source
WARNING: Missing strong random number source; using weak randomBytes
Warning: XMLHttpRequest is not defined

ReferenceError: self is not defined
    at /Users/liangliang/code/truffle/packages/contract/dist/truffle-contract.js:514061:11
```

packages/contract/package.json
```diff
@@ -16,7 +16,7 @@
   "main": "index.js",
   "scripts": {
     "build": "yarn compile",
-    "compile": "sh -c \"mkdir -p ./dist\" && browserify --debug ./index.js | exorcist ./dist/truffle-contract.js.map > ./dist/truffle-contract.js && uglifyjs ./dist/truffle-contract.js -o ./dist/truffle-contract.min.js",
+    "compile": "NODE_ENV=production webpack --config ./webpack.config.js && NODE_ENV=development webpack --config ./webpack.config.js",
```
cliffoo, line 19: How does this along with the webpack config make browser / node install separate versions of the truffle contract package?

cliffoo (commented):
I don't really understand why node is not installing the webpacked version. Why do the build files' suffix matter?

I also discovered a couple of things:
- As @eggplantzzz suggested, the tests in packages/contract-tests run OK.
- Package works fine in node (using the non-webpacked version, but again I don't know how it's targeting that)
- A react app using webpack 5 doesn't work out of the box, as it's expected because web3 needs a bunch of polyfill
- I thought a react app using webpack 4 would work ok, but
  - It ran out of memory! Gc complained and called it quits, even with --max-old-space-size set
  - This is expected, but you need a .d.ts to explicity type Truffle contract things in a ts project
- Another observation but very minor: The bundle webpack produced for me was larger than the one you had, but only slightly
```sh
$ du -sh * 
 16M    truffle-contract.js
 17M    truffle-contract.js.map
7.8M    truffle-contract.min.js
4.0K    truffle-contract.min.js.LICENSE.txt
 20M    truffle-contract.min.js.map
```

packages/contract/package.json
eggplantzzz: The environment variable is read by webpack I believe which alters its output.

eggplantzzz:
@lsqproduction You will need to test it in the browser. You can create an html file and add a script tag targeting the bundled file. Then in the terminal you should find `TruffleContract` with value of the export from @truffle/contract.

eggplantzzz:
So something to note is that this new browser bundle will not work in the Node environment any more.

cliffoo (approved):
My confusions have been cleared up thanks to Mr. @eggplantzzz. I tested it again: Node install is unaffected and the webpack bundle works well with a script tag! LGTM (ci passing, ofc)

cds-amal (commented):
Looks good @eggplantzzz

Issue #4726, opened 2022-02-11
Update/replace chromafi


wbt:
---------------------------
## Issue
@truffle/contract depends on @truffle/debug-utils which depends on @trufflesuite/chromafi which depends on ansi-mark which is [vulnerable](https://github.com/F1LT3R/ansi-mark/issues/2), throws [warnings on every run](https://github.com/F1LT3R/ansi-mark/issues/3) (as does super-split, also a dependency of chromafi from the same developer), and isn't maintained.

It would be better if either:
A. Trufflesuite forks and maintains ansi-mark, updates the reference from chromafi, and issues an update on chromafi, or 
B. Truffle/debug-utils finds a way to drop chromafi.

haltman-at:
Should be straightforward enough to strip out the `ansi-mark` dependency, since I'm pretty sure we're not even using that.  Heh, our fork of `chromafi` will get even further away from the original...

haltman-at:
OK, I've put up trufflesuite/chromafi#7 to get rid fo that dependency, and once we merge/release that we can update debug-utils to use it.

wbt:
@haltman-at Do you want to try doing the same for super-split, which has a similar issue and is also a chromafi dependency?

haltman-at:
OK, PR for this is merged, it'll go out in this week's release.

PR #4731, opened 2022-02-14
Dependency update: Update @trufflesuite/chromafi to 3.0.0

haltman-at:
Addresses #4726.  Updates our chromafi fork to the one I just published that gets rid of a bunch of dependencies.  It's 3.0.0 because one of those dependencies I removed was providing actual functionality, so it was a breaking change... but it doesn't break anything for us, as the removed functionality (the `highlight` option) wasn't anything we were using.  (No, that option did *not* control whether syntax highlighting is on or not; it's confusingly named and frankly I don't really understand *what* it did.)  I also removed the comment in debug-utils about that option since it's gone now.

### DIFF
--- a/packages/debug-utils/index.js
+++ b/packages/debug-utils/index.js
@@ -34,59 +34,59 @@ const verbosePanicTable = {
 };
 
 const commandReference = {
-  o: "step over",
-  i: "step into",
-  u: "step out",
-  n: "step next",
+  "o": "step over",
+  "i": "step into",
+  "u": "step out",
+  "n": "step next",
   ";": "step instruction (include number to step multiple)",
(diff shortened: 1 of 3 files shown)


haltman-at:
(I also did a quick manual test to be sure that yes, there really still is syntax highlighting on this branch! :P )

packages/debug-utils/index.js
```diff
@@ -34,59 +34,59 @@ const verbosePanicTable = {
 };
 
 const commandReference = {
-  o: "step over",
-  i: "step into",
-  u: "step out",
-  n: "step next",
+  "o": "step over",
+  "i": "step into",
+  "u": "step out",
```
cds-amal, line 39: heh. Husky barked.

cds-amal (approved):
Nice. LGTM!

PR #4734, opened 2022-02-14
core: use debug module for spawnResult error handling

lsqproduction:
A quick change from console.log to debug module, as console logging the stderr was adding extra lines to the end of the migration output.

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -14,6 +14,7 @@ const path = require("path");
 const EventEmitter = require("events");
 const spawnSync = require("child_process").spawnSync;
 const Require = require("@truffle/require");
+const debug = require("debug")("console");
 
 const processInput = input => {
   const inputComponents = input.trim().split(" ");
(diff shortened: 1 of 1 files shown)


packages/core/lib/console.js
```diff
@@ -272,7 +273,7 @@ class Console extends EventEmitter {
       // Theoretically stderr can contain multiple errors.
       // So let's just print it instead of throwing through
       // the error handling mechanism. Bad call?
-      console.log(spawnResult.stderr.toString());
+      debug(spawnResult.stderr.toString());
```
cds-amal, line 276: [spawnSync](https://nodejs.org/dist/latest-v16.x/docs/api/child_process.html#child_processspawnsynccommand-args-options) returns `status`, `signal` and `error` components in its return object. Do you think it's worth putting the debug behind a guard that checks iff there was an error? It might even be worth it to include those values when logging

cds-amal (commented):
How can I test this?

packages/core/lib/console.js
eggplantzzz: We would get more info this way, right? If that's the case I don't think we should.

cds-amal: There could be more information, but it would only be presented if DEBUG environment is configured

eggplantzzz:
For reference, the PR that implemented this logging in the first place is https://github.com/trufflesuite/truffle/pull/3794.

cds-amal:
> For reference, the PR that implemented this logging in the first place is #3794.

welp, I forgot about that. 🤔

cds-amal (approved):
@eggplantzzz convinced me it's ok to merge this PR, leaving the **ora** investigation as follow-on work.

PR #4736, opened 2022-02-15
Internal improvement: Add license field to core package.json

haltman-at:
Just noticed this field was missing.  Adding it.


eggplantzzz (approved):
bing, bam, boom

PR #4737, opened 2022-02-15
Remove homepage from dashboard package.json because it affects build

haltman-at:
Partial undoing of #4733... turns out that `homepage` affects how `dashboard` builds, so, uh, taking it out.

I'd put in a comment saying not to put it back in, but, not clear that comments are allowed in `package.json`, so, not doing that. :-/


gnidan (approved):
Well that's freakin' weird

haltman-at:
It's a React thing, specifically.  Apparently React makes use of the `homepage` field.

Issue #4616, opened 2022-01-12
Debugger produces error message when printing state while debugging mainnet transaction `0x1c796d1e9867ee9b5f3ae57fa9e7a8623a5baf7eb2419ca7c1a371071a12896a`


haltman-at:
## Issue

(Thanks to @gnidan for finding this issue.)  While debugging mainnet transaction `0x1c796d1e9867ee9b5f3ae57fa9e7a8623a5baf7eb2419ca7c1a371071a12896a` with `--fetch-external`, the debugger encounters an error while trying to print the state (i.e., the source code).

## Steps to Reproduce

In one terminal, using Ganache 7, fork mainnet with `ganache --fork`.

In another terminal, using Node 14, run `truffle debug -x 0x1c796d1e9867ee9b5f3ae57fa9e7a8623a5baf7eb2419ca7c1a371071a12896a --url http://localhost:8545/`.  Make sure you use Node 14!  If you use Node 12, you will instead encounter issue #4617, which will prevent you from reproducing this one!

Type `; 8011` to advance to step 8012, then type `n` to advance again.  Observe the error message.  Or, if you prefer to skip directly to the error, it first occurs at step 8240, so you can get there with `; 8239` from the start of the transaction.

## Expected Behavior

The debugger should correctly print the source location, if possible; if impossible, it should do better than erroring.

## Actual Results

```
debug(inline_config:0x1c796d1e...)> 

ChiToken.sol:

Uncaught TypeError: Cannot read property 'length' of undefined
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at DebugInterpreter.interpreter (/home/sniffnoy/truffle/truffle/packages/core/lib/debug/interpreter.js:684:24)
    at DebugPrinter.printState (/home/sniffnoy/truffle/truffle/packages/core/lib/debug/printer.js:178:18)
    at Object.formatRangeLines (/home/sniffnoy/truffle/truffle/packages/debug-utils/index.js:499:20)
    at Object.formatLinePointer (/home/sniffnoy/truffle/truffle/packages/debug-utils/index.js:420:30)
```

## Environment

* Truffle version (`truffle version`): 5.4.28
* node version (`node --version`): v14.18.2

dongmingh:
The issue can be reproduced with Node v16.10.0 and Truffle v5.4.29.

The cause is that the source code range are `null` at this location
```
debug(inline_config:0x1c796d1e...)> :!<controller.current.location.sourceRange>
{
[...]

PR #4742, opened 2022-02-16
Allow debugger to handle an instruction with invalid source range

dongmingh:
This PR is to address the crash reported in issue #4616.

The root cause of the issue is that an instruction's byte-offset to the start of the range in the source code is past the end of source code. This is impossible since the byte-offset should always be within the range of the source code. Then the debugger sets the source range to null for such an instruction since it does not exist in the source code. And the null source range causes crash when printing the source code.  However, all instructions along with their attributes including the byte-offset are given by the source map which is provided by the compiler. Hence, this issue is caused by the invalid input from the compiler.

However, the debugger needs to be protected from crash with such an invalid input or even does something more than just not crash. Two options are considered when such an instruction is detected. The first option is to issue a warning message and not print any source code. Hence, there will be no crash. But this option is simply to protect debugger from crash. The second option is to print part of the source code, such as the end of source code, in addition to the warning message. The reason to choose the end of source code because the byte-offset is past the end of source code. The second option is selected because it protects the debugger from crash, issues a warning message, and provides some information of the source code.

A note about implementation:
When an invalid byte-offset of an instruction is detected, we only set its source range to the end of source code for the purpose of displaying source code. Other attributes of the instruction including the invalid byte-offset are remained intact. The warning message is issued when printing the source code.

### DIFF
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -172,22 +172,39 @@ class DebugPrinter {
     const colorizedLines = splitLines(colorizedSource);
 
     this.config.logger.log("");
-    // We clone range though only "lines" is used here.
-    let rangePrintout = JSON.parse(JSON.stringify(range));
+    // We create printoutRange with range.lines as initial value for printing.
+    let printoutRange = {
+      start: {
+        line: range.lines.start.line,
+        column: range.lines.start.column
+      },
+      end: {
+        line: range.lines.end.line,
+        column: range.lines.end.column
+      }
+    };
+
     // We print a warning message and display the end of source code when the
     // instruction's byte-offset to the start of the range in the source code
-    // is past the end of source code
-    const instruction = this.session.view(solidity.current.instruction);
-    if (instruction.start >= source.length) {
+    // is past the end of source code.
+    if (range.start >= source.length) {
       this.config.logger.log(
-        `${colors.bold("Warning:")} Past end of source, displaying end.`
+        `${colors.bold(
+          "Warning:"
+        )} Location is past end of source, displaying end.`
       );
       this.config.logger.log("");
-      // We set the lines of the cloned range with the end of source code.
+      // We set the printoutRange with the end of source code.
       // Note that "lines" is the split lines of source code as defined above.
-      rangePrintout.lines = {
-        start: { line: lines.length - 1, column: 0 },
-        end: { line: lines.length - 1, column: 0 }
+      printoutRange = {
+        start: {
+          line: lines.length - 1,
+          column: 0
+        },
+        end: {
+          line: lines.length - 1,
+          column: 0
+        }
       };
     }
 
@@ -196,7 +213,7 @@ class DebugPrinter {
     this.config.logger.log(
       DebugUtils.formatRangeLines(
         colorizedLines,
-        rangePrintout.lines,
+        printoutRange,
         lines,
         contextBefore,
         contextAfter
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -172,16 +172,21 @@ class DebugPrinter {
     const colorizedLines = splitLines(colorizedSource);
 
     this.config.logger.log("");
-    // printout a warning message and display the end of source code when the instruction's
-    // byte-offset to the start of the range in the source code is past the end of source code
+    // We print a warning message and display the end of source code when the
+    // instruction's byte-offset to the start of the range in the source code
+    // is past the end of source code
     const instruction = this.session.view(solidity.current.instruction);
-    if (instruction.start > source.length - 1) {
+    if (instruction.start >= source.length) {
       this.config.logger.log(
-        `${colors.bold(
-          "Warning:"
-        )} The end of source code is displayed since the instruction's source range is past the end of source code.`
+        `${colors.bold("Warning:")} Past end of source, displaying end.`
       );
       this.config.logger.log("");
+      // We set the source range to the end of source code for printing purpose only
+      // Note that the object "lines" is the split lines of source code as defined above
+      range.lines = {
+        start: { line: Object.keys(lines).length - 1, column: 0 },
+        end: { line: Object.keys(lines).length - 1, column: 0 }
+      };
     }
 
     //HACK -- the line-pointer formatter doesn't work right with colorized
--- a/packages/core/lib/debug/printer.js
+++ b/packages/core/lib/debug/printer.js
@@ -172,6 +172,8 @@ class DebugPrinter {
     const colorizedLines = splitLines(colorizedSource);
 
     this.config.logger.log("");
+    // We clone range though only "lines" is used here.
+    let rangePrintout = JSON.parse(JSON.stringify(range));
     // We print a warning message and display the end of source code when the
     // instruction's byte-offset to the start of the range in the source code
     // is past the end of source code
@@ -181,11 +183,11 @@ class DebugPrinter {
         `${colors.bold("Warning:")} Past end of source, displaying end.`
       );
       this.config.logger.log("");
-      // We set the source range to the end of source code for printing purpose only
-      // Note that the object "lines" is the split lines of source code as defined above
-      range.lines = {
-        start: { line: Object.keys(lines).length - 1, column: 0 },
-        end: { line: Object.keys(lines).length - 1, column: 0 }
+      // We set the lines of the cloned range with the end of source code.
+      // Note that "lines" is the split lines of source code as defined above.
+      rangePrintout.lines = {
+        start: { line: lines.length - 1, column: 0 },
+        end: { line: lines.length - 1, column: 0 }
       };
     }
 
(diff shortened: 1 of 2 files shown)


packages/core/lib/debug/printer.js
```diff
@@ -171,6 +171,17 @@ class DebugPrinter {
     const colorizedLines = splitLines(colorizedSource);
 
     this.config.logger.log("");
+    // printout a warning message and display the end of source code when the instruction's
+    // byte-offset to the start of the range in the source code is past the end of source code
+    const instruction = this.session.view(solidity.current.instruction);
+    if (instruction.start > source.length - 1) {
```
haltman-at, line 177: Again, I would suggest `instruction. start >= source.length` rather than `instruction.start > source.length - 1`.

packages/core/lib/debug/printer.js
```diff
@@ -171,6 +171,17 @@ class DebugPrinter {
...
+      this.config.logger.log(
+        `${colors.bold(
+          "Warning:"
+        )} The end of source code is displayed since the instruction's source range is past the end of source code.`
```
haltman-at, line 181: I think the wording here could be tightened up a bit; perhaps just something like "Warning: Past end of source, displaying end"?  This is a computer printout aimed at fellow programmers, we don't need to be too elaborate here. :)

haltman-at (commented):
So, my big question here is, is this being handled in the right place?

Like, you chose to put the bulk of the handling in `SourceMapUtils.getProcessedInstructionsForBinary`.  I'm not certain if that's the correct place?  Like, maybe this would be better handled in `getSourceRange`?  (That might require changing the arguments to `getSourceRange`, and appropriately adjusting the invocation.)

Or another possibility would be to not handle this in `SourceMapUtils` at all, but rather to simply allow the `null`s, and perform all the handling in `core/lib/debug` and `debug-utils`.  IDK, that's maybe a bit hacky.

I don't know, what do you think of those possibilities?  Or do you think the code should stay where it is?

haltman-at:
OK!  That's likely the easiest approach, since it lets you avoid the missing-sources problem.  Just be sure that if you take any shortcuts they're appropriately commented.

packages/core/lib/debug/printer.js
```diff
@@ -171,6 +171,22 @@ class DebugPrinter {
...
+    const instruction = this.session.view(solidity.current.instruction);
+    if (instruction.start >= source.length) {
+      this.config.logger.log(
+        `${colors.bold("Warning:")} Past end of source, displaying end.`
+      );
+      this.config.logger.log("");
+      // We set the source range to the end of source code for printing purpose only
+      // Note that the object "lines" is the split lines of source code as defined above
+      range.lines = {
+        start: { line: Object.keys(lines).length - 1, column: 0 },
```
haltman-at, line 186: The object `lines` is already an array, right?  So there isn't a need to use `Object.keys` here; you could just do `lines.length`, unless I've misunderstood.

packages/core/lib/debug/printer.js
```diff
@@ -171,6 +171,22 @@ class DebugPrinter {
...
+    // is past the end of source code
+    const instruction = this.session.view(solidity.current.instruction);
+    if (instruction.start >= source.length) {
+      this.config.logger.log(
+        `${colors.bold("Warning:")} Past end of source, displaying end.`
+      );
+      this.config.logger.log("");
+      // We set the source range to the end of source code for printing purpose only
+      // Note that the object "lines" is the split lines of source code as defined above
+      range.lines = {
```
haltman-at, line 185: I think you should avoid mutating `range.lines` here, seeing as `range` is a field on an argument that was passed in.  While I don't think it should cause a problem in the cases that will actually come up, to avoid potential problems in the future you should avoid making this modification... this should be a pure function and not mutate its inputs.

haltman-at (commented):
I believe this should do the job, but I think there's a mutation that should be removed to avoid future problems.

packages/core/lib/debug/printer.js
```diff
@@ -171,13 +171,31 @@ class DebugPrinter {
     const colorizedLines = splitLines(colorizedSource);
 
     this.config.logger.log("");
+    // We clone range though only "lines" is used here.
+    let rangePrintout = JSON.parse(JSON.stringify(range));
```
haltman-at, line 175: Please don't perform deep clones this way.  This is a hack and, if there were truly a reason you needed to resort to it, it ought to be labeled as such.  You generally shouldn't need to perform deep cloning at all though, and you certainly don't need to here.  If you did need to, I would suggest importing `lodash.clonedeep` to do it -- but, again, there's absolutely no reason to perform a clone here, whether deep or shallow.

As best I can tell, the reason for this clone is... to preserve the existing structure of variables that's used?  That's not something you need to preserve.  Rather than performing a deep clone to preserve the existing structure of variables used, you should change the variables used to something more appropriate for the new situation.  You should just be extracting and using the relevant information, not cloning irrelevant information.

(Btw: You can even do this without using any mutation.  Avoiding mutation is nice when possible as it makes the meaning of variables clearer.  I mean you don't want to go too hard on that, don't avoid mutation if that will make things difficult, but...)

In a more minor issue that occurs on this line, I must also point out that this variable's name is confusing.  You named it `rangePrintout`; that sounds like a printout for a range, when you meant a range for a printout.  If you keep this variable (although I imagine there's a good chance you'll remove it), you'd probably want to rename it to `printoutRange` or `rangeForPrintout`.  Remember, English is a head-second language; if you write a noun-noun compound, the first noun will be read as modifying the second, not vice-versa.  Now you will often see variables named with modifiers coming after the noun, but importantly, in those cases, the modifiers are *adjectives*, not other nouns, which makes the result unambiguous regardless of order.  With two nouns, though, you can't use part of speech to disambiguate, so you have to use order; thus the first noun will be read as modfiying the second.  You'll want to keep that in mind when naming variables.

packages/core/lib/debug/printer.js
```diff
@@ -171,13 +171,31 @@ class DebugPrinter {
...
+    // We print a warning message and display the end of source code when the
+    // instruction's byte-offset to the start of the range in the source code
+    // is past the end of source code
+    const instruction = this.session.view(solidity.current.instruction);
```
haltman-at, line 179: I only just noticed something here: This isn't correct!  You're looking at `solidity.current.instruction`, but remember, `location` was passed in.  If the location that was passed in was the current location, this is fine.  But what if the location that was passed in isn't the current location?  You should go by the passed-in location, not the current instruction.

packages/core/lib/debug/printer.js
```diff
@@ -171,13 +171,31 @@ class DebugPrinter {
...
+    if (instruction.start >= source.length) {
+      this.config.logger.log(
+        `${colors.bold("Warning:")} Past end of source, displaying end.`
```
haltman-at, line 182: Hm, so I know I was the one who suggested this message to you, but I'm realizing now that when I did so, I forgot about the case where the location passed in isn't the current location.  Maybe change "Past end of source" to "Location is past end of source"?  To my mind, the former reads as necessarily being about the current location, whereas the latter doesn't, so I'd suggest changing that.

haltman-at (changes requested):
I just noticed this contains a logic error that will need to be fixed.  Also, that cloning hack should be removed, especially as there isn't a need to clone here at all.  There's also some other minor stuff.

packages/core/lib/debug/printer.js
dongmingh: I consider to keep the variable `rangePrintout` and rename it to `printoutRange`.  This variable will contain `sourceRange.lines` only.  The initial value of this variable is `range.lines` from input and is set to the end of source code if the invalid byte-offset is detected.

haltman-at: That's really needlessly indirect, don't you think?  If it's only going to contain `lines`, shouldn't you just use `lines`?  Why would you have a variable just to contain it?

haltman-at: Oops, it seems I misunderstood @dongmingh's comment here -- he meant contain *as the value*, not contain *as a subfield*.  Sorry about that!  Yes, this approach makes sense given what was actually meant.

packages/core/lib/debug/printer.js
```diff
@@ -172,13 +172,48 @@ class DebugPrinter {
     const colorizedLines = splitLines(colorizedSource);
 
     this.config.logger.log("");
+    // We create printoutRange with range.lines as initial value for printing.
+    let printoutRange = {
```
haltman-at, line 176: Why not just do `let printoutRange = range.lines`?  You're not going to be altering its components, so there isn't a need to do a clone like this.

haltman-at (commented):
Seems basically good, thanks Dongming!  One small nitpick.

packages/core/lib/debug/printer.js
```diff
@@ -172,13 +172,48 @@ class DebugPrinter {
...
+        `${colors.bold(
+          "Warning:"
+        )} Location is past end of source, displaying end.`
+      );
+      this.config.logger.log("");
+      // We set the printoutRange with the end of source code.
+      // Note that "lines" is the split lines of source code as defined above.
+      printoutRange = {
+        start: {
+          line: lines.length - 1,
```
haltman-at, line 201: Also, thanks for finding a better way to get the final line number.

haltman-at (approved):
OK, let's get this in, thanks Dongming!

PR #4744, opened 2022-02-16
contract-tests: fix detetNetwork test

cds-amal:
This PR is necessary to fix an error hidden behind a bad test. [assert.rejects](https://nodejs.org/api/assert.html#assertrejectsasyncfn-error-message) is a promise that should be awaited. Upon fixing this test, it exposed an error in `detectNetwork()` that didn't guard use of `this.interfaceAdapter`. 

- await  in tests as per documentation
- remove 3rd positional parameter, message, to  so expected and got comparison is shown on failures
- guard this.interfactAdapter in  implementation


packages/contract/lib/contract/constructorMethods.js
```diff
@@ -119,6 +119,10 @@ module.exports = Contract => ({
   },
 
   async detectNetwork() {
+    // guard interfaceAdapter!
+    if (!Boolean(this.interfaceAdapter)) {
```
haltman-at, line 123: This is just a longer way of saying `if (!this.interfaceAdapter)` that, IMO, doesn't really add any clarity to it.  I would suggest either the shorter way or the more explicit `if (this.interfaceAdapter == null)`.

haltman-at (commented):
Seems basically good, but I'm a bit confused about the reasoning for removing the messages... how would you get error messages if the tests *pass* on an error?  Since the failure case is when there's *no* error, there won't be any pre-existing error message to pass on, right?

Also one really minor style nitpick. :P

packages/contract/lib/contract/constructorMethods.js
cds-amal: > Seems basically good, but I'm a bit confused about the reasoning for removing the messages... how would you get error messages if the tests _pass_ on an error? Since the failure case is when there's _no_ error, there won't be any pre-existing error message to pass on, right?
> 
Correct, no pre-existing error message to pass on, however there's still an indication that an expected failure didn't happen, like below where I simulated a failure in a test. I think this is acceptable, but will defer to you whether it's acceptable.

>   1) TruffleContract.deployed()
       throws if called before setting a provider:
     AssertionError [ERR_ASSERTION]: Missing expected rejection (Error).
      at async Context.<anonymous> (test/contract/constructorMethods.js:72:5)

cds-amal:
> Seems basically good, but I'm a bit confused about the reasoning for removing the messages... how would you get error messages if the tests _pass_ on an error? Since the failure case is when there's _no_ error, there won't be any pre-existing error message to pass on, right?
> 
Correct, no pre-existing error message to pass on, however there's still an indication that an expected failure didn't happen, like below where I simulated a failure in a test. I think this is acceptable, but will defer to you whether it's acceptable.

>   1) TruffleContract.deployed()
       throws if called before setting a provider:
     AssertionError [ERR_ASSERTION]: Missing expected rejection (Error).
      at async Context.<anonymous> (test/contract/constructorMethods.js:72:5)

haltman-at (approved):
Oh, I see, so the message is redundant because it wasn't saying anything not in the default message.  Cool.  Approved.

Issue #4722, opened 2022-02-11
Solidity tests on Solidity versions <0.7.6 can pass incorrectly and have useless warnings printed when they pass


haltman-at:
## Issue

This issue is being split off from #4696.  So I'm just going to copy the relevant text from that:

> But the thing that poses a real problem here is how this [#4696] interacts with Truffle's Solidity assertion library.
> 
> The smaller problem is that it means that if you use the Solidity assertion library with --show-events on, you'll get a bunch of warnings about undecodeable events, because passing TestEvents have "" as their message. These are supposed to be filtered out... but this relies on decoding them, which the decoder currently fails at.
> 
> The bigger problem is that it could result in Solidity tests passing incorrectly. If you use an assertion and give the message as "", and it isn't one of our assertions that automatically includes a message of some sort, then the decoder will fail to decode it and as such the test runner won't throw. Oops!
> 
> Fixing the underlying problem with the decoder seems hard [i.e. #4696], so we may want a stopgap solution to the TestEvent problem that can go out quicker. A simple way would be to just restore the old way of handling TestEvents, that didn't rely on @truffle/decoder.

## Environment

* Truffle version (`truffle version`): 5.4.33

lsqproduction:
Addressed in the release v5.5.2.

Issue #4696, opened 2022-02-05
Decoder can't decode events from old Solidity versions with misencoded empty string (can potentially cause Solidity tests to pass incorrectly)


haltman-at:
## Issue

So it turns out that Solidity versions prior to 0.7.6, when using ABI encoder v1, would incorrectly encode the empty string when emitting events or returning values or revert strings.

This is a problem for the decoder, because when decoding events or return values (including revert strings), it requires that the input be encoded *exactly* correctly for the particular decoding.  As such, the decoder will fail to decode such events, return values, or revert strings.  (The problem doesn't apply to custom errors for the simple reason that they weren't introduced until Solidity 0.8.4) The reason the decoder does this is to prevent it from returning nonsensical interpretations in various cases, so this isn't an easy assumption to remove.

(Note I say "the decoder", but of course this problem really lies in Truffle Codec, and so all this applies to the debugger as well.)

Now, one possibility would be to say, this is a Solidity bug, we won't handle it... but we generally try to handle Solidity bugs.  This one seems particularly like a pain to handle, so maybe this is low-priority, I don't know.

But the thing that poses a real problem here is how this interacts with Truffle's Solidity assertion library.

The smaller problem is that it means that if you use the Solidity assertion library with `--show-events` on, you'll get a bunch of warnings about undecodeable events, because passing `TestEvent`s have `""` as their message.  These are supposed to be filtered out... but this relies on decoding them, which the decoder currently fails at.

The bigger problem is that it could result in Solidity tests passing incorrectly.  If you use an assertion and give the message as `""`, and it isn't one of our assertions that automatically includes a message of some sort, then the decoder will fail to decode it and as such the test runner won't throw.  Oops!

Fixing the underlying problem with the decoder seems hard, so we may want a stopgap solution to the `TestEvent` problem that can go out quicker.  A simple way would be to just restore the old way of handling `TestEvent`s, that didn't rely on `@truffle/decoder`.  On the hackier side, another possibility would be to alter our assertion libraries to *always* append some sort of message. :)  (And, uh, use `" "` instead of `""` in the success case...?)  ...yeah, that actually might be a good idea anyway, but I think we should probably do the former...

## Environment

* Truffle version (`truffle version`): 5.4.32

haltman-at:
Note to self: Go back and split this in two.

haltman-at:
[...]

PR #4745, opened 2022-02-17
Bug fix: Remove incorrect warnings from `truffle test --show-events` (and prevent Solidity tests from incorrectly passing?)

haltman-at:
Addresses #4722.  This PR adds the `disableChecks` option codec and decoder when decoding events.  If this option is passed, codec will be a bit looser about checking that its event decodings are strictly accurate.  This can, in some cases, lead to events being incorrectly decoded as ambiguous (where, without the option set, they would correctly be decoded as unambiguous).

So why add such an option?  Well, it turns out that Solidity versions older than 0.7.6, when using ABI encoder V1, sometimes incorrectly encodes the empty string (using an encoding for it that has more padding than it should).  And this is particularly a problem for us because the `TestEvent`s that we use for Solidity testing often include empty strings in their arguments!  In the case of a passing test, this would cause a bunch of incorrect warnings to show up when using `--show-events`.  In the case of a failing test, this notionally could cause it to incorrectly be treated as passing, although oddly I haven't been able to get this case to actually occur (more on this in a bit).

As for the problem of what to do about this generally (#4696), I don't really know.  But @gnidan and I came up with this stopgap to address #4722, the problems this causes for Solidity testing.

The stopgap is this: First, we create the `disableChecks` flag, described above.  Then, we have Solidity testing use that flag.  As such, it will be able to correctly `TestEvent`s even if they were decoded wrong.  That's it!  This does mean that other events emitted during Solidity testing could be incorrectly decoded as ambiguous, but @gnidan and I figured this was the lesser of two evils, as extra encodings appearing from `disableChecks` being set ought to be pretty rare.

What the `disableChecks` flag actually does, see, is that codec, when decoding events, checks its work afterward by *re-encoding* the event to make sure it matches what was passed in.  If the `disableChecks` flag turned off that check entirely, it would be pretty easy to construct nasty examples.  However, I made it so that the `disableChecks` flag turns off that check as applied to the event *data*, but leaves it on for the event *topics*.  (Since the empty string issue can't come up in the event topics!)  The result is that examples that will fool codec with `disableChecks` set have to be fairly contrived.  (If you want to see one, though, take a look at the `StrangeEventTest` I added.  The particular strings used are arbitrary... but those `96`s are not! :)  I've engineered this event so that `StrangeEventTest.StrangeEvent(96, "ABC", "123", 96)` will, if `disableChecks` is on, get an incorrect alternate decoding of `StrangeEventLibrary.StrangeEvent(96, "123", "ABC", 96)`... note how the order of the strings is swapped.  If people *really* want, I can explain how that happens, but I think I'll skip going into the binary for now!)

I do have to make a note here... as mentioned above, I was not actually able to observe any cases where a *failing* `TestEvent` got encoded wrong, even when its message was empty.  I have no idea why.  I had the same contract emitting both -- not in the same transaction admittedly but that shouldn't matter -- and could see in the binary that the failing ones got encoded right and the passing ones got encoded wrong.  I have no idea what's up with this, and I'm not sure if this suggests that a different solution would be better.  Well, I've put up this PR 

In addition to adding this functionality, I of course also added tests of it.  (Well, one test that makes three checks... sorry, it was just easier than splitting things up.)  Rather than add new tests to `decoder`, I decided to add these as tests in `truffle`, testing the `truffle test --show-events` command.  The test checks that:
1. We don't get any warnings due to undecodeable events as a result of `TestEvent`s;
2. The failing test fails as it should;
3. That after the Solidity tests run, the `disableChecks` flag is properly reset, so that the Javascript test doesn't get fooled by my tricksy event and correctly realizes it's unambiguous (despite running after Solidity tests).

And that's it!  Well, and `prettier` also made some changes... the real changes to `codec/lib/core.js` are the ones from lines 420 to 441, the rest are just due to prettier and can be ignored.

### DIFF
--- a/packages/core/lib/testing/TestRunner.js
+++ b/packages/core/lib/testing/TestRunner.js
@@ -39,11 +39,11 @@ class TestRunner {
     this.TEST_TIMEOUT = (options.mocha && options.mocha.timeout) || 300000;
   }
 
-  disableChecks() {
+  disableChecksOnEventDecoding() {
     this.disableChecks = true; //used by Solidity testing due to empty string problem on Solidity <0.7.6
   }
 
-  reEnableChecks() {
+  reEnableChecksOnEventDecoding() {
     this.disableChecks = false;
   }
 
--- a/packages/core/lib/testing/TestRunner.js
+++ b/packages/core/lib/testing/TestRunner.js
@@ -39,6 +39,14 @@ class TestRunner {
     this.TEST_TIMEOUT = (options.mocha && options.mocha.timeout) || 300000;
   }
 
+  disableChecks() {
+    this.disableChecks = true; //used by Solidity testing due to empty string problem on Solidity <0.7.6
+  }
+
+  reEnableChecks() {
+    this.disableChecks = false;
+  }
+
   async initialize() {
     debug("initializing");
     this.config.resolver = new Resolver(this.config, {
@@ -169,7 +177,8 @@ class TestRunner {
       //should be fine, but should change this once decoder
       //accepts more general types for blocks
       fromBlock: this.currentTestStartBlock.toNumber(),
-      extras: "necessary" //include weird decodings if usual ones fail :P
+      extras: "necessary", //include weird decodings if usual ones fail :P
+      disableChecks: this.disableChecks //for Solidity testing
     });
 
     const userDefinedEventLogs = logs.filter(log => {
--- a/packages/truffle/test/sources/mixed_testing/.eslintrc.json
+++ b/packages/truffle/test/sources/mixed_testing/.eslintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": ["../../../../../.eslintrc.truffle.json"]
+}
--- a/packages/truffle/test/sources/mixed_testing/test/js-test.js
+++ b/packages/truffle/test/sources/mixed_testing/test/js-test.js
@@ -1,0 +1,8 @@
+const StrangeEventTest = artifacts.require("StrangeEventTest");
+
+contract("StrangeEventTest", function () {
+  it("should emit a strange event", async function () {
+    const instance = await StrangeEventTest.deployed();
+    await instance.run();
+  });
+});
--- a/packages/codec/lib/core.ts
+++ b/packages/codec/lib/core.ts
@@ -4,7 +4,6 @@ const debug = debugModule("codec:core");
 import type * as Abi from "@truffle/abi-utils";
 import * as Ast from "@truffle/codec/ast";
 import * as AbiData from "@truffle/codec/abi-data";
-import * as Compiler from "@truffle/codec/compiler";
 import * as Topic from "@truffle/codec/topic";
 import type * as Pointer from "@truffle/codec/pointer";
 import type {
@@ -19,7 +18,6 @@ import type {
   LogDecoding,
   LogOptions
 } from "@truffle/codec/types";
-import type { ConstructorReturndataAllocation } from "@truffle/codec/abi-data/allocate";
 import * as Evm from "@truffle/codec/evm";
 import * as Contexts from "@truffle/codec/contexts";
 import { abifyType, abifyResult } from "@truffle/codec/abify";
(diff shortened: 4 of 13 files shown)


packages/core/lib/testing/TestRunner.js
```diff
@@ -39,6 +39,14 @@ class TestRunner {
     this.TEST_TIMEOUT = (options.mocha && options.mocha.timeout) || 300000;
   }
 
+  disableChecks() {
```
eggplantzzz, line 42: What would you think about renaming this option so as to make it clear that it is related to codec checks? Something like `disableCodecChecks` and so forth?

haltman-at: Yeah, that would probably be clearer; I'll go do that.

packages/core/lib/testing/TestRunner.js
```diff
@@ -39,11 +39,11 @@ class TestRunner {
     this.TEST_TIMEOUT = (options.mocha && options.mocha.timeout) || 300000;
   }
 
-  disableChecks() {
+  disableChecksOnEventDecoding() {
     this.disableChecks = true; //used by Solidity testing due to empty string problem on Solidity <0.7.6
```
eggplantzzz, line 43: sorry and maybe this is pedantic, but does it make sense to change these variable names too?

haltman-at:
Looking at this again, I decided that the `StrangeEvent` business is going to be horribly unclear to anyone looking at the code later, so I decided to put an explanation in comments in the code itself.  It's not a full explanation, but I think it'll do.

packages/core/lib/testing/TestRunner.js
eggplantzzz: right but looked at as a generic property of `TestRunner` (like `TestRunner.disableChecks`) it is not clear what it means

haltman-at: I mean, that's true, but when is that situation going to come up?  Everywhere it appears in the code, it's made clear what it means.

eggplantzzz: Ok, if you are convinced that it doesn't really matter then I won't hold it back.

packages/truffle/test/sources/mixed_testing/test/js-test.js
```diff
@@ -0,0 +1,8 @@
+const StrangeEventTest = artifacts.require("StrangeEventTest");
+
+contract("StrangeEventTest", function () {
+  it("should emit a strange event", async function () {
+    const instance = await StrangeEventTest.deployed();
+    await instance.run();
```
eggplantzzz, line 6: should this have an assertion or is this just here to test that it doesn't throw or something?

eggplantzzz (approved):
I like it!

packages/truffle/test/sources/mixed_testing/.eslintrc.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": ["../../../../../.eslintrc.truffle.json"]
```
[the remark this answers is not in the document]
haltman-at: I mean, that's the same as everywhere we have this...

packages/core/lib/testing/TestRunner.js
haltman-at: I mean the field can be renamed in the future, it's a purely internal thing, it's not like that would break anything.

packages/truffle/test/sources/mixed_testing/test/js-test.js
haltman-at: Ah, sorry, that perhaps warrants some explanation.  That's not a test of our code at all.  That's a test *within the Truffle project which is a test fixture*.  The test of our code is elsewhere; it runs `truffle test --show-events`, causing this test to run, and then and checks that the correct things get logged as a result.

packages/truffle/test/sources/mixed_testing/.eslintrc.json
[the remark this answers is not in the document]
eggplantzzz: yep! it's still ugly everywhere!

packages/core/lib/testing/TestRunner.js
eggplantzzz: sounds good to me!

PR #4748, opened 2022-02-17
Prepare @truffle/dashboard* packages for release

gnidan:
This PR "downgrades" the package.json versions for the three new packages, so as to facilitate `lerna version` setting these to v0.1.0 in the next release.

This is just playing it safe; from what I recall, `lerna version` doesn't like it when you try to bump a package's version to itself.


gnidan:
> Thanks for getting this!

Just glad I thought of it!

PR #4749, opened 2022-02-17
Await `assert.rejects` in box tests

cds-amal:
This PR awaits `assert.rejects` as per spec.


packages/box/test/utils-unbox.js
```diff
@@ -43,10 +43,11 @@ describe("utils", () => {
       },
       message: errorMessage
     };
+    let url;
```
haltman-at, line 46: Yikes! How was this missing...?

haltman-at (approved):
Thanks Amal!

Issue #2838, opened 2020-02-19
Link an address to a contract as a library in deployer


bwheeler96:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

My contract has a `libary Lib {}` in it. I would like to use an existing library in the migrations script.

Something like:

`deployer.linkAddressAsLibrary('0xabcde...', 'AnExistingLibraryOnChain')`

gnidan:
Looks like the missing piece is an equivalent for `deployer.deploy()`... right now the process is:

```javascript
deployer.deploy(MyLibrary);
deployer.link(MyLibrary, MyContract);
deployer.deploy(MyContract);
```

But you already have a deployed library... you just need that instance to be recorded in the artifact JSON. So I think a better solution is a deployer method to record the already-deployed library. Maybe something like this:

```javascript
deployer.recordDeployed(MyLibrary, "<address>");
deployer.link(MyLibrary, MyContract);
deployer.deploy(MyContract);
```

Something like that?

bwheeler96:
@gnidan something like that yes. Here is the simplest use case I can think of:

`Library A` exists on main chain/kovan wherever, but for local development you want to use your own.

In local development, you link your own `Library A`, but then for mainnet migrations you use the on-chain version.

There are more complex cases, like what I am trying to do saving code by embedding constant addresses that don't belong in the code itself, but also don't belong in the constructor of the contract. So I use `library` as a way to embed constant addresses dynamically.

bwheeler96:
@gnidan This actually already works, the key is that `.link` returns a thenable, so it must be awaited. Here is what works for me

```
await MyContract.link('CustomLibraryName', '0xabcde...')
```

Alternatively you could use `.then()` syntax.

Also, it should be noted that my deploy script must be setup like this for `async` to work:

```
async function doDeploy() {
  await ...
}

module.exports = (deployer, network) => {
  deployer.then(async () => {
    await doDeploy(deployer, network)
  })
}
```

eggplantzzz:
Do note @bwheeler96 that the exported function in your migration can be `async`. So you can do
```
module.exports = async (deployer, network) => {
  const instance = await deployer.deploy(MyContract);
  ... // do other stuff
```

bwheeler96:
@eggplantzzz I agree that this **should** be true, but someone, somewhere showed an example where there appeared to be some issue with exporting an async function.

[...]

PR #4752, opened 2022-02-18
Enhancement: Allow deployer to link library instances

haltman-at:
Finally does away with #2838!

This PR allows `deployer.link` to take a library *instance* in the first argument, rather than just a library *constructor*.  This allows for linking to a library at a specified address in deployment:
```
const libraryInstance = await Library.at(address);
await deployer.link(libraryInstance, Contract);
```
thus finally solving issue #2838.

This PR is mostly pretty straightforward; the primary functional change is that now, when looking for the library name, we check `library.constructor.contractName` if `library.contractName` doesn't exist (and `library.constructor` does exist).  You may be wondering how it's possible that I was able to seamlessly replace `library.contract_name` with `library.contractName`; the answer is that these getters are synonyms, so I changed it to the more modern one.

In addition, there was a buggy ternary I had to change -- the condition was simply `typeof library.isDeployed`, which, of course, will always be truthy!  It looks like what was intended was `typeof library.isDeployed === "function"`, which I changed it to.

Now you may notice that I did *not* change the final call to `destination.link(library)`.  What's up with that?  Don't I have to update that call?  The answer is, no!  Unlike `deployer.link()`, `contract.link()` is actually already prepared to handle this case, so I didn't have to do any work there.

Finally, I added a test of this functionality (in addition to giving it a few manual tests).  It seems to work! :)  I also renamed the helper function `linkingOccurredForName` to simply `linkingOccurred`, as it's now checking address as well, so the name no longer seemed appropriate.

### DIFF
--- a/packages/deployer/test/deployer.js
+++ b/packages/deployer/test/deployer.js
@@ -11,7 +11,7 @@ const {
   getAllEventsByName,
   preDeployOccurredForNames,
   postDeployOccurredForNames,
-  linkingOccurredForName
+  linkingOccurred
 } = require("./helpers/eventSystem");
 
 describe("Deployer (sync)", function () {
@@ -228,7 +228,32 @@ describe("Deployer (sync)", function () {
 
       assert(preDeployOccurredForNames(options, ["IsLibrary", "UsesLibrary"]));
       assert(postDeployOccurredForNames(options, ["IsLibrary", "UsesLibrary"]));
-      assert(linkingOccurredForName(options, "UsesLibrary", "IsLibrary"));
+      assert(
+        linkingOccurred(options, "UsesLibrary", "IsLibrary", IsLibrary.address)
+      );
+    });
+
+    it("deployer.link with library instance", async function () {
+      const newLibrary = await IsLibrary.new();
+
+      await deployer.start();
+      const migrate = async function () {
+        await deployer.deploy(IsLibrary); //just for comparison below
+        const isLibrary = await IsLibrary.at(newLibrary.address);
+        await deployer.link(isLibrary, UsesLibrary);
+        await deployer.deploy(UsesLibrary);
+      };
+
+      await migrate();
+
+      assert(UsesLibrary.address !== null);
+      assert(IsLibrary.address !== null);
+      assert(newLibrary.address !== null);
+      assert(newLibrary.address !== IsLibrary.address);
+
+      assert(
+        linkingOccurred(options, "UsesLibrary", "IsLibrary", newLibrary.address)
+      );
     });
 
     // There's a chain like this in the @truffle/core solidity-tests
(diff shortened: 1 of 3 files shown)


packages/deployer/test/deployer.js
```diff
@@ -228,7 +228,32 @@ describe("Deployer (sync)", function () {
...
+        await deployer.link(isLibrary, UsesLibrary);
+        await deployer.deploy(UsesLibrary);
+      };
+
+      await migrate();
+
+      assert(UsesLibrary.address !== null);
+      assert(IsLibrary.address !== null);
+      assert(newLibrary.address !== null);
+      assert(newLibrary.address !== IsLibrary.address);
```
eggplantzzz, line 252: What is the point of this check?

eggplantzzz: I guess it ensures that `at` doesn't update the address.

packages/deployer/test/helpers/eventSystem.js
```diff
@@ -24,12 +24,18 @@ const postDeployOccurredForNames = (options, contractNames) => {
   }, true);
 };
 
-const linkingOccurredForName = (options, contractName, libraryName) => {
+const linkingOccurred = (
```
cds-amal, line 27: Is there any way to get this to distinguish the type of linking that occurred?

haltman-at: What do you mean?  Like instance vs class?  That'd have to be emitted in the event, and right now the event only gives names and addresses.  Is that information we want to put in the event?  It's not clear to me what a reporter would do with that, or why we'd want to check that here... what matters ultimately is the address.

cds-amal: I think the interface distinction matters because it could expose a different intention by the user. It is something I would want to see in the migration log/reporter, preferring to err on the side of presenting more information to the user. Perhaps I'm wrong, and I also don't want to hold your PR back. We can always revisit what we present later on if it becomes an issue

haltman-at: Hm, I wonder if other people have opinions on this, regarding if this is something that should be added, if the reporter should display this... @gnidan? @eggplantzzz? @lsqproduction? @sukanyaparashar?

(And yeah I'll probably just merge this later, but there's no hurry; nothing depends on this and the release is days away, so I'm OK with waiting a bit longer while people discuss. :) )

eggplantzzz: Offhand I can't really think of something useful to print about this. How would you word it @cds-amal? It seems pretty unhelpful to me to say something like "Linking class instance" versus something else. What kind of different intention might the  user have? I'm kind of with @haltman-at on this one. But if you have a more specific idea of what and how to implement this I'm all ears.

haltman-at: OK since nobody else has responded I'm going to go ahead and merge this shortly, if you want to respond do so quickly!

PR #4753, opened 2022-02-18
Internal improvement: Remove listener from debugger trace submodule

haltman-at:
This PR removes the indirect system of communication between `next` and `advance` in the debugger's `trace` submodule.  I noticed that this was overly complicated while going through the code with @dongmingh.  So, now, `next` is gone; the whole thing is replaced by `advance`.  The listener is of course gone too, which means that `session` no longer has to start the listener.  And, that's it.  I don't know if this will meaningfully improve debugger performance, but it's definitely simpler to read, at least.

(Also, `prettier` made some changes as well.)

Edit: Oops, I forgot to remove the definition of the `NEXT` action itself!  Went back and did that now.

Edit again: One more thing -- I figured while I was at it, I would rename `BACKTICK` to `TOCK` and `TOCK` to `ADVANCE`.  I think that makes things clearer! :)

### DIFF
--- a/packages/debugger/lib/trace/sagas/index.js
+++ b/packages/debugger/lib/trace/sagas/index.js
@@ -31,16 +31,13 @@ export function* advance() {
   let waitingForSubmodules = 0;
 
   if (remaining > 0) {
-    debug("putting TICK");
     // updates state for current step
     waitingForSubmodules = yield select(trace.application.submoduleCount);
     yield put(actions.tick());
-    debug("put TICK");
 
     //wait for all backticks before continuing
     while (waitingForSubmodules > 0) {
-      yield take(actions.BACKTICK);
-      debug("got BACKTICK");
+      yield take(actions.TOCK);
       waitingForSubmodules--;
     }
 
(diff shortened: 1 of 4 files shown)


packages/debugger/lib/trace/sagas/index.js
```diff
@@ -25,51 +24,36 @@ export function* addSubmoduleToCount(increment = 1) {
...
     // updates state for current step
     waitingForSubmodules = yield select(trace.application.submoduleCount);
     yield put(actions.tick());
-    debug("put TICK");
 
     //wait for all backticks before continuing
     while (waitingForSubmodules > 0) {
-      yield take(actions.BACKTICK);
-      debug("got BACKTICK");
+      yield take(actions.TOCK);
```
cds-amal, line 40: Does it make sense to keep the debug logs for tracing `actions.tick()`, `actions.TOCK()` and `actions.endTrace()`?

cds-amal (approved):
Looks good to me!

packages/debugger/lib/trace/sagas/index.js
haltman-at: Not really to my mind; I removed them because this part of the code has been unproblematic for quite a while now, and these debugs are quite noisy if turned on.

Issue #4739, opened 2022-02-15
Debugger's stacktraces show extra, non-existent stackframes


haltman-at:
## Issue

When you make an external function call, you get *two* stackframes instead of one.  One for the external call (with the address but no function name), and then another one for the function itself (but w/o the address).

These should be combined.  Originally they *were* combined (when they were added in #2965), but then I decombined them (in #2976) due to some problems with how fallback and receive functions are handled.  However, now that #4710 is merged, those problems shouldn't get in the way anymore, and it should be possible to combine them again.

## Environment

* Truffle version (`truffle version`): 5.4.33

leeftk:
Fixed in latest release.

Issue #3878, opened 2021-02-25
Add redundant method for determining external function calls in stacktraces


haltman-at:
## Issue

Currently, function names for external functions are determined purely by looking at AST nodes when stacktracing.  One could put in a redundant system here where we also check the selector (like `txlog` does).  Note this might require some reorganizing; currently only `data` and `txlog` are set up to go checking selectors like that.

PR #4755, opened 2022-02-19
Re-add system to skip uninformative frames in stacktraces

haltman-at:
Addresses #4739.  This PR re-adds the system that I removed in #2976.  As such there's not really much new here, it's mostly just adding the deleted code back in and then adjusting the tests appropriately.  I did rename some of the fields and variables because I feel like the old names could have been clearer.  There is one new thing I had to add -- but I'll get to that.

Anyway, to explain: As of Solidity 0.5.1, when we make an external call, there's a jump-in marking, so without any additional work, `stacktrace` would generate two stackframes; one for the external call, then one for the jump-in.  In the `solidity` submodule, we have the phantom guard system to ensure that these get combined.  It wasn't being applied here.  Now we apply it here.  (The reason it wasn't being applied here before was that due to #4697, the phantom guard was activated inappropriately in some situations, and so applying it here would have caused some stackframes to be skipped.  Now that #4697 is resolved, this is no longer a problem.)

So, when adding a frame to the stack in `stacktrace`, we now add two additional fields: `combineWithNextInternal`, and `sourceIsInternal`.  These are both booleans; the first is set for external calls for which the phantom guard is set, indicating that, when generating the final stacktrace, this frame should be combined with the one after it, so long as the one after it is an internal call (and not to an internal source, i.e., unmapped code or a generated source).  The second, `sourceIsInternal`, we actually only include for internal calls (since it's meaningless for external ones), and records whether the source being jumped to is an internal source (meaning, again, unmapped code or a generated source).

Then, in the `generateReport` function in the selectors, we now add a preprocessing step where we go over the list of stackframes and combine any pair where the lower is marked `combineWithNextInternal` and the upper is of `type: "internal"` (i.e., an internal call) and not marked `sourceIsInternal` (not to an internal source).  (Sorry about the terminological overloading, but, that's what we've been calling these.)  "Combining" them means we take most of the information from the upper one (since that's the actual function call), but we take the `calledFromLocation` information from the lower one (since we want to treat the whole thing as a single call).  You might think we'd take `address` from the lower one as well, but actually we don't need to, since they'll already have the same address.

Actually, I said above that #4697 is now solved, but that's not exactly true; there is one case where the phantom guard will still be set inappropriately, and that's when calling a getter.  However, since getters don't call other functions (other than perhaps generated functions that won't get combined anyway), this ends up not causing any problems.  (Other than that the getter won't have its name displayed... this is making me wonder whether we should do something about #3878.)

Of course I had to adjust the tests.  Note I didn't bother adding a test for stacktraces starting in an internal source, to check that those don't get combined; that would have been difficult to write a test for, so I just tested it manually instead.  I also added three new tests, to test that stackframes that truly shouldn't get combined don't.  While I was at it, I also noticed that the existing tests didn't test the `stacktrace` submodule's reporting of the revert message, so I added that to the tests too.

Also note that I didn't need to make any changes to debug-utils; the `formatStacktrace` function in debug-utils is already prepared to handle the case of a stackframe with both a `functionName` and an `address`, even though that has never existed previously (since the `address` field was only added after I removed this system in #2976).

(Edit: I'm wrong, it did previously exist, I forgot that for a while we printed address on every stackframe!)

### DIFF
--- a/packages/debugger/test/stacktrace.js
+++ b/packages/debugger/test/stacktrace.js
@@ -50,7 +50,7 @@ contract StacktraceTest {
 
   function runRequire(bool succeed) public {
     emit Num(1); //EMIT
-    require(succeed); //REQUIRE
+    require(succeed, "requirement failed"); //REQUIRE
   }
 
   function runPay(bool succeed) public {
@@ -83,8 +83,10 @@ contract StacktraceTest {
     fail(); //CANTFALLBACK
   }
 
+  error CustomFailure();
+
   function fail() public {
-    revert("Nope!"); //FALLBACKFAIL
+    revert CustomFailure(); //FALLBACKFAIL
   }
 
   function runFallback(bool succeed) public {
@@ -220,6 +222,7 @@ describe("Stack tracing", function () {
     let prevLocation = report[report.length - 2].location;
     assert.strictEqual(location.sourceRange.lines.start.line, failLine);
     assert.strictEqual(prevLocation.sourceRange.lines.start.line, callLine);
+    assert.strictEqual(report[0].message, "requirement failed");
   });
 
   it("Generates correct stack trace at an intermediate state", async function () {
@@ -386,6 +389,7 @@ describe("Stack tracing", function () {
     let prevLocation = report[report.length - 4].location; //similar
     assert.strictEqual(location.sourceRange.lines.start.line, failLine);
     assert.strictEqual(prevLocation.sourceRange.lines.start.line, callLine);
+    assert.strictEqual(report[0].panic.toNumber(), 0x51);
   });
 
   it("Generates correct stack trace on unexpected self-destruct", async function () {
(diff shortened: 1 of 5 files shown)


cds-amal:
> Of course I had to adjust the tests. Note I didn't bother adding a test for stacktraces starting in an internal source, to check that those don't get combined; that would have been difficult to write a test for, so I just tested it manually instead.

I'd like to learn how your technique to test this.

haltman-at:
Oh, good question.  I tested that case by debugging a transaction that uses generated sources, stepping into the generated sources, and then using the `s` command to view the current callstack.  (Not exactly a stacktrace per se, but it's using the same mechanisms.)

packages/debugger/test/stacktrace.js
```diff
@@ -80,21 +110,44 @@ contract Boom {
...
+  }
+}
+
+library Library {
+  function outerFail() external {
+    innerFail(); //CANTLIB
+  }
+
+  function innerFail() internal {
+    revert("Nope!"); //LIBFAIL
```
cds-amal, line 130: This is neat! It took me a moment to figure out.

cds-amal (approved):
Looks good @haltman-at.

Issue #4679, opened 2022-02-01
First line of deployment reports are displaying incorrectly


haltman-at:
## Issue

Something's up wit the first lines of deployment reports; see example below.

## Steps to Reproduce

Try deploying a contract, that should suffice!

## Actual Results

Every deployment report looks like this at the start:

```
1_initial_migration.js
======================

   Deploying 'Migrations'
   ----------------------
   ⠋ Blocks: 0            Seconds: 0   > transaction hash:    0x11f1e20b9fe2a3d465a517356ecb2eafe5c6ef70a85747ac43beabf6c31e3f22
   > Blocks: 0            Seconds: 0
   > contract address:    0x513Dfe39C05BFbEcb22AF30e681D1e08C2256A2b
```

The rest of the report is fine.  Note that this error occurs for each contract deployed, not just the first one, nor just the first one in a file.

## Environment

* Truffle version (`truffle version`): 5.4.31
* node version (`node --version`): 12.22.9

lsqproduction:
Is it the formating of the report - how transaction hash is on the side instead of below Blocks and Seconds?
Or is it something else?

haltman-at:
Yes, the transaction hash is on the side; but also, blocks and seconds appears twice; and moreover, the first time it appears, it has an incomplete spinner next to it, rather than a `>`.

cds-amal:
@eggplantzzz also pointed out that there's extra crud at the end of migration.

lsqproduction:
Did this got fixed?
my initial migration looks pretty. Or are there other reproduction steps? I did truffle develop and then migrate.
```
1_initial_migration.js
======================

   Deploying 'Migrations'
   ----------------------
   > transaction hash:    0xad5f99688f993f3c04f1020171329bfc158e449c6713fdb8c79f0f6a6ebc4a4c
   > Blocks: 0            Seconds: 0
   > contract address:    0x8c234caaE6Dd6bB7E189FA20Aa9440D358ee5164
   > block number:        1
   > block timestamp:     1643997505
   > account:             0x512292D7C303E288dAeFB68607f8ae9b7A972DBE
   > balance:             99.999441521875
   > gas used:            165475 (0x28663)
   > gas price:           3.375 gwei
   > value sent:          0 ETH
   > total cost:          0.000558478125 ETH


   > Saving migration to chain.
   > Saving artifacts
   -------------------------------------
   > Total cost:      0.000558478125 ETH
```

the end of migration is still ugly. 

```
[...]

PR #4759, opened 2022-02-21
Remove junk from migration output

lsqproduction:
Addresses #4679 and other undefined in the migration output.


gnidan (approved):
Looks good! Thanks for the digging @lsqproduction

cds-amal (approved):
I approve! The tests passed and I verified a migration against ropsten and ganache with a 15s blocktime.

PR #4760, opened 2022-02-21
Dependency update: Update solc/wrapper to 0.8.12

haltman-at:
Apparently, as of Solc 0.8.12, `solc/wrapper` has now been made considerably smaller in size?  So, that seems like a good thing to update to.  This PR does that.

### DIFF
--- a/packages/compile-solidity/package.json
+++ b/packages/compile-solidity/package.json
@@ -36,7 +36,7 @@
     "original-require": "^1.0.1",
     "require-from-string": "^2.0.2",
     "semver": "^7.3.4",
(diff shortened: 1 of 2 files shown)


eggplantzzz:
And this stuff still works with older versions of the Solidity compiler?

haltman-at:
Yes, solc in this package is just used for `solc/wrapper`, not solc itself.  Note that if the new `solc/wrapper` version didn't work with old Solidity versions, all of our many tests using old Solidity versions would have failed.

eggplantzzz (approved):

Issue #3742, opened 2021-01-27
Feature request: Custom autocompletes for Truffle CLI


haltman-at:
## Issue

Currently the `truffle` CLI command has no custom autocompletes.  It would be nice to have these.  Ideally we'd have them both for bash and for zsh.  Definitely there should be autocompletes for what truffle command to use (so that `truffle migr` would autocomplete to `truffle migrate`, not `truffle migrations/`), and then perhaps individual commands could also have autocompletes for their individual options.  One could possibly even go further, autocompleting names of contract, networks, test files, boxes, ethpm packages, `exec`/`run` commands, solc versions... but at the least there should be autocompletes for the commands themselves, so that `truffle migr` autocompletes correctly. :)

## Environment

* Truffle version (`truffle version`): 5.1.63

Xuefeng-Zhu:
@davidmurdoch Hi David, I am interesting in working on truffle Ethdenver bounty. Is there a way that I can reach out truffle dev on EthDenver Discord? Thanks

davidmurdoch:
@Xuefeng-Zhu thanks for the interest in the bounty! I'll be away from my desk all weekend, so unfortunately I won't be able to help. I'll ping some people in our internal chat to try to find someone to assist you further.

cds-amal:
Hi @Xuefeng-Zhu I'll be attending this weekend and will be lurking on the ethDenver discord channel.

Xuefeng-Zhu:
@cds-amal cool. what is your discord handler?

cds-amal:
@Xuefeng-Zhu I'm this [discord server](https://discord.gg/JfTdgdYv)

Issue #3527, opened 2020-11-12
Feature request: Customized tab-completion for bash and zsh (possibly other shells?)


haltman-at:
## Issue

Truffle has no customized tab-completion for its commands; adding this would make it easier to use.

## Expected Behavior

Probably just tab-completion of commands to start.  Not clear that further customized tab-completion is worth it.

ragnarok22:
Excelent issue! @truffle-teams

haltman-at:
Closing this as duplicate of (the later but more informative) #3742.  Oops!

PR #4761, opened 2022-02-21
CLI tab-complete

hughes-ch:
Addresses #3527 (also #3742 - didn't see the associated PR until today)

This uses the yargs tab-complete feature for bash and zsh. Tab-complete is installed automatically during a postInstall script, so theoretically all you need to do is reload your shell after installing truffle. You can also install manually with:

    truffle autocomplete install

To uninstall, use 

    truffle autocomplete uninstall

A lot of the changes are to make the tab-complete more responsive by loading modules within functions instead of at the top of modules. It's still not as quick as I'd like, but optimizing cli.bundled.js seemed to be out of the scope of this PR. I can look further into that as a separate issue.

Other changes include two new commands (shown above) and updates to the latest version of yargs to get the latest tab-complete functionality.


haltman-at:
Oh, whoa, I totally didn't notice that we had both #3527 and #3742, oops!  I'll go close one of them.  (...yes that's not a comment on your actual PR, you only just opened that so I can't claim to have read it yet!  But uh thanks for pointing out that duplication that we'd missed...)

haltman-at:
Btw, you may have to update your PR to fix the merge conflicts before people will be willing to review it. :P

hughes-ch:
Yeah, I should have merged before opening this PR. Will do that now.

haltman-at:
Hm, so I've rerun the tests here a few times and I keep getting the same failures... I don't know why this PR would interfere with those tests, but whatever's going on, it looks like there is a real problem and it's not just spurious failures.

hughes-ch:
Bummer - I'll take a look and see what's going on

hughes-ch:
Looks like the newer version of yargs was exiting the process early during this.args.parse (in command.js). For some reason, it didn't recognize an argument to compile (at least in the unit test I was troubleshooting) and just printed the help menu rather than trying to compile. 

Turns out I don't even need the updated version of yargs. Rather than going down the rabbit hole of trying to update versions, I can just use the version that's currently on the develop branch.

hughes-ch:
The `autocomplete` command relies on `process.env.SHELL` to determine whether to update bash or zsh settings. Looks like `process.env.SHELL` is not set in docker containers. I've updated the unit tests so that they do not break in that case.

haltman-at:
Oh, huh.  Thanks!  Seems there's some new problem now though...

hughes-ch:
Yeah, this one does not want to go through! 

But it looks like fs/promises might not be available on older versions of Node, so one of my unit tests are encountering an error. I ran out of time today to look at this, but I can try to address it again tomorrow.

haltman-at:
Hm, well Node 12 does have `fs.promises`, you just can't import it as `fs/promises`; you have to get at it via `fs`.  So, if you want to continue using `fs.promises`, you can do it that way.

hughes-ch:
That looks like it did it. Thanks for the help!

haltman-at:
Hi @hughes-ch sorry to bug you about this, but we just recently updated our CI to include tests on Node 16... would you perhaps mind merging/rebasing again so that this can run with the Node 16 tests?  Thanks!  (And hopefully we can get this reviewed soon...)

gnidan:
This PR is freaking awesome. I'll make sure we get reviews on your PR over the next week or so. Thank you so much for taking this on!

hughes-ch:
I hope you're still as excited about it after looking at my code!

@haltman-at I'll try to get to it as soon as I can. I'm expecting to have some free time either this weekend or early next week.

hughes-ch:
Well it looks like I pulled a ton more commits in than I intended on the rebase. Let me see if there's anything I can do to just get my changes in this PR...

gnidan:
It's hard to review this now with all the extra commits, but I recall from looking briefly at this last week that this approach writes to `.bashrc` (et al.). Is that right?

If so, I think there's a more proper way to do things... I think it uses like a `autocomplete.d/` directory? Would prefer not to write to user config files like `.bashrc` if we can avoid it.

hughes-ch:
Yes, it writes to `.bashrc` (and similar). 

I found [this stack exchange question](https://serverfault.com/questions/506612/standard-place-for-user-defined-bash-completion-d-scripts) which points to using `.bash_completion` (I'm sure there's an equivalent for zsh). I just tested it out and it seems to work. I agree that editing this would be a lot better than `.bashrc`.

gnidan:
> I found [this stack exchange question](https://serverfault.com/questions/506612/standard-place-for-user-defined-bash-completion-d-scripts) which points to using `.bash_completion` (I'm sure there's an equivalent for zsh). I just tested it out and it seems to work. I agree that editing this would be a lot better than `.bashrc`.

Hm yeah I'll need to search more thoroughly for zsh (which is what I use)... does zsh provide such a mechanism by default? I found [https://github.com/marlonrichert/zsh-autocomplete](https://github.com/marlonrichert/zsh-autocomplete) but it looks third-party.

Anyway, maybe we target `.bash_completion` instead of `.bashrc` to start. Could you update your PR to do that instead of `.bashrc`? And in the meantime, I'll keep reading about zsh and see if I can figure out a proper recommendation.

hughes-ch:
Ah, I didn't see that when originally looking through the Server Fault question. I agree that using the directory would be much better than `.bash_completion`. I'll give it a shot and update the PR.

gnidan:
Thank you @hughes-ch!

haltman-at:
Yes thank you! :D

hughes-ch:
So I've tested out placing completion settings in `~/.local/share/bash-completion/completions` and it seems to work well. There's no corresponding directory that I can find for zsh, though. The best I can come up with is to create a similar `~/.local/share/zsh-completion/completions` that zsh users would need to add to their `$fpath`.

hughes-ch:
I'll try to at least have the PR updated for bash completion settings tomorrow, though. Sorry for the slow turnaround time! I was hoping I'd have more time this week.

haltman-at:
No problem, there's no hurry. :)

hughes-ch:
Updated to use completion directory for bash. Will continue to look into zsh next week.

hughes-ch:
After looking into zsh further, I don't think there's any support for automatically adding tab-complete without modifying `.zshrc` directly, unlike with bash. Actually, neither bash nor zsh support the [XDG standard](https://wiki.archlinux.org/title/XDG_Base_Directory) to load from `.local/share/`. However, most Unix machines come with the `bash-completion` project which adds this functionality to bash, which I'm leveraging in my last commit. 

Can we accept modifying `.zshrc` for zsh? Or should I code as much as I can, then document what zsh users need to add to their `.zshrc` to get tab-completions working?

cds-amal:
Nice work on this @hughes-ch, I'd be ok with printing the commands with instructions to add it to `.zshrc`. @gnidan, what do you think?

Adding [this to Reddit thread to supplement](https://www.reddit.com/r/zsh/comments/t94p9o/does_zsh_use_local/) this discussion, the OP's name checks out.

haltman-at:
@hughes-ch Huh, I thought zsh had a similar mechanism, where you don't have to write directly to `.zshrc` to install autocompletions?  Going by [this link](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command) it sounds like it does.  That said, I am no way an expert on zsh, so it's possible I've misunderstood here.

hughes-ch:
Unfortunately zsh-completions doesn't seem to be installed by default on many distributions (unlike bash-completions). I didn't want to have to rely on something that would have to be installed after the fact. I'm not a zsh expert either, so I don't know if zsh-completions is common enough that I can assume that most zsh users will already be using it. If it turns out it *is* really common, then I'll certainly support it.

haltman-at:
Ah, I see.  Oh well.  Thanks for clarifying that.  I guess this may just be the way to go, then.

gnidan:
> I don't know if zsh-completions is common enough that I can assume that most zsh users will already be using it. If it turns out it _is_ really common, then I'll certainly support it.

Hm, is there a way to implement this to use zsh-completions if it's available, and otherwise fallback to .zshrc?

gnidan:
also @hughes-ch... if your appetite for this PR is waning, I don't want to add additional burden. I'm also fine with an initial implementation that only uses .zshrc

and not to be a pestering middle manager, but I'd be remiss if I didn't mention that Truffle's hiring for [a number of positions](https://consensys.net/open-roles/?discipline=32535&role=), and if you're available/interested, you could apply to work on this sort of thing for money 💸

hughes-ch:
> Hm, is there a way to implement this to use zsh-completions if it's available, and otherwise fallback to .zshrc?
I can do that. I can try to have a PR ready this weekend or early next week. 

> Truffle's hiring for a number of positions
I'd actually love to talk to someone about this. Do you mind if I message you on the Twitter handle you have listed on your profile? I can't take a full time job at the moment, but I was wondering if there's something that I could help out with part time.

gnidan:
> I can do that. I can try to have a PR ready this weekend or early next week.

Sweet!

> I'd actually love to talk to someone about this. Do you mind if I message you on the Twitter handle you have listed on your profile? I can't take a full time job at the moment, but I was wondering if there's something that I could help out with part time.

Yep, I'm `@gnidan` most places :). Definitely you can confirm the Twitter account is me via the slick 1-of-1 NFT depicting me and my parents' neighbors' late cat. We're totally open to part-time contractors! Looking forward to hearing from you!

hughes-ch:
Looked further into `zsh-completions`. Unfortunately, unlike `bash-completions`, it doesn't use a completions directory in the user's home directory. Zsh and all of its popular frameworks rely on completions being placed in `$fpath` as @haltman-at [mentioned above](https://github.com/zsh-users/zsh-completions/blob/master/zsh-completions-howto.org#telling-zsh-which-function-to-use-for-completing-a-command). By default, `$fpath` only has paths in the `/usr` directory, so placing completions there would require elevated privileges. Even if we appended to `$fpath`, we'd have to write to `.zshrc` anyways. 

I've decided to stick with a similar method to how yargs' completion mechanism is set up instead of modifying `$fpath`. I'm still certainly open to other ideas, but I'm not able to find any ways to add completion without modifying `.zshrc` directly.

cds-amal:
> I've decided to stick with a similar method to how yargs' completion mechanism is set up instead of modifying `$fpath`. I'm still certainly open to other ideas, but I'm not able to find any ways to add completion without modifying `.zshrc` directly.

@hughes-ch, @gnidan  thoughts on printing the commands for the user to copy their `.zshrc`? Or ask the user for permission before modifying their `.zshrc`  ?

hughes-ch:
I'll defer to @gnidan here, but I'd be OK with something similar. 

Since tab completion is added to `.zshrc` during install of the truffle npm package, there's no opportunity for a user to interact. We won't be able to ask permission, and I don't think `stdout` is shown to the user during an `npm install`. I think the best solution here if not directly editing `.zshrc` would be to update the README about how to install tab completion for zsh users.

benjamincburns:
Sorry guys, I'm late to the party here. Could I maybe suggest a more straightforward approach that's likely better for the user?

There's a fairly well-established pattern for how to accomplish this w/o having to worry about how the user setup their dotfiles.

The gist is that a program that brings its own completions should have a `completion` or `completions` command, and the completions script should simply be written to stdout. Typically this command allows you to specify your particular shell as the first positional argument (e.g. `truffle completion bash`).

This allows users of `bash` and `zsh`, for example, to just chuck a `source <(truffle completion $(basename $(echo $SHELL)))` into their `.bashrc`, `.profile`, `.zprofile`, `.zshrc`, or wherever else they like.

This approach has the following benefits:

- works with however I've set up my dotfiles
- doesn't require me to change any of my dotfiles to get latest completions when I upgrade `truffle`
- allows the user to explicitly request a specific type of shell


Some examples of tools that do it this way:

- [`helm`](https://helm.sh/docs/helm/helm_completion/#see-also)
- [`kubectl`](https://kubernetes.io/docs/tasks/tools/included/)

hughes-ch:
Good idea @benjamincburns I might have been too ambitious trying to do this automatically. I'll make sure to add instructions to the README. 

One question I had - the CLI logic has command inference built in (so that "truffle c" is inferred as "truffle compile"). I didn't want to break "truffle compile" inference with "truffle completion" so I chose "truffle autocomplete" as the command. Is that worth breaking naming convention here?

gnidan:
> One question I had - the CLI logic has command inference built in (so that "truffle c" is inferred as "truffle compile"). I didn't want to break "truffle compile" inference with "truffle completion" so I chose "truffle autocomplete" as the command. Is that worth breaking naming convention here?

Good point! Shame that those two share so many letters :)

I think `truffle autocomplete` sounds good. Should we namespace it further? I was thinking `truffle autocomplete install`... not that we're likely to add `truffle autocomplete uninstall`, but still, might be nice for it to read like a more complete verb phrase. Thoughts?

hughes-ch:
I think I'd prefer just using `truffle autocomplete`. There's a lot of boilerplate to add a subcommand (including further nested directories). I'd like to avoid paths that look like `core/lib/commands/autocomplete/commands/install/commands...` if possible :)

gnidan:
> I think I'd prefer just using `truffle autocomplete`. There's a lot of boilerplate to add a subcommand (including further nested directories). I'd like to avoid paths that look like `core/lib/commands/autocomplete/commands/install/commands...` if possible :)

Yeah no one wants that. Sounds good!

hughes-ch:
Stuck with `truffle autocomplete`, but took out the auto-install feature and added documentation for the manual process. Folks will be able to find how to use the new command via the help menu or [via the online docs](https://github.com/trufflesuite/trufflesuite.com/pull/1232).

benjamincburns:
On their face most of these changes look fine, but there appear to be a bunch of changes in this PR that aren't directly related to the goal of providing a CLI tab-completion feature.

I think normally the process would be for us to ask that you please only submit the changes pertaining to the goal of the PR, however I see @gnidan has reviewed some of this already, so I'll defer to him.

hughes-ch:
If you are referring to the changes in `packages/core/cli.js` and `packages/core/lib/command.js`, it is true that most of them are not needed for basic CLI tab-completion to work. I included them because they improve performance of the tab-completion feature. Anecdotally, without those changes, it took a second or two longer for the tab-completion to populate. I figured that the changes would be beneficial enough to include them in the first round of changes because of the huge user experience improvement. 

If it's decided, I can certainly remove them and add them in a second, separate PR.

benjamincburns (approved):
Thanks for the explanation about what appeared to be unrelated changes. Performance improvements alongside of new functionality is always a bonus, IIMO.

I would've thought that moving the `require`s out of the top level of the module would mess with webpack somewhat, but the green build says otherwise.

benjamincburns:
Would merge, but I'm new to the team and still learning the ropes around process for accepting outside contributions. Per our [dev handbook](https://github.com/trufflesuite/dev-handbook/blob/master/git-practices.md#reviews), it seems we need two approved reviews before merging.

Most of the team should be back next week, so hopefully we can get this merged in time for the next release!

haltman-at:
Wait, so there's going to be absolutely no auto-install, not even for bash?  To my mind that kind of defeats the point... but I guess if we're doing this that can always be added in later...?

gnidan:
> Wait, so there's going to be absolutely no auto-install, not even for bash? To my mind that kind of defeats the point... but I guess if we're doing this that can always be added in later...?

If we want to add it later, sure!

But I think the explicit `truffle autocomplete` is perhaps the better approach, anyway. Maybe later we can add functionality to suggest `truffle autocomplete` every so often, to users that don't have it enabled.

I'll review this later this week. Thanks for doing all you could to get this over the finish line, @hughes-ch!

hughes-ch:
Just rebased to fix a merge error that I noticed this weekend. With this last rebase, As I was walking through it, I realized some of my commits had been duplicated so I cleaned up the history a little bit and had to force-push. Now this should be a lot easier to merge in. 

Also, 584ac58 includes a change to bring a long-running `require` statement into `module.exports`. This should allow truffle's CLI to start much quicker. It's also really necessary to the CLI tab-completion since a long lead-time between pressing the "tab" and seeing the completion populate makes it almost unusable.

gnidan:
Haven't forgotten about this, sorry for the delay! I will try to make sure we get another review on this by the end of the week.

eggplantzzz:
Closing this as it went stale and now there are more problems that we can't resolve; in addition to having a lack of resources at the moment.

Issue #4756, opened 2022-02-19
Phantom call guard should not be set when entering a Yul contract


haltman-at:
## Issue

Oops, looks I haven't yet quite gotten function depth correct once and for all.  The phantom call guard will be set when entering a Yul contract, causing the next jump-in to be ignored.  That shouldn't happen; Yul contracts should not have the guard set.  Their first jump-in should take them to 1, not 0.

## Environment

* Truffle version (`truffle version`): 5.5.0

lsqproduction:
Addressed in the release v5.5.2.

PR #4762, opened 2022-02-22
Bug fix: Fix function depth for Yul contracts

haltman-at:
Addresses #4756.  It turns out I wasn't *quite* done with fixing the function depth -- I still hadn't fixed it for Yul contracts!  Yul contracts shouldn't have the phantom call guard set, as when you call them, you are legitimately running outside a function at first.  So, we want to restrict the phantom call guard to when the language is Solidity.  This PR does that.

(What about Vyper, you may ask?  Vyper also shouldn't have the guard set, because like Solidity <0.5.1 the initial function call lacks a jump marking; but this was already handled by the check that the guard only applies to things compiled with `solc`.)

To do this, we have to check the language of the context's primary source; we only set the guard if the language is `"Solidity"`.  Problem: How can we check this?  Once we're *in* the context, we can look up the context's primary source and check its language... but the call guard has to be determined *before* we enter the context.  And until we enter the context, we don't have its sources available.  (OK, now that I look at it more, actually we *could* make those sources available if I had taken an alternate approach, but I didn't think of that possibility until just now, so, uh, I didn't do that.)

Instead, the way I decided to handle it was to attach a new `primaryLanguage` field to each context, determined at startup.  So we can just check `context.primaryLanguage`.

While I was doing this, I decided to simplify the code for adding contexts, because having to maintain that each time I add something new is annoying.  Not that I add new things often, but!

I also added a test to ensure that function depth hits 1 when inside a Yul function call.  Anyway, now function depth should *finally* be correct.  (Until Vyper decides to change things or we add a new language or something.)

### DIFF
--- a/packages/debugger/lib/evm/actions/index.js
+++ b/packages/debugger/lib/evm/actions/index.js
@@ -1,34 +1,8 @@
 export const ADD_CONTEXT = "EVM_ADD_CONTEXT";
-export function addContext({
-  context,
-  contractName,
-  binary,
-  sourceMap,
-  primarySource,
-  immutableReferences,
-  compiler,
-  compilationId,
-  abi,
-  contractId,
-  contractKind,
-  isConstructor,
-  linearizedBaseContracts
-}) {
+export function addContext(context) {
   return {
-    type: ADD_CONTEXT,
-    context,
-    contractName,
-    binary,
-    sourceMap,
-    primarySource,
-    immutableReferences,
-    compiler,
-    compilationId,
-    abi,
-    contractId,
-    contractKind,
-    isConstructor,
-    linearizedBaseContracts
+    ...context,
+    type: ADD_CONTEXT
   };
 }
 
--- a/packages/codec/lib/contexts/types.ts
+++ b/packages/codec/lib/contexts/types.ts
@@ -46,6 +46,7 @@ export interface DebuggerContext {
   abi?: Abi.Abi;
   sourceMap?: string;
   primarySource?: number;
+  primaryLanguage?: string;
   compiler?: Compiler.CompilerVersion;
   compilationId?: string;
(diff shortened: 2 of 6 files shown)


cds-amal:
> To do this, we have to check the language of the context's primary source; we only set the guard if the language is `"Solidity"`. Problem: How can we check this? Once we're _in_ the context, we can look up the context's primary source and check its language... but the call guard has to be determined _before_ we enter the context. And until we enter the context, we don't have its sources available. (OK, now that I look at it more, actually we _could_ make those sources available if I had taken an alternate approach, but I didn't think of that possibility until just now, so, uh, I didn't do that.)
> 
> Instead, the way I decided to handle it was to attach a new `primaryLanguage` field to each context, determined at startup. So we can just check `context.primaryLanguage`.
> 

Thanks for that context :P <3

packages/debugger/lib/evm/actions/index.js
```diff
@@ -1,34 +1,8 @@
...
-  immutableReferences,
-  compiler,
-  compilationId,
-  abi,
-  contractId,
-  contractKind,
-  isConstructor,
-  linearizedBaseContracts
-}) {
+export function addContext(context) {
```
cds-amal, line 2: Would you consider `createContextAction` as this seems like an action creator?

cds-amal (approved):
Thanks @haltman-at Looks good to me. I left a suggestion for a name change, but it is your call.

packages/debugger/lib/evm/actions/index.js
haltman-at: Yeah, I don't really want to go renaming this.  Also, all of our actions work like this?  Like, I'd have to rename every single action function to `createEtcAction`, and then use those names everywhere... that would be so wordy!

Issue #4327, opened 2021-09-24
No way to set Truffle network config options for Truffle Test's or Truffle Develop's internal Ganache network


haltman-at:
## Issue

Ordinarily, in Truffle, it is possible to set per-network configuration options in one's Truffle config.  However, as best I can tell, there is no way to do this for the ephemeral networks created by Truffle Test and Truffle Develop.  (The reason is different for each, so arguably this should be two issues, but I'm grouping them together.)  The Truffle Test issue has come up twice recently in other discussions (#3780, #4319) so I'm making an issue for it.

Let's start with Truffle Develop, as that's the easier case.  In the [`Environment.develop` function](https://github.com/trufflesuite/truffle/blob/develop/packages/environment/environment.js#L73), you can see that `config.networks[network]` is overwritten with a custom, barebones, network config.  Presumably, the new network config should keep other existing config options in place, and only overwrite `network_id`, `provider`, and other ways of specifying a provider; leaving in place such things as `gas`, `disableConfirmationListener`, etc.  This problem actually also applies to Truffle Test, as it also uses `Environment.develop`.

But troublesome problem with Truffle Test, and the one people have been asking about, is a different one.  Let's look at the [current behavior](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/commands/test/index.js#L174).  Currently (a few lines above where I linked), if no network is specified, it will default to `"development"` or `"test"` (depending on if a config for `"development"` has been defined), and then (at the point where I linked), it will start up an internal Ganache (ordinarily called `"test"`) precisely if... no config is defined for the current network (whether explicitly selected or defaulted to).

This makes it impossible to configure this internal network.  A user might think, oh, well, the internal network is called `"test"`, so to configure it I'll add a config for the network `"test"`.  But as soon as they do so, Truffle will now say, oh, well since you have a config for `"test"`, you must have been intending to connect to some external network called `"test"`, and we won't start up Ganache for you.  That's not what the user wanted!  And people have been asking about this.  But right now there is no way to configure that internal network without Truffle concluding that want an external network instead.

[...]

PR #4772, opened 2022-02-23
Set Truffle network config options for Truffle test's internal ganache network

sukanyaparashar:
### ISSUE
No way to set Truffle network config options for Truffle Test's internal Ganache network. See issue #4327 

### SOLUTION
This PR solves the issue of non-configurable Truffle test's managed ganache network from the truffle-config.js.
If the configured network is "test" and provider, host or url doesn't exist, or there is no configured network defined, then managed ganache is started overriding with the user specified configurations or the default configurations.
If there is a configured network in the config, then it connects to that external network.

### DIFF
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -82,7 +82,6 @@ module.exports = async function (options) {
     !configuredNetwork["url"];
   const port = await require("get-port")();
   const ipcOptions = { network: "test" };
-  let numberOfFailures;
 
   if (testNetworkDefined && noProviderHostOrUrlConfigured) {
     const ganacheOptions = {
@@ -97,19 +96,21 @@ module.exports = async function (options) {
       },
       ...config.networks[config.network]
     };
-    numberOfFailures = await startGanacheAndRunTests(
+    const numberOfFailures = await startGanacheAndRunTests(
       ipcOptions,
       ganacheOptions,
       config
     );
+    return numberOfFailures;
   } else if (configuredNetwork) {
     await Environment.detect(config);
     const { temporaryDirectory } = await copyArtifactsToTempDir(config);
-    numberOfFailures = await prepareConfigAndRunTests({
+    const numberOfFailures = await prepareConfigAndRunTests({
       config,
       files,
       temporaryDirectory
     });
+    return numberOfFailures;
   } else {
     const ganacheOptions = {
       host: "127.0.0.1",
@@ -122,13 +123,13 @@ module.exports = async function (options) {
         instamine: "strict"
       }
     };
-    numberOfFailures = await startGanacheAndRunTests(
+    const numberOfFailures = await startGanacheAndRunTests(
       ipcOptions,
       ganacheOptions,
       config
     );
+    return numberOfFailures;
   }
-  return numberOfFailures;
 
   // Start internal ganache network
   async function startGanacheAndRunTests(ipcOptions, ganacheOptions, config) {
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -68,7 +68,42 @@ module.exports = async function (options) {
     inputFile: file
   });
 
-  if (config.networks[config.network]) {
+  /**
+   * If test network exists and provider, host or url exists, then it connects to that external network.
+   * If test network exists and provider, host or url doesn't exist, then internal ganache is started with the user specified
+   * configurations in the config.
+   * If no network is specified in the config, then internal ganache is started with default configuration.
+   * If development network exists, then it connects to the development network specified in the config.
+   */
+  if (
+    config.networks[config.network] &&
+    config.network == "test" &&
+    !config.networks[config.network]["provider"] &&
+    !config.networks[config.network]["host"] &&
+    !config.networks[config.network]["url"]
+  ) {
+    const ipcOptions = { network: "test" };
+    const port = await require("get-port")();
+
+    const ganacheOptions = {
+      host: "127.0.0.1",
+      port,
+      network_id: 4447,
+      mnemonic:
+        "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat",
+      time: config.genesis_time,
+      miner: {
+        instamine: "strict"
+      },
+      ...config.networks[config.network]
+    };
+    const numberOfFailures = await startGanache(
+      ipcOptions,
+      ganacheOptions,
+      config
+    );
+    return numberOfFailures;
+  } else if (config.networks[config.network]) {
     await Environment.detect(config);
     const { temporaryDirectory } = await copyArtifactsToTempDir(config);
     const numberOfFailures = await prepareConfigAndRunTests({
@@ -92,6 +127,16 @@ module.exports = async function (options) {
         instamine: "strict"
       }
     };
+    const numberOfFailures = await startGanache(
+      ipcOptions,
+      ganacheOptions,
+      config
+    );
+    return numberOfFailures;
+  }
+
+  // Start internal ganache network
+  async function startGanache(ipcOptions, ganacheOptions, config) {
     const { disconnect } = await Develop.connectOrStart(
       ipcOptions,
       ganacheOptions
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -69,22 +69,22 @@ module.exports = async function (options) {
   });
 
   /**
-   * If test network exists and provider, host or url exists, then it connects to that external network.
    * If test network exists and provider, host or url doesn't exist, then internal ganache is started with the user specified
    * configurations in the config.
+   * If development or test network with provider, host or url exists, then it connects to that network specified in the config.
    * If no network is specified in the config, then internal ganache is started with default configuration.
-   * If development network exists, then it connects to the development network specified in the config.
    */
-  if (
-    config.networks[config.network] &&
-    config.network == "test" &&
-    !config.networks[config.network]["provider"] &&
-    !config.networks[config.network]["host"] &&
-    !config.networks[config.network]["url"]
-  ) {
-    const ipcOptions = { network: "test" };
-    const port = await require("get-port")();
+  const configuredNetwork = config.networks[config.network];
+  const testNetworkDefined = configuredNetwork && config.network == "test";
+  const noProviderHostOrUrlConfigured =
+    !configuredNetwork["provider"] &&
+    !configuredNetwork["host"] &&
+    !configuredNetwork["url"];
+  const port = await require("get-port")();
+  const ipcOptions = { network: "test" };
+  let numberOfFailures;
 
+  if (testNetworkDefined && noProviderHostOrUrlConfigured) {
     const ganacheOptions = {
       host: "127.0.0.1",
       port,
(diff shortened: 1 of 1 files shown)


packages/core/lib/commands/test/run.js
```diff
@@ -68,7 +68,42 @@ module.exports = async function (options) {
     inputFile: file
   });
 
-  if (config.networks[config.network]) {
+  /**
+   * If test network exists and provider, host or url exists, then it connects to that external network.
+   * If test network exists and provider, host or url doesn't exist, then internal ganache is started with the user specified
+   * configurations in the config.
+   * If no network is specified in the config, then internal ganache is started with default configuration.
+   * If development network exists, then it connects to the development network specified in the config.
```
cds-amal, line 76: This comment describes 4 conditions, but the code has three. I think the first conditional isn't in the implementation.

packages/core/lib/commands/test/run.js
```diff
@@ -68,7 +68,42 @@ module.exports = async function (options) {
...
+   */
+  if (
+    config.networks[config.network] &&
+    config.network == "test" &&
+    !config.networks[config.network]["provider"] &&
+    !config.networks[config.network]["host"] &&
+    !config.networks[config.network]["url"]
```
cds-amal, line 83: It would be nice to refactor for readability? Some conditional fragments may be useful further down the logic branches.
```javascript
  const configuredNetwork = config.networks[config.network];
  const testNetworkDefined = configuredNetwork && config.network == "test";
  const noProviderHostOrUrlConfigured = !configuredNetwork["provider"] && !configuredNetwork["host"] && !configuredNetwork["url"];

  if (testNetworkDefined && noProviderHostOrUrlConfigured) {
    // do the things
```

packages/core/lib/commands/test/run.js
```diff
@@ -68,7 +68,42 @@ module.exports = async function (options) {
...
+  ) {
+    const ipcOptions = { network: "test" };
+    const port = await require("get-port")();
```
cds-amal, line 86: `port` is used in the other branches and could be factored out so it is only initialized once.

packages/core/lib/commands/test/run.js
```diff
@@ -92,6 +127,16 @@ module.exports = async function (options) {
...
+    const numberOfFailures = await startGanache(
+      ipcOptions,
+      ganacheOptions,
+      config
+    );
+    return numberOfFailures;
+  }
+
+  // Start internal ganache network
+  async function startGanache(ipcOptions, ganacheOptions, config) {
```
cds-amal, line 139: Maybe rename this function to `StartGanacheAndRunTests` is a better name for this function?

packages/core/lib/commands/test/run.js
```diff
@@ -68,7 +68,42 @@ module.exports = async function (options) {
...
+        instamine: "strict"
+      },
+      ...config.networks[config.network]
+    };
+    const numberOfFailures = await startGanache(
+      ipcOptions,
+      ganacheOptions,
+      config
+    );
+    return numberOfFailures;
```
cds-amal, line 105: The definition of `numberOfFailures` could be hoisted above the if statement, and each branch would set `numberOfFailures` and there would be one return statement at the end.

cds-amal (changes requested):
Thanks for this @sukanyaparashar. I noticed a discrepancy between the comment and code and made some refactoring/readability comments.

packages/core/lib/commands/test/run.js
sukanyaparashar: The first and the last comment can be joined I guess, because that is the 2nd condition in the else if branch. I will change that. Thanks @cds-amal .

packages/core/lib/commands/test/run.js
sukanyaparashar: Your suggestion in this code makes sense @cds-amal . I will make it more cleaner as suggested.

packages/core/lib/commands/test/run.js
sukanyaparashar: Sure, this name will make more sense.

packages/core/lib/commands/test/run.js
```diff
@@ -68,7 +68,41 @@ module.exports = async function (options) {
...
+   * configurations in the config.
+   * If development or test network with provider, host or url exists, then it connects to that network specified in the config.
+   * If no network is specified in the config, then internal ganache is started with default configuration.
+   */
+  const configuredNetwork = config.networks[config.network];
+  const testNetworkDefined = configuredNetwork && config.network == "test";
+  const noProviderHostOrUrlConfigured =
+    !configuredNetwork["provider"] &&
+    !configuredNetwork["host"] &&
+    !configuredNetwork["url"];
```
cds-amal, line 82: Oops, should verify `configuredNetwork` is not undefined! Sorry about that!
```suggestion
  const noProviderHostOrUrlConfigured =
    configuredNetwork &&
    !configuredNetwork.provider &&
    !configuredNetwork.host &&
    !configuredNetwork.url;
```

packages/core/lib/commands/test/run.js
```diff
@@ -68,19 +68,50 @@ module.exports = async function (options) {
...
+  const noProviderHostOrUrlConfigured =
+    configuredNetwork &&
+    !configuredNetwork.provider &&
+    !configuredNetwork.host &&
+    !configuredNetwork.url;
+  const port = await require("get-port")();
+  const ipcOptions = { network: "test" };
+  let numberOfFailures;
+
+  if (testNetworkDefined && noProviderHostOrUrlConfigured) {
```
haltman-at, line 88: This condition fixes the behavior with `test`, but I don't think it preserves the existing behavior in other cases.  To preserve current behavior, it seems to me that this case case should also be invoked if `config.networks[config.netowork]` is undefined.  (This is the reason I hit "request changes".)

packages/core/lib/commands/test/run.js
```diff
@@ -68,19 +68,50 @@ module.exports = async function (options) {
...
+   * If no network is specified in the config, then internal ganache is started with default configuration.
+   */
+  const configuredNetwork = config.networks[config.network];
+  const testNetworkDefined = configuredNetwork && config.network == "test";
+  const noProviderHostOrUrlConfigured =
+    configuredNetwork &&
+    !configuredNetwork.provider &&
+    !configuredNetwork.host &&
+    !configuredNetwork.url;
+  const port = await require("get-port")();
```
haltman-at, line 84: We probably shouldn't be getting a port unless we're actually in the managed case; I don't know how much work it actually takes to get a port, but it strikes me as probably something to be avoided in the cases where we don't need it.

packages/core/lib/commands/test/run.js
```diff
@@ -68,19 +68,50 @@ module.exports = async function (options) {
...
 
-  if (config.networks[config.network]) {
+  /**
+   * If test network exists and provider, host or url doesn't exist, then internal ganache is started with the user specified
+   * configurations in the config.
+   * If development or test network with provider, host or url exists, then it connects to that network specified in the config.
+   * If no network is specified in the config, then internal ganache is started with default configuration.
+   */
+  const configuredNetwork = config.networks[config.network];
+  const testNetworkDefined = configuredNetwork && config.network == "test";
```
haltman-at, line 78: I think this variable could use a better name; `testNetworkDefined` sounds like it refers to the `test` network, well, being defined, whereas what you're actually checking for is if it's defined *and* used.

haltman-at (dismissed):
Seems mostly good, but there's one part that looks wrong to me; hitting "request changes" as a result.  Also two more minor points about variable naming & where code is placed.

packages/core/lib/commands/test/run.js
sukanyaparashar: The last else branch actually checks for the undefined network. @cds-amal and I talked about this and I will make changes in the comments to make it more understandable.

packages/core/lib/commands/test/run.js
eggplantzzz: I would say that generally in the Truffle codebase prefer `===` over `==` in most cases unless you actually want the behavior of `==`.

sukanyaparashar: Sorry @eggplantzzz . I forgot about this again. I will correct it.

packages/core/lib/commands/test/run.js
sukanyaparashar: Yes, @eggplantzzz . I think you are correct. We can just make it to if/else. I am working on it and testing it. I will make a PR with this change in a few minutes.

packages/core/lib/commands/test/run.js
cds-amal: Nice rework on the refactoring @sukanyaparashar !

packages/core/lib/commands/test/run.js
cds-amal: ooh, yea, that's a good point on the variable name

packages/core/lib/commands/test/run.js
```diff
@@ -68,43 +86,48 @@ module.exports = async function (options) {
...
-      }
-    };
-    const { disconnect } = await Develop.connectOrStart(
+  if (
+    (testNetworkDefined && noProviderHostOrUrlConfigured) ||
+    !configuredNetwork
+  ) {
+    // Use managed ganache with overriding user specified config or without any specification in the config
+    port = await require("get-port")();
+    ganacheOptions = { ...ganacheOptions, port, ...configuredNetwork };
```
cds-amal, line 116: Perhaps leave a comment to indicate `configuredNetwork` is will only spread if it's defined and be ignored otherwise.

sukanyaparashar: Sure, I will do that @cds-amal . Thanks.

packages/core/lib/commands/test/run.js
```diff
@@ -68,43 +86,50 @@ module.exports = async function (options) {
...
-      config,
-      files,
-      temporaryDirectory
-    });
-    return numberOfFailures;
-  } else {
-    const ipcOptions = { network: "test" };
-    const port = await require("get-port")();
+  const configuredNetwork = config.networks[config.network];
+  const testNetworkDefined = configuredNetwork && config.network === "test";
```
haltman-at, line 90: This variable could still use a renaming like I suggested, I think.

packages/core/lib/commands/test/run.js
```diff
@@ -68,43 +86,50 @@ module.exports = async function (options) {
...
-        instamine: "strict"
-      }
-    };
-    const { disconnect } = await Develop.connectOrStart(
+  if (
+    (testNetworkDefined && noProviderHostOrUrlConfigured) ||
+    !configuredNetwork
+  ) {
+    // Use managed ganache with overriding user specified config or without any specification in the config
+    port = await require("get-port")();
```
haltman-at, line 115: You only use `port` in this block, so rather than declaring it with `let` up above, I think it would make more sense to declare it `const` right here.

haltman-at (commented):
Cool, logic seems good now.  I'm dismissing my request changes.  I like the reorganization, too. :)  Two comments still on style though, hope you don't mind!

cds-amal (commented):
This is shaping up nicely!

packages/core/lib/commands/test/run.js
sukanyaparashar: Can we use "testNetworkDefinedAndUsed" as the variable name, @haltman-at ? Also suggested by @cds-amal . As it is related to the "test" network, so I am also in support of that. Any other suggestions anyone @eggplantzzz , @gnidan , @fainashalts ?

haltman-at: Sure, that sounds fine.

sukanyaparashar: Okay, I will change the it and push another commit. Thanks @haltman-at .

haltman-at (approved):
Seems good to me!

sukanyaparashar:
> Seems good to me! Now we just have to see if @cds-amal still wants changes. :)

Sure, I will wait for @cds-amal to review it before merging.

cds-amal (approved):
LGTM! Nice work on this @sukanyaparashar .

Miladinho:
Have the docs been updated to reflect this change? I tries to follow the PR description and have had no luck

PR #4781, opened 2022-02-25
Update Solidity version in Truffle Init to 0.8.12

haltman-at:
Just realize we didn't update this for 0.8.12.  Since we still have no automation for this, here's a manual PR.  Again, I used `--no-verify` to keep prettier from altering the comment indentation on the db stuff.


cds-amal (approved):
Looks good! Thanks @haltman-at

Issue #3873, opened 2021-02-24
Support of solidity new overflow and divide by 0 error messages


sirhill:
## Issue

Solidity are now preventing overflow and division by 0 errors which were previously managed by `SafeMath.sol` lib.
As stated by @chriseth, solidity should now returns nice error messages.

It's interesting to note that ganache support partially this as it may returns out-of-bounds in the case of positive overflows (both for `add` and `mul`).
 

## Steps to Reproduce

```
contract SafeMath {

  function mul(uint256 a, uint256 b) public pure returns (uint256 c) {
    return a * b;
  }

  function div(uint256 a, uint256 b) public pure returns (uint256) {
    return a / b;
  }

  function sub(uint256 a, uint256 b) public pure returns (uint256) {
    return a - b;
  }

  function add(uint256 a, uint256 b) public pure returns (uint256 c) {
    return a + b;
  }
}
```

## Expected Behavior

A nice error message saying the call or the transaction reverted due of an overflow error or a division by 0.

## Actual Results

In truffle (rpc@ropsten):

As a call:
```
truffle(goerli01)> safeMath.add(MAX_UINT, 1).then((x) => x.toString())
'35408467139433450592217433187231851964531694900788300625387963629091585785856'
```

As a transaction:
``` reason: undefined,
  hijackedStack: 'StatusError: Transaction: 0x......  exited with an error (status 0). \n' +
    '     Please check that the transaction:\n' +
    '     - satisfies all conditions set by Solidity `require` statements.\n' +
    '     - does not trigger a Solidity `revert` statement.\n' +
    '\n' +
    '    at Object.receipt (/home/node/.config/yarn/global/node_modules/truffle/build/webpack:/packages/contract/lib/handlers.js:128:1)\n' +
    '    at runMicrotasks (<anonymous>)\n' +
    '    at processTicksAndRejections (internal/process/task_queues.js:93:5)\n' +
    '    at Function.start (/home/node/.config/yarn/global/node_modules/truffle/build/webpack:/packages/contract/lib/override.js:49:1)'
}
```

In Ganache
```
  1) Contract: SafeMath
       should prevent add overflow:
     AssertionError: Expected 'invalid opcode', got Error: value out-of-bounds (argument="a", value=-1, code=INVALID_ARGUMENT, version=abi/5.0.0-beta.153) instead
      at module.exports (test/helpers/assertJump.js:12:5)
      at runMicrotasks (<anonymous>)
[...]

PR #4791, opened 2022-03-02
Enhancement: Support panic codes in Truffle Contract

haltman-at:
Addresses #3873.  This PR adds support for panic codes in Truffle Contract.  More specifically, `reason.js` will now return a reason not only for revert strings, but also for panic codes.  Of course there's still no support for custom errors, which wouldn't be reasonable to support in Truffle Contract right now, but I did add a case for these with the generic message `"Custom error (could not decode)"`.

Anyway, as for the actual decoding, it's pretty straightforward.  Of course I could have used Truffle Codec here, but I think probably we shouldn't introduce that dependency yet?  So instead it's just done manually in line with the current way.  While I was doing this I took the time to try to make some of the variables a bit clearer.

Note that getting the string for the panic code is done with `debug-utils`, but we already use `debug-utils` in contract, so, that's not a new dependency!

Note, importantly, that this PR kind of just overloads the existing revert reason interface!  It does *not* attempt to draw a hard distinction between panics and revert strings in what it returns; it just has the panics start with `"Panic: "`.  If someone has a revert string start with `"Panic: "`, it may look like a panic!  This is kind of unfortunate, but I'm not sure if there's a good alternative for now?  That said, if this is unacceptable, we can try to find a different way of doing this.  (@gnidan, any opinion on this?)

Also, I didn't include the numerical panic code in the message.  Not sure if that's OK.  Please let me know if that should be changed.  (Again, @gnidan, any opinion?)

I also added tests of this; in order to do this, I also updated the tests in `contract-tests` from Solidity 0.5.0 to 0.8.12.  This did increase gas costs for some tests, so I had to turn down the number of iterations in that one deliberately-expensive test to keep it still deliberate-expensive but not more-than-the-block-limit expensive.  Note that the new panic tests used the existing `triggerAssertError()` function (since those now panic), while I changed the invalid opcode tests to use a new `triggerInvalidOpcode()` function that uses assembly.  I also improved the style a little in the tests I touched.

### DIFF
--- a/packages/contract-tests/test/deploy.js
+++ b/packages/contract-tests/test/deploy.js
@@ -211,18 +211,26 @@ describe("Deployments", function () {
     // Constructor in this test consumes ~6388773 (ganache) vs blockLimit of 6721975.
     it("doesn't multiply past the blockLimit", async function () {
       this.timeout(50000);
-      let iterations = 6000; // # of times to set a uint in a loop, consuming gas.
+      let iterations = 5000; // # of times to set a uint in a loop, consuming gas.
 
       const estimate = await Example.new.estimateGas(iterations);
       const block = await web3.eth.getBlock("latest");
       const multiplier = Example.gasMultiplier;
 
-      assert(multiplier === 1.25, "Multiplier should be initialized to 1.25");
+      assert.equal(
+        multiplier,
+        1.25,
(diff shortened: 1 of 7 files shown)


gnidan (approved):
Sweet!

lsqproduction (approved):
looks good

PR #4797, opened 2022-03-03
Add .eslintrc to truffledb

tenthirtyone:
Ran into this with @cds-amal this week while working on unit tests. Husky will bark at you after making changes to existing tests without this change.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,7 +2,9 @@
   "root": true,
   "env": {
     "es2021": true,
-    "node": true
+    "node": true,
+    "mocha": true,
+    "jest": true
   },
   "globals": {
     "BUNDLE_VERSION": "readonly",
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,9 +2,7 @@
   "root": true,
   "env": {
     "es2021": true,
-    "node": true,
-    "mocha": true,
-    "jest": true
+    "node": true
   },
   "globals": {
     "BUNDLE_VERSION": "readonly",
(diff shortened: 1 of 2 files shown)


.eslintrc.json
```diff
@@ -2,7 +2,9 @@
   "root": true,
   "env": {
     "es2021": true,
-    "node": true
+    "node": true,
+    "mocha": true,
+    "jest": true
```
gnidan, line 7: you can have both of these???

haltman-at (commented):
I don't think this is the right way to do this, because this will affect everything.  Or, at the least, it's not the way we're already handling this currently!  This is what `.eslintc.package.json` is for; the way we currently handle this problem is to put, in packages that need it, a `.eslintrc.json` that imports the global `.eslintrc.package.json`.  You can see this in `box`, in `debugger`, etc.  So, whatever package you need this for, just add the same thing there!

...or, if you really think this is the right way to do it, then this PR should in addition delete `.eslintrc.package.json` and all the files that reference it. :P

tenthirtyone:
@haltman-at no strong feelings at all. Your ears must have been burning @cds-amal and I were on a call talking about doing the exact same thing with `tsconfig.json`. I did not know it was already setup for eslint.

I'll make the change and send it up.

.eslintrc.json
cds-amal: From the [docs](https://eslint.org/docs/user-guide/configuring/language-options#specifying-environments)

> These environments are not mutually exclusive, so you can define more than one at a time.

haltman-at (approved):
Thanks for getting this!

PR #4799, opened 2022-03-04
Internal improvement: Add url to debugger usage line

haltman-at:
Just noticed this was missing.  Ideally I'd like to do this as `[--network <network>|--url <provider_url>]` but the way that `network` gets tacked on at the end prevents that.  Oh well?  Let me know if there's a better way to do this.

### DIFF
--- a/packages/core/lib/commands/debug/meta.js
+++ b/packages/core/lib/commands/debug/meta.js
@@ -35,7 +35,7 @@ module.exports = {
   },
   help: {
     usage:
-      "truffle debug [<transaction_hash>] [--fetch-external|-x]" +
+      "truffle debug [<transaction_hash>] [--fetch-external|-x] [--url <provider_url>]" +
       OS.EOL +
       "                             [--compile-tests|--compile-all|--compile-none]",
     options: [
--- a/packages/core/lib/commands/debug/meta.js
+++ b/packages/core/lib/commands/debug/meta.js
@@ -35,7 +35,9 @@ module.exports = {
   },
   help: {
     usage:
-      "truffle debug [<transaction_hash>] [--fetch-external|-x] [--url <provider_url>]" +
+      "truffle debug [<transaction_hash>] [--fetch-external|-x]" +
+      OS.EOL +
+      "                             [--network <network>|--url <provider_url>]" +
       OS.EOL +
       "                             [--compile-tests|--compile-all|--compile-none]",
     options: [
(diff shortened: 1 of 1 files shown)


cds-amal (commented):
I think this change is confusing because it doesn't convey `--network` and `--url` are mutually exclusive options. We should present `[--network <network> | --url <provider_url>]` as you suggested...

Maybe we can exclude `network` from `allowedGlobalOptions` and manually add the network option for this command. However, not sure if modifying `allowedGlobalOptions` has implications on the command line parsing :/

haltman-at:
Yes, does anyone know about the effects of that?  @eggplantzzz, @fainashalts?

fainashalts:
@haltman-at I think you could exclude network from `allowedGlobalOptions` and add it manually (afaik the allowed options just get tacked on to whatever you list manually anyway), but that doesn't really solve the problem here, does it? It would appear as an option still. 

I don't believe we have a way to present mutually exclusive options, other than showing it in the usage as `|`.

haltman-at:
Well, we can do it with `|`, then. :)   As long as it's OK to remove it from `allowedGlobalOptions`, we can manually set the usage line and put that in there!  The normal options appearing in the usage line are controlled by `usage`, not `options`, so that's not a problem.  I'll go do that.

packages/core/lib/commands/debug/meta.js
```diff
@@ -70,6 +77,6 @@ module.exports = {
           "Forces the debugger to use artifacts even if it detects a problem.  Dangerous; may cause errors."
       }
     ],
-    allowedGlobalOptions: ["network", "config"]
+    allowedGlobalOptions: ["config"] //network is allowed but is handled separately above so usage can be combined w/url
```
cds-amal, line 80: It might be useful to document this is a workaround (or needed?) to handle a mutually exclusive inherited option?

cds-amal (approved):
Looks good to me, as @fainashalts signed off on this approach.

packages/core/lib/commands/debug/meta.js
haltman-at: This comment is intended to compactly express that! :P  Do you think it's worth writing more here?

cds-amal: I think it's worth the extra context to document the motivation because this change introduces a new technique to handle the case of displaying a mutually exclusive global and (local?) option.

eggplantzzz: Are you saying to explicitly state that this is done to document the fact that "url" and "network" are mutually exclusive?

cds-amal: yes @eggplantzzz

haltman-at: OK, I've updated the comment to be more explicit.  I'll merge this once it passes CI.

PR #4800, opened 2022-03-04
Bump DB Deps

tenthirtyone:
I hit up the warnings and I tried to get ahead of some `yarn upgrade-interactive` packages.

This should take care of most of the `db` and `db-kit` warnings except for the the `@gql2ts/from-schema` warning since it requires graphql < 15 and I bumped up apollo to v3 and it requires >=15.3. 

Build/tests still passed regardless of `gql2ts` squawking the warning.


tenthirtyone:
:facepalm: I forgot we rebase instead of merge. It's late and I just laid the changes on top of a new branch

Issue #3836, opened 2021-02-20
Rename debugger's "solidity" submodule to "sources" or "sourcemap" something


haltman-at:
Since it's no longer just Solidity and Yul but Vyper too.  Note that we could keep `solidity` as an alias for the renamed selector, to keep this from being a breaking change.

leeftk:
Addressed in latest release.

PR #4808, opened 2022-03-08
Rename debugger `solidity` submodule to `sourcemapping`

haltman-at:
Addresses #3836.  Finally getting this renaming out of the way!

Not much to say about this one, it's mostly a straight renaming.  The `solidity` selectors are left in to keep this from being a breaking change, though.  (Possibly we can remove those later?)  Also due to `eslint` I had to remove an unused variable at one point.

Also `prettier` made some changes.


gnidan (approved):
lgtm!

PR #4810, opened 2022-03-09
Display "new" before contract name in stacktraces for constructors

haltman-at:
This PR makes it so that when printing out Solidity stacktraces, calls to constructors appear as `new Contract` rather than just `Contract`.

To do this, I added an `isConstructor` field into the frames of the stacktrace object.  This is a boolean and describes whether that stackframe takes place in a constructor.  (For external calls it's taken from the context; for internal calls it copies the stackframe below.)  Then, when creating the test stacktrace, for stackframes of type `"external"` and with `isConstructor` set, and with no associated function name (not sure how that would happen), we print `new Contract` instead of `Contract`.  Pretty simple!  Note the function in the stacktrace selectors that performs the shifting of the locations also had to be updated to include this in its shifting.

I also updated the tests to check the `isConstructor` field... and fixed the description of one of the tests, oops!

(Also, `prettier` made some changes.  Ugh.)

### DIFF
--- a/packages/debug-utils/index.js
+++ b/packages/debug-utils/index.js
@@ -34,59 +34,59 @@ const verbosePanicTable = {
 };
 
 const commandReference = {
-  o: "step over",
-  i: "step into",
-  u: "step out",
-  n: "step next",
+  "o": "step over",
+  "i": "step into",
+  "u": "step out",
+  "n": "step next",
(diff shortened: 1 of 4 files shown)


cds-amal (approved):
Looks good @haltman-at.

PR #4815, opened 2022-03-09
Bug fix: Make hdwallet-provider's `initialize` method always resolve

eggplantzzz:
There is an interesting behavior that hdwallet-provider has when, for example, a bad url is passed. In the constructor, the package goes to fetch the chain id asynchronously. If the connection or url is bad, then the package will throw an error asynchronously which can cause peoples' apps problems. We want to allow them to handle the errors in their code. So now when this call for the chain id fails, we resolve the Promise (instead of previously rejecting) with a `success` field and an `error` field. This way we can wait until the user goes to interact with hdwallet-provider before throwing an error. This will allow them to handle the error without crashing their process.

### DIFF
--- a/packages/hdwallet-provider/src/index.ts
+++ b/packages/hdwallet-provider/src/index.ts
@@ -49,7 +49,7 @@ class HDWalletProvider {
   private chainId?: ChainId;
   private chainSettings: ChainSettings;
   private hardfork: Hardfork;
-  private initialized: Promise<void>;
+  private initialized: Promise<{ error: any; success: boolean }>;
 
   public engine: ProviderEngine;
 
@@ -95,7 +95,7 @@ class HDWalletProvider {
         [
           `No provider or an invalid provider was specified: '${providerToUse}'`,
           "Please specify a valid provider or URL, using the http, https, " +
-          "ws, or wss protocol.",
+            "ws, or wss protocol.",
           ""
         ].join("\n")
       );
@@ -128,7 +128,10 @@ class HDWalletProvider {
       (chainSettings && typeof chainSettings.chainId !== "undefined")
     ) {
       this.chainId = chainId || chainSettings.chainId;
-      this.initialized = Promise.resolve();
+      this.initialized = Promise.resolve({
+        error: null,
+        success: true
+      });
     } else {
       this.initialized = this.initialize();
     }
@@ -154,7 +157,9 @@ class HDWalletProvider {
           }
         },
         async signTransaction(txParams: any, cb: any) {
-          await self.initialized;
+          const { success, error } = await self.initialized;
+          if (!success) throw error;
+
           // we need to rename the 'gas' field
           txParams.gasLimit = txParams.gas;
           delete txParams.gas;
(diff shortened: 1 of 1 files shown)


packages/hdwallet-provider/src/index.ts
```diff
@@ -271,21 +276,36 @@ class HDWalletProvider {
...
+            });
             return;
           } else if (response.error) {
-            reject(response.error);
+            resolve({
+              success: false,
+              error: response.error
+            });
             return;
           }
```
cds-amal, line 293: This is a minor style nit that you are free to ignore... the code could be written as
```javascript
if (error) {
  return resolve({});
}

if (response.error) {
  return resolve({})
}

// process response

File: packages/hdwallet-provider/src/index.ts [hunk 2]
```diff
@@ -271,21 +276,36 @@ class HDWalletProvider {
...
           if (isNaN(parseInt(response.result, 16))) {
             const message =
               "When requesting the chain id from the node, it" +
               `returned the malformed result ${response.result}.`;
-            throw new Error(message);
+            const error = new Error(message);
+            resolve({
+              success: false,
+              error
+            });
```
cds-amal, line 302: Should there be a return in this branch? it would seem if `isNaN` is true, then resolve will be invoked twice, the second time setting `this.chainId` to NaN

cds-amal (commented):
This is a really neat approach @eggplantzzz. I left a small style nit and a logic question. How should we test this?

packages/hdwallet-provider/src/index.ts
cds-amal: IIRC, the second resolve will have an actual impact, but it's still weird to rely on that behavior.

packages/hdwallet-provider/src/index.ts
eggplantzzz: Yeah that looks a lot nicer. I was avoiding this because I actually didn't want to return anything explicitly. But this just seems a lot cleaner.

packages/hdwallet-provider/src/index.ts
eggplantzzz: Yep I'll add the `return`s.

packages/hdwallet-provider/src/index.ts
```diff
@@ -279,30 +279,28 @@ class HDWalletProvider {
           // always resolve the Promise - if an error occurs, throw it when
           // this.initialized is checked so errors can be handled by the user
           if (error) {
-            resolve({
+            return resolve({
               success: false,
               error
             });
-            return;
           } else if (response.error) {
```
cds-amal, line 286: You can also get rid of the else here ... and use `if (response.error) ...`, your call, though.

cds-amal (commented):
Is it possible to have tests for these?

packages/hdwallet-provider/src/index.ts
eggplantzzz: I actually don't really like that style-wise but thanks for the suggestion!

PR #4816, opened 2022-03-09
Bug fix: Improve hdwallet-provider argument validation

eggplantzzz:
Currently hdwallet-provider's checks for input arguments is a bit primitive. It just checks to see if you passed one argument as the check for "new-style input arguments". If you were trying to use old-style arguments (positional) and provided just one argument (let's say a string for the provider url), it will then fail with an unhelpful "cannot use in with string"-type error. This PR adds one more check to make sure that that argument is an object.

This PR also adds some information to the message for the error that is thrown when the input arguments do not validate successfully, telling the user to make sure they provide both provider info as well as private keys/mnemonic which are required.

### DIFF
--- a/packages/hdwallet-provider/src/constructor/getOptions.ts
+++ b/packages/hdwallet-provider/src/constructor/getOptions.ts
@@ -135,9 +135,9 @@ export const getOptions = (
     return fromArguments(args);
   } else {
     const message =
-      "Unknown arguments format passed to new HDWalletProvider." +
-      " Please ensure you passed both provider information and a menmonic or " +
-      "private keys.";
+      "Unknown arguments format passed to new HDWalletProvider. " +
+      "Please ensure you passed provider information along with either a " +
+      "mnemonic or private keys.";
     throw new Error(message);
   }
 };
--- a/packages/hdwallet-provider/src/constructor/getOptions.ts
+++ b/packages/hdwallet-provider/src/constructor/getOptions.ts
@@ -9,21 +9,18 @@ import { wordlist } from "ethereum-cryptography/bip39/wordlists/english";
 const isMnemonicPhrase = (
   credentials: LegacyConstructor.Credentials
 ): credentials is MnemonicPhrase =>
-  typeof credentials === "string" &&
-  validateMnemonic(credentials, wordlist);
+  typeof credentials === "string" && validateMnemonic(credentials, wordlist);
 
 // check that the first argument is a list of private keys
 const isPrivateKeys = (
   credentials: LegacyConstructor.Credentials
-): credentials is PrivateKey[] =>
-  credentials instanceof Array;
+): credentials is PrivateKey[] => credentials instanceof Array;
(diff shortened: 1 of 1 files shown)


packages/hdwallet-provider/src/constructor/getOptions.ts
```diff
@@ -102,12 +101,9 @@ const fromArguments = (
...
   args: ConstructorArguments
 ): args is [Constructor.InputOptions] => {
-  // new-style means exactly one argument
-  if (args.length !== 1) {
-    return false;
-  }
-
   const [options] = args;
+  // new-style means exactly one argument and an object
+  if (args.length !== 1 || typeof options !== "object") return false;
```
haltman-at, line 106: This (and the message at the end) is the only real change, right?  The rest is all `prettier`?

packages/hdwallet-provider/src/constructor/getOptions.ts
```diff
@@ -140,8 +134,10 @@ export const getOptions = (
   } else if (matchesLegacyArguments(args)) {
     return fromArguments(args);
   } else {
-    throw new Error(
-      "Unknown arguments format passed to new HDWalletProvider. Please check your configuration and try again"
-    );
+    const message =
+      "Unknown arguments format passed to new HDWalletProvider." +
+      " Please ensure you passed both provider information and a menmonic or " +
```
haltman-at, line 139: This message is a little ambiguous.  Potential rewording of the second sentence to be less ambiguous (asusming I've got the right reading!): "Please ensure you passed provider information along with either a mnemonic or private keys."  OTOH if you want to leave it alone it's probably fine.

haltman-at (approved):
Seems basically good.  One minor suggestion about the wording of the message but if you'd prefer to just ignore that and merge this that seems fine.

Btw, I'd say this is more of a bug fix rather than an internal improvement, right?

packages/hdwallet-provider/src/constructor/getOptions.ts
eggplantzzz: yes correct

packages/hdwallet-provider/src/constructor/getOptions.ts
eggplantzzz: Yeah I'll reword this

packages/hdwallet-provider/src/constructor/getOptions.ts
eggplantzzz: I mean, it's not really a bug - it is just an improvement of the validation and clearer messaging. If you feel strongly that this is a fix then we can go ahead and change it.

haltman-at: Well this isn't an internal improvement; it has actual external effects!  So it's either an enhancement or a bug fix; I'd call it a bug fix, as it's causing something that ought to have worked to actually work.

PR #4817, opened 2022-03-09
Add sanitize function for network_id in truffle test

sukanyaparashar:
### ISSUE
`truffle test` command was breaking when invalid or "*" is passed as network_id from `truffle-config.js`.

### SOLUTION
Added a sanitize function for the ganache options passed from the `truffle-config.js`.

### DIFF
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -122,15 +122,18 @@ module.exports = async function (options) {
     }
   }
 
+  if (configuredNetwork) {
+    configuredNetwork.network_id = sanitizeNetworkID(
+      configuredNetwork.network_id
+    );
+  }
+
   if (
     (testNetworkDefinedAndUsed && noProviderHostOrUrlConfigured) ||
     !configuredNetwork
   ) {
     // Use managed ganache with overriding user specified config or without any specification in the config
     const port = await require("get-port")();
-    configuredNetwork.network_id = sanitizeNetworkID(
-      configuredNetwork.network_id
-    );
 
     // configuredNetwork will spread only when it is defined and ignored when undefined
     ganacheOptions = { ...ganacheOptions, port, ...configuredNetwork };
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -34,7 +34,7 @@ const parseCommandLineFlags = options => {
 
 // Sanitize ganache options specified in the config
 function sanitizeGanacheOptions(ganacheOptions) {
-  let network_id = ganacheOptions.network_id;
+  const network_id = ganacheOptions.network_id;
 
   // Use default network_id if "*" is defined in config
   if (network_id === "*") {
@@ -137,8 +137,6 @@ module.exports = async function (options) {
 
     // configuredNetwork will spread only when it is defined and ignored when undefined
     ganacheOptions = { ...ganacheOptions, port, ...configuredNetwork };
-
-    // Sanitize the ganache options if required
     const sanitizedGanacheOptions = sanitizeGanacheOptions(ganacheOptions);
 
     numberOfFailures = await startGanacheAndRunTests(
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -30,6 +30,30 @@ const parseCommandLineFlags = options => {
   }
 };
 
+// Sanitize ganache options specified in the config
+function sanitizeGanacheOptions(ganacheOptions) {
+  let network_id = ganacheOptions.network_id;
+  let ganacheOptionsChanged;
+
+  // Use default network_id if "*" is defined in config
+  if (network_id === "*") {
+    network_id = 4447;
+    ganacheOptionsChanged = { ...ganacheOptions, network_id };
+    return ganacheOptionsChanged;
+  }
+
+  const parsedNetworkId = parseInt(network_id, 10);
+  if (isNaN(parsedNetworkId)) {
+    const error =
+      `The network id specified in the truffle config ` +
+      `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
+    throw new Error(error);
+  }
+  network_id = parsedNetworkId;
+  ganacheOptionsChanged = { ...ganacheOptions, network_id };
+  return ganacheOptionsChanged;
+}
+
 module.exports = async function (options) {
   const Config = require("@truffle/config");
   const { Environment, Develop } = require("@truffle/environment");
@@ -107,27 +131,6 @@ module.exports = async function (options) {
     }
   };
 
-  function sanitizeGanacheOptions(ganacheOptions) {
-    let network_id = ganacheOptions.network_id;
-
-    // Use default network_id if "*" is defined in config
-    if (network_id === "*") {
-      network_id = 4447;
-      let ganacheOptionsChanged = { ...ganacheOptions, network_id };
-      return ganacheOptionsChanged;
-    }
-    const parsedNetworkId = parseInt(network_id, 10);
-    if (isNaN(parsedNetworkId)) {
-      const error =
-        `The network id specified in the truffle config ` +
-        `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
-      throw new Error(error);
-    }
-    network_id = parsedNetworkId;
-    let ganacheOptionsChanged = { ...ganacheOptions, network_id };
-    return ganacheOptionsChanged;
-  }
-
   if (
     (testNetworkDefinedAndUsed && noProviderHostOrUrlConfigured) ||
     !configuredNetwork
@@ -141,10 +144,9 @@ module.exports = async function (options) {
     // Sanitize the ganache options if required
     const sanitizedGanacheOptions = sanitizeGanacheOptions(ganacheOptions);
 
-    ganacheOptions = { ...ganacheOptions, ...sanitizedGanacheOptions };
     numberOfFailures = await startGanacheAndRunTests(
       ipcOptions,
-      ganacheOptions,
+      sanitizedGanacheOptions,
       config
     );
   } else {
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -33,13 +33,11 @@ const parseCommandLineFlags = options => {
 // Sanitize ganache options specified in the config
 function sanitizeGanacheOptions(ganacheOptions) {
   let network_id = ganacheOptions.network_id;
-  let ganacheOptionsChanged;
 
   // Use default network_id if "*" is defined in config
   if (network_id === "*") {
     network_id = 4447;
-    ganacheOptionsChanged = { ...ganacheOptions, network_id };
-    return ganacheOptionsChanged;
+    return { ...ganacheOptions, network_id };
   }
 
   const parsedNetworkId = parseInt(network_id, 10);
@@ -50,8 +48,7 @@ function sanitizeGanacheOptions(ganacheOptions) {
     throw new Error(error);
   }
   network_id = parsedNetworkId;
-  ganacheOptionsChanged = { ...ganacheOptions, network_id };
-  return ganacheOptionsChanged;
+  return { ...ganacheOptions, network_id };
 }
 
 module.exports = async function (options) {
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -1,3 +1,5 @@
+const defaultNetworkIdForTestCommand = 4447;
+
 const parseCommandLineFlags = options => {
   // parse out command line flags to merge in to the config
   const grep = options.grep || options.g;
@@ -36,7 +38,7 @@ function sanitizeGanacheOptions(ganacheOptions) {
 
   // Use default network_id if "*" is defined in config
   if (network_id === "*") {
-    network_id = 4447;
+    network_id = defaultNetworkIdForTestCommand;
     return { ...ganacheOptions, network_id };
   }
 
@@ -119,7 +121,7 @@ module.exports = async function (options) {
   let numberOfFailures;
   let ganacheOptions = {
     host: "127.0.0.1",
-    network_id: 4447,
+    network_id: defaultNetworkIdForTestCommand,
     mnemonic:
       "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat",
     time: config.genesis_time,
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -48,8 +48,7 @@ function sanitizeGanacheOptions(ganacheOptions) {
       `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
     throw new Error(error);
   }
-  network_id = parsedNetworkId;
-  return { ...ganacheOptions, network_id };
+  return { ...ganacheOptions, network_id: parsedNetworkId };
 }
 
 module.exports = async function (options) {
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -107,12 +107,30 @@ module.exports = async function (options) {
     }
   };
 
+  function sanitizeNetworkID(network_id) {
+    if (network_id !== "*") {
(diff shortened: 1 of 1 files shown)


packages/core/lib/commands/test/run.js
```diff
@@ -107,12 +107,30 @@ module.exports = async function (options) {
...
+          `The network id specified in the truffle config ` +
+          `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
+        throw new Error(error);
+      }
+      return network_id;
+    } else {
+      // We have a "*" network. Return the default.
+      return 4447;
+    }
+  }
```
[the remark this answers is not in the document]
cds-amal: A couple of notes on [`parseInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt). 

- It could return `0`, or `NaN` both of which are falsey. If you intended this behavior, it warrants a comment.  
- has a [peculiar behavior which allows `parseInt("23*blahasdf#$", 10) === 23`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#using_parseint).  The question is how strict do you want to be?

cds-amal (commented):
Thanks for this @sukanyaparashar. I left a comment on complexity/readability and wonder if  network_id needs validation. For example, what's the behavior for the following values of network_id:
- "*"
- "123*"
- "12dos"
- "*123"

packages/core/lib/commands/test/run.js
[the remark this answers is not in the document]
sukanyaparashar: > This logic could be simplified if checking for `*` first.
> 
> ```js
> // use default network_id if *
> if (network_id === "*") return 4447;
> 
> const parsedNetworkId = parseInt(network_id, 10);
> if (isNaN(parsedNetworkId)) { //throw appropriate error }
> return parsedNetworkId;
> ```

Thanks for the suggestion @cds-amal . I will simplify the function as you suggested. It is the same function that is written for truffle develop. So, once it is simplifed, I think we can simplify the sanitize function there too.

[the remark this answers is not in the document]
sukanyaparashar: > A couple of notes on [`parseInt`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt).
> 
> * It could return `0`, or `NaN` both of which are falsey. If you intended this behavior, it warrants a comment.
> * has a [peculiar behavior which allows `parseInt("23*blahasdf#$", 10) === 23`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#using_parseint).  The question is how strict do you want to be?

I am not that sure how far should we validate this to check for junk values. Any suggestions @eggplantzzz ?

packages/core/lib/commands/test/run.js
```diff
@@ -107,6 +107,27 @@ module.exports = async function (options) {
     }
   };
 
+  function sanitizeNetworkID(network_id) {
```
eggplantzzz, line 110: I would actually rather see this function be a bit more general, `sanitizeGanacheOptions`. In there we could handle `network_id`, I can foresee in the future a time that we might concern ourselves with sanitizing other options as well.

packages/core/lib/commands/test/run.js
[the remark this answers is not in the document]
eggplantzzz: I like @cds-amal's suggestion for re-writing this. I think we probably should do that and implement it with `truffle develop` as well. As for how far we should go with this, I'm not too fired up about going hard on this. I think what Amal has suggested should be good enough for now in my opinion.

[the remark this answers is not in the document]
sukanyaparashar: I agree with you @eggplantzzz . Lets modify this in this way.

packages/core/lib/commands/test/run.js
sukanyaparashar: Okay, I will do that.

packages/core/lib/commands/test/run.js
```diff
@@ -107,6 +107,22 @@ module.exports = async function (options) {
     }
   };
 
+  function sanitizeGanacheOptions(ganacheOptions) {
+    if (ganacheOptions.network_id === "*") {
+      ganacheOptions.network_id = 4447;
```
haltman-at, line 112: I would like to suggest that this function should probably be pure and not mutate the input; it would probably be better to clone with spreads rather than to directly mutate like this.

Granted, in this case it's harmless because it's only used in one place, but unnecessary mutation makes code harder to understand, so it's good to avoid it when reasonable.

packages/core/lib/commands/test/run.js
```diff
@@ -116,6 +132,11 @@ module.exports = async function (options) {
 
     // configuredNetwork will spread only when it is defined and ignored when undefined
     ganacheOptions = { ...ganacheOptions, port, ...configuredNetwork };
```
haltman-at, line 134: As an example of what I'm talking about -- and sorry for not catching this on your previous PR! -- the way you set up `ganacheOptions` earlier, and then mutate it here, is confusing; it would be much more readable if it was all set up in one place.  Like, I completely missed that the `configuredNetwork` options get dumped into the Ganache options, because it happens separately later after setup, and not just after but quite a bit after, where I wouldn't think to look for it.  Having variables change their value well after they're set up is generally not something you expect to happen!

That said, that's obviously a separate issue from this PR, so you don't need to change it here.

haltman-at (commented):
This isn't a complete review as I didn't look at all of this, but there's one particular thing I wanted to comment on.

packages/core/lib/commands/test/run.js
sukanyaparashar: Thanks @haltman-at for your suggestion. I will make the code better.

packages/core/lib/commands/test/run.js
sukanyaparashar: Sure, I will change this thing in another PR.

packages/core/lib/commands/test/run.js
```diff
@@ -107,6 +107,27 @@ module.exports = async function (options) {
     }
   };
 
+  function sanitizeGanacheOptions(ganacheOptions) {
+    let network_id = ganacheOptions.network_id;
+
+    // Use default network_id if "*" is defined in config
+    if (network_id === "*") {
+      network_id = 4447;
```
haltman-at, line 115: Looking at this a bit more -- where did this `4447` come from?  I would think that the thing to do would be, if `network_id` in the config is `"*"`, simply not *put* a network ID in the Ganache options, and let it use its own default, rather than us setting it for it.  Is there a reason for putting 4447 instead?

(Sorry, still not a proper review, but I thought that was worth commenting on!)

sukanyaparashar: Thanks @haltman-at for raising this. 4447 is the default network_id we use for ganache in truffle test. When the user specifies network_id in config, it overrides the default ganache options set in the `if` block of the code and then we are calling the `sanitizeGanacheOptions` function. If the network_id is `*` then it again returns the default the ganache network_id, that is 4447.

eggplantzzz: Should we just leave it out in the case "*" is specified? Seems like it might be a bit simpler.

cds-amal: It makes sense that `*` should not assign a network_id for ganacheOptions (for test) and allow Ganache to use its default, current timestamp.

sukanyaparashar: Since, we are calling the sanitize function after overriding the default ganache options, the network_id will return "*" again if we leave it out.

haltman-at: @sukanyaparashar You can *remove* the network ID if `"*"` was passed.  Your options aren't limited to setting a number or letting `"*"` pass through.  You can either overwrite it as `undefined`, or, if that's not sufficient, use the `delete` operator to remove it entirely.

cds-amal: What about setting `network_id = Date.now()`  instead of `4777` to align with Ganache's default when `network_id === '*'`?

sukanyaparashar: > @sukanyaparashar You can _remove_ the network ID if `"*"` was passed. Your options aren't limited to setting a number or letting `"*"` pass through. You can either overwrite it as `undefined`, or, if that's not sufficient, use the `delete` operator to remove it entirely.

https://github.com/trufflesuite/truffle/blob/2e85aa659d680a6ea7feb5bc7089556b3889557f/packages/environment/environment.js#L132

We have a validation check for network_id == null.

sukanyaparashar: > What about passing along `network_id = Date.now()` instead of `4777` to align with Ganache's default when the `networks.test.network_id === '*'`?

That can be done @cds-amal . But should we change it only when `network_id === "*"` or should we set the default ganache `network_id = Date.now()` when even no network is specified in the config?

cds-amal: Good question. If network_id = '*' we set it to Date.now(), if network_id is unspecified in the config, then Ganache will set its networkId to Date.now(). I'm ok leaving the special case of only setting it when `config.network_id = '*'` and let Ganache handle the other case. What do others think?

haltman-at: I don't think we should use `Date.now()`, because that -- like `4447` or any other attempt on our part to match Ganache's default -- is a fragile solution.  If the network_id is `"*"`, we should let Ganache do all the deciding, not us.  That's the robust solution.  If we try to match, it could change in the future.  (`Date.now()` is particularly worrying because that changes over time...)

@sukanyaparashar I don't see why the check you describe is relevant.  That runs as part of `Environment.detect(config)`.  I'm not saying you should remove the network ID from the config when network_id is `"*"`, but rather that it should be removed from the ganache options.  I don't believe there's any reason we can't do the latter?

sukanyaparashar: https://github.com/trufflesuite/truffle/blob/01c1d42d72564bf58b698742bf2ffe4ed0678901/packages/environment/environment.js#L75

I think there is a catch here @haltman-at . If I comment out the line in the `develop` function here, then ganache starts up even without the network_id.

haltman-at: I am not suggesting you comment out the line.  I am suggesting that, specifically when `network_id` from the config is equal to `"*"`, you actively remove `network_id` from the ganache options, i.e., by setting it to `undefined` or deleting it.  This requires checking for this condition and acting on it, not just commenting out a line.

haltman-at: Oh!  OK the catch has been explained in a meeting -- the problem isn't `Environment.detect`, it's a particular manual copy in `Environment.develop`.  That can be fixed, but I've been convinced that `4447` (or `Date.now()`) is fine.  I'll mark this resolved.

packages/core/lib/commands/test/run.js
```diff
@@ -107,6 +107,27 @@ module.exports = async function (options) {
     }
   };
 
+  function sanitizeGanacheOptions(ganacheOptions) {
```
eggplantzzz, line 110: One other thing to consider is pulling this out of the main body of this function. Style-wise I dislike having functions defined like this in the middle of another function as it disturbs the narrative and makes it harder to understand the code flow.

sukanyaparashar: Sure, @eggplantzzz . I will do that. Thanks.

packages/core/lib/commands/test/run.js
```diff
@@ -30,6 +30,30 @@ const parseCommandLineFlags = options => {
...
 };
 
+// Sanitize ganache options specified in the config
+function sanitizeGanacheOptions(ganacheOptions) {
+  let network_id = ganacheOptions.network_id;
+  let ganacheOptionsChanged;
+
+  // Use default network_id if "*" is defined in config
+  if (network_id === "*") {
+    network_id = 4447;
```
sukanyaparashar, line 40: network_id is returned as 4447 (default ganache network_id) when "*" is specified in the truffle-config.js "test" network. I skipped the Date.now() for now to keep the code in sync with `truffle develop`.

packages/core/lib/commands/test/run.js
```diff
@@ -30,6 +30,30 @@ const parseCommandLineFlags = options => {
   }
 };
 
+// Sanitize ganache options specified in the config
+function sanitizeGanacheOptions(ganacheOptions) {
+  let network_id = ganacheOptions.network_id;
+  let ganacheOptionsChanged;
```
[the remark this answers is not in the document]
sukanyaparashar: Yes, that can be done @eggplantzzz . Thanks for your suggestion.

packages/core/lib/commands/test/run.js
cds-amal: Can you replace the magic number with an appropriately commented constant?

sukanyaparashar: > Can you replace the magic number with an appropriately commented constant?

Or should I go with Date.now() as you suggested?

sukanyaparashar: > Can you replace the magic number with an appropriately commented constant?

Sure, @cds-amal . Thanks for the comment. :)

packages/core/lib/commands/test/run.js
```diff
@@ -116,9 +139,13 @@ module.exports = async function (options) {
 
     // configuredNetwork will spread only when it is defined and ignored when undefined
     ganacheOptions = { ...ganacheOptions, port, ...configuredNetwork };
+
+    // Sanitize the ganache options if required
```
cds-amal, line 143: Small nit. This comment is redundant, as your choice in function name and variable serves as adequate documentation.

packages/core/lib/commands/test/run.js
```diff
@@ -30,6 +32,27 @@ const parseCommandLineFlags = options => {
...
 };
 
+// Sanitize ganache options specified in the config
+function sanitizeGanacheOptions(ganacheOptions) {
+  let network_id = ganacheOptions.network_id;
+
+  // Use default network_id if "*" is defined in config
+  if (network_id === "*") {
+    network_id = defaultNetworkIdForTestCommand;
+    return { ...ganacheOptions, network_id };
```
cds-amal, line 42: This could be simplified. 

```suggestion
    return { ...ganacheOptions, network_id: defaultNetworkIdForTestCommand };
```

packages/core/lib/commands/test/run.js
```diff
@@ -30,6 +32,27 @@ const parseCommandLineFlags = options => {
...
+
+  const parsedNetworkId = parseInt(network_id, 10);
+  if (isNaN(parsedNetworkId)) {
+    const error =
+      `The network id specified in the truffle config ` +
+      `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
+    throw new Error(error);
+  }
+  network_id = parsedNetworkId;
+  return { ...ganacheOptions, network_id };
```
cds-amal, line 53: This can be simplified
```suggestion
  return { ...ganacheOptions, network_id: parsedNetworkId };
```

packages/core/lib/commands/test/run.js
```diff
@@ -116,9 +139,13 @@ module.exports = async function (options) {
 
     // configuredNetwork will spread only when it is defined and ignored when undefined
```
cds-amal, line 140: Good comment!

cds-amal (commented):
This reads a lot cleaner! I left some minor style comment.

packages/core/lib/commands/test/run.js
```diff
@@ -30,6 +32,25 @@ const parseCommandLineFlags = options => {
   }
 };
 
+// Sanitize ganache options specified in the config
```
cds-amal, line 35: Small nit. I think this comment is unnecessary because intent is already clearly communicated by the names you chose.

cds-amal (approved):
Looks good to me!

packages/core/lib/commands/test/run.js
sukanyaparashar: Thanks @cds-amal for the suggestion. I will remove this in my upcoming PR.

gnidan:
Nice work on this @sukanyaparashar!

PR #4827, opened 2022-03-09
Disallow bytes0 as valid Solidity type

wbt:
It does not appear valid based on https://docs.soliditylang.org/en/v0.8.12/types.html#fixed-size-byte-arrays


gnidan (approved):
Sweet! I'll wait for @haltman-at to come in and say why we need to keep this for silly legacy reasons, but you have a first approval from me at least :)

haltman-at (approved):
Thanks!  No, there are no silly legacy reasons why we need `bytes0`; `bytes0` has never been legal.

PR #4867, opened 2022-03-14
Internal improvement: Simplify Truffle's crypto.getRandomValues shim

eggplantzzz:
In https://github.com/trufflesuite/truffle/pull/4128 we shimmed the crypto library as webpack strips it out and ensjs requires the `getRandomValues` function. If it is not present it will print warnings to the console. This PR attempts to simplify the `crypto` shim as we don't need the entire `crypto` package.


cds-amal (approved):
Nice!

Issue #2928, opened 2020-04-03
rinkeby-fork not using specified address from HD Wallet Provider


pizzarob:
HD Wallet Provider in the truffle.js config file doesn't seem to be working properly when doing a dry-run using rinkeby-fork. The HD wallet provider has no effect on the accounts array during a dry run migration.

```js
  "rinkeby-fork": {
      provider() {
        // Has no affect on accounts array. Still populated with 10 random addresses when 
        // I expect only one.
        return new HDWalletProvider(
          privateKey,
          "https://rinkeby.infura.io/v3/API_KEY"
        );
      },
      network_id: 4,
    },
```

haltman-at:
Thanks for the report!  What Truffle version did this occur in?

pizza-r0b:
@haltman-at  5.1.2

haltman-at:
OK.  Do you know if it still occurs in the most recent version?

pizzarob:
Yeah looks like it does @haltman-at - updated to version Truffle v5.1.22

gnidan:
My hunch is that the code that sets up the dry-run probably doesn't respect the accounts in the provider. Hopefully we can get a quick look into this soon and spot the error quickly.

Thanks for raising this!

dongmingh:
@pizzarob can you confirm if the issue is fixed in the latest Truffle? We think this is fixed. Thanks in advance!

MicaiahReid:
We haven't heard anything back on this, so we're calling it fixed! Please open another issue if you're finding that this is still happening.

PR #4876, opened 2022-03-16
Bug fix: Don't autogenerate accounts when forking for a dry run

eggplantzzz:
addresses https://github.com/trufflesuite/truffle/issues/2928

When Truffle forks for a dry run, it initializes a Ganache instance. That Ganache instance is passed a provider, presumably with the user's unlocked accounts. On initialization, however, Ganache generates another set of 10 accounts that are unlocked which pushes any accounts the user has unlocked to the second half of the array of unlocked accounts. This behavior is probably not what we want, as the user should probably expect that his unlocked accounts will start at index 0.

To remedy this, this PR adds `wallet: { totalAccounts: 0 }` to the list of options passed to Ganache. This causes Ganache to refrain from unlocking additional accounts.

To test this, you can simply create a network configuration that has a provider with unlocked accounts (@truffle/hdwallet-provider). In your migrations, use the `accounts` variable (`module.exports = (deployer, network, accounts) => {...` to print out the list of accounts while using the currently published Truffle. You should receive an array with 20 accounts, the accounts you unlocked being at the end of the array. Use this branch to run the same migration and you should get an array of 10 accounts: only the ones unlocked in your provider in the network configuration.

### DIFF
--- a/packages/core/lib/commands/migrate/index.js
+++ b/packages/core/lib/commands/migrate/index.js
@@ -1,6 +1,8 @@
 module.exports = {
   run: require("./run"),
   meta: require("./meta"),
+  runMigrations: require("./runMigrations"),
+  setUpDryRunEnvironmentThenRunMigrations: require("./setUpDryRunEnvironmentThenRunMigrations"),
   determineDryRunSettings: require("./determineDryRunSettings"),
   prepareConfigForRealMigrations: require("./prepareConfigForRealMigrations")
 };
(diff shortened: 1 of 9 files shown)


gnidan (approved):
Whoa! This might be the fix for that elusive "dry run is totally broken" thing I've been on about for months!

eggplantzzz:
I think I'm going to write a couple of tests for this PR.

gnidan (approved):
sweet!

gnidan:
> Should we worry about this breaking people's projects that rely on this faulty behavior?

I'd be surprised if anyone is relying on the broken behavior... I'm fine waiting for people to complain, then we can sort it out then.

PR #4877, opened 2022-03-16
Dependency update: Update highlightjs-solidity to 2.0.5

haltman-at:
Updates our syntax highlighting for Solidity 0.8.13; see [here](https://github.com/highlightjs/highlightjs-solidity/pull/65) for the actual PR on the highlightjs-solidity repo that this is updating us to have.


cds-amal (approved):
Looks good!

Issue #4863, opened 2022-03-13
Truffle solc pragma causes the exception Cannot destructure property 'contractName'


shadmatic:
- [X] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue. Ticket No: 821498

---------------------------

## Issue

In our repo codebase we are using many contracts dependencies from different solidity versions. Thus, we are using the truffle config `pragma`. Problem is happening in our CI, we are experiencing the error: `TypeError: Cannot destructure property 'contractName' of 'contract' as it is undefined.`.

## Steps to Reproduce

Create and import a few contracts with different versions, for instance:
- ChainLink LinkToken uses ^0.4.11
- ChainLink VRFMock uses 0.6.6

Create a basic contract using `pragma solidity ^0.8.0;` and import above contracts which has mixed versions.

In your `truffle-config.js` file, just set your solc to pragma:
```js
//...
  compilers: {
    solc: {
      version: "pragma", // Fetch exact version from solc-bin (default: truffle's version)
// ...
```

## Expected Behavior

The contracts to be built successfully. I realized that in truffle `v5.4.26` it actually works

## Actual Results

```sh
TypeError: Cannot destructure property 'contractName' of 'contract' as it is undefined.
    at forContract (/usr/local/lib/node_modules/truffle/build/webpack:/packages/compile-common/dist/src/shims/NewToLegacy.js:5:12)
    at Array.map (<anonymous>)
    at Object.save (/usr/local/lib/node_modules/truffle/build/webpack:/packages/workflow-compile/index.js:[123](https://github.com/StarterXyz/effective-system/runs/5463281701?check_suite_focus=true#step:5:123):1)
    at Object.compileAndSave (/usr/local/lib/node_modules/truffle/build/webpack:/packages/workflow-compile/index.js:96:1)
    at Object.compileContractsWithTestFilesIfNeeded (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/testing/Test.js:239:38)
    at Object.run (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/testing/Test.js:97:27)
    at Object.module.exports [as run] (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/commands/test/run.js:102:1)
    at Command.run (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/command.js:189:1)
[...]

PR #4878, opened 2022-03-16
Bug fix: Shim output from compileWithPragmaAnalysis

eggplantzzz:
In [this commit](https://github.com/trufflesuite/truffle/commit/34e716be929f3a53e6e3b7bd83d245887e59f5a4) the compilation results were shimmed and converted to `WorkflowCompileResult`s. The output from `compileWithPragmaAnalysis` was accidentally left out. This PR shims that method as well.

### DIFF
--- a/packages/compile-solidity/src/index.js
+++ b/packages/compile-solidity/src/index.js
@@ -186,9 +186,7 @@ const Compile = {
       paths,
       options
     });
-    return Compilations.promoteCompileResult({
-      compilations: compilationResult.compilations
-    });
+    return Compilations.promoteCompileResult(compilationResult);
   }
 };
 
--- a/packages/compile-solidity/src/index.js
+++ b/packages/compile-solidity/src/index.js
@@ -182,7 +182,13 @@ const Compile = {
   },
 
   async sourcesWithPragmaAnalysis({ paths, options }) {
-    return compileWithPragmaAnalysis({ paths, options });
+    const compilationResult = await compileWithPragmaAnalysis({
+      paths,
+      options
+    });
+    return Compilations.promoteCompileResult({
+      compilations: compilationResult.compilations
+    });
   }
 };
 


packages/compile-solidity/src/index.js
```diff
@@ -182,7 +182,17 @@ const Compile = {
   },
 
   async sourcesWithPragmaAnalysis({ paths, options }) {
-    return compileWithPragmaAnalysis({ paths, options });
+    const compilationResult = await compileWithPragmaAnalysis({
+      paths,
+      options
+    });
+    return compilationResult.compilations.every(
```
haltman-at, line 189: Is there a reason for the switch here?  I don't see any reason not to just promote whatever you get.

packages/compile-solidity/src/index.js
```diff
@@ -182,7 +182,17 @@ const Compile = {
...
+      compilation => compilation.contracts.length === 0
+    )
+      ? Compilations.emptyWorkflowCompileResult()
+      : Compilations.promoteCompileResult({
+          compilations: compilationResult.compilations
```
haltman-at, line 194: Similarly, why the shallow clone here?  You can just promote directly.

haltman-at (commented):
Two comments about what seem like overcomplications...

packages/compile-solidity/src/index.js
eggplantzzz: Oh, I kind of assumed the empty one was a special case and this might be necessary. Happy to remove this!

haltman-at (approved):
Thanks for finding this!

haltman-at:
Btw, this PR fixes #4863, right?

eggplantzzz:
Yes it does @haltman-at, thanks for referencing that.

haltman-at:
(OK but @eggplantzzz you should edit that into the first comment, otherwise whoever's doing the release might miss it.)

PR #4879, opened 2022-03-16
Bug fix: Set exit code instead of calling process.exit

eggplantzzz:
According to Node, it is perhaps preferable to set the exit code and "exit gracefully" rather than calling `process.exit`. See [this article](https://nodejs.dev/learn/how-to-exit-from-a-nodejs-program).

Also, command `run` functions probably should not return anything unless they explicitly want to return an exit code.

**Update:** For whatever reason setting `process.exitCode` causes a lot of tests to fail. I subtracted that change from this PR and left it calling `process.exit`.

### DIFF
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -56,7 +56,9 @@ if (userWantsGeneralHelp) {
 
 command
   .run(inputArguments, options)
-  .then(returnStatus => process.exit(returnStatus))
+  .then(returnStatus => {
+    returnStatus ? (process.exitCode = returnStatus) : (process.exitCode = 0);
+  })
   .catch(error => {
     if (error instanceof TaskError) {
       analytics.send({
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -57,7 +57,9 @@ if (userWantsGeneralHelp) {
 command
   .run(inputArguments, options)
   .then(returnStatus => {
-    returnStatus ? (process.exitCode = returnStatus) : (process.exitCode = 0);
+    typeof returnStatus !== undefined
+      ? (process.exitCode = returnStatus)
+      : (process.exitCode = 0);
   })
   .catch(error => {
     if (error instanceof TaskError) {
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -57,9 +57,7 @@ if (userWantsGeneralHelp) {
 command
   .run(inputArguments, options)
   .then(returnStatus => {
-    returnStatus !== undefined
-      ? (process.exitCode = returnStatus)
-      : (process.exitCode = 0);
+    returnStatus !== undefined ? process.exit(returnStatus) : process.exit();
   })
   .catch(error => {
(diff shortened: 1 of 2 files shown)


packages/core/cli.js
```diff
@@ -56,7 +56,11 @@ if (userWantsGeneralHelp) {
 
 command
   .run(inputArguments, options)
-  .then(returnStatus => process.exit(returnStatus))
+  .then(returnStatus => {
+    typeof returnStatus !== undefined
```
haltman-at, line 60: I'm assuming you mean `typeof returnStatus !== "undefined"`?  If so, it's simpler to just write `returnStatus !== undefined`.  (But either way, `typeof returnStatus` will never equal `undefined`!)

Btw, is this code implicitly assuming that `returnStatus` will always either be either `undefined` or an integer?  If so you might want to add a comment to that effect.

eggplantzzz: Yeah sorry I messed this up. So maybe we should eventually change Truffle's architecture here. Currently each command's `run` function does not really have a standardized system for returning values to this context here. Right now it just completes, sometimes returns something, and throws if there is an error.

Would it make sense to devise some sort of a system where each `run` function never throws and instead returns some sort of object describing what happened. Perhaps it would have fields that would describe whether it was successful, a message if there was an error, miscellaneous data related to any given error, etc. What do you think of that?

haltman-at (commented):
Oh hey there's another case of `process.exit(1)` lower down in this file... assuming we're making this change, presumably that should also be changed?

Although, uh, given how CI keeps failing, I'm wondering if this change is actually valid...?

packages/core/cli.js
haltman-at: Huh, interesting idea, but I have no idea!  Seems like a thing for later regardless...

eggplantzzz:
> Although, uh, given how CI keeps failing, I'm wondering if this change is actually valid...?

Will changing the other `process.exit(1)` affect CI?

haltman-at:
CI is seemingly not runining again...

haltman-at:
CI seems to have gotten lost on the way to this one.

packages/core/cli.js
```diff
@@ -56,7 +56,9 @@ if (userWantsGeneralHelp) {
 
 command
   .run(inputArguments, options)
-  .then(returnStatus => process.exit(returnStatus))
+  .then(returnStatus => {
+    returnStatus !== undefined ? process.exit(returnStatus) : process.exit();
```
benjamincburns, line 60: I might be missing something, but isn't this change equivalent to the previous version? That is, calling `process.exit(undefined)` should be identical to calling `process.exit()`, right?

haltman-at: Those are equivalent, yeah.

eggplantzzz: Ok so I can probably just remove that change and just remove the return value from `compile`. I'll close this and open a new PR.

PR #4880, opened 2022-03-16
Dependency update: Update node-ipc to 9.2.1 and pin it

haltman-at:
This PR does two things.  First, it updates node-ipc to 9.2.1, because may as well update.  Second and more importantly, it *pins* it there, because node-ipc 9.2.2 has [this](https://github.com/RIAEvangelist/node-ipc/issues/233) ghastly thing in it, and we have to make sure users of `@truffle/environment` don't encounter that...


fainashalts (approved):
LGTM! Pending tests running, of course. :)

cds-amal (approved):
Yay! You saved the patient, doc!  If the tests pass, release her!

PR #4875, opened 2022-03-16
show a message in the terminal when the browser is waiting for signature

benjamincburns:
See #4824 and #4850

PR #4885, opened 2022-03-17
events: better encapsulation of --quiet flag logic

benjamincburns:
This was prompted by feedback in #4875 regarding `if` statements w/o curly braces.

However, the bulk of the `if` statements in question appear to have been added due to the concern of silencing console output not being encapsulated well in the default event handlers. As a result, I did a bit of refactoring and pushed that concern down into the `Subscriber` type.

This does introduce a bit of a cohesion issue now, as ideally the `Subscriber` type wouldn't need to know about these things, however I think the ultimate solution for this would be to introduce a proper presentation layer for Truffle.

### DIFF
--- a/packages/events/Subscriber/index.js
+++ b/packages/events/Subscriber/index.js
@@ -1,5 +1,7 @@
+const ora = require("ora");
 const helpers = require("./helpers");
 const { createLookupTable, sortHandlers, validateOptions } = helpers;
+const Writable = require("stream").Writable;
 
 class Subscriber {
   constructor({ emitter, options, config }) {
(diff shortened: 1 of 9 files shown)


benjamincburns:
There are a bunch of other occurrences of this pattern in our code, however. I'd recommend implementing a uniform set of eslint (or similar) rules across all projects to clean this up.

gnidan:
We'll want to pull @eggplantzzz into this :)

benjamincburns:
> There are a bunch of other occurrences of this pattern in our code, however. I'd recommend implementing a uniform set of eslint (or similar) rules across all projects to clean this up.

Incidentally, this is another case where we'd be benefitted by a move over to NX - it makes it somewhat easy to create workspace-level configs for tools like eslint, but thankfully doesn't force conformity across all packages in the project.

packages/events/Subscriber/index.js
```diff
@@ -11,12 +13,35 @@ class Subscriber {
     this.unsubscribeListeners = {};
     this.quiet = config.quiet;
     this.config = config;
-    if (config.logger) this.logger = config.logger;
-    if (initialization) initialization.bind(this)(config);
+
+    Object.defineProperty(this, "logger", {
+      get: () => {
+        if (!this._logger) {
+          this.logger = this.config.logger || console;
```
benjamincburns, line 20: note: this line is intentionally `this.logger =` and not `this._logger =`, as I wanted to invoke the wrapper logic in the setter.

benjamincburns:
There are a bunch of problems with this PR and it's not currently blocking anything, so I'm going to revisit it at a later time.

PR #4889, opened 2022-03-17
Internal improvement: Use typed imports for web3-utils in codec

haltman-at:
The `codec` package has a bunch of places where it imports `web3-utils` untyped.  I think this was necessary at one point?  Well, it definitely isn't any longer.  This PR cleans that up and replaces them with typed imports.  It also eliminates a few unused imports in the process.

Also, `prettier` made some changes too, which actually ends up being the bulk of the PR.  Sorry!

...and then also, `prettier` introduced a syntax error at one point, and I had to change a coercion from `<>` to `as` to get it to something `prettier` would handle properly...


cds-amal (approved):
I didn't know about untyped and typed imports and had to read up on it. Thank you Harry. LGTM

PR #4890, opened 2022-03-17
Add dependency-graph script

benjamincburns:
Generates dependency-graph.pdf in the root of the repo that shows
dependencies between the packages in our workspace. Solid lines reflect
runtime dependencies, dashed lines reflect dev dependencies, and dotted
lines reflect peer dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "test": "lerna run test --stream --concurrency=1 -- --colors",
     "ci": "./scripts/ci.sh",
     "geth": "./scripts/geth.sh",
+    "dependency-graph": "lerna-dependency-graph -f pdf -o dependency-graph.pdf",
     "solc-bump": "node ./scripts/solc-bump.js",
     "update": "lernaupdate"
   },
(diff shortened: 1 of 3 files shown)


benjamincburns:
Example of the image it generates:

dependency-graph

package.json
```diff
@@ -9,6 +9,7 @@
     "test": "lerna run test --stream --concurrency=1 -- --colors",
     "ci": "./scripts/ci.sh",
     "geth": "./scripts/geth.sh",
+    "dependency-graph": "lerna-dependency-graph -f png -o dependency-graph.png",
```
gnidan, line 12: Would PDF be better? PNG looks so raster

benjamincburns: I'll have a look, but I thought PDF would just encapsulate the raster. If that's the case, I'll also check for a SVG output, as that might be a bit nicer.

benjamincburns: Ah, I see what you mean now - PDF is indeed better, as the text is text, so cmd-f works, etc... wilco

gnidan (approved):
nice to have this easily accessible. thanks!

PR #4891, opened 2022-03-18
Refactor code in truffle test command run.js

sukanyaparashar:
### ISSUE:
The code written in `truffle/packages/core/lib/commands/test/run.js` needed some refactoring.

### SOLUTION:
The code is refactored for better readability/understanding.

### DIFF
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -7,6 +7,9 @@ const parseCommandLineFlags = options => {
   const reporter = options.reporter || options.r;
 
   /**
+   * If the reporter is returned as undefined, it ignores the specification of any reporter type in the
+   * config and displays the default mocha reporter "spec", as opposed to reporter completely being absent
+   * which results in checking for the reporter type specified in the config.
    * Reporter value passed on the command line has precedence over a reporter value in the config.
    * If neither exist, then nothing is passed to mocha and it uses its default reporter type "spec".
    */
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -7,11 +7,10 @@ const parseCommandLineFlags = options => {
   const reporter = options.reporter || options.r;
 
   /**
-   * If the reporter is returned as undefined, it ignores the specification of any reporter type in the
-   * config and displays the default mocha reporter "spec", as opposed to reporter completely being absent
-   * which results in checking for the reporter type specified in the config.
    * Reporter value passed on the command line has precedence over a reporter value in the config.
    * If neither exist, then nothing is passed to mocha and it uses its default reporter type "spec".
+   * Note: It is important that reporter be completely omitted, and not be set to undefined!
+   * As setting it to undefined will ignore the reporter value specified in the config.
    */
   return reporter === undefined
     ? { mocha: { grep, bail } }
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -7,10 +7,8 @@ const parseCommandLineFlags = options => {
   const reporter = options.reporter || options.r;
 
   /**
-   * If the reporter type is undefined in the command line, it results in checking and returning the
-   * reporter type in config if exists, otherwise returns the default mocha reporter type "spec".
-   * If the reporter type is defined in the command line, it results in overriding the reporter type in
-   * config if exists.
+   * Reporter value passed on the command line has precedence over a reporter value in the config.
+   * If neither exist, then nothing is passed to mocha and it uses its default reporter type "spec".
    */
   return reporter === undefined
     ? { mocha: { grep, bail } }
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -7,32 +7,16 @@ const parseCommandLineFlags = options => {
   const reporter = options.reporter || options.r;
 
   /**
-   * This if-else condition is explicitly written to avoid the overlapping of
-   * the config by the default mocha reporter type when user specifies a mocha reporter type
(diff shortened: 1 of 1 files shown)


packages/core/lib/commands/test/run.js
```diff
@@ -7,32 +7,16 @@ const parseCommandLineFlags = options => {
...
 
   /**
-   * This if-else condition is explicitly written to avoid the overlapping of
-   * the config by the default mocha reporter type when user specifies a mocha reporter type
-   * in the config and doesn't specify it as the command line argument.
-   * If the reporter is returned as undefined, it ignores the specification of any reporter type in the
-   * config and displays the default mocha reporter "spec", as opposed to reporter completely being absent
-   * which results in checking for the reporter type specified in the config.
+   * If the reporter type is undefined in the command line, it results in checking and returning the
+   * reporter type in config if exists, otherwise returns the default mocha reporter type "spec".
```
eggplantzzz, line 11: This comment is not correct. Mocha defaults to "spec", we don't specify it anywhere. What is the point of this comment exactly? Does it mean to say something like, "Truffle gives precedence to a reporter value passed on the command line, then to a value specified in the config, and if neither exist then we don't pass it to mocha and it uses its default one"?

packages/core/lib/commands/test/run.js
```diff
@@ -7,32 +7,16 @@ const parseCommandLineFlags = options => {
...
-  } else {
-    return {
-      mocha: {
-        grep,
-        bail,
-        reporter
-      }
-    };
-  }
+  return reporter === undefined
```
eggplantzzz, line 15: This looks a lot less chonky.

packages/core/lib/commands/test/run.js
sukanyaparashar: > This comment is not correct. Mocha defaults to "spec", we don't specify it anywhere. What is the point of this comment exactly? Does it mean to say something like, "Truffle gives precedence to a reporter value passed on the command line, then to a value specified in the config, and if neither exist then we don't pass it to mocha and it uses its default one"?

This is exactly what it supposed to mean. I agree that this comment needs to be changed.

packages/core/lib/commands/test/run.js
```diff
@@ -7,32 +7,14 @@ const parseCommandLineFlags = options => {
   const reporter = options.reporter || options.r;
 
   /**
-   * This if-else condition is explicitly written to avoid the overlapping of
-   * the config by the default mocha reporter type when user specifies a mocha reporter type
-   * in the config and doesn't specify it as the command line argument.
-   * If the reporter is returned as undefined, it ignores the specification of any reporter type in the
-   * config and displays the default mocha reporter "spec", as opposed to reporter completely being absent
-   * which results in checking for the reporter type specified in the config.
+   * Reporter value passed on the command line has precedence over a reporter value in the config.
```
haltman-at, line 10: I agree that the earlier coment was overlong, but the way that you simplified it erased the key reason that the comment was necessary -- namely, that in this case the distinction between `reporter` being undefined, and `reporter` being absent, **is relevant**, and that the code must not be simplified to remove this distinction.  IMO, the comment should still indicate that (it just doesn't need to take so many words to do it).

packages/core/lib/commands/test/run.js
```diff
@@ -115,30 +97,34 @@ module.exports = async function (options) {
...
     // Use managed ganache with overriding user specified config or without any specification in the config
-    const port = await require("get-port")();
+    const defaultPort = await require("get-port")();
+    const defaultHost = "127.0.0.1";
+    const defaultMnemonic =
+      "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
 
     // configuredNetwork will spread only when it is defined and ignored when undefined
-    ganacheOptions = { ...ganacheOptions, port, ...configuredNetwork };
+    const ganacheOptions = {
```
haltman-at, line 113: This is quite a bit clearer, thanks!

haltman-at (commented):
Thanks for this!  One quick note on the comment.

packages/core/lib/commands/test/run.js
sukanyaparashar: Thanks for pointing that out @haltman-at . I think it is worth mentioning. I will add that before merging.

sukanyaparashar: @haltman-at I, have added the previous comment which was left out. I thought of making the comment a bit shorter but I think these many lines in the comment are important to completely understand the logic.

haltman-at: Hm, I don't think the addition does the job.  What's non-obvious here isn't the **logic**.  The logic is clear!  What's non-obvious here is the **reason** for the logic.  Simply put: If a refactorer were to see this code, they might incorrectly attempt to simplify it by replacing it with `return { mocha: { grep, bail, reporter }}`, which wouldn't work!  They would do this because they would expect `undefined`, and complete absence, to act exactly the same, because they do so in many contexts.  But they don't here.  The key thing to get at is that this distinction, which is normally irrelevant, is in this case relevant.  I'd suggest a comment like "Note: It is important that `reporter` be completely omitted, and not be set to `undefined`!"  That's the sort of thing that would express that the distinction is important and that you not remove it.  (It might also be nice to say why -- i.e., that it being present as `undefined` will take precedence as over the default -- but the most important thing is to express the distinction and it's importance.)

sukanyaparashar: Okay, I got your point now. I will add a note like you suggested. Thanks for the suggestion @haltman-at . Sometimes it is difficult to convey the words in a proper way :)

haltman-at (approved):
I'll give this a redundant second approve. :)

PR #4893, opened 2022-03-18
Remove clobbered `paths` key from tsconfig

cds-amal:
The encoder and decoder packages had a duplicate `paths` key in their `tsconrfig.json`. This PR removes the first occurrence, leaving the one that JavaScript would use in this case.


haltman-at (approved):
Thanks for catching this!

Issue #4824, opened 2022-03-09
When using Truffle Dashboard, Truffle should show a message in the terminal when the browser is waiting for signature and delete the message after


gnidan:
## Issue

Right now, when you use Truffle with Truffle Dashboard, you have to keep your eye on the Dashboard browser tab, since all signature requests go there.

This is a bit jarring, and some users (including @Pclaesen) have raised the concern that Truffle Dashboard should maybe steal focus or such, so they don't stare blindly at `truffle migrate` output, erroneously waiting for activity when they actually need to do something themselves. After some discussion, it seems the better course of action would be for the terminal to alert the user to the effect of "Hey, alt-tab your way over to Truffle Dashboard, cause it's waiting for you!".

This is a bit tricky because it requires:
- Truffle must know it's connected to Truffle Dashboard
- Truffle must know when the Dashboard is waiting for response and display a message at that point
- Ideally, Truffle would remove the message from the terminal when the user has processed the request, so Truffle must also know when that happens to do this console cleanup

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.5.3
* node version (`node --version`):
* npm version (`npm --version`):

MicaiahReid:
Some notes:
 - We probably need to add a new default subscriber to truffle that is used by truffle dashboard.

PR #4901, opened 2022-03-21
Fix dashboard subscriber

benjamincburns:
Fixes that should've been included in #4875.

### DIFF
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -286,6 +286,7 @@ export const configProps = ({
 
         const options = configObject.network_config;
         options.verboseRpc = configObject.verboseRpc;
+        options.events = configObject.events;
 
         return Provider.create(options);
       },
--- a/packages/events/defaultSubscribers/dashboard.js
+++ b/packages/events/defaultSubscribers/dashboard.js
@@ -1,16 +1,9 @@
 const Spinner = require("@truffle/spinners").Spinner;
 
 module.exports = {
-  initialization: function (config) {
-    this._logger = {
-      log: ((...args) => {
-        if (config.quiet) {
-          return;
-        }
-
-        (this.logger || config.logger || console).log(...args);
-      }).bind(this)
-    };
+  initialization: function () {
+    this.logger = this.logger || console;
+    this.pendingTransactions = [];
   },
   handlers: {
     "rpc:request": [
@@ -28,14 +21,18 @@ module.exports = {
     ],
     "rpc:result": [
       function (event) {
-        const { payload, error, result } = event;
+        let { error } = event;
+        const { payload, result } = event;
 
         if (payload.method === "eth_sendTransaction") {
+          error = error || result.error;
           if (error) {
             const errMessage = `Transaction submission failed with error ${error.code}: '${error.message}'`;
             this.spinner.fail(errMessage);
           } else {
-            this.spinner.succeed(`Transaction submitted successfully. Hash: ${result.result}`);
+            this.spinner.succeed(
+              `Transaction submitted successfully. Hash: ${result.result}`
+            );
           }
 
           delete this.pendingTransactions[payload.id];
--- a/packages/truffle/test/scenarios/migrations/migrate.js
+++ b/packages/truffle/test/scenarios/migrations/migrate.js
@@ -7,7 +7,7 @@ const Reporter = require("../reporter");
 const sandbox = require("../sandbox");
 const Web3 = require("web3");
 
-describe("migrate (success)", function() {
+describe("migrate (success)", function () {
   let config;
   let web3;
   let networkId;
@@ -17,7 +17,7 @@ describe("migrate (success)", function() {
   before(done => Server.start(done));
   after(done => Server.stop(done));
 
-  before(async function() {
+  before(async function () {
     this.timeout(10000);
     config = await sandbox.create(project);
     config.network = "development";
@@ -33,7 +33,7 @@ describe("migrate (success)", function() {
     networkId = await web3.eth.net.getId();
   });
 
-  it("runs migrations (sync & async/await)", async function() {
+  it("runs migrations (sync & async/await)", async function () {
     this.timeout(70000);
 
     await CommandRunner.run("migrate", config);
@@ -62,7 +62,7 @@ describe("migrate (success)", function() {
     console.log(output);
   });
 
-  it("forces a migration with the -f option", async function() {
+  it("forces a migration with the -f option", async function () {
     this.timeout(70000);
 
     await CommandRunner.run("migrate -f 3", config);
--- a/packages/provider/wrapper.js
+++ b/packages/provider/wrapper.js
@@ -60,8 +60,9 @@ module.exports = {
         );
       }
 
-      if (options.events) options.events.emit("rpc:request", { payload });
-
+      if (options.events) {
+        options.events.emit("rpc:request", { payload });
+      }
       return payload;
     };
   },


packages/truffle/test/scenarios/migrations/migrate.js
```diff
@@ -46,7 +46,6 @@ describe("migrate (success)", function() {
     assert(output.includes("Replacing 'UsesExample'"));
     assert(output.includes("PayableExample"));
     assert(output.includes("1 ETH"));
-    assert(output.includes("Saving migration"));
```
gnidan, line 49: Wait but what happened to this output?

benjamincburns: It was shown as an ora spinner, which conflicted with the one that we needed to show.

Also it wasn't true "output," as it was sent to stderr and deleted on the `endTransaction` event (see edits to `Reporter.js` and `Message.js` in `events`)

gnidan: Ah, so we're just not doing that behavior anymore. Seems fine, but might want a second opinion from @eggplantzzz

benjamincburns: Another option would be to retain this, but strip out `ora` and use a lib like [`spinnies`](https://github.com/jcarpanelli/spinnies) in its place.

benjamincburns: Has been updated to use `spinnies` now (see `packages/truffle/spinners`)

packages/core/lib/debug/cli.js
```diff
@@ -78,7 +80,12 @@ class CLIDebugger {
           )}.`
         );
       }
-      fetchSpinner.warn(warningStrings.join("  "));
+      // simulate ora's "warn" feature
+      spinners.update({
+        text: `⚠ ${warningStrings.join("  ")}`,
+        color: "yellow",
+        status: "stopped"
+      });
```
cds-amal, line 88: Shouldn't this take a name, specifying the spinner to update?

packages/events/defaultSubscribers/dashboard.js
```diff
@@ -20,35 +13,25 @@ module.exports = {
...
-            })
+            color: "red"
           });
         }
       }
     ],
     "rpc:result": [
       function (event) {
-        const { payload, error, result } = event;
+        let { payload, error, result } = event;
```
cds-amal, line 25: Do all variables have to be mutable?

packages/events/defaultSubscribers/migrate/Messages.js
```diff
@@ -219,8 +221,8 @@ class Messages {
 
       deployed: () => {
         let stopText;
-        if (reporter.blockSpinner) {
-          reporter.blockSpinner.stop();
+        if (spinners.pick("migration-reporter")) {
```
cds-amal, line 224: `pick` confused me because of `lodash.pick`, maybe `get` instead?

packages/events/defaultSubscribers/migrate/Messages.js
```diff
@@ -275,7 +277,9 @@ class Messages {
 
       // Transactions
       endTransaction: () => {
-        if (reporter.blockSpinner) reporter.blockSpinner.stop();
+        if (spinners.pick("migration-reporter")) {
+          spinners.remove("migration-reporter");
+        }
```
cds-amal, line 282: Thoughts on moving this guard to the `spinners.remove` function?

packages/spinners/lib/index.ts
```diff
@@ -0,0 +1,96 @@
...
+export function add(
+  name: string,
+  options: SpinnerOptions | string
+): SpinnerOptions {
+  if (typeof options === "string") {
+    options = {
+      text: options
+    };
+  }
+  return spinnies.add(name, options);
```
cds-amal, line 17: How does `spinnies` handle collisions?

packages/spinners/lib/index.ts
```diff
@@ -0,0 +1,96 @@
...
+ */
+export function pick(name: string): SpinnerOptions {
+  return spinnies.pick(name);
+}
+
+/**
+ * Remove spinner with name.
+ */
+export function remove(name: string): SpinnerOptions {
+  return spinnies.remove(name);
```
cds-amal, line 31: Does this need to be guarded?

cds-amal (commented):
Thanks @benjamincburns ! I left some minor nits and one concern for spinners usage. I think making the does-it-exist guard an API concern would simplify the code, and make it easier to use.

packages/truffle/test/scenarios/migrations/migrate.js
eggplantzzz: What are the tradeoffs ora vs. spinnies? I'm not in love with ora, is spinnies a better package you think?

packages/config/src/configDefaults.ts
```diff
@@ -286,6 +286,7 @@ export const configProps = ({
 
         const options = configObject.network_config;
         options.verboseRpc = configObject.verboseRpc;
+        options.events = configObject.events;
```
gnidan, line 289: Why'd this need to be added here now?

packages/spinners/README.md
```diff
@@ -0,0 +1,8 @@
+# `@truffle/spinners`
```
gnidan, line 1: Is there a way to make this package more general purpose than just spinners? Maybe @truffle/console-utils?

packages/spinners/package.json
```diff
@@ -0,0 +1,51 @@
+{
+  "name": "@truffle/spinners",
+  "version": "0.0.1",
```
gnidan, line 3:
```suggestion
  "version": "0.1.0-0",
```

New packages should start at v0.1.0, and we put the `-0` there for the time being, until the first publish (when we increment to 0.1.0)

packages/spinners/package.json
```diff
@@ -0,0 +1,51 @@
...
+    "etherscan",
+    "ipfs",
+    "solidity",
+    "verify-source",
+    "sourcify",
+    "compile"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
```
gnidan, line 39: came here to make sure this was here :)

packages/core/lib/debug/cli.js
gnidan: On that note... could this "warn" feature be a dedicated interface provided by the new package you're adding here?

packages/events/defaultSubscribers/migrate/Messages.js
gnidan: :heavy_plus_sign:

gnidan (commented):
This looks good to me, apart from a couple questions/comments.

General question though: should we split this up into two PRs? i.e., an initial PR to swap ora for spinnies, then a subsequent PR to "fix dashboard subscriber"

Thanks for this!

packages/events/defaultSubscribers/migrate/Messages.js
benjamincburns: I just copied the interface from spinnies, as the goal was pretty much to make a static wrapper. I'll add a `get` alias, however.

FWIW, if I wanted to spend more time on this I probably would've written a more abstracted wrapper. I don't like that every operation has to be run through the top-level module functions. I wish it was just returning a `Spinner` object, for example.

packages/events/defaultSubscribers/migrate/Messages.js
benjamincburns: good call!

packages/events/defaultSubscribers/obtain.js
```diff
@@ -25,47 +28,60 @@ module.exports = {
...
     ],
     "downloadCompiler:succeed": [
-      function() {
-        if (this.quiet) return;
-        this.spinner.succeed();
+      function () {
+        if (this.quiet) {
+          return;
+        }
+        spinners.succeed("obtain-handler");
```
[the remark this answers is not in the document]
benjamincburns: It would likely throw, and I think in this case throwing on an unknown spinner name is desirable for catching spinner lifecycle bugs.

packages/spinners/lib/index.ts
benjamincburns: Good question - I'm 99% sure it just overwrites, as internally it just keeps each spinner config in a dictionary-like object.

benjamincburns: Yeah, per [the `spinnies` code](https://github.com/jcarpanelli/spinnies/blob/90e33ee6384eab6d4783a506301ab3dbaee24b00/index.js#L36-L51), it just overwrites.

packages/spinners/lib/index.ts
benjamincburns: Good question. It [throws](https://github.com/jcarpanelli/spinnies/blob/90e33ee6384eab6d4783a506301ab3dbaee24b00/index.js#L75-L81) if the spinner name isn't known. However I think per my remark above, this is probably desirable in some cases, as it will help us to catch spinner lifecycle bugs at dev time.

packages/events/defaultSubscribers/obtain.js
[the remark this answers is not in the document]
benjamincburns: Does throw from w/i [this function](https://github.com/jcarpanelli/spinnies/blob/90e33ee6384eab6d4783a506301ab3dbaee24b00/index.js#L105-L112) (called by `succeed` [here](https://github.com/jcarpanelli/spinnies/blob/90e33ee6384eab6d4783a506301ab3dbaee24b00/index.js#L62)).

packages/config/src/configDefaults.ts
benjamincburns: Without this line the `events` field wasn't set on the `config` object at the time that the `wrap` function is called in `@truffle/provider`. When that's the case, the `preHook` and `postHook` don't fire off their respective `rpc:*` events.

benjamincburns: [block where event is emitted](https://github.com/trufflesuite/truffle/blob/12c76a0ec9647c94400101e6dd1fc2249aeb8806/packages/provider/wrapper.js#L87-L92)

packages/spinners/README.md
benjamincburns: I considered that, but decided against it for two reasons:

1. Keeping it in a dedicated module keeps the dependency tree lighter for modules that only want this functionality (though hopefully this can be improved via tree shaking when we switch over to ESMs)
2. I didn't want to create a module with a fuzzy boundary around its functional scope that might grow organically in such a way that would thwart a more considered presentation-layer design/architecture effort.

packages/events/defaultSubscribers/dashboard.js
benjamincburns: No - could break it out, but I like the readability of the destructuring syntax.

packages/core/lib/debug/cli.js
benjamincburns: Missing name is indeed a bug. bit odd that the tests didn't catch that... 🤔

benjamincburns: Re: moving it to the dedicated interface, this is a feature gap between `ora` and `spinnies`, and this is the only place where the `ora` `warn` feature was used. Given that this is a bit of a bodge, I didn't want to encourage its use by adding it to our wrapper.

That said, this would be a fairly straightforward PR for `spinnies`. Perhaps I'll do that and we can just add that to the wrapper when that PR is accepted?

benjamincburns:
> General question though: should we split this up into two PRs? i.e., an initial PR to swap ora for spinnies, then a subsequent PR to "fix dashboard subscriber"

@gnidan what's the value in splitting the PRs out given that we don't squash commits? My inclination would've been to split them but I decided not to because the `spinnies` work was direct fallout from the review on this PR, so it seemed like it'd be splitting context to create a new PR for it.

gnidan:
> @gnidan what's the value in splitting the PRs out given that we don't squash commits? My inclination would've been to split them but I decided not to because the `spinnies` work was direct fallout from the review on this PR, so it seemed like it'd be splitting context to create a new PR for it.

Mostly I want to make sure that we separate Dashboard Task Force–focused work from general Truffle changes. Switching out spinners logic is a substantial change that, IMO, deserves a separate PR that everyone on the team can scrutinize.

benjamincburns:
> Switching out spinners logic is a substantial change that, IMO, deserves a separate PR that everyone on the team can scrutinize.

Ah, I was thinking that since it blocks #4824, it fell under the umbrella of task force work.

packages/truffle/test/scenarios/migrations/migrate.js
benjamincburns: spinner work has been moved to #4934

packages/core/lib/debug/cli.js
benjamincburns: Wound up doing that refactoring work, as it made the module much nicer to consume. see #4934 for details on that, as I've split it out from this PR, per @gnidan's request.

packages/events/defaultSubscribers/migrate/Messages.js
benjamincburns: fixed in #4934

packages/events/defaultSubscribers/migrate/Reporter.js
```diff
@@ -210,8 +208,9 @@ class Reporter {
     this.currentBlockWait =
       `Blocks: ${data.blocksWaited}`.padEnd(21) +
       `Seconds: ${data.secondsWaited}`;
-    if (this.blockSpinner) {
-      this.blockSpinner.text = this.currentBlockWait;
+
+    if (spinners.pick("migration-reporter")) {
+      spinners.updateText("migration-reporter", this.currentBlockWait);
```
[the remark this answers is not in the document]
benjamincburns: fixed in #4934

gnidan (approved):

PR #4904, opened 2022-03-21
fix: Improve code and readability

aymen94:
Improve code, using a constant, init variable to improve readability and clean code

### DIFF
--- a/scripts/geth-accounts.js
+++ b/scripts/geth-accounts.js
@@ -8,6 +8,7 @@
  */
 
 function createAccounts() {
+  let acc;
   for (var i = 0; i < 10; i++) {
     acc = personal.newAccount("");
     personal.unlockAccount(acc, "");


scripts/geth-accounts.js
```diff
@@ -8,6 +8,7 @@
  */
 
 function createAccounts() {
+  let acc;
```
haltman-at, line 11: Thanks for catching this!

scripts/wait-for-quorum-network.js
```diff
@@ -24,12 +24,13 @@ async function retry(fn, retries = 5, interval = 1000) {
...
 const web3 = new Web3();
+const interval = 5000;
 web3.setProvider(
   `http://${developmentNetworkConfig.host}:${developmentNetworkConfig.port}`
 );
 
 console.log("Waiting for the quorum network to be ready...");
 
-retry(web3.eth.getAccounts, (interval = 5000))
+retry(web3.eth.getAccounts, interval)
```
haltman-at, line 34: This change just seems neutral to me.

scripts/solc-bump.js
```diff
@@ -13,12 +13,14 @@ const PACKAGES_DIR = path.resolve(__dirname, "..", "packages");
...
   const packages = fs.readdirSync(PACKAGES_DIR);
+  const packageJsonName = "package.json";
+  const encoding = "utf8";
   return packages
     .map(packageName => {
       // parse the package.json file for `packageName`
       const packagePath = path.resolve(PACKAGES_DIR, packageName);
-      const packageJsonPath = path.resolve(packagePath, "package.json");
-      const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, "utf8"));
+      const packageJsonPath = path.resolve(packagePath, packageJsonName);
```
haltman-at, line 22: I don't think there's any reason to turn `"package.json"` into a named constant instead of a plain string literal; I think it makes the code less clear, not more.  Everyone knows what `"package.json"` means; it's quite clear on its own, it's not like some magic numbers whose meaning is unclear until it's given a name.  Whereas it's a lot clear what `packageJsonName` means (and makes you think that there might be multiple possibilities for it).  Then you look up the definition and it's just `"package.json"`... it would have been clearer to do without the indirection.  And right now, we do.  I don't think we should make this change.

scripts/solc-bump.js
```diff
@@ -13,12 +13,14 @@ const PACKAGES_DIR = path.resolve(__dirname, "..", "packages");
...
+      const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, encoding));
```
haltman-at, line 23: My opinion for this one is similar; there's no reason to write `encoding` rather than `"utf8"`.  Again, everyone knows what `"utf8" means; there's no magic number problem here.  Putting it in a constant called `encoding` once again just makes it look like there might be other encodings we have to deal with; but in fact, we never use any encoding but `"utf8"`.  As with the previous one, I think this change decreases clarity, not increases it.

haltman-at (changes requested):
Hm, I'm not a fan of all of these changes.  I think some of thse are good but some of these are detrimental.  I'd certainly approve a PR with some of these changes, but I don't think we should incorporate all of them.

scripts/wait-for-quorum-network.js
cds-amal: The retry function defines `interval` as the 3rd positinal argument and `async function retry(fn, retries = 5, interval = 1000)` and this invocation mistakenly results in 5000 retries at 1 sec intervals! whoa!

cds-amal (changes requested):
Thanks for bringing attention to these files @aymen94. Unfortunately, I have to hold this PR while we investigate the purpose of these files, which are currently unused by the project. The correct refactor might be to delete them. We'll find out and update here. Sorry about that!

We do appreciate the contribution, and I'd like to point out we have `good first issues` on the [Truffle repo](https://github.com/trufflesuite/truffle/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) as well our [website repo](https://github.com/trufflesuite/trufflesuite.com/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22).

scripts/wait-for-quorum-network.js
haltman-at: Oh, yikes, so there's a problem here that needs fixing then? (In the original code, I mean, rather than in the PR.)

aymen94:
@haltman-at @haltman-at, no worry, the important thing is that the issues have been noticed
If you indicate to me the PR where you will do the refactoring, I could make my contribution.

eggplantzzz:
@gnidan What do you think about removing the quorum script and the script that bumps solc for all packages? They don't seem to be useful any more...?

sukanyaparashar:
Thanks for your contribution @aymen94 . I have removed the unused scripts `solc-bump.js` and `wait-for-quorum-network.js` in this PR https://github.com/trufflesuite/truffle/pull/4952. Can you please rebase your branch to get approval for the change you have made in `get-accounts.js` script?

eggplantzzz (approved):
I like it!

eggplantzzz (approved):
I like it!

cds-amal (approved):
@aymen94, we sincerely appreciate your work that helped us identify and remove dead code.  We love these contributions! Thanks so much!

PR #4906, opened 2022-03-21
Internal improvement: Increase timeout on docker tags test

haltman-at:
We're getting a lot of timeouts recently!  This test in particular has been timing out a lot recently!  So I'm upping it!

If this PR fails CI I may add other timeout increases to it as well.


cds-amal (approved):
Thanks @haltman-at, those CI timeouts are annoying! Let's try this.

Issue #4842, opened 2022-03-11
Add analytics message type to message bus


MicaiahReid:

MicaiahReid:
@haltman-at Reopening because the PR only merged into a feature branch, not into `develop`

lsqproduction:
We didn't need message type in the end. Work done in [v5.7.3](https://github.com/trufflesuite/truffle/releases/tag/v5.7.3). Closing.

PR #4907, opened 2022-03-21
dashboard-message-bus:add analytics message type

lsqproduction:
address #4842

### DIFF
--- a/packages/dashboard-message-bus/lib/message/types.ts
+++ b/packages/dashboard-message-bus/lib/message/types.ts
@@ -46,7 +46,7 @@ export interface InvalidateMessage extends Message {
  * The payload indicate whether or not the user opts in to share anonymous usage data with Truffle.
  * This message payload is a truffle command request to update user config that should be forwarded.
  */
-export interface SetAnalytics extends Message {
+export interface SetAnalyticsMessage extends Message {
   type: "analytics";
   payload: boolean;
 }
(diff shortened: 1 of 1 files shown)


fainashalts (approved):
Looks good to me! Nice use of type predicate!

packages/dashboard-message-bus/lib/message/types.ts
```diff
@@ -41,6 +41,16 @@ export interface InvalidateMessage extends Message {
   payload: number;
 }
 
+/**
+ * Message intended to set analytic collection preference for user.
+ * The payload indicate whether or not the user opts in to share anonymous usage data with Truffle.
+ * This message payload is a truffle command request to update user config that should be forwarded.
+ */
+export interface SetAnalytics extends Message {
```
gnidan, line 49: Looks like the other message types have `Message` in the name:

```suggestion
export interface SetAnalyticsMessage extends Message {
```

gnidan (changes requested):
Looks like we should keep the name style consistent, at least in this PR. Otherwise, looks great!

packages/dashboard-message-bus/lib/message/types.ts
gnidan: (You'll have to update the other part of the code, also)

lsqproduction: thank you @gnidan 
I will update it.

gnidan (approved):
Awesome, thanks!

PR #4914, opened 2022-03-22
Add network switching capabilities to Truffle Dashboard

MicaiahReid:


haltman-at:
Hm, CI isn't running for some reason...

packages/dashboard-message-bus/lib/message/types.ts
```diff
@@ -31,6 +31,16 @@ export interface LogMessage extends Message {
   };
 }
 
+/**
+ * Message to log in Dashboard browser console
+ */
+export interface DebugMessage extends Message {
+  type: "debug";
```
cds-amal, line 38: Perhaps it's time to refine the Message interface to keep track of valid types. @gnidan ?

```typescript

type MessageType = "debug" | "provider" | "log" // | etc...
export interface Message {
  id: number;
  type: MessageType;
  payload: any;
}
```

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -76,8 +101,11 @@ export class DashboardServer {
 
     this.expressApp.get("/ports", this.getPorts.bind(this));
 
+    this.subscribeSocket = await this.connectToSubscribePort();
+
     if (this.rpc) {
```
MicaiahReid, line 106: This probably is legacy code that should be removed. In what circumstances would a user want RPC disabled here?

packages/dashboard/src/Dashboard.tsx
```diff
@@ -19,36 +23,40 @@ import ConfirmNetworkChanged from "./components/ConfirmNetworkChange";
...
     } else {
       setConnectedChainId(chainId);
     }
-  }, [chainId, connectedChainId, socket]);
+  }, [chainId, connectedChainId, subscribeSocket]);
 
-  const initializeSocket = async () => {
-    if (socket && socket.readyState === WebSocket.OPEN) return;
+  const initializeSockets = async () => {
+    await initializeSubSocket();
```
MicaiahReid, line 49: Put these in a Promise.all so they can happen RACE.

packages/dashboard/src/Dashboard.tsx
```diff
@@ -70,31 +78,82 @@ function Dashboard() {
...
+
+  const initializePubSocket = async () => {
+    if (publishSocket && publishSocket.readyState === WebSocket.OPEN) return;
 
-    connectedSocket.send("ready");
+    const { publishPort } = await getPorts();
+    const socket = await initializeSocket(
+      publishPort,
+      (event: WebSocket.MessageEvent) => {
+        if (typeof event.data !== "string") {
```
MicaiahReid, line 97: The publish socket probably doesn't need any event handler at this point. We can use the socket to make a request for data but we don't need every message.

packages/dashboard/src/Dashboard.tsx
```diff
@@ -70,31 +78,82 @@ function Dashboard() {
...
+              JSON.stringify(logMessage.payload.message)
+          );
+        }
+      }
+    );
+
+    // since our socket is open, request some initial data from the server
+    const message = createMessage("initialize", ""); // no payload needed
+    const response = await sendAndAwait(socket, message);
+    if (response.payload.error) {
```
MicaiahReid, line 116: Verify this is actually how the error would show up.

packages/dashboard/lib/DashboardServer.ts
MicaiahReid: Removed.

gnidan:
Hm there's a `package-lock.json` that shouldn't be here

MicaiahReid:
Closing in favor of #5045

PR #4916, opened 2022-03-22
Stop forcibly exiting Truffle at end of command

gnidan:
Let's see if CI finishes here ;)


haltman-at:
Isn't this redundant with #4879...?

gnidan:
> Isn't this redundant with #4879...?

Oh! It's completely redundant. Closing this now for two reasons (the other being that my test here failed)

PR #4923, opened 2022-03-23
Hook up example events -> dashboard communication

gnidan:
This PR does a few things:
- Add a new `"debug"` message type, intended for logging in the browser console. This message type behaves like `"provider"` messages, in that it waits for a response from a subscriber.
- Handle this new message type in the frontend by logging and then issuing a blank response to clear it from the message-bus requests queue
- Initialize a quick DashboardMessageBusClient for use in @truffle/events. Define a `sendAndAwait` method to behave similarly to the @truffle/dashboard-message-bus util function
- Use this client to send an example debug message to the browser upon `"compile:start"`

### DIFF
--- a/packages/dashboard-message-bus/lib/message/types.ts
+++ b/packages/dashboard-message-bus/lib/message/types.ts
@@ -32,6 +32,16 @@ export interface LogMessage extends Message {
 }
 
 /**
+ * Message to log in Dashboard browser console
+ */
+export interface DebugMessage extends Message {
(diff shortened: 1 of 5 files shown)


cds-amal (approved):
Nice and quick. I like things that unblock.

benjamincburns (approved):

gnidan:
Thanks for the fix @benjamincburns... your changes look good to me, so I think we can merge this!

PR #4926, opened 2022-03-23
Allow connecting Truffle Dashboard via WalletConnect

michaeljohnbennett:
Initial POC on integrating WAGMI and WalletConnect into the dashboard codebase.

Still to do some testing on the walletconnect side against some of the wallets but normal metamask code is working as expected no we fixed a few things with the provider library selectors/hooks in the code.


michaeljohnbennett:
@dkillen will be doing a final bit of log cleanup and then this will be good to go.

Generally the code is good for review. We will be cleaning up a few log elements prior to it being read to merge.

packages/dashboard/src/utils/utils.ts
```diff
@@ -104,21 +126,20 @@ export const getNetworkName = async (chainId: number) => {
 };
 
 export const getDisplayName = async (
-  library: providers.Web3Provider,
+  provider: providers.BaseProvider,
   address: string
 ) => {
-  const ensName = await reverseLookup(library, address);
+  const ensName = await reverseLookup(provider, address);
```
MicaiahReid, line 132: Did you get this working? It seems to only be returning `undefined` for me, and I have reverse lookup setup.

MicaiahReid (changes requested):

MicaiahReid (commented):
There were some stylistic changes (increasing the Button size which increases the header height) that I'm not too picky about, but I'd definitely want more eyes on. Since we'll have a lot of dashboard changes coming through from the task force, I want to be sure we're being stylistically consistent.

packages/dashboard/src/components/common/Button.tsx
```diff
@@ -8,7 +8,7 @@ interface Props {
 function Button({ onClick, text }: Props) {
   return (
     <button
-      className="rounded p-2 bg-truffle-blue text-truffle-brown uppercase hover:bg-white"
+      className="rounded p-2 mx-2 bg-truffle-blue text-truffle-brown uppercase hover:bg-white"
```
michaeljohnbennett, line 11: @MicaiahReid reverted the change in this button.

michaeljohnbennett:
@MicaiahReid prettier ran, button sizing fixed.

packages/dashboard/src/components/Header/Header.tsx
```diff
@@ -1,37 +1,49 @@
...
+    updateAccountDisplay(connectData.connector?.getProvider(), accountData.address);
+  }, [connectData, accountData]);
+
 
   return (
     <header className="grid grid-cols-2 py-2 px-4 border-b-2 border-truffle-light text-md uppercase">
       <div className="flex justify-start items-center">
         <span className="inline-flex items-center gap-3">
-          <img src="/truffle-logomark.svg" width="32px" />
+          <img src={"/truffle-logomark.svg"} width="32px" alt={'Truffle Logo'}/>
```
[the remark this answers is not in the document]
michaeljohnbennett: all fixed.

michaeljohnbennett (commented):
fixes made as per @MicaiahReid's feedback.

MicaiahReid (approved):
Looks good to me! If we want to get this merged sooner, maybe we could pull out the ENS name bit for another PR? I suppose there's no harm done in it not working currently, so maybe we can leave it in and add an issue to get it working ¯\_(ツ)_/¯

michaeljohnbennett:
lets leave it in and get people trying it, it might be a simple fix once we have a few people playing with it, we can pair on this once merged if you have an ENS we can test with.

PR #4927, opened 2022-03-23
Pin faker to latest non-malicious version

FrederikBolding:
https://github.com/advisories/GHSA-5w9c-rv96-fr7g

### DIFF
--- a/packages/abi-utils/lib/arbitrary.ts
+++ b/packages/abi-utils/lib/arbitrary.ts
@@ -1,5 +1,5 @@
 import * as fc from "fast-check";
-import faker from "@faker-js/faker";
+import faker from "faker";
 import { camelCase, pascalCase } from "change-case";
 
 import type * as Types from "./types";
--- a/packages/abi-utils/lib/arbitrary.ts
+++ b/packages/abi-utils/lib/arbitrary.ts
@@ -1,5 +1,5 @@
 import * as fc from "fast-check";
-import faker from "faker";
+import faker from "@faker-js/faker";
 import { camelCase, pascalCase } from "change-case";
 
 import type * as Types from "./types";
--- a/packages/abi-utils/lib/faker.d.ts
+++ b/packages/abi-utils/lib/faker.d.ts
@@ -1,0 +1,5 @@
+// faker.d.ts
+declare module '@faker-js/faker' {
+    import faker from 'faker';
+    export default faker;
+}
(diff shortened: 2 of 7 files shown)


eggplantzzz:
Hey @FrederikBolding, can you explain why you want to make this change? As of right now I see no reason for this.

FrederikBolding:
@eggplantzzz Sorry, I forgot to add the details to the GitHub issue.

Essentially it looks like the `faker.js` dev has decided to nuke the project. Which prompted GitHub to create this advisory (https://github.com/advisories/GHSA-5w9c-rv96-fr7g) which is flagged in the MetaMask extension build process (when running `yarn audit`).

FrederikBolding:
It does seem that the advisory is wrong in saying that all versions are affected though, I've attempted to submit a suggestion to remove that. But nevertheless I imagine you guys might want to switch 😅

haltman-at:
Yeah this just seems unnecessary to me given that we've already pinned our `faker` version to one that's not affected.  Maybe if the fork is actually maintained in the future there might be a reason to switch later?

FrederikBolding:
> Yeah this just seems unnecessary to me given that we've already pinned our `faker` version to one that's not affected. Maybe if the fork is actually maintained in the future there might be a reason to switch later?

The advisory was fixed to limit the scope of the issue.

I guess the only potential issue is the author releasing a malicious package with a minor version bump instead of a major one. Then there is a chance it can be installed. But if that's not an issue, feel free to close this 👍 

Also I do think the fork is being maintained, seems they released a TypeScript version etc.

haltman-at:
> I guess the only potential issue is the author releasing a malicious package with a minor version bump instead of a major one. Then there is a chance it can be installed.

Oh, huh, looks like we didn't actually pin the version after all, we just relied on the fact that the change was in a major version!  I forgot about that... well, we can actually pin it if it comes to that.

eggplantzzz:
So @haltman-at, what do you want to do for now? I would say maybe we pin it for now until at some point in the future we can verify that it will be maintained by someone reliable, what do you think?

FrederikBolding:
Happy to pin it if that is your chosen solution 👍

eggplantzzz:
Let me just confirm with @haltman-at that we can pin it for now. Thanks for doing this @FrederikBolding!

eggplantzzz:
Ok @FrederikBolding, could you go ahead and pin the version of `faker` being used? We'll consider moving to a forked version later on if it becomes necessary. Thank you for doing this!

FrederikBolding:
@eggplantzzz Done!

eggplantzzz:
Thanks again!

eggplantzzz (approved):

cds-amal (approved):
LGTM

Issue #4928, opened 2022-03-23
Unhelpful error when using pragma setting and source does not contain pragma expression


eggplantzzz:
If you are using the "pragma" setting while compiling, Truffle will fail unhelpfully if a "direct source" (one supplied by the user in their `contracts` directory) does not contain a pragma expression. The error is something like 
```
TypeError: Cannot read properties of null (reading '1')
at getSemverExpression (/usr/local/lib/node_modules/truffle/build/webpack:/packages/compile-solidity/dist/compileWithPragmaAnalysis.js:19:46)
```

Truffle should error, explaining to the user that their "direct sources" must contain pragma expressions while using this setting.

haltman-at:
OK, this should be fixed in 5.5.6.

PR #4930, opened 2022-03-23
Bug fix: Check matching result for null values with pragma setting

eggplantzzz:
addresses https://github.com/trufflesuite/truffle/issues/4928

There is currently a bug where Truffle throws an unhelpful error when a user utilizes the "pragma" compiler setting and doesn't include a pragma expression in one of their sources. Though this is legal Solidity, Truffle requires the user to have a pragma expression in the sources the user creates because Truffle needs to load a version of the Solidity parser to find imports for the source.

There is code meant to handle this case but it turns out that it fails before running this code as it tries to index into a regex match when that result could possibly be `null`. This PR adds a check that determines whether a result is present before attempting to index into the result.

I'm going to try and write (a) test(s) for this bug fix.

### DIFF
--- a/packages/compile-solidity/test/compileWithPragmaAnalysis.js
+++ b/packages/compile-solidity/test/compileWithPragmaAnalysis.js
@@ -116,6 +116,16 @@ describe("compileWithPragmaAnalysis", function () {
       assert.equal(compilations.length, 3);
     });
 
+    it.only("compiles files with imports without pragma expressions", async function () {
+      const { compilations } = await compileWithPragmaAnalysis({
+        options: config,
+        paths: paths.concat([
+          path.join(sourceDirectory, "withImports", "D.sol")
+        ])
+      });
+      assert.equal(compilations.length, 1);
+    });
+
     it("finds a version that satisfies all pragmas if it exists", async function () {
       const { compilations } = await compileWithPragmaAnalysis({
         options: config,
--- a/packages/compile-solidity/test/compileWithPragmaAnalysis.js
+++ b/packages/compile-solidity/test/compileWithPragmaAnalysis.js
@@ -86,7 +86,7 @@ describe("compileWithPragmaAnalysis", function () {
 
     // note that it will find the newest version of Solidity that satisifes
     // each pragma expression and then do one compilation per version
-    it("will make one compilation per compiler version", async function () {
+    it("makes one compilation per compiler version", async function () {
       const { compilations } = await compileWithPragmaAnalysis({
         options: config,
         paths
@@ -94,7 +94,7 @@ describe("compileWithPragmaAnalysis", function () {
       assert.equal(compilations.length, 3);
     });
 
-    it("will compile files with the same version together", async function () {
+    it("compiles files with the same version together", async function () {
       const { compilations } = await compileWithPragmaAnalysis({
         options: config,
         paths: paths.concat(
@@ -116,12 +116,10 @@ describe("compileWithPragmaAnalysis", function () {
       assert.equal(compilations.length, 3);
     });
 
-    it.only("compiles files with imports without pragma expressions", async function () {
+    it("compiles files with imports without pragma expressions", async function () {
       const { compilations } = await compileWithPragmaAnalysis({
         options: config,
-        paths: paths.concat([
-          path.join(sourceDirectory, "withImports", "D.sol")
-        ])
+        paths: [path.join(sourceDirectory, "withImports", "D.sol")]
       });
       assert.equal(compilations.length, 1);
     });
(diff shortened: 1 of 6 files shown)


packages/compile-solidity/test/compileWithPragmaAnalysis.js
```diff
@@ -149,6 +144,25 @@ describe("compileWithPragmaAnalysis", function () {
...
+          options: config,
+          paths: [path.join(__dirname, "sources", "badSources", "NoPragma.sol")]
+        });
+        assert.fail("The function should have thrown.");
+      } catch (error) {
+        const expectedMessage = "Could not find a valid pragma expression";
+        if (error.message.includes(expectedMessage)) {
+          return "all good";
+        }
+        assert.fail(
```
haltman-at, line 159: With this and the other unexpected error fails you added, I think it probably makes more sense to directly rethrow the unexpected error rather than using `assert.fail`; that way we get better stacktrace information.  (It's worth noting that doing things this way, we'll actually never get a stacktrace pointing to the `assert.fail` on line 153... because it'll be caught by the `catch` on line 154, and it's this `assert.fail` on line 159 that the stacktrace will actually point to!  Just rethrowing instead would avoid this problem, and other similar problems.)

packages/compile-solidity/test/compileWithPragmaAnalysis.js
```diff
@@ -157,11 +171,14 @@ describe("compileWithPragmaAnalysis", function () {
         });
         assert.fail("The function should have thrown.");
       } catch (error) {
-        const expectedMessage = "Could not find a pragma expression";
+        const expectedMessage = "Could not find a valid pragma expression";
         if (error.message.includes(expectedMessage)) {
           return "all good";
         }
-        throw error;
+        assert.fail(
```
haltman-at, line 178: Similarly here.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
```diff
@@ -179,7 +196,10 @@ describe("compileWithPragmaAnalysis", function () {
...
```
haltman-at, line 199: Similarly here.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
```diff
@@ -149,6 +144,25 @@ describe("compileWithPragmaAnalysis", function () {
...
+    it("throws an error when a 'direct source' lacks a pragma expression", async function () {
+      try {
+        await compileWithPragmaAnalysis({
+          options: config,
+          paths: [path.join(__dirname, "sources", "badSources", "NoPragma.sol")]
+        });
+        assert.fail("The function should have thrown.");
+      } catch (error) {
+        const expectedMessage = "Could not find a valid pragma expression";
+        if (error.message.includes(expectedMessage)) {
```
haltman-at, line 156: Rather than using this awkward construct with `return;` (and if you are going to do this I would suggeste eliminating the `"all good"` return value as placing that value there suggests it's going to be used for something and it isn't), I think it would just make sense to invert the conditional and put the failure in the `if`.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
```diff
@@ -149,6 +144,25 @@ describe("compileWithPragmaAnalysis", function () {
   });
 
   describe("when there is a semver expression error", function () {
+    it("throws an error when a 'direct source' lacks a pragma expression", async function () {
```
haltman-at, line 147: Question: Do we currently have a test that it does *not* error when an *imported* source lacks a pragma expression?  That might also be good to add.

haltman-at (approved):
Thanks Tyler!  I have some style comments on the tests, and one test I think should possibly be added, but I'm going to hit approve because I don't know that any of that is all really that important (all of my comments are about the tests, not the functional code :P ).

haltman-at:
Also hey this fixes #4928 right?  You should mention that in the description so whoever's on release doesn't miss it.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
eggplantzzz: Yeah I thought about removing that, it sounds like a good idea.

packages/compile-solidity/test/compileWithPragmaAnalysis.js
eggplantzzz: For all of these I thought it was more proper to use mocha mechanics to determine whether the tests failed or not. In other words, it seems slightly janky to throw an error in the test itself to make it fail as it is slightly less direct. But I opted to print the error in the "fail reason string" in order to keep the reason. You still think it would be better to throw? If so I don't really feel strongly enough about this to argue more :)

eggplantzzz: I went ahead and just did a straight up assertion that the message was in the error and printed the error in the reason string. Let me know what you think of that.

haltman-at: The new version seems to work by essentially the same mechanics as the old version?  It still has the same problems.  I'm not seeing any substantial difference here.

>  In other words, it seems slightly janky to throw an error in the test itself to make it fail as it is slightly less direct.

Well, it's not throwing a *new* error, it's *rethrowing* an existing error.  That's useful to preserve information; doing an assertion is actually throwing a new error and will obscure the source!  As I mentioned, currently one of your assertions will *never* cause the ultimate error because it just gets caught... that's confusing!

PR #4932, opened 2022-03-24
Internal improvement: Allow a tiny amount of tolerance on float-to-decimal tests

haltman-at:
This PR fixes a problem that would occur in the encoder tests where it would occasionally fail due to tiny imprecisions in float-to-decimal conversions.

I don't know of any good way to ensure that the encoder gets these cases exactly right, so instead, I'm allowing a little bit of wiggle room.  Note that we only allow the wiggle room when passing in a `number` or `Number`; for other inputs, which are decimal to begin with, we require exact matching.  But when passing in a binary floating-point input, i.e. `number` or `Number`, I've allowed a tolerance of 10^-24.  That's probably as good as we're going to get, since I don't see a good way right now to make `encoder` handle floating-point inputs better.  Note that because we need to detect `Number`s as well as `number`s, we use lodash's `isNumber` rather than just using `typeof`.

Anyway, the tests seem to consistently pass now, after having run them a number of times.

Note: `prettier` made a lot of changes on this PR, so most of the changes you'll see here aren't real and can be ignored.  The real changes are the ones before line 29, and after line 388; everything else is just `prettier`.

### DIFF
--- a/packages/encoder/test/wrap-fc.test.ts
+++ b/packages/encoder/test/wrap-fc.test.ts
@@ -9,6 +9,7 @@ import {
   BigNumber as EthersBigNumber,
   FixedNumber
 } from "@ethersproject/bignumber";
+import isNumber from "lodash/isNumber";
 
 import * as fc from "fast-check";
 import { testProp } from "jest-fast-check";
@@ -20,6 +21,8 @@ const Bytes = () => fc.nat(31).map(k => 32 - k);
 const Bits = () => Bytes().map(k => 8 * k);
 const DecimalPlaces = () => fc.nat(79).map(k => ((k + 17) % 80) + 1);
 const Precision = () => fc.record({ bits: Bits(), places: DecimalPlaces() });
+
+const tolerance = 1e-24; //for encoding of JS numbers only!
 
 const addBoxes = (inputs: any[]) => {
   let boxed: any[] = [];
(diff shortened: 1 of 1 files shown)


packages/encoder/test/wrap-fc.test.ts
```diff
@@ -277,48 +346,81 @@ describe("Wrapping of elementary values", () => {
...
 
   testProp("Unsigned decimals", [UfixedInfo()], async info => {
     const wrapped = <Codec.Format.Values.UfixedValue>(
       await encoder.wrapElementaryValue(info.type, info.input)
     );
-    assert(
-      wrapped.value.asBig.eq(info.value),
-      `Wrapped Big ${wrapped.value.asBig.toFixed()} did not match original value ${info.value.toFixed()}`
-    );
+    if (isNumber(info.input)) {
```
eggplantzzz, line 392: so what are the type options for input here? numbers and what else?

eggplantzzz (approved):
I like it!

packages/encoder/test/wrap-fc.test.ts
haltman-at: This is defined in `FixedInput`; currently we test MikeMcl's `Big`s and `BigNumber`s, ethers's `FixedNumber`s and `BigNumber`s, strings, `BN`s, `BigInt`s, and `number`s.

...that's odd, I don't see `Number`s.  Oops, guess I got mixed up there.  You think I should change it back to use `typeof`?

eggplantzzz: which ones do we want this to apply to, just `number`s?

I'll leave it up to you, I'm not really sure

haltman-at: Yeah, I think I just won't bother changing it...

PR #4935, opened 2022-03-24
Add pending tx view and account dialog to Truffle Dashboard

michaeljohnbennett:
Addition of Pending TXs and account view dialogs to code.

Emulating other dapps where pending transactions are visible when submitted to the network provider.

Addition of an account dialog using code similar to the sushiswap interface.

Builds on WAGMI code and elements added in other PR awaiting merge (Walletconnect #4926).

Draft right now, mostly due to a lot of features to be added in but here now for reference to other team members as a possible way forward in terms of elements/components to use.


michaeljohnbennett:
All work done now, ready for review and testing from other team members.

Toasts
State of Transactions held in the app.
Hooks for successful TXs and resulting updates of reducer store.
Refactoring and cleanup.

PR #4936, opened 2022-03-24
Add dashboard-message-bus-client module

benjamincburns:


packages/dashboard-message-bus-client/lib/client.ts
```diff
@@ -0,0 +1,223 @@
+import { DashboardMessageBusClientOptions } from "./types";
```
gnidan, line 1: This should probably just be defined in this file, since it's only used in this file and since it's going to be the first thing one looks for when reading this file

packages/dashboard-message-bus-client/lib/client.ts
```diff
@@ -0,0 +1,223 @@
...
+      () => {
+        const message = createMessage(type, payload);
+        return this._publishSocket.send(message);
+      },
+      this._options.connectRetryCount,
+      bestEffort
+    );
+  }
+
+  async sendAndWait<MessageType extends Message, ResponseType extends Message>({
```
gnidan, line 73: There's a mismatch here with the dashboard-message-bus util, which is called `sendAndAwait`

packages/dashboard-message-bus-client/lib/client.ts
```diff
@@ -0,0 +1,223 @@
...
+    this.cause = cause;
+  }
+  toString(): string {
+    return this.cause
+      ? `${super.toString()}\ncaused by: ${this.cause.toString()}`
+      : super.toString();
+  }
+}
+
+export class DashboardMessageBusClient {
```
gnidan, line 25: Let's get this at the top of the file, with only the Options type defined above it :)

packages/dashboard-message-bus-client/lib/client.ts
```diff
@@ -0,0 +1,223 @@
...
+        return this._publishSocket.sendAndWait<MessageType, ResponseType>(
+          message as MessageType
+        );
+      },
+      this._options.connectRetryCount,
+      bestEffort
+    );
+  }
+
+  async getMessage<ResponseType extends Message>(
```
gnidan, line 90: This method's name wasn't clear to me at first what it does. It's also very polymorphic... can we avoid that by giving it its own Options type?

gnidan (commented):
Hm... I see what you mean in asking for review of interface. Maybe we should sync and brainstorm about that `getMessage` flow... it seems to be doing a lot of disparate heavy lifting, maybe we could come up with something that's more succinct to explain?

packages/dashboard-message-bus-client/lib/client.ts
benjamincburns: haha, yep - looks like the changes I pushed last night covered that off already :-)

packages/dashboard-message-bus-client/lib/client.ts
benjamincburns: I think our discussion negated this one, but as an aside, ideally the stuff that's currently in the `dashboard-message-bus/util` module will become internal to the `dashboard-message-bus` package. Not to say that naming consistency isn't useful all the same, however.

benjamincburns:
I think this is ready for review now (provided CI goes green). The `dash` branch has had the most recent `develop` branch merged in, and this has been rebased against `dash`, so it should merge cleanly. Also I think I've pulled out most of the improvements that weren't _strictly_ necessary for this PR, though I'm sure a few things remain. I'm happy to pull those out, if necessary, however.

benjamincburns:
Ah, it looks like I missed some code that was still depending on the holdover client impl

packages/dashboard-message-bus-client/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2022 Truffle Blockchain Group, Inc.
```
eggplantzzz, line 1: Should we change this to ConsenSys?

packages/promise-tracker/LICENSE
```diff
@@ -0,0 +1,19 @@
+Copyright (c) 2022 Truffle Blockchain Group, Inc.
```
eggplantzzz, line 1: Same about the company.

packages/promise-tracker/test/tracker.test.ts
```diff
@@ -0,0 +1,135 @@
...
+      target.throwingTrackedMethod();
+
+      assert.strictEqual(throwCount, 0);
+      await waitForOutstandingPromises({ catchHandler: () => throwCount++ });
+      assert.strictEqual(throwCount, 2);
+    });
+  });
+});
+
+class TrackedObject {
```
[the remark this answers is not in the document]
cds-amal: >...classes must be defined before they can be constructed.

TrackerObject is referenced, not constructed, before the class definition. By the time, its constructor is invoked, TrackerObject has already defined.

My mental model is (open to correction @davidmurdoch!)... Javascript runs to completion, so  it will first process the module, building up references, where the `it` function will have closed over the TrackerObject class definition.  When the `it` function runs, all should be good.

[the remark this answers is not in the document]
davidmurdoch: It's technically fine here... but i think it is super weird to put helpers at the bottom.

packages/promise-tracker/lib/index.ts
```diff
@@ -0,0 +1,104 @@
...
+  if (typeof descriptor.value === "function") {
+    const impl = descriptor.value;
+
+    // wrapper function that keeps a list of outstanding promises for each
+    // instance of DashboardMessageBusClient
+    function newImpl(...args: any[]) {
+      const returnValue = impl.apply(this, args);
+
+      // only track async functions
+      if (returnValue.then && typeof returnValue.then === "function") {
```
davidmurdoch, line 18: Can `returnValue` be `null` or `undefined` here? If so, this will throw.

packages/promise-tracker/lib/index.ts
```diff
@@ -0,0 +1,104 @@
...
+        const trackedTask = returnValue
+          .then((result: any) => {
+            _cleanUpTrackedTask(this, trackedTask);
+            return result;
+          })
+          .catch((err: any) => {
+            _cleanUpTrackedTask(this, trackedTask);
+            throw err;
+          });
```
davidmurdoch, line 27: would `finally` be better?

packages/promise-tracker/lib/index.ts
```diff
@@ -0,0 +1,104 @@
+const _outstandingPromiseInstanceMap = new WeakMap<Object, Set<Promise<any>>>();
+const _outstandingPromiseMap = new Map<Promise<any>, boolean>();
+
+export function tracked<T extends Object>(
+  target: T,
+  key: string | Symbol,
+  descriptor: PropertyDescriptor
+) {
+  if (typeof descriptor.value === "function") {
```
davidmurdoch, line 9: should @tracked be allows on non-functions and non-promises at all?

packages/promise-tracker/lib/index.ts
```diff
@@ -0,0 +1,104 @@
...
+) {
+  const { target } = options || {};
+  let { catchHandler } = options || {};
+
+  const iterable =
+    (target
+      ? _outstandingPromiseInstanceMap.get(target)?.values()
+      : _outstandingPromiseMap.keys()) || [];
+
+  catchHandler = catchHandler || (() => {});
```
davidmurdoch, line 61: Should the default be to swallow errors? Perhaps we should use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AggregateError and throw at the end?

packages/promise-tracker/lib/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  await Promise.all(Array.from(iterable, p => p.catch(catchHandler)));
```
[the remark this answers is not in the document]
davidmurdoch: I think this should be `Promise.allSettled`, as `Promise.all` may resolve as soon as a promise throws, and since `catchHandler` can be supplied by the caller it can throw and cause an early exit.

packages/promise-tracker/test/tracker.test.ts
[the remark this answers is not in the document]
benjamincburns: It runs just fine (build would be red, otherwise).

As a style choice, I tend to put helpers at the bottom to aid in human readability, as when you're reading through that file, the helper class is unlikely to be the thing you're looking for first. IIRC, @gnidan has the same preference, but I could've misinterpreted.

I think it's fair to stick this in the "style choice" bucket, as I'm not getting any `ReferenceError`s, etc, when running the code here. Personally in that case I don't really care much about which particular style we go with here. I just want it to be consistent with the rest of the codebase. With that in mind, if this is enough of an issue to request a change, I'd prefer that we encode and enforce these sort of style rules as prettier or eslint config.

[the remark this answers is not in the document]
davidmurdoch: I don't really care that much (though i'd like to take this moment to point out that imports are basically helpers and so now you've got them at the top _and_ the bottom 😁 :troll: )

packages/promise-tracker/lib/index.ts
benjamincburns: good catch - will add a test for this and fix it

packages/promise-tracker/lib/index.ts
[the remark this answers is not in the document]
benjamincburns: Agreed re: `Promise.allSettled` I might need to shim that for compatibility, however. But I'll look into it.

> What happens if new tracked promises are started after `await waitForOutstandingPromises()` is called?

`waitForOutstandingPromises` is meant to be called immediately prior to terminating the process via `process.exit`. Honestly it's kind of a crap workaround for the fact that we sometimes have some handles left open at the end of a CLI run. Primarily we want to ensure that anything firing off to the message bus has had enough time to succeed, but we don't want to depend on a static delay (e.g. `await delay(1000); process.exit(exitCode);`).

A more ideal choice would be to ensure that there aren't any open handles left dangling when the CLI finishes executing, but this was proving rather difficult to debug.

packages/promise-tracker/lib/index.ts
benjamincburns: No need for it so far. Also at the moment it's strictly a method decorator. I thought I addressed that in the README, but yeah - if people need it to decorate something other than a method, they'll need to implement that.

packages/promise-tracker/lib/index.ts
benjamincburns: Yes, per above comment. The goal is to allow any "best effort" tasks to complete, even if they complete unsuccessfully (e.g. sending events to the dashboard message bus). The message bus might not be running, however. That would cause the client to throw, but nothing needs to be done with that error, as it's not considered exceptional.

packages/promise-tracker/lib/index.ts
benjamincburns: I'll also see if I can't make it throw when decorating something other than a method.

packages/promise-tracker/lib/index.ts
benjamincburns: yes, I think

packages/promise-tracker/lib/index.ts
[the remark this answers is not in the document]
benjamincburns: Actually now that I look at this again I remembered that effectively removed the need for a shim for `Promise.allSettled`. I even checked for this problem w/ [a test](https://github.com/trufflesuite/truffle/pull/4936/files/9f79abcd47364dcb729bccdd51ec5557dd798e25#diff-cf37491629209eac850cb0b88a1914fee649ef2152aebdcd22cd35a17de3e2fbR83-R97).

benjamincburns:
Will push review fixes as individual commits, but I'll squash them prior to merge so the merged commits are composable.

packages/promise-tracker/lib/index.ts
benjamincburns: > I'll also see if I can't make it throw when decorating something other than a method.

Changes that I'll be pushing shortly make it so that I get compiler errors when trying to decorate anything other than a method. I've still added a `throw` check just for good measure, however.

packages/promise-tracker/lib/index.ts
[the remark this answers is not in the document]
benjamincburns: I tried `allSettled` but the existing implementation is effectively equivalent. With the existing implementation I map the catch handler to the individual promises during aggregation.

To preserve the property where `catchHandler` is called once per rejected promise, I'd need to map the `catchHandler` to the returned array of status objects, filtering on those with `result.status === "rejected"`.

The current implementation achieves the same behaviour for the caller in a single line.

packages/promise-tracker/lib/index.ts
benjamincburns: done

packages/promise-tracker/lib/index.ts
benjamincburns: yes, and fixed in next push

packages/promise-tracker/test/tracker.test.ts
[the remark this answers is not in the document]
benjamincburns: I mean, if you squint hard enough, all code is helper code! 😆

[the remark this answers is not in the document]
davidmurdoch: 🤣

tenthirtyone:
Had a really great dialogue with @benjamincburns tonight that gave me a perspective into the PR and changes.

I am suggesting one of the following paths be taken:

1) The work be done to remove the original, underlying tech debt, that required the changes in this PR to move the project forward. <--- This is my preference.

2) Merge this PR, accept the tech debt, and move the dashboard project forward at break neck speed with a plan to address the tech debt or spin up a future version that rolls this into a lessons learned and allows for corrections to take place in that version.

I will support either path forward and back that choice fully and without judgement, but leaving the PR hanging is not a productive option.

Issue #4758, opened 2022-02-21
Truffle Dashboard's "Process" button doesn't provide feedback upon being clicked


gnidan:
## Issue

When you hit "Process" on the Truffle Dashboard, it defers to MetaMask for handling the signature, but the button itself doesn't change text or become disabled. With complex flows (e.g. involving HD wallets), there can be a significant delay, and there's nothing in-page to show that Truffle Dashboard is waiting.

The behavior should likely be to disable the button and change the text to something like "Processing..."

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.5.0
* node version (`node --version`):
* npm version (`npm --version`):

eggplantzzz:
This has been addressed and released with version 5.5.8 https://github.com/trufflesuite/truffle/releases/tag/v5.5.8

PR #4938, opened 2022-03-24
Allow the truffle dashboard to disable the process button after click

dongmingh:
This PR allows the truffle dashboard to disables the "PROCESS" button and to display
"PROCESSING..." after click the "PROCESS" button as reuqested in issue #4758.

The truffle dashboard displays two buttons: PROCESS and REJECT in the
"INCOMING REQUESTS" page. The user can click one of the buttons to process or reject
the request.  But, after click the "PROCESS" button, it may take some time for the
transaction to complete. In particular, when involving HD wallets, the delay could be
significant.  During the delay period, the truffle dashboard shall disable the buttons
and change the text to "PROCESSING...".

### DIFF
--- a/packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
+++ b/packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
@@ -4,6 +4,7 @@ import { handleDashboardProviderRequest, respond } from "../../utils/utils";
 import Button from "../common/Button";
 import Card from "../common/Card";
 import { DashboardProviderMessage } from "@truffle/dashboard-message-bus";
+import React from "react";
 
 interface Props {
   request: DashboardProviderMessage;
@@ -17,6 +18,8 @@ interface Props {
 }
 
 function IncomingRequest({ provider, socket, request, setRequests }: Props) {
+  const [disable, setDisable]=React.useState(false);
+
   const removeFromRequests = () => {
     setRequests(previousRequests =>
       previousRequests.filter(other => other.id !== request.id)
@@ -104,12 +107,26 @@ function IncomingRequest({ provider, socket, request, setRequests }: Props) {
 
   const body = <div>{formatDashboardProviderRequestParameters(request)}</div>;
 
-  const footer = (
-    <div className="flex justify-start items-center gap-2">
-      <Button onClick={process} text="Process" />
-      <Button onClick={reject} text="Reject" />
-    </div>
-  );
+  var footer;
+  if (disable) {
+    footer = (
+      <div className="flex justify-start items-center gap-2">
+        <button
+          className="rounded p-2 text-truffle-brown uppercase hover:bg-white"
+          disabled={disable}
+        >
+          Processing...
+        </button>
+      </div>
+    );
+  } else {
+    footer = (
+      <div className="flex justify-start items-center gap-2">
+        <Button onClick={() => {process(); setDisable(true);}} text="Process" />
+        <Button onClick={reject} text="Reject" />
+      </div>
+    );
+  }
 
   return (
     <div key={request.id} className="flex justify-center items-center">


packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
```diff
@@ -104,12 +107,26 @@ function IncomingRequest({ provider, socket, request, setRequests }: Props) {
...
+      </div>
+    );
+  } else {
+    footer = (
+      <div className="flex justify-start items-center gap-2">
+        <Button onClick={() => {process(); setDisable(true);}} text="Process" />
+        <Button onClick={reject} text="Reject" />
+      </div>
+    );
+  }
```
gnidan, line 129: You can do it like this to avoid mutable variables, which allows the definition of a function to read more sequentially (in contrast to `var` or `let`, which add cognitive complexity to future readers by forcing readers to look out for updates):

```suggestion
  const footer = disable
    ? <div className="flex justify-start items-center gap-2">
          <button
            className="rounded p-2 text-truffle-brown uppercase hover:bg-white"
            disabled={disable}
          >
            Processing...
          </button>
        </div>
    : <div className="flex justify-start items-center gap-2">
          <Button onClick={() => {process(); setDisable(true);}} text="Process" />
          <Button onClick={reject} text="Reject" />
        </div>;
```

This style suggests "there will always be a footer, and that footer differs based on `disabled` as follows." As you can see, it even draws attention to the specific differences between these two alternatives, which I'll add a separate comment for.

packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
```diff
@@ -104,12 +107,26 @@ function IncomingRequest({ provider, socket, request, setRequests }: Props) {
...
+  var footer;
+  if (disable) {
+    footer = (
+      <div className="flex justify-start items-center gap-2">
+        <button
+          className="rounded p-2 text-truffle-brown uppercase hover:bg-white"
+          disabled={disable}
+        >
+          Processing...
+        </button>
```
gnidan, line 119: I noticed that this uses `<button>` whereas the existing code uses `<Button>`. Is there a way to achieve the disable behavior with `<Button>`, and could you please switch to that if so?

gnidan (commented):
Nice, thanks for this! Got some style notes, but glad this is pretty straightforward.

packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
dongmingh: A separate comment will make the code neat, thanks for suggestion.

packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
dongmingh: I will add `disabled` to `<Button>` so that we can use it here. And add `disabled` to all places where `<Button>` is called.

packages/dashboard/src/components/ConfirmNetworkChange.tsx
```diff
@@ -30,7 +30,7 @@ function ConfirmNetworkChanged({
     </div>
   );
 
-  const confirmButton = <Button text="Confirm" onClick={confirm} />;
+  const confirmButton = <Button text="Confirm" disabled={false} onClick={confirm} />;
```
gnidan, line 33: I don't think we should include this if we're just passing false

```suggestion
  const confirmButton = <Button text="Confirm" onClick={confirm} />;
```

packages/dashboard/src/components/ConnectNetwork.tsx
```diff
@@ -17,7 +17,7 @@ function ConnectNetwork({ confirm }: Props) {
     "Please connect your wallet to use the Truffle Dashboard Provider.";
 
   const connectButton = (
-    <Button text="Connect Wallet" onClick={() => activate(injectedConnector)} />
+    <Button text="Connect Wallet" disabled={false} onClick={() => activate(injectedConnector)} />
```
gnidan, line 20: Same as above
```suggestion
    <Button text="Connect Wallet" onClick={() => activate(injectedConnector)} />
```

packages/dashboard/src/components/ConnectNetwork.tsx
```diff
@@ -32,7 +32,7 @@ function ConnectNetwork({ confirm }: Props) {
     </div>
   );
 
-  const confirmButton = <Button text="Confirm" onClick={confirm} />;
+  const confirmButton = <Button text="Confirm" disabled={false} onClick={confirm} />;
```
gnidan, line 35:
```suggestion
  const confirmButton = <Button text="Confirm" onClick={confirm} />;
```

packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
```diff
@@ -104,12 +107,14 @@ function IncomingRequest({ provider, socket, request, setRequests }: Props) {
...
 
-  const footer = (
-    <div className="flex justify-start items-center gap-2">
-      <Button onClick={process} text="Process" />
-      <Button onClick={reject} text="Reject" />
-    </div>
-  );
+  const footer = disable
+    ? <div className="flex justify-start items-center gap-2">
+        <Button disabled={disable} onClick={() => {}} text="Processing..." />
```
gnidan, line 112: For properties that you're setting to true, you can just do it like this:
```suggestion
        <Button disabled onClick={() => {}} text="Processing..." />
```

packages/dashboard/src/components/DashboardProvider/IncomingRequest.tsx
```diff
@@ -104,12 +107,14 @@ function IncomingRequest({ provider, socket, request, setRequests }: Props) {
...
+      </div>
+    : <div className="flex justify-start items-center gap-2">
+        <Button disabled={false} onClick={() => {process(); setDisable(true);}} text="Process" />
+        <Button disabled={false} onClick={reject} text="Reject" />
```
gnidan, line 116:
```suggestion
        <Button onClick={() => {process(); setDisable(true);}} text="Process" />
        <Button onClick={reject} text="Reject" />
```

packages/dashboard/src/components/common/Button.tsx
```diff
@@ -1,14 +1,19 @@
 import type { MouseEventHandler } from "react";
 
 interface Props {
+  disabled: boolean;
```
gnidan, line 4: Maybe this needs to be optional:
```suggestion
  disabled?: boolean;
```

MicaiahReid (approved):
This looks good for showing the initial feedback, but hard coding the initial state of the buttons probably isn't the best bet long-term.

Out of curiosity I ran the dashboard and went to migrate a contract. After clicking "Process" it shows "Processing...". But if you refresh the page while the transaction is still processing, it won't show that a transaction is processing - it will allow you to process or reject that same transaction again. Eventually we'd want to store this state somehow, or get it from truffle. @gnidan should I just make another issue for this?

gnidan (approved):
Looking good! Thanks @dongmingh

gnidan:
> Eventually we'd want to store this state somehow, or get it from truffle. @gnidan should I just make another issue for this?

Definitely worth a separate issue for this :)

PR #4939, opened 2022-03-25
chore: use `@achrinza/node-ipc`

achrinza:
`node-ipc` contains nested transient dependencies that is managed by the same author (e.g. `easy-stack`). This change ensures that all packages maintained by @/riaevangelist in the dependency tree are pinned.

More info about the fork: https://github.com/achrinza/node-ipc/issues/1

Signed-off-by: Rifa Achrinza <25147899+achrinza@users.noreply.github.com>

### DIFF
--- a/packages/environment/chain.js
+++ b/packages/environment/chain.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 require("source-map-support/register");
 
-const IPC = require("node-ipc").IPC;
+const IPC = require("@achrinza/node-ipc").IPC;
 const Ganache = require("ganache");
 const path = require("path");
 const debug = require("debug");
--- a/packages/environment/develop.js
+++ b/packages/environment/develop.js
@@ -1,4 +1,4 @@
-const { IPC } = require("node-ipc");
+const { IPC } = require("@achrinza/node-ipc");
 const path = require("path");
 const { spawn } = require("child_process");
 const debug = require("debug");
--- a/packages/environment/package.json
+++ b/packages/environment/package.json
@@ -20,13 +20,13 @@
     "test": "mocha ./test/** ./test/**/*"
   },
   "dependencies": {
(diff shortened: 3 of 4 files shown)


cds-amal (approved):
Thanks @achrinza. I love that this maintenance fork defined scope and governance! Can we get another opinion on this?

eggplantzzz (dismissed):
Seems cool to me. Thanks @achrinza!

cds-amal:
Hi @achrinza, after a team discussion, we've decided to decline this PR. Perhaps in the future we can revisit that choice.

I support the preventative nature of this maintenance fork and hope the project gets to a point where it is adopted by the community. While the code changes are reviewable, there are several changes to the `package-lock.json` that are not quite as easy to evaluate and sign off. The numerous unverified commits are also a concern, for a maintenance fork to address a supply-chain concern.

achrinza:
No worries, I hope the team can revisit it in the future.

To address some concerns for future consideration:

> there are several changes to the package-lock.json that are not quite as easy to evaluate and sign off.

`package-lock.json` is only used for `@achrinza/node-ipc`'s CI pipeline. This does not get published in the NPM registry and [is ignored when installed as a dependency](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#package-lockjson-vs-npm-shrinkwrapjson); Only truffle's own `yarn.lock` is considered when resolving the dependency tree.

However if its still a concern, we do use [lockfile-lint](https://github.com/achrinza/node-ipc/blob/3a909b05c20980cfc4f0ab24112db70eb755cb76/.github/workflows/ci.yaml#L45-L54) to verify our `package-lock.json`.

> The numerous unverified commits are also a concern, for a maintenance fork to address a supply-chain concern.

This is unfortunately not fixable without rewriting the Git History, which would break  existing provenance. To some, the signed Git Tags may be suffice in acting as the final sign-off on a release.

PR #4941, opened 2022-03-25
Add dialogue to Truffle Dashboard to encourage friendly opt-in for better product plz

fainashalts:
This PR adds some frontend components for asking users whether they are willing to enable Truffle's analytics. Some screenshots below (please note the language used, happy to reword it!): 
Screen Shot 2022-03-25 at 9 42 26 AM
Screen Shot 2022-03-25 at 9 42 34 AM
Screen Shot 2022-03-25 at 9 42 41 AM

### DIFF
--- a/packages/dashboard/src/components/common/AnalyticsToggle.tsx
+++ b/packages/dashboard/src/components/common/AnalyticsToggle.tsx
@@ -12,7 +12,7 @@ export default function AnalyticsToggle() {
           checked={enabled}
           onChange={setEnabled}
           className={`${enabled ? 'bg-truffle-blue' : 'bg-truffle-brown'
-            } relative inline-flex items-center h-6 rounded-full w-11 transition-colors focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-truffle-blue`}
+            } sr-only>{{"This switch enables anonymous analytics tracking"}} relative inline-flex items-center h-6 rounded-full w-11 transition-colors focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-truffle-blue`}
         >
           <span
             className={`${enabled ? 'translate-x-6' : 'translate-x-1'
@@ -24,3 +24,33 @@ export default function AnalyticsToggle() {
     </Switch.Group>
   );
 }
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
--- a/packages/dashboard/src/components/common/AnalyticsToggle.tsx
+++ b/packages/dashboard/src/components/common/AnalyticsToggle.tsx
@@ -1,0 +1,26 @@
+import { Switch } from '@headlessui/react';
+import { useState } from 'react';
+
+export default function AnalyticsToggle() {
+  const [enabled, setEnabled] = useState(false);
+
+  return (
+    <Switch.Group>
+      <div className="flex justify-center mt-4">
+        <span className="pr-3 float-left">Disabled</span>
+        <Switch
+          checked={enabled}
+          onChange={setEnabled}
+          className={`${enabled ? 'bg-truffle-blue' : 'bg-truffle-brown'
+            } relative inline-flex items-center h-6 rounded-full w-11 transition-colors focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-truffle-blue`}
+        >
+          <span
+            className={`${enabled ? 'translate-x-6' : 'translate-x-1'
+              } inline-block w-4 h-4 transform bg-white rounded-full transition-transform`}
+          />
+        </Switch>
+        <span className="pl-3 float-right">Enabled</span>
+      </div>
+    </Switch.Group>
+  );
+}
--- a/packages/dashboard/src/components/common/AnalyticsToggle.tsx
+++ b/packages/dashboard/src/components/common/AnalyticsToggle.tsx
@@ -7,7 +7,7 @@ export default function AnalyticsToggle() {
   return (
     <Switch.Group>
       <div className="flex justify-center mt-4">
-        <span className="pr-3 float-left">Disabled</span>
+        <span className="pr-3 float-left">No</span>
         <Switch
           checked={enabled}
           onChange={setEnabled}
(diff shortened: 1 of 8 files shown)


tenthirtyone:
Lookin good with the styles. 

When the page height is > 100% of the view height the icon does not scroll with the page (Not sure if this will ever happen in its current form)

image

tenthirtyone:
This css to the parent container will get the sticky effect with page scroll (bot/right pixels can be anything)

image

packages/dashboard/src/components/common/Popover.tsx
```diff
@@ -0,0 +1,22 @@
...
+export default function SettingsPopover() {
+  return (
+    <Popover>
+      <Popover.Button className="absolute bottom-2 right-2"> <SettingsIcon analyticsSet={false} /></Popover.Button >
+      <Popover.Overlay className="bg-black opacity-30 fixed inset-0" />
+      <Popover.Panel className="absolute z-10 bottom-14 right-2 mb-2">
+        <div className="bg-truffle-light border-2 mb-4 block z-50 font-normal leading-normal text-lg max-w-xs no-underline break-words rounded-lg">
+          <div className="bg-truffle-blue border-2 text-center text-truffle-brown font-semibold p-3 mb-2 uppercase rounded-t-lg">Help Improve Truffle!</div>
+          <div className="bg-truffle-light text-truffle-brown m-2">Truffle uses anonymous opt-in analytics to determine how we could make our
+            product better. Are you willing to opt-in?
```
davidmurdoch, line 14: The question "Are you willing to opt-in?" cannot be answered by the words "Enabled" or "Disabled".

Also, if the options were "Yes" or "No" to this question a "Yes" answer would not actually give us permission to track them. It only lets us know that if we did ask for permission they would be willing to opt in.

I propose something like:

> You can help us improve Truffle by allowing us to track how you use our tools. For details on what information we collect and how we use it visit our [tracking policy].
> Would you like to allow tracking?
> [✔️] Allow Tracking

packages/dashboard/src/components/common/AnalyticsToggle.tsx
```diff
@@ -0,0 +1,26 @@
...
+    <Switch.Group>
+      <div className="flex justify-center mt-4">
+        <span className="pr-3 float-left">Disabled</span>
+        <Switch
+          checked={enabled}
+          onChange={setEnabled}
+          className={`${enabled ? 'bg-truffle-blue' : 'bg-truffle-brown'
+            } relative inline-flex items-center h-6 rounded-full w-11 transition-colors focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-truffle-blue`}
+        >
+          <span
```
davidmurdoch, line 17: Accessibility audit:

This Switch element should contain a ` <span className="sr-only">{{ << PUT WHAT ENABLING THIS ENABLES HERE >> }}</span>` element. It might contain something like "Enable tracking". The value here may be used by screen readers (`sr-only` cl;ass here means "screen readers only").

packages/dashboard/src/components/common/SettingsIcon.tsx
```diff
@@ -0,0 +1,21 @@
+import { MdSettings } from "react-icons/md";
+// import PopoverRender from "./SettingsPopover";
```
davidmurdoch, line 2: Remove?

```suggestion
```

packages/dashboard/src/components/common/SettingsIcon.tsx
```diff
@@ -0,0 +1,21 @@
...
+interface Props {
+  analyticsSet: boolean;
+  // onClick: typeof PopoverRender;
```
davidmurdoch, line 5: Remove?

```suggestion
```

fainashalts:
> Lookin good with the styles.
> 
> When the page height is > 100% of the view height the icon does not scroll with the page (Not sure if this will ever happen in its current form)
> 
> image


Added the sticky and also changed the color. Thanks!

packages/dashboard/src/components/common/Popover.tsx
fainashalts: Do we have a tracking policy link? I wasn't aware of one. We had discussed the format and had decided on a slider. I see what you mean about disabled/enabled and will change it to yes/no. I like your rewording, only open question being whether we actually have a tracking policy!

packages/dashboard/src/components/common/AnalyticsToggle.tsx
fainashalts: Ah! You know, I read up on accessibility with the popover component, and completely missed covering the switch element. This makes sense. Added -- is this what you meant? I checked in the developer tools but don't actually know how to make sure an screen reader will catch this!

packages/dashboard/src/components/common/Popover.tsx
fainashalts: Oh also, yes there is a conditional for the red dot, hardcoded to include it for now. Liang is taking this over to add the backend component at which point the conditional will actually be hooked into something!

fainashalts: I reworded this but didn't add anything about the tracking policy because I do not believe we have one yet. Definitely think we should add a link when we can though.

packages/dashboard/src/components/common/AnalyticsToggle.tsx
davidmurdoch: I'm just starting to get into the accessibility stuff myself, and I'm sure there is so much else we are doing wrong all over the place. I haven't tried it out on my mac yet, but I think you can enable "VoiceOver" in your system settings and try it out by navigating with the Tab key, or by hovering with the mouse.

> This switch enables anonymous analytics tracking

Something else I just thought of: the toggle doesn't only enable tracking, it also disables it. Something like "Enable anonymized tracking" and "Disable anonymized tracking" is probably enough. ¯\_(ツ)_/¯ Anything is better than nothing, of course!

fainashalts:
Analytics policy PR: https://github.com/trufflesuite/trufflesuite.com/pull/1209

Once merged, we can add a link to it in this dialog box.

davidmurdoch (commented):
Looks like some space characters for removed in the latest commit.

davidmurdoch (approved):

PR #4942, opened 2022-03-25
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### DIFF
--- a/packages/dashboard/src/utils/utils.ts
+++ b/packages/dashboard/src/utils/utils.ts
@@ -125,5 +125,5 @@ export const reverseLookup = async (
 };
 
 export const shortenAddress = (address: string) => {
-  return `${address.substr(0, 6)}...${address.substr(address.length - 4, 4)}`;
+  return `${address.slice(0, 6)}...${address.slice(-4)}`;
 };
--- a/packages/events/Subscriber/helpers.js
+++ b/packages/events/Subscriber/helpers.js
@@ -54,8 +54,8 @@ const convertHandlerNameToRegex = name => {
     start += match.index + matchLength;
     str += cleanString + starRegex;
   }
-  str += name.substr(start).replace(reRegExpChar, "\\$&");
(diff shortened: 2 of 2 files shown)


haltman-at (approved):
Thanks for this!  (Now can we get a second reviewer on this?)

cds-amal (approved):
LGTM

packages/dashboard/src/utils/utils.ts
```diff
@@ -125,5 +125,5 @@ export const reverseLookup = async (
 };
 
 export const shortenAddress = (address: string) => {
-  return `${address.substr(0, 6)}...${address.substr(address.length - 4, 4)}`;
```
eggplantzzz, line 128: Looks like you will need to adjust the 2nd argument for the `slice` method here. In `substr` the 2nd argument is the length of the string you want to extract. In slice it is the index to extract up to.

packages/dashboard/src/utils/utils.ts
```diff
@@ -125,5 +125,5 @@ export const reverseLookup = async (
...
+  return `${address.slice(0, 6)}...${address.slice(-4)}`;
```
eggplantzzz, line 128: Looks like you will need to adjust the 2nd argument for the `slice` method here. In `substr` the 2nd argument is the length of the string you want to extract. In slice it is the index to extract up to.

eggplantzzz (approved):

eggplantzzz:
Thanks @CommanderRoot!

Issue #1005, opened 2018-06-13
TypeError: Cannot read property 'match' of undefined


gunvant:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

MyService.deployed().then(function(instance) { meta = instance;  meta.allocate.sendTransacti
on("0xf17f52151ebef6c7334fad080c5704d77216b732",20,{from: "0x627306090abab3a6e1400e9345bc60c78a8bef57"});})

throws below error

TypeError: Cannot read property 'match' of undefined
    at parse (C:\Users\jambug\AppData\Roaming\npm\node_modules\truffle\build\webpack:\~\yargs\~\yargs-parser\index.js:125:
1)
    at Function.Parser.detailed (C:\Users\jambug\AppData\Roaming\npm\node_modules\truffle\build\webpack:\~\yargs\~\yargs-p
arser\index.js:761:1)
    at Object.Yargs.self._parseArgs (C:\Users\jambug\AppData\Roaming\npm\node_modules\truffle\build\webpack:\~\yargs\yargs
.js:938:1)
    at Object.Yargs.self.parse (C:\Users\jambug\AppData\Roaming\npm\node_modules\truffle\build\webpack:\~\yargs\yargs.js:5
32:1)

But, the same is executed successfully from test.js file

* Operating System: Windows
* Ethereum client:
* Truffle version (`truffle version`): Truffle v4.1.11 (core: 4.1.11)
* node version (`node --version`): v7.4.0
* npm version (`npm --version`): 5.5.1

cgewecke:
@gunvant I wasn't able to reproduce this on my own and your example is specific to your project. It's hard to tell what's causing the error. 

If you have a chance could you provide reproduction steps - e.g. a simple example of a call sequence that succeeds in a test but triggers this error in the console?

Could you also specify whether you're running `truffle console` or `truffle develop`?

vendor77777:
Das liegt wahrscheinlich das ich Clone Repo habe Datei Crash wie korrigiere
ich das Beispiel auf der haubt Seite habe ich 2 mal web3 angezeigt einmal
mit mein Profil und Original aber wenn ich drauf gehe zeigt mir eine und
die selbe Seite gipst da ein Plan?

Am 17.06.2018 04:22 schrieb "c-g-e-w-e-k-e->" <notifications@github.com>:

[...]

Issue #399, opened 2017-05-09
Parsing issue with multiple spaces


illuzen:
- [ ] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

Not sure when this was introduced, but in the yargs-parser node module, we now break when double spaces are introduced between certain tokens, so 

`MultiSigWallet.new(web3.eth.accounts, 3 ).then(function(i){wallet = i}) 
`
is fine, but

`MultiSigWallet.new(web3.eth.accounts, 3 ).then(function(i){wallet  = i})
`
throws 

> TypeError: Cannot read property 'match' of undefined
>     at parse (/usr/local/lib/node_modules/truffle/node_modules/yargs-parser/index.js:122:12)
>     at Function.Parser.detailed (/usr/local/lib/node_modules/truffle/node_modules/yargs-parser/index.js:756:10)
>     at parseArgs (/usr/local/lib/node_modules/truffle/node_modules/yargs/yargs.js:873:27)
>     at Object.Yargs.self.parse (/usr/local/lib/node_modules/truffle/node_modules/yargs/yargs.js:499:18)
>     at Command.getCommand (/usr/local/lib/node_modules/truffle/lib/command.js:19:24)
>     at TruffleInterpreter.interpret (/usr/local/lib/node_modules/truffle/lib/repl.js:104:20)
>     at bound (domain.js:280:14)
>     at REPLServer.runBound [as eval] (domain.js:293:12)
>     at REPLServer.onLine (repl.js:536:10)
>     at emitOne (events.js:96:13)
>     at REPLServer.emit (events.js:191:7)
>     at REPLServer.Interface._onLine (readline.js:238:10)
>     at REPLServer.Interface._line (readline.js:581:8)
>     at REPLServer.Interface._ttyWrite (readline.js:860:14)
>     at REPLServer.self._ttyWrite (repl.js:609:7)
>     at ReadStream.onkeypress (readline.js:117:10)
> 

Printing out args from yargs-parser.parse shows it is tokenizing funny

```
[ 'MultiSigWallet.new(web3.eth.accounts,',
  '3',
  ').then(function(i){wallet',
  ,                     // <--------- empty string is token?
  '=',
  'i})' ]


```
sorry I don't have time right now to debug further, it's not critical, just spent half an hour chasing it down. It might be an issue in yargs-parser?

##  to Reproduce

truffle console

[...]

PR #4944, opened 2022-03-25
tokenize with greedy match instead of single space

percentcer:
fix for https://github.com/trufflesuite/truffle/issues/1005

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -17,7 +17,7 @@ const Require = require("@truffle/require");
 const debug = require("debug")("console");
 
 const processInput = input => {
-  const inputComponents = input.trim().split(" ");
+  const inputComponents = input.trim().split(/\s+/);
   if (inputComponents.length === 0) return input;
 
   if (inputComponents[0] === "truffle") {


percentcer:
pinging @cgewecke as requested in the discussion thread

haltman-at (commented):
While the change makes sense, I don't understand why this would cause the issue you claim it does.  It looks to me like the existing function should work fine; if there's a problem with multiple spaces causing crashes, I'm guessing the offending code is elsewhere.  Can you demonstrate that this is actually the problem?  The comment you left on the issue shows the difference in tokenization, but not that this leads to the error; I'm skeptical that it does lead to that error.

haltman-at:
OK @percentcer I've looked into this some more and while the issue does exist (I've reopened it, thanks for pointing this out to us), this PR doesn't fix it; it targets the wrong thing.  The error is thrown by `getCommand`, not `processInput`; `processInput` successfully returns a string.  When I try making the change here on my local copy, the bug still occurs.

haltman-at:
OK I guess the problem is implicitly in `processInput` after all (since we can't really affect `getCommand`), but the thing is that the stripping of extra spaces has to happen *all* the time, not only when it starts with `truffle`.  I'll put up a review showing you what I mean.

haltman-at:
Ugh, it won't let me do it as a review suggestion because the necessary change is too far away from the line you changed.  Blech.

Basically, this PR will work if, in addition to the change you already made, you also change line 26 to
```javascript
  return inputComponents.join(" ");
```

It might also be worth adding a comment noting why you're stripping out extra spaces (namely, because `yargs` has a problem with them).

haltman-at (changes requested):
See my comments above; this PR doesn't quite fix the bug, but with just a tiny bit more work it will.

haltman-at:
Oh, wait, crap, my solution is no good either -- it wrecks string literals!  Ugh.  Really, so does your original PR, but I imagine it's less likely to come up in the specific context where the line starts with `truffle`...

OK, I guess this is just going to require more investigation. :-/  Will likely have to actually look at just how `getCommand` works...

eggplantzzz:
This issue is probably related https://github.com/trufflesuite/truffle/issues/399?

haltman-at:
Yes, that appears to be the same issue.  It looks like #399 and #1005 are duplicates of one another.

haltman-at:
Oh, looks like `getCommand` is us, not yargs.  So yay, we can presumably fix it after all!  It lives [here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/command.js).  I'll take a more detailed look.

haltman-at:
Oh, silly me, though -- the problem comes from a call to `yargs.parse()`.  OK, going to have to look at that, then...

haltman-at:
Oh!  It turns out updating `yargs` to the latest version solves the problem.  Going to open a PR for that.  Hope you don't mind!

haltman-at:
OK, I've put up #5035 to solve the problem by updating `yargs`.  Note we can't update all the way to the latest version, as that changes that break our tests; but 13.3.0 solves the problem without breaking tests.  So, hopefully that's all OK and we can get that merged!

haltman-at:
Closing as superseded but thanks a bunch @percentcer for putting the spotlight on this!

Issue #4835, opened 2022-03-10
Opcode listings should be in hexadecimal


wbl:
Jump targets are identified by hex numbers, but the listing counts lines in decimal. This makes it tricky to see where the jump target is: you have to convert.

haltman-at:
Sorry, what's the context for this?  There's a lot of information missing from this issue.  Is this about `truffle opcode`?  Or about the `p` display in `truffle debug`?  Or something else I'm not thinking of?

Pobepto:
Hi @haltman-at, I want to help with this issue, but I don't quite understand the context. Can you explain a little more about what needs to be done?

haltman-at:
Actually, this issue contains a misconception -- currently, `truffle opcode` doesn't print the instruction PC values, it prints the instruction *indices*!  Which is different, and much less useful... it should presumably print the PC and in hex, as @wbl says.

haltman-at:
Sure, @Pobepto.  This issue pertains to the `truffle opcode` command, which acts as a disassembler.  Currently, it prints out each instruction prefixed with a number.  You might think (as @wbl did) that it's the PC value in decimal.  But actually, it's the instruction index in decimal.  (These can be different because `PUSH` instructions take up multiple bytes.  So if e.g. a contract starts with `PUSH1 0x80 POP`, then the `POP` instruction would have an index of 1 (because there's only one previous instruction) but a PC value of 2 (because there are two previous bytes).)

The instruction index isn't very useful; the PC value would be more useful, and it would be even *more* useful if, as @wbl requested, it were in hex, since that would allow one to see at a glance where jumps go.  So, to resolve this issue would mean changing that printout.

There is the question of if the old printout should be preserved in some form for those who want that (any opinion on this, @gnidan?) but I don't really see any value in the old way (and I believe `truffle opcode` is little-used anyway) so it's probably OK to just change it, is my guess?

Hope that explains things!

gnidan:
> There is the question of if the old printout should be preserved in some form for those who want that (any opinion on this, @gnidan?) but I don't really see any value in the old way (and I believe truffle opcode is little-used anyway) so it's probably OK to just change it, is my guess?

I concur with this reasoning. If we can keep the existing info and still have the output look nice, then that's probably ideal, but I don't think it's a problem if we just get rid of the decimal instruction index info.

Pobepto:
[...]

PR #4946, opened 2022-03-26
Show program counter instead of instruction index in `truffle opcode`

Pobepto:
## Task
The problem was that the `truffle opcode` gave instruction indexes instead of PC values. Issue: #4835.

Works done:
- [x] Corrected output for `truffle opcode`.

## Environment
- Truffle version (`truffle version`): 5.5.6

### DIFF
--- a/packages/core/lib/commands/opcode/run.js
+++ b/packages/core/lib/commands/opcode/run.js
@@ -31,7 +31,10 @@ module.exports = async function (options) {
   const opcodes = CodeUtils.parseCode(bytecode, numInstructions);
 
   if (opcodes.length === 0) {
-    throw new TruffleError("List of instructions is empty!");
+    console.log(
+      "Contract has no bytecode. Please check to make sure it's not an `abstract contract` or an `interface`."
+    );
+    return;
   }
 
   const lastPCByteLength = Conversion.toBytes(
--- a/packages/core/lib/commands/opcode/run.js
+++ b/packages/core/lib/commands/opcode/run.js
@@ -30,15 +30,15 @@ module.exports = async function (options) {
   }
   const opcodes = CodeUtils.parseCode(bytecode, numInstructions);
 
-  const indexLength = (opcodes.length + "").length;
+  const lastPCByteLength = Conversion.toBytes(
+    opcodes[opcodes.length - 1].pc
+  ).byteLength;
 
   opcodes.forEach(opcode => {
-    let strIndex = Conversion.toHexString(opcode.pc) + ":";
-
-    while (strIndex.length < indexLength + 1) {
-      strIndex += " ";
-    }
-
-    console.log(strIndex + " " + opcode.name + " " + (opcode.pushData || ""));
+    console.log(
+      Conversion.toHexString(opcode.pc, lastPCByteLength) + ":",
+      opcode.name,
+      opcode.pushData || ""
+    );
   });
 };
--- a/packages/core/lib/commands/opcode/run.js
+++ b/packages/core/lib/commands/opcode/run.js
@@ -30,6 +30,10 @@ module.exports = async function (options) {
   }
   const opcodes = CodeUtils.parseCode(bytecode, numInstructions);
 
+  if (opcodes.length === 0) {
+    throw new TruffleError("List of instructions is empty!");
+  }
+
   const lastPCByteLength = Conversion.toBytes(
     opcodes[opcodes.length - 1].pc
   ).byteLength;
--- a/packages/core/lib/commands/opcode/run.js
+++ b/packages/core/lib/commands/opcode/run.js
@@ -3,6 +3,8 @@ module.exports = async function (options) {
   const TruffleError = require("@truffle/error");
   const WorkflowCompile = require("@truffle/workflow-compile");
   const CodeUtils = require("@truffle/code-utils");
+  const Codec = require("@truffle/codec");
+  const Conversion = Codec.Conversion;
 
   if (options._.length === 0) {
     throw new TruffleError("Please specify a contract name.");
@@ -31,8 +33,8 @@ module.exports = async function (options) {
 
   const indexLength = (opcodes.length + "").length;
 
-  opcodes.forEach((opcode, index) => {
-    let strIndex = index + ":";
+  opcodes.forEach(opcode => {
+    let strIndex = Conversion.toHexString(opcode.pc) + ":";
 
     while (strIndex.length < indexLength + 1) {
       strIndex += " ";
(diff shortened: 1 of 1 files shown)


packages/core/lib/commands/opcode/run.js
```diff
@@ -3,6 +3,8 @@ module.exports = async function (options) {
   const TruffleError = require("@truffle/error");
   const WorkflowCompile = require("@truffle/workflow-compile");
   const CodeUtils = require("@truffle/code-utils");
+  const Codec = require("@truffle/codec");
+  const Conversion = Codec.Conversion;
```
lsqproduction, line 7: Thank you for the PR, @Pobepto. This looks great!
Would you like to achieve this by combining the two lines and destructing?
```suggestion
  const { Conversion } = require("@truffle/codec");
```

packages/core/lib/commands/opcode/run.js
```diff
@@ -31,8 +33,8 @@ module.exports = async function (options) {
 
   const indexLength = (opcodes.length + "").length;
 
-  opcodes.forEach((opcode, index) => {
-    let strIndex = index + ":";
+  opcodes.forEach(opcode => {
+    let strIndex = Conversion.toHexString(opcode.pc) + ":";
 
     while (strIndex.length < indexLength + 1) {
```
haltman-at, line 39: The thing I think most needs changing here is this loop for padding the displayed value.  It left-pads with spaces to get things to a length of `indexLength + 1`.  (I guess it was written back before `str.padStart()` existed.)  But, I think that's not really what we want, for two reasons:

1. (The more important reason) The variable `indexLength` is based on the length of the largest *index* in decimal; but here we're no longer going based on indices but rather based on PC values.  So the length to pad to should be based on the largest PC value, not the largest index.

2. (The less important reason)  I think the following looks a bit awkward:

```
  0xff: POP
0x0100: POP
```

Right?  Instead, it should probably be:

```
0x00ff: POP
0x0100: POP
```

So, instead of left-padding with spaces to the left of the `0x`, it would make more sense to left-pad with 0s to the right of the `0x`.

Of course the good news is that, since you're already using `Codec.Conversion.toHexString`, you can use its functionality for that; it allows a second argument which is a minimum length to use.  Note however that this length is given in bytes, rather than hex digits, so make sure you account for that if you want to do it that way.  (Or you could always do it manually, obviously, but that seems more awkward to me...)

haltman-at (commented):
Thanks @Pobepto!  I have two comments on this, one relating to a variable naming issue and the other relating to how the printout is padded.

packages/core/lib/commands/opcode/run.js
Pobepto: Yeah, sure, I will change it!

packages/core/lib/commands/opcode/run.js
```diff
@@ -31,8 +33,8 @@ module.exports = async function (options) {
 
   const indexLength = (opcodes.length + "").length;
 
-  opcodes.forEach((opcode, index) => {
-    let strIndex = index + ":";
+  opcodes.forEach(opcode => {
+    let strIndex = Conversion.toHexString(opcode.pc) + ":";
```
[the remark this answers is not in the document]
Pobepto: You're right, I will fix it

packages/core/lib/commands/opcode/run.js
Pobepto: Done! Maybe the naming is a little strange, if there is a better option I will change 🙂

The idea is that we look at the last PC, find out its length in bytes and make padding based on it.

haltman-at (dismissed):
This looks good to me.  Thanks @Pobepto! (Now can we get another reviewer on this? :) )

packages/core/lib/commands/opcode/run.js
```diff
@@ -29,15 +30,15 @@ module.exports = async function (options) {
   }
   const opcodes = CodeUtils.parseCode(bytecode, numInstructions);
 
-  const indexLength = (opcodes.length + "").length;
+  const lastPCByteLength = Conversion.toBytes(
+    opcodes[opcodes.length - 1].pc
```
haltman-at, line 34: Oh wait!  Sorry, one more thing!  I just realized this might cause problems if `opcodes` is an empty array.  Could you throw in something to handle that case?  Thank you!

haltman-at:
(Note: I haven't dismissed my approve despite noticing a problem because I'm going to be gone for the next few days and don't want to hold things up.  Just assume my approval is valid once the problem I noticed is fixed! :)  Meanwhile we still need a second reviewer on this.)

packages/core/lib/commands/opcode/run.js
Pobepto: Oh right, I'll fix that

packages/core/lib/commands/opcode/run.js
```diff
@@ -29,15 +30,19 @@ module.exports = async function (options) {
   }
   const opcodes = CodeUtils.parseCode(bytecode, numInstructions);
 
-  const indexLength = (opcodes.length + "").length;
-
-  opcodes.forEach((opcode, index) => {
-    let strIndex = index + ":";
+  if (opcodes.length === 0) {
+    throw new TruffleError("List of instructions is empty!");
```
haltman-at, line 34: Oh, hey, popping in here to say that this shouldn't error.  Running this on empty bytecode isn't an error.  (For isnstance, a contract may have empty bytecode if it's an interface or abstract.)  Truffle shouldn't error just because the user did something unusual.  You can have some sort of printout about empty bytecode and what this might indicate, but there's no reason for it to be an error.

Pobepto: Would this solution work?

```suggestion
    console.log("List of instructions is empty!");
    return;
```

Pobepto: @cds-amal, looks good to me! If everyone else agrees, I'll correct it like this

gnidan: How about this:
```
Contract has no bytecode. Please check to make sure it's not `abstract` or an `interface`
```

cds-amal: I like @gnidan 's suggestion

sukanyaparashar: @gnidan 's suggestion looks good to me too.

cds-amal (approved):
LGTM!

packages/core/lib/commands/opcode/run.js
eggplantzzz: Do we need to tell them to check? I kind of like @cds-amal's as it just informs them. Although it seems like we should say, "No opcodes..." because they are requesting opcodes, right?

gnidan: Yeah, but if they're requesting opcodes for something that has no opcodes, it seems prudent to let them know about the disconnect there.

haltman-at (approved):
Giving this the second approval now that I'm back. :)  Thanks a bunch @Pobepto!

PR #4950, opened 2022-03-28
spinners: fix remove logic

benjamincburns:
Works around a bug in the `spinnies` module (jcarpanelli/spinnies#34) where removal of the final spinner may cause the spinner text to not disappear and/or may cause the process to hang on termination due to an interval not being cleared.

### DIFF
--- a/packages/events/defaultSubscribers/migrate/Messages.js
+++ b/packages/events/defaultSubscribers/migrate/Messages.js
@@ -220,7 +220,7 @@ class Messages {
       deployed: () => {
         let stopText;
         if (reporter.blockSpinner) {
-          reporter.blockSpinner.stop();
+          reporter.blockSpinner.remove();
           stopText = `   > ${reporter.currentBlockWait}`;
         }
 
@@ -277,7 +277,7 @@ class Messages {
       // Transactions
       endTransaction: () => {
         if (reporter.blockSpinner) {
-          reporter.blockSpinner.stop();
+          reporter.blockSpinner.remove();
         }
         return `   > ${data.message}`;
       },
--- a/packages/spinners/lib/spinner.ts
+++ b/packages/spinners/lib/spinner.ts
@@ -45,6 +45,7 @@ export class Spinner {
       return;
     }
     spinnies.remove(this.name);
+    spinnies.checkIfActiveSpinners();
   }
 
   /**
--- a/packages/spinners/lib/spinner.ts
+++ b/packages/spinners/lib/spinner.ts
@@ -5,9 +5,9 @@ const spinnies = new Spinnies();
 export class Spinner {
   name: string;
 
-  constructor(name: string, options: SpinnerOptions);
+  constructor(name: string, options: Partial<SpinnerOptions>);
   constructor(name: string, text: string);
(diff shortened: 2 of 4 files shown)


packages/events/defaultSubscribers/migrate/Messages.js
```diff
@@ -220,7 +220,7 @@ class Messages {
       deployed: () => {
         let stopText;
         if (reporter.blockSpinner) {
-          reporter.blockSpinner.stop();
+          reporter.blockSpinner.remove();
```
cds-amal, line 223: Does `blockSpinner.remove()` have a meaningful implication to `stopText`? It also seems weird that, in the pre-commit `stop()` is invoked before setting `stopText`

cds-amal (changes requested):
> Works around a bug in the spinnies module where removal of the final spinner may cause the spinner text to not disappear and/or may cause the process to hang on termination due to an interval not being cleared.

That's quite the premise! This should probably link to the bug/issue described.

packages/spinners/lib/spinner.ts
```diff
@@ -45,6 +45,7 @@ export class Spinner {
       return;
     }
     spinnies.remove(this.name);
+    // spinnies.checkIfActiveSpinners();
```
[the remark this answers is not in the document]
benjamincburns: Oof, I accidentally committed it like that while doing some manual testing to ensure that the change did in fact work around the issue.

packages/events/defaultSubscribers/migrate/Messages.js
benjamincburns: `stopText` is unrelated to the spinner - I didn't write that code, but it appears to be used for logging output.

cds-amal: oh! I see. Thanks!

packages/spinners/typings/spinnies/index.d.ts
```diff
@@ -131,5 +131,10 @@ declare module "spinnies" {
      * @returns false if all spinners have succeeded, failed or have been stopped
      */
     hasActiveSpinners: () => boolean;
+
+    /**
+     * Updates the spinning interval if no spinners are left active
+     */
+    checkIfActiveSpinners: () => void;
```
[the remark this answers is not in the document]
benjamincburns: Per other comment, this was necessary, as the invocation shouldn't have been commented out

benjamincburns:
> This should probably link to the bug/issue described.

No such issue exists, and I didn't bother to create one as `spinnies` appears to be unmaintained. I'll go ahead and create one now, however.

I did reach out to the repo owner and offer to volunteer as a maintainer however, as it's a project that gets 100k+ weekly downloads and appears to see very few bug reports and/or pull requests.

cds-amal:
> ... Issue created: [jcarpanelli/spinnies#34](https://github.com/jcarpanelli/spinnies/issues/34)
> 
> I did reach out to the repo owner and offer to volunteer as a maintainer however, as it's a project that gets 100k+ weekly downloads and appears to see very few bug reports and/or pull requests.

It's a better experience to ora, imo. If the author abandoned it, perhaps we can switch to a maintained fork, or fork our own.

PR #4953, opened 2022-03-28
Add custom inspect alias to codec inspectors, to support browser-util-inspect

gnidan:
[browser-util-inspect](https://www.npmjs.com/package/browser-util-inspect) is pretty cool because it provides the Node.js `util.inspect` functionality in the browser! It even supports custom inspect functions...

... except instead of looking for the symbol `util.inspect.custom`, it looks for a method called `inspect`.

This PR aliases the existing symbol-defined method with an additional `inspect` interface for use by this package in the browser.


haltman-at (approved):
Makes sense!

gnidan:
Thanks!

PR #4954, opened 2022-03-28
Remove return in the compile run method

eggplantzzz:


haltman-at:
Sorry, what's the reason for removing this?  Is it that if there's a return value it should be an exit code, but this isn't an exit code?  or what?

eggplantzzz:
Yes, currently if anything is returned by the commands' run methods, `cli.js` treats it as an exit status. This command returns some kind of compilation result (I don't remember what it is offhand but it is definitely not an exit status) which is incorrect.

cds-amal (approved):
Please put the context/reason up in the PR description. The change looks good.

Issue #4133, opened 2021-06-23
Add support for detailed decoding of multicalls in decoder


haltman-at:
## Issue

@gnidan has suggested that when decoding a multicall transaction, we should not just leave the individual bytestrings undecoded, but should also decode them and return that information.  We don't want to put that information on the `Value` object itself, so @gnidan has suggested this could go in an `interpretations` field (or something) on the overall transaction decoding.

We also need a new interface for decoding the return values for these.  @gnidan has suggested this could be a new interface that takes, instead of an ABI entry and a return value, a transaction object and a return value.

Note that this would be handled purely in `decoder`, this wouldn't involve adding anything new to codec.

## Environment

* Truffle version (`truffle version`): 5.3.11

kevinbluer:
Available in [v5.6.18](https://github.com/trufflesuite/truffle/releases/tag/v5.5.18) 👍

PR #4955, opened 2022-03-29
Enhancement: Add decoding of multicalls to `@truffle/decoder`

haltman-at:
Addresses #4133.  This PR adds `interpretations` to decodings returned by `@truffle/decoder`, with one interpretation currently existing: multicall.

This is mostly pretty straightforward -- the decoder, after decoding a transaction, will check whether it's a multicall (as in: name `multicall`, arguments a single `bytes[]`), and, if it is, will then recurse over the subcalls and decode them, sticking the result in `interpretations.multicall`.  There's not much to say about the logic here!  I also added a test of this functionality.

I also updated `CalldataDecodingInspector` so that it will check for `interpretations.multicall`, and, if it finds it present, print out the the calls made rather than the raw bytestrings for each entry.  In the process I found that the existing logic for `formatFunctionLike` wasn't really working like I wanted so I fixed that up a bit too.

Really, the thing on this PR that really needs looking over is the types.  Here are the questions I have for @gnidan:
1. I made `interpretations` optional, and only included it when there's something to put in there.  Is this right?  Should it be required?  (Note that if we make it required, and then later also make it required on individual values when we do reverse ENS resolution, I'm going to have to go updating all the wrap functions... not that that'd actually be hard, just annoying, but...)
2. You'll notice that `multicall` is not `CalldataDecoding[]`, but rather `(CalldataDecoding | null)[]`.  This is because, what if one of the individual `bytes` decodes to an error?  Yeah, that shouldn't normally happen, but, like, we should be prepared for the possibility.  I realize this isn't ideal but I couldn't think of a better way of handling this.  Like, if you want the actual error, you can always look in `arguments`!  In the inspector, I just had these `null`s turn into a "could not decode" message... should I perhaps have it actually look in `arguments` so I can have it display the actual error message in case of an error?

But yeah, mostly nothing complicated here!

### DIFF
--- a/packages/codec/lib/core.ts
+++ b/packages/codec/lib/core.ts
@@ -60,13 +60,15 @@ export function* decodeCalldata(
       return {
         kind: "create" as const,
         decodingMode: "full" as const,
-        bytecode: Conversion.toHexString(info.state.calldata)
+        bytecode: Conversion.toHexString(info.state.calldata),
+        interpretations: {}
       };
     } else {
       return {
         kind: "unknown" as const,
         decodingMode: "full" as const,
-        data: Conversion.toHexString(info.state.calldata)
+        data: Conversion.toHexString(info.state.calldata),
+        interpretations: {}
       };
     }
   }
@@ -111,7 +113,8 @@ export function* decodeCalldata(
       class: contractType,
       abi: abiEntry,
       data: Conversion.toHexString(info.state.calldata),
-      decodingMode: "full" as const
+      decodingMode: "full" as const,
+      interpretations: {}
     };
   }
   let decodingMode: DecodingMode = allocation.allocationMode; //starts out this way, degrades to ABI if necessary
@@ -180,7 +183,8 @@ export function* decodeCalldata(
       bytecode: Conversion.toHexString(
         info.state.calldata.slice(0, allocation.offset)
       ),
-      decodingMode
+      decodingMode,
+      interpretations: {}
     };
   } else {
     return {
@@ -189,7 +193,8 @@ export function* decodeCalldata(
       abi: <Abi.FunctionEntry>allocation.abi, //we know it's a function, but typescript doesn't
       arguments: decodedArguments,
       selector,
-      decodingMode
+      decodingMode,
+      interpretations: {}
     };
   }
 }
@@ -466,7 +471,8 @@ export function* decodeEvent(
         class: emittingContractType,
         abi: allocation.abi,
         arguments: decodedArguments,
-        decodingMode
+        decodingMode,
+        interpretations: {}
       };
     } else {
       decoding = {
@@ -476,7 +482,8 @@ export function* decodeEvent(
         abi: allocation.abi,
         arguments: decodedArguments,
         selector,
-        decodingMode
+        decodingMode,
+        interpretations: {}
       };
     }
     decodings.push(decoding);
@@ -682,7 +689,8 @@ export function* decodeReturndata(
         kind: "returnmessage" as const,
         status: true as const,
         data: Conversion.toHexString(info.state.returndata),
-        decodingMode: allocation.allocationMode
+        decodingMode: allocation.allocationMode,
+        interpretations: {}
       };
       decodings.push(decoding);
       continue;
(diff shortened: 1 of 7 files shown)


gnidan:
> 1. I made `interpretations` optional, and only included it when there's something to put in there.  Is this right?

I'm leaning towards thinking it should be required and empty when there's nothing. This way it will be in front of people's faces when they integrate the format... if it's optional, it feels a bit too surprise-y when it does show up. But I could be convinced otherwise.



> 2\. You'll notice that `multicall` is not `CalldataDecoding[]`, but rather `(CalldataDecoding | null)[]`.  This is because, what if one of the individual `bytes` decodes to an error?  Yeah, that shouldn't normally happen, but, like, we should be prepared for the possibility.  I realize this isn't ideal but I couldn't think of a better way of handling this.  Like, if you want the actual error, you can always look in `arguments`!  In the inspector, I just had these `null`s turn into a "could not decode" message... should I perhaps have it actually look in `arguments` so I can have it display the actual error message in case of an error?

Hm, we could make it `(CalldataDecoding | Error)[]` or provide some other representation that can communicate error cases.

Alternatively... maybe we should treat it as a bad interpretation altogether, and indicate the error in the interpretation itself... like `CalldataDecoding[] | MulticallDecodingError`.

All of these sound kinda clunky, though. Maybe errors won't be so common and we can just stick with your "or null"... not sure.

haltman-at:
> I'm leaning towards thinking it should be required and empty when there's nothing. This way it will be in front of people's faces when they integrate the format... if it's optional, it feels a bit too surprise-y when it does show up. But I could be convinced otherwise.

OK, I've made it required.

gnidan:
Additional point: my recent Solidity Summit demo featured this PR's behavior prominently, and in doing so I realized a gap in what this supports... namely:
- multicall v2 (see [repo](https://github.com/makerdao/multicall)). there are a couple different forms within this v2, so you might want to check for all those more comprehensively (like how they define `aggregate` and `tryAggregate` separately)
- Uniswap v3's "nonstandard" multicall, which defines a function that takes a preceding `deadline` argument (see [IMulticallExtended](https://github.com/Uniswap/swap-router-contracts/blob/main/contracts/interfaces/IMulticallExtended.sol))

I did some hacking to add these in [this commit](https://github.com/trufflesuite/truffle/commit/84be17029ac8b62aa43cc61f72ee44ddc1ce06d8), but the variety of forms here suggests to me that we want to handle the differences more robustly.

(Is it possible to be less strict when detecting? Like can we say "use the interpretation if it looks like it might be a multicall?" I'm guessing this is hard)

haltman-at:
> I did some hacking to add these in [this commit](https://github.com/trufflesuite/truffle/commit/84be17029ac8b62aa43cc61f72ee44ddc1ce06d8), but the variety of forms here suggests to me that we want to handle the differences more robustly.

I mean, these could just be different interpretations and we could update it as need be...? :-/  Yeah I dunno I don't really have a better idea offhand.

> (Is it possible to be less strict when detecting? Like can we say "use the interpretation if it looks like it might be a multicall?" I'm guessing this is hard)

Yeah I don't really have any idea how that would work.

haltman-at:
OK, I've added the additional multicall interpretations (together with tests for most of them and updates to the inspector).  Notes on these:

1. I grouped `blockAndAggregate` into `aggregate` and `tryBlockAndAggregate` into `tryAggregate` rather than giving these separate interpretations, since they have the same arguments, and only the return values differ.  If you want to know which this one is, well, again, the full decoding is still there; you can just look at `decoding.abi.name`.
2. Not sure what you'll think of the format; hoping it's OK.  Like e.g. I included `requireSuccess` as a direct boolean, and `deadline` as a BN.
3. The way I handled `aggregate` in the inspectors is a little hacky.  Oh well? :-/

So uh yeah hope this is good?

gnidan:
> Moreover, as I mentioned, you can always look outside the `interpretations` field if you really want to know about the errors.
> 
> All this still leaves the question though @gnidan: Should I have the inspector do just that, to handle such errors better?

This is probably worth doing, but if you want to open an issue for it, I don't think it needs to be this PR

gnidan (approved):
This looks good. I think the additional interpretation is probably the right call. Thank you!

PR #4958, opened 2022-03-29
92% reduction in RPC API calls during migrations

tenthirtyone:
This is a mess! It's just a test to see what the change breaks.


packages/config/src/configDefaults.ts
```diff
@@ -2,6 +2,8 @@ import * as path from "path";
 import Provider from "@truffle/provider";
 import TruffleConfig from "./";
 
+let provider: any;
```
eggplantzzz, line 5: Hmm, in Truffle what scope is this available in??

tenthirtyone:
Closing in favor of https://github.com/trufflesuite/truffle/pull/5029

PR #4961, opened 2022-03-30
Set yarn version to 1.19.0

gnidan:
... to get around a "expected workspace package" that doesn't make sense.

(Marking this as draft because discussion is probably warranted)

cc @fainashalts since you were running into the same problem


fainashalts:
Thanks for the cc, I think this is what we should probably do here! Doesn't seem like we need any features from newer yarn versions, right? [Here's](https://github.com/yarnpkg/yarn/issues/7807) a link to the issue discussion that led to me using yarn v1.19.0, for further reading. 

It does seem that someone deleted their yarn.lock file and re-attempted successfully 21 days ago. I tried that with Truffle just now and ran into this error when running `yarn bootstrap` after: 

```
@truffle/blockchain-utils: index.ts(1,58): error TS2307: Cannot find module 'web3/providers' or its corresponding type declarations.
```

I am in the midst of backlog grooming and don't have time to look further but wanted to bring it up! I don't mind using an older version of yarn but if the `yarn.lock` trick might work it is probably worth trying.

benjamincburns:
@fainashalts I also ran into the same error while troubleshooting this issue just now after running `npx yarn-deduplicate`. I believe it's caused by an undeclared dependency on `web3` in `@truffle/blockchain-utils`. It appears that package has the types dep, but not the actual module.

benjamincburns:
@gnidan FWIW, I think `npx yarn-deduplicate` is probably the most correct way to fix this, as I believe this is caused by some detritus in our `yarn.lock`. I'll attempt a PR in the morning for comparison purposes.

cds-amal:
I noticed that error when using `yarn add ...` in a package instead of lerna's API. It's weird because I recall being able to do this in the past. Ben likely found the root cause. Any luck on that @benjamincburns ?

benjamincburns:
Ah thanks for the reminder. I think it's just old duplicate packages in the yarn.lock there's a tool that can be run to deduplicate them. I think it's literally just `npx yarn-deduplicate`.

PR #4962, opened 2022-03-31
Update solc to the latest version in contract-schema dependencies

sukanyaparashar:
### ISSUE
`solc` in `contract-schema` dependencies has an outdated version.

### SOLUTION
This PR updates the `solc` to the latest version in **contract-schema** `package.json` file.

### DIFF
--- a/packages/contract-schema/package.json
+++ b/packages/contract-schema/package.json
@@ -30,7 +30,7 @@
   "devDependencies": {
     "json-schema-to-typescript": "^5.5.0",
     "mocha": "8.1.2",
-    "solc": "0.6.0"
+    "solc": "0.8.13"
   },
   "keywords": [
     "artifacts",


packages/compile-solidity/test/test_supplier.js
```diff
@@ -122,7 +122,9 @@ describe("CompilerSupplier", function () {
     });
 
     it("compiles w/ local path solc when options specify path", async function () {
-      const pathToSolc = path.join(__dirname, "../node_modules/solc/index.js");
+      // If multiple child projects have same solc version dependency, then yarn hoists the solc dependency 
+      // in the truffle root node_modules folder rather than the local package node_modules folder
+      const pathToSolc = path.join(__dirname, "../../../node_modules/solc/index.js");
```
gnidan, line 127: oh hey it moved :)

packages/contract-schema/test/solc.js
```diff
@@ -89,7 +90,7 @@ contract B {
 
     //check that ast and legacyAST have the correct form
     assert.equal(A.ast.nodeType, "SourceUnit");
-    assert.equal(A.legacyAST.name, "SourceUnit");
+    assert.equal(A.legacyAST.nodeType, "SourceUnit");
```
gnidan, line 93: this stands out as being suspect to me... can we verify that the `legacyAST` is actually the legacy AST format?

gnidan (commented):
nice! just a quick question about the legacy AST stuff

packages/contract-schema/test/solc.js
sukanyaparashar: Hey @gnidan . Thanks for the comment. I tried printing out the `legacyAST` which gives me the below format in JSON.
`{
  absolutePath: 'A.sol',
  exportedSymbols: { A: [ 12 ], B: [ 17 ] },
  id: 18,
  license: 'MIT',
  nodeType: 'SourceUnit',
  nodes: [
    {
      id: 1,
      literals: [Array],
      nodeType: 'PragmaDirective',
      src: '34:23:0'
    },
    {
      abstract: false,
      baseContracts: [],
      canonicalName: 'A',
      contractDependencies: [],
      contractKind: 'contract',
      fullyImplemented: true,
      id: 12,
      linearizedBaseContracts: [Array],
      name: 'A',
      nameLocation: '70:1:0',
      nodeType: 'ContractDefinition',
      nodes: [Array],
      scope: 18,
      src: '61:80:0',
      usedErrors: []
    },
    {
      abstract: false,
      baseContracts: [],
      canonicalName: 'B',
      contractDependencies: [],
      contractKind: 'contract',
      fullyImplemented: true,
      id: 17,
      linearizedBaseContracts: [Array],
      name: 'B',
      nameLocation: '154:1:0',
      nodeType: 'ContractDefinition',
      nodes: [Array],
      scope: 18,
      src: '145:60:0',
      usedErrors: []
    }
  ],
  src: '34:174:0'
}`
If this is what you are mentioning about or please do correct me if I am wrong somewhere.

gnidan: hm that looks like the new AST format. could you compare and see?

if newer Solidity gets rid of the legacy format, we probably want to remove it from our tests, and possibly see about removing that field from @truffle/contract in such cases (otherwise we'd be lying)

sukanyaparashar: I think they have removed the legacy AST stuff from `0.8.0` onwards. I will check that. Thanks.

gnidan: we can probably treat it as a separate PR to fix the contract behavior, but we probably want to remove this line from the test here (if it is, in fact, removed.)

thanks @sukanyaparashar!

cc @haltman-at in case you have opinions when you get back

cds-amal: Is this helpful? https://docs.soliditylang.org/en/v0.8.13/080-breaking-changes.html?highlight=legacyAST#interface-changes

sukanyaparashar: Yes @gnidan . I think I can remove this line in this test and then make another PR for removing the legacy AST in the `@truffle/contract` as both `A.ast` and `A.legacyAST` are yielding the same results. We can wait for @haltman-at 's comment on this.

sukanyaparashar: > Is this helpful? https://docs.soliditylang.org/en/v0.8.13/080-breaking-changes.html?highlight=legacyAST#interface-changes

Thanks @cds-amal . This link is helpful. Is the `compact AST` mentioned there same as we are referring to in `A.ast`?

packages/contract-schema/test/solc.js
```diff
@@ -88,9 +88,8 @@ describe("solc", function () {
       );
     });
 
-    //check that ast and legacyAST have the correct form
+    //check that ast has the correct form
```
sukanyaparashar, line 91: The check for the legacyAST is removed from this file because the legacy AST has been removed from the **solc**  **0.8.0** onwards. Can you please check if this can be done, @haltman-at ?

haltman-at: Not sure what you're asking.  I agree it makes sense to remove that check though.

packages/contract-schema/test/solc.js
gnidan: > Whether that's a good idea, I don't know, but just confirming that that is the current behavior, even if it ought to be changed.

Definitely should be changed! That's a bug :)

packages/contract-schema/test/solc.js
```diff
@@ -4,20 +4,21 @@ const Schema = require("../");
 const debug = require("debug")("test:solc");
 
 describe("solc", function () {
-  const exampleSolidity = `pragma solidity ^0.6.0;
+  const exampleSolidity = `// SPDX-License-Identifier: MIT
+  pragma solidity ^0.8.0;
```
haltman-at, line 8: Worth noting that adding in those extra spaces will mean the lines of the actual Solidity source will begin with those extra spaces. :P  That's presumably why they weren't there before.  However since this has no actual effect, it's not a problem to change it.

haltman-at (approved):
Seems good to me!  Thanks Sukanya!

packages/contract-schema/test/solc.js
sukanyaparashar: Oh, sorry if I am not clear about what I asked. I just wanted your confirmation if it is okay to remove the legacy AST check. Thanks for explaining that @haltman-at .

packages/contract-schema/test/solc.js
sukanyaparashar: Sure, @gnidan and @haltman-at . We can change this behavior in another PR.

packages/contract-schema/test/solc.js
sukanyaparashar: Thanks for explaining about the spaces @haltman-at .

Issue #4971, opened 2022-04-04
Create consistent Button elements for use across the UI


michaeljohnbennett:
Add in a consistent react component or buttons to allow singular source of truth for button design across the UI.

Work done in other branch (pending tx views) extracted out into a single PR.

cliffoo:
Resolved in the [newest Truffle release](https://github.com/trufflesuite/truffle/releases/tag/v5.5.19). Thanks @michaeljohnbennett!

PR #4972, opened 2022-04-04
Unify Truffle Dashboard <Button /> component and provide new style properties - 4971

michaeljohnbennett:
Addresses #4971 

As per discussion with @MicaiahReid on adding in the refactored button code in a separate PR.

- Cleaned up imports to be all package relative.
- Also DashboardProvider.tsx renamed to be default packed export.
- Buttons cleaned up around site and disconnect button made a small red button.

We now have a Button element that will take styling of size/variants(filled/empty/etc)/colour.


michaeljohnbennett:
going to revert the package relative imports as they fail on GitHub but work locally.

packages/dashboard/src/components/common/Button.tsx
```diff
@@ -1,25 +1,108 @@
...
-interface Props {
-  disabled?: boolean;
-  onClick: MouseEventHandler<HTMLElement>;
-  text: string;
-}
+export type ButtonColor = "blue" | "red" | "brown" | "default";
+export type ButtonSize = "xs" | "sm" | "lg";
+export type ButtonVariant = "outlined" | "filled" | "empty";
+
+const DIMENSIONS = {
```
fainashalts, line 8: Should these style-related aspects be in a separate file? I don't feel super strongly about it but that might be nicer.

gnidan: Do they relate to anything outside this file? If they're only related to `<Button />` then they should probably live in this file

michaeljohnbennett: its all button centric config. sorry I wasn't getting notices for this PR.

PR #4980, opened 2022-04-04
Bug fix: Allow functions as input for functions and contracts in `@truffle/encoder`

haltman-at:
A thing I forgot while writing encoder: If `x` is callable, then `typeof x` will return `"function"`, not `"object"`!  This means that callable objects wouldn't be accepted as input for external functions.  It was always my intention that those would work (because of the intended use case with Truffle Contract, that with a rewritten version of Truffle Contract one could actually pass in the methods), but they didn't.  So, I made that work, and also added a test.

I also did the same for contract input, because, well, why not; figured it makes the most sense for it to act analogously.

### DIFF
--- a/packages/encoder/test/address.test.ts
+++ b/packages/encoder/test/address.test.ts
@@ -173,6 +173,18 @@ describe("Encoding", () => {
       );
     });
 
+    it("Encodes callable objects with an address field", async () => {
+      let input: any = () => undefined;
+      input.address = "0x10ca7e901d10CA7E901D10Ca7e901D10CA7e901D";
+      input.garbate = "garbage";
+      const { data } = await encoder.encodeTxNoResolution(abi, [input]);
+      assert.strictEqual(
+        data,
+        selector +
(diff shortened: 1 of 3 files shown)


gnidan (approved):
Huh alright!

packages/encoder/test/address.test.ts
```diff
@@ -173,6 +173,18 @@ describe("Encoding", () => {
       );
     });
 
+    it("Encodes callable objects with an address field", async () => {
+      let input: any = () => undefined;
+      input.address = "0x10ca7e901d10CA7E901D10Ca7e901D10CA7e901D";
+      input.garbate = "garbage";
```
cds-amal, line 179: @haltman-at is `garbate` a typo?

haltman-at: Oops, thanks, it is!  Fortunately it's a deliberate garbage field so it's irrelevant. :)

PR #4983, opened 2022-04-05
Bug fix: correct error in function name

eggplantzzz:
In the [PR to remove the reporters package](https://github.com/trufflesuite/truffle/pull/4436) (I think), there was an error where a function name changed and the location where it is called was not. This PR updates the code to call the correct function.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -17655,7 +17655,7 @@ inquirer-autocomplete-prompt@^1.4.0:
     run-async "^2.4.0"
     rxjs "^6.6.2"
 
-inquirer@8.2.2:
+inquirer@8.2.2, inquirer@^8.2.0:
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -17655,7 +17655,7 @@ inquirer-autocomplete-prompt@^1.4.0:
     run-async "^2.4.0"
     rxjs "^6.6.2"
 
-inquirer@8.2.2:
+inquirer@8.2.2, inquirer@^8.2.0:
```
eggplantzzz, line 17658: I'm not sure why this needed to be updated but I regenerated it because the yarn.lock check in CI wouldn't pass it.

haltman-at (approved):
Thanks Tyler!

PR #4984, opened 2022-04-05
Remove `fs` adapter from @truffle/db

tenthirtyone:
Removes jsondown as a dep on truffle db.


haltman-at (approved):
Thanks @tenthirtyone!  I mean I'm not *exactly* qualified to approve this, I'm taking your word for it that we don't need this functionality, but... :)

tenthirtyone:
@haltman-at 
image

PR #4986, opened 2022-04-05
Internal improvement: Swap in mvdir for cpr package

eggplantzzz:
Truffle has a `copy` utility responsible for copying files or directories. It uses a very old package called `cpr` which this PR swaps out for a package with very few dependencies called `mvdir`. A lot of the code in the `copy` utility was also updated to be more modern and simple.

### DIFF
--- a/packages/core/lib/commands/migrate/setUpDryRunEnvironmentThenRunMigrations.js
+++ b/packages/core/lib/commands/migrate/setUpDryRunEnvironmentThenRunMigrations.js
@@ -1,5 +1,4 @@
 const { Environment } = require("@truffle/environment");
-const { promisify } = require("util");
 const Artifactor = require("@truffle/artifactor");
 const Resolver = require("@truffle/resolver");
 const copy = require("../../copy");
@@ -24,7 +23,7 @@ module.exports = async function (config) {
     prefix: "migrate-dry-run-"
   }).name;
 
-  await promisify(copy)(config.contracts_build_directory, temporaryDirectory);
+  await copy(config.contracts_build_directory, temporaryDirectory);
 
   config.contracts_build_directory = temporaryDirectory;
   // Note: Create a new artifactor and resolver with the updated config.
--- a/packages/core/lib/commands/migrate/setUpDryRunEnvironmentThenRunMigrations.js
+++ b/packages/core/lib/commands/migrate/setUpDryRunEnvironmentThenRunMigrations.js
@@ -23,9 +23,7 @@ module.exports = async function (config) {
     prefix: "migrate-dry-run-"
   }).name;
 
-  await copy(config.contracts_build_directory, temporaryDirectory, {
-    overwrite: true
-  });
+  await copy(config.contracts_build_directory, temporaryDirectory);
 
   config.contracts_build_directory = temporaryDirectory;
   // Note: Create a new artifactor and resolver with the updated config.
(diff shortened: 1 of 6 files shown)


packages/core/lib/commands/test/copyArtifactsToTempDir.js
```diff
@@ -19,11 +18,11 @@ const copyArtifactsToTempDir = async config => {
     return {temporaryDirectory};
   }
 
-  await promisify(copy)(config.contracts_build_directory, temporaryDirectory);
+  await copy(config.contracts_build_directory, temporaryDirectory);
```
haltman-at, line 21: Any reason this doesn't use `overwrite` like the other one?

packages/core/lib/commands/migrate/setUpDryRunEnvironmentThenRunMigrations.js
```diff
@@ -24,7 +23,15 @@ module.exports = async function (config) {
...
   }).name;
 
-  await promisify(copy)(config.contracts_build_directory, temporaryDirectory);
+  // copy returns an error if there is an error, undefined if successful
+  const error = await copy(
+    config.contracts_build_directory,
+    temporaryDirectory,
+    { overwrite: true }
+  );
+  if (error) {
```
haltman-at, line 32: Rather than having this check here, might it make more sense to put this check inside the `copy` function itself?  You didn't put it on the other invocations, but I'm assuming they should probably have it, right?  (And if they shouldn't, they should probably use a `try`/`catch` to explicitly swallow the error rather than implicitly swallowing it like they do in this version.)

packages/core/lib/copy.js
```diff
@@ -1,62 +1,16 @@
...
-
-  readStream.on("error", function(err) {
-    callback(err);
-    callback = function() {};
-  });
-
-  writeStream.on("error", function(err) {
-    callback(err);
-    callback = function() {};
+const copy = async function (from, to, options) {
```
haltman-at, line 9: So much code gone, hooray! :)

haltman-at (commented):
Thanks Tyler!  A few comments.

packages/core/lib/commands/migrate/setUpDryRunEnvironmentThenRunMigrations.js
eggplantzzz: Ah yes, this is a good idea.

packages/core/lib/commands/test/copyArtifactsToTempDir.js
eggplantzzz: Nice catch, this will definitely need it if a temporary directory is being created first!

eggplantzzz: I'm confused, this actually seems to work without the `overwrite: true` on `truffle test` on the metacoin box. I have no idea why!

eggplantzzz:
I decided to make `overwrite: true` the default for the copy util because every usage of it relies on this behavior.

eggplantzzz:
I decided to make `overwrite: true` the default for the copy util because every usage of it relies on this behavior.

haltman-at (approved):

packages/core/lib/commands/test/copyArtifactsToTempDir.js
eggplantzzz: Ah I think it ends up picking up and compiling the artifacts later when it goes to do all the test-specific stuff like `Assert.sol`.

PR #4988, opened 2022-04-05
Dependency update: Bump ethers to 5.6.2

eggplantzzz:


packages/contract/lib/statuserror.js
```diff
@@ -11,7 +11,9 @@ class StatusError extends TruffleError {
 
     if (reason) reasonString = `Reason given: ${reason}.`;
 
-    if (utils.bigNumberify(receipt.gasUsed).eq(utils.bigNumberify(gasLimit))) {
+    if (
+      utils.BigNumber.from(receipt.gasUsed).eq(utils.BigNumber.from(gasLimit))
```
benjamincburns, line 15: I feel like our prettier config could do a bit better in cases like this 🤔 

not suggesting that fixing that should be in scope for this review, however

benjamincburns (approved):
LGTM, though I wonder if we've taken an unnecessary dependency on `ethers` in some places in order to make use of its `BigNumber` utility, when making better use of an existing dependency (e.g. `bn.js`, which `ethers` uses internally), or perhaps even `BigInt`, would do.

I'm not suggesting that you should address that here, however - just an observation.

PR #4989, opened 2022-04-06
Enhancement: Indicate which storage variables have unknown values in debugger

haltman-at:
Note: This PR is not quite up to my usual standards, but I'm PRing it anyway because we're trying to get this out fast!  I'm intending to address the problems in a later PR.  (Possibly I'll just handle this when I implement reverse ENS resolution, or possibly I'll do it separately first.)

This PR adds an option to the debugger's `variables` and `variable` methods so that it will indicate which storage variables are unknown, as opposed to just filling in the missing values with zero.  It then also sets the CLI debugger to use that option, and display such missing values with gray question marks.  (Note that when dealing with complex values, parts can be known and parts unknown!)  It also makes it so that the debugger will print an explanatory note when there is at least one such gray question mark.

Naturally, this involves some changes to codec.  First, there's a new type of error, `ReadErrorStorageDeliberate`.  (Alternative name suggestions are appreciated -- `NoStorageSuppliedError`, perhaps?)  This error is used for values that are unknown because the storage is unknown.  This error is thrown if the storage returned in response to a request is `null` instead of a `Uint8Array`.  Meanwhile, the debugger's `decode` saga gets an option to pass `null` instead of zero when it doesn't know the storage, and the `variables` and `variable` methods get options that are forwarded to the `decode` option.

Also, `ResultInpsector` has been updated for the new error type, and displays it as a gray question mark, as mentioned.

There's also a function added to detect when a variable contains any of these errors (however deep), so we can add the note to the debugger printout when necessary.  Note that due to practical limitations, we only apply this note to `v` and to `:variable`, but not to more general expressions, as we don't have a good way of determining what variables are used in more general expressions.

Also I added a test.

OK, so what's the thing that's wrong with this PR that I'll need to go back and change later?

Well, it's that whole `null` thing -- the response type for decoding is supposed to be `Uint8Array`, not `Uint8Array | null`.  Of course, since codec doesn't currently have `strictNullChecks` on, this doesn't actually matter.  Still, I should avoid relying on that!  So I should update the types to have `| null`.  But by the same logic that also means I should account for the possibility of a `null` in response to a code request, and add a similar error there (`ReadErrorCodeDeliberate`? `NoCodeSuppliedError`?).  So yeah.  But in the interest of speed that's been dropped from this PR... so this will come later, I guess!

Btw this is going to cause merge conflicts with #4882 because this edits code that that moved. :-/

### DIFF
--- a/packages/codec/lib/format/errors.ts
+++ b/packages/codec/lib/format/errors.ts
@@ -814,6 +814,7 @@ export type ReadError =
   | ReadErrorStack
   | ReadErrorBytes
   | ReadErrorStorage
+  | ReadErrorStorageDeliberate
   | UnusedImmutableError;
 
 /**
@@ -908,6 +909,18 @@ export interface ReadErrorBytes {
  */
 export interface ReadErrorStorage {
   kind: "ReadErrorStorage";
+  range: Storage.Range;
+}
+
+/**
+ * A read error in storage, but one triggered deliberately to indicate
+ * that that particular area of storage is unknown, rather than due to
+ * an unexpected error condition.
+ *
+ * @Category Generic errors
+ */
+export interface ReadErrorStorageDeliberate {
+  kind: "ReadErrorStorageDeliberate";
   range: Storage.Range;
 }
 
--- a/packages/codec/lib/format/errors.ts
+++ b/packages/codec/lib/format/errors.ts
@@ -814,7 +814,7 @@ export type ReadError =
   | ReadErrorStack
   | ReadErrorBytes
   | ReadErrorStorage
-  | ReadErrorStorageDeliberate
+  | StorageNotSuppliedError
   | UnusedImmutableError;
 
 /**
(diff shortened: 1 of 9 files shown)


packages/codec/lib/format/errors.ts
```diff
@@ -814,6 +814,7 @@ export type ReadError =
   | ReadErrorStack
   | ReadErrorBytes
   | ReadErrorStorage
+  | ReadErrorStorageDeliberate
```
gnidan, line 817: Can we call this something that indicates the read error is due to unhydrated trace? Maybe `ReadErrorStorageNotTraced`?

haltman-at: Well, notionally it would occur whenever someone decided to pass in `null`, for whatever reason.  That's why my other suggestion was `NoStorageSuppliedError`.

gnidan: Ah cool yes I prefer that!

haltman-at: OK I did that, although instead of `NoStorageSuppliedError` I called it `StorageNotSuppliedError`.

Note I left unchanged the name of the function that digs for these.

gnidan (approved):
merge merge merge

PR #4993, opened 2022-04-06
Dependency updates: Bump TypeScript and Jest packages

eggplantzzz:


packages/abi-utils/package.json
```diff
@@ -2,51 +2,51 @@
...
     "prepare": "yarn build",
     "test": "jest lib/**"
   },
-  "bugs": {
-    "url": "https://github.com/trufflesuite/truffle/issues"
+  "types": "dist/lib/index.d.ts",
+  "dependencies": {
+    "change-case": "3.0.2",
+    "faker": "5.5.3",
+    "fast-check": "^2.12.1"
```
cds-amal, line 29: I realize this isn't part of you PR but wonder why these aren't dev deps.

packages/db-kit/package.json
```diff
@@ -74,5 +52,28 @@
     "meow": "^9.0.0",
     "react": "^16.8.0",
     "source-map-support": "^0.5.19"
```
cds-amal, line 54: Should this be a dev dep?

cds-amal (commented):
The code change themselves look good but as a matter of git readability there are two commits that have a `subject: "y"` which devalues `git log`'s value and affected my ability to easily get an overview of the PR using normal git tooling. Could you update those commit messages before merging?

packages/abi-utils/package.json
cds-amal: Lernaaaaaaaaaaaa

cds-amal (approved):
Thanks @eggplantzzz ! I like it

xhulz (approved):

PR #4994, opened 2022-04-06
Dependency update: Bump @truffle/error and @types/node

eggplantzzz:


packages/compile-common/package.json
```diff
@@ -21,7 +21,7 @@
   },
   "types": "dist/src/index.d.ts",
   "dependencies": {
-    "@truffle/error": "^0.1.0",
+    "@truffle/error": "0.1.0",
```
haltman-at, line 24: Is there some reason you pinned these...?  Our dependencies on our own packages aren't normally pinned.

packages/plugins/package.json
```diff
@@ -33,9 +35,7 @@
     "ts-jest": "^26.1.0",
     "typescript": "^4.1.4"
   },
-  "dependencies": {
-    "@truffle/error": "^0.0.8",
```
haltman-at, line 37: Yikes, thanks for catching this!

haltman-at (commented):
Thanks for catching that outdated `@truffle/error` dependency!  This doesn't seem to involve `@types/node` at all though as best I can tell?

packages/compile-common/package.json
eggplantzzz: yeah we should probably `^` this - sorry, `lernaupdate` pins it by default and I neglected to doctor it up

PR #4997, opened 2022-04-07
core: fix debugger -x crash due to missing spinner

benjamincburns:
Bit of an embarrassing one. Seems I've come to lean too heavily on the TS compiler 😊


gnidan (approved):
d'oh!

PR #4998, opened 2022-04-07
Fix unbox event output

benjamincburns:
Fixes some minor errors in the output of the `unbox` command. Namely duplicate messages on failure due to the `unbox:fail` event firing twice, and a bad spinner success message due to a missed string when migrating from `spinnes` to `@truffle/spinners`.

### DIFF
--- a/packages/box/lib/utils/index.ts
+++ b/packages/box/lib/utils/index.ts
@@ -2,8 +2,6 @@ import unbox from "./unbox";
 import fs from "fs";
 import config from "../config";
 import tmp from "tmp";
-import process from "process";
-const cwd = require("process").cwd();
 import path from "path";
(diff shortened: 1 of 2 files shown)


packages/box/lib/utils/index.ts
```diff
@@ -40,17 +33,12 @@ export = {
...
-    } catch (error) {
-      events.emit("unbox:fail");
-      throw error;
-    }
+    const tmpDir = tmp.dirSync(options);
+    events.emit("unbox:preparingToDownload:succeed");
+    return {
+      path: path.join(tmpDir.name, "box"),
+      cleanupCallback: tmpDir.removeCallback
+    };
```
eggplantzzz, line 41: Ah this makes sense, I didn't realize `emit` is called in `Box` as well. Good catch!

PR #5001, opened 2022-04-07
spinners: fix spinner color issues

benjamincburns:
- [x] replace the `spinnies` dependency with a truffle-maintained fork (`@trufflesuite/spinnies`)
- [x] removes workaround for jcarpanelli/spinnies#34 (fixed by trufflesuite/spinnies#1) 
- [x] implements `warn` feature on `@truffle/spinners`
- [x] changes spinner text color back to terminal default

### DIFF
--- a/packages/spinners/lib/spinner.ts
+++ b/packages/spinners/lib/spinner.ts
@@ -1,4 +1,8 @@
-import Spinnies, { SpinnerOptions, StopAllStatus } from "spinnies";
+import Spinnies, {
+  SpinnerOptions,
+  StopAllStatus,
+  Color
+} from "@trufflesuite/spinnies";
 
 const spinnies = new Spinnies();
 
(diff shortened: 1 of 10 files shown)


benjamincburns:
Before this can be brought out of draft status, I need to:

- [ ] Get trufflesuite/spinnies#1 approved and merged
 - Needs a code review ideally, though that doesn't necessarily need to block this
- [ ] Release & publish the `@trufflesuite/spinnies` package
 - I don't believe that I have the necessary rights in the trufflesuite NPM org to do this
- [ ] Update `packages/spinners/package.json` to  use the properly released `@trufflesuite/spinnies` package

When the current build shows green I'll edit this comment to include a link to the green build, and preemptively push the updated `package.json` file. Unfortunately this will cause the build to go red until the package mentioned above is published.

packages/spinners/package.json
```diff
@@ -12,7 +12,7 @@
   "bugs": {
     "url": "https://github.com/trufflesuite/truffle/issues"
   },
-  "version": "0.1.2",
+  "version": "0.2.0-0",
```
haltman-at, line 15: Oh, interesting, actually changing this in the PR itself... we haven't typically done this, but it might be a good idea for us to start!

packages/spinners/lib/spinner.ts
```diff
@@ -45,55 +49,54 @@ export class Spinner {
       return;
     }
     spinnies.remove(this.name);
-    spinnies.checkIfActiveSpinners();
```
haltman-at, line 48: Sorry, as someone not familiar with all the spinner changes, what's the significance of this change?

haltman-at (commented):
Mostly makes sense to me.  Only one line I don't quite understand.

packages/spinners/lib/spinner.ts
benjamincburns: The line that was removed was a workaround for a known issue in the official `spinnies` module. I fixed that issue in our fork, and removed the workaround.

benjamincburns: Bug in question was: jcarpanelli/spinnies#34

haltman-at: OK, thanks!

haltman-at (approved):

PR #5005, opened 2022-04-08
Bug fix: Replace for/in loop with C-style loop to prevent VSCode problems

haltman-at:
Once again, it seems that my use of a `for ... in` loop is causing problems for VSCode.  Since in this case I was using it to loop over a non-sparse array, I just changed it to a C-style loop.  Hopefully that should get rid of the problem.

It may be worth doing a more thorough sweep of the codebase (or at least debugger and codec) for these loops if they're going to keep causing problems for VSCode, but this PR just fixes this particular case.

### DIFF
--- a/packages/codec/lib/format/utils/circularity.ts
+++ b/packages/codec/lib/format/utils/circularity.ts
@@ -17,7 +17,7 @@ function tieWithTable(
   let reference: number;
   switch (untied.type.typeClass) {
     case "array":
-      let untiedAsArray = <Format.Values.ArrayValue>untied; //dammit TS
+      const untiedAsArray = <Format.Values.ArrayValue>untied; //dammit TS
       reference = untiedAsArray.reference;
       if (reference === undefined) {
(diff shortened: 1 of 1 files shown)


michaeljohnbennett (approved):
fixes look ok to me, will test shortly.

haltman-at:
OK, I'll wait to merge until you've tested it (or until we're close to release :P ).

michaeljohnbennett:
I've tested it locally, linking into my build and I'm getting the old error message that I cannot connect to the Ethereum RPC Endpoint: 

```Could not connect to your Ethereum client.
Please check that your Ethereum client:
- is running
- is accepting RPC connections (i.e., "--rpc" or "--http" option is used in geth)
- is accessible over the network
- is properly configured in your Truffle configuration file (truffle-config.js)
```

So, perhaps there are other places breaking now that you fixed this one but I cannot see any traces/debug output to see where.

Issue #4425, opened 2021-11-12
Update the CompilerSupplier's default compilerRoots to be more canonical


gnidan:
> https://ethereum.github.io/solc-bin/bin/ is deprecated but it's just not receiving new binaries. Old ones are still there. Keeping it here should be fine for the time being and could actually provide an additional fallback in case the other options go down. It would only be a problem if we needed to rebuild one of the binaries and replace it but these cases are rare (currently there's just one that might end up being rebuilt eventually: https://github.com/ethereum/solidity/issues/11671).
> 
> Also, you might want to add https://solc-bin.ethereum.org/wasm/ (or rather https://binaries.soliditylang.org/wasm/ but it's just a cosmetic difference as they both point at the same thing) to the list above `bin/`. It has wasm rather than asm.js builds and uses the same naming convention so it can be used as the primary source with `bin/` being a fallback. Wasm builds offer better performance and don't have problems like https://github.com/ethereum/solidity/issues/11671. In 0.6.2 we switched to doing wasm builds and putting them in `bin/` but older binaries were not replaced. Instead wasm rebuilds of older versions are in `wasm/`.
> 
> To make this more apparent there are new dirs that contain the same binaries but correctly separated (`emscripten-asmjs/` and `emscripten-wasm32/`). I'd recommend using them if possible, though unfortunately they are not drop-in replacements since the platform name is now a part of the naming scheme.
> 
> Note also that each directory, including `wasm/`, has a separate `list.json` and wasm binaries obviously have different hashes than the asm.js ones so maybe it does not fit into this system after all. At least if you do not download `list.json` from the same entry you use to download binaries.

_Originally posted by @cameel in https://github.com/trufflesuite/truffle/pull/4222#discussion_r689001206_

haltman-at:
Hm, this seeems to be a duplicate of #4252.

haltman-at:
OK, I think we can finally close this with 5.5.29 released!

PR #5008, opened 2022-04-08
Fix: official repo to list Solidity's releases

Neurone:
Added the official repo for Solidity's Compiler releases, before the Truffle custom one.
Added a note for the last two deprecated repo.

### DIFF
--- a/packages/compile-solidity/test/test_supplier.js
+++ b/packages/compile-solidity/test/test_supplier.js
@@ -152,8 +152,9 @@ describe("CompilerSupplier", function () {
       );
 
       // Delete if it's already there.
-      if (await fse.exists(compilerCacheDirectory))
-        await fse.removeSync(compilerCacheDirectory);
+      if (fse.existsSync(compilerCacheDirectory)) {
+        fse.removeSync(compilerCacheDirectory);
+      }
 
       options.compilers = {
         solc: { version: "0.4.21" }
@@ -172,8 +173,9 @@ describe("CompilerSupplier", function () {
         return filename.includes("v0.4.21+commit.dfe3193c");
       });
 
-      assert(
-        compilerFilename !== undefined,
+      assert.equal(
+        compilerFilename,
+        undefined,
         "The compiler should have been cached but wasn't"
       );
 
--- a/packages/compile-solidity/test/test_supplier.js
+++ b/packages/compile-solidity/test/test_supplier.js
@@ -1,7 +1,7 @@
 const debug = require("debug")("compile:test:test_supplier");
 const fse = require("fs-extra");
 const path = require("path");
-const assert = require("assert");
+const { assert } = require("chai");
 const { Resolver } = require("@truffle/resolver");
 const { Compile } = require("@truffle/compile-solidity");
 const Config = require("@truffle/config");
@@ -173,9 +173,8 @@ describe("CompilerSupplier", function () {
         return filename.includes("v0.4.21+commit.dfe3193c");
       });
 
-      assert.equal(
+      assert.isDefined(
         compilerFilename,
-        undefined,
         "The compiler should have been cached but wasn't"
       );
 
--- a/packages/compile-solidity/test/test_supplier.js
+++ b/packages/compile-solidity/test/test_supplier.js
@@ -150,13 +150,10 @@ describe("CompilerSupplier", function () {
         Config.getTruffleDataDirectory(),
         "compilers/node_modules"
       );
-      const expectedCache = path.resolve(
-        compilerCacheDirectory,
-        "soljson-v0.4.21+commit.dfe3193c.js"
-      );
 
       // Delete if it's already there.
-      if (await fse.exists(expectedCache)) await fse.unlink(expectedCache);
+      if (await fse.exists(compilerCacheDirectory))
+        await fse.removeSync(compilerCacheDirectory);
 
       options.compilers = {
         solc: { version: "0.4.21" }
@@ -170,10 +167,23 @@ describe("CompilerSupplier", function () {
         options: cachedOptions
       });
 
-      assert(await fse.exists(expectedCache), "Should have cached compiler");
+      const cachedCompilerFilenames = fse.readdirSync(compilerCacheDirectory);
+      const compilerFilename = cachedCompilerFilenames.find(filename => {
+        return filename.includes("v0.4.21+commit.dfe3193c");
+      });
+
+      assert(
+        compilerFilename !== undefined,
+        "The compiler should have been cached but wasn't"
+      );
+
+      const cachedCompilerPath = path.join(
+        compilerCacheDirectory,
+        compilerFilename
+      );
 
       // Get cached solc access time
-      initialAccessTime = (await fse.stat(expectedCache)).atime.getTime();
+      initialAccessTime = (await fse.stat(cachedCompilerPath)).atime.getTime();
 
       // Wait a second and recompile, verifying that the cached solc
       // got accessed / ran ok.
@@ -184,7 +194,7 @@ describe("CompilerSupplier", function () {
         options: cachedOptions
       });
 
-      finalAccessTime = (await fse.stat(expectedCache)).atime.getTime();
+      finalAccessTime = (await fse.stat(cachedCompilerPath)).atime.getTime();
       const NewPragma = findOne("NewPragma", compilations[0].contracts);
 
       assert(NewPragma.contractName === "NewPragma", "Should have compiled");
--- a/packages/truffle/test/scenarios/commands/obtain.js
+++ b/packages/truffle/test/scenarios/commands/obtain.js
@@ -6,55 +6,52 @@ const path = require("path");
 const assert = require("assert");
 const sandbox = require("../sandbox");
 
-let logger, config, project, expectedPath;
+let logger, config, project, compilersCacheDirectory;
 
 describe("truffle obtain", function () {
   project = path.join(__dirname, "../../sources/obtain");
 
   beforeEach(async function () {
-    expectedPath = path.join(
+    compilersCacheDirectory = path.join(
       Config.getTruffleDataDirectory(),
       "compilers",
-      "node_modules",
-      "soljson-v0.7.2+commit.51b20bc0.js"
+      "node_modules"
     );
     this.timeout(10000);
     config = await sandbox.create(project);
     logger = new MemoryLogger();
     config.logger = logger;
+
+    // ensure the compiler is not cached beforehand
+    if (fse.existsSync(compilersCacheDirectory)) {
+      fse.removeSync(compilersCacheDirectory);
+    }
+  });
+
+  afterEach(() => {
+    if (fse.existsSync(compilersCacheDirectory)) {
+      fse.removeSync(compilersCacheDirectory);
+    }
   });
 
   it("fetches the solc version specified", async function () {
     this.timeout(70000);
-    // ensure the compiler does not yet exist
-    try {
-      fse.unlinkSync(expectedPath);
-    } catch (error) {
-      // unlink throws when file doesn't exist
-      if (error.code !== "ENOENT") {
-        throw error;
-      }
-    }
     await CommandRunner.run("obtain --solc=0.7.2", config);
-    assert(fse.statSync(expectedPath), "The compiler was not obtained!");
-    fse.unlinkSync(expectedPath);
+    const cachedCompilersFilenames = fse.readdirSync(compilersCacheDirectory);
+
+    assert(
+      cachedCompilersFilenames.some(filename => {
+        return filename.includes("v0.7.2+commit.51b20bc0");
+      }),
+      "The compiler was not obtained!"
+    );
   });
 
   it("respects the `quiet` option", async function () {
     this.timeout(80000);
-    // ensure the compiler does not yet exist
-    try {
-      fse.unlinkSync(expectedPath);
-    } catch (error) {
-      // unlink throws when file doesn't exist
-      if (error.code !== "ENOENT") {
-        throw error;
-      }
-    }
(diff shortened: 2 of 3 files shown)


Neurone:
This PR fixes problems you can encounter if the truffle relay node is down for some reason. See this StackExchange question: https://ethereum.stackexchange.com/questions/125523/truffle-compile-list-not-showing-latest-versions-of-compiler

gnidan:
Looks like our relay was pointing to the deprecated URLs. Fixing that now!

I think for this PR, though, we want to keep relay.trufflesuite.com as the primary source, since that will allow us to fix issues without requiring a Truffle release.

gnidan:
Thanks for bringing this to our attention! Will discuss with the team to confirm the behavior we want moving forward.

haltman-at:
Note: This is related to issue #4425.

Neurone:
@gnidan I set that as the first repo because I found this note in the comment: _this relay address exists so that we have a backup option in case more official distribution mechanisms fail_, but I see your point :)

cds-amal:
This precedence and motivation, should probably be documented

eggplantzzz:
@cameel can you let us know which urls that Truffle should use when obtaining the solc compiler? Is `https://binaries.soliditylang.org/bin/` correct or should we be using a different one?

cameel:
My recommendation would be `https://binaries.soliditylang.org/emscripten-wasm32/` with a fallback to `https://binaries.soliditylang.org/emscripten-asmjs/`.

Overall, this should be clear from [Static Binaries](https://docs.soliditylang.org/en/latest/installing-solidity.html#static-binaries) section in the docs. Let me know if you think it still leaves something out.

Neurone:
@eggplantzzz sure, no problem, I will send an update to the PR soon.

MicaiahReid:
Hi @Neurone, is this PR ready for review? I see that you've made some commits, but it isn't currently passing CI. Let us know if you are able to look into what's causing this to fail.

Neurone:
Hi, yes, it's ready.

I didn't change anything aside from the URLs in the arrays. The error is related to the content of the loaded list: the old CI for node 12 tries to find a nightly build in the list (`0.4.16-nightly.2017.8.9+commit.81887bc7`), but new URLs contain the release one only (`0.4.16+commit.d7661dd9`).

If you agree, I can change the test here https://github.com/trufflesuite/truffle/blob/3acd56290e1019ca5e3a739e3d3a597ff00a8fdc/packages/compile-solidity/test/test_supplier.js#L100 to use version `0.4.16+commit.d7661dd9`

cameel:
> Should we keep https://solc-bin.ethereum.org/bin/ and https://ethereum.github.io/solc-bin/bin/ as backups for the backups?

It would not hurt. At least for `https://ethereum.github.io`. It no longer receives new versions but it's a completely different host so in case S3 goes down, you might still be able to get binaries for some older versions from there. `https://solc-bin.ethereum.org` on the other hand is pointing at the same S3 bucket as `https://binaries.soliditylang.org` and I don't expect that to change ever so that one's probably pointless. At most it might help if there's an issue with DNS or certificates so that one resolves and the other does not.

cameel:
> The error is related to the content of the loaded list: the CI tries to find a nightly build in the list (`0.4.16-nightly.2017.8.9+commit.81887bc7`), 

Right. That's actually one difference between the new and old URLs - the nightlies are only available in `bin/`. Do you think we should add them to the new dirs too? I'd expect tools to be handing nightlies separately anyway.

Neurone:
> > The error is related to the content of the loaded list: the CI tries to find a nightly build in the list (`0.4.16-nightly.2017.8.9+commit.81887bc7`),
> 
> Right. That's actually one difference between the new and old URLs - the nightlies are only available in `bin/`. Do you think we should add them to the new dirs too? I'd expect tools to be handing nightlies separately anyway.

I think it's the cleanest path to follow. But currently, both relays redirect to `https://binaries.soliditylang.org`, so if this is the way to go, we should convince the Solidity team to update their files.

cameel:
Personally, I really don't like the fact that nightlies sit in the same dir as releases since there are tons of them. If you go to the github repo backing the S3 bucket, github won't even show you all the releases because it hits the limit of 1000 files.

But if you think that it's still better to have them in the new dirs, this is something we can definitely change. Could you submit an issue in the Solidity repo?

Neurone:
> Personally, I really don't like the fact that nightlies sit in the same dir as releases since there are tons of them. If you go to the github repo backing the S3 bucket, github won't even show you all the releases because it hits the limit of 1000 files.
> 
> But if you think that it's still better to have them in the new dirs, this is something we can definitely change. Could you submit an issue in the Solidity repo?

Sure, no problem. Another solution is to add a separate file filled with release versions. This way, we have a _clean_ list (`list-releases.json`) and a complete one (`list.json`). I'll prepare an issue for the Solidity repo soon.

cameel:
For me it's only really an issue when browsing the dir manually so I don't think splitting the list would be a solution. When I do that I don't even look at the list :)

Neurone:
I see :) I cannot list dir contents from the outside, so I didn't notice that behavior.

If you don't mind having duplicated files, another option is to use the main folders for release versions only and a specific subfolder for all versions, i.e. `/emscripten-wasm32/` only releases, `/emscripten-wasm32/full-list/` all releases.

Neurone:
@cameel I cannot open issues on https://github.com/ethereum/solc-bin, so I opened it on https://github.com/ethereum/solidity

cameel:
> I see :) I cannot list dir contents from the outside, so I didn't notice that behavior.

See https://github.com/ethereum/solc-bin/tree/gh-pages/bin. It's the repo backing the S3 bucket. The fact that there are so many files in there makes it kinda hard to browse that dir.

> I cannot open issues on https://github.com/ethereum/solc-bin, so I opened it on https://github.com/ethereum/solidity

That's fine. solc-bin does not have issues exactly because we want to have them all in one place - in the main repo, Maybe we should just put that in the README.

kevinweaver:
@neurone it looks like this still isn't passing CI, we'll restart it to see if it's flaky. Otherwise, let us know if we can help resolve the issue.

Neurone:
@kevinweaver we need to keep this PR pending until the Solidity team reaches an agreement about how to update the management of release and nightly builds (see [here](https://github.com/ethereum/solidity/issues/13015)). When that thread is closed, I can update this PR accordingly.

cds-amal:
Maintenance note: leaving this in `waiting` for more visibility

haltman-at:
> So it looks like axios throws an error and we retry the next url in the case that the server response is something other than 2xx. See [handling errors](https://www.npmjs.com/package//axios#handling-errors). This is relevant because it seems like not all urls formats will have all versions of the Solidity compiler.

Hm, so since it errors on a 404, that means that actually it's OK to use incomplete lists, right?  Since if it doesn't find it on one, it'll just go to the next!  So the case I was worried about (that it finds it's not on one list and then gives up) doesn't seem to actually be a concern, right?

davidmurdoch:
@eggplantzzz is this a good summary: we're okay with the changes as proposed, but CI is failing, so those failures need to be investigated and resolved before we can merge this?

eggplantzzz:
I think there may be something minor still here. Truffle fetches the list of compiler versions and uses that list to determine what is available. If it fetches it from a source without all versions and the version requested is not in it, it will say "hey there is no compiler version for the semver expression you specified". This may not be a problem, however, as long as we have the more complete sources checked before the incomplete ones.

eggplantzzz:
So it fetches all the version ranges [here](https://github.com/trufflesuite/truffle/blob/bc911496cdb3019a91398199dc9b260c6007df7f/packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts#L187-L211). That step is decoupled from the fetching of the actual compiler. Assuming it is in the initial fetched list, it will proceed to try and find it in all the sources. But it also uses this list to get the filenames. There will also need to be some work done if each source has files with different names. So for example, the wasm builds have a different filename than the asm. So this is totally plausible but the `VersionRange` strategy will need some re-writing.

haltman-at:
Oh, looks like you're right.  My mistake!  Yes, that means that we have to be careful about the order of things, unless we go changing the way this works...

haltman-at:
Oh, wow, it looks like the reality is even worse than we realized -- instead of going down the list of sources to find a working source, and then using that source for both the version list and the compiler itself, it goes down the list of sources to find a working source for the version list, uses that to get the filename for the compiler, and then *goes down the list again* to find the compiler, even though filenames may differ between sources!  Yikes!  This whole process needs to be reworked to be actually sensible...

Neurone:
@eggplantzzz sure! This would be very helpful also because I forgot almost everything about this PR, and you can do a better job for sure :D I just sent you an invitation.

eggplantzzz:
Thanks @Neurone for doing this work!

packages/compile-solidity/test/test_supplier.js
```diff
@@ -150,13 +150,10 @@ describe("CompilerSupplier", function () {
...
         "compilers/node_modules"
       );
-      const expectedCache = path.resolve(
-        compilerCacheDirectory,
-        "soljson-v0.4.21+commit.dfe3193c.js"
-      );
 
       // Delete if it's already there.
-      if (await fse.exists(expectedCache)) await fse.unlink(expectedCache);
+      if (await fse.exists(compilerCacheDirectory))
```
haltman-at, line 155: Minor note here: `fs.exists` is deprecated.  You should probably use `fs.existsSync` instead.

packages/compile-solidity/test/test_supplier.js
```diff
@@ -150,13 +150,10 @@ describe("CompilerSupplier", function () {
...
+        await fse.removeSync(compilerCacheDirectory);
```
haltman-at, line 156: Style note: You should probably put braces around the body of this block, indented blocks without braces can be a source of confusion for later editors.

Also, since you're using `removeSync`, the `await` is redundant.

packages/compile-solidity/test/test_supplier.js
```diff
@@ -170,10 +167,23 @@ describe("CompilerSupplier", function () {
         options: cachedOptions
       });
 
-      assert(await fse.exists(expectedCache), "Should have cached compiler");
+      const cachedCompilerFilenames = fse.readdirSync(compilerCacheDirectory);
+      const compilerFilename = cachedCompilerFilenames.find(filename => {
+        return filename.includes("v0.4.21+commit.dfe3193c");
```
haltman-at, line 172: This is truly nitpicking and you don't need to change this, but `filename => { return foo; }` could instead just be `filename => foo`.

packages/compile-solidity/test/test_supplier.js
```diff
@@ -170,10 +167,23 @@ describe("CompilerSupplier", function () {
...
+      });
+
+      assert(
```
haltman-at, line 175: Another style note -- it's better to use `assert.somethingMoreSpecific` when it exists; in this case, that could be `assert.notEqual` or `assert.notStrictEqual`.

packages/truffle/test/scenarios/commands/obtain.js
```diff
@@ -6,55 +6,52 @@ const path = require("path");
...
-        throw error;
-      }
-    }
     await CommandRunner.run("obtain --solc=0.7.2", config);
-    assert(fse.statSync(expectedPath), "The compiler was not obtained!");
-    fse.unlinkSync(expectedPath);
+    const cachedCompilersFilenames = fse.readdirSync(compilersCacheDirectory);
+
+    assert(
+      cachedCompilersFilenames.some(filename => {
```
haltman-at, line 43: Same note here as earlier about `filename => { return foo; }`.

haltman-at (commented):
OK, seems good on functionality!  I have a few minor comments on style.

packages/compile-solidity/test/test_supplier.js
eggplantzzz: I'll go ahead and address these concerns in https://github.com/trufflesuite/truffle/pull/5445 as it moves the compile-solidity tests to a new package. It should make dealing with merge conflict stuff easier.

eggplantzzz: Oh actually I will need to address these here I think since some of these are related to necessary test suite changes for this particular PR.

packages/compile-solidity/test/test_supplier.js
```diff
@@ -170,10 +168,24 @@ describe("CompilerSupplier", function () {
         options: cachedOptions
       });
 
-      assert(await fse.exists(expectedCache), "Should have cached compiler");
+      const cachedCompilerFilenames = fse.readdirSync(compilerCacheDirectory);
+      const compilerFilename = cachedCompilerFilenames.find(filename => {
+        return filename.includes("v0.4.21+commit.dfe3193c");
+      });
+
+      assert.equal(
```
haltman-at, line 176: (Hey @eggplantzzz looks like you inverted this one)

eggplantzzz: oh I did this algorithmically and didn't think about the value - `isUndefined` should be used

eggplantzzz: oh no that's also wrong, it should be defined :P

haltman-at: Well, you're just using `assert` and not `chai` here, right?  Node's builtin `assert` doesn't have `isDefined`, that's a `chai` thing...

eggplantzzz: oh I thought it was using chai...easily remedied

haltman-at (approved):
Woohoo!

eggplantzzz:
Thanks again @Neurone!

PR #5010, opened 2022-04-09
Dependency update: Bump ethereum-cryptography to 1.0.3

eggplantzzz:


cds-amal (commented):
Why pin this [dep](https://www.npmjs.com/package/ethereum-cryptography)?

eggplantzzz:
I thought we made it kind of a practice to pin external deps. That way we can be sure that no upgrades will break Truffle as has happened occasionally in the past.

haltman-at:
I don't think we have a general practice of pinning things?  I think we normally only pin things that need to be pinned or that have a history of breaking.

gnidan (approved):

PR #5011, opened 2022-04-09
Internal improvement: Remove hdkey from core

eggplantzzz:
The library hdkey seems to be unused in core. This PR removes it.


haltman-at (approved):
Thanks Tyler!

Issue #5006, opened 2022-04-08
Log spam while running tests in Github Action


TomiOhl:
## Issue

Tons of `Waiting for transaction signature. Please check your wallet for a transaction approval message.` log entries while running unit tests. This makes reading through the results pretty problematic, especially if there are many transactions in your tests or their before/beforeEach hooks.

## Steps to Reproduce

- write a smart contract and a test
- commit it to github, to a repository which runs the tests in a github action
- run the test and check the output

## Expected Behavior

See the output of truffle test (mocha).
![output_5 5 6](https://user-images.githubusercontent.com/6696659/162433951-9c46cd5f-0a4e-4e79-8c1e-b21a4b7feb9b.png)
(screenshot from Truffle 5.5.6)

## Actual Results

I see not only the usual test output, but "`Waiting for transaction signature. Please check your wallet for a transaction approval message.`" entries too.
![output_5 5 8](https://user-images.githubusercontent.com/6696659/162502025-c40d2461-d142-47fe-aa35-3f43539f2918.png)
(screenshot from Truffle 5.5.8)
Note that a similar issue was present in Truffle 5.5.7 when running locally too, but that was fixed in 5.5.8, thus I edited the issue.

## Environment

* Operating System: Ubuntu 20.04.3
* Ethereum client: Ganache v7.0.4
* Truffle version (`truffle version`): 5.5.8
* node version (`node --version`): 16

eggplantzzz:
Hmm, it looks like that logging is meant to be printed while using Truffle Dashboard to sign transactions. It shouldn't print when not, seems to be a bug. Thanks for reporting this @TomiOhl!

gnidan:
dang I was hoping this wouldn't be a problem. sorry about this; we'll get it fixed!

benjamincburns:
Thanks for the bug report @TomiOhl - this will very likely be fixed in the next Truffle release, assuming #5018 passes code review.

benjamincburns:
This was fixed as of the release one hour ago, so I'll close this.

TomiOhl:
Nice, just tested it and can confirm it's fixed. Thanks!

PR #5018, opened 2022-04-13
events: fix logspam from dashboard transaction spinner

benjamincburns:
addresses #5006

### DIFF
--- a/packages/events/defaultSubscribers/dashboard.js
+++ b/packages/events/defaultSubscribers/dashboard.js
@@ -8,9 +8,12 @@ module.exports = {
   handlers: {
     "rpc:request": [
       function (event) {
+        if (!isDashboardNetwork(this.config)) {
+          return;
+        }
+
         const { payload } = event;
         if (payload.method === "eth_sendTransaction") {
-          // TODO: Do we care about ID collisions?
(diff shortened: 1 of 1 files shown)


packages/events/defaultSubscribers/dashboard.js
```diff
@@ -9,8 +9,10 @@ module.exports = {
     "rpc:request": [
       function (event) {
         const { payload } = event;
-        if (payload.method === "eth_sendTransaction") {
-          // TODO: Do we care about ID collisions?
+        if (
+          isDashboardNetwork(this.config) &&
```
gnidan, line 13: heh sure you can make this wrapper :)

gnidan: by why not instead just check the negation at the very start and short circuit return early if it's not the dashboard network?

gnidan (commented):
quick question before i approve

packages/events/defaultSubscribers/dashboard.js
benjamincburns: happy to do that as well - one sec.

gnidan (approved):
nice and clean!

PR #5019, opened 2022-04-13
remove extra version fields

eggplantzzz:
I think during a merge conflict an extra `version` field was left in the encoder and db-kit packages. This PR removes the extra `version` field.


cds-amal (approved):
LGTM pending tests

PR #5021, opened 2022-04-14
fix slow ethers tests

benjamincburns:
Before this change the dashboard and dashboard provider tests took about 2 minutes each to run. Now they take around 20 seconds each.


cds-amal (approved):
LGTM. How much time does it save?

PR #5022, opened 2022-04-14
Internal improvement: Eliminate another for-in loop

haltman-at:
Did a review of other for-in loops in debugger and codec.  This one was the only one I found that seemed likely to cause trouble, so I'm getting rid of it pre-emptively.  @michaeljohnbennett let me know if you encounter problems with any others and I can get rid of those too, but this was the only one that seemed worth changing pre-emptively.


cds-amal (approved):
Looks good @haltman-at, and 200 points for the branch name!

Issue #5007, opened 2022-04-08
Debugger crashes with -x for contracts written for Solidity v0.4.4


gnidan:
## Issue
 
There's a TypeError for this transaction: 0xea524441c46fd574aac1d38fa83f3af3cbd6d27cb6e083cb4fa02edf1eee8604

## Steps to Reproduce

`truffle debug -x --network=mainnet 0xea524441c46fd574aac1d38fa83f3af3cbd6d27cb6e083cb4fa02edf1eee8604`

## Expected Behavior

I expect to be able to debug without source for this contract

## Actual Results

```
TypeError: Cannot read property 'language' of undefined
    at Function.normalize (/Users/fainashalts/truffle-dev/truffle/packages/debugger/lib/session/index.js:199:69)
    at fa.addExternalCompilations (/Users/fainashalts/truffle-dev/truffle/packages/debugger/lib/session/index.js:535:41)
    at DebugRecognizer.<anonymous> (/Users/gnidan/.nvm/versions/node/v14.19.0/lib/node_modules/truffle/build/webpack:/packages/fetch-and-compile/dist/debug.js:96:1)
    at Generator.next (<anonymous>)
    at /Users/gnidan/.nvm/versions/node/v14.19.0/lib/node_modules/truffle/build/webpack:/packages/fetch-and-compile/dist/debug.js:27:1
    at new Promise (<anonymous>)
    at exports.modules.687440.__awaiter (/Users/gnidan/.nvm/versions/node/v14.19.0/lib/node_modules/truffle/build/webpack:/packages/fetch-and-compile/dist/debug.js:23:1)
    at DebugRecognizer.addCompiledInfo (/Users/gnidan/.nvm/versions/node/v14.19.0/lib/node_modules/truffle/build/webpack:/packages/fetch-and-compile/dist/debug.js:92:1)
    at /Users/gnidan/.nvm/versions/node/v14.19.0/lib/node_modules/truffle/build/webpack:/packages/fetch-and-compile/dist/fetch.js:133:1
    at Generator.next (<anonymous>)
    at fulfilled (/Users/gnidan/.nvm/versions/node/v14.19.0/lib/node_modules/truffle/build/webpack:/packages/fetch-and-compile/dist/fetch.js:5:43)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
```

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.5.8
* node version (`node --version`):
[...]

PR #5024, opened 2022-04-14
Bug fix: Fix debugger crash on ancient Solidity versions

### LINKED ISSUE #1679: TypeError: Cannot read property 'legacyAST' of undefined
## Issue

Solidity compilers version 0.4.8 and older do not deliver a legacyAST as it is expected by truffle. This means that older sources cannot be tested anymore with the appropriate compiler version.

## Steps to Reproduce

```
mkdir test
cd test
truffle init
echo "module.exports = { compilers: { solc: { version: \"0.4.8\" } } }" > truffle-config.js 
echo "contract Test { }" > contracts/Test.sol 
cat - <<'end' > contracts/Migrations.sol
contract Migrations {
  address public owner;
  uint public last_completed_migration;
  modifier restricted() { if (msg.sender == owner) _; } 
  function Migrations() { owner = msg.sender; } 
  function setCompleted(uint completed) restricted { last_completed_migration = completed; }
  function upgrade(address new_address) restricted { Migrations upgraded = Migrations(new_address); upgraded.setCompleted(last_completed_migration);
  }
}
end
truffle test
```
## Expected Behavior

A truffle test run that does not throw exceptions.

## Actual Results

```
+ mkdir test
+ cd test
+ truffle init

✔ Preparing to download
✔ Downloading
✔ Cleaning up temporary files
✔ Setting up box

Unbox successful. Sweet!

Commands:

  Compile:        truffle compile
  Migrate:        truffle migrate
  Test contracts: truffle test

+ echo 'module.exports = { compilers: { solc: { version: "0.4.8" } } }'
+ echo 'contract Test { }'
+ cat -
+ truffle test
Compiling ./contracts/Migrations.sol...
Compiling ./contracts/Test.sol...

Compilation warnings encountered:

/Users/pieter/test/test/contracts/Migrations.sol:1:1: Warning: Source file does not specify required compiler version!Consider adding "pragma solidity ^0.4.8
contract Migrations {
^
Spanning multiple lines.
,/Users/pieter/test/test/contracts/Test.sol:1:1: Warning: Source file does not specify required compiler version!Consider adding "pragma solidity ^0.4.8
contract Test { }
^
Spanning multiple lines.

TypeError: Cannot read property 'legacyAST' of undefined
    at /Users/pieter/node_modules/truffle/build/webpack:/packages/truffle-compile/index.js:200:1
    at Array.forEach (<anonymous>)
    at /Users/pieter/node_modules/truffle/build/webpack:/packages/truffle-compile/index.js:187:1
    at Array.forEach (<anonymous>)
    at supplier.load.then.solc (/Users/pieter/node_modules/truffle/build/webpack:/packages/truffle-compile/index.js:184:1)
Truffle v5.0.2 (core: 5.0.2)
Node v10.15.0
```

## Environment

* Operating System: Darwin Kernel Version 18.2.0
* Ethereum client: built into truffle
* Truffle version (`truffle version`): v5.0.2 (core: 5.0.2)
* node version (`node --version`): v10.15.0
* npm version (`npm --version`):  6.4.1

haltman-at:
Addresses #5007.

This PR fixes a debugger crash that would occur when using `--fetch-external` to debug contracts written with Solidity versions prior to 0.4.9.

Now, we basically don't support Solidity versions that old... but ideally the debugger wouldn't crash, at least.  Especially as a person using `--fetch-external` doesn't necessarily know in advance which Solidity versions would be involved!

(Frankly I'm surprised that `compile-solidity` doesn't run into problems in such cases... I thought that's where the problem was...)

Because we don't really support Solidity that old, this PR isn't intended as any sort of full fix; it's basically just meant to patch over the problem.  Actually truly making such old versions work, issue #1679, is another matter.

Long story short, it turns out that the branch of the `shimContracts` function in codec that gets taken if `sources` or `files` is present relies on `sourcePath` also being present, but with these ancient Solidity versions it's missing (with the current `compile-solidity`, that is; I don't remember offhand if this would still necessarily be the case if we were to do something about #1679).  So, I added a check for `sourcePath`; if it's missing, we go down the other code path instead, because that one doesn't make use of `sourcePath`.  And then after that, things work out.

Ideally, of course, I'd fix the problem by going back even further, and actually getting `sourcePath` to be present; but since that comes from `compile-solidity`, I'd consider that to be a problem for when we handle #1679, and out of scope for this PR.  So, apologies if this doesn't go quite as far back as would be ideal.

### DIFF
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -145,7 +145,9 @@ export function shimContracts(
     //ast needs to be coerced because schema doesn't quite match our types here...
 
     //if files or sources was passed, trust that to determine the source index
-    //(assuming we have a sourcePath!)
+    //(assuming we have a sourcePath! currently it will be absent when dealing with
+    //Solidity versions <0.4.9; presumably we will fix this if we ever properly
+    //support versions that old, but for now this is necessary to get debug -x to work)
     if ((files || inputSources) && sourcePath) {
       //note: we never set the unreliableSourceOrder flag in this branch;
       //we just trust files/sources.  If this info is bad, then, uh, too bad.
@@ -266,12 +268,8 @@ export function getContractNode(
   contract: Contract,
   compilation: Compilation
 ): AstNode {
-  const {
-    contractName,
-    sourceMap,
-    deployedSourceMap,
-    primarySourceId
-  } = contract;
+  const { contractName, sourceMap, deployedSourceMap, primarySourceId } =
+    contract;
   const { unreliableSourceOrder, sources } = compilation;
 
   let sourcesToCheck: Source[];
@@ -545,7 +543,7 @@ export function collectUserDefinedTypesAndTaggedOutputs(
                 subNode.nodeType === "EventDefinition" ||
                 subNode.nodeType === "ErrorDefinition"
               ) {
-                  references[compilation.id][subNode.id] = subNode;
+                references[compilation.id][subNode.id] = subNode;
               }
             }
           }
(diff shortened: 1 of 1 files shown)


packages/codec/lib/compilations/utils.ts
```diff
@@ -145,9 +145,13 @@ export function shimContracts(
     //ast needs to be coerced because schema doesn't quite match our types here...
 
     //if files or sources was passed, trust that to determine the source index
-    if (files || inputSources) {
+    //(assuming we have a sourcePath!)
```
cds-amal, line 148: I think it's worth explaining the correlation between sourcePath and Solidity version as an extended comment.

cds-amal (approved):
LGTM. I do think the comment could be expanded to give further context. I'll leave that up to you as the information can be retrieved from the PR notes.

packages/codec/lib/compilations/utils.ts
dongmingh: I agree with Amal's comment of explaining the correlation between the two but up to your determination.

dongmingh (approved):

packages/codec/lib/compilations/utils.ts
haltman-at: OK, I'll add something quickly before merge.

PR #5027, opened 2022-04-15
Internal improvement: Re-fix sporadically failing encoder test

haltman-at:
So, #4932 was supposed to fix this, but didn't quite.  Hopefully this one actually does.  So why do I think that this one will work where the previous one failed?

Well, the problem with #4932 is that it assumed that the discrepancy would be at most a certain *absolute* size.  However, after looking at more counterexamples, it's clearer what's actually going on.  If there is a discrepancy, it always begins in the 15th decimal place after the first nonzero one.  It's the relative error, not the absolute error, that matters.

So why the 15th decimal place?  Well, why are discrepancies occurring at all?  Well, it's because we're converting from `Big` (decimal) to `number` (binary) and then back to `Big`.  That's necessarily going to be a bit lossy, especially as `number` has limited precision.  How much precision?  53 bits.  That's inbetween 15 and 16 decimal places.  Thus the problem; sometimes there isn't enough precision and things will slip a bit.  (Of course, things shouldn't slip any when dealing with integers, so I've altered the check to only allow this wiggle room when dealing with non-integers.)

Anyway I altered the test to go based on how far out the first differing decimal place is compared to the first nonzero decimal place, and check that it's at least 15 decimal places further out (or that they're equal).  The resulting code looks a bit wonky, because `Big`s are a bit wonky, but oh well.

Anyway hopefully this is fixed now!  As a check, I ran this test 512 times...

### DIFF
--- a/packages/encoder/test/wrap-fc.test.ts
+++ b/packages/encoder/test/wrap-fc.test.ts
@@ -312,17 +312,10 @@ const BoolInput = (value: boolean) => {
 };
 
 function testDecimalAgreementWithTolerance(actual: Big, expected: Big): void {
-  const negativeLog10OfRelativeTolerance = Math.floor(
-    Math.log10(Number.MAX_SAFE_INTEGER)
-  ); //15
-  const firstDecimalPlace = expected.e; //note: decreases, not increases, to the right of the decimal point
-  const firstDifferingDecimalPlace = actual.minus(expected).e; //note: will be 0 if difference is 0, so don't trust it in that case
-  const approximateNegativeLog10OfDifference =
-    firstDecimalPlace - firstDifferingDecimalPlace; //don't trust this if they're equal!
   assert(
-    actual.eq(expected) ||
-      approximateNegativeLog10OfDifference >= negativeLog10OfRelativeTolerance,
-    `Wrapped Big ${actual.toFixed()} was not within relative tolerance 10^-${negativeLog10OfRelativeTolerance} of original value ${expected.toFixed()}, difference approximately 10^-${approximateNegativeLog10OfDifference} of original value`
+    actual.eq(expected) || //next check will fail for zero, so need this one
+      actual.minus(expected).abs().lt(expected.abs().times(Number.EPSILON)),
+    `Wrapped Big ${actual.toFixed()} was not within relative epsilon of original value ${expected.toFixed()}`
   );
 }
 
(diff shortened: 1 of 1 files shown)


packages/encoder/test/wrap-fc.test.ts
```diff
@@ -313,6 +311,19 @@ const BoolInput = (value: boolean) => {
   return fc.oneof(...addBoxes(representations).map(fc.constant));
 };
 
+function testDecimalAgreementWithTolerance(actual: Big, expected: Big): void {
+  const negativeLog10OfRelativeTolerance = (2 ** 53).toString().length - 1; //15
```
eggplantzzz, line 315: What is the significance of this number?

haltman-at: As mentioned above, it's the precision of a `number` expressed in decimal places (approximately) as opposed to bits.

haltman-at: I guess really, instead of using `2 ** 53`, I should use the constant `Number.MAX_SAFE_INTEGER`.  I guess I'll go change when I have time.  (Also, arguably, instead of `num.toString().length - 1`, I should use `Math.floor(Math.log10(num))`... I might change that as well.)

haltman-at: OK, I've replaced this with an expression that is hopefully clearer.

benjamincburns (approved):
This is a bit confusing, but it fixes the problem so I'm inclined to approve.

That said, this feels more like an "epsilon problem." There's a constant (`Number.EPSILON`) to make these sort of comparisons easier when dealing with floating point errors, but I'm not certain that it's relevant in this case, as we're dealing with an equivalent "bit slippage" sort issue here w/ big number libs.

I leave it up to you as to whether or not it can/should be expressed in terms of `Number.EPSILON` or some other constant.

haltman-at:
Aha, thanks!  Yes, I think it's exactly an epsilon problem.  I'll try rewriting it like you suggest.  If for some reason that doesn't work, then I'll probably just switch it back to the current way... but I'll still use `Number.EPSILON`, rather than `Number.MAX_SAFE_INTEGER`, for clarity.  Thanks for remindng me that that constant exists!

haltman-at:
(And yeah, I reran that test 512 times once again to be sure...)

benjamincburns:
> (And yeah, I reran that test 512 times once again to be sure...)

Legend!

Issue #5026, opened 2022-04-14
Truffle instantiates duplicate providers after every contract in a deployment is mined, linear increase in API calls per contract deployed


tenthirtyone:
- [ ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

For every contract deployed by truffle, a new provider appears to be instantiated. This is a memory leak but it all results in a chattier application with unnecessary calls the the JSON RPC endpoint. 

## Steps to Reproduce

1. Create a new truffle project (MetaCoin works)
2. Pull down `ganache` (not required but helps illustrate the point)
3. Change 
`logging-options.ts L102  log: config.quiet ? () => {} : console.count`
4. In your migrations, Copy/Paste the deploy line so you are deploying multiple times:
![image](https://user-images.githubusercontent.com/6285466/163443497-7ad985fa-6a30-4a6b-a612-d4b2f9608fce.png)
5. Start a local ganache from cli with 8s block mining interval
6. define and deploy to the local ganache network

Tested against truffle stable and the `provider-changes` branch, confirmed on both. Logged output:

Logs on stable: https://gist.github.com/tenthirtyone/e9e30ec9d6a5520e7ed2d91a15908aee
Logs on provider-change: https://gist.github.com/tenthirtyone/a0d1d232984d4b8662012edda3d3bdec

## Expected Behavior

The number of JSON RPC endpoint calls between contract deploys should be consistent. 

## Actual Results

JSON RPC calls increase linearly for each contract deployed. 

Truffle Stable RPC call growth per contract deployed:
![image](https://user-images.githubusercontent.com/6285466/163443893-4914466b-fde5-402f-a016-2088363a2e31.png)

`provider-changes` RPC call growth per contract deployed:
![image](https://user-images.githubusercontent.com/6285466/163443935-7957d997-0f43-4ab5-b85d-61fc44232a8e.png)

I measured the number of `eth_getBlockByNumber` calls made between every contract deployment

E.g.
RPC calls made between contract 1 & contract 2 and a constant block mine interval of 8s: 45
RPC calls made between contract 2 & contract 3 and a constant block mine interval of 8s: 70
RPC calls made between contract 3 & contract 4 and...: 94

The change is consistently at 24.

This contributes to why users experience the polling timeout and soft caps us on the number of contracts we can deploy per migration file. I have not run this test against multiple migration files. 

## Environment

* Operating System: UBUNTU
[...]

Issue #4814, opened 2022-03-09
Connection issues during large migrations


venkatesh16031999:
- [ ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

What the issue is, in broad strokes.

1. We are using a upgradeable proxy contract in truffle suite to deploy and upgrade using proxy patterns, when we deploy new proxy contract it is getting successful, but when we upgrade any contract and try to deploy an upgraded contract, it is getting failed every time. 
2. We have tried deploying on both Polygon Mainnet and Testnet
3. We have also tried by changing the RPC. Nothing seems to work.

## Steps to Reproduce

1. Deploy a proxy upgradeable contract using Openzeppelin SDK
2. Update the contract code
3. Deploy the upgraded Proxy contract, error will thrown / [object Object] output will be displayed instead of deploying the contract

Please provide the shortest amount of steps to reproduce your issue.

## Expected Behavior

Contract should be deployed and transaction hash and other summary should be resulted as output

What you expected to happen.

## Actual Results

Attaching the error logs for the reference
1. https://mclo.gs/29w2HoT
2. https://mclo.gs/lhsa0Id

What actually happened. Please give examples and support it with screenshots, copied output or error messages.

## Environment
Windows Operating System Environment

* Operating System: 
* Ethereum client: Polygon | Web3.js
* Truffle version (`truffle version`): Truffle v5.4.16 (core: 5.4.16)
* node version (`node --version`): Node v14.17.4
* npm version (`npm --version`): 6.14.14

MicaiahReid:
@venkatesh16031999 Do you have any additional information that could point to this being a problem with Truffle? People tend to have a hard time with deployments to Polygon mainnet, so this could be related to that. Also, we've seen similar errors to the one you're showing when people have connection issues.

Regardless, @tenthirtyone will do some investigating on this issue. Thanks for opening it!

tenthirtyone:
@venkatesh16031999 Hey there, can you confirm if you are following the instructions [here](https://docs.polygon.technology/docs/develop/truffle). 

If not, can you please share your truffle network config you are using for matic? It will look similar to this:

```
matic: {
      provider: () => new HDWalletProvider(mnemonic, `https://rpc-mumbai.maticvigil.com`),
      network_id: 80001,
      confirmations: 2,
      timeoutBlocks: 200,
      skipDryRun: true
    },
```

[...]

Issue #4757, opened 2022-02-19
How to prevent PollingBlockTracker from crashing my nodejs


stranger-games:
- I've opened a support ticket before filing this issue.

---------------------------

## Issue

RPC endpoint errors causing nodejs process to crash.
I am connecting to HDWalletProvider on a nodejs backend. It's typical to receive once in a while an error from RPC end point like time out or PollingBlockTracker. This is understandable and not the issue. The issue is that any RPC error cause my nodejs server to crash.

Is there a way to catch RPC errors to prevent crashing and handle it by skipping the block or retrying or whatever.

Please note that this issue is not about the timeout error or polling block tracker error. This issue is about the process crashing when those errors occur.

## Steps to Reproduce

Simple initialization of web3 with RPC end point like this.
`const web3_56 = new Web3(new HDWalletProvider(process.env.PRIVATE_PRODUCTION, `${process.env.BSC_MAINNET_NODE}`));//bscmainnet
`
Then initializing contracts like this. If the contract has some events the errors are more likely to occur.

`var contract = new web3.eth.Contract(abi, address);`

Leave the process even without activity until some kind of RPC end point error occurs (timeout or polling block tracker or whatever).

## Expected Behavior

I expect to have a way to catch those errors and decide in code what to do without having my nodejs process crash.

## Actual Results

My nodejs process crash if any RPC error occurs.

> C:\prj\server\node_modules\safe-event-emitter\index.js:74
>       throw err
>       ^
> 
> Error: PollingBlockTracker - encountered an error while attempting to update latest block:
> undefined
>     at PollingBlockTracker._performSync (C:\prj\server\node_modules\eth-block-tracker\src\p    at runMicrotasks (<anonymous>)
> Emitted 'error' event on Web3ProviderEngine instance at:
>     at safeApply (C:\prj\server\node_modules\safe-event-emitter\index.js:70:5)
>     at PollingBlockTracker.SafeEventEmitter.emit (C:\prj\server\node_modules\safe-event-emitter\index.js:56:5)
>     at PollingBlockTracker._performSync (C:\prj\server\node_modules\eth-block-tracker\src\polling.js:53:16)
>     at runMicrotasks (<anonymous>)
>     at processTicksAndRejections (internal/process/task_queues.js:95:5)

## Environment

* Operating System: 
* Windows
* Ethereum client:
* hdwallet-provider
* Truffle version:
* @truffle/hdwallet-provider": "^2.0.0
* node version (`node --version`):
[...]

Issue #4288, opened 2021-08-28
Can not deploy to Binance Smart Chain


aliveli186:
`truffle migrate --network bsctestnet`
fails with the following error message:
```
> Compiled successfully using:
   - solc: 0.8.7+commit.e28d00a7.Emscripten.clang


C:\Projects\node_modules\request\request.js:848
          var e = new Error('ETIMEDOUT')
                  ^
Error: PollingBlockTracker - encountered an error while attempting to update latest block:
Error: ETIMEDOUT
    at Timeout.<anonymous> (C:\Projects\node_modules\request\request.js:848:19)
    at listOnTimeout (internal/timers.js:557:17)
    at processTimers (internal/timers.js:500:7)
    at PollingBlockTracker._performSync (C:\Projects\node_modules\eth-block-tracker\src\polling.js:51:24)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at runNextTicks (internal/process/task_queues.js:64:3)
    at listOnTimeout (internal/timers.js:526:9)
    at processTimers (internal/timers.js:500:7)
UnhandledRejections detected
Promise {
  <rejected> {
    code: -32603,
    message: 'ETIMEDOUT',
    data: { originalError: [Object] },
    stack: 'Error: ETIMEDOUT\n' +
      '    at Timeout.<anonymous> (F:\\Projects\\Blockchain\\node_modules\\request\\request.js:848:19)\n' +
      '    at listOnTimeout (internal/timers.js:557:17)\n' +
      '    at processTimers (internal/timers.js:500:7)'
  }
}
```

`truffle migrate --network ropsten` works.

The only difference is the one which fails is a HTTPS the other is a WebSocket address.

Tool Versions:
```
Truffle v5.4.5 (core: 5.4.5)
Solidity - 0.8.7 (solc-js)
Node v14.17.5
Web3.js v1.5.1
npm v6.14.14
```

truffle-config.js:
```
    ropsten: {
      provider: () => new HDWalletProvider(mnemonic, `wss://ropsten.infura.io/ws/v3/${infuraKey}`),
      websockets: true,
      network_id: 3,       // Ropsten's id
      gas: 5500000,        // Ropsten has a lower block limit than mainnet
      confirmations: 0,    // # of confs to wait between deployments. (default: 0)
      timeoutBlocks: 200,  // # of blocks before a deployment times out  (minimum/default: 50)
[...]

PR #5029, opened 2022-04-18
Partial memoize provider to reduce number of RPC calls on deployments

tenthirtyone:
### Description
From the investigating that has stemmed from #4814, this PR reduces the number of RPC calls during a migration by 80% - 90%. Working with @cds-amal, we've identified places in the config that will require optimization because parts of code depend on instantiating new providers. 

The issue was originally reported by a user deploying to Polygon's Mumbai testnet. (https://github.com/trufflesuite/truffle/issues/4814)

### Steps to Reproduce
1. Create an Infura account
2. `init` a truffle project, this test was run against truffle version `5.5.3`
3. Define a `matic` network pointing to your Infura endpoint using the recommended `matic` config.
``` javascript
matic: {
      provider: () => new HDWalletProvider(mnemonic, `https://polygon-mumbai.infura.io/v3/[INFURA_PROJECT_KEY]`),
      network_id: 80001,
      confirmations: 2,
      timeoutBlocks: 200,
      skipDryRun: true
    },
```
4. Run `truffle migrate --network matic`
5.  Observe the API calls:
image

6. Pull down and checkout this branch
7. `use-truffle-core` to set system truffle to project folder & this branch
8. Run `truffle migrate --network matic` again
9. Observe the API calls:
image

ER: API calls drop from ~400 (391 actual) to ~150 (151 actual)

10. Set Truffle back to stable `use-truffle-stable`
11. Change your `matic` network in truffle config:
``` javascript
      network_id: 80001,
      confirmations: 0,
      disableConfirmationListener: true,
      timeoutBlocks: 200,
      pollingInterval: 60000,
      skipDryRun: true,
    },
```
12. Run `truffle migrate --network matic`
13. Observe the API calls:
image
 
14. Set Truffle back to core `use-truffle-core`
15. Run `truffle migrate --network matic`
16. Observe the API calls:
image

ER: API calls drop from 224 to 54. ~87% reduction from the first test (54 vs 391).

### Related Issues
This new issue was discovered during work on this PR:
https://github.com/trufflesuite/truffle/issues/5026

These open issues mention similar problems:
https://github.com/trufflesuite/truffle/issues/852
https://github.com/trufflesuite/truffle/issues/3356
https://github.com/trufflesuite/truffle/issues/3468
https://github.com/trufflesuite/truffle/issues/3646
https://github.com/trufflesuite/truffle/issues/3471
https://github.com/trufflesuite/truffle/issues/4757

These closed issues may be related:
https://github.com/trufflesuite/truffle/issues/3200
https://github.com/trufflesuite/truffle/issues/4013
https://github.com/trufflesuite/truffle/issues/4288

### DIFF
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -2,7 +2,11 @@ import * as path from "path";
 import Provider from "@truffle/provider";
 import TruffleConfig from "./";
 
-let provider: any;
+let provider: any = null;
+
+export const resetProvider = () => {
+  provider = null;
+};
 
 export const getInitialConfig = ({
   truffleDirectory,
@@ -293,6 +297,8 @@ export const configProps = ({
         if (!provider) {
           console.log("No memo provider");
           provider = Provider.create(options);
+        } else {
+          console.log("Using memo provider");
         }
         return provider;
 
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -295,14 +295,9 @@ export const configProps = ({
         options.events = configObject.events;
 
         if (!provider) {
-          console.log("No memo provider");
           provider = Provider.create(options);
-        } else {
-          console.log("Using memo provider");
         }
         return provider;
-
-        //return Provider.create(options);
       },
       set() {
         throw new Error(
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -287,7 +287,7 @@ export const configProps = ({
         const options = configObject.network_config;
         options.verboseRpc = configObject.verboseRpc;
         options.events = configObject.events;
-
+        console.log(new Error("valuevaluevalue").stack);
         return Provider.create(options);
       },
       set() {
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -2,6 +2,8 @@ import * as path from "path";
 import Provider from "@truffle/provider";
 import TruffleConfig from "./";
 
+let provider: any;
+
 export const getInitialConfig = ({
   truffleDirectory,
   workingDirectory,
@@ -287,8 +289,14 @@ export const configProps = ({
         const options = configObject.network_config;
         options.verboseRpc = configObject.verboseRpc;
         options.events = configObject.events;
-        console.log(new Error("valuevaluevalue").stack);
-        return Provider.create(options);
+
+        if (!provider) {
+          console.log("No memo provider");
+          provider = Provider.create(options);
+        }
+        return provider;
+
+        //return Provider.create(options);
       },
       set() {
         throw new Error(
--- a/packages/config/test/methods.test.ts
+++ b/packages/config/test/methods.test.ts
@@ -38,6 +38,28 @@ describe("TruffleConfig.detect", () => {
   });
 });
 
+describe(".load", () => {
+  const config1 = `${__dirname}/testConfig.js`;
+  const config2 = `${__dirname}/testConfig2.js`; // identical to config1
+  const config3 = `${__dirname}/testConfig3.js`; // different mnemonic from 1&2
+
+  it("providers should not be ===", () => {
+    const Config1 = TruffleConfig.load(config1);
+    const Config2 = TruffleConfig.load(config2);
+    const Config3 = TruffleConfig.load(config3);
+
+    const prov1 = Config1.networks.develop.provider();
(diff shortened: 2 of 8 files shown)


packages/core/lib/networks.js
```diff
@@ -198,13 +198,14 @@ const Networks = {
...
   // Try to connect to every named network except for "test" and "development"
-  asURIs: async function(options, networks) {
+  asURIs: async function (options, networks) {
     const result = {
       uris: {},
       failed: []
     };
 
     for (const networkName of networks) {
+      console.log("Networks asURIs");
```
cds-amal, line 208: Missed one in the log culling!

cds-amal (commented):
We can use Ganache to validate this work to show improved network usage in a dev friendlier way imho.

packages/config/src/configDefaults.ts
```diff
@@ -2,6 +2,12 @@ import * as path from "path";
 import Provider from "@truffle/provider";
 import TruffleConfig from "./";
 
+let provider: any = null;
+
+export const resetProvider = () => {
+  provider = null;
+};
+
```
davidmurdoch, line 10: I believe a test like this would now fail:

```
const config1 = Config.load(pathToConfig);
const config2 = Config.load(pathToOtherConfig);
assert.notStrictEqual(config1.networks.develop.provider, config2.networks.develop.provider); // this fails, no?
```

I think we should treat Truffle as a whole as a library, not only as a CLI tool with short lived commands. `@truffle/config` _is_ a major versioned (stable API) library now (with 31000+ weekly downloads): https://www.npmjs.com/package/@truffle/config

I think the memoization needs to be solved another way.

packages/db/src/project/test/integration.spec.ts
```diff
@@ -54,37 +54,6 @@ const compilationConfig = {
   all: true
 };
 
-const migratedArtifacts = [
-  require(path.join(
-    __dirname,
-    "compilationSources",
-    "build",
-    "contracts",
```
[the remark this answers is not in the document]
tenthirtyone: Good call: #5062 is up. I'll yank this out

packages/config/src/configDefaults.ts
eggplantzzz: Can you elaborate a bit further? I'm not sure I'm following you here, why would that fail and what does that have to do with your comments about @truffle/config having a lot of downloads and treating Truffle as a whole library?

tenthirtyone: This is great and I think it's the exact kind of insight we need from the team suss these things out. I will make some additional unit tests to cover these cases and it would be helpful to keep addressing them as this saga continues. 

Since a lot of these libs were built organically and issues were addressed as they came in, this would really help us get a bigger picture of the tradeoffs so we can perhaps formalize something. E.g. 1 dev did _this_ and another did _that_ and now to a change to this impacts that, or vice versa and it reaches deadlock.

I'm convinced that this is impossible to solve with a silver bullet that satisfies every implementation that is in place and we'll have to get the big picture if we want to solve this problem.

eggplantzzz: Ok, is that a problem?

cds-amal: Yea, it would be bad. David raised a good issue. I'm personally guilty, of thinking `@truffle/config` is an internal project.

tenthirtyone:
This will require #5062 now

packages/config/test/methods.test.ts
```diff
@@ -38,6 +38,28 @@ describe("TruffleConfig.detect", () => {
   });
 });
 
+describe(".load", () => {
+  const config1 = `${__dirname}/testConfig.js`;
+  const config2 = `${__dirname}/testConfig2.js`; // identical to config1
+  const config3 = `${__dirname}/testConfig3.js`; // different mnemonic from 1&2
+
+  it("providers should not be ===", () => {
```
tenthirtyone, line 46: @davidmurdoch I'm seeing the same results for this on `develop` and on this branch.

davidmurdoch: I thought provider was a getter, not a function, per https://github.com/trufflesuite/truffle/pull/5029/files#diff-180639c74cbd9b1cd8f6bd257ccdfd81d6ea56545c59f9e93fb76f3225f661d2R288-R301

cds-amal:
@tenthirtyone can you rebase this PR's actual changes against develop? This PR reports 54 commits across 60 files, which doesn't seem right.

tenthirtyone:
@eggplantzzz nope. We can close

Issue #4807, opened 2022-03-07
Setting Ganache options for `truffle test` and `truffle develop` should be consistent and correct


haltman-at:
## Issue

Follow-up to #3147.  It is now possible to set Ganache options for `truffle test` with the `test` network; however, the way this works is inconsistent with how `truffle develop` does it, and has some problems.  Moreover, the way that `truffle develop` does it also has problems.  The two of these should be made consistent with one another, and also better.  Specifically:

1. Currently, `truffle test` simply dumps all the options from the test network into the Ganache options (with one exception, to be discussed below).  We might want to possibly restrict these in some way, to only valid Ganache options?  Not sure of this.  (Will Ganache cause an error on getting options like `skipDryRun`?  Are there *valid* options that could cause problems as they have a different intention?)

2. By contrast, `truffle develop` only allows specific fields; we should check whether there are additional fields it should allow, and whether it should possibly (like `truffle test`) simply dump in all the options.

3. However, `truffle develop` also renames some fields.  One passes in `accounts` rather than `total_accounts`, `defaultEtherBalance` rather than `default_balance_ether`, `genesis_time` rather than `time`, and `gas` rather than `gasLimit`.  Let's ignore that last one for now as it's got its own problems.  For the others, both `truffle develop` and `truffle test` should accept *both* names for each of these; like, you should be able to pass in either `accounts` or `total_accounts` and have it work, both for `truffle develop` and for `truffle test`.

4. Actually, `truffle test` does accept *one* synonym... it accepts `genesis_time` as a synonym for `time`... *sort of*.  In reality, this doesn't work, because it looks at `config.genesis_time` rather than `config.networks.test.genesis_time` (and no, there's no getter for this that would cause it to work).  So, this bug should be fixed (and also the code should be moved to live with the rest of where the user settings are set; currently it's some distance away which is confusing).

5. Finally, in `truffle develop`, `gas` should **not** be accepted as a synonym for `gasLimit`.  This is because the `gas` field is supposed to set the default *transaction* gas limit, whereas Ganache's `gasLimit` option sets the *block* gas limit, an entirely different thing.  @fainashalts already fixed this bug in `truffle test` back in #4323; it should be fixed in `truffle develop` as well.

## Environment

* Truffle version (`truffle version`): 5.5.3

sukanyaparashar:
[...]

PR #2236, opened 2019-07-19
Feature: Enable `genesis_time` config option for `truffle develop` & `truffle test`

krsnnik:
Feature request #1891 

Allows users to specify a timestamp through the `genesis_time` option in a truffle-config file when running `truffle test` & `truffle develop`. Will warn the users if an invalid date has been passed and default to using the current timestamp.

@CruzMolina held my hand through this first PR to trufflesuite, couldn't have done it without him.

PR #5030, opened 2022-04-18
Refactor truffle develop and truffle test codes

sukanyaparashar:
### ISSUE
Setting Ganache options for `truffle test` and `truffle develop `are not consistent and correct. See #4807 .

### SOLUTION
A new file `configureGanacheOptions.js` is added which consists of an adapter function `getGanacheOptions` which can be  used by both `truffle test` and `truffle develop` to configure managed **Ganache**.

### DIFF
--- a/packages/core/lib/commands/develop/run.js
+++ b/packages/core/lib/commands/develop/run.js
@@ -1,5 +1,6 @@
 const emoji = require("node-emoji");
 const mnemonicInfo = require("../../mnemonics/mnemonic");
+const configureGanacheOptions = require("../configureGanacheOptions");
 
 const runConsole = async (config, ganacheOptions) => {
   const Console = require("../../console");
@@ -28,7 +29,7 @@ module.exports = async options => {
   const customConfig = config.networks.develop || {};
 
   const { mnemonic, accounts, privateKeys } = mnemonicInfo.getAccountsInfo(
-    customConfig.accounts || 10
+    customConfig.accounts || customConfig.total_accounts || 10
   );
 
   const onMissing = () => "**";
@@ -39,41 +40,11 @@ module.exports = async options => {
     "Ensure you do not use it on production blockchains, or else you risk losing funds.";
 
   const ipcOptions = { log: options.log };
-
-  const ganacheOptions = {
-    host: customConfig.host || "127.0.0.1",
-    port: customConfig.port || 9545,
-    network_id: customConfig.network_id || 5777,
-    total_accounts: customConfig.accounts || 10,
-    default_balance_ether: customConfig.defaultEtherBalance || 100,
-    blockTime: customConfig.blockTime || 0,
-    fork: customConfig.fork,
-    mnemonic,
-    gasLimit: customConfig.gas || 0x6691b7,
-    gasPrice: customConfig.gasPrice || 0x77359400,
-    time: config.genesis_time
-  };
-
-  if (customConfig.hardfork !== null && customConfig.hardfork !== undefined) {
-    ganacheOptions["hardfork"] = customConfig.hardfork;
-  }
-
-  function sanitizeNetworkID(network_id) {
-    if (network_id !== "*") {
-      if (!parseInt(network_id, 10)) {
-        const error =
-          `The network id specified in the truffle config ` +
-          `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
-        throw new Error(error);
-      }
-      return network_id;
-    } else {
-      // We have a "*" network. Return the default.
-      return 5777;
-    }
-  }
-
-  ganacheOptions.network_id = sanitizeNetworkID(ganacheOptions.network_id);
+  const ganacheOptions = configureGanacheOptions.getGanacheOptions(
+    config,
+    customConfig,
+    mnemonic
+  );
 
   const { started } = await Develop.connectOrStart(ipcOptions, ganacheOptions);
   const url = `http://${ganacheOptions.host}:${ganacheOptions.port}/`;
--- a/packages/core/lib/commands/test/run.js
+++ b/packages/core/lib/commands/test/run.js
@@ -1,5 +1,3 @@
-const defaultNetworkIdForTestCommand = 4447;
-
 const parseCommandLineFlags = options => {
   // parse out command line flags to merge in to the config
   const grep = options.grep || options.g;
@@ -17,30 +15,13 @@ const parseCommandLineFlags = options => {
     : { mocha: { grep, bail, reporter } };
 };
 
-function sanitizeGanacheOptions(ganacheOptions) {
-  const network_id = ganacheOptions.network_id;
-
-  // Use default network_id if "*" is defined in config
-  if (network_id === "*") {
-    return { ...ganacheOptions, network_id: defaultNetworkIdForTestCommand };
-  }
-
-  const parsedNetworkId = parseInt(network_id, 10);
-  if (isNaN(parsedNetworkId)) {
-    const error =
-      `The network id specified in the truffle config ` +
-      `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
-    throw new Error(error);
-  }
-  return { ...ganacheOptions, network_id: parsedNetworkId };
-}
-
 module.exports = async function (options) {
   const Config = require("@truffle/config");
   const { Environment, Develop } = require("@truffle/environment");
   const { copyArtifactsToTempDir } = require("./copyArtifactsToTempDir");
   const { determineTestFilesToRun } = require("./determineTestFilesToRun");
   const { prepareConfigAndRunTests } = require("./prepareConfigAndRunTests");
+  const configureGanacheOptions = require("../configureGanacheOptions");
 
   const optionsToMerge = parseCommandLineFlags(options);
   const config = Config.detect(options).merge(optionsToMerge);
@@ -105,31 +86,18 @@ module.exports = async function (options) {
     (testNetworkDefinedAndUsed && noProviderHostOrUrlConfigured) ||
     !configuredNetwork
   ) {
-    // Use managed ganache with overriding user specified config or without any specification in the config
-    const defaultPort = await require("get-port")();
-    const defaultHost = "127.0.0.1";
     const defaultMnemonic =
       "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
-
-    // configuredNetwork will spread only when it is defined and ignored when undefined
-    const ganacheOptions = {
-      host: defaultHost,
-      port: defaultPort,
-      network_id: defaultNetworkIdForTestCommand,
-      mnemonic: defaultMnemonic,
-      time: config.genesis_time,
-      miner: {
-        instamine: "strict"
-      },
-      ...configuredNetwork
-    };
-
-    const sanitizedGanacheOptions = sanitizeGanacheOptions(ganacheOptions);
+    const ganacheOptions = configureGanacheOptions.getGanacheOptions(
+      config,
+      configuredNetwork,
+      defaultMnemonic
+    );
 
     const ipcOptions = { network: "test" };
     numberOfFailures = await startGanacheAndRunTests(
       ipcOptions,
-      sanitizedGanacheOptions,
+      ganacheOptions,
       config
     );
   } else {
--- a/packages/core/lib/commands/configureGanacheOptions.js
+++ b/packages/core/lib/commands/configureGanacheOptions.js
@@ -1,0 +1,52 @@
+const defaultNetworkIdForGanache = 5777;
+
+function sanitizeGanacheOptions(ganacheOptions) {
+  const network_id = ganacheOptions.network_id;
+
+  // Use default network_id if "*" is defined in config
+  if (network_id === "*") {
+    return { ...ganacheOptions, network_id: defaultNetworkIdForGanache };
+  }
+
+  const parsedNetworkId = parseInt(network_id, 10);
+  if (isNaN(parsedNetworkId)) {
+    const error =
+      `The network id specified in the truffle config ` +
+      `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
+    throw new Error(error);
+  }
+  return { ...ganacheOptions, network_id: parsedNetworkId };
+}
+
+const ConfigureGanache = {
+  getGanacheOptions: function (config, customConfig, mnemonic) {
+    const ganacheOptions = {
+      host: customConfig.host || "127.0.0.1", // Default host for managed ganache
+      port: customConfig.port || 9545, // Default port for managed ganache
+      network_id: customConfig.network_id || defaultNetworkIdForGanache,
+      total_accounts:
+        customConfig.accounts || customConfig.total_accounts || 10,
+      default_balance_ether:
+        customConfig.defaultEtherBalance ||
+        customConfig.default_balance_ether ||
+        100,
+      blockTime: customConfig.blockTime || 0,
+      fork: customConfig.fork,
+      mnemonic: mnemonic,
+      gasPrice: customConfig.gasPrice || 0x77359400,
+      time: config.genesis_time,
+      miner: {
+        instamine: "strict"
+      }
+    };
+
+    if (customConfig.hardfork !== null && customConfig.hardfork !== undefined) {
+      ganacheOptions["hardfork"] = customConfig.hardfork;
+    }
+
+    const sanitizedGanacheOptions = sanitizeGanacheOptions(ganacheOptions);
+    return sanitizedGanacheOptions;
+  }
+};
+
+module.exports = ConfigureGanache;


packages/core/lib/commands/develop/run.js
```diff
@@ -28,7 +29,7 @@ module.exports = async options => {
   const customConfig = config.networks.develop || {};
 
   const { mnemonic, accounts, privateKeys } = mnemonicInfo.getAccountsInfo(
-    customConfig.accounts || 10
+    customConfig.accounts || customConfig.total_accounts || 10
```
haltman-at, line 32: Aha, the synonyms! :)

One note here -- and I realize that this problem is pre-existing and not due to your changes, but as long as you're getting the rest of this, you might want to handle this here -- this code means that `accounts` is zero, it'll end up as 10 instead.  Perhaps this should be changed so that an input of `0` will be respected?  (There isn't some problem with this, is there @davidmurdoch?)

packages/core/lib/commands/test/run.js
```diff
@@ -91,7 +72,7 @@ module.exports = async function (options) {
     inputFile: file
   });
 
-  const configuredNetwork = config.networks[config.network];
+  let configuredNetwork = config.networks[config.network];
```
haltman-at, line 75: Is there some reason this was changed from `const` to `let`?  I'm not seeing one.

packages/core/lib/configureGanacheOptions.js
```diff
@@ -0,0 +1,48 @@
...
+  return { ...ganacheOptions, network_id: parsedNetworkId };
+}
+
+function configureManagedGanache(config, networkConfig, mnemonic) {
+  const ganacheOptions = {
+    host: networkConfig.host || "127.0.0.1", // Default host for managed ganache
+    port: networkConfig.port || 9545, // Default port for managed ganache
+    network_id: networkConfig.network_id || Date.now(),
+    total_accounts:
+      networkConfig.accounts || networkConfig.total_accounts || 10,
```
haltman-at, line 25: Same note here as above about respecting zeroes.

packages/core/lib/configureGanacheOptions.js
```diff
@@ -0,0 +1,48 @@
...
+    throw new Error(error);
+  }
+  return { ...ganacheOptions, network_id: parsedNetworkId };
+}
+
+function configureManagedGanache(config, networkConfig, mnemonic) {
+  const ganacheOptions = {
+    host: networkConfig.host || "127.0.0.1", // Default host for managed ganache
+    port: networkConfig.port || 9545, // Default port for managed ganache
+    network_id: networkConfig.network_id || Date.now(),
```
haltman-at, line 23: ~~Hm, it looks to me like this doesn't sanitize out the case of `network_id = "*"`... that has to be handled, doesn't it?  Or did I just miss where that happens?~~ Nevermind, you handled that separately!  Wondering if it makes sense to combine those?  The way you have it is fine though I think.

packages/core/lib/configureGanacheOptions.js
```diff
@@ -0,0 +1,48 @@
...
+    total_accounts:
+      networkConfig.accounts || networkConfig.total_accounts || 10,
+    default_balance_ether:
```
haltman-at, line 26: This is another case where zeroes should presumably be respected.

packages/core/lib/configureGanacheOptions.js
```diff
@@ -0,0 +1,48 @@
...
+      networkConfig.defaultEtherBalance ||
+      networkConfig.default_balance_ether ||
+      100,
+    blockTime: networkConfig.blockTime || 0,
+    fork: networkConfig.fork,
+    mnemonic: mnemonic,
+    gasPrice: networkConfig.gasPrice || 0x77359400,
+    time: networkConfig.genesis_time || config.genesis_time,
```
haltman-at, line 34: I think we can probably omit `config.genesis_time`... I guess that's technically breaking, but the whole thing is kind of a bug and I doubt people are actually using it.  Although, IDK, maybe it's best to leave it in just for safety.  (@gnidan, any opinion by any chance?)  I guess probably it's best to leave it in after all, but it might be worth adding a comment noting that this is for compatibility with old bugged versions. :P

packages/core/lib/configureGanacheOptions.js
```diff
@@ -0,0 +1,48 @@
...
+    total_accounts:
+      networkConfig.accounts || networkConfig.total_accounts || 10,
+    default_balance_ether:
+      networkConfig.defaultEtherBalance ||
+      networkConfig.default_balance_ether ||
+      100,
+    blockTime: networkConfig.blockTime || 0,
+    fork: networkConfig.fork,
+    mnemonic: mnemonic,
+    gasPrice: networkConfig.gasPrice || 0x77359400,
```
haltman-at, line 33: Another possible place where zeroes should be respected?  I don't know how that will affect things though (@davidmurdoch, do you know?)

packages/core/lib/configureGanacheOptions.js
haltman-at: Also, `time` should be allowed here as a synonym, not just `genesis_time`.  That is something that should actually change.

packages/core/lib/configureGanacheOptions.js
```diff
@@ -0,0 +1,48 @@
...
+    const error =
+      `The network id specified in the truffle config ` +
+      `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
+    throw new Error(error);
+  }
+  return { ...ganacheOptions, network_id: parsedNetworkId };
+}
+
+function configureManagedGanache(config, networkConfig, mnemonic) {
+  const ganacheOptions = {
```
haltman-at, line 20: I notice `gasLimit` is missing... that needs to be added.  We shouldn't accept `gas` as a synonym for it, but we should accept `gasLimit`.

haltman-at (commented):
Thanks for getting this Sukanya!  I have a number of nitpicks, hope you don't mind...

packages/core/lib/commands/test/run.js
sukanyaparashar: Oh! I forgot to change that. Previously I was trying to alter the `configuredNetwork` later in the code for some testing purpose and so changed it from `const` to `let`. I will change it to `const` again. Thanks for pointing that out @haltman-at .

packages/core/lib/configureGanacheOptions.js
sukanyaparashar: I tried omitting `config.genesis_time` and the CI failed in a few tests which pointed out invalid time if there is no `genesis_time` defined in the config. So, I added that again. Not exactly sure about the behavior though. May be @davidmurdoch can better explain that.

I will add `time` as a synonym here. Thanks.

packages/core/lib/configureGanacheOptions.js
sukanyaparashar: Sure, I will add that. May be I misunderstood that it should be removed like it was previously done in `truffle test`.

packages/core/lib/configureGanacheOptions.js
haltman-at: Oh hm.  OK, guess we better not break that then, yikes!

packages/core/lib/commands/develop/run.js
sukanyaparashar: I am not very sure of this. But in my opinion if `accounts` is `0`, then the commands like `test` and `migrate` won't work as those commands need at least one account to work as expected.

davidmurdoch: As of 7.0 ganache _allows_ 0 accounts on startup, and users can then add others later via RPC. They can also start ganache with unlocked accounts, though I don't know if truffle allows that option to be passed through from truffle config to ganache.

haltman-at: > I am not very sure of this. But in my opinion if accounts is 0, then the commands like test and migrate won't work as those commands need at least one account to work as expected.

I mean, so what?  To my mind, if the user requests an operation that will error, they should get the operation and error they requested!

sukanyaparashar: Okay, sure. I agree with you @haltman-at . We should put checks for these and throw the errors as required.

packages/core/lib/configureGanacheOptions.js
sukanyaparashar: @davidmurdoch, Is there any minimum `gasPrice` requirement for **Ganache** network ? I see the default is set to 2000000000.

packages/core/lib/commands/develop/run.js
haltman-at: Oh, I wasn't suggesting putting a check... I was suggesting we just let the consequences play out if that's what the user really wants. :)  After all, Truffle Develop and Truffle Test don't necessarily *require* accounts, right?  They *usually* do, but not always, so I can imagine someone wanting to set it to 0...

But maybe a check is a good idea...?  IDK, @gnidan, any opinion here?

sukanyaparashar: I was thinking of throwing appropriate errors for each case (`accounts`, `defaultEtherBalance` and `gasPrice`) so that user becomes aware of the managed ganache network requirements. Any suggestions on this @cds-amal @eggplantzzz @davidmurdoch ?

haltman-at: Is it always true that they won't work, though?

I'm not sure what the effect of `gasPrice: 0` would be.  Does it just mean gas is free?

Setting `defaultEtherBalance` to 0 seems like a problem... but if it's OK to also set `gasPrice` to 0, so that gas is free, then it doesn't matter if you also have no ether, right? :)

Similarly, `accounts: 0` means you can't use the unlocked accounts... but there's nothing stopping you from making your own private key and making your own account.  Aside perhaps from a lack of ether, but I imagine that can be addressed either with `evm_mine` calls, or, again, by making gas free so that ether is unnecessary.

These may be fringe use cases, but I don't see a reason to bar them.  Unless `gasPrice: 0` doesn't actually make gas free, anyway.

sukanyaparashar: I have tested with `gasPrice: 0` which seems to make no difference. `migrate` and `test` works with only `gasPrice: 0`. And if `gasPrice` and `defaultBalanceEther` are both 0, then the transaction fails with **insufficient funds**. May be that means `gasPrice: 0` is not making gas free.

haltman-at: Oh, interesting...

packages/core/lib/commands/develop/run.js
```diff
@@ -27,10 +27,20 @@ module.exports = async options => {
 
   const config = Config.detect(options);
   const customConfig = config.networks.develop || {};
+  let mnemonic = "";
```
haltman-at, line 30: I would avoid assigning default values that aren't meaningful.  If you just want to declare the variable and you're going to give it a value later, you can just write `let mnemonic;`.  By instead writing `let mnemonic = "";`, you're implying that the value `""` is meaningful and will be used at some point, which is misleading.

haltman-at (commented):
Functionality looks good, still some style things though (some of which should possibly change, some of which should definitely change).

packages/core/lib/commands/develop/run.js
```diff
@@ -27,10 +27,20 @@ module.exports = async options => {
...
+  let accounts = [];
+  let privateKeys = [];
 
-  const { mnemonic, accounts, privateKeys } = mnemonicInfo.getAccountsInfo(
-    customConfig.accounts || customConfig.total_accounts || 10
-  );
+  if (customConfig.accounts === 0 || customConfig.total_accounts === 0) {
```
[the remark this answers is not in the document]
sukanyaparashar: Thanks for your comments @haltman-at . I agree that this can be written in a more elegant way. I will work on this.

packages/core/lib/commands/develop/run.js
```diff
@@ -1,5 +1,6 @@
 const emoji = require("node-emoji");
 const mnemonicInfo = require("../../mnemonics/mnemonic");
+const configureGanacheOptions = require("../../configureGanacheOptions");
```
eggplantzzz, line 3: Just a style thing but how would you feel about just destructuring the method we want to use? It looks a little bit cleaner...like
```
const { configureManagedGanache } = require(....
```

packages/core/lib/commands/develop/run.js
```diff
@@ -27,9 +28,15 @@ module.exports = async options => {
   const config = Config.detect(options);
   const customConfig = config.networks.develop || {};
 
-  const { mnemonic, accounts, privateKeys } = mnemonicInfo.getAccountsInfo(
-    customConfig.accounts || 10
-  );
+  // Respect user's number of accounts choice to 0 in truffle config
+  const isZeroAccount =
```
eggplantzzz, line 32: Small detail, but how would you feel about renaming this to `isZeroAccounts`? When I first see this I think of THE zero account.

packages/core/lib/commands/develop/run.js
eggplantzzz, line 3: Also, the name of this module - how would you expect it to be expanded in the future? Is the name appropriate for that file if we do expand it? Is there a better name that would be appropriate?

packages/core/lib/commands/test/run.js
```diff
@@ -105,31 +86,26 @@ module.exports = async function (options) {
...
-    // Use managed ganache with overriding user specified config or without any specification in the config
     const defaultPort = await require("get-port")();
-    const defaultHost = "127.0.0.1";
     const defaultMnemonic =
       "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
-
     // configuredNetwork will spread only when it is defined and ignored when undefined
-    const ganacheOptions = {
-      host: defaultHost,
+    const configuredManagedNetwork = {
```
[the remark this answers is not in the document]
sukanyaparashar: Thanks for the suggestion @eggplantzzz . Naming variables can sometimes be confusing :) This one makes sense to me. I will change that.

packages/core/lib/commands/develop/run.js
sukanyaparashar: If we are going to expand this file for some other networks in future, then the module name could change. How about something like `configureManagedNetworks`? Any suggestions @eggplantzzz @cds-amal ?

eggplantzzz: What other kinds of utilities do you imagine might belong in that file?

sukanyaparashar: I am not able to think of anything specific right now to be included in this file. But in my opinion, this file can be used as a common adapter for the `core/lib` package.

packages/core/lib/commands/test/run.js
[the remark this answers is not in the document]
eggplantzzz: Yeah, naming is hard.

packages/core/lib/commands/develop/run.js
eggplantzzz: I'm just trying to think of a name that will describe all the things that will go in here.

sukanyaparashar: Does something like `managedNetworkAdapter` work?

haltman-at:
OK, so, I've taken a look at the recent commit, and it still bugs me.  I think maybe I didn't make clear enough precisely what my criticism was.  I'm going to attempt to make that explicit here.

Code should, whenever feasible, reflect the intended behavior as directly as possible.  (When it's not possible, it needs to be commented.  Possibly with a `//HACK` comment.)  Code that does not do so is code that will confuse people.  To my mind, the code you've written here does not achieve that goal.  Instead of structuring it around the fundamental behavior, you've structured it around a quirk of the language.

To understand what I mean by that, let's ask: What *is* the intended behavior here?  Well, the intended behavior is that if the user specifies a value for some field, that value should be passed on; and if they don't, if they leave it undefined, then a default value should be used.  Notice how zero is *not* a special case anywhere in this description!  The value `undefined` is a special case, but the value zero is not.  And yet this code treats it as a special case.  This suggests something is wrong.

Now, the original code attempted to accomplish this behavior with expressions of the form `userSuppliedValue || defaultValue`.  However, this expression does not accomplish the desired behavior when `userSuppliedValue` is zero, because zero is considered by the language to be falsy.  So, the expression `userSuppliedValue || defaultValue` is wrong because, in addition to the intended special case at `undefined`, it has an additional, unintended, special case at zero, whereas the intended behavior does not have this, and has a special case only at `undefined`.

Therefore, the ideal solution is to replace this expression with code that does not have a special case at zero, but rather has one only at `undefined`.  As I'd mentioned above, if we didn't have to worry about Node 12 compatibility, the way to do this would be `userSuppliedValue ?? defaultValue`; but since we need to still be Node 12 compatible for now, this is not an option.  (Yes obviously that also has a special case at `null` but that doesn't concern us here.)  Regardless, there are still plenty of ways to do this.  For instance, one might write `userSupliedValue !== undefined ? userSuppliedValue : defaultValue`.  That gets a bit awkward when there are multiple possible places the user supplied value might live, so you might not want to go with the most straightforward extension of that, but you get the idea.

However, this code, instead of removing the special case at zero by not using the troublesome `||`, instead leaves in the troublesome `||` with its special case at zero, and then adds additional special handling at zero to cause it to behave correctly in this case.  This is what I mean when I say you have structured this code around a quirk of the language, rather than the fundamental intended behavior.

Now, functionally, this may be correct -- you may have eliminated the special case in the behavior.  But you have not eliminated the special case in the expression of the behavior.  Again, code should reflect the intended behavior in as direct a manner as possible whenever this can be managed; so code here that makes a special case for zero is, to my mind, not doing that.  This code confused me when I read it, and it will confuse others in the future as well.  I think this code would be much clearer if you simply eliminated all special handling of zero and did not use `||`.

sukanyaparashar:
> OK, so, I've taken a look at the recent commit, and it still bugs me. I think maybe I didn't make clear enough precisely what my criticism was. I'm going to attempt to make that explicit here.
> 
> Code should, whenever feasible, reflect the intended behavior as directly as possible. (When it's not possible, it needs to be commented. Possibly with a `//HACK` comment.) Code that does not do so is code that will confuse people. To my mind, the code you've written here does not achieve that goal. Instead of structuring it around the fundamental behavior, you've structured it around a quirk of the language.
> 
> To understand what I mean by that, let's ask: What _is_ the intended behavior here? Well, the intended behavior is that if the user specifies a value for some field, that value should be passed on; and if they don't, if they leave it undefined, then a default value should be used. Notice how zero is _not_ a special case anywhere in this description! The value `undefined` is a special case, but the value zero is not. And yet this code treats it as a special case. This suggests something is wrong.
> 
> Now, the original code attempted to accomplish this behavior with expressions of the form `userSuppliedValue || defaultValue`. However, this expression does not accomplish the desired behavior when `userSuppliedValue` is zero, because zero is considered by the language to be falsy. So, the expression `userSuppliedValue || defaultValue` is wrong because, in addition to the intended special case at `undefined`, it has an additional, unintended, special case at zero, whereas the intended behavior does not have this, and has a special case only at `undefined`.
> 
> Therefore, the ideal solution is to replace this expression with code that does not have a special case at zero, but rather has one only at `undefined`. As I'd mentioned above, if we didn't have to worry about Node 12 compatibility, the way to do this would be `userSuppliedValue ?? defaultValue`; but since we need to still be Node 12 compatible for now, this is not an option. (Yes obviously that also has a special case at `null` but that doesn't concern us here.) Regardless, there are still plenty of ways to do this. For instance, one might write `userSupliedValue !== undefined ? userSuppliedValue : defaultValue`. That gets a bit awkward when there are multiple possible places the user supplied value might live, so you might not want to go with the most straightforward extension of that, but you get the idea.
> 
> However, this code, instead of removing the special case at zero by not using the troublesome `||`, instead leaves in the troublesome `||` with its special case at zero, and then adds additional special handling at zero to cause it to behave correctly in this case. This is what I mean when I say you have structured this code around a quirk of the language, rather than the fundamental intended behavior.
> 
> Now, functionally, this may be correct -- you may have eliminated the special case in the behavior. But you have not eliminated the special case in the expression of the behavior. Again, code should reflect the intended behavior in as direct a manner as possible whenever this can be managed; so code here that makes a special case for zero is, to my mind, not doing that. This code confused me when I read it, and it will confuse others in the future as well. I think this code would be much clearer if you simply eliminated all special handling of zero and did not use `||`.

Thanks for clarifying this @haltman-at . I was only focussing on the special case at zero. Sure, I will change the logic according to your suggestion.

packages/core/lib/configureGanacheOptions.js
```diff
@@ -18,27 +18,27 @@ function sanitizeGanacheOptions(ganacheOptions) {
...
-    const userAccounts = networkConfig.accounts || networkConfig.total_accounts;
-    return isZeroAccount ? 0 : userAccounts || 10;
+    if ("accounts" in networkConfig) {
+      return networkConfig.accounts;
+    }
+    if ("total_accounts" in networkConfig) {
+      return networkConfig.total_accounts;
+    }
+    return 10; // Default number of accounts when undefined in the truffle config
   };
```
sukanyaparashar, line 28: @haltman-at , I found this way a bit cleaner to understand the intended behavior. Let me know if this makes sense to you.

packages/core/lib/commands/develop/run.js
```diff
@@ -27,9 +28,17 @@ module.exports = async options => {
   const config = Config.detect(options);
   const customConfig = config.networks.develop || {};
 
-  const { mnemonic, accounts, privateKeys } = mnemonicInfo.getAccountsInfo(
-    customConfig.accounts || 10
-  );
+  const getAccounts = customConfig => {
+    if ("accounts" in customConfig) {
+      return mnemonicInfo.getAccountsInfo(customConfig.accounts);
```
eggplantzzz, line 33: Ugh, this will cause a problem if the user supplies `null` or `undefined` for that value in the config. Should we first provide a guard here to check that `accounts` or `total_accounts` is a number or string representation of a number?

sukanyaparashar: Should we throw an error to the user that the supplied value should be a number?

haltman-at: I'm not sure how much this is actually something to worry about?  Like, the previous method used in `develop`, where the code did `{...defaults, ...userOptions}` (not the exact code but YKWIM) also has this problem.

I would say the ideal behavior for `undefined` would be the same as if it isn't there, but I don't know if it's a big deal.  Whereas `null` strikes me as wholly not worth worrying about -- yeah if you do that you'll get problems, but why would you do that?

packages/core/lib/configureGanacheOptions.js
```diff
@@ -0,0 +1,71 @@
...
+  const ganacheOptions = {
+    host: networkConfig.host || "127.0.0.1", // Default host for managed ganache
+    port: networkConfig.port || 9545, // Default port for managed ganache
+    network_id: networkConfig.network_id || Date.now(),
+    total_accounts: calcTotalAccounts(networkConfig),
+    default_balance_ether: calcTotalEtherBalance(networkConfig),
+    blockTime: networkConfig.blockTime || 0,
+    fork: networkConfig.fork,
+    mnemonic: mnemonic,
+    gasLimit: networkConfig.gasLimit || 0x6691b7,
```
haltman-at, line 53: This line is probably worth a comment -- since you're just using `||` here, it's probably worth explaining why that's OK in this case but not in others (unless you think this should be changed too).

packages/core/lib/configureGanacheOptions.js
```diff
@@ -0,0 +1,71 @@
...
+    gasPrice: calcGasPrice(networkConfig),
+    // config.genesis_time is for compatibility with older versions
+    time:
+      networkConfig.genesis_time || networkConfig.time || config.genesis_time,
```
haltman-at, line 57: Similarly here, worth putting a comment explaining why `||` is OK here.

haltman-at (commented):
I think this approach is probably good (although @eggplantzzz might not agree!), but I think it could use some more comments in the code.

packages/core/lib/commands/develop/run.js
sukanyaparashar: I was also wondering why user will specify `null` as we didn't have any check for `null` previously too in our code. @eggplantzzz , what do you think about skipping the `null` check?

packages/core/lib/configureGanacheOptions.js
sukanyaparashar: Thanks for reviewing this @haltman-at . I am still working on making the logic a little bit more simpler and reduce the redundancies. In my opinion, we should have the same pattern for each of the properties (get rid of `||`).

haltman-at: That's another possibility, sure!

packages/core/lib/configAdapter.js
```diff
@@ -0,0 +1,103 @@
...
+      `The network id specified in the truffle config ` +
+      `(${network_id}) is not valid. Please properly configure the network id as an integer value.`;
+    throw new Error(error);
+  }
+
+  return { ...ganacheOptions, network_id: parsedNetworkId };
+}
+
+// This function returns the first defined argument value and throws if all arguments are undefined
+const getFirstDefinedValue = (...values) => {
```
haltman-at, line 21: Btw, if you want to make this shorter, you can just do `const getFirstDefinedValue = (...values) => values.find(value => value !== undefined)`. :)  Well, that version won't throw an error if nothing's passed in, it'll return `undefined`, but I don't see why we need to throw an error in that case anyway.

packages/core/lib/configAdapter.js
```diff
@@ -0,0 +1,103 @@
...
+    gasLimit,
+    gasPrice,
+    time: genesisTime,
+    miner: {
+      instamine: "strict"
+    }
+  };
+
+  // Set hardfork if defined, otherwise rely on Ganache's default which will always be correct
+  if (networkConfig.hardfork != null) {
```
haltman-at, line 95: Is there a reason this particular option is handled separately at the end, rather than inside the main body of `ganacheOptions` like the others?

(Btw, `ganacheOptions.hardfork` would read a little more idiomatic than `ganacheOptions["hardfork"]`, IMO... that's nitpicking, though.)

packages/truffle/test/scenarios/genesis_time/genesis_time_invalid.js
```diff
@@ -1,50 +0,0 @@
-const MemoryLogger = require("../MemoryLogger");
```
haltman-at, line 1: Can you explain in more detail just why you deleted these two tests?  I'm not sure I follow.

haltman-at (commented):
Thanks @sukanyaparashar, I think this version is a lot cleaner!   I see you more or less reinvented the `??` operator (without the short-circuiting)... which isn't a bad thing, seeing as `??` would be the right way to do this if we didn't have to worry about Node 12 compatibility!

I do have some style comments but most of this stuff is not that important, just me pointing out some further ways this can be cleaned up a bit.

packages/truffle/test/scenarios/genesis_time/genesis_time_invalid.js
haltman-at: Hm, I was going to ask if @gnidan knows any more about this, but it looks like the PR that originally added this option was #2236, an external contribution reviewed only by Cruz.  So, uh, I guess nobody here anymore is likely to know much about this... :-/

packages/core/lib/configAdapter.js
```diff
@@ -0,0 +1,103 @@
+function sanitizeGanacheOptions(ganacheOptions) {
```
cds-amal, line 1: This function name doesn't convey what the function does, which is operating on `network_id`. It made sense in previous revisions, and now that you're settling on a new approach, this could be refactored. Instead of:

```javascript
const sanitizedG = sanitizeGanacheOptions(config);
```
consider this:

```javascript
config.network_id = resolveNetworkId(config.network_id);
```

packages/core/lib/configAdapter.js
cds-amal: Harry's comment makes sense. Returning undefined may be the better way to proceed. Ganache already has sensible defaults which we could rely on.

packages/core/lib/configAdapter.js
cds-amal: Same nit: prefer idiomatic :)

packages/core/lib/configAdapter.js
```diff
@@ -0,0 +1,103 @@
...
+    }
+  };
+
+  // Set hardfork if defined, otherwise rely on Ganache's default which will always be correct
+  if (networkConfig.hardfork != null) {
+    ganacheOptions["hardfork"] = networkConfig.hardfork;
+  }
+
+  const sanitizedGanacheOptions = sanitizeGanacheOptions(ganacheOptions);
+  return sanitizedGanacheOptions;
```
cds-amal, line 100: This needlessly creates a new object when all it does is set/validate `network_config`. See previous comment.

cds-amal (changes requested):
Nice work @sukanyaparashar, this is a lot easier to read and reason about. There's just one piece I'd like to see refactored, given the implementation you've landed on.

packages/core/lib/configAdapter.js
```diff
@@ -0,0 +1,103 @@
...
+
+  const gasPrice = getFirstDefinedValue(
+    networkConfig.gasPrice,
+    0x77359400 // Use default gas price 2000000000 wei
+  );
+
+  const genesisTime = getFirstDefinedValue(
+    networkConfig.time,
+    networkConfig.genesis_time,
+    Date.now() // Use current time as default
```
haltman-at, line 74: Also, are you sure this default is safe?  I imagine it might cause problems if `fork` is passed.  Wouldn't it be better to just use no default?  (This would of course necessitate removing that throw like I mentioned.)

packages/core/lib/configAdapter.js
sukanyaparashar: This makes sense to me too. There is probably no need of throwing an error and we can leave it undefined.

packages/core/lib/configAdapter.js
sukanyaparashar: As it was written outside the main body of `ganacheOptions` previously, so I thought there is some significant difference but I don't think it is necessary. I will test it putting inside the main `ganacheOptions` body and note if anything changes.

packages/core/lib/configAdapter.js
sukanyaparashar: Yes, this makes sense. I will refactor this. Thanks @cds-amal .

packages/truffle/test/scenarios/genesis_time/genesis_time_invalid.js
sukanyaparashar: @cds-amal and I figured it out that if we are providing a default value for time, then the `genesis_time_undefined.js` doesn't make any sense since it will get a default value of current timestamp whenever time is not configured in the `truffle-config.js`. However, if we are going to pass an undefined value as default for time as suggested by @haltman-at , then `genesis_time_undefined.js` test is going to make sense.

sukanyaparashar: And `genesis_time_invalid.js` test looks for a default value and fails the test if there is none. So, if we want to remove `config.genesis_time`(which is always undefined) since we are now passing undefined as the default value when there is no defined value for time in the `truffle-config.js`, then we probably need to remove this test to pass the CI.

packages/core/lib/configAdapter.js
```diff
@@ -38,7 +32,7 @@ function configureManagedGanache(config, networkConfig, mnemonic) {
 
   const network_id = getFirstDefinedValue(
     networkConfig.network_id,
-    Date.now() // Use current time as default
+    5777 // Use as default network_id
```
haltman-at, line 35: Any particular reason for this change?

packages/core/lib/configAdapter.js
```diff
@@ -53,25 +47,21 @@ function configureManagedGanache(config, networkConfig, mnemonic) {
     100 // Use as default ether balance for each account
   );
 
-  const blockTime = getFirstDefinedValue(
-    networkConfig.blockTime,
-    0 // Use as default block time
-  );
+  const blockTime = getFirstDefinedValue(networkConfig.blockTime);
```
haltman-at, line 50: These lines of code, where `getFirstDefinedValue` is called on a single argument with no default, don't make sense.  If you only have a single input and no defaults, there's no need for the function; you can just write `const blockTime = networkConfig.blockTime`, etc.  Also, is there a reason these defaults were deleted?

haltman-at (commented):
Thanks Sukanya!  Still a few questions.

packages/core/lib/configAdapter.js
sukanyaparashar: Sure, @haltman-at . In my opinion, these values can be left undefined as we can rely on ganache's default for these values.

packages/core/lib/configAdapter.js
sukanyaparashar: I found out a different behavior with this. In `environment.js`, there is a logic to set the `network_id` to 5777 if it is undefined in the `config.network`. So when we use `Date.now()` as default, it conflicts with the set `network_id` 5777 and throws an error when we use the `migrate` command.

haltman-at: Oh, interesting.  Makes sense to stick with 5777 then.

packages/core/lib/configAdapter.js
haltman-at: Makes sense!  Do you know whether those defaults are the same as what we have here, though?  If they're different, it's possible people might be relying on the defaults here.  Indeed even if they are the same, it might make sense to leave in the defaults in case Ganache ever changes them (so that our defaults don't change).

Regardless, if you do take out the defaults, you should also take out the function call.

packages/core/lib/configAdapter.js
```diff
@@ -0,0 +1,89 @@
...
+
+  const default_balance_ether = getFirstDefinedValue(
+    networkConfig.defaultEtherBalance,
+    networkConfig.default_balance_ether,
+    100 // Use as default ether balance for each account
+  );
+
+  const blockTime = getFirstDefinedValue(networkConfig.blockTime);
+
+  const fork = getFirstDefinedValue(networkConfig.fork);
```
cds-amal, line 52: Agree with Harry that it doesn't make sense to invoke `getFirstDefinedValue` with a single argument.

packages/core/lib/configAdapter.js
cds-amal: The preference to rely on Ganache's defaults, with exceptions, should be documented in the code

sukanyaparashar: > Makes sense! Do you know whether those defaults are the same as what we have here, though? If they're different, it's possible people might be relying on the defaults here. Indeed even if they are the same, it might make sense to leave in the defaults in case Ganache ever changes them (so that our defaults don't change).
> 
> Regardless, if you do take out the defaults, you should also take out the function call.

I will make the changes you suggested @haltman-at . I thought of keeping the scope of adding the defaults again after the discussion :) The `blockTime` and `gasPrice` are same as Ganache's defaults. So, what should be desired behavior for this, should we rely on Ganache's defaults or should we keep our defaults? Any suggestions on this @gnidan @eggplantzzz ?

haltman-at: IMO we should probably leave the defaults as-is in this PR, and we can change or remove them later if we think that's warranted?  Just in the interest of keeping the scope of this PR from expanding...

sukanyaparashar: I agree with you @haltman-at . We can change them sometime later if desired.

packages/core/lib/configAdapter.js
```diff
@@ -0,0 +1,100 @@
...
+    // Higher precedence is given to the networkConfig.time or networkConfig.genesis_time
+    networkConfig.time,
+    networkConfig.genesis_time,
+    config.time,
+    config.genesis_time
+  );
+
+  const fork = networkConfig.fork;
+
+  // If not defined in the config, default hardfork value is "london"
```
haltman-at, line 75: I'd remove this comment.  It's true at the moment, but it will quickly become out of date.

packages/core/lib/configAdapter.js
```diff
@@ -0,0 +1,100 @@
...
+    mnemonic,
+    gasLimit,
+    gasPrice,
+    time: genesisTime,
+    miner: {
+      instamine: "strict"
+    }
+  };
+
+  ganacheOptions.network_id = resolveNetworkId(network_id);
```
haltman-at, line 96: Once again, makes more sense to move this earlier so that `ganacheOptions` is defined all at once?

haltman-at (approved):
OK, seems basically good to me.  I have two last nitpicks, but I'm going to go ahead and hit approve anyway. :)  (Not like you can actually merge it until @cds-amal dismisses his "changes requested", anyway. :P )

packages/core/lib/configAdapter.js
sukanyaparashar: Oh! Okay. If that changes quickly, then it is better to remove that. Thanks @haltman-at .

haltman-at: Well, maybe not "quickly", maybe only every few months... but nobody's going to remember to update this comment when they don't need to update the code.

cds-amal (approved):
Looks good to me. Nice work getting this through @sukanyaparashar.

PR #5032, opened 2022-04-19
Internal improvement: Remove esdoc from debugger

haltman-at:
We're no longer using this, @gnidan suggested removing it.


cds-amal (approved):
LGTM

PR #5035, opened 2022-04-19
Dependency update: Update yargs to ^13.3.0

haltman-at:
Addresses #399 and its duplicate #1005; supersedes #4944 (apologies @percentcer, but thanks for bringing our attention to the issue!).

I tried updating `yargs` to a more recent version, and, hey, that solved the problem!  Why to 13.3.0 specifically, and not an even more recent version?  Simple: More recent versions didn't pass CI!  Note we're currently on 8.x, so going to 13.3.0 is blowing past a number of breaking releases... however, it seems to work, and this version passes CI, so I guess it's OK?

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -83,7 +83,7 @@
     "web3": "1.5.3",
     "web3-utils": "1.5.3",
     "xregexp": "^4.2.4",
-    "yargs": "^8.0.2"
+    "yargs": "^13.3.0"
   },
   "devDependencies": {
     "@truffle/blockchain-utils": "^0.1.3",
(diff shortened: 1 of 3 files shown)


haltman-at:
OK, this breaks things apparently... will try other versions and see how they do.

cds-amal:
Hey @haltman-at, did you manually test/verify, `truffle develop` and `truffle console`?

haltman-at:
I manually tested `truffle console`, but not `truffle develop`.  I assume they should work the same way -- they use the same code, right? -- but I can try it just to be certain.

haltman-at:
OK @cds-amal, I did some more manual testing with `truffle develop`, this time testing a few more cases!  Specifically I tested putting double spaces within and before:
1. Ordinary expressions,
2. Truffle commands,
3. Truffle commands with the `truffle` prefix

They all work!

sukanyaparashar (approved):
Looks good to me since you have tested everything and it is working as expected.

PR #5038, opened 2022-04-20
Make most tests use ganache provider instead of server

benjamincburns:
Makes it so that the tests in question don't require open ports to listen on, and are therefore easier to execute in parallel.

### DIFF
--- a/packages/db/src/project/test/compilationSources/.eslintrc.json
+++ b/packages/db/src/project/test/compilationSources/.eslintrc.json
@@ -1,3 +1,3 @@
 {
-  "extends": ["../../../../../.eslintrc.truffle.json"]
+  "extends": ["../../../../../../.eslintrc.truffle.json"]
 }
--- a/packages/db/src/project/test/compilationSources/truffle-config.js
+++ b/packages/db/src/project/test/compilationSources/truffle-config.js
@@ -1,13 +1,21 @@
+const Ganache = require("ganache");
+
+const provider = Ganache.provider({
+  // note instamine must be set to eager (default) with vmErrorsOnRPCResponse enabled
+  vmErrorsOnRPCResponse: true,
+  logging: {
+    quiet: true
+  }
+});
+
 module.exports = {
   networks: {
     development: {
-      host: "127.0.0.1",
-      port: 8545,
+      provider,
       network_id: "*"
     },
     test: {
-      host: "127.0.0.1",
-      port: 8545,
+      provider,
       network_id: "*"
     }
   },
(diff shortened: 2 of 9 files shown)


benjamincburns:
There are still a couple of tests using `Ganache.server`, namely the `@truffle/provider` tests, and the scenario tests.

The `@truffle/provider` tests could probably be refactored to work w/o the server by using mocks, but I didn't want to dive into it too deeply. The scenario tests are a whole other kettle of fish, however. Given the nature of those tests (e2e integration tests), I didn't feel it was appropriate to refactor them, especially as part of this PR.

packages/hdwallet-provider/test/provider.test.ts
```diff
@@ -224,19 +222,19 @@ describe("HD Wallet Provider", function () {
...
-      assert.ok(provider.engine,
-        "Web3ProviderEngine instantiated");
-      // @ts-ignore - _blockTracker is not officially part of the interface
-      assert.ok(provider.engine._blockTracker,
-        "PollingBlockTracker instantiated");
+      assert.ok(provider.engine, "Web3ProviderEngine instantiated");
+      assert.ok(
+        (provider.engine as any)._blockTracker,
+        "PollingBlockTracker instantiated"
+      );
```
benjamincburns, line 232: formatting hooks wrapped this line, which made the `@ts-ignore` no longer apply to the correct place, so I opted to replace it w/ a cast to `any`, as that's at least enshrined in proper syntax.

packages/db/src/project/test/compilationSources/.eslintrc.json
```diff
@@ -1,3 +1,3 @@
 {
-  "extends": ["../../../../../.eslintrc.truffle.json"]
+  "extends": ["../../../../../../.eslintrc.truffle.json"]
```
haltman-at, line 2: Oops, looks like I miscounted here, thanks!

packages/db/src/project/test/compilationSources/truffle-config.js
```diff
@@ -1,13 +1,21 @@
+const Ganache = require("ganache");
+
+const provider = Ganache.provider({
+  // note instamine must be set to eager (default) with vmErrorsOnRPCResponse enabled
+  vmErrorsOnRPCResponse: true,
+  logging: {
+    quiet: true
+  }
+});
```
cds-amal, line 9: Nit/confustion: the comment is confusing because it implies a code action and mentions the default, which nullifies the purpose of the comment? Perhaps explicitly setting miner.instamine = 'strict' with a comment to explain the pairing would be better?

cds-amal (approved):
Looks good to me! I left a nit, which you could ignore.

PR #5040, opened 2022-04-22
Bug fix: Ensure all contracts get primary source id attached in shimContracts (fix phantom stackframes, again)

haltman-at:
So, it looks like I accidentally broke the debugger's phantom stackframe system.  Again.  But really the problem wasn't with the phantom stackframe system... it was actually a problem with `shimContracts` which I accidentally exposed.

The real problem was that if multiple contracts were declared in the same source, but neither `sources` nor `files` was passed in, they wouldn't all get a primary source ID.  The first one to get processed would, and then the rest wouldn't.  This is because I overloaded the `index` return value from `getIndexToAddAt`.  I had it return a number to indicate "put this source at this index", and `null` to indicate "we already have this source, we don't need to add it".  But, actually, we need that index even if we already have the source, because it's used for the contract's `primarySourceId`!  So returning `null` in that case wasn't cutting it.

To fix this, I made `index` always a number, and I added another return value, `needsAdding`, which is a boolean.  Now instead of returning `index: null`, we return `needsAdding: false`, and `index` is always a number.  Moreover, we set `primarySourceId` regardless of the value of `needsAdding`.

So why was this causing the phantom stackframe problem?  Well, without a `primarySourceId`, a contract wouldn't get a `primarySourceIndex`, and without a `primarySourceIndex`, it wouldn't get a `primaryLanguage`, and you may recall that I recently restricted the phantom stackframe system to only apply to contracts whose `primaryLanguage` is `"Solidity"`.  Oops!

Anyway, this should fix that.  Sorry about the mess...

### DIFF
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -166,7 +166,8 @@ export function shimContracts(
       contractObject.primarySourceId = index.toString(); //HACK
     } else {
       //if neither was passed, attempt to determine it from the ast
-      let index;
+      let index: number;
+      let needsAdding: boolean;
       if (sourceObject.ast) {
         //note: this works for both Solidity and Vyper
         index = sourceIndexForAst(sourceObject.ast); //sourceObject.ast for typing reasons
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
Looks good to me! Thanks @haltman-at

PR #5047, opened 2022-04-22
improve default truffle-config.js readability

davidmurdoch:
This PR isn't trying to make it perfect, just a little bit better. :-)

### DIFF
--- a/packages/core/lib/commands/init/initSource/truffle-config.js
+++ b/packages/core/lib/commands/init/initSource/truffle-config.js
@@ -1,7 +1,7 @@
 /**
  * Use this file to configure your truffle project. It's seeded with some
  * common settings for different networks and features like migrations,
- * compilation and testing. Uncomment the ones you need or modify
+ * compilation, and testing. Uncomment the ones you need or modify
  * them to suit your project as necessary.
  *
  * More information about configuration can be found at:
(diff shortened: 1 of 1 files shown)


haltman-at (approved):
Makes sense to me.  Thanks!

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -46,34 +46,37 @@ module.exports = {
...
+    //   network_id: 1342,       // Custom network
+    //   gas: 8500000,           // Gas sent with each transaction (default: ~6700000)
+    //   gasPrice: 20000000000,  // 20 gwei (in wei) (default: 100 gwei)
+    //   from: <address>,        // Account to send transactions from (default: accounts[0])
+    //   websocket: true         // Enable EventEmitter interface for web3 (default: false)
     // },
+    //
     // Useful for deploying to a public network.
-    // NB: It's important to wrap the provider as a function.
+    // Note: It's important to wrap the provider as a function to ensure truffle uses a new provider every time.
```
cds-amal, line 61: Uses a new provider every time is ambiguous. I would rather leave it as the suggestion below and expand the reasoning in our documentation. 

```suggestion
    // Note: It's important to wrap the provider as a function.
```

cds-amal (changes requested):
Oops, missed this.

eggplantzzz (approved):
I like this.

PR #5051, opened 2022-04-24
Update create contract template

Rassska:
* the constructor visibility has been removed from Solidity since v0.7 and above. *
    
    Read more about this right here: 
    https://docs.soliditylang.org/en/v0.7.0/070-breaking-changes.html#functions-and-events

### DIFF
--- a/packages/core/lib/commands/create/templates/Example.sol
+++ b/packages/core/lib/commands/create/templates/Example.sol
@@ -2,6 +2,6 @@
 pragma solidity >=0.4.22 <0.9.0;
 
 contract Example {
-  constructor() public {
+  constructor() {
   }
 }


haltman-at:
It's not true that constructor visibility is no longer permitted; including it is a warning, not an error.  Presumably it will be removed at some point in the future, but for now it's still legal.

It's not obvious to me whether we want to remove this.  Notice the version pragma up top -- this starting contract is meant to work with a wide variety of versions.  If we removed the constructor visibility, it would only work with 0.7.x and up.

I don't know, do other people have opinions on this? @gnidan? @eggplantzzz? @cds-amal?

cds-amal:
I'm not in favor of changing for the multiple Solidity versions specified by the pragma.

eggplantzzz:
Thanks for the proposed change @Rassska! I agree with @haltman-at that we should leave this as it is until Truffle by default uses a version greater than `0.7.0`.

lsqproduction:
Close per @eggplantzzz 's comment above.

Issue #4485, opened 2021-12-03
@truffle/fetch-and-compile should not require a @truffle/config object


gnidan:
## Issue

The `fetchAndCompile` function accepts two arguments: an address and a TruffleConfig. The second argument **must** be a TruffleConfig because it relies heavily on that package's defined defaults for things like precedence order for which fetchers to use, supplying no defaults of its own.

This is fiddly. The interface here should be expanded to accept either the existing `config` object or a new options argument with the relevant parameters that are actually used in this flow. This alternate form should _strictly_ contain that which actually matters in this context (e.g. there's no need to allow configuring network name, network id, or anything about solc other than whether to use Docker – none of that "The `solc` option is an object of the form supplied to a `truffle-config.js` file's `compilers` configuration", I say! 🙌 )

gnidan:
@haltman-at writes about the information that fetch-and-compile uses from the config:

> 1. `sourceFetchers` -- list of source fetchers to check, in what order; defaults as you know to `["etherscan", "sourcify"]`
> 2. `network_id` -- the network ID. of course as you know on an actual config this is a getter that returns `config.networks[config.network].network_id`
> 3. `etherscan`, `sourcify`, etc -- each fetcher name is taken as a config field w/options for that fetcher; currently only `etherscan` is currently used, it's where you can put your etherscan `apiKey` (of course you have noted that this sort of thing probably really belongs in user config rather than project config, but I don't think that's really relevant here and right now it's in project config)
> 4. `compilers.solc.docker` -- are you using docker for solc? (and possibly in the future a similar option might exist for `vyper`?? not clear how that'd work, like what the alternative would be)
> 5. (not checked deliberately by fetch-and-compile but I believe this accidentally has an effect... we should probably override this in fetch-and-compile to avoid issues, I should probably throw up a PR to fix this) it looks like `strict` gets passed through to compile-solidity and will cause errors on warnings (like I said, that's not deliberate and I should fix it)
>     
>     oh that one's probably meant to be an option rather than a config field though isn't it? so I guess actually I won't bother with that because people don't set that in the config (and it will not in fact make it through from the command-line options).

Thanks for this, @haltman-at!

gnidan:
This being said, what about this:

```typescript
export interface FetchAndCompileOptions {
  network: {
    networkId: number;
  };
  fetch?: {
    sourceFetchers?: string[]; // eg. ["etherscan", "sourcify"]
    etherscan?: {
      apiKey?: string;
    };
[...]

PR #5054, opened 2022-04-26
Enhancement: Add new interface for fetch-and-compile that doesn't require a Truffle config

haltman-at:
Addresses #4485.

This PR alters the interface for `fetch-and-compile` so that its functions now, instead of a Truffle Config, take either a Truffle Config *or* a new type of object containing the necessary info.  All the exported functions now accept this alternate input (which is intended to now be the primary input).  The new input type is based on the one suggested by @gnidan in the commentson #4485, although there have been some minor alterations for typing convenience.

To accomplish this, all the functions now run their input first through a `normalizeInput` function which normalizes things to the new format.  Since `Config` is a class rather than just a type, we detect if the input is a `Config` with a simple `instanceof` check.

The rest of the code has been altered appropriately.  Out of an abundance of caution, I'm still avoiding using `?.` for now, even though it should be fine here. :P  But, uh, I can change that and just use it if people would prefer...

I also had to use a `//@ts-ignore` in one place where TS couldn't tell that something was defined but it very obviously is.

There is still one bit of remaining interface awkwardness -- `getSupportedNetworks`.  I updated it to also take this new form of input; but this new input type still requires specifying a network ID, which is irrelevant for `getSupportedNetworks`.  I've suggested in the past that `getSupportedNetworks` should allow passing in a simple `string[]` instead of this more complicated input; I didn't go and implement that here, but I could go back and do so?

(We could also make it so that `network` and `networkId` aren't actually required in `FetchAndCompileOptions`, and instead we explicitly check for them and throw if they're not present?  That would resolve a little bit of the awkwardness, anyway, without having to go accept a whole third type of input for this one function...)

### DIFF
--- a/packages/fetch-and-compile/lib/fetch.ts
+++ b/packages/fetch-and-compile/lib/fetch.ts
@@ -12,15 +12,18 @@ import Config from "@truffle/config";
 const { Compile } = require("@truffle/compile-solidity"); //sorry for untyped import!
 import type { Recognizer, FailureType, FetchAndCompileOptions } from "./types";
 import type { WorkflowCompileResult } from "@truffle/compile-common";
-import { normalizeInput } from "./utils";
+import {
+  normalizeFetchAndCompileOptions,
+  normalizeFetcherNames
+} from "./utils";
 
 export async function fetchAndCompileForRecognizer(
   recognizer: Recognizer,
   options: FetchAndCompileOptions | Config
 ): Promise<void> {
-  const normalizedOptions = normalizeInput(options);
+  const normalizedOptions = normalizeFetchAndCompileOptions(options);
   const fetcherConstructors: FetcherConstructor[] =
-    getSortedFetcherConstructors(normalizedOptions);
+    getSortedFetcherConstructors(normalizeFetcherNames(normalizedOptions));
   const fetchers = await getFetchers(
     fetcherConstructors,
     normalizedOptions,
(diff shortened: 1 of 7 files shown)


haltman-at:
OK, after talking to @gnidan, I went back and added support for `string[]` input to `getSupportedNetworks`.  So that's substantially easier to use now, yay!  I also added a test of that functionality.

cds-amal:
The Readme examples should reflect these changes.

packages/fetch-and-compile/lib/fetch.ts
```diff
@@ -118,13 +129,16 @@ async function tryFetchAndCompileAddress(
       break;
     }
     //set up the config
-    let externalConfig: Config = config.with({
+    let externalConfig: Config = Config.default().with({
```
cds-amal, line 132: This could be created before being used in this loop.

cds-amal (changes requested):
Nice work @haltman-at. I left one minor note on the logic, and think the Readme should reflect the new interface/usage before merging.

packages/fetch-and-compile/lib/fetch.ts
haltman-at: I am confused, no it can't?  It depends on `options`, which has only just been determined.

haltman-at:
Oh, the readme!  I always forget that exists, thanks for catching that.

haltman-at:
Oops, I just noticed another problem -- `fetcherOptions` wasn't made optional!  Will fix that too.

packages/fetch-and-compile/lib/fetch.ts
cds-amal: You are not confused, sir! I was wrong.

cds-amal (approved):
Looks good to me @haltman-at

PR #5057, opened 2022-04-27
Bug fix: Allow use of ENS with mapping key watching in decoder

haltman-at:
Thanks to @gnidan for catching this.  So here's a fairly dumb mistake on my part.  The decoder uses the encoder for watching mapping keys.  The encoder allows the use of ENS for addresses (if this is turned on), and the decoder, by default, turns it on.  Yet ENS wasn't working; in the encoder, it wasn't actually getting turned on.  What's the problem?

Well, so, the decoder uses a special route for creating the encoder that avoids repeating certain work.  And, I accidentally restricted the encoder's ENS setup code to the normal route, rather than running it uncondtionally.  Oops.  Anyway, this PR moves that code out of the conditional so it'll run regardless.

I had meant to include an automated test of this, but I ran into some trouble setting it up, so I just excluded it for now.  But if people think it's worth it I can try again; I will likely need to get some help with that though.  It worked when I tried it manually, at least -- well, OK, it didn't exactly *work* as such, because I ran into some connection issues, but it made the call; and I've seen the encoder's ENS functionality work plenty of other times, so, um, this ought to be correct?

### DIFF
--- a/packages/encoder/lib/encoders.ts
+++ b/packages/encoder/lib/encoders.ts
@@ -89,6 +89,7 @@ export class ProjectEncoder {
   constructor(info: Types.EncoderInfoInternal) {
     //first, set up the basic info that we need to run
     if (info.userDefinedTypes && info.allocations) {
+      debug("internal route!");
       this.userDefinedTypes = info.userDefinedTypes;
       this.allocations = info.allocations;
     } else {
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
Looks good @haltman-at, even though I couldn't verify correct behavior. This path should be tested in CI, but I don't feel strongly about including it in this PR or as soonish followup work, given you've tested manually. So, uh, your choice :)

haltman-at:
Cool, I'm going to just go ahead and merge it... I'd say "unless @gnidan objects", but, well, he's out!

PR #5067, opened 2022-04-28
Dependency update: Update ganache to ^7.1.0

haltman-at:
Since nobody else has done this so far!

(Also, oops, `encoder` was on a different version from the other packages.  Now it's not.)


sukanyaparashar (approved):
Looks good to me.

PR #5079, opened 2022-05-04
Enhancement: Add ability to compile only specified sources

eggplantzzz:
This PR is using work from https://github.com/trufflesuite/truffle/pull/2713/files to implement a compilation flow where a user can specify specific files to compile.

### DIFF
--- a/packages/core/lib/commands/compile/run.js
+++ b/packages/core/lib/commands/compile/run.js
@@ -23,6 +23,33 @@ module.exports = async function (options) {
     );
   }
 
+  if (config._ && config._.length > 0) {
+    // set paths based on command-line inputs, transforming to absolute
+    // paths where appropriate
+    config.paths = config._.map(specifiedPath => {
+      // convert relative paths to absolute paths based on whether
+      // the naive absolute path exists on disk
+      //
+      // NOTE in case of collision where the specified path refers to some
+      // non-FS source (e.g. `truffle/Assert.sol`) and where that specified
+      // path corresponds to an existing file relative to the working dir.,
+      // this selects the latter as priority over the former.
+
+      const absolutePath = path.resolve(
+        config.working_directory,
+        specifiedPath
+      );
+
+      // i.e., pass the absolutePath if it's a real file, otherwise just
+      // pass whatever was specified.
+      if (fse.existsSync(absolutePath)) {
+        return absolutePath;
+      } else {
+        return specifiedPath;
+      }
+    });
+  }
+
   const compilationOutput = await WorkflowCompile.compile(config);
   if (options.saveIntermediate) {
     // Get the filename the user provided to save the compilation results to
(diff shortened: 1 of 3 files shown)


haltman-at:
This is intended to completely supersede #2713, right?

eggplantzzz:
Yes this will make that PR irrelevant.

packages/core/lib/commands/compile/run.js
```diff
@@ -23,6 +23,33 @@ module.exports = async function (options) {
     );
   }
 
+  if (config._ && config._.length > 0) {
```
cds-amal, line 26: what is `config._`?

packages/workflow-compile/index.js
```diff
@@ -26,12 +26,17 @@ async function compile(config) {
...
+        return await Compile.all(config);
+      } else if (config.paths && config.paths.length > 0) {
+        // compile only user specified sources
+        return await Compile.sourcesWithDependencies({
+          options: config,
+          paths: config.paths
+        });
+      } else {
+        return await Compile.necessary(config);
+      }
```
cds-amal, line 39: Style observation (no need to change). Just pointing out the else clauses aren't necessary when preceded by a return.

packages/workflow-compile/index.js
```diff
@@ -26,12 +26,17 @@ async function compile(config) {
...
 
-      const compileMethod =
-        config.all === true || config.compileAll === true
-          ? Compile.all
-          : Compile.necessary;
-
-      return await compileMethod(config);
+      if (config.all === true || config.compileAll === true) {
+        return await Compile.all(config);
+      } else if (config.paths && config.paths.length > 0) {
```
cds-amal, line 31: There's an opportunity for a stricter test. 
`Array.isArray(config.paths) && config.paths.length > 0)` would fail if `configs.paths` is not an Array, whereas the current test would pass if `config.paths` is truthy and doesn't guarantee `paths` has a `length` property.

cds-amal (approved):
Looks good to me. I had a question, and some minor nits, for your consideration. Nothing to hold back the PR.

packages/core/lib/commands/compile/run.js
eggplantzzz: `_` is an array where arguments from the command line get put. In this case it will be an array populated by source paths.

packages/workflow-compile/index.js
eggplantzzz: Do you think this is worth doing? Do you think it will provide value? You are really just providing a stricter validation step?

cds-amal: It's your call, I just wanted to bring it to your attention. I think it is more correct to check for an Array type because a truthy value doesn't imply Array. OTOH, since we control how `config.paths` is populated, does it really matter?

packages/core/lib/commands/compile/run.js
cds-amal: thanks!

PR #5080, opened 2022-05-04
Specify --platform when running solc Docker image

gnidan:
I noticed a warning when using Dockerized solc on M1 Macs, since the solc images are distributed for x64, and Docker throws a warning to suggest specifying `--platform` directly.

This PR does just that, and the warning goes away!

(In the future, if solc distributes Docker images for other platforms, we might want to revisit this... but I'm guessing that's pretty unlikely)

### DIFF
--- a/packages/compile-solidity/src/compilerSupplier/loadingStrategies/Docker.ts
+++ b/packages/compile-solidity/src/compilerSupplier/loadingStrategies/Docker.ts
@@ -39,7 +39,7 @@ export class Docker {
 
     const versionString = await this.validateAndGetSolcVersion();
     const command =
-      "docker run --rm -i ethereum/solc:" +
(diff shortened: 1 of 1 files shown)


haltman-at (approved):
I guess this makes sense... @gnidan, you say this works and makes the warning go away, so I don't see any problem with this.

gnidan:
> I guess this makes sense... @gnidan, you say this works and makes the warning go away? If so I don't see any problem with this.

Yep! As you might've guessed, I did this work because I wanted to avoid an ugly warning in a live demo, and this change successfully achieved that goal :)

Thanks!

gnidan:
> Is it possible to add the actual warning, so it's available for the audit trail?

Sure, where do you want me to add it? Just in the PR description?

cds-amal:
The PR description is a good backup. Having it in a git-commit body would have been the best :)

PR #5093, opened 2022-05-09
Internal improvement: More test optimization

eggplantzzz:


packages/debugger/package.json
```diff
@@ -17,9 +17,9 @@
   "scripts": {
     "build": "webpack --config webpack/webpack.config.js",
     "prepare": "yarn build",
-    "test": "mochapack --webpack-config webpack/webpack.config-test.js --recursive",
-    "test:coverage": "nyc mochapack --webpack-config webpack/webpack.config-test.js --recursive",
-    "test:debug": "NODE_ENV=testing node --inspect ./node_modules/.bin/mochapack --webpack-config webpack/webpack.config-test.js --recursive"
+    "test": "mochapack --webpack-config webpack/webpack.config-test.js \"./test/*.js\" \"./test/data/*.js\"",
+    "test:coverage": "nyc mochapack --webpack-config webpack/webpack.config-test.js \"./test/*.js\" \"./test/data/*.js\"",
+    "test:debug": "NODE_ENV=testing node --inspect ./node_modules/.bin/mochapack --webpack-config webpack/webpack.config-test.js \"./test/*.js\" \"./test/data/*.js\""
```
cds-amal, line 22: What's the difference between `--recursive` and glob?

packages/debugger/test/sources/init/truffle-config.js
```diff
@@ -0,0 +1,116 @@
+/**
+ * Use this file to configure your truffle project. It's seeded with some
+ * common settings for different networks and features like migrations,
```
cds-amal, line 3: Let's nix the comments. They add too much noise around actual settings.

packages/debugger/webpack/webpack.config-test.js
```diff
@@ -28,7 +28,7 @@ module.exports = merge(commonConfig, {
             plugins: ["@babel/plugin-transform-runtime"]
           }
         },
-        exclude: [path.resolve(__dirname, "..", "node_modules")],
+        exclude: path.resolve(__dirname, "..", "node_modules"),
```
cds-amal, line 31: Is this equivalent? `exclude` was previously an Array

packages/truffle/test/sources/init/contracts/Migrations.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.0;
+
+contract Migrations {
+  address public owner = msg.sender;
```
cds-amal, line 5: we can speed up tests if we remove migrations. Though we should have 1 integration test to cover that path.

packages/truffle/test/sources/init/truffle-config.js
```diff
@@ -0,0 +1,116 @@
+/**
+ * Use this file to configure your truffle project. It's seeded with some
+ * common settings for different networks and features like migrations,
+ * compilation and testing. Uncomment the ones you need or modify
+ * them to suit your project as necessary.
```
cds-amal, line 5: Noisy file :(

cds-amal (commented):
Looks good, I just have a few questions

packages/debugger/package.json
eggplantzzz: `--recursive` runs its algorithm on all directories. We don't want it going and trying to run on the `sources` directory because it causes errors. Those fixtures are not meant for mochapack consumption.

packages/debugger/webpack/webpack.config-test.js
eggplantzzz: Oh this must have been a mistake on my part. I'll change it back, I assume it is equivalent? I think I've seen both cases.

packages/debugger/package.json
cds-amal: oh, cool! thanks!

packages/truffle/test/scenarios/solidity_testing/solidityTests.js
```diff
@@ -62,7 +62,7 @@ describe("Solidity Tests", function () {
     });
 
     it("throws errors correctly", function () {
-      this.timeout(70000);
+      this.timeout(100000);
```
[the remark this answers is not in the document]
sukanyaparashar: Yeah! Got it. :)

packages/truffle/test/sources/init/migrations/1_initial_migration.js
```diff
@@ -0,0 +1,5 @@
+const Migrations = artifacts.require("Migrations");
```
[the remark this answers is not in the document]
eggplantzzz: That's true! Currently we don't have a way to cleanly share these between packages. I mean, we could just require from outside each package with relative paths like `require("../../core/test/sources...` but that is a dirty-feeling solution. Maybe we should create a package with test helpers to be used internally that contains fixtures and some of the sandbox-type utilities? I'm liking that idea!

PR #5102, opened 2022-05-11
Internal improvement: Remove node check in cli.js

eggplantzzz:
In the interest of cleaning up and optimizing Truffle, this PR removes the Node version check in `cli.js`.

### DIFF
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -1,10 +1,8 @@
 #!/usr/bin/env node
 require("source-map-support/register");
 
-const semver = require("semver"); // to validate Node version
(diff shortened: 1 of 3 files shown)


haltman-at:
How much does this actually help?  The check seems useful; is it so slow that it's worth getting rid of it?

eggplantzzz:
So putting some timers in the code reveals that the benefit is significant! Running from the beginning of `cli.js` up to the point where the commands load in each case (running `truffle version`) showed that that code went from taking around .46 seconds to taking .01 seconds. However the time the whole command flow took to run didn't seem to decrease for some reason...

PR #5114, opened 2022-05-16
Internal improvement: Restructure bundle for Truffle

eggplantzzz:
This PR aims to speed up Truffle's command flow. It restructures the entry points for Truffle meaning it creates a bundle for each command and its `run` method. Previously, all commands were included in the main bundle causing them to all get loaded on every command. This new architecture only loads the command being run and there have been some significant time reductions during initial tests.

Another side effect of this is that the `Command` class responsible for running the commands has been removed in favor of some helper methods used to parse out commands, parse out options, and then run the `run` method for the command that we care about.

It also removes the webpack entry point for the `commands` folder as that is no longer necessary now that we bundle each individual command.

**Update**: Just running a quick benchmark (run on a bare metacoin project) on my machine I got the following results --
  _with published Truffle_ 
    - `time truffle compile` yielded ~ 2.2 total seconds
    - `time truffle version` yielded ~ 1.8 total seconds
  _running new bundle_    
    - `time truffle compile` yielded ~ 1.7 total seconds
    - `time truffle version` yielded ~ 1 total second


eggplantzzz:
The size of the `build` folder after using `npm pack` is 53.3 MB. The `build` folder on the current Truffle distribution is 51.8 MB.

packages/core/lib/commands/commands.js
```diff
@@ -0,0 +1,27 @@
...
+  "obtain",
+  "opcode",
+  "preserve",
+  "publish",
+  "run",
+  "test",
+  "unbox",
+  "version",
+  "watch"
+];
```
gnidan, line 27: is it okay that this is now static?

eggplantzzz: One thought I had is that we could generate it at build time - that might be a little strange though.

eggplantzzz:
I think I'm going to work on this a bit more, I think the calls to `getCommand` and `runCommand` can be simplified and made clearer somehow.

packages/core/lib/command-utils.js
```diff
@@ -0,0 +1,209 @@
...
+const { bundled, core } = require("../lib/version").info();
+const OS = require("os");
+const analytics = require("../lib/services/analytics");
+const { extractFlags } = require("./utils/utils"); // contains utility methods
+const globalCommandOptions = require("./global-command-options");
+const debugModule = require("debug");
+const debug = debugModule("core:command:run");
+const commands = require("./commands/commands");
+
+const getCommand = (inputStrings, options, noAliases) => {
```
gnidan, line 11: Couple things:
- Can we turn this into a single options argument?
- Is there a better name than `getCommand`? Or maybe this just needs a docstring explaining the function

packages/core/lib/command-utils.js
```diff
@@ -0,0 +1,209 @@
...
+
+  return {
+    name: chosenCommand,
+    options: commandOptions,
+    run: command.run,
+    meta: command.meta
+  };
+};
+
+const prepareOptions = (command, inputStrings, options) => {
```
gnidan, line 90: (Same comments about `options` arg here, and request for docstring)

packages/core/lib/command-utils.js
```diff
@@ -0,0 +1,209 @@
...
+    }
+  }
+
+  return {
+    ...clone,
+    ...command.options
+  };
+};
+
+const runCommand = async function (command, options) {
```
gnidan, line 146: Is `command` here the same thing that is returned by `getCommand`?

gnidan (commented):
Couple fly-by style comments

packages/core/lib/command-utils.js
```diff
@@ -0,0 +1,209 @@
...
+const OS = require("os");
+const analytics = require("../lib/services/analytics");
+const { extractFlags } = require("./utils/utils"); // contains utility methods
+const globalCommandOptions = require("./global-command-options");
+const debugModule = require("debug");
+const debug = debugModule("core:command:run");
+const commands = require("./commands/commands");
+
+const getCommand = (inputStrings, options, noAliases) => {
+  if (inputStrings.length === 0) {
```
lsqproduction, line 12: Is this `inputStrings` here the same as the one in `cli.js`? if so, when the `inputString` length is 0, wouldn't it qualify as `userWantsGeneralHelp` instead of throwing the error?

lsqproduction, line 12: Is this `inputStrings` here the same as the one in `cli.js`? if so, when the `inputString` length is 0, wouldn't it qualify as `userWantsGeneralHelp` instead of throwing the error?

packages/core/lib/command-utils.js
eggplantzzz: So this function returns the command name, its options, its meta info, and the run method for that command. Is there a better name that comes to mind for this stuff? I don't mind if they are named, I'll change them.

packages/core/lib/command-utils.js
eggplantzzz: Yes it is.

packages/core/lib/command-utils.js
eggplantzzz: Yes, that's true! I guess we also need to consider this utility function on its own as you never know where it might be used later on. It is probably ok for this to throw when no input is given, eh? 

That makes me take note though, I think we need to put the same kind of check for 0 input strings in the console context. I don't think that exists right now.

packages/core/cli.js
```diff
@@ -35,27 +35,33 @@ if (!semver.gte(process.version, minimumNodeVersion)) {
...
 
-command
-  .run(inputArguments, options)
+const {
+  getCommand,
+  prepareOptions,
+  runCommand
+} = require("./lib/command-utils");
+
+const command = getCommand(inputStrings, {}, false);
```
eggplantzzz, line 61: I'm thinking maybe that instead of passing an array of input strings to `getCommand` that we should have another function call before this that parses the input and determines which command to get. Seems like a cleaner interface.

packages/core/lib/commands/develop/run.js
```diff
@@ -9,10 +9,10 @@ const runConsole = async (config, ganacheOptions) => {
   const Console = require("../../console");
   const { Environment } = require("@truffle/environment");
 
-  const commands = require("../index");
+  const commands = require("../commands");
   const excluded = new Set(["console", "develop", "unbox", "init"]);
```
cds-amal, line 13: Thoughts on adding `db` to the exclusion list? Currently, it allows for `db serve` which will lock up the develop console and is a terrible user experience.

packages/core/lib/console-child.js
```diff
@@ -1,11 +1,11 @@
-const Command = require("../lib/command");
 const TruffleError = require("@truffle/error");
 const Config = require("@truffle/config");
 const Web3 = require("web3");
 const yargs = require("yargs");
 
+// we split off the part Truffle cares about and need to convert to an array
 const input = process.argv[2].split(" -- ");
-const inputStrings = input[1];
+const inputStrings = input[1].split(" ");
```
cds-amal, line 8: Why split on space? I don't know if it's possible in this code path, but if `inputStrings` were to have a value that's quoted spaces, this would break.

packages/core/lib/console.js
```diff
@@ -284,9 +282,10 @@ class Console extends EventEmitter {
   }
 
   interpret(input, context, filename, callback) {
+    // processInput returns a sanitized string
     const processedInput = processInput(input);
     if (
-      this.command.getCommand(processedInput, this.options.noAliases) != null
+      getCommand(processedInput.split(" "), {}, this.options.noAliases) !== null
```
cds-amal, line 288: Are quoted strings a concern?

cds-amal (changes requested):
This is promising! We should have tests to verify the static commands list reflect the commands on disk. I also had some questions on argument parsing.

packages/core/lib/console-child.js
eggplantzzz: The way it is structured now, `inputStrings` is an array with each member being the arguments passed to Truffle. In other words, if a user types `truffle compile --all false` this should be `["compile", "--all", "false"]`. This is what `getCommand` and `prepareOptions` takes. We could re-do how Truffle ingests this stuff if you have another idea, I haven't thought about it too much and tried to leave it similar to how it has been in the past.

packages/core/lib/commands/develop/run.js
eggplantzzz: I have no problem with that as long as we are sure we don't want it running in this context. The real question is 'do we want to be able to run db in the repl?'.

packages/core/lib/console.js
eggplantzzz: I'm not sure, can you give an example of what you mean?

eggplantzzz:
@cds-amal I definitely like having a test to ensure a 1-1 between whats in `lib/commands/commands.js` and `lib/commands/index.js`. I'll add a quick one here.

packages/core/lib/console.js
cds-amal: Command line arguments aren't strictly whitespace separated and should account for quoted arguments.  Here's a pretty contrived example.

```
"--breakfast 'ham and eggs' "
split by space:  [ '--breakfast', "'ham", 'and', "eggs'", '' ]
```

packages/core/lib/commands/develop/run.js
cds-amal: I vote `no` because it stops REPL behavior, and currently we can't stop the process with ctrl-c, ctrl-z, or other keystrokes within the locked-up terminal.

packages/core/lib/console.js
eggplantzzz: I think this problem must exist in current Truffle as well.

packages/core/lib/commands/develop/run.js
eggplantzzz: Ok, I'm going to blacklist this here unless someone else wants to voice a dissenting opinion here.

packages/core/lib/console.js
cds-amal: The changes introduce a split on space though.

eggplantzzz: It only changes where the split is performed. [Here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/command.js#L117-L119) is where it is done in the current code. This will change what `getCommand` receives but shouldn't affect the outcome since that function is not concerned with anything but the first string.

lsqproduction (approved):
This is looking good to me.

packages/core/lib/console-child.js
cds-amal: ok!

packages/core/lib/console.js
cds-amal: Oh, it used a split on space-sep before. :+1:

cds-amal (approved):
Looks good @eggplantzzz. CI says it's ready to go!

PR #5113, opened 2022-05-16
Enhancement: Add cronos testnet to etherscan source fetcher

haltman-at:
Looks like Etherscan just added this network.  Adding it here.

PR #5115, opened 2022-05-17
Enhancement: Add more networks to sourcify source fetcher

haltman-at:
Looks like Sourcify just added a bunch of networks.  (Including finally adding Oneledger mainnet; I don't know why previously they only supported the testnet.)  Adding them.  (Would have combined this with #5113 had I realized!)


cds-amal (approved):
lgtm

PR #5119, opened 2022-05-17
Internal improvement: Fix name of network in source-fetcher

haltman-at:
Sorry, just a quick renaming, sorry about all these tiny source-fetcher PRs!  But it seems that this "wagmi" network is part of Avalanche?  So I figured I'd alter our name for it appropriately.  That's all.


cds-amal (approved):
Makes sense. Thanks @haltman-at

PR #5122, opened 2022-05-18
Update truffle init solc version to 0.8.14

haltman-at:
I used `--no-verify` to get it to leave the comma in after `compilers`. :P


sukanyaparashar (approved):
Wow! New solc version. Nice! Looks good to me 👍

Issue #4825, opened 2022-03-09
Types are not readily usable on Contract object


wbt:
## Issue
When attempting to use @truffle/contract in Typescript, the default types don't permit use as documented, and have assorted other redundancies and issues.

## Steps to Reproduce
In a new folder/project, 
`npm i @truffle/contract`
Create a file `example.ts` with the following contents:
````typescript
import TruffleContract from '@truffle/contract';
const getDeployedInstanceFromContract = function(truffleContract: TruffleContract.Contract) {
	return new Promise(function(resolve, reject) {
		truffleContract.deployed()
		.then(function(instance) {
			//Initial processing here which motivates use of a wrapper omitted for minimal example.
			resolve(instance);
		}).catch(function(err: unknown) {
			//Error processing here omitted for minimal example.
			reject(err);
		});
	});
};
````
## Expected Behavior
Based on the [documentation](https://www.npmjs.com/package/@truffle/contract), I would expect this to compile and run without a problem. 

## Actual Results
The call to .deployed() has a compilation error: 
>This expression is not callable.
  No constituent of type 'string | number | boolean | any[] | { [k: string]: any; }' is callable.ts(2349)

The next line also has an error:
>Parameter 'instance' implicitly has an 'any' type.ts(7006)

The second is more easily overcome but raises the issue that there should be a separate type for this (e.g. TruffleContract.Instance) and there doesn't appear to be one, though the "Contract Instance API" is broken out under a [separate heading in documentation](https://www.npmjs.com/package/@truffle/contract#user-content-contract-instance-api) that presents a separate set of available functions.  The documentation there also notes that a function call returns an object that "looks like this" but doesn't have a Type specification available in the Truffle package.

## Cause Analysis
The typings for @truffle/contract create a namespace TruffleContract with just one member: 'Contract', which is a renamed version of `@truffle/contract-schema/ContractObject` which is a passthrough from `@truffle/contract-schema/spec/ContractObject`.
The index.d.ts file there disables the linter and says the following:
> This file was automatically generated by json-schema-to-typescript.
> DO NOT MODIFY IT BY HAND. Instead, modify the source JSONSchema file,
> and run `yarn prepare` inside your local packages/contract-schema

This typing file is not very useful.  It has a lot of repetition in typings, for example from `ABI`:
````typescript
//... 
 | (
      | {
          [k: string]: any;
        }
      | {
          [k: string]: any;
        }
    )
  | (
      | {
          [k: string]: any;
        }
      | {
[...]

PR #5142, opened 2022-05-25
Bug fix: Remove incorrect typings from contract

eggplantzzz:
In response to https://github.com/trufflesuite/truffle/issues/4825, this PR removes the types from @truffle/contract.

### DIFF
--- a/packages/contract/typings/index.d.ts
+++ b/packages/contract/typings/index.d.ts
@@ -1,7 +1,0 @@
-declare module "@truffle/contract" {
-  import { ContractObject } from "@truffle/contract-schema";
-  namespace TruffleContract {
-    export type Contract = ContractObject;
-  }
-  export default TruffleContract;
-}


haltman-at (approved):
We said we're doing this, right? So let's do it!  (Although, should this be marked as a breaking change to `contract`?)

eggplantzzz:
@haltman-at oh man, I don't know. I guess they never worked so maybe not?

cds-amal (approved):
Outta here!

wbt:
Yep...this was a breaking change on @truffle/contract put out on a [patch-level bump](https://github.com/trufflesuite/truffle/compare/v5.5.15...v5.5.16#diff-88b68240d0400d2e6be5407cf556e6c243a84926340fb220afca4766690f97d6).  I'd really hoped this cluster of projects would get away from that practice, but it's unfortunately still fairly common. 

Prior to merging this, there was a way to get Typescript to compile a TypeScript file with the line "import TruffleContract from '@truffle/contract';" (though it would drop that import in the transpiled code) and now previously working code can't even get past that line in failing to compile:

>error TS7016: Could not find a declaration file for module '@truffle/contract'. '/project/path/node_modules/@truffle/contract/index.js' implicitly has an 'any' type.   Try `npm i --save-dev @types/truffle__contract` if it exists or add a new declaration (.d.ts) file containing `declare module '@truffle/contract';`

PR #5143, opened 2022-05-26
Bug fix: Fix behavior of stepOver and stepOut on function definitions

haltman-at:
Addresses #4561, supersedes #4677.

This PR fixes some longstanding problems with the debugger's stepOver and stepOut functionality.  Specifically:
1. `stepOut` now behaves consistently; the unnecessary special case it had has been removed
2. When `stepOver` is used while on a function definition, the function will (usually) be stepped over rather than into

How does this work?  Well, (1) there's not much to say about... I just removed the special case that shouldn't have been there.

As for (2), I made it so that, if you are on a function definition, then `stepOver` will defer to `stepOut` instead.  Note that this relies on being able to *determine* that you're on a function definition, so this won't work with Vyper (or with Solidity older than 0.4.12, or assembly/Yul older than 0.6.0).

It also requires some wonky logic for Yul functions, to ensure that this is only applied when we're actually *entering* the Yul function, but this is just the same logic we were already using for this purpose in `data`; said logic has now been factored out into a selector in `sourcemapping`.

There is a special case on top of this special case, though.  If you're on the definition of a base constructor, then we don't have a good way to step over *only* that base constructor; so as a fallback we just don't apply our special case here.  (You might expect we could use modifier depth for this, but, well, modifier depth has a number of problems...)  That means we'll step into the base constructor rather than over it in any way, which is not entirely satisfactory, but I don't see a better alternative.  (Stepping over the entire constructor would be going too far, I think; better to not go far enough than to do that.)

To check for this, we check whether you're on a constructor definition, and whether the ID of the contract you're in (in terms of source location) matches the contract ID of the context you're in; if you're on a constructor definition but the IDs differ, you're in a base constructor.  (Since this means that you're on the definition of a constructor for a contract other than the one you're actually constructing.)

Also note that since we can't use `data` in the controller, to get the contract definition, we use the same hack we use in `stacktrace` (which has been moved over to `sourcemapping`, and both `stacktrace` and `controller` pull it from there, so we don't have the same logic in two places).

Other changes:
1. I clarified the help text for `(i)`.  Yeah, OK, that probably should have been a separate PR, but whatever, it's so minor...
2. I added tests!  We just didn't really have any tests for stepping.  So I added some rudimentary tests for `stepNext` and `stepInto`, and tests of the improved `stepOver` and `stepOut` functionality, both for Soliidty and for Yul.  There could probably be more here, but I figured this was a good start.

### DIFF
--- a/packages/debug-utils/index.js
+++ b/packages/debug-utils/index.js
@@ -35,7 +35,7 @@ const verbosePanicTable = {
 
 const commandReference = {
   "o": "step over",
-  "i": "step into",
+  "i": "step line / step into",
   "u": "step out",
   "n": "step next",
   ";": "step instruction (include number to step multiple)",
--- a/packages/debugger/lib/controller/sagas/index.js
+++ b/packages/debugger/lib/controller/sagas/index.js
@@ -188,13 +188,16 @@ function* stepOver() {
   let finished;
 
   //special case: what if you're on a function definition?
-  //note that we exclude base constructors here, because we don't have a
-  //good way to go to the end of a base constructor; we'll just perform
-  //an ordinary stepOver in that case
   if (
     startingNode &&
-    startingNode.nodeType === "FunctionDefinition" &&
-    !(yield select(controller.current.onBaseConstructorDefinition))
+    ((startingNode.nodeType === "FunctionDefinition" &&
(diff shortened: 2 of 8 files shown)


cds-amal (approved):
Looks good @haltman-at. Let's get it in!

PR #5156, opened 2022-05-31
Add support for more networks to sourcify source fetcher

haltman-at:
Sourcify added more networks, updating!

### DIFF
--- a/packages/source-fetcher/lib/sourcify.ts
+++ b/packages/source-fetcher/lib/sourcify.ts
@@ -59,6 +59,8 @@ const SourcifyFetcher: FetcherConstructor = class SourcifyFetcher
     "avalanche",
     "fuji-avalanche",
     "wagmi-avalanche",
+    //sourcify does *not* support non-testnet dfk?
+    "testnet-dfk-avalanche",
     "telos",
     "testnet-telos",
     "ubiq",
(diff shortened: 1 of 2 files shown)


packages/source-fetcher/lib/sourcify.ts
```diff
@@ -59,6 +59,8 @@ const SourcifyFetcher: FetcherConstructor = class SourcifyFetcher
     "avalanche",
     "fuji-avalanche",
     "wagmi-avalanche",
+    //sourcify does *not* support non-testnet dfk?
```
cds-amal, line 62: I'm confused by this question in `sourcify.ts` that precedes something named, `testnet-dfk-avalanche`.

cds-amal (commented):
Hey @haltman-at, I have a question about the `non-testnet dfk` comment, otherwise lgtm.

packages/source-fetcher/lib/sourcify.ts
haltman-at: Yeah so DFK and WAGMI are Avalanche "subnets"?  I don't know what that actually means.  But you'll notice (per my practice of always including the mainnet -- or in this case the mainsubnet? -- in `networks.ts` even if it's not supported yet) I put both `dfk-avalanche` and `testnet-dfk-avalanche` in `networks.ts`.  These are different networks with different chain IDs.  However Sourcify only supports the latter atm?  So yeah I dunno.

cds-amal: Thanks @haltman-at

cds-amal (approved):

Issue #4900, opened 2022-03-20
`require` statement does not emit the given error-message


barakman:
## Issue

I believe that this is a bug in the compiler, in which, under very specific circumstances, a `require` statement does not emit the given error-message.

So I have initially posted it on [Solc GitHub](https://github.com/ethereum/solidity/issues/12819).

However, the issue might actually be on Truffle and/or Ganache side, so I am posting it here as well.

## Environment

- Compiler version: 0.8.12
- Ganache version: 7.0.1
- Truffle version: 5.4.33
- Operating system: MacOS
- node version (`node --version`): v16.13.0
- npm version (`npm --version`): 8.1.0
- yarn version (`yarn --version`): 1.22.10

## Steps to Reproduce

```solidity
contract MyContract {
    struct Set {
        address[] _values;
        mapping(address => uint256) _indexes;
    }

    Set private addressSet;

    function test(address[] calldata addresses) external {
        for (uint256 i = 0; i < addresses.length; i++)
            require(add(addresses[i]), "already exists");
    }

    function add(address value) internal returns (bool) {
        return _add(addressSet, value);
    }

    function _add(Set storage set, address value) private returns (bool) {
        if (set._indexes[value] == 0) {
            set._values.push(value);
            set._indexes[value] = set._values.length;
            return true;
        } else {
            return false;
        }
    }
}
```
When function `test(addresses)` is executed with an address which has already been passed to it in a previous execution, it should revert with an error-message `"already exists"`.

Indeed, this error-message is emitted when we do so in two different transactions:
```
await myContract.test([ADDRESS1, ADDRESS2]);
await myContract.test([ADDRESS1, ADDRESS2]); // reverts with `already exists`
```
However, when we do so in a single transaction:
```
await myContract.test([ADDRESS1, ADDRESS1]);
```
The transaction reverts with `exited with an error (status 0) after consuming all gas`.
I suppose that this error-message is a Truffle and/or Ganache implementation detail, so theoretically, the bug is there and not in the compiler.

However, each one of the following observations leads me to believe that this is in fact a compiler issue:
1. The correct error-message is emitted if we change the input type from a dynamic array `address[] calldata addresses` to a static array `address[2] calldata addresses`.
[...]

Issue #5131, opened 2022-05-20
ExpectRevert not working when using HDWalletProvider


cipaterson:
- [x ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

When using HDWalletProvider in the truffle-config.js file, the tests are not working as expected.
When a call reverts, the promise doesn't fail as expected but is resolved with the reverted transaction
As soon as the development network is not user HDWalletProvider the tests will be successful

## Steps to Reproduce

git clone https://github.com/cipaterson/HDWalletProviderIssue
cd HDWalletProviderIssue
npm install
ganache -m "ahead nasty leopard trick demise trap ten spoil festival comic unlock shallow"
truffle test # Will fail
truffle test --network dev # will succeed

## Expected Behavior

~/HDWalletProviderIssue master !1 ?1 > truffle test --network dev                                           7s 18:38:07
Using network 'dev'.

Compiling your contracts...
===========================
> Compiling ./contracts/Test.sol
> Compiling @openzeppelin/contracts/access/Ownable.sol
> Compiling @openzeppelin/contracts/utils/Context.sol
> Artifacts written to /var/folders/b2/_77sbqqj34d1h8f91dz9229h0000gn/T/test--80927-jlutVldRrnS6
> Compiled successfully using:
   - solc: 0.8.13+commit.abaa5c0e.Emscripten.clang

  Contract: Stacking Test Suite
    Owner: Test to attach tokens
owner 0x64729B74061b020Adfd7Bbc5148C4f3CD516044E
user1 0x842d479e85a854A74a736Ba6E2cC888b9C68b887
catch
      ✔ should reject for not owner caller

  1 passing (69ms)

## Actual Results

~/HDWalletProviderIssue master !1 ?1 > truffle test                                                        57s 18:37:52
Using network 'development'.

Compiling your contracts...
===========================
> Compiling ./contracts/Test.sol
> Compiling @openzeppelin/contracts/access/Ownable.sol
> Compiling @openzeppelin/contracts/utils/Context.sol
> Artifacts written to /var/folders/b2/_77sbqqj34d1h8f91dz9229h0000gn/T/test--80629-XGUzFlNdOGPY
> Compiled successfully using:
   - solc: 0.8.13+commit.abaa5c0e.Emscripten.clang

  Contract: Stacking Test Suite
    Owner: Test to attach tokens
[...]

PR #5158, opened 2022-06-01
Bug fix: Properly error out when estimateGas fails (and stacktracing is disabled)

haltman-at:
Addresses #4900.

So, this one's a bit messy.  I am hoping that I have done this right.  Or as right as it can be, anyway; there's some real hackery I've put in here, and I don't think there's any way to truly do this *properly*, but I'm hoping this is as close as reasonably possible, or at the very least close enough.

The problem: If `estimateGas` failed (and no explicit gas value was set), Truffle Contract wouldn't error out, but would continue ahead with a default gas value.  This is correct behavior in the special case of stacktracing mode, but incorrect behavior generally (again, per #4900).

The basic cause of the problem is that in `execute.js`, in `getGasEstimate`, the `catch()` function simply accepted, rather than rejecting with any sort of error.  So the easy fix is to simply reject with the error.  However, this is not a satisfactory solution on its own.  The problem here is that the error one will get from `estimateGas` is *different* from the error one would get from actually sending a reverting transaction.

Now, I should be clear -- to a large extent, that's just how things are going to be now; and, as I'll get to in a bit, I ended up deleting some tests as a result.  But, ideally, we would like to include some of the same information in this new error that we had in the previous error.  Like, there should be a `reason` field if applicable, and the error message (and stack trace) should include the decoded reason... including if it's a `Panic(uint)`, not only if it's an `Error(string)`.

So, this is where things get hacky.  We take the error, and then we Frankenstein additional properties onto it (and yes, we mutate... no cloning here, how are you going to properly clone an error?).  First, we use `reason.js` to get the reason; you'll notice we pass in `{ error }`, rather than `error`, in order to hit the format that `reason.js` expects.  Then we set `error.reason = reason`.  And then -- this is where things get real hacky -- we even modify `error.message` and `error.stack` (if applicable; that's what those dollar signs are for).  That last part obviously can potentially have false positives, but I'm betting on it not really coming up too much in practice.  Relying on `message` or `stack`  is already hacky business; just distinguishing a `Panic(uint)` from an `Error(string)` that way is already not perfectly reliable, after all.

Still, the result is, I hope, OK.  By the way, I should note that this won't work very well on Ganache 2, since it won't give you the binary encoded revert message.  So there you're just stuck with the decoding that Ganache gives you.  Still, better than nothing!  Like, you will still get its decoding, nothing I do here interferes with that; although it's worth noting that `reason` won't get attached in that case; its decoding will only appear in the message and stacktrace.

This was enough to get most of the contract tests to pass.  But there were a few sticklers that I wound up simply removing (well, removing particular checks in the tests, not removing the tests wholesale).  I'm hoping this is acceptable; I'll describe the remaining ones below and why I thought they were OK to remove.

Of the five removed checks, four of them were checking that we got "intrinsic gas too low" errors in certain situations.  As best I can tell, `estimateGas` just doesn't produce this error -- after all, how would it?  (Also, uh, they were checking that we got that when a constructor reverted, rather than when the intrinsic gas was actually too low...?  What...?  I'm assuming someone can explain this but I sure don't understand.)  Since I didn't see any reasonable way to fix this, I simply removed those checks.

The other removed check was about checking that we got the receipt back in the error.  Obviously, that isn't going to be possible if you didn't actually send the transaction.  So, that had to go.

You'll notice I also altered the debugger tests... there are a number of tests of debugging failed transactions, so I added gas values to those to get the transactions to actually occur.  While I was at it, I did a bit of factoring of the gas limits used in those tests.

Additionally, while I was doing all this, I also made some minor simplifications to the code.

So... I'm hoping this is OK.  It's a bit of a mess, but it's the best I could come up with...?


haltman-at:
Also it only occurred to me that instead of deleting those checks, I could duplicate those tests -- have one version without gas specified, and one with, and the one with could keep the old check in.  However when I tried that just now it didn't work for some reason.  But, I could investigate and try that again if people think it's worth having those checks!

eggplantzzz:
It's kind of complicated but this may be related to https://github.com/trufflesuite/truffle/issues/5131

packages/contract/lib/execute.js
```diff
@@ -208,18 +225,12 @@ const execute = {
...
-              constructor,
-              params,
-              network.blockLimit,
-              stacktrace
-            );
-          } catch (error) {
-            promiEvent.reject(error);
-            return;
-          }
+          params.gas = await execute.getGasEstimate.call(
```
eggplantzzz, line 228: Do we not need to reject the `promiEvent` any more? I'm not sure offhand how exactly this should work. If it throws here it will just stop all execution so the `promiEvent` shouldn't matter any more?

packages/debugger/test/data/returnvalue.js
```diff
@@ -297,7 +297,7 @@ describe("Return value decoding", function () {
     let instance = await abstractions.ReturnValues.deployed();
     let txHash;
     try {
-      await instance.fail(); //web3 throws on failure
+      await instance.fail({ gas: gasLimit }); //web3 throws on failure
```
eggplantzzz, line 300: Do we only deal with `gas`? Should we eventually handle `gasLimit`? [Here](https://ethereum.org/en/developers/docs/transactions/) in the params section they list `gasLimit` and not `gas`.

packages/debugger/test/data/returnvalue.js
```diff
@@ -322,7 +322,7 @@ describe("Return value decoding", function () {
     let instance = await abstractions.ReturnValues.deployed();
     let txHash;
     try {
-      await instance.failNoisy(); //web3 throws on failure
+      await instance.failNoisy({ gas: gasLimit }); //web3 throws on failure
```
eggplantzzz, line 325: We need to do this because previously it would fail on tx execution and now it will fail on gas estimation, correct?

packages/contract/lib/execute.js
```diff
@@ -56,7 +57,23 @@ const execute = {
               : accept(bestEstimate.toHexString());
           }
         })
-        .catch(() => accept());
+        .catch(error => {
```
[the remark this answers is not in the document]
haltman-at: Yeah -- I agree that this way of doing things is pretty wonky.  I don't really think it makes a lot of sense that `stacktrace` is a parameter to `interfaceAdapter.estimateGas`; IMO it would make sense to have the catch here, not there.  But since it already is that way, I decided not to do extra work by changing it.  However, I can totally change it if you think that would be a good idea.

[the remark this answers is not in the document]
haltman-at: (See discussion below.)

packages/debugger/test/data/returnvalue.js
haltman-at: Yup, exactly.  Putting in `gas` because we want the tx to happen and get reverted, rather than not to happen at all.

packages/contract/lib/execute.js
haltman-at: As best I can tell, simply allowing this to throw works fine.  Doing this didn't break any tests, after all.  I was not able to discern any difference between throwing and having the `promiEvent` reject.

packages/debugger/test/data/returnvalue.js
haltman-at: Regarding the first, Truffle and web3 call it `gas`.  That's surprising that that documentation calls it `gasLimit`.  I've never seen it called that before.

As for the second, yes, it is.  Why this value?  I dunno, but honestly, it doesn't particularly matter.  I left the actual value unchanged from what it already was.

packages/contract/lib/execute.js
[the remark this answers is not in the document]
haltman-at: Well, no, it's not a matter of what error it would throw; it's a matter of whether stacktracing is on or not.  My point is about where that should be checked; to my mind, that should be checked here, not in `estimateGas`.

packages/contract/lib/execute.js
haltman-at: Oh, duh -- the reason it makes no difference is because the whole thing is wrapped in `.catch(promiEvent.reject)`!

haltman-at:
OK, so, I went back and made a few changes to this.  Specifically:

1. Those tests I removed -- I put them back in and split them in two.  For each one, there's a no-gas-specified version with the check removed, and a version with the check still in but with gas specified.
2. However, some of those tests still weren't passing!  I had to make a change to `override.js` so that it wouldn't limit the gas when getting the reason string.  (Is that OK?  Should it be a high default instead of nothing?  IDK.)
3. Finally, inspired by `override.js`, I changed my own error-mangling to do what it does; I changed it to leave `stack` alone, to have it alter `message` in a less hacky way.

packages/contract/lib/execute.js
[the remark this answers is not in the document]
eggplantzzz: Yes that is what I was trying to say. But I'm also saying that we need to check in the `catch` to make sure it is throwing for the right reason...unless the only case in which `estimateGas` fails is when something reverts.

So if something reverts, `estimateGas` will throw, right? In the case that stacktracing is on, we can ignore the error and proceed. Will `estimateGas` ever fail for a different reason? In that case we would want to throw and not continue regardless of whether stacktracing is on. Does that seem correct?

packages/debugger/test/data/returnvalue.js
eggplantzzz: Maybe we should set another constant named something like `lowTxGasLimit` or something to avoid future confusion eh?

packages/contract/lib/execute.js
eggplantzzz: oh haha, so it will reject anyways...cool - this is confusing

packages/contract/lib/execute.js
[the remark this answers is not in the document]
haltman-at: Oh, that's a good point, I didn't think of that.  Yeah I agree there's a potential problem there.  Perhaps that can be saved for a later PR, though, since the problem is pre-existing?  Detecting the error type may require some trickiness in general.

packages/debugger/test/data/returnvalue.js
haltman-at: Yeah to my knowledge it doesn't matter... I think `gasLimit` is probably fine as a name?  Calling it `arbitraryGasLimit` doesn't sound to me like it really adds much meaningful.

packages/contract/lib/execute.js
haltman-at: Yeah :-/

packages/contract/lib/execute.js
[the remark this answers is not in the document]
eggplantzzz: Sounds good to me, I'll leave it up to you!

packages/debugger/test/data/returnvalue.js
eggplantzzz: Yeah but I kind of feel like people (including myself) get confused enough with the gas thing as it is that it is worth making it a separate case. I guess if you really don't want to do it I'll back off.

haltman-at: I guess I could rename it to `blockGasLimit`, if you think that would be clearer?  Of course it would then *also* be used for the transaction gas limit, but that does at least make it more explicit that we *are* reusing the block gas limit as the tx gas limit.  Does that sound good to you?

Otherwise I'd rather leave it alone; it's just test code, after all.

packages/contract/lib/execute.js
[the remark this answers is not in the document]
haltman-at: Cool, I'll leave it for a later PR, and file an issue once this is merged.

packages/debugger/test/data/returnvalue.js
eggplantzzz: Why couldn't we just export two variables with different names even if they have the same value so it is clear exactly what each value is for? Doesn't it kind of seem hacky to re-use a block gas limit constant for a tx gas limit? I mean, those two terms have similar names but they are very different things.

I know it is just test code but a lot of people are going to be working on this in the future and I can foresee confusion when trying to reason about tests.

haltman-at: > Why couldn't we just export two variables with different names even if they have the same value so it is clear exactly what each value is for?

Hm, I guess I could go with that?  Seems a bit more than is necessary, but it is at least explicit.  I'll go do that I guess.

eggplantzzz: thanks @haltman-at!

haltman-at: OK, I've gone ahead and made this change.

eggplantzzz (approved):

PR #5164, opened 2022-06-02
Add non-testnet DFK to sourcify fetcher

haltman-at:
Didn't include this one in #5156 because Sourcify didn't support it yet... but now it does.  So there you go.


cds-amal (approved):
Nice! Sourcify is consuming all the things!

Issue #5134, opened 2022-05-21
DISABLE_ESLINT_PLUGIN: command not found


swisstackle:
---------------------------
I figured I would transfer this problem to github issues because I dont believe that thats what the discussions are for.

## Issue
I am currently trying to build all packages on my local machine to contribute to truffle:

When yarn/lerna (yarn bootstrap) tries to build the truffle dashboard package, I am getting the error 'DISABLE_ESLINT_PLUGIN' is not recognized as an internal or external command, which is used in the package.json to run react-scripts without eslint.

According to react-scripts, this tag was added with the version 4.0.2, but the package.json in the dashboard package references version ^4.0.0 for react-scripts. This is a little bit confusing to me, because how could you guys build the dashboard package with this tag, if you guys use version ^4.0.0? But then again, the carrot makes sure that the latest version is installed, so the issue might clearly be on my side.

Here is the PR that added the 'DISABLE_ESLINT_PLUGIN' tag to react-scripts: https://github.com/facebook/create-react-app/pull/10170

## Steps to Reproduce

1. Clone the develop branch
2. run yarn bootstrap
## Expected Behavior

All packages are being built

## Actual Results

Error: ['DISABLE_ESLINT_PLUGIN' is not recognized as an internal or external command]

## Environment

* Operating System: Windows 10 (x64)
* Truffle version (`truffle version`): latest develop status
* node version (`node --version`): 16.14
* npm version (`npm --version`): 8.10.0

cds-amal:
Hi @swisstackle, truffle dashboard [uses craco for script management](https://github.com/trufflesuite/truffle/blob/develop/packages/dashboard/package.json#L24-L32). See their [website for more information](https://github.com/gsoft-inc/craco).

swisstackle:
I just installed craco globally, but I am still getting the same error. Technically that shouldnt make a difference anyways, because without having it installed globally, it should just use the craco that is being installed by yarn in the dashboard package.
Ill investigate further.

cds-amal:
Thanks, @swisstackle. Please let me know what you discover.

benjamincburns:
I think we should make the build work on Windows. We want people to be able to contribute code regardless of what OS they use.

@swisstackle do you happen to have a WSL install? If so, would you mind testing whether or not your changes also build correctly under WSL? If they do, I'd be really happy if you'd be willing to submit those changes as a PR to unbreak the build for Windows users.

swisstackle:
[...]

PR #5169, opened 2022-06-02
Transfered environment variables from package.json to .env file

swisstackle:
As of #5134 I was asked to create a PR to push the changes I have made to resolve the Dashboard Windows capability issues.

Perhaps we could just write "set DISABLE_ESLINT_PLUGIN=true  && set SKIP_PREFLIGHT_CHECK=true && craco build" as this is the  commandline syntax for Windows but I think that an env file is easier and cross platform compatible.

@benjamincburns

### DIFF
--- a/packages/source-fetcher/lib/networks.ts
+++ b/packages/source-fetcher/lib/networks.ts
@@ -24,6 +24,8 @@ export const networkNamesById: { [id: number]: string } = {
   43114: "avalanche",
   43113: "fuji-avalanche",
   11111: "wagmi-avalanche",
+  53935: "dfk-avalanche", //not presently supported by either fetcher, but...
+  335: "testnet-dfk-avalance",
   40: "telos",
   41: "testnet-telos",
   8: "ubiq",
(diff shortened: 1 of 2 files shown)


benjamincburns (approved):
Thanks for this change, @swisstackle!

benjamincburns (changes requested):
Actually would you mind changing this to use [`cross-env`](https://www.npmjs.com/package/cross-env) instead of the `.env` file?

This is because in the cases where we tend to use `.env` files we tend to use them for local-only config that shouldn't be committed. Using such a file here in another context might introduce a future security problem where someone thinks they've added a secret to their local `.env` and winds up committing it.

Also thanks to @michaeljohnbennett for catching this on review!

swisstackle:
Ok will do. Should I add cross-env as a dependency with yarn in /truffle or only in /truffle/packages/dashboard/package.json?

benjamincburns:
> @benjamincburns Ok will do. Should I add cross-env as a dependency with yarn in /truffle or only in /truffle/packages/dashboard/package.json?

Just in the dashboard package is fine, thanks!

PR #5173, opened 2022-06-03
Cross-Env added to Dashboard for Windows build compatibility.

swisstackle:
As of #5134 .
Tested on local windows machine.


benjamincburns:
Hi @swisstackle - thanks again for this change. It looks like the build is failing because you might have forgotten to include the changes to the `yarn.lock` file. We use this file to keep our dependency versions from changing on us unexpectedly. Would you mind pushing that as well?

swisstackle:
@benjamincburns  I don't believe that there are any changes to the yarn lock file because I only installed cross-env in the Dashboard directory with npm. Should I use yarn instead (inside the Dashboard directory)?

swisstackle:
I added it as a dev dependency with yarn.

cds-amal (changes requested):
Hi @swisstackle,

Package dependencies should be listed in the associated project's `package.json`.  Since Truffle uses lerna to manage our monorepo, the process to add a dependency is to use [lerna-add](https://github.com/lerna/lerna/tree/main/commands/add#readme) which will modify the correct `package.json` and resolve sibling package dependencies.

So, if you `git revert 5000cce6f` or `git reset --hard HEAD~1` and then did `.\node_modules\.bin\lerna add --scope @truffle/dashboard cross-env` from the truffle repo-root folder, it should work. You will have to `git push --force`, but since you're the only one working on the branch, it's fine.

Apologies, hopefully we'll get this documented in our `CONTRIBUTING.md` file.

benjamincburns:
> @benjamincburns I don't believe that there are any changes to the yarn lock file because I only installed cross-env in the Dashboard directory with npm. Should I use yarn instead?

As @cds-amal said, you'll want to add the `cross-env` dependency to the package-specific `package.json` rather than the top one. Truffle uses `yarn` exclusively for dev purposes, so you'll want to make sure that you then run `yarn bootstrap` in the base directory. To see what this is doing, have a look at the `bootstrap` script in the root `package.json`. After you run that, you should notice modifications to the `yarn.lock` file. You'll want to add those to your commit along with the other changes.

If you aren't familiar with `yarn.lock` (or `package-lock.json` from npm), I think [this StackOverflow answer](https://stackoverflow.com/a/39992365/203705) has a really good succinct explanation of what it is, and why it's important.

cds-amal (approved):
Thanks, @swisstackle! This is very helpful!

haltman-at:
Oops, looks like we left this one a bit too long and now there's a merge conflict that'll have to be fixed. >_>  Um, is there anyone who can be second reviewer on this...?  I mean I could approve it on the basis that it looks fine to me (although I'm wondering, should `cross-env` be a dev dep rather than a regular dep?), but as I don't know this stuff at all that seems maybe a little silly...

cds-amal:
Good catch @haltman-at! `cross-env` should be a dev dependency.

PR #5177, opened 2022-06-04
Upgrade to Web3 1.7.4

cds-amal:
This PR upgrades to latest web3 (1.7.4)

It turns out Truffle had a Ganache teardown issue with the decoder tests because the test providers used `eager instamine mode`. Truffle should use `instamine strict` mode, as it models how GETH and other clients behave. A good summary of how the different modes behave is [outlined in this doc](https://hackmd.io/wv7c5wn5Q0ebUWUH-N1H5A)

### DIFF
--- a/packages/decoder/package.json
+++ b/packages/decoder/package.json
@@ -32,7 +32,7 @@
     "@truffle/source-map-utils": "^1.3.87",
     "bn.js": "^5.1.3",
     "debug": "^4.3.1",
-    "web3-utils": "1.5.3"
+    "web3-utils": "1.7.4"
   },
   "devDependencies": {
     "@truffle/config": "^1.3.30",
@@ -50,7 +50,7 @@
     "mocha": "9.2.2",
     "tmp": "^0.2.1",
     "typescript": "^4.1.4",
-    "web3": "1.5.3"
+    "web3": "1.7.4"
   },
   "keywords": [
     "abi",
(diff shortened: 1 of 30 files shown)


packages/decoder/package.json
```diff
@@ -21,7 +21,7 @@
     "build": "tsc",
     "prepare": "yarn build",
     "start": "tsc --watch",
-    "test": "mocha ./test/current/test/* ./test/legacy/test/*"
+    "test": "mocha ./test/current/test/* ./test/legacy/test/* --exit"
```
gnidan, line 24: Can we make sure this isn't necessary when using the decoder as a library? (i.e., not via tests)

gnidan (dismissed):
(Requesting changes until it's clear what's going on with those open listeners that we're observing with the decoder tests)

packages/decoder/test/current/test/compatible-nativize.js
```diff
@@ -11,31 +10,34 @@ const Codec = require("@truffle/codec");
...
   let abstractions;
-  let compilations;
   let web3;
 
-  let Contracts;
-
   before("Create Provider", async function () {
     provider = Ganache.provider({
+      miner: {
+        blockTime: 1
```
cds-amal, line 22: This should be replaced with 
`instamine: "strict"`

packages/decoder/package.json
wbt: Is there a missing `provider.disconnect()` somewhere in there?

cds-amal: There were missing `await providerr.disconnect()` statements, as well as our tests using the incorrect Ganache instamine mode! Thanks to @MicaiahReid for the assist on tracking this down.

haltman-at:
Note that in addition to the conflict listed here, this also needs to have `db-kit` updated now that #5192 is merged.

haltman-at (approved):

Issue #4768, opened 2022-02-23
Dry-run public network check should go by chain ID rather than network ID


haltman-at:
## Issue

Currently, the dry-run public networks list works by checking the network ID.  It should probably instead check the chain ID, assuming that's possible.

## Environment

* Truffle version (`truffle version`): 5.5.1

Issue #4720, opened 2022-02-11
Dry-run public networks list is out of date


haltman-at:
## Issue

The [dry-run network whitelist](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/commands/migrate/determineDryRunSettings.js) has been sitting unmaintained for a long time.  It should probably be updated somehow; or, alternatively, we should find some other mechanism to replace it.

Now, this whitelist was introduced in #1028.  Going that PR, as well as the condition about `config.production` (which needs to be documented btw, but [that's its own issue](https://github.com/trufflesuite/truffle/issues/4721)), it sounds like the intention was for this to be a list of all public networks (as opposed to private development networks you might create with Ganache or similar).  But if that's the case, it's badly out of date!

We may need to start maintaining a more exhaustive list... our own version of the various `chains.json` files that [exist](https://github.com/ethereum/sourcify/blob/master/services/core/src/chains.json) out [there](https://chainlist.org/chains.json).  There's a little bit of this already in [source-fetcher](https://github.com/trufflesuite/truffle/blob/develop/packages/source-fetcher/lib/networks.ts), but that only covers the networks supported by at least one of Sourcify or Etherscan (or Etherscan's clones for other networks); it's not intended to cover *all* public networks.  And indeed, there are already networks on the dry-run whitelist that don't appear in source-fetcher, such as `99` (core), `7762959` (Musiccoin), and `61717561` (Aquachain).  So I don't think we can just use the list from source-fetcher!

(Although, if we are making our own `chains.json` or similar, then `source-fetcher` could use *that*, instead of having its own separate list!)

See also: #4768, #4769

## Environment

* Truffle version (`truffle version`): 5.4.33

lsqproduction:
We decided to use the new Ganache feature.

kevinbluer:
Available in [v5.6.18](https://github.com/trufflesuite/truffle/releases/tag/v5.5.18) 👍

eggplantzzz:
Looks like this can be closed.

PR #5183, opened 2022-06-07
Replace dry-run supported networks list with one from Ganache

haltman-at:
Addresses #4720.  Does **not** address #4768, I figured that can be handled by a future PR.

Anyway not much to say about this one, it's pretty straightforward execution of what was discussed earlier.


packages/core/lib/commands/migrate/determineDryRunSettings.js
```diff
@@ -1,19 +1,9 @@
...
-    99, // Core
+const Ganache = require("ganache");
 
-    7762959, // Musiccoin
-    61717561 // Aquachain
-  ];
+module.exports = function (config, options) {
+  //note: this is a list of chain IDs but we're still using
+  //network ID.  This should be fixed later.
+  const supportedChainIds = Ganache.__experimental_info().fork.knownChainIds;
```
cds-amal, line 6: `Ganache.__experimental_info()` implies the API behavior could change or disappear.

cds-amal (commented):
Thanks @haltman-at . It makes sense to consolidate Network/ChainId's source of truth. I'd feel a lot better if this information wasn't behind an experimental name. Like, what's the rule for semver and experimental public APIs? Can they change in a patch?

packages/core/lib/commands/migrate/determineDryRunSettings.js
eggplantzzz: Well it's not like this is an external package. Talk to the Ganache team and see what the plans are.

haltman-at: Yeah, this is the name @davidmurdoch & co settled on; note that this feature was added to Ganache, because I asked for it, specifically for it to be used for this *exact* purpose!  So I intend to use it regardless of what David and company call it!

cds-amal: The PR mentions the API will likely change in the future. 

> This introduces a new experimental and private (this will likely change in a future release!) __experimental_info export:

My concern is the risk of a Ganache patch release changing the behavior. I'm unsure how SemVer defines version bumps regarding experimental API changes.

haltman-at: So, are you suggesting that we pin Ganache alongside making this change, then?  Or what?

cds-amal: @gnidan, thoughts?

gnidan: Since this might break semver, I can see the argument for why we might want to pin.

But I'd still rather get that name switched before this PR gets merged, and then lean on semver (and expectations provided by Ganache) to stick with caret.

haltman-at: I mean, waiting for the name to get switched seems like it would significantly delay things, right?  I'd prefer to pin simply so this can be merged quicker.

gnidan: My thinking is that we can resolve the name thing by next week, but others may not be so optimistic.

Anyway, to be more explicit: if we go with the experimental, semver-abandoning name, then yes we should pin.

cds-amal: I vote for pinning for now, as Ganache has imminent releases queued up, and we're about to be OOO for a week.

davidmurdoch: I do not think that this breaks semver for Ganache; this is clearly marked as a private member, as denoted by the `_`. And hopefully the extra `_` followed by "experimental" communicates that it is going to break. It is also undocumented.

If a property prefixed with `__experimental` does not sufficiently communicate that this one thing will change in the future then humanity has no hope of making it through The Great Filter; may God have mercy on our souls.

gnidan: Yeah, @davidmurdoch don't take my words too harshly; I just mean that the use experimental features means that we can't rely on semver, and thus pinning is logically more warranted.

davidmurdoch: On last point: if removing something labeled `__experimental` without a major version bump is abandoning semver than fixing bugs is a much worse offender. :-)

davidmurdoch: relevant: https://github.com/semver/semver/issues/238#issuecomment-1149260160

davidmurdoch: > For this system to work, you first need to declare a public API. This may consist of documentation or be enforced by the code itself. Regardless, it is important that this API be clear and precise.

https://github.com/semver/semver/blob/master/semver.md

Per semver, it seems that documenting this endpoint as private (or at least "not public") in a clear and consider manner is all Ganache needs to do to be in absolute compliance. I still think__experimental_ is clear though. 🤷‍♂️

haltman-at: So... I guess I pin for now, then?  Or what?

gnidan: > So... I guess I pin for now, then?  Or what?

Sounds like!

haltman-at: OK, pin commit added.

cds-amal (approved):
Thanks @haltman-at, Looks good!

PR #5190, opened 2022-06-10
Internal improvement: Remove seemingly unused methods

eggplantzzz:
There are some methods in the deployer that seem unused. This PR gets rid of them. They look like "private" methods and I can't find any reference to them.


cds-amal (approved):
Ya, get it out of here!

PR #5192, opened 2022-06-15
Change transaction type to match web3-core transaction 

lsqproduction:
I was getting a type error during the `yarn bootstrap` process which was a legit error. 

```
@truffle/db-kit: src/cli/decodeTransaction/useDecodedTransaction.ts(50,31): error TS2345: Argument of type 'import("/Users/liangliang/code/truffle/node_modules/web3/node_modules/web3-core/types/index").Transaction' is not assignable to parameter of type 'import("/Users/liangliang/code/truffle/packages/decoder/dist/types").Transaction'.
@truffle/db-kit:   Types of property 'gas' are incompatible.
@truffle/db-kit:     Type 'number' is not assignable to type 'string | undefined'.
@truffle/db-kit: error Command failed with exit code 2.
```


packages/decoder/lib/types.ts
```diff
@@ -260,9 +260,9 @@ export interface Transaction {
    */
   gasPrice?: string;
   /**
-   * Gas provided by the sender, as numeric string.
+   * Gas provided by the sender, as number.
    */
-  gas?: string;
+  gas?: number;
```
haltman-at, line 265: Hm, so, I just checked what web3 actually returns, and it's a string, not a number.  As such I'm wary of changing this type -- the reason that this type is defined here, after all, rather than just using web3's types, is because web3's types are so frequently wrong!

So rather than changing this type, could there be another solution?  Like, maybe we just don't use the type from `web3-core` in the first place?  (What if we import from `@types/web3`?  Or from `web3` directly, which I assume would take from `@types/web3`?)  Or could one just coerce if really necessary?  (Can the `web3-core` dependency just go away, perhaps...?)

haltman-at (commented):
Hm, a bit uncertain if this is the right way to handle this problem.

gnidan (approved):
looks good to me! (although maybe @haltman-at should also take a look to confirm)

packages/db-kit/src/cli/decodeTransaction/useDecodedTransaction.ts
```diff
@@ -47,7 +47,7 @@ export function useDecodedTransaction({
   }
 
   useEffect(() => {
-    decoder.decodeTransaction(transaction).then(decoding => {
+    decoder.decodeTransaction(transaction as unknown as DecoderTransaction).then(decoding => {
```
cds-amal, line 50: Is casting to unknown required here?

cds-amal (commented):
Thanks @lsqproduction, I only have a casting question.

packages/db-kit/src/cli/decodeTransaction/useDecodedTransaction.ts
cds-amal: Why @gnidan? What would happen if the interim cast to unknown?

lsqproduction: yeah, sinch `transaction` and `DecoderTransaction` were different and it couldn't be cast directly, we need to first type it as `unknow` then to `DecoderTransaction`

haltman-at: I may need to update the `DecoderTransaction` type a bit with some of the new fields, dunno if that would help.  (Or @lsqproduction I don't know if you want to include that on this PR... looks to me like new fields are `type: number`, `maxFeePerGas: string`, and `maxPriorityFeePerGas: string`?  Dunno if that will actually improve things, as mentioned.)

cds-amal: What's the typescript error without the interim cast? I removed the interim cast locally, and it compiled without errors. I'm not sure what I'm doing wrong. OR is this a case where we're concerned with converting primitive types that we need the interim cast to unknown?

cds-amal: @lsqproduction and I discovered the truffle repo had an ancestor `$HOME/node_modules`, which breaks npm/yarn dependency guarantees. After removing the ancestor `node_modules`, we observed `yarn bootstrap` is consistent with other dev-machines. 

If you're curious you can try installing an older version of `web3` in an ancestor folder (`$HOME` should work!). 
```console
cd $HOME
npm install web3@1.5.0
cd path/to/your/truffle/repo
yarn bootstrap # should fail in db-kit
```

haltman-at (changes requested):
Hey, what happened to the web3-core dev dependency?  Either that should be added back in, or the line importing it should be deleted and the types should be taken from somewhere else.  (Wondering if we can do the latter, it seems to me to be the better solution... like can we get the types from `@types/web3` (presumably by importing from `web3` instead of `web3-core`)?  Because their types are better present, and their types are better...)

lsqproduction:
> Hey, what happened to the web3-core dev dependency? Either that should be added back in, or the line importing it should be deleted and the types should be taken from somewhere else. (Wondering if we can do the latter, it seems to me to be the better solution... like can we get the types from `@types/web3` (presumably by importing from `web3` instead of `web3-core`)? Because their types are better present, and their types are better...)


Thanks @haltman-at for the reminder. I've added web3-core as a dev dep in db-kit now. I've considered the latter options:
`@types/web3` is deprecated, and it does exist in db-kit dev dep already but not being used. Maybe I should remove it altogether. 

`web3` also just uses the type from `web3-core` unless I'm not understanding it correctly? `web3-core` lives under a package in `web3`.

haltman-at:
> Thanks @haltman-at for the reminder. I've added web3-core as a dev dep in db-kit now. I've considered the latter options: `@types/web3` is deprecated, and it does exist in db-kit dev dep already but not being used. Maybe I should remove it altogether.
> 
> `web3` also just uses the type from `web3-core` unless I'm not understanding it correctly? `web3-core` lives under a package in `web3`.

Oh, I think maybe the confusion here is that you're not aware of how `@types` packages work.  Sorry, I should explain that.  If you look around the codebase, you'll see that `@types` packages are frequently listed as dev dependencies, yet nowhere are they explicitly imported from.  That's because packages with these names are treated specially by Typescript.

The way it works is that `@types/foo` is used to supply types for `foo`.  So you never import explicitly from `@types/foo`; you import from `foo` and `@types/foo` is used instead.  (I may have the exact workings wrong here, this isn't really my area, but this is my understanding?)

So -- as I understand it -- if you import from `web3`, you'll get the type from `@types/web3`, not the actual type from `web3`.  Whereas if you import from `web3-core`, then, since there's no such package as `@types/web3-core`, you bypass this mechanism and get the actual type from `web3-core` (and thus `web3`).  So you should import from `web3` in order to *not* get the type from `web3`. :)

The problem here is that the types from `web3` are often wrong, and the types from `@types/web3` are typically right, which is why we use it.  I didn't realize it was deprecated, I have to say.  But I wouldn't suggest getting rid of it, not unless we want to have to deal with a bunch of incorrect types...

Sorry for the inclarity!

packages/decoder/lib/types.ts
```diff
@@ -267,6 +267,18 @@ export interface Transaction {
    * Data sent with the transaction, as hex string.
    */
   input: string;
+  /**
+   * A positive unsigned 8-bit number between 0 and 0x7f that represents the type of the transaction.
+   */
+  type?: number;
+  /**
+   * Defaulted to (2 * block.baseFeePerGas) + maxPriorityFeePerGas. The maximum fee per gas that the transaction is willing to pay in total.
```
haltman-at, line 275: Aha, thanks for these comments!  I'd suggest removing the bits about the defaults, they don't make sense in the decoder context.

packages/db-kit/src/cli/decodeTransaction/useDecodedTransaction.ts
haltman-at: Just to be clear, this really is necessary?

haltman-at (commented):
Seems good to me!  Two minor comments.  Going to go dismiss my request changes meanwhile. :)

packages/decoder/lib/types.ts
lsqproduction: That makes sense. thank you for catching it.

Issue #5083, opened 2022-05-05
Truffle should allow people to specify their `truffle exec` scripts as typescript, without compilation


benjamincburns:
## Feature

Similar to #4461, it would be nice if people could `truffle exec` typescript without needing to run a compilation step.

## Steps to Reproduce

Write a script for `truffle exec` in TypeScript

## Desired Behavior

Truffle would use `ts-node` to execute said script

## Current Behavior

Users must first compile said script, and `truffle exec` the compilation output.

ganganimaulik:
can you assign me this issue?

Issue #4328, opened 2021-09-25
Support ESM


soulofmischief:
## Issue

Truffle doesn't support ESM.

## Steps to Reproduce

1. Set `"type": "module"` in `package.json`.
2. Run truffle.

## Expected Behavior

The config loads, and ESM imports are correctly handled.

## Actual Results

The configuration doesn't load, because the `original-require` package which ingests `truffle-config.js` uses `require` and not `import`.
The error suggests to change the extension to `.cjs`, but this just causes a cascade of needing to change extensions until the entire project is `.cjs` files and no longer a module.

It's impossible to use latest versions of popular tools such as `node-fetch` which have switched over to ESM, and it's causing dependency rot.

## Environment

* Operating System: Linux (Fedora)
* Ethereum client: N/A
* Truffle version: 5.4.12
* node version: 16.5.0
* npm version: 7.19.1

dongmingh:
We will look into this issue. Thanks for raising this!

soulofmischief:
Thank you!

For anyone snooping for an intermediate fix for ESM packages like `node-fetch`:

**esm.js**
```
const
  fetchPromise = import( 'node-fetch' ).then( mod => mod.default ),
  fetch = ( ...args ) => fetchPromise.then( fetch => fetch( ...args ))


module.exports = { fetch }
```

Fardinak:
Any temporary workarounds for this? Changing the truffle-config file extension to `.cjs` and invoking commands returns the following:

```
Could not find suitable configuration file.
Truffle v5.5.3 (core: 5.5.3)
Node v17.6.0
```

Fardinak:
Just found a (seemingly undocumented) `--config` CLI argument that helped with that.

eggplantzzz:
@Fardinak so Truffle looks explicitly for `truffle-config.js`, if you rename it it will not find it. But you found the workaround; namely to pass the `--config` flag to the file.

Fardinak:
@eggplantzzz Yes, but unfortunately that did not prove useful in my case after all; since I'm also using Ganach. I gave up when I couldn't add my project.

In the end, I had to move the Truffle directory structure into a subdirectory and add a `package.json` with `type: "commonjs"` to work around the issue.

benjamincburns:
I ran into this while implementing TypeScript support for migrations.

[...]

PR #5222, opened 2022-06-22
Enhancement: Support writing migrations in TypeScript

benjamincburns:
Makes it so that users can write their migrations as TypeScript files.

Also includes a port of the `@truffle/require` package over to TypeScript.

Remaining work for full support of TypeScript migrations (not in this PR/merge):

- Figure out a sane (or at least the best possible) `tsconfig.json` setup so that IDEs can provide type hints for migrations
- Modify the default box so that `truffle init` or `truffle init --typescript` produces a sane default config
- Document TypeScript support on the website

We will likely tackle these items after baseline TypeScript support for `truffle exec` and `truffle test` are enabled.


packages/truffle/test/sources/migrations/typescript/migrations/global.d.ts
```diff
@@ -0,0 +1,4 @@
+export {};
+declare global {
+  const artifacts: any;
+}
```
benjamincburns, line 4: This file is just a temporary hack for now - this won't be necessary in user code when this is finished

packages/truffle/test/sources/migrations/typescript/migrations/2_migrations_sync.ts
```diff
@@ -0,0 +1,10 @@
+/// <reference types="./global" />
+/* global artifacts */
```
benjamincburns, line 2: The `reference types` line here won't be necessary in the final PR. I'm also going to try to find a way around the `global artifacts` comment so users who are using ESLint don't need to add that, but I have lower confidence on that one.

packages/require/require.js
```diff
@@ -131,4 +132,29 @@ const Require = {
...
+  const fileExtension = path.extname(file).toLocaleLowerCase();
+  const isTs = [".ts", ".tsx"].includes(fileExtension.toLocaleLowerCase());
+  if (isTs) {
+    try {
+      require("ts-node");
+    } catch {
+      throw new TruffleError(`Attempted to execute script with extension
+        ${fileExtension}, but the 'ts-node' module, or one of its required peers
+        has not been installed.`);
+    }
```
benjamincburns, line 146: This is necessary because `ts-node` is an optional peer dependency, and people who aren't using TS shouldn't need to install it.

packages/truffle/test/sources/migrations/typescript/migrations/2_migrations_sync.ts
cds-amal: We have `.eslint` [rules for this that target `*.js` files](https://github.com/trufflesuite/truffle/blob/develop/.eslintrc.truffle.json#L8-L16), perhaps extending the glob to include typescript would suffice?

packages/require/lib/compiler.ts
```diff
@@ -0,0 +1,94 @@
...
+import TruffleConfig from "@truffle/config";
+import { readFileSync, existsSync } from "fs";
+
+import originalRequire from "original-require";
+
+import vm from "vm";
+
+/**
+ * @returns compiler function from user's installation of `ts-node` if script is
+ * a ts or tsx file, otherwise `undefined`.
```
benjamincburns, line 14: I need to fix this comment. Will do before merge.

packages/require/lib/compiler.ts
```diff
@@ -0,0 +1,94 @@
...
+ */
+export function compile(conf: TruffleConfig, scriptPath: string): string {
+  const source = readFileSync(scriptPath, { encoding: "utf-8" });
+
+  // we only compile TS files, so just return the source unless we are dealing
+  // with one of those
+  if (![".ts", ".tsx", ".cts"].includes(path.extname(scriptPath))) {
+    return source;
+  }
```
benjamincburns, line 23: In theory we could pass `js` files through this, especially in the case where the user has `allowJs: true` in their `tsconfig.json`.

I decided against that however because 1) I'd need to add logic to detect this condition, and 2) I figured it'd be best to avoid performance degradation for existing projects.

packages/require/lib/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    sandbox[key] = conf.context[key];
+  });
+
+  const context = vm.createContext(sandbox);
+
+  const old_cwd = process.cwd();
+  const cwd = path.dirname(scriptPath);
+  process.chdir(cwd);
+
+  registerTsNode(conf, scriptPath, context);
```
benjamincburns, line 83: I think this chunk of code is probably the lowest-hanging fruit as far as performance improvement goes, but I left it alone for now to preserve behavior for JS code.

That is, to make this faster we could cache the script global context and only run `registerTsNode` once for the lifetime of that context.

packages/require/lib/compiler.ts
```diff
@@ -0,0 +1,94 @@
...
+        }
+      };
+
+      _compilationService = originalRequire("ts-node").create(createOptions);
+
+      const globalDeclarationsPath = existsSync(
+        path.join(__dirname, "scriptGlobals.ts")
+      )
+        ? path.join(__dirname, "scriptGlobals.ts")
+        : path.join(__dirname, "..", "scriptGlobals.ts");
```
benjamincburns, line 76: This ternary was necessary to make it so I could run truffle via `cli.js` as well as via the bundle.

packages/require/lib/compiler.ts
```diff
@@ -0,0 +1,94 @@
...
+      _compilationService!.compile(
+        readFileSync(globalDeclarationsPath, { encoding: "utf-8" }),
+        globalDeclarationsPath
+      );
```
benjamincburns, line 80: This adds the typings for the global context to the cached `compilationService` instance. I tested doing this a few different ways, and this method gave me the fewest (possibly even zero?) compromises as compared to the others.

Doing it this way and caching the service lets us read in the typings only once (rather than for each TS file we that we interpret), and it doesn't prepend data to the user's source, which avoids needing to deal with line number offsets when reporting type check, compilation, and runtime errors.

benjamincburns: Also note that for now these typings are static and don't change depending on what context the caller passes in. For now I'd like to take that as tech debt, as solving that issue wasn't necessary to make any of the existing code paths work.

**Please advise if you disagree and think that it should be included as part of this change.** I just didn't want to bite it off now, as while I have what looks to be a straightforward plan-of-attack for it, I encountered a bunch of unexpected problems while making just this change work, and I didn't want to delay getting TS migrations out to users any longer than absolutely necessary.

packages/require/lib/index.ts
```diff
@@ -0,0 +1,130 @@
+import path from "path";
+import TruffleConfig from "@truffle/config";
+import { options as expectOptions } from "@truffle/expect";
+import { Web3Shim, createInterfaceAdapter } from "@truffle/interface-adapter";
+
+import vm from "vm";
+import Module from "module";
+
+import originalRequire from "original-require";
```
benjamincburns, line 9: This is a holdover from the way that we were doing this before. Referencing `require` this way allows us to mark this module as external in our WebPack config, which gives us a reference to NodeJS's builtin `require` rather than WebPack's one.

packages/require/lib/index.ts
```diff
@@ -0,0 +1,130 @@
...
+  });
+  const web3 = new Web3Shim({
+    provider: options.provider,
+    networkType: options.networks[options.network].type
+  });
+
+  try {
+    const fn = this.file({
+      file: options.file,
+      context: { web3, interfaceAdapter },
```
benjamincburns, line 122: I just realized that while this codepath will attempt to compile TypeScript `truffle exec` scripts, it won't include the necessary type definitions for the `web3` and `interfaceAdapter` globals here.

I think that's fine personally, as we aren't claiming support for TypeScript in `truffle exec` here.

packages/require/lib/types.ts
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   *  Path to file to execute. Must be a module that exports a function.
+   */
+  file: string;
+
+  /**
+   * Arguments passed to the exported function within file. If a callback is not
+   * included in args, exported function is treated as synchronous.
+   */
+  args: any[];
```
benjamincburns, line 20: I kept this from the original code, but I couldn't see anywhere that this was actually referenced. Should I just remove it?

packages/require/lib/types.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export interface Callback<ResultType> {
+  (error: Error): void;
+  (error: null, val: ResultType): void;
+}
+
+export interface Provider {
+  send(payload: JsonRPCRequest, callback: Callback<JsonRPCResponse>): any;
+}
```
benjamincburns, line 60: Ideally these would live in some canonical types package somewhere

packages/require/lib/index.ts
```diff
@@ -0,0 +1,130 @@
...
+    exports: scriptModule.exports,
+    global,
+    process,
+    setImmediate,
+    setInterval,
+    setTimeout,
+    module: scriptModule,
+    config: conf,
+    artifacts: conf.resolver,
+    require: (id: string) => {
```
davidmurdoch, line 41: This implementation seems fine, probably.

I'd love an explanation as to why it is necessary in the first place. Why can't we pass in `require` (i mean `original-require`, once webpacked) directly?

At first I thought the answer was because the `require` available here shares a global scope with the parent script, and that could be a problem, but this doesn't seem to change that.

Is this all because of relative path issues? If so, I think there are built-in methods to resolve those we could use instead of all this. https://nodejs.org/docs/latest/api/modules.html#requiremain

packages/require/lib/index.ts
davidmurdoch: I wonder if you could do something quick and easy like `if (!context.process[Symbol.for("ts-node.register.instance")]) { registerTsNode(conf, scriptPath, context); }`?

packages/require/lib/types.ts
davidmurdoch: /shrug

packages/require/lib/compiler.ts
```diff
@@ -0,0 +1,94 @@
...
+    } catch (err) {
+      if (err.code === "MODULE_NOT_FOUND") {
+        throw new TruffleError(`Attempted to execute script with extension
```
davidmurdoch, line 83: For debugging it might be nice to have the original error's context as well. Not a blocker for me, but this definitely swallows some information that might be useful. Could `MODULE_NOT_FOUND` have been bubbled up from userland (a user `required` something that isn't installed)?

davidmurdoch (commented):
Wasn't able to do a super in depth review today. but wanted to shared what I have so far!

cds-amal (commented):
An initial review @benjamincburns. I will follow up later in the week once I test it locally. Would this support truffle-develop console? It would be nice to get a typescript test in that scenario

packages/require/lib/index.ts
eggplantzzz: Will this assignment work as desired if you do that? Why wouldn't it still use webpack's `require`?

davidmurdoch: Well, depending on where you to it you might have to do shenanigans like `(0, eval)("require")`. But the reason is works is because it gets `require` from the global scope, which belongs do node.

eggplantzzz: Oh so it kind of "bypasses" webpack's `require`?

davidmurdoch: right

packages/require/lib/index.ts
eggplantzzz: It looks like this function is here primarily to handle node modules that are installed in the same directory as the script?

packages/config/src/configDefaults.ts
```diff
@@ -129,7 +129,7 @@ export const configProps = ({
       transform: resolveDirectory
     },
     migrations_file_extension_regexp() {
-      return /^\.(js|es6?)$/;
+      return /^\.(js|ts|es6?)$/;
```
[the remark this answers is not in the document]
benjamincburns: I don't think that's a necessary feature, personally. I suppose it's easy enough for people to enable in their own truffle configs if they want it.

packages/require/.gitignore
```diff
@@ -1,3 +1,6 @@
 node_modules
-yarn.lock
-package-lock.json
+dist
+.vscode
```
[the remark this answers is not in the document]
benjamincburns: Ah I just copied this over from another package. I think it makes sense to leave the `*.vscode` (as vs code supports directory-level config, IIRC), but I can remove the `*.abi` and `*.bin` as they're not relevant.

packages/require/lib/compiler.ts
```diff
@@ -0,0 +1,94 @@
...
+import type { CreateOptions, Service } from "ts-node";
+import TruffleConfig from "@truffle/config";
+import { readFileSync, existsSync } from "fs";
+
+import originalRequire from "original-require";
+
+import vm from "vm";
+
+/**
+ * @returns compiler function from user's installation of `ts-node` if script is
```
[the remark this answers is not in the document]
benjamincburns: I thought I commented on this previously, but it appears to have gone missing. But yeah you're correct - I need to fix up the doc comments for all of this code.

packages/require/lib/compiler.ts
```diff
@@ -0,0 +1,94 @@
...
+ * a ts or tsx file, otherwise `undefined`.
+ */
+export function compile(conf: TruffleConfig, scriptPath: string): string {
+  const source = readFileSync(scriptPath, { encoding: "utf-8" });
+
+  // we only compile TS files, so just return the source unless we are dealing
+  // with one of those
+  if (![".ts", ".tsx", ".cts"].includes(path.extname(scriptPath))) {
```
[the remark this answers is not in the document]
benjamincburns: By "a constant," I assume you mean a constant variable rather than a literal?

I tend to follow the rule of three for things like this, but it's easy enough to move it for future use.

packages/require/lib/compiler.ts
benjamincburns: Good question. I purposefully excluded `.mts` files for now, as I didn't want the added complexity of dealing with EJS modules. I suppose I can imagine a world in which users will want EJS support, but for now I'm not sure what'd be driving that desire in migration script code other than a preference to use what's new.

packages/require/lib/compiler.ts
benjamincburns: > I wonder if our webpack process should tweaked to either put the compiled source in the "correct" relative location, or update these lines during bundling to point to the right place.

I think the detail that you might be missing here is that I don't want the compiled source. I need to be able to read the _uncompiled_ source and pass it through the compiler host to augment the existing global types.

I started looking at the road of a `raw` loader for this file so that I could read in just the file contents, but then I'd still need to handle the unbundled case separately. Otherwise I think this is a bit more self-contained of a solution as-is than adding a "magic" var (aka a hidden input) that's set by the webpack bundler process.

As for the idiom of trying to read the file and handling the exception, I suppose I can move this into a separate `_getGlobalContextSource()` function and then follow that idiom there. I was just a bit lazy doing it inline, and I wanted a `const` rather than a `let` variable declaration.

benjamincburns: > I need to be able to read the uncompiled source and pass it through the compiler host to augment the existing global types.

I'll add a comment to clarify this as well, as it's a subtle but important issue. My desire was to make use of a `.d.ts` file here rather than a `.ts` file, but there's no good way to hand a `.d.ts` file to the compiler host to modify the global context for all files going forward. You can point the compiler at it with a `typeRoots` value in the `compilerOptions` but then that will override the default `typeRoots`, or any `typeRoots` defined by the user's `tsconfig.json`.

packages/require/lib/compiler.ts
benjamincburns: Agreed - will do on both points

packages/require/lib/index.ts
benjamincburns: `eval("require")` works fine, but I get weird results with `global.require`. I see that `original-require` is used in a number of Truffle packages however, so I think I’ll leave removal of that package for another time

packages/require/lib/index.ts
benjamincburns: > I'd love an explanation as to why it is necessary in the first place.

This code is just ported over from the JS version, and it's 5 years old, so I don't have all of the context.

My understanding is that it's basically attempting to resolve webpacked modules first, then it falls back to the typical node built-in require and attempts to resolve them externally as per usual node behavior.

However as you sort of alluded to, it's a somewhat poor approximation of the actual `require`. I did attempt to rewrite it at one stage (with the goal of executing modules required by scripts running in the child context within that same child context) but ran into some complexities around relative paths for transitive dependencies.

The `vm2` module has solved this to some extent, but then they have hardened their VM sandbox to prevent modification to builtins (`source-map-support` apparently modifies built-ins), and annoyingly they always force strict mode, even if your source doesn't include the `"use strict";` directive.

packages/require/lib/index.ts
benjamincburns: I think that once contextified the `context` needs to be treated like an opaque type. But that said, `context.process` should be the same as `process`, which makes me wonder if all of these hoops that I'm jumping through to only register it in the child context are for nought.

packages/require/lib/index.ts
benjamincburns: I think TS support for `truffle exec` just needs its own feature ticket, yeah

packages/require/lib/types.ts
benjamincburns: I'll remove it and any other stuff that isn't referenced

packages/require/lib/compiler.ts
benjamincburns: If only error causes had been added earlier - I might see about adding a polyfill w/i `@truffle/error` to support them, as that'd be much nicer than rolling our own error chaining...

https://v8.dev/features/error-cause
https://node.green/#ES2022-features-Error-cause-property

benjamincburns: Ah I suppose I don't need to polyfill it - I can just attempt to use it and if it's an early version of node the second argument will be ignored...

packages/config/src/configDefaults.ts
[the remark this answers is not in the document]
cds-amal: Extending to include `ts` is fine. I missed (the obvious) that it was for matching migration extension.

packages/require/lib/index.ts
```diff
@@ -0,0 +1,130 @@
...
+export function file(options: RequireOptions) {
+  const scriptPath = path.resolve(options.file);
+
+  expectOptions(options, ["file"]);
+
+  const conf = TruffleConfig.default().with(options);
+
+  const scriptModule = new Module(scriptPath);
+
+  // Provide all the globals listed here: https://nodejs.org/api/globals.html
```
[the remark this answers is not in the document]
benjamincburns: Technically we provide all of them as we pass the `global` object itself. Though I'm not a fan of that, and I've removed it in my WIP code that I'm about to push.

I'll revise the comment to make it sound less comprehensive.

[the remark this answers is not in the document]
benjamincburns: Really I'm just trying my darndest to avoid this becoming a `vm2` rewrite.

packages/require/lib/compiler.ts
benjamincburns: resolving w/o modification, but please comment if you feel that `mts` should be in the list.

packages/require/lib/index.ts
benjamincburns: I changed up the flow of this and now we only `registerTsNode` one time.

packages/require/lib/compiler.ts
benjamincburns: In order to make this work I need to add error chaining to `@truffle/error`. I think this might be best handled in a separate PR. Should be a quick/easy thing to do, however.

packages/require/lib/types.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  /**
+   * A reference to a network defined in the user's `@truffle/config`.
+   */
+  network: string | number;
+
+  /**
+   * The id of the network as reported by the `net_version` JSON RPC method.
+   */
+  network_id: string | number;
```
davidmurdoch, line 52: The JSON-RPC type is an integer returned as a string; I don't think `number` belongs here.

Maybe something in truffle converts it to a number? I don't believe it should be normalized to `number`, as the value may be greater than `MAX_SAFE_INTEGER`.

davidmurdoch (changes requested):

packages/require/lib/types.ts
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * The body of a valid JSON RPC response
+ */
+export interface JsonRpcResponse {
+  jsonrpc: string;
+  id: number;
+  result?: any;
+  error?: string;
```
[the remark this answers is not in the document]
benjamincburns: I'd agree, however this needs to be permissive to providers from Web3, and web3 doesn't have this restriction.

packages/require/lib/types.ts
```diff
@@ -0,0 +1,112 @@
...
+  params: any[];
+  id: number;
+}
+
+/**
+ * The body of a valid JSON RPC response
+ */
+export interface JsonRpcResponse {
+  jsonrpc: string;
+  id: number;
```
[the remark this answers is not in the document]
benjamincburns: I think that while this is a more permissive type restriction for this field, it would actually increase the degree of restriction over the objects that can be assigned to this interface. As I mentioned in my other comment, these types need to be assignable from web3 providers, so I don't think I can further restrict them beyond what's already here.

That said, I do think it would be beneficial for Truffle to formally define a provider type (and subsequent subtypes, such as this one) along with a set of wrapper/transformer/converter/etc types/utilities/proxies/whatevers that allow Truffle to accept providers from various libraries, while still presenting them with a consistent abstraction internally. I just don't think that doing that is in scope for this issue.

packages/truffle/test/sources/migrations/typescript/truffle-config.js
```diff
@@ -0,0 +1,13 @@
+module.exports = {
```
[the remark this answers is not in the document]
benjamincburns: no

packages/truffle/test/scenarios/migrations/typescript.js
```diff
@@ -0,0 +1,69 @@
+const MemoryLogger = require("../MemoryLogger");
```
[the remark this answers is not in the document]
benjamincburns: No. The changes made in this PR don't change how our internal scenario tests are executed, and additional work would need to be done (outside of the scope of this feature) to change that.

packages/require/lib/typescript.ts
```diff
@@ -0,0 +1,171 @@
...
+export function compile(
+  conf: TruffleConfig,
+  sourceFilePath: string,
+  context: Object
+): string {
+  const source = readFileSync(sourceFilePath, { encoding: "utf-8" });
+
+  // we only compile TS files, so just return the source unless we are dealing
+  // with one of those
+  if (!_tsExtensions.includes(path.extname(sourceFilePath))) {
```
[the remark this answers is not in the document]
benjamincburns: Good idea re: regex. I'll change that.

Re: `mts` files, I don't believe that they are a concern, as I don't think that support for EJS targets is considered to be in scope for this change. Please do let me know if you think it should be, though!

packages/require/lib/typescript.ts
```diff
@@ -0,0 +1,171 @@
+import path from "path";
+
+import type { CreateOptions, Service } from "ts-node";
+import TruffleConfig from "@truffle/config";
+import { readFileSync } from "fs";
```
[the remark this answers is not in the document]
benjamincburns: > Maybe we should make these reads (and the whole compilation pipeline) asynchronous?

I think the effort involved in doing that would be far far greater than the value gained _in this specific context_.

That said, I'd be _very_ strongly in support of efforts to make the entire CLI execution flow async, but I don't think it's within the scope of the problem that I'm currently trying to solve.

packages/require/lib/typescript.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    try {
+      // Ensure that ts-node is installed by attempting to resolve its module ID.
+      // We use `originalRequire` here to ensure that we use ts-node from the
+      // user's environment rather than resolving our own (should we
+      // accidentally bundle ts-node)
+      originalRequire.resolve("ts-node");
+    } catch (err) {
+      // TODO: add error chaining support
+      throw new TsNodeDependencyError(path.extname(sourceFilePath));
```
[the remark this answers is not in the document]
benjamincburns: Have a look at the constructor of `TsNodeDependencyError`. It's a specialized error with a normalized message (the message string is constructed from the extension, but it is a full sentence that informs the user what happened).

packages/truffle/test/scenarios/migrations/typescript.js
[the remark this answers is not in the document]
benjamincburns: Er, let me revise that to say that yes they _should_ be, but this PR doesn't enable that to happen :-p

packages/require/lib/typescript.ts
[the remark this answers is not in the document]
benjamincburns: I just did a bit more research regarding `.mts` files, and I think this should also be regarded as a separate task.

Reviewing the node docs on [differences between ES modules and CommonJS](https://nodejs.org/api/esm.html#differences-between-es-modules-and-commonjs), it would seem that in order to properly support ES modules we'd need to detect whether we're loading such, and in that case we'd need to provide a slightly different global context (e.g. `import` should be present instead of `require`, `__dirname` and `__filename` should not be present, `import.meta.url` should be populated, etc).

packages/require/lib/types.ts
benjamincburns: The source of this is usually the user's `truffle-config.js` file. [Examples in our documentation](https://trufflesuite.com/docs/truffle/reference/configuration/#networks) specify `network_id` as a number in `truffle-config.js`, and I suspect there is quite a lot of code in the wild that does this.

packages/require/lib/types.ts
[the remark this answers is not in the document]
davidmurdoch: I'm not following your reasoning.

JSON-RPC 2.0 spec: https://www.jsonrpc.org/specification#request_object

Here is the type in web3: https://github.com/ChainSafe/web3.js/blob/1.x/packages/web3-core-helpers/types/index.d.ts#L225

packages/truffle/test/sources/migrations/typescript/truffle-config.js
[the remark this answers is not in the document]
davidmurdoch: :+1:, but sad day

packages/require/lib/typescript.ts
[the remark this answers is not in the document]
davidmurdoch: Ok. Can you make an issue for handling ejs/mts and then resolve this comment thread?

packages/require/lib/index.ts
```diff
@@ -0,0 +1,128 @@
...
+    provider: options.provider,
+    networkType: options.networks[options.network].type
+  });
+  const web3 = new Web3Shim({
+    provider: options.provider,
+    networkType: options.networks[options.network].type
+  });
+
+  try {
+    const fn = this.file({
```
davidmurdoch, line 118: What is `this` here? The global scope?

packages/require/lib/types.ts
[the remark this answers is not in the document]
benjamincburns: Yeah, you're totally right, and I definitely had my wires crossed yesterday when writing that.

But also I swear I copied these types directly out of the web3 code for the version that's currently in Truffle's `yarn.lock`. I wonder if they modified it since then?

Regardless as long as we're not making it so that providers from earlier versions are not assignable to variables/args that are typed with our provider type, I'm happy.

packages/require/lib/types.ts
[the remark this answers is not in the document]
benjamincburns: Thanks again for the quick references. I really wish the web3 codebase would finish converting over to TypeScript so these types were all generated from source. The amount of change from version-to-version is a bit annoying.

packages/require/lib/typescript.ts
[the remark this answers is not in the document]
benjamincburns: Oh I see. You're absolutely correct. I'll improve that.

packages/require/lib/typescript.ts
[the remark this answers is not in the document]
benjamincburns: Will do.

packages/require/lib/typescript.ts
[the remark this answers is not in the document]
benjamincburns: The commit that you linked changes the return type of the `file` function from `any` to `Promise<any>`, which is a breaking change. Resolving that breaking change is where the bulk of the effort would be on this, as it will force the caller to create an asynchronous context.

packages/require/lib/index.ts
benjamincburns: Good catch! I really don't know why it's like that. The only reason why this function is showing up as changed is because I decided to take this as an opportunity to port this entire module over to TS, and the only changes that I made to it were to support that change.

I have a sneaking suspicion that it's a very ugly ugly hack that someone put into place so that they could bind `this` to cause `exec` to call some local implementation of `file`, but I'm not certain. I'll dig into it and remove the `this` if I can.

packages/require/lib/typescript.ts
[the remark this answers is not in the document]
davidmurdoch: Ah, I see! Thanks for the explanation. I hadn't realized this existed before. I should have noticed the original implementation was deleted and rewritten in a new location (`git mv` would have helped here, but I should have caught it regardless).

You reasoning definitely makes sense now, since this exported file already (mis)behaves this way. Thanks for clarifying!

Note: if it is as easy as fixing up any uses of `Require.file` (doubted it's this easy) it might not be too complex after all as those contexts seem to be async already.

I'm marking this as Resolved. Thanks!

packages/require/lib/index.ts
davidmurdoch: Weird. I'm fine with it as is. I was just wondering if you had already dug into it.

I'll let you mark this one as Resolved in case you want to look into it (you might want to `git mv` the old implementation to this new location, just in case there is some good history there).

packages/require/lib/index.ts
benjamincburns: #5083

packages/require/lib/typescript.ts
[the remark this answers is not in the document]
benjamincburns: #4328

packages/require/lib/types.ts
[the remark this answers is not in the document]
benjamincburns: Figured out where the weirdness came from. We're depending on `@types/web3@1.0.20` in a number of our packages. Of course the types in that package are totally wrong, and converting to the correct types here causes compilation errors. 😞

I'll figure something out, though.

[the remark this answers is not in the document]
benjamincburns: OK - got that sorted. Wound up using the `provider` type from `web3-core/types` (but imported/exported it as `Provider` because who names their types with a lower case first letter?)

packages/require/lib/index.ts
benjamincburns: FWIW, I did do a `git mv` when I set this up, but it seems that enough of the file has changed that it didn't track its history. I'll see if I can insert a commit that does the move as a separate commit.

davidmurdoch, line 118: One more thing. Does the `this` needs to be typed here? Otherwise I think `file` will be of type `any`.

packages/interface-adapter/lib/adapter/web3/index.ts
```diff
@@ -80,9 +82,13 @@ export class Web3InterfaceAdapter implements InterfaceAdapter {
     const gas = new BN(receipt.gasUsed);
     const value = new BN(tx.value);
     const cost = gasPrice.mul(gas).add(value);
+    const timestamp =
+      typeof block.timestamp === "string"
+        ? parseInt(block.timestamp)
+        : block.timestamp;
```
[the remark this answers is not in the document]
benjamincburns: No idea, sorry. I just handled the case because the type check failed as it was a `string | BN` being assigned to a `BN`.

packages/require/lib/index.ts
benjamincburns: Sorry, I forgot to push the change that removed the `this`. Fixed now.

packages/interface-adapter/lib/adapter/web3/index.ts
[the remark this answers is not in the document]
benjamincburns: On further reflection, I think it's because the return type for `web3.eth.getBlock` is also used elsewhere in web3 as an input type, and they accept both `string` and `BN` inputs. Chances are that it's always returned as a BN, however.

benjamincburns:
I realized that the local tests for `@truffle/require` weren't executing, so I went ahead and fixed that. That in turn caused me to encounter and fix a bug where we'd fail to register `ts-node` properly in the case when `ts-node` has already been registered (e.g. by the mocha test runner being executed with `--register ts-node/register`).

packages/require/lib/index.ts
```diff
@@ -0,0 +1,143 @@
...
+import Module from "module";
+
+import originalRequire from "original-require";
+
+import { ExecOptions, RequireOptions } from "./types";
+import { compile } from "./typescript";
+
+import Debug from "debug";
+
+const debug = Debug("@truffle/require:index");
```
cds-amal, line 16: I don't think I've seen the debug string prefixed by `@ truffle`, it usually starts with the package name, like  `dashboard-message-bus-client:connection`
```suggestion
const debug = Debug("require:index");
```

packages/require/lib/typescript.ts
```diff
@@ -0,0 +1,137 @@
...
+import TruffleConfig from "@truffle/config";
+import { readFileSync } from "fs";
+
+import originalRequire from "original-require";
+
+import { TsNodeDependencyError } from "./types";
+
+import Debug from "debug";
+
+const debug = Debug("@truffle/require:typescript");
```
cds-amal, line 13:
```suggestion
const debug = Debug("require:typescript");
```

packages/require/lib/typescript.ts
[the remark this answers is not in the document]
cds-amal: > That said, I'd be very strongly in support of efforts to make the entire CLI execution flow async, but I don't think it's within the scope of the problem that I'm currently trying to solve

This should probably be an issue

packages/require/test/lib/typescript_module_with_local_file_import.ts
```diff
@@ -0,0 +1,6 @@
+// const module_with_global_import = require("./typescript_module_with_global_import");
```
cds-amal, line 1: This should go.

```suggestion
```

packages/require/lib/typescript.ts
```diff
@@ -0,0 +1,137 @@
...
+ * @param context the global context in which the compiled script will run
+ *
+ * @returns an instance of a TypeScript compiler service (TS Compiler API)
+ */
+function _getOrCreateCompilationService(
+  conf: TruffleConfig,
+  sourceFilePath: string
+): TSNode.Service {
+  if (!tsNode) {
+    // TODO: add error chaining support
```
cds-amal, line 72: What does error chaining support entail? Maybe this can become an issue instead of a TODO so it's not lost in the codebase, and we get eyes on it.

cds-amal (commented):
Looks good to me @benjamincburns. I'm excited to see how this evolves, and looking forward to us generating TS scaffolding. I left some very minor nits. The only thing that stands out is whether the `@truffle/require` TS conversion can be extracted to its own PR.

packages/require/lib/typescript.ts
benjamincburns: #5335 5335

cds-amal (approved):
We decided that extracting `@truffle/require` TS conversion as a separate PR wouldn't add value, so I approve this!

davidmurdoch (approved):
Really great work @benjamincburns! Huge congrats for getting this done!

PR #5224, opened 2022-06-23
Enhancement: Support v4 of signTypedData in hdwallet-provider

eggplantzzz:


cds-amal:
I don't have a good idea, but was thinking of integration tests. How can we test this new feature/behavior?

cds-amal (approved):
Good good, not bad!

PR #5227, opened 2022-06-23
Dependency update: Update solc version to 0.8.15

sukanyaparashar:
This PR updates the `solc` version to latest `0.8.15` in the `compile-solidity` and `contract-schema` packages.

### DIFF
--- a/packages/compile-solidity/package.json
+++ b/packages/compile-solidity/package.json
@@ -1,4 +1,4 @@
-{
+  {
   "name": "@truffle/compile-solidity",
   "description": "Compiler helper and artifact manager for Solidity files",
   "license": "MIT",
@@ -35,7 +35,7 @@
     "original-require": "^1.0.1",
     "require-from-string": "^2.0.2",
     "semver": "^7.3.4",
-    "solc": "0.8.12"
+    "solc": "0.8.15"
   },
   "devDependencies": {
     "@truffle/artifactor": "^4.0.158",
--- a/packages/compile-solidity/package.json
+++ b/packages/compile-solidity/package.json
@@ -1,4 +1,4 @@
-  {
+{
   "name": "@truffle/compile-solidity",
   "description": "Compiler helper and artifact manager for Solidity files",
   "license": "MIT",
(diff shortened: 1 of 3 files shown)


haltman-at:
Oh, thanks for getting this!  It would probably be a good idea to update it in `contract-schema` as well.

haltman-at:
I think it would make more sense as a single PR, but if you want to split it up I guess it's no harm.  They're already on two different versions, so it's not like this PR would be increasing the number of versions.

Side note: Do you know about lerna-update-wizard?  That can be very useful for doing this sort of thing.

sukanyaparashar:
> > I think it would make more sense as a single PR, but if you want to split it up I guess it's no harm. They're already on two different versions, so it's not like this PR would be increasing the number of versions.
> > Side note: Do you know about lerna-update-wizard? That can be very useful for doing this sort of thing.
> 
> I will have to look into lerna-update-wizard. Thanks for letting me know about it @haltman-at. For the time being, I will update the dependency in `contract-schema` package in this PR itself.

sukanyaparashar:
> Yeah, if you just go into the project root and run `npx lerna-update-wizard`, it'll bring up this command-line tool for updating dependencies. :) It can be annoying to use, so I don't necessarily recommend it for small updates, but if you've got a whole lot of packages to update, it can be useful. Also, it helpfully identifies for you when you have more than one version of the same package! Like we do here. :P

Thanks a lot for explaining this in details @haltman-at. :)

packages/compile-solidity/package.json
```diff
@@ -1,4 +1,4 @@
-{
+  {
```
haltman-at, line 1: Really minor note, but, you seem to have accidentally introduced an extra space here?

haltman-at (approved):
Seems good to me!  There's a minor formatting issue you may want to fix before merge but I'm just going to go ahead and hit approve. :)  Thanks again for getting this!

packages/compile-solidity/package.json
sukanyaparashar: Thanks for catching this @haltman-at. I have fixed this in my new commit.

haltman-at (approved):
(I already approved this, but re-approving for good measure. :P )

sukanyaparashar:
> (I already approved this, but re-approving for good measure. :P )

Yeah! Just to make sure. Thanks @haltman-at.

PR #5239, opened 2022-06-28
Internal improvement: Simplify contract-sources's default export

eggplantzzz:
**THIS IS A BREAKING CHANGE FOR contract-sources**

This removes the callback from the default export's signature.


haltman-at (dismissed):
Makes sense to me, I dunno why this needs to be async.  Wondering if @gnidan has an opinion.

gnidan:
probably worth keeping as `async` just in case (is there a discrepancy with the description?)

PR #5240, opened 2022-06-28
Internal improvement: Remove callback argument for Migrate.run

eggplantzzz:
**THIS IS A BREAKING CHANGE FOR migrate**

### DIFF
--- a/packages/migrate/index.js
+++ b/packages/migrate/index.js
@@ -66,39 +66,29 @@ const Migrate = {
     return migrations;
   },
 
-  run: async function (options, callback) {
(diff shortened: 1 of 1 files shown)


haltman-at:
@eggplantzzz The decoder tests also do the same things the debugger tests do so they're going to need this changed as well.

haltman-at (approved):
Let's get rid of this junk!

cds-amal:
What purpose did the call back provide before?

eggplantzzz:
It was just old architecture. No other purpose.

Issue #5050, opened 2022-04-24
Truffle Dashboard is a blank page


Bodiebuilds:
- [ ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

Running truffle dashboard opens a blank page

## Steps to Reproduce

nvm use 14
npm i -g truffle
npm truffle dashboard

tried rolling back versions but still same issue
npm uninstall -g truffle 
npm i -g truffle@v5.5.3

## Expected Behavior

open truffle dashboard webpage 

## Actual Results

node version is good and tried rolling back various versions of truffle (5.5.3 ; 5.5.4 and latest version) with no luck. Updated VS code as well. running dashboard command brings me to a blank white page. It has worked before with no problem but has been giving me this same result now.

## Environment

* Operating System: macOS
* Ethereum client:
* Truffle version (`truffle version`):5.5.11
* node version (`node --version`): 14.9.1
* npm version (`npm --version`): 6.14.16

lsqproduction:
@Bodiebuilds Thank you for opening the issue.  Please note `npm truffle dashboard` is not the correct command. It should just be `truffle dashboard`. May I also know what are the terminal outputs you were getting after running this command? 
Attaching a guide for [truffle dashboard](https://trufflesuite.com/blog/introducing-truffle-dashboard/)

dashboarduser:
I have the same problem after truffle reinstallation
npm uninstall -g truffle
npm install -g truffle
truffle dashboard — opens blank in Chrome (my default browser)
sudo truffle dashboard — opens in Safari all good, but can't use MetaMask

Environment:
macOS
truffle version
Truffle v5.5.12 (core: 5.5.12)
Ganache v^7.1.0
Solidity - 0.8.13 (solc-js)
Node v14.18.2
Web3.js v1.5.3

gnidan:
Thanks for reporting this! We've heard of a couple users running into this periodically, but we haven't been able to get the information we need to understand what's causing this behavior. If this is still occurring for anyone, would you mind please checking the browser console (like, right-click -> Inspect, or see [these instructions for Chrome](https://developer.chrome.com/docs/devtools/open/)) and share any errors you see?

Also, if the `truffle dashboard` terminal window has any odd-looking output, please share that as well! Any additional information would be enormously helpful, because otherwise we're a bit stumped.

Thanks all. 🙏

sukanyaparashar:
[...]

PR #5045, opened 2022-04-22
Add network switching capabilities to Truffle Dashboard

MicaiahReid:

Issue #5020, opened 2022-04-13
Dashboard shows odd network name when connecting with Ganache


benjamincburns:
## Issue

When I connect the dashboard to Ganache via Metamask, it tells me that I've connected to a network called "CENNZNET OLD."

## Steps to Reproduce

1. Start truffle dashboard
2. Configure MetaMask to point at localhost:8545
3. Observe network name shown in dashboard

## Expected Behavior

Network name should ideally be resolved in the following order (with each step falling back to the next):

1. User's `truffle-config.js`
2. Metamask's name for the network
3. https://chainid.network/chains.json (but pulled from the Truffle relay so we can unbreak old versions if that disappears/moves)

## Actual Results

![image](https://user-images.githubusercontent.com/803016/163287408-b1425d48-37a3-4b59-a770-4288f635c00d.png)

## Environment
 
* Truffle version (`truffle version`): v5.5.9

haltman-at:
Note that the reason this is happening at present is due to the prioritization of (3) over the others (together with the use of chain ID 1337).

cliffoo:
Some updates on this issue as I work through [pull#5241](https://github.com/trufflesuite/truffle/pull/5241).

In fear of potential supply chain attack (no hard feelings to [lists.eth](https://github.com/ethereum-lists/chains) or [defillama](https://github.com/DefiLlama/chainlist)), I'd like the new Dashboard to minimize all network requests that are not strictly related to processing Dashboard messages.

So chainid.network/chains.json appears to come from [ethereum-lists/chains](https://github.com/ethereum-lists/chains), and since:

- For a given chain, we only need its id and name for now.
- I'd like this id-to-name mapping to go into version control (like how we update sourcify fetcher).

I made a [little script](https://github.com/cliffoo/parse-chains) to parse the chain data in ethereum-lists/chains, which outputs a json small enough that I want to bundle it into the Dashboard frontend, and use like 
```js
import chainIDtoNameJson from "src/assets/chainIDtoName.json";
function getChainNameByID(id) {
  return chainIDtoName[id];
}
getChainNameByID(1337) // => "Ganache"
```

This, on top of my discovery that MetaMask doesn't seem to provide network names, I think the expected behavior can be modified as such:
Resolve chain name by:
1. `truffle-config.js`
2. `chainIDtoName.json` in `packages/dashboard/src/assets`

[...]

Issue #4982, opened 2022-04-05
Load state of "Processing..." display on page refresh


MicaiahReid:
This looks good for showing the initial feedback, but hard coding the initial state of the buttons probably isn't the best bet long-term.

Out of curiosity I ran the dashboard and went to migrate a contract. After clicking "Process" it shows "Processing...". But if you refresh the page while the transaction is still processing, it won't show that a transaction is processing - it will allow you to process or reject that same transaction again. Eventually we'd want to store this state somehow, or get it from truffle. @gnidan should I just make another issue for this?

_Originally posted by @MicaiahReid in https://github.com/trufflesuite/truffle/pull/4938#pullrequestreview-925036114_

PR #5241, opened 2022-06-28
Dashboard frontend

cliffoo:
I revamped dashboard frontend.

### Why?

This started when I looked into fleshing out features and adding them to dashboard ([decoded tx](https://github.com/trufflesuite/truffle/pull/4933), [contract interaction](https://github.com/trufflesuite/truffle/pull/4940),[ network + wallet management](https://github.com/trufflesuite/truffle/pull/5045), and more), and I realized various problems with the existing react code, both the code itself and also more meta (file structure, maintainability) things. So I started from scratch, picked up a component library ([mantine](https://github.com/mantinedev/mantine)) I've been eyeing for a long time, and rebuilt the dashboard frontend.

### Some problems and their solutions

<details>
<summary>
Performance
</summary>

- Problem

  There are examples of things that cause sub-optimal runtime (take a look at [this](https://github.com/trufflesuite/truffle/blob/v5.5.23/packages/dashboard/src/components/dashboardprovider/index.tsx), <code>ctrl + f</code> "requests" and see how many times it's looping.)

- Solution

  I've made things as efficient as I could, bearing readability in mind. This specific example above (looping requests) is [fixed with a hashmap](https://github.com/trufflesuite/truffle/blob/dash-woah/packages/dashboard/src/contexts/DashContext/state.ts#L40) so constant lookup time.

</details>

<details>
<summary>
State management
</summary>

- Problem

  State is _everywhere_ right now, it seeps through components at different levels and it's hard to keep track / probably causing extra renders.

- Solution

  App state is managed better in general. There is a <code>\<DashContext \/></code> close to the root that the entire app relies on (to read and mutate state).

</details>

<details>
<summary>
UX
</summary>

- Problem

  Right now:
  - Wallet is disconnected on refresh.
  - Without having wallet connected, you cannot connect to the dashboard network.
  - The "processing" state for requests is hard-coded. Sometimes on refresh a processed request re-appears.

- Solution

  - Wallet is only disconnected on explicit user request.
  - While wallet is waiting to be connected, you can connect to the dashboard network and even queue up RPC requests to be processed.
  - Once "Confirm" is clicked the RPC request leaves the frontend state, so it won't show again on refresh (though you can see it on Metamask).

</details>

<details>
<summary>
Wagmi
</summary>

- Problem

  Wagmi is used weird, it might be because wagmi has evolved and its usage has improved? Take a look at how [wagmi is currently provided to the app](https://github.com/trufflesuite/truffle/blob/v5.5.23/packages/dashboard/src/App.tsx). I think it's not very readable / might be doing the wrong thing.

- Solution

  [New implementation](https://github.com/trufflesuite/truffle/blob/dash-woah/packages/dashboard/src/components/wrappers/WagmiWrapper.tsx) is clean and easy to understand.

</details>

<details>
<summary>
Code organization
</summary>

  This is a big one for me, but it's rather hard to articulate, since there is no _right_ way to organize things in react. I guess an exercise you could do is start from the root (`src/App.tsx`), dig into how each component relate to each other until you've traversed through all of them; then do the same for this branch and see how they compare.

  Here's how I've organized things:

  - <code>App.tsx</code> <- Only context providers and routing go here, no real app logic
  - <code>contexts/</code>
    - <code>DashContext/</code> <- (Currently single) source for app's state
  - <code>components/</code>
    - <code>common/</code> <- Reusable components
    - <code>composed/</code> <- Distinct parts of the app, such as <code>\<Sidebar \/></code> and <code>\<Txs \/></code>, "composed" by nesting smaller components in larger ones, and reusing components from <code>components/common</code>
    - <code>wrappers</code> <- Context providers by 3rd party libs
  - <code>assets/</code> <- Static things like fonts and images
  - <code>hooks/</code> <- All custom hooks
  - <code>utils/</code> <- Convenience things

</details>

<details>
<summary>
Craco
</summary>

- Problem:
  
  So it looked like craco was running out of community support, though someone seems to have picked it up recently. 
  
  I don't see a convincing argument for us to use it over CRA. Craco doesn't support CRA 5 yet, we don't need to override any webpack configs, and we're currently [only using craco to get tailwind working](https://github.com/trufflesuite/truffle/blob/v5.5.23/packages/dashboard/craco.config.js) (tailwind is no longer needed, as mantine uses [emotion](https://github.com/emotion-js/emotion) for styling).

- Solution:

  CRA.

</details>

<details>
<summary>
Eslint & preflight checks
</summary>

- Problem:

  Currently [eslint and preflight checks are disabled](https://github.com/trufflesuite/truffle/blob/v5.5.23/packages/dashboard/package.json#L25) during build and dev processes. This means important checks like [`exhaustive-deps`](https://github.com/trufflesuite/truffle/blob/v5.5.23/packages/dashboard/package.json#L25) are suppressed, which means components are likely [not handling updates consistently](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies).

- Solution:

  Enable 'em. The build / start commands are now just `react-scripts build` and `react-scripts start`. Proper checks are in place and not generating warnings.

</details>

<details>
<summary>
Security
</summary>

  I'm probably being more cautious than necessary here, but since a big reason why you'd use dashboard in the first place is security (no need to hand off mnemonic, just let Metamask take care of it), as I mentioned [here](https://github.com/trufflesuite/truffle/issues/5020#issuecomment-1182894979) I want to minimize / elimiate all network requests that are not used for processing RPC requests.

  This means no more requests to the following:
  
  - Google fonts cdn (fonts bundled now)
  - chainid.network/chains.json (chain info bundled also)

</details>

### Other things
- Re-renders are minimized. Lifecycles and states are efficiently managed.
- `import` / `export` `type` wherever possible. This is good because `type` imports are [fully erased during runtime](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html).
- Dev deps that should be deps, are now deps.

### Fun things
- Dark mode.
- Smoother experience in general.

### Why this is good
The biggest win for me is that this sets up a good foundation for future improvements. There is a clearer path forward on integrating new features, reasons in the lists above.

### How to review
The best way is probably to ignore how dashboard frontend is _now_. I.e. When going through the changes, ignore all deletions (red) and look only at additions (green). There is almost no inherited code.

### Problems I've introduced
- [Fixed by upgrading to mantine 5] ~Mantine + react dependency clash. See [this](https://github.com/mantinedev/mantine/issues/1201).~
- [Fixed by upgrading to mantine 5] ~`flushSync` warning on clipboard copy in modal. See [this](https://github.com/mantinedev/mantine/issues/1823).~
- When lerna links dependencies
  - [Fixed by adding necessary babel plugins and core] ~Babel peer deps warning, but things work ok?~
  - React native peer deps warning, but no I'm not adding that.

### Issues that are conveniently resolved
- Blank page [issue#5050](https://github.com/trufflesuite/truffle/issues/5050) `=>` This shouldn't happen, unless there's no browser wallet (like Metamask). (Aside: I'll probably add an ErrorBoundary in the future to catch errors like no wallet found.)
- Bad chain names [issue#5020](https://github.com/trufflesuite/truffle/issues/5020) `=>` 1337 resolves to "Dev Network" now.
- Hard-coded processing state [issue#4982](https://github.com/trufflesuite/truffle/issues/4982) => Once tx / call is processed it is removed from dashboard state. You won't see a processed request on refresh.
- Chain names caching [issue#4949](https://github.com/trufflesuite/truffle/issues/4949) `=>` No need to cache, chain names are now bundled.
- Outdated react-scripts [issue#4874](https://github.com/trufflesuite/truffle/issues/4874) `=>` Updated to latest.

There may be more? I just searched `is:issue is:open dashboard` and filtered through the result.

### Future
This PR was originally more ambitious, but in favor of better git practices and PR reviewers' time, new features will ensue in the form of smaller, more digestible PRs after this one merges against develop.

### Progress
- [x] Bomb
- [x] ~Clear outdated `caniuse-lite` warning~ (already made it to develop)
- [x] Add and setup mantine
- [x] Add and setup react-router-dom
- [x] Layout and nav
- [x] Wagmi context
- [x] Basic network management
  - [x] (Dis)connect
  - [x] Display chain name + id
- [x] `<DashContext />`
  - [x] Setup
  - [x] Message bus client setup
  - [x] Notice system
  - [x] Wire up wagmi and message bus client
  - [x] Chain name
- [x] Message queue
- [x] Message modal
- [x] Confirm network (initially and on switch)
- [x] Make things look nice
- [x] Clean up commit messages
- [x] Rebase
- [x] Mantine 5 came out very recently, upgrade
- [x] [Fix ci](https://github.com/trufflesuite/truffle/pull/5359)


packages/dashboard/src/contexts/DashContext/state.ts
```diff
@@ -0,0 +1,122 @@
...
+    id: null,
+    name: null
+  },
+  notice: {
+    show: false,
+    type: "LOADING"
+  }
+};
+
+const reducer = (state: stateType, action: actionType): stateType => {
```
michaeljohnbennett, line 72: wouldn't it be better to use immer/zustand instead of this huge set of case statements/reducer code?

packages/dashboard/package.json
```diff
@@ -22,60 +22,56 @@
   ],
   "scripts": {
```
michaeljohnbennett, line 23: if you want style/lint enforcement, should you setup husky here with eslint configured correctly? But this is hard in a mono repo.

michaeljohnbennett (commented):
overall looks fine, seems a major rewrite.

Mantine is yet another framework to learn and I guess the main issue is how long it will live/be maintained for.

Not a huge fan of the emotion styling as it seems super verbose compared to tailwind where its more declarative but I understand this is tied into mantine.

State seems overly complex, reducers could be made a lot simpler by trying something like zustand/immer as a lightweight alternative.

cliffoo:
Yea I mean frontend things in general come and go, Mantine happens to have stellar community + docs.

Tailwind sometimes hurts my eyes with dozens of classnames, but to each their own.

Here's the state shape that currently handles the entire app. It'd be the same complexity with any other state management lib.

```ts
type stateType = {
  host: string;
  port: number;
  client: DashboardMessageBusClient | null;
  providerMessages: Map<
    number,
    ReceivedMessageLifecycle<DashboardProviderMessage>
  >;
  chainInfo: {
    id: number | null;
    name: string | null;
  };
  notice: {
    show: boolean;
    type: noticeContentType | null;
  };
};
```

This PR isn't about change of tooling as that'd be mundane, it's just a by-product of fixing the problems I outlined above.

packages/dashboard/src/contexts/DashContext/state.ts
cliffoo: It's a switch statement with 4 cases. Reducer suffices.

packages/dashboard/package.json
cliffoo: Husky hook (to lint on precommit) _does_ run.

packages/dashboard/src/components/composed/Contracts.tsx
```diff
@@ -0,0 +1,5 @@
+function Contracts(): JSX.Element {
+  return <div>interact with contracts</div>;
```
MicaiahReid, line 2: If we're keeping this path maybe we can center it and change it to `Contract interactions coming soon` or something

MicaiahReid: Though I know this path isn't advertised, so maybe we don't care

MicaiahReid (commented):
Very cool stuff! Nothing particularly important or blocking from me, but I don't think I looked in depth enough to warrant a full approval. Great work!

packages/dashboard/src/components/composed/Txs/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+        align="center"
+        justify="center"
+        pb="4%"
+        className={classes.maxSize}
+      >
+        <LogoImg size={200} variant="dimmed" />
+        <Text size="sm" color="dimmed">
+          Nothing to see for now... Get started with&nbsp;
+          <Anchor
+            href="https://trufflesuite.com/docs/truffle/getting-started/using-the-truffle-dashboard"
```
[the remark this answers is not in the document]
cliffoo: I can definitely make an update once this makes it to develop / gets released. The file to update is [here](https://github.com/trufflesuite/trufflesuite.com/blob/main/src/docs/truffle/getting-started/using-the-truffle-dashboard.md).

packages/dashboard/src/App.tsx
```diff
@@ -1,45 +1,43 @@
...
+          <MantineGlobal />
+          <WagmiWrapper>
+            <DashProvider>
+              <BrowserRouter>
+                {/*  Everything maps to /txs except for /contracts  */}
+                <Routes>
+                  <Route path="/" element={<Layout />}>
+                    <Route index element={<Navigate to="/txs" replace />} />
+                    <Route path="txs" element={<Txs />} />
+                    <Route path="contracts" element={<Contracts />} />
```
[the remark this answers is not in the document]
cliffoo: Good point! I've removed `contracts` route and made `colors` route available during development only.

packages/dashboard/src/components/composed/Contracts.tsx
cliffoo: Got rid of the route, so this isn't reachable anymore 🙈

packages/dashboard/src/contexts/DashContext/state.ts
michaeljohnbennett: your not counting the cyclomatic complexity of your last case and the several if statements embedded within that are are really a full other set of logic.

cliffoo: Could you explain how a different state management lib would make it easier / cleaner to determine dashboard message type? 
```ts
if (isDashboardProviderMessage(message)) {
  // ...
} else if (isInvalidateMessage(message)) {
  // ...
} else if (isLogMessage(message)) {
  // ...
} else if (isDebugMessage(message)) {
  // ...
}
```

packages/dashboard/src/contexts/DashContext/state.ts
```diff
@@ -0,0 +1,122 @@
...
+} from "@truffle/dashboard-message-bus-common";
+import {
+  messageNeedsInteraction,
+  messageIsUnsupported,
+  rejectMessage,
+  confirmMessage
+} from "src/utils/dash";
+import type noticeContentType from "src/components/composed/Notice/noticeContentType";
+
+type actionType =
```
gnidan, line 24: Types should be capitalized, since they're akin to classes, not instances

gnidan: I'd also say that you could do something like this:

```typescript
interface Action {
  type: string;
}
```

and just do:

```typescript
interface SetClientAction extends Action {
  type: "set-client";
  data: DashboardMessageBusClient;
}
```

(`extends Action` is maybe unnecessary here, but it might be helpful cause then it will typecheck)

packages/dashboard/src/contexts/DashContext/state.ts
```diff
@@ -0,0 +1,122 @@
...
+  | { type: "set-notice"; data: Partial<stateType["notice"]> }
+  | {
+      type: "handle-message";
+      data: {
+        lifecycle: ReceivedMessageLifecycle<Message>;
+        provider: providers.JsonRpcProvider;
+      };
+    };
+
+type stateType = {
```
gnidan, line 36: (same here with the capitalization)

Also perhaps worth noting... it's better to use `interface` when you can, since it's a simpler concept... sometimes you can't, but here I'm sure `interface` is fine

packages/dashboard/src/contexts/DashContext/state.ts
gnidan: Using a state management library definitely seems like overkill here... perhaps in the future, we'll change our mind, but let's not add dependencies in this PR.

Also, if we do want to move in the direction of bringing in a first-class state management solution, then we should probably stick with redux because we use it elsewhere already.

Note that if we do use a state management solution, we will likely have our work cut out for us with these action types... i.e., Redux pretty strongly demands that all actions are serializable plain objects, and `DashboardMessageBusClient` is extremely not serializable.

gnidan (commented):
Just a couple comments about type styles

packages/dashboard/src/App.tsx
```diff
@@ -1,45 +1,41 @@
...
+            <DashProvider>
+              <BrowserRouter>
+                <Routes>
+                  <Route path="/" element={<Layout />}>
+                    <Route index element={<Navigate to="/txs" replace />} />
+                    <Route path="txs" element={<Txs />} />
+                  </Route>
+                  {process.env.NODE_ENV === "development" && (
+                    <Route path="colors" element={<Palette />} />
+                  )}
```
gnidan, line 30: This should probably appear in the nav if we're keeping this here (appear in nav only in development, of course)

packages/dashboard/src/components/composed/Txs/Tx/index.tsx
```diff
@@ -0,0 +1,60 @@
...
+    opacity: 0.2,
+    transition: "opacity 0.2s"
+  }
+}));
+
+type TxProps = {
+  lifecycle: ReceivedMessageLifecycle<DashboardProviderMessage>;
+};
+
+function Tx({ lifecycle }: TxProps): JSX.Element {
```
gnidan, line 29: Maybe better not to abbreviate this? Existing style is that we don't abbreviate anything, ever.

packages/dashboard/src/contexts/DashContext/DashProvider.tsx
```diff
@@ -0,0 +1,108 @@
...
+        });
+      }
+
+      dispatch({ type: "set-chain-info", data });
+    };
+
+    updateChangeInfo();
+  }, [chain]);
+
+  const ops = {
```
gnidan, line 90: it might be more readable to spell this out as `operations` without abbreviating

packages/dashboard/src/contexts/DashContext/state.ts
cliffoo: Yup I've [capitalized types and interfaces](https://github.com/trufflesuite/truffle/commit/486efcb9cae84609c6966dc64a8a9963384a17a3)! I did end up having [a `BaseAction` that others extend off of](https://github.com/trufflesuite/truffle/blob/dash-woah/packages/dashboard/src/contexts/DashContext/types/Action.ts) to imply that `AnAction` and `AnotherAction` are "similar", and not unrelated.

packages/dashboard/src/App.tsx
cliffoo: [Nav button added](https://github.com/trufflesuite/truffle/commit/191c6a8572901b9cb3fb8626ba6c80cb7bfac637).

packages/dashboard/src/components/composed/Txs/Tx/index.tsx
cliffoo: I've [removed all abbreviations (except for "RPC")](https://github.com/trufflesuite/truffle/commit/9ad1ad3500d54161ed10028d9bf252b0cee50d23), some file / dir names were changed.

packages/dashboard/src/components/common/LogoImg.tsx
```diff
@@ -0,0 +1,88 @@
...
+  if (typeof size !== "number") {
+    size = sizes[size || "sm"];
+  }
+
+  let colors = colorVariants.normal;
+  if (variant === "dimmed") {
+    colors = colorVariants.dimmed[colorScheme];
+  } else if (colorVariants.hasOwnProperty(variant as string)) {
+    colors = colorVariants[variant!];
+  }
```
gnidan, line 53: Looks like this block is causing some trouble for me... maybe re-write this so it's just `const colors = ...`?

gnidan:
Is there more we could do here to signify to the user that this needs their action? Previously, the confirm button was self-evident, but now the user has to click to see confirm/reject.

Screen Shot 2022-08-02 at 4 46 32 PM

packages/dashboard/src/components/common/LogoImg.tsx
gnidan: ah the problem is what happens if `colorScheme` isn't defined... since this was my first load, I had nothing in local storage, so I didn't have any "saved color scheme"... thus `colorVariants.dimmed[colorScheme]` was coming back undefined.

also, how's this work with respect to system color scheme? I'd expect the default to match my system light/dark mode.

packages/dashboard/src/components/composed/Sidebar/Middle/index.tsx
```diff
@@ -0,0 +1,43 @@
...
+      sx={{ opacity: 0.4 }}
+    >
+      coming soon
+    </Badge>
+  );
+
+  return (
+    <Navbar.Section grow py="sm">
+      <NavButton
+        label="Transactions / Calls"
```
gnidan, line 21: Calls won't actually prompt the user, since they don't need a signature... maybe better to call this "Signature Requests" or somesuch

gnidan: It would also probably be a nice affordance to include a badge with the number of outstanding requests

packages/dashboard/src/components/common/LogoImg.tsx
cliffoo: This is a bug indeed, the first load would fail because of this. I've fixed it [here](https://github.com/trufflesuite/truffle/pull/5241/commits/3fecc13ddde63bcac230cb6ffd3b4bb371f99a21).

It respects system color scheme until you put something in local storage (on toggle button click). The priority is:
1. Local storage, if defined
2. System, if detected (using [Window.matchMedia api](https://developer.mozilla.org/en-US/docs/Web/API/Window/matchMedia))
3. Fallback to light

packages/dashboard/src/components/composed/Sidebar/Middle/index.tsx
cliffoo: I wasn't sure on how to label this, because a given request may not be a transaction request (e.g. `eth_signTransaction`, `personal_sign`, etc.), but I erroneously put "calls". "Signature Requests" is much better! Commit [here](https://github.com/trufflesuite/truffle/pull/5241/commits/9a8fb1c2d72ebfd00eeef6edb68c1a978d6a439d).

I've also added a dot to indicate the number of outstanding requests (capping at 99, >99 shows "99+"). Commit [here](https://github.com/trufflesuite/truffle/pull/5241/commits/7996f8a3195ca6dac3a9f109e313c7a06a6bf892).

cliffoo:
@gnidan Re comment about a clear request for user to do something; +33% clicks

Does the arrow that lights up teal on hover count lol? I imagined users going like "Oh wow look something just popped up, I wonder if that's the transaction I just sent to the dashboard network let me go click that and inspect the details".

Anyway I tucked each request into a modal because without modals, a few requests can flood your screen pretty quickly. Also the modal is always centered so that's nice. Things are looking homogeneous at the moment (you cannot differentiate requests of the same type), but once decoding comes in this won't be a problem.

I can definitely move / duplicate the buttons from the modal so users can acknowledge a request without an extra click, or should I keep it as is to coerce users into affording an extra click + looking at the request details?

\* As I'm writing this, I realize there may be a solution to this. What say you? Is this a better way?
Screen Shot 2022-08-04 at 06 44 51

gnidan:
> * As I'm writing this, I realize there may be a solution to this. What say you? Is this a better way?

This is the better way I, too, was beginning to realize :)

Let's go with that. Users have already complained about the number of clicks they have to make when running, e.g., `truffle migrate` for a complex application. That 33% increase could be like 20-30 extra clicks in some cases!

packages/dashboard/src/components/composed/Sidebar/Middle/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+      sx={{ opacity: 0.4 }}
+    >
+      coming soon
+    </Badge>
+  );
+
+  return (
+    <Navbar.Section grow py="sm">
+      <NavButton
+        label="Signature Requests"
```
gnidan, line 26:
```suggestion
        label="Signature requests"
```

Maybe it's personal preference but I always felt like Section Headers Using All Caps was just a bit overdoing it ;)

packages/dashboard/src/components/composed/RPCs/RPC/Overview.tsx
```diff
@@ -0,0 +1,140 @@
...
+      position="apart"
+      spacing={50}
+      pl={42}
+      pr={35}
+      py="lg"
+      className={`${classes.container} ${
+        active ? classes.activeContainer : ""
+      }`}
+      tabIndex={0}
+    >
```
gnidan, line 110: I really dig the signifiers you have for expanding / rejecting confirming! The mouse-over effect suggesting "click to expand" looks beautiful and clearly guides usage, and the accept/reject check/x are nice touches :)

I noticed however that this disappears when expanded - confirm/reject don't show up anymore, and it'd be great to have the reverse chevrons-up to show I can collapse the view.

<details>
<summary>See what I mean</summary>

the-signifiers-duke-the-signifiers

</details>

packages/dashboard/src/components/composed/Sidebar/Top/ColorSchemeToggle.tsx
```diff
@@ -0,0 +1,41 @@
...
+    }
+  };
+});
+
+function ColorSchemeToggle(): JSX.Element {
+  const { classes } = useStyles();
+  const { colorScheme, toggleColorScheme } = useMantineColorScheme();
+  const featherIconProps = { size: 16 };
+
+  const handleClick = () => void toggleColorScheme();
```
gnidan, line 28: How do I switch back to system theme?

packages/dashboard/src/components/composed/Notice/content/ConfirmChain.tsx
```diff
@@ -0,0 +1,83 @@
...
+    operations: { toggleNotice }
+  } = useDash()!;
+  const { colorScheme } = useMantineColorScheme();
+  const { classes } = useStyles();
+
+  const title = (
+    <Title order={4}>
+      <Info size={18} />
+      <Space w={6} />
+      Confirm Network
```
gnidan, line 41:
```suggestion
      Confirm network
```

gnidan (changes requested):
I think these are the rest of my comments. This looks so good!

benjamincburns:
> This looks so good!

I apologize as I feel I haven't been adequately able to review this as I lack the expertise in frontend development. But that said, I really appreciate the design changes that you've made here @cliffoo - I think this is heading in a really exciting direction!

packages/dashboard/src/components/composed/Sidebar/Top/ColorSchemeToggle.tsx
gnidan: What if we just remove it from local storage and default to system theme upon page load? It's okay if there's no way to unset something if a refresh serves as a painless way to do the same thing!

cliffoo: A ternary toggle would probably be confusing. Users might think: I'm setting it to light... dark... and woah! Did I just unset it? 🤔

Anyway I think it might be better to keep it in local storage for now, even though there's no UI to reset it to system. If someone wants the dashboard to have the opposite color scheme than their system color scheme, they'd have to toggle it every time the page loads!

Would it be better, in the future when there's a settings panel / modal of some kind, to have a button there to reset it to system?

packages/dashboard/src/components/composed/RPCs/RPC/Overview.tsx
cliffoo: Yea I omitted it because I thought it'd make the expanded view cleaner, but I think this makes sense! I've [made the change](https://github.com/trufflesuite/truffle/pull/5241/commits/2d94c4775b82d70de0bd00bdbccb9423d0a370a8), still clean :)

packages/dashboard/src/components/composed/Sidebar/Top/ColorSchemeToggle.tsx
gnidan: We'll figure this out later :)

packages/dashboard/src/components/composed/RPCs/RPC/Overview.tsx
gnidan: Amazing

gnidan (approved):
This looks good to go once you sort out the conflicts. Merge when green :)

Nice job @cliffoo

PR #5087, opened 2022-05-05
Add --url option in truffle console command

sukanyaparashar:
### ISSUE
`truffle console` should allow --url option like `truffle debug` does.

### SOLUTION
This PR adds an --url option to the `truffle console` command that connects to a specified url and can be used outside of Truffle project (with web3 object and accounts).

PR #5245, opened 2022-06-29
Refactor console-child.js to make the network settings more robust

sukanyaparashar:
### ISSUE
PR #5087 broke `truffle console --network` because it [hardcoded the `network name` to be `develop`
](https://github.com/trufflesuite/truffle/blob/b7896bb42b932eaf506369315f43405e5762535f/packages/core/lib/console-child.js#L45)
This was because `console-child.js` had logic exclusively used by the `truffle develop` command. When `truffle console --network <network_name>` is used, it adds the `develop` network into `config` which is not required. This behaviour doesn't make sense after PR #5087 because `console-child.js` has to satisfy more code paths such as:
- `truffle console --network`
- `truffle console --url` 
- `truffle develop`

### SOLUTION
This PR  makes `console-child.js` more robust to handle the above cases. And also adds an integration test for the `truffle console --network` option.

**NOTE:** It needs to be merged before the next release so that it doesn't break the use of `--network` options with the `truffle console` command.

### DIFF
--- a/packages/core/lib/console-child.js
+++ b/packages/core/lib/console-child.js
@@ -23,33 +23,43 @@ global.crypto = {
 const { network, config, url } = yargs(input[0]).argv;
 const detectedConfig = Config.detect({ network, config });
 
-if (url) {
-  detectedConfig.networks[network] = {
-    network_id: "*",
-    provider: function () {
-      return new Web3.providers.HttpProvider(url, { keepAlive: false });
-    }
-  };
-} else {
-  const customConfig = detectedConfig.networks.develop || {};
-
-  //need host and port for provider url
-  const ganacheOptions = {
+function getConfiguredNetworkUrl(customConfig) {
+  const configuredNetworkOptions = {
     host: customConfig.host || managedGanacheDefaultHost,
     port: customConfig.port || managedGanacheDefaultPort
   };
-  const ganacheUrl = `http://${ganacheOptions.host}:${ganacheOptions.port}/`;
+  return `http://${configuredNetworkOptions.host}:${configuredNetworkOptions.port}/`;
+}
 
-  //set up the develop network to use, including setting up provider
-  detectedConfig.networks.develop = {
+let configuredNetwork;
+
+if (url) {
+  // implies not "develop"
+  // Use "url" to configure network
+  configuredNetwork = {
+    network_id: "*",
+    url
+  };
+} else {
+  // Otherwise derive network settings
+  const customConfig = detectedConfig.networks[network] || {};
+
+  configuredNetwork = {
     host: customConfig.host || managedGanacheDefaultHost,
     port: customConfig.port || managedGanacheDefaultPort,
     network_id: customConfig.network_id || managedGanacheDefaultNetworkId,
-    provider: function () {
-      return new Web3.providers.HttpProvider(ganacheUrl, { keepAlive: false });
-    }
+    url: getConfiguredNetworkUrl(customConfig)
   };
 }
+
+detectedConfig.networks[network] = {
+  ...configuredNetwork,
+  provider: function () {
+    return new Web3.providers.HttpProvider(configuredNetwork.url, {
+      keepAlive: false
+    });
+  }
+};
 
 const { getCommand, prepareOptions, runCommand } = require("./command-utils");
 const command = getCommand({ inputStrings, options: {}, noAliases: false });
--- a/packages/core/lib/console-child.js
+++ b/packages/core/lib/console-child.js
@@ -34,8 +34,7 @@ function getConfiguredNetworkUrl(customConfig) {
 let configuredNetwork;
 
 if (url) {
-  // implies not "develop"
-  // Use "url" to configure network
+  // Use "url" to configure network (implies not "develop")
   configuredNetwork = {
     network_id: "*",
     url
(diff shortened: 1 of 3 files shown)


packages/truffle/test/scenarios/commands/console.js
```diff
@@ -1,23 +1,32 @@
...
+      reporter: new Reporter(logger)
+    };
   });
 
-  describe("when run with --url option", () => {
-    it("displays the hostname in the prompt", async () => {
+  describe("when run with options", () => {
+    it("displays the url hostname in the prompt", async () => {
       const url = "http://localhost:8545";
       const parsedUrl = new URL(url);
```
cds-amal, line 31: These tests should cover the cases
  - url
    - runs in folder with a truffle-config
    - runs in folder without a truffle-config
  - network
    - runs in folder with a truffle-config

cds-amal (changes requested):
Looks good @sukanyaparashar. A few tests should be added for better coverage.

packages/core/lib/console-child.js
```diff
@@ -23,33 +23,32 @@ global.crypto = {
...
   };
-  return `http://${ganacheOptions.host}:${ganacheOptions.port}/`;
+  return `http://${configuredNetworkOptions.host}:${configuredNetworkOptions.port}/`;
 }
 
 let configuredNetwork;
 
 if (url) {
+  // implies not "develop"
   // Use "url" to configure network
```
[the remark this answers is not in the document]
sukanyaparashar: That makes sense! @cds-amal . Thanks :)

packages/truffle/test/scenarios/commands/console.js
sukanyaparashar: Sure, thanks for the suggestions @amal. I will add the tests to cover the mentioned cases.

packages/truffle/test/scenarios/commands/console.js
```diff
@@ -1,41 +1,108 @@
...
+          output.includes(expectedValue),
+          `Expected "${expectedValue}" in output`
+        );
+      }).timeout(90000);
+    });
+
+    describe("without a config", () => {
+      let sandlot;
+      before(() => {
+        sandlot = tmp.dirSync({ unsafeCleanup: true });
```
eggplantzzz, line 76: what is sandlot? :) did you mean sandbox?

sukanyaparashar: :) Yeah!! Just a different variable other than `sandbox` to use if the case is without a config.

packages/truffle/test/scenarios/commands/console.js
```diff
@@ -1,41 +1,108 @@
...
+        config = { working_directory: sandlot.name };
+        config.logger = logger;
+        config.mocha = {
```
eggplantzzz, line 79: I'm not sure you need to configure `config.mocha`. Do the tests work with setting this value? I ask because I recently removed a bunch of these as the whole `Reporter` thing was unnecessary.

packages/truffle/test/scenarios/commands/console.js
```diff
@@ -1,41 +1,108 @@
...
+          reporter: new Reporter(logger)
+        };
+      });
+      after(() => {
+        sandlot.removeCallback();
```
eggplantzzz, line 84: You probably copied this from another test as well but you probably don't need this as `tmp` cleans up after itself. Check out the "Synchronous file creation" section of [this](https://www.npmjs.com/package/tmp).

packages/truffle/test/scenarios/commands/console.js
sukanyaparashar: Yeah! I just checked that it works without `config.mocha`. I will remove those.

packages/truffle/test/scenarios/commands/console.js
sukanyaparashar: Thanks for the link @eggplantzzz . It is clearly written that it will clean itself after.

packages/truffle/test/scenarios/commands/console.js
```diff
@@ -1,41 +1,108 @@
...
-    config = Config.default();
+  project = path.join(__dirname, "../../sources/console");
+
+  before(async function () {
+    this.timeout(10000);
+    config = await sandbox.create(project);
+    config.network = "development";
     config.logger = logger;
+    config.mocha = {
+      reporter: new Reporter(logger)
```
eggplantzzz, line 25: You might not need to set this property for all these tests. In other tests I've found this doesn't do anything and I was able to get rid of 
```
config.mocha = {
  reporter: new Reporter(logger)
}
```

sukanyaparashar: Yes, sure. I also checked removing those. And it seems that it doesn't do anything. I will get rid of these.

cds-amal (approved):
Looks good to me, though @eggplantzzz may have more comments.

packages/core/lib/console-child.js
```diff
@@ -23,33 +23,31 @@ global.crypto = {
...
+  // Use "url" to configure network (implies not "develop")
   configuredNetwork = {
-    network,
     network_id: "*",
     url
   };
 } else {
   // Otherwise derive network settings
-  const customConfig = detectedConfig.networks.develop || {};
+  const customConfig = detectedConfig.networks[network] || {};
```
eggplantzzz, line 44: Should we use `develop` as a fallback? Because the settings in `config.networks.develop` are meant for the develop console. Or do we pass the name "develop" as a network in the case when it is the development console? Like maybe it should be
```
detectedConfig.networks[network] || detectedConfig.develop || {};
```
?

eggplantzzz (approved):
I like it

Issue #5235, opened 2022-06-27
truffle workflow-compile lost dashboard-message-bus-client dependency


moshmage:
## Issue

The package lost dependencies but still relies on them

## Steps to Reproduce

Visit this public github action log, check build stage.
[Please provide the shortest amount of steps to reproduce your issue.](https://github.com/taikai/dappkit/runs/7069036809?check_suite_focus=true)

## Expected Behavior

The compile action to work

## Actual Results

truffle/workflow-compile complains about missing dependency,
```
Error: Cannot find module '@truffle/dashboard-message-bus-client'
```

## Environment

* Operating System: 
* Ethereum client:
* Truffle version: "@truffle/workflow-compile": "4.0.20",
* node version:  v14.17.0
* npm version: 6.14.13

moshmage:
after providing `dashboard-message-bus-client` manually via package.json dependencies, this shows up

```
internal/modules/cjs/loader.js:885
  const err = new Error(message);
              ^
Error: Cannot find module './messages'
Require stack:
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/dashboard-message-bus-common/dist/index.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/dashboard-message-bus-client/dist/client.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/dashboard-message-bus-client/dist/index.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/events/defaultSubscribers/dashboard.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/events/defaultSubscribers/index.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/events/EventManager.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/events/index.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/config/dist/index.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/workflow-compile/utils.js
- /home/runner/work/dappkit/dappkit/node_modules/@truffle/workflow-compile/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:885:15)
```

Something _very wrong_ happened on your last deploy - This is essentially blocking us from deploying a new version.

cds-amal:
Hi @moshmage, thanks for raising this dependency issue. Hopefully #5230 resolves this.

moshmage:
[...]

PR #5246, opened 2022-06-29
Fix broken imports

cliffoo:
## Related
[issue#5235](https://github.com/trufflesuite/truffle/issues/5235), [discussion#5237](https://github.com/trufflesuite/truffle/discussions/5237)

## At a glance
Some Dashboard-related packages were newly introduced in the latest release (v5.5.19), and due to misconfigured `package.json`, the following Truffle packages fail to import when individually installed.
<details>
<summary>Installable, but fail on import</summary>
<code>
@truffle/box @truffle/compile-michelson @truffle/compile-smartpy @truffle/compile-solidity @truffle/compile-vyper @truffle/config @truffle/dashboard @truffle/dashboard-message-bus @truffle/dashboard-message-bus-client @truffle/dashboard-message-bus-common @truffle/dashboard-provider @truffle/db @truffle/db-kit @truffle/environment @truffle/events @truffle/external-compile @truffle/fetch-and-compile @truffle/migrate @truffle/require @truffle/resolver @truffle/workflow-compile
</code>
</details>

This PR resolves this issue, packages should now be individually installable. Tested using Verdaccio.

This PR builds on [pull#5230](https://github.com/trufflesuite/truffle/pull/5230).
## Post-mortem
Two things:
1. To prevent broken imports.
    - Include other Truffle packages as deps in Truffle packages.
    - Dev deps should only be used in tests, not bundled into prod.
    - `package.json` should have `files` field properly configured, encompassing all necessary parts of the bundle and not just the entry point.
    - Remember peer deps.
3. To catch broken imports.
    - Currently CI doesn't include simulating install from registry. This is problematic because dep issue such as this are overlooked when testing from a Truffle dev environment, since package managers and lerna use magic to find the deps.
    - We should simulate installing from registry in CI, importing them, and testing compile when used in ts project.

## How to test
<details>
<summary>Setup Verdaccio</summary>
<ul>
<li><code>npm install -g verdaccio</code></li>
<li><code>htpasswd -bc htpasswd-file username password</code></li>
<li>Copy <a href="https://gist.github.com/cliffoo/df2d6f51179502fb4e2492ad2c8a75b9">this config</a>.</li>
<li><code>verdaccio -c path-to-config-file-you-copied</code></li>
</ul>
You can also reference <a href="https://verdaccio.org">their docs</a>.
</details>

<details>
<summary>Publish to Verdaccio</summary>
<ul>
<li><code>npm logout && yarn logout</code></li>
<li><code>npm adduser --registry http://localhost:4873</code></li>
<li><code>lerna publish from-package --no-push --registry http://localhost:4873</code></li>
</ul>
</details>

<details>
<summary>Test install</summary>
<ul>
<li>Copy <a href="https://gist.github.com/cliffoo/ad44137004c90e3fe6bba2a67590bebd">these two files</a> into an empty directory. Make sure directory is close enough to root that node can't traverse to other places to find deps.</li>
<li>In that directory <code>cat pkgs | xargs -n 1 ./auto-test-imports.sh</code></li>
<li>In another terminal <code>tail -f test.out</code></li>
</ul>
</details>

## Credits 🙏
Huge thanks to @cds-amal, @lsqproduction, and @haltman-at!

### DIFF
--- a/packages/dashboard-message-bus-client/package.json
+++ b/packages/dashboard-message-bus-client/package.json
@@ -42,12 +42,13 @@
     "axios": "^0.24.0",
     "debug": "^4.3.1",
     "delay": "^5.0.0",
-    "isomorphic-ws": "^4.0.1"
+    "isomorphic-ws": "^4.0.1",
+    "tiny-typed-emitter": "^2.1.0",
+    "ws": "^7.2.0"
   },
   "devDependencies": {
     "@types/debug": "^4.1.5",
     "@types/node": "^16.11.6",
-    "tiny-typed-emitter": "^2.1.0",
     "typescript": "^4.1.4"
   }
 }
--- a/packages/dashboard-message-bus-common/package.json
+++ b/packages/dashboard-message-bus-common/package.json
@@ -5,8 +5,11 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
-    "dist/lib"
+    "dist"
   ],
+  "directories": {
+    "dist": "dist"
+  },
   "scripts": {
     "build": "tsc",
     "prepare": "yarn build"


packages/dashboard-message-bus-common/package.json
```diff
@@ -5,8 +5,11 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
-    "dist/lib"
```
haltman-at, line 8: Yikes, thanks for catching this!

packages/dashboard-message-bus-common/package.json
```diff
@@ -5,8 +5,11 @@
...
+    "dist"
   ],
+  "directories": {
+    "dist": "dist"
```
haltman-at, line 11: I don't understand what this is doing... I can't even find documentation for `directories.dist` in the [package.json documentation](https://docs.npmjs.com/cli/v6/configuring-npm/package-json).  What is the intention of this?  Should this be here?

packages/dashboard-message-bus-client/package.json
```diff
@@ -42,12 +42,13 @@
     "axios": "^0.24.0",
     "debug": "^4.3.1",
     "delay": "^5.0.0",
-    "isomorphic-ws": "^4.0.1"
+    "isomorphic-ws": "^4.0.1",
+    "tiny-typed-emitter": "^2.1.0",
```
haltman-at, line 46: Thanks for catching this!

packages/dashboard-message-bus-client/package.json
```diff
@@ -42,12 +42,13 @@
...
+    "ws": "^7.2.0"
```
haltman-at, line 47: Confused why this dependency was added; I can't find it being used anywhere in the package, and the new depcheck script doesn't complain here either.  What's the reason for adding this?

haltman-at (commented):
Seems mostly good, confused about two things.

packages/dashboard-message-bus-common/package.json
haltman-at: Actually it looks like a bunch of our `package.json` files do weird things with `directories` that I don't understand.  (Including some of mine, due to my use of copy-paste...)  So uh it seems likely I just don't know what I'm talking about here. :P  Or alternatively maybe we should clean that up later...

packages/dashboard-message-bus-client/package.json
cds-amal: [Seems to be a peer dep of `isomorphic-ws`](https://github.com/heineiuo/isomorphic-ws/blob/master/package.json#L26)

haltman-at: Oh, huh!  Are there other places we have to deal with peer deps like this?  Could there be more of these that are missing?  (...not that those need to go in this same PR, mind you).

cds-amal (changes requested):
Hey @cliffoo, I'm hitting the red button because I want to test these packages are indeed passing. IIRC you and @gnidan discovered `@truffle/events` was missing a `dashboard-message-bus-client` (or some other dashboard dep) and this PR doesn't address the events package which should have a significant impact on the codebase.

cliffoo:
> Hey @cliffoo, I'm hitting the red button because I want to test these packages are indeed passing. IIRC you and @gnidan discovered `@truffle/events` was missing a `dashboard-message-bus-client` (or some other dashboard dep) and this PR doesn't address the events package which should have a significant impact on the codebase.

Yea I think [this commit](https://github.com/trufflesuite/truffle/commit/126b82eb1c6145f958484c9cf73aad50a02397fd) fixed this in [pull#5230](https://github.com/trufflesuite/truffle/pull/5230), the dep is there now.

packages/dashboard-message-bus-common/package.json
cliffoo: A clean-up after may be necessary. I don't understand the purpose of `directories`, but since `files` and `directories` are the two relevant `package.json` properties that are declared in some other Truffle packages, which are installable and importable, I thought it'd be prudent to include both.

I'm pretty sure `directories` can be omitted though 😜

packages/dashboard-message-bus-client/package.json
cliffoo: So I caught this because one of the import was breaking without it, can't recall which.

I think `isomorphic-ws` across the monorepo now has `ws` peer dep, so that's fine. But I'm not sure on how to find other peer dep issues like this, @cds-amal might know? Does [`dependency-check`](https://github.com/dependency-check-team/dependency-check) have a suitable flag for this?

Related: [NPM v7 automatically installs peer deps](https://github.blog/2021-02-02-npm-7-is-now-generally-available/#peer-dependencies)

haltman-at: It doesn't seem like `dependency-check` handles this, unfortunately.  Troublesome.

haltman-at (approved):
OK, I'm going to go ahead and approve this.  I'm really not sure `directories` should be there, but if you don't want to bother removing it it seems that it's probably harmless, and also we can go clean up `directories` later more generally if people care about that.

cds-amal (approved):
Nice work on this @cliffoo, let's get it merged.

packages/dashboard-message-bus-client/package.json
cds-amal: I'm thinking we can be more confident if we utilize `dependency-check` along with the `verdaccio` testing outlined in the PR's `How to test` section.

PR #5250, opened 2022-06-30
pin geth to v1.10.19

eggplantzzz:
Geth just released a new version, yay! 🥇 However, in https://github.com/ethereum/go-ethereum/releases/tag/v1.10.20 they removed the `js` subcommand. 🤕 Now CI is broken, oh my!

This pins Geth to v1.10.19 until we can find out what the proper fix is.


cds-amal (approved):
Nice one @eggplantzzz! Approved, pending tests.

haltman-at (approved):
Why does Geth break things in patch releases so often...?

PR #5253, opened 2022-06-30
Internal improvement: Further simplify scenario tests

eggplantzzz:
This PR aims to further simplify the scenario tests as they are by far the slowest tests in Truffle's codebase.

One thing this PR does is use a fixture in a test that was previously unboxing from GitHub. This will lessen the amount of network traffic required to run CI.

This also consolidates some test setup, grouping together `before` hook content as well. In the test setups there are also some places where a provider is unnecessarily created that is also removed.

There is also a lot of redundancy in the scenario tests with migrations being run many times. This also seeks to remove some amount of this redundancy as migrations are also a very costly thing to run in CI.


packages/truffle/test/scenarios/commands/console.js
```diff
@@ -69,9 +69,6 @@ describe("truffle console", () => {
         config = { working_directory: sandlot.name };
         config.logger = logger;
```
cliffoo, line 70: Santlot 🤨

cliffoo (approved):
Well I'm all in for faster CI

packages/truffle/test/scenarios/happy_path/happypath.js
```diff
@@ -1,21 +1,19 @@
...
 const CommandRunner = require("../commandRunner");
-const contract = require("@truffle/contract");
 const fs = require("fs");
 const path = require("path");
-const assert = require("assert");
+const { assert } = require("chai");
 const Server = require("../server");
 
-describe("Happy path (truffle unbox)", function () {
+describe("basic commands on a fresh project", function () {
```
cds-amal: Does it make sense to test this happy path with and without a Migrations contract?

packages/truffle/test/scenarios/migrations/production.js
```diff
@@ -75,12 +71,7 @@ describe("production", function () {
       config = await sandbox.create(project);
       config.network = "fakeRopsten";
       config.logger = logger;
-
-      const provider = new Web3.providers.HttpProvider(
-        "http://localhost:8545",
-        { keepAlive: false }
-      );
-      web3 = new Web3(provider);
+      web3 = new Web3("http://localhost:8545");
```
cds-amal, line 74: Is this needed since `web3` was created in the before (line 23)?

packages/truffle/test/sources/toyProject/contracts/Migrations.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.4.22 <0.9.0;
+
+contract Migrations {
+  address public owner = msg.sender;
+  uint public last_completed_migration;
+
+  modifier restricted() {
```
cds-amal, line 8: Does it make sense to have a fixture project with and without a Migrations contract?

cds-amal (commented):
Looks good @eggplantzzz, though I have a few questions. Does it make sense to lower the timeouts if there is a speed increase from not making network requests?

PR #5255, opened 2022-07-01
Internal improvement: Move several tests to a better location

eggplantzzz:
There is a test called "happy path" that seems out of place. It tests basic commands on a simple project. These seem more appropriately housed in the commands folder in "scenarios".


cds-amal (approved):
LGTM!

PR #5256, opened 2022-07-01
Add Sourcify networks for July 2022

haltman-at:
This PR adds support to the Sourcify fetcher for the new networks that Sourcify has added support for this month (it looks to me like they may have gone on to a monthly update schedule? We'll see if that holds up).  I excluded Crystaleum, however, as it once again has network ID different from chain ID (it uses network ID 1), which we still can't handle.

As with the way I've done things in the past, since Sourcify has added support for the godwoken testnet, I've also put the godwoken mainnet in the networks file, even though neither fetcher presently supports it.

Sourcify has also partially *removed* support for ropsten and kovan with this update, but I've left those in; unlike when Etherscan removed support for hoo, Sourcify says existing verified contracts on ropsten or kovan will remain available (you just won't be able to verify new ones), so I don't see any reason to remove our support there.

### DIFF
--- a/packages/source-fetcher/lib/networks.ts
+++ b/packages/source-fetcher/lib/networks.ts
@@ -25,6 +25,7 @@ export const networkNamesById: { [id: number]: string } = {
   43113: "fuji-avalanche",
   11111: "wagmi-avalanche",
   53935: "dfk-avalanche",
+  432201: "dexalot-avalanche",
   335: "testnet-dfk-avalance",
   40: "telos",
   41: "testnet-telos",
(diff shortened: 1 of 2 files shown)


cds-amal:
> Sourcify has also partially removed support for ropsten and kovan with this update, but I've left those in; unlike when Etherscan removed support for hoo, Sourcify says existing verified contracts on ropsten or kovan will remain available (you just won't be able to verify new ones), so I don't see any reason to remove our support there.

Do they respond with something like tx-too-new for requests they no longer support?

PR #5261, opened 2022-07-05
Use newer ethereumjs-util

### LINKED ISSUE #5260: hdwallet-provider uses old ethereumjs-util
## Issue

@truffle/hdwallet-provider [specifies ethereumjs-util@6](https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/package.json#L29). This uses an older version of @types/bn.js which aren't compatible with the new types incorporated by #5177, a breaking change rolled out in [Truffle's latest patch-level release](https://github.com/trufflesuite/truffle/releases/tag/v5.5.20).

wbt:
Hopefully fixes #5260

### DIFF
--- a/packages/hdwallet-provider/src/constructor/getOptions.ts
+++ b/packages/hdwallet-provider/src/constructor/getOptions.ts
@@ -6,7 +6,9 @@ import type * as LegacyConstructor from "./LegacyConstructor";
 // check that the first argument is a mnemonic phrase
 const isMnemonicPhrase = (
   credentials: LegacyConstructor.Credentials
-): credentials is MnemonicPhrase => typeof credentials === "string";
+): credentials is MnemonicPhrase => {
+  return typeof credentials === "string" && credentials.includes(" ");
+};
 
 // check that the first argument is a list of private keys
 const isPrivateKeys = (
@@ -19,9 +21,11 @@ const isPrivateKey = (
 ): credentials is PrivateKey =>
   typeof credentials === "string" &&
   credentials.length === 64 &&
-  // this is added since parseInt(mnemonic) should equal NaN and private keys
-  // should parse into a valid number
-  parseInt(credentials) !== NaN;
+  // this is added since parseInt(mnemonic) should equal NaN (unless it starts
+  // with a-f) and private keys should parse into a valid number - this will
+  // also parse with the largest hex value, namely "f" * 64
+  parseInt(credentials, 16) !== NaN &&
+  !credentials.includes(" ");
 
 // turn polymorphic first argument into { mnemonic } or { privateKeys }
 const getSigningAuthorityOptions = (
--- a/packages/hdwallet-provider/src/constructor/getOptions.ts
+++ b/packages/hdwallet-provider/src/constructor/getOptions.ts
@@ -4,7 +4,7 @@ import type * as Constructor from "./Constructor";
 import type * as LegacyConstructor from "./LegacyConstructor";
 
 // check that the first argument is a mnemonic phrase
-const isMnemonicPhrase = (
+const isMnemonicLike = (
   credentials: LegacyConstructor.Credentials
 ): credentials is MnemonicPhrase => {
   return typeof credentials === "string" && credentials.includes(" ");
@@ -16,7 +16,7 @@ const isPrivateKeys = (
 ): credentials is PrivateKey[] => credentials instanceof Array;
 
 // check that the first argument is a single private key (default case for invalid mnemonics)
-const isPrivateKey = (
+const isPrivateKeyLike = (
   credentials: LegacyConstructor.Credentials
 ): credentials is PrivateKey =>
   typeof credentials === "string" &&
@@ -31,7 +31,7 @@ const isPrivateKey = (
 const getSigningAuthorityOptions = (
   credentials: LegacyConstructor.Credentials
 ): Constructor.SigningAuthority => {
-  if (isPrivateKey(credentials)) {
+  if (isPrivateKeyLike(credentials)) {
     return {
       privateKeys: [credentials]
     };
@@ -39,7 +39,7 @@ const getSigningAuthorityOptions = (
     return {
       privateKeys: credentials
     };
-  } else if (isMnemonicPhrase(credentials)) {
+  } else if (isMnemonicLike(credentials)) {
     return {
       mnemonic: {
         phrase: credentials
--- a/packages/hdwallet-provider/src/constructor/getOptions.ts
+++ b/packages/hdwallet-provider/src/constructor/getOptions.ts
@@ -2,14 +2,11 @@ import type { MnemonicPhrase, PrivateKey } from "./types";
 import type { ConstructorArguments } from "./ConstructorArguments";
 import type * as Constructor from "./Constructor";
 import type * as LegacyConstructor from "./LegacyConstructor";
-import { validateMnemonic } from "ethereum-cryptography/bip39";
-import { wordlist } from "ethereum-cryptography/bip39/wordlists/english";
 
 // check that the first argument is a mnemonic phrase
 const isMnemonicPhrase = (
   credentials: LegacyConstructor.Credentials
-): credentials is MnemonicPhrase =>
-  typeof credentials === "string" && validateMnemonic(credentials, wordlist);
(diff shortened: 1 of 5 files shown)


gnidan:
Hey @wbt thanks for this! Looks like you need to commit the `yarn.lock` changes (in addition to whatever that other failure is)

wbt:
Yeah, it looks like it might actually be more involved with having to adapt to some breaking changes in the dependency.
Unfortunately, I don't have the bandwidth to take that deep dive in right now.

wbt:
The changelog about known-breaking changes is [here](https://github.com/ethereumjs/ethereumjs-monorepo/blob/master/packages/util/CHANGELOG.md#700---2020-04-30).

gnidan:
> Yeah, it looks like it might actually be more involved with having to adapt to some breaking changes in the dependency.
> Unfortunately, I don't have the bandwidth to take that deep dive in right now.

dang! maybe someone else will be able to pick this up quickly. anyway, always appreciate your contributions in any capacity 🙏

eggplantzzz:
So it looks like there was some issue where things that were "mnemonic-like" but invalid were parsed as private keys. I updated the parsing in hdwallet-provider to fix this issue and resolve the test that was failing.

packages/hdwallet-provider/src/constructor/getOptions.ts
```diff
@@ -20,29 +17,31 @@ const isPrivateKeys = (
 const isPrivateKey = (
   credentials: LegacyConstructor.Credentials
 ): credentials is PrivateKey =>
-  !isPrivateKeys(credentials) && !isMnemonicPhrase(credentials);
+  typeof credentials === "string" &&
+  credentials.length === 64 &&
+  // this is added since parseInt(mnemonic) should equal NaN and private keys
+  // should parse into a valid number
```
cds-amal, line 23: Is this true? You can pass ganache a mnemonic like that looks like a private-key and it would generate priv/pub keys from that.  `ganache --wallet.mnemonic 2cf177c279d1b414ff675b914adee1349d040dd21dd6f64d335dce83da582c8a`

cds-amal (commented):
Looks good, tests are passing. I have a question about sniffing out a private-key.

packages/hdwallet-provider/src/constructor/getOptions.ts
```diff
@@ -2,14 +2,11 @@ import type { MnemonicPhrase, PrivateKey } from "./types";
...
 import type * as LegacyConstructor from "./LegacyConstructor";
-import { validateMnemonic } from "ethereum-cryptography/bip39";
-import { wordlist } from "ethereum-cryptography/bip39/wordlists/english";
 
 // check that the first argument is a mnemonic phrase
 const isMnemonicPhrase = (
   credentials: LegacyConstructor.Credentials
-): credentials is MnemonicPhrase =>
-  typeof credentials === "string" && validateMnemonic(credentials, wordlist);
+): credentials is MnemonicPhrase => typeof credentials === "string";
```
haltman-at, line 9: Why was the validation removed here?  I assume this is related to what you said on the PR comments but the explanation wasn't really clear... could you explain in more detail?

packages/hdwallet-provider/src/constructor/getOptions.ts
```diff
@@ -20,29 +17,31 @@ const isPrivateKeys = (
 const isPrivateKey = (
   credentials: LegacyConstructor.Credentials
 ): credentials is PrivateKey =>
-  !isPrivateKeys(credentials) && !isMnemonicPhrase(credentials);
+  typeof credentials === "string" &&
+  credentials.length === 64 &&
+  // this is added since parseInt(mnemonic) should equal NaN and private keys
+  // should parse into a valid number
+  parseInt(credentials) !== NaN;
```
haltman-at, line 24: This should be `parseInt(credentials, 16)`, right?  Without the 16, it won't work unless there's a `0x` in front, and I don't think we want to require that.

I also think using `parseInt` in this way is pretty iffy, because the number is 32 bytes long.  It technically works, but I think if you want to do this it's at least worth making a comment saying something like "yes, this works even if the number is outside the safe integer range!".

packages/hdwallet-provider/src/constructor/getOptions.ts
haltman-at: Also, I think this function should be renamed... or actually, removed.  The name is misleading now because it's no longer performing validation.  Looking at the changes below, this makes sense, but the function name is misleading, so I think the function should be done away with.

packages/hdwallet-provider/src/constructor/getOptions.ts
```diff
@@ -20,29 +17,31 @@ const isPrivateKeys = (
...
-      }
+      privateKeys: [credentials]
     };
   } else if (isPrivateKeys(credentials)) {
     return {
       privateKeys: credentials
     };
-  } else if (isPrivateKey(credentials)) {
-    // if(...) included for explicitness
+  } else if (isMnemonicPhrase(credentials)) {
```
haltman-at, line 38: Per my comments above, I think the function `isMnemonicPhrase` should be eliminated if it's not going to actually perform any validation, and instead this condition should just be `typeof credentials === "string"`; you could then add a comment noting that in this case you assume it's a mnemonic.

haltman-at (commented):
I think there's a bug in this relating to the private key parsing.  I also have some style comments.

packages/hdwallet-provider/src/constructor/getOptions.ts
eggplantzzz: Consider the case where I use "I am having fun" as my mnemonic. In that case this check will say "hey, that is not a valid mnemonic and therefore is not a mnemonic - so it must be a private key". Then it returns the `signingAuthority` stuff in an incorrect form: namely like
```
{ privateKey: "I am having fun" }
```
It seems to me that here we just want to figure out whether what they entered is "mnemonic-like". Later we do this validation again anyways.

packages/hdwallet-provider/src/constructor/getOptions.ts
eggplantzzz: Yup, the 16 is definitely missing! Mnemonics technically can parse into a number if they start with the letters a-f so @cds-amal recommended checking for " "s as well. I'll add a comment here mentioning that "f" *64 still parses into a number in JS.

packages/hdwallet-provider/src/constructor/getOptions.ts
eggplantzzz: For reference, we validate the mnemonic again [here](https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/src/index.ts#L312).

packages/hdwallet-provider/src/constructor/getOptions.ts
haltman-at: Makes sense!

packages/hdwallet-provider/src/constructor/getOptions.ts
haltman-at: Yeah if you add the spaces check that makes sense!

haltman-at (approved):
Hitting approve on this, seems good to me!  I think `isMnemonicLike` would be a better name than `isMnemonicPhrase` but now that there's a least *some* validation it doesn't bother me. :P

packages/hdwallet-provider/src/constructor/getOptions.ts
```diff
@@ -19,9 +21,11 @@ const isPrivateKey = (
...
   typeof credentials === "string" &&
   credentials.length === 64 &&
-  // this is added since parseInt(mnemonic) should equal NaN and private keys
-  // should parse into a valid number
-  parseInt(credentials) !== NaN;
+  // this is added since parseInt(mnemonic) should equal NaN (unless it starts
+  // with a-f) and private keys should parse into a valid number - this will
+  // also parse with the largest hex value, namely "f" * 64
+  parseInt(credentials, 16) !== NaN &&
+  !credentials.includes(" ");
```
cds-amal, line 28: thanks for this!

haltman-at:
Oh, wait, this is failing CI for some reason...

cds-amal (approved):
LGTM pending tests ofc :)

cds-amal:
The updated function names are much better.

PR #5262, opened 2022-07-05
resolver: update dependency: abi-to-sol@^0.6.2

cds-amal:
`abi-to-sol` had a minor bump a few releases ago and semver range wouldn't resolve to its new version. This PR fixes that.

abi-to-sol
  * @truffle/resolver: ^0.5.2 → ^0.6.2


haltman-at (approved):
Oh yikes, thanks for catching this!

Issue #5236, opened 2022-06-27
Add GHA to fail CI on missing dependencies


cds-amal:
We should add a GHA script to utilize #5230

cliffoo:
Fixed in [pull#5264](https://github.com/trufflesuite/truffle/pull/5264) and in the [latest release v5.5.22](https://github.com/trufflesuite/truffle/releases/tag/v5.5.22). Thanks @cds-amal!

PR #5264, opened 2022-07-06
Add GHA dependency check

cds-amal:
This PR builds on #5230 to address #5236 by adding a GitHub action to CI. I didn't create a separate job for the dependency check because it would require the expensive `yarn bootstrap` command and I don't want CI to take longer. Instead, I inserted the dependency-check into the matrix strategy `build` job to take advantage of the matrix's short-circuiting behavior.  If one matrix job fails, it cancels all others. I also updated the Slack-app config and made the matrix-build job a dependency of `yarncheck`. `yarncheck` resolves in ~2.5 minutes and then the matrix job runs, followed by Slack notification.

If this is approved, I will rebase to remove all simulated errors

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -56,7 +56,7 @@ jobs:
 
     - run: yarn bootstrap
     
-    - run: yarn depcheck || (echo 'Please add missing dependencies to appropriate package'; false)
+    - run: yarn depcheck
     
     - run: ${{ matrix.env }} yarn ci
       env:
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           channel: '#truffle-ci-notifications'
           status: ${{ job.status }}
-          fields: commit,author,pullRequest
+          fields: job,commit,author,pullRequest
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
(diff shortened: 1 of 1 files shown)


.github/workflows/nodejs.yml
```diff
@@ -44,29 +53,33 @@ jobs:
       run: npm -g install npm@6
 
     - run: npm install -g yarn
+
     - run: yarn bootstrap
+    
+    - run: yarn depcheck || (echo 'Please add missing dependencies to appropriate package'; false)
```
haltman-at, line 59: I think we're probably better off without the explicit custom error message here; I think the existing error is clear enough and the method for including it is kind of hacky.  I only put an explicit custom error message on yarncheck because otherwise it's just not clear at all with that one.

(If `yarn --frozen-lockfile` worked like it was supposed to, and I didn't need the custom test, then that would probably be clear and I could omit the custom error... but it doesn't, so yeah.)

cds-amal:
> Hm, would it make more sense to just put the test in the `yarncheck` job? My understanding was that a failure of `yarncheck` would also cancel the others... is that not the case? That's what I've always observed...

I would prefer doing these build/lint type checks in one job, but `dependency-check` needs `yarn bootstrap` to build out the entry points; I wanted to avoid adding another 12-13 minutes to the `yarncheck` job. I'm also unsure why the cancel behavior changed, as I remember it working as you did, but testing showed otherwise.

.github/workflows/nodejs.yml
cds-amal: Sure, I'll remove that.

haltman-at:
Makes sense!

.github/workflows/nodejs.yml
cds-amal: Simulated an error [and it looks ok](https://github.com/trufflesuite/truffle/runs/7254092415).

cds-amal:
Will rebase to remove the simulated errors

haltman-at (approved):

Issue #5186, opened 2022-06-08
truffle --version crashes, truffle help doesn't crash


agro1986:
## Issue

`truffle --version` crashes

## Steps to Reproduce

run this:

```
truffle --version
```

## Actual Results

```
TypeError: Cannot read properties of null (reading 'name')
    at prepareOptions (/Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/webpack:/packages/core/lib/command-utils.js:86:39)
    at Object.586806 (/Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/webpack:/packages/core/cli.js:65:1)
    at __webpack_require__ (/Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/webpack:/webpack/bootstrap:19:1)
    at /Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/webpack:/webpack/bootstrap:37:1
    at Function.__webpack_require__.O (/Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/webpack:/webpack/runtime/chunk loaded:23:1)
    at __webpack_require__.x (/Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/webpack:/webpack/bootstrap:38:1)
    at Function.__webpack_require__.x (/Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/webpack:/webpack/runtime/startup chunk dependencies:36:1)
    at /Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/webpack:/webpack/startup:3:1
    at Object.<anonymous> (/Users/xxx/.nvm/versions/node/v16.15.1/lib/node_modules/truffle/build/cli.bundled.js:628:12)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:77:12)
    at node:internal/main/run_main_module:17:47
```

## Environment

* Operating System: MacOS 12.4 (21F79)
* Truffle version (`truffle version`): v5.5.17
[...]

PR #5266, opened 2022-07-07
core: error better on scenarios with incorrect truffle command

lsqproduction:
Address #5186, truffle crashes when an invalid truffle command is used. 
For example `truffle --version` or `truffle liang`
Added a better error and prompts the user to use `truffle help` to see the list of available valid commands. 

I thought it is cleaner looking to just display the error message instead of the error message and stack trace together, do let me know if you think otherwise. Or if you have a better suggestion on how to handle this error.

### DIFF
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -63,6 +63,15 @@ const command = getCommand({
   options: {},
   noAliases: false
 });
+
+//getCommand() will return null if a command not recognized by truffle is used.
+if (command === null) {
+  console.log(
+    `\`truffle ${inputStrings}\` is not a valid truffle command. Please see \`truffle help\` for available commands.`
+  );
+  process.exit(1);
+}
+
 const options = prepareOptions({
   command,
   inputStrings,
--- a/packages/core/lib/command-utils.js
+++ b/packages/core/lib/command-utils.js
@@ -47,7 +47,7 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
     }
   }
 
-  if (chosenCommand == null) {
+  if (chosenCommand === null) {
     return null;
   }
 


packages/core/lib/command-utils.js
```diff
@@ -48,7 +48,12 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
   }
 
   if (chosenCommand == null) {
-    return null;
+    //error out and exit when we cannot find any command that matches.
+    let err = new Error(
+      `\`truffle ${firstInputString}\` is not a valid truffle command. Please see \`truffle help\` for available commands.`
+    );
+    console.log(err.message);
+    process.exit(err);
```
eggplantzzz, line 56: in general you don't call `process.exit` with an `Error` object. you should call it with an exit code: `0` if everything went successfully and `1` if not. you can check out the docs on this [here](https://nodejs.dev/learn/how-to-exit-from-a-nodejs-program)

packages/core/cli.js
```diff
@@ -63,11 +63,22 @@ const command = getCommand({
   options: {},
   noAliases: false
 });
-const options = prepareOptions({
-  command,
-  inputStrings,
-  options: {}
-});
+const options = (function () {
```
cds-amal, line 66: Does this have to be an [IFFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE)?

cds-amal (commented):
Thanks for this. I have one question.

packages/core/lib/command-utils.js
cds-amal: Note: this file already uses `process.exit` in other places.

packages/core/lib/command-utils.js
```diff
@@ -48,7 +48,12 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
   }
 
   if (chosenCommand == null) {
```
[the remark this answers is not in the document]
lsqproduction: done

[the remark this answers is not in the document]
eggplantzzz: thanks!

packages/core/cli.js
eggplantzzz: Yeah I don't know why this should be constructed inside a function like this. It seems like it should read something like
```
const command = getCommand(....
if (command === null) {
  // deal with error stuff
}
const options = prepareOptions(...
```

Or maybe you have a reason for writing it the way you did?

packages/core/cli.js
```diff
@@ -63,11 +63,22 @@ const command = getCommand({
...
+  //check if command returned null
```
eggplantzzz, line 67: Just a general note here; if you leave comments I would say make them explain something that is not obvious in the code. :) Like, it is pretty clear that we are checking for a `null` value. What might be more useful is explaining why. I think comments should attempt to add more context or provide extra insight that is not obvious.

packages/core/cli.js
```diff
@@ -63,6 +63,16 @@ const command = getCommand({
   options: {},
   noAliases: false
 });
+
+//getCommand() will return null if a command not recognized by truffle is used.
+if (command === null) {
+  let err = new Error(
```
eggplantzzz, line 69: If we are not going to throw this `Error`, there is no sense in creating it. Just log and `exit`.

packages/core/cli.js
lsqproduction: agreed. I changed it.

packages/core/cli.js
lsqproduction: okay. will be making the change.

eggplantzzz (approved):
I like this

PR #5274, opened 2022-07-08
Internal improvement: Add missing `buffer` import in `dashboard-message-bus-common`

cliffoo:
It runs fine in node but not in the browser, which creates problems when dashboard's frontend uses it.

### DIFF
--- a/packages/dashboard-message-bus-common/lib/utils.ts
+++ b/packages/dashboard-message-bus-common/lib/utils.ts
@@ -1,4 +1,5 @@
 import { Message } from "./messages";
+import { Buffer } from "buffer";
 
 /**
  * Convert any JS object or value to a base64 representation of it


cds-amal (approved):
LGTM! Good to be explicit!

eggplantzzz:
This code runs in the browser? How does that happen? I can't see any kind of webpacking or whatnot that would allow it to.

cliffoo:
> This code runs in the browser? How does that happen? I can't see any kind of webpacking or whatnot that would allow it to.

🤷‍♂️ There is no bundling, but I get import errors without this in the browser, and no errors with it.
Maybe ts compile does some magic to not bundle / polyfill `"buffer"` and resolve it somehow?

cliffoo:
Not 100% sure but here's how I understand it to work:

1. Module _m_ imports a node built-in module that cannot be used directly in the browser.
2. Web app _w_ installs _m_ as dependency.
3. Web app's bundler resolves _m_'s dependencies using a bundler like webpack, so something like `"buffer"` will point to an isomorphic implementation that works in both the browser and node.

I guess without step 1 this won't work? Because the bundler just sees an undefined var like `Buffer`, without knowing it's supposed to come from `import { Buffer } from "buffer"`

eggplantzzz:
I'm really not asking specifically about `Buffer`, I was just wondering how this runs in the browser in general. You are probably correct that some other package bundles it in!

Issue #5160, opened 2022-06-01
Deprecate/remove eth-pm support


cds-amal:
Our maintenance/dependency cost is quite high for an ecosystem that doesn't seem to have community support. We should deprecate/remove `truffle install` and `truffle publish`.

cliffoo:
Fixed in [pull#5283](https://github.com/trufflesuite/truffle/pull/5283) and in the [latest release v5.5.22](https://github.com/trufflesuite/truffle/releases/tag/v5.5.22). Thanks @sukanyaparashar!

PR #5283, opened 2022-07-12
Deprecate/remove ethpm support

sukanyaparashar:
### ISSUE
`truffle install` and `truffle publish` are broken since a long time and we needed to remove the `ethpm` support. See #5160.

### SOLUTION
This PR disables/removes the `ethpm` support from truffle and also removes the `ethpm` and `ethpm-registry` dependencies from `package.json` in `truffle core`.
 - [x] Remove `install` and `publish` from global `help`.
 - [x] Show helpful message when `install` or `publish` is invoked.
 - [x] Remove `help` for both `install` and `publish`.


packages/core/lib/commands/publish/run.js
```diff
@@ -1,7 +1,8 @@
...
-  const Package = require("../../package");
-
-  const config = Config.detect(options);
-  return await Package.publish(config);
+module.exports = async function () {
+  const emoji = String.fromCodePoint(0x1f61e);
+  console.log(
+    emoji +
+      " Oops! ethpm has moved on faster than truffle's integration. " +
+      "This feature has been broken for some time, so we've disabled it."
```
cds-amal, line 6: We should probably create a discussion and provide the url for the curious.

packages/core/package.json
```diff
@@ -59,8 +59,6 @@
     "del": "^2.2.0",
     "ethereum-cryptography": "1.0.3",
     "ethereumjs-wallet": "^1.0.1",
-    "ethpm": "0.0.19",
-    "ethpm-registry": "0.1.0-next.3",
```
cds-amal, line 63: adios!

packages/core/lib/commands/install/run.js
```diff
@@ -1,9 +1,7 @@
...
-
-  if (options._ && options._.length > 0) options.packages = options._;
-
-  const config = Config.detect(options);
-  return await Package.install(config);
+module.exports = async function () {
+  const emoji = String.fromCodePoint(0x1F61E);
+  console.log(
+    emoji + " Oops! ethpm has moved on faster than truffle's integration. " +
+      "This feature has been broken for some time, so we've disabled it."
```
cds-amal, line 5: Should link to the same discussion url

packages/core/lib/commands/install/meta.js
```diff
@@ -1,9 +1,10 @@
 module.exports = {
   command: "install",
-  description: "Install a package from the Ethereum Package Registry",
+  description:
+    "[deprecated] Install a package from the Ethereum Package Registry",
   builder: {},
   help: {
-    usage: "truffle install <package_name>[@<version>]",
+    usage: "[deprecated] truffle install <package_name>[@<version>]",
```
cds-amal, line 7: It seems odd to describe how to use a command that a user can't because it isn't implemented. Does a deprecated command need `usage/description/options`? What do y'all think?

```console
$ truffle help install

  Usage:        [deprecated] truffle install <package_name>[@<version>]
  Description:  [deprecated] Install a package from the Ethereum Package Registry
  Options:
                package_name
                    Name of the package as listed in the Ethereum Package Registry. (required)
                <@version>
                    When specified, will install a specific version of the package, otherwise will install
                    the latest version.
```

cds-amal (commented):
Thanks for taking this on @sukanyaparashar! There should be a link to a discussion (to be created). I also have a couple of questions on the deprecation messaging that others should weigh in on.

packages/core/lib/commands/install/meta.js
sukanyaparashar: I guess @cds-amal, we can just remove `help` and only keep the command `description` and mention as `deprecated`.

cds-amal: @eggplantzzz thoughts?

packages/core/lib/commands/install/run.js
```diff
@@ -1,9 +1,7 @@
-module.exports = async function (options) {
-  const Config = require("@truffle/config");
-  const Package = require("../../package");
-
-  if (options._ && options._.length > 0) options.packages = options._;
-
-  const config = Config.detect(options);
-  return await Package.install(config);
+module.exports = async function () {
```
lsqproduction, line 1: curious about the async here. Is it needed?

sukanyaparashar: Thanks for pointing this out @lsqproduction. `async` is not needed now since we have removed the `await` call. I will remove it.

packages/core/lib/commands/install/meta.js
kevinweaver: I'd be curious what other libraries do, but if the command no longer exists, I'd personally say it's reasonable to remove the `help` as well.

cds-amal: > I guess @cds-amal, we can just remove `help` and only keep the command `description` and mention as `deprecated`.

Where would the `description` be used if `help` is removed? Maybe it's better to not display help for these commands and indicate deprecation in the stubbed implementation.

eggplantzzz: My first thought is that @cds-amal's suggestion of removing it from the help and providing messaging in the stubbed implementation is good. We probably shouldn't advertise it if it is broken/deprecated.

packages/core/lib/commands/install/run.js
```diff
@@ -1,9 +1,7 @@
...
+  const emoji = String.fromCodePoint(0x1F61E);
+  console.log(
+    emoji + " Oops! ethpm has moved on faster than truffle's integration. " +
```
eggplantzzz, line 4: cute :)

eggplantzzz:
So overall I like the idea of removing `install` and `publish` from the list of commands and just having messaging where the command implementation used to be. Anyone disagree?

cds-amal:
I do not disagree

packages/core/lib/command-utils.js
```diff
@@ -187,7 +187,9 @@ const runCommand = async function (command, options) {
 const displayGeneralHelp = () => {
   const yargs = require("yargs/yargs")();
   commands.forEach(command => {
-    yargs.command(require(`./commands/${command}/meta`));
+    if (command !== "install" && command !== "publish") {
```
eggplantzzz, line 190: maybe leave a comment here about why we're doing this

eggplantzzz (approved):
other than leaving the comment I think this should be good to go

packages/core/lib/command-utils.js
sukanyaparashar: Yes, that makes sense. I will add a comment.

cds-amal (approved):
LGTM, though @eggplantzzz may have some feedback

eggplantzzz (approved):
seems cool

cliffoo:
[Integration tests are removed](https://github.com/trufflesuite/truffle/pull/5317), reflected in the latest release v5.5.23

PR #5286, opened 2022-07-13
Enhancement: Add `debug_storageRangeAt` integration to debugger library

haltman-at:
Partly, but not fully, addresses #2894.  It's not fully because the PR deliberately leaves out the CLI side of things, and only adds the feature to the library.  The reason for this is that for now we're holding off on the CLI side of things until trufflesuite/ganache#3338 is fixed.  (I have the code for it written, it's just not in this PR, and also I suspect I might want to change it some after discussing things with @gnidan.)

(**Edit**: This PR is now a **breaking change** to `debugger`; see later comments about why.)

(Also because #2894 contains some other stuff which should really be split out into its own issue (or multiple such), but, I'll do that later. :P )

Anyway!  This PR adds to the debugger library a new option, `storageLookup`.  This option can be passed to `forTx` or to `load` (this may seem an odd interface but the reasoning will be explained below); if it is, then unknown storage, rather than being reported as zero or unknown, will be looked up using `debug_storageRangeAt`.  If the client does not support this method, then `forTx` or `load` will throw if this option is passed.

Note that any storage fetched with `debug_storageRangeAt` *will* be recorded in the codex, meaning that what storage the debugger knows about can be affected by a call to `decode`; moreover, the debugger will continue to know about such storage after a reset (though not an unload).  This may seem a bit strange, but with the planned CLI implementation, this quirk won't be visible to the CLI user.  Again, this PR does not include this implementation, so the interface I've created here may seem a bit annoying in places; I assure you, however, that such an implementation exists and smooths over these issues for the users.  Library users however should be aware of these potential sharp corners!  (Anyway this phenomenon already occurs with storage touched during the transaction!  With the planned CLI implementation, things will actually be *more* consistent now.)

So, how does all this work?

Well, first off, I've added a method to the adapter for using `debug_storageRangeAt`.  There's then also corresponding sagas for using that method, `obtainStorage`, much like we have `obtainBinaries`.  Note that because it's important for detection purposes that `obtainStorage` throw a clean, catchable error if `debug_storageRangeAt` is not supported (and because no we are not putting the adapter in the state), we have to do some trickery here... I'm not going to recap all the code, but yes, that is indeed a `return error;`.  Ultimately though it all works out, and `obtainStorage` will throw as expected if `debug_storageRangeAt` is not supported.

Secondly, the `decode` saga in the debugger now checks if `storageLookup` is turned on, and, if it is, instead of returning zero or unknown when storage is unknown, calls `storageLookup`, which in turn calls `obtainStorage`.  Note that the result is stored in the codex.

Now, the thing about `requestStorage` is that it needs to know the current block hash and transaction index, which previously we didn't keep track of; so, the `evm` submodule now records that information.

So how does all this work on debugger startup / transaction load?  Basically the main thing here is we need to determine if `debug_storageRangeAt` is supported.  How do we do that?  We request some storage and see if we get an error or not!

The problem here is, in order to ensure we don't get an error if it's supported, we need to make our request about a block that *has at least one transaction in it*.  This is why `storageLookup` is an option to `forTx` and to `load`, and not to say `forProject` or `startFullMode` (as I'd originally written things) -- because we use the block that the transaction we're debugging is in, which means we have to be debugging a transaction!  I don't know of any other good way to find a block with at least one transaction in it.

Admittedly, we could do without this if instead our detection method checked the type of the error, to distinguish between a "not supported" error and other types of errors; however, I didn't do that as I was worried about potential variance between clients here.  I'm just not sure that such a method will be robust enough (without a bunch of per-client special cases that would have to be updated as we support more clients, which I don't like the idea of).

Anyway, on first transaction load, we test whether `debug_storageRangeAt` is supported, and set a flag in the state accordingly.  Once it's set, we don't bother rechecking on subsequent transaction loads.  Note that this is separate from the flag that says whether `debug_storageRangeAt` is being *used*, since after all that can differ from load to load.  Also note that we do this even if we're in light mode, since the user may choose to start full mode later.  (Remember, light mode turns off the data listener, but it's still possible to invoke data sagas manually, and we do!)

Because this check requires knowing the block hash, it (under the current setup) has to occur after `evm.begin()`, which puts it pretty late in the setup process... I'm actually not entirely comfortable with this, as it means that maybe it could take a while before it errors out.  Possibly all of this should be moved earlier?  @gnidan, any opinions on that?

Anyway, I think that mostly covers it.  I also added a test.  Note that since for the test we want two transactions in a single block, I used `miner_stop` to turn off automining and then `evm_mine` to manually perform mines.  I also changed `prepareContracts` so that if you pass `null` (not `undefined`) for the `migrations`, it'll skip the migrations... we use this here to avoid having to deal with the interaction of migrations and manual mining.

And I think that sums it up!

### DIFF
--- a/packages/debugger/lib/data/actions/index.js
+++ b/packages/debugger/lib/data/actions/index.js
@@ -99,20 +99,3 @@ export function allocate(
     state
   };
 }
-
-export const SET_STORAGE_VISIBILITY = "DATA_SET_STORAGE_VISIBILITY";
-export function setStorageVisibility(visibility) {
-  return {
-    type: SET_STORAGE_VISIBILITY,
-    visibility
-  };
-}
-
-export const SET_STORAGE_VISIBILITY_SUPPORT =
-  "DATA_SET_STORAGE_VISIBILITY_SUPPORT";
-export function setStorageVisibilitySupport(visibility) {
-  return {
-    type: SET_STORAGE_VISIBILITY_SUPPORT,
-    visibility
-  };
-}
--- a/packages/debugger/lib/data/actions/index.js
+++ b/packages/debugger/lib/data/actions/index.js
@@ -99,3 +99,20 @@ export function allocate(
     state
   };
 }
+
+export const SET_STORAGE_VISIBILITY = "DATA_SET_STORAGE_VISIBILITY";
+export function setStorageVisibility(visibility) {
+  return {
+    type: SET_STORAGE_VISIBILITY,
+    visibility
+  };
+}
+
+export const SET_STORAGE_VISIBILITY_SUPPORT =
+  "DATA_SET_STORAGE_VISIBILITY_SUPPORT";
+export function setStorageVisibilitySupport(visibility) {
+  return {
+    type: SET_STORAGE_VISIBILITY_SUPPORT,
+    visibility
+  };
+}
--- a/packages/debugger/lib/data/reducers.js
+++ b/packages/debugger/lib/data/reducers.js
@@ -310,14 +310,36 @@ function mappedPaths(state = DEFAULT_PATHS, action) {
   }
 }
 
+function allStorageVisible(state = null, action) {
+  if (action.type === actions.SET_STORAGE_VISIBILITY) {
+    return Boolean(action.visibility); //force Boolean to prevent undefined
+  } else {
+    return state;
+  }
+}
+
+function storageVisibilitySupported(state = null, action) {
+  if (action.type === actions.SET_STORAGE_VISIBILITY_SUPPORT) {
+    return action.visibility;
+  } else {
+    return state;
+  }
+}
+
 const proc = combineReducers({
   assignments,
   mappedPaths
 });
 
+const application = combineReducers({
+  storageVisibilitySupported,
+  allStorageVisible
+});
+
 const reducer = combineReducers({
   info,
-  proc
+  proc,
+  application
 });
 
 export default reducer;
--- a/packages/debugger/lib/data/reducers.js
+++ b/packages/debugger/lib/data/reducers.js
@@ -310,36 +310,14 @@ function mappedPaths(state = DEFAULT_PATHS, action) {
   }
 }
 
-function allStorageVisible(state = null, action) {
-  if (action.type === actions.SET_STORAGE_VISIBILITY) {
-    return Boolean(action.visibility); //force Boolean to prevent undefined
-  } else {
-    return state;
-  }
-}
-
-function storageVisibilitySupported(state = null, action) {
-  if (action.type === actions.SET_STORAGE_VISIBILITY_SUPPORT) {
-    return action.visibility;
-  } else {
-    return state;
-  }
-}
-
 const proc = combineReducers({
   assignments,
   mappedPaths
 });
 
-const application = combineReducers({
-  storageVisibilitySupported,
-  allStorageVisible
-});
-
 const reducer = combineReducers({
   info,
-  proc,
-  application
+  proc
 });
 
 export default reducer;
--- a/packages/debugger/lib/data/sagas/index.js
+++ b/packages/debugger/lib/data/sagas/index.js
@@ -1043,13 +1043,10 @@ export function* decode(
   const internalFunctionsTable = yield select(
     data.current.functionsByProgramCounter
   );
-  const storageLookup = yield select(data.application.storageLookup);
 
   debug("definition: %o");
   debug("ref: %o");
   debug("compilationId: %s", compilationId);
-
-  const ZERO_WORD = new Uint8Array(Codec.Evm.Utils.WORD_SIZE); //automatically filled with zeroes
 
   const decoder = Codec.decodeVariable(
     definition,
@@ -1074,18 +1071,7 @@ export function* decode(
     let response;
     switch (request.type) {
       case "storage":
-        if (storageLookup) {
-          response = yield* evm.requestStorage(request.slot);
-        } else if (indicateUnknown) {
-          //the debugger supplies all storage it knows at the beginning.
-          //so anything it gets a request for can be presumed to be
-          //unknown.
-          response = null;
-        } else {
-          //if indicateUnknown isn't set, all unknown storage is assumed
-          //to be zero.
-          response = ZERO_WORD;
-        }
+        response = yield* evm.requestStorage(request.slot, indicateUnknown);
         break;
       case "code":
         response = yield* evm.requestCode(request.address);
--- a/packages/debugger/lib/data/sagas/index.js
+++ b/packages/debugger/lib/data/sagas/index.js
@@ -1044,6 +1044,7 @@ export function* decode(
   const internalFunctionsTable = yield select(
     data.current.functionsByProgramCounter
   );
+  const allStorageVisible = yield select(data.application.allStorageVisible);
 
   debug("definition: %o");
   debug("ref: %o");
@@ -1074,11 +1075,18 @@ export function* decode(
     let response;
     switch (request.type) {
       case "storage":
-        //the debugger supplies all storage it knows at the beginning.
-        //any storage it does not know is presumed to be zero.
-        //(unlesss indicateUnknown is passed, in which case we use
-        //null as a deliberately invalid response)
-        response = indicateUnknown ? null : ZERO_WORD;
+        if (allStorageVisible) {
+          response = yield* requestStorage(request.slot);
+        } else if (indicateUnknown) {
+          //the debugger supplies all storage it knows at the beginning.
+          //so anything it gets a request for can be presumed to be
+          //unknown.
+          response = null;
+        } else {
+          //if indicateUnknown isn't set, all unknown storage is assumed
+          //to be zero.
+          response = ZERO_WORD;
+        }
         break;
       case "code":
         response = yield* requestCode(request.address);
@@ -1087,6 +1095,7 @@ export function* decode(
         debug("unrecognized request type!");
     }
     debug("sending response");
+    debug("response: %O", response);
     result = decoder.next(response);
   }
   //at this point, result.value holds the final value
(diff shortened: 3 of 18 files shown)


gnidan:
> Because this check requires knowing the block hash, it (under the current setup) has to occur after `evm.begin()`, which puts it pretty late in the setup process... I'm actually not entirely comfortable with this, as it means that maybe it could take a while before it errors out. Possibly all of this should be moved earlier? @gnidan, any opinions on that?

It does strike me as the sort of thing that should happen earlier in the startup process, but it's probably not worth moving it earlier unless it cleans something up along the way.

packages/debugger/lib/data/actions/index.js
```diff
@@ -99,3 +99,20 @@ export function allocate(
     state
   };
 }
+
+export const SET_STORAGE_VISIBILITY = "DATA_SET_STORAGE_VISIBILITY";
```
gnidan, line 103: "storage visibility" is a weird name... wouldn't this be better named something like "storage fetching" or "intertransactional storage" or something?

packages/debugger/lib/web3/sagas/index.js
```diff
@@ -184,6 +194,59 @@ function* receiveBinary(address) {
   return binary;
 }
 
+export function* obtainStorage(address, slot, blockHash, txIndex) {
```
gnidan, line 197: this flow all makes sense 👌

packages/debugger/test/data/fetch-storage.js
```diff
@@ -0,0 +1,130 @@
...
+  async function mine() {
+    await promisify(provider.send.bind(provider))({
+      jsonrpc: "2.0",
+      method: "evm_mine",
+      id: Date.now(),
+      params: [Date.now()]
+    });
+  }
+
+  async function waitForSent(promiEvent) {
```
gnidan, line 50: a better name for this might be `waitForTransactionHash`, since that's what it returns?

packages/debugger/test/data/fetch-storage.js
```diff
@@ -0,0 +1,130 @@
...
+
+    const prepared = await prepareContracts(provider, sources, null); //skip migrations, we're doing this manually!
+    abstractions = prepared.abstractions;
+    compilations = prepared.compilations;
+  });
+
+  it("Displays values from previous transactions", async function () {
+    this.timeout(12000);
+
+    const instancePromiEvent = abstractions.SimpleStorage.new(1); //don't await it yet!
```
gnidan, line 89: i hate having to do this sorta thing with @truffle/contract

packages/debugger/test/data/fetch-storage.js
```diff
@@ -0,0 +1,130 @@
...
+    });
+
+    await bugger.stepNext(); //just step into the contract
+
+    const variables = Codec.Format.Utils.Inspect.unsafeNativizeVariables(
+      await bugger.variables()
+    );
+
+    const expectedResult = {
+      x: 3 //previous value
```
gnidan, line 125: hey that's cool :)

packages/debugger/lib/session/sagas/index.js
```diff
@@ -133,16 +134,20 @@ function* startFullMode() {
 
   yield* trace.addSubmoduleToCount(listenersToActivate.length);
 
-  //begin any full-mode modules that need beginning
-  yield* txlog.begin();
+  if (yield select(session.status.loaded)) {
```
gnidan, line 137: what happens if this isn't true here?

packages/debugger/lib/data/reducers.js
```diff
@@ -310,14 +310,36 @@ function mappedPaths(state = DEFAULT_PATHS, action) {
   }
 }
 
+function allStorageVisible(state = null, action) {
```
gnidan, line 313: do these two reducers (and the corresponding actions/selectors) belong in the data app-module? it seems like they should part of evm alongside the actual retrieved storage records.

(also, you mention [here](https://github.com/trufflesuite/truffle/pull/5286/files#diff-c904fc92588b6e51e83081a80402bf454bc2936777a7d010c7249a6507f874b1R225) that this gets set even in light mode, and data normally doesn't run in light mode IIRC)

packages/debugger/lib/data/sagas/index.js
```diff
@@ -1286,6 +1295,64 @@ function* requestCode(address) {
...
+  //NOTE: this is also assuming that no individual variable ever requires
+  //reading the same storage value twice in the course of decoding!
+  //this assumption is presently true, but if it ever becomes false, then it
+  //will become necessary to add that check here (not that the result would
+  //be *wrong* without such a check, but it would cause unnecessary network
+  //requests)
+  const address = yield select(data.current.address);
+  const blockHash = yield select(data.views.blockHash); //cannot use number here!
+  const txIndex = yield select(data.views.txIndex);
+  const word = yield* web3.obtainStorage(address, slot, blockHash, txIndex);
```
gnidan, line 1313: what happens if this returns an error?

haltman-at:
> edit: Looking at the code, I think it makes sense for to be part of the EVM module there. Block hash + tx index is part of the EVM

Oh, my question wasn't whether that should be moved out of EVM, but rather whether `evm.begin()` should be moved earlier.

packages/debugger/lib/data/reducers.js
gnidan: I guess the counterargument is that the data saga is the one that does the orchestration of storage fetching... thus the data module should be responsible for the associated flags.

It's just a bit weird, no? Like, it's related to raw EVM data, not decoded data, _and_ light mode touches it. Seems altogether lower-level.

gnidan (commented):
Looks good! Got some questions :)

gnidan:
> Oh, my question wasn't whether that should be moved out of EVM, but rather whether `evm.begin()` should be moved earlier (to allow this to be moved earlier).

Oh! Yeah, maybe! It seems to be more natural for it to be earlier in the startup.

packages/debugger/test/data/fetch-storage.js
haltman-at: :shrug:

packages/debugger/test/data/fetch-storage.js
haltman-at: :)

packages/debugger/lib/data/actions/index.js
gnidan: Hm, yeah, I see where you're coming from. I just don't think it'll make as much sense to someone trying to make sense of the code, vs. something that says more specifically what it relates to.

Along those lines, I think "storage fetching" might not even be sufficient, because that might lead one to believe it's `eth_getStorageAt`, and so I'd want to discriminate the contrast there as best we can. But "intertransactional storage access" is a mouthful (or maybe not, I kinda like it?)

packages/debugger/lib/session/sagas/index.js
gnidan: Saw that it's the same, it just looked to me like a strange guard with no explanation.

packages/debugger/lib/data/reducers.js
haltman-at: Yeah, it is a bit weird.  You've figured out the reason I put it in data, but it could also go in EVM... and, as you point out, putting it in evm would make the light mode thing less awkward.  I guess I'll go change that.

gnidan: I'd even say that maybe `requestStorage` could be moved to evm altogether... but that would mean you'd have to add arguments to that function, so that data could do the lookups it needs before invoking it. Thoughts?

packages/debugger/lib/data/sagas/index.js
haltman-at: You mean if this throws?  Well, then we throw.  I don't know how we'd recover from an error here!

If your question is, what happens if this throws for the specific reason that `debug_storageRangeAt` isn't supported, rather than for any other reason, then the answer is, that won't happen because if `debug_storageRangeAt` isn't supported we won't end up on this code path!

haltman-at:
> > Oh, my question wasn't whether that should be moved out of EVM, but rather whether `evm.begin()` should be moved earlier (to allow this to be moved earlier).
> 
> Oh! Yeah, maybe! It seems to be more natural for it to be earlier in the startup.

OK, I'll have to see about potentially moving it earlier.  Note that if I put this state in `evm` rather than `data`, then potentially -- and actually offhand I think this is the right thing to do -- this could be *part* of `evm.begin()`, and we wouldn't have to make a separate call for it at all...

packages/debugger/lib/data/sagas/index.js
gnidan: I think I was getting myself confused. `web3.obtainStorage` apparently throws... for some reason I thought I saw that it returned its errors.

packages/debugger/lib/data/reducers.js
haltman-at: It actually wouldn't require adding arguments, all that info is available in the EVM state.  Note that right now `requestCode` is in data as well... do you think that should be moved too?

packages/debugger/lib/session/sagas/index.js
haltman-at: Hm, I'd ask if it's worth adding a comment, but I'm not sure that belongs in this PR...

packages/debugger/lib/data/actions/index.js
gnidan: or maybe like... "precision storage lookup"

I feel like we need to decide what we will be calling this feature, publicly speaking, and go with that as the basis of our naming.

packages/debugger/lib/data/reducers.js
gnidan: Oh yeah! Totally! (evm module has access to web3 module?)

packages/debugger/lib/data/actions/index.js
haltman-at: (After discussing this more with @gnidan, we decided to go with "storage lookup", and to rename `fetchStorage` to `storageLookup`.)

packages/debugger/lib/data/reducers.js
haltman-at: Yup, should be no problem!

packages/debugger/lib/session/sagas/index.js
gnidan: hm, I don't recall my feelings from last night in depth anymore, so I guess I'll just resolve this for now... if this line continues to confuse me after you move the various things around, I'll ask for the comment then :)

haltman-at:
OK, I've made these changes... except that moving the test earlier turned out to be more of a problem than I realized. :-/  It could be done, but it would require more reorganization than the simple moving-code-around that I thought.

As such, I've also opted *not* to move the test into `evm.begin()`, because, eh, the order here is a bit incomplete, y'know?  We may want to revisit this later.

packages/debugger/lib/data/sagas/index.js
```diff
@@ -1074,19 +1074,27 @@ export function* decode(
     let response;
     switch (request.type) {
       case "storage":
-        //the debugger supplies all storage it knows at the beginning.
-        //any storage it does not know is presumed to be zero.
-        //(unlesss indicateUnknown is passed, in which case we use
-        //null as a deliberately invalid response)
-        response = indicateUnknown ? null : ZERO_WORD;
+        if (storageLookup) {
+          response = yield* evm.requestStorage(request.slot);
```
gnidan, line 1078: hey this is slick, ain't it!

packages/debugger/lib/data/sagas/index.js
```diff
@@ -1074,19 +1074,27 @@ export function* decode(
...
+          response = null;
+        } else {
+          //if indicateUnknown isn't set, all unknown storage is assumed
+          //to be zero.
+          response = ZERO_WORD;
+        }
         break;
       case "code":
-        response = yield* requestCode(request.address);
+        response = yield* evm.requestCode(request.address);
```
gnidan, line 1091: like, of course you ask the evm to request storage or code!

packages/debugger/lib/data/selectors/index.js
```diff
@@ -186,6 +186,16 @@ function spoofThisDefinition(contractName, contractId, contractKind) {
...
   state: state => state.data,
 
+  /**
+   * data.application
+   */
+  application: {
+    /**
+     * data.application.storageLookup
+     */
+    storageLookup: createLeaf([evm.application.storageLookup], identity)
```
gnidan, line 196: does this not belong in `data.views`? I don't know how you do things these days, I guess

packages/debugger/lib/session/sagas/index.js
```diff
@@ -216,6 +221,16 @@ function* fetchTx(txHash) {
     //full-mode-only modules
     yield* txlog.begin();
   }
+  try {
+    //finally, set storage visibility.  We do this even in light mode, since
```
gnidan, line 225:
```suggestion
    //finally, enable storage lookup.  We do this even in light mode, since
```

gnidan (approved):
looks good :)

packages/debugger/lib/data/selectors/index.js
haltman-at: Man, I dunno, it's all a big mess.  But this selector is going away in the additional change I'm making anyway. :P

packages/debugger/lib/data/sagas/index.js
haltman-at: :)

packages/debugger/lib/session/sagas/index.js
haltman-at: Oops, thanks!

haltman-at:
Oh, wait, I just realized -- `requestStorage` is not inconsistent with `requestCode`!  It should perform the switch (on whether `indicateUnknown` is true and on whether `storageLookup` is true) itself.

This also means that I should have it check the codex... currently it doesn't because that's never necessary in the contexts where it's called.  But if I'm making it a more abstract unit, then, well, that'd be a good thing, y'know...?  So it doesn't trip people up later.

This is a little awkward due to the need for an `indicateUnknown` argument, but oh well.

haltman-at:
OK after talking to @gnidan we decided to eliminate the `indicateUnknown` option; now it's just always on.  That makes this PR a **breaking change**.  While I was at it, I also made some other changes:

1. What happens if a mapping key decodes to an error?  We didn't have anything to handle this case.  Now we discard it (unless of course it's an error that gets handled by `cleanBool`).  This was already a problem, I just noticed it now because of the increased (but still probably low?) probability of it coming up.

2. That special case in `requestCode` for handling the zero address has been unnecessary for quite some time now.  That was there because we used to not have a separation between `instances` and `affectedInstances`; but now these are separate, and also the zero address is now always in the codex.  So this is gone.

gnidan (approved):
alright, nice!

PR #5287, opened 2022-07-13
Enhancement: Add support for XinFin Apothem testnet to sourcify fetcher

haltman-at:
I guess I was at least partly wrong about Sourcify going to a monthly schedule, because they just added this network.  As with other testnets where they only support the testnet but not the mainnet, I listed both the mainnet and the testnet in `networks.ts` even though only the testnet got added to `sourcify.ts`.


cds-amal (approved):
LGTM

PR #5291, opened 2022-07-13
Internal improvement: Up timeout on decoder test

haltman-at:
This test has been timing out a bunch recently for whatever reason, so I'm turning up the timeout to 4 seconds.  Hopefully that's enough!


cds-amal (approved):
LGTM, pending tests ofc! Thanks, @haltman-at.

PR #5307, opened 2022-07-15
Dep update: bump ensjs to 2.1.0 in encoder

eggplantzzz:
When bumping ensjs I forgot to bump it in 1 package!

### DIFF
--- a/packages/encoder/package.json
+++ b/packages/encoder/package.json
@@ -25,7 +25,7 @@
   },
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@ensdomains/ensjs": "^2.0.0",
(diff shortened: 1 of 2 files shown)


haltman-at (approved):

cliffoo:
woawoawoa don't merge yet, release in progress

haltman-at:
Hm I think this is hitting a yarn hoisting problem?

eggplantzzz:
Yeah, I had to update a test in the previous PR for updating ensjs. Since I updated ensjs in all packages it now is putting the solc in the original location again and so the test didn't need to be updated (if I had updated ensjs in all 3 packages instead of only the 2).

cds-amal (approved):
LGTM

PR #5324, opened 2022-07-19
Enhancement: Include raw event data in txlog

haltman-at:
This one is pretty straightforward. It adds a `raw` field to `event` actions in the txlog, so there will be info even if decoding failed.  The `raw` field has four subfields: `address`, `codeAddress`, `topics`, and `data`.  The `topics` and `data` fields are what you expect.  The `address` is the address that officially omitted the event, and `codeAddress` is the address of the *code* that emitted the event.  These can be different if the event was emitted from a delegatecall (e.g. a library), as officially it's the caller of the library that emits the event, but I thought it was good to include the `codeAddress` as well to tell you what contract's *code* emitted the event.

Note that if an event is emitted from a constructor, both `address` and `codeAddress` will be equal to the address that the contract was being created at (barring some cases where we can't determine it, in which case they'll show as zero instead, sorry -- although this will only ever happen for reverted events).

I also expanded the existing tests to test this new functionality.

(Actually, wait, thought, now that I've done this -- should `address` and `codeAddress` be dropped?  They're derivable from the rest of the txlog, which seems to go against how txlog works in general; it generally doesn't include derivable info like that.  Like if you think about the original purpose of `txlog` as a visualization tool, this address info wouldn't be used, because it would already be implicit from the diagram.  Should I remove those?)

### DIFF
--- a/packages/debugger/lib/txlog/actions/index.js
+++ b/packages/debugger/lib/txlog/actions/index.js
@@ -181,12 +181,13 @@ export function identifyFunctionCall(
 }
 
 export const LOG_EVENT = "TXLOG_LOG_EVENT";
-export function logEvent(pointer, newPointer, decoding) {
+export function logEvent(pointer, newPointer, decoding, rawEventInfo) {
   return {
     type: LOG_EVENT,
     pointer,
     newPointer, //does not actually affect current pointer!
-    decoding
+    decoding,
+    rawEventInfo
   };
 }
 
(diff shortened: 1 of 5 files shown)


gnidan:
@haltman-at I concur... we should remove the redundancy and replace it with an easy-to-use selector

haltman-at:
Cool. Should the `flattedEvents` selector be in this PR, or be a separate one?

haltman-at:
Cool, that's what I had intended anyway. :)

PR #5329, opened 2022-07-19
Fix console for test and migrate to work with public testnets and dashboard

sukanyaparashar:
### ISSUE
This issue was introduced by the merged PR #5245. `test` and `migrate` commands are not working for public testnets like `ropsten` or `goerli` with `truffle console --network <testnet or dashboard>`, which requires a `provider` property to connect to the testnet specified. This is because [the `provider` property of the network config is set here](https://github.com/trufflesuite/truffle/blob/54d1f3779e5bbf5d476f36a96b317f1cfee46183/packages/core/lib/console-child.js#L54), which doesn't have a check for the `provider` property's existence in a network config.

### SOLUTION
This PR adds a condition to check if `provider` is specified in a given network config and uses that `provider` to connect to the public testnet (`ropsten`, `goerli`...etc). It also checks if the network specified is `dashboard`.


cds-amal (commented):
[Truffle docs](https://trufflesuite.com/docs/truffle/reference/configuration/#networks) show that there can be other optional values for a network. Should those be preserved, like the provider?

sukanyaparashar:
> [Truffle docs](https://trufflesuite.com/docs/truffle/reference/configuration/#networks) show that there can be other optional values for a network. Should those be preserved, like the provider?

@amal, thanks for the comment. I guess we will need to have special consideration for the `dashboard` stuff since the rpc endpoint for `dashboard` is by default `http://localhost:24012/rpc`.

packages/core/lib/console-child.js
```diff
@@ -23,41 +24,71 @@ global.crypto = {
...
+    provider: function () {
+      return new Web3.providers.HttpProvider(url, {
+        keepAlive: false
+      });
+    }
   };
 } else {
   // Otherwise derive network settings
   const customConfig = detectedConfig.networks[network] || {};
+  const isDashboardNetwork = network === "dashboard" ? true : false;
```
cds-amal, line 65:
```suggestion
  const isDashboardNetwork = network === "dashboard";
```
The result of the comparison is a Boolean

packages/core/lib/console-child.js
```diff
@@ -23,41 +24,71 @@ global.crypto = {
...
   };
-  return `http://${configuredNetworkOptions.host}:${configuredNetworkOptions.port}/`;
+  const urlSuffix = isDashboardNetwork ? "/rpc" : "";
+  return `http://${configuredNetworkOptions.host}:${configuredNetworkOptions.port}${urlSuffix}`;
 }
 
 let configuredNetwork;
 
-if (url) {
+const configNetworkWithProvider =
```
cds-amal, line 41: Is `configDefinesProvider` clearer than `configNetworkWithProvider`?

packages/core/lib/console-child.js
```diff
@@ -23,41 +24,71 @@ global.crypto = {
...
+  detectedConfig.networks[network] && detectedConfig.networks[network].provider;
+
+if (configNetworkWithProvider) {
+  // Use "provider" specified in the network config
```
cds-amal, line 45: Should this be updated to indicate other configured network properties are used as well?

cds-amal (changes requested):
Thanks, @sukanyaparashar, I left some minor fedback. Can you also add a couple of unit tests for this logic?

packages/core/lib/console-child.js
sukanyaparashar: Your suggestion makes sense. `configDefinesProvider` sounds like a better alternative.

packages/core/lib/console-child.js
sukanyaparashar: Yes, this comment needs to be updated I guess, to indicate spreading out of the other specified network properties. Thanks for this @cds-amal.

sukanyaparashar:
> Thanks, @sukanyaparashar, I left some minor fedback. Can you also add a couple of unit tests for this logic?

I like your idea of adding tests for this logic, @amal. Do we need to add tests for the `dashboard` stuff too?

packages/core/lib/console-child.js
```diff
@@ -23,41 +24,73 @@ global.crypto = {
...
+
+if (configDefinesProvider) {
+  // Use "provider" specified in the config to connect to the network
+  // along with the other specified network properties
+  configuredNetwork = {
+    ...detectedConfig.networks[network],
+    network_id: "*",
+    provider: detectedConfig.networks[network].provider
+  };
+} else if (url) {
```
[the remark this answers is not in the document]
sukanyaparashar: Yeah! It should be!

packages/core/lib/command-utils.js
```diff
@@ -206,9 +212,85 @@ const displayGeneralHelp = () => {
     .showHelp();
 };
 
+const getConfiguredNetworkUrl = function (customConfig, isDashboardNetwork) {
```
eggplantzzz, line 215: Would it be slightly more flexible if the second argument were more general, like just `urlSuffix`? It probably does't matter too much in this case though.

packages/core/lib/command-utils.js
```diff
@@ -206,9 +212,85 @@ const displayGeneralHelp = () => {
...
+    const defaultPort = isDashboardNetwork
+      ? managedDashboardDefaultPort
+      : managedGanacheDefaultPort;
+    const defaultNetworkId = isDashboardNetwork
+      ? "*"
+      : managedGanacheDefaultNetworkId;
+
+    configuredNetwork = {
+      // customConfig will spread only when it is defined and ignored when undefined
+      ...customConfig,
```
eggplantzzz, line 270: Do we not want to spread after adding the defaults? That way user settings override the default ones.

packages/core/lib/command-utils.js
```diff
@@ -206,9 +212,85 @@ const displayGeneralHelp = () => {
...
+      host: customConfig.host || defaultHost,
+      port: customConfig.port || defaultPort,
+      network_id: customConfig.network_id || defaultNetworkId,
+      provider: function () {
```
eggplantzzz, line 274: I'm not sure it makes sense to specify both `host` and `url` along with `provider`. Shouldn't we only specify one or the other?

Also, if we spread the user's settings in here there may be a situation where we end up with both. In that case, what happens? Does one take precedence over the other?

packages/core/lib/console-child.js
```diff
@@ -23,60 +18,33 @@ global.crypto = {
...
-    // Perform error handling ourselves.
-    if (error instanceof TruffleError) {
-      console.log(error.message);
-    } else {
-      // Bubble up all other unexpected errors.
-      console.log(error.stack || error.toString());
-    }
-    process.exit(1);
-  });
+main();
```
eggplantzzz, line 50: Why did you need to put all of the above to be executed in a function here at the end?

packages/core/test/sources/console-child/truffle-config.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+  networks: {
+    funTimeNetwork: {
+      host: "127.0.0.1",
+      port: 5555,
+      network_id: "*",
+      gas: 5500000,
+      confirmations: 2
+    },
+    crazyTimeNetwork: {
```
eggplantzzz, line 10: What a crrraaaaazy network!

packages/core/lib/command-utils.js
sukanyaparashar: I think your suggestion makes sense @eggplantzzz. We can spread the `customConfig` after the defaults.

packages/core/lib/command-utils.js
sukanyaparashar: In this case,`provider` takes precedence over the other. Because our logic says that if there is a `provider` specified in the config then we use the `provider` to connect to the network and ignore `host` and `port` (The first `if` block of the main function will be executed). Does that make sense?

packages/core/lib/console-child.js
sukanyaparashar: Oh! This is just for sometime in the future if we want this function to be exported for some reason.

packages/core/test/sources/console-child/truffle-config.js
sukanyaparashar: I guess it is not necessary. I will remove the `gas` property. Thanks @eggplantzzz.

packages/core/lib/console-child.js
cds-amal: IIRC, this came about from refactoring the module to make unit testing easier. Why would it be easier? It's because this module is executed as a child-process and those tests are more complicated; factoring the logic to a different module simplifies writing unit tests. 

It could be extended to use a `require.main === module ? main() : throw new Error("console-child should not be imported")` check to prevent incorrect import, but I don't think it's necessary.

packages/core/lib/command-utils.js
eggplantzzz: Yes it does! But that seems to mean that we only want to add a provider here in the case that there is one since `host` and `port` will be ignored **every time** since a provider is always specified here, right?

sukanyaparashar: That makes sense to me! We can remove the `host` and `port` settings as the provider will always be here. Thanks for catching this @eggplantzzz.

packages/core/lib/command-utils.js
```diff
@@ -206,9 +212,80 @@ const displayGeneralHelp = () => {
...
+
+  const configDefinesProvider =
+    detectedConfig.networks[network] &&
+    detectedConfig.networks[network].provider;
+
+  if (configDefinesProvider) {
+    // Use "provider" specified in the config to connect to the network
+    // along with the other specified network properties
+    configuredNetwork = {
+      ...detectedConfig.networks[network],
```
cds-amal, line 238: This line is not tested in the test, or did I miss it?

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,68 @@
...
+    );
+  });
+
+  describe("deriveConfigEnvironment", function () {
+    it("network has provider", function () {
+      config.network = "crazyTimeNetwork";
+      const cfg = deriveConfigEnvironment(config, config.network, "");
+      const expectedNetworkConfig = config.networks.crazyTimeNetwork;
+      assert.equal(cfg.network, config.network);
+      assert.equal(cfg.networks.crazyTimeNetwork, expectedNetworkConfig);
```
sukanyaparashar, line 37: Does this line satisfies your query @cds-amal ? If the `crazyTimeNetwork` has other properties, then it should be included in the `cfg.networks.crazyTimeNetwork`. Please correct me if I am wrong somewhere :)

sukanyaparashar: Hey @amal. Please ignore the above comment. I realised that there were some mistakes in the unit tests. I have updated the logic for each of the use cases.

sukanyaparashar: Hey @cds-amal. I realised that there were a few mistakes in the unit tests. I have updated the logic. Please ignore my above comment.

packages/core/lib/command-utils.js
```diff
@@ -233,11 +233,11 @@ const deriveConfigEnvironment = function (detectedConfig, network, url) {
 
   if (configDefinesProvider) {
     // Use "provider" specified in the config to connect to the network
-    // along with the other specified network properties
+    // along with the other network properties
     configuredNetwork = {
-      ...detectedConfig.networks[network],
       network_id: "*",
-      provider: detectedConfig.networks[network].provider
+      provider: detectedConfig.networks[network].provider,
```
cds-amal, line 239: Is this necessary if the entire network object, which should include the provider property, is spread out?

cds-amal (commented):
The tests look better! I will have to take a look tomorrow.

packages/core/lib/command-utils.js
sukanyaparashar: Yeah! This is unnecessary as the network object will include the `provider` specified. I will remove this line. Thanks @cds-amal.

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+
+describe("console-child", function () {
+  before("Create a sandbox", function () {
+    config = createSandbox(
+      path.join(__dirname, "..", "sources", "console-child")
+    );
+  });
+
+  describe("deriveConfigEnvironment", function () {
+    it("network has provider", function () {
```
eggplantzzz, line 32: These strings should read as plain english descriptions of what these things do. An example is something like
```
it("returns a config with...<whatever you expect>...
```

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+  fs.copySync(source, tempDir.name);
+  const config = TruffleConfig.load(
+    path.join(tempDir.name, "truffle-config.js"),
+    {}
+  );
+  return config;
+}
+
+describe("console-child", function () {
+  before("Create a sandbox", function () {
```
eggplantzzz, line 25: Get rid of "Create a sandbox", it is obvious from `createSandbox` that that is what's happening :)

Just omit the string like `before(function () {...`

packages/core/test/sources/console-child/truffle-config.js
eggplantzzz: Oh, I guess that stuff would be useful if you wanted to test that user-specified parameters end up in the return for this function.

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+    config = createSandbox(
+      path.join(__dirname, "..", "sources", "console-child")
+    );
+  });
+
+  describe("deriveConfigEnvironment", function () {
+    it("network has provider", function () {
+      config.network = "crazyTimeNetwork";
+      const expectedNetworkConfig = config.networks[config.network];
+      const cfg = deriveConfigEnvironment(config, config.network, "");
```
eggplantzzz, line 35: Here you should probably just omit the third argument or explicitly pass `undefined` as I think that will most closely model the situation where `url` is not present. In other words, if no `url` is present, an empty string will not be passed but rather `undefined`.

Also a small stylistic nit; I would kind of rather specify the string literal `"crazyTimeNetwork"` as the second argument and not use `config.network`. I think it is slightly clearer and I'd rather be clearer than use another value that just happens to work.

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+        cfg.networks[config.network].confirmations,
+        expectedNetworkConfig.confirmations
+      );
+    });
+
+    it("url is specified instead of network", function () {
+      config.network = "localhost:5555";
+      const cfg = deriveConfigEnvironment(
+        config,
+        config.network,
```
eggplantzzz, line 47: Same here about using a string literal for the network name.

packages/core/test/sources/console-child/truffle-config.js
cds-amal: ah, this is related to my comment as well. You could stick a `customUserProperty` in the network config and assert it shows up in the verification step.

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+      );
+      const expectedNetworkConfig = config.networks[config.network];
+      assert.equal(cfg.network, config.network);
+      assert.equal(cfg.networks[config.network], expectedNetworkConfig);
+    });
+
+    it("network has host and port", function () {
+      config.network = "funTimeNetwork";
+      const expectedNetworkConfig = config.networks[config.network];
+      const cfg = deriveConfigEnvironment(config, config.network, "");
```
eggplantzzz, line 58: Same here about the string literal and `undefined` for the `url` parameter.

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+      assert.equal(cfg.network, config.network);
+      assert.equal(
+        cfg.networks[config.network].confirmations,
+        expectedNetworkConfig.confirmations
+      );
+    });
+
+    it("url is specified instead of network", function () {
+      config.network = "localhost:5555";
```
eggplantzzz, line 44: I kind of feel like you should change this network name so it doesn't match the url provivded as a third argument. It seems it is likely to confuse people in that they might think there is some relationship between the third argument provided and this network name when, in fact, they are totally unrelated.

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+describe("console-child", function () {
+  before("Create a sandbox", function () {
+    config = createSandbox(
+      path.join(__dirname, "..", "sources", "console-child")
+    );
+  });
+
+  describe("deriveConfigEnvironment", function () {
+    it("network has provider", function () {
+      config.network = "crazyTimeNetwork";
```
eggplantzzz, line 33: In all these cases I would prefer that you not mutate `config` and instead use string literals as arguments as I've mentioned in a couple of places.

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+      const expectedNetworkConfig = config.networks[config.network];
+      assert.equal(cfg.network, config.network);
+      assert.equal(cfg.networks[config.network], expectedNetworkConfig);
+    });
+
+    it("network has host and port", function () {
+      config.network = "funTimeNetwork";
+      const expectedNetworkConfig = config.networks[config.network];
+      const cfg = deriveConfigEnvironment(config, config.network, "");
+      assert.equal(cfg.network, config.network);
```
eggplantzzz, line 59: Just use a string literal instead of `config.network`.

eggplantzzz:
In the tests for this, I think `config` should just remain an unmutated object. Just use the config as input to the function that you are testing and instead of setting `config.network` to something different for each test (this is unnecessary and confusing I think), just use a string with the network name.

packages/core/test/sources/console-child/truffle-config.js
```diff
@@ -0,0 +1,19 @@
...
+  networks: {
+    funTimeNetwork: {
+      host: "127.0.0.1",
+      port: 5555,
+      network_id: "*",
+      confirmations: 2
+    },
+    crazyTimeNetwork: {
+      network_id: "*",
+      provider: () => "http://localhost:5555",
```
eggplantzzz, line 11: This should be a provider if it is set as `provider` and not a string.

packages/core/test/lib/console-child.js
```diff
@@ -0,0 +1,86 @@
...
+        expectedNetworkConfig.port
+      );
+      assert.equal(
+        cfg.networks[config.network].confirmations,
+        expectedNetworkConfig.confirmations
+      );
+    });
+
+    it("network is dashboard", function () {
+      config.network = "dashboard";
```
eggplantzzz, line 75: Same with this test with just using the network name and not setting `config.network`.

packages/core/test/lib/console-child.js
sukanyaparashar: I will change the strings according to your suggestions @eggplantzzz. Thanks !

packages/core/test/sources/console-child/truffle-config.js
sukanyaparashar: Yeah! I agree with you @cds-amal. I will add a `customUserProperty` in the `network` config. Thanks for your suggestions @cds-amal and @eggplantzzz.

packages/core/test/lib/console-child.js
sukanyaparashar: When we specify the `--url` option with `truffle console` command, then the `network` name happens to be the `host` of the url. It is the logic that goes behind `mergeConfigNetwork.js` inside the `core/lib` package.

eggplantzzz: Yeah I understand, but the network name in this case is arbitrary and has no bearing on the results (at least in this function). In other words, you could name it `"sukanyaNetwork"` and this test will correctly pass. I'm suggesting you change it to something like that so people won't think it has something to do with the `url` parameter.

sukanyaparashar:
> In the tests for this, I think `config` should just remain an unmutated object. Just use the config as input to the function that you are testing and instead of setting `config.network` to something different for each test (this is unnecessary and confusing I think), just use a string with the network name.

Okay! That makes sense to me. Thanks @eggplantzzz.

packages/core/test/lib/console-child.js
sukanyaparashar: I got your point @eggplantzzz. The network name doesn't matter here much. I will apply your suggestions :)

packages/core/test/lib/command-utils.js
```diff
@@ -0,0 +1,93 @@
...
+        cfg.networks.crazyTimeNetwork.customUserProperty,
+        expectedNetworkConfig.customUserProperty
+      );
+    });
+
+    it("returns a config with a url host network object having the url property ", function () {
+      const cfg = deriveConfigEnvironment(
+        config,
+        "anyTimeNetwork",
+        "http://localhost:5555"
```
cds-amal, line 53: This value is used 2x, it can be factored out as a constant.

cds-amal (changes requested):
Thanks, @sukanyaparashar! Aren't unit tests great? I only have a request to document a few functions, and the other comments are stylistic and can be ignored.

packages/core/lib/console-child.js
```diff
@@ -1,82 +1,49 @@
...
-const { network, config, url } = yargs(input[0]).argv;
-const detectedConfig = Config.detect({ network, config });
-
-function getConfiguredNetworkUrl(customConfig) {
-  const configuredNetworkOptions = {
-    host: customConfig.host || managedGanacheDefaultHost,
-    port: customConfig.port || managedGanacheDefaultPort
-  };
-  return `http://${configuredNetworkOptions.host}:${configuredNetworkOptions.port}/`;
+function inheritConfig() {
```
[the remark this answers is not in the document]
sukanyaparashar: I guess we can go with the name `deriveConfig` then?

packages/core/test/lib/command-utils.js
sukanyaparashar: Yeah! I missed that. Thanks @cds-amal.

sukanyaparashar:
> Thanks, @sukanyaparashar! Aren't unit tests great? I only have a request to document a few functions, and the other comments are stylistic and can be ignored.

I agree with you @cds-amal. Unit tests are really great ! :)

packages/core/lib/console-child.js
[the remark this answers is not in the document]
cds-amal: This was a stylistic comment, feel free to ignore it.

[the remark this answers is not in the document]
sukanyaparashar: I thought `deriveConfig` will be a better name, so I changed it in my recent commit. :)

eggplantzzz (approved):
Looks great! I guess @cds-amal has something else to be addressed? I am good with this.

sukanyaparashar:
> Looks great! I guess @cds-amal has something else to be addressed? I am good with this.

Thanks for the approval @eggplantzzz. I will wait for @cds-amal's comment and approval as well. :)

cds-amal (approved):
Looks good!

PR #5330, opened 2022-07-20
Implement encoder/decoder-based `truffle call` command

gnidan:
This draft PR adds a `truffle call` command leveraging Truffle's encoding and decoding functionality (including overload resolution :)
Screen Shot 2022-07-19 at 9 12 26 PM

### DIFF
--- a/packages/core/lib/commands/call/meta.js
+++ b/packages/core/lib/commands/call/meta.js
@@ -21,7 +21,7 @@ module.exports = {
   },
   help: {
     usage:
-      "truffle call [<contract-address>] <contract-name> <function-name>|<function-signature>\n" +
+      "truffle call <contract-address>|<contract-name> <function-name>|<function-signature>\n" +
       "                             " + // spacing to align with previous line
       "[<args...>] [--fetch-external|-x] [--network <network>|--url <provider_url>]\n" +
       "                             " + // spacing to align with previous line
@@ -36,10 +36,11 @@ module.exports = {
         description:
           "The address of the contract to be called.\n" +
           "                    Mandatory if --fetch-external is passed."
-      },  
+      },
       {
         option: "<function-signature>",
-        description: "The function signature inside the specified contract to be called."
+        description:
+          "The function signature inside the specified contract to be called."
       },
       {
         option: "<args...>",
--- a/packages/core/lib/commands/call/meta.js
+++ b/packages/core/lib/commands/call/meta.js
@@ -7,30 +7,43 @@ module.exports = {
       type: "string"
     },
     "fetch-external": {
-      describe: "Allow debugging of external contracts",
+      describe: "Allow calling functions of external contracts",
       alias: "x",
       type: "boolean",
       default: false
+    },
+    "block-number": {
+      describe: "Allows calling functions of a contract from a specific block",
+      alias: "b",
+      type: "string",
+      default: "latest"
     }
   },
   help: {
     usage:
-      "truffle call [<contract-name>] [<function-name>] [<args...>]\n" +
+      "truffle call <contract-name>|<contract-address> <function-name>|<function-signature>\n" +
       "                             " + // spacing to align with previous line
-      "[--fetch-external|-x] [--network <network>|--url <provider_url>]\n" +
-      "                             ",
+      "[<args...>] [--fetch-external|-x] [--network <network>|--url <provider_url>]\n" +
+      "                             " + // spacing to align with previous line
+      "[--block-number|-b <block_number>]",
     options: [
       {
         option: "<contract-name>",
-        description: "The specified contract"
+        description: "The specified contract to be called."
+      },
+      {
+        option: "<contract-address>",
+        description:
+          "The address of the contract to be called.\n" +
+          "                    Mandatory if --fetch-external is passed."
       },
       {
         option: "<function-name>",
-        description: "The function inside the specified contract to be called"
+        description: "The function inside the specified contract to be called."
       },
       {
         option: "<args...>",
-        description: "List of arguments of the function to be called"
+        description: "List of arguments of the function to be called."
       },
       {
         option: "--fetch-external|-x",
@@ -47,6 +60,10 @@ module.exports = {
         option: "--network",
         description:
           "The network to connect to, as specified in the Truffle config."
+      },
+      {
+        option: "--block-number|-b",
+        description: "The block number from which the contract is to be called."
       }
     ],
     allowedGlobalOptions: ["from", "config"]
--- a/packages/core/lib/commands/call/meta.js
+++ b/packages/core/lib/commands/call/meta.js
@@ -33,9 +33,7 @@ module.exports = {
       },
       {
         option: "<contract-address>",
-        description:
-          "The address of the contract to be called.\n" +
-          "                    Mandatory if --fetch-external is passed."
+        description: "The address of the contract to be called.\n"
       },
       {
         option: "<function-name>",
--- a/packages/core/lib/commands/call/meta.js
+++ b/packages/core/lib/commands/call/meta.js
@@ -1,15 +1,54 @@
 module.exports = {
   command: "call",
   description: "Call read-only contract function with arguments",
-  builder: {},
+  builder: {
+    "url": {
+      describe: "Use specified URL for provider",
+      type: "string"
+    },
+    "fetch-external": {
+      describe: "Allow debugging of external contracts",
+      alias: "x",
+      type: "boolean",
+      default: false
+    }
+  },
   help: {
     usage:
-      "truffle call <contract-name> <function-name> [<function-args...>] [--network=<network-name>",
-    usage:
-      "truffle call <contract-name> <function-name> [<args...>]\n" +
-      "                                " + // spacing to align with previous line
-      "[--network=<network-name>] [--from=<account>]",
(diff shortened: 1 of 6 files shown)


packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const contractNames = fs
+    .readdirSync(config.contracts_build_directory)
+    .filter(filename => filename.endsWith(".json"))
+    .map(filename => filename.slice(0, -".json".length));
+  const contracts = contractNames
+    .map(contractName => ({
+      [contractName]: config.resolver.require(contractName)
+    }))
+    .reduce((a, b) => ({ ...a, ...b }), {});
```
gnidan, line 24: I really hope this isn't the best way to list all artifacts...

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const settings = {
+    provider: config.provider,
+    projectInfo: {
+      artifacts: Object.values(contracts)
```
gnidan, line 29: Technically these are getting passed as contract abstraction objects, not as plain JSON artifacts, but it's probably fine.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+    .readdirSync(config.contracts_build_directory)
+    .filter(filename => filename.endsWith(".json"))
+    .map(filename => filename.slice(0, -".json".length));
+  const contracts = contractNames
+    .map(contractName => ({
+      [contractName]: config.resolver.require(contractName)
+    }))
+    .reduce((a, b) => ({ ...a, ...b }), {});
+
+  const settings = {
```
gnidan, line 26: Will these settings differ for @truffle/encoder and @truffle/decoder? cc @haltman-at

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+    provider: config.provider,
+    projectInfo: {
+      artifacts: Object.values(contracts)
+    }
+  };
+
+  const contract = contracts[contractName];
```
gnidan, line 33: This needs handling for failure conditions - e.g. we might want to remind the user to run `truffle compile` first

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+  const instance = await contract.deployed();
```
gnidan, line 34: Similarly, this needs error-handling in case the user didn't run `truffle migrate` or what-have-you.

packages/core/lib/commands/call/meta.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  command: "call",
+  description: "Call read-only contract function with arguments",
+  builder: {},
+  help: {
+    usage:
+      "truffle call <contract-name> <function-name> [<function-args...>] [--network=<network-name>",
```
gnidan, line 7: Although this just says `<contract-name>`, an initial release of this should include support for `--fetch-external`, and thus you should be allowed to specify either a contract name or an address.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const encoder = await Encoder.forContractInstance(instance, settings);
+
+  const { abi: functionEntry, tx: transaction } =
+    await encoder.encodeTransaction(functionName, args);
+
+  // wrap provider for lazy EIP-1193 compatibility
+  const provider = new Encoder.ProviderAdapter(config.provider);
```
gnidan, line 42: I'd advocate against merging until we figure out what to do about this.

For this quick hacking, I just wanted to do the EIP-1193 standard `provider.request`, so I sorta just... made the ProviderAdapter into an EIP-1193 provider (ish)

If we decide that this behavior is nice, then it probably should be a separate PR first to sort that out. 

(Of course, we could otherwise just use whatever gross syntax `config.provider.send()` gives us :)

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const result = await provider.request({
+    method: "eth_call",
+    params: [transaction]
+  });
+
+  const decoder = await Decoder.forContractInstance(instance, settings);
+
+  const [decoding] = await decoder.decodeReturnValue(functionEntry, result);
```
gnidan, line 51: This will need to handle the cases where there is not exactly 1 decoding returned... obviously error nicely if there's 0, and then maybe give a warning if there's >1?

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+
+  config.logger.log(
+    util.inspect(new Codec.Export.ReturndataDecodingInspector(decoding), {
+      colors: true,
+      depth: null,
+      maxArrayLength: null,
+      breakLength: 79
+    })
+  );
```
gnidan, line 60: We probably don't want to use `ReturndataDecodingInspector` here, since the output should be hand-crafted based on the context of the running command...

So instead of just outputting "Returned values: ...", we should probably use `ResultInspector` for each of the returned values and control how it looks tightly.

gnidan: Note that there's some work here... we want to make sure we can handle situations where there's exactly one return value, vs. more than one, as well as handle situations where return values have names vs. not.

packages/core/lib/global-command-options.js
```diff
@@ -4,6 +4,11 @@ const options = {
     description:
       "Specify the network to use. Network name must exist in the configuration."
   },
+  from: {
+    option: "--from <account>",
+    description:
+      "Specify the account from which to make calls or send transactions."
+  },
```
gnidan, line 11: Does this belong here? Should I just move this to a quick separate PR?

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+  };
+
+  const contract = contracts[contractName];
+  const instance = await contract.deployed();
+
+  const encoder = await Encoder.forContractInstance(instance, settings);
+
+  const { abi: functionEntry, tx: transaction } =
+    await encoder.encodeTransaction(functionName, args);
```
gnidan, line 39: Once the encoder makes its overload determination, this should probably check the `stateMutability` and throw a warning if it's not a readonly function

gnidan (commented):
This was definitely just a quick hack, and I'm probably missing things that will need further consideration, but here's what I knew I was neglecting as I was writing this :)

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,61 @@
...
+
+  // wrap provider for lazy EIP-1193 compatibility
+  const provider = new Encoder.ProviderAdapter(config.provider);
+
+  const result = await provider.request({
+    method: "eth_call",
+    params: [transaction]
```
gnidan, line 46: Oh, oops, I forgot to hook up `--from` here!

Note that this currently just relies on @truffle/encoder producing this parameter comprehensively, and further relies on Ganache for filling in gaps with whatever defaults.

This should probably handle the various options more explicitly, or at least check to make sure this works against non-Ganache clients

gnidan:
Oh, probably should also mention... once there are actual well-understood edge cases to test, this PR should probably also include tests for those :)

packages/core/lib/commands/call/meta.js
gnidan: and for `<function-name>`, this should also allow the raw signature (e.g. `truffle call MyContract 'myFunction(string,uint)' ...`

the reason for this is: consider a contract that defines function `f` with two overloads (`f(string)` and `f(uint256)`). Today, if you pass a numeric string to the encoder, then the encoder's overload resolution system will always favor the latter overload, thus there's no good way currently to explicitly specify to use the string overload.

in addition, this situation probably even warrants a warning of some kind (e.g. "We notice you're calling a weirdly overloaded function... you might want to specify explicitly which overload you want")

packages/core/lib/commands/call/meta.js
```diff
@@ -0,0 +1,15 @@
...
+    usage:
+      "truffle call <contract-name> <function-name> [<args...>]\n" +
+      "                                " + // spacing to align with previous line
+      "[--network=<network-name>] [--from=<account>]",
+    options: [],
+    allowedGlobalOptions: ["from", "network", "config"]
```
gnidan, line 13: oh I just remembered... I also forgot about a blockNumber parameter (default `"latest"`)

gnidan: @haltman-at says "oh so you'll need CLI flags for every damn tx option"

gnidan: we can probably skip access lists tho

packages/core/lib/global-command-options.js
sukanyaparashar: Yeah! @gnidan. I think this can be a separate PR since it is adding a new property for global options, which of course, can later be used by other commands if needed. I will make another PR for adding the `from` property to the global options list.

sukanyaparashar:
Closing this PR as a new one is opened https://github.com/trufflesuite/truffle/pull/5647.

PR #5336, opened 2022-07-21
Internal improvement: Remove hardcoded path to `webpack` in packages/truffle

cliffoo:
Currently the truffle package is built with this command:

`node --max-old-space-size=8192 ./node_modules/webpack/bin/webpack.js --config ./webpack.config.js`

But wait! When lerna realizes that another package in the monorepo has `webpack` as a dependency, that relative path to `webpack.js` won't work, because the dependecy would be moved one level up to the root `node_modules` directory.

Naturally we want this instead:

`webpack --config ./webpack.config.js`

Of course we still want the build process to have enough memory, so:

`NODE_OPTIONS="--max-old-space-size=8192" webpack --config ./webpack.config.js`

That's what this PR changes the build command to!

\* Long branch name to reflect the frustrating way I found out about this issue

### DIFF
--- a/packages/truffle/package.json
+++ b/packages/truffle/package.json
@@ -20,7 +20,7 @@
   "scripts": {
     "analyze": "./scripts/analyze.sh",
     "build": "yarn build:cli",
-    "build:cli": "node --max-old-space-size=8192 ./node_modules/webpack/bin/webpack.js --config ./webpack.config.js",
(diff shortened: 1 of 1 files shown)


benjamincburns (approved):
LGTM provided that checks pass

cliffoo:
I tested setting `--max-old-space-size` using `NODE_OPTIONS` to a very low number and the build failed; I set it to 8 gb and the build succeeded. Therefore the env variable must be effective.

cds-amal (approved):
Nice one @cliffoo. LGTM

PR #5343, opened 2022-07-21
Internal improvement: Edit folder name for browser dist files

eggplantzzz:
It seems to me that if we are going to bundle @truffle/contract for browser use, we should be a little bit clearer. Currently the bundled files are distributed in a folder called "dist". I propose renaming this folder "browser-dist" to be more explicit about the nature of the files contained inside.

### DIFF
--- a/packages/contract/.gitignore
+++ b/packages/contract/.gitignore
@@ -1,4 +1,4 @@
 node_modules
 yarn.lock
 package-lock.json
-dist
+browser-dist
--- a/packages/contract/index.js
+++ b/packages/contract/index.js
@@ -13,7 +13,3 @@ const contract = (json = {}) => {
 contract.version = truffleContractVersion;
 
 module.exports = contract;
-
-if (typeof window !== "undefined") {
-  window.TruffleContract = contract;
-}
--- a/packages/contract/notes.md
+++ b/packages/contract/notes.md
@@ -1,0 +1,26 @@
+## ./dist artifacts
+
+TODO
+
+- [ ] verify these assets
+
+### browserify
+
+```
+total 31M
+  11M Feb 13 13:37 truffle-contract.js                | browserified output
+  978 Feb 13 13:30 truffle-contract.js.LICENSE.txt    | license
+  13M Feb 13 13:37 truffle-contract.js.map            | sourcemap
+ 7.1M Feb 13 13:37 truffle-contract.min.js            | minified / uglified
+```
+
+### webpack
+
+```
+total 55M
+ 15M Feb 13 21:09 truffle-contract.js                   | browserified output
+ 16M Feb 13 21:09 truffle-contract.js.map               | sourcemap for browserified
+6.8M Feb 13 21:09 truffle-contract.min.js               | minified / uglified
+ 978 Feb 13 21:09 truffle-contract.min.js.LICENSE.txt   | license
+ 19M Feb 13 21:09 truffle-contract.min.js.map           | sourcemap for minified
+```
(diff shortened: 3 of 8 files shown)


lsqproduction (approved):
Make sense :-)

cliffoo (approved):
Thanks @eggplantzzz !

wbt:
Well, this looks like another breaking change on a patch-number release, with the build pipeline looking for this in /dist to copy it to where it gets served for the browser no longer able to find the file in that location.  I'm now back in the mode of expecting that sort of thing from Truffle.

gnidan:
> Well, this looks like another breaking change on a patch-number release, with the build pipeline looking for this in /dist to copy it to where it gets served for the browser no longer able to find the file in that location.

The determination we made was that this isn't a breaking change, since the NPM package's `"main"` field still resolves to the same interface. Sorry if this tripped you up... if you feel strongly, we can undo the change (or major bump the package).

> I'm now back in the mode of expecting that sort of thing from Truffle.

Perhaps it's not clear how seriously we take semver! We'll do what we can in the future to communicate our thought processes around the practicals of what is breaking vs. what isn't.

wbt:
Thanks for the response. 
> Sorry if this tripped you up... if you feel strongly, we can undo the change (or major bump the package).

In retrospect and for next time, I would major bump the package, but doing that now will leave this version as the last on the prior major-version number.  I've already updated my code to handle it, but have no idea how many others might also have to adapt.

> Perhaps it's not clear how seriously we take semver! We'll do what we can in the future to communicate our thought processes around the practicals of what is breaking vs. what isn't.

No, it's not clear, because of the number of recent times I've had to make code changes to accommodate breaking changes in Truffle released in patch level updates.  Recent examples that spring immediately to mind include breaking workarounds for [Truffle's types](https://github.com/trufflesuite/truffle/pull/5142#issuecomment-1165843255) and around the [bn.js type-breaking update](https://github.com/trufflesuite/truffle/issues/5260) and now [this one](https://github.com/trufflesuite/truffle/pull/5343#issuecomment-1206517223).  I'm not saying it's easy, but there might need to be some updates in the thought processes around what is and isn't breaking.

gnidan:
> Recent examples that spring immediately to mind include [...]

Oof, that's too many for my comfort. Apologies for the poor oversight on this matter when it comes to our typings (and now this).

As I hope you understand (in the abstract), there's often a judgment call. Since Truffle has more than its fair share of legacy code/requirements, there's occasionally an opportunity to make a meaningful impact for our users at the cost of a technically-breaking change. The biggest example that comes to mind of us doing this knowingly is when we removed the default `truffle preserve` plugins (in [Truffle v5.5.17](https://github.com/trufflesuite/truffle/releases/tag/v5.5.17))... these default plugins were the source of a large number of support requests, and we believed that no one was even using this feature. As a result, we decided it was safe to release this breakage, and fortunately, so far no one has said anything (probably because you weren't affected by it 😉 🤗).

But this situation here isn't that. This situation here is... well, we just didn't anticipate that this would break anyone's setup. This strikes me as a failure of process - it's one thing to say "the pros outweigh the cons", but it's another to miss something entirely. I'm truly sorry about this, and I am very interested in making sure this doesn't happen again.

If you have any ideas on how to establish/communicate a semver-adherence policy, or if you know of automated detection techniques we haven't considered, maybe those are two avenues to take. I will give this some thought and discuss with the team. Let's continue this in a new Github Discussion thread. We'll plan to open a thread sometime next week, after team discussion, unless you want to start the discussion yourself.

Thank you for this feedback @wbt.

wbt:
More at https://github.com/trufflesuite/truffle/discussions/5417 (not yet automatically linked).

PR #5348, opened 2022-07-22
Update celoscan url

haltman-at:
Celoscan is moving from celoscan.xyz to celoscan.io; this PR updates those URLs.


cds-amal (approved):
Nice catch!

PR #5349, opened 2022-07-23
Internal improvement: Add a comment and update some others

haltman-at:
The first commit of this PR adds an explanatory comment that I noticed was missing while going over this code with @cds-amal.  The second updates the style of a bunch of comments for consistency.  Yeah that second one doesn't actually matter at all (or doesn't anymore now that the debugger docs are dead), but it was bugging me, so why not.

...also, `prettier` made some changes again.  Not sure why it keeps going back and forth on this matter, may have to look into that...

### DIFF
--- a/packages/debug-utils/index.js
+++ b/packages/debug-utils/index.js
@@ -34,59 +34,59 @@ const verbosePanicTable = {
 };
 
 const commandReference = {
-  "o": "step over",
-  "i": "step line / step into",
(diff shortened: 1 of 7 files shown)


cds-amal (commented):
Looks good to me.

cds-amal (approved):
It really does look good to me! Thanks @haltman-at

PR #5359, opened 2022-07-26
Internal improvements: Custom jest resolver for db

cliffoo:
I'm not too sure why this isn't happening on develop, but take a look at [this failed ci run](https://github.com/trufflesuite/truffle/runs/7512994995).

We recently upgraded to jest 28, and it's trying to use an esm version of uuid (a pouchdb dep), things break because of that. So I added a custom jest resolver that forces jest to use the cjs version of uuid.

Others have run into this, here are some related resources:
- [Comment on microsoft/accessibility-insights-web pull#5421](https://github.com/microsoft/accessibility-insights-web/pull/5421#issuecomment-1109168149)
- [Jest's upgrade guide from 27 to 28 calls out uuid](https://jestjs.io/docs/upgrading-to-jest28#packagejson-exports)

### DIFF
--- a/packages/db/package.json
+++ b/packages/db/package.json
@@ -130,6 +130,7 @@
       "<rootDir>/test/**/test/index.(ts|js)",
       "<rootDir>/src/**/test/*\\.(spec|test)\\.(ts|js)",
       "<rootDir>/test/**/test/*\\.(spec|test)\\.(ts|js)"
-    ]
+    ],
+    "resolver": "<rootDir>/test/jest-resolver.js"
   }
 }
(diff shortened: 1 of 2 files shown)


haltman-at (approved):
Yikes!  I have to admit that I don't really understand this, but since it's seemingly the officially-recommended workaround, I guess I'll hit approve on this?

cliffoo:
Thanks @haltman-at!

cds-amal (commented):
Thanks for catching this @cliffoo! We should probably have a check for esm/cjs when evaluating version bumps, since module authors can choose ECMAScript modules now.

cliffoo:
Okay it appears this custom resolver is no longer needed according to [this thread](https://github.com/uuidjs/uuid/pull/616). I'll make a note to test if ci passes when I remove it.

Issue #5297, opened 2022-07-15
Empty error description when deploying a smart contract (Detailed trace)


SmartArray:
## Issue
When deploying a smart contract, sometimes we get a stack trace like
```
Error: 
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/deployer/src/deployment.js:331:1
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Migration._deploy (/usr/local/lib/node_modules/truffle/build/webpack:/packages/migrate/Migration.js:69:1)
    at Migration._load (/usr/local/lib/node_modules/truffle/build/webpack:/packages/migrate/Migration.js:55:1)
    at Migration.run (/usr/local/lib/node_modules/truffle/build/webpack:/packages/migrate/Migration.js:203:1)
    at Object.runMigrations (/usr/local/lib/node_modules/truffle/build/webpack:/packages/migrate/index.js:143:1)
    at Object.runFrom (/usr/local/lib/node_modules/truffle/build/webpack:/packages/migrate/index.js:108:4)
    at Object.runAll (/usr/local/lib/node_modules/truffle/build/webpack:/packages/migrate/index.js:112:4)
    at Object.run (/usr/local/lib/node_modules/truffle/build/webpack:/packages/migrate/index.js:85:1)
    at Object.run (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/testing/Test.js:114:1)
    at startGanacheAndRunTests (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/commands/test/run.js:49:1)
    at Object.module.exports [as run] (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/commands/test/run.js:106:1)
    at runCommand (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/command-utils.js:184:1)
```

This is not very developer friendly and should be changed.

Fortunately, I was able to dig into the webpack-compiled bundles and take a look what's going on.

https://github.com/trufflesuite/truffle/blob/b9ed05e806b5de524cc1c5489cdd28e0fced573f/packages/deployer/src/sanitizeMessage.js#L6-L7

just tries to print the first element of the error array. By utilizing the environment variable ```DEBUG=deploy*```, I got 
```bash
deployer:sanitizeMessage processing the following message - [ undefined, [], [], [], [], [], [], [] ] +0ms
```

[...]

PR #5361, opened 2022-07-26
Bug fix: Ensure deployment error events always return the message

haltman-at:
Addresses #5297.

If `quiet` is set, then errors during deployment will have no message.  This PR fixes this by having the relevant events *return* the message regardless of the `quiet` setting, just only *print* it when `quiet` is false.  (I thought about using shorter messages when `quiet` is turned on, as the full messages can be kind of long sometimes, but this seemed simpler.)

### DIFF
--- a/packages/deployer/test/errors.js
+++ b/packages/deployer/test/errors.js
@@ -75,7 +75,8 @@ describe("Error cases", function () {
           provider
         }
       },
-      network: "test"
+      network: "test",
+      quiet: true //want to test that errors still work when quiet turned on
     });
     await Environment.detect(options);
     deployer = new Deployer(options);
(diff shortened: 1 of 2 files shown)


sukanyaparashar (approved):
Thanks for getting this done @haltman-at . It definitely needed a better error handling. Looks good to me !

cds-amal (commented):
I see there's already an approval, but is it possible to add a test for this?

haltman-at:
Oh, yeah, I could probably add a test in `deployer` for this.

haltman-at:
OK @cds-amal I've updated this PR.  I didn't actually add any new tests!  I just turned on `quiet` in the existing error tests to ensure they passed. :)  (And yes if you turn on `quiet` on develop, the tests fail.)

cds-amal (approved):
Thanks, @haltman-at

PR #5362, opened 2022-07-27
Use latest node types, remove unused typeRoots

benjamincburns:
Also bumps the `conf` version to `v10.1.2` in `@truffle/config` for compatibility with latest node types.


benjamincburns:
Turns out I need a few more changes.

PR #5363, opened 2022-07-27
Bump typescript to 4.7.2 and @types/node to 18.6.1

benjamincburns:
Also includes minimal code & dependency changes necessary to support latest TypeScript and @types/node


kevinweaver (approved):
👍

Issue #5335, opened 2022-07-21
Enhancement: support error chaining in @truffle/error


benjamincburns:
As of node v14.0.0 the `Error` type supports error chaining by passing an [optional second argument of type `{ cause: any }` to the constructor of `Error`](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-error-message), with the value of that `cause` property appearing as the value of the [`cause` property on the resulting `Error` instance](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause).

We should update the constructor of `@truffle/error` to implement the same pattern, and where possible, pass the underlying cause to the super constructor.

Optionally we should also consider polyfilling the Error class to support error chaining for versions of Node prior to v14.0.0.

eggplantzzz:
According to [Node.green](https://node.green/#ES2022-features-Error-cause-property-Error-has-cause) this feature is supported from Node 16 on.

benjamincburns:
> According to [Node.green](https://node.green/#ES2022-features-Error-cause-property-Error-has-cause) this feature is supported from Node 16 on.

Ah you're correct. Since 16.9.0 [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#browser_compatibility). I'm not sure where I got v14 from... 🤔

Regardless, there's a nice polyfill for it in [`core-js`](https://github.com/zloirock/core-js#ecmascript-error), so we should be able to use it on all node versions.

PR #5367, opened 2022-07-28
Error chaining

benjamincburns:
Supports error chaining for subtypes of `TruffleError`. Addresses issue #5335.

Leverages [modifications to the `Error` type to support error chaining](https://tc39.es/ecma262/multipage/fundamental-objects.html#sec-error-message) as of ES2022, with the added benefit that the `stack` property for `TruffleError` and all of its subtypes will also include the message and stack of the full error chain (unfortunately the base `Error` type does not do this by default).

This PR also contains a breaking change for the `@truffle/error` package. Specifically, it changes the module exports such that the `require` line for JavaScript code that depends on this module must change. This is an unfortunate consequence of having converted this module to TypeScript as part of this change.

Before this change, `@truffle/error` was required by JS code with a line like the following:

```javascript
```

Now it must be required in one of the following ways:

```javascript
const { TruffleError } = require("@truffle/error"); // object destructuring
const TruffleError = require("@truffle/error").default; // default export
```


benjamincburns:
Draft status for the moment because I'm being lazy and letting CI run the tests for me.

Could use a couple of unit tests to verify that the chained `stack` property is working correctly (though I did test it manually and it appears to be working correctly).

Also need to actually do the polyfil part for node <= 16.9.0.

I also think that the value added here is sufficient that we should modify all of our existing `Error` subtypes to instead extend `TruffleError` and chain where appropriate. I see 42 custom error types on a search of our codebase for `extends Error`, and I think there are also quite a few places where we simply `throw new Error(...)`. I'll do this in a subsequent PR, however.

benjamincburns:
Note to reviewers: since I added duplicated code in #5423 to load a polyfill in a few different places, this PR should introduce a new _internal_ (aka `private: true` in the `package.json`, or whatever the idiomatic way to prevent publish is these days) moduled called `@truffle/polyfills` or similar. That module should encapsulate any polyfill logic that we need, and have specific exports for each polyfill that we use.

benjamincburns:
This is currently blocked due to there being scenario tests in this repo that are meant to be testing the `@truffle/preserve-*` plugins.

I'll address that in a separate PR and then return back to this one.

benjamincburns:
> Note to reviewers: since I added duplicated code in [#5423 (comment)](https://github.com/trufflesuite/truffle/pull/5423#discussion_r943074306) to load a polyfill in a few different places, this PR should introduce a new _internal_ (aka `private: true` in the `package.json`, or whatever the idiomatic way to prevent publish is these days) modules called `@truffle/polyfills` or similar. That module should encapsulate any polyfill logic that we need, and have specific exports for each polyfill that we use.

I submitted #5492 to handle this

eggplantzzz:
This PR is still relevant.

eggplantzzz:
Closing this for PR maintenance.

PR #5375, opened 2022-07-28
Enhancement: Allow encoder to look up functions by full signature

haltman-at:
Previously, `encodeTransaction` allowed either a list of abi entries as input, or a function name.  I've expanded it to allow function signature as well.  This is pretty straightforward so I don't have much to say about this.  I also added tests of this functionality.

This is intended to support #5330, providing a way of performing manual resolution of overloads on the command line.

I thought about also allowing this for `encodeTxNoResolution`, but decided against it... it'd be a bit awkward.

Note that the error you get if lookup fails is still called a `NoFunctionByThatNameError`; however, the message has been updated.

Also `prettier` made some changes.

### DIFF
--- a/packages/encoder/lib/encoders.ts
+++ b/packages/encoder/lib/encoders.ts
@@ -892,10 +892,10 @@ export class ContractEncoder {
    * select an overload by other means.  For enums you may also specify the
    * enum type as documented in [[encodeTransaction]].
    *
-   * @param abisOrName The ABI entries for the overloads, or the name of the
-   *   function.  Note that if you are inputting ABI entries, they must be
-   *   for functions, not constructors.  The entries must be ones associated
-   *   with this contract.
+   * @param abisOrNameOrSig The ABI entries for the overloads, or the name or
+   *   full signature of the function.  Note that if you are inputting ABI
+   *   entries, they must be for functions, not constructors.  The entries must
+   *   be ones associated with this contract.
    * @param inputs An array of the inputs to the transaction.  May include a
    *   transaction options argument on the end if the `allowOptions` flag is
    *   set.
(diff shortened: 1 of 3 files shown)


cds-amal:
> I thought about also allowing this for encodeTxNoResolution, but decided against it... it'd be a bit awkward.

Can you explain the awkwardness, please?

gnidan (approved):

haltman-at:
> Can you explain the awkwardness, please?

Sure!

So, it helps a bit to know the history here.  The original idea was that there would be `encodeTransaction`, which did not perform resolution, and took a single ABI entry; and `resolveAndEncode`, which did perform resolution, and took a list of them to resolve between.  Now, `encodeTransaction` could take an ABI entry either for a function or for a constructor; but since the purpose of `resolveAndEncode` was to perform resolution on top of encoding, it didn't allow constructor ABIs as input, because there's never any resolution to perform with those.

Then some changes got made.  The `resolveAndEncode` method got renamed to `encodeTransaction`, and `encodeTransaction` got renamed to `encodeTxNoResolution`.  Also, the new `encodeTransaction` got the ability to take a function name in place of a list of ABI entries.  And now it's getting the ability to take a signature.

So what if we did the same with `encodeTxNoResolution`?  Obviously that can't take a function *name* because the whole point is that it *doesn't* perform resolution (notionally it could error out if there are multiple ABI entries for that name but uh I think that's stretching things).  But what about a function *signature*?  Those are unique!  So I could allow it to take that!  But it just seems kind of awkward to me because it's complicating the interface, it's not necessary for `truffle call`, and also like -- you can't encode a constructor that way?  So it's a bit awkward.  I guess constructors are already a bit awkward, what with how you can only do them with `encodeTxNoResolution` rather than `encodeTransaction`.  (Maybe there should be a separate method for them...?)

I dunno.  There are reasons to include it and not to include it.  I don't know if @gnidan has an opinion here.  I guess he's already approved this, so presumably he's OK with this PR as it is!  Ultimately it's not necessary for `truffle call`, so maybe that can just be a later thing if we want to do it.

PR #5376, opened 2022-07-29
Enhancement: Add `allowJson` and `strictBooleans` options to encoder

haltman-at:
This PR adds two new options to the encoder, intended for use with `truffle call` (#5330).  It also adds tests of these and updates the documentation.  (Edit: **This PR now contains breaking changes to `@truffle/codec` and `@truffle/encoder`.**)

The first option, `allowJson`, allows arrays and tuples/structs to be given as JSON strings.  Note that this *only applies* to arrays and tuples, not more elementary types like strings!  So you can pass `'["abc","def"]'` for `["abc","def"]`, but not `'"abc"'` for `"abc"` on its own.  For tuples/structs, the JSON can be either for an array or for an object (since both arrays and objects are allowed as input for these).  The JSON cannot be for some other form of input; e.g., you can't put in JSON for a type/value pair or a codec `Value` or what have you.

Note that the individual elements can, within the JSON, be given in any form allowed; e.g., you can pass `'[1, "2"]'` for an array of two uints, for instance.

Also, note that JSON input must be a `string` and not a `String`; I didn't see any reason to bother with the latter given the intended use case. :P

The second option, `strictBooleans`, makes it so that the only strings that will be accepted as booleans are `"true"` and `"false"` (ignoring case).  I considered also allowing `"0"` and `"1"`, but decided against it for now... @gnidan, let me know if you think that should be changed.  Note that having `"0"` be treated as false would be a breaking change, but I don't think that should stop us if you think it's the right thing.  Turning on `strictBooleans` also makes it so that you can't just use arbitrary input as a boolean like you could previously.

Now, because the catch-all I turned off was previously used to handle all sorts of things, that actually made `strictBooleans` stricter than intended... you couldn't use an actual `boolean` anymore!  So I added a new case to handle that.  It also makes it so that you can't use a `number` either, since those were also handled by the catch-all!  I didn't bother adding a case for that, though, because the intended use-case doesn't require it.  I mean, OK, the intended use-case doesn't require `boolean`s to work as input either, but rejecting those would just be ridiculous, so, yeah.  Again, @gnidan, let me know if you want this changed.

Anyway, this is mostly pretty straightforward -- just adding cases to `wrap` in the existing style.  I think if `truffle call` passes these two options, it should be able to use `encoder` pretty well!

### DIFF
--- a/packages/codec/lib/wrap/bool.ts
+++ b/packages/codec/lib/wrap/bool.ts
@@ -70,18 +70,18 @@ function* boolFromString(
   const lowerCasedInput = input.toLowerCase();
   if (
     wrapOptions.strictBooleans &&
-    !["true", "false"].includes(lowerCasedInput)
+    !["true", "false", "1", "0"].includes(lowerCasedInput)
   ) {
     throw new TypeMismatchError(
       dataType,
       input,
       wrapOptions.name,
       5,
-      "Input was neither 'true' nor 'false'"
+      "Input was not 'true', 'false', '1', or '0'"
     );
   }
-  //strings are true unless they're falsy or the case-insensitive string "false"
-  const asBoolean = Boolean(input) && input.toLowerCase() !== "false";
+  //strings are true unless they're falsy or the case-insensitive strings "false" or "0"
+  const asBoolean = Boolean(input) && !["false", "0"].includes(lowerCasedInput);
   return {
     type: dataType,
     kind: "value" as const,
--- a/packages/codec/lib/wrap/bool.ts
+++ b/packages/codec/lib/wrap/bool.ts
@@ -14,6 +14,7 @@ const boolCasesBasic: Case<
   Format.Values.BoolValue,
   never
 >[] = [
+  boolFromBoolean, //needed due to strictBooleans mode
   boolFromString,
   boolFromBoxedPrimitive,
   boolFromCodecBoolValue,
@@ -27,10 +28,30 @@ export const boolCases: Case<
   Format.Types.BoolType,
   Format.Values.BoolValue,
   never
->[] = [
-  boolFromTypeValueInput,
-  ...boolCasesBasic
-];
+>[] = [boolFromTypeValueInput, ...boolCasesBasic];
+
+function* boolFromBoolean(
+  dataType: Format.Types.BoolType,
+  input: unknown,
+  wrapOptions: WrapOptions
+): Generator<never, Format.Values.BoolValue, WrapResponse> {
+  if (typeof input !== "boolean") {
+    throw new TypeMismatchError(
+      dataType,
+      input,
+      wrapOptions.name,
+      1,
+      "Input was not a boolean"
+    );
+  }
+  return {
+    type: dataType,
+    kind: "value" as const,
+    value: {
+      asBoolean: input
+    }
+  };
+}
 
 function* boolFromString(
   dataType: Format.Types.BoolType,
@@ -44,6 +65,19 @@ function* boolFromString(
       wrapOptions.name,
       1,
       "Input was not a string"
+    );
+  }
+  const lowerCasedInput = input.toLowerCase();
+  if (
+    wrapOptions.strictBooleans &&
+    !["true", "false"].includes(lowerCasedInput)
+  ) {
+    throw new TypeMismatchError(
+      dataType,
+      input,
+      wrapOptions.name,
+      5,
+      "Input was neither 'true' nor 'false'"
     );
   }
   //strings are true unless they're falsy or the case-insensitive string "false"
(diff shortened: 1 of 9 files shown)


cds-amal:
>The second option, strictBooleans, makes it so that the only strings that will be accepted as booleans are "true" and "false" (ignoring case). I considered also allowing "0" and "1", but decided against it for now... @gnidan, let me know if you think that should be changed. Note that having "0" be treated as false would be a breaking change, but I don't think that should stop us if you think it's the right thing. Turning on strictBooleans also makes it so that you can't just use arbitrary input as a boolean like you could previously.

I'm interested to know about @gnidan's design choices here! 

>Now, because the catch-all I turned off was previously used to handle all sorts of things, that actually made strictBooleans stricter than intended... you couldn't use an actual boolean anymore! So I added a new case to handle that. It also makes it so that you can't use a number either, since those were also handled by the catch-all! I didn't bother adding a case for that, though, because the intended use-case doesn't require it. I mean, OK, the intended use-case doesn't require booleans to work as input either, but rejecting those would just be ridiculous, so, yeah. Again, @gnidan, let me know if you want this changed.

Fun and interesting!

gnidan:
> The second option, `strictBooleans`, makes it so that the only strings that will be accepted as booleans are `"true"` and `"false"` (ignoring case). I considered also allowing `"0"` and `"1"`, but decided against it for now... @gnidan, let me know if you think that should be changed. Note that having `"0"` be treated as false would be a breaking change, but I don't think that should stop us if you think it's the right thing. Turning on `strictBooleans` also makes it so that you can't just use arbitrary input as a boolean like you could previously.

It's not uncommon to use `0`/`1` as booleans in the CLI, right? Maybe for that reason, we want to allow that. @cds-amal does this mesh with your understanding of CLI interface patterns?

gnidan:
> Now, because the catch-all I turned off was previously used to handle all sorts of things, that actually made `strictBooleans` stricter than intended... you couldn't use an actual `boolean` anymore! So I added a new case to handle that. It also makes it so that you can't use a `number` either, since those were also handled by the catch-all! I didn't bother adding a case for that, though, because the intended use-case doesn't require it. I mean, OK, the intended use-case doesn't require `boolean`s to work as input either, but rejecting those would just be ridiculous, so, yeah. Again, @gnidan, let me know if you want this changed.

No I think this is fine.

haltman-at:
> > The second option, `strictBooleans`, makes it so that the only strings that will be accepted as booleans are `"true"` and `"false"` (ignoring case). I considered also allowing `"0"` and `"1"`, but decided against it for now... @gnidan, let me know if you think that should be changed. Note that having `"0"` be treated as false would be a breaking change, but I don't think that should stop us if you think it's the right thing. Turning on `strictBooleans` also makes it so that you can't just use arbitrary input as a boolean like you could previously.
> 
> It's not uncommon to use `0`/`1` as booleans in the CLI, right? Maybe for that reason, we want to allow that. @cds-amal does this mesh with your understanding of CLI interface patterns?

OK.  OK to do this is as a breaking change?  So that `"0"` will be interpreted as `false` more generally?

gnidan:
> > > The second option, `strictBooleans`, makes it so that the only strings that will be accepted as booleans are `"true"` and `"false"` (ignoring case). I considered also allowing `"0"` and `"1"`, but decided against it for now... @gnidan, let me know if you think that should be changed. Note that having `"0"` be treated as false would be a breaking change, but I don't think that should stop us if you think it's the right thing. Turning on `strictBooleans` also makes it so that you can't just use arbitrary input as a boolean like you could previously.
> > 
> > 
> > It's not uncommon to use `0`/`1` as booleans in the CLI, right? Maybe for that reason, we want to allow that. @cds-amal does this mesh with your understanding of CLI interface patterns?
> 
> OK. OK to do this is as a breaking change? So that `"0"` will be interpreted as `false` more generally?

Makes sense!

haltman-at:
OK, I'll go back and change that, then!

haltman-at:
OK, I've made those changes.  **This PR now contains breaking changes to `@truffle/codec` and `@truffle/encoder`!**

cds-amal:
> It's not uncommon to use `0`/`1` as booleans in the CLI, right? Maybe for that reason, we want to allow that. @cds-amal does this mesh with your understanding of CLI interface patterns?

Weeeeeel, it's `0` for success and not `0` for failure where the not `0` range encodes some meaningful error status. [See Bash Exit Status](https://www.gnu.org/software/bash/manual/html_node/Exit-Status.html#:~:text=When%20a%20command%20terminates%20on,the%20return%20status%20is%20126.)

haltman-at:
I mean, we're talking about command line user input, not exit statuses...

gnidan (approved):
lgtm!

haltman-at:
Cool, gonna merge this then!

PR #5380, opened 2022-07-29
Update @truffle/core in truffle and update yarn.lock

davidmurdoch:
There are some issues with Truffle release 5.5.24:

 * @truffle/core's version was not bumped in truffle/package.json devDeps appropriately
 * both truffle develop and master (post release) fail in CI, not getting past yarncheck

I think this PR will fix both of these issues and get things working again.

(but a proper post-mortem should be done)

PR #5384, opened 2022-08-01
Verify versions of @truffle dependencies on commit

benjamincburns:
Adds a script that runs as a pre-commit hook whenever any package's `package.json` is changed. Script checks that all dependencies on packages in the `@truffle` namespace resolve to the local version of the package in question.


benjamincburns:
Note that on 7a9c574 (bad merge commit per #5380), I see the following output:

```
 Package "truffle" depends on "@truffle/core@^5.5.23", but local "@truffle/core" has version 5.5.22, which is not in range ^5.5.23
```

scripts/check-truffle-namespace-dependency-versions.js
```diff
@@ -0,0 +1,136 @@
...
+          `${candidate.name} requires ${name}@${version} (${rawRange} is up-to-date for version ${version})`
+        );
+      }
+    } else {
+      debug(`${candidate.name} does not require ${name}`);
+    }
+  }
+}
+
+if (require.main === module) {
```
eggplantzzz, line 134: Can you explain why this is here?

eggplantzzz:
I think this is cool

scripts/check-truffle-namespace-dependency-versions.js
```diff
@@ -0,0 +1,136 @@
...
+    doneColor("Done."),
+    `${errors.length} errors, ${warnings.length} warnings found for @truffle namespace version dependencies.`
+  );
+
+  if (errors.length > 0) {
+    process.exit(1);
+  }
+}
+
+function getPackageNames() {
```
gnidan, line 59: Should this be named something like `getPackageDirectories`? Since we're not actually reaching into `package.json` to find the `name` field here.

scripts/check-truffle-namespace-dependency-versions.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+function checkDependencyVersion(candidate, dependency) {
+  const name = dependency.name;
+  const version = dependency.version;
+
+  for (const depType of [
+    "dependencies",
+    "devDependencies",
+    "peerDependencies"
```
gnidan, line 85: what about `optionalDependencies`?

scripts/check-truffle-namespace-dependency-versions.js
```diff
@@ -0,0 +1,136 @@
...
+}
+
+function checkDependencyVersionRange(candidate, dependency) {
+  const name = dependency.name;
+  const version = dependency.version;
+
+  for (const depType of [
+    "dependencies",
+    "devDependencies",
+    "peerDependencies"
```
gnidan, line 113: also `optionalDependencies` here

gnidan (commented):
few comments

scripts/check-truffle-namespace-dependency-versions.js
benjamincburns: I'll just move it to a const list so it's not redefined

scripts/check-truffle-namespace-dependency-versions.js
benjamincburns: Just a convention - makes it so that this code can be imported in another module (once exports are added, anyway)

scripts/check-truffle-namespace-dependency-versions.js
gnidan: nice. also probably check `package.json` spec, cause I might be forgetting others. they're always adding things

scripts/check-truffle-namespace-dependency-versions.js
benjamincburns: Apparently there's also a `bundleDependencies`, so I included that as well.

scripts/check-truffle-namespace-dependency-versions.js
benjamincburns: I'd like to say that I checked the spec before GH Copilot filled in `bundleDependencies` for me, but alas Copilot was faster (I did double check the spec, however).

benjamincburns:
Updated it to drop the range check altogether, as we decided on the release postmortem call that we'd error if the caret version range isn't a tight constraint on the version

benjamincburns:
Updated output when ran over commit 7a9c574:

```Error: Package "@truffle/core" depends on "@truffle/artifactor@^4.0.161", but range has not been updated for version 4.0.162
Error: Package "@truffle/core" depends on "@truffle/box@^2.1.52", but range has not been updated for version 2.1.53
Error: Package "@truffle/core" depends on "@truffle/codec@^0.13.2", but range has not been updated for version 0.13.3
Error: Package "@truffle/core" depends on "@truffle/compile-solidity@^6.0.35", but range has not been updated for version 6.0.36
Error: Package "@truffle/core" depends on "@truffle/config@^1.3.32", but range has not been updated for version 1.3.33
Error: Package "@truffle/core" depends on "@truffle/contract@^4.5.17", but range has not been updated for version 4.5.18
Error: Package "@truffle/core" depends on "@truffle/dashboard@^0.1.9", but range has not been updated for version 0.1.10
Error: Package "@truffle/core" depends on "@truffle/debug-utils@^6.0.27", but range has not been updated for version 6.0.28
Error: Package "@truffle/core" depends on "@truffle/debugger@^11.0.0", but range has not been updated for version 11.0.1
Error: Package "@truffle/core" depends on "@truffle/decoder@^5.2.20", but range has not been updated for version 5.2.21
Error: Package "@truffle/core" depends on "@truffle/deployer@^3.2.88", but range has not been updated for version 3.2.89
Error: Package "@truffle/core" depends on "@truffle/environment@^0.2.115", but range has not been updated for version 0.2.116
Error: Package "@truffle/core" depends on "@truffle/expect@^0.1.1", but range has not been updated for version 0.1.2
Error: Package "@truffle/core" depends on "@truffle/fetch-and-compile@^0.5.10", but range has not been updated for version 0.5.11
Error: Package "@truffle/core" depends on "@truffle/migrate@^3.3.2", but range has not been updated for version 3.3.3
Error: Package "@truffle/core" depends on "@truffle/plugins@^0.2.7", but range has not been updated for version 0.2.8
Error: Package "truffle" depends on "@truffle/core@^5.5.23", but range has not been updated for version 5.5.22
Done. Found 17 errors for @truffle namespace version dependencies.```

File: scripts/check-truffle-namespace-dependency-versions.js (hunk 4 above)
Comment (cds-amal) [reply scripts/check-truffle-namespace-dependency-versions.js]: `bundled?Dependencies` could be a boolean.

>If this is spelled "bundledDependencies", then that is also honored.
>
>Alternatively, "bundleDependencies" can be defined as a boolean value. A value of true will bundle all dependencies, a value of false will bundle none.

Comment (benjamincburns) [reply scripts/check-truffle-namespace-dependency-versions.js]: Ah thanks - good catch!

File: scripts/check-truffle-namespace-dependency-versions.js [hunk 5]
```diff
@@ -67,7 +68,7 @@ function* checkDependencyVersionRange(candidate, dependency) {
 
   for (const depType of depTypes) {
     const deps = candidate[depType];
-    if (deps && deps[name]) {
+    if (typeof deps === "object" && deps[name]) {
```
cds-amal, line 71: Note [according to specs](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#bundledependencies), `bundled?Dependencies` is an Array.

scripts/check-truffle-namespace-dependency-versions.js
cds-amal: It's a guard that invokes the `main` function only if this file is being run as a script, i.e. `node check-truffle-namespace-dependency-versions.js`. If another module imported this module, then `main` will not be invoked/executed.

scripts/check-truffle-namespace-dependency-versions.js
benjamincburns: Consistentcy is underrated, eh?

benjamincburns: Turns out it also doesn't include version info, so it's moot in this context (dropped it from the checks).

benjamincburns:
@cds-amal @gnidan @eggplantzzz @haltman-at - I just realized that I've committed one of my biggest software peeves in this change: this script complains about errors that it likely has enough information to just fix for itself.

Is there some reason not to introduce a `--fix` flag and have husky run it that way?

scripts/check-truffle-namespace-dependency-versions.js
```diff
@@ -0,0 +1,87 @@
...
+  "optionalDependencies"
+];
+
+function* checkDependencyVersionRange(candidate, dependency) {
+  const name = dependency.name;
+  const version = dependency.version;
+
+  for (const depType of depTypes) {
+    const deps = candidate[depType];
+    if (deps && Object.keys(deps).includes(name)) {
```
cds-amal, line 69: Not a request to change but I'm curious why you chose `Object.keys` over `deps.hasOwnProperty(name)` or the `in` operator here.

scripts/check-truffle-namespace-dependency-versions.js
```diff
@@ -0,0 +1,87 @@
...
+function getPackageDirectoryNames() {
+  return fs.readdirSync(path.join(__dirname, "..", "packages"));
+}
+
+function readPackages() {
+  const packages = getPackageDirectoryNames();
+  const packageSpecs = {};
+  for (const packageName of packages) {
+    const packageFilePath = path.join(
+      path.resolve(__dirname, "..", "packages", packageName),
```
cds-amal, line 46: The CWD for scripts defaults to the root of the package folder, is there something with how the script is invoked (lerna) that made this necessary?

cds-amal (commented):
I'm so happy this is in @benjamincburns, thank you! I had some questions (not request to change) about some of your choices. The more important thing, though, is what is the implication of [this comment](https://github.com/trufflesuite/truffle/pull/5384#issuecomment-1201882943)? I can't dig into it now, but will look at it first thing in the morning.

scripts/check-truffle-namespace-dependency-versions.js
benjamincburns: Ah, that's a better approach. To answer the question - in the moment I preferred `Object.keys` as it will just return an empty list for any type without enumerable properties, including `null` and `undefined`. But the test for a truthy value of `deps` here makes that concern moot, so I can either drop that test, or convert to `deps.hasOwnProperty(...)`.

scripts/check-truffle-namespace-dependency-versions.js
benjamincburns: No, just prudence on my part. In scripts like these that run over the entire project I tend to prefer to resolve paths relative to the script location, as that makes it possible to invoke the script manually from any CWD and have it affect the repo that it's running within.

cds-amal:
> @cds-amal @gnidan @eggplantzzz @haltman-at - I just realized that I've committed one of my biggest software peeves in this change: this script complains about errors that it likely has enough information to just fix for itself.
> 
> Is there some reason not to introduce a `--fix` flag and have husky run it that way?

This PR starts with the codebase in a bad state, so errors are expected? I think we should leave it without the `--fix` option and if it becomes a future nuisance, then it would be appropriate to consider.

benjamincburns:
> The more important thing, though, is what is the implication of https://github.com/trufflesuite/truffle/pull/5384#issuecomment-1201882943? I can't dig into it now, but will look at it first thing in the morning.

@cds-amal that comment is just to show the output that the script would have given me on the bad merge commit from the TypeScript migrations PR. I just included it as a sanity check to show that the script would have helped to mitigate the problem.

benjamincburns:
> This PR starts with the codebase in a bad state, so errors are expected? I think we should leave it without the `--fix` option and if it becomes a future nuisance, then it would be appropriate to consider.

@cds-amal I fixed the warnings in this PR - see commit fad0098

scripts/check-truffle-namespace-dependency-versions.js
cds-amal: That's your call :)

scripts/check-truffle-namespace-dependency-versions.js
cds-amal: Oh, thanks!

cds-amal (approved):
This is really cool! Definitely an improvement to the codebase. Thanks, @benjamincburns

scripts/check-truffle-namespace-dependency-versions.js
eggplantzzz: Ah ok gotcha

eggplantzzz (approved):

Issue #4261, opened 2021-08-13
@truffle/compile-solidity's CompilerSupplier does too much manual retry logic


gnidan:
The VersionRange loading strategy is most notably guilty of this. It (and any other instances) should be updated to use [axios-retry](https://www.npmjs.com/package/axios-retry), already used by the Docker strategy.

_Originally posted by @haltman-at in https://github.com/trufflesuite/truffle/pull/4222#discussion_r687292057_

gnidan:
@haltman-at can you please summarize? That PR is a bit hard to catch up on

haltman-at:
Basically, the thing there is that when the compiler supplier retries, it's not retrying with the same URL but with a different one.  That likely makes something like `axios-retry` inappropriate, since the point of that is to perform appropriate backoff when retrying against the same server.

eggplantzzz:
So basically the whole flow needs to be retried which entails getting the list of available versions for the source, finding the correct one (if available), and then fetching it.

gnidan:
Is it no longer true that the CompilerSupplier does too much manual retry logic? It sounds like the amount of retry logic is still too much.

Ah but this issue is specifically about moving to axios-retry or some other automated retry mechanism... and I concur that that's probably not feasible, considering the necessary behavior. Bam, I love using the close-button dropdown!

PR #5388, opened 2022-08-03
Bug fix: Fix retry mechanism for fetching the Solidity compiler

eggplantzzz:
Currently Truffle has an issue with how it downloads the Solidity compiler. It currently has 2 sources (technically there are 3 url roots that Truffle uses but the first is a Truffle link that redirects to the second) that it fetches the compiler from. First, it fetches a list of the Solidity compilers available from the first source that successfully returns a list. After it has this list, it extracts the filename for the one it wants and sends it over to a function that makes the request. Here is where there is a bit of a problem.

Once this function receives the filename, it makes the request to the first source. If it fails, then it retries using the 2nd and 3rd url (if the 2nd fails). However, this could be a problem because different sources have different compiler versions available and the urls also may be different (for example, wasm vs. asm builds).

So this PR attempts to put these two steps together so that first a list is downloaded from a source. Once it gets the filename from the list it can request from the same source. If this fails, then it will retry starting with fetching the list from the next source, and so on. This should eliminate cases where there are discrepancies between what sources have available.


cds-amal:
How should this be smoke tested?

eggplantzzz:
If you want to smoke test it then maybe you could put typos in the code for the `compilerRoots` urls to ensure that the later ones get used?

haltman-at:
Btw, may not be worth fixing in this PR, but probably worth taking note of issue #4261 as long as you're looking at this stuff.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -124,90 +141,89 @@ export class VersionRange {
...
-    } catch (error) {
-      throw new NoRequestError(versionConstraint, error);
+  async getSolcFromCacheOrUrl(versionConstraint: string, index: number = 0) {
+    // go through all sources (compilerRoots) trying to locate a
+    // suitable version of the Solidity compiler
+    const { compilerRoots, events } = this.config;
+    if (index > compilerRoots.length - 1) {
+      throw new CompilerFetchingError(compilerRoots);
+    }
+    if (!compilerRoots || compilerRoots.length < 1) {
```
haltman-at, line 184: The `if (!compilerRoots)` check needs to go before the `index >= compilerRoots.length` check; otherwise an absent `compilerRoots` would cause an error when you attempt to get `compilerRoots.length`.  **This is the reason I'm hitting request changes**.

Also, you don't need a separate case for `compilerRoots.length < 1`; that will be handled automatically by the `index >= compilerRoots.length` check.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -52,10 +52,27 @@ export class VersionRange {
...
+        `Failed to fetch the list of Solidity compilers from the following ` +
+          `sources: ${this.config.compilerRoots}. Make sure you are connected ` +
+          `to the internet.`
+      );
+    }
+    let data;
+    try {
+      data = await this.getSolcVersionsForSource(
+        this.config.compilerRoots[index],
+        index + 1
```
haltman-at, line 67: So, I was really confused here why you were adding 1 to the `index` in cases when you weren't retrying.  This had me really confused.  Eventually I realized it was because `getSolcVersionsForSource` only uses that number for printing.

I think this is confusing and it would be better to *not* add 1 to the index here, but rather instead to add it at the last minute, when emitting the event, since it's purely a display concern (like happens on line 163).

There are several places where you do this; I'll mark them for convenience.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -124,90 +141,89 @@ export class VersionRange {
...
-      ? this.findNewestValidVersion(versionConstraint, allVersions)
-      : versionConstraint;
-    const fileName = this.getSolcVersionFileName(versionToUse, allVersions);
 
-    if (!fileName) throw new NoVersionError(versionToUse);
+    let allVersionsForSource: string[], versionToUse: string | null;
+    try {
+      allVersionsForSource = await this.getSolcVersionsForSource(
+        compilerRoots[index],
+        index + 1
```
haltman-at, line 193: Same comment

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -124,90 +141,89 @@ export class VersionRange {
...
 
-    if (this.cache.has(fileName))
-      return this.getCachedSolcByFileName(fileName);
-    return this.getAndCacheSolcByUrl(fileName);
+      if (this.cache.has(fileName)) {
+        return this.getCachedSolcByFileName(fileName);
+      }
+      return await this.getAndCacheSolcByUrl(fileName, index);
+    } catch (error) {
+      return await this.getSolcFromCacheOrUrl(versionConstraint, index + 1);
```
haltman-at, line 210: Same comment

haltman-at (changes requested):
Thanks Tyler!  However I noticed a bug in `getSolcFromCacheOrUrl`, so I'm hitting request changes.  I also have some comments on the style of the rest, but that's less important, obviously!

(Also see my comment above about #4261. :P )

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -80,7 +97,7 @@ export class VersionRange {
     }
   }
 
-  findNewestValidVersion(version, allVersions) {
+  findNewestValidVersion(version: string, allVersions: any) {
```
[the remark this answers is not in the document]
eggplantzzz: Sorry I made some further changes and ended up giving the compiler list its own type.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -124,90 +141,89 @@ export class VersionRange {
...
-    let allVersions, versionToUse;
-    try {
-      allVersions = await this.getSolcVersions();
-    } catch (error) {
-      throw new NoRequestError(versionConstraint, error);
+  async getSolcFromCacheOrUrl(versionConstraint: string, index: number = 0) {
+    // go through all sources (compilerRoots) trying to locate a
+    // suitable version of the Solidity compiler
+    const { compilerRoots, events } = this.config;
+    if (index > compilerRoots.length - 1) {
```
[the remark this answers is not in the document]
eggplantzzz: Yeah I was ambivalent about this, happy to change it to suit your preference.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
eggplantzzz: Yep I ended up switching the order of this as well.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
eggplantzzz: So there is an `attemptNumber` variable in `getSolcVersionsForSource`. I don't think it would make sense to just pass the index as the second argument. What if I set `const attemptNumber = index + 1;` before using it as an argument? That would make it clear what is going on eh?

haltman-at: Oh yeah, that also works... or just add a comment, honestly.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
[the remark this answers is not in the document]
haltman-at: ...wait, you've got two PRs up at once altering this stuff simultaneously?  Uh, that's confusing...

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
haltman-at: ...wait, you've got two PRs up at once altering this stuff simultaneously?  Uh, that's confusing... I don't really know how to handle that...

eggplantzzz: With respect to the last point you made in your first comment here, I think we should leave the `compilerRoots.length < 1` check because it yields a different error message. In that case it will tell the user they have no compiler root urls.

haltman-at: Makes sense!  I think writing `compilerRoots.length === 0` would be clearer, but it also doesn't really matter. :P

haltman-at (approved):
OK, seems good to me!  One minor style comment but you could also just ignore that. :P

haltman-at:
Also now that I've thought about this more... maybe #4261 should be closed?  I think I misunderstood the logic earlier -- I think I put up that issue because it was just retrying.  But actually it's retrying from a different URL!  So maybe we don't want to use a retry library after all, since that's more for retrying the same operation?  Maybe the way it's done here is just fine.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
eggplantzzz: agreed!

eggplantzzz:
Yeah you are probably right about closing that issue. It seems necessary since we definitely need to do something a bit more complex than just retrying the same URL.

Issue #5334, opened 2022-07-20
Debugger txlog should include raw data for external calls somehow


haltman-at:
## Issue

I recently added raw data for events in the txlog (#5324).  This suggests that the same should happen for calls and returns (presumably just external calls/returns, the idea doesn't make a lot of sense for internal calls and returns).

How this would work would have to be figured out, though.  For calls it makes sense to attach it to the `callexternal` action.  But what about for returns?  One could attach it to the same action, but that doesn't mesh well with how txlog is designed -- consider where error decodings go; they don't go on the external call, they go on the internal call that threw!  This stems from the original intent that txlog would be used as the basis for a visualizer.

So assuming we're not redoing that (and I don't want to redo that!), we'd need to determine where the raw return data would go.  Although I suppose I could possibly implement raw call data without that...?

## Environment

* Truffle version (`truffle version`): 5.5.22

haltman-at:
(The call data part of this is now done; see #5391 for more details.)

PR #5391, opened 2022-08-04
Enhancement: Include raw call data in txlog in all cases

haltman-at:
Does part, but not all, of #5334.  Specifically the calls part; ignored the returns part for the reasons discussed there.

Adds to `externalreturn` actions a `raw` field, with either `raw.calldata` or `raw.binary` (for creations).  Originally I was going to make it just a single field, but I thought it was worth distinguishing?  IDK.

Btw, a thing I'd forgotten: For calls and creations that can't be decoded, we actually already include the raw data!  So uh now we're including it twice in such cases.  Oh well?

I wish I'd remembered that though when thinking about how to handle return values / errors / events in the first place!  It would have been sensible to do those that way initially... oh well.  Well, we can figure out exactly how we want this to work...

Also I updated two tests to test this stuff.

### DIFF
--- a/packages/debugger/lib/txlog/reducers.js
+++ b/packages/debugger/lib/txlog/reducers.js
@@ -185,6 +185,14 @@ function transactionLog(state = DEFAULT_TX_LOG, action) {
           (kind === "function" || kind === "library") &&
           action.absorbNextInternalCall;
       }
+      //include raw data regardless
+      call.raw = {};
+      if (calldata) {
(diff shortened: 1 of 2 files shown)


gnidan (approved):
Makes sense! I concur about separating the two fields for calls vs. creates.

gnidan:
(For the redundant fields, let's just leave those for now, and we'll coordinate about it later)

PR #5394, opened 2022-08-04
Internal improvement: Add more types in compile-solidity

eggplantzzz:


gnidan (approved):
Heckin' yeah

(Is this a breaking change to that package?)

eggplantzzz:
Hmm, why would it be breaking?

Issue #5106, opened 2022-05-11
truffle --version returns "1.0.0" or "unknown"


kevinweaver:
## Issue

The `node`, `npm`, and `nvm` commands have a `--version` flag, but the truffle equivalent is `truffle version`.
If the user runs `truffle --version`, the output is unexpected. 

## Steps to Reproduce
Given truffle  v5.5.13 and
node 14:
```
> truffle --version
1.0.0
```

node 16:
```
> truffle --version
unknown
```

## Expected Behavior

`truffle --version` should be an alias of `truffle version` and return the expected output. Something like:
```
Truffle v5.5.13 (core: 5.5.13)
Ganache v^7.1.0
Solidity v0.5.16 (solc-js)
Node v16.9.1
Web3.js v1.5.3
```

## Actual Results

The user hurts itself in confusion! 

## Environment

* Operating System: MacOS 12.3.1
* Ethereum client:
* Truffle version (`truffle version`):  v5.5.13
* node version (`node --version`): 14 or 16
* npm version (`npm --version`):  8.10.0

MicaiahReid:
Possibly related to #5105.

swisstackle:
Is this still an issue? I have node version 16.14 and truffle version 5.5.15. When I run "truffle --version", I get 5.5.15 as the output.

eggplantzzz:
Well actually `truffle --version` is not really officially allowed. Actually on 5.5.16 it is broken (from the bundling changes in the last release) and yields an error. We should probably write in some code to support some of these common formats for requesting the version and help info.

PR #5395, opened 2022-08-04
Bug Fix: add command handling for '--version' & '-v'

lsqproduction:
This addresses #5106 .
Added handling for `-v` and `--version` in `getCommand()`, so it will work both inside and outside of the console.

Turn off Yargs' default behavior for `--version` in `prepareOptions()` so our logic to display version info will kick in.

### DIFF
--- a/packages/core/lib/command-utils.js
+++ b/packages/core/lib/command-utils.js
@@ -27,7 +27,9 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
 
   // If the command wasn't specified directly, go through a process
   // for inferring the command.
-  if (commands.includes(firstInputString)) {
+  if (firstInputString === "-v" || firstInputString === "--version") {
+    chosenCommand = "version";
+  } else if (commands.includes(firstInputString)) {
     chosenCommand = firstInputString;
   } else if (noAliases !== true) {
     let currentLength = 1;
@@ -89,7 +91,10 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
 // the input options, merges it with the input options, and returns the result
 const prepareOptions = ({ command, inputStrings, options }) => {
   const yargs = require("yargs/yargs")();
-  yargs.command(require(`./commands/${command.name}/meta`));
+  yargs
+    .command(require(`./commands/${command.name}/meta`))
+    //Turn off yargs' default behavior when handling "truffle --version"
+    .version(false);
   const commandOptions = yargs.parse(inputStrings);
 
   // remove the task name itself put there by yargs
--- a/packages/core/lib/command-utils.js
+++ b/packages/core/lib/command-utils.js
@@ -112,8 +112,14 @@ const prepareOptions = ({ command, inputStrings, options }) => {
   });
 
   // method `extractFlags(args)` : Extracts the `--option` & `-option` flags from arguments
-  const inputOptions = extractFlags(inputStrings);
+  let inputOptions = extractFlags(inputStrings);
 
+  //prevent invalid option warning for `truffle -v` & `truffle --version`
+  if (command.name === "version") {
+    inputOptions = inputOptions.filter(
+      opt => opt !== "-v" && opt !== "--version"
+    );
+  }
   // adding allowed global options as enumerated in each command
   const allowedGlobalOptions = command.meta.help.allowedGlobalOptions
     .filter(tag => tag in globalCommandOptions)


packages/core/lib/command-utils.js
```diff
@@ -22,7 +22,9 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
     return null;
   }
 
-  const firstInputString = inputStrings[0];
+  const firstInputString = ["-v", "--version"].includes(inputStrings[0])
```
eggplantzzz, line 25: I feel like this is a slightly confusing way to do this. By that I mean it takes a minute to figure out what the code is attempting to do. You probably Maybe something like the following would be clearer?
```
const firstInputString = inputStrings[0];
if (firstInputString === "-v" || firstInputString === "--version") {
  return {
    // and then explicitly return the whole object here
```

What do you think of that? I will also say that my intuition tells me to avoid messing with `firstInputString` like you are doing just to handle one or two cases that will occur. I'm having trouble explaining exactly why but I don't like swapping out what the user entered and labeling it as `firstInputString`, as in the case where they entered `-v` it isn't the first input string. Let the first input string be the first input string.

Or probably a better idea might be to just add a special case in the `if/else` that is below. Then you can set `chosenCommand` to "version". That is probably a more eloquent way to handle this than what I wrote above.

lsqproduction: I do agree it is probably a better idea to not mess with `firstInputString`. I will work on moving it to the `if/else` below.

packages/core/lib/command-utils.js
```diff
@@ -25,6 +25,9 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
   const firstInputString = inputStrings[0];
   let chosenCommand = null;
 
+  if (firstInputString === "-v" || firstInputString === "--version") {
```
eggplantzzz, line 28: I guess this works but I think it is more clear if this is part of the main `if/else` below.

lsqproduction: Ah I see. Thank you for taking the time to explain this. I was keeping this outside thinking too many if and else if's can be confusing, but I do follow your logic since it is all checking the input to figure out the user's desired command.

eggplantzzz: Yeah, my idea is kind of that they seem much more connected if they are all part of the same one. Then it's more like, "ok, either this, this, or this happens". :)

packages/core/lib/command-utils.js
```diff
@@ -134,7 +139,11 @@ const prepareOptions = ({ command, inputStrings, options }) => {
   let invalidOptions = inputOptions.filter(opt => !validOptions.includes(opt));
 
   // TODO: Remove exception for 'truffle run' when plugin options support added.
-  if (invalidOptions.length > 0 && command.name !== "run") {
+  if (
+    invalidOptions.length > 0 &&
+    command.name !== "run" &&
+    command.name !== "version"
```
eggplantzzz, line 145: Hmm, maybe it is better if we stick `--version` and `-v` in the allowed global options for the version command and make a comment about why it is there? I don't know, they are both hacks and I don't know which is better. Anyone else have opinions on this?

eggplantzzz: Or you could maybe remove the `--version` or `-v` from the result that `extractFlags` returns in the case that we are running the version command. It's kind of ugly writing so much code for this exception but I'm not sure about what the alternative should be.

On the other hand you could just do the same exact thing we do with `--help` and `help` in `cli.js` and create a similar method to `displayGeneralHelp` that would allow us to avoid all this stuff.

lsqproduction: if we do a util function for displaying version info like `displayGeneralHelp` we would need to call this logic as well inside of the `main()` in `console-child.js`. This will require changes in two files. 

If we filter out `--v` out of the `inputOptions` (result of `extractFlags`), then only changes are on `command-util` but still in two functions `getCommand` and `prepareOptions`. 

Which one of these routes would be better?

sukanyaparashar: @lsqproduction, I think the better way might be to filter out the `--version` or `-v` from the `extractFlags` result as @eggplantzzz suggested.

lsqproduction: Yes, I also prefer this method. I will get this up.

packages/core/lib/command-utils.js
```diff
@@ -107,7 +112,9 @@ const prepareOptions = ({ command, inputStrings, options }) => {
   });
 
   // method `extractFlags(args)` : Extracts the `--option` flags from arguments
-  const inputOptions = extractFlags(inputStrings);
+  const inputOptions = extractFlags(inputStrings).filter(
```
eggplantzzz, line 115: We don't want to do this filter for all commands. It is reasonable that we would have a `--version` or `-v` flag for a command and in that case we wouldn't want to remove it here. We only would want to remove it in the case that they are trying to get the version info.

Hmm, it also seems like maybe `extractFlags` is not correct as it won't extract flags that begin with a single `-` such as `-v`.

lsqproduction: Can you help to share some examples of when the command would have `--version` or `-v`? Is it to get the version info for a certain truffle command?
When that happens, we will then remove the filter logic, no? And below it will check it as valid options. 

`extractFlags` only works with `--` right now. Should I add this change to this PR? or separate PR?

eggplantzzz: We don't have a `--version` flag for any other commands. That being said, this is not the correct behavior. If we happen to add a `--version` flag to some other command in the future, we don't want to have to remember, "hey, there was that thing we added a bit ago that will now cause incorrect behavior and now we have to remember to remove it". That wouldn't be great. Instead we can just build in the correct behavior now.

We are making an exception for the `version` command here. We want this code to do exactly what we want and we don't want to create code that "just happens to work" for doing the thing we want in the short term. Make sense?

eggplantzzz: As for the change to `extractFlags`, I would say make it a separate PR. It is easier when PRs are smaller and more focussed subject-wise :)

packages/core/lib/command-utils.js
```diff
@@ -107,8 +112,14 @@ const prepareOptions = ({ command, inputStrings, options }) => {
   });
 
   // method `extractFlags(args)` : Extracts the `--option` & `-option` flags from arguments
-  const inputOptions = extractFlags(inputStrings);
+  let inputOptions = extractFlags(inputStrings);
 
+  //prevent invalid option warning for `-v` & `--version`
```
eggplantzzz, line 117: One more nit and then this looks good; can you make it clear here that it is for `truffle -v` and `truffle --version`? The comment as it is written still seems like it would be confusing for someone. Other than that, if you have tested this I think we are good to go!

eggplantzzz (approved):
I like this

eggplantzzz:
I mean, I'm not too concerned about you messing with the commits. I can see the history of changes made easily enough since there aren't too many.

lsqproduction:
cleaned up the commits anyways :-) good to practice my rebasing.

PR #5398, opened 2022-08-05
Remove abi-util node types warning

benjamincburns:
Removes warning:

```
warning "workspace-aggregator-aed90a07-068f-443d-8404-55b7f884493e > @truffle/abi-utils > ts-node@10.7.0" has unmet peer dependency "@types/node@*".
```

### DIFF
--- a/packages/abi-utils/package.json
+++ b/packages/abi-utils/package.json
@@ -33,6 +33,7 @@
     "@types/faker": "^5.1.2",
     "@types/jest": "27.4.1",
     "@types/jest-json-schema": "^2.1.2",
+    "@types/node": "12.12.21",
     "jest": "28.1.3",
     "jest-extended": "^0.11.5",
     "jest-fast-check": "2.0.0",


haltman-at:
So why this particular `@node/types` version?  I should note we currently have 6 (!) different versions of that package, so I'm wondering if there was any particular reason for selecting this one.

benjamincburns:
> So why this particular `@types/node` version? I should note we currently have 6 (!) different versions of that package, so I'm wondering if there was any particular reason for selecting this one.

Like on the other PR w/ `typescript@4.7.4`, it's because I originally wrote it on my `typescript-4.7` branch and cherry-picked it off of there. `@node/types@12.20.55` is the latest `12.x` node types version, with node 12.x being the earliest node version that we support.

I'd have gone for an earlier patch version (e.g. `^12.0.0`) but I don't think they bugfix old minor type versions, so I couldn't be confident that wouldn't introduce new issues.

In the meantime if you'd prefer this be consistent with something that's already there, please nominate one of the six for me to choose from @haltman-at.

benjamincburns:
Also it looks like [node 12 wasn't LTS until `v12.13.0`](https://nodejs.org/es/blog/release/v12.13.0/), so ideally we'd be using the caret range of that version across the board 🤨

I'll consider that for the TS branch and in the meantime I'll drop this back to `12.12.21`

haltman-at:
No I don't really think it matters that much, that seems perfectly fine...

haltman-at (approved):

PR #5402, opened 2022-08-05
Remove tsutils peer dep warning on bootstrap

benjamincburns:
Removes the following warning:

```
warning "@typescript-eslint/eslint-plugin > tsutils@3.21.0" has unmet peer dependency "typescript@>=2.8.0 || >= 3.2.0-dev || >= 3.3.0-dev || >= 3.4.0-dev || >= 3.5.0-dev || >= 3.6.0-dev || >= 3.6.0-beta || >= 3.7.0-dev || >= 3.7.0-beta".
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "lint-staged": "^12.1.2",
     "nyc": "15.1.0",
     "prettier": "^2.5.1",
-    "prs-merged-since": "^1.1.0"
+    "prs-merged-since": "^1.1.0",
+    "typescript": "^4.1.4"
   },
   "workspaces": {
     "packages": [


cds-amal (commented):
Thanks, @benjamincburns, LGTM!

cds-amal (dismissed):
*clicks the correct button this time* :)

benjamincburns:
> Hold up! I see some runaway npm processes

@cds-amal - not sure I understand - is there something you want me to follow up on?

benjamincburns:
Happy to make this the `4.1.4` version that's currently in use, but I just figured with the `typescript-4.7` branch following close behind this I'd save the extra step.

cds-amal:
I'm not sure what happened. I was testing your PRs when I noticed I had a run-away set of `npm` processes on my machine and decided to retract my approval until I understood what happened.

```
lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.4 LTS
Release:        20.04
Codename:       focal
```

lsqproduction:
> I had a run-away set of `npm` processes

@cds-amal is this same as node processes? or how do you check for the npm processes?

cds-amal:
> @cds-amal is this same as node processes? or how do you check for the npm processes?

You can use [`ps -ef | grep npm`](https://explainshell.com/explain?cmd=ps+-ef+%7C+grep+npm), or `pgrep npm`, if you have [`pgrep`](https://www.unix.com/man-page/osx/1/pgrep/) installed.

haltman-at:
> Happy to make this the `4.1.4` version that's currently in use, but I just figured with the `typescript-4.7` branch following close behind this I'd save the extra step.

IMO it's better to keep things in sync at any given time; so here that would mean using 4.1.4, and then, once this is merged, alert the author of `typescript-4.7` to go update their PR.  After all, things can go wrong and that other PR might not happen as expected!

benjamincburns:
Will do @haltman-at, though for what it's worth there's already version inconsistencies for `typescript` in `develop`. Specifically `compile-solidity` uses `typescript@4.3.5` while the others use `typescript@4.1.4`.

That's definitely not a good reason to drive _further_ inconsistency, however. In fact, I think this is yet more reason why we should introduce a CI check that gives us red builds on dependency version consistency issues like this.

haltman-at:
Yeah right now we only have 2 versions of TS pretty sure.  Was looking through lerna update wizard recently; there were a decent number of 2's, a single 3 (ethers -- should probably do something about that), and then a single 6 O_o (@types/node).

benjamincburns:
`typescript-4.7` fixes the `@types/node` one, thankfully

haltman-at (approved):

Issue #5065, opened 2022-04-28
Switch truffle's default from "strict" to "eager" until WebSockets is the default communication mode.


davidmurdoch:
Update to Ganache v7 should be faster than v6, but many users don't experience this in many (most?) cases as Truffle's web3 must poll for changes (example: https://github.com/OpenZeppelin/openzeppelin-upgrades/pull/523).

If truffle updates to using WebSockets this polling behavior won't occur. In the meantime, I think we should switch our default mode to "eager" (which is also Ganache v7's default).

There _is_ a way to make Ganache behave as if it was in eager mode for messages sent over HTTP connections only, but doing it is very hacky and gross, as it makes the RPC logic change based on the connection kind. So I'd rather not do this (I actually [did do it already](https://github.com/trufflesuite/ganache/compare/3acc35e2e121e8ef2bcebbb3aa5eeec360fc58cf...8b15e3c1aebc4c99a217438fad3d1c208c662f5e), and it really is horrible).

gnidan:
The issue here is that Web3.js's polling behavior means that with HTTP providers, **each transaction in `truffle test` takes _at least one-thousand milliseconds_**, oh my.

We'll want to reevaluate the need for the strict->eager switch after #4637 resolves

kevinbluer:
Heads up that this is addressed in https://github.com/trufflesuite/truffle/pull/5407 👌

PR #5407, opened 2022-08-05
switch ganache to use 'eager' mode in test environment

eggplantzzz:
Ganache 7 introduced 2 instamine modes: 'strict' and 'eager'. In eager mode, Ganache would receive a tx, mine it, and then send a response. This is not an accurate representation of how nodes behave in the wild so 'strict' mode was implemented in the test flow in Truffle in order to more accurately reflect this behabvior. However, this caused test suites to run a lot slower because of polling behavior in web3. We believe it does something like poll every 1 second to see if the tx has been included (after a tx is submitted). So this would cause each tx in a test to take at least that polling delay to complete. With 'eager' mode enabled (even though this behavior is innaccurate), it was always immediately available and therefore ran much faster. In the interest of speed for testing, we have decided to re-enable 'eager' mode

### DIFF
--- a/packages/core/lib/configAdapter.js
+++ b/packages/core/lib/configAdapter.js
@@ -88,7 +88,7 @@ function configureManagedGanache(config, networkConfig, mnemonic) {
     gasPrice,
     time: genesisTime,
     miner: {
-      instamine: "strict"
+      instamine: "eager"
     }
   };
 


cds-amal (approved):
Looks good to me.

cliffoo (approved):
LGTM as well! Ganache v7 release notes for reference [here](https://github.com/trufflesuite/ganache/releases/tag/v7.0.0) (`miner.instamine` under New Startup Options).

eggplantzzz:
this PR addresses https://github.com/trufflesuite/truffle/issues/5065

Issue #3830, opened 2021-02-20
Support Vyper debugging on Vyper versions back to 0.1.0-beta.7 (old-style Vyper sourcemaps)


haltman-at:
## Issue

Currently, the debugger only supports Vyper back to version 0.1.0-beta.12, because that's when Solidity-style sourcemaps were added.  Vyper however has its own style of sourcemaps that go back to 0.1.0-beta.7; we should support this.

PR #5408, opened 2022-08-06
Various improvements to `truffle opcode` and `@truffle/code-utils`

haltman-at:
OK, I maybe stuffed a bit too many different things into this one. XD  (Note: **Contains breaking changes to `@truffle/code-utils`.**)

This PR:
1. Fixes various bugs in `truffle opcode` and `@truffle/code-utils`
2. Reworks the interface to `@truffle/code-utils`
3. Adds handling to `@truffle/code-utils` for Vyper 0.3.4
4. Adds handling to `truffle opcode` for really old Vyper versions :)
5. A few smaller things

In order to understand how this PR came about, it's useful to understand a bit of the history here.  Originally, the `parseCode` function exported by `@truffle/code-utils` took only one argument, the bytecode to be disassembled.  It also attempted to strip the metadata off the end.  However, metadata stripping is unreliable, so at some point I added a second argument.  The second argument is the number of instructions expected -- as might be gotten from a source map -- and if not passed it behaves the old way, where it attempts to strip the metadata.  I also altered the various invocations of `parseCode` to pass this second argument, so we wouldn't actually be using the metadata stripping.

This is a terrible interface!  I had planned to make a breaking change and redo it when it came time to add EOF support, but, um, I ended up doing it now.  See, the reason this PR came about is I noticed a bug in Truffle Opcode: When I changed it to use the source map, I also made it so that it assumed the source map was present; if it was absent, it would crash.  Oops.

So, first things first, I've changed Truffle Opcode so it no longer makes that assumption.  Now, if the source map is present, it will pass the number of instructions, but if it's not present, it will attempt to perform metadata stripping instead.

Of course, as mentioned, I've also reworked the interface to `parseCode`; now, instead of the old way where metadata stripping is turned on if you don't pass an instruction count, instead the second argument is now an options argument, and both of these are now options.  Obviously, this is a breaking change.  (OK, I guess I could have preserved the old interface, but I didn't really see any reason to.)  Since I was already making breaking changes to `@truffle/code-utils`, I also went ahead and removed some exports that were only there for compatibility.  Of course I had to update the various invocations of `parseCode` as well, but that was straightforward.

(Note that my intent is that, when we implement EOF support, EOF bytecode will simply ignore these options, as the purpose of both options is for separating code from data, something that EOF does on its own.)

While I was at it, I also made it so that Truffle Opcode can use old-style Vyper sourcemaps as well as Solidity sourcemaps... OK, semi-old-style, ones recent enough that our shim can handle it.  Sorry, the really old ones are still getting ignored!  I also added special handling to `parseCode` so that, when metadata stripping is turned on, it will also strip out Vyper 0.3.4's idiosyncratic metadata format (that is now gone in Vyper 0.3.5, hooray).  Note I did *not* make any changes to `@truffle/codec`'s metadata handling for this; Codec's concern is bytecode matching, not disassembly, and since this special metadata is *fixed*, Codec doesn't need to give it any special treatment.

I also added two tests to `@truffle/code-utils`; one of the special metadata stripping for Vyper 0.3.4, and one of the more normal Solidity metadata stripping (which wasn't there before!).  And it's a good thing I added that latter test, because it turns out that, uh, our Solidity metadata stripping was actually broken.  (Good thing we were no longer using it anywhere!  But now we will be again...)  The problem is that we were using an old version of `cbor` which didn't accept `Uint8Array`s.  (This problem didn't affect Codec, as it was passing in hex strings instead.) ~~So I updated `cbor` to 7.0.6; not to 8.x because 8.1.0 failed to build for some reason (I guess I could have tried an earlier 8.x version, but I didn't bother).  That solved the problem.  It also let me ditch the `@types/cbor` dependency, as `cbor` now supplies its own types.~~  (Scratch that, see comments below -- I ended up *not* updating `cbor` and using a `Buffer.from()` instead.)

Also, it turns out that the `@truffle/code-utils` dependency was missing in core, so I added it.  I wonder why depcheck didn't catch that...? @cds-amal you may want to know about this.

And that (together with some minor style and comment improvements) basically covers it!  Note that I only added tests of `parseCode`, not of `truffle opcode` itself; I did the testing of the new `truffle opcode` cases (no sourcemap, old Vyper-style sourcemap) manually.


haltman-at:
Ugh, dammit, looks like cbor 7.0.6 is causing problems elsewhere.  OK, I'll have to figure out what to do about that...

haltman-at:
Ugh -- if I downgrade to `cbor` 6.0.1, then the bootstrap succeeds... except it takes forever and the resulting bundle size of Truffle Contract is *huge*. O_o

Meanwhile, if I downgrade to `cbor` 5.2.0, we hit the original problem.

Possible options here:
1. Switch to a different `cbor` package;
2. Stay on `cbor`, but work around the bug with a conversion.

I'm going to go with (2) for now, but wondering if other people have opinions!

packages/code-utils/package.json
```diff
@@ -21,7 +21,7 @@
   },
   "types": "./dist/src/index.d.ts",
   "dependencies": {
-    "cbor": "^5.1.0"
+    "cbor": "^5.2.0"
```
benjamincburns, line 24: was the minor bump here intentional?

packages/code-utils/src/index.ts
```diff
@@ -1,53 +1,72 @@
...
- * map, then add one) this is used to exclude metadata instead.
+ * These options can be combined, although I'm not sure why you'd want to.
  *
  * @param  {String} hexString Hex string representing the code
  * @return Array               Array of instructions
  */
 export function parseCode(
   hexString: string,
-  numInstructions: number = null
+  { maxInstructionCount, attemptStripMetadata }: DisassemblyOptions = {}
```
benjamincburns, line 27: Do we care about breaking changes affecting external users of this package? I think not, but I'm not certain.

On the off chance that we do, would it be impossible or a bad idea to provide an overloaded signature so it's not a breaking change?

packages/code-utils/src/index.ts
```diff
@@ -1,53 +1,72 @@
...
+ * constructors that have had arguments attached to the end!
  *
- * If you pass in numInstructions (hint: count the semicolons in the source
- * map, then add one) this is used to exclude metadata instead.
+ * These options can be combined, although I'm not sure why you'd want to.
  *
  * @param  {String} hexString Hex string representing the code
  * @return Array               Array of instructions
  */
 export function parseCode(
```
benjamincburns, line 25: How much do we care about cyclomatic complexity? This function had what I'd consider to be high cyclomatic complexity prior to this change, and then this change increases it substantially. Perhaps this should be broken out into a few child functions for easier readability?

packages/code-utils/src/index.ts
```diff
@@ -76,17 +95,15 @@ export function parseCode(
...
 function isValidCBOR(metadata: Uint8Array) {
   try {
     //attempt to decode but discard the value
     //note this *will* throw if there's data left over,
     //which is what we want it to do
-    cbor.decodeFirstSync(metadata);
+    //HACK: this version of cbor doesn't accept Uint8Arrays,
+    //but it does accept Buffers.  (Unfortunately newer versions
+    //cause problems. :-/ )
+    cbor.decodeFirstSync(Buffer.from(metadata));
```
benjamincburns, line 106: What problems does the latest `cbor` cause? My gut feel on things like this is that it's usually best to go for the latest version of a package unless it's just broken, as we're likely to want to move to that version later on for various reasons anyway (e.g. due to CVEs, etc).

That said, if we revert the minor version change in this PR I suppose that's fine, too.

benjamincburns: Alternatively is it possible to refactor the code above to use `Buffer` instead of `UInt8Array` so that we don't need the hack here? If you think doing so would be painful, cause performance or readability issues, please disregard.

packages/code-utils/test/index.test.ts
```diff
@@ -17,21 +19,43 @@ describe("CodeUtils.parseCode", function () {
...
   });
 
   it("doesn't strip invalid metadata", function () {
-    const parsedCode = CodeUtils.parseCode(
-      "0x600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600060000100"
-    );
+    const bytecode =
+      "0x600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600050600060000100";
```
benjamincburns, line 37: I don't think that you necessarily need to fix this as part of this PR because this problem is quite pervasive in our tests, but I think tests like this would be a lot easier for people such as myself to understand these tests if the code that generated these hex strings was checked in somewhere and referenced by this test suite.

To be clear I think this bytecode is generated code that _should_ be checked in, and I don't think that we should take the extra performance or reliability hit to regenerating this bytecode on every run of the test suite. So take the word "referenced" above to have a very loose definition - even a comment pointing at a script that generates the bytecode from a source file would be fine.

benjamincburns: Though now that I wrote that, I think I might've commented on a hex string that would've needed to be hand-edited, as it's the invalid case, but the point still remains. :-)

Although even in this case it'd be good to have it be mutating a different correct bytestring so that it's clear _how_ the metadata here is invalid.

packages/codec/lib/contexts/utils.ts
```diff
@@ -252,6 +252,9 @@ interface CborInfo {
   cborSegment: string;
 }
 
+//note: we *don't* bother handling Vyper 0.3.4's idiosyncratic CBOR
+//here, because it's always fixed, so there just isn't any need to;
+//we're OK with just not normalizing that
```
benjamincburns, line 257: I assume this is a comment for future package maintainers who might be looking for Viper 0.3.4 CBOR handling here?

If it's meant to help consumers of the function, I'm afraid that it isn't prescriptive enough for my feeble brain to understand. Specifically it's not clear to me if I should avoid calling this with binaries produced by Vyper 0.3.4, or if I should expect it to error or return null if I do call it with said Vyper metadata.

If it's easy to detect that the binary contains said CBOR (e.g. with a call to `isValidCBOR` as done in `packages/code-utils/src/index.ts`, perhaps this function should throw some `InvalidCBORError` with a message that indicates that Viper 0.3.4 isn't supported here.

Alternatively if we're already just returning `null` in that case, perhaps this should just be converted to typedoc comments and that can be noted in the `@returns` line.

packages/codec/package.json
```diff
@@ -29,7 +29,7 @@
     "@truffle/compile-common": "^0.7.32",
     "big.js": "^6.0.3",
     "bn.js": "^5.1.3",
-    "cbor": "^5.1.0",
```
benjamincburns, line 32: Same here re minor version bump - was that intentional?

packages/core/lib/commands/opcode/run.js
```diff
@@ -21,14 +21,22 @@ module.exports = async function (options) {
...
   }
 
-  let bytecode = Contract.deployedBytecode;
-  let numInstructions = Contract.deployedSourceMap.split(";").length;
+  const bytecode = options.creation
+    ? Contract.bytecode
+    : Contract.deployedBytecode;
+  const sourceMap = Compilations.Utils.simpleShimSourceMap(
+    options.creation ? Contract.sourceMap : Contract.deployedSourceMap
+  ); //the shim is in case of an old-style Vyper source map
```
benjamincburns, line 29: Is this shim useful elsewhere? If so, perhaps it should be the responsibility of either `@truffle/contract` or `@truffle/resolver` to conditionally apply it when applicable?

packages/code-utils/src/index.ts
```diff
@@ -1,53 +1,72 @@
 import parseOpcode from "./opcodes";
-import type { Instruction, OpcodeTable, opcodeObject, opcodes } from "./types";
-export type { Instruction, OpcodeTable, opcodeObject, opcodes };
+import type { Instruction, OpcodeTable, DisassemblyOptions } from "./types";
+export type { Instruction, OpcodeTable, DisassemblyOptions };
 import * as cbor from "cbor";
```
benjamincburns, line 4: My guess is that this (importing the entire package contents) was the source of the bundle bloat issue that you observed, though I'm betting that `cbor` isn't a TS module, so the TS compiler probably can't do proper tree-shaking here.

Perhaps it's possible that webpack can, but I also doubt it.

benjamincburns (commented):
I'm not super familiar with this code, so I have a big pile of questions that probably more represents my lack of a deep knowledge of this part of our codebase moreso than shortcomings in this PR. Much gratitude in advance for bearing with me there! 😄

packages/code-utils/package.json
haltman-at: I mean, it wasn't *intentional* per se, but it wasn't unintentional either.  When I downgraded, I figured I might as well go to 5.2.0 (the latest 5.x) rather than what we had previously.

packages/codec/lib/contexts/utils.ts
benjamincburns: I think I lean toward wanting it to throw, as that's a bit more aligned with a "fail fast and fail loudly" mindset than returning `null` would be. Assuming it does return `null`, prior to the addition of your comment it wouldn't have been very clear as to why that was happening to anyone other than someone very deeply familiar with the ins-and-outs of the various compilers and CBOR.

packages/code-utils/src/index.ts
haltman-at: Yeah, I figured it was probably OK to just break this and that the extra work to make it nonbreaking wasn't worth it... but that's what review is for!  I should note that I discussed this with @gnidan earlier and he thought it was OK to just break it, so that's at least one more person in favor of just breaking it. :)

packages/codec/package.json
haltman-at: See above.

packages/codec/lib/contexts/utils.ts
benjamincburns: All of that said - the comment is a net improvement over the state of things as you found it, so I'm happy for that to merge as-is.

packages/code-utils/src/index.ts
benjamincburns: I mean this very much looks to be an internal package (ideally we wouldn't publish those 😬), so I certainly have no qualms. Just wanted to make sure it was a conscious choice.

packages/core/lib/commands/opcode/run.js
haltman-at: I mean, we already use this shim elsewhere!  This PR didn't create the shim, it just used the one I'd already written.

If we were going to do something like what you suggest, I imagine it would be in `@truffle/config`.  I don't think that would be a good idea though, because I think it could potentially limit us in the future.  Way way down on my list of priorities somewhere is issue #3830, which that sort of auto-shimming would make impossible.  Granted, I am pretty skeptical I'll ever *actually* bother to implement a solution for #3830 (I'm surprised it got tagged priority 6 instead of 7!), but the potential for other issues of that sort in the future make me wary of doing that sort of thing.

packages/code-utils/src/index.ts
benjamincburns: Also just to be clear I'm not suggesting that you must either revert back to 5.1.0 or use 7.x. I think saying "if you change the version you _must_ do so in the most optimal fashion possible" is a bit pennywise, pound foolish.

Just mentioned all of the above in case you encountered the issues with 7.x and "noped out" of the change before determining whether it'd be high or low effort to fix said issues.

packages/code-utils/src/index.ts
haltman-at: Hm yeah my understanding is that Webpack's tree-shaking capabilities are pretty limited. :-/  But I also don't know this stuff very well so I could be wrong?

packages/code-utils/src/index.ts
haltman-at: Oh yeah as I mentioned above, it had been my intent for a while to make that breaking change; I just hadn't really had any reason to touch this code anytime recently.  I figured I'd end up doing it when Shanghai hit and I needed to rewrite large chunks of this, but since I ended up going over this code *now* I figured I'd do it now instead!

packages/core/lib/commands/opcode/run.js
benjamincburns: Hopefully in that case the shim detects if it's already been applied and bails early, and/or the lifecycle of a shimmed contract is always quite short when this is in use?

I just tend to view mutations like this as difficult-to-untangle consistency issues lying in wait.

benjamincburns: Ah but I'm a dolt - you're not mutating the contract instance - it's returning the shimmed `sourceMap`. Nevermind me.

packages/code-utils/src/index.ts
haltman-at: I don't actually remember offhand what the problems with 7.x were, good question.  I think they were build problems in `@truffle/contract`, similar but not identical to what I got with 8.x?  I'd have to try it again to be certain, though.

packages/code-utils/src/index.ts
benjamincburns: I'm in a similar boat. I've had to acquire deep knowledge of what webpack can and can't do in that area, and then I quickly lose it the moment I get away from the problem. From what I remember it has a bunch of facilities for manually pruning out code trees in cases like this, but in doing so you have to be quite confident that you'll never branch into said pruned code. Rather ugly, IMO.

packages/code-utils/package.json
benjamincburns: Yeah, I suppose that's fine. I'll leave it up to you as to whether it's worth the effort to pursue further changes for 7.x, though by the fact that you didn't do that already I assume that you've already determined that it isn't.

packages/code-utils/src/index.ts
benjamincburns: Ah sorry, I keep saying `7.x` when I mean whatever the latest version is. My apologies if that's `8x` or something else.

benjamincburns (approved):
Approving this with a couple of caveats.

It's my strong belief that if a code change doesn't introduce new bugs and reflects a net positive in value, we shouldn't block it from merging. We can always make further improvements later. That said, I think if we're seeing maintenance issues with the `cbor` package, we should investigate moving to one of the other CBOR parser packages. I don't think it's wise or necessary to block this change on completion of that effort, however.

As for my other unresolved comments, I think the same applies. We don't have a stated max cyclomatic complexity metric in our dev-handbook, so I consider that requested change optional.

Similarly, I don't think it's within the scope of the problem that's being solved here to fix the opaque/unreadable/write-only-code hex strings in the test suite, so I regard that requested improvement as optional as well.

packages/code-utils/src/index.ts
haltman-at: OK, here's the error message:
```
@truffle/contract: $ yarn compile
@truffle/contract: $ NODE_ENV=production webpack --config ./webpack.config.js && NODE_ENV=development webpack --config ./webpack.config.js
@truffle/contract: assets by status 7.81 MiB [cached] 1 asset
@truffle/contract: orphan modules 2.42 MiB [orphan] 483 modules
@truffle/contract: runtime modules 1.28 KiB 7 modules
@truffle/contract: modules by path ../../node_modules/ 11.9 MiB 2821 modules
@truffle/contract: modules by path ../codec/ 928 KiB 297 modules
@truffle/contract: modules by path ../contract-schema/ 234 KiB 47 modules
@truffle/contract: modules by path ./ 416 KiB 23 modules
@truffle/contract: modules by path ../interface-adapter/ 285 KiB 22 modules
@truffle/contract: optional modules 300 bytes [optional] 20 modules
@truffle/contract: modules by path ../debug-utils/ 152 KiB 10 modules
@truffle/contract: modules by path ../compile-common/dist/src/ 11.8 KiB 8 modules
@truffle/contract: modules by path ../abi-utils/dist/lib/*.js 13.9 KiB
@truffle/contract:   ../abi-utils/dist/lib/index.js 1.34 KiB [built] [code generated]
@truffle/contract:   + 3 modules
@truffle/contract: + 7 modules
@truffle/contract: ERROR in ../../node_modules/cbor/lib/utils.js 22:9-42
@truffle/contract: Module not found: Error: Can't resolve 'node-inspect-extracted' in '/home/sniffnoy/truffle/truffle/node_modules/cbor/lib'
@truffle/contract: resolve 'node-inspect-extracted' in '/home/sniffnoy/truffle/truffle/node_modules/cbor/lib'
@truffle/contract:   Parsed request is a module
@truffle/contract:   using description file: /home/sniffnoy/truffle/truffle/node_modules/cbor/package.json (relative path: ./lib)
@truffle/contract:     resolve as module
@truffle/contract:       /home/sniffnoy/truffle/truffle/node_modules/cbor/lib/node_modules doesn't exist or is not a directory
@truffle/contract:       /home/sniffnoy/truffle/truffle/node_modules/cbor/node_modules doesn't exist or is not a directory
@truffle/contract:       /home/sniffnoy/truffle/truffle/node_modules/node_modules doesn't exist or is not a directory
@truffle/contract:       looking for modules in /home/sniffnoy/truffle/truffle/node_modules
@truffle/contract:         single file module
@truffle/contract:           using description file: /home/sniffnoy/truffle/truffle/package.json (relative path: ./node_modules/node-inspect-extracted)
@truffle/contract:             no extension
@truffle/contract:               Field 'browser' doesn't contain a valid alias configuration
@truffle/contract:               /home/sniffnoy/truffle/truffle/node_modules/node-inspect-extracted doesn't exist
@truffle/contract:             .js
@truffle/contract:               Field 'browser' doesn't contain a valid alias configuration
@truffle/contract:               /home/sniffnoy/truffle/truffle/node_modules/node-inspect-extracted.js doesn't exist
@truffle/contract:             .json
@truffle/contract:               Field 'browser' doesn't contain a valid alias configuration
@truffle/contract:               /home/sniffnoy/truffle/truffle/node_modules/node-inspect-extracted.json doesn't exist
@truffle/contract:             .wasm
@truffle/contract:               Field 'browser' doesn't contain a valid alias configuration
@truffle/contract:               /home/sniffnoy/truffle/truffle/node_modules/node-inspect-extracted.wasm doesn't exist
@truffle/contract:         /home/sniffnoy/truffle/truffle/node_modules/node-inspect-extracted doesn't exist
@truffle/contract:       /home/sniffnoy/truffle/node_modules doesn't exist or is not a directory
@truffle/contract:       looking for modules in /home/sniffnoy/node_modules
@truffle/contract:         single file module
@truffle/contract:           No description file found in /home/sniffnoy/node_modules or above
@truffle/contract:           no extension
@truffle/contract:             /home/sniffnoy/node_modules/node-inspect-extracted doesn't exist
@truffle/contract:           .js
@truffle/contract:             /home/sniffnoy/node_modules/node-inspect-extracted.js doesn't exist
@truffle/contract:           .json
@truffle/contract:             /home/sniffnoy/node_modules/node-inspect-extracted.json doesn't exist
@truffle/contract:           .wasm
@truffle/contract:             /home/sniffnoy/node_modules/node-inspect-extracted.wasm doesn't exist
@truffle/contract:         /home/sniffnoy/node_modules/node-inspect-extracted doesn't exist
@truffle/contract:       /home/node_modules doesn't exist or is not a directory
@truffle/contract:       /node_modules doesn't exist or is not a directory
@truffle/contract:  @ ../../node_modules/cbor/lib/decoder.js 6:14-32
@truffle/contract:  @ ../../node_modules/cbor/lib/cbor.js 6:0-38
@truffle/contract:  @ ../codec/dist/lib/contexts/utils.js 31:26-41
@truffle/contract:  @ ../codec/dist/lib/contexts/index.js 29:27-45
@truffle/contract:  @ ../codec/dist/lib/index.js 139:30-51
@truffle/contract:  @ ../debug-utils/index.js 4:14-39
@truffle/contract:  @ ./lib/promievent.js 2:19-50
@truffle/contract:  @ ./lib/execute.js 2:19-42
@truffle/contract:  @ ./lib/contract/index.js 4:16-37
@truffle/contract:  @ ./index.js 2:17-42
@truffle/contract: webpack 5.73.0 compiled with 1 error in 90166 ms
```

packages/codec/lib/contexts/utils.ts
haltman-at: > I assume this is a comment for future package maintainers who might be looking for Viper 0.3.4 CBOR handling here?

Yup!

> If it's meant to help consumers of the function, I'm afraid that it isn't prescriptive enough for my feeble brain to understand. Specifically it's not clear to me if I should avoid calling this with binaries produced by Vyper 0.3.4, or if I should expect it to error or return null if I do call it with said Vyper metadata.

Ah, sorry.  TBH I wasn't really expecting there to be other future consumers of this function; you'll notice it's not exported.

To answer your question: It will return `null`.  What this function does is, attempt to find CBOR info on the end of a binary, return information about it if it's present, and return `null` if it's not.  Since Vyper 0.3.4 does its CBOR in a weird way, this function will fail to detect it and return `null`.  Due to it only being used in this one specific context, however, that's OK, and the point of the comment is to explain why.

Throwin an error instead of returning `null` would of course have been another way to do it, but not what I went with.  Note that detecting the Vyper 0.3.4 situation and doing something different in that case is no easier than detecting the Vyper 0.3.4 situation and returning meaningful information, so doing that would not save any work.

packages/code-utils/src/index.ts
haltman-at: Perhaps we can save that for when this function will be need to be substantially rewritten come Shanghai? :)

packages/code-utils/test/index.test.ts
haltman-at: Oh, that's an interesting idea!

Some of these tests predate me and I simply don't know what the source code is that generated them.  And as for the ones I generated... I'm not sure I know just which source code of mine I generated all those of those from, either!

(The one you commented on I actually hand-wrote... not just hand-edited, hand-wrote. :)  It just repeatedly pushes a zero onto the stack and then pops it off; finally it pushes two zeroes onto the stack, adds them, and then terminates!  I don't think any of the ones made by compiling actual source code have any edits in them.)

I agree actually having the source code on hand would be nice.  That said, compiling Vyper in Truffle does have some complications, especially if you need to do so on a specific version of Vyper such as 0.3.4, especially especially if you maybe want multiple specific versions, as we do here.  (The "long input" test was compiled on a pre-0.3.4 version of Vyper, and that's essential to the test.)  So unfortunately I don't know if that's practical.

But yeah I could add some sort of comment at least... where would the actual source file go, though?  (For the ones whose source files we can determine, anyway.)

packages/code-utils/src/index.ts
haltman-at: > Also if we have to stay on 5.1.0 or 5.2.0, is it possible to refactor the code above to use Buffer instead of UInt8Array so that we don't need the hack here? If you think doing so would be painful, cause performance or readability issues, please disregard.

No, it would be a pretty straightforward change.  I could totally go do that.

haltman-at: I guess there's some extent to which I'm wary of expanding the use of the Node-specific Buffer class... but I guess I'm using it anyway here, and we use it in other places without problem, so I guess it doesn't really matter. :P

cds-amal:
> Also, it turns out that the @truffle/code-utils dependency was missing in core, so I added it. I wonder why depcheck didn't catch that...? @cds-amal you may want to know about this.

Whoa! Thanks, @haltman-at!

packages/source-map-utils/index.js
```diff
@@ -109,8 +109,10 @@ var SourceMapUtils = {
     }
 
     //because we might be dealing with a constructor with arguments, we do
-    //*not* remove metadata manually
-    let instructions = CodeUtils.parseCode(binary, numInstructions);
+    //*not* pass attemptStripMetadata under any circumstances
```
cds-amal, line 112: I think it would be helpful to explain why metadata is preserved for potential cstor with artuments.

cds-amal (approved):
Looks good to me. I'm approving pending resolution of @benjamincburns 's unresolved conversations.

packages/code-utils/src/index.ts
benjamincburns: This would be my suggestion if you even want to make the change at all.

I don't think eliminating the `UInt8Array` adds any functional value, and the minor performance and code quality improvements that we get for the change are arguably quite low value by comparison to the rest of the changes here, so I see no reason to block the merge of this PR.

packages/code-utils/src/index.ts
benjamincburns: Same.

packages/code-utils/src/index.ts
benjamincburns: Re: the webpack error above, I think solving that is also not worth doing in this PR.

That said, without having looked at the `cbor` lib at all, I'd bet an arbitrary modest sum that the error in question is webpack choking on a conditional require for a missing (hopefully optional) peer dependency, and you very likely can add `node-expect-extracted` to the list of externals in the webpack config to resolve it.

You'll need to have a look at the code to verify that's the case, however.

packages/codec/lib/contexts/utils.ts
benjamincburns: I think it's fine as-is. As purely a style preference I tend to like to avoid returning `null`, as I feel that throwing errors communicates the design intent a bit better, however I think it's perfectly acceptable practice here, _especially_ since this is a local function.

packages/code-utils/test/index.test.ts
benjamincburns: Again, I think this is fine to merge as-is, however I think we need to raise an issue about this with some priority, and at some stage (pending team consensus) we should probably draw a line in the sand and prohibit new tests that contain opaque/unreadable binaries.

I'll go ahead and create that issue (assuming one hasn't been created already).

benjamincburns: Issue created as #5419

packages/code-utils/src/index.ts
haltman-at: Interesting.  Yeah I'm not very familiar with webpack errors.

Yeah 8.x gave a different build error which I didn't post.  But I can post that error message too if people care...

haltman-at: FWIW, here's the error we get on 8.x:

```
@truffle/code-utils: $ yarn build
@truffle/code-utils: $ tsc
@truffle/code-utils: ../../node_modules/cbor/types/lib/encoder.d.ts(102,40): error TS2304: Cannot find name 'URL'.
@truffle/code-utils: error Command failed with exit code 2.
@truffle/code-utils: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
@truffle/code-utils: error Command failed with exit code 2.
@truffle/code-utils: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

haltman-at: Oh interesting -- I'm trying different 7.x versions and they don't all lead to that error!  Like 7.0.0 doesn't.  It does bloat the bundle though. :-/  Oh well.

packages/source-map-utils/index.js
haltman-at: ...so, I was going to explain why, but on writing out the explanation, I realized this doesn't actually make sense anymore!  So I'll explain why I did it this way, and then I'll go change it. :P

Suppose you have a constructor, and you attempt to remove the metadata.  But there are arguments on the end of the constructor -- meaning it's the arguments on the end, not the metadata!  So the attempt to remove the metadata could go very badly.  More specifically, if the last two bytes get interpreted as a large length, then a whole lot could be incorrectly removed, way more than just metadata!

The reason why this no longer applies is because these days we only remove valid metadata.  The probability of us removing code by accident is quite low.  That said, if the last two bytes are *small*, we could notionally remove parts of the *arguments* by accident... but who cares, arguments are data and not code!

Hm, I dunno.  Maybe it's best to leave this as-is after all.  This case honestly doesn't really matter that much.  But I can at least explain it more?

PR #5410, opened 2022-08-08
Gracelessly shove Hardhat support into `truffle debug`

gnidan:
wip... this is a hack and a half

### DIFF
--- a/packages/compile-common/src/shims/FromHardhat.ts
+++ b/packages/compile-common/src/shims/FromHardhat.ts
@@ -1,0 +1,183 @@
+import type * as Hardhat from "hardhat/types";
+
+import type {
+  Compilation,
+  CompiledContract,
+  Bytecode,
+  LinkReference
+} from "../types";
+
+const supportedFormats = new Set(["hh-sol-build-info-1"]);
+
+export function buildInfoCompilation(
+  buildInfo: Hardhat.BuildInfo
+): Compilation {
+  const { _format } = buildInfo;
+
+  if (!supportedFormats.has(_format)) {
+    throw new Error(`Unsupported build info format: ${_format}`);
+  }
+
+  const sourceIndexes = buildInfoSourceIndexes(buildInfo);
+
+  return {
+    sourceIndexes,
+    sources: buildInfoSources(buildInfo, sourceIndexes),
+    contracts: buildInfoContracts(buildInfo),
+    compiler: {
+      name: "solc",
+      version: buildInfo.solcLongVersion
+    }
+  };
+}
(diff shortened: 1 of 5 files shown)


packages/core/lib/commands/debug/run.js
```diff
@@ -1,21 +1,50 @@
...
+    config = loadConfig(options);
+  } catch {
+    config = Config.default();
+    config.network = "development";
+    config.networks["development"] = {
+      url: "http://127.0.0.1:8545",
+      network_id: "*"
+    };
+
+    require(`${config.working_directory}/node_modules/hardhat/register`);
```
alcuadrado, line 30: Not needed if you then import hardhat

packages/core/lib/commands/debug/run.js
```diff
@@ -1,21 +1,50 @@
...
+    const {
+      artifacts
+    } = require(`${config.working_directory}/node_modules/hardhat`);
+    const buildInfoPaths = await artifacts.getBuildInfoPaths();
+    const buildInfos = await Promise.all(
```
alcuadrado, line 35: This will lead to an OOM. I'd read one build info at the time, as they can be huge. They have the entire solc output, so they can take 100mb+ in some large projects.

packages/compile-common/src/shims/FromHardhat.ts
```diff
@@ -0,0 +1,183 @@
...
+
+const supportedFormats = new Set(["hh-sol-build-info-1"]);
+
+export function buildInfoCompilation(
+  buildInfo: Hardhat.BuildInfo
+): Compilation {
+  const { _format } = buildInfo;
+
+  if (!supportedFormats.has(_format)) {
+    throw new Error(`Unsupported build info format: ${_format}`);
```
alcuadrado, line 18: <3 

It's great that you are validating this!

packages/core/lib/commands/debug/run.js
gnidan: I tried it without this and this line was necessary... I think I'm going to go with the approach of shelling out to `node -r hardhat/register` instead of this.

packages/core/lib/commands/debug/run.js
gnidan: Good call!

packages/compile-common/src/shims/FromHardhat.ts
gnidan: Of course :D

gnidan:
I want this branch name back.

PR #5411, opened 2022-08-08
Remove HECO testnet from Etherscan fetcher

haltman-at:
Etherscan seems to have removed support for the HECO testnet?  Removing it here as well.

(More accurately, it seems like they've dumped hecoinfo.com entirely... except someone is still keeping it running with the same API, so we can still use it!  But they've removed support for the testnet.  IDK.)


sukanyaparashar (approved):
Looks good to me, Harry !

cds-amal:
> Etherscan seems to have removed support for the HECO testnet? Removing it here as well.
> 
> (More accurately, it seems like they've dumped hecoinfo.com entirely... except someone is still keeping it running with the same API, so we can still use it! But they've removed support for the testnet. IDK.)

What a weird situation!

cds-amal (approved):
LGTM

PR #5413, opened 2022-08-08
Dependency update: Update solc dependencies to 0.8.16 in truffle packages

sukanyaparashar:
This PR upgrades the `solc` version dependencies to latest **0.8.16** in `contract-schema` and `compile-solidity` packages.


eggplantzzz (approved):
i like it

PR #5420, opened 2022-08-10
Devise new experimental @truffle/from-hardhat package for compatibility translation

gnidan:
_(spun out from #5410... please see #5434 for the work to actually provide `truffle debug` support)_

## Overview

This PR introduces a new, _experimental_ package **@truffle/from-hardhat** to provide the following interfaces:

- `expectHardhat()`, which returns successfully iff the working directory is part of a Hardhat project, and iff the installed Hardhat version is supported (`^2.10.1`, current as of this PR). Exceptions are thrown to distinguish these failure modes. (This function is intended for use inside a `try`/`catch`) 

- `prepareCompilations()` that translates Hardhat "BuildInfo" into our native @truffle/compile-common "Compilation"

- `prepareConfig()` which _minimally_ reads the Hardhat config and converts it into a @truffle/config. Basically this only contains plain `url` networks. 🔍 **See documented caveats**: https://github.com/trufflesuite/truffle/blob/e9da6d97810d12041e6ecc21a7a62bd932b72fe5/packages/from-hardhat/src/api.ts#L65-L94

## Intended use

This PR intends for this package to be used by specific **read-only** Truffle commands, primarily `truffle debug`. At this time, this PR does not seek to provide any kind of support for `truffle compile`, `migrate`, `test`, etc. This package does not likely even provide enough information for most of other read-only commands, e.g. `truffle networks`, because no attempt is made to encode a means by which to convert deployment information into Truffle's artifacts.

🤔 Why so limited? Because even `truffle debug` is heckin' useful, and **I posit that we should do this despite its being scrappy and experimental because it's** 🔥 🔥 🔥

## Approach

This PR seeks to avoid intermixing Truffle's and Hardhat's Node.js runtimes. This new package includes Hardhat as a `devDependency` for its types, and keeps this dev- guarantee by not exposing any reference to these types. This ensures that we're not at risk of users' Truffle installations inadvertently including a full install of Hardhat as well.

How? This shells out to `npx hardhat`! Information about the Hardhat project is requested via `npx hardhat console` and passed along stdio via minified JSON. This ensures that Truffle's [likely polluted] Node.js runtime does not cause any conflicts with Hardhat's own [likely polluted] Node.js runtime.

These are necessary because neither Truffle nor Hardhat were really built to be library-first, and it seems prudent to treat Hardhat like a regular system executable on the PATH. (Well, `npx` is the executable on the PATH, but Hardhat specifically only supports running via `npx`)

### DIFF
--- a/packages/from-hardhat/package.json
+++ b/packages/from-hardhat/package.json
@@ -1,0 +1,47 @@
+{
+  "name": "@truffle/from-hardhat",
+  "description": "Import Hardhat project information into Truffle-native formats",
+  "license": "MIT",
+  "author": "g. nicholas d'andrea <gnidan@trufflesuite.com>",
+  "homepage": "https://github.com/trufflesuite/truffle/tree/master/packages/from-hardhat#readme",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/trufflesuite/truffle.git",
+    "directory": "packages/from-hardhat"
+  },
+  "bugs": {
+    "url": "https://github.com/trufflesuite/truffle/issues"
+  },
+  "version": "0.1.0-0",
+  "main": "dist/src/index.js",
+  "scripts": {
+    "build": "ttsc",
+    "prepare": "yarn build",
+    "test": "exit 0"
+  },
+  "dependencies": {
+    "@truffle/compile-common": "^0.7.32",
+    "@truffle/compile-solidity": "^6.0.38",
+    "@truffle/config": "^1.3.34",
+    "debug": "^4.3.1",
+    "find-up": "^2.1.0",
+    "semver": "^5.7.1"
+  },
+  "devDependencies": {
+    "@types/find-up": "^2.1.0",
+    "@types/node": "^18.6.5",
+    "hardhat": "^2.10.1",
+    "ttypescript": "1.5.13",
+    "typescript": "^4.3.5",
+    "typescript-transform-paths": "3.3.1"
+  },
+  "keywords": [
+    "ethereum",
+    "solidity",
+    "truffle",
+    "hardhat"
+  ],
+  "publishConfig": {
+    "access": "public"
+  }
+}
--- a/packages/from-hardhat/src/api.ts
+++ b/packages/from-hardhat/src/api.ts
@@ -5,7 +5,10 @@ import type * as Hardhat from "hardhat/types";
 import type * as Common from "@truffle/compile-common";
 import type TruffleConfig from "@truffle/config";
 
-import { supportedHardhatVersionRange } from "./constants";
+import {
+  supportedHardhatVersionRange,
+  supportedHardhatBuildInfoFormats
+} from "./constants";
 import * as Compilation from "./compilation";
 import * as Config from "./config";
 
@@ -38,7 +41,7 @@ export const expectHardhat = async (
   const hardhatVersion = await askHardhatVersion(options);
 
   if (!semver.satisfies(hardhatVersion, supportedHardhatVersionRange)) {
-    throw new IncompatibleHardhatError(hardhatVersion);
+    throw new IncompatibleHardhatVersionError(hardhatVersion);
   }
 };
 
@@ -54,7 +57,7 @@ export class NotHardhatError extends Error {
 /**
  * Thrown when Hardhat was detected but with an incompatible version
  */
-export class IncompatibleHardhatError extends Error {
+export class IncompatibleHardhatVersionError extends Error {
   constructor(detectedVersion: string) {
     super(
       `Expected Hardhat version compatible with ${supportedHardhatVersionRange}, got: ${detectedVersion}`
@@ -133,6 +136,12 @@ export const prepareCompilations = async (
       (await fs.readFile(buildInfoPath)).toString()
     );
 
+    const { _format } = buildInfo;
+
+    if (!supportedHardhatBuildInfoFormats.has(_format)) {
+      throw new IncompatibleHardhatBuildInfoFormatError(_format);
+    }
+
     const compilation = Compilation.fromBuildInfo(buildInfo);
 
     compilations.push(compilation);
@@ -140,3 +149,16 @@ export const prepareCompilations = async (
 
   return compilations;
 };
+
+/**
+ * Thrown when the build-info format detected has an incompatible version
+ */
+export class IncompatibleHardhatBuildInfoFormatError extends Error {
+  constructor(detectedFormat: string) {
+    super(
+      `Expected build-info to be one of ["${[
+        ...supportedHardhatBuildInfoFormats
+      ].join('", "')}"], got: "${detectedFormat}"`
+    );
+  }
+}
--- a/packages/from-hardhat/src/api.ts
+++ b/packages/from-hardhat/src/api.ts
@@ -1,0 +1,142 @@
+import { promises as fs } from "fs";
+import semver from "semver";
+
+import type * as Hardhat from "hardhat/types";
+import type * as Common from "@truffle/compile-common";
+import type TruffleConfig from "@truffle/config";
+
+import { supportedHardhatVersionRange } from "./constants";
+import * as Compilation from "./compilation";
+import * as Config from "./config";
+
+import {
+  checkHardhat,
+  askHardhatConsole,
+  askHardhatVersion
+} from "./ask-hardhat";
+import { EnvironmentOptions } from "./options";
+
+/**
+ * Checks for the existence of a Hardhat project configuration and asserts
+ * that the local installed version of Hardhat matches this package's
+ * supported version range.
+ *
+ * @param options to control process environment (e.g. working directory)
+ * @return Promise<void> when expectation holds
+ * @throws NotHardhatError when not in a Hardhat project directory
+ * @throws IncompatibleHardhatError if Hardhat has unsupported version
+ */
+export const expectHardhat = async (
+  options?: EnvironmentOptions
+): Promise<void> => {
+  const isHardhat = await checkHardhat(options);
+
+  if (!isHardhat) {
+    throw new NotHardhatError();
+  }
+
+  const hardhatVersion = await askHardhatVersion(options);
+
+  if (!semver.satisfies(hardhatVersion, supportedHardhatVersionRange)) {
+    throw new IncompatibleHardhatError(hardhatVersion);
+  }
+};
+
+/**
+ * Thrown when no Hardhat project is found
+ */
+export class NotHardhatError extends Error {
+  constructor() {
+    super("Current working directory is not part of a Hardhat project");
+  }
+}
+
+/**
+ * Thrown when Hardhat was detected but with an incompatible version
+ */
+export class IncompatibleHardhatError extends Error {
+  constructor(detectedVersion: string) {
+    super(
+      `Expected Hardhat version compatible with ${supportedHardhatVersionRange}, got: ${detectedVersion}`
+    );
+  }
+}
+
+/**
+ * Constructs a @truffle/config object based on the Hardhat config.
+ *
+ * WARNING: except for fields documented here, the values present on the
+ * returned @truffle/config object MUST be regarded as unsafe to use.
+ *
+ * The returned `config` is defined to contain the following:
+ *
+ *   - `config.networks` with configurations for all Hardhat-configured
+ *     networks, provided:
+ *       - The configured network is not the built-in `hardhat` network
+ *       - The configured network defines a `url` property
+ *
+ *     Note: this function ignores all properties other than `url`,
+ *     including any information that can be used for computing
+ *     cryptographic signatures. THIS FUNCTION DOES NOT READ PRIVATE KEYS.
+ *
+ * Suffice to say:
+ *
+ * THIS FUNCTION'S BEHAVIOR IS EXPERIMENTAL AND SHOULD ONLY BE USED IN
+ * SPECIFICALLY KNOWN-SUPPORTED USE CASES (like reading for configured
+ * network urls)
+ *
+ * @param options to control process environment (e.g. working directory)
+ * @return Promise<TruffleConfig>
+ *
+ * @dev This function shells out to `npx hardhat console` to ask the Hardhat
+ *      runtime environment for a fully populated config object.
+ */
+export const prepareConfig = async (
+  options?: EnvironmentOptions
+): Promise<TruffleConfig> => {
+  const hardhatConfig = (await askHardhatConsole(
+    `hre.config`,
+    options
+  )) as Hardhat.HardhatConfig;
+
+  return Config.fromHardhatConfig(hardhatConfig);
+};
+
+/**
+ * Constructs an array of @truffle/compile-common `Compilation` objects
+ * corresponding one-to-one with Hardhat's persisted results of each solc
+ * compilation.
+ *
+ * WARNING: this function only supports Hardhat projects written entirely
+ * in solc-compatible languages (Solidity, Yul). Behavior of this function
+ * for Hardhat projects using other languages is undefined.
+ *
+ * @param options to control process environment (e.g. working directory)
+ * @return Promise<Compilation[]> from @truffle/compile-common
+ *
+ * @dev This function shells out to `npx hardhat console` to ask the Hardhat
+ *      runtime environment for the location of the project build info
+ *      files
+ */
+export const prepareCompilations = async (
+  options?: EnvironmentOptions
+): Promise<Common.Compilation[]> => {
+  const compilations = [];
+
+  const buildInfoPaths = (await askHardhatConsole(
+    `artifacts.getBuildInfoPaths()`,
+    options
+  )) as string[];
+
+  for (const buildInfoPath of buildInfoPaths) {
+    const buildInfo: Hardhat.BuildInfo = JSON.parse(
+      (await fs.readFile(buildInfoPath)).toString()
+    );
+
+    const compilation = Compilation.fromBuildInfo(buildInfo);
+
+    compilations.push(compilation);
+  }
+
+  return compilations;
+};
--- a/packages/from-hardhat/src/ask-hardhat.ts
+++ b/packages/from-hardhat/src/ask-hardhat.ts
@@ -1,0 +1,108 @@
+import { spawn } from "child_process";
+
+import findUp from "find-up";
+
+import { validHardhatConfigFilenames } from "./constants";
+import { EnvironmentOptions, withDefaultEnvironmentOptions } from "./options";
+
+/**
+ * Returns a Promise to a boolean that is true if and only if
+ * the detected or specified environment is part of a Hardhat project.
+ *
+ * (i.e., if the working directory or any of its parents has a Hardhat config)
+ */
+export const checkHardhat = async (
+  options?: EnvironmentOptions
+): Promise<boolean> => {
+  const { workingDirectory } = withDefaultEnvironmentOptions(options);
+
+  // search recursively up for a hardhat config
+  const hardhatConfigPath = await findUp(validHardhatConfigFilenames, {
+    cwd: workingDirectory
+  });
+
+  return !!hardhatConfigPath;
+};
+
+/**
+ * Reads version information via `npx hardhat --version`
+ */
+export const askHardhatVersion = async (
+  options?: EnvironmentOptions
+): Promise<string> =>
+  new Promise((accept, reject) => {
+    const { workingDirectory } = withDefaultEnvironmentOptions(options);
+
+    const hardhat = spawn(`npx`, ["hardhat", "--version"], {
+      stdio: "pipe",
+      cwd: workingDirectory
+    });
+
+    let output = "";
+    hardhat.stdout.on("data", data => {
+      output = `${output}${data}`;
+    });
+
+    // setup close event before writing to stdin because we're sending eof
+    hardhat.on("close", code => {
+      if (code !== 0) {
+        return reject(new Error(`Hardhat exited with non-zero code ${code}`));
+      }
+
+      return accept(output);
+    });
+  });
+
+export interface AskHardhatConsoleOptions {
+  // turn off json stringify/parse
+  raw?: boolean;
+}
+
+export const askHardhatConsole = async (
+  expression: string,
+  {
+    raw = false,
+    ...options
+  }: AskHardhatConsoleOptions & EnvironmentOptions = {}
+): Promise<string | unknown> =>
+  new Promise((accept, reject) => {
+    const { workingDirectory } = withDefaultEnvironmentOptions(options);
+
+    const hardhat = spawn(`npx`, ["hardhat", "console"], {
+      stdio: ["pipe", "pipe", "inherit"],
+      cwd: workingDirectory
+    });
+
+    // we'll capture the stdout
+    let output = "";
+    hardhat.stdout.on("data", data => {
+      output = `${output}${data}`;
+    });
+
+    // setup close event before writing to stdin because we're sending eof
+    hardhat.on("close", code => {
+      if (code !== 0) {
+        return reject(new Error(`Hardhat exited with non-zero code ${code}`));
+      }
+
+      if (raw) {
+        return accept(output);
+      }
+
+      try {
+        return accept(JSON.parse(output));
+      } catch (error) {
+        return reject(error);
+      }
+    });
+
+    hardhat.stdin.write(`
+      Promise.resolve(${expression})
+        .then(${
+          raw
+            ? `console.log`
+            : `(resolved) => console.log(JSON.stringify(resolved))`
+        })
+    `);
+    hardhat.stdin.end();
+  });
--- a/packages/from-hardhat/src/ask-hardhat.ts
+++ b/packages/from-hardhat/src/ask-hardhat.ts
@@ -43,7 +43,7 @@ export const askHardhatVersion = async (
       output = `${output}${data}`;
     });
 
-    hardhat.on("close", code => {
+    hardhat.once("close", code => {
       if (code !== 0) {
         return reject(new Error(`Hardhat exited with non-zero code ${code}`));
       }
@@ -79,7 +79,7 @@ export const askHardhatConsole = async (
     });
 
     // setup close event before writing to stdin because we're sending eof
-    hardhat.on("close", code => {
+    hardhat.once("close", code => {
       if (code !== 0) {
         return reject(new Error(`Hardhat exited with non-zero code ${code}`));
       }
--- a/packages/from-hardhat/src/ask-hardhat.ts
+++ b/packages/from-hardhat/src/ask-hardhat.ts
@@ -34,7 +34,7 @@ export const askHardhatVersion = async (
     const { workingDirectory } = withDefaultEnvironmentOptions(options);
 
     const hardhat = spawn(`npx`, ["hardhat", "--version"], {
-      stdio: "pipe",
+      stdio: ["pipe", "pipe", "inherit"],
       cwd: workingDirectory
     });
 
@@ -43,7 +43,6 @@ export const askHardhatVersion = async (
       output = `${output}${data}`;
     });
 
-    // setup close event before writing to stdin because we're sending eof
     hardhat.on("close", code => {
       if (code !== 0) {
         return reject(new Error(`Hardhat exited with non-zero code ${code}`));
--- a/packages/from-hardhat/src/compilation.ts
+++ b/packages/from-hardhat/src/compilation.ts
@@ -1,0 +1,121 @@
+import type * as Hardhat from "hardhat/types";
+
+import type { Compilation, CompiledContract } from "@truffle/compile-common";
+import * as CompileSolidity from "@truffle/compile-solidity";
+
+import { supportedHardhatBuildInfoFormats } from "./constants";
+
+export const fromBuildInfo = (buildInfo: Hardhat.BuildInfo): Compilation => {
+  const { _format } = buildInfo;
+
+  if (!supportedHardhatBuildInfoFormats.has(_format)) {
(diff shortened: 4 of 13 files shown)


packages/from-hardhat/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "@truffle/compile-common": "^0.7.32",
+    "@truffle/compile-solidity": "^6.0.38",
+    "@truffle/config": "^1.3.34",
+    "debug": "^4.3.1",
+    "find-up": "^2.1.0"
+  },
+  "devDependencies": {
+    "@types/find-up": "^2.1.0",
+    "@types/node": "^18.6.5",
+    "hardhat": "^2.10.1",
```
gnidan, line 32: @cds-amal @fainashalts FYI I realized that we can benefit from Hardhat's own types by keeping this as a devDependency safely... if we just don't actually export anything with these types in the signature. This is the approach I'm taking here.

Just letting you know because this breaks the assumptions of our earlier conversation.

gnidan:
cc @alcuadrado this @truffle/from-hardhat PR is now ready for review - I believe I handled the concerns you stated in #5410, but would love if you could take another look! Thank you!

packages/from-hardhat/package.json
benjamincburns: how does this compare to making it an optional peerDepenency?

gnidan: > how does this compare to making it an optional peerDepenency?

Good question! Firstly, it's worth noting that [`package.json`'s `peerDependenciesMeta` field](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#peerdependenciesmeta) (which is how you mark peerDeps as optional) was added in NPM v7, and we need to support v6 still. But this is a minor quibble.

The main concern here, IMO, is the question "what does Truffle's Node.js-land need from the `"hardhat"` NPM package?"
- Types means _at least_ a `devDep`
- We need more than `devDep` if-and-only-if we want to have a shared Node.js runtime between Truffle and Hardhat... as in, iff we want to use Hardhat as a library

I looked at using HH as a library, but, like Truffle, it's really not intended for use via straight `require()` (or `import`). I got it working in #5410... so this approach is viable albeit awkward. But think about what we're saying when Truffle does `require("hardhat")`... like, _"here you go, Node! have another, different one of **me**!"_. Even without the risk of cross-clobbering each other's globals / other runtime modification shenanigans, the memory footprint and whatnot... oy!

So my conclusion was basically... screw `peerDependencies`, HH is prescriptive: you use it via `npx hardhat`. Truffle can use HH via `npx hardhat`. Keep the playgrounds separate :)

gnidan: Probably worth following up that there is a counter-argument that we can just use `peerDependencies` for the `$(npm bin)` or whatever. But still... I think there's value in treating HH as an executable... like, you're not going to try to shove Docker into your `package.json` - you just shell out to `docker` and error nicely if it's not there.

packages/from-hardhat/src/compilation.ts
```diff
@@ -0,0 +1,121 @@
...
+    sources: Sources.fromBuildInfo(buildInfo, sourceIndexes),
+    contracts: Contracts.fromBuildInfo(buildInfo),
+    compiler: {
+      name: "solc",
+      version: buildInfo.solcLongVersion
+    }
+  };
+};
+
+namespace SourceIndexes {
```
benjamincburns, line 28: Hmm, so I'm not sure how I feel about the use of namespaces in TS code. I generally regard them as strictly for the purpose of writing type declarations for pure JS modules that do complicated things that would otherwise be inexpressible via the TS type system.

My main concern with them is pretty minor, however. Basically it just feels a bit odd reading `SourceIndexes.fromBuildInfo` in the code above, then discovering that it's not a static method on a class, but rather a namespaced local function.

I think also given that the namespaces here aren't exported, it _really_ is quite a minor matter of stylistic opinion.

packages/from-hardhat/src/compilation.ts
```diff
@@ -0,0 +1,121 @@
...
+namespace Sources {
+  export const fromBuildInfo = (
+    buildInfo: Hardhat.BuildInfo,
+    sourceIndexes: Compilation["sourceIndexes"]
+  ): Compilation["sources"] =>
+    sourceIndexes.map(sourcePath => {
+      // to handle if sourceIndexes is a sparse array
+      if (!sourcePath) {
+        return;
+      }
```
benjamincburns, line 52: Doesn't this also return a sparse array with an `undefined` at this index? Maybe you want to remove sparse indices with a `.filter` ahead of the map? I think you could also avoid the extra `O(n)` pass of a `filter` line by using a `reduce` that only accumulates defined `sourcePath` values into an array.

Personally I tend to find the `filter` approach more readable, however.

packages/from-hardhat/src/compilation.ts
```diff
@@ -0,0 +1,121 @@
...
+  export const fromBuildInfo = (
+    buildInfo: Hardhat.BuildInfo
+  ): CompiledContract[] => {
+    const contracts = [];
+    for (const [sourcePath, sourceContracts] of Object.entries(
+      buildInfo.output.contracts
+    )) {
+      for (const [contractName, compilerOutputContract] of Object.entries(
+        sourceContracts
+      )) {
```
benjamincburns, line 76: Are we not worried about sparse arrays here?

Also if you want to remain consistent with the functional style above, you could do something like 

```typescript
Object.entries(buildInfo.output.contracts).map([sourcePath, sourceContracts] => 
  Object.entries(sourceContracts).map([contractName, compilerOutputContract] => return { 
    /* contents of CompiledContract here */
  }).flat();
```

packages/from-hardhat/src/compilation.ts
```diff
@@ -0,0 +1,121 @@
...
+namespace SourceIndexes {
+  export const fromBuildInfo = (
+    buildInfo: Hardhat.BuildInfo
+  ): Compilation["sourceIndexes"] => {
+    const sourceIndexes = [];
+    for (const { index, sourcePath } of Object.entries(
+      buildInfo.output.sources
+    ).map(([sourcePath, source]) => ({ index: source.id, sourcePath }))) {
+      sourceIndexes[index] = sourcePath;
+    }
```
benjamincburns, line 37: Same question about sparse arrays here as well

packages/from-hardhat/src/compilation.ts
benjamincburns: Though that assumes that we aren't working with sparse arrays.

packages/from-hardhat/src/api.ts
```diff
@@ -0,0 +1,142 @@
...
+ * @return Promise<TruffleConfig>
+ *
+ * @dev This function shells out to `npx hardhat console` to ask the Hardhat
+ *      runtime environment for a fully populated config object.
+ */
+export const prepareConfig = async (
+  options?: EnvironmentOptions
+): Promise<TruffleConfig> => {
+  const hardhatConfig = (await askHardhatConsole(
+    `hre.config`,
```
gnidan, line 98: Worth noting about this approach: this only works because `hre.config` is serializable.

I am so grateful and so jealous that they made this serializable :)

packages/from-hardhat/tsconfig.json
```diff
@@ -0,0 +1,39 @@
+{
+  "compilerOptions": {
```
benjamincburns, line 2: Do we want things like `strict` and `noImplicitAny` in here? Excluding now it is going to make it rather complicated to standardize on it in the future if we move to TypeScript project builds w/ a single inherited base config.

packages/from-hardhat/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@truffle/compile-solidity": "^6.0.38",
+    "@truffle/config": "^1.3.34",
+    "debug": "^4.3.1",
+    "find-up": "^2.1.0",
+    "semver": "^5.7.1"
+  },
+  "devDependencies": {
+    "@types/find-up": "^2.1.0",
+    "@types/node": "^18.6.5",
+    "hardhat": "^2.10.1",
```
benjamincburns, line 33: I asked elsewhere, but should we consider making this an optional peer dependency?

packages/from-hardhat/src/config.ts
```diff
@@ -0,0 +1,36 @@
...
+export const fromHardhatConfig = (hardhatConfig: HardhatConfig): Config => {
+  return Config.default().merge({
+    networks: Networks.fromHardhatConfig(hardhatConfig)
+  });
+};
+
+namespace Networks {
+  export const fromHardhatConfig = (hardhatConfig: HardhatConfig) =>
+    Object.entries(hardhatConfig.networks)
+      .flatMap(([networkName, networkConfig]) => {
```
benjamincburns, line 14: Oh hey, TIL `flatMap` is a thing. Just pretend that I used it in my example code above :-p

packages/from-hardhat/src/options.ts
```diff
@@ -0,0 +1,14 @@
...
+   * @default process.cwd()
+   */
+  workingDirectory?: string;
+}
+
+export const withDefaultEnvironmentOptions = ({
+  workingDirectory = process.cwd()
+}: EnvironmentOptions = {}): EnvironmentOptions => ({
+  workingDirectory
+});
```
benjamincburns, line 14: I notice that you seem to prefer arrow function definitions. Why not use the `function withDefaultEnvironmentOptions(...) { ... }` syntax?

Personally I find plain old functions to be slightly more readable (less boilerplate) and they make it easier to inspect function references in a debugger, as you can often check their `name` property to find out where they originated from.

Curious to know what the benefits of using arrow functions like this might be. Perhaps it's that they break the global `this` binding?

packages/from-hardhat/src/config.ts
```diff
@@ -0,0 +1,36 @@
...
+        return [
+          {
+            [networkName]: {
+              url,
+              network_id: "*"
+            }
+          }
+        ];
+      })
+      .reduce((a, b) => ({ ...a, ...b }), {});
```
benjamincburns, line 35: I wonder if there's some more readable way to write this line specifically? I was initially thinking maybe `Object.assign` with a spread operator, but I dunno, the function chaining does kinda win out here over nested calls, IMO.

benjamincburns (changes requested):
Requested changes are mostly a bunch of small things, with a much larger pile of questions.

Also I do think it would be nice if this had tests, but as discussed on our 1:1, I'd agree that writing _fast_ tests for this could take some nontrivial test engineering effort, so I'd be happy enough to agree that tests could be done in another PR. That said, if you're likely going to put this down for a while before returning to it, perhaps we should capture a ticket to that effect.

packages/from-hardhat/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@truffle/from-hardhat",
+  "description": "Import Hardhat project information into Truffle-native formats",
+  "license": "MIT",
+  "author": "g. nicholas d'andrea <gnidan@trufflesuite.com>",
```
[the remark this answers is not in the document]
gnidan: I just put my name whenever I realize I've contributed serious authorship to some package. Either if I'm first to make the package, or if I rewrite it heavily.

I don't have strong feelings about it. I could see why we'd want to move away from that.

packages/from-hardhat/package.json
benjamincburns: Ah that makes sense, and feel free to ignore my repeat question in the review below.

I was just thinking that it'd be very nice if there was some way to import _just_ the types from the hardhat package without needing to add all of their transitive dependencies into our `node_modules` tree and our `yarn.lock` (even if it is just for dev purposes).

If our usage of their types is minimal, perhaps it'd be worth extracting _only_ the types that we need into a local type declaration? Though I could also easily argue against that approach due to the potential effort involved in doing that, along with the increased maintenance burden that will be required.

Sure would be nice if there was a code-gen utility for that sort of thing, though!

packages/from-hardhat/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "typescript-transform-paths": "3.3.1"
+  },
+  "keywords": [
+    "ethereum",
+    "solidity",
+    "truffle",
+    "hardhat"
+  ],
+  "publishConfig": {
+    "access": "public"
```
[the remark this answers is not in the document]
gnidan: You're asking a lot of questions that are bigger than this PR ;)

packages/from-hardhat/src/api.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  if (!semver.satisfies(hardhatVersion, supportedHardhatVersionRange)) {
+    throw new IncompatibleHardhatError(hardhatVersion);
+  }
+};
+
+/**
+ * Thrown when no Hardhat project is found
+ */
+export class NotHardhatError extends Error {
```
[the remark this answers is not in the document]
gnidan: Oh yeah I meant to note that I extended `Error` with the plan that it will be later chained via your #5367 

Seems better to do that vs. import @truffle/error here.

packages/from-hardhat/src/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const validHardhatConfigFilenames = [
+  "hardhat.config.js",
+  "hardhat.config.ts"
+];
+export const supportedHardhatVersionRange = "^2.10.1";
+export const supportedHardhatBuildInfoFormats = new Set([
+  "hh-sol-build-info-1"
+]);
```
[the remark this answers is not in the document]
gnidan: I was torn on this, too. Normally I'm a staunch believer of not doing this sort of thing, but here I wanted it to be visible and reduce the changes of forgetting to mark the version bump or whatever.

I'll consider this for a bit (and probably switch it.)

packages/from-hardhat/src/ask-hardhat.ts
```diff
@@ -0,0 +1,108 @@
...
+      cwd: workingDirectory
+    });
+
+    let output = "";
+    hardhat.stdout.on("data", data => {
+      output = `${output}${data}`;
+    });
+
+    // setup close event before writing to stdin because we're sending eof
+    hardhat.on("close", code => {
```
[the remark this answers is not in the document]
gnidan: Hm probably. I'll try it!

packages/from-hardhat/src/api.ts
benjamincburns: Part of the struggle of forging our own path in the early days of the Ethereum ecosystem, I suppose. Also the node ecosystem and what is considered best practice has advanced just a _touch_ since 2015!

Like, I think at the time that the original `truffle-config.js` was designed executable config was the hotness, as it was (and still is) super flexible. Now it's somewhat frowned upon for multi-modal projects like ours due this exact concern.

packages/from-hardhat/src/ask-hardhat.ts
```diff
@@ -0,0 +1,108 @@
...
+  {
+    raw = false,
+    ...options
+  }: AskHardhatConsoleOptions & EnvironmentOptions = {}
+): Promise<string | unknown> =>
+  new Promise((accept, reject) => {
+    const { workingDirectory } = withDefaultEnvironmentOptions(options);
+
+    const hardhat = spawn(`npx`, ["hardhat", "console"], {
+      stdio: ["pipe", "pipe", "inherit"],
```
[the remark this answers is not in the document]
gnidan: For an experimental release, I think it's better to reveal these kinds of underpinnings... otherwise it's like magic, but it's... broken magic?

packages/from-hardhat/src/ask-hardhat.ts
```diff
@@ -0,0 +1,108 @@
...
+    // we'll capture the stdout
+    let output = "";
+    hardhat.stdout.on("data", data => {
+      output = `${output}${data}`;
+    });
+
+    // setup close event before writing to stdin because we're sending eof
+    hardhat.on("close", code => {
+      if (code !== 0) {
+        return reject(new Error(`Hardhat exited with non-zero code ${code}`));
```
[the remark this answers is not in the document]
gnidan: Ah yeah that's neat! Certainly adds work though to make it bubble up nice. Seems like it could just be a follow-on task?

(Also see my other comment about why I think @truffle/core should just use your work in #5367 to connect these errors to TruffleError)

packages/from-hardhat/src/compilation.ts
```diff
@@ -0,0 +1,121 @@
...
+import type { Compilation, CompiledContract } from "@truffle/compile-common";
+import * as CompileSolidity from "@truffle/compile-solidity";
+
+import { supportedHardhatBuildInfoFormats } from "./constants";
+
+export const fromBuildInfo = (buildInfo: Hardhat.BuildInfo): Compilation => {
+  const { _format } = buildInfo;
+
+  if (!supportedHardhatBuildInfoFormats.has(_format)) {
+    throw new Error(`Unsupported build info format: ${_format}`);
```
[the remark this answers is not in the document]
gnidan: Oh yeah forgot a class for the errors in this file. I'll add them!

packages/from-hardhat/src/ask-hardhat.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    let output = "";
+    hardhat.stdout.on("data", data => {
+      output = `${output}${data}`;
+    });
+
+    // setup close event before writing to stdin because we're sending eof
+    hardhat.on("close", code => {
+      if (code !== 0) {
+        return reject(new Error(`Hardhat exited with non-zero code ${code}`));
```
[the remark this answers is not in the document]
gnidan: Will also make this one a class.

packages/from-hardhat/src/ask-hardhat.ts
```diff
@@ -0,0 +1,108 @@
...
+      }
+
+      if (raw) {
+        return accept(output);
+      }
+
+      try {
+        return accept(JSON.parse(output));
+      } catch (error) {
+        return reject(error);
```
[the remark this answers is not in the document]
gnidan: It's definitely an internal error here... the idea is that we're going to ask HH for some JSON, and if it spits out invalid JSON, that's not a known failure mode. I'm okay with a vague panic here until someone complains.

packages/from-hardhat/src/compilation.ts
gnidan: Yeah I did this only for stylistic reasons (I like the function name consistency `fromBuildInfo`, and I like that it means I can use `export const fromBuildInfo = () => new Promise(/* ... */)`)

packages/from-hardhat/src/compilation.ts
gnidan: Actually this was probably just overly defensive programming. I realized afterwards that a lot of stuff would break if this was actually sparse.

gnidan: But it is more important to get `sourceIndexes` correct, and `sources` could be in any order

packages/from-hardhat/src/compilation.ts
gnidan: Yeah I should just remove the sparseness support.

(I added it because @truffle/db allows everything to be sparse, and I had forgotten that that wasn't true of compile-common's spec)

packages/from-hardhat/tsconfig.json
gnidan: I had to turn `strictNullChecks` off because the types in compile-common aren't suitable for use with that flag

(@eggplantzzz is looking into fixing that)

gnidan: I don't care about `noImplicitAny`. Happy to add it if you think it's useful.

packages/from-hardhat/package.json
gnidan: (See conversation above)

packages/from-hardhat/src/options.ts
gnidan: I literally only did it because I don't like `function() { return new Promise() }` as much as `() => new Promise()`

packages/from-hardhat/src/config.ts
gnidan: `Object.assign()` is just ugly... this is my favorite way to do this, but nowadays maybe there's other sugar to consider? Happy to take suggestions.

packages/from-hardhat/package.json
gnidan: Hm, I see your concern. Not sure copy+pasting the types is acceptable? Sounds like follow-on work? Or maybe we could get HH to publish types separately?

packages/from-hardhat/src/config.ts
gnidan: `flatMap` is great

packages/from-hardhat/package.json
```diff
@@ -0,0 +1,47 @@
...
+  },
+  "bugs": {
+    "url": "https://github.com/trufflesuite/truffle/issues"
+  },
+  "version": "0.1.0-0",
+  "main": "dist/src/index.js",
+  "scripts": {
+    "build": "ttsc",
+    "prepare": "yarn build",
+    "test": "exit 0"
```
gnidan, line 20: As per @benjamincburns's [review comment](https://github.com/trufflesuite/truffle/pull/5420#pullrequestreview-1073370698), we need to open an issue to add tests to this package.

packages/from-hardhat/package.json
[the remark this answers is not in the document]
gnidan: Honestly though I think `prepareCompilations` is a fine export that people might want to have access to, even in this experimental state. Like, keeping this public means it's super easy to run @truffle/decoder inside a HH script!

packages/from-hardhat/src/api.ts
gnidan: There's trade-offs... like, I also had the thought about how easy it is to read private keys from memory. Our approach clamps that down, since we only store mnemonics inside closures.

packages/from-hardhat/package.json
benjamincburns: Yeah, I think _if_ we make changes to this approach, it's definitely something that's a future concern.

packages/from-hardhat/src/constants.ts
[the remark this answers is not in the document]
benjamincburns: I wouldn't get too hung up on it, as I think either way is totally defensible.

packages/from-hardhat/src/api.ts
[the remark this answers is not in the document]
gnidan: Wait, error chaining requires that you extend @truffle/error?

I thought the idea was "throw whatever error you like and @truffle/core will decide if it should be wrapped as TruffleError"

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
benjamincburns: Yeah, makes sense as follow-on

packages/from-hardhat/src/compilation.ts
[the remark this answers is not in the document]
benjamincburns: Feel free to resolve these when you push that change.

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
benjamincburns: I think my main concern on this one is that a user seeing this error will just see a JSON parsing issue without much additional context and assume Truffle is just broken.

That said, it's an experimental feature, so I'm totally fine with this being reconsidered during a later "polishing" phase.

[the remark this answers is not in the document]
gnidan: Good point!

packages/from-hardhat/src/compilation.ts
benjamincburns: I suppose it might be better in that case to just remove this block and the associated comment.

packages/from-hardhat/tsconfig.json
benjamincburns: I tend to think that `noImplicitAny` is useful, as otherwise it's hard to tell when a type is correctly inferred, vs implicitly `any`.

packages/from-hardhat/src/config.ts
benjamincburns: Nah don't worry about it. I was just wondering aloud here.

packages/from-hardhat/package.json
benjamincburns: #5442

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
benjamincburns: Yeah fair enough - we'll certainly learn more about its failure modes that way.

packages/from-hardhat/src/api.ts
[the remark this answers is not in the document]
benjamincburns: FWIW, extending from `TruffleError` makes my work in #5367 substantially easier. This is because I've overridden the `stack` property on `TruffleError` in that PR to spit out stack traces for the underlying causes in addition to the top-level one, as personally I find this to be the most useful reason to chain errors. Unfortunately we don't get that unless we inherit from `TruffleError`, as the standard `stack` property doesn't include the traces from the underlying causes.

That said, if you'd rather not do that now, I'll be happy to rebase on `develop` when this merges and I can handle it then.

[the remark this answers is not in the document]
benjamincburns: > Wait, error chaining requires that you extend @truffle/error?
> 
> I thought the idea was "throw whatever error you like and @truffle/core will decide if it should be wrapped as TruffleError"... like, that neatly makes error display a concern of @truffle/core, no?

Nah, that's a concern that's separate from error chaining. Error chaining allows you to catch one error, throw another, but retain a reference to the error that was the original underlying cause of the one that you're throwing. It makes wrapped errors much easier to source back to their root cause.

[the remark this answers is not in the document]
benjamincburns: So like

```
} catch(err) {
  throw new WrappedFooError("Foo failed real bad, guys", { cause: err });
}
```

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
benjamincburns: Not going to capture an issue for this one, as I think we're almost certain to revisit it as we polish this up for a less-experimental release.

packages/from-hardhat/src/compilation.ts
benjamincburns: I don't think it's at all necessary to "fix" this, as it's not really an issue. Feel free to resolve immediately if you don't wanna fix it, or resolve after pushing if you do.

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
gnidan: Decided against custom classes in this module because I'd rather not have to expose it at all right now. Let's leave this as follow-on work?

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
gnidan: Looking at this, I'll save this one for later since error-handling there requires moving stuff around and ensuring we export errors and whatnot.

packages/from-hardhat/src/compilation.ts
gnidan: Resolving :)

packages/from-hardhat/tsconfig.json
gnidan: Nice, I love turning on a thing and having no added type-checks. Will push shortly :)

packages/from-hardhat/src/compilation.ts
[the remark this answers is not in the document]
gnidan: Done!

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
gnidan: done!

packages/from-hardhat/src/api.ts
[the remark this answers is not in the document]
gnidan: I'll leave the error class hierarchy stuff for you with #5367 ;)

packages/from-hardhat/package.json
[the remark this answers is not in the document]
gnidan: We can debate the `author` field elsewhere ;)

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
benjamincburns: yeah, fine by me, especially since we expect this to be a very rare case.

packages/from-hardhat/package.json
benjamincburns: Ah, I resolved this, but realized that neither of the two people who you pinged at the start have replied, so I've unresolved it.

benjamincburns: Otherwise please consider my part in this resolved.

packages/from-hardhat/src/ask-hardhat.ts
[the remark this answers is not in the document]
benjamincburns: sure - I may or may not pick it up in the context of #5367, depending on how far I go with that.

packages/from-hardhat/src/options.ts
benjamincburns: I think this is fine as-is, I do prefer the plain old `function` syntax, but we do this in _many_ other places in the code, so it's not at all out of line.

benjamincburns (approved):
LGTM!

gnidan:
Thanks @benjamincburns! Will merge when green.

Issue #5421, opened 2022-08-10
Typedoc missing on numerous types in @truffle/db


benjamincburns:
When I run `typedoc@0.22.18` in the `@truffle/db` package, it gives me the following warnings.

Note that these are valid warnings, but to avoid dropping a pile of new warnings into the db build I've disabled these warnings in the PR that updates to this version of `typedoc`. As a result, please be sure to remove the `validation: { notExported: false }` config from the `typedoc` config in `packages/truffle/db/typedoc.js`

```
Warning: CollectionProperty, defined at src/meta/collections.ts:57, is referenced by Meta.IdFields but not included in the documentation.
Warning: FilteredCollectionName, defined at src/meta/collections.ts:112, is referenced by Meta.MutableCollectionName but not included in the documentation.
Warning: ResourceFilter, defined at src/meta/collections.ts:77, is referenced by Meta.Resource but not included in the documentation.
Warning: Collection, defined at src/meta/collections.ts:34, is referenced by Meta.Resource but not included in the documentation.
Warning: Structure, defined at src/meta/batch.ts:162, is referenced by Meta.Batch.Inputs but not included in the documentation.
Warning: IdField, defined at src/meta/id/types.ts:10, is referenced by Meta.Id.StrictIdInput but not included in the documentation.
Warning: CollectionMutation, defined at src/meta/process/types.ts:165, is referenced by Meta.Process.Mutation but not included in the documentation.
Warning: ProcessResponse, defined at src/meta/process/types.ts:88, is referenced by Meta.Process.Process but not included in the documentation.
Warning: GraphQlRequest, defined at src/meta/process/types.ts:65, is referenced by Meta.Process.ProcessRequest but not included in the documentation.
Warning: Web3Request, defined at src/meta/process/types.ts:73, is referenced by Meta.Process.ProcessRequest but not included in the documentation.
Warning: CollectionQuery, defined at src/meta/process/types.ts:132, is referenced by Meta.Process.Query but not included in the documentation.
Warning: GraphQlRequestType, defined at src/meta/process/types.ts:36, is referenced by Meta.Process.RequestType but not included in the documentation.
Warning: Web3RequestType, defined at src/meta/process/types.ts:43, is referenced by Meta.Process.RequestType but not included in the documentation.
Warning: Common.Compilation, defined at ../compile-common/dist/src/types.d.ts:2, is referenced by Project.LoadCompile.Compilation but not included in the documentation.
[...]

PR #5423, opened 2022-08-10
Standardize `@types/node` version and upgrade all TS packages to use `typescript@4.7.2`

benjamincburns:
So the main goal here was to update to latest TypeScript, as that has support for the ES2022 lib, which is required to support error chaining (#5367).

In the process of performing this upgrade, I noticed that even though we have a very well-known minimum supported version of NodeJS, five of the six different versions of `@types/node` upon which we depend were targeted at the wrong node version, and included API features that aren't present in our minimum supported node version.

[Edit: corrected above paragraph]

These changes shipped in two different commits, with each commit being the complete set of changes necessary to deliver the stated goal. That is, each commit contains the obvious change to support the primary motive, and then whatever follow-on changes were necessary to ensure that I wasn't going to break CI.

Please see the commit messages for those two commits for the rationale behind each of the follow-on changes.

As for why this isn't two separate commits, it's primarily because testing this was a lot more effort than I initially expected, and I didn't want to have to potentially triple that effort by having to test these as both independent _and_ integrated changes (which is what would be required for two separate PRs).

### DIFF
--- a/packages/codec/package.json
+++ b/packages/codec/package.json
@@ -46,12 +46,12 @@
     "@types/lodash": "^4.14.179",
     "@types/semver": "^6.0.0",
     "@types/utf8": "^2.1.6",
-    "@zerollup/ts-transform-paths": "^1.7.3",
     "ts-node": "10.7.0",
     "ttypescript": "1.5.13",
-    "typedoc": "0.20.37",
+    "typedoc": "0.22.18",
     "typedoc-plugin-remove-references": "^0.0.5",
-    "typescript": "^4.1.4"
+    "typescript": "^4.7.4",
+    "typescript-transform-paths": "3.3.1"
   },
   "keywords": [
     "abi",
--- a/packages/compile-solidity/src/polyfill.ts
+++ b/packages/compile-solidity/src/polyfill.ts
@@ -1,0 +1,11 @@
+// must polyfill AbortController to use axios >=0.20.0, <=0.27.2 on node <= v14.x
+import { AbortController, AbortSignal } from "node-abort-controller";
+
+// we're using the "dom" lib in this package, so these types are already
+// defined, but their implementations are nonexistent when running under node <=
+// v15.x, so we polyfill them here anyway
+
+if (typeof (global as any).AbortController === "undefined") {
+  (global as any).AbortController = AbortController;
+  (global as any).AbortSignal = AbortSignal;
+}
--- a/packages/dashboard-message-bus-client/lib/polyfill.ts
+++ b/packages/dashboard-message-bus-client/lib/polyfill.ts
@@ -1,0 +1,12 @@
+// must polyfill AbortController to use axios >=0.20.0, <=0.27.2 on node <= v14.x
+import { AbortController, AbortSignal } from "node-abort-controller";
+
+declare global {
+  type AbortController = typeof AbortController;
+  type AbortSignal = typeof AbortSignal;
+}
+
+if (typeof (global as any).AbortController === "undefined") {
+  (global as any).AbortController = AbortController;
+  (global as any).AbortSignal = AbortSignal;
+}
--- a/packages/db/typedoc.js
+++ b/packages/db/typedoc.js
@@ -1,6 +1,4 @@
-const useNeo = false;
-
-const common = {
+module.exports = {
   entryPoints: ["src/index.ts"],
   categorizeByGroup: false,
   categoryOrder: [
@@ -17,44 +15,12 @@ const common = {
     "Internal boilerplate",
     "Internal"
   ],
-  readme: "none",
   includes: "dist",
   media: "docs",
-  out: "dist/docs"
-};
-
-const plugins = [];
-
-// for typedoc-neo-theme
-const neo = {
-  theme: "../../node_modules/typedoc-neo-theme/bin/default",
-  plugin: [...plugins, "typedoc-neo-theme"],
-  outline: [
-    {
-      "Core namespaces": {
-        DataModel: "datamodel",
-        Resources: "resources",
-        GraphQl: "graphql",
-        Process: "process",
-        Batch: "batch"
-      },
-      "Truffle integration": {
-        "Class: Project": "classes/project",
-        "Class: Project.ConnectedProject": "classes/project.connectedproject"
-      },
-      "Meta system": {
-        "Meta": "meta",
-        "Meta.Pouch": "meta.pouch",
-        "Meta.GraphQl": "meta.graphql",
-        "Meta.Process": "meta.process",
-        "Meta.Batch": "meta.batch"
-      }
-    }
-  ]
-};
-
-module.exports = {
-  ...common,
-  plugin: plugins.length > 0 ? plugins : ["none"],
-  ...(useNeo ? neo : {})
+  out: "dist/docs",
+  plugin: [],
+  readme: "none",
+  validation: {
+    notExported: false
+  }
 };
--- a/packages/hdwallet-provider/test/provider.test.ts
+++ b/packages/hdwallet-provider/test/provider.test.ts
@@ -65,7 +65,7 @@ describe("HD Wallet Provider", function () {
           "http://localhost:8545"
         );
         assert.fail("Should throw on invalid mnemonic");
-      } catch (e) {
+      } catch (e: any) {
         assert(e.message.includes("Mnemonic invalid or undefined"));
       }
     });
@@ -314,7 +314,7 @@ describe("HD Wallet Provider", function () {
             provider: { junk: "in", an: "object" }
           });
           assert.fail("Should throw on invalid provider");
-        } catch (e) {
+        } catch (e: any) {
           assert(e.message.includes("invalid provider was specified"));
         }
       });
@@ -329,7 +329,7 @@ describe("HD Wallet Provider", function () {
             url: "justABunchOfJunk"
           });
           assert.fail("Should throw on invalid url");
-        } catch (e) {
+        } catch (e: any) {
           assert(e.message.includes("invalid provider was specified"));
         }
       });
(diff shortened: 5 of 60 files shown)


benjamincburns:
ICYMI in the dense commit messages, re: the new `typedoc` warnings on build of docs for `db` and `db-kit`, I've disabled those successfully now and raised #5421 and #5420 to cover the effort required to re-enable the warnings.

cds-amal:
> In the process of performing this upgrade, I noticed that even though we have a very well-known minimum supported version of NodeJS, exactly zero of the six different versions of `@types/node` upon which we depend were targeted at this version, and all of the versions that we _were_ using included API features that aren't present in our minimum supported node version.
> 

Yikes! Thanks so much for catching this!

> These changes shipped in two different commits, with each commit being the complete set of changes necessary to deliver the stated goal. That is, each commit contains the obvious change to support the primary motive, and then whatever follow-on changes were necessary to ensure that I wasn't going to break CI.
> 
> Please see the commit messages for those two commits for the rationale behind each of the follow-on changes.
> 

This makes sense. FWIW, It's easier to review the distinct, and "complete", commits.

> As for why this isn't two separate commits, it's primarily because testing this was a lot more effort than I initially expected, and I didn't want to have to potentially triple that effort by having to test these as both independent _and_ integrated changes (which is what would be required for two separate PRs).

Strongly agree here.

packages/compile-solidity/src/polyfill.ts
```diff
@@ -0,0 +1,11 @@
+// must polyfill AbortController to use axios >=0.20.0, <=0.27.2 on node <= v14.x
```
cds-amal, line 1: Minor nit: I think the filename should reflect the type of polyfill, like `abort-controller-pollyfill.ts` or something similar, instead of `polyfill.ts`

packages/dashboard-message-bus-client/lib/polyfill.ts
```diff
@@ -0,0 +1,12 @@
+// must polyfill AbortController to use axios >=0.20.0, <=0.27.2 on node <= v14.x
+import { AbortController, AbortSignal } from "node-abort-controller";
```
cds-amal, line 2: Second use; duplicate, grin and bear it  :)

packages/codec/package.json
```diff
@@ -46,12 +46,12 @@
...
     "@types/utf8": "^2.1.6",
-    "@zerollup/ts-transform-paths": "^1.7.3",
     "ts-node": "10.7.0",
     "ttypescript": "1.5.13",
-    "typedoc": "0.20.37",
+    "typedoc": "0.22.18",
     "typedoc-plugin-remove-references": "^0.0.5",
-    "typescript": "^4.1.4"
+    "typescript": "^4.7.4",
+    "typescript-transform-paths": "3.3.1"
```
eggplantzzz, line 54: So this is a valid drop-in for @zerorollup/ts-transform-paths?

packages/source-fetcher/lib/polyfill.ts
```diff
@@ -0,0 +1,12 @@
+// must polyfill AbortController to use axios >=0.20.0, <=0.27.2 on node <= v14.x
+import { AbortController, AbortSignal } from "node-abort-controller";
```
cds-amal, line 2: 3rd time, perhaps consider factoring this out to a utility module? Maybe this warrants a new truffle-package to handle these utility functions? Would be interested to hear what others think.

packages/db/typedoc.js
```diff
@@ -17,44 +15,12 @@ const common = {
...
-};
-
-const plugins = [];
-
-// for typedoc-neo-theme
-const neo = {
-  theme: "../../node_modules/typedoc-neo-theme/bin/default",
-  plugin: [...plugins, "typedoc-neo-theme"],
-  outline: [
-    {
```
cds-amal, line 33: Good riddance!

cds-amal (commented):
This looks good to me! I left one nit and noticed polyfill logic was duplicated.

packages/hdwallet-provider/test/provider.test.ts
```diff
@@ -65,7 +65,7 @@ describe("HD Wallet Provider", function () {
           "http://localhost:8545"
         );
         assert.fail("Should throw on invalid mnemonic");
-      } catch (e) {
+      } catch (e: any) {
```
eggplantzzz, line 68: Do these even need to be typed? They all should be of type `Error`, no?

benjamincburns: IIRC, typed catch handlers are now required when `noImplicitAny` is `true` (as it is in all of our `tsconfig.json`) files.

benjamincburns: Also FWIW, I've verified that if any code changes here were reverted, the build would break.

eggplantzzz: Ok that makes sense. But even so, they should be of type `Error` right?

benjamincburns:
Polyfill logic was duplicated to ensure that it would load properly under each of the different command bundles.

An alternative would have been to make webpack handle injecting the polyfill, but then it wouldn't be present on the unpacked build.

packages/hdwallet-provider/test/provider.test.ts
benjamincburns: No, there is no requirement to throw an object of type `Error` in JavaScript. You can throw whatever you like. This is why the TS compiler only allows a type of `any` here.

packages/source-fetcher/lib/polyfill.ts
eggplantzzz: So I don't know what is better here; 6 lines of code duplicated 4 times or 30 separate packages which each contain  6 
 lines of code that now need to be maintained for instances like this.

benjamincburns:
I could see making a shared polyfill module, but I think that would be a better fit for another PR. I suspect there are other polyfills that we'd want to pull into said module (e.g. the error chaining one that's currently in draft status will need to polyfill the `Error` type to chain properly on `node@<18`.

cds-amal:
> Polyfill logic was duplicated to ensure that it would load properly under each of the different command bundles.
> 
> An alternative would have been to make webpack handle injecting the polyfill, but then it wouldn't be present on the unpacked build.

:thinking: The only clients affected would be us. It could impact our ability to debug scenarios under Node 12. I don't think anyone else bootstraps to run truffle, so maybe there's merit in a webpack approach. @eggplantzzz, @haltman-at, @gnidan, @sukanyaparashar, @lsqproduction thoughts?

benjamincburns:
> > Polyfill logic was duplicated to ensure that it would load properly under each of the different command bundles.
> > 
> > An alternative would have been to make webpack handle injecting the polyfill, but then it wouldn't be present on the unpacked build.
> 
> :thinking: The only clients affected would be us. It could impact our ability to debug scenarios under Node 12. I don't think anyone else bootstraps to run truffle, so maybe there's merit in a webpack approach. @eggplantzzz, @haltman-at, @gnidan, @sukanyaparashar, @lsqproduction, @cliffoo thoughts?

@cds-amal I have a fairly strong preference against using webpack to load the polyfills, simply because I think it's dangerous precedent to establish the pattern of doing so. It would be fine for this change, but the next one that will be added will be for a much more modern feature (error chaining, introduced in ES2022). With a more modern polyfill we are far more likely to see differences in behavior between the packed and unpacked node versions.

packages/source-fetcher/lib/polyfill.ts
benjamincburns: I'd be opposed to a `@truffle/common` on the basis of the naming not really defining a well-defined scope (these are sometimes referred to as "god modules").

That said, I think a `@truffle/polyfills` that behaves like the module here (applies polyfills as a side-effect of module evaluation) is probably warranted.

This PR is currently blocking #5367. My request is that you guys allow this bit of cruft through on this PR, and then let me clean it up in #5367, as that PR will introduce a second polyfill, and personally I find it easier to structure modules like the one I'm proposing here (`@truffle/polyfills`) when I have more than a single concern to place on them.

If you guys agree, I'll add a note to #5367 with a link back to this convo so that it won't be missed on review of that one.

cds-amal: That sounds like a good idea @benjamincburns.

cds-amal:
>@cds-amal I have a fairly strong preference against using webpack to load the polyfills, simply because I think it's dangerous precedent. It would be fine for this change, but the next one that will be added will be for a much more modern feature (error chaining, introduced in ES2022). With a more modern polyfill we are far more likely to see differences in behavior between the packed and unpacked node versions.

I agree. My preference is a polyfill module, as we'll have to massage the environment when we drop/add the next set of node versions.

packages/compile-solidity/src/polyfill.ts
cds-amal: I guess we shouldn't worry about this If the intention is to revisit and have a polyfill module as followup work.

cds-amal (approved):
Thanks, @benjamincburns! I'm approving, though I think @eggplantzzz may have something to add.

packages/hdwallet-provider/test/provider.test.ts
benjamincburns: I stand corrected, it can also be type `unknown`.

benjamincburns: Apparently [best practice](https://typescript.tv/best-practices/error-ts1196-catch-clause-variable-type-annotation/) is to use `unknown` with type guards when you find yourself wishing that you had Java or C# like typed error handlers, but given that this is existing code that isn't throwing due to catching a type with a missing `message` property, I think that'd be complete overkill here.

packages/source-fetcher/lib/polyfill.ts
benjamincburns: Note added. https://github.com/trufflesuite/truffle/pull/5367#issuecomment-1211540122

benjamincburns:
@eggplantzzz I'm going to go ahead and merge this, as I think the only dangling thread was the question about the `catch` type, and I doubt that you want to block this on the typed error handler pattern that I mentioned there. That said, if you had other concerns please raise them to me and I'll be happy to create a new PR to address them, with priority.

PR #5425, opened 2022-08-10
Bug fix: Prevent crash that occurs when processing certain bad jumps

haltman-at:
So @cds-amal happened to hit this crash while doing some unrelated work.  I haven't reproduced it myself, but I'm pretty sure the correct solution here is to just throw a guard around it, so I'm doing that.

So what's going on here in more detail?  This code is part of the function that attempts to determine whether a given `JUMPDEST` instruction is the start of the designated invalid function that Solidity sometimes inserts into contracts as the place that uninitialized internal function pointers will jump to.  In recent Solidity versions, though, determining this is complicated by the fact that an uninitialized internal function pointer will not jump *directly* to the code that throws the error, rather it jumps to code that *jumps* to code that throws the error.  So, we have to look at the destination of that second jump!

Thing is, the code I wrote assumes that said destination exists as an instruction, i.e., it's not past the end of the code and it's not in the data part of a `PUSH` instruction.  If it is, you could get a crash.  Instead, we should just return `false` in such cases.  So I threw a guard around it to check that the instruction exists, and if it doesn't, then we'll fall to the end where we return `false`.

I didn't add a test.  Coming up with a test for this seems difficult!  But, @cds-amal could test it with his example (assuming he can reproduce it, which he might not be able to), or he could send that to me for manual testing?  I thought I'd throw this up here anyway though in this form; normally I'd test first, but I figured in this case I'd just get it up here since it's so simple.

### DIFF
--- a/packages/source-map-utils/index.js
+++ b/packages/source-map-utils/index.js
@@ -493,29 +493,31 @@ var SourceMapUtils = {
       const jumpIndex = instructions.findIndex(
         instruction => instruction.pc === jumpAddress
       );
-      if (checkAgainstTemplate(instructions, jumpIndex, newSequence)) {
-        return true;
-      }
-      debug("indirect: %O", instructions.slice(index, index + 4));
-      debug("jumpAddress: %d", jumpAddress);
-      debug("jumpIndex: %d", jumpIndex);
-      debug("instr count: %d", instructions.length);
-      const jumpInstruction = instructions[jumpIndex];
-      const jumpFile = jumpInstruction.file;
-      if (jumpFile !== -1) {
-        const findOverlappingRange = overlapFunctions[jumpFile];
-        const range = SourceMapUtils.getSourceRange(jumpInstruction);
-        const { node: jumpNode } = SourceMapUtils.findRange(
-          findOverlappingRange,
-          range.start,
-          range.length
-        );
-        if (
-          jumpNode &&
-          jumpNode.nodeType === "YulFunctionDefinition" &&
-          jumpNode.name === "panic_error_0x51"
-        ) {
+      if (jumpIndex !== undefined) {
+        if (checkAgainstTemplate(instructions, jumpIndex, newSequence)) {
           return true;
+        }
+        debug("indirect: %O", instructions.slice(index, index + 4));
+        debug("jumpAddress: %d", jumpAddress);
+        debug("jumpIndex: %d", jumpIndex);
+        debug("instr count: %d", instructions.length);
+        const jumpInstruction = instructions[jumpIndex];
+        const jumpFile = jumpInstruction.file;
+        if (jumpFile !== -1) {
+          const findOverlappingRange = overlapFunctions[jumpFile];
+          const range = SourceMapUtils.getSourceRange(jumpInstruction);
+          const { node: jumpNode } = SourceMapUtils.findRange(
+            findOverlappingRange,
+            range.start,
+            range.length
+          );
+          if (
+            jumpNode &&
+            jumpNode.nodeType === "YulFunctionDefinition" &&
+            jumpNode.name === "panic_error_0x51"
+          ) {
+            return true;
+          }
         }
       }
     }


benjamincburns (approved):
I think it's fine to merge this as-is.

That said, I think that problems like this (where you can't repro it easily due to the complexity of the integrated system) make a strong case for unit tests, but the cyclomatic complexity of this code makes that somewhat difficult.

At your discretion (or in a separate PR) you might consider moving the chunk of code in the block that you isolated here into its own function, and then you could write a simple unit test that proves the crash is no longer present.

benjamincburns:
Argh sorry - don't merge this yet - one sec

packages/source-map-utils/index.js
```diff
@@ -493,30 +493,32 @@ var SourceMapUtils = {
...
-          findOverlappingRange,
-          range.start,
-          range.length
-        );
-        if (
-          jumpNode &&
-          jumpNode.nodeType === "YulFunctionDefinition" &&
-          jumpNode.name === "panic_error_0x51"
-        ) {
+      if (jumpIndex !== undefined) {
```
benjamincburns, line 496: What's the return type if `jumpIndex` is `undefined`? If I'm reading it correctly, I think there's an implicit `undefined` return, while it returns `bool` if `jumpIndex` _is_ defined?

What are the implications to the caller if this function returns `undefined`?

Also I assume that you verified that `findIndex` strictly returns `undefined` when the search predicate is never satisfied? I know some library authors like to follow the pattern of `string.indexOf` and return `-1`.

benjamincburns (changes requested):
Sorry, I hit approve too soon on the last review. Please see comments below.

Also this function is a friggin' monster w.r.t. cyclomatic complexity. Regard this as optional, but I think since you're further increasing its cyclomatic complexity here I'd strongly recommend breaking it apart. Specifically I think it'd be quite trivial to break out the major blocks here into a small pile of non-exported internal functions with readable names, and that would make the logic flow here a _lot_ more readable.

packages/source-map-utils/index.js
haltman-at: > What's the return type if jumpIndex is undefined? If I'm reading it correctly, I think there's an implicit undefined return, while it returns bool if jumpIndex is defined?

No, in this case, as I mentioned, it'll fall down and hit the `return false;` at the end.

> Also I assume that you verified that findIndex strictly returns undefined when the search predicate is never satisfied? I know some library authors like to follow the pattern of string.indexOf and return -1.


Oops!  You're absolutely right, it returns `-1`.  My mistake.  The problem with writing things you didn't test! :grimacing: Thanks for catching that, will go change that.

benjamincburns: To be clear, I think unless it's part of the design of this function to return `undefined` sometimes, it's likely better that it always return `bool`.

benjamincburns: I think we were commenting at the same time. Honestly I'm having some serious trouble even seeing where this function starts and where it ends given all of the nested control flow structure!

benjamincburns: Also my bad - I thought the end of the diff was the end of the file. I'm still having a "fun" time spotting the beginning of the function, however.

benjamincburns: ok, found it with confidence at 399. I'm apparently a bit cross-eyed today, but also yeah - cyclomatic complexity like woah.

benjamincburns (approved):
OK, I think with the index change we're back into "better than it was before the change" territory, but I'd still _strongly_ recommend a follow-up PR that breaks the logic in this entire module into more bite-sized chunks.

haltman-at:
So, I guess I'll merge this shortly, but first I want to argue against this line of reasoning -- that the cyclomatic complexity is too high and that therefore it should be broken up.  Note I'm not necessarily arguging against breaking up this particular function; what I'm arguing against is the the argument used to support it, that of cyclomatic complexity.  It's a quite generic argument that pays little attention to the specifics of the function.  Whether the function should be broken up is, to my mind, a matter of those specifics, and it's that general point that I want to argue, not the case of this function in particular.  This came up also recently in the case of `CodeUtils.parseCode` on #5408, where you also argued that the cyclomatic complexity was too high, so I really want to respond to the general argument.

So, firstly: I think that if you want to argue that the cyclomatic complexity is too high, you are going to have to provide some numbers -- both the complexities of the functions in question, and some calibration numbers for what you're considering low or medium or high.  Ideally the calibration numbers would come with example functions illustrating those numbers; I know what I consider a complex or simple function, I don't know what I consider a high or low cyclomatic complexity.  Without examples for what the numbers correspond to, it's hard for me or anyone else to make a judgement as to whether your calibration is reasonable or not.

Which brings me to my second point: This just isn't that complex!  It's not, like, *simple*, but it's, y'know, medium.  It's within reason to my mind.  The same is even more true of `CodeUtils.parseCode`, which I wouldn't consider that complex at all.  Now I realize I'm arguing by flat assertion here -- but your arguments were also by flat assertion!  You say it's high complexity, I say it's not, evidently our calibrations are different here.  This is obviously not particularly decisive; honestly the function of this paragraph is largely just to make explicit the point that judgements of simplicity or complexity are subjective and that it can be useful to keep that in mind.

But, thirdly, I think looking at complexity of *functions* may be a bad way to look at things.  You request that the function be split up, but it is split up, just in a different way; it's just split up into sections rather than into functions.  Each section has a comment explaining its purpose.  This accomplishes much the same purpose as splitting it into functions; and function names have a length constraint, so trying to fit the explanation into that, and relying on the readability of function names rather than the readability of comments, strikes me as a potential way to lose information.  Meanwhile, splitting the function up into smaller functions would reduce the maximum cyclomatic complexity of any given function, but it would reduce the cyclomatic complexity of the broader program not at all!  If these smaller functions make sense on their own, then this is an improvement; but if they only make sense in context, then you haven't actually reduced the maximum cyclomatic complexity of the relevant readable segment.

Which brings me to my final point -- so far I've only argued that splitting up a function is not necessarily helpful; but in fact I what I want to argue is that splitting up a function can often be actively detrimental.  Taking a code segment and turning it into a function is only helpful if that code segment has independent meaning.  But many code segments that one might want to split out do not have independent meaning, and are only understandable in context.  In this case, splitting it out can easily reduce readability, by delocalizing the reading process.

When a code path is written linearly, one reads it from top to bottom, mentally keeping track of the meanings of the variables, and placing each meaning into its appropriate context.  When it is split into functions, and those functions have independent meaning, this simplifies the reading, as one can now skip over substantial segments whose meaning is now summed up in a function name; one no longer needs to read the implementing code.  But when the functions do not have independent meaning, skipping over their implementations is not possible if one wishes to achieve understanding.  The result is that you still need to read the implementation, but now the code for that implementation is delocalized from its context.  Instead of being able to read linearly, you have to jump up and down between the new function and its old context.  This makes determining what's going on harder, not easier, and it gets worse the more of these you split out!

And indeed, we have examples of this in our codebase.  Take a look at [VersionRange.ts](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts).  It's split into a number of small functions... and it's terribly confusing.  What's funny here is that it doesn't *need* to be confusing like this.  The overall procedure being performed isn't that complex.  It's just that it was split into a bunch of functions, but most of those functions only make sense in the context of the other functions.  So in trying to read it, you have to jump up and down and try to piece it all together.  This is, to my mind, a prime example of code that would be made more readable if written in a more direct and linear style with fewer and longer functions, with what are currently functions being turned into commented segments.

So considering all of the above, I maintain that if you want to argue that a function should be split up, such arguments need to be based on the specifics of the function, and take into account the factors above, and not be based solely on reducing the maximum cyclomatic complexity among functions.  Because the relevant unit of code isn't the function, but the separately readable segment; if you reduce the maximum cyclomatic complexity of functions in a way that fails to affect that of separately readable segments, then one has not improved matters, and by delocalizing things may have made things worse.

benjamincburns:
You've raised a bunch of points there @haltman-at, and it will take me a fair amount of time to reply to all of them. In the interim, please do feel free to merge this unless you're waiting on comments/review from someone else, as I'd hate for this discussion to hold up a good improvement to our code.

haltman-at:
Yup!  I wasn't going to hold it up, I just was holding off on merging for the purpose of visibility. :)  I'll go merge it now!

benjamincburns:
> I think that if you want to argue that the cyclomatic complexity is too high, you are going to have to provide some numbers -- both the complexities of the functions in question, and some calibration numbers for what you're considering low or medium or high. Ideally the calibration numbers would come with example functions illustrating those numbers; I know what I consider a complex or simple function, I don't know what I consider a high or low cyclomatic complexity. Without examples for what the numbers correspond to, it's hard for me or anyone else to make a judgement as to whether your calibration is reasonable or not.

That's a totally fair request. I use the [CodeMetrics VSCode plugin](https://marketplace.visualstudio.com/items?itemName=kisstkondoros.vscode-codemetrics) to show me a close approximation of cyclomatic complexity in my editor. It calculates a complexity score of 40 for the `isDesignatedInvalid` function here. The complexity prior to this change was 39.

As to whether that's high or low, I'm struggling to find any rigorous studies on this (though I'm sure that at least one exists _somewhere_ and I'll add a new reply if I find one), but it seems from a google search that there's a loose industry consensus around max complexity threshold values ranging between 10 and 15.

From personal experience, the two teams that I've been part of who enforced a maximum cyclomatic complexity metric used a critical threshold value of 10, and one of them had a "warning" threshold as well. I think it was 8, but I'm not 100% certain. I've also worked on a few other teams that didn't track/enforce it strictly, but who had a culture of giving the sort of feedback that I gave here and on #5408 when they had readability/testability concerns.

> Which brings me to my second point: This just isn't that complex! It's not, like, _simple_, but it's, y'know, medium. It's within reason to my mind. The same is even more true of `CodeUtils.parseCode`, which I wouldn't consider that complex at all.

I think we both can agree that the more independent paths there are through a functional code unit, the more working memory is required on behalf of the reader to interpret and reason about it, as there's literally just more stuff that you have to hold in your head.

I know that your remark wasn't intended as any sort of commentary on anyone's abilities, but I do have a cognitive disability that disadvantages me where working memory is concerned. Personally I don't find this code to be either "simple" or "medium" levels of complexity when it comes to readability.

I gave you the feedback about cyclomatic complexity in both cases because I felt that reading and reasoning about the code in question was more challenging than it needed to be. Furthermore, my general experience has been that I tend to have higher task success when working in codebases where metrics like cyclomatic complexity are regularly considered.

But my personal issues aside, I also tried in earnest to "red team" my personal experience here and I've not found anything that suggests that a cyclomatic complexity score of 40 is anything other than unacceptably high for a single function. You're correct that `CodeUtils.parseCode` with its score of 16 is _better_, but again it is still above the higher industry consensus threshold of 15.

But don't just take my word for it. The screenshots below show what the default settings for the aforementioned CodeMetrics plugin has to say on the matter (see the small text above the function names):

image

image

If you haven't already, I'd encourage you to go and research the topic on your own, as you might have more success in finding something that contradicts my findings (as well as the default settings for CodeMetrics).

> Now I realize I'm arguing by flat assertion here -- but your arguments were also by flat assertion! You say it's high complexity, I say it's not, evidently our calibrations are different here. This is obviously not particularly decisive; honestly the function of this paragraph is largely just to make explicit the point that judgements of simplicity or complexity are subjective and that it can be useful to keep that in mind.

I think we can get away from flat assertion if we're slightly more precise on this topic. That is, I think we'd both agree that with the assertion that complexity metrics are highly _objective_ measures of _only_ the specific things that they claim to measure, and the farther one strays from that definition, the less objectively useful they become.

With that in mind, I think what cyclomatic complexity has to say about the _testability_ of your code is a much more objective statement as compared to what it says as a proxy measurement for readability. This is because measuring testability is actually [its intended use](https://users.csc.calpoly.edu/~jdalbey/206/Lectures/BasisPathTutorial/index.html#HD_NM_19). Specifically, cyclomatic complexity is a measure of the number of linearly independent paths through a code unit (pretty much always a method or function in practice), so it stands to reason that as a code unit's cyclomatic complexity grows, it becomes more difficult to test.

My suspicion is that the testability concern is what drives the industry norm threshold of 10-15 moreso than the readability concern.

Either way, the inability to strictly decouple the objective aspects of this from the subjective ones is indeed _precisely_ the rub on this sort of thing. However it's also why I think that it would be helpful if we as a team had an agreed-upon complexity metric and threshold. Not because I think there's some perfect threshold, but rather because having some _good enough_ threshold would help us to _consistently_ enforce a certain (albeit imperfect) standard of readability and testability without the overhead of long comment threads like this one.

I'm also not too fussed about what that metric and threshold should be however, so if you're aware of [some](https://www.sonarsource.com/docs/CognitiveComplexity.pdf) [alternative](https://sourcery.ai/blog/working-memory/) that might work better for our team than cyclomatic complexity, please do point me in that direction! I just tend to use cyclomatic complexity because it's what I'm used to and because it's worked well for me in the past.

> But, thirdly, I think looking at complexity of _functions_ may be a bad way to look at things. You request that the function be split up, but it is split up, just in a different way; it's just split up into sections rather than into functions. Each section has a comment explaining its purpose. This accomplishes much the same purpose as splitting it into functions; and function names have a length constraint, so trying to fit the explanation into that, and relying on the readability of function names rather than the readability of comments, strikes me as a potential way to lose information. Meanwhile, splitting the function up into smaller functions would reduce the maximum cyclomatic complexity of any given function, but it would reduce the cyclomatic complexity of the broader program not at all! If these smaller functions make sense on their own, then this is an improvement; but if they only make sense in context, then you haven't actually reduced the maximum cyclomatic complexity of the relevant readable segment.

So I think in this case having the comment blocks is a _tremendous_ help for readability, but it doesn't do much for testability.

My suggestion to break up the code is generally my default suggestion for high cyclomatic complexity. As you point out here, you are correct that it doesn't do anything to reduce the overall complexity of the functionality that the original block of code encapsulates/encapsulated (tenses are hard), but it does mean that the resulting components are more readable and more testable pieces of the original whole (see code samples below for an example).

An arguably better approach is to actually refactor the logic to reduce the total complexity, but this is significantly more risky and significantly higher effort. On the topic of risk, the likelihood of introducing unintentional logic changes is high unless you have unit tests in place ahead of time that cover each [basis path](https://en.wikipedia.org/wiki/Basis_path_testing) (also see link above) in the code. This is generally why I don't suggest this approach as a good first-pass method of improving readability. 

> Which brings me to my final point -- so far I've only argued that splitting up a function is not necessarily helpful; but in fact I what I want to argue is that splitting up a function can often be actively detrimental. Taking a code segment and turning it into a function is only helpful if that code segment has independent meaning. But many code segments that one might want to split out do not have independent meaning, and are only understandable in context. In this case, splitting it out can easily reduce readability, by delocalizing the reading process.

I agree and I disagree. I think there are cases when this is absolutely true, but it's _generally_ been my experience that if you follow good naming practices for your internal/private functions, you tend to wind up with an extremely readable high-level process description in the original function, and then all of the details behind that process are encapsulated by the internal functions that you create when going through the refactoring process.

I'd be very happy to rework the code in this module along those lines so that we can have a concrete example to discuss. Do you think that would be useful? Also please see my example below as a preview.

> When a code path is written linearly, one reads it from top to bottom, mentally keeping track of the meanings of the variables, and placing each meaning into its appropriate context. When it is split into functions, and those functions have independent meaning, this simplifies the reading, as one can now skip over substantial segments whose meaning is now summed up in a function name; one no longer needs to read the implementing code. But when the functions do not have independent meaning, skipping over their implementations is not possible if one wishes to achieve understanding.

I strongly agree with all of your points here, but I think where we differ in the general case is on the assumption of independence. Personally I think that if a block of code can be given a short descriptive comment that describes what that code does, it can also be given a function name that describes that meaning just as well. It might not be generically useful outside of the module that contains it, but that's why you don't export it.

As one quick cherry-picked example from this PR, consider the following segment:

```
    //if it's panic_error_0x51, return true
    if (
      node &&
      node.nodeType === "YulFunctionDefinition" &&
      node.name === "panic_error_0x51"
    ) {
      return true;
    }
```

Now compare it to this one:

```javascript
    if (node.isWhatever0x51ActuallyMeansPanicError) return true;
```

Or the more functional version:

```javascript
    if (isWhatever0x51ActuallyMeansPanicError(node)) return true;
```

Which is more readable? Which allows you to write a simple and fast test for whether or not the module can correctly determine whether an `0x51` panic has occurred?

> The result is that you still need to read the implementation, but now the code for that implementation is delocalized from its context. Instead of being able to read linearly, you have to jump up and down between the new function and its old context. This makes determining what's going on harder, not easier, and it gets worse the more of these you split out!

I think this is very context dependent. Without breaking this apart I need to go and read the _entire_ implementation, no matter what my objective is in trying to read this code. With the code broken out into a function that represents a single high-level process, and several lower-level functions that implement that process, I only need to go and read the details that I care about in support of whatever my objective might be.

> And indeed, we have examples of this in our codebase. Take a look at [VersionRange.ts](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts). It's split into a number of small functions... and it's terribly confusing. What's funny here is that it doesn't _need_ to be confusing like this. The overall procedure being performed isn't that complex. It's just that it was split into a bunch of functions, but most of those functions only make sense in the context of the other functions. So in trying to read it, you have to jump up and down and try to piece it all together. This is, to my mind, a prime example of code that would be made more readable if written in a more direct and linear style with fewer and longer functions, with what are currently functions being turned into commented segments.

I'm sorry but I just strongly disagree with your assessment of the readability of the `VersionRange` code readability. The class at the top of the file is the clear exported entrypoint into the code, and in reading that class I can see clearly the high-level approach, and I can drill down into the details as needed by viewing each of the lower-level functions that are referenced by the methods in the `VersionRange` class.

That said, I do agree that when this sort of complexity metric thresholding is taken to an extreme (and sometimes even when it isn't), it can produce code that results in what you described. I also concede that there are times when you need to trace through the _complete_ logic for a given piece of functionality, and this can increase the effort involved when that's the case.

However it's been my experience that when your entrypoint code serves as a higher-level process description the need to read through an entire large segment of code becomes far far less frequent. That's because ultimately reading code is an exercise in tree traversal, and being able to prune out large sections of the tree is far more often a benefit than it is a hinderance when I'm trying to find the code segment that I care about.

But yeah, here I am once again arguing a flat assertion based on personal experience. I don't think that should invalidate the argument, but all the same I'll do my best to try and think of a better way to measure how often these situations occur, even if it's just keeping some notes about my experiences as I go about my usual work. If you have any ideas on how to go about this, please let me know.

> So considering all of the above, I maintain that if you want to argue that a function should be split up, such arguments need to be based on the specifics of the function, and take into account the factors above, and not be based solely on reducing the maximum cyclomatic complexity among functions. Because the relevant unit of code isn't the function, but the separately readable segment; if you reduce the maximum cyclomatic complexity of functions in a way that fails to affect that of separately readable segments, then one has not improved matters, and by delocalizing things may have made things worse.

I think I addressed all of the points that you covered in this last paragraph in my comments above. Please do let me know if I missed something, however.

I think the only other thing that's maybe worth mentioning here is that if we can agree on a complexity metric and how it should be applied, I'd be a proponent of that metric being enforced via automated code checks, even if the result of a "violation" is just an automated comment on the PR, similar to a coveralls report.

Finally, I just wanted to say that regardless of where we end up on this, I really appreciate you taking the time to write up your thoughts and giving me the opportunity to respond here.

Issue #5374, opened 2022-07-28
Contract instances should have `instance.estimateGas` (and probably `instance.call`) similar to `instance.sendTransaction`


haltman-at:
## Issue

Truffle Contract instances let you do `instance.sendTransaction({...})` to easily send an arbitrary transaction to the instance; this is useful, for instance, for deliberately invoking the contract's fallback or receive functions.  However there's no corresponding `instance.estimateGas({...})` for if you just want a gas estimate.  There's also no `instance.call({...})`, although it's not clear how useful that would be, as the return value would necessarily have to be undecoded hex.  Still, likely we should support it as well.

## Environment

* Truffle version (`truffle version`): 5.5.23

haltman-at:
A note, this issue arose out of discussion #5353.

haltman-at:
This is now released!

PR #5426, opened 2022-08-11
Enhancement: Add `instance.estimateGas()` and `instance.call()`

haltman-at:
Addresses #5374.  Also, while I was at it, I improved some existing tests and fixed a bug.

This PR adds support for `instance.estimateGas()` and `instance.call()`, much like we currently have `instance.sendTransaction()`.  This was mostly pretty straightforward.  First off, in `contract/index.js`, where `instance.sendTransaction` and `instance.send` are set up, I also set up `estimateGas` and `call` in the same way.  Note I put `call` behind a guard like `send` to make sure it doesn't overwrite a user method of the same name, because A. that might plausibly exist, and B. `instance.call` honestly isn't that useful, I just included it for completeness :P  I didn't bother with such a thing for `estimateGas` as I figured that is pretty unlikely to break anything, although we could if we want to be certain.

You'll notice that the way that `instance.sendTransaction` works is by calling `execute.send` but with passing `null` for the `fn` and `methodABI` arguments.  That strikes me as not a great way of doing things, but I didn't want to redo it, so I did the others in the same way for consistency.

However, this meant that `call` and `estimate` had to be updated to account for the possibility that `fn` and `methodABI` might be null.  Also, `estimate` had to take an `address` parameter like the others, which previously it didn't.  (Although, `method.estimateGas()` was passing it one... there just wasn't any parameter there to receive it!)  This is basically all straightforward -- if `fn` is present we do the existing thing, if it isn't we just invoke web3.  Note that in `estimate` we make sure to set the address appropriately; this was already accounted for in `call`, but needed to be set up explicitly in `estimate`.  There are other appropriate adjustments which I won't detail.

There are two more particular things I want to note, though.  First, note that I put explicit `await`s in `estimate` where they weren't any before.  This may not seem to matter, but it's actually important for improving error behavior (there should have been one all along).  Second, I *didn't* put an `if (methodABI)` guard around the call event emission.  On inspecting the code that uses it, I'm pretty sure it should work fine without one (and adding one would reduce its functionality, anyway!).

Anyway, with that done, it was just a matter of adding tests... and improving the existing tests!  I noticed that the existing tests of `sendTransaction` and `send` checked that `fallbackTriggered` was `false` before, but not that it was `true` afterward, so I added checks for that.  (I also changed the asserts to make better use of chai!)  And then I also added similar tests for `call` (altering the fallback function so that it returns a value) and `estimateGas` (I didn't know what to checked for this so I just checked that the output is a finite positive integer).

And that's it!  Nothing too complicated. :)

(And yeah, I deleted the geth tag when running the tests locally so that these tests would actually run. :P )

### DIFF
--- a/packages/contract-tests/test/methods.js
+++ b/packages/contract-tests/test/methods.js
@@ -589,7 +589,7 @@ describe("Methods", function () {
     });
   });
 
-  describe("sendTransaction() / send() [ @geth ]", function () {
+  describe("sendTransaction() / call() / estimateGas() / send() [ @geth ]", function () {
     it("should trigger the fallback function when calling sendTransaction()", async function () {
       const example = await Example.new(1);
       let triggered = await example.fallbackTriggered();
@@ -624,6 +624,26 @@ describe("Methods", function () {
 
       triggered = await example.fallbackTriggered();
       assert.isTrue(triggered, "Fallback should now have been triggered");
+    });
+
+    it("should trigger the fallback function when calling call()", async function () {
+      const example = await Example.new(1);
+
+      const result = await example.call({ data: "0x01234567" });
+      assert.strictEqual(
+        result,
+        "0x0123" //fallback function cuts the input in half
+      );
+    });
+
+    it("should trigger the fallback function when calling estimateGas()", async function () {
+      const example = await Example.new(1);
+
+      const result = await example.estimateGas({});
+      //I'm not sure what to test about this, so let's just test that it's a finite positive integer
+      assert.isFinite(result, "Gas estimate should be a finite number");
+      assert(result > 0, "Gas estimate should be positive");
+      assert(Number.isInteger(result), "Gas estimate should be an integer");
     });
 
     it("should accept tx params (send)", async function () {
--- a/packages/contract-tests/test/methods.js
+++ b/packages/contract-tests/test/methods.js
@@ -592,7 +592,7 @@ describe("Methods", function () {
   describe("sendTransaction() / send() [ @geth ]", function () {
     it("should trigger the fallback function when calling sendTransaction()", async function () {
       const example = await Example.new(1);
-      const triggered = await example.fallbackTriggered();
+      let triggered = await example.fallbackTriggered();
 
       assert.isFalse(triggered, "Fallback should not have been triggered yet");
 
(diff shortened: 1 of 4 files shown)


fainashalts (commented):
I think the code here works just fine but I am a little confused about the usefulness of an estimate gas function that doesn't indicate the method for which we need an estimate. I am likely missing something but wanted to throw in my two cents in case I am not!

haltman-at:
> I think the code here works just fine but I am a little confused about the usefulness of an estimate gas function that doesn't indicate the method for which we need an estimate. I am likely missing something but wanted to throw in my two cents in case I am not!

So, this originates from discussion #5353.  The user was looking for a convenient way to estimate the gas used by a fallback function invocation.

packages/contract/lib/execute.js
```diff
@@ -124,11 +124,12 @@ const execute = {
    * @param  {Array}  inputs     ABI segment defining method arguments
    * @return {Boolean}           true if final argument is `defaultBlock`
    */
-  hasDefaultBlock: function (args, lastArg, inputs) {
+  hasDefaultBlock: function (args, lastArg, methodABI) {
+    const expectedArgs = methodABI ? methodABI.inputs.length : 0;
```
eggplantzzz, line 128: A small nit but choose to ignore it; should we name this `expectedNumberArgs` or something? It probably doesn't matter too much but I was confused for a split second.

haltman-at: Sure, I can go clarify that.

fainashalts:
> fallback function invocation.

Ohhh, I see. I saw #5353 but didn't realize he was looking for a way to estimate for a fallback function. I rescind my question. :)

packages/contract-tests/test/methods.js
```diff
@@ -589,40 +589,61 @@ describe("Methods", function () {
...
+      assert.strictEqual(balance, web3.utils.toWei("1", "ether"));
+
+      triggered = await example.fallbackTriggered();
+      assert.isTrue(triggered, "Fallback should now have been triggered");
+    });
+
+    it("should trigger the fallback function when calling call()", async function () {
+      const example = await Example.new(1);
+
+      const result = await example.call({ data: "0x01234567" });
```
eggplantzzz, line 632: Can you explain the meaning of the value of `data` here?

haltman-at: It's the binary data passed to the call, just like in a `instance.sendTransaction`.  This one is obviously just an example.  I made it so that the fallback function takes the passed in data and cuts off the second half, so I passed in `0x01234567` and checked that I got `0x0123` back.  (Since the 4 bytes `0x01234567` doesn't correspond to any function selector, it'll hit the fallback function rather than any of the contract's defined functions.)

Does that answer your question?

fainashalts (approved):
LGTM! Thanks Harry!

PR #5431, opened 2022-08-12
Enhancement: Prepare debugger's `block.difficulty` reporting for the merge

haltman-at:
Since the merge has now happened on Goerli, I figured I'd go ahead and do this!  Note that I'm adding this PR without tests, since Ganache doesn't yet support Paris and the debugger tests rely on Ganache.  But I can go back and modify tests appropriately once Ganache does support Paris!  (Indeed, I'll likely have to...)

Other than the question of what to do about `0x44` in disassembly, which I'm intending to continue to put off for now, the debugger should only need one update for the merge: Determining the value of `block.difficulty`.  See, `block.difficulty` compiles to the opcode `0x44`, which pre-merge returns the difficulty but post-merge will return the mix hash.  As such, I've altered the place where we get the difficulty so that it will use the mix hash if it's nonzero, and the difficulty otherwise.  (Note that current versions of web3 will always return *something* for the mix hash, even if you're on a pre-merge block; it'll just be zero in that case.)  As a bit of future-proofing, I accounted for the possibility that the mix hash field might in a future web3 version be renamed from `mixHash` to `prevRandao`, as suggested in [EIP-4399](https://eips.ethereum.org/EIPS/eip-4399).

OK, but you may ask... what if the mix hash is legitimately zero on a post-merge block?  Won't we incorrectly look at the difficulty in that case?  Well, in that extraordinarily unlikely case, because it's a post-merge block, the difficulty will also be zero!  So there's no problem.

Anyway, the existing tests of `block.difficulty` continue to pass with this, at least. :)

### DIFF
--- a/packages/debugger/lib/web3/sagas/index.js
+++ b/packages/debugger/lib/web3/sagas/index.js
@@ -69,10 +69,25 @@ function* fetchTransactionInfo(adapter, { txHash }) {
   yield put(session.saveReceipt(receipt));
   yield put(session.saveBlock(block));
 
+  //Solidity's block.difficulty uses the opcode 0x44.
+  //This will return the difficulty (pre-merge) or the mixHash (post-merge).
+  //So if the mixHash is present and nonzero, we use that.  Otherwise, we use
+  //the difficulty.
+  let difficultyOrMixHash;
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
Looks good @haltman-at.

PR #5432, opened 2022-08-12
Internal improvement: Fully convert compile-solidity to use TypeScript

eggplantzzz:
NOTE THAT THIS IS A BREAKING CHANGE FOR @truffle/compile-common.

Note that types were moved from `compile-common/src/types.ts` to `compile-common/src/sources.ts` in order to prevent a clobbering issue. `src/index.ts` is the following:
```
export * as Shims from "./shims";
export * as Sources from "./sources";
export * as Errors from "./errors";
export * as Compilations from "./compilations";
export * from "./types";
```

If the type `Sources` is exported from `./types` then it will conflict with the `Sources` namespace.


packages/compile-common/src/types.ts
```diff
@@ -64,12 +48,12 @@ export type CompiledContract = {
...
-  legacyAST: object;
-  ast: object;
+  legacyAST?: object;
+  ast?: object;
   abi: Abi;
   metadata: string;
-  bytecode: Bytecode;
-  deployedBytecode: Bytecode;
+  bytecode?: Bytecode;
+  deployedBytecode?: Bytecode;
```
gnidan, line 56: This makes this PR a breaking change to compile-common, yes?

eggplantzzz: Will this be breaking? It makes the type more lenient meaning that it shouldn't break anything, right?

benjamincburns: Yeah, should be considered non-breaking, IMO

gnidan: "Making the type more lenient" only applies if this optionality is an input type. Like, if we're talking about exporting some function`(options: { /* ... */ }) => any`, then adding a `?` to one of the `options` is non-breaking.

Here, it's different... to a package consuming the type, upgrading to the version that includes this change means handling a new case: previously, (e.g.) you would only need to handle `object`, but now you must handle both sides of `object | undefined`.

packages/compile-common/src/sources.ts
```diff
@@ -1,6 +1,21 @@
...
+//original paths by transformed path
+export interface PathMapping {
+  [sourcePath: string]: string;
+}
+
+export interface CollectedSources {
+  sources: Sources;
+  targets: string[];
+  originalSourcePaths: PathMapping;
+}
```
benjamincburns, line 17: Why remove these from the `types` file?

packages/compile-solidity/src/profiler/loadParser.ts
```diff
@@ -32,12 +36,15 @@ async function loadParser({ events, compilers: { solc: solcConfig } }) {
   }
 
   // determine normal solc version and then load that version as solcjs
-  const { version } = semver.coerce(solc.version());
+  const result: SemVer | null = semver.coerce(solc.version());
```
benjamincburns, line 39: Is the return type of `semver.coerce` strictly `Semver | null`? If not, you likely either want `Optional<SemVer>` or `SemVer?` here.

packages/compile-solidity/src/profiler/loadParser.ts
```diff
@@ -46,8 +53,8 @@ async function loadParser({ events, compilers: { solc: solcConfig } }) {
   return makeParseImports(parserSolc);
 }
 
-function makeParseImports(parser) {
-  const parseImports = body => {
+function makeParseImports(parser: any) {
```
benjamincburns, line 56: Is there not a better type for this? If not, maybe this should be a generic?

packages/compile-solidity/src/profiler/loadParser.ts
```diff
@@ -56,12 +63,14 @@ function makeParseImports(parser) {
         // if there's a match provide the helpful error, otherwise return solc's error output
         if (contractSolcPragma) {
           const contractSolcVer = contractSolcPragma[0];
-          const configSolcVer = semver.valid(solc.version());
+          const configSolcVer = semver.valid(parser.version());
```
benjamincburns, line 66: Logic change. Are we sure that we want `parser.version` instead of `solc.version` here? Can they ever be different versions?

packages/compile-solidity/src/profiler/loadParser.ts
benjamincburns: If there really is no better type for this apart from `any`, or if defining a valid type constraint as part of this PR would be high effort, you might also consider defining a `SolidityParser` type alias that is still `any`, as that way it'll be easy to update in the future should better types become available.

packages/compile-solidity/src/run.ts
```diff
@@ -140,19 +205,13 @@ function orderABI({ abi, contractName, ast }) {
   ];
 }
 
-/**
- * @param sources - { [sourcePath]: contents }
- * @param targets - sourcePath[]
- * @param setings - subset of Solidity settings
- * @return solc compiler input JSON
- */
```
benjamincburns, line 148: Why remove the typedoc?

packages/compile-solidity/src/run.ts
```diff
@@ -1,14 +1,81 @@
...
+  sources: Common.Sources.Sources;
+  originalSourcePaths: Common.Sources.PathMapping;
+  solcVersion: string;
+};
+
+type Contracts = {
+  [path: string]: {
+    [name: string]: any;
+  };
+};
```
benjamincburns, line 69: Minor nit - you might consider moving these to a local `types.ts` file, which would be in line with other TS packages in the repo.

packages/compile-solidity/src/run.ts
```diff
@@ -1,14 +1,81 @@
...
 // however, there is a check here that returns null if no sources exist
-async function run(rawSources, options, internalOptions = {}) {
+export async function run(
+  rawSources: { [name: string]: string },
+  options: Config,
+  internalOptions: InternalOptions = {}
+): Promise<Compilation | null> {
   if (Object.keys(rawSources).length === 0) {
     return null;
   }
```
benjamincburns, line 81: If we are okay with breaking changes here, you _might_ want to consider making this throw instead of returning `null`, depending on what else calls this. That would be my recommendation, unless we do want this to be a strictly homomorphic change.

packages/compile-solidity/src/run.ts
benjamincburns: Ah, I guess they weren't all that useful before. Might be nice if you were to update them, though.

That said, consider this an optional request.

packages/compile-solidity/src/run.ts
```diff
@@ -171,7 +230,11 @@ function prepareCompilerInput({
  * @param sources - { [sourcePath]: string }
  * @return { [sourcePath]: { content: string } }
  */
-function prepareSources({ sources }) {
+function prepareSources({
+  sources
+}: {
+  sources: Common.Sources.Sources;
+}): PreparedSources {
```
benjamincburns, line 237: You might want to define `XOptions` interfaces for these args, as personally I find this syntax to be far less readable. There are a few other occurrences of this, but I'll just flag this one to cut down on the noise.

packages/compile-solidity/src/run.ts
```diff
@@ -216,7 +279,10 @@ function prepareOutputSelection({ targets = [] }) {
 /**
  * Load solc and perform compilation
  */
-async function invokeCompiler({ compilerInput, options, solc }) {
+async function invokeCompiler({ compilerInput, options, solc }): Promise<{
+  compilerOutput: CompilerOutput;
+  solcVersion: string;
+}> {
```
benjamincburns, line 285: How does this compile without a type definition for the args? I think you may have a bunch of things that are implicitly typed as `any`.

I'd recommend turning on the `noImplicitAny` flag in `tsconfig.json` and seeing what pops up.

packages/compile-solidity/src/run.ts
```diff
@@ -282,7 +348,7 @@ function detectErrors({
   const infos = rawInfos.map(({ formattedMessage }) => formattedMessage);
 
   if (errors.includes("requires different compiler version")) {
-    const contractSolcVer = errors.match(/pragma solidity[^;]*/gm)[0];
+    const contractSolcVer = errors.match(/pragma solidity[^;]*/gm)![0];
```
benjamincburns, line 351: I don't think the use of the `!` operator is safe here. I'd recommend using a type assertion instead.

packages/compile-solidity/src/run.ts
```diff
@@ -412,55 +480,58 @@ function processContracts({
...
+              linkReferences: formatLinkReferences(linkReferences)
+            }),
+            deployedBytecode: zeroLinkReferences({
+              bytes: (deployedBytecodeInfo || {}).object,
+              linkReferences: formatLinkReferences(
+                (deployedBytecodeInfo || {}).linkReferences
+              )
+            }),
+            immutableReferences: (deployedBytecodeInfo || {})
+              .immutableReferences,
```
benjamincburns, line 507: this would be a good place to use the coalescing operator
```suggestion
            immutableReferences: deployedBytecodeInfo?.immutableReferences,
```

packages/compile-solidity/src/run.ts
```diff
@@ -412,55 +480,58 @@ function processContracts({
...
 }
 
-function repairOldContracts(contracts) {
-  const contractNames = [].concat(
-    ...Object.values(contracts).map(source => Object.keys(source))
-  );
+function repairOldContracts(contracts: Contracts): Contracts {
+  const contractNames = Object.values(contracts)
+    .map(source => Object.keys(source))
+    .flat();
```
benjamincburns, line 526:
```suggestion
    .mapFlat(source => Object.keys(source));
```

packages/core/lib/testing/Test.js
```diff
@@ -13,7 +13,7 @@ const SolidityTest = require("./SolidityTest");
 const RangeUtils = require("@truffle/compile-solidity/dist/compilerSupplier/rangeUtils");
 const expect = require("@truffle/expect");
 const Migrate = require("@truffle/migrate");
-const Profiler = require("@truffle/compile-solidity/dist/profiler");
+const { Profiler } = require("@truffle/compile-solidity/dist/profiler");
```
benjamincburns, line 16: This is an example of a change that wouldn't be necessary if you were to use the `export =` syntax in `compile-solidity/dist/profiler`

yarn.lock
```diff
@@ -6526,7 +6526,7 @@
   dependencies:
     "@types/node" "*"
 
-"@types/node@*":
+"@types/node@*", "@types/node@12.12.21":
```
benjamincburns, line 6529: BTW it's always worth reading your `yarn.lock` to see what actually changed.

If you see a new line for a module that's already in use elsewhere in the repo, that's a _strong_ indication that you're introducing/increasing dependency version inconsistency.

packages/compile-solidity/src/run.ts
```diff
@@ -1,14 +1,81 @@
...
+      id: number;
+    };
+  };
+  errors?: any[];
+};
+
+type ProcessAllSourcesArgs = {
+  sources: Common.Sources.Sources;
+  compilerOutput: CompilerOutput;
+  originalSourcePaths: any;
```
benjamincburns, line 34: Can this be given a more specific type?

packages/compile-solidity/src/run.ts
```diff
@@ -1,14 +1,81 @@
...
+  language: string;
+};
+
+type PrepareCompilerInputArgs = {
+  sources: Common.Sources.Sources;
+  targets: Targets;
+  language: string;
+  settings: any;
+  modelCheckerSettings: any;
```
benjamincburns, line 43: Can these be given a more specific type?

packages/compile-solidity/src/run.ts
```diff
@@ -1,14 +1,81 @@
...
+};
+
+type InternalOptions = {
+  language?: string;
+  noTransform?: boolean;
+  solc?: any;
```
benjamincburns, line 49: Can this be given a more specific type?

packages/compile-solidity/src/compileWithPragmaAnalysis.ts
```diff
@@ -140,7 +146,7 @@ const compileWithPragmaAnalysis = async ({ paths, options }) => {
 
   reportSources({ paths: filteredPaths, options });
 
-  const compilations = [];
+  const compilations: any[] = [];
```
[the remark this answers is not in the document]
benjamincburns: I think it can be typed as `Compilation[]`, given the return type of the `run` function that populates this array.

benjamincburns (changes requested):
Nice work overall - I _really_ love seeing JS code swapped over.

I understand that some of the `any` types that I flagged might be challenging, so please do feel free to drop a note on any of them that are too challenging to fix now. I think my only ask would be that where that's the case, we should raise issues so that we can improve the types here over time.

Otherwise I think the only thing that I didn't address in my comments is that I think it'd be very nice if the tests were also converted over to TS. Totally not necessary as part of this PR, but I just wanted to give a nudge all the same.

packages/compile-common/src/types.ts
benjamincburns: Oh, very nice catch! I should've realized that this is a return type.

packages/compile-solidity/src/index.ts
```diff
@@ -191,8 +208,4 @@ const Compile = {
   }
 };
 
-module.exports = {
-  Compile,
-  CompilerSupplier,
-  Shims
-};
+export { Compile, CompilerSupplier };
```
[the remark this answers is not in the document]
benjamincburns: On further review, I don't think that this change in export syntax _does_ introduce a breaking change _here_. As such, it might be a better idea to delete the explicit `export` line and switch to using the `export` keyword inline with the definitions.

Please do take this as a nudge to review the export syntax in other modules (such as in `compile-solidity/dist/profiler`)

benjamincburns:
Oh one other thing - I'd recommend squashing these commits immediately prior to merging (but after review approval). I see at least one `update yarn.lock` commit, which leads me to believe that there are probably commits added by this PR that don't build properly, and that makes tools like `git bisect` a lot harder to use.

packages/compile-common/src/sources.ts
eggplantzzz: `Sources` was being clobbered in that other location due to this namespace being named `Sources`. So in other words, these are the exports for everything in compile-common/src...
```
export * as Shims from "./shims";
export * as Sources from "./sources";
export * as Errors from "./errors";
export * as Compilations from "./compilations";
export * from "./types";
```

The `Sources` type was being clobbered by the `Sources` namespace. @gnidan advised me that this is a perfectly good new location.

packages/compile-solidity/src/compileWithPragmaAnalysis.ts
```diff
@@ -149,13 +155,9 @@ const compileWithPragmaAnalysis = async ({ paths, options }) => {
 
     const config = Config.default().with(compilationOptions);
     const compilation = await run(versionsAndSources[compilerVersion], config);
-    if (compilation.contracts.length > 0) {
+    if (compilation && Object.keys(compilation.contracts).length > 0) {
```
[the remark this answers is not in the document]
eggplantzzz: This along with the other instances were necessary to handle `compilation` being `null`.

packages/compile-solidity/src/index.ts
```diff
@@ -27,7 +40,7 @@ async function compileYulPaths(yulPaths, options) {
     });
     debug("Yul compiled successfully");
 
-    if (compilation.contracts.length > 0) {
+    if (compilation && compilation.contracts.length > 0) {
```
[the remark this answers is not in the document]
eggplantzzz: So I'm not sure we want to throw in this instance. The `run` method returns a `Compilation` or `null` if there is nothing to compile. So if there was nothing to compile in `run`, `null` is returned and we don't push the result into the array.

packages/compile-solidity/src/index.ts
```diff
@@ -56,16 +69,16 @@ const Compile = {
...
     if (solidityNames.length > 0) {
       debug("Compiling Solidity (specified sources)");
       const compilation = await run(soliditySources, options, {
         noTransform: true
       });
       debug("Compiled Solidity");
-      if (compilation.contracts.length > 0) {
-        solidityCompilations = [compilation];
+      if (compilation && compilation.contracts.length > 0) {
+        solidityCompilations.push(compilation);
```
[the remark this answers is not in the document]
eggplantzzz: This seems like the same to me, TS didn't like doing it the other way but I forget why.

packages/compile-solidity/src/profiler/loadParser.ts
eggplantzzz: Yup, it is strictly `Semver | null`. https://github.com/npm/node-semver/blob/main/functions/coerce.js

packages/compile-solidity/src/index.ts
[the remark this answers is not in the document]
eggplantzzz: So I'll just respond to all the comments of this type here. I think the old behavior is buggy. Should it return `null` it would throw in the old code as you can't access `contracts` on `null`. So I was considering this fixing a bug.

packages/compile-solidity/src/index.ts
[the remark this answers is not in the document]
eggplantzzz: okie doke

packages/compile-solidity/src/profiler/loadParser.ts
eggplantzzz: This was an error. There is no `solc` in this scope and rather `parser` is the variable we want to use. If you look at the call to this method ~ line 24 you can see `solc` is passed in here (a portion of the result of calling `supplier.load`). Looks to me like someone got their names confused.

packages/compile-solidity/src/run.ts
eggplantzzz: It seemed redundant now that we have the types...plus TS told me to get rid of it :)

packages/compile-solidity/src/run.ts
eggplantzzz: That would be nice but I didn't implement that because `flatMap` isn't supported in Node 12. I assume that ends up in the built code.

packages/compile-solidity/src/run.ts
eggplantzzz: So in a lot of these situations I just tried to maintain current behavior. You think I should make it throw? I suppose that would be an improvement.

packages/compile-solidity/src/run.ts
eggplantzzz: sure, I don't mind

packages/compile-solidity/src/run.ts
eggplantzzz: `Compile.Sources.PathMapping`

packages/compile-solidity/src/run.ts
eggplantzzz: I'm not certain about this as this is something returned from `solcWrap`. If you follow it back, it returns whatever gets returned [here](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts#L119-L129). I don't know what that is exactly, any suggestions?

packages/compile-solidity/src/run.ts
eggplantzzz: I changed these to `object` for now as I'm not certain how accurate I can be about what will be in here to be honest.

packages/compile-solidity/src/index.ts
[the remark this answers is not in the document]
benjamincburns: Yeah if you're confident that everything works fine this way, then it's fine by me.

packages/compile-solidity/src/index.ts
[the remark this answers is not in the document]
benjamincburns: Ah my bad, for some reason I didn't catch that these were meant to be one-and-only pushes to the array.

I do find the old syntax to be more clear as to that intention, though.

If TS complained, it's probably something that the `as` keyword could fix, but I'm not super fussed about this now that I understand that this is logically correct.

packages/compile-common/src/sources.ts
benjamincburns: Makes sense. Might be worth adding a comment, but I don't think it's strictly necessary.

packages/compile-solidity/src/profiler/loadParser.ts
benjamincburns: That begs the question as to whether this code is even reachable... Not something to worry about in this PR if we're aiming for a homomorphic change here, however.

packages/compile-solidity/src/run.ts
benjamincburns: Yeah if there isn't more to say other than type details, I suppose it would be.

packages/compile-solidity/src/run.ts
benjamincburns: Not important to me. I wanted to be sure that you'd considered it, but you _definitely_ have now, and it does seem reasonable to leave it as is.

packages/compile-common/src/sources.ts
eggplantzzz: Probably the only thing useful in the future will be the conversation on this PR. Nobody is going to know that these types used to be in `./types.ts`, right?

packages/compile-solidity/src/run.ts
benjamincburns: Pretty sure it down-compiles to a call to `map(...).flat()` thanks to the target being `ES2016`, but to be sure you'd have to check the JS output. That said, it's not really worth the extra time.

packages/compile-solidity/src/run.ts
benjamincburns: I'm not too fussed.

Worst case you could define a custom interface here that at least captures the members of the type on which you depend, but I don't see a ton of value in that, as it would make more sense for there to be actual types coming from `solcWrap`.

I don't think there is a good answer to this, but the only lingering concern that I have is over what would happen if there was suddenly proper typing for `solcWrap`? Ideally this dependency would be immediately discoverable due to a compiler error, but that won't happen if it's `any`.

I suppose a custom local type would break if it had structural conflicts with the hypothetical typed return value of `solcWrap`, but I dunno - I don't like this package assuming authority over another package's types.

So yeah, this is probably just a really long-winded way of saying "yeah, fair enough." 😆

packages/compile-solidity/src/run.ts
benjamincburns: If this code doesn't depend on any properties of these (sorry, I didn't go look again to see if it does), then that's maybe fine? Alternatively you could consider typing anything that is opaque to this package as `unknown`, as that would force a consumer to declare a type for it before actually consuming it.

I'll leave the decision as to whether or not that is appropriate to you, however.

packages/compile-common/src/sources.ts
benjamincburns: Yeah fair enough.

packages/compile-solidity/src/run.ts
benjamincburns: Yeah, it throws either way if the `!` is invalid. If you throw on a null assertion, at least you can provide some context.

Otherwise if we leave it strictly identical in behavior in cases like this, we're negating the primary value that the TS type checker provides, IMO.

benjamincburns: Also an explicit throw is more readable, IMO, as that makes it a lot more clear that this code _can_ throw.

packages/core/lib/testing/Test.js
eggplantzzz: I like named imports :)

eggplantzzz:
Thanks for the review @benjamincburns! Oh, and as for converting the tests to TS, see https://github.com/trufflesuite/truffle/pull/5445 where the tests get removed from this package altogether :)

packages/compile-solidity/src/profiler/loadParser.ts
eggplantzzz: Do you think I'm mistaken? I think this code will crash if executed. I would argue that this is a small enough change that it shouldn't be a big deal to include it. What do you think about that point of view?

benjamincburns: > Do you think I'm mistaken?

Oh not at all! My reply was more academic, sorry. It's more that if the original code would crash if executed, but we weren't ever seeing that, this whole path may be effectively dead code.

I don't think there's any further action required here. Sorry for the confusion!

packages/core/lib/testing/Test.js
benjamincburns: Yeah totally fair since this is an internal module. Just highlighted it more as an example of the difference caused by the way that TS handles exports. Apologies if this was already well understood on your end. I only recently encountered this "restriction" of the TS export syntax.

packages/compile-solidity/src/run.ts
benjamincburns: FWIW, since this code bumps up against a bunch of stuff that's untyped, please don't interpret my suggestion here as trying to require you to enable `noExplicitAny` in order to merge this. I do think this code would probably be improved if `noImplicitAny` were `true`, but I was more just suggesting temporarily enabling it to spot any implicit `any` types to make it easier to decide whether it's possible/appropriate to make them explicitly typed.

Net net, I think this PR is already well over the threshold of value added to our code, so I'm likely to approve once you're done pushing changes.

packages/compile-solidity/src/run.ts
```diff
@@ -70,17 +92,15 @@ async function run(rawSources, options, internalOptions = {}) {
...
     sources,
     compilerOutput,
     originalSourcePaths,
     language
   });
-  const sourceIndexes = outputSources
+  const sourceIndexes: string[] = outputSources
     ? outputSources.map(source => source.sourcePath)
-    : undefined; //leave undefined if sources undefined
+    : [];
```
benjamincburns, line 103: Sorry that I missed this on my first pass, but I think this change is breaking? That is, assuming that we consider the `run` function part of the public interface of the package.

I'd say it's breaking because code that previously expected an `undefined` for `sourceIndexes` in the return data must now be changed to expect an empty array.

packages/compile-solidity/src/run.ts
```diff
@@ -124,8 +144,8 @@ function orderABI({ abi, contractName, ast }) {
 
   // Put function names in a hash with their order, lowest first, for speed.
   const functionIndexes = orderedFunctionNames
-    .map((functionName, index) => ({ [functionName]: index }))
-    .reduce((a, b) => Object.assign({}, a, b), {});
+    .map((functionName: string, index: number) => ({ [functionName]: index }))
+    .reduce((a: object, b: object) => Object.assign({}, a, b), {});
```
benjamincburns, line 148: I don't think `object` is an appropriate type here, is it? That is, shouldn't it be an object with the same index signature that was defined by the map?

```suggestion
    .reduce((a: { [functionName: string]: number } , b: { [functionName: string]: number }) => Object.assign({}, a, b), {});
```

packages/compile-solidity/src/run.ts
eggplantzzz: I would say that it probably isn't part of the interface. It is not exposed by it. It is exposed, however, in the work done to move the tests [here](https://github.com/trufflesuite/truffle/pull/5445). But that is purely so that it can be tested by this separate packages. What do you think @gnidan? Maybe you have a different opinion.

packages/compile-solidity/src/run.ts
eggplantzzz: Ok, that is a bit more precise.

packages/compile-solidity/src/run.ts
benjamincburns: Assuming that the consensus is that these _aren't_ part of the public interface, but that they still need to be exported for testing purposes, I think we should consider using some naming prefix to indicate to would-be consumers that these are not actually part of the public interface.

Historically I've used an underscore or dunder prefix for that, but I'm not really all that fussed about what we actually use - so long as it's communicated externally.

benjamincburns: Alternatively maybe we just communicate that unless otherwise documented, only the things exported by the module's entrypoint should be considered part of the public interface?

packages/compile-solidity/src/run.ts
eggplantzzz: Do we want to force the user to do that?

packages/compile-solidity/src/run.ts
gnidan: > Alternatively maybe we just communicate that unless otherwise documented, only the things exported by the module's entrypoint should be considered part of the public interface? This is probably much easier to enforce in practice.

This is my strong preference - don't force us to use underscores ;)

Particularly following @wbt's excellent concerns raised in https://github.com/trufflesuite/truffle/discussions/5417 (aside: sorry wbt; I know I still owe you a response there), I think we should probably write something similar to how Hardhat has its ["stability guarantees" page](https://hardhat.org/hardhat-runner/docs/reference/stability-guarantees), where it says at the top that they don't "follow semver strictly."

This will probably hurt a bit at first, but it seems best to maintain our public interfaces explicitly as "what is documented is what you can rely on". Sure, sometimes people like you, @wbt, will have to fix breakages as they happen, but that's unavoidable... the only user-facing difference with strict adherence to semver would be that each breakage gets a major release; the work to upgrade remains the same. But it's costly to do major releases, and it's also dangerous for us to let our public interfaces remain so loosely defined as "whatever is exported"... maybe in a few years we'll have the project maturity to support that, but right now this sort of thing keeps causing us to get in our own way.

gnidan: (Oh and please don't fret @wbt; I'm not talking about your recently raised issues about our breaking types. Those are absolutely mea culpa and should've been major releases or not at all. I'm specifically talking here about the part of the interface surface area that we believe is internal)

benjamincburns: > This is my strong preference - don't force us to use underscores ;)

Ha, I'm not sure that I could even if I wanted to!

benjamincburns: I'm going to resolve this thread under the assumption that this isn't a change to public interface, and that @gnidan and/or @eggplantzzz will handle figuring out how to communicate our stability guarantees elsewhere.

packages/compile-solidity/src/run.ts
benjamincburns: I'd vote yes. Using `unknown` for code that hasn't been typed yet creates a sort of boundary line for unchecked types in our code. If we use `any`, the unchecked types propagate out far-and-wide, and it can become very difficult to reign it in later.

That said, if there's a difference of opinion here, I'm not staunch enough about this to block the entire changeover to TS just for this one concern.

Here's some reading on the subject, should you want to consider it further.
https://dmitripavlutin.com/typescript-unknown-vs-any/

benjamincburns (approved):
Approved on the assumption that there's consensus that the `run` function is internal code. If this isn't a correct assumption, please don't merge until that's resolved.

I also left a few comment threads unresolved, but I don't think that resolving them is necessary to get this merged.

Super happy to be seeing more and more of our code converting over to TS!

wbt:
I'm not sure why I got a mention notification on merge, but please consider this a polite reminder that breaking changes should be marked as such in the [semver](https://github.com/orgs/trufflesuite/discussions/5417) update!

Issue #5296, opened 2022-07-14
Use a `TruffleError` instead of `Error` error for options parsing errors


davidmurdoch:
See #5295 for additional context.

kevinbluer:
Addressed in https://github.com/trufflesuite/truffle/pull/5433 👌

PR #5433, opened 2022-08-12
Use TruffleError instead of Error while parsing mutually exclusive options

sukanyaparashar:
### ISSUE
Truffle uses the generic `Error` type in some situations that provide a stack trace to the users that is unnecessary. 
For example: Using mutually exclusive options, `--url` and `--network` with `truffle console` and `truffle debug` includes a stack trace, which is irrelevant to the users, as Truffle doesn't allow these two options to be specified together. See #5296.

### SOLUTION
This PR resolves this issue by using `TruffleError` instead of the generic `Error` type while parsing the mutually exclusive options, `--url` and `--network` with `truffle console` and `truffle debug`.

### DIFF
--- a/packages/core/lib/commands/console/run.js
+++ b/packages/core/lib/commands/console/run.js
@@ -1,11 +1,21 @@
 module.exports = async function (options) {
+  const OS = require("os");
   const { Console, excludedCommands } = require("../../console");
   const { Environment } = require("@truffle/environment");
+  const TruffleError = require("@truffle/error");
   const commands = require("../commands");
   const loadConfig = require("../../loadConfig");
 
   if (options.url && options.network) {
-    throw new Error("'url' and 'network' are mutually exclusive options");
+    const message =
+      "" +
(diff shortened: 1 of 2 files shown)


haltman-at (commented):
Thanks Sukanya!  Do you know, are there other instances of this in our codebase, or is this it?

sukanyaparashar:
> Thanks Sukanya! Do you know, are there other instances of this in our codebase, or is this it?

Hey @haltman-at, I will check the codebase once more if there happens to be other cases where we can use `TruffleError` instead of `Error`.

haltman-at:
Well if you checked once, I don't know if you need to check a second time!  I was just making sure you remembered to check in the first place. :)

lsqproduction:
looks good to me. 👍

lsqproduction (approved):

haltman-at (approved):

cds-amal (approved):
Looks good! Thanks, @sukanyaparashar

PR #5434, opened 2022-08-13
Hook up `truffle debug` to @truffle/from-hardhat

gnidan:
This PR builds on #5420 and uses that PR's introduced @truffle/from-hardhat package to provide native Hardhat support for the `truffle debug` command.


benjamincburns (approved):
LGTM - pretty easy one to approve.

That said, this one could probably also use a test. I think unfortunately anything other than a scenario test would require a lot of additional refactoring.

Feel free to pile this onto #5442 if you'd like.

gnidan:
Thanks @benjamincburns!

PR #5443, opened 2022-08-16
allow extractOptions to work with -option

lsqproduction:
extractOption at the current state only recognized `--options`, this adds `-option`.

### DIFF
--- a/packages/core/lib/command-utils.js
+++ b/packages/core/lib/command-utils.js
@@ -106,7 +106,7 @@ const prepareOptions = ({ command, inputStrings, options }) => {
     }
   });
 
-  // method `extractFlags(args)` : Extracts the `--option` flags from arguments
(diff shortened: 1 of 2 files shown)


haltman-at (approved):
It's probably worth noting just what this affects.  As best I can tell, this only affects whether a given option triggers the unsupported options warning, right?

lsqproduction:
> 

you're correct @haltman-at. Do you think this should be noted in the comments in the code?

eggplantzzz:
I don't think that should be mentioned in the code - just maybe noted on this PR. Looks like it is already here :)

haltman-at:
Yes agree with Tyler, this doesn't need a code comment, would have been helpful to note in the PR though.

lsqproduction:
got cha. Thank you for explaining.

PR #5444, opened 2022-08-16
Enhancement: Add godwoken mainnet to sourcify fetcher

haltman-at:
Sourcify added one more network, adding it!  (I already had it in `networks.ts` so I just needed to add it to `sourcify.ts`.)


sukanyaparashar (approved):
Looks good, @haltman-at !

cds-amal (approved):
Nice!

PR #5445, opened 2022-08-16
Internal improvement: Move compile-solidity tests to a new separate package

eggplantzzz:
While bootstrapping Truffle, you will notice a warning printed by lerna about a circular dependency. The tests for @truffle/compile-solidity use @truffle/resolver (as it is required in the config for certain flows) and @truffle/resolver has a dependency on @truffle/compile-solidity. To eliminate this circular dependency, these tests have been moved to their own package (and converted to TypeScript for kicks!).

Note that this PR is against the `moreTS` branch which is the branch where @truffle/compile-solidity is fully converted (almost, not the tests) to TypeScript in [this PR](https://github.com/trufflesuite/truffle/pull/5432).


packages/compile-solidity-tests/test/compileWithPragmaAnalysis.ts
```diff
@@ -129,8 +133,11 @@ describe("compileWithPragmaAnalysis", function () {
         options: config,
         paths: [path.join(sourceDirectory, "withImports", "C.sol")]
       });
+      if (!compilations) {
```
haltman-at, line 136: If you're doing `if (...) { assert.fail() }`, you're not using assertions properly.  I think the right way to do this in chai would be `assert.isDefined(compilations), "Compilations were not created")`.

packages/compile-solidity-tests/test/compileWithPragmaAnalysis.ts
```diff
@@ -129,8 +133,11 @@ describe("compileWithPragmaAnalysis", function () {
...
+        assert.fail("Compilations were not created.");
+      }
       assert.equal(compilations.length, 1);
-      assert(compilations[0].compiler.version.startsWith("0.6.12"));
+      assert(compilations[0].compiler.version!.startsWith("0.6.12"));
```
haltman-at, line 140: What is `!.`?  I've never seen that before, I can't find any documentation on it...

packages/compile-solidity-tests/test/compilerSupplier/index.ts
```diff
@@ -196,51 +211,39 @@ describe("CompilerSupplier", () => {
...
-            assert(solc === "called Local");
-            done();
-          })
-          .catch(() => {
-            assert(false);
-            done();
-          });
+      it("calls load on the Local strategy", async function () {
+        const { solc } = await supplier.load();
+        assert(solc.version() === "called Local");
```
haltman-at, line 246: This should be an `assert.equal` or `assert.strictEqual`.

packages/compile-solidity-tests/test/compilerSupplier/index.ts
```diff
@@ -63,15 +76,18 @@ describe("CompilerSupplier", () => {
...
       });
       afterEach(() => {
-        Docker.prototype.load.restore();
+        unStub(Docker.prototype, "load");
       });
 
       it("calls load on the Docker strategy", async function () {
         const { solc } = await supplier.load();
-        assert(solc === "called Docker");
+        assert(solc.version() === "called Docker");
```
haltman-at, line 90: Same comment here, this should be an `assert.equal` or `assert.strictEqual`.

packages/compile-solidity-tests/test/compilerSupplier/index.ts
```diff
@@ -81,59 +97,54 @@ describe("CompilerSupplier", () => {
...
       });
       afterEach(() => {
-        Native.prototype.load.restore();
+        unStub(Native.prototype, "load");
       });
 
       it("calls load on the Native strategy", async function () {
         const { solc } = await supplier.load();
-        assert(solc === "called Native");
+        assert(solc.version() === "called Native");
```
haltman-at, line 109: (another one)

packages/compile-solidity-tests/test/compilerSupplier/index.ts
```diff
@@ -81,59 +97,54 @@ describe("CompilerSupplier", () => {
...
       });
       afterEach(() => {
-        VersionRange.prototype.load.restore();
+        unStub(VersionRange.prototype, "load");
       });
 
       it("calls load on the VersionRange strategy", async function () {
         const { solc } = await supplier.load();
-        assert(solc === "called VersionRange");
+        assert(solc.version() === "called VersionRange");
```
haltman-at, line 128: (and another)

packages/compile-solidity-tests/test/compilerSupplier/index.ts
```diff
@@ -81,59 +97,54 @@ describe("CompilerSupplier", () => {
...
-            assert(solc === "called VersionRange");
-            done();
-          })
-          .catch(() => {
-            assert(false);
-            done();
-          });
+      it("calls load on the VersionRange strategy", async function () {
+        const { solc } = await supplier.load();
+        assert(solc.version() === "called VersionRange");
```
haltman-at, line 147: (and another)

packages/compile-solidity-tests/test/compilerSupplier/index.ts
```diff
@@ -196,51 +211,39 @@ describe("CompilerSupplier", () => {
...
```
haltman-at, line 225: (and another)

packages/compile-solidity-tests/test/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -58,32 +58,38 @@ const allVersions = {
...
-      instance.getSolcFromCacheOrUrl.restore();
-      instance.versionIsCached.restore();
+      unStub(instance, "getCachedSolcByVersionRange");
+      unStub(instance, "getSolcFromCacheOrUrl");
+      unStub(instance, "versionIsCached");
     });
 
     it("calls getCachedSolcByVersionRange when single solc is specified", async () => {
       await instance.load("0.5.0");
+      // @ts-ignore - TS not smart enough to recognize stubbed methods
```
haltman-at, line 87: OK, this one (and the one right after it) were pre-existing, but I'm guessing these should be using `assert.isDefined`?

packages/compile-solidity-tests/test/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -142,14 +150,16 @@ describe("VersionRange loading strategy", () => {
...
-        instance.cache.add.restore();
-        instance.compilerFromString.restore();
+        unStub(instance.cache, "has");
+        unStub(instance.cache, "add");
+        unStub(instance, "compilerFromString");
       });
 
       it("eventually calls add and compilerFromString", async () => {
         await instance.getSolcFromCacheOrUrl("0.5.1");
+        // @ts-ignore
```
haltman-at, line 160: (same comment again)

packages/compile-solidity-tests/test/run.ts
```diff
@@ -26,12 +27,11 @@ describe("async run(rawSources, options)", () => {
...
-      try {
-        const { contracts } = await run(rawSources, options);
-        assert(contracts);
-      } catch (error) {
-        assert.fail(error);
+      const result = await run(rawSources, options);
+      if (result === null) {
+        throw new Error("Compilation failed.");
       }
+      assert(result.contracts);
```
haltman-at, line 34: Another one that should be `assert.isDefined`.

packages/compile-solidity-tests/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "private": true,
+  "version": "0.1.0",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "ttsc",
+    "prepare": "yarn build",
+    "test": "./scripts/test.sh"
+  },
+  "devDependencies": {
+    "@truffle/artifactor": "^4.0.165",
```
haltman-at, line 24: These versions will have to be updated before merge, of course.

haltman-at (commented):
My main note here is that there are a bunch of places where `assert` is used in a way that isn't the closest match to the intent, I would suggest changing those.  A bunch of those are pre-existing so maybe that's not entirely necessary?  But I thought it was worth pointing out at least.

packages/compile-solidity-tests/test/compileWithPragmaAnalysis.ts
cds-amal: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

eggplantzzz: Yeah, that just tells TS "hey, I know what I'm doing and this will be defined". It's just so I don't have to do something like
```
if (compilations[0] && compilations[0].compiler && compilations[0].compiler.version...
```

eggplantzzz:
Sounds good to me @haltman-at about the massaging of the chai assertions!

eggplantzzz:
@haltman-at it seems like chai has no preference for `assert(a === b)` or `assert.equal(a, b)`. They have examples of both kinds of usage in the docs https://www.chaijs.com/api/assert/. I changed some of them over. Do you have a preference for `assert.equal`?

packages/compile-solidity-tests/test/compileWithPragmaAnalysis.ts
haltman-at: Oh, OK.  Thanks!

packages/compile-solidity-tests/test/compilerSupplier/loadingStrategies/VersionRange.ts
haltman-at: Oh is it?  I didn't get that.  For booleans there's also `isTrue` which is more specific. :)

haltman-at:
> @haltman-at it seems like chai has no preference for `assert(a === b)` or `assert.equal(a, b)`. They have examples of both kinds of usage in the docs https://www.chaijs.com/api/assert/. I changed some of them over. Do you have a preference for `assert.equal`?

Yeah the latter generates much better error messages when it fails.

eggplantzzz:
Oh that's when it gives you the actual values of the things being tested on failure in the test report. That's always super helpful.

packages/compile-solidity-tests/test/compileWithPragmaAnalysis.ts
haltman-at, line 136: (There is still this which I think should be changed to use an `assert.isDefined` or something rather than `if (...) { assert.fail() }`, but I think that's the last one.)

packages/compile-solidity-tests/test/compilerSupplier/index.ts
```diff
@@ -250,18 +253,14 @@ describe("CompilerSupplier", () => {
...
-            assert(error.message.includes(expectedMessageSnippet));
-            done();
-          });
+      it("throws an error", async function () {
+        try {
+          await supplier.load();
+          assert.fail();
+        } catch (error) {
+          let expectedMessageSnippet = "version matching globbity gloop";
+          assert(error.message.includes(expectedMessageSnippet));
```
haltman-at, line 262: Oh, actually, this could be an `assert.include`.

packages/compile-solidity-tests/test/supplier.ts
```diff
@@ -224,9 +227,10 @@ describe("CompilerSupplier", function () {
           "Version8Pragma",
           compilations[0].contracts
         ); //update when necessary
-        assert(VersionLatestPragma.compiler.version.includes("0.8.")); //update when necessary
-        assert(
-          VersionLatestPragma.contractName === "Version8Pragma", //update when necessary
+        assert.isTrue(VersionLatestPragma.compiler.version.includes("0.8.")); //update when necessary
```
haltman-at, line 230: I would also suggest changing this to use `assert.include`.  I don't think it makes sense to use `assert.isTrue` here; that's more for, like, checking that a particular boolean has been computed correctly.  For a weird custom assertion, it makes more sense to just use `assert`.  But in this case there's no need to do that, because there's `assert.include`. :)

packages/compile-solidity-tests/test/supplier.ts
```diff
@@ -318,8 +327,8 @@ describe("CompilerSupplier", function () {
           error = err;
         }
 
-        assert(error);
-        assert(error.message.includes("option must be"));
+        assert.isDefined(error);
+        assert.isTrue(error.message.includes("option must be"));
```
haltman-at, line 331: (same comment)

packages/compile-solidity-tests/test/supplier.ts
```diff
@@ -344,8 +353,8 @@ describe("CompilerSupplier", function () {
           error = err;
         }
 
-        assert(error);
-        assert(error.message.includes(imageName));
+        assert.isDefined(error);
+        assert.isTrue(error.message.includes(imageName));
```
haltman-at, line 357: (same comment)

haltman-at (commented):
(Actually, just noticed a few asserts that could use `assert.include`!)

packages/compile-solidity-tests/package.json
```diff
@@ -0,0 +1,53 @@
...
+  "private": true,
+  "version": "0.1.0",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "ttsc",
+    "prepare": "yarn build",
+    "test": "./scripts/test.sh"
+  },
+  "devDependencies": {
+    "@truffle/artifactor": "^4.0.167",
```
haltman-at, line 24: These need to be updated again.

haltman-at (commented):
OK, seems good to me, except package versions need to be updated again due to the release last week!

haltman-at (approved):

PR #5448, opened 2022-08-17
Fix exception on deploying a contract with both constructor arguments and deploy options

ebrahimmfadae:
## Remove the line that deletes overwrite property from options object

In later steps, we use the existence of `overwrite` property in the `isTxParams()` function. If we delete it, the `options` object will be considered as a constructor argument; Thus we will hit a runtime error because of an extra or wrong constructor argument.

### Environment
```sh
$ truffle version
Truffle v5.5.26 (core: 5.5.26)
Ganache v7.4.0
Solidity - 0.8.4 (solc-js)
Node v16.16.0
Web3.js v1.7.4
```

### Steps to reproduce

1. Write a smart contract that takes at least one argument in the constructor.
2. Add `deployer.deploy(MyContract, arg0, { overwrite: false })` to migrations setup.
3. Execute migration command.
4. Wait for the extra constructor argument exception to be thrown.

* The value of `arg0` is optional. (Except `null` and `undefined` which throws another exception)

### DIFF
--- a/packages/deployer/src/deployment.js
+++ b/packages/deployer/src/deployment.js
@@ -29,7 +29,7 @@ class Deployment {
   /**
    * Helper to parse a deploy statement's overwrite option
    * @private
-   * @param  {Arry}    args        arguments passed to deploy
+   * @param  {Array}    args       arguments passed to deploy
    * @param  {Boolean} isDeployed  is contract deployed?
    * @return {Boolean}             true if overwrite is ok
    */
@@ -39,7 +39,6 @@ class Deployment {
 
     const overwrite = isObject && isDeployed && lastArg.overwrite === false;
 
-    isObject && delete lastArg.overwrite;
     return !overwrite;
   }
 
--- a/packages/deployer/src/deployment.js
+++ b/packages/deployer/src/deployment.js
@@ -29,9 +29,9 @@ class Deployment {
   /**
    * Helper to parse a deploy statement's overwrite option
    * @private
-   * @param  {Array}    args       arguments passed to deploy
-   * @param  {Boolean} isDeployed  is contract deployed?
-   * @return {Boolean}             true if overwrite is ok
+   * @param  {Array}   args       arguments passed to deploy
+   * @param  {Boolean} isDeployed is contract deployed?
+   * @return {Boolean}            true if overwrite is ok
    */
   _canOverwrite(args, isDeployed) {
     const lastArg = args[args.length - 1];


haltman-at:
Thanks for catching this!

I'm wondering though if this is actually the best way to handle this.  I'm assuming that `overwrite` is deleted because it's not actually a standard option.  That said, web3 doesn't actually care if you include nonstandard options, so not deleting it is fine.  But also, a more comprehensive solution in case we ever do have to worry about that might be to still delete `overwrite`, but to go further and change `{}` to `undefined` if this results in `{}`.  Such a solution as that obviously could not live entirely within this function, however.

So probably that's more trouble than it's worth and what you did is the best way.  Wondering if say @gnidan has an opinion though.

(Also, wow, that `overwrite` variable is named backwards. O_o  But that's not your fault obviously, that was pre-existing, you don't need to change that...)

haltman-at (approved):

packages/deployer/src/deployment.js
```diff
@@ -39,7 +39,6 @@ class Deployment {
 
     const overwrite = isObject && isDeployed && lastArg.overwrite === false;
 
-    isObject && delete lastArg.overwrite;
```
cds-amal, line 42: What an odd side effect! 😞

cds-amal (commented):
Thanks for catching this @ebrahimmfadae. It makes sense here, but I'm unable to reproduce the issue you described with a contract that requires constructor arguments. Can you share reproduction steps, please?

cds-amal:
> (Also, wow, that `overwrite` variable is named backwards. O_o But that's not your fault obviously, that was pre-existing, you don't need to change that...)

OMG, I stared at that for a really long time! The `overwrite` variable and it's negated return is confusing!

packages/deployer/src/deployment.js
ebrahimmfadae: Thanks for your response. Reproducing steps added to the description.

ebrahimmfadae:
> Thanks for catching this!
> 
> I'm wondering though if this is actually the best way to handle this. I'm assuming that `overwrite` is deleted because it's not actually a standard option. That said, web3 doesn't actually care if you include nonstandard options, so not deleting it is fine. But also, a more comprehensive solution in case we ever do have to worry about that might be to still delete `overwrite`, but to go further and change `{}` to `undefined` if this results in `{}`. Such a solution as that obviously could not live entirely within this function, however.
> 
> So probably that's more trouble than it's worth and what you did is the best way. Wondering if say @gnidan has an opinion though.
> 
> (Also, it's scary how this wasn't a pure function before, so I guess it's good you changed that even if we want to do more! ...you know what, I'm going to go ahead and hit approve on this, I think it's best to get this in even if potentially we might want to improve it later.)
> 
> (Also, wow, that `overwrite` variable is named backwards. O_o But that's not your fault obviously, that was pre-existing, you don't need to change that...)

Thanks for your response. I agree with you that this function must be pure. Also, I have another suggestion. We can clone the `options` object and prune it right before passing it to the web3's deploy function.

packages/deployer/src/deployment.js
cds-amal: Thank you! I was able to reproduce the error and verify the fix! Previously, I had a `value` property in the `txParams` object.

cds-amal (approved):
Looks good! Thanks, @ebrahimmfadae.

PR #5451, opened 2022-08-17
Add --from to global command options

sukanyaparashar:
This PR adds `--from` option to truffle global command options list, which can be used to specify an account to make calls or send transactions. This is related to #5330, where it was discussed to be a separate PR.

### DIFF
--- a/packages/core/lib/global-command-options.js
+++ b/packages/core/lib/global-command-options.js
@@ -7,7 +7,7 @@ const options = {
   from: {
     option: "--from <account>",
     description:
-      "Specify the account from which to make calls or send transactions."
+      "Specify the default account from which to make calls or send transactions."
   },
   config: {
     option: "--config <file>",
--- a/packages/core/lib/global-command-options.js
+++ b/packages/core/lib/global-command-options.js
@@ -4,6 +4,11 @@ const options = {
     description:
       "Specify the network to use. Network name must exist in the configuration."
   },
+  from: {
+    option: "--from <account>",
+    description:
+      "Specify the account from which to make calls or send transactions."
+  },
   config: {
     option: "--config <file>",
     description:


packages/core/lib/global-command-options.js
```diff
@@ -4,6 +4,11 @@ const options = {
     description:
       "Specify the network to use. Network name must exist in the configuration."
   },
+  from: {
+    option: "--from <account>",
+    description:
+      "Specify the account from which to make calls or send transactions."
```
gnidan, line 10: Not sure if it's worth adding this word here:

```suggestion
      "Specify the default account from which to make calls or send transactions."
```

gnidan (approved):
Leaving this approval for you after you decide how you feel about my suggestion :)

gnidan:
Although actually we may want this PR to include the removal of `from` from the options in each command separately

packages/core/lib/global-command-options.js
sukanyaparashar: Adding the `default` word makes sense. Thanks for the suggestion, @gnidan.

sukanyaparashar:
> Although actually we may want this PR to include the removal of `from` from the options in each command separately

We have the `allowedGlobalOptions` property for each command. Am I missing something what you meant by removal of `--from` from the options separately, @gnidan?

sukanyaparashar:
Hey @gnidan. I checked the commands in the `core` package and couldn't find any other command having `--from` option. I think `truffle call` is the only one that will use this option for now.

gnidan:
> Hey @gnidan. I checked the commands in the `core` package and couldn't find any other command having `--from` option. I think `truffle call` is the only one that will use this option for now.

Oh, interesting. Most commands do support this flag today... like `truffle test --from=<account>` will override things, but I guess it's just not listed as an explicit option everywhere, so you'll get a warning. This is probably fine for now, but might want to revisit this later.

sukanyaparashar:
I didn't know about that. Thanks @gnidan for letting me know about it. I think we should later on(in a separate PR) add the`--from` option to the `allowedGlobalOptions` array for the options that support it.

gnidan (approved):

Issue #5452, opened 2022-08-17
Truffle scenario tests exercises `@truffle/preserve`, but that package is no longer part of the truffle monorepo.


benjamincburns:
This is currently blocking #5335 (and PR #5367) due to there being scenario tests in this repo that are meant to be testing the `@truffle/preserve-*` plugins.

Specifically in this case, `@truffle/preserve` depends on `@truffle/error`, and #5367 introduces breaking changes to that package. As a result, there's a chicken/egg problem where we have to publish an update for `@truffle/error` in order to publish an update for `@truffle/preserve`, but with the current state of affairs I can't publish `@truffle/error` because the integration tests for the current version of `@truffle/preserve` produce a red build of `@truffle/error`.

I could work around this by avoiding the breaking change in #5367 (and I may do that as well), but this issue will just pop up again in some other way. The underlying problem is that we effectively haven't decoupled `truffle` from `@truffle/preserve` until we've stopped testing that the current version of `@truffle/preserve` integrates with in-flight dev changes _from within in the `truffle` repo_.

PR #5455, opened 2022-08-18
Revive `@truffle/preserve`

benjamincburns:
Brings `@truffle/preserve` back into the project, but leaves the plugins in the [trufflesuite/preserves](https://github.com/trufflesuite/preserves) repo. Attempts to preserve history where possible, although it gets a bit weird in places.

Addresses #5452.


benjamincburns:
This is a draft for the moment as I need to rework the history one more time. I just wanted to push _something_ up that matches what I expect the final state will be so that I can start getting CI feedback.

benjamincburns:
This is ready for review now, I think. I'll push a second PR that removes the `spinnies` dep in favor of `@truffle/spinners`, but otherwise that should be it 🤞

cds-amal (approved):
Thanks, @benjamincburns. Looks good to me!

PR #5456, opened 2022-08-18
Update spinners to latest spinnies version

benjamincburns:
The latest version of @trufflesuite/spinnies includes a few minor typing bug fixes. This change updates to the latest version of that package, and fixes a few other typing errors.

This change is necessary to remove the `spinnies` library from the `preserve` package, which makes it consistent with our other spinner logic (properly stacked spinners, etc). Unfortunately when attempting that change without this one, I run into a typing error that this change fixes.

Also introduces an `indent` property that should've been added to the original version of this package.


cliffoo (approved):
Looks good to me! Thanks @benjamincburns. Just a few questions from my side:

- For `stoppedPrefix` and the new `indent` methods, why allow get / set `undefined`? Why not be strict like the other methods (just `string`)?
- Speaking of arg / returns type, would `Color | string` be better than just `string`?
- Is the optional chaining after `spinnies.pick` necessary? Does `spinnies` ever remove `this.name`?

Apologies if the answers to these questions are trivial / obvious due to my lack of context!

benjamincburns:
> Apologies if the answers to these questions are trivial / obvious due to my lack of context!

No need to apologize, @cliffoo! Thanks for the review!

Personally I feel somewhat strongly that reviewers should review what's in front of them and if they need additional context it's totally fine to ask as many questions as you like!

> * For `stoppedPrefix` and the new `indent` methods, why allow get / set `undefined`? Why not be strict like the other methods (just `string`)?

Assigning `undefined` to these properties just causes `spinnies` to revert to its internal defaults for that option. For example, here's [the application of the default for `stoppedPrefix`](https://github.com/trufflesuite/spinnies/blob/6070d82185a55896baf3f602962eadd1ad65a374/index.js#L260-L261), and [here's the application of the default for `indent`](https://github.com/trufflesuite/spinnies/blob/6070d82185a55896baf3f602962eadd1ad65a374/index.js#L238)).

> * Speaking of arg / returns type, would `Color | string` be better than just `string`?

The properties that I changed from `Color | string` to `string` were just incorrectly typed to begin with - probably bad copypasta on my part. The `Color` type is just a union of string literal types that match up to the color strings in the version of `chalk` that `spinnies` uses. Those "prefix" properties don't set the color of the prefix, but the actual prefix text itself (e.g. ❌  is the default for `failPrefix`, IIRC).


> * Is the optional chaining after `spinnies.pick` necessary? Does `spinnies` ever remove `this.name`?

Yeah, when the `remove` method is called on the `Spinner` object, it calls `spinnies.remove(this.name)`, and that deletes all knowledge of that spinner from spinnies's internal module-level state. If you then call `spinnies.pick(this.name)`, it returns `undefined` rather than throwing.

PR #5458, opened 2022-08-18
Internal improvement: Convert workflow-compile to TypeScript

eggplantzzz:


eggplantzzz:
I'm wondering if it is better to do the initial TS conversion work first and then refine later in a separate PR, or whether I should just refine in the same PR. Any opinions on this?

haltman-at:
CI didn't run...?

cds-amal:
> I'm wondering if it is better to do the initial TS conversion work first and then refine later in a separate PR, or whether I should just refine in the same PR. Any opinions on this?

Conversion then refinement is my vote.

packages/workflow-compile/legacy/index.js
```diff
@@ -3,15 +3,37 @@ const fse = require("fs-extra");
...
+
+  if (!config.resolver) config.resolver = new Resolver(config);
+
+  if (!config.artifactor) {
+    config.artifactor = new Artifactor(config.contracts_build_directory);
+  }
+
+  return config;
+}
+
```
cliffoo, line 36: This appears to be identical to `prepareConfig` in utils.js. Maybe this legacy file can also be ts so it can import instead? Save some imports and duplicate code.

cliffoo (commented):
Nice one @eggplantzzz ! Just one comment about some duplicate code. Are there reasons to keep files under /legacy js? If not I think we should convert that too. But... If there are reasons, then maybe extract that function as a separate js file to be imported by both src/utils.ts and legacy/index.js? Your call!

I think the `require("@truffle/workflow-compile").default` is fine. One of these days, the core package will be ts!

packages/workflow-compile/legacy/index.js
eggplantzzz: So the legacy stuff is eventually going to be removed. I think it might be a breaking change to do so currently because some people might be using it. My approach was to leave it alone quarantined in its own folder (as it is sort of deprecated) as there doesn't seem to be much value in updating it if it is going to be removed the next time we are able to create a breaking change. What do you think?

cliffoo: That's fair. I guess we just have to make a note of this and chuck it when workflow-compile is ready for a major bump.

cliffoo (approved):
Looks like there's a conflict because v5.5.30 was released, but otherwise everything LGTM! Thanks @eggplantzzz 👍

PR #5463, opened 2022-08-18
Fix typo in `package.json`

cliffoo:


packages/dashboard-message-bus-common/package.json
```diff
@@ -17,7 +17,7 @@
   "repository": {
     "type": "git",
     "url": "https://github.com/trufflesuite/truffle.git",
-    "directory": "packages/dashboard-mesasge-bus"
+    "directory": "packages/dashboard-message-bus"
```
haltman-at, line 20: Oh hey while you're at it, this should be `dashboard-message-bus-common`, right?

cliffoo: You're right, I need 💤

sukanyaparashar (approved):
This looks good! @cliffoo

haltman-at (approved):

PR #5465, opened 2022-08-19
Internal improvement: Move selector computation from `codec` into `abi-utils`

haltman-at:
This PR moves code for computing signatures and selectors from `codec` into `abi-utils`.  Note that we continue to re-export these functions from Codec, in order to avoid a breaking change.  I also added a few tests of these functions in `abi-utils` -- not property-based, sorry, I couldn't figure out a good way to do that.  On top of that, I expanded the `abi-utils` README to cover this new functionality.

Since this PR is mostly just moving code around and not changing it, there's not a lot to say here about the code itself.  However, I wanted to record why I'm doing this.

Basically, it comes back to the abi-to-sol problem.  I want to be able to make breakiing changes to Codec without a repeat of the problem that caused last time (see #5396 for an explanation).  This is the only code from Codec that abi-to-sol uses.  Moving it to abi-utils will of course not get rid of the problem, but it reduces it, since it means that the problem will only come up when we break abi-utils rather than when we break codec.  Since we break abi-utils hardly ever, and break codec somewhat frequently, this acts as a stopgap to allow us to avoid problems until we figure out a better solution.

### DIFF
--- a/packages/abi-utils/lib/signature.ts
+++ b/packages/abi-utils/lib/signature.ts
@@ -1,0 +1,46 @@
+import { Parameter, FunctionEntry, EventEntry, ErrorEntry } from "./types";
+import { soliditySha3 } from "web3-utils";
+
+export const ShortSelectorSize = 4;
+
+//NOTE: this function returns the written out SIGNATURE, not the SELECTOR
+export function abiSignature(
+  abiEntry: FunctionEntry | EventEntry | ErrorEntry
+): string {
+  return abiEntry.name + abiTupleSignature(abiEntry.inputs);
+}
+
+export function abiTupleSignature(parameters: Parameter[]): string {
+  const components = parameters.map(abiTypeSignature);
+  return "(" + components.join(",") + ")";
+}
+
+export function abiTypeSignature(parameter: Parameter): string {
+  const tupleMatch = parameter.type.match(/tuple(.*)/);
+  if (tupleMatch === null) {
+    //does not start with "tuple"
+    return parameter.type;
+  } else {
+    const tail = tupleMatch[1]; //everything after "tuple"
+    const tupleSignature = abiTupleSignature(
+      parameter.components as Parameter[]
+    ); //it won't be undefined
+    return tupleSignature + tail;
+  }
+}
+
+export function abiSelector(
+  abiEntry: FunctionEntry | EventEntry | ErrorEntry
+): string {
+  const signature = abiSignature(abiEntry);
+  //NOTE: web3's soliditySha3 has a problem if the empty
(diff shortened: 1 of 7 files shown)


cds-amal:
>Basically, it comes back to the abi-to-sol problem. I want to be able to make breakiing changes to Codec without a repeat of the problem that caused last time (see https://github.com/trufflesuite/truffle/pull/5396 for an explanation). This is the only code from Codec that abi-to-sol uses. Moving it to abi-utils will of course not get rid of the problem, but it reduces it, since it means that the problem will only come up when we break abi-utils rather than when we break codec. Since we break abi-utils hardly ever, and break codec somewhat frequently, this acts as a stopgap to allow us to avoid problems until we figure out a better solution.

This makes sense! Good call.

haltman-at:
More specifically the reason it came up is because I'm planning a substantial Codec breakage to deal with this internal function pointers problem. :P

Btw, you may notice I added the `new-feature: @truffle/abi-utils` label and then removed it.  I added it because, duh, it's a new feature for abi-utils!  I removed it because, oh right, abi-utils is still in 0.x, so there's no distinction between a new feature and a breaking change... so if we increment the minor version number, we'll hit the exact problem I'm trying to avoid. >_>  So much for semver...

eggplantzzz (approved):
I like it

eggplantzzz:
I didn't scour the code that looks like it was moved - are all those utility functions just moved wholesale? There were no changes to them?

packages/abi-utils/lib/signature.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function abiTupleSignature(parameters: Parameter[]): string {
+  const components = parameters.map(abiTypeSignature);
+  return "(" + components.join(",") + ")";
+}
+
+export function abiTypeSignature(parameter: Parameter): string {
+  const tupleMatch = parameter.type.match(/tuple(.*)/);
+  if (tupleMatch === null) {
+    //does not start with "tuple"
```
cds-amal, line 21: The comment says does not start with `tuple` but the regex doesn't use `^`. Is that correct?

cds-amal (commented):
I have a question about the regex type matching, but otherwise looks good!

packages/abi-utils/lib/signature.ts
haltman-at: Oops!  Good catch.  This code will work anyway, because there's no other place that `tuple` can occur, but I should change it to not rely on that, thanks.

haltman-at:
Putting the label back on at @gnidan's insistence.

PR #5469, opened 2022-08-19
Enhancement: Add gnosis network to etherscan fetcher

haltman-at:
Etherscan has created gnossiscan.io for the Gnosis blockchain, so, this PR adds support for it.  Note that we had the chain listed as "xdai" but that name seems to be out of date so I changed it to "gnosis" (which meant updating sourcify.ts as well).

Also prettier triggered again but hopefully this is the last time!


sukanyaparashar (approved):
Looks good!

cds-amal (approved):

PR #5470, opened 2022-08-19
Compiler plugin architecture

fainashalts:
This PR sets the stage for compiler plugins. If a developer wishes to use a compiler that is not currently supported by Truffle core, they will be able to bring it themselves, via a plugin!

In order to work, a compiler plugin will need to expose particular compiler functions, namely:  `all()`, `sourcesWithDependencies()`, and `necessary()`. 

To indicate that a compiler plugin should be used, indicate so in project config: 

```
plugins: ["compile-zksolc"],
compilers: {
// this information should provide anything the plugin needs to use the compiler plugin, like version
}
```

This is still a draft, and there may be significant issues that I haven't considered. Creating this Draft PR so that @dongmingh can test with the plugin architecture when he builds the `compile-zksolc` plugin.

### DIFF
--- a/packages/plugins/.eslintrc.json
+++ b/packages/plugins/.eslintrc.json
@@ -1,0 +1,3 @@
+{
+  "extends": ["../../.eslintrc.package.json"]
+}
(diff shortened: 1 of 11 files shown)


eggplantzzz:
We need to re-evaluate whether this work is still a good start to this work or whether we should begin afresh.

PR #5479, opened 2022-08-22
Internal improvement: Properly time REPL inputs

cds-amal:
Currently, `runInREPL` sends all queued commands to the child process on the first detected REPL prompt and then closes the input stream, which terminates the child process (bad).

This PR reworks the logic to send one queued command per REPL prompt.

### DIFF
--- a/packages/truffle/test/scenarios/commandRunner.js
+++ b/packages/truffle/test/scenarios/commandRunner.js
@@ -50,7 +50,7 @@ module.exports = {
     inputCommands = [],
     config,
     executableCommand,
-    executableArgs,
+    executableArgs = "",
     displayHost
   } = {}) {
     const cmdLine = `${this.getExecString()} ${executableCommand} ${executableArgs}`;
@@ -63,13 +63,12 @@ module.exports = {
     // seems safe to escape parens only, as the readyprompt is constructed from
     // [a-zA-Z] strings and wrapping parens.
     const escapedPrompt = readyPrompt.replace("(", "\\(").replace(")", "\\)");
-    const readyPromptRex = new RegExp(`^${escapedPrompt}`, "gm");
+    const readyPromptRex = new RegExp(`^${escapedPrompt}`, "m");
 
     let outputBuffer = "";
 
     return new Promise((resolve, reject) => {
       const child = exec(cmdLine, { cwd: config.working_directory });
-      let numSeenPrompts = 0;
 
       if (child.error) return reject(child.error);
 
@@ -81,10 +80,11 @@ module.exports = {
         // accumulate buffer from chunks
         outputBuffer += data;
 
-        // count prompt
-        const foundPrompts = (outputBuffer.match(readyPromptRex) || []).length;
-        if (foundPrompts > numSeenPrompts) {
-          numSeenPrompts = foundPrompts;
+        if (readyPromptRex.test(outputBuffer)) {
+          // set outputBuffer to remaining segment after final match
+          const segments = outputBuffer.split(readyPromptRex);
+          outputBuffer = segments.pop();
+
           if (inputCommands.length === 0) {
             // commands exhausted, close stdin
             child.stdin.end();
@@ -94,7 +94,6 @@ module.exports = {
             child.stdin.write(nextCmd + EOL);
           }
         }
-
         config.logger.log("OUT: ", data);
       });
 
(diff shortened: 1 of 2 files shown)


packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -74,17 +79,20 @@ module.exports = {
...
-        if (!seenChildPrompt && outputBuffer.includes(readyPrompt)) {
-          seenChildPrompt = true;
-          inputCommands.forEach(command => {
-            child.stdin.write(command + EOL);
-          });
-          child.stdin.end();
+        // count prompt
+        const foundPrompts = (outputBuffer.match(readyPromptRex) || []).length;
+        if (foundPrompts > numSeenPrompts) {
+          numSeenPrompts++;
```
davidmurdoch, line 87: This should probably be:

```suggestion
          numSeenPrompts = foundPrompts;
```

cds-amal: The most correct! Thanks, @davidmurdoch.

packages/truffle/test/scenarios/commands/develop.js
```diff
@@ -25,20 +25,29 @@ describe("truffle develop", function () {
...
-        inputCommands: [input],
+        inputCommands: inputs,
         config,
         executableCommand: "develop",
         displayHost: "develop"
       });
       output = logger.contents();
     });
 
+    it("Sends multiple commands to REPL", () => {
```
cds-amal, line 44: I landed on placing a test here because it's quicker to piggyback in this scenario than breakout another file to validate our testing library. Please let me know if you think this belongs elsewhere.

packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -74,17 +79,20 @@ module.exports = {
...
-        if (!seenChildPrompt && outputBuffer.includes(readyPrompt)) {
-          seenChildPrompt = true;
-          inputCommands.forEach(command => {
-            child.stdin.write(command + EOL);
-          });
-          child.stdin.end();
+        // count prompt
+        const foundPrompts = (outputBuffer.match(readyPromptRex) || []).length;
+        if (foundPrompts > numSeenPrompts) {
+          numSeenPrompts = foundPrompts;
```
cliffoo, line 87: Might this be better? It
- Doesn't tire as `outputBuffer` grows
- Removes equivocation between `found` and `numSeen`

```suggestion
        const numNewPrompts = (data.match(readyPromptRex) || []).length;
        if (numNewPrompts > 0) {
          numSeenPrompts = += numNewPrompts;
```

packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -60,11 +60,16 @@ module.exports = {
         ? `debug(${displayHost})>`
         : `truffle(${displayHost})>`;
 
-    let seenChildPrompt = false;
+    // seems safe to escape parens only, as the readyprompt is constructed from
+    // [a-zA-Z] strings and wrapping parens.
+    const escapedPrompt = readyPrompt.replace("(", "\\(").replace(")", "\\)");
+    const readyPromptRex = new RegExp(`^${escapedPrompt}`, "gm");
```
sukanyaparashar, line 66: Sorry @cds-amal if it is a silly question, but what is this `"gm"` flag does? `"g"` is for global search and `"m"` is multi-line search if I am understanding correctly.

cds-amal: Yes, "g" is for global. It means the meta characters, `^`, and `$` will match the beginning and end of each line instead of the beginning and end of the input string (which can contain multiple lines)

packages/truffle/test/scenarios/commandRunner.js
davidmurdoch: @cliffoo `data` could be a _partial_ line, so this method could potentially miss lines completely.

In other words: the `data` event could be emitted multiple times for a line like "truffle(develop)>\n":

1. `truffle<de`
2. `velop>\n`

davidmurdoch (approved):

packages/truffle/test/scenarios/commandRunner.js
```diff
@@ -74,19 +78,22 @@ module.exports = {
...
-          seenChildPrompt = true;
-          inputCommands.forEach(command => {
-            child.stdin.write(command + EOL);
-          });
-          child.stdin.end();
-        }
+        if (readyPromptRex.test(outputBuffer)) {
+          // set outputBuffer to remaining segment after final match
+          const segments = outputBuffer.split(readyPromptRex);
+          outputBuffer = segments.pop();
```
davidmurdoch, line 86: Truffle's test data won't fail on this, but as a general purpose solution this might not work. For example:

```
`this is a line
truffle(develop)>
this is a line
truffle(develop)>
this is a line
`
```

You'll skip one of the `truffle(develop)>` lines, whereas I think you previous version would have properly handled it.

PR #5481, opened 2022-08-22
Internal improvement: Update @truffle/migrate to TypeScript

eggplantzzz:
**Note**: This is a breaking change for migrate due to the necessity of changing `require`-style imports as follows:
```
const Migrate = require("@truffle/migrate");

// the above now needs to be changed to
const Migrate = require("@truffle/migrate").default;
```


packages/migrate/src/index.ts
```diff
@@ -1,22 +1,23 @@
...
  *  the `.runMigrations` method.
  */
-const Migrate = {
+export default {
   Migration: Migration,
   logger: null,
 
   promptToAcceptDryRun: async function (options) {
-    const prompt = [
+    const prompt: Question[] = [
```
cds-amal, line 20: Should `options` above be typed? Ah, just noticed `promptToAcceptDryRun` is invoked without `options`.

packages/require/lib/types.ts
```diff
@@ -14,7 +14,7 @@ export interface RequireOptions {
   /**
    * The `@truffle/config` for the user's project
    */
-  config: TruffleConfig;
+  config?: TruffleConfig;
```
cds-amal, line 17: Can you explain why this is optional?

cds-amal (commented):
Thanks, @eggplantzzz. It seems straightforward enough (though I'm not TS maven). I did notice most of the exported functions in `migrate/src/index.ts` are untyped and default to implicit `any`.  Do we care about this? ~Hrm, there's no tsconfig file~. Perhaps `noImplicitAny` should be set.

benjamincburns:
> Perhaps `noImplicitAny` should be set.

So I was out in the garden and I felt a weird sensation... I ran inside to check, and sure enough, someone was talking about `noImplicitAny`!

Kidding aside, I just wanted to pipe up in favor of everything using `noImplicitAny` and all exports being typed. Contract stuff is pretty much impossible to type properly at the moment, but otherwise there's really not much value of converting to TS if we're not going to make use of the type system.

packages/core/lib/commands/migrate/run.js
```diff
@@ -38,6 +38,7 @@ module.exports = async function (options) {
     );
     if (proceed) await runMigrations(preparedConfig);
   } else {
+    console.log("about to run them");
```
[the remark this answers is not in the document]
eggplantzzz: oh whoops, this is definitely my test code

eggplantzzz:
Yeah I'm not sure how far to take this. I was going to do the bare minimum to convert and decided I would consider adding `noImplicitAny` to be a "refinement" :) I'll go ahead and see how far I can take it with `noImplicitAny` set to `true`.

packages/require/lib/types.ts
eggplantzzz: Because that is how we use it https://github.com/trufflesuite/truffle/blob/develop/packages/migrate/Migration.js#L35-L42

packages/migrate/src/index.ts
eggplantzzz: I think it probably should. It looks like it is a mistake not to include it because `emitEvent` uses it. Is this currently broken?

benjamincburns:
> Yeah I'm not sure how far to take this. I was going to do the bare minimum to convert and decided I would consider adding `noImplicitAny` to be a "refinement" :) I'll go ahead and see how far I can take it with `noImplicitAny` set to `true`.

Yeah, I think in this case it's best to still set `noImplicitAny` to `true`, but then _explicitly_ use the `any` and `unknown` (with preference for the latter in exports) so that 1) the problem doesn't grow over time, and 2) they're easier to spot for said future refinements.

Don't let me bully you into it, though - if things are super ugly with `noImplicitAny`, it's still better to just do what you can for now.

packages/truffle/test/scenarios/commands/migrate.js
```diff
@@ -19,7 +25,13 @@ describe("truffle migrate", () => {
 
   describe("when run on the most basic truffle project", () => {
     it("doesn't throw", async () => {
-      await CommandRunner.run("migrate", config);
+      try {
+        await CommandRunner.run("migrate", config);
+      } catch (error) {
+        console.log("the logged stuff -- %o", config.logger.loggedStuff);
```
fainashalts, line 31: Did you mean to leave this? If so I'd probably leave a more descriptive message in the logging? What is config.logger.loggedStuff usually conveying?

fainashalts (commented):
I think I'm also in the `noImplicitAny` camp. I think its ok to have implicit any if its really hard to convert everything, but I'd say add it to the tsconfig and see how bad it is maybe? If its a mess then it makes more sense to leave it as follow-on work.

packages/migrate/src/index.ts
cds-amal: I'm unsure if it is broken, though we should raise an issue and fix in follow-up work.

packages/require/lib/types.ts
cds-amal: thanks!

cds-amal (commented):
Just a partial review. I didn't notice CI was still red.

packages/db/package.json
```diff
@@ -64,6 +64,7 @@
     "@truffle/compile-common": "^0.8.1",
     "@truffle/contract-schema": "^3.4.10",
     "@truffle/resolver": "^9.0.16",
+    "@truffle/migrate": "^3.3.9",
```
benjamincburns, line 67:
```suggestion
    "@truffle/migrate": "^3.3.10",
```

eggplantzzz: Thanks!

benjamincburns: I'm really not sure how you managed to commit this, as the version check (see husky stuff in top-level package.json) should've caught it 🤔

benjamincburns: @eggplantzzz can you please confirm whether you committed with the `--no-verify` flag? If not, I'll raise a new issue for myself to go and fix that script.

packages/migrate/test/index.ts
```diff
@@ -1,8 +1,14 @@
...
+import Config from "@truffle/config";
+import assert from "assert";
+import * as sinon from "sinon";
+import { Migration } from "../src/Migration";
+let options: Config, migrations: Migration[], fakeMigration: Migration;
+
+const unstub = (stubbedThing: object, methodName: string) => {
+  // @ts-ignore
+  stubbedThing[methodName].restore();
+};
```
dongmingh, line 11: I like the idea of unstub to pair up with sinon.stub.  This makes the code easy to follow.

eggplantzzz: For some reason `sinon` stubs don't seem to play well with TypeScript. I actually made this method so I would only have to write `// @ts-ignore` once instead of 20 times lol

packages/migrate/package.json
```diff
@@ -25,13 +26,19 @@
     "@truffle/expect": "^0.1.3",
     "@truffle/interface-adapter": "^0.5.22",
     "@truffle/require": "^2.1.7",
+    "@truffle/resolver": "^9.0.17",
```
cds-amal, line 29: Is `debug` used in this package?

packages/truffle/test/scenarios/commands/migrate.js
```diff
@@ -19,7 +25,13 @@ describe("truffle migrate", () => {
 
   describe("when run on the most basic truffle project", () => {
     it("doesn't throw", async () => {
-      await CommandRunner.run("migrate", config);
+      try {
+        await CommandRunner.run("migrate", config);
+      } catch (error) {
+        console.log("the logger contents -- %o", config.logger.loggedStuff);
+        console.log("the following error occurred -- %o", error.message);
+        assert.fail();
```
cds-amal, line 33: Nice!

cds-amal (commented):
Nice @eggplantzzz, I have a few questions.

packages/box/package.json
```diff
@@ -27,13 +27,13 @@
...
     "ignore": "^5.1.8",
-    "inquirer": "8.2.2",
+    "inquirer": "8.2.4",
     "tmp": "^0.2.1",
     "vcsurl": "^0.1.1"
   },
   "devDependencies": {
     "@types/fs-extra": "^8.1.0",
-    "@types/inquirer": "6.0.3",
+    "@types/inquirer": "9.0.1",
```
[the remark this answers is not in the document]
eggplantzzz: I'm into the idea of, where possible, using the same types package version. Although we won't be able to judiciously apply this principal everywhere; for example, we use @types/web3 version `^1.0.20`. Seems like searching about this yields results saying something like "they should correspond, but they are best effort attempts at maintaining parity". I'll go ahead and give the types package the same version and ensure that it passes.

packages/core/lib/commands/migrate/prepareConfigForRealMigrations.js
```diff
@@ -1,7 +1,7 @@
 module.exports = async function (buildDir, options) {
   const Artifactor = require("@truffle/artifactor");
   const { Resolver } = require("@truffle/resolver");
-  const Migrate = require("@truffle/migrate");
+  const Migrate = require("@truffle/migrate").default;
```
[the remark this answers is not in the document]
eggplantzzz: Looks like it does to me.

packages/db/typings/schema.d.ts
```diff
@@ -0,0 +1,826 @@
+// tslint-disable
+// graphql typescript definitions
+
```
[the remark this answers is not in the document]
eggplantzzz: Actually the path to this file was updated. Previously, the TS typings folder was named `types`. To be consistent across packages I renamed it `typings`.

packages/migrate/package.json
eggplantzzz: nope, doesn't look like it! I'll get rid of it

packages/box/package.json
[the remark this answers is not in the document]
cds-amal: I like the idea of using the same types package version when possible, given the best effort attempts to maintain parity.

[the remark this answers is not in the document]
eggplantzzz: Actually looks like there are 2 versions of inquirer. I vote we consolidate these in a future PR. For now I say we leave this type upgrade as is. Opposed?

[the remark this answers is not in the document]
cds-amal: Agreed!

packages/migrate/src/index.ts
```diff
@@ -29,15 +33,17 @@ const Migrate = {
     if (answer.proceed) {
       return true;
     }
-    await emitEvent(options, "migrate:dryRun:notAccepted");
+    if (options) {
+      await emitEvent(options, "migrate:dryRun:notAccepted");
+    }
```
cliffoo, line 38: Why `if (options)`?

packages/migrate/src/index.ts
```diff
@@ -170,8 +180,8 @@ const Migrate = {
     };
   },
 
-  lastCompletedMigration: async function (options) {
-    let Migrations;
+  lastCompletedMigration: async function (options: Config): Promise<number> {
+    let Migrations: any; // I don't think we have a good type for this yet
```
cliffoo, line 184: Is it still ok to type this with `import type { Migration } from "./Migration"` for now?

packages/migrate/src/index.ts
```diff
@@ -182,14 +192,14 @@ const Migrate = {
...
-    const migrationsOnChain = async migrationsAddress => {
+    const migrationsOnChain = async (migrationsAddress: string) => {
       return (
         (await Migrations.interfaceAdapter.getCode(migrationsAddress)) !== "0x"
       );
     };
 
     // Two possible Migrations.sol's (lintable/unlintable)
-    const lastCompletedMigration = migrationsInstance => {
+    const lastCompletedMigration = (migrationsInstance: any) => {
```
cliffoo, line 202: Same as above

cliffoo (approved):
So this migrates our code to... migrate? LGTM! Thanks @eggplantzzz

packages/migrate/src/index.ts
eggplantzzz: No, `Migrations` here refers to a Truffle contract abstraction which we don't really have a type for yet.

packages/migrate/src/index.ts
eggplantzzz: `options` in this method is an optional parameter so we need to make sure it is defined.

packages/migrate/src/index.ts
eggplantzzz: Same response as the one related to typing `Migrations`.

cds-amal (approved):
Muy bien @eggplantzzz, gracias!

packages/migrate/src/emitEvent.ts
```diff
@@ -0,0 +1,7 @@
+import type Config from "@truffle/config";
```
dongmingh, line 1: You use `import type` here but not on other file for the same Config. Is there anything special here?

eggplantzzz: Yes, in this file the class `Config` is not used. The only thing used here is the type `Config`. This is useful because the types do not make it into the compiled, distributed files.

PR #5484, opened 2022-08-23
Internal improvement: Remove faker from abi-utils and generate arbitraries manually

eggplantzzz:
One of the MetaMask team pointed out that faker is a big dependency of abi-utils and asked Truffle to remove it as they are trying to cut down the size of their extension. Truffle does not really use faker for much besides generating random names for when it creates arbitrary abis. This PR implements a simple utility that generates these names manually.


packages/abi-utils/lib/arbitrary.ts
```diff
@@ -393,29 +392,46 @@ const reservedWords = new Set([
...
 
-// borrowed from https://runkit.com/dubzzz/faker-to-fast-check
-const fakerToArb = (template: string, transform = camelCase) => {
+type WordListKey = keyof typeof wordLists;
+const getArb = (
+  wordTypes: WordListKey[],
+  transform = camelCase
+): fc.Arbitrary<string> => {
+  const results: fc.Arbitrary<string>[] = [];
+  for (const wordType of wordTypes) {
```
haltman-at, line 401: This seems overly complicated to me in two ways:

1. Rather than having these separate noun-vs-verb branches, you could just use `wordLists[wordType]` everywhere with no branch.
2. Instead of doing `fc.integer(...).noBias().noShrink().map(...)`, you could just use `fc.constantFrom(...)`, if I'm understanding correctly?

It looks to me like doing those would simplify this function a lot.

haltman-at (commented):
Seems basically good, but one part seems overly complicated, you might want to simplify it.

Btw, some history you may be interested in: When I added errors, I was originally going to do them as adjective-noun, but I ultimately went with noun-noun because one of faker's adjectives was "1080p" and that obviously caused problems if it went first!  (Although I guess I could have filtered out things that started with digits.)

packages/abi-utils/lib/arbitrary.ts
gnidan: > 2\. Instead of doing `fc.integer(...).noBias().noShrink().map(...)`, you could just use `fc.constantFrom(...)`, if I'm understanding correctly?

OMG I was looking for this very thing!

@eggplantzzz I'll address this concern and push up a new commit, since this is my mess that Harry's showing issue with.

gnidan: Updated. Wow that's so much cleaner. Thanks @haltman-at!

packages/abi-utils/lib/arbitrary.ts
```diff
@@ -393,29 +392,30 @@ const reservedWords = new Set([
...
     { arbitrary: fc.constant(""), weight: 1 }
   );
-const EventName = () =>
-  fakerToArb("{{hacker.verb}} {{hacker.noun}}", pascalCase);
-const ErrorName = () =>
-  fakerToArb("{{hacker.noun}} {{hacker.noun}}", pascalCase);
-const FunctionName = () => fakerToArb("{{hacker.verb}} {{hacker.noun}}");
+const EventName = () => Name(["verb", "noun"], pascalCase);
+const ErrorName = () => Name(["noun", "noun"], pascalCase);
+const FunctionName = () => Name(["verb", "noun"]);
```
gnidan, line 418: I really like how this makes it seem like fast-check just wants to play Mad Libs

haltman-at:
(Note this needs to be rebased against develop to solve the yarncheck problem.)

haltman-at (approved):
Hooray!

PR #5485, opened 2022-08-24
Internal improvement: use `@truffle/spinners` in `@truffle/preserve` instead of `spinnies`

benjamincburns:
Makes the `@truffle/preserve` spinners code consistent with other spinners in the truffle codebase. This is necessary to prevent cases where the two different spinner libraries may have active spinners at the same time and therefore may overwrite one another.

Also removes the accidentally-committed `api.json` doc file. That file was ignored (but still accidentally committed) during the move of the `@truffle/preserve` module back into this monorepo.


packages/preserve/.eslintrc.json
```diff
@@ -0,0 +1,23 @@
...
+    "node": true
+  },
+  "overrides": [
+    {
+      "files": ["*.ts", "*.mts", "*.cts", "*.tsx"],
+      "rules": {
+        // ESLint can't resolve global types like Console, and TypeScript
+        // already errors when you use things that aren't defined, so this check
+        // can be safely disabled
+        "no-undef": "off"
```
benjamincburns, line 13: Per the comment this is a safe exclusion. If anyone has concerns, please see [the recommendation from ESLint to disable `no-undef` for TypeScript](https://github.com/typescript-eslint/typescript-eslint/blob/main/docs/linting/TROUBLESHOOTING.md#i-get-errors-from-the-no-undef-rule-about-global-variables-not-being-defined-even-though-there-are-no-typescript-errors)

haltman-at: Makes sense.  You think we should do this more generally in our Typescript packages?

haltman-at: (Also wondering if people will object to comments in `.eslintrc.json`. :P )

benjamincburns: > Makes sense. You think we should do this more generally in our Typescript packages?

Yes. I think it likely should just go into our base `.eslintrc.json` variants, with the file type guard in tact. Happy to PR that change if you're in support of it.

> (Also wondering if people will object to comments in .eslintrc.json. :P)

Haha, I had a similar thought when writing this, but I think if people object too strongly my response might be a light-hearted "come at me, bro!"

FWIW, comments are totally encouraged in `.eslintrc.json`, per [the eslint config file documentation](https://eslint.org/docs/latest/user-guide/configuring/configuration-files#comments-in-configuration-files). There's also a trend toward using [JSON5](https://json5.org/), for human-authored/human-readable files like this. For example, VSCode has supported JSON5 for its various config files for years. Sadly I don't think that `.eslint.json` supports full JSON5, but comment support is good enough, IMHO.

benjamincburns: > Yes. I think it likely should just go into our base .eslintrc.json variants, with the file type guard in tact. Happy to PR that change if you're in support of it.

Alternatively I can include that change in this PR if you think it's warranted? As far as I understand, it should be a completely benign change.

haltman-at: I think either way is fine!  I'm in favor of making the change but don't really have an opinion on how it should be done.

cds-amal: >Alternatively I can include that change in this PR if you think it's warranted? As far as I understand, it should be a completely benign change.

I think it should be a separate PR because it is a different concern from the intended scope of this one.

cds-amal, line 13: This is cool, and having comments is useful here!

cds-amal (approved):
Thanks, @benjamincburns. Looks good to me. I think promoting the `no_undef` should be a separate PR.

packages/preserve/.eslintrc.json
benjamincburns: Sure thing, thanks for looking this over!

PR #5492, opened 2022-08-25
Consolidate polyfills into new `@truffle/polyfills` package

benjamincburns:
- Establishes a repeatable pattern for adding and depending upon polyfills in our monorepo.
- Replaces usage of `node-abort-controller` with `@truffle/polyfills/nodeAbortController`
- Adds a polyfill for error causes to be used in an upcoming PR (#5367)


benjamincburns:
Sorry for the immediate force push - had to fix a spelling error in my commit message. This is good to review now.

benjamincburns:
I think I was a bit too ambitious on this one by pulling in the error chaining polyfill. It passes the tests I have here, but causes annoying test failures in #5367. Going to close this for now. Tomorrow I will pull the `errorCause` stuff out and reopen. Better to include that in #5367, as then I'll know it's working 100% correctly.

PR #5496, opened 2022-08-26
Add optimistic goerli to etherscan fetcher

haltman-at:
It looks like https://goerli-optimistic.etherscan.io/ is now up!  It doesn't seem to be entirely public yet, Etherscan has yet to link it from their other pages, like the main optimistic site... but it's there and it works. :)  So why not add it here? :)


cliffoo (approved):

cds-amal (approved):
LGTM! Glass half full, I say.

PR #5497, opened 2022-08-26
Add hecoinfo testnet back in to source-fetcher

haltman-at:
I removed this in #5411, but, uh, it's back now?  So uh yean, adding it back in.

(Noting once again that hecoinfo is no longer actually an Etherscan project, but it still continues to use the same API, so we can still use our Etherscan fetcher with it, unlike say with hooscan.)


cds-amal (approved):
Looks good to me.

Issue #5203, opened 2022-06-16
Implement Node 18 Support


fainashalts:
We need to deprecate support for Node 12 soon, and we should be sure developers can upgrade to the newest Node, which is v18.

haltman-at:
See also #5305.

fainashalts:
Moving DB to its own repo, like preserves, seems reasonable to me. We can always move it back when it is being actively worked on. Or develop it as a plug-in instead perhaps.

cliffoo:
Jest cannot be upgraded because they dropped support for node 12, and we might run into similar trouble with other dependencies since node 12 reach EOL > 4 months ago. Just adding a note here since #5507 closed.

davidmurdoch:
Fixed by https://github.com/trufflesuite/truffle/pull/5543

PR #5507, opened 2022-09-06
Upgrade jest to v29

cliffoo:


cliffoo:
Oops nvm [jest 29 doesn't support node 12 anymore](https://jestjs.io/blog/2022/08/25/jest-29). Closing.

cds-amal:
@cliffoo I think we should tag this as blocked, so it's not forgotten when Truffle eventually drops node 12 support.

cliffoo:
True true

haltman-at:
Hm rather than leaving this open on the PR board where it takes up space while blocked, it would be better to just have this as an issue for now?  (Since *what* to do is easy, it just needs to be recorded that we have to do it.)

cliffoo:
That makes sense. Actually I'll just add a note to the [node 18 issue](https://github.com/trufflesuite/truffle/issues/5203) rather than opening a new one. Thanks @haltman-at

cds-amal (commented):
Jest is up to [29.1.2, and the changelog looks yummy](https://github.com/facebook/jest/blob/main/CHANGELOG.md). Is there a reason not to bump up?

cliffoo:
Nope! This PR just aged. Pushed changes to use latest.

eggplantzzz (approved):
great, I like!

PR #5508, opened 2022-09-06
Use canonical URL for goerli-optimistic

haltman-at:
It turns out that Etherscan's explorer for Optimistic Goerli canonically lives at https://goerli-optimism.etherscan.io/ rather than https://goerli-optimistic.etherscan.io/ ; both work but the former seems to be the canonical one.  Yes, this is inconsistent, but that's the way it is!  So I changed it to that.

Also Sourcify added support for kekchain but since that's another one with chain ID different from network ID I have left it out and added a comment about it.

### DIFF
--- a/packages/source-fetcher/lib/etherscan.ts
+++ b/packages/source-fetcher/lib/etherscan.ts
@@ -63,7 +63,7 @@ const EtherscanFetcher: FetcherConstructor = class EtherscanFetcher
       "sepolia": "api-sepolia.etherscan.io",
       "optimistic": "api-optimistic.etherscan.io",
       "kovan-optimistic": "api-kovan-optimistic.etherscan.io",
-      "goerli-optimistic": "api-goerli-optimistic.etherscan.io",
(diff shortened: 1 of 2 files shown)


cds-amal:
How did you determine `https://goerli-optimism.etherscan.io/` is the canonical one?

cds-amal (approved):
LGTM!

haltman-at:
> How did you determine `https://goerli-optimism.etherscan.io/` is the canonical one?

Go visit it, you'll see that one redirects to the other.

cds-amal:
> > How did you determine `https://goerli-optimism.etherscan.io/` is the canonical one?
> 
> Go visit it, you'll see that one redirects to the other.

Ah, I thought it was from some documentation. Where the browser ends up isn't necessarily canonical, as the service could employ a public API which resolves/redirects elsewhere.

haltman-at:
Oh, hm.  Well, it's the best we have in this case...? <shrug>

Issue #5459, opened 2022-08-18
Not able to debug transaction in hardhat project


tomasz90:
## Issue

Not able to debug transaction in hardhat project.

## Steps to Reproduce

1. Run node at localhost using hardhat
2. Make any transaction
3. Grab the hash
4. Try to debug this transaction using accuired hash

## Expected Behavior

I am able to debug transaction

## Actual Results

Not able to debug transaction, error:

```
SyntaxError: Unexpected token N in JSON at position 0
    at JSON.parse (<anonymous>)
    at ChildProcess.<anonymous> (/usr/local/lib/node_modules/truffle/build/webpack:/packages/from-hardhat/dist/src/ask-hardhat.js:91:1)
    at Object.onceWrapper (events.js:520:26)
    at ChildProcess.emit (events.js:400:28)
    at ChildProcess.emit (domain.js:475:12)
    at maybeClose (internal/child_process.js:1088:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:296:5)
Truffle v5.5.27 (core: 5.5.27)
```

## Environment

* Operating System: MacOS Monterey
* Hardhat version: 2.10.1
* Truffle version: 5.5.27
* node version: 14.20.0
* npm version: 6.14.17

gnidan:
Oh no! Thanks for reporting this. I haven't seen this before. I don't suppose your project is open source so I can try to reproduce locally? Or if there's any additional information you can provide?

gnidan:
(In the meantime, I'll add some extra error logging and release a dist tag later today for you to try.)

tomasz90:
> Oh no! Thanks for reporting this. I haven't seen this before. I don't suppose your project is open source so I can try to reproduce locally? Or if there's any additional information you can provide?

yep, unfortunately the project is not open source :(

I can add that when I call "truffle debug" without hash I am getting similar message:

`SyntaxError: Unexpected token N in JSON at position 0
    at JSON.parse (<anonymous>)
    at ChildProcess.<anonymous> (/usr/local/lib/node_modules/truffle/build/webpack:/packages/from-hardhat/dist/src/ask-hardhat.js:91:1)
    at Object.onceWrapper (events.js:520:26)
    at ChildProcess.emit (events.js:400:28)
    at ChildProcess.emit (domain.js:475:12)
    at maybeClose (internal/child_process.js:1088:16)
    at Socket.<anonymous> (internal/child_process.js:446:11)
[...]

PR #5510, opened 2022-09-07
Wrap meaningful `hardhat console` output in sigils

gnidan:
This addresses the concern surfaced in #5459, where we discovered that many Hardhat project configurations are setup to produce console output while executing ("reading") their hardhat config.

This breaks the existing experimental implementation of Hardhat support in `truffle debug` because the @truffle/from-hardhat support relies on the stdout produced by `npx hardhat console`. Since Hardhat config files might produce stdout of their own, and the Hardhat console necessarily executes those config files, we need some way of distinguishing Truffle-specific output from any config noise.

The solution this PR takes is to use sigils: Truffle instructs `npx hardhat console` to wrap the answers to Truffle's questions in a start/end delimiter. On the receiving side, Truffle then only interprets the response between those delimiters. To minimize risk of string collision with user code, this PR uses the strings `truffle-start` and `truffle-end`, wrapped with control characters from the Latin-1 supplemental block.

### DIFF
--- a/packages/from-hardhat/src/ask-hardhat.ts
+++ b/packages/from-hardhat/src/ask-hardhat.ts
@@ -67,6 +67,12 @@ export const askHardhatConsole = async (
   new Promise((accept, reject) => {
     const { workingDirectory } = withDefaultEnvironmentOptions(options);
 
+    // Latin-1 Supplemental block control codes
+    const sos = String.fromCodePoint(0x0098); // start of string
+    const st = String.fromCodePoint(0x009c); // string terminator
+    const prefix = `${sos}truffle-start${st}`;
+    const suffix = `${sos}truffle-end${st}`;
+
     const hardhat = spawn(`npx`, ["hardhat", "console"], {
       stdio: ["pipe", "pipe", "inherit"],
       cwd: workingDirectory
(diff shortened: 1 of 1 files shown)


benjamincburns (approved):
I think this is fine as long as our write to stdout flushes fully and some other part of the process doesn't intervene.

However this is a great example of why CLI tools like this should prefer stderr for informational output, and why REPLs should always have a `--command` or `--expression` flag.

packages/from-hardhat/src/ask-hardhat.ts
```diff
@@ -99,8 +112,12 @@ export const askHardhatConsole = async (
...
         .then(${
           raw
-            ? `console.log`
-            : `(resolved) => console.log(JSON.stringify(resolved))`
+            ? `(resolved) => console.log(
+                \`${prefix}$\{resolved}${suffix}\`
+              )`
+            : `(resolved) => console.log(
+                \`${prefix}$\{JSON.stringify(resolved)}${suffix}\`
+              )`
```
cds-amal, line 120: Please document this `write` for our future selves and new maintainers.

cds-amal (changes requested):
Nice solution @gnidan wrt to sentinels. I am on the fence using control characters instead of declaring app-specific ones. The only argument for app-specific ones are it would be easier to debug because they be visible. Your call, however.

Also, what canary tests can be added to this feature to catch breaking changes in the HH ecosystem?

gnidan:
> I think this is fine as long as our write to stdout flushes fully and some other part of the process doesn't intervene.

Yeah, I'm hoping this won't be a problem. But hey, it's still experimental! Guess we'll see if users complain!

gnidan:
> Also, what canary tests can be added to this feature to catch breaking changes in the HH ecosystem?

Not sure... since we're still marketing this as experimental, I guess our canary tests can be our GH issues for the time being. I'd like to avoid installing+running HH inside our CI, which is slow enough already.

cds-amal (approved):
That is beautiful!

gnidan:
Thanks both of you! Will merge when green.

Issue #5495, opened 2022-08-25
`truffle migrate` `--save` flag doesn't work and is not documented


eggplantzzz:
While digging through migrations code, we found out that there is the undocumented option `save`. If you set this to `false`, Truffle will omit saving the set migrations on chain. This option should probably be documented on the website and in the Truffle help.

Additionally there is a bug with this option. If you pass it as a flag (`--save`) on the command line, it gets parsed as a string and will not work correctly. We should tell yargs that this option is a boolean so that it will get parsed properly and work.

lsqproduction:
Addresses in [Truffle v5.5.32]( https://github.com/trufflesuite/truffle/releases/tag/v5.5.32).

PR #5516, opened 2022-09-08
enable migration's  `save` option for yargs.

lsqproduction:
Addresses #5495 
Previously, `Migration.js` allows users to run a migration without saving it on chain, but the option wasn't added for yargs. 

Since this is added in yargs now, `truffle migrate help` will also show this option, but do we want to document this on truffle website? 

I do feel these migrations are a bit abrupt at the end.  Should we add a nice message too to indicate again that the migrations weren't saved on chain? Similar to the ones for on chain migrations like this 

```
Summary
=======
 Migrations not saved on chain
```

### DIFF
--- a/packages/core/lib/commands/migrate/meta.js
+++ b/packages/core/lib/commands/migrate/meta.js
@@ -18,6 +18,7 @@ module.exports = {
     },
     "--verbose-rpc": {
       describe: "Log communication between Truffle and the Ethereum client.",
+      type: "boolean",
       default: false
     },
     "dry-run": {
--- a/packages/core/lib/commands/migrate/meta.js
+++ b/packages/core/lib/commands/migrate/meta.js
@@ -48,7 +48,7 @@ module.exports = {
       describe: "Specify whether the migration will save on chain",
       type: "boolean",
       default: true,
-      hidden: true
+      hide: true
     }
   },
   help: {
--- a/packages/core/lib/commands/migrate/meta.js
+++ b/packages/core/lib/commands/migrate/meta.js
@@ -47,7 +47,8 @@ module.exports = {
     "save": {
       describe: "Specify whether the migration will save on chain",
       type: "boolean",
-      default: true
+      default: true,
+      hidden: true
     }
   },
   help: {
--- a/packages/core/lib/commands/migrate/meta.js
+++ b/packages/core/lib/commands/migrate/meta.js
@@ -48,7 +48,7 @@ module.exports = {
       describe: "Specify whether the migration will save on chain",
       type: "boolean",
       default: true,
-      hide: true
+      hidden: true
     }
   },
   help: {
@@ -111,8 +111,8 @@ module.exports = {
       },
       {
         option: "--save",
-        describe: "Specify whether the migration will save on chain",
-        hide: true
+        description: "Specify whether the migration will save on chain",
+        hidden: true
       }
     ],
     allowedGlobalOptions: ["network", "config", "quiet"]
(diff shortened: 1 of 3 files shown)


haltman-at:
> Since this is added in yargs now, `truffle migrate help` will also show this option

Hm, it looks like it wasn't added to the `help` section though... doesn't it also needd to be added there for it to show up?

> but do we want to document this on truffle website?

I'd say so, yes!

eggplantzzz:
I think @gnidan and I talked about this and decided to fix it but leave it undocumented. Do you still feel the same @gnidan?

lsqproduction:
@eggplantzzz can you help me understand how this is different than dry run?

eggplantzzz:
Oh sorry I misunderstood, this PR is for fixing the bug and not documenting it in the help system. This should be good!

packages/core/lib/commands/migrate/meta.js
```diff
@@ -43,6 +43,11 @@ module.exports = {
       describe: "Adds extra verbosity to the status of an ongoing migration",
       type: "boolean",
       default: false
+    },
+    "save": {
+      describe: "Specify whether the migration will save on chain",
+      type: "boolean",
+      default: true
```
eggplantzzz, line 50: I think this should be good to default to `true` since that will maintain current expected behavior.

eggplantzzz (approved):

gnidan:
> I think @gnidan and I talked about this and decided to fix it but leave it undocumented. Do you still feel the same @gnidan?

Yes I still feel this way

lsqproduction:
Since we don't want to document this, I've added `hide` property for this option so it won't show up in `truffle migrate help` .

lsqproduction:
> > Since this is added in yargs now, `truffle migrate help` will also show this option
> 
> Hm, it looks like it wasn't added to the `help` section though... doesn't it also needd to be added there for it to show up?
> 
> > but do we want to document this on truffle website?
> 
> I'd say so, yes!

interesting discovery, the `help` section is used by `displayCommandHelp()` and the builder is used by yargs default command help. We have two different ways of displaying command help. `truffle cmd help` and `truffle help cmd`

lsqproduction:
> Can you elaborate a little bit more on what the `hide` and `hidden` properties do here? I'm not familiar with what they do.

good catch. My bad, the `hidden` is supposed to be `hide` this hides the key from the generated usage information

eggplantzzz:
Well do note that yargs does not ingest stuff in the `help` property so that won't actually have any effect on them. However it looks like we have a mechanism for hiding options [here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/commands/help/displayCommandHelp.js#L38-L40). Looks like if an option gets an `internal` property that is truthy it ignores it while printing. A better name for this would be `hide` like you have here.

I would suggest you change this property name to `hide` and update all options that have the `internal` field set to `true` to use `hide` instead. Do you feel like doing that?

lsqproduction:
Nice catch @eggplantzzz on the `internal`. I can't seem to find much documentation for this one. 
I do see only `compile` has this `internal` key as truthy. I can change this out too. Am I missing any other instances with `internal`?

lsqproduction:
`hidden` is actually the correct property to use.

packages/core/lib/commands/migrate/meta.js
```diff
@@ -51,7 +57,7 @@ module.exports = {
       "                                " + // spacing to align with previous line
       "[--compile-all] [--compile-none] [--verbose-rpc] [--interactive]\n" +
       "                                " + // spacing to align with previous line
-      "[--skip-dry-run] [--describe-json] [--dry-run]",
+      "[--skip-dry-run] [--describe-json] [--dry-run] [--save <boolean>]",
```
eggplantzzz, line 60: So since we want to leave this undocumented you should probably leave it out here.

eggplantzzz:
Cool, this is looking good.

PR #5521, opened 2022-09-09
Upgrade solc version to latest 0.8.17 in truffle init

sukanyaparashar:
This PR upgrades the `solc` version to latest **v0.8.17** in `truffle-config.js` when `truffle init` command is run.


haltman-at (approved):
Thanks for getting this!

PR #5535, opened 2022-09-13
Compiler plugins returns compiler object

dongmingh:
This PR is for compiler plugin to return the plugin compiler object.

Currently, the truffle compiler plugin returns the caller with plugin module and definition and the caller needs to determine the compiler object in order to execute compilation.  From design view point, it might make more sense for truffle compiler plugin to return the compiler object instead.  In this way, the caller can invoke the object to execute compilation without knowing the plugin module and definition.

### DIFF
--- a/packages/plugins/lib/Plugins.ts
+++ b/packages/plugins/lib/Plugins.ts
@@ -102,7 +102,7 @@ export class Plugins {
     const definitions = Plugins.loadPluginDefinitions(plugins);
 
     return Object.entries(definitions).map(
-      ([module, definition]) => new Plugin({ module, definition })
+      ([module, definition]) => require(path.join(module, definition.compile)).Compile
     );
   }
 }


packages/plugins/lib/Plugins.ts
```diff
@@ -102,7 +102,7 @@ export class Plugins {
     const definitions = Plugins.loadPluginDefinitions(plugins);
 
     return Object.entries(definitions).map(
-      ([module, definition]) => new Plugin({ module, definition })
+      ([module, definition]) => require(path.join(module, definition.compile)).Compile
```
cds-amal, line 105: This new logic bypasses `originalRequire` which is needed to resolve modules not bundled with Truffle.

cds-amal (commented):
I'm not sure of the compiler plugin intention, but should mention that this plugin system uses `original-require` to handle the case where web-packed/bundled-truffle needs to load non bundled modules.

haltman-at:
Is this a breaking change...?  This sounds like a breaking change...

packages/plugins/lib/Plugins.ts
dongmingh: Thanks for pointing this out.  I can see another issue with the new logic is losing the APIs defined in the class Plugin.  I will rework this.

dongmingh:
@haltman-at Thanks for reviewing.  I will add logic in workflow-compile to handle the required object to avoid braking change.

packages/plugins/lib/Plugins.ts
fainashalts: Thanks @cds-amal and @dongmingh -- let me know when this is read for re-review!

haltman-at:
> @haltman-at Thanks for reviewing. I will add logic in workflow-compile to handle the required object to avoid braking change.

It would still be a breaking change to the particular package, however, if I'm understanding correctly.  If so, that needs to be marked.

dongmingh:
@haltman-at I meant to undo the changes in plugin and add logic to workflow-compile to handle zksolc.

fainashalts:
> Is this a breaking change...? This sounds like a breaking change...


@haltman-at thanks for the review! Are you saying this specific change is breaking, or the overall branch being merged into (`compiler-plugin-system`) is breaking? The compiler plugin logic would only get triggered if the user lists a compiler plugin in their config, so I'm actually not sure if that makes this breaking. We're adding one more type of compiler but it shouldn't affect how the rest of the `plugins` module works. 

I may be misunderstanding the discussion here so feel free to set me right!

haltman-at:
Oh, oops!  I missed that this PR isn't into `develop`.  Silly me!  I guess ignore my comment, because I haven't actually looked at the branch this is being merged into.

eggplantzzz:
Closing this as @dongmingh will submit a new one.

PR #5537, opened 2022-09-14
bump lerna to ^5.6.1

eggplantzzz:
lerna is now being maintained again, yay! Let's see what happens when I bump this...

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "eslint": "^8.4.1",
     "eslint-plugin-react-hooks": "^4.2.0",
     "husky": "^7.0.4",
-    "lerna": "^4.0.0",
+    "lerna": "^5.5.1",
     "lerna-dependency-graph": "^1.1.0",
     "lerna-update-wizard": "1.1.2",
     "lint-staged": "^12.1.2",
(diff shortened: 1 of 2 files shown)


cds-amal:
[Lerna 5.0 breaks support for node v12](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#500-2022-05-24)

cds-amal (dismissed):
We'll have to hold off until we drop support for Node 12 (and support Node 18) because [Lerna 5.0 breaks support for Node 12](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#500-2022-05-24)

cliffoo:
Note that lerna has had a few patch releases since this PR opened https://www.npmjs.com/package/lerna

cliffoo (approved):
Yes please! 

Nx:
[Integrate with lerna](https://nx.dev/recipe/lerna-and-nx)
[Migrate](https://blog.nrwl.io/migrating-from-lerna-to-nx-better-dev-ergonomics-much-faster-build-times-da76ff14ccbb)
Just imagine the speed!

Issue #5244, opened 2022-06-29
Circular dependency between `@truffle/resolver` and `@truffle/compile-solidity`


haltman-at:
## Issue

We currently have a circular dependency between `@truffle/resolver` and `@truffle/compile-solidity`.

The package `@truffle/compile-solidity` imports `@truffle/resolver` as a dev dependency, because it uses it in tests.  Meanwhile, the package `@truffle/resolver` imports the range utils from `@truffle/compile-solidity`.  (Yes, one of the legs of this circularity is a dev dependency, but even that will cause problems for lerna.)

Since the range utils are the only part of `@truffle/compile-solidity` that's imported, the simplest solution might be to simply split the range utils off into their own package?  Since that sounds easier to me than eliminating `@truffle/resolver` from the compile-solidity tests...

## Environment

* Truffle version (`truffle version`): 5.5.19

haltman-at:
Huh, I thought #5445 was supposed to resolve this, but it's still around...

PR #5540, opened 2022-09-15
Eliminate circular dependency between compile-solidity and resolver

haltman-at:
Addresses #5244.  I believe #5445 was supposed to do this, but it looks like Tyler forgot to actually remove the dependency.  So, why not, I did it here.  And that should be that!


cds-amal (approved):
Thanks, @haltman-at.

eggplantzzz (approved):
thanks for cleaning up after me

PR #5541, opened 2022-09-15
Update web3 to 1.8.0

haltman-at:
Web3 1.8.0 is out, so this PR updates all of our packages to that (for web3, web3-utils, and etc).


haltman-at:
Hm, looks like they accidentally broke some things.  Closing this for now.

PR #5542, opened 2022-09-15
Internal improvement: Add ResolverIntercept logic to Resolver

eggplantzzz:
@truffle/migrate has a class called `ResolverIntercept` that is used as a wrapper class for `Resolver`. This class keeps track of the contracts required by a user and caches them. It seems to make more sense that this logic should just exist as a part of `Resolver` instead of existing in @truffle/migrate. voila -


eggplantzzz:
Decided to close this as it might be more complexities that we need to better research involving the cacheing of contracts.

PR #5545, opened 2022-09-16
Update 0x44 to PREVRANDAO|DIFFICULTY

haltman-at:
So, the merge is complete, and 0x44 no longer means `DIFFICULTY`, rather it means `PREVRANDAO`.  Except, our disassembler is still likely to be used on pre-merge code, so just updating it to `PREVRANDAO` seems wrong.  Instead I've updated it to `PREVRANDAO|DIFFICULTY` to cover both.  I figure in the future, in cases where we know more about the code being disassembled, we can narrow it down using extra information (e.g., we can always return `PREVRANDAO` for code that's written in EOF, once that exists, and potentially the debugger could narrow things down further); but for now, we can just produce this ambiguous return to get up to date? :shrug:

### DIFF
--- a/packages/code-utils/src/opcodes.ts
+++ b/packages/code-utils/src/opcodes.ts
@@ -56,7 +56,7 @@ const codes: OpcodeTable = {
   0x41: "COINBASE",
   0x42: "TIMESTAMP",
   0x43: "NUMBER",
-  0x44: "DIFFICULTY",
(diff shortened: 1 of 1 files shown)


gnidan (approved):
Sure, a reasonable approach!

gnidan:
Sorry the all core devs didn't care

PR #5557, opened 2022-09-21
Type opcodes table as const in @truffle/code-utils

gnidan:
I didn't like how this was needlessly clobbering type information, so this PR modifies @truffle/code-utils to preserve `const`-ness in its definition of the opcodes table.

### DIFF
--- a/packages/code-utils/src/opcodes.ts
+++ b/packages/code-utils/src/opcodes.ts
@@ -167,4 +167,4 @@ const codes = {
 export type OpcodeTable = typeof codes;
 
 export const parseOpcode = (op: number) =>
-  codes[op in codes ? (op as keyof OpcodeTable) : 0xfe];
+  op in codes ? codes[op as keyof OpcodeTable] : "INVALID";
--- a/packages/code-utils/src/opcodes.ts
+++ b/packages/code-utils/src/opcodes.ts
@@ -1,6 +1,4 @@
-import type { OpcodeTable } from "./types";
-
-const codes: OpcodeTable = {
+const codes = {
   0x00: "STOP",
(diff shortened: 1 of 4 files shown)


packages/code-utils/src/opcodes.ts
```diff
@@ -162,8 +160,11 @@ const codes: OpcodeTable = {
...
+  0xfe: "INVALID",
   0xff: "SELFDESTRUCT"
-};
+} as const;
+
+export type OpcodeTable = typeof codes;
 
-export = (op: number) => (codes[op] ? codes[op] : "INVALID");
+export const parseOpcode = (op: number) =>
+  codes[op in codes ? (op as keyof OpcodeTable) : 0xfe];
```
haltman-at, line 170: Using `0xfe` as a default before the lookup, rather than `"INVALID"` after, is pretty confusing.  I think it would be better to do this the old way (not necessarily the same code, but at least doing the defaulting after), or, failing that, there should at least be a comment to explain this.

haltman-at (commented):
Not really clear to me that this is useful, but, uh, OK, I guess?

packages/code-utils/src/opcodes.ts
gnidan: Okee! Yeah I was on the fence. I'll switch it back.

gnidan:
> Not really clear to me that this is useful, but, uh, OK, I guess?

As usual, this work is driven by some offhand hacking :)

haltman-at (approved):
Sure, I guess we can do this?

gnidan:
Thanks!

PR #5560, opened 2022-09-22
Dev improvement: update prepare-release script to create a separate branch for releasing

eggplantzzz:
This PR update the prepare-release script. After pulling the latest from the `develop` branch, the script will now switch to a new branch to use for the release.

### DIFF
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -10,11 +10,11 @@ git checkout master
 git pull origin master
 git checkout develop
 git pull origin develop
-git checkout -b release-branch-$(date +%s)
+git checkout -b release-branch-"$(date +%s)"
 
 ## Build
 yarn bootstrap
 
 ## Get output of changes for release notes
-prs-merged-since --repo trufflesuite/truffle --tag v$LAST_PUBLISHED_TAG --format markdown
+prs-merged-since --repo trufflesuite/truffle --tag v"$LAST_PUBLISHED_TAG" --format markdown
 lerna changed
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -10,7 +10,7 @@ git checkout master
 git pull origin master
 git checkout develop
 git pull origin develop
-git checkout -b release-branch-"$(date +%s)"
+git checkout -b release/"$(date +%m-%d-%y.%S)"
 
 ## Build
 yarn bootstrap
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -10,6 +10,7 @@ git checkout master
 git pull origin master
 git checkout develop
 git pull origin develop
+git checkout -b release-branch-$(date +%s)
 
 ## Build
 yarn bootstrap
--- a/scripts/publish-release.sh
+++ b/scripts/publish-release.sh
@@ -4,8 +4,7 @@
 set -ex
 
 ## Obtain/check npm access
-npm whoami
-if [ "$?" != "0" ];
+if ! npm whoami;
   then
     npm login
 fi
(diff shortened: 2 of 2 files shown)


cds-amal (commented):
Since you're in here, it might be useful to lint this through https://www.shellcheck.net/.  I try to do this with the bash scripts I change.

Will this change impact today's release? (if merged?)

haltman-at:
It's not enough to just change this, right?  Wouldn't `publish-release.sh` also have to be changed to merge this new branch back in to `master` and `develop`?

eggplantzzz:
Oh I forgot about the version bump changes, doh! Will update!

eggplantzzz:
Ok so I updated the publish-release script and ran the code through the linter as you suggested @cds-amal!

scripts/publish-release.sh
```diff
@@ -4,20 +4,20 @@
...
 fi
 node ./scripts/npm-access.js
 
 ## Publish packages to npm
 lerna version
-lerna publish from-package 
+lerna publish from-package
 
 ## Update git branches
+RELEASE_BRANCH_NAME=$(git branch --show-current)
```
haltman-at, line 18: Hm, I was going to say "this bothers me because it depends on people being on the correct branch", but now that I look I guess that was actually already true, so, uh, I guess that's probably not a problem?

If you do want to do something about this, you could alter the release branch to just be based on the current day rather than time, that way one could switch to it reliably (well, unless you're doing the release around midnight).  I dunno, probably this isn't worth worrying about, and the existing code already has this caveat, but I thought I'd point it out at least.

scripts/publish-release.sh
```diff
@@ -4,20 +4,20 @@
...
-npm whoami
-if [ "$?" != "0" ];
+if ! npm whoami;
   then
     npm login
 fi
 node ./scripts/npm-access.js
 
 ## Publish packages to npm
 lerna version
```
cds-amal, line 14: The release branch is the active branch at this point, and I'm unsure about the behavior of `lerna version` at this point.
- previously the command ran on active branch `develop`, which had an `origin/develop` remote. Will it work without the `origin/release/EPOCH` remote?
- assuming `version` works, the changes committed to the remote branch gets merged into `origin/master`, however `origin/develop` still needs reconciliation. Note, local `develop` and `origin/develop` could have diverged if someone merged a PR during the release process.

scripts/prepare-release.sh
```diff
@@ -10,10 +10,11 @@ git checkout master
 git pull origin master
 git checkout develop
 git pull origin develop
+git checkout -b release-branch-"$(date +%s)"
```
cds-amal, line 13: Should this release branch be pushed to `origin` to support `lerna version`?

cds-amal (changes requested):
I have concerns about how to reconcile `origin/develop` with the release branch changes at the end of the process. At this point, `master` has the release changes, and `origin/develop` could be unchanged, or have accepted PRs.

scripts/publish-release.sh
eggplantzzz: [lerna readme](https://github.com/lerna/lerna/tree/main/commands/version#readme)

It says that it bumps the packages "that have been updated since the previous tagged release". It doesn't seem like it cares about what branch you are on or take that into account at all.

And yep, `develop` could have changed. So it is possible someone will have to resolve a conflict. Is there any alternative to this?

scripts/prepare-release.sh
gnidan, line 13: I think we should use `release/<date>` as the branch name. Use those namespaces!

eggplantzzz: changed it to release/"$(date +%m-%d-%y.%S)"

scripts/publish-release.sh
eggplantzzz: Yeah, but I also wanted to ensure that the branch would have a unique name. I vote we don't worry about this at this time. Let me know if you decide to disagree :)

PR #5561, opened 2022-09-23
Add raw storage writes to txlog

haltman-at:
This PR adds a new type of action to txlog, `write`.  Like events, it does not have subactions.  It represents a write to a storage variable -- or at least, the intent is that it eventually will; right now it just represents a write to a storage slot.  Decoding storage writes is hard, so this PR just contains the raw storage info, with nothing decoded.

Each `write` action presently has two fields, `raw` and `steps`.  The `raw` field contains an object mapping storage slots to the values written to them; presently, each `write` action represents only a single storage write, so each `raw` field will only have a single entry.  However, the intent is that in the future, multiple raw storage writes may be consolidated into a single `write` action when they represent a single logical write to a storage variable (e.g. if you store a new value in a string and it takes up multiple storage slots).

The `steps` field similarly contains an object keyed with storage slots, but this time the value is the step when that slot was written to.

And that's it; the hard part is all going to be in decoding but this PR doesn't have that.

There's also a test.

### DIFF
--- a/packages/debugger/lib/txlog/actions/index.js
+++ b/packages/debugger/lib/txlog/actions/index.js
@@ -211,6 +211,19 @@ export function logEvent(pointer, newPointer, step, decoding, rawEventInfo) {
   };
 }
 
+//this may be replaced once decoding info was added
+export const STORE = "TXLOG_STORE";
+export function store(pointer, newPointer, step, rawSlot, rawValue) {
+  return {
+    type: STORE,
+    pointer,
+    newPointer, //does not actually affect current pointer!
+    step,
+    rawSlot,
+    rawValue
+  };
+}
+
 export const RECORD_ORIGIN = "TXLOG_RECORD_ORIGIN";
 export function recordOrigin(pointer, address) {
   return {
--- a/packages/debugger/lib/txlog/actions/index.js
+++ b/packages/debugger/lib/txlog/actions/index.js
@@ -211,7 +211,7 @@ export function logEvent(pointer, newPointer, step, decoding, rawEventInfo) {
   };
 }
 
-//this may be replaced once decoding info was added
+//this may be replaced once decoding info is added
 export const STORE = "TXLOG_STORE";
 export function store(pointer, newPointer, step, rawSlot, rawValue) {
   return {
--- a/packages/debugger/lib/txlog/reducers.js
+++ b/packages/debugger/lib/txlog/reducers.js
@@ -58,6 +58,29 @@ function transactionLog(state = DEFAULT_TX_LOG, action) {
         }
       };
 
+    case actions.STORE:
+      //this case will likely need to be redone when decoding is added
+      //(likely split into multiple cases)
+      //for now, there is no combining of different writes, but when decoding
+      //is added there will need to be!
+      return {
+        byPointer: {
+          ...state.byPointer,
+          [pointer]: {
+            ...node,
+            actions: [...node.actions, newPointer]
+          },
+          [newPointer]: {
+            type: "write",
+            raw: {
+              [action.rawSlot]: action.rawValue
+            },
+            steps: {
+              [action.rawSlot]: step
+            }
+          }
+        }
+      };
     case actions.INTERNAL_CALL:
       return {
         byPointer: {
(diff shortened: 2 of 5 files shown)


packages/debugger/lib/txlog/actions/index.js
```diff
@@ -211,6 +211,19 @@ export function logEvent(pointer, newPointer, step, decoding, rawEventInfo) {
   };
 }
 
+//this may be replaced once decoding info was added
+export const STORE = "TXLOG_STORE";
+export function store(pointer, newPointer, step, rawSlot, rawValue) {
+  return {
+    type: STORE,
+    pointer,
+    newPointer, //does not actually affect current pointer!
```
gnidan, line 220: What do these names mean? Maybe we should use `storagePointer` and `underlyingPointer`?

gnidan: or like source/dest dichotomy

cds-amal: Yea, would like to know what these names mean. I was thinking pointers as in linkedlist, or other graph DS.

cds-amal (commented):
First pass. I didn't have the stamina to finish.

packages/debugger/lib/txlog/actions/index.js
haltman-at: The `pointer` and `newPointer` terminology here is the same as for every other txlog action.  The `pointer` is the JSON pointer to the current location in the txog, and the `newPointer` is the JSON pointer for the new node that is getting created.  They're pointers into the txlog tree; they've got nothing to do with storage pointers.

Likely these terms could be clearer but they're the same as for the other actions; if you think they should be changed, then I would prefer to change them uniformly for all the actions and not just for this one.

packages/debugger/lib/txlog/actions/index.js
```diff
@@ -211,6 +211,19 @@ export function logEvent(pointer, newPointer, step, decoding, rawEventInfo) {
   };
 }
 
+//this may be replaced once decoding info was added
```
cds-amal, line 214: This comment is confusing.

packages/debugger/lib/txlog/actions/index.js
cds-amal: Makes sense to me. I have to remember the code uses unorthodox pointer names whenever I read this code, and was thrown off when @gnidan interpreted the abstractions differently.

packages/debugger/lib/txlog/actions/index.js
```diff
@@ -211,6 +211,19 @@ export function logEvent(pointer, newPointer, step, decoding, rawEventInfo) {
...
+export const STORE = "TXLOG_STORE";
```
cds-amal, line 215: Would you consider `TXLOG_STORAGE` or `TXLOG_STORAGE_SLOT` since it's bundling a storage slot and value?

cds-amal (approved):
This looks good to me. I had a suggestion about variable naming, but don't feel it should block this PR.

packages/debugger/lib/txlog/actions/index.js
haltman-at: Oh, oops, incorrect tense on that verb (should be "is added")... is that the confusing part?  I'll go fix that...

packages/debugger/lib/txlog/actions/index.js
cds-amal: I like `RECORD_STORAGE`, but am ok if you think it's fine.

haltman-at: I guess probably it should be `RECORD_STORE`, since it's recording a particular store.

haltman-at: Yeah I think I'll just go ahead and merge this as-is. :P

cds-amal: > I don't really understand the basis of your criticism.

Sorry, @haltman-at, I should have explained this better. It stems from this being a Redux action creator for `store`, which is how Redux refers to its state tree. However, I don't have a good alternate suggestion so, I can't complain.

haltman-at: Oh wow -- yeah that's really far away from anything that was on my mind.  That's just on a totally different level of abstraction entirely.  I don't expect people to get confused by that collision because I don't expect people to be thinking about those simultaneously?

PR #5570, opened 2022-09-27
Add new Sourcify networks for late September

haltman-at:
Sourcify added a few more networks.  They also removed one, but Sourcify's removed networks remain available (unlike Etherscan's), so I don't remove them from the fetcher in case someone wants to make use of them.


sukanyaparashar (approved):
Looks good !

PR #5571, opened 2022-09-28
Bug fix: Add `disableChecks` parameter to decoder type

eggplantzzz:
[Here](https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/testing/TestRunner.js#L149-L156) there is a call to a function which includes a parameter not included in the type [here](https://github.com/trufflesuite/truffle/blob/develop/packages/decoder/lib/types.ts#L146-L174). This PR adds the missing type.

### DIFF
--- a/packages/decoder/lib/types.ts
+++ b/packages/decoder/lib/types.ts
@@ -171,6 +171,10 @@ export interface EventOptions {
    * the event -- should be returned.  Defaults to `"off"`.
    */
   extras?: ExtrasAllowed;
+  /**
+   * Allows decodings that don't pass the re-encoding test.  Don't turn
+   * this on unless you know what you're doing!
+   */
   disableChecks?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


packages/decoder/lib/types.ts
```diff
@@ -171,6 +171,7 @@ export interface EventOptions {
    * the event -- should be returned.  Defaults to `"off"`.
    */
   extras?: ExtrasAllowed;
+  disableChecks?: boolean;
```
haltman-at, line 174: The option needs a documentation comment, we use typedoc in the decoder and make a point of documenting everything like this.  Fortunately you can just copy/adapt the comment from [here](https://github.com/trufflesuite/truffle/blob/develop/packages/codec/lib/types.ts#L735).

eggplantzzz: Thanks @haltman-at!

haltman-at (approved):
Seems good, thanks!

Issue #3329, opened 2020-08-31
Problems when contract names conflict with internal node objects.


mktcode:
I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) and was asked to file an issue here.

---------------------------

## Issue

Truffle console crashes if there is a contract called Buffer.

From Gitter: 
> If you have one named "Buffer" it overwrites the native "Buffer" object. This will cause problems. I assume somewhere in your project there could be a contract named "Buffer"?

## Steps to Reproduce

```
truffle init
yarn add @chainlink/contracts
```
MyTest.sol
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.6.0;

import "@chainlink/contracts/src/v0.6/ChainlinkClient.sol";

contract MyTest is ChainlinkClient {}

```

## Expected Behavior

I expect the truffle console (`truffle develop`) to still work after compiling the contracts.

## Actual Results

Truffle console crashes with following exception:
```
truffle(develop)> version
Truffle v5.1.42 (core: 5.1.42)
Solidity - ^0.6.0 (solc-js)
Node v12.18.2
Web3.js v1.2.1
truffle(develop)> TypeError: Buffer.isBuffer is not a function
    at Function.writeFileSync [as sync] (/usr/local/lib/node_modules/truffle/build/webpack:/node_modules/write-file-atomic/index.js:213:1)
    at Configstore.set all [as all] (/usr/local/lib/node_modules/truffle/build/webpack:/node_modules/configstore/index.js:58:1)
    at new Configstore (/usr/local/lib/node_modules/truffle/build/webpack:/node_modules/configstore/index.js:25:1)
    at Function.getUserConfig (/usr/local/lib/node_modules/truffle/build/webpack:/packages/config/dist/index.js:171:1)
    at Object.send (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/services/analytics/index.js:5:1)
    at Command.run (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/command.js:137:1)
    at Console.interpret (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/console.js:155:1)
    at ReplManager.interpret (/usr/local/lib/node_modules/truffle/build/webpack:/packages/core/lib/repl.js:129:1)
    at bound (domain.js:426:14)
    at REPLServer.runBound [as eval] (domain.js:439:12)
[...]

PR #5573, opened 2022-09-28
Bug fix: don't clobber repl context variables when setting up repl context

eggplantzzz:
As mentioned in https://github.com/trufflesuite/truffle/issues/3329, if a user has a contract name that conflicts with one of Node's native objects (such as `Buffer`) then it will overwrite that object. This could cause things to break. In order to fix this, this PR provides an implementation where Truffle will not overwrite objects found in the context the first time it loads the repl and sets up the environment. The `provision` method is where these things are set up, and on subsequent calls to `provision` it will overwrite things it finds in the repl context as it will need to update contracts that have changed etc.

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -78,7 +78,7 @@ class Console extends EventEmitter {
       this.repl.displayPrompt();
 
       // hydrate the environment with the user's contracts
-      this.provision();
+      this.provision(false);
 
       this.repl.on("exit", () => {
         process.exit();
@@ -187,7 +187,9 @@ class Console extends EventEmitter {
     };
   }
 
-  provision() {
+  // overwriteReplContextVars is provided to allow us to avoid clobbering native
+  // Node objects (e.g. Buffer, number) on the first provision
+  provision(overwriteReplContextVars = true) {
     let files;
     try {
       const unfilteredFiles = fse.readdirSync(
@@ -222,11 +224,11 @@ class Console extends EventEmitter {
       return abstraction;
     });
 
-    this.resetContractsInConsoleContext(abstractions);
+    this.resetContractsInConsoleContext(abstractions, overwriteReplContextVars);
     return abstractions;
   }
 
-  resetContractsInConsoleContext(abstractions) {
+  resetContractsInConsoleContext(abstractions, overwriteReplContextVars) {
     abstractions = abstractions || [];
 
     const contextVars = {};
@@ -237,7 +239,13 @@ class Console extends EventEmitter {
 
     // make sure the repl gets the new contracts in its context
     Object.keys(contextVars || {}).forEach(key => {
-      this.repl.context[key] = contextVars[key];
+      if (overwriteReplContextVars) {
+        this.repl.context[key] = contextVars[key];
+      } else {
+        if (this.repl.context[key] === undefined) {
+          this.repl.context[key] = contextVars[key];
+        }
+      }
     });
   }
 
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -47,7 +47,7 @@ class Console extends EventEmitter {
     this.options = options;
 
     this.repl = null;
-    // we need to keep tract of name conflicts that occur between contracts and
+    // we need to keep track of name conflicts that occur between contracts and
     // repl context objects so as not to overwrite them - this is to prevent
     // overwriting Node native objects like Buffer, number, etc.
     this.replContextNameConflicts = [];
@@ -62,7 +62,7 @@ class Console extends EventEmitter {
     });
   }
 
-  detectNameConflicts(abstractions) {
+  recordNameConflicts(abstractions) {
     for (const abstraction of abstractions) {
       if (
         Object.getOwnPropertyNames(this.repl.context.global).includes(
@@ -243,7 +243,7 @@ class Console extends EventEmitter {
     });
 
     if (initialProvision) {
-      this.detectNameConflicts(abstractions);
+      this.recordNameConflicts(abstractions);
     }
 
     this.resetContractsInConsoleContext(abstractions);
@@ -256,7 +256,7 @@ class Console extends EventEmitter {
     const contextVars = {};
 
     abstractions.forEach(abstraction => {
-      // don't overwrite Node's native objects - we detect name conflicts
+      // don't overwrite Node's native objects - we record name conflicts
       // on the first call to `provision`
       if (!this.replContextNameConflicts.includes(abstraction.contract_name)) {
         contextVars[abstraction.contract_name] = abstraction;
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -50,7 +50,7 @@ class Console extends EventEmitter {
     // we need to keep track of name conflicts that occur between contracts and
     // repl context objects so as not to overwrite them - this is to prevent
     // overwriting Node native objects like Buffer, number, etc.
-    this.replContextNameConflicts = [];
+    this.replContextNameConflicts = {};
 
     this.interfaceAdapter = createInterfaceAdapter({
       provider: options.provider,
@@ -62,14 +62,27 @@ class Console extends EventEmitter {
     });
   }
 
+  // we only want to record a conflict the first time we try and load a contract
+  // into the repl context because we need to overwrite them when they change
   recordNameConflicts(abstractions) {
     for (const abstraction of abstractions) {
       if (
+        this.replContextNameConflicts[abstraction.contract_name] ===
+          undefined &&
         Object.getOwnPropertyNames(this.repl.context.global).includes(
           abstraction.contract_name
         )
       ) {
-        this.replContextNameConflicts.push(abstraction.contract_name);
+        this.replContextNameConflicts[abstraction.contract_name] = {
+          conflict: true,
+          warned: false
+        };
+      } else if (
+        this.replContextNameConflicts[abstraction.contract_name] === undefined
+      ) {
+        this.replContextNameConflicts[abstraction.contract_name] = {
+          conflict: false
+        };
       }
     }
   }
@@ -93,7 +106,7 @@ class Console extends EventEmitter {
       this.repl.setPrompt("truffle(" + this.options.network + ")> ");
 
       // hydrate the environment with the user's contracts
-      this.provision(true);
+      this.provision();
 
       // provision first before displaying prompt so that if
       // there is a warning the user will end up at the prompt
@@ -207,7 +220,7 @@ class Console extends EventEmitter {
     };
   }
 
-  provision(recordNameConflicts = false) {
+  provision() {
     let files;
     try {
       const unfilteredFiles = fse.readdirSync(
@@ -242,9 +255,7 @@ class Console extends EventEmitter {
       return abstraction;
     });
 
-    if (recordNameConflicts) {
-      this.recordNameConflicts(abstractions);
-    }
+    this.recordNameConflicts(abstractions);
 
     this.resetContractsInConsoleContext(abstractions);
     return abstractions;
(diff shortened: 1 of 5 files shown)


cliffoo (approved):
Nicely done @eggplantzzz ! I verified that:
- If contract name doesn't conflict, hydrate to repl.
- If conflict, contract available via `artifacts.require`

Which I think is very sensible! We definitely need to document this though.

Two more things:
- We prob want tests for this? Though I think that can come later and this PR shouldn't be held back.
- Release: I consider this breaking because there may be users relying on the current behavior to overwrite node's built-in objects.

packages/core/lib/console.js
```diff
@@ -237,7 +240,13 @@ class Console extends EventEmitter {
 
     // make sure the repl gets the new contracts in its context
     Object.keys(contextVars || {}).forEach(key => {
-      this.repl.context[key] = contextVars[key];
+      if (overwriteReplContextVars) {
+        this.repl.context[key] = contextVars[key];
+      } else {
+        if (this.repl.context[key] === undefined) {
```
cds-amal, line 246: This can be combined as an `else if`

eggplantzzz: seems reasonable

packages/core/lib/console.js
```diff
@@ -237,7 +240,13 @@ class Console extends EventEmitter {
...
+          this.repl.context[key] = contextVars[key];
+        }
+      }
```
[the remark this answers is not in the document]
eggplantzzz: Related to the release portion of your comment, @gnidan voted that we consider this a bug that is being fixed as opposed to breaking.

Tests sound like a great idea!

eggplantzzz:
So I found an issue with the current implementation here that needs to be fixed. Currently it won't clobber names in the context the first call to `provision` but will subsequent times. This is partially necessary because we have to provide fresh copies of all artifact objects after each command as they could have changed/been updated.

What we really need to do is make a note of all name conflicts on the first provision. On subsequent calls to `provision` we should reference this list and not overwrite any variables that conflicted on the first call (as those will be the objects that we do not want to overwrite).

eggplantzzz:
Hmmm @dongmingh you shouldn't be getting warnings for things like MetaCoin. Can you tell me more exactly what you did?

dongmingh:
@eggplantzzz let me know if need me to show you how I saw this.

eggplantzzz:
Thanks @dongmingh, I found a bug with how contracts are detected and fixed it. You should no longer be getting conflicts for things like MetaCoin. However, the prompt still does not appear when the warning is present.

eggplantzzz:
Ok I also caused the prompt to display after provisioning which will cause the warning to display before the prompt. So the other bug you found @dongmingh should be fixed now.

dongmingh:
I retest it and no longer see conflict contracts like MetaCoin.  Also the truffle prompt displays as expected. Thanks.

packages/core/lib/console.js
```diff
@@ -47,6 +47,10 @@ class Console extends EventEmitter {
     this.options = options;
 
     this.repl = null;
+    // we need to keep tract of name conflicts that occur between contracts and
```
cds-amal, line 50: typo: s/tract/track/

packages/core/lib/console.js
```diff
@@ -58,6 +62,18 @@ class Console extends EventEmitter {
     });
   }
 
+  detectNameConflicts(abstractions) {
```
cds-amal, line 65: `detect` doesn't imply a side effect. How about `record`, or `capture` instead?

packages/core/lib/console.js
```diff
@@ -232,9 +256,28 @@ class Console extends EventEmitter {
...
+          ? this.replContextNameConflicts[0]
+          : this.replContextNameConflicts.join(", ");
+      console.log(
+        `\n > Warning: One or more of your contract(s) has a name conflict ` +
+          `with something in the current repl context and was not loaded by ` +
+          `default. \n > You can use 'artifacts.require("<artifactName>")' ` +
+          `to obtain a reference to your contract(s). \n > Truffle recommends ` +
+          `that you rename your contract to avoid problems. \n > The following ` +
+          `name conflicts exist: ${contractNames}.`
+      );
```
cds-amal, line 278: Can you add a test for this?

cds-amal (changes requested):
Thanks, @eggplantzzz, this looks better! I left a few nits and requesting changes for testing.

packages/core/lib/console.js
eggplantzzz: sure, I like `recordNameConflicts`

packages/core/lib/console.js
eggplantzzz: yep!

packages/core/lib/console.js
```diff
@@ -187,7 +207,7 @@ class Console extends EventEmitter {
     };
   }
 
-  provision() {
+  provision(initialProvision = false) {
```
cds-amal, line 210: Would this cover the case when develop is started and there are no artifacts in the build folder? The build folder can be populated on a `migrate` or `compile` after this first check.

cds-amal (commented):
Had a scenario question.

packages/core/lib/console.js
cds-amal: Just verified this is still a clobber path

## Reprod steps

### Create project and start develop REPL
```console
truffle init
truffle create all Buffer
truffle develop
```

## REPL commands to clobber Buffer
```console
truffle(develop)> Buffer
[Function: Buffer] {
  poolSize: 8192,
  from: [Function: from],
  of: [Function: of],
  alloc: [Function: alloc],
  allocUnsafe: [Function: allocUnsafe],
  allocUnsafeSlow: [Function: allocUnsafeSlow],
  isBuffer: [Function: isBuffer],
  compare: [Function: compare],
  isEncoding: [Function: isEncoding],
  concat: [Function: concat],
  byteLength: [Function: byteLength],
  [Symbol(kIsEncodingSymbol)]: [Function: isEncoding]
}
truffle(develop)> migrate
## deployment steps...

truffle(develop)> Buffer.abi
[ { inputs: [], stateMutability: 'nonpayable', type: 'constructor' } ]
truffle(develop)>


```

cds-amal:
> I consider this breaking because there may be users relying on the current behavior to overwrite node's built-in objects.
@clifoo, what if we consider clobbering a bug, so this can be a patch instead of breaking? What do you think?

cliffoo:
I believe that's the conclusion @eggplantzzz mentioned earlier, though I'm not fully convinced. This reminds me of the recent [integer/string conversion change](https://discuss.python.org/t/int-str-conversions-broken-in-latest-python-bugfix-releases/18889) pushed to python as a patch release ([short video for context](https://youtu.be/eTucYT2LpNU)). Yes it was a bug, they fixed it, yet it broke things anyway.

Similarly, this is bug fix _and_ it's breaking. (My `Math` and `Buffer` contracts were just fine, what happened!)

eggplantzzz:
If we consider this a breaking change we can't get it in until v6. Originally through discussion with @gnidan we decided this was a bug fix. We can revisit and discuss this again.

packages/core/lib/console.js
```diff
@@ -47,6 +47,10 @@ class Console extends EventEmitter {
     this.options = options;
 
     this.repl = null;
+    // we need to keep track of name conflicts that occur between contracts and
+    // repl context objects so as not to overwrite them - this is to prevent
+    // overwriting Node native objects like Buffer, number, etc.
+    this.replContextNameConflicts = {};
```
cds-amal, line 53: What do you think of using a Map instead of an Object? I think it would enable less awkward conditional expressions.

packages/truffle/test/scenarios/commands/console.js
```diff
@@ -39,6 +40,39 @@ describe("truffle console", () => {
...
+        "../../sources/consoleWithConflicts"
+      );
+      config = await sandbox.create(projectWithConflicts);
+      config.network = "development";
+      config.logger = logger;
+    });
+
+    it("warns the user", async function () {
+      const networkName = config.network;
+      await CommandRunner.runInREPL({
```
cds-amal, line 59: I love that we're getting mileage out of this interface!

packages/core/lib/console.js
```diff
@@ -232,9 +267,41 @@ class Console extends EventEmitter {
...
+        false
+      ) {
+        contextVars[abstraction.contract_name] = abstraction;
+      }
     });
 
+    if (
+      Object.entries(this.replContextNameConflicts).some(
+        ([_key, value]) => value.conflict === true && value.warned === false
+      )
```
cds-amal, line 283: This condition is determined in the `recordNameConflict` function and can be stored in state, like `alertUserOfNewConflicts` or a better name.

cds-amal (changes requested):
I verified this works, which is really cool. I left some minor nits, and requesting a rework of the display conflict alert logic.

packages/core/lib/console.js
```diff
@@ -58,6 +64,26 @@ class Console extends EventEmitter {
     });
   }
 
+  // we only want to record a conflict the first time we try and load a contract
+  // into the repl context because we need to overwrite them when they change
+  recordNameConflicts(abstractions) {
```
cds-amal, line 69: It seems this function detects new collisions and stores them opaquely. The object DS has to be mined to extract those new collisions, to alert the user. It might be better to maintain the collection of new collisions that can be directly used later on.

packages/core/lib/console.js
```diff
@@ -58,6 +64,26 @@ class Console extends EventEmitter {
...
+    for (const abstraction of abstractions) {
+      const globals = Object.getOwnPropertyNames(this.repl.context.global);
+      const name = abstraction.contract_name;
+      if (this.replContextNameConflicts[name] === undefined) {
+        const conflict = globals.includes(name);
+        // note that we need to warn the user later
+        if (conflict) {
+          this.warnUserOfConflicts = true;
```
cds-amal, line 77: This could be an actual collection that builds the new contracts, that can be used to alert, as well as update the known collisions.

packages/core/lib/console.js
```diff
@@ -232,9 +264,38 @@ class Console extends EventEmitter {
...
+      if (
+        this.replContextNameConflicts[abstraction.contract_name].conflict ===
+        false
+      ) {
+        contextVars[abstraction.contract_name] = abstraction;
+      }
     });
 
+    if (this.warnUserOfConflicts) {
+      let contractNames = [];
```
cds-amal, line 278: `recordNameConflicts` could build this list, and this function could just use it without having to mine the `replContextNameConflicts` object.

packages/core/lib/console.js
```diff
@@ -232,9 +264,38 @@ class Console extends EventEmitter {
...
+      for (const name in this.replContextNameConflicts) {
+        if (
+          this.replContextNameConflicts[name].conflict === true &&
+          this.replContextNameConflicts[name].warned === false
+        ) {
+          this.replContextNameConflicts[name].warned = true;
+          contractNames.push(name);
+        }
+      }
```
cds-amal, line 287: This piece of logic reconstructs the information `recordNameConflicts` already knew, as well as update the repl's known collisions.

cds-amal (commented):
I kept coming back to this because something was bugging me. Unless I'm missing something, this logic can be simplified if we track two pieces of state:

1. `knownCollisionsSoFar` 
2. `newlyDiscoveredCollisions` 

- `fn recordNameConflicts` would add update the `newlyDiscoveredCollisions` container IFF new abstraction isn't in `knownCollisionsSoFar`
- `fn resetContractsInConsoleContext` would:
  1. load the abstraction in context if the name isn't in both `newlyDiscoveredCollisions` and `knownCollisionsSoFar`
  2. alert user if `newlyDiscoveredCollisions` has items
  3. move `newlyDiscoveredCollisions` to `knownCollisionsSoFar`
  4. reset `newlyDiscoveredCollisions`

Of course, please correct me if I missed a key piece of the logic.

packages/core/lib/console.js
```diff
@@ -264,27 +261,21 @@ class Console extends EventEmitter {
...
       let contractNames = [];
-      for (const name in this.replContextNameConflicts) {
-        if (
-          this.replContextNameConflicts[name].conflict === true &&
-          this.replContextNameConflicts[name].warned === false
-        ) {
-          this.replContextNameConflicts[name].warned = true;
-          contractNames.push(name);
-        }
+      for (const name of this.newReplNameConflicts.entries()) {
```
cds-amal, line 274: I think `this.newReplNameConflicts.keys()`, and `name` would be the actual name, so it wouldn't have to be dereferenced.

cds-amal (changes requested):
This is much easier to follow! I left a stylistic comment, feel free to ignore; and a change request to use `Set.keys()`  instead of `Set.entries()` in the `for of` loop.

packages/core/lib/console.js
cds-amal: Oh, we can go directly from the `Set` to a list by:
`[...this.newReplNameConflicts.keys()].join(', ')` which is the same as looping and collecting the name field in an array.

eggplantzzz: Oh ok sure.

packages/core/lib/console.js
```diff
@@ -64,22 +64,14 @@ class Console extends EventEmitter {
...
-        }
-        this.replContextNameConflicts[name] = {
-          conflict,
-          warned: false
-        };
+      if (this.knownReplNameConflicts.has(name)) {
+        continue;
+      } else if (this.replGlobals.has(name)) {
+        this.newReplNameConflicts.add(name);
+        this.knownReplNameConflicts.add(name);
```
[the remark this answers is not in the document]
eggplantzzz: By now you should know that I often prefer using `if/else if` rather than two `if` statements :)

packages/core/lib/console.js
cds-amal: `this.newReplNameConflicts` makes more sense to be a list since it's only used to collect and display (join) strings.

eggplantzzz: I'm going to vote we leave it as a set as it has the api you originally requested (`clear`, `add`, etc.) and thus you will interact with it the same way as `replGlobals` and `knownNameConflicts`. I don't really think we gain much by having it be an array.

cds-amal (approved):
LGTM. Thanks, @eggplantzzz!

PR #5575, opened 2022-09-28
Bug fix: move web3 and web3 types into regular dependencies for hdwallet-provider

eggplantzzz:
Since Truffle exposes web3 types through the constructor arguments (https://github.com/trufflesuite/truffle/blob/develop/packages/hdwallet-provider/src/constructor/types.ts), web3 and its types should be part of the regular dependencies.


haltman-at (approved):
Ooh, good catch!

Issue #5578, opened 2022-09-29
truffle develop/console does not exclude `console` or `develop` commands


cds-amal:
When this condition is hit, you have to enter multiple `.exit`s to leave the REPL.

## Reproduction steps
- `reprod`
- `truffle init`
- `truffle develop`
- `console`
- `console`
- `console`
- You will need to enter `.exit` 4 times to leave the REPL

gnidan:
I've also observed this behavior with bash and zsh ;)

cds-amal:
> I've also observed this behavior with bash and zsh ;)

`truffle develop` behaves differently than Node's REPL.  Is this the expected behavior @gnidan?

```console
$ node
Welcome to Node.js v16.17.1.
Type ".help" for more information.
> console
Object [console] {
  log: [Function: log],
  warn: [Function: warn],
  dir: [Function: dir],
  time: [Function: time],
  timeEnd: [Function: timeEnd],
  timeLog: [Function: timeLog],
  trace: [Function: trace],
  assert: [Function: assert],
  clear: [Function: clear],
  count: [Function: count],
  countReset: [Function: countReset],
  group: [Function: group],
  groupEnd: [Function: groupEnd],
  table: [Function: table],
  debug: [Function: debug],
  info: [Function: info],
  dirxml: [Function: dirxml],
  error: [Function: error],
  groupCollapsed: [Function: groupCollapsed],
  Console: [Function: Console],
  profile: [Function: profile],
  profileEnd: [Function: profileEnd],
  timeStamp: [Function: timeStamp],
  context: [Function: context]
}
> .exit

$ truffle develop
bigint: Failed to load bindings, pure JS will be used (try npm run rebuild?)
Truffle Develop started at http://127.0.0.1:9545/

< -- SNIP Accounts header -- >

truffle(develop)> console
truffle(develop)> .exit
truffle(develop)> .exit
╭─amal@faifai  ~/dev/scratch/2022/09/29/test-nesting ‹node-16.17.1› ‹truffle-core›
╰─$

gnidan:
It used to be that Truffle wouldn't let you run `console` or `develop` from within its repl. We should probably go back to that behavior.

cds-amal:
[...]

PR #5579, opened 2022-09-30
Fix: set and enforce allowed truffle commands in REPL

cds-amal:
This PR sets and enforces allowed truffle commands that should run in the `console` and `develop` REPLS. It should address #5578. Happy to include a test if anyone feels it is necessary.

## How to test

1. create new truffle project
1. `truffle develop`
    1. verify `develop` returns an error.
    1. verify `console` returns the JavaScript console object.
    1. verify  `truffle <unknown command> presents the appropriate message.
    1. verify `truffle [ console | develop ]` presents the appropriate message.

### DIFF
--- a/packages/core/lib/commands/console/run.js
+++ b/packages/core/lib/commands/console/run.js
@@ -20,13 +20,14 @@ module.exports = async function (options) {
 
   let config = loadConfig(options);
 
-  const consoleCommands = commands.reduce((acc, name) => {
-    return !excludedCommands.has(name)
-      ? Object.assign(acc, { [name]: commands[name] })
-      : acc;
-  }, {});
+  const allowedConsoleCommands = commands.filter(
+    cmd => !excludedCommands.has(cmd)
+  );
 
   await Environment.detect(config);
-  const c = new Console(consoleCommands, config.with({ noAliases: true }));
+  const c = new Console(
+    allowedConsoleCommands,
+    config.with({ noAliases: true })
+  );
   return await c.start();
 };
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -15,19 +15,24 @@ const spawnSync = require("child_process").spawnSync;
 const Require = require("@truffle/require");
 const debug = require("debug")("console");
 const { getCommand } = require("./command-utils");
+const validTruffleCommands = require("./commands/commands");
+
+// Create an expression that returns a string when evaluated
+// by the REPL
+const makeIIFE = str => `(() => "${str}")()`;
 
 const processInput = (input, allowedCommands) => {
   const words = input.trim().split(/\s+/);
   if (words.length === 0) return input;
 
   if (words[0] === "truffle") {
-    if (allowedCommands.includes(words[1])) {
-      return words.slice(1).join(" ");
+    const cmd = words[1];
+    if (validTruffleCommands.includes(cmd)) {
+      return allowedCommands.includes(cmd)
+        ? words.slice(1).join(" ")
+        : makeIIFE("ℹ️ : 'truffle ${cmd}' is not allowed within Truffle REPL");
     }
-    // Log friendly information
-    // return console.info expression which, when processed, will not modify
-    // the `_` variable.
-    return `console.info("ℹ️ : 'truffle ${words[1]}' is not allowed within Truffle REPL")`;
+    return makeIIFE("ℹ️ : 'truffle ${cmd}' is not a valid Truffle command");
   }
 
   return input.trim();
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -17,7 +17,7 @@ const debug = require("debug")("console");
 const { getCommand } = require("./command-utils");
 
 const processInput = (input, allowedCommands) => {
-  const positionals = input.trim().split(" ");
+  const positionals = input.trim().split(/\s+/);
   if (positionals.length === 0) return input;
 
   if (positionals[0] === "truffle") {
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -23,18 +23,30 @@ const makeIIFE = str => `(() => "${str}")()`;
 
 const processInput = (input, allowedCommands) => {
   const words = input.trim().split(/\s+/);
+
+  // empty input
   if (words.length === 0) return input;
 
-  if (words[0] === "truffle") {
+  // maybe truffle command
+  if (words[0].toLowerCase() === "truffle") {
     const cmd = words[1];
-    if (validTruffleCommands.includes(cmd)) {
-      return allowedCommands.includes(cmd)
-        ? words.slice(1).join(" ")
-        : makeIIFE(`ℹ️ : 'truffle ${cmd}' is not allowed within Truffle REPL`);
+
+    if (cmd === undefined) {
+      return makeIIFE(
+        `ℹ️ : 'Missing truffle command. Please include a valid truffle command.`
+      );
     }
-    return makeIIFE(`ℹ️ : 'truffle ${cmd}' is not a valid Truffle command`);
+
+    const normalizedCommand = cmd.toLowerCase();
+    if (validTruffleCommands.includes(normalizedCommand)) {
+      return allowedCommands.includes(normalizedCommand)
+        ? words.slice(1).join(" ")
+        : makeIIFE(`ℹ️ : '${cmd}' is not allowed within Truffle REPL`);
+    }
+    return makeIIFE(`ℹ️ : '${cmd}' is not a valid Truffle command`);
   }
 
+  // an expression
   return input.trim();
 };
 
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -30,9 +30,9 @@ const processInput = (input, allowedCommands) => {
     if (validTruffleCommands.includes(cmd)) {
       return allowedCommands.includes(cmd)
         ? words.slice(1).join(" ")
-        : makeIIFE("ℹ️ : 'truffle ${cmd}' is not allowed within Truffle REPL");
+        : makeIIFE(`ℹ️ : 'truffle ${cmd}' is not allowed within Truffle REPL`);
     }
-    return makeIIFE("ℹ️ : 'truffle ${cmd}' is not a valid Truffle command");
+    return makeIIFE(`ℹ️ : 'truffle ${cmd}' is not a valid Truffle command`);
   }
 
   return input.trim();
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -27,7 +27,7 @@ const processInput = input => {
 };
 
 class Console extends EventEmitter {
-  constructor(tasks, options) {
+  constructor(allowedCommands, options) {
     super();
     EventEmitter.call(this);
 
@@ -44,6 +44,7 @@ class Console extends EventEmitter {
       "build_directory"
     ]);
 
+    this.allowedCommands = allowedCommands;
     this.options = options;
 
     this.repl = null;
@@ -285,6 +286,7 @@ class Console extends EventEmitter {
     // processInput returns a sanitized string
     const processedInput = processInput(input);
     if (
+      processedInput in this.allowedCommands &&
       getCommand({
         inputStrings: processedInput.split(" "),
         options: {},
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -16,13 +16,16 @@ const Require = require("@truffle/require");
 const debug = require("debug")("console");
 const { getCommand } = require("./command-utils");
 
-const processInput = input => {
-  const inputComponents = input.trim().split(" ");
-  if (inputComponents.length === 0) return input;
+const processInput = (input, allowedCommands) => {
+  const positionals = input.trim().split(" ");
+  if (positionals.length === 0) return input;
 
-  if (inputComponents[0] === "truffle") {
-    return inputComponents.slice(1).join(" ");
+  if (positionals[0] === "truffle") {
+    return allowedCommands.includes(positionals[1])
+      ? positionals.slice(1).join(" ") // remove truffle prefix
+      : ""; // treat disallowed command as empty string
   }
+
   return input.trim();
 };
 
@@ -283,8 +286,7 @@ class Console extends EventEmitter {
   }
 
   interpret(input, context, filename, callback) {
-    // processInput returns a sanitized string
-    const processedInput = processInput(input);
+    const processedInput = processInput(input, this.allowedCommands);
     if (
       this.allowedCommands.includes(processedInput.split(" ")[0]) &&
       getCommand({
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -17,13 +17,17 @@ const debug = require("debug")("console");
 const { getCommand } = require("./command-utils");
 
 const processInput = (input, allowedCommands) => {
-  const positionals = input.trim().split(/\s+/);
-  if (positionals.length === 0) return input;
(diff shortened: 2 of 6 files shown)


packages/core/lib/console.js
```diff
@@ -27,7 +27,7 @@ const processInput = input => {
 };
 
 class Console extends EventEmitter {
-  constructor(tasks, options) {
+  constructor(allowedCommands, options) {
```
cds-amal, line 30: `tasks` was previously unused. This should have been caught by `husky`'s `eslint`. I assume this got in because of `--no-verify`. Please be careful when using that commit option.

cds-amal (commented):
Leave notice for reviewers

lsqproduction (commented):
Can we have a better error message for when `develop` is used inside the repl?

packages/core/lib/commands/console/run.js
```diff
@@ -20,13 +20,14 @@ module.exports = async function (options) {
 
   let config = loadConfig(options);
 
-  const consoleCommands = commands.reduce((acc, name) => {
-    return !excludedCommands.has(name)
-      ? Object.assign(acc, { [name]: commands[name] })
-      : acc;
-  }, {});
+  const allowedConsoleCommands = commands.filter(
```
eggplantzzz, line 23: yeah much simpler

packages/core/test/lib/console.js
```diff
@@ -8,11 +8,10 @@ const Web3 = require("web3");
 const { Resolver } = require("@truffle/resolver");
 const config = new Config();
 
-const consoleCommands = Object.keys(commands).reduce((acc, name) => {
-  return !excludedCommands.has(name)
-    ? Object.assign(acc, { [name]: commands[name] })
-    : acc;
-}, {});
+console.log(commands);
```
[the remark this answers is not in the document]
gnidan: good catch :)

packages/core/lib/console.js
```diff
@@ -285,6 +286,7 @@ class Console extends EventEmitter {
     // processInput returns a sanitized string
     const processedInput = processInput(input);
     if (
+      processedInput in this.allowedCommands &&
```
eggplantzzz, line 289:
```suggestion
      this.allowedCommands.includes(processedInput.split(" ")[0]) &&
```

gnidan (changes requested):
If you type `truffle console`, apparently that also returns the `Console` object

packages/core/lib/console.js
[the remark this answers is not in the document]
cds-amal: Thanks @eggplantzzz !

cds-amal:
> Can we have a better error message for when `develop` is used inside the repl?

I think the evaluation step in REPL is sufficient. `develop` should return `Uncaught ReferenceError: develop is not defined`.  OTOH, `truffle [console | develop]` could provide better feedback:

```
truffle(develop)> truffle develop
'truffle develop' is not allowed within Truffle REPL
undefined
truffle(develop)> truffle console
'truffle console' is not allowed within Truffle REPL
undefined
truffle(develop)>
```

Thoughts? @eggplantzzz, @gnidan, @dongmingh?

eggplantzzz:
What could a user possibly be trying to do by running `truffle develop/console` in the repl? I don't know, I'm not sure we need to provide a special error message but I don't feel strongly.

dongmingh:
If I were the user, I would expect to see a message that shows why this does not work in addition to `undefined`.  Hence I would like to see that truffle can provide more information to the user if possible.  So, I like the suggestion: add a message like `'truffle develop' is not allowed within Truffle REPL` in addition to `undefined`.

cds-amal:
Two votes for feedback! Let's see what this would look like.

sukanyaparashar:
> What could a user possibly be trying to do by running `truffle develop/console` in the repl? I don't know, I'm not sure we need to provide a special error message but I don't feel strongly. Would people find it helpful?

I kind of agree with @eggplantzzz. Users using `truffle develop` or `truffle console` must be knowing that they are already within a `truffle REPL`.

eggplantzzz:
Ok, in the case that we want to display a message, we should probably activate it for all non-allowed Truffle commands.

cds-amal:
> Ok, in the case that we want to display a message, we should probably activate it for all Truffle commands not allowed in the repl context.

The code should display the appropriate message from: `ℹ️ : 'truffle [console | db | develop | init | watch]' is not allowed within Truffle REPL` and return `undefined`

packages/core/lib/console.js
```diff
@@ -16,18 +16,25 @@ const Require = require("@truffle/require");
...
-  if (inputComponents[0] === "truffle") {
-    return inputComponents.slice(1).join(" ");
+  if (words[0] === "truffle") {
+    if (allowedCommands.includes(words[1])) {
+      return words.slice(1).join(" ");
+    }
+    // Log friendly information
+    // return console.info expression which, when processed, will not modify
+    // the `_` variable.
+    return `console.info("ℹ️ : 'truffle ${words[1]}' is not allowed within Truffle REPL")`;
```
eggplantzzz, line 30: We also need to check for disallowed commands in the case the user enters input in the following form: `<command> --<option> <value>...`.

Besides that I think this is also the wrong place for this check, no? I feel like the input should be processed (removing the "truffle") and then the check should occur. Do you disagree?

packages/core/lib/console.js
```diff
@@ -16,18 +16,25 @@ const Require = require("@truffle/require");
...
-  const inputComponents = input.trim().split(" ");
-  if (inputComponents.length === 0) return input;
+const processInput = (input, allowedCommands) => {
+  const words = input.trim().split(/\s+/);
+  if (words.length === 0) return input;
 
-  if (inputComponents[0] === "truffle") {
-    return inputComponents.slice(1).join(" ");
+  if (words[0] === "truffle") {
+    if (allowedCommands.includes(words[1])) {
```
eggplantzzz, line 24: Ah, there is another problem with this method of checking. If you implement it this way, it will display this message every time a user enters something that is not a command after "truffle". So `truffle tyler` should trigger this message. We probably want to just check for other Truffle commands that are not "allowed", right?

packages/core/lib/console.js
```diff
@@ -16,18 +16,25 @@ const Require = require("@truffle/require");
 const debug = require("debug")("console");
 const { getCommand } = require("./command-utils");
 
-const processInput = input => {
-  const inputComponents = input.trim().split(" ");
-  if (inputComponents.length === 0) return input;
+const processInput = (input, allowedCommands) => {
+  const words = input.trim().split(/\s+/);
```
lsqproduction, line 20: neat 😸

packages/core/lib/console.js
lsqproduction: would moving the validation part to a different function make sense here?

packages/core/lib/console.js
dongmingh: It is a good idea to add a notice.

packages/core/lib/console.js
cds-amal: Oh yea, you're correct!

dongmingh:
An inconsistent return between truffle command and non-truffle supported command.

For example,
`truffle(develop)> truffle compile` and `truffle(develop)> compile` return the same result.
But `truffle(develop)> truffle develop` and  `truffle(develop)> develop` return different messages.

```
truffle(develop)> truffle develop
ℹ️ : 'truffle develop' is not allowed within Truffle REPL
undefined
```
and
```
truffle(develop)> develop
evalmachine.<anonymous>:0
develop
^

Uncaught ReferenceError: develop is not defined
    at evalmachine.<anonymous>
    at sigintHandlersWrap (node:vm:276:12)
    at Script.runInContext (node:vm:139:14)
    at runScript (/Users/dongminghwang/work/trufflesuite/truffle/packages/core/lib/console.js:389:21)
    at Console.interpret (/Users/dongminghwang/work/trufflesuite/truffle/packages/core/lib/console.js:404:21)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:902:10)
    at REPLServer.emit (node:events:513:28)
    at REPLServer.emit (node:domain:489:12)
```  

For a non-truffle supported command, shall both `truffle cmd` and `cmd` return the same message?

packages/core/lib/console.js
cds-amal: Yea, this will have to be refactored, to handle validation at this seam, and `Console`'s clients and tests will have to be updated as well. 🦆 🦆 🦆 :/

packages/core/lib/console.js
cds-amal: > If you type `truffle console`, apparently that also returns the `Console` object

Thanks, @gnidan. I updated the behavior so
- `truffle <restricted-command>` will log the command isn't allowed in the REPL.
- `truffle <unknown-command>` will log the command isn't a valid Truffle command.
- `any expression that doesn't start with truffle` will be interpreted by the REPL

packages/core/lib/console.js
cds-amal: Turns out that there wasn't a need to refactor as we can grab the list of valid truffle commands, use it to determine branching logic.

cds-amal:
> For a non-truffle supported command, shall both truffle cmd and cmd return the same message?

@dongmingh, I don't think they should be the same. `truffle <unsupported command` has enough information to know that the user intended to run a truffle command. However, `<unsupported truffle command>` is not distinguishable between an unsupported command and a variable the user is free to use in a REPL. I chose to handle the former as a special case, and leave the latter to the REPL.

dongmingh:
That makes sense.  I agree with the way you handle it. Thanks for explanation.

dongmingh (approved):
The returned messages of `truffle develop` and `develop` in `truffle(develop)` make sense.  The changes of the codes look good.  I approve the PR since the CI failure is not relevant to this PR.

lsqproduction (dismissed):
Looks good.

dongmingh (changes requested):
The returned message for `truffle console`, `truffle develop` and `truffle bad_command` has the same error, like this
"ℹ️ : 'truffle ${cmd}' is not allowed within Truffle REPL"
The message contains ${cmd} instead of the actual command name.  The issue is caused by the message in processInput, see core/lib/console.js lines 33 and 35.  The ${cmd} needs to be enclosed with "`".

dongmingh (approved):
The changes look good.
The tests show that the command names are displayed correctly in the returned messages and the results are satisfactory:
```
truffle(develop)> truffle develop
"ℹ️ : 'truffle develop' is not allowed within Truffle REPL"

truffle(develop)> truffle console
"ℹ️ : 'truffle console' is not allowed within Truffle REPL"

truffle(develop)> truffle bad_cmd
"ℹ️ : 'truffle bad_cmd' is not a valid Truffle command"

truffle(develop)> truffle
"ℹ️ : 'truffle undefined' is not a valid Truffle command"

truffle(develop)> bad_cmd
evalmachine.<anonymous>:0
bad_cmd
^

Uncaught ReferenceError: bad_cmd is not defined
    at evalmachine.<anonymous>
    at sigintHandlersWrap (node:vm:276:12)
    at Script.runInContext (node:vm:139:14)
    at runScript (/Users/dongminghwang/work/trufflesuite/truffle/packages/core/lib/console.js:394:21)
    at Console.interpret (/Users/dongminghwang/work/trufflesuite/truffle/packages/core/lib/console.js:409:21)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:902:10)
    at REPLServer.emit (node:events:513:28)
    at REPLServer.emit (node:domain:489:12)
```

Also, compilation and migration work as expected.

eggplantzzz:
So the 'truffle undefined' message above is not the best I guess? Should we add some code to account for when only `truffle` is entered in the repl?

dongmingh:
@cds-amal We may want to show `truffle help` along with a list of supported commands when the user enters an invalid command or no command. With this, we have a common point of answer, i.e., `truffle help`, and the user can see all supported commands and choose the one he needs.  If you and/or @lsqproduction think this is too much, we can tell the user `a command is needed` and type `truffle help` for help.

lsqproduction:
> @cds-amal We may want to show `truffle help` along with a list of supported commands when the user enters an invalid command or no command. With this, we have a common point of answer, i.e., `truffle help`, and the user can see all supported commands and choose the one he needs. If you and/or @lsqproduction think this is too much, we can tell the user `a command is needed` and type `truffle help` for help.

This makes the best sense to me as well.

lsqproduction:
After our discussion, we've decided the below two cases are outside of the scope of this PR.  New issues are opened.
#5620 
```
truffle(develop)> truffle
"ℹ️ : 'truffle undefined' is not a valid Truffle command"
```
#5619 
`truffle(develop)> bad_cmd`

packages/truffle/test/scenarios/commands/develop.js
```diff
@@ -68,6 +68,50 @@ describe("truffle develop", function () {
...
+  describe("Improper truffle commands", async function () {
+    this.timeout(7000);
+
+    [
+      {
+        cmd: "console",
+        expectedError: `ℹ️ : 'console' is not allowed within Truffle REPL`
+      },
+      {
+        cmd: "CONSOLE",
```
lsqproduction, line 80: Curious about how you thought to add `CONSOLE` as well.

lsqproduction (approved):
I like it.

packages/core/lib/console.js
```diff
@@ -15,19 +15,43 @@ const spawnSync = require("child_process").spawnSync;
...
+      );
+    }
+
+    const normalizedCommand = cmd.toLowerCase();
+    if (validTruffleCommands.includes(normalizedCommand)) {
+      return allowedCommands.includes(normalizedCommand)
+        ? words.slice(1).join(" ")
+        : makeIIFE(`ℹ️ : '${cmd}' is not allowed within Truffle REPL`);
+    }
+    return makeIIFE(`ℹ️ : '${cmd}' is not a valid Truffle command`);
```
cds-amal, line 46: @lsqproduction, `validTruffleCommands` are lowercased, so we should convert a potential `cmd` to lowercase for comparison. While the comparison is done in lowercase, the output uses the exact casing the user entered.

packages/truffle/test/scenarios/commands/develop.js
cds-amal: [See this comment](https://github.com/trufflesuite/truffle/pull/5579/files#r1002058952)

dongmingh (approved):
The changes look good. The test results are good:

truffle(develop)> truffle
"ℹ️ : 'Missing truffle command. Please include a valid truffle command."
truffle(develop)> truffle develop
"ℹ️ : 'develop' is not allowed within Truffle REPL"
truffle(develop)> truffle console
"ℹ️ : 'console' is not allowed within Truffle REPL"
truffle(develop)> bad_cmd
evalmachine.<anonymous>:0
bad_cmd
^

Uncaught ReferenceError: bad_cmd is not defined
    at evalmachine.<anonymous>
    at sigintHandlersWrap (node:vm:276:12)
    at Script.runInContext (node:vm:139:14)
    at runScript (/Users/dongminghwang/work/trufflesuite/truffle/packages/core/lib/console.js:406:21)
    at Console.interpret (/Users/dongminghwang/work/trufflesuite/truffle/packages/core/lib/console.js:421:21)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:902:10)
    at REPLServer.emit (node:events:513:28)
    at REPLServer.emit (node:domain:489:12)

Also, compilation and migration work as expected.

gnidan (approved):

PR #5588, opened 2022-10-03
Correctly check for non-integer credentials

RyanCavanaugh:
Values in JavaScript are never equal to `NaN`, so this check wasn't doing anything. Found with TypeScript 4.9.


haltman-at (approved):
Yikes, thanks, good catch!

cds-amal:
`NaN` is a valid outcome of `parseInt`. I don't understand this change.

cds-amal (approved):
Oops, took me a second! Good catch. Thanks!

Issue #5466, opened 2022-08-19
Truffle debugger & decoder should handle new-style internal function pointers


haltman-at:
## Issue

So, if you compile Solidity with `viaIR` turned on, it uses a different representation for internal function pointers.  Instead of them containing actual program counter values, internal functions are now represented by an abstract index.

Fortunately, this index isn't hard to compute.  The order that the functions are given in is just the order they appear in the contract's source, and if they're inherited, base contracts appear before derived contracts.  Note that functions marked `external` are skipped, and so are functions marked `private` if they're in a base contract.  Indexing starts at 1, as 0 is used for the designated invalid function.

(Thought: I just realized I didn't account for free functions.  Crud.  I'll have to determine how those work; hopefully those don't screw this up too much.  I should probably also test whether internal library functions affect things in any way.  I'm hoping I don't end up having to file issues with Solidity...)

(They're still 8 bytes, thankfully, even though that was excessive even with the old format and especially excessive now.  Guess they wanted to limit the breakage.)

Anyway, debugger and decoder should have functionality for handling these.  Note that the main difference will not be in Codec itself, seeing as internal function pointers are decoded with a passed-in lookup table; it's constructing that table that will be different.  (Possibly much of the new code will go in source-map-utils... or maybe Codec's AST utils, it could go there too.)  However, I think we'd also probably want to revise the output format for internal function pointers, which previously assumed that internal function pointers would necessarily be given in terms of PC value.  This would of course be a breaking change to the output format, but since I don't think much external to us is thinking about internal function pointers, I don't think it'd be a big problem.

## Environment

* Truffle version (`truffle version`): 5.5.27

haltman-at:
So, this has turned out to be more of an issue than I realized -- it turns out that indices are only assigned to functions that get pointed to, which is substantially harder to determine.  We may have to wait for https://github.com/ethereum/solidity/issues/13425 before properly addressing this, although determining that isn't necessarily *impossible*, just a bunch of work.  Still, I'm putting aside for now the problem of finding a proper solution to this.

Smithbull001:
> ## Issue
> So, if you compile Solidity with `viaIR` turned on, it uses a different representation for internal function pointers. Instead of them containing actual program counter values, internal functions are now represented by an abstract index.
> 
[...]

PR #5596, opened 2022-10-05
Disable internal function pointer decoding for contracts compiled with `viaIR`

haltman-at:
So, the original plan for this PR was for it to address #5466.  Unfortunately, that turned out to be rather more difficult than anticipated.  As such, I've put up this PR as a stopgap.  (I did do a bunch of work on the original plan before I realized the problem; if people are interested, I can push that branch as well so people can see it.  I *am* intending to still use a lot of that work in the future when we attempt to handle this more properly, but the time for that is not now.)

So, anyway, as a stopgap, if `viaIR` is turned on, instead of decoding using indices like I was hoping to do, we just disable decoding of internal function pointers.  That way at least we don't get silly errors that don't apply.

There's more to that than it sounds, though, because -- just as if we wanted to switch to index based-decoding when appropriate -- we still have to actually determine if `viaIR` is turned on!  We've often in the past made switches based on compiler version, but never before on compilation settings.  That means that we need to get that information to the debugger and decoder, where we've never had to before.

Doing this isn't that hard, because after all we can parse it out of the `metadata` (if available).  Note that for now we default to `viaIR` off, but in the future we may default it differently for different Solidity versions.  But the big thing to note here is that I've made alterations to the `Compilation`, `Contract`, and `Source` types, so @gnidan take note!

Note I didn't use a `Settings` type from elsewhere, I just made a custom one and included `viaIR` as the only setting because, well, it's the only one that matters right now.

There is one caveat to this PR, which is that, when decoding of internal function pointers is disabled, the unhelpful value you get still lists a pair of PC values, even though in fact that doesn't apply in these cases.  But I didn't see that as being worth changing until we're actually decoding index-based values.  Again, on my `newtable` branch I did do the work to support this kind of thing, so if you want to see that branch, I can push it and you can see what that looks like.  In this PR though I left all that out.

So yeah, a bit of a disappointment, but better than doing nothing for now I guess...

(I didn't add any tests; the way the debugger and decoder tests are set up make it a pain to use multiple different compilation settings.  But I did test this manually.)

(Oh, also, I fixed an unrelated crash that came up in testing.  Um, that maybe should have been a separate PR.  Oh well.  It's a separate commit at least, so we can always cherry-pick it if need be.)

### DIFF
--- a/packages/source-map-utils/index.js
+++ b/packages/source-map-utils/index.js
@@ -450,12 +450,20 @@ var SourceMapUtils = {
         return null;
       }
       index++;
-      while (instructions[index].name.match(/^PUSH\d*/)) {
+      while (
+        instructions[index] &&
+        instructions[index].name.match(/^PUSH\d*/)
+      ) {
         index++;
         if (index > startingIndex + 3) {
           //check: are there more than 2 PUSHes?
           return null;
         }
+      }
+      if (!instructions[index]) {
+        //covers both the case where we ran off already,
+        //and where we're about to run off
+        return null;
       }
       if (instructions[index].name === "JUMP") {
         if (index === startingIndex + 1) {
--- a/packages/codec/lib/compilations/types.ts
+++ b/packages/codec/lib/compilations/types.ts
@@ -43,6 +43,11 @@ export interface Compilation {
    * specified on each source and contract, but please don't actually do that.
    */
   compiler?: Compiler.CompilerVersion;
+  /**
+   * The settings used in the compilation.  Works similarly to the compiler field.
+   * May be omitted.
+   */
+  settings?: Compiler.Settings;
 }
 
 /**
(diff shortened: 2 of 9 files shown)


packages/source-map-utils/index.js
```diff
@@ -450,13 +450,21 @@ var SourceMapUtils = {
         return null;
       }
       index++;
-      while (instructions[index].name.match(/^PUSH\d*/)) {
+      while (
+        instructions[index] &&
+        instructions[index].name.match(/^PUSH\d*/)
```
cds-amal, line 455: Should this be `/^PUSH\d+/` ?

cds-amal (approved):
Makes sense, and aligns with your PR explanation; thanks for that! 

One request, please; can you add an issue for the test setup? I'd add it myself, but I wouldn't be able to describe it fully. It seems like tech debt we should address some time.

> (I didn't add any tests; the way the debugger and decoder tests are set up make it a pain to use multiple different compilation settings. But I did test this manually.)

packages/source-map-utils/index.js
haltman-at: I mean, it doesn't really matter in this context?  There's no such instruction as plain `PUSH`.

haltman-at:
> One request, please; can you add an issue for the test setup? I'd add it myself, but I wouldn't be able to describe it fully. It seems like tech debt we should address some time.

Honestly, I don't think this would be that difficult to change, I just didn't want to do it together with this.  I'll put up an issue for it, yeah.

haltman-at:
OK, put up #5627 for that.

haltman-at:
OK, I showed @gnidan the part I was worried about and he was OK with it, so I'm merging this. :)

PR #5598, opened 2022-10-06
Add support for compiler plugin

dongmingh:
This PR adds two APIs, definesCompiler() and loadCompiler(), to the Plugin class to support compiler plugin. These APIs are exposed to workflow-compile that it can access the plugin compiler.  There are on-going work on workflow-compile to obtain the Plugin object and access compiler.

### DIFF
--- a/packages/plugins/lib/Plugin.ts
+++ b/packages/plugins/lib/Plugin.ts
@@ -69,7 +69,7 @@ export class Plugin {
   loadCompiler(): any {
     if (!this.definesCompiler()) {
       throw new TruffleError(
-        `Plugin ${this.module} does not define a \`truffle plugin\` compiler.`
+        `Plugin ${this.module} does not define a \`truffle compiler plugin\`.`
       );
     }
 
--- a/packages/plugins/lib/Plugin.ts
+++ b/packages/plugins/lib/Plugin.ts
@@ -63,7 +63,7 @@ export class Plugin {
    */
 
   definesCompiler(): boolean {
-    return !!this.definition.compile;
+    return !!this.definition.compiler;
   }
 
   loadCompiler(): any {
@@ -73,7 +73,7 @@ export class Plugin {
       );
     }
 
-    return this.loadModule(this.definition.compile).Compile;
+    return this.loadModule(this.definition.compiler).Compile;
   }
 
   /*
--- a/packages/plugins/lib/Plugin.ts
+++ b/packages/plugins/lib/Plugin.ts
@@ -73,7 +73,11 @@ export class Plugin {
       );
     }
 
-    return this.loadModule(this.definition.compiler).Compile;
+    const compilerPath = this.loadModule(this.definition.compiler);
+    if (!compilerPath.Compile) {
+      throw new TruffleError(`Plugin ${this.module}: Compile is not defined.`);
+    }
+    return compilerPath.Compile;
   }
 
   /*
--- a/packages/plugins/lib/Plugin.ts
+++ b/packages/plugins/lib/Plugin.ts
@@ -59,6 +59,24 @@ export class Plugin {
   }
 
   /*
+   * compiler
+   */
+
+  definesCompiler(): boolean {
+    return !!this.definition.compile;
+  }
+
+  loadCompiler(): any {
+    if (!this.definesCompiler()) {
+      throw new TruffleError(
+        `Plugin ${this.module} does not define a \`truffle plugin\` compiler.`
+      );
+    }
+
+    return this.loadModule(this.definition.compile).Compile;
+  }
+
+  /*
    * internals
    */
 
--- a/packages/plugins/test/Plugin.test.ts
+++ b/packages/plugins/test/Plugin.test.ts
@@ -199,4 +199,38 @@ describe("Plugin", () => {
       expect(() => plugin.loadRecipe()).toThrow(expectedError);
     });
   });
+
+  describe("loadCompiler()", () => {
+    it("should load compiler defined in the plugin definition", () => {
+      const plugin = new Plugin({
+        module: "dummy-compiler",
+        definition: { compile: "index.js" }
+      });
+
+      const loadedCompiler = plugin.loadCompiler();
+      const expectedResult = "Successfully called dummy-compiler:compile()";
+      expect(loadedCompiler.compile()).toEqual(expectedResult);
+    });
+
+    it("should throw when compiler plugin definition is an absolute path", () => {
+      const plugin = new Plugin({
+        module: "dummy-compiler",
+        definition: { compile: "/index.js" }
+      });
+
+      const expectedError = "Absolute paths not allowed!";
+      expect(() => plugin.loadCompiler()).toThrow(expectedError);
+    });
+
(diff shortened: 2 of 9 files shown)


packages/plugins/test/Plugin.test.ts
```diff
@@ -199,4 +199,38 @@ describe("Plugin", () => {
       expect(() => plugin.loadRecipe()).toThrow(expectedError);
     });
   });
+
+  describe("loadCompiler()", () => {
```
gnidan, line 203: It's great that you're adding tests for this functionality!

packages/plugins/lib/Plugin.ts
```diff
@@ -58,6 +58,24 @@ export class Plugin {
...
+  }
+
+  loadCompiler(): any {
+    if (!this.definesCompiler()) {
+      throw new TruffleError(
+        `Plugin ${this.module} does not define a \`truffle plugin\` compiler.`
+      );
+    }
+
+    return this.loadModule(this.definition.compile).Compile;
```
dkillen, line 76:
```suggestion
    return this.loadModule(this.definition.compiler).Compiler;
```

packages/plugins/lib/Plugin.ts
```diff
@@ -58,6 +58,24 @@ export class Plugin {
     return this.loadModule(this.definition.preserve.recipe).Recipe;
   }
 
+  /*
+   * compiler
+   */
+
+  definesCompiler(): boolean {
+    return !!this.definition.compile;
```
dkillen, line 66:
```suggestion
    return !!this.definition.compiler;
```

packages/plugins/test/Plugin.test.ts
```diff
@@ -199,4 +199,38 @@ describe("Plugin", () => {
...
+    it("should load compiler defined in the plugin definition", () => {
+      const plugin = new Plugin({
+        module: "dummy-compiler",
+        definition: { compile: "index.js" }
```
dkillen, line 207:
```suggestion
        definition: { compiler: "index.js" }
```

packages/plugins/test/Plugin.test.ts
```diff
@@ -199,4 +199,38 @@ describe("Plugin", () => {
...
+
+      const loadedCompiler = plugin.loadCompiler();
+      const expectedResult = "Successfully called dummy-compiler:compile()";
+      expect(loadedCompiler.compile()).toEqual(expectedResult);
+    });
+
+    it("should throw when compiler plugin definition is an absolute path", () => {
+      const plugin = new Plugin({
+        module: "dummy-compiler",
+        definition: { compile: "/index.js" }
```
dkillen, line 218:
```suggestion
        definition: { compiler: "/index.js" }
```

packages/plugins/test/Plugins.test.ts
```diff
@@ -32,6 +37,10 @@ describe("Plugins", () => {
             tag: "dummy-recipe",
             preserve: { tag: "dummy-recipe", recipe: "." }
           }
+        }),
+        new Plugin({
+          module: "dummy-compiler",
+          definition: { compile: "index.js" }
```
dkillen, line 43:
```suggestion
          definition: { compiler: "index.js" }
```

packages/plugins/test/Plugins.test.ts
```diff
@@ -185,4 +195,24 @@ describe("Plugins", () => {
...
+        working_directory: __dirname,
+        plugins: ["dummy-compiler"]
+      };
+
+      const foundPlugins = Plugins.listAllCompilers(config);
+
+      const expectedPlugins = [
+        new Plugin({
+          module: "dummy-compiler",
+          definition: { compile: "index.js" }
```
dkillen, line 211:
```suggestion
          definition: { compiler: "index.js" }
```

packages/plugins/test/fixture/dummy-compiler/index.js
```diff
@@ -0,0 +1,7 @@
+const Compile = {
```
dkillen, line 1:
```suggestion
const Compiler = {
```

dkillen (commented):
Hi @dongmingh I have made some suggestions for minor changes which appear to me to be consistent with the way other plugins are defined. In that regard, it seems to me that you want to refer to a `compiler/Compiler` rather than `compile/Compile`.

packages/plugins/test/fixture/dummy-compiler/index.js
```diff
@@ -0,0 +1,7 @@
...
+  compile() {
+    return "Successfully called dummy-compiler:compile()";
+  }
+};
+
+module.exports = { Compile };
```
dkillen, line 7:
```suggestion
module.exports = { Compiler };
```

fainashalts (approved):
This looks good, Dongming! I had one suggestion and it looks like Dave had a couple as well. Otherwise I think you can merge this into the `compiler-plugin-system` branch as we'll be giving that one a thorough review also before we merge it into develop when its time!

packages/plugins/lib/Plugin.ts
dongmingh: @dkillen thanks for reviewing the PR.

1. `compile` in definition.compile is used because `compile` is defined in truffle-plugin.json.  @gnidan 
and @fainashalts can we use `compiler` instead of `compile` as definition in truffle-plugin.json?

2. `Compile` in ().Compile is used because Truffle compilers use `Compile` as the name of API's, e.g., see packages/compile-solidity/src/index.js.  The plugin needs to be consistent with Truffle implementation.  Hence, we cannot change it.

cds-amal:
@dongmingh @fainashalts There have been significant changes in production and it might be worthwhile to rebase `compiler-plugin-system` against latest `develop`, and then rebase `zksolc-compiler` against the result.

dongmingh:
@cds-amal Thanks, I will rebase.

packages/plugins/lib/Plugin.ts
dongmingh: Per the discussion in the config meeting, we will change `compile` to `compiler` in the definition.

packages/plugins/test/Plugin.test.ts
dongmingh: Thx

packages/plugins/lib/Plugin.ts
```diff
@@ -58,6 +58,24 @@ export class Plugin {
...
+  }
+
+  loadCompiler(): any {
+    if (!this.definesCompiler()) {
+      throw new TruffleError(
+        `Plugin ${this.module} does not define a \`truffle compiler plugin\`.`
+      );
+    }
+
+    return this.loadModule(this.definition.compiler).Compile;
```
fainashalts, line 76: Should we have error handling for if the `Compile` method is missing?

dongmingh: The error handling is added.

fainashalts:
Hey @dongmingh I believe this can be merged into my branch! This doesn't include the changes we discussed on Monday, right? I think we can merge this in, and then you can create a new branch for those changes.

PR #5604, opened 2022-10-07
delete dependabot.yml

davidmurdoch:
The existence of this file with this configuration disables all security updates. See: https://github.com/dependabot/dependabot-core/issues/5845


.github/dependabot.yml
```diff
@@ -1,20 +0,0 @@
...
-      - dependency-name: "ganache"
-    reviewers:
-      - "trufflesuite/truffle-core-devs"
-  - package-ecosystem: "npm" # this section is for security update PRs
-    directory: "/" # Location of package manifests
-    schedule:
-      interval: "weekly"
-    # Disable version updates for npm dependencies
-    # see https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates#overriding-the-default-behavior-with-a-configuration-file
-    open-pull-requests-limit: 0
```
cds-amal, line 20: @davidmurdoch what happens if this setting is removed?

davidmurdoch: The problem with the current config is that github doesn't allow us to specify `npm` twice. You are looking at the 2nd of 2 npm configs. This `open-pull-requests-limit: 0` setting is the official documented ~hack~ way to enable security-only updates; setting `open-pull-requests-limit` _actually_ sets it to `10` if the Security Updates setting is enabled in the GitHub UI.

haltman-at:
This is no longer necessary, right?

davidmurdoch:
> This is no longer necessary, right?

Have we confirmed that hack from the other PR has re-enabled dependabot?

haltman-at:
Yup!  I've been running it to find dependencies to update and even made one update on that basis. :)

PR #5599, opened 2022-10-06
migrate migrate to TS

eggplantzzz:
revert commit 913784336c2d1be5d1baf3ae9c93da57bd284716

PR #5608, opened 2022-10-10
Bug fix: update gitignore for db and delete accidentally committed file

eggplantzzz:
I believe in https://github.com/trufflesuite/truffle/pull/5599 the `types` directory in db got renamed to `typings` to be consistent with other Truffle packages. The types file that is generated was in the .gitignore, but now with the rename it accidentally got committed. This PR fixes this by adding the correct path in the gitignore.


cds-amal (approved):
Likes! Thanks, @eggplantzzz.

NOTE: CI will be red because of the `dependabot.yml`

haltman-at:
> NOTE: CI will be red because of the `dependabot.yml` issue.

I believe that only affects `develop` itself and not PRs?  Like the PR that introduced that issue wasn't red.

haltman-at (approved):

cds-amal:
> 
> I believe that only affects `develop` itself and not PRs?  Like the PR that introduced that issue wasn't red.

Thanks @haltman-at ! Makes sense.

PR #5609, opened 2022-10-10
Add a check to help prevent generated files being checked into source control

haltman-at:
We had a problem recently where a generated file was accidentally checked into source control and it caused a problem (see #5608) where bootstrapping resulted in changes to tracked files.  I figured it would be good to add this measure to help prevent this.  So I added a check for this in CI.  Note that because this check requires bootstrapping, like `depcheck` it goes in the jobs that *aren't* `yarncheck`.  Obviously this won't catch every case, but it should help, I figure.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -53,6 +53,7 @@ jobs:
     - run: npm install -g yarn
 
     - run: yarn bootstrap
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
Looks good.  Nice one, @haltman-at

PR #5611, opened 2022-10-11
Add arbitrum goerli support to Etherscan fetcher

haltman-at:
Seems Etherscan added support for Arbitrum Goerli at some point, but I didn't notice till now.  Well, I've updated the source fetcher with it.


cds-amal (approved):
Looks good!

Issue #5600, opened 2022-10-06
There is a dependency cycle involving db


eggplantzzz:
@truffle/db has a devDependency on @truffle/migrate even though it is not listed in the package.json. See [this](https://github.com/trufflesuite/truffle/blob/develop/packages/db/src/project/test/integration.spec.ts#L9).

When you add @truffle/migrate to the devDependencies for @truffle/db, lerna complains about the dependency cycle. We should get rid of this somehow.

lsqproduction:
released in truffle [v5.6.4](https://github.com/trufflesuite/truffle/releases/tag/v5.6.4)

PR #5612, opened 2022-10-11
Silence Lerna's cycle dependency warnings

cds-amal:
Remove `@truffle/db`'s devDep on `@truffle/migrate` to get rid of lerna's cycle dep error.
Note: uses CommonJS import to make get past compile and test.

This dependency exists only for a single integration test in truffle-db, and we feel it's better to not impact Lerna's dependency/ordering until we [remove the circular dependency in the near future](https://github.com/trufflesuite/truffle/issues/5600).


cliffoo (approved):
👍

eggplantzzz (approved):

PR #5613, opened 2022-10-12
Update webpack-bundle-analyzer to ^4.6.1

haltman-at:
This PR updates webpack-bundle-analyzer from 3.0.3 to 4.6.1.  Hopefully that doesn't break anything, but given that webpack-bundle-analyzer is only used in one script that doesn't do anything fancy with it, it would be very surprising if it did.  Note, the real point of this PR is to get rid of a dependabot alert about the `ejs` package (since `webpack-bundle-analyzer` was the limiting factor there, and the version I'm updating to doesn't even use `ejs`).  Hopefully it doesn't introduce any new problems!

### DIFF
--- a/packages/truffle/package.json
+++ b/packages/truffle/package.json
@@ -60,7 +60,7 @@
     "tmp": "^0.2.1",
     "web3": "1.7.4",
     "webpack": "^5.73.0",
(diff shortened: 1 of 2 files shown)


cds-amal (approved):
Looks good to me. I verified the `analyze` script still works, presenting information.

Issue #3412, opened 2020-10-05
--verbose-rpc flag on truffle console has no effect


grantwest:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

`--verbose-rpc` flag of `truffle console` command has no effect

## Steps to Reproduce

```
truffle console --network testnet --verbose-rpc
truffle(testnet)> (await web3.eth.getBalance(accounts[0])).toString()
'49935967720960000'
truffle(testnet)>
```

## Expected Behavior

I expect to see an RPC call logged.

## Actual Results

The only thing printed is the balance as shown in the Steps to Reproduce section above.

## Environment

* Operating System: Debian
* Truffle version: v5.1.42
* node version: v12.18.3
* npm version: 6.14.6

eggplantzzz:
Very strange, when I connect to an existing `truffle develop` session I seem to get the rpc calls printed but not when I connect to Ropsten. We'll have to investigate, thank you for reporting this!

shunjizhan:
same for me when using a custom local network, `--verbose-rpc` prints nothing extra

sukanyaparashar:
This should be fixed in [v5.6.2](https://github.com/trufflesuite/truffle/releases/tag/v5.6.2)!

PR #5618, opened 2022-10-14
Bug fix: patch/wrap the `sendAsync` method when wrapping provider

lsqproduction:
Addresses #3412, to log for `--verbose-rpc` with HDWalletProvider.

Currently, the logging from `--verbose-rpc` is added onto the provider from wrapper.js's `pre` and `post` hook methods by patching the `send` method. Sometimes web3 uses the  `sendAsync` method, which we have not pathed, causing the lack of logging. 

We need to execute the `pre` and `post` hooks when `sendAsync` is called.

For testing:
Set up goerli via HDWalletProvider
run `truffle console --verbose-rpc --network goerli`

### DIFF
--- a/packages/provider/wrapper.js
+++ b/packages/provider/wrapper.js
@@ -27,7 +27,7 @@ module.exports = {
 
     /* overwrite method */
     provider.send = this.send(originalSend, preHook, postHook);
-    // path sendAsync when sendAsync is used.
+    // overwrite sendAsync only when sendAsync is used.
     if (provider.sendAsync) {
       const originalSendAsync = provider.sendAsync.bind(provider);
       provider.sendAsync = this.send(originalSendAsync, preHook, postHook);
--- a/packages/provider/wrapper.js
+++ b/packages/provider/wrapper.js
@@ -27,7 +27,7 @@ module.exports = {
 
     /* overwrite method */
     provider.send = this.send(originalSend, preHook, postHook);
-
+    provider.sendAsync = this.send(originalSend, preHook, postHook);
     /* mark as wrapped */
     provider._alreadyWrapped = true;
 
--- a/packages/provider/wrapper.js
+++ b/packages/provider/wrapper.js
@@ -1,5 +1,5 @@
-var debug = require("debug")("provider:wrapper"); // eslint-disable-line no-unused-vars
-var ProviderError = require("./error");
+const debug = require("debug")("provider:wrapper"); // eslint-disable-line no-unused-vars
+const ProviderError = require("./error");
 
 module.exports = {
   /*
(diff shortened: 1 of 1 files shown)


packages/provider/wrapper.js
```diff
@@ -27,7 +27,7 @@ module.exports = {
 
     /* overwrite method */
     provider.send = this.send(originalSend, preHook, postHook);
-
+    provider.sendAsync = this.send(originalSend, preHook, postHook);
```
davidmurdoch, line 30: Using `provider.sendAsync.bind(provider)` instead of `originalSend` seems like it'd be more correct.

davidmurdoch:
Monkey patching is always iffy, but in this case it's the replacement of the `error` value with truffle's own Error class that could _potentially_ cause problems.

Reason: an `Error` object _may_ have additional properties, and truffle doesn't copy those properties to its new custom `ProviderError` object. Other tools or application that expect such properties to exist may not behave correctly.

I'm not sure this fact can cause issues in actual real world use, some smoke additional testing may be needed to be sure.

* note: if this does become an issue because of this PR, it was _already_ an issue, but we've likely been getting away with it because wherever we use `send` (instead of `sendAsync`) may just be safe from this `ProviderError` side effect.
* note 2: I really doubt this is an actual problem; just theoretical :-)

packages/provider/wrapper.js
lsqproduction: yes you're right. let me make the change.

eggplantzzz: @davidmurdoch can you explain your reasoning? I don't understand why this is the case, should we be doing that with line 29 as well where we patch the `send` method?

lsqproduction: Yes, maybe I do not understand it correctly. Line 26 is where `originalSend` was assigned `this.send(originalSend, preHook, postHook)`.

So I thought for reliability it would be better to do `this.send(originalSendAsync, preHook, postHook)'. If I do the above `provider.sendAsync.bind(provider)`, would that just patch the provider's sendAsync method with the provider's send async method?

packages/provider/wrapper.js
```diff
@@ -19,15 +19,19 @@ module.exports = {
...
+    const postHook = this.postHook(options);
 
-    var originalSend = provider.send.bind(provider);
+    const originalSend = provider.send.bind(provider);
 
     /* overwrite method */
     provider.send = this.send(originalSend, preHook, postHook);
-
+    // path sendAsync when sendAsync is used.
+    if (provider.sendAsync) {
```
eggplantzzz, line 31: just a small style nit: it might make this a little more clear if you assign `originalSendAsync` right next to `originalSend` and then assign `provider.sendAsync` right next to `provider.send`. this is not something I feel strongly about so feel free to ignore me

packages/provider/wrapper.js
davidmurdoch: @eggplantzzz  we are doing that with line 29 as well.

packages/provider/wrapper.js
lsqproduction: I tried this. If I used the same pattern on line 29, it would break when using the development network like ganache, since the `provider.sendAsync` method wouldn't be available.  So `truffle console --verbose-rpc` would break.

eggplantzzz: Include it in an `if` so it won't like you've done.

eggplantzzz (approved):
nice

PR #4933, opened 2022-03-24
Prototype transaction decoding in Truffle Dashboard

gnidan:

Issue #4897, opened 2022-03-18
Track latest workflow-compile results on dashboard


MicaiahReid:

MicaiahReid:
See @gnidan for more info :)

cliffoo:
Released in [v5.6.7](https://github.com/trufflesuite/truffle/releases/tag/v5.6.7). Dashboard now persists workflow-compile results.

PR #5621, opened 2022-10-16
Decode requests in dashboard

cliffoo:
### Addressed
#4897

### Supported methods
`eth_sendTransaction`, `personal_sign`, `eth_signTypedData_v3`, `eth_signTypedData_v4`

### Unsupported methods
`eth_decrypt`: MM [says it's deprecated](https://docs.metamask.io/guide/rpc-api.html#eth-decrypt-deprecated).
`eth_signTypedData`: I just couldn't get a valid request to send.
`eth_signTypedData_v1`: This is [supposed to be the same](https://docs.metamask.io/guide/signing-data.html#a-brief-history) as `eth_signTypedData`, but MM seems to think it's not a valid method.

### Screenshots
screenshot

screenshot

screenshot

screenshot

### Idea
diagram

### Other things
- Using the decoder means various polyfills are required, which the latest react-scripts cannot provide. The CRA setup is replaced with webpack + babel + eslint, with minimal loaders and plugins. I also took this opportunity to freshen up the dev server scripts.
- Removed cpy-cli devDep. This is no longer needed as the build command targets the dist folder directly.

### Thanks
@gnidan, @haltman-at for answering my questions! This [PR](https://github.com/trufflesuite/truffle/pull/4933) was helpful also.


haltman-at:
> Using the decoder means web3js is now a transitive dependency, which the latest react-scripts cannot handle.

Oh hm, how'd that happen?  We took `web3` out of the decoder proper a while ago, although it's still in there as a devdep for testing.  We are still using `web3-utils`.  Wondering where the problem is coming from?

cliffoo:
I misspoke... It's not `web3`, but other deps (including `web3-utils`) that require polyfills when webpacking decoder. I've updated my comments.

packages/dashboard-message-bus-common/lib/messages.ts
```diff
@@ -25,9 +27,18 @@ export interface DashboardProviderMessage extends Message {
...
 
+export type WorkflowCompileResultMessageType = "workflow-compile-result";
+export const workflowCompileResultMessageType = "workflow-compile-result";
+/**
+ * Message to inform subscribers of new WorkflowCompileResult
+ */
+export interface WorkflowCompileResultMessage extends Message {
+  type: WorkflowCompileResultMessageType;
+  payload: WorkflowCompileResult;
+}
```
benjamincburns, line 38: I think it's incohesive to define such a specific message type here.

The message bus common package is meant to house logic and types that are common between the message bus service and the message bus client. In general I think it makes sense to define a generic message envelope in here, but I think defining the structure of messages that are specific to some other package (compile-common in this case) is probably a code smell.

The easy test for why it's problematic is that you need to take a dependency on `compile-common` here. Perhaps this could be avoided by defining a more generic and extensible message envelope type here that you could use to embed this same information? For example, for messages to do with RPC calls we just have the `DashboardProviderMessage` type. We don't have one message type per type of RPC call, however.

Perhaps there's a more generic `CLIEventMessageType` structure that would work better here?

If you aren't familiar with how other message busses work, it might be worth having a look at the message envelope types used by MQTT and AMQP.

packages/dashboard/dev-server/all.ts
```diff
@@ -0,0 +1,48 @@
...
+      notice("Stopping servers...");
+      kill();
+      try {
+        await result;
+      } catch (err) {
+        /* We don't really care. */
+      }
+    }
+  });
+}
```
benjamincburns, line 46: It might be better to just register the `kill` function as a `SIGINT` handler via `process.on("SIGINT", kill);` That also doesn't require you to put stdin in raw mode.

https://nodejs.org/api/process.html#signal-events

packages/dashboard/dev-server/dashboard.ts
```diff
@@ -7,6 +7,13 @@ const options = {
...
 };
-
 const dashboardServer = new DashboardServer(options);
-dashboardServer.start();
+
+async function startServer() {
+  await dashboardServer.start();
+  console.log(
+    `Dashboard server running at http://${options.host}:${options.port}`
+  );
```
benjamincburns, line 16: Does this also print when you run `truffle dashboard` in the shipped CLI?

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -70,12 +70,7 @@ export class DashboardServer {
     this.rpc = options.rpc ?? true;
     this.verbose = options.verbose ?? false;
     this.autoOpen = options.autoOpen ?? true;
-    this.frontendPath = path.join(
-      __dirname,
-      ".",
-      "dashboard-frontend",
-      "build"
-    );
+    this.frontendPath = path.join(__dirname, "dashboard-frontend");
```
benjamincburns, line 73: Yeah, so like I was saying on Slack - I think that when we break this out into a separate package, rather than outputting the frontend's webpack build into the server package's directory, we just refactor this line to resolve the frontend package's path via `require.resolve` instead of using a path that's relative to `__dirname`.

benjamincburns: Alternatively you could even have the index of the frontend module export a string array that's a list of paths to be served up by the backend/server package. Or maybe it's an object that defines a set of route keys and paths?

Either way, that allows the frontend some way to tell the backend what it contains, and allows you to add/remove/modify stuff to the frontend without changing the server logic.

packages/dashboard/package.json
```diff
@@ -39,42 +37,66 @@
     "@mantine/hooks": "^5.0.0",
     "@mantine/notifications": "^5.0.0",
     "@mantine/prism": "^5.0.0",
+    "@truffle/codec": "^0.14.8",
+    "@truffle/compile-common": "^0.9.0",
```
benjamincburns, line 41: FWIW, I think it's totally reasonable to take a dep on `compile-common` here.

packages/dashboard/tsconfig.json
```diff
@@ -29,5 +29,10 @@
     "isolatedModules": true,
     "noEmit": true
   },
+  "ts-node": {
+    "compilerOptions": {
+      "module": "commonjs"
+    }
+  },
```
benjamincburns, line 36: Ha, I wound up having to do the same in my branch.

packages/dashboard/webpack.config.ts
```diff
@@ -0,0 +1,103 @@
...
+    },
+    fallback: {
+      crypto: require.resolve("crypto-browserify"),
+      os: require.resolve("os-browserify/browser"),
+      path: require.resolve("path-browserify"),
+      stream: require.resolve("stream-browserify"),
+      vm: require.resolve("vm-browserify")
+    }
+  },
+  devtool: "source-map",
```
benjamincburns, line 59: might want to use the same value for `devtool` as the one that we used in the truffle build.

packages/dashboard/webpack.config.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+  },
+  stats: "minimal",
+  module: {
+    rules: [
+      {
+        test: [/\.jsx?$/, /\.tsx?$/],
+        use: ["babel-loader"],
+        exclude: /node_modules/
+      },
```
benjamincburns, line 78: I think babel-loader is maybe fine, but for consistency with the work that I'm doing, you might want to consider going with ts-loader.

That said, babel loader doesn't check types at all, so unless you add a separate build step to check for type errors, CI will be completely blind to any issues there.

packages/events/defaultSubscribers/dashboard.js
```diff
@@ -39,6 +39,17 @@ module.exports = {
...
+      async function ({ result }) {
+        try {
+          const publishLifecycle = await this.messageBusClient.publish({
+            type: "workflow-compile-result",
+            payload: result
+          });
+          publishLifecycle.abandon();
+        } catch (err) {}
+      }
+    ],
```
benjamincburns, line 52: Yeah, given that this is being sent from the `events` package, I think it would be wise to change this out for a more generic event envelope.

benjamincburns: Nice catch on the `.abandon()`, too.

benjamincburns (changes requested):
Looks good overall - I think there's just one change that I'd really like to see before approving (the message type stuff), and while I'd also like to swap babel-loader for ts-loader, I think I could be convinced to leave it as-is.

packages/dashboard-message-bus-common/lib/messages.ts
benjamincburns: > In general I think it makes sense to define a generic message envelope in here, but I think defining the structure of messages that are specific to some other package (compile-common in this case) is probably a code smell.

Put differently - there's no reason why any of the core message bus service logic _must_ know anything at all about `WorkflowCompileResult` in order do its job effectively. It should be able to treat that as opaque data that it's just shuffling around.

cliffoo: Thanks Ben. Initially I was uncertain about having an envelope message type that represents any CLI event data, mostly because the payload must necessarily be typed `any`. But I figured that I can just type it `WorkflowCompileResult`  in the frontend, which admittedly isn't as nice as the given type guarantees ensured by `isWorkflowCompileResultMessage`, but this is a worthwhile tradeoff compared to taking `compile-common` as a dependency and making more specific message types in the future as dashboard grows, which does smell.

I've made [updates](https://github.com/trufflesuite/truffle/commit/35a9980fa901c2d1add466f0927541ae44d8abed) to use `CliEventMessage` and got rid of `WorkflowCompileResultMessage`.

packages/dashboard/dev-server/all.ts
cliffoo: That would be cleaner but the reason for this is that I found not using raw mode / handling SIGINT directly can cause something like this:
```sh
$ yarn start
[dashboard] Dashboard server running at http://localhost:24012
[webpack] Compiling: 3%
[webpack] ...
^C  # Send kill signal
[dashboard] yarn start:server exited with code SIGINT                                                                                                                                      
--> Sending SIGTERM to other processes..
$ 
# Some time later
[webpack] Finished compiling
[webpack] yarn start:react exited with code SIGINT
```

packages/dashboard/lib/DashboardServer.ts
cliffoo: Resolving the frontend build is definitely the clean way to do this when server becomes its own package. This also means we can avoid copying in the `truffle` webpack config [here](https://github.com/trufflesuite/truffle/blob/v5.6.5/packages/truffle/webpack.config.js#L271-L282).

That list of paths to export _might_ be useful? At the moment the server listens on purpose-specific paths like `/rpc` and `/ports`, and does a [catch-all at the end](https://github.com/trufflesuite/truffle/blob/v5.6.5/packages/dashboard/lib/DashboardServer.ts#L102-L105) to let react browser router handle the rest.

packages/dashboard/package.json
cliffoo: Yea the types are important for type-checking and schema definition. Your point about `WorkflowCompileResultMessage` being too narrow still perfectly stands though.

packages/dashboard/tsconfig.json
cliffoo: I made this change per [this directive](https://webpack.js.org/configuration/configuration-languages/#typescript).

packages/dashboard/webpack.config.ts
cliffoo: Good call! I've [changed](https://github.com/trufflesuite/truffle/commit/7d3729a7c62603536a065f1fb3a6286f5395f180) it to `nosources-source-map`.

packages/dashboard/webpack.config.ts
cliffoo: I certainly wasn't aware of the nuances (babel just throws the types away?? I was fooled by my assumption about [this plugin](https://babeljs.io/docs/en/babel-preset-typescript)). I've [updated](https://github.com/trufflesuite/truffle/commit/2f898673baff2df7dfe63c2893b13d75eadcf64e) it to ts-loader to align with your stuff (and you know, to check types).

packages/events/defaultSubscribers/dashboard.js
cliffoo: Yup! It's now `CliEventMessage`.

packages/dashboard/dev-server/all.ts
benjamincburns: Ah fair enough. Just wanted to make sure that you'd considered it.

packages/dashboard/lib/DashboardServer.ts
benjamincburns: Yeah, that makes sense. Going to resolve this however as it's really just a tangent to this review.

packages/dashboard/webpack.config.ts
benjamincburns: Thanks heaps - that'll make my life a lot easier for the remainder of the typescript support/bundling stuff!

packages/dashboard/src/utils/dash.ts
```diff
@@ -59,14 +70,74 @@ export async function confirmMessage(
...
 }
 
+export async function decodeMessage(
+  lifecycle: ReceivedMessageLifecycle<DashboardProviderMessage>,
+  decoder: ProjectDecoder
+) {
+  const method = lifecycle.message.payload.method as DecodableRpcMethod;
+
+  switch (method) {
+    case "eth_sendTransaction": {
```
benjamincburns, line 88: Something to consider for future work - do we want to decode/display pre-signed messages (`eth_sendRawTransaction`)? Or maybe we even want to prevent them from getting sent through and tell the user to use the browser wallet for better key management?

packages/dashboard/webpack.config.ts
```diff
@@ -0,0 +1,103 @@
...
+    alias: {
+      src: path.resolve("src")
+    },
+    fallback: {
+      crypto: require.resolve("crypto-browserify"),
+      os: require.resolve("os-browserify/browser"),
+      path: require.resolve("path-browserify"),
+      stream: require.resolve("stream-browserify"),
+      vm: require.resolve("vm-browserify")
+    }
```
benjamincburns, line 57: Do we actually _need_ any of these packages in the dashboard frontend? If not, maybe we should use `false` instead of `require.resolve(polyfillPackageName)`. That would help reduce the weight of the dependency tree.

packages/dashboard/webpack.config.ts
```diff
@@ -0,0 +1,103 @@
...
+        errors: true,
+        warnings: false
+      }
+    }
+  },
+  stats: "minimal",
+  module: {
+    rules: [
+      {
+        test: [/\.jsx?$/, /\.tsx?$/],
```
benjamincburns, line 75: IIRC it's substantially more efficient if you only pass `\.tsx?` through `ts-loader`, though this project doesn't have many (any?) pure-JS files anyway, so maybe it's a non-issue.

packages/dashboard/src/contexts/DashContext/DashProvider.tsx
```diff
@@ -23,46 +36,198 @@ function DashProvider({ children }: DashProviderProps): JSX.Element {
...
+        stateRef.current.decoderCompilationHashes
+      );
+
+      // Iterate over incoming compilations and determine if they are useful to
+      // in-memory decoder or db.
+      // Do not try to optimize by:
+      // - Batch inserting.
+      // - Mirroring some kind of db state.
+      for (const compilation of compilations) {
+        const hash = sha1(compilation);
```
benjamincburns, line 102: How do we know whether we've chosen the correct version of the workflow compile result for the transaction in question? I see that everything is keyed on this sha1 hash of the entire `WorkflowCompileResult`, but that won't allow for matching to a specific version of a given contract, will it?

A better approach might be to store off this hash, but also an index of compliations that's keyed by the hash of the deployed contract bytecode. That would let you verify if you're decoding against the correct/deployed version by hashing the result of `eth_getCode` and checking for it in the database.

This would also allow for less arbitrary cache invalidation strategies, as you could treat the database as an LRU cache and only keep around the most recently used N versions of each contract, where N is ideally some user-configurable value.

packages/dashboard/src/contexts/DashContext/types/Schema.ts
```diff
@@ -0,0 +1,16 @@
...
+    value: {
+      dataHash: string;
+      data: Compilation;
+      timeAdded: number;
+    };
+    indexes: {
+      TimeAdded: number;
+    };
+  };
+}
```
benjamincburns, line 16: Have you considered how we'll update this schema in the future? For example, let's assume that we decide to release this with the compilation result matching and cache invalidation logic as-is, but then later on we want to come back in the future and improve it. How will we detect the version change? Will we migrate users forward, or will we just wipe their data and ask them to recompile the contract versions that they care about?

I think at a minimum you might want to consider adding a schema version field here so that we can easily detect what version users are on, report on the schema version via error messages, analytics, etc if necessary, etc. It'll also serve as a hint to future devs that they need to consider existing installs when updating the schema.

benjamincburns: That said, a lazy version detection and data migration strategy is totally valid in a lot of cases, so if you already considered that and want to opt to go that way, that's totally fine by me.

benjamincburns (approved):
Approved, but with a few open questions. Please read through my new comments prior to merging, as you may want to make further updates first.

Also when you do opt to merge, I'd very much be a fan of selectively or even completely squashing some of the 79 commits in this PR. I know some people aren't fans of that practice, but I have to wonder if all 79 commits build successfully. If you're confident that they do, feel free to ignore me. However if you think they might not - please bear in mind that commits that don't build properly break `git bisect`, which is a very powerful debugging tool.

packages/dashboard-message-bus-common/lib/messages.ts
```diff
@@ -11,6 +9,20 @@ export interface Response {
...
+ * Message to inform subscribers of Truffle CLI event data.
+ * The message payload label helps identify the type of event data.
+ */
+export interface CliEventMessage extends Message {
+  type: CliEventMessageType;
+  payload: {
+    label: string;
+    data: any;
+  };
+}
```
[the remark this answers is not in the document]
cliffoo: Thanks @benjamincburns. I'll definitely dig into other message brokers (especially MQTT) in the near future as dashboard's messaging requirements grow. Though I do wonder if adding another dashboard package is better than squishing the types into `@truffle/dashboard-message-bus-common`

> The message bus common package is meant to house logic and types that are common between the message bus service and the message bus client

Something like `@truffle/event-messages` would serve the purpose of providing types to all things message bus, which falls into the domain of the common package?

Ah and about using generics here: I considered it but since the entire message lifecycle for `CliEventMessage` isn't useful at the moment, (this may very well change; right now it's just `abandon`ed after publish, so only `message.payload.data` is useful), I opted to simply pass the payload data to a function (`handleWorkflowCompileResult`) whose params are typed.

packages/dashboard-message-bus-common/package.json
```diff
@@ -28,6 +28,7 @@
   "publishConfig": {
     "access": "public"
   },
+  "dependencies": {},
```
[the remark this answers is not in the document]
cliffoo: Done 👍 Thanks for catching that.

packages/dashboard/src/utils/dash.ts
cliffoo: From a cursory search I'm unable to verify that MM is open to broadcasting an `eth_sendRawTransaction` request, but I will note this down, thanks.

Maybe we can do both! Display it and show users a notification about wallet safekeeping.

packages/dashboard/webpack.config.ts
cliffoo: I believe so. The graph goes `dashboard -> decoder -> web3-utils`. The specifics of how `web3-utils` uses node core modules I'm not sure, but I imagine setting these to `false` would render it unpredictable / unworkable.

packages/dashboard/src/contexts/DashContext/DashProvider.tsx
cliffoo: Yea so the db essentially maps compilation hash to compilation + timestamp.

I see two things here:

> What if workflow-compile changes?

At the moment it's only dependent on the shape of `Compilation`, so it that changes, it's probably time to migrate db.

> The current schema means we can't match, hence decode specific version of a given contract?

So the reason for storing compilations is that one compilation is the unit; you cannot break it into e.g. sources, contracts, etc. and reconstruct it at a later time. That being said, all compilations are fed to one decoder, which means if you write contract A, compile, modify A to be A', compile, the decoder will use compilation for A' to decode.

One caveat as noted by [this comment](https://github.com/trufflesuite/truffle/pull/5646#issuecomment-1294246853) is that in some cases the bytecodes are the same despite having different contracts (like a change in param name), in which case with this week's release the decoder will prefer the "latest" compilation.

cliffoo:
Thanks @benjamincburns! I'm fairly confident that each commit builds successfully, but I'm totally fine with squashing also.

packages/dashboard/src/contexts/DashContext/types/Schema.ts
cliffoo: Yes the schema is [versioned](https://github.com/trufflesuite/truffle/blob/35a9980fa901c2d1add466f0927541ae44d8abed/packages/dashboard/src/contexts/DashContext/state.ts#L18). And `idb` provides a nice way to run an `upgrade` function when the code specifies a version higher than that in the browser, to which you can add transactions to. I think ideally migrations in the future preserve as much compilation (and other) data as possible, and wipe only as the last resort if we cannot guarantee integrity.

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
```diff
@@ -0,0 +1,49 @@
...
+    <>
+      <Group spacing={6}>
+        <Text size="sm" color="teal" weight={700}>
+          Decoded parameters
+        </Text>
+        {!decodingSucceeded && (
+          <AlertTriangle size={16} className={classes.alert} />
+        )}
+      </Group>
+      <Prism language={language} copyLabel="Copy to clipboard">
```
gnidan, line 42: Couple things:

- Is the coloring we get here worth including all of Prism?
- Note we use Highlightjs in other places because @haltman-at is the maintainer of highlightjs-solidity, so we've kind of coalesced around that highlighting system instead of Prism.
- What's the value of copying this decoded representation to the clipboard? Won't this go away when we have codec-components?

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
```diff
@@ -0,0 +1,49 @@
...
+import type { DecodableRpcMethod } from "src/utils/constants";
+
+const useStyles = createStyles((theme, _params, _getRef) => ({
+  alert: {
+    color: theme.colors.orange[6]
+  }
+}));
+
+type DecodedParamsProps = {
+  decodingInspected: string;
```
gnidan, line 13: Should we pass actual Decoding objects instead of `string`? Like, seems like this should probably be a [`Codec.CalldataDecoding`](https://trufflesuite.com/docs/truffle/codec/modules/_truffle_codec.html#calldatadecoding), and then we just convert to string as-needed.

Seems like this would prevent things from breaking in the future. Strings are dumb.

packages/dashboard/src/utils/constants.ts
```diff
@@ -3,32 +3,43 @@ import chainIDtoNameJson from "src/assets/chainIDtoName.json";
 export const EMOTION_KEY = "trfl";
 export const COLOR_SCHEME_KEY = "trfl.dash.color-scheme";
 
-export const INTERACTIVE_RPC_METHODS_ARR = [
-  "eth_signTransaction",
+export const decodableRpcMethodsArr = [
```
gnidan, line 6: Ooh maybe your Prism method "language" detection thing should use these constants instead of hardcoding the methods as string literals

packages/dashboard/src/utils/dash.ts
```diff
@@ -1,21 +1,32 @@
...
+} from "src/utils/constants";
+import type {
+  DecodableRpcMethod,
+  UnsupportedRpcMethod,
   knownChainID
 } from "src/utils/constants";
-import type { UNSUPPORTED_RPC_METHOD } from "src/utils/constants";
+
+export function messageIsDecodable(message: DashboardProviderMessage) {
+  return (decodableRpcMethods as Set<string>).has(message.payload.method);
```
gnidan, line 21: You really need the coercion here?

packages/events/defaultSubscribers/dashboard.js
```diff
@@ -39,6 +39,20 @@ module.exports = {
         }
       }
     ],
+    "compile:succeed": [
+      async function ({ result }) {
+        try {
+          const publishLifecycle = await this.messageBusClient.publish({
+            type: "cli-event",
```
gnidan, line 46: What's the value of a generic `cli-event`, vs. making a dedicated workflow-compile-result message type? Do we really want to shoehorn a bunch of different events like this? I know Ben was talking about making things generic to avoid the use of `any`, but we could also... just make message types for message types we expect.

gnidan (commented):
Some comments/questions for you. Thanks for shepherding this with so much care!

packages/dashboard/package.json
```diff
@@ -39,42 +37,62 @@
...
```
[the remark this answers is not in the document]
cliffoo: > for the shims

Nah just using the types for now.

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
cliffoo: Ah prism was added back in #5241, I wasn't aware that we were using highlightjs somewhere else (and our involvement with it). I picked it because it's provided by mantine (`@mantine/prism`) so it's in-line with other components in terms of props it accepts and styling.

Yes this will leave when codec-components come. With the button you can click once and send decodings to your friends :) jk this `copyLabel` prop just changes what shows on hover, the copy button is there by default.

I've noted this down as follow-up work! Makes sense to align with the rest of the monorepo and switch prism out.

packages/dashboard/src/utils/constants.ts
cliffoo: It's [using the type of these constants](https://github.com/trufflesuite/truffle/blob/68d9d5fcc1014a6aa5da99372373479139ec54b7/packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx#L15) (`type DecodableRpcMethod` is just `typeof decodableRpcMethodsArr[number]`). 

I figured `method === "eth_sendTransaction"` is more understandable than `method === decodableRpcMethodsArr[0]`.

packages/events/defaultSubscribers/dashboard.js
cliffoo: I'm torn, Ben's most convincing comment imo is [this one](https://github.com/trufflesuite/truffle/pull/5621#discussion_r1020108728). Maybe we can reconsider this one when dashboard starts receiving CLI events of more types?

As an aside, I just thought about the plugin system and it will probably require a generic message type similar to this one.

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
cliffoo: Hm I figured it's nice to `import inspect from "object-inspect"` once, `inspect` once and pass the string around.

Like we won't do anything with the decoding other than `inspect`ing it before codec-components, and codec-components will replace much of this stuff anyway (it consumes `Codec.CalldataDecoding` and that's it, no need to pass that off to other components). Is this lazy?

packages/events/defaultSubscribers/dashboard.js
benjamincburns: FWIW, my biggest (albeit not my only) concern about the specific message type was that it was causing us to take a dependency on `workflow-compile` from `dashboard-message-bus-common`.

My thought there is that over time as the scope of the dashboard grows, it's likely going to require data from a bunch of different packages. My fear is that if we establish a pattern of coupling generic/core/common packages to specific/functional packages we're likely going to run into a bunch of obnoxious problems (e.g. circular dependencies) down the line. Admittedly in this case the logic wasn't tightly coupled, but in my view that's more reason to avoid taking the direct dependency.

As a secondary concern, I think it's also usually a bad sign for the maintainability of an architecture when "transport" systems need to be made aware of the detailed structure of the data that it's passing. Ideally we'd have a single envelope type for all messages, there would be some topic key in that type, and the bus would just route messages to relevant subscribers based on that topic key. The payload field of that message envelope would also ideally be typed `unknown` so that we'd get compiler errors if anything w/i the message bus logic tried to interact with it directly.

packages/dashboard-message-bus-common/lib/messages.ts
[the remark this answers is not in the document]
benjamincburns: > Something like @truffle/event-messages would serve the purpose of providing types to all things message bus, which falls into the domain of the common package?

My thinking in the case of a hyptothetical `@truffle/event-messages` was that the only consumers of that package would be event message producers or consumers. `@truffle/dashboard-message-bus-common` is really only meant to contain code that's common to both the message bus client and the message bus service itself, and my thought was that it should only contain logic and/or types that is _strictly required_ for the proper functioning of both. In my ideal world the only direct consumers of `@truffle/dashboard-message-bus-common` would be `@truffle/dashboard-message-bus` and `@truffle/dashboard-message-bus-client`.

Honestly it's probably a mistake on my part to have created the common package without moving the message payload types elsewhere.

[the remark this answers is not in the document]
benjamincburns: > Ah and about using generics here: I considered it but since the entire message lifecycle for CliEventMessage isn't useful at the moment, (this may very well change; right now it's just abandoned after publish, so only message.payload.data is useful), I opted to simply pass the payload data to a function (handleWorkflowCompileResult) whose params are typed.

You may want to consider changing the type of `data` from `any` to `unknown`, as that will force an explicit cast. I don't feel strongly about that preference, however. As a pattern it make the code that consumes the `unknown`-typed field more explicit and readable, but in this specific case I don't think that it's likely to add a ton of extra value over `any`.

packages/dashboard/package.json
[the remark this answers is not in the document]
gnidan: ah so then maybe it can be a devDep!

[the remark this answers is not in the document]
eggplantzzz: @gnidan I thought your opinion on these types (which are exposed through TS magic) was that they should really be regular dependencies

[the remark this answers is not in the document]
gnidan: Wait, I assumed that dashboard *didn't* expose these... are they exposed?

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
gnidan: Don't know about lazy, but it definitely seems to make sense to pass around the raw thing rather than the thing that's missing information

packages/dashboard/package.json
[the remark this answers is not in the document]
cliffoo: Nope it's all lost at compile time. It's used [here](https://github.com/trufflesuite/truffle/blob/68d9d5fcc1014a6aa5da99372373479139ec54b7/packages/dashboard/src/contexts/DashContext/DashProvider.tsx#L15-L18) and [here](https://github.com/trufflesuite/truffle/blob/68d9d5fcc1014a6aa5da99372373479139ec54b7/packages/dashboard/src/contexts/DashContext/types/Schema.ts#L2).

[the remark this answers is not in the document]
cliffoo: Updated to be devDep!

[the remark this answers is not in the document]
cds-amal: I thought the reason to use deps was to make sure types don't break when the project is added as a dependency of another project via npm/yarn.

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
cliffoo: Updated to pass around `decoding` (typed `CalldataDecoding | string`, `string` because it might be a signature request, not `eth_sendTransaction`)

packages/dashboard-message-bus-common/lib/messages.ts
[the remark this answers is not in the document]
cliffoo: > You may want to consider changing the type of data from any to unknown

Done!

packages/dashboard-message-bus-common/lib/messages.ts
```diff
@@ -66,6 +78,12 @@ export interface InvalidateMessage extends Message {
   payload: number;
 }
 
+export const isCliEventMessage = (
+  message: Message
+): message is CliEventMessage => {
```
gnidan, line 83: Can you make this generic to `T` the same way `CliEventMessage<T>` is?

Even if we have to specify a default value, probably still worth keeping the parameter from being obscured by this type guard.

packages/dashboard-message-bus-common/lib/messages.ts
```diff
@@ -87,6 +105,7 @@ export const isInvalidateMessage = (
 };
 
 export type MessageType =
+  | CliEventMessage
```
gnidan, line 108: :( goodbye generic param

packages/dashboard/src/utils/dash.ts
gnidan: ah yeah that makes sense

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
gnidan: Ah yeah hrm, that's a bit tricky. The `string` here strikes me as a code smell... ideally I wish it were easy to make EIP-712 signatures the responsibility of the codec (cc @haltman-at), but this doesn't really make sense.

Maybe we should wrap `string` in some Dashboard-defined type? That way the type gets a proper name, at least.

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
```diff
@@ -0,0 +1,52 @@
...
+}: DecodedParamsProps): JSX.Element {
+  const decodingInspected = inspectDecoding(decoding);
+  const { classes } = useStyles();
+
+  const language =
+    method === "eth_sendTransaction"
+      ? "javascript"
+      : method === "eth_signTypedData_v3" || method === "eth_signTypedData_v4"
+      ? "json"
+      : "markdown";
```
gnidan, line 33: I guess maybe this should at least be its own function (e.g. `highlightLanguageForRpcMethod`)

packages/dashboard/src/contexts/DashContext/DashProvider.tsx
```diff
@@ -23,46 +37,200 @@ function DashProvider({ children }: DashProviderProps): JSX.Element {
...
+        compilationStore.length
+      );
+      const decoderCompilationHashes: State["decoderCompilationHashes"] =
+        new Set();
+      compilationStore.forEach((row, index) => {
+        decoderCompilations[index] = row.data;
+        decoderCompilationHashes.add(row.dataHash);
+      });
+
+      const decoder = await forProject({
```
gnidan, line 209: Apologies if I missed it, but is this now using the new interface to add compilations to the decoder?

(If it's not already using this interface, it can probably be a separate task if you want, but should at least open an issue to do that switch.)

gnidan (commented):
Couple more comments. I'm getting excited :D

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
cliffoo: Makes sense. I replaced `Calldata | string` with a `Decoding` type defined [here](https://github.com/trufflesuite/truffle/blob/e8618ce9a68eca4240d95d327b531b08c73d86e1/packages/dashboard/src/utils/dash.ts#L134).

packages/dashboard-message-bus-common/lib/messages.ts
cliffoo: Oh should I specify a param here? It defaults to `unknown` so I omitted it.

packages/dashboard/src/components/composed/RPCs/RPC/Details/Expanded/DecodedParams.tsx
cliffoo: Makes sense. [Updated](https://github.com/trufflesuite/truffle/commit/ecc8ea83001946da7e2398853d8941425e75f101)

packages/dashboard-message-bus-common/lib/messages.ts
gnidan: It's unfortunate... you'll note that if you try to specify a param, you'll have to add it to the main `MessageType` type, where it doesn't really belong (since it's not like `T` applies elsewhere)

But basically, I hate generic defaults because you can very quickly lose track of the parameter without realizing it (like here... anything that uses `MessageType` has no way of specifying what kind of CliEvent payload)

packages/dashboard/src/contexts/DashContext/DashProvider.tsx
cliffoo: Yea I'll leave that as a follow-up, we still re-init the decoder on new compilations.
I'll open an issue once this is merged, thanks for the reminder!

packages/dashboard-message-bus-common/lib/messages.ts
gnidan: This comment is not to say we should change this... @benjamincburns makes a good argument for the use of generics in the first place, it's just you end up with stuff like this unless you do a whole bunch of work up front.

(This sort of thing is why I was saying not to bother and instead just keep making new types, but I'm happy to lose that argument if this example is the worst of it.)

cliffoo: Ah I see. Would this be better?

```ts
// Rid default
interface CliEventMessage<T> extends Message { // ...

// Specify param
type MessageType = CliEventMessage<unknown> | // ...
```

gnidan: Up to you if you want to keep the default... if I were you, I'd go either way, but I probably would take care not to rely on the default personally (like do at least "specify param" if not "rid default")

packages/dashboard/package.json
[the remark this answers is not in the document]
gnidan: > I thought the reason to use deps was to make sure types don't break when the project is added as a dependency of another project via npm/yarn.

Not sure what you mean here, but we only need a regular dep if the package runs the dependency's code (trivial case) or if the package exports something that references one of the dependency's types, like as argument or return value for an exported function (case we're discussing)

gnidan (approved):
This one's good for me if it's good for others :D

gnidan:
Oh shoot... one more thing, although I won't say it should be blocking:

The "could not decode" error suggests running `truffle compile --all`. This doesn't make sense for Hardhat users using Truffle Dashboard.

I guess medium term we should only suggest the Truffle command when we know it's a Truffle project... and slightly mediumer term we could just use @truffle/from-hardhat :)

packages/dashboard-message-bus-common/lib/messages.ts
cliffoo: I see, I removed the default param and let `MessageType` specify `unknown`. Update [here](https://github.com/trufflesuite/truffle/pull/5621/commits/822a619edf169dc12874d7fe9ba262dbe3754f47).

I think this isn't very important at the moment: `MessageType` isn't actually used anywhere in the dashboard package, only its extended interfaces are.

Issue #5617, opened 2022-10-14
Tests fail when including multiple Openzeppelin artifacts


ryanc414:
- [ x ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

All Truffle tests fail before they start when including multiple artifacts from Openzeppelin's contracts repo from the migrations via `artifact.require()`.

## Steps to Reproduce

Please see the minimal repro here: https://github.com/ryanc414/truffle-openzeppelin-bug

In my case, including both the `TimelockController` and `ERC20PresetFixedSupply` contracts from `@openzeppelin/contracts@4.7.3` via `artifact.require()` was enough to trigger the bug. Note that neither contract was actually deployed - merely including them in the migration was enough.

Including either contract on its own does not trigger the bug, only the combination of both. They don't have to be included from the same migrations file. I haven't experimented with other contracts from Openzeppelin.

## Expected Behavior

Tests should run and pass.

## Actual Results

```
 npm test

> truffle-openzeppelin-bug@1.0.0 test
> truffle test


Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.


  Contract: Test
    1) "before all" hook: prepare suite for "testcase"


  0 passing (100ms)
  1 failing

  1) Contract: Test
       "before all" hook: prepare suite for "testcase":
     TypeError: Cannot read properties of undefined (reading 'toString')
```

## Environment

* Operating System: MacOS
* Ethereum client: N/A
* Truffle version (`truffle version`):  v5.6.1
* node version (`node --version`): v16.15.1
* npm version (`npm --version`): 8.11.0

cds-amal:
Hi @ryanc414,

Thanks for the reproduction repo! The error happens because the project has no declared Solidity contracts and therefore during the compilation phase, needed before running tests, no sources are compiled and truffle will have no sources to test. Another minor issue is the path to the OZ contracts are incorrect. 

You can apply this patch to the project to address the issues:

1. create a file from the diff below, call it `fix.diff` in the same folder as the reproduction repo.
2. apply the diff: `git apply fix.diff`
3. rerun truffle test: `truffle test`

### patch
```diff
diff --git a/contracts/PlaceHolder.sol b/contracts/PlaceHolder.sol
new file mode 100644
index 0000000..6e95318
--- /dev/null
+++ b/contracts/PlaceHolder.sol
@@ -0,0 +1,8 @@
[...]

PR #5622, opened 2022-10-16
codec: handle undefined index

ryanc414:
## PR description

Bug fix - handle the case when `index` may be undefined in `shimContracts` to avoid an exception being thrown. This may occur under the circumstances described in https://github.com/trufflesuite/truffle/issues/5617

## Documentation

- No doc changes required.

### DIFF
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -166,7 +166,7 @@ export function shimContracts(
       contractObject.primarySourceId = index.toString(); //HACK
     } else {
       //if neither was passed, attempt to determine it from the ast
-      let index: number;
+      let index: number | undefined;
       let needsAdding: boolean;
       if (sourceObject.ast) {
         //note: this works for both Solidity and Vyper
@@ -196,7 +196,7 @@ export function shimContracts(
         debug("else; index: %d", index);
       }
       //whether needed adding or not, set the source ID on the contract object
-      contractObject.primarySourceId = index?.toString(); //HACK
+      contractObject.primarySourceId = index.toString(); //HACK
       debug("(no index unless mentioned)");
     }
 
@@ -398,7 +398,7 @@ function inferLanguage(
 
 function getIndexToAddAt(
   sourceObject: Source,
-  index: number,
+  index: number | undefined,
   sources: Source[],
   unreliableSourceOrder: boolean
 ): { index: number; needsAdding: boolean; unreliableSourceOrder: boolean } {
@@ -411,38 +411,38 @@ function getIndexToAddAt(
     "sources: %o",
     sources.map(source => source.sourcePath)
   );
-  if (
-    sources.every(
-      existingSource =>
-        existingSource.sourcePath !== sourceObject.sourcePath ||
-        (!sourceObject.sourcePath &&
-          !existingSource.sourcePath &&
-          existingSource.source !== sourceObject.source)
-    )
-  ) {
+  const existingIndex = sources.findIndex(
+    existingSource =>
+      existingSource.sourcePath !== sourceObject.sourcePath ||
+      (!sourceObject.sourcePath &&
+        !existingSource.sourcePath &&
+        existingSource.source !== sourceObject.source)
+  );
+  if (existingIndex === -1) {
+    //it's not already there, let's add it
     if (unreliableSourceOrder || index === undefined || index in sources) {
       //if we can't add it at the correct spot, set the
       //unreliable source order flag
       debug("collision!");
       unreliableSourceOrder = true;
     }
-    //otherwise, just leave things alone
     if (unreliableSourceOrder) {
       //in case of unreliable source order, we'll ignore what indices
       //things are *supposed* to have and just append things to the end
       index = sources.length;
     }
+    //otherwise, just leave things alone
     return {
       index,
       needsAdding: true,
       unreliableSourceOrder
     };
   } else {
-    //return index: null indicates don't add this because it's
-    //already present
     debug("already present, not adding");
+    //note that index may be undefined in this case if it was passed
+    //in as undefined
     return {
-      index,
+      index: existingIndex,
       needsAdding: false,
       unreliableSourceOrder
     };
--- a/packages/codec/lib/compilations/utils.ts
+++ b/packages/codec/lib/compilations/utils.ts
@@ -413,7 +413,8 @@ function getIndexToAddAt(
   //contents if not)
   const existingIndex = sources.findIndex(
     existingSource =>
-      existingSource &&
+      existingSource && //findIndex treats absent as undefined, so we need this guard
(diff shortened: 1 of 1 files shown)


eggplantzzz:
Hey @ryanc414, thanks for the proposed change! However, it seems like `getIndexToAdd` will always return an index that is a number. So I'm not sure we need this in here. Thoughts @haltman-at?

ryanc414:
> Hey @ryanc414, thanks for the proposed change! However, it seems like `getIndexToAdd` will always return an index that is a number. So I'm not sure we need this in here. Thoughts @haltman-at?

Hi there @eggplantzzz, no problem!

Yes, I agree that the type signature for `getIndexToAdd` would imply that a number should always be returned for the value of `index`. However my testing shows that practically this is not always the case.

Taking a closer look at the Typescript configuration, it seems that the `strictNullChecks` flag is not enabled in the `tsconfig.json` for the `codec` package - meaning that every type implicitly includes both `null` and `undefined` as possible values. To avoid these kinds of bugs in future I would recommend enabling this flag (as it is already enabled for some other packages in this repo) and fixing up the resulting compiler errors - however such a change would be quite large and beyond the scope of this PR!

Following the calling code through, you can see that if `index` is undefined when passed into `getIndexToAdd` it may be returned unchanged as an undefined value.

Zooming out a bit, the context of the original issue was that this bug was hit when using artifacts from third-party dependcies, where the source code may not be readily available. Therefore, I think that any fix should bear in mind the fact that the source code may not always be available when processing compiled artifacts.

eggplantzzz:
Yep, you're right. I wonder if we should add a `?` on line 194 as well...actually maybe we should turn on the `strictNullChecks` check and tighten the function signature. @haltman-at should be the one to make the call here though.

haltman-at:
Yup, `strictNullChecks` is off (sorry), so this sort of thing can absolutely happen.  It's not obvious to me whether this is the correct fix though.  I'll take a look.  Thanks for raising the issue!

haltman-at (commented):
OK, after looking at this a bit, I've concluded that you're right -- this is, esentially, the correct fix.  While juggling all this, I seem to have dropped the case where `index` is undefined.  Thanks for catching that.  That said, would you be OK with me quickly doing this instead?  I want to while I'm at it update some of the comments that have gotten out of date (and note that this change makes more of them out-of-date!) and also make the style a little more consistent.  Thanks!

ryanc414:
Hi @haltman-at, that's great thanks for taking a look. Sure, I'm happy for you to close this PR and come up with a more robust fix!

haltman-at:
OK, I'll do that then.  Well, not really "more robust" -- it's going to be essentially the same as what you did -- just better explained in the comments. :)

Thanks again for finding this!

haltman-at:
Or actually instead I'm going to add a second commit to this so you can have your commit make it in. :)

haltman-at:
OK, after looking at this some more I actually realize that wasn't the right fix after all!  Rather, the right fix is to have `getIndexToAddAt` always return an index; it was a bug that it could return `undefined`.  Now both cases will always return an index -- if it's not already there, the logic is the same as before; and if it is already there, then it just returns the index that it found.

With this, the test passes.

haltman-at:
Oops!  The `decoder` tests fail though.  Guess I need to take another look.

haltman-at:
...and now `downgrade-test.js` is failing?  Geez, this has turned rather more complicated than I anticipated...

haltman-at:
D'oh, I reversed some checks.  Silly me...

haltman-at:
OK, should have this working now!  Latest commit fixes the bug and passes the decoder tests, hopefully nothing else goes wrong!  Just needs some review. :)

ryanc414:
> OK, should have this working now! Latest commit fixes the bug and passes the decoder tests, hopefully nothing else goes wrong! Just needs some review. :)

Great, still seems to fix the original issue in my testing 👌

packages/codec/lib/compilations/utils.ts
```diff
@@ -398,51 +398,52 @@ function inferLanguage(
...
-  if (
-    sources.every(
-      existingSource =>
-        existingSource.sourcePath !== sourceObject.sourcePath ||
+  //first: is this already there? only add it if it's not.
+  //(we determine this by sourcePath if present, and the actual source
+  //contents if not)
+  const existingIndex = sources.findIndex(
+    existingSource =>
+      existingSource &&
```
cds-amal, line 416: Is it possible for an element of  `sources` to be false-y?

packages/codec/lib/compilations/utils.ts
```diff
@@ -398,51 +398,52 @@ function inferLanguage(
...
     sources.map(source => source.sourcePath)
   );
-  if (
-    sources.every(
-      existingSource =>
-        existingSource.sourcePath !== sourceObject.sourcePath ||
+  //first: is this already there? only add it if it's not.
+  //(we determine this by sourcePath if present, and the actual source
+  //contents if not)
+  const existingIndex = sources.findIndex(
```
eggplantzzz, line 414: I didn't know about `findIndex`

packages/codec/lib/compilations/utils.ts
```diff
@@ -398,51 +398,52 @@ function inferLanguage(
...
       index,
       needsAdding: true,
       unreliableSourceOrder
     };
   } else {
-    //return index: null indicates don't add this because it's
-    //already present
     debug("already present, not adding");
+    //note that index may be undefined in this case if it was passed
+    //in as undefined
```
cds-amal, line 444: Is this true since `index` is now set to `existingIndex` in this clause's return?

cds-amal (commented):
Hey @haltman-at, just have a couple of questions.

packages/codec/lib/compilations/utils.ts
haltman-at: Oops!  Thanks, this is exactly the sort of thing I meant to fix.  Will fix.

packages/codec/lib/compilations/utils.ts
haltman-at: Sort of.  No actual element will be falsy, but the array may be sparse.  Different array functions differ in whether they just skip non-existent elements, or treat them as `undefined`; `findIndex`, it seems, does the latter, which is why I had to add the guard.

cds-amal: That's good to know! It's worth leaving a comment to explain the guard.

haltman-at: OK, I can add that.

haltman-at: Added that!

cds-amal (approved):
Thanks, @haltman-at! Changes LGTM, pending CI, of course. 

@ryanc414, thanks so much for bringing this bug to our attention and getting us started on PR fix! :bow:

PR #5624, opened 2022-10-18
Turn on strict null checks in compile-common

haltman-at:
People thought this was on, but it wasn't!  The `skipLibCheck` option is also turned on to prevent the usual problems with `contract-schema`.

As part of this, I had to make two more changes to get it to compile.  One was eliminating the `return undefined;` at the beginning of `LegacyToNew.forBytecode`, which meant also changing where it was invoked.  Fortunately, this function, though exported, is only presently used in `LegacyToNew.forContract`, so that was simple to update.

(*Arguably* that's a breaking change...?  I'm going to just not worry about it unless someone thinks I should.)

The other change was a nonfunctional change made just to satisfy Typescript.  That's it!

### DIFF
--- a/packages/compile-common/src/shims/LegacyToNew.ts
+++ b/packages/compile-common/src/shims/LegacyToNew.ts
@@ -38,8 +38,10 @@ export function forContract(contract: any): CompiledContract {
     ast,
     abi,
     metadata,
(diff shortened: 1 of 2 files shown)


gnidan (approved):
Assuming this passes :)

PR #5625, opened 2022-10-19
Add POA Core to sourcify fetcher

haltman-at:
Sourcify has added support for POA Core, so this PR adds that.  Note I already had it listed in `networks.ts` even though neither fetcher supported it.  I decided to rename it though from "poa" to "core-poa" as I've never seen it referred to as just "POA", only ever as "POA Core".  So yeah!

...also I added a comment about how Etherscan still supports Arbitrum Rinkeby even if they've hidden it. :P


cds-amal (approved):
Looks good!

PR #5626, opened 2022-10-19
added one comma, removed one comma

daviddoswell:
**Reasoning**

Added comma to make sentence structure flow better. Truffle has 3 distinct value propositions, each requiring their own name space, or separation of concerns.

Removed a comma to make second sentence flow directly into the content. Since we begin with a preposition, and the opening is short, we can go right into what a user gets with Truffle.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@
 
 ---
 
-Truffle is a development environment, testing framework and asset pipeline for Ethereum, aiming to make life as an Ethereum developer easier. With Truffle, you get:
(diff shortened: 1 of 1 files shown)


gnidan (approved):
I am ashamed that this README was missing an Oxford comma. Thank you for this!

haltman-at (approved):

daviddoswell:
> I am ashamed that this README was missing an Oxford comma. Thank you for this!

Happens to the best of us! Thanks for approving.

PR #5631, opened 2022-10-20
Remove redundant scripts and ts-node-dev dependency

haltman-at:
This PR removes redundant scripts from the `db` package; these scripts are no longer needed now that the `truffle db serve` command exists.  It also removes the `ts-node-dev` devDependency, which was only used for these scripts, and was the real reason for this PR, in the hopes that it will help us resolve some dependabot alerts. :)


packages/db/package.json
```diff
@@ -29,10 +29,6 @@
     "clean": "rm -rf ./dist ./types/schema.d.ts",
     "madge": "madge ./src/index.ts",
     "prepare": "yarn build",
-    "start": "ts-node-dev -r tsconfig-paths/register ./bin/server",
-    "start:debug": "DEBUG=pouchdb:api ts-node-dev -r tsconfig-paths/register ./bin/server",
-    "start:drizzle": "ts-node-dev -r tsconfig-paths/register ./bin/server ./test/truffle-projects/drizzle-box",
-    "start:drizzle:debug": "DEBUG=pouchdb:api ts-node-dev -r tsconfig-paths/register ./bin/server ./test/truffle-projects/drizzle-box",
```
cds-amal, line 35: We can lose the `truffle-projects/` folder now.

cds-amal (changes requested):
Thanks for following up on this @haltman-at. Now there's unused test fixtures we can remove.

packages/db/package.json
haltman-at: Oh, I didn't realize, thanks!  @gnidan didn't mention that so I didn't know offhand whether that was used elsewhere.  (I guess I should likely check myself to be sure...)

haltman-at: Yup, looks like you're right!  Made that change.

cds-amal (approved):
Looks good!

PR #5633, opened 2022-10-21
Increase timeout on docker tags test

haltman-at:
CI has been timing out a bunch recently, but this particular test seems to be doing so especially often.  Upping it.


cds-amal (approved):
Ok! LGTM

PR #5637, opened 2022-10-22
update gh-actions to use node 16

cds-amal:
## PR description

This PR bumps our action dependencies to address the following warnings for the `build` and `yarncheck` jobs:

> [yarncheck | build](https://github.com/trufflesuite/truffle/actions/runs/3304923000/jobs/5454669430)
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-node, actions/setup-node, actions/checkout

## Notes
- [x] This PR also ups the timeout of [a unit test](https://github.com/trufflesuite/truffle/blob/gh/actions-bump/packages/core/test/lib/commands/compile.js#L103) because of #5639 

## Documentation

- [x] no documentation updates are required.

### DIFF
--- a/packages/core/test/lib/commands/compile.js
+++ b/packages/core/test/lib/commands/compile.js
@@ -100,7 +100,9 @@ describe("compile", function () {
     });
 
     it("prints a list of docker tags", async function () {
-      this.timeout(20000);
+      this.timeout(
+        60000 * 2  // oh boy!
+      );
       const options = {
         list: "docker"
       };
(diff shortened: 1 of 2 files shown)


packages/core/test/lib/commands/compile.js
```diff
@@ -100,7 +100,9 @@ describe("compile", function () {
     });
 
     it("prints a list of docker tags", async function () {
-      this.timeout(20000);
+      this.timeout(
+        60000 * 2  // oh boy!
+      );
```
cds-amal, line 105: 2 minutes is an egregious limit for a test. OTOH, these tests shouldn't hit the network. I opened #5639 to capture that.

haltman-at (approved):
Thanks for getting this!

Issue #5602, opened 2022-10-06
Cannot add compilation to `ProjectDecoder` instance


cliffoo:
Current workaround is to re-instantiate when new compilations come. It would be nice if I can keep feeding that one instance.

```
truffle -v
5.6.0
```

haltman-at:
Question, which of the three decoder classes is this referring to?  Just the `ProjectDecoder`?  All of them?

cliffoo:
Yea only `ProjectDecoder` for this one. I don't know if the same need will arise for contract and contract instance decoders; if they do I guess I'll open issues separately.

haltman-at:
OK cool, if it's just `ProjectDecoder` I don't think this should be too hard.

dongmingh:
Close the issue since this is fixed in Truffle v5.6.6.

PR #5646, opened 2022-10-26
Allow adding compilations to an existing `ProjectDecoder`

haltman-at:
This PR addresses #5602 by making it possible to add compilations to an already existing `ProjectDecoder`.  (It doesn't do this for `ContractDecoder` or `ContractInstanceDecoder`.)  To do this, it adds two methods to `ProjectDecoder`, namely, `addCompilations` and `addAdditionalProjectInfo`.

The `addCompilations` interface assumes you are adding codec-style compilations.  It also assumes that none of these has an ID that collides with the IDs of any of the compilations it already knows about.  It's on the user to ensure that.

The `addAdditionalProjectInfo` interface is more general and allows also adding compile-common style compilations, or just a list of artifacts.  If you use it to add codec-style compilations, it's still up to you to avoid ID collisions.  If you use compile-common style compilations, or just a list of artifacts, then an ID will be assigned; to prevent collisions, I've given the `ProjectDecoder` a new state variable which serves as a nonce to be incorporated into the ID and which is incremented each time you call `addAdditionalProjectInfo`.  I'm not sure how truly robust this system is, but it shouldn't encounter collisions in normal use at least, I figure.

Anyway, how does `addCompilations` work under the hood?  Well, it's mostly pretty simple.  It takes the new compilations and merges them with the existing ones.  Then it performs the various decoder startup computations on the new compilations, and merges the results with what it already has.  Usually the form of the merge is very simple, just `[...old, ...new]` or `{ ...old, ...new}` (actually those latter cases are done as `Object.assign(old, new)` since we intend to alter things here!).  There is one case where it's slightly more complicated, which is calldata allocations, but it's still pretty simple.

There were two cases, however -- returndata allocations and event allocations -- where correctly performing a merge on the results would have been quite difficult.  So to handle those, instead of running the allocation computations on just the new compilations and then merging the results, it instead reruns the relevant computations on the whole set of both old and new compilations.  Oh well, it'll do I figure.

And that's it!  This PR also adds a quick test of this new capability.

### DIFF
--- a/packages/decoder/lib/decoders.ts
+++ b/packages/decoder/lib/decoders.ts
@@ -41,7 +41,8 @@ import {
   VariableNotFoundError,
   MemberNotFoundError,
   ArrayIndexOutOfBoundsError,
-  NoProviderError
+  NoProviderError,
+  RepeatCompilationIdError
 } from "./errors";
 import { Shims } from "@truffle/compile-common";
 //sorry for the untyped import, but...
@@ -80,6 +81,14 @@ export class ProjectDecoder {
   ) {
     if (!provider) {
       throw new NoProviderError();
+    }
+    //check for repeat compilation IDs
+    for (let i = 0; i < compilations.length; i++) {
+      for (let j = i + 1; j < compilations.length; j++) {
+        if (compilations[i].id === compilations[j].id) {
+          throw new RepeatCompilationIdError(compilations[i].id);
+        }
+      }
     }
     this.providerAdapter = new ProviderAdapter(provider);
     this.compilations = compilations;
@@ -148,6 +157,27 @@ export class ProjectDecoder {
   public async addCompilations(
     compilations: Compilations.Compilation[]
   ): Promise<void> {
+    //first: make sure we're not adding a compilation with an existing ID
+    const existingIds = new Set(
+      this.compilations.map(compilation => compilation.id)
+    );
+    //we use a find() rather than a some() so that we can put the ID in the error
+    const conflictingCompilation = compilations.find(compilation =>
+      existingIds.has(compilation.id)
+    );
+    if (conflictingCompilation !== undefined) {
+      throw new RepeatCompilationIdError(conflictingCompilation.id);
+    }
+    //also: check for repeats among the ones we're adding
+    for (let i = 0; i < compilations.length; i++) {
+      for (let j = i + 1; j < compilations.length; j++) {
+        if (compilations[i].id === compilations[j].id) {
+          throw new RepeatCompilationIdError(compilations[i].id);
+        }
+      }
+    }
+
+    //now: checks are over, start adding stuff
     this.compilations = [...this.compilations, ...compilations];
 
     const {
@@ -239,8 +269,9 @@ export class ProjectDecoder {
    * contracts to be decoded.  This may come in several forms; see the type
    * documentation for more information.  If passing in `{ compilations: ... }`,
    * take care that the compilations have different IDs from others passed in
-   * so far.  If passed in in another form, an ID will be assigned automatically,
-   * which should generally avoid any collisions.
+   * so far, otherwise this will error.  If passed in in another form, an ID
+   * will be assigned automatically, which should generally avoid any
+   * collisions.
    */
   public async addAdditionalProjectInfo(
     projectInfo: Compilations.ProjectInfo
--- a/packages/decoder/lib/decoders.ts
+++ b/packages/decoder/lib/decoders.ts
@@ -167,7 +167,9 @@ export class ProjectDecoder {
 
     const { contexts, deployedContexts, contractsAndContexts, allocationInfo } =
       AbiData.Allocate.Utils.collectAllocationInfo(compilations);
-    Object.assign(this.contexts, contexts);
+    this.contexts = Object.assign(contexts, this.contexts); //HACK: we want to
+    //prioritize new contexts over old contexts, so we do this.  a proper timestamp
+    //system would be better, but this should hopefully do for now.
     Object.assign(this.deployedContexts, deployedContexts);
     this.contractsAndContexts = [
       ...this.contractsAndContexts,
--- a/packages/decoder/lib/decoders.ts
+++ b/packages/decoder/lib/decoders.ts
@@ -68,6 +68,8 @@ export class ProjectDecoder {
 
   private ensSettings: DecoderTypes.EnsSettings;
 
+  private addProjectInfoNonce: number = 0;
+
   /**
    * @protected
    */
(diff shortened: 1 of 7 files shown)


packages/decoder/lib/decoders.ts
```diff
@@ -133,6 +135,122 @@ export class ProjectDecoder {
...
+      compilations
+    );
+
+    Object.assign(this.referenceDeclarations, referenceDeclarations);
+    Object.assign(
+      this.userDefinedTypesByCompilation,
+      userDefinedTypesByCompilation
+    );
+    Object.assign(this.userDefinedTypes, userDefinedTypes);
+
```
cliffoo, line 167: Is there anything to be done with `this.codeCache` and `this.ensSettings`? Or have they nothing to do with adding new compilations?

cliffoo (commented):
YES, thanks @haltman-at ! `addAdditionalProjectInfo` looks perfect for dashboard's needs. I understand your point about merging / recomputing different parts of the decoder state. I don't know enough about codec stuff to know if all the utils functions are called and handled correctly inside `addCompilations`, I guess I'll wait for someone to take a proper look at it 😅

packages/decoder/lib/decoders.ts
haltman-at: No, those should be unaffected.

haltman-at:
OK, so (from talking to @gnidan, although it was apparently @cliffoo who first reported it) it seems that there was a problem where the decoder preferred to match bytecode to older contracts (in terms of the order they were added in) rather than newer ones.  Ideally we'd fix this with some sort of timestamp system, but that would be a bunch of work (in that it would probably require changing a whole bunch of stuff internally).  So as a quick-and-dirty fix, I've just reversed the order of assignment here.  That said, we probably should add some sort of proper timestamp system in the future.  I guess I'll put up an issue for that after this is merged.

packages/decoder/lib/decoders.ts
```diff
@@ -133,6 +135,124 @@ export class ProjectDecoder {
...
+   *
+   * Adds additional compilations to the decoder like [[addCompilations]],
+   * but allows it to be specified in more general forms.
+   *
+   * @param projectInfo Information about the additional compilations or
+   * contracts to be decoded.  This may come in several forms; see the type
+   * documentation for more information.  If passing in `{ compilations: ... }`,
+   * take care that the compilations have different IDs from others passed in
+   * so far.  If passed in in another form, an ID will be assigned automatically,
+   * which should generally avoid any collisions.
```
cds-amal, line 243: This sounds like it would be a difficult thing for users to debug. Is there a way to give feedback that the ID exists? Am I missing something?

cds-amal (commented):
I have a question about the developer experience. What happens in the event a user provides an ID that already exists, will it be non-trivial to determine the cause?

packages/decoder/lib/decoders.ts
haltman-at: OK, I'll add that!

haltman-at: OK, I've added this.  In addition to checking for collisions on adding new ones, I also added checks for repeats among the ones you're adding, and among the ones you start the decoder with.  I also added tests of this.

Although, hm -- I just realized I didn't add a check for repeats when starting *en*coder... do you think that should be added too?

cds-amal: Sure, it makes sense to be thorough with the feedback.

packages/decoder/lib/decoders.ts
```diff
@@ -79,6 +82,14 @@ export class ProjectDecoder {
     if (!provider) {
       throw new NoProviderError();
     }
+    //check for repeat compilation IDs
+    for (let i = 0; i < compilations.length; i++) {
+      for (let j = i + 1; j < compilations.length; j++) {
+        if (compilations[i].id === compilations[j].id) {
```
cds-amal: Will repeats be next to each other? Is it possible for i, and i +2 to share the same id if i+1 does not?

packages/decoder/lib/decoders.ts
```diff
@@ -133,6 +144,146 @@ export class ProjectDecoder {
...
+   * have IDs distinct from those the decoder already has.
+   */
+  public async addCompilations(
+    compilations: Compilations.Compilation[]
+  ): Promise<void> {
+    //first: make sure we're not adding a compilation with an existing ID
+    const existingIds = new Set(
+      this.compilations.map(compilation => compilation.id)
+    );
+    //we use a find() rather than a some() so that we can put the ID in the error
```
cds-amal, line 161: There could be more conflicts. Maybe we share them all in the error?

packages/decoder/lib/decoders.ts
```diff
@@ -133,6 +144,146 @@ export class ProjectDecoder {
...
+    const conflictingCompilation = compilations.find(compilation =>
+      existingIds.has(compilation.id)
+    );
+    if (conflictingCompilation !== undefined) {
+      throw new RepeatCompilationIdError(conflictingCompilation.id);
+    }
+    //also: check for repeats among the ones we're adding
+    for (let i = 0; i < compilations.length; i++) {
+      for (let j = i + 1; j < compilations.length; j++) {
```
cds-amal: Oh, they're expected to be in order! Can we assume the client breaks this guarantee?

cds-amal (commented):
I have a couple of questions.

packages/decoder/lib/decoders.ts
haltman-at: No, there's no reason repeats have to be next to each other.  This should find any repeats; why is being next to each other relevant?

packages/decoder/lib/decoders.ts
haltman-at: They're not expected to be in order.  Why do you say they are?

packages/decoder/lib/decoders.ts
haltman-at: Oh, hm, yeah, that's a thing we could do.  It's more trouble though... do you think that's worth it?

packages/decoder/lib/decoders.ts
cds-amal: Oh, I read that wrong. Please disregard, I have no good excuse there. :/

packages/decoder/lib/decoders.ts
cds-amal: Yes, I think it's best to give full information feedback when we can.

haltman-at: OK, I've updated it so that reporting here is now based around arrays and sets of IDs, instead of just stopping as soon as there's a single conflict.

cds-amal (approved):
Thanks! Looks good to me, pending CI

PR #5647, opened 2022-10-26
Implement new `truffle call` command using @truffle/encoder and @truffle/decoder

sukanyaparashar:
## PR description

This PR adds the `truffle call` command leveraging Truffle's encoding and decoding functionality (including overload resolution). Prior work credit: @gnidan (#5330).

## Testing instructions

To test this new feature, write a smart contract inside a truffle project with a function that returns some value or you can clone this [Test Project with a simple storage contract](https://github.com/sukanyaparashar/truffle-test-project).

1. Run `truffle develop` inside a truffle project.
2. In the REPL, run `migrate` to deploy the contract.
3. In the REPL, run `call <contract_name|contract_address> <function_name>`.
       
    onlyfunction

4.  In the REPL, run `call <contract_name|contract_address> <function_name> <arg>`.
        
    functionwitharg

5. In the REPL, run `call <contract_name|contract_address> <function_name>` that returns multiple values of different types.

    multiplevalues

6.  In the REPL, run `call <contract_name|contract_address> <function_signature> <arg>`. NOTE: ABI-style signatures doesn't allow spaces.
     
    functionsigwitharg

7. In the REPL, run `call <contract_name|contract_address> <function_name_that_reverts>`.
    
    revertmessage

8. In the REPL, run `call <contract_name|contract_address> <function_name_that_panics>`.

    panicmessage

**NOTE:** This can also be tested in a truffle project outside of truffle REPL with the following steps -

1. Run a ganache instance in a separate terminal.
2. Run `truffle migrate`.
3. Run `truffle call <contract_name|contract_address> <function_name>`.

    onlyfunction    

4. Run `truffle call <contract_name|contract_address> <function_name> <arg>`.  
   
    functionwitharg

5. In the REPL, run `truffle call <contract_name|contract_address> <function_name>` that returns multiple values of different types.

    multiplevalues

6. Run `call <contract_name|contract_address> <function_signature> <arg>`. NOTE: Function signature should be in `""` and without spaces.

    functionsigwitharg

7. Run `truffle call <contract_name|contract_address> <function_name_that_reverts>`.

    revertmessage

8. Run `truffle call <contract_name|contract_address> <function_name_that_panics>`.

    panicmessage


packages/provider/index.js
```diff
@@ -30,6 +32,39 @@ module.exports = {
         { keepAlive: false }
       );
     }
+    if (
+      typeof provider.request === "undefined" &&
+      typeof provider.send !== "undefined"
+    ) {
```
gnidan, line 38: Instead of doing this here, I think it might make more sense to do this for `request` near the same place we wrap `send` (https://github.com/trufflesuite/truffle/blob/develop/packages/provider/wrapper.js#L26-L34)

This way all the wrapping logic is cohesive to that `wrapper.js` module.

Unless you think that this is a better place to handle EIP-1193 separately? I'm not sure.

sukanyaparashar: I, kind of agree with you @gnidan. `wrapper.js` might make more sense to handle EIP-1193 wrapping. Any thoughts on this @cds-amal @davidmurdoch @eggplantzzz @lsqproduction ?

packages/provider/index.js
```diff
@@ -4,6 +4,8 @@ const { createInterfaceAdapter } = require("@truffle/interface-adapter");
 const wrapper = require("./wrapper");
 const DEFAULT_NETWORK_CHECK_TIMEOUT = 5000;
 
+let id = 0;
```
davidmurdoch, line 7: I don't think the `id` should be a module singleton. We should probably declare it in the scope in which we add the `.request` method onto the `provider`.

The issue with the singleton is that a) it is only used if we add the `request` method to the `provider` ourselves. b) it makes for potentially non-deterministic JSON-RPC ids if `packages/provider` is ever used multiple times in a single node session. The second point likely isn't an issue with typical Truffle usage today, since Truffle (and the provider package) isn't used as a user library, but it's likely worth thinking about.

packages/provider/index.js
davidmurdoch: I don't have a preference.

packages/provider/index.js
sukanyaparashar: Thanks for the explanation, @davidmurdoch . I agree that the `id` should be declared within the scope of the method.

packages/provider/index.js
sukanyaparashar: Should we make a separate PR to handle the implementation of EIP-1193?

sukanyaparashar: I have moved the EIP-1193 wrapping to `wrapper.js` as suggested by @gnidan. I thought of keeping this implementation in this PR itself, since it is related to `truffle call` command.

gnidan: Oh sorry, I missed your question. I think a separate PR for the provider stuff would be good because it'll likely get more eyes that way.

sukanyaparashar: I will make a separate PR for this. Thanks for clarifying, @gnidan. :)

haltman-at:
I notice your testing instructions only mention `truffle develop`... you tested this outside of `truffle develop` as well, right?

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+      ]
+    });
+  } catch (error) {
+    console.log(`Error Message: ${error.message}\nError Code: ${error.code}`);
+  }
+
+  if (result !== null && result !== undefined) {
+    const [decoding] = await decoder.decodeReturnValue(functionEntry, result);
+
+    for (const { value: result } of decoding.arguments) {
```
haltman-at, line 67: This seemingly doesn't handle different kinds of decodings?  Like, this seems OK for decodings with `kind = "return"`, but there are other kinds as well.  If it's not a kind that has `arguments`, this will error.  Not only that, but if you get back a revert string (on a client that doesn't error on such things), it will print it out as if it were a valid returned result!

One way to simplify this would be to use `Codec.Export.ReturndataDecodingInspector`, since that already handles all these cases.  That said, maybe you don't like how it formats its output. :)  Well -- if that's the case, uh, talk to me and maybe we can come up with something?

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+    } else {
+      const settings = {
+        provider: config.provider,
+        projectInfo: {
+          artifacts: Object.values(contracts)
+        }
+      };
+
+      const contract = contracts[contractNameOrAddress];
+      // Error handling to remind users to run truffle migrate first
```
haltman-at, line 117: The error handling mentioned here appears to be missing??  (This comment is blocking, I'm hitting request changes based on this.)

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+      const instance = await contract.deployed();
+      encoder = await Encoder.forContractInstance(instance, settings);
+      decoder = await Decoder.forContractInstance(instance, settings);
+    }
+    return { encoder, decoder };
```
haltman-at, line 122: This is a really minor stylistic note, but I think this code would be easier to follow if each branch returned separately, rather than having each branch setting `{ encoder, decoder }` and having that returned at the end.  E.g., in the first case, you could just write `return await ...` and never need to explicitly set `encoder` and `decoder` at all. :)

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+    if (contractNameOrAddress.startsWith("0x")) {
+      const projectInfo = {
+        artifacts: Object.values(contracts)
+      };
+
+      ({ encoder, decoder } = await getEncoderDecoderForContractAddress(
+        contractNameOrAddress,
+        projectInfo
+      ));
+    } else {
```
haltman-at, line 108: As long as I'm making stylistic notes -- this code could use more comments!  Like, label the two branches here as the address case vs the name case.  I can see that's what you're doing, sure, but explicit labels are still helpful to the reader.  Alternatively, another way of making this more explicit would be to rename the variable in each case, so e.g. in the first case you could open with writing `const address = contractNameOrAddress;`.  But really I just think writing `//address case` would help a fair bit!

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+      const settings = {
+        provider: config.provider,
+        projectInfo: {
+          artifacts: Object.values(contracts)
+        }
+      };
+
+      const contract = contracts[contractNameOrAddress];
```
haltman-at, line 116: Or like, speaking of comments, a comment here like `//remember, this is a name` would help prevent confusion here!  (That or the variable rename I suggested.)  This is even *if* you mark this branch with `//name case`!

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+      // Error handling to remind users to run truffle migrate first
+      const instance = await contract.deployed();
+      encoder = await Encoder.forContractInstance(instance, settings);
+      decoder = await Decoder.forContractInstance(instance, settings);
+    }
+    return { encoder, decoder };
+  }
+
+  async function sourceFromExternal(contractNameOrAddress, config) {
```
haltman-at, line 125: You had better be in the address case if this is being called, right??  It doesn't seem right that this variable is called `contractNameOrAddress`.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+    const { compileResult } = await fetchAndCompile(
+      contractNameOrAddress,
```
haltman-at, line 127: I think you had better have some validation here on the address before calling `fetchAndCompile`!  (This comment is blocking, hitting request changes based on this.)

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+    const projectInfo = {
+      commonCompilations: compileResult.compilations
+    };
+
+    const { encoder, decoder } = await getEncoderDecoderForContractAddress(
+      contractNameOrAddress,
+      projectInfo
+    );
+
+    return { encoder, decoder };
```
haltman-at, line 140: Really minor style note, but you can tighten this up with just `return await ...`. :)  Why create a variable when you're just about to return it anyway? :)

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  async function getEncoderDecoderForContractAddress(
+    contractNameOrAddress,
```
haltman-at, line 144: If this function is for an address specifically, this variable shouldn't be called `contractNameOrAddress`.  That's quite a misleading name.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+    projectInfo
+  ) {
+    const projectEncoder = await Encoder.forProject({
+      provider: config.provider,
+      projectInfo
+    });
+    const encoder = await projectEncoder.forAddress(
+      contractNameOrAddress,
+      config.blockNumber
```
haltman-at, line 153: (Again, we need validation on blockNumber before use, although I assume this will be handled in a unified manner rather than separately.)

haltman-at (changes requested):
OK, looked this over, the bones are there, but I think there are a number of details that need fixing up, particularly as relates to input validation and error handling.  A few of these are such that I think I need to hit "request changes".  There are also some stylistic improvements I think could be made although that's obviously less important.  More comments throughout though would likely be helpful.

sukanyaparashar:
> I notice your testing instructions only mention `truffle develop`... you tested this outside of `truffle develop` as well, right?

Yes, it works outside the REPL too. I will add that to the testing instructions too. Thanks!

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+  if (functionEntry.stateMutability !== "view") {
+    console.log(
+      `WARNING!!! Not a view function\n  ABI stateMutability: ${functionEntry.stateMutability}`
+    );
+  }
+
+  const provider = new Encoder.ProviderAdapter(config.provider);
+  let result;
+  try {
+    result = await provider.request({
```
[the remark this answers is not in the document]
haltman-at: Btw, another reason to do this is that `request` returns `Promise<any>`.  That's dangerous!  You should use something better-typed than that.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+          data: transaction.data
+        },
+        config.blockNumber
+      ]
+    });
+  } catch (error) {
+    console.log(`Error Message: ${error.message}\nError Code: ${error.code}`);
+  }
+
+  if (result !== null && result !== undefined) {
```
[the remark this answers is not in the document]
haltman-at: Oh, wait -- actually, I have that last part wrong.  On looking at this a second time, there's actually some important error-handling missing here.  This is now a blocking comment, sorry, as there's some important stuff not handled here.

So, first off -- the thing I was originally going to write was possibly correct.  An actual *error* from `request`/`sendRequest`, as it's currently written, should be rare, not something expected; as such, we shouldn't be catching and logging such errors, we should just be bubbling them up -- there shouldn't be a `try`/`catch` at all.

But, we *can* get an error back from the client; it just won't be *thrown*.  Rather, it will consist of the return having a different form than you expect.  So that's a case you need to handle.

Now, `request` currently has no handling for that case, so that's something you'll need to write.  Perhaps the right thing to do is to convert this non-thrown error into a thrown one?  In that case, you will need that `try`/`catch` block after all!  But you'll also need quite a bit more handling.

See, in the case that an error is returned from a call, you're going to have to do some work to extract the raw return value so you can decode it.  It's not enough to just print the error message -- that might be OK for a revert string, sure, but what about a panic code?  Or a custom error?  You're going to need the decoder to handle those, and that means you're going to need to get the raw data.

How do you get that raw data?  Well... unfortunately, that varies a bit.  You can see that we have code in Truffle Contract to do that function [here](https://github.com/trufflesuite/truffle/blob/develop/packages/contract/lib/reason.js#L15) -- it may not be the clearest, feel free to bug me if you don't understand what's going on there.  So you'll likely have to do basically that; ideally that functionality should be factored out into some utility, but that may not be practical here, IDK.  Something to discuss with @gnidan, I imagine.

But this is a case that will need to be handled.

packages/core/lib/commands/call/run.js
haltman-at: Btw, for the particular case of panic codes, you may want to also consider making use of the `panicString` function in `debug-utils` to provide a readable interpretation of the panic code to the user.

packages/core/lib/commands/call/meta.js
```diff
@@ -0,0 +1,82 @@
...
+    "fetch-external": {
+      describe: "Fetch referenced verified contracts as needed",
+      alias: "x",
+      type: "boolean",
+      default: false
+    },
+    "block-number": {
+      describe: "Calls specified contract function at specific block",
+      alias: "b",
+      type: "integer",
```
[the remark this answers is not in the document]
sukanyaparashar: This is wrong, I guess. The type should be a string here. Thanks for catching this @haltman-at .

packages/core/lib/commands/call/meta.js
```diff
@@ -0,0 +1,82 @@
...
+    "contract-name": {
+      describe: "The contract name to be called",
+      type: "string"
+    }
+  },
+  help: {
+    usage:
+      "truffle call <contract-address>|<contract-name> <function-name>|<function-signature>\n" +
+      "                             " + // spacing to align with previous line
+      "[<arg>...] [--fetch-external|-x] [--network <network>|--url <provider_url>]\n" +
```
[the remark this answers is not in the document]
sukanyaparashar: I think your comment makes sense here. Function arguments should not be optional.

packages/core/lib/commands/call/meta.js
```diff
@@ -0,0 +1,82 @@
...
+      },
+      {
+        option: "<function-name>",
+        description:
+          "The function name inside the specified contract to be called."
+      },
+      {
+        option: "<function-signature>",
+        description:
+          "The function signature inside the specified contract to be called."
```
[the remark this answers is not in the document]
sukanyaparashar: Does this mean that people can think of "function signature" as "function selector" which is the hash of the function signature? In that case, explicitly specifying it in the description to avoid confusion makes sense.

packages/core/lib/commands/call/meta.js
```diff
@@ -0,0 +1,82 @@
...
+      },
+      {
+        option: "<arg>...",
+        description: "List of arguments of the function to be called."
+      },
+      {
+        option: "--fetch-external|-x",
+        description: "Fetch referenced verified contracts as needed."
```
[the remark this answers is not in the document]
sukanyaparashar: I will add this in the description!

packages/core/lib/commands/call/meta.js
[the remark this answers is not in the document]
haltman-at: Yes, "signature" is unfortunately sometimes abused to mean "selector".

packages/core/lib/commands/call/meta.js
```diff
@@ -0,0 +1,82 @@
...
+          "                    This can be used outside of a Truffle project."
+      },
+      {
+        option: "--network",
+        description:
+          "The network to connect to, as specified in the Truffle config."
+      },
+      {
+        option: "--block-number|-b",
+        description: "The block number from which the contract is to be called."
```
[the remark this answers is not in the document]
sukanyaparashar: Yeah! The sentence seems wrong. I will change this.

dongmingh (commented):
Could you grant me access to your [test repo](https://github.com/sukanyaparashar/truffle-test-project) that I can test this PR?

sukanyaparashar:
> Could you grant me access to your [test repo](https://github.com/sukanyaparashar/truffle-test-project) that I can test this PR?

I have made the repository public for anyone to access. Thanks @dongmingh for reminding me of that. :)

packages/truffle/test/scenarios/commands/call.js
```diff
@@ -0,0 +1,109 @@
...
+        inputCommands: ["migrate", "call Sample bad"],
+        config,
+        executableCommand: "console",
+        executableArgs: `--network ${networkName}`,
+        displayHost: networkName
+      });
+      const output = logger.contents();
+      const expectedValue = "You are a failure";
+      assert(
+        output.includes(expectedValue),
```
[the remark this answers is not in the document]
sukanyaparashar: Sure, I will use chai's assertion library for the tests.

dongmingh (commented):
Thanks @sukanyaparashar for granting me access to your test repo.

I tested with both inside and outside of REPL with your test repo and the test results are the same as `testing instructions` suggests.

In addition, I tested an un-migrated contract:
```
truffle(develop)> call Sample0 get
TypeError: Cannot read properties of undefined (reading 'deployed')
```
The message is correct.

And an undefined function:
```
truffle(develop)> call Sample get0
NoFunctionByThatNameError: Contract Sample has no function with name or signature get0
```
The message is correct.

And function `get(uint256, uint256)`, 
```
truffle(develop)> call Sample get(uint256,uint256) 10 20
230
```
The result is correct.

If I add a space to separate the function arguments
```
truffle(develop)> call Sample get(uint256, uint256) 10 20
NoFunctionByThatNameError: Contract Sample has no function with name or signature get(uint256,
```

We might want to include this test scenarios to the `testing instructions`.  And consider if we can parse spaces inside the parentheses correctly.

haltman-at:
> We might want to include this test scenarios to the `testing instructions`.  And consider if we can parse spaces inside the parentheses correctly.

To be clear, ABI-style signatures don't include spaces.  Encoder won't handle it with spaces.  While it might be nice to allow spaces in the future as a convenience feature, I think it makes perfect sense to reject such spaces, certainly in an initial version.

gnidan:
> > We might want to include this test scenarios to the `testing instructions`.  And consider if we can parse spaces inside the parentheses correctly.
> 
> To be clear, ABI-style signatures don't include spaces. Encoder won't handle it with spaces. While it might be nice to allow spaces in the future as a convenience feature, I think it makes perfect sense to reject such spaces, certainly in an initial version.

Agree with @haltman-at here, but @dongmingh great suggestion! I had not considered that we might want to allow spaces in the future. Seems like a nice feature!

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+    : await sourceFromLocal(contractNameOrAddress, config);
+
+  ({ abi: functionEntry, tx: transaction } = await encoder.encodeTransaction(
+    functionNameOrSignature,
+    args
+  ));
+
+  if (functionEntry.stateMutability !== "view") {
+    console.log(
+      `WARNING!!! Not a view function\n  ABI stateMutability: ${functionEntry.stateMutability}`
```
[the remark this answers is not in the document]
sukanyaparashar: Yeah! That makes sense. I will add this to the warning.

packages/core/lib/commands/call/run.js
sukanyaparashar: Yeah! I guess removing unnecessary variable declaration makes sense. Thanks @haltman-at .

packages/core/lib/commands/call/run.js
sukanyaparashar: I guess comments are good enough to understand but renaming the variables in each branch will be a better idea.

packages/core/lib/commands/call/run.js
sukanyaparashar: I have missed this error handling. May be catch the error and display it with `TruffleError` is a good idea!

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+  let functionEntry, transaction;
+
+  const { encoder, decoder } = config.fetchExternal
+    ? await sourceFromExternal(contractNameOrAddress, config)
+    : await sourceFromLocal(contractNameOrAddress, config);
+
+  ({ abi: functionEntry, tx: transaction } = await encoder.encodeTransaction(
+    functionNameOrSignature,
+    args
+  ));
```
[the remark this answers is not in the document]
sukanyaparashar: `TruffleError` makes sense here. Should we display a generalized error for function name, function signature and function argument types? Like `The function name, function signature or function arguments you entered are invalid! Please run the command again with valid function name, function signature and function arguments (if any)!`.

sukanyaparashar:
> > > We might want to include this test scenarios to the `testing instructions`.  And consider if we can parse spaces inside the parentheses correctly.
> > 
> > 
> > To be clear, ABI-style signatures don't include spaces. Encoder won't handle it with spaces. While it might be nice to allow spaces in the future as a convenience feature, I think it makes perfect sense to reject such spaces, certainly in an initial version.
> 
> Agree with @haltman-at here, but @dongmingh great suggestion! I had not considered that we might want to allow spaces in the future. Seems like a nice feature!

So for the initial release, should we check for spaces in the function signature string and remove them to convert into ABI-style signatures?

packages/truffle/test/scenarios/commands/call.js
```diff
@@ -0,0 +1,109 @@
...
+    cleanupSandboxDir();
+  });
+
+  beforeEach(async () => {
+    ({ config, cleanupSandboxDir } = await sandbox.create(project));
+    config.network = "development";
+    config.logger = logger;
+  });
+
+  describe("when runs with basic contract", () => {
```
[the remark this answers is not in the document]
sukanyaparashar: I will add the tests for invalid arguments.

packages/core/lib/commands/call/run.js
sukanyaparashar: I believe the validation you are mentioning on the `address` means to check if the address is 40 character hex string with "0x" prefix?

haltman-at:
> So for the initial release, should we check for spaces in the function signature string and remove them to convert into ABI-style signatures without spaces?

I don't think so, no.  Note that in order to even get a signature string with spaces, the user would have to put that argument in quotes on the command line!

gnidan:
Yeah if we decide we want to support spaces later, we should try to do it entirely within the encoder itself

packages/core/lib/commands/call/run.js
haltman-at: More than that.  Addresses should be validated for a correct checksum (unless the user wants to bypass that additional layer of validation).  The usual way to validate an address is to use the `isAddress` function from web3-utils.  (I assume there's an equivalent function in ethers, too; I don't know that library as well.)  This function will check that the input both has the correct form for an address, and that its checksum... unless the address is all lowercase or all uppercase, in which case the checksum valiation is bypassed; it's understood that in these cases the user is deliberately bypassing checksum validation.  That's generally how you want to validate addresses.

sukanyaparashar:
> > So for the initial release, should we check for spaces in the function signature string and remove them to convert into ABI-style signatures without spaces?
> 
> I don't think so, no. Note that in order to even get a signature string with spaces, the user would have to put that argument in quotes on the command line!

Yeah! I agree with you @haltman-at . So, if the user enters spaces between the arguments, then should we throw a `TruffleError` saying that `ABI-style function signature doesn't allow spaces`?

haltman-at:
> Yeah! I agree with you @haltman-at . So, if the user enters spaces between the arguments, then should we throw a `TruffleError` saying that `ABI-style function signature doesn't allow spaces`?

I think that is substantially overcomplicating things.  For an initial version, if they enter spaces, and get a function not found error as a result, it should be the same function not found error they'd get for any other way they could mess up the signature.  I don't see any particular need for a special spaces-based error.

(Again, remember that normally you can't even *detect* the user entering spaces; you'll only *actually* get spaces if they also use quotes.)

packages/core/lib/commands/call/run.js
haltman-at: Actually, hm -- if it weren't for the necessity of accepting contract *names*, I'd say that arguably address validation should be done not with `isAddress` but rather by putting these addresses into the options (and using `allowOptions`) during encoding!  Then encoder will do the validation for you. :)  That way is probably also more extensible in the future.  And actually I guess doing it that way is perfectly viable if you introduce a separate check for contract names.

I think doing things that way does have some pitfalls, though, so you may want to talk to me if you decide to take that route.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+  await Environment.detect(config);
+
+  const [contractNameOrAddress, functionNameOrSignature, ...args] = config._;
+  let functionEntry, transaction;
+
+  const { encoder, decoder } = config.fetchExternal
+    ? await sourceFromExternal(contractNameOrAddress, config)
+    : await sourceFromLocal(contractNameOrAddress, config);
+
+  ({ abi: functionEntry, tx: transaction } = await encoder.encodeTransaction(
```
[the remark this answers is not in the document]
sukanyaparashar: Thanks for mentioning this @haltman-at . I didn't have much knowledge about this. This is really cool that encoder handles these things too. :)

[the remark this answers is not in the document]
haltman-at: Ah yeah sorry I guess you weren't following the initial discussions between me and @gnidan about `truffle call`.  The thing is that Truffle Encoder was originally meant to support Truffle Contract, and so I made it with that in mind -- the idea that what would be passed in would be JS objects, rather than necessarily strings.  This led to two problems when considering Truffle Call:

1. There was no way to specify complex types like arrays or structs; thus the `allowJson` option had to be added
2. For compatibility with ethers, *any* input was allowed as a boolean.  On the command line there's no need for that and it would only cause confusion, thus, `strictBooleans`.

Based on the discussion of addresses below, possibly I might want to add another new option for recognizing contract names... well, we can discuss this.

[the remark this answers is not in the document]
sukanyaparashar: Thanks for explaining this, @haltman-at.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,167 @@
...
+  try {
+    result = await provider.request({
+      method: "eth_call",
+      params: [
+        {
+          from: config.from,
+          to: transaction.to,
+          data: transaction.data
+        },
+        config.blockNumber
```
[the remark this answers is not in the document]
sukanyaparashar: @haltman-at, what are the cases should we consider for a block number validation?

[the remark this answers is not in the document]
haltman-at: Well -- according to my understanding -- it should either be a number (presumably converted from a numeric string), or else it should be one of the special values (`"latest"`, etc), right?

packages/core/lib/commands/call/run.js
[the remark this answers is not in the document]
sukanyaparashar: `call` can be added as an adapter method but why is that more preferable than a raw request?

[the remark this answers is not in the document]
haltman-at: Because that's the point of the adapter!  The whole point of the adapter is to contain reusable methods that other components can -- in contexts where for whatever reason we don't want to use web3 or ethers -- use to perform RPC methods without having to muck about with raw requests.  Using the adapter just so you can perform a request yourself, rather than calling a convenient method to have it construct and perform the request for you, defeats the point of it!

Alternatively, rather than use the adapter at all, maybe it's acceptable to just use web3 here.  I assumed that you used the adapter because after talking to @gnidan you concluded there was some reason not to use web3 here.  But I'm not actually seeing such a reason offhand?  So likely you could just use web3 and spare yourself some trouble.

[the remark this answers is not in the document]
sukanyaparashar: I gave it a thought and I think it might be a better idea to add `call` as an adapter method so that it can be reused. I am going to try that. Thanks @haltman-at .

packages/core/lib/commands/call/run.js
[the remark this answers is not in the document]
sukanyaparashar: When implementing the 1193 provider type [here](https://github.com/trufflesuite/truffle/blob/37db8eb29b936443eea59a8feeabfdaf246c6eec/packages/provider/wrapper.js#L11), the error handling has been done when the call response returns an error. With a `try/catch` block, I am able to get the raw data with `error.data` in the `catch` block. I guess decoding that raw data (as you mentioned [can be done like this](https://github.com/trufflesuite/truffle/blob/develop/packages/contract/lib/reason.js#L15)) will solve the cases related with revert strings, panic codes and custom errors.

dongmingh (commented):
I tested with both contract name and contract address and they work as the `test instructions` says.  Then I tested with `-b` option and see this:
```
$ truffle call Sample get -b 0
The contract self-destructed.
dongminghwang@1200116  ~/work/trufflesuite/test/reprod/2023/01/17/implement-truffle-call/truffle-test-project ‹node-16.18.1›  ‹› (main) 
$ truffle call Sample get -b 1
Returned values: (
  200 (type: uint256)
)
dongminghwang@1200116  ~/work/trufflesuite/test/reprod/2023/01/17/implement-truffle-call/truffle-test-project ‹node-16.18.1›  ‹› (main) 
$ truffle call Sample get -b 2
Returned values: (
  200 (type: uint256)
)
dongminghwang@1200116  ~/work/trufflesuite/test/reprod/2023/01/17/implement-truffle-call/truffle-test-project ‹node-16.18.1›  ‹› (main) 
$ truffle call Sample get -b 3
TypeError: Cannot read properties of undefined (reading 'toTwos')
    at Object.toBytes (/Users/dongminghwang/work/trufflesuite/truffle/packages/codec/lib/conversion.ts:195:17)
    at ContractDecoder.<anonymous> (/Users/dongminghwang/work/trufflesuite/truffle/packages/decoder/lib/decoders.ts:1327:30)
    at fulfilled (/Users/dongminghwang/work/trufflesuite/truffle/packages/decoder/dist/decoders.js:28:58)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
Truffle (unbundled) (core: 5.7.0)
Node v16.18.1
```
The last one might be caused by the contract is undefined in block 3.

sukanyaparashar:
> I tested with both contract name and contract address and they work as the `test instructions` says. Then I tested with `-b` option and see this:
> 
> ```
> $ truffle call Sample get -b 0
> The contract self-destructed.
> dongminghwang@1200116  ~/work/trufflesuite/test/reprod/2023/01/17/implement-truffle-call/truffle-test-project ‹node-16.18.1›  ‹› (main) 
> $ truffle call Sample get -b 1
> Returned values: (
>   200 (type: uint256)
> )
> dongminghwang@1200116  ~/work/trufflesuite/test/reprod/2023/01/17/implement-truffle-call/truffle-test-project ‹node-16.18.1›  ‹› (main) 
> $ truffle call Sample get -b 2
> Returned values: (
>   200 (type: uint256)
> )
> dongminghwang@1200116  ~/work/trufflesuite/test/reprod/2023/01/17/implement-truffle-call/truffle-test-project ‹node-16.18.1›  ‹› (main) 
> $ truffle call Sample get -b 3
> TypeError: Cannot read properties of undefined (reading 'toTwos')
>     at Object.toBytes (/Users/dongminghwang/work/trufflesuite/truffle/packages/codec/lib/conversion.ts:195:17)
>     at ContractDecoder.<anonymous> (/Users/dongminghwang/work/trufflesuite/truffle/packages/decoder/lib/decoders.ts:1327:30)
>     at Generator.next (<anonymous>)
>     at fulfilled (/Users/dongminghwang/work/trufflesuite/truffle/packages/decoder/dist/decoders.js:28:58)
>     at processTicksAndRejections (node:internal/process/task_queues:96:5)
> Truffle (unbundled) (core: 5.7.0)
> Node v16.18.1
> ```
> 
> The last one might be caused by the contract is undefined in block 3.

Yes, @dongmingh. The last error occurred because there is no block number 3. I don't know if we need to handle that and throw a `TruffleError` instead.

dongmingh:
It will be nice to display a message like "contract is undefined in the block."

sukanyaparashar:
Please let me know if I should make a new PR for this work to be completed :)

haltman-at:
> It will be nice to display a message like "contract is undefined in the block."

That seems to me like a sufficiently rare error case that we don't need to worry about it.  Making sure the block *exists*, however, is possibly the sort of thing we might want to handle during block number validation.  That said, it doesn't seem like a big priority compared to everything else here.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,218 @@
...
+  const provider = new Encoder.ProviderAdapter(config.provider);
+  let decoding;
+
+  // Get the first defined "from" address
+  const fromAddress = getFirstDefinedValue(
+    options.from,
+    config.networks[config.network].from,
+    config.from
+  );
+  checkValidAddress(fromAddress);
```
haltman-at, line 77: Why is validation performed on `from` but not `to`?  I think I know why, but that at least deserves a comment.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,218 @@
...
+    [decoding] = await decoder.decodeReturnValue(functionEntry, result, {
+      status: true
+    });
+  } catch (error) {
+    [decoding] = await decoder.decodeReturnValue(functionEntry, error.data, {
+      status: false
+    });
+
+    // Gives a readable interpretation of the panic code returned from the call
+    if (decoding.abi.name === "Panic") {
```
haltman-at, line 96: This should be outside of the catch block, seems to me.  Also, why does this error case result in a `TruffleError`, but other error cases don't?

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,218 @@
...
+  config.logger.log(
+    util.inspect(new Codec.Export.ReturndataDecodingInspector(decoding), {
+      colors: true,
+      depth: null,
+      maxArrayLength: null,
+      breakLength: 79
+    })
+  );
+
+  async function sourceFromLocal(contractNameOrAddress, config) {
```
haltman-at, line 113: Why is this being declared as an inner function?  I find that confusing; it disrupts reading the code, because the reader doesn't know whether there might be actual things happening afterward.  I'd suggest moving it outside.  Similarly with the others.

packages/core/lib/commands/call/run.js
haltman-at: Also, I think `checkValidAddress` would be clearer if inlined... it's such a minor thing.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,218 @@
...
+  let config = loadConfig(options);
+  await Environment.detect(config);
+
+  const [contractNameOrAddress, functionNameOrSignature, ...args] = config._;
+  let functionEntry, transaction;
+
+  // Validates the block number and returns the hex string representation of a number or
+  // an error if the string is other than "latest", "pending" and "earliest" or
+  // a hex string representation of a number
+  const blockToFetch = formatBlockSpecifier(config.blockNumber);
```
haltman-at, line 38: Well, it *throws* an error if it's not formatted right, right?  But such an error is an expected error, so it should probably get turned into a `TruffleError`.  Maybe that's not so important, idk.

packages/encoder/lib/adapter.ts
```diff
@@ -172,7 +175,9 @@ export class ProviderAdapter {
             if (error) {
               return reject(error);
             }
-
+            if (response.error) {
```
haltman-at, line 178: What is this for? Is this to handle the various errors that different clients might throw?  I think you've missed some in that case.

packages/core/lib/commands/call/run.js
haltman-at: Hm, this still needs error handling for the case where they didn't enter a valid contract name.  Right now you're going to get an unhelpful error message in that case -- it'll tell you the contract wasn't deployed, when in fact it didn't have a valid name in the first place.

haltman-at (commented):
It's not entirely clear what's going to happen to this PR -- are you going to finish it? Am I going to take it over? -- so it's not entirely clear if there's any point in me rereviewing it, but Nick (IIRC) was of the opinion that I should, so, I've rereviewed it.

Anyway, this improves a lot!  However I still think a fair bit was missed with the error handling, and that will have to be fixed.

There's also the question of whether we want to include the potential encoder changes we discussed earlier; I guess coordinating on that is more difficult now. :-/

gnidan:
> are you going to finish it? Am I going to take it over?

@sukanyaparashar / @haltman-at, seems that it makes sense for Harry to take this over, since we'd like to get this done and recognize that Sukanya, you have other priorities right now. Of course we'll keep your credit, since Harry's just going to be bringing it over the finish line.

Thanks all!

sukanyaparashar:
> > are you going to finish it? Am I going to take it over?
> 
> @sukanyaparashar / @haltman-at, seems that it makes sense for Harry to take this over, since we'd like to get this done and recognize that Sukanya, you have other priorities right now. Of course we'll keep your credit, since Harry's just going to be bringing it over the finish line.
> 
> Thanks all!

That absolutely makes sense, @gnidan @haltman-at . Thank you for letting me know and would love to see this feature out soon! :)

packages/core/lib/commands/call/run.js
haltman-at: Oh, I see -- it's an inner function so it can make use of the imports.  Blech.  Well,  I guess there's no avoiding that then.  I'm adding a comment though to make this clearer.

haltman-at:
OK, I've made a bunch of improvements to the functionality.  However, I still need to make a pass over the tests, as well as to perform manual testing.

haltman-at:
Oops, I also need to just straight-up update the tests, it's failing CI because I didn't...

haltman-at:
OK, I have substantially revised this and I think it should be OK now!

I had to leave out a few tests due to the framework making them difficult.

Also sorry for how slow the tests are.  We can cut some if you like.  Or if you know a better way to do it that would be quick to implement...

gnidan:
> Also sorry for how slow the tests are. We can cut some if you like.

Yeah, let's cut all the slow tests for now and open an issue about the lack of automated testing as a result.

packages/core/lib/commands/call/run.js
```diff
@@ -0,0 +1,317 @@
...
+
+  let config = loadConfig(options);
+  await Environment.detect(config);
+
+  const [contractNameOrAddress, functionNameOrSignature, ...args] = config._;
+  let functionEntry, transaction;
+
+  const fromAddress =
+    options.from ?? config.networks[config.network]?.from ?? config.from;
+  if (!web3Utils.isAddress(fromAddress)) {
```
gnidan, line 36: Can we skip the need to specify a `from` field?

At the very least this error should indicate which address is the problem, since this error now suggests that it might be the contract address that's wrong

gnidan: But I'm thinking... maybe we can just use `0x0` as the `from` address

gnidan (changes requested):
Looks good apart from the requirement for `from`. Let's use the zero-address for `from` instead of erroring.

gnidan (approved):
Looks good, thanks!

Issue #5323, opened 2022-07-19
Debugger CLI should have event log


haltman-at:
## Issue

Now that #3824 is done and events are added to the txlog, there's still the task of adding an event log to the CLI (when you press `e`, say).

## Environment

* Truffle version (`truffle version`): 5.5.22

lsqproduction:
released in truffle [v5.6.4](https://github.com/trufflesuite/truffle/releases/tag/v5.6.4)

PR #5651, opened 2022-10-27
Add event log to debugger CLI

haltman-at:
Addresses #5323.  This PR adds an event log to the debugger CLI.  It's accessed with the new `e` command.  By default it shows the 3 most recent events.  (Not sure if this is a good default -- anyone have opinions?)  But you can put a number (or `all`) after the `e` to show more or fewer; this setting is then remembered while the debugger stays open.

There's not a lot to say about this PR; the hard work was done a while back.  This just adds a CLI interface.  As mentioned we print the N most recent events; note that this does include reverted events.  For each event we print the decoding (assuming we can decode it) and the address that emitted it.  If it was emitted from a delegatecall, we print both the address of the code that did the emitting, and the address of the contract it was emitted on behalf of.  And if it was reverted, we print "Reverted event:" before it in yellow.

I didn't include any means to print the raw event info.  Maybe that can come later?

There a few other things perhaps worth changing in here.  The decodings are printed in color, but if there's no arguments, there's not a lot of color.  Perhaps the emitting addresses should be colorized?

Also, it would be nice if the addresses had their contract classes beside them.  That might require going back and actually altering the `flattedEvents` selector that this relies on, but it might be worth doing.  Thoughts?  (Edit: Actually, I think I can do it without that, so maybe I should just do this...)

Regardless, I figured I'd put this up in this form and see what people think.

While I was at it, I simplified the non-repeatable-command check.  That was really getting out of hand.

Because this is a debugger CLI PR, I just did manual testing.

### DIFF
--- a/packages/core/lib/debug/interpreter.js
+++ b/packages/core/lib/debug/interpreter.js
@@ -619,6 +619,22 @@ class DebugInterpreter {
           await this.printer.printReturnValue();
         }
         break;
+      case "e":
+        if (cmdArgs) {
+          const eventsCount = parseInt(cmdArgs);
+          if (!isNaN(eventsCount) && eventsCount > 0) {
+            this.printer.eventsCount = eventsCount;
+          } else if (cmdArgs === "all") {
+            this.printer.eventsCount = Infinity;
+          } else {
+            this.printer.print(
+              'Invalid event count given, must be positive integer or "all"'
+            );
+            break;
+          }
+        }
+        this.printer.printEvents();
+        break;
       case ":":
         watchExpressionAnalytics(cmdArgs);
         this.printer.evalAndPrintExpression(cmdArgs);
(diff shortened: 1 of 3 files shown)


cds-amal (dismissed):
Thanks for this! It's nice to be able to observe events. <3

cds-amal (changes requested):
Thanks for explaining @haltman-at. I'm requesting changes because I agree with your suggestion for the contract Name to be displayed with the emitting address.

haltman-at:
OK @cds-amal, I made that change.  Names of contracts are now listed (when available), not just addresses.  Also, the addresses are now colorized.  I also split up the "event emitted by A on behalf of B" message over two lines.

cds-amal (approved):
Looks good! Manually tested, this is exciting!

PR #5655, opened 2022-10-28
Update ganache to 7.5.0

haltman-at:
Since it seems Dependabot isn't handling this like @davidmurdoch tried to set it up to do, I've put up this manual PR. :shrug:


cds-amal (approved):
Thanks for getting to this quickly! Approved, pending CI, of course.

PR #5657, opened 2022-10-28
Instrument commandRunner

cds-amal:
## PR description

I've used this technique to debug on CI and it's a pain to hack it up every time. This is an effort to save myself some typing, and share the technique. This PR enables setting the `DEBUG` environment when using `commandRunner`. This would be useful to enable and capture truffle `DEBUG` output in CI to examine `commandRunner`'s `stdout` and `stderr`

<details><summary>Expand to see sample output</summary>

```console
truffle:     truffle unbox [ @standalone ]
truffle:       when run without arguments
truffle: 	---truffle commandRunner stdout--- |	CommandRunner
truffle: 	---truffle commandRunner stdout--- |	execString: node /home/runner/work/truffle/truffle/packages/truffle/build/cli.bundled.js unbox --force
truffle: 	---truffle commandRunner stderr--- |	2022-10-28T16:02:50.758Z unbox {
truffle: 	---truffle commandRunner stderr--- |	  in: 'git@github.com:trufflesuite/truffle-init-default',
truffle: 	---truffle commandRunner stderr--- |	  out: 'https://github.com:trufflesuite/truffle-init-default'
truffle: 	---truffle commandRunner stderr--- |	}
truffle: 	---truffle commandRunner stdout--- |	
truffle: 	---truffle commandRunner stdout--- |	Starting unbox...
truffle: 	---truffle commandRunner stdout--- |	=================
truffle: 	---truffle commandRunner stdout--- |	
truffle: 	---truffle commandRunner stderr--- |	2022-10-28T16:02:50.765Z follow-redirects options {
truffle: 	---truffle commandRunner stderr--- |	  maxRedirects: 50,
truffle: 	---truffle commandRunner stderr--- |	  maxBodyLength: 10485760,
truffle: 	---truffle commandRunner stderr--- |	  protocol: 'https:',
truffle: 	---truffle commandRunner stderr--- |	  path: '/trufflesuite/truffle-init-default/master/truffle-box.json',
truffle: 	---truffle commandRunner stderr--- |	  method: 'HEAD',
truffle: 	---truffle commandRunner stderr--- |	  headers: {
truffle: 	---truffle commandRunner stderr--- |	    Accept: 'application/json, text/plain, */*',
truffle: 	---truffle commandRunner stderr--- |	    'User-Agent': 'axios/0.27.2'
truffle: 	---truffle commandRunner stderr--- |	  },
truffle: 	---truffle commandRunner stderr--- |	  agent: undefined,
[...]

PR #5658, opened 2022-10-28
Test verbose runner

cds-amal:
test command runner


cds-amal:
Successfully tested #5657

PR #5661, opened 2022-10-28
Simplify/fix when to check for designated invalid function

haltman-at:
I was going over this code with @cds-amal and realized it could be simplified.  I noticed that when the index is not `-1`, but we don't have an AST, we didn't check for the designated invalid function.  I was like, wait, why don't we?  I should add that check.  And then I realized these could be combined -- if the index is `-1`, then `ast` will necessarily be undefined, so checking if the index is `-1` really doesn't need to be separate.  So, now it's simplified a bit.

I only did manual testing of this since it requires old Solidity versions to hit this code.  I tested it on 0.5.0, it worked fine.


cds-amal (approved):
Makes sense.

PR #5664, opened 2022-10-31
Add Arbitrum Nova to Etherscan fetcher

haltman-at:
Etherscan has added Arbitrum Nova, so I've added it to the fetcher.

Etherscan has also added the Aptos network, along with its testnet and devnet; but as best I can tell those don't presently allow contract verification (assuming this is even a concept that applies to them), so I haven't added those.


gnidan (approved):

cds-amal (approved):
Looks good.

PR #5317, opened 2022-07-18
Remove ethpm integration test

cds-amal:
This test was marked as skipped, so I decided to remove it since #5283 is merged.

PR #5665, opened 2022-11-01
Remove ethpm-v1 concerns

cds-amal:
## PR description

Remove `ethpm` logic since Truffle no longer supports it (see: #5283 and #5317)

Packages modified
- compile-solidity-test
- config
- resolver
- truffle

## Documentation
- [x] no documentation updates are required.


lsqproduction (approved):
Nice catch. thank you for removing these dead codes.

sukanyaparashar (approved):
Thanks for getting rid of these @cds-amal.

eggplantzzz (approved):
Yeah this all looks cool to me. The only thing is that we should probably verify that `config.ethpm` values aren't used by anything. I searched and didn't see anything that was using it so we're probably good!

PR #5668, opened 2022-11-02
remove db from dev deps in truffle

eggplantzzz:
This PR removes @truffle/db from the `devDependencies` in truffle. It is already in `optionalDependencies`.


lsqproduction (approved):
i like

cds-amal (approved):
Looks good to me

PR #5669, opened 2022-11-02
Update apollo-server to ^3.11.0

haltman-at:
We got a dependabot alert about this... except dependabot won't generate an update because it says it's no longer vulnerable?  I was going to just dismiss the alert, but since in this case it seems easy to just upgrade, I figured I'd just go ahead and do that.


cds-amal (approved):
Makes sense, tests pass, and manually verified `truffle db serve` works.

While reviewing, I noticed they switched to a monorepo and scoped npm packages. I created an issue #5670 so it doesn't get lost.

PR #5684, opened 2022-11-09
Running truffle debug with appropriate flag (--vscode)

xhulz:
## PR description

This PR allows the user to debug a transaction using VSCode.

## Documentation

### DIFF
--- a/packages/core/lib/commands/debug/run.js
+++ b/packages/core/lib/commands/debug/run.js
@@ -67,7 +67,7 @@ module.exports = async function (options) {
     compilations
   });
 
-  // Checks if the user wants to debug in vscode
+  // Checks if the user wants to open the debugger in vscode
   if (config.vscode) {
     await cliDebugger.openVSCodeDebug();
     return;
--- a/packages/core/lib/commands/debug/run.js
+++ b/packages/core/lib/commands/debug/run.js
@@ -60,9 +60,17 @@ module.exports = async function (options) {
   if (config.compileAll && config.compileNone) {
     throw new Error("Incompatible options passed regarding what to compile");
   }
-  const interpreter = await new CLIDebugger(config, {
+
+  const cliDebugger = new CLIDebugger(config, {
     txHash,
     compilations
-  }).run();
+  });
+
+  if (config.vscode) {
+    await cliDebugger.openVSCodeDebug();
+    return;
+  }
+
+  const interpreter = await cliDebugger.run();
   return await promisify(interpreter.start.bind(interpreter))();
 };
--- a/packages/core/lib/commands/debug/run.js
+++ b/packages/core/lib/commands/debug/run.js
@@ -61,16 +61,19 @@ module.exports = async function (options) {
     throw new Error("Incompatible options passed regarding what to compile");
   }
 
+  // Create a new CLIDebugger instance
   const cliDebugger = new CLIDebugger(config, {
     txHash,
     compilations
   });
 
+  // Checks if the user wants to debug in vscode
   if (config.vscode) {
     await cliDebugger.openVSCodeDebug();
     return;
   }
 
+  // Starts the cli debugger
   const interpreter = await cliDebugger.run();
   return await promisify(interpreter.start.bind(interpreter))();
 };
--- a/packages/core/lib/debug/cli.js
+++ b/packages/core/lib/debug/cli.js
@@ -207,16 +207,17 @@ class CLIDebugger {
   }
 
   async openVSCodeDebug() {
-    // Sets the parameters
-    const params = new URLSearchParams({
-      txHash: this.txHash,
-      workingDirectory: this.config.working_directory,
-      providerUrl: this.config.provider.host
-    });
+    // Sets the URL
+    const url = new URL("/debug", "vscode://trufflesuite-csi.truffle-vscode");
+
+    // Sets the query parameters
+    url.searchParams.set("txHash", this.txHash);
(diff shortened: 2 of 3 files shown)


packages/core/lib/commands/debug/run.js
```diff
@@ -60,9 +60,20 @@ module.exports = async function (options) {
...
+  // Create a new CLIDebugger instance
+  const cliDebugger = new CLIDebugger(config, {
     txHash,
     compilations
-  }).run();
+  });
+
+  // Checks if the user wants to open the debugger in vscode
+  if (config.vscode) {
+    await cliDebugger.openVSCodeDebug();
```
gnidan, line 72: Shouldn't this method not be part of the thing called the "CLI" debugger? Seems better to keep the CLIDebugger just for the CLI and push this VS Code logic elsewhere.

xhulz: Hi @gnidan, Thank you for that

what place do you suggest?

xhulz: @kevinbluer for visibility

kevinbluer: Thanks for the heads up @xhulz! Also sounds like we've got a way forward for this from our earlier conversation 👍

packages/core/lib/debug/cli.js
```diff
@@ -204,6 +205,42 @@ class CLIDebugger {
...
+   * This function is responsible for opening the debugger in vscode.
+   */
+  async openVSCodeDebug() {
+    // Sets the URL
+    const url = new URL("/debug", "vscode://trufflesuite-csi.truffle-vscode");
+
+    // Sets the query parameters
+    url.searchParams.set("txHash", this.txHash);
+    url.searchParams.set("workingDirectory", this.config.working_directory);
+    url.searchParams.set("providerUrl", this.config.provider.host);
```
gnidan, line 219: Note that this is brittle, since we can't always rely on being able to get a URL from a provider

packages/core/lib/commands/debug/run.js
gnidan: Hm, maybe just have a `VSCodeDebugger` class?

xhulz: Hey @gnidan @kevinbluer 

I just moved all the vscode debug function to a new class called `VSCodeDebugger`.

Thank you all again

packages/core/lib/debug/vscode.js
```diff
@@ -0,0 +1,73 @@
...
+
+    // Sets the query parameters
+    url.searchParams.set("txHash", this.txHash);
+    url.searchParams.set("workingDirectory", this.config.working_directory);
+    url.searchParams.set("providerUrl", this.getProviderUrl());
+    url.searchParams.set("network", this.config.network);
+    url.searchParams.set("disableFetchExternal", disableFetchExternal);
+
+    // Opens VSCode based on OS
+    const openCommand = process.platform === "win32" ? `start ""` : `open`;
```
acuarica, line 26: Does `open` also works in Linux?

packages/core/lib/debug/vscode.js
```diff
@@ -0,0 +1,73 @@
...
+    // Checks if there is a provider in the config
+    if (this.config.provider) {
+      return this.config.provider.host;
+    }
+
+    // Creates the provider URL from host and port
+    if (this.config.network_config) {
+      return new URL(
+        `http://${this.config.network_config.host}:${this.config.network_config.port}`
+      ).toString();
```
acuarica, line 66: Is it worth to do `new URL(string).toString()`? In other words, what's the issue with returning just `http://[...]`?

packages/core/lib/debug/vscode.js
```diff
@@ -0,0 +1,73 @@
...
+class VSCodeDebugger {
+  constructor(config, txHash) {
+    this.config = config;
+    this.txHash = txHash;
+  }
+
+  /**
+   * This function is responsible for opening the debugger in vscode.
+   */
+  async run() {
```
acuarica, line 12: Given the purpose of this method, I would name it `open` instead of `run`.

packages/core/lib/debug/vscode.js
```diff
@@ -0,0 +1,73 @@
+const childProcess = require("child_process");
+
+class VSCodeDebugger {
```
acuarica, line 3: Just a small detail, the name `VSCodeDebugger` might be a bit misleading. What about `VSCodeOpenDebugger` or `VSCodeDebugAdapter`?

acuarica (approved):
lgtm, although please keep in mind that my review is mostly about style rather than correctness, since I'm not that familiar with Truffle codebase.

xhulz:
Hey @acuarica, thank you for you review!

Your suggestions make sense. I'll wait for the other reviews to implement them all at once.

Thank you!

packages/core/lib/debug/vscode.js
cds-amal: IIRC, `win32: start`, `osX: open`, and *nix is `xdg-open or similar resource openers mimeopen, mimeo, linopen, handlr etc`.

cds-amal (commented):
Partial review, I didn't get a chance to run this yet.

eggplantzzz:
Sorry for neglecting this for so long, I didn't know that you were waiting on a review for me!

It looks like there is some kind of problem when importing `truffle/console`. I get this when executing debug with `--vscode`:
Screenshot 2023-01-17 at 3 16 47 PM

Issue #5282, opened 2022-07-12
console.sol


davidmurdoch:
- [x] trufflesuite/ganache#3327
- [x] trufflesuite/truffle#5193
- [x] trufflesuite/truffle#5195
- [x] trufflesuite/trufflesuite.com#1267

Issue #5195, opened 2022-06-15
Truffle doesn't hook into Ganache's console.log event to handle log output


davidmurdoch:
The current event name is "ganache:vm:tx:console.log", but this will likely change before it is released in ganache. See also: #5193

haltman-at:
Released in 5.7.0.

Issue #5193, opened 2022-06-15
Truffle's resolver doesn't hook into ganache's console.sol


davidmurdoch:
Which isn't truffle's fault, since Ganache hasn't yet published it yet. But it will Soon:tm:!

cliffoo:
To add some clarifications: This can be closed when `@truffle/console.sol` can be imported inside solidity contracts.

haltman-at:
Released in 5.7.0.

PR #5687, opened 2022-11-09
Implement console.log in Truffle

cds-amal:
## PR description

Integrates [`@ganache/console.sol`](https://github.com/trufflesuite/ganache/tree/develop/src/chains/ethereum/console.log) Smart contract logging into Truffle's test and develop workflow. 

Note: Ganache already provides console logging, but it isn't integrated into Truffle (if you start a ganache instance and use truffle to interact with your smart contract on the ganache instance, logging output will be in the Ganache's TTY)

This is part of the effort outlined in #5282 and should address: #5193 and #5195

<details><summary>Click to see sample output using truffle test</summary>

Note every line prefixed with a `:` is a console.log output.

```console
$ truffle test
Using network 'test'.


Compiling your contracts...
===========================
> Everything is up to date, there is nothing to compile.
  : CSTR::Simple
  :     secret: 13n


  Contract: Simple
    Works with migrations deployment
      ✔ Should have a deployed contract
      ✔ has the secret
  : The passphrase is: 13n
  : Leak it a hundred times, plus one more: 1313n
      ✔ has the derived Secret


    Works with new deployment
  : CSTR::Simple
  :     secret: 551n
      ✔ Should have a deployed contract
      ✔ has the secret
  : The passphrase is: 551n
  : Leak it a hundred times, plus one more: 55651n
      ✔ has the derived Secret


  6 passing (139ms)

```

</details>

### How to use in a project

- Import `truffle/Console.sol` in your contract and use its interface. See [Ganache docs for usage details](https://github.com/trufflesuite/ganache/tree/develop/src/chains/ethereum/console.log)
  ```solidity
   pragma solidity >=0.4.25 <0.9.0;
   import "truffle/Console.sol";
   
   ...
   // in some function...
   console.log("hello msg.sender: %o", msg.sender);
   ```
- Configuration: There are 2 options for configuration, both set in `truffle-config.js`. 
  1. `solidityLog.displayPrefix` sets the prefix for every console.log line. The
     default is `""`; however, it is a nice option when you want to differentiate
     console.log output from extra verbose text output.
  2. `solidityLog.preventConsoleLogMigration` a safety measure to disable accidentally
     migrating contracts that use console.log to MAINNET. The default is `false`,
     you would have to opt in, setting to `true`, should you need it. 

## Testing instructions

This is more of a smoke test scenario. Try adding console.log in a contract, or if you prefer, [try one of these](https://github.com/cds-amal/console-log-tests). Does logging work? Is it intuitive? Do reverts log how you expect? Please try to break it. This feature only works when using `truffle develop` or `truffle test`. 

- To test `preventConsoleLogMigration` you will need to run the project against a network that simulates Mainnet. This can be done by starting ganache w/ `ganache  --chain.networkId 1`

  The linked projects define a `mainfork` network in their `truffle-config.js`,
  so specify it in the migrate command (`truffle migrate --network mainfork`).

  |mainnet | preventConsoleLogMigration| expectation|
  |---|---|---|
  |false |false| it should deploy|
  |false |true| it should deploy |
  |true |false| it should deploy|
  |true |true| it should NOT deploy, and print a helpful, actionable message.|
  
  1. Start ganache w/ `ganache --chain.networkId 1` (simulate mainnet)
  2. Set `preventConsoleLogMigration` appropriately. test true/false on mainnet and not mainnet
  2. `truffle migrate --network mainfork` for testing on mainnet, if using one the linked projects.

## Documentation

### DIFF
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -30,7 +30,7 @@ export const getInitialConfig = ({
     verboseRpc: false,
     solidityLog: {
       displayPrefix: "",
-      disableMigration: false
+      disableMigrate: false
     },
     gas: null,
     gasPrice: null,
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -30,7 +30,7 @@ export const getInitialConfig = ({
     verboseRpc: false,
     solidityLog: {
       displayPrefix: "",
-      disableMigration: true
+      disableMigration: false
     },
     gas: null,
     gasPrice: null,
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -28,6 +28,9 @@ export const getInitialConfig = ({
       }
     },
     verboseRpc: false,
+    solidityLog: {
+      prefix: ""
+    },
     gas: null,
     gasPrice: null,
     maxFeePerGas: null,
@@ -93,6 +96,7 @@ export const configProps = ({
     network() {},
     networks() {},
     verboseRpc() {},
+    solidityLog() {},
     build() {},
     resolver() {},
     artifactor() {},
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -29,7 +29,8 @@ export const getInitialConfig = ({
     },
     verboseRpc: false,
     solidityLog: {
-      prefix: ""
+      displayPrefix: "",
+      disableMigration: true
     },
     gas: null,
     gasPrice: null,
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -30,7 +30,7 @@ export const getInitialConfig = ({
     verboseRpc: false,
     solidityLog: {
       displayPrefix: "",
-      disableMigrate: false
+      preventConsoleLogMigration: false
     },
     gas: null,
     gasPrice: null,
--- a/packages/config/test/unit.test.ts
+++ b/packages/config/test/unit.test.ts
@@ -12,7 +12,7 @@ describe("TruffleConfig unit tests", async () => {
 
     it("solidityLog", () => {
       const expectedSolidityLog = {
-        displyPrefix: "",
+        displayPrefix: "",
         disableMigration: false
       };
       assert.deepStrictEqual(truffleConfig.solidityLog, expectedSolidityLog);
--- a/packages/config/test/unit.test.ts
+++ b/packages/config/test/unit.test.ts
@@ -3,52 +3,65 @@ import TruffleConfig from "../dist";
 import { describe, it } from "mocha";
 
 describe("TruffleConfig unit tests", async () => {
-  describe("with", async () => {
-    let truffleConfig: TruffleConfig;
+  let truffleConfig: TruffleConfig;
 
-    beforeEach(() => {
+  describe("Defaults", async () => {
+    before(() => {
       truffleConfig = TruffleConfig.default();
     });
 
-    it("a simple object", async () => {
-      const expectedRandom = 42;
-      const expectedFoo = "bar";
-      const obj = {
-        random: expectedRandom,
-        foo: expectedFoo
+    it("solidityLog", () => {
+      const expectedSolidityLog = {
+        displyPrefix: "",
+        disableMigration: false
       };
-      const newConfig = truffleConfig.with(obj);
-      assert.equal(expectedRandom, newConfig.random);
-      assert.equal(expectedFoo, newConfig.foo);
+      assert.deepStrictEqual(truffleConfig.solidityLog, expectedSolidityLog);
     });
-
-    it("overwrites a known property", () => {
-      const expectedProvider = { a: "propertyA", b: "propertyB" };
-      const newConfig = truffleConfig.with({ provider: expectedProvider });
-      assert.deepEqual(expectedProvider, newConfig.provider);
-    });
-
-    it("ignores properties that throw", () => {
-      const expectedSurvivor = "BatMan";
-      const minefield = { who: expectedSurvivor };
-
-      const hits = ["boom", "pow", "crash", "zonk"];
-      hits.forEach(hit => {
-        Object.defineProperty(minefield, hit, {
-          get() {
-            throw new Error("BOOM!");
-          },
-          enumerable: true //must be enumerable
-        });
+  }),
+    describe("with", async () => {
+      beforeEach(() => {
+        truffleConfig = TruffleConfig.default();
       });
 
-      const newConfig = truffleConfig.with(minefield);
+      it("a simple object", async () => {
+        const expectedRandom = 42;
+        const expectedFoo = "bar";
+        const obj = {
+          random: expectedRandom,
+          foo: expectedFoo
+        };
+        const newConfig = truffleConfig.with(obj);
+        assert.strictEqual(expectedRandom, newConfig.random);
+        assert.strictEqual(expectedFoo, newConfig.foo);
+      });
 
-      //one survivor
-      assert.equal(expectedSurvivor, newConfig.who);
+      it("overwrites a known property", () => {
+        const expectedProvider = { a: "propertyA", b: "propertyB" };
+        const newConfig = truffleConfig.with({ provider: expectedProvider });
+        assert.deepStrictEqual(expectedProvider, newConfig.provider);
+      });
 
-      //these jokers shouldn't be included
-      hits.forEach(hit => assert.equal(undefined, newConfig[hit]));
+      it("ignores properties that throw", () => {
+        const expectedSurvivor = "BatMan";
+        const minefield = { who: expectedSurvivor };
+
+        const hits = ["boom", "pow", "crash", "zonk"];
+        hits.forEach(hit => {
+          Object.defineProperty(minefield, hit, {
+            get() {
+              throw new Error("BOOM!");
+            },
+            enumerable: true //must be enumerable
+          });
+        });
+
+        const newConfig = truffleConfig.with(minefield);
+
+        //one survivor
+        assert.strictEqual(expectedSurvivor, newConfig.who);
+
+        //these jokers shouldn't be included
+        hits.forEach(hit => assert.strictEqual(undefined, newConfig[hit]));
+      });
     });
-  });
 });
--- a/packages/config/test/unit.test.ts
+++ b/packages/config/test/unit.test.ts
@@ -13,7 +13,7 @@ describe("TruffleConfig unit tests", async () => {
     it("solidityLog", () => {
       const expectedSolidityLog = {
         displayPrefix: "",
-        disableMigration: false
+        disableMigrate: false
       };
       assert.deepStrictEqual(truffleConfig.solidityLog, expectedSolidityLog);
     });
--- a/packages/config/test/unit.test.ts
+++ b/packages/config/test/unit.test.ts
@@ -13,7 +13,7 @@ describe("TruffleConfig unit tests", async () => {
     it("solidityLog", () => {
       const expectedSolidityLog = {
         displayPrefix: "",
-        disableMigrate: false
+        preventConsoleLogMigration: false
       };
       assert.deepStrictEqual(truffleConfig.solidityLog, expectedSolidityLog);
     });
--- a/packages/contract/lib/contract/index.js
+++ b/packages/contract/lib/contract/index.js
@@ -5,6 +5,7 @@ const execute = require("../execute");
 const bootstrap = require("./bootstrap");
 const constructorMethods = require("./constructorMethods");
 const properties = require("./properties");
+const util = require("util");
 
 // For browserified version. If browserify gave us an empty version,
 // look for the one provided by the user.
@@ -155,6 +156,10 @@ if (typeof Web3 === "object" && Object.keys(Web3).length === 0) {
   }
 
   Contract._constructorMethods = constructorMethods(Contract);
+
+  Contract.prototype[util.inspect.custom] = function (/* _depth , _options */) {
+    return `Contract at: ${this.address}`;
+  };
 
   // Getter functions are scoped to Contract object.
   Contract._properties = properties;
--- a/packages/contract/lib/contract/index.js
+++ b/packages/contract/lib/contract/index.js
@@ -5,7 +5,6 @@ const execute = require("../execute");
 const bootstrap = require("./bootstrap");
 const constructorMethods = require("./constructorMethods");
 const properties = require("./properties");
-const util = require("util");
 
 // For browserified version. If browserify gave us an empty version,
 // look for the one provided by the user.
@@ -157,7 +156,7 @@ if (typeof Web3 === "object" && Object.keys(Web3).length === 0) {
 
   Contract._constructorMethods = constructorMethods(Contract);
 
-  Contract.prototype[util.inspect.custom] = function (/* _depth , _options */) {
+  Contract.prototype[Symbol.for("nodejs.util.inspect.custom")] = function () {
     return `Contract at: ${this.address}`;
   };
 
--- a/packages/contract/lib/contract/index.js
+++ b/packages/contract/lib/contract/index.js
@@ -156,10 +156,6 @@ if (typeof Web3 === "object" && Object.keys(Web3).length === 0) {
 
   Contract._constructorMethods = constructorMethods(Contract);
 
-  Contract.prototype[Symbol.for("nodejs.util.inspect.custom")] = function () {
-    return `Contract at: ${this.address}`;
-  };
-
   // Getter functions are scoped to Contract object.
   Contract._properties = properties;
 
--- a/packages/core/lib/commands/migrate/runMigrations.js
+++ b/packages/core/lib/commands/migrate/runMigrations.js
@@ -71,7 +71,7 @@ module.exports = async function (config) {
 
       debugLog(`${dirtyArtifacts.length} consoleLog artifacts detected`);
       debugLog(
-        `config.solidityLog.disableMigrate: ${config.solidityLog.disableMigrate}`
+        `config.solidityLog.preventConsoleLogMigration: ${config.solidityLog.preventConsoleLogMigration}`
       );
 
       if (dirtyArtifacts.length) {
@@ -81,11 +81,11 @@ module.exports = async function (config) {
         console.warn(dirtyArtifacts.join(", "));
         console.warn();
 
-        if (config.solidityLog.disableMigrate) {
+        if (config.solidityLog.preventConsoleLogMigration) {
           throw new TruffleError(
             "You are trying to deploy contracts that use console.log." +
               os.EOL +
-              "Please fix, or disable this check by setting solidityLog.disableMigrate to false" +
+              "Please fix, or disable this check by setting solidityLog.preventConsoleLogMigration to false" +
               os.EOL
           );
         }
--- a/packages/core/lib/commands/migrate/runMigrations.js
+++ b/packages/core/lib/commands/migrate/runMigrations.js
@@ -41,44 +41,46 @@ async function findArtifactsThatUseConsoleLog(buildDir) {
   const debugLog = debug.extend("dirty-files");
   const files = await readdir(buildDir);
 
-  let promises = [];
-  files.forEach(fn => {
-    if (fn.endsWith(".json")) {
-      promises.push(usesConsoleLog(path.join(buildDir, fn)));
-    }
-  });
-
-  const settled = await Promise.allSettled(promises);
-  const dirtyArtifacts = [];
-  for (const s of settled) {
-    if (s.status === "rejected") {
-      debugLog("promise failure: %o", s.reason);
-      continue;
-    }
-    dirtyArtifacts.push(s.value);
-  }
-  return dirtyArtifacts;
+  const artifacts = [];
+  await Promise.allSettled(
+    files.map(async fn => {
+      if (fn.endsWith(".json")) {
+        try {
+          const itLogs = await usesConsoleLog(path.join(buildDir, fn));
+          if (itLogs) {
+            artifacts.push(fn);
+          }
+        } catch (e) {
+          debugLog("Promise failure: %o", e.message);
+        }
+      }
+    })
+  );
+  return artifacts;
 }
 
 module.exports = async function (config) {
   const debugLog = debug.extend("guard");
   // only check if deploying on MAINNET
+  // NOTE: this includes Ethereum Classic as well as Ethereum as they're only
+  // distinguishable by checking their chainIds, 2 and 1 respectively.
   if (config.network_id === 1) {
     debugLog("solidityLog guard for mainnet");
     try {
       const buildDir = config.contracts_build_directory;
-      const dirtyArtifacts = await findArtifactsThatUseConsoleLog(buildDir);
+      const loggingArtifacts = await findArtifactsThatUseConsoleLog(buildDir);
 
-      debugLog(`${dirtyArtifacts.length} consoleLog artifacts detected`);
+      debugLog(`${loggingArtifacts.length} consoleLog artifacts detected`);
       debugLog(
-        `config.solidityLog.preventConsoleLogMigration: ${config.solidityLog.preventConsoleLogMigration}`
+        "config.solidityLog.preventConsoleLogMigration: " +
+          config.solidityLog.preventConsoleLogMigration
       );
 
-      if (dirtyArtifacts.length) {
+      if (loggingArtifacts.length) {
         console.warn(
           `${os.EOL}Solidity console.log detected in the following assets:`
         );
-        console.warn(dirtyArtifacts.join(", "));
+        console.warn(loggingArtifacts.join(", "));
         console.warn();
 
         if (config.solidityLog.preventConsoleLogMigration) {
@@ -91,6 +93,8 @@ module.exports = async function (config) {
         }
       }
     } catch (err) {
+      if (err instanceof TruffleError) throw err;
+
       debugLog("Unexpected error %o:", err);
       // Something went wrong while inspecting for console log.
       // Log warning and skip the remaining logic in this branch
--- a/packages/core/lib/commands/migrate/runMigrations.js
+++ b/packages/core/lib/commands/migrate/runMigrations.js
@@ -61,9 +61,8 @@ async function getDirtyArtifactFiles(buildDir) {
 }
 
 module.exports = async function (config) {
-  // only check if deploying on MAINNET
   const debugLog = debug.extend("guard");
-
+  // only check if deploying on MAINNET
   if (config.network_id === 1) {
     debugLog("solidityLog guard for mainnet");
     try {
--- a/packages/core/lib/commands/migrate/runMigrations.js
+++ b/packages/core/lib/commands/migrate/runMigrations.js
@@ -1,46 +1,54 @@
-const { open } = require("fs/promises");
+const { createReadStream } = require("fs");
 const { readdir } = require("fs/promises");
 const path = require("path");
 const JSONStream = require("JSONStream");
 const Migrate = require("@truffle/migrate").default;
 const TruffleError = require("@truffle/error");
 const os = require("os");
+const debug = require("debug")("migrate:run");
 
 // Search for Push1 (60) to Push(32) 7F + console.log
 //                      60 ---- 7F  C O N S O L E . L O G
 const consoleLogRex = /[67][0-9a-f]636F6e736F6c652e6c6f67/i;
 
 async function usesConsoleLog(artifactJson) {
-  const fd = await open(artifactJson);
+  const debugLog = debug.extend("test");
+  debugLog("Artifact: %o", artifactJson);
 
   //create a parser to get the value of jsonpath .deployedBytecode
   const parser = JSONStream.parse(["deployedBytecode"]);
-  const stream = fd.createReadStream().pipe(parser);
+  const stream = createReadStream(artifactJson).pipe(parser);
 
   return new Promise((resolve, reject) => {
     stream.on("data", data => {
       //JSONParse will emit the entire string/value
       //so initiate stream cleanup here
       stream.destroy();
-      resolve(consoleLogRex.test(data));
+      const usesConsoleLog = consoleLogRex.test(data);
+      debugLog("usesConsoleLog:", usesConsoleLog);
+      resolve(usesConsoleLog);
     });
 
     stream.on("error", err => {
-      fd.close();
+      stream.destroy();
+      debugLog("onError: %o", err);
       reject(err);
     });
-
-    stream.on("close", () => fd.close());
   });
 }
 
 module.exports = async function (config) {
   // only check if deploying on MAINNET
+  const dirtyArtifacts = [];
+  let continueTesting = true;
+  const debugLog = debug.extend("guard");
+  debugLog("config.network_id", config.network_id);
+  debugLog("config.solidityLog", config.solidityLog);
   if (config.network_id === 1) {
+    debugLog("solidityLog guard for mainnet");
     try {
       const buildDir = config.contracts_build_directory;
       const maybeArtifacts = await readdir(buildDir);
-      const dirtyArtifacts = [];
       for (let i = 0, L = maybeArtifacts.length; i < L; i++) {
         const fn = maybeArtifacts[i];
         if (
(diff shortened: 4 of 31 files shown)


packages/contract/lib/contract/index.js
```diff
@@ -156,6 +157,10 @@ if (typeof Web3 === "object" && Object.keys(Web3).length === 0) {
 
   Contract._constructorMethods = constructorMethods(Contract);
 
+  Contract.prototype[util.inspect.custom] = function (/* _depth , _options */) {
+    return `Contract at: ${this.address}`;
+  };
+
```
cds-amal, line 163: This might be controversial. I did this to make the REPL experience better because the giant wall of text to display a contract instance is a massive wall of text, pages long.

packages/config/src/configDefaults.ts
```diff
@@ -28,6 +28,10 @@ export const getInitialConfig = ({
       }
     },
     verboseRpc: false,
+    solidityLog: {
+      displayPrefix: "",
+      disableMigration: false
```
cds-amal, line 33: Should be `disableMigrate`, oops. Open to better names.

packages/config/test/unit.test.ts
```diff
@@ -3,52 +3,65 @@ import TruffleConfig from "../dist";
...
-    it("a simple object", async () => {
-      const expectedRandom = 42;
-      const expectedFoo = "bar";
-      const obj = {
-        random: expectedRandom,
-        foo: expectedFoo
+    it("solidityLog", () => {
+      const expectedSolidityLog = {
+        displayPrefix: "",
+        disableMigration: false
```
cds-amal, line 16: Should be `disableMigrate`

davidmurdoch:
This should be tested by using `import "@truffle/console.sol";`, in a contract, right?

haltman-at:
Hey Amal could you put an actual title on this PR?  I had no idea what this was.

packages/core/lib/console.js
```diff
@@ -323,38 +320,47 @@ class Console extends EventEmitter {
...
     );
 
-    if (spawnResult.stderr) {
-      // Theoretically stderr can contain multiple errors.
-      // So let's just print it instead of throwing through
-      // the error handling mechanism. Bad call?
-      debug(spawnResult.stderr.toString());
-    }
+    // Theoretically stderr can contain multiple errors.
+    // So let's jsut print it instaed of throwing through
```
davidmurdoch, line 338:
```suggestion
    // So let's just print it instead of throwing through
```

packages/core/lib/console.js
```diff
@@ -323,38 +320,47 @@ class Console extends EventEmitter {
...
+    // the error handling mechanism. Bad call?
+    sid.stderr.on("data", data => {
+      debug(data.toString());
+    });
+
+    sid.stdout.on("data", data => {
+      console.log(data.toString().trim());
```
davidmurdoch, line 345: Can you add a comment why this is now needed when it wasn't before?

packages/core/lib/console.js
```diff
@@ -323,38 +320,47 @@ class Console extends EventEmitter {
...
-      // Theoretically stderr can contain multiple errors.
-      // So let's just print it instead of throwing through
-      // the error handling mechanism. Bad call?
-      debug(spawnResult.stderr.toString());
-    }
+    // Theoretically stderr can contain multiple errors.
+    // So let's jsut print it instaed of throwing through
+    // the error handling mechanism. Bad call?
+    sid.stderr.on("data", data => {
+      debug(data.toString());
```
davidmurdoch, line 341: IIRC, stderr's `data` event might emit in chunks; `data` might be in the middle of a line. Here is a StackOverflow question about it: https://stackoverflow.com/questions/18243810/handling-chunked-responses-from-process-stdout-data-event

packages/core/lib/console.js
```diff
@@ -323,38 +320,47 @@ class Console extends EventEmitter {
...
+      sid.on("close", code => {
+        if (!code) {
+          // re-provision to ensure any changes are available in the repl
+          this.provision();
 
-    //display prompt when child repl process is finished
-    this.repl.displayPrompt();
+          //display prompt when child repl process is finished
+          this.repl.displayPrompt();
+          void resolve();
```
davidmurdoch, line 356: While `reject` after a `resolve` doesn't do anything, as the Promise will ignore it, it's best to avoid it.

```suggestion
          return void resolve();
```

cds-amal:
> This should be tested by using `import "@truffle/console.sol";`, in a contract, right? No heavy_check_mark from me without this :-)

I'll have to research this a bit as I'm a bit unclear on how to manage `@ganache/console.log` dependency, as well as resolving its `console.sol` contract to work in both webpack and non-webpack scenarios. We already do this for other contracts in `@truffle/resolver` so I should probably start there.

packages/core/lib/console.js
cds-amal: Yikes, thanks!

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,107 @@
...
 
 module.exports = async function (config) {
+  // only check if deploying on MAINNET
+  const debugLog = debug.extend("guard");
+
+  if (config.network_id === 1) {
+    debugLog("solidityLog guard for mainnet");
+    try {
+      const buildDir = config.contracts_build_directory;
+      const dirtyArtifacts = await getDirtyArtifactFiles(buildDir);
```
MicaiahReid, line 71: I feel like this function name could be more specific. It's not immediately clear what "dirty artifacts" means, but it could be with a better name. Just a suggestion though, my opinions aren't very strong here

packages/config/src/configDefaults.ts
```diff
@@ -28,6 +28,10 @@ export const getInitialConfig = ({
       }
     },
     verboseRpc: false,
+    solidityLog: {
+      displayPrefix: "",
+      disableMigrate: false
```
MicaiahReid, line 33: This has probably already been discussed, but I don't tend to not dislike :wink: boolean settings with a negative in the name... it's just confusing. Is there a reason not to name it `enableMigrate`?

MicaiahReid (commented):
Great work and super stoked for this. Thanks @cds-amal!

packages/config/src/configDefaults.ts
cds-amal: Good point @MicaiahReid. What do you think of `guardMigrate`, or `preventConsoleLogMigration` ?

packages/core/lib/commands/migrate/runMigrations.js
cds-amal: What do you think of `getArtifactsUsingConsoleLog`?

MicaiahReid: Works for me!

packages/config/src/configDefaults.ts
MicaiahReid: To me `preventConsoleLogMigration` is very clear. I wonder if there are other opinions, but I'd hate for this to get bogged down with this bike shedding.

cds-amal:
Needed to rebase to run CI. Apologies.

packages/config/src/configDefaults.ts
lsqproduction: i like `preventConsoleLogMigration` as well

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,106 @@
...
+  const settled = await Promise.allSettled(promises);
+  const dirtyArtifacts = [];
+  for (const s of settled) {
+    if (s.status === "rejected") {
+      debugLog("promise failure: %o", s.reason);
+      continue;
+    }
+    dirtyArtifacts.push(s.value);
+  }
+  return dirtyArtifacts;
```
davidmurdoch, line 60: This is ever so slightly more performant, as it avoids looping over the files twice. But there is an actual _fix_ here, in think, which is that `dirtyArtifacts` now contains the actual file names of the files that contain console.log, instead of `true`/`false`:

I didn't test this lol

```
  const dirtyArtifacts = [];
  await Promise.allSettled(files.map(async fn => {
    if (fn.endsWith(".json")) {
      try {
        const usedConsoleLog = await usesConsoleLog(path.join(buildDir, fn)));
        if (usedConsoleLog) {
          dirtyArtifacts.push(fn);
        }
      } catch(e) {
        debugLog("promise failure: %o", e.message);
      }
    }
  }));

  return dirtyArtifacts;
```

davidmurdoch: p.s., please add a test for this :-)

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,106 @@
...
+}
 
 module.exports = async function (config) {
+  const debugLog = debug.extend("guard");
+  // only check if deploying on MAINNET
+  if (config.network_id === 1) {
```
davidmurdoch, line 66: Just a note, this covers Ethereum Classic (chainId 2 and Ethereum (chainId 1).

packages/core/lib/console.js
davidmurdoch: @gnidan Any thoughts on this corner case? I don't think `console-child.js` is actually capable of causing the above issue, so it's probably fine to ignore this case.

packages/resolver/lib/sources/truffle/index.ts
```diff
@@ -68,6 +73,30 @@ export class Truffle implements ResolverSource {
...
+      let unbundledGanacheConsoleSol;
+      // @ts-ignore
+      if (typeof BUNDLE_VERSION === "undefined") {
+        unbundledGanacheConsoleSol = path.resolve(
+          path.join(
+            require.resolve("@ganache/console.log"),
+            "..",
+            "..",
+            "console.sol"
+          )
```
davidmurdoch, line 87: I think this should work:

```suggestion
          require.resolve("@ganache/console.log/console.sol")
```

packages/resolver/lib/sources/truffle/index.ts
```diff
@@ -68,6 +73,30 @@ export class Truffle implements ResolverSource {
...
+        );
+      }
+      const actualImportPath =
+        // @ts-ignore
+        typeof BUNDLE_VERSION !== "undefined"
+          ? path.resolve(path.join(__dirname, `console.sol`))
+          : unbundledGanacheConsoleSol;
+      const body = fse.readFileSync(actualImportPath, { encoding: "utf8" });
```
davidmurdoch, line 95: YUNO async? Or at least cache it? Otherwise this file read will happen for not just every _file_ that imports `console.sol`, but every separate compilation of ever file that imports `console.sol` (like if a helper util uses `console.sol` which is then used by multiple contracts).

Optionally, you could inline `console.sol` when webpacking by using the define plugin to replace this `body` with the contents at bundle time. But that's weird.

packages/environment/develop.js
```diff
@@ -59,35 +63,56 @@ const Develop = {
...
-    if (options.log) {
-      debugRPC.enabled = true;
-
-      loggers.ganache = function () {
+    const sanitizeAndCallFn =
+      fn =>
+      (...args) => {
         // HACK-y: replace `{}` that is getting logged instead of ""
-        var args = Array.prototype.slice.call(arguments);
+        // var args = Array.prototype.slice.call(arguments);
```
davidmurdoch, line 75: should this line be deleted?

```suggestion
```

packages/resolver/lib/resolver.ts
```diff
@@ -53,15 +53,16 @@ export class Resolver {
     import_path: string,
     search_path?: string
   ): ReturnType<typeof contract> {
-    let abstraction;
-    this.sources.forEach((source: ResolverSource) => {
+    for (const source of this.sources) {
       const result = source.require(import_path, search_path);
       if (result) {
-        abstraction = contract(result);
+        let abstraction = contract(result);
```
davidmurdoch, line 59:
```suggestion
        const abstraction = contract(result);
```

packages/core/lib/console.js
gnidan: Oh yeah definitely should make sure we buffer this data!

cds-amal: Additional context. The data is from stderr and was previously fed to debug, a conditional, decorated console.error. Someone would have to invoke `DEBUG=console truffle develop` to hook into it.

packages/core/lib/commands/migrate/runMigrations.js
cds-amal: nice.

packages/contract/lib/contract/index.js
eggplantzzz: I actually don't understand what this is doing. I assume from what you said it has something to do with providing custom output when logging contracts. Can you explain to me a little bit how this works?

cds-amal: When an object is evaluated in the REPL it will be inspected. For example, when you get the deployed instance for a contract (`meta = await MetaCoin.deployed()`), that contract instance is inspected, and the REPL dumps a wall of text to represent the instance object and its numerous properties. This alters the representation to be a single line:  `Contract at: ..address..`

eggplantzzz: Ok so `util.inspect.custom` is the symbol property on prototype that defines the result of inspecting the object in question?

cds-amal: yes

packages/resolver/lib/sources/truffle/index.ts
cds-amal: I think caching is the best approach, and would prefer to cache at the resolver interface, this function's caller. Trade memory for disk-access for all the sources.

packages/contract/lib/contract/index.js
cds-amal: @eggplantzzz I didn't think of the implication to webpacked @truffle/contract :/. I'll investigate Monday.

packages/resolver/lib/sources/truffle/index.ts
davidmurdoch: Ok. Make an issue for this and we can consider this comment resolved!

cds-amal: #5757, if you please.

emilyJLin95:
I didn't look at the code but the flow/output makes sense for the most part. Some follow up questions that might be relevant for the docs team

1. Will this work with Vyper print?
2. Will a user need to `npm i truffle/console.sol` like they would have to for `ganache/console.log/console.sol`?
3. Does the displayLogPrefix work for `ganache/console.log/console.sol` as well? As in, will it show up on the ganache transaction logs too?

davidmurdoch:
@emilyJLin95 

1. Will this work with Vyper print?
Yes, Vyper effectively uses the same EVM bytecode as console.sol; Ganache doesn't know what the original language is.
2. Will a user need to npm i truffle/console.sol like they would have to for ganache/console.log/console.sol?
No, it's already included (just like `truffle/Assert.sol`, and friends)
3. Does the displayLogPrefix work for ganache/console.log/console.sol as well? As in, will it show up on the ganache transaction logs too?
I'll let @cds-amal answer this one.

cds-amal:
@emilyJLin95 

>Does the displayLogPrefix work for ganache/console.log/console.sol as well? As in, will it show up on the ganache transaction logs too?

The prefix only appears in the context of `truffle test` and the truffle develop console. They do not appear in Ganache's TTY output. Is that what you meant by Ganache transaction logs?

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,110 @@
...
+  });
+}
+
+async function findArtifactsThatUseConsoleLog(buildDir) {
+  const debugLog = debug.extend("dirty-files");
+  const files = await readdir(buildDir);
+
+  const artifacts = [];
+  await Promise.allSettled(
+    files.map(async fn => {
```
eggplantzzz, line 46: `fn` here is "filename"? small thing but could we use the whole word since `fn` is commonly used for function?

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,110 @@
...
+              os.EOL
+          );
+        }
+      }
+    } catch (err) {
+      if (err instanceof TruffleError) throw err;
+
+      debugLog("Unexpected error %o:", err);
+      // Something went wrong while inspecting for console log.
+      // Log warning and skip the remaining logic in this branch
```
eggplantzzz, line 100: do we want to continue with migrations if the user has this turned on and something goes wrong?

packages/core/lib/commands/test/run.js
```diff
@@ -38,16 +38,21 @@ module.exports = async function (options) {
   }
 
   // Start managed ganache network
-  async function startGanacheAndRunTests(ipcOptions, ganacheOptions, config) {
+  async function startGanacheAndRunTests(
+    ipcOptions,
+    ganacheOptions,
+    truffleConfig
```
eggplantzzz, line 44: why change this variable to `truffleConfig`? we don't really have a convention of doing this

packages/core/lib/console.js
```diff
@@ -243,10 +243,7 @@ class Console extends EventEmitter {
   provision() {
     let files;
     try {
-      const unfilteredFiles = fse.readdirSync(
-        this.options.contracts_build_directory
-      );
-      files = unfilteredFiles.filter(file => file.endsWith(".json"));
+      files = fse.readdirSync(this.options.contracts_build_directory);
```
eggplantzzz, line 246: I'm not sure it is good to remove that filter. I think it was added at some point to prevent errors when there were other things in the `build` folder.

packages/resolver/lib/resolver.ts
```diff
@@ -16,7 +13,10 @@ export class Resolver {
   options: any;
   sources: ResolverSource[];
 
-  constructor(options: any, resolverOptions: ResolverOptions = {}) {
+  constructor(
+    options: any,
+    resolverOptions: ResolverOptions = { includeTruffleSources: true }
```
eggplantzzz, line 18: do we really want to include Truffle sources by default? if so, I would say we should just get rid of `includeTruffleSources` (and the default here) completely and just add them

packages/resolver/lib/resolver.ts
```diff
@@ -53,15 +53,16 @@ export class Resolver {
     import_path: string,
     search_path?: string
   ): ReturnType<typeof contract> {
-    let abstraction;
-    this.sources.forEach((source: ResolverSource) => {
+    for (const source of this.sources) {
       const result = source.require(import_path, search_path);
       if (result) {
-        abstraction = contract(result);
+        const abstraction = contract(result);
```
eggplantzzz, line 59: ooh nice catch

packages/truffle/test/scenarios/solidity_console/Printf.sol
```diff
@@ -0,0 +1,39 @@
...
+  address public owner = msg.sender;
+
+  constructor() payable { }
+
+  function log_string() view public {
+    console.log("String! I am not a twine!");
+  }
+
+  function log_unicode() view public {
+    console.log(unicode"This is unicode: ☮");
```
eggplantzzz, line 16: whoa, I've never seen this before - how does `unicode` work here?

packages/core/lib/commands/migrate/runMigrations.js
cds-amal: Ah, good question. Originally, my thinking was to stop if detecting consoleLog usage, letting the workflow's next steps handle the other exceptions. What do you think?

packages/core/lib/console.js
cds-amal: Good catch! I'll revert this change.

packages/core/lib/console.js
```diff
@@ -323,38 +320,55 @@ class Console extends EventEmitter {
...
+    const spawnOptions = { stdio: "pipe" };
     const settings = ["config", "network", "url"]
       .filter(setting => options[setting])
       .map(setting => `--${setting} ${options[setting]}`)
       .join(" ");
 
     const spawnInput = `${settings} -- ${inputStrings}`;
 
-    const spawnResult = spawnSync(
+    const sid = spawn(
```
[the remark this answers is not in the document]
cds-amal: Spawned id, or a vicious reference to a punk child process.

packages/core/lib/console.js
cds-amal: Actually, I think we can do the filtering in one loop

packages/core/lib/commands/test/run.js
cds-amal: It made the code less confusing for me. Since `startGanacheAndRunTests` only uses positional arguments, the renaming won't impact the callers.

packages/core/lib/commands/migrate/runMigrations.js
eggplantzzz: I'm not really sure what we should do. Maybe this is fine for now.

packages/core/lib/commands/test/run.js
eggplantzzz: Well I'm not going to fight you on it, but we use `config` like literally everywhere else in the codebase.

eggplantzzz:
This works with multiple variables specified in the form `console.log("%o - %o", var1, var2)`, nice!

One thing, numbers end up getting printed with a trailing `n`. In other words, the number `1000` will be printed as `1000n`. I don't really think this should hold up this PR but in the future we may want to consider removing this before printing?

eggplantzzz (approved):
i think this is cool

fainashalts (commented):
So excited to see this feature go out! Looks great @cds-amal. Just a few comments from me.

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,110 @@
...
+    stream.on("data", data => {
+      //JSONParse will emit the entire string/value
+      //so initiate stream cleanup here
+      stream.destroy();
+      const usesConsoleLog = consoleLogRex.test(data);
+      debugLog("usesConsoleLog:", usesConsoleLog);
+      resolve(usesConsoleLog);
+    });
+
+    stream.on("error", err => {
```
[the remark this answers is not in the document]
eggplantzzz: I think that would be a pretty non-standard implementation for this kind of a thing :/ I would vote we stick with the textbook usage described by Node in this case. Any other opinions?

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,110 @@
...
+    });
+  });
+}
+
+async function findArtifactsThatUseConsoleLog(buildDir) {
+  const debugLog = debug.extend("dirty-files");
+  const files = await readdir(buildDir);
+
+  const artifacts = [];
+  await Promise.allSettled(
```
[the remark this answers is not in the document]
eggplantzzz: You can't really use an `async` function with `forEach`. I mean, you can but it is probably not doing what you think it is doing.

I'm pretty sure `forEach` doesn't "respect" `async` functions in that it will execute each one and the code flow will continue on without waiting for the `async` stuff to resolve. `for` loops, however, do "respect" `await`s :)

packages/core/lib/commands/migrate/runMigrations.js
cds-amal: This logic tests if console.log code exists in the `bytecode` before migration, and I wanted the errors to reflect that concern. Migration logic still has to kick in and errors are already handled. In the future, we could have a static analysis step during compilation that would persist its outcome, so we query the artifact instead of analyzing on the fly.

packages/resolver/lib/resolver.ts
cds-amal: I think this should be in a separate PR.

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,110 @@
...
+async function usesConsoleLog(artifactJson) {
+  const debugLog = debug.extend("test");
+  debugLog("Artifact: %o", artifactJson);
+
+  //create a parser to get the value of jsonpath .deployedBytecode
+  const parser = JSONStream.parse(["deployedBytecode"]);
+  const stream = createReadStream(artifactJson).pipe(parser);
+
+  return new Promise((resolve, reject) => {
+    stream.on("data", data => {
```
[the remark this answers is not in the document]
cds-amal: If the handlers were more complex, It would make sense to extract them. As it is, I think the code is idiomatic, small as it is.

packages/core/lib/commands/migrate/runMigrations.js
```diff
@@ -1,6 +1,110 @@
...
+
+      debugLog(`${loggingArtifacts.length} consoleLog artifacts detected`);
+      debugLog(
+        "config.solidityLog.preventConsoleLogMigration: " +
+          config.solidityLog.preventConsoleLogMigration
+      );
+
+      if (loggingArtifacts.length) {
+        console.warn(
+          `${os.EOL}Solidity console.log detected in the following assets:`
```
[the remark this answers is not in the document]
cds-amal: I split it into multiple lines to account for multiple contracts being joined and displayed on a console. I figured it's best to list the contracts on a separate line.  Using a template string is less readable after prettier does its thing.

packages/core/lib/commands/migrate/runMigrations.js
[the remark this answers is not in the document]
fainashalts: @eggplantzzz good point, I was just quickly writing up an example and you are right that you wouldn't want to use `foreach` if needing to wait for each line to resolve before continuing. I guess I'd still prefer filter and then a for loop because it is more readable but I get @cds-amal's point re: looping through twice vs. once and this is an efficient way to use `Promise.allSettled()` + `map`.

packages/core/lib/commands/migrate/runMigrations.js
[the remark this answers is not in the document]
fainashalts: Fair enough. If the handlers ever grow we can always pull them out then.

packages/truffle/test/scenarios/solidity_console/Printf.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.4.22 <0.9.0;
+
+import "truffle/Console.sol";
+
+contract Printf {
+  address public owner = msg.sender;
```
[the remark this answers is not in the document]
cds-amal: Ownership isn't needed for this test. I declared owner state out of habit. I could add a test that utilizes it.

packages/core/lib/commands/migrate/runMigrations.js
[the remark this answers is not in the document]
cds-amal: This is how this API works. The `error` event rejects the promise, and the `data` event resolves a bool.

packages/core/lib/commands/migrate/runMigrations.js
cds-amal: I'm going to resolve this for now, feel free to reopen if I misunderstood your intents.

packages/resolver/lib/resolver.ts
cds-amal: I opened #5771 to track this.

packages/core/lib/commands/migrate/runMigrations.js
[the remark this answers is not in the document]
eggplantzzz: But it seems plausible that you could add a `try/catch` in the "data" callback and handle everything there. I don't really like it as it kind of feels like you are subverting the API. Does that miss any other errors that this might catch?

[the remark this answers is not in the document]
davidmurdoch: An errors that occur in the data callback  function are not related to the stream and the stream won't know about those.

The pattern used here is correct.

packages/truffle/test/scenarios/solidity_console/Printf.sol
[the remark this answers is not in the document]
davidmurdoch: @cds-amal If anything I think you should remove the `owner`, since it isn't related to the functionality you are testing.

cliffoo (commented):
This is awesome!

What do you think about `displaySuffix`?

packages/config/src/configDefaults.ts
```diff
@@ -28,6 +28,10 @@ export const getInitialConfig = ({
       }
     },
     verboseRpc: false,
+    solidityLog: {
+      displayPrefix: "",
+      preventConsoleLogMigration: false
+    },
```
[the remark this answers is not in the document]
cds-amal: I created #5775 to follow up on that.

cds-amal:
> What do you think about `displaySuffix`?

What do you have in mind @cliffoo?

packages/contract/lib/contract/index.js
cds-amal: > * Let `a = Telephone.new(1)` and `b = Telephone.new(1)`, they both print `Contract at: ...`! Pointing out whether or not it's a Promise can avoid potential confusion.

> * Since we're changing this symbol, maybe there's more basic information other than just the address that might be useful? ContractName, methods, etc. That way it's not a wall of text and you still get only the info you'll likely need.

I've never had to inspect contract properties beyond `abi`, though many contracts' `abi` could be multi-pagers. What exactly should we present?

> * What if someone wants the default inspector? Like a `contract.asObject` to expose that? Maybe this is already in place via another interface so please correct me if I'm wrong.

Good question! Maybe `delete Contract.prototype[Symbol.for("nodejs.util.inspect.custom")]`, or `console.log(util.inspect(a, {customInspect: false}))`. I'd have to test the former when I get to a computer. 

What do people think? Should we remove this change, and let users monkey patch themselves? Leave it in, or PR separately and wait for user feedback for guidance? @eggplantzzz, @cliffoo, @sukanyaparashar, @fainashalts, @dongmingh, @lsqproduction?

cds-amal: Removing this functionality because, as @cliffoo pointed out, it is not well-defined. I'll consider the feedback and put up another PR.

packages/resolver/lib/sources/truffle/index.ts
```diff
@@ -68,6 +73,25 @@ export class Truffle implements ResolverSource {
         this.options.compilers
       );
       return { body: addressSource, filePath: importPath };
+    } else if (importPath === `truffle${path.sep}Console.sol`) {
```
davidmurdoch, line 76: I already :heavy_check_mark: the PR, but I do think this should also accept `truffle${path.sep}console.sol`

cds-amal: `Console.sol` is more consistent with the other truffle sources. What's your thinking?

davidmurdoch: While the consistency is nice, the other truffle sources are only to be used in Solidity-based tests. This is the first contract of its kind, and it aligns with HH's, Ganache's, and maybe most importantly, the name of the contract library itself.

> Contract and library names should also match their filenames.

- https://docs.soliditylang.org/en/v0.8.17/style-guide.html#contract-and-library-names

The library name is `console`. So I think the file should be `console.sol` to match.

davidmurdoch: cc @gnidan

sukanyaparashar (approved):
Thanks for this great logging feature, @cds-amal . Looks good to me! :)

packages/environment/develop.js
```diff
@@ -116,20 +140,31 @@ const Develop = {
...
-    const ipcNetwork = options.network || "develop";
+  /**
+   * Connect to a managed Ganache service. This will connect to an existing
+   * Ganache service if one exists, or, create a new one to connect to.
+   *
+   * @param {Object} ipcOptions
+   * @param {string} ipcOptions.network the network name.
+   * @param {Object} ganacheOptions to be used if starting a Ganache service is
+   *        necessary.
+   * @param {TruffleConfig} truffleConfig the truffle config.
```
gnidan, line 151: I'm not super keen on adding config as an argument here... could we, possibly as follow-on work, just pull out the specific fields we want instead?

gnidan (approved):
Found no dragons. Did not check for wyverns.

packages/environment/develop.js
cds-amal: Yep, added #5784 for that.

packages/resolver/lib/sources/truffle/index.ts
gnidan: Definitely prefer aligning with OZ's use of PascalCase for Solidity files. Unless they're just matching the exported symbol.

Seems like `library console` is bad style? For clear reasons, but still.

davidmurdoch: Yeah, `library console`, is not idiomatic Solidity, but `console.log` isn't either.

Since library names should match their filenames I think Truffle must either align its import filename with the library's name (`console`), or rewrite the library to match the proposed filename of `truffle/Console.sol` (so usage would be `Console.log` instead of `console.log` -- preventing drop in replacement).

cds-amal: How about allowing both `@truffle/[Cc]ontract.sol`, promote?

davidmurdoch (approved):

PR #5689, opened 2022-11-09
Add sourcify networks for November

haltman-at:
Sourcify has added more networks.  Updated this.

Because this is such a small routine change, I didn't bother testing it.  But, uh, you could try using `fetch-and-compile` with a contract from one of these networks, I guess?

Sourcify's own tests use addresses `0xA7e70Be8A6563DCe75299c30D1566A83fC63BC37` and `0x571bb36009bB26D5313244B30397D2a2341a2A11` for chain 592, and addresses `0x662749a392CeB1b5973a90FB2c388a2C18B8812c` and `0x9FEc9e780c422916E21845748e001E949A5ddD57` for chain 1001, so those seem like ones to use for this purpose!


eggplantzzz (approved):
i like

Issue #5690, opened 2022-11-10
Dashboard sometimes only listens on ipv6 loopback by default


benjamincburns:
## Issue

While helping @OnlyOneJMJQ troubleshoot an issue with the truffle dashboard, we encountered a problem where the dashboard's HTTP server was listening only on the IPv6 loopback interface, while his `truffle migrate` command was attempting to connect to the IPv4 interface. This resulted in a mix of `ECONNREFUSED` and `ECONNTIMEOUT` errors from the `truffle migrate` process as it was attempting to connect to the dashboard server process.

## Steps to Reproduce

1. Ensure that you don't have `dashboard.host` specified in your `truffle-config.js`
2. In one terminal instance run `truffle dashboard` to spin up the dashboard process
3. In another terminal run `truffle migrate --network dashboard`

Note: this may only reproduce on MacOS and/or with Node v18

## Expected Behavior

The `truffle migrate` command should succeed as expected.

## Actual Results

A mix of `ECONNREFUSED` and `ECONNTIMEOUT` errors are reported by the `truffle migrate` process, and that process exits with a failure code.

## Short-term workaround

This issue can be worked around by explicitly setting `dashboard.host` to "127.0.0.1" or "::1" in `truffle-config.js`, and ensuring that the value of the dashboard network config (the string, function, or object assigned to `networks.dashboard`) matches.

## Underlying cause

Most modern operating systems have two entries for `localhost` in their `/etc/hosts` (or in the case of Windows, `C:\Windows\system32\drivers\etc\hosts`) files. The first entry usually resolves to the ipv4 loopback address (`127.0.0.1`), and the second usually resolves to the ipv6 loopback address (`::1`).

For some reason when we initialize the dashboard's instance of `http.Server` via express with a hostname value of `localhost`, it's electing to listen _only_ on the ipv6 loopback interface. To address this, back in node v11.4.0 they added [an `ipv6Only` field to the `options` object that's passed into the `net.Server.listen` method](https://nodejs.org/docs/latest-v11.x/api/net.html#net_server_listen_options_callback).

What's confusing here is that `ipv6Only` defaults to `false`, which should trigger "dual stack" support for `net.Server.listen`, and by extension, `http.Server.listen`, as `http.Server.listen` is identical to `net.Server.listen`.

[...]

PR #5699, opened 2022-11-14
Make dashboard bind to IPv4 & IPv6 when available

benjamincburns:
## PR description

Addresses issue #5690.

This change makes it so that the dashboard server and dashboard message bus listens on both IPv4 and IPv6 for a given hostname. In the event that one of the two protocols is not available for a given hostname, it will only listen on the protocol that is available.

## Testing instructions

1. Run `truffle dashboard` in your project of choice.
2. Ensure that you can connect to it via http://127.0.0.1:24012 (note the port number may be different on your machine)
3. Ensure that you can connect to it via http://[::1]:24012 (note the port number may be different on your machine)
4. Configure the `dashboard` network to connect to host `127.0.0.1` and ensure that `truffle migrate --reset --network dashboard` succeeds under this configuration.
5. Configure the `dashboard` network to connect to host `::1` and ensure that `truffle migrate --reset --network dashboard` succeeds under this configuration.
6. Configure the `dashboard` network to connect to host `localhost` and ensure that `truffle migrate --reset --network dashboard` succeeds under this configuration.

## Documentation

## Note to reviewers

There's a bit of unfortunate code duplication here. I would have preferred to put the logic for resolving the IPs of a given local hostname into the `@truffle/dashboard-message-bus-common` package, but the webpack operation in the build for the dashboard frontend choked on the node-specific modules that are used by this logic (the `os` and `net` modules). As a result, I decided it was easier to _temporarily_ duplicate this code between the `utils` module in the dashboard server, and the `utils` module in the `dashboard-message-bus` module.

That said, we're still not breaking "the rule of three" here, so hopefully this can be approved as-is. However if #5621 happens to be ready to merge before this PR, I can rebase this code on `develop`, move the logic to `@dashboard-message-bus-common`, and drop the `net` and `os` packages in the dashboard's newly-ejected webpack config.


OnlyOneJMJQ:
Will give this a try today and get back to you!

benjamincburns:
I think it's safe to review this while we're waiting on feedback from Josh, as it's a fairly small change and I've already verified by other means (`lsof -i -P | grep LISTEN`) that it has the intended result (listening on both IPv6 _and_ IPv4 when available).

benjamincburns:
Just confirming that @OnlyOneJMJQ got back to me via Slack and confirmed that this change does indeed fix the issue he observed in #5690

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -104,15 +107,26 @@ export class DashboardServer {
...
-        if (this.autoOpen) {
-          const host = this.host === "0.0.0.0" ? "localhost" : this.host;
-          open(`http://${host}:${this.port}`);
-        }
-        resolve();
-      });
-    });
+    const bindIpAddresses = await resolveBindHostnameToAllIps(this.host);
+    this.httpServers = await Promise.all(
+      bindIpAddresses.map(async bindIpAddress => {
```
davidmurdoch, line 112:
```suggestion
      bindIpAddresses.map(bindIpAddress => {
```

packages/dashboard-message-bus/lib/utils.ts
```diff
@@ -79,3 +92,21 @@ export const sendAndAwait = (socket: WebSocket, message: Message) => {
     socket.send(encodedMessage);
   });
 };
+
+const getLocalInterfaceAddresses = () => {
+  return Object.keys(networkInterfaces()).flatMap(name =>
+    networkInterfaces()[name].map(iface => iface.address)
+  );
```
davidmurdoch, line 99: I believe `networkInterfaces` ends up in a system call, so the results can change between calls, even within a single node.js event loop. I could be wrong, but I'd unless you want to investigate i'd just cache the return of `networkInterfaces()` and use it for the keys and within the `flatMap`. (but if you do investigate and learn anything interesting please teach me your learnings!)

davidmurdoch (changes requested):

packages/dashboard-message-bus/lib/utils.ts
benjamincburns: Ah good catch - will do.

benjamincburns:
Argh, so this is frustrating... I added better error handling for the case when the user mistypes their desired hostname, but it turns out that the error handling code is unreachable due to [how we provision additional ports for the websocket servers](https://github.com/trufflesuite/truffle/blob/6af18fad1fae0b9c5d4316c4280b3face6fa48c0/packages/dashboard/lib/DashboardServer.ts#L156-L159). That is, when I purposefully test with a mistyped hostname, it throws `Error: getaddrinfo ENOTFOUND` from that line.

The bigger issue is that the `getPort` function that we're using can't find an unused port that's common between multiple local IP addresses. If I call it twice (once for IPv6 and once for IPv4) it may return different ports.

I think this means that in order to make this change properly I need to refactor the message bus to reuse the `http.Server` instance(s) that `DashboardServer` sets up to serve up the dashboard webpage.

cc @cliffoo as FYI

eggplantzzz:
Closing this for now as the changes here are too big for something that we are not able to reliably reproduce on our end. If we can reproduce this in the future we'll revisit it.

PR #5708, opened 2022-11-16
Early return dashboard event handlers if network isn't dashboard

cliffoo:
So @cds-amal discovered that sometimes event handlers for dashboard are run even if dashboard isn't the targeted network, causing a failed request to the message bus on every such occasion. The fix is to check that the current network is in fact dashboard, then proceed with the event handler logic.

### DIFF
--- a/packages/events/defaultSubscribers/dashboard.js
+++ b/packages/events/defaultSubscribers/dashboard.js
@@ -26,6 +26,8 @@ module.exports = {
   handlers: {
     "compile:start": [
       async function () {
+        if (!isDashboardNetwork(this.config)) return;
+
         try {
           const publishLifecycle = await this.messageBus.publish({
             type: "debug",
@@ -41,9 +43,7 @@ module.exports = {
     ],
     "rpc:request": [
       function (event) {
-        if (!isDashboardNetwork(this.config)) {
-          return;
-        }
+        if (!isDashboardNetwork(this.config)) return;
 
         const { payload } = event;
         if (payload.method === "eth_sendTransaction") {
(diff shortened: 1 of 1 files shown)


packages/events/defaultSubscribers/dashboard.js
```diff
@@ -41,9 +43,7 @@ module.exports = {
     ],
     "rpc:request": [
       function (event) {
-        if (!isDashboardNetwork(this.config)) {
-          return;
-        }
+        if (!isDashboardNetwork(this.config)) return;
```
cds-amal, line 46: Do you think this is the right place for this, or should it be closer to where the `rpc:request` is emitted? If filtering can be done close to the source, then we don't have to expend an event.

packages/events/defaultSubscribers/dashboard.js
```diff
@@ -57,9 +57,7 @@ module.exports = {
     ],
     "rpc:result": [
       function (event) {
-        if (!isDashboardNetwork(this.config)) {
-          return;
-        }
+        if (!isDashboardNetwork(this.config)) return;
```
cds-amal, line 60: Same as above

cds-amal (commented):
A quick question, @cliffoo.

packages/events/defaultSubscribers/dashboard.js
cliffoo: I believe it belongs here. Wherever `rpc:request` (or any other) event is emitted, the primary concern is that the action of emitting is carried out, not what subscribes to the events and what they decide to do with it.

More importantly I think it complicates the logic at source. E.g. for `rpc:request`, checking that the network isn't dashboard doesn't mean the event shouldn't be emitted, it just means dashboard should ignore it. It should still be emitted because there are other workflows that depend on this event and have event handlers setup for it.

cliffoo: On a side note I do think repeatedly writing `if (!isDashboardNetwork(this.config)) return;` isn't ideal. Maybe when dashboard has more event handlers that need such setup we can extract this logic into a decorator / wrapper.

cds-amal (approved):
Looks good! Thanks, @cliffoo

Issue #5705, opened 2022-11-15
Truffle packages' `package.json`s have a bunch of extra fields that shouldn't exist


davidmurdoch:
Like this one: https://github.com/trufflesuite/truffle/blob/275115e1ce358aad3050a7583c034e5d19d885a1/packages/environment/package.json#L40

These extra fields should be identified and removed.

lsqproduction:
@davidmurdoch, thanks for catching these. Any thoughts on how one might identify these?

davidmurdoch:
I think @eggplantzzz added them all in a single commit long ago. Doing a git blame on that line above should reveal the commit that caused it.

haltman-at:
Huh, I thought these got added by our release process?

davidmurdoch:
Per this comment https://github.com/lerna/lerna/issues/1880#issuecomment-456160357:

1. `gitHead` provides metadata for packages in the registry, and is based on the [existing behavior](https://github.com/npm/read-package-json/blob/6bac747004d5bc334a9f37c3799a965954d16641/read-json.js#L363) of npm publish.
2. It is not meant to be versioned. `lerna publish` should always be [resetting those changes](https://github.com/lerna/lerna/blob/5a7392e59e6ae5ee8aea10467ffb4b12f3b1e3d1/commands/publish/index.js#L180-L182), unless `--no-git-reset` is passed.
3. `gitHead` is a temporary property only meant to exist during `lerna publish` when it is constructing the package tarballs.

eggplantzzz:
I don't really know how these get committed to source control to be honest. I assume they accidentally get committed along the way.

cliffoo:
#5710 is merged and included in the [v5.6.7](https://github.com/trufflesuite/truffle/releases/tag/v5.6.7) release. Closing.

PR #5710, opened 2022-11-16
get rid of unnecessary gitHead property from package.jsons

eggplantzzz:
works towards resolving https://github.com/trufflesuite/truffle/issues/5705


cds-amal (approved):
Wow, get that outta here! (pending tests ofc)
...from Thu Sep 24 16:09 2020, truffle@5.1.46!

PR #5717, opened 2022-11-18
Remove custom jest resolver for db

cliffoo:
So if I understand [this thread](https://github.com/uuidjs/uuid/pull/616) correctly, now that we're using jest@29 we can remove the custom jest resolver introduced in [this PR](https://github.com/trufflesuite/truffle/pull/5359). Hopefully CI passes.


cds-amal (approved):
Thanks, @cliffoo. LGTM

PR #5734, opened 2022-11-24
Hack config and provider - what will break?

cds-amal:
We learned @truffle/config (`delete require.cache[...]`) doesn't invalidate the cache because webpack redefines require.  This means production usage of `@truffle/config` utilizes node require caching, and only loads 
`@truffle/config` once per process invocation.

This suggests we can modify config to cache its provider. `config.provider`  can be a thunk (`provider: () => ...`) which results in a new provider instance every time the thunk is invoked (many times!). Now, when we evaluate it the first time, we store the instance instead of the thunk.

This *Draft* PR will try to ferret out issues in our understanding. Currently, we think:
- there may be some nonce issue with hdwallet-provider, or its use in one of its sub-providers (per @davidmurdoch)
- tests that rely on the old behavior may need to be updated. 

## Documentation

### DIFF
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -190,7 +190,7 @@ class TruffleConfig {
     // The require-nocache module used to do this for us, but
     // it doesn't bundle very well. So we've pulled it out ourselves.
     //@ts-ignore
-    delete require.cache[Module._resolveFilename(file, module)];
+    delete originalRequire.cache[Module._resolveFilename(file, module)];
     const staticConfig = originalRequire(file);
 
     config.merge(staticConfig);
--- a/packages/config/src/index.ts
+++ b/packages/config/src/index.ts
@@ -190,7 +190,9 @@ class TruffleConfig {
     // The require-nocache module used to do this for us, but
     // it doesn't bundle very well. So we've pulled it out ourselves.
     //@ts-ignore
-    delete originalRequire.cache[Module._resolveFilename(file, module)];
+    //this doesn't do anything in the webpack build
+    //BUT may have an impact in non-webpack scenarios, ie, unit tests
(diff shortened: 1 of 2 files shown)


packages/config/src/index.ts
```diff
@@ -190,6 +190,8 @@ class TruffleConfig {
     // The require-nocache module used to do this for us, but
     // it doesn't bundle very well. So we've pulled it out ourselves.
     //@ts-ignore
+    //this doesn't do anything in the webpack build
+    //BUT may have an impact in non-webpack scenarios, ie, unit tests
```
cds-amal, line 194: @gnidan, @davidmurdoch CI passing supports removing this cache/clean code. Still have to test develop/console behavior for contra cases.

packages/provider/index.js
```diff
@@ -17,7 +17,9 @@ module.exports = {
   getProvider: function (options) {
     let provider;
     if (options.provider && typeof options.provider === "function") {
+      console.log("Unthunking provider:", ++count);
       provider = options.provider();
+      options.provider = provider;
```
cds-amal, line 22: Cache the provider, so there's only ever one for the duration of the process.

cds-amal (commented):
CI didn't suss out anything. Excitement grows.

packages/provider/index.js
davidmurdoch: oh yeaaahhh

davidmurdoch: side effect: the user _wants_ it to be called a bajillion times, and now it doesn't.

I think this was a bug in Truffle, but it's been around for so long users may be relying on it.

cds-amal: Well, no user was relying on it because they didn't have access to the functionality because webpack/original-require didn't allow it :)

PR #5739, opened 2022-11-29
Add --url option to the `truffle migrate` command

sukanyaparashar:
## PR description

This PR adds `--url` option to the `truffle migrate` command that connects to a specified url.

## Testing instructions

To test this option -
1. Run a ganache instance in a terminal.
2. Run `truffle migrate --url http://127.0.0.1:8545` inside a truffle project. The contracts inside the truffle project will be deployed in Ganache. 

## Documentation

### DIFF
--- a/packages/core/lib/commands/migrate/meta.js
+++ b/packages/core/lib/commands/migrate/meta.js
@@ -54,6 +54,10 @@ module.exports = {
       type: "boolean",
       default: true,
       hidden: true
+    },
+    "url": {
+      describe: "Use specified URL for provider",
+      type: "string"
     }
   },
   help: {
@@ -62,7 +66,7 @@ module.exports = {
       "                                " + // spacing to align with previous line
       "[--compile-all] [--compile-none] [--verbose-rpc] [--interactive]\n" +
       "                                " + // spacing to align with previous line
-      "[--skip-dry-run] [--describe-json] [--dry-run]",
+      "[--skip-dry-run] [--describe-json] [--dry-run] [--network <network>|--url <provider_url>]",
     options: [
       {
         option: "--reset",
@@ -118,8 +122,18 @@ module.exports = {
         option: "--save",
         description: "Specify whether the migration will save on chain",
         hidden: true
+      },
+      {
+        option: "--url",
+        description:
+          "Creates a provider using the given url and connects to the network."
+      },
+      {
+        option: "--network",
+        description:
+          "The network to connect to, as specified in the Truffle config."
       }
     ],
-    allowedGlobalOptions: ["network", "config", "quiet"]
+    allowedGlobalOptions: ["config", "quiet"]
   }
 };
--- a/packages/core/lib/commands/migrate/run.js
+++ b/packages/core/lib/commands/migrate/run.js
@@ -1,15 +1,29 @@
 module.exports = async function (options) {
   const WorkflowCompile = require("@truffle/workflow-compile").default;
   const { Environment } = require("@truffle/environment");
-  const Config = require("@truffle/config");
   const determineDryRunSettings = require("./determineDryRunSettings");
   const prepareConfigForRealMigrations = require("./prepareConfigForRealMigrations");
   const runMigrations = require("./runMigrations");
   const setUpDryRunEnvironmentThenRunMigrations = require("./setUpDryRunEnvironmentThenRunMigrations");
+  const loadConfig = require("../../loadConfig");
+  const OS = require("os");
+  const TruffleError = require("@truffle/error");
   const tmp = require("tmp");
   tmp.setGracefulCleanup();
 
-  const config = Config.detect(options);
+  if (options.url && options.network) {
+    const message =
+      "" +
+      "Mutually exclusive options, --url and --network detected!" +
+      OS.EOL +
+      "Please use either --url or --network and try again." +
+      OS.EOL +
+      "See: https://trufflesuite.com/docs/truffle/reference/truffle-commands/#migrate" +
+      OS.EOL;
+    throw new TruffleError(message);
+  }
+
+  let config = loadConfig(options);
   if (config.compileNone || config["compile-none"]) {
     config.compiler = "none";
   }
--- a/packages/truffle/test/scenarios/commands/migrate.js
+++ b/packages/truffle/test/scenarios/commands/migrate.js
@@ -34,5 +34,15 @@ describe("truffle migrate", () => {
         assert.fail();
       }
     }).timeout(20000);
+
+    it("doesn't throw when --url option is passed", async () => {
+      try {
+        await CommandRunner.run("migrate --url http://127.0.0.1:8545", config);
+      } catch (error) {
+        console.log("the logger contents -- %o", config.logger.loggedStuff);
+        console.log("the following error occurred -- %o", error.message);
+        assert.fail();
+      }
+    }).timeout(20000);
   });
 });


packages/core/lib/commands/migrate/meta.js
```diff
@@ -62,7 +66,7 @@ module.exports = {
       "                                " + // spacing to align with previous line
       "[--compile-all] [--compile-none] [--verbose-rpc] [--interactive]\n" +
       "                                " + // spacing to align with previous line
-      "[--skip-dry-run] [--describe-json] [--dry-run]",
+      "[--skip-dry-run] [--describe-json] [--dry-run] [--network <network>|--url <provider_url>]",
```
lsqproduction, line 69: Can we add the spacing to align with the previous line?

packages/core/lib/commands/migrate/meta.js
```diff
@@ -118,8 +122,18 @@ module.exports = {
...
         description: "Specify whether the migration will save on chain",
         hidden: true
+      },
+      {
+        option: "--url",
+        description:
+          "Creates a provider using the given url and connects to the network."
+      },
+      {
+        option: "--network",
```
lsqproduction, line 132: I'm curious about why you prefer `--network` lives here vs with the other global options.

Actually, can `url` be a global option as well?

packages/core/lib/commands/migrate/meta.js
```diff
@@ -54,6 +54,10 @@ module.exports = {
       type: "boolean",
       default: true,
       hidden: true
+    },
+    "url": {
+      describe: "Use specified URL for provider",
```
lsqproduction, line 59: let's have the same wording as the description under the `help` below.

packages/core/lib/commands/migrate/meta.js
sukanyaparashar: Sure @lsqproduction. I will add the spacing.

packages/core/lib/commands/migrate/meta.js
sukanyaparashar: The global options generally appends the options at the end of the usage. The reason behind keeping `--network` and `--url` options here rather than in the global options, is to show that they are mutually exclusive.

lsqproduction: Okie. thank you for the explanation.

lsqproduction (approved):

packages/core/lib/commands/migrate/run.js
```diff
@@ -1,15 +1,29 @@
...
+    const message =
+      "" +
+      "Mutually exclusive options, --url and --network detected!" +
+      OS.EOL +
+      "Please use either --url or --network and try again." +
+      OS.EOL +
+      "See: https://trufflesuite.com/docs/truffle/reference/truffle-commands/#migrate" +
+      OS.EOL;
+    throw new TruffleError(message);
+  }
```
cds-amal, line 24: I think Truffle sets`options.network = develop` when using the `truffle develop` REPL console that would trigger this error condition if the user enters `migrate --url`. We could either error better here, **or** filter the command in the console logic. What do you think?

cds-amal (changes requested):
Thanks, @sukanyaparashar. I have a question about how this feature would work in `truffle develop`.  I suspect we should error better, or not allow the user to enter this command in the REPL. What do you think?

sukanyaparashar:
Thanks for the approval @lsqproduction. However, I am holding off to merge it now till all the commands have the `--url` option added to it. I will be opening a PR per command to add the option to it. @cds-amal and I discussed about it and thought that it might be a better idea to add the `--url` option altogether to all the commands.

packages/core/lib/commands/migrate/run.js
sukanyaparashar: When we are already inside the REPL, then it ignores the `--url` option and does the migrations on the `develop` network (managed ganache with http://127.0.0.1:9545). It doesn't trigger the error condition. I think its better to error out the cause for that to let the users know that this option is not applicable inside the REPL.

sukanyaparashar: Please ignore my previous comment @cds-amal. I wasn't testing this with the correct branch. Your interpretation is correct that inside the develop REPL , `migrate --url` will trigger the above mentioned error. We can error better here to display something like "`migrate --url` is not supported here".

cds-amal: Yea, I prefer error better at the REPL input processing seam.

packages/core/lib/commands/migrate/meta.js
eggplantzzz: should we add a note in each of these letting the user know that these options are mutually exclusive? do you think that would be helpful?

packages/core/lib/commands/migrate/run.js
```diff
@@ -1,15 +1,29 @@
...
+
+  let config = loadConfig(options);
```
eggplantzzz, line 26: why not use `const` here?

sukanyaparashar: Oh! Yes, we can use `const` here.

packages/core/lib/commands/migrate/meta.js
sukanyaparashar: `Usage` section shows that they are mutually exclusive but if adding a note in the descriptions is preferable, then we can add it here too.

packages/core/lib/console.js
```diff
@@ -46,6 +46,16 @@ const processInput = (input, allowedCommands) => {
...
+  if (words.includes("--url")) {
+    const message = "url option is not supported within Truffle REPL";
+    return makeIIFE(`ℹ️ : ${message}`);
+  }
+
+  if (words.includes("--network")) {
+    const message = "network option is not supported within Truffle REPL";
+    return makeIIFE(`ℹ️ : ${message}`);
+  }
+
```
cds-amal, line 58: Let's add tests for these two

cds-amal (commented):
I think this command is good, though we should add tests for the REPL command processing.

packages/core/lib/commands/migrate/meta.js
eggplantzzz: whatever you think, I don't feel strongly about it :)

packages/core/lib/console.js
sukanyaparashar: I will add the tests. Thanks @cds-amal.

dongmingh (commented):
I ran some tests and see the results:

1. start `ganache` in one terminal, and run `truffle migrate --url http://127.0.0.1:8545` on the other terminal, the migration works properly.

2. start `truffle develop` and run the following tests:

2.1. `migrate --url http://127.0.0.1:8545`
```
            truffle(develop)> migrate --url http://127.0.0.1:8545
            'ℹ️ : url option is not supported within Truffle REPL'
```

2.2. `migrate --network develop`
```
           truffle(develop)> migrate --network develop
           'ℹ️ : network option is not supported within Truffle REPL'
```
2.3 `truffle migrate --network develop`
           migration works properly

2.4 `truffle migrate --url http://127.0.0.1:8545`
```
           truffle(develop)> truffle migrate --url http://127.0.0.1:8545
           Mutually exclusive options, --url and --network detected!
           Please use either --url or --network and try again.
           See: https://trufflesuite.com/docs/truffle/reference/truffle-commands/#migrate
```

In the last test, the message says both `--url` and `--network` are detected though only option `--url` is provided.

packages/core/lib/console.js
cds-amal: The tests should cover both logic branches; with and without the `truffle` prefix.

sukanyaparashar:
> I ran some tests and see the results:
> 
> 1. start `ganache` in one terminal, and run `truffle migrate --url http://127.0.0.1:8545` on the other terminal, the migration works properly.
> 2. start `truffle develop` and run the following tests:
> 
> 2.1. `migrate --url http://127.0.0.1:8545`
> 
> ```
>             truffle(develop)> migrate --url http://127.0.0.1:8545
>             'ℹ️ : url option is not supported within Truffle REPL'
> ```
> 
> 2.2. `migrate --network develop`
> 
> ```
>            truffle(develop)> migrate --network develop
>            'ℹ️ : network option is not supported within Truffle REPL'
> ```
> 
> 2.3 `truffle migrate --network develop` migration works properly
> 
> 2.4 `truffle migrate --url http://127.0.0.1:8545`
> 
> ```
>            truffle(develop)> truffle migrate --url http://127.0.0.1:8545
>            Mutually exclusive options, --url and --network detected!
>            Please use either --url or --network and try again.
>            See: https://trufflesuite.com/docs/truffle/reference/truffle-commands/#migrate
> ```
> 
> In the last test, the message says both `--url` and `--network` are detected though only option `--url` is provided.

Thanks for the thorough testing @dongmingh. I will make the required changes to address the issues with `truffle migrate` command in the REPL.

PR #5740, opened 2022-11-29
dashboard: Fix decoding text overflow

cliffoo:
So say your decoding is `"AAAAAAAAAA AAAAAAAAAA AAAAAAAAAA AAAAAAAAAA"`, the display is ok:

But if we remove the whitespaces it pushes the buttons away:

This PR is fix:


gnidan (approved):
Ah the lazy way :)

Thanks for fixing this so I can get back to my on-chain shenanigans!

PR #5743, opened 2022-11-29
Make @truffle/provider compatible with EIP-1193

sukanyaparashar:
## PR description

This PR adds EIP-1193 provider standard implementation of `.request` method to truffle's `provider` package.

### DIFF
--- a/packages/provider/wrapper.js
+++ b/packages/provider/wrapper.js
@@ -5,7 +5,7 @@ function make1193ProviderRequest(provider) {
   /*
   NOTE: If truffle switches to use WebSockets instead of HTTP requests in the future
         then we will need to keep separate requestId counters for .request and .send
-        methods to avoid the race condition on parallel requests.
+        methods to avoid collision while generating requestIds on parallel requests.
   */
   let requestId = 0;
   const modifiedRequest = async function ({ method, params }) {
--- a/packages/provider/wrapper.js
+++ b/packages/provider/wrapper.js
@@ -2,6 +2,11 @@ const debug = require("debug")("provider:wrapper"); // eslint-disable-line no-un
 const ProviderError = require("./error");
 
 function make1193ProviderRequest(provider) {
+  /*
+  NOTE: If truffle switches to use WebSockets instead of HTTP requests in the future
+        then we will need to keep separate requestId counters for .request and .send
+        methods to avoid the race condition on parallel requests.
+  */
   let requestId = 0;
   const modifiedRequest = async function ({ method, params }) {
     return await new Promise((accept, reject) => {
--- a/packages/provider/wrapper.js
+++ b/packages/provider/wrapper.js
@@ -1,6 +1,38 @@
 const debug = require("debug")("provider:wrapper"); // eslint-disable-line no-unused-vars
 const ProviderError = require("./error");
 
+function make1193ProviderRequest(provider) {
+  let requestId = 0;
+  const modifiedRequest = async function ({ method, params }) {
+    return await new Promise((accept, reject) => {
+      provider.send(
+        {
+          jsonrpc: "2.0",
+          id: ++requestId,
+          method,
+          params
+        },
+        (error, response) => {
+          if (error) {
+            return reject(error);
+          }
+          if (response.error) {
+            const error = new Error(response.error.message);
+            error.code = response.error.code;
+            if (response.error.data) {
+              error.data = response.error.data;
+            }
+            return reject(error);
+          }
+          const { result: res } = response;
+          accept(res);
+        }
+      );
+    });
+  };
+  return modifiedRequest.bind(provider);
+}
+
 module.exports = {
   /*
    * Web3.js Transport Wrapper
@@ -31,6 +63,11 @@ module.exports = {
     if (provider.sendAsync) {
       const originalSendAsync = provider.sendAsync.bind(provider);
       provider.sendAsync = this.send(originalSendAsync, preHook, postHook);
+    }
+
+    // EIP-1193 provider wrapping
+    if (provider.request === undefined && provider.send !== undefined) {
+      provider.request = make1193ProviderRequest(provider);
     }
     /* mark as wrapped */
     provider._alreadyWrapped = true;


packages/provider/wrapper.js
```diff
@@ -32,6 +32,43 @@ module.exports = {
...
+      typeof provider.request === "undefined" &&
+      typeof provider.send !== "undefined"
+    ) {
+      provider.request = async function ({ method, params }) {
+        let id = 0;
+        return await new Promise((accept, reject) => {
+          provider.send(
+            {
+              jsonrpc: "2.0",
+              id: ++id,
```
cds-amal, line 47: `id` should probably be declared outside the scope of `provider.request` because it will always be 1, at this point.

cds-amal (commented):
Thanks, @sukanyaparashar. Small PRs are GREAT!

packages/provider/wrapper.js
gnidan: probably better just to use the timestamp, because otherwise there's risk that someone will intersperse `.request` with `.send` and the IDs might collide.

packages/provider/wrapper.js
```diff
@@ -32,6 +32,43 @@ module.exports = {
       const originalSendAsync = provider.sendAsync.bind(provider);
       provider.sendAsync = this.send(originalSendAsync, preHook, postHook);
     }
+
+    // EIP-1193 provider wrapping
+    if (
+      typeof provider.request === "undefined" &&
+      typeof provider.send !== "undefined"
+    ) {
+      provider.request = async function ({ method, params }) {
```
gnidan, line 41: Is there any way to clean this wrapper up while you're at it? Not necessary, but I think we'd be a lot happier in the future if this whole function was divided into two equal parts (e.g., it could defer to separate wrapper functions for both legacy and EIP-1193 cases)

packages/provider/wrapper.js
davidmurdoch: It was changed from `timestamp` to `id` because timestamps are a) non-deterministic, and b) not granular enough -- this matters if we ever want to switch to websockets, which matches responses to requests by `id`.

gnidan: Ah thank you for the context @davidmurdoch! So we do have a race condition and need to track the IDs across `.send` and `.request` invocations?

davidmurdoch: If we switch to WebSockets _and_ send parallel requests, potentially yes. This is what the JSON-RPC `id` field is for. We could also use a UUID, but those are slower to generate and aren't deterministic (not that determinism is actually important here -- it's just a nice to have). In early development of forking websocket URL support I ran into this race condition.

HTTP doesn't care though, and that's all that Truffle can use internally today, last I checked, so this is only for a future where Truffle switches to using a WebSocket provider.

packages/provider/wrapper.js
sukanyaparashar: It is a nice idea to clean up the wrapper function. I will look into it. Thanks for the suggestion @gnidan.

packages/provider/wrapper.js
```diff
@@ -32,6 +32,43 @@ module.exports = {
       const originalSendAsync = provider.sendAsync.bind(provider);
       provider.sendAsync = this.send(originalSendAsync, preHook, postHook);
     }
+
+    // EIP-1193 provider wrapping
+    if (
+      typeof provider.request === "undefined" &&
```
eggplantzzz, line 38: I don't believe you need to do the `typeof` thing here. I think you can just do `provider.request === undefined`.

packages/provider/wrapper.js
```diff
@@ -32,6 +32,43 @@ module.exports = {
...
+      typeof provider.send !== "undefined"
+    ) {
+      provider.request = async function ({ method, params }) {
+        let id = 0;
```
fainashalts, line 42: Can `id` be something more descriptive? Like `requestId`?

packages/provider/wrapper.js
```diff
@@ -32,6 +32,43 @@ module.exports = {
...
+        return await new Promise((accept, reject) => {
```
fainashalts, line 43: Is there a reason we need to use `new Promise`? I prefer `async/await` syntax for async operations, to make the code more concise and easier to read. Perhaps something like: 

```
const response = await provider.send({
        jsonrpc: "2.0",
        id: ++requestId,
        method,
        params
      });
```

packages/provider/wrapper.js
sukanyaparashar: `requestId` makes sense to me.

packages/provider/wrapper.js
davidmurdoch: I think it is because `response` would be `undefined` here since `provider.send` uses callbacks

sukanyaparashar: I agree with @davidmurdoch here. `provider.send` uses callbacks and we are wrapping this function in a callback style way to use `await` for the response/error.

packages/provider/wrapper.js
```diff
@@ -32,6 +64,13 @@ module.exports = {
       const originalSendAsync = provider.sendAsync.bind(provider);
       provider.sendAsync = this.send(originalSendAsync, preHook, postHook);
     }
+
+    // EIP-1193 provider wrapping
+    if (provider.request === undefined && provider.send !== undefined) {
+      provider.request = make1193ProviderRequest(provider);
+    }
```
sukanyaparashar, line 71: @gnidan , @cds-amal and I thought of making a separate function for the 1193 type.

packages/provider/wrapper.js
```diff
@@ -32,6 +64,13 @@ module.exports = {
...
+    provider.request = provider.request.bind(provider);
```
haltman-at, line 72: Hm, is this line necessary uncondtionally?  It seems odd to include in the case where `provider.request` is already defined.  Should this be moved inside the conditional?  If so, perhaps `make1193ProviderRequest` could be modified to return `modifiedRequest.bind(provider)`, rather than setting it in two steps (it's nice to avoid mutation when it's reasonable to do so :) ).

haltman-at (approved):
This seems good to me.  I have one minor note but nothing I think is essential.

packages/provider/wrapper.js
```diff
@@ -1,6 +1,38 @@
...
+            }
+            return reject(error);
+          }
+          const { result: res } = response;
+          accept(res);
+        }
+      );
+    });
+  };
+  return modifiedRequest;
```
cds-amal, line 33:
```suggestion
  return modifiedRequest.bind(provider);
```

packages/provider/wrapper.js
```diff
@@ -1,6 +1,38 @@
 const debug = require("debug")("provider:wrapper"); // eslint-disable-line no-unused-vars
 const ProviderError = require("./error");
 
+function make1193ProviderRequest(provider) {
```
sukanyaparashar, line 4: @haltman-at @cds-amal @fainashalts @eggplantzzz , I ran into a couple of github issues on the branch. So, I had to undo the last commit. But this latest commit includes all the suggested changes.

haltman-at (approved):
(I already approved this, but I'll re-approve it for good measure :P )

packages/provider/wrapper.js
cds-amal: Were the issues related to GH not allowing CI to run? I've noticed this with long-lived PRs that lag behind develop, and you have to rebase against develop to enable CI.

sukanyaparashar: Yes, @cds-amal. CI was failing on @truffle/db package. I realised it that this branch needed a rebase against develop.

packages/provider/wrapper.js
```diff
@@ -1,6 +1,38 @@
...
+  let requestId = 0;
```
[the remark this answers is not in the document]
sukanyaparashar: In my opinion, since this will only matter when we use WebSockets instead of HTTP, we can keep things simple for now having only one `requestId` counter. And come back to it once we are done releasing the `truffle call` feature. I think I need some more clarity on this.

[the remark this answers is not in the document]
cds-amal: If that's the case, a note in the code wrt `requestId` would be prudent.

cds-amal (approved):
Looks good to me.

packages/provider/wrapper.js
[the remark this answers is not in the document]
sukanyaparashar: It makes sense to add a note in there to remind of any future work related to the `requestId ` to be done. I will do that, thanks @cds-amal .

cds-amal (approved):
I like it.

gnidan (approved):
Nice work! Excited to see this get into the code :D

Issue #5746, opened 2022-11-30
TS compilation errors in contract-schema


wbt:
In addition to issues duplicating #4825 (closed & I can't reopen, but for a different package) about types being messed up, in @truffle/contract-schema the [index.d.ts](https://npm.runkit.com/@truffle/contract-schema/spec/index.d.ts?t=1669818862275) file triggers 21 errors on compilation, related to possibly-undefined values not being assignable to a definitely-defined index type: 

```
node_modules/@truffle/contract-schema/spec/index.d.ts:79:3 - error TS2411: Property 'contractName' of type 'string | undefined' is not assignable to 'string' index type 'string | number | boolean | any[] | { [k: string]: any; }'.

79   contractName?: string;
     ~~~~~~~~~~~~

node_modules/@truffle/contract-schema/spec/index.d.ts:81:3 - error TS2411: Property 'metadata' of type 'string | undefined' is not assignable to 'string' index type 'string | number | boolean | any[] | { [k: string]: any; }'.

81   metadata?: Metadata;
     ~~~~~~~~

node_modules/@truffle/contract-schema/spec/index.d.ts:85:3 - error TS2411: Property 'bytecode' of type 'string | undefined' is not assignable to 'string' index type 'string | number | boolean | any[] | { [k: string]: any; }'.

85   bytecode?: string;
     ~~~~~~~~

node_modules/@truffle/contract-schema/spec/index.d.ts:89:3 - error TS2411: Property 'deployedBytecode' of type 'string | undefined' is not assignable to 'string' index type 'string | number | boolean | any[] | { [k: string]: any; }'.

89   deployedBytecode?: string;
     ~~~~~~~~~~~~~~~~

node_modules/@truffle/contract-schema/spec/index.d.ts:93:3 - error TS2411: Property 'sourceMap' of type 'string | undefined' is not assignable to 'string' index type 'string | number | boolean | any[] | { [k: string]: any; }'.

93   sourceMap?: string;
     ~~~~~~~~~

node_modules/@truffle/contract-schema/spec/index.d.ts:97:3 - error TS2411: Property 'deployedSourceMap' of type 'string | undefined' is not assignable to 'string' index type 'string | number | boolean | any[] | { [k: string]: any; }'.

97   deployedSourceMap?: string;
     ~~~~~~~~~~~~~~~~~

[...]

PR #5747, opened 2022-11-30
Attempt to fix types on contract-schema

wbt:
## PR description

This PR demonstrates a conceptual approach to closing #5746 by broadening the string-indexed type to allow for undefined, which is needed for the TypeScript type declaration file generator, as that generator marks all the specified "properties" above as optional.   Another strategy could be to not mark those specified properties as optional, but at first glance that seems more likely to break things.
 
I don't expect this to actually work because `undefined` isn't a primitive type in the vocabulary of the [instance data model](https://json-schema.org/draft/2020-12/json-schema-core.html#name-instance-data-model). 

## Testing instructions
Run steps involved in building/preparing for release and inspect contract-schema/spec/index.d.ts to see if it includes the `|undefined` possibility that the Issue notes fixes the issue when manually added in.

## Documentation

This should just broaden a type and not break existing code that uses narrower types, unless that code is somehow relying on the optional string-index value to be defined for any given string. 


wbt:
As expected, this failed validation. I think https://github.com/bcherny/json-schema-to-typescript/pull/497 is probably the better solution, but leave this draft here in case it is helpful in advancing the conversation toward a better solution.

haltman-at:
(I've retitled this to be more descriptive of its contents; feel free to change it again if I've gotten it wrong!)

eggplantzzz:
Thanks @wbt for the POC but I'll go ahead and close this for now since this might not be the right approach. We'll have to re-group and figure something else out eventually.

PR #5759, opened 2022-12-06
Remove Arbitrum Rinkeby and Optimistic Kovan from Etherscan fetcher

haltman-at:
It looks like Etherscan has removed support for Arbitrum Rinkeby; the old URL `testnet.arbiscan.io` is still there, sure, but now it's just a duplicate of `goerli.arbiscan.io`.  So, I'm removing support from this network from the Etherscan fetcher.  (Note that it's still supported by the Sourcify fetcher, since while Sourcify will disable new networks and stop letting you verify on them, it doesn't remove access to the already-verified contracts.)

Edit: Looks like Optimistic Kovan is gone too, removed that too.  (Along with the appropriate tests.)

Boba Rinkeby is also gone and replaced by Boba Goerli, but I'm going to save handling that for a separate PR as I need more information there.

### DIFF
--- a/packages/fetch-and-compile/test/fetch.test.ts
+++ b/packages/fetch-and-compile/test/fetch.test.ts
@@ -167,15 +167,6 @@ describe("Etherscan single-source Solidity case", function () {
     );
   });
 
-  it("verifies contract from optimism kovan", async function () {
-    await runTestBody(
-      69,
-      "0x5bb6699ef885ca997d1467380ff9e51c606a07e1",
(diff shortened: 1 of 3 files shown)


fainashalts:
LGTM but failing tests for some reason!

haltman-at:
Oh, duh!  It's failing tests because we had explicit tests for that network.  Guess I'd better remove those.

cds-amal (approved):
Gracias, looks good to me.

PR #5760, opened 2022-12-06
POC for getting fetch-and-compile functioning in the browser

eggplantzzz:
This is a POC for getting fetch-and-compile working in the browser. The webpack configs created for this are kind of hacky with all the ridiculous shimming and a very narrow code path is attempted to avoid using the filesystem.


packages/compile-solidity/src/compilerSupplier/index.ts
```diff
@@ -28,17 +40,22 @@ export class CompilerSupplier {
...
     const userSpecification = this.version;
 
     let strategy: CompilerSupplierStrategy;
     const useDocker = this.docker;
     const useNative = userSpecification === "native";
-    const useSpecifiedLocal =
-      userSpecification &&
-      (fs.existsSync(userSpecification) || path.isAbsolute(userSpecification));
+    let useSpecifiedLocal: boolean | string | undefined;
+    if (!userSpecification) {
```
eggplantzzz, line 53: There should be a better way to do this; currently this hack is here just to prevent disk access.

PR #5765, opened 2022-12-07
Fix: `truffle unbox` use `spawnSync` to run the post-install hook

GuillaumeRx:
## PR description

This replace `execSync` in favor of `spawnSync`  to run the post-install hook avoiding a `ENOBUFS` error due to Yarn 3 filling the 200 ko Buffer limit of the spawned process.

## Testing instructions

- Run the `truffle unbox metamask/snap-box` command
- Observe that the command succeeds.

## Demo

With a `npm` box:

https://user-images.githubusercontent.com/13910212/206447417-0debb5e4-b2a8-450a-afac-9c1e93db96f8.mp4

With a `yarn` box

https://user-images.githubusercontent.com/13910212/206448780-06e49e1a-58ee-4ab1-976e-4714140b8647.mp4

### DIFF
--- a/packages/box/lib/utils/unbox.ts
+++ b/packages/box/lib/utils/unbox.ts
@@ -4,7 +4,7 @@ import download from "download-git-repo";
 import axios from "axios";
 import vcsurl from "vcsurl";
 import { parse as parseURL } from "url";
-import { execSync } from "child_process";
+import { spawnSync } from "child_process";
 import inquirer from "inquirer";
 import type { Question } from "inquirer";
 import type { boxConfig, unboxOptions } from "typings";
@@ -148,7 +148,12 @@ function installBoxDependencies({ hooks }: boxConfig, destination: string) {
   const postUnpack = hooks["post-unpack"];
 
   if (postUnpack.length === 0) return;
-  execSync(postUnpack, { cwd: destination, stdio: "ignore" });
+
+  spawnSync(postUnpack, {
+    cwd: destination,
+    shell: true,
+    stdio: ["ignore", process.stdout, process.stderr]
+  });
 }
 
 export = {
--- a/packages/box/lib/utils/unbox.ts
+++ b/packages/box/lib/utils/unbox.ts
@@ -148,7 +148,7 @@ function installBoxDependencies({ hooks }: boxConfig, destination: string) {
   const postUnpack = hooks["post-unpack"];
 
   if (postUnpack.length === 0) return;
-  execSync(postUnpack, { cwd: destination });
+  execSync(postUnpack, { cwd: destination, stdio: "ignore" });
 }
 
 export = {


cds-amal:
Thanks, @GuillaumeRx!

cds-amal (dismissed):
Tested locally and LGTM! Thanks, @GuillaumeRx!

We need another approval, and CI passing ofc.

packages/box/lib/utils/unbox.ts
```diff
@@ -148,7 +148,7 @@ function installBoxDependencies({ hooks }: boxConfig, destination: string) {
   const postUnpack = hooks["post-unpack"];
 
   if (postUnpack.length === 0) return;
-  execSync(postUnpack, { cwd: destination });
+  execSync(postUnpack, { cwd: destination, stdio: "ignore" });
```
cliffoo, line 151: Would increasing `maxBuffer` be better? Ignoring stdio altogether hides warnings.

Or use spawn instead of exec.

cliffoo (commented):
Thanks @GuillaumeRx !

sukanyaparashar (dismissed):
Works locally! Thanks @GuillaumeRx.

packages/box/lib/utils/unbox.ts
cds-amal: Would prefer using spawn to get around the buffer issue. Thanks, @cliffoo

GuillaumeRx: Thanks for the review ! I'll push an update using `spawn` tomorrow 😃

cliffoo: Looks like parsing can be avoided by setting `shell` to `true` ([gist](https://gist.github.com/cliffoo/2bd76a69fbff35263e1fd32349d25d83))

GuillaumeRx: `execSync` will only output `stderr` to the parent process stderr. Should I replicate this behaviour with `spawn` ?

packages/box/lib/utils/unbox.ts
```diff
@@ -148,7 +148,12 @@ function installBoxDependencies({ hooks }: boxConfig, destination: string) {
   const postUnpack = hooks["post-unpack"];
 
   if (postUnpack.length === 0) return;
-  execSync(postUnpack, { cwd: destination, stdio: "ignore" });
+
+  spawnSync(postUnpack, {
+    cwd: destination,
+    shell: true,
+    stdio: ["ignore", process.stdout, process.stderr]
```
GuillaumeRx, line 155: Ignore `stdin` and pipe `stdout` / `stderr` to parent process. I also piped `stdout` because - after looking at some boxes - most of them use `--loglevel=error` to elevate the logging to `stderr`.

`execSync` was previously left at its default config that didn't pipe `stdout`, now that it is with `spawnSync` the log level flag is not required anymore.

cliffoo:
hhmvvsmmmmm so this prompted us to revisit the post-install hook workflow as a whole: the security implications of allowing box creators to execute arbitrary commands on the user end (e.g. `rm <path-to-ur-favorite-file>`) in a (sub)shell are grave. I think this PR should brew while we decide on a solution that thwarts malicious commands (with regex perhaps to define what's  allowed).

And for `metamask/snap-box` specifically, `yarn install --silent` should get around this problem for now.

Thanks again @GuillaumeRx :+1:

cds-amal (changes requested):
Hi @GuillaumeRx,

We decided that since this PR doesn't affect the vulnerability issue, we could merge it. However, not sanitizing inputs when using `spawnSync` and `shell`  is a new concern, e.g. expanding `*` and `$` in the shell. To be honest, I'm unsure how to proceed and need some time to research this.

OnlyOneJMJQ:
Update since we've discussed this internally. This PR can be merged once we restrict use of the `post-install` hook to only those boxes we control (those in the `truffle-box` repo). That way we can be sure no malicious code is being executed, but leave flexibility for our trusted first-party boxes.

[Tracking this work in a GitHub project here.](https://github.com/orgs/trufflesuite/projects/33/views/1)

Issue #3777, opened 2021-02-07
Broken TypeScript typings on fresh install


Butterneck:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue.

---------------------------

## Issue

After a fresh install of `@truffle/hdwallet-provider` (other packages possibly affected as well) TypeScript typings are not working correctly and compilation fails.

## Steps to Reproduce

1. `npm install @truffle/hdwallet-provider web3`
2. Put inside `src/index.ts`:
```js
import HDWalletProvider from "@truffle/hdwallet-provider";
console.log("Hello world!")
```
3. `npx typescript --esModuleInterop src/index.ts`

## Expected Behavior

Compilation succeeds without any issues.

## Actual Results

Typecheck fails with the following errors:
```
node_modules/@truffle/hdwallet-provider/dist/index.d.ts:2:61 - error TS2307: Cannot find module 'ethereum-protocol'.
import { JSONRPCRequestPayload, JSONRPCErrorCallback } from "ethereum-protocol";
                                                             ~~~~~~~~~~~~~~~~~~~
node_modules/@truffle/hdwallet-provider/dist/index.d.ts:3:43 - error TS2307: Cannot find module 'web3/providers'.
import { Callback, JsonRPCResponse } from "web3/providers";
                                           ~~~~~~~~~~~~~~~~
```

## Environment

* Operating System: Tested on Arch Linux
* Ethereum client:
* Truffle version (`truffle version`): `Truffle v5.1.65 (core: 5.1.65)
Solidity v0.5.16 (solc-js)
Node v15.7.0
Web3.js v1.2.9`
* node version (`node --version`): `v15.7.0`
* npm version (`npm --version`): `7.4.3`

gnidan:
Yikes! We'll have to look into this. Hopefully this is as simple as adding `skipLibCheck` to hdwallet-provider's tsconfig.

Thanks for reporting!

cds-amal:
@gnidan looks like a `npm 7`, comes with node 15 by default

binzydev:
> @gnidan looks like this may be related to `npm 7`, which comes with node 15 by default

Same problem with NPM 6 and node 14.16.0 here

cds-amal:
Hi @binzydev,

We might have to move some types from `devDependencies` to `dependencies` so that they make it to the declarations file. As a workaround you can manually install them.

```console
[...]

PR #5767, opened 2022-12-07
Bug fix: fix hdwallet types

eggplantzzz:
## PR description

In https://github.com/trufflesuite/truffle/issues/3777 it was reported that hdwallet-provider could not be included in TS projects because the compiler would choke on some missing types. This turned out to be both because it was using the wrong version of @types/web3 and did not contain some necessary types packages in the regular dependencies section. The types were being exposed but were only listed under `devDependencies` in the `package.json`.

This PR pins @types/web3 in all of Truffle to `1.0.20` (which was sometimes resolved to `1.2.2` which is a stub package containing no types). It also moves a couple of dependencies into the `dependencies` section of `package.json` as those types are exposed in that package.

## Testing instructions

If you would like to test this you can do the following:
- create a new npm project (`npm init -y`) 
- create a tarball using `npm pack` in hdwallet-provider after checking out this PR's latest commit and bootstrapping
- add an `index.ts` that imports hdwallet-provider
- set up a rudimentary `tsconfig.json` and install TS
- move the tarball into your new project's directory and install from it `npm install -S <yourTarballFilename>`
- compile the project using typescript

## Documentation


benjamincburns:
This might be a dumb question, but why do we need `@types/web3` at all now that `web3` ships with types included?

eggplantzzz:
I think the types are broken...at one point they were. I'll try removing them and see if it breaks.

eggplantzzz:
Ah, also web3 doesn't contain those provider types we use that are included in @types/web3.

benjamincburns:
> Ah, also web3 doesn't contain those provider types we use that are included in @types/web3.

If that's the only missing piece, I'd be an advocate of providing our own version of those via `@truffle/provider` along with a PR back to web3 that exposes good types for providers.

eggplantzzz:
Yeah I'd be into getting rid of that dep: even if it means simplifying the types.

haltman-at:
> This might be a dumb question, but why do we need `@types/web3` at all now that `web3` ships with types included?

Yeah while I don't know the current status, for quite a while at least it was the case that the types included in `web3` were simply wrong and unusable, and so you still had to use `@types/web3` despite `web3` notionally including types.  This may be fixed now, though; I haven't checked.

haltman-at (approved):
Makes sense to me as a quick fix, but as @benjamincburns we should really review whether we still need `@types/web3` at all.  I guess I'll open an issue for that?

benjamincburns:
Yeah all good, guys - glad you didn't interpret my comments as blocking this merge!

Issue #5749, opened 2022-12-01
Encoder doesn't specify any easy way to encode create transactions


gnidan:
## Issue

@truffle/encoder's `ContractEncoder` and `ContractInstanceEncoder` currently provide a means of encoding creates via their `encodeTxNoResolution`, which requires specifying the constructor ABI entry. Since the encoder already has this information, and since the caller very well might not :), there ought to be an easier way!

(Aside: would we also want to provide an easy way to do `CREATE2`?)

## Steps to Reproduce

Look through @truffle/encoder docs.

## Expected Behavior

Find method to make it easy to encode a create transaction.

## Actual Results

Find only `encodeTxNoResolution`, which works, but requires that you pass the ABI around along with the encoder.

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.6.8
* node version (`node --version`):
* npm version (`npm --version`):

haltman-at:
> (Aside: would we also want to provide an easy way to do `CREATE2?`)

`CREATE2` isn't something you can do as a transaction; it only exists as an opcode.

gnidan:
> `CREATE2` isn't something you can do as a transaction; it only exists as an opcode.

You can if you do it as a `CREATE` transaction that deploys a contract that invokes `CREATE2` :)

haltman-at:
OK but that is obviously not a concern for the encoder...

gnidan:
I wouldn't say "obviously". Arguably it's not, but I think there's an argument that it could be!

haltman-at:
Released in 5.7.0.

PR #5777, opened 2022-12-10
Add `encodeCreation()` method to Encoder

haltman-at:
Addresses #5749.

This PR adds an `encodeCreation()` method to the `ContractEncoder` and `ContractInstanceEncoder` classes.  (Using it from the latter ignores the particular instance; it's included for consistency.)

The encoder already had the functionality of encoding contract creations, but accessing it wasn't convenient; you had to pass in the ABI for the constructor, and make one up if none existed.  This PR just factors out the relevant functionality to make it accessible.  It adds an `encodeCreation()` method to the contract encoder (and also contract instance encoder); it doesn't take an `abi` argument because none is necessary.

The existing functionality for getting constructor information, which used to just live in the constructor case of `getMethod()`, has been factored out into `getConstructorMethod()`.  Note that this method accepts an optional ABI argument with the constructor ABI; this is just to avoid re-finding it in cases where it's already known.  The new `encodeCreation()` method doesn't pass this in, but when it's being called from `getMethod(abi)`, it will pass in this argument to avoid recomputation.  In the case where `abi` is not passed in, `getConstructorMethod()` will call `getConstructorAbi()`, which will find the relevant ABI entry if it exists and use a default one if it doesn't.

I modified the existing test of constructor encoding to use this new method.  I also added one more, which tests the case of a default constructor; we didn't have a test for that case before.

Testing instructions: The tests I added to CI should be sufficient, I think, but let me know if you think they're not...

### DIFF
--- a/packages/encoder/lib/encoders.ts
+++ b/packages/encoder/lib/encoders.ts
@@ -1597,4 +1597,17 @@ export class ContractInstanceEncoder {
     encoded.tx.to = this.toAddress;
     return encoded;
   }
+
+  /**
+   * **This method is asynchronous.**
+   *
+   * This method functions identically to [[ContractEncoder.encodeCreation]].
+   * The particular contract instance is ignored, only its class is used.
+   */
+  public async encodeCreation(
+    inputs: unknown[],
+    options: Types.ResolveOptions = {}
+  ): Promise<Codec.Options> {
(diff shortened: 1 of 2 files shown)


gnidan (approved):
Sweet!

PR #5780, opened 2022-12-12
remove `require` cache invalidation

eggplantzzz:
THIS IS A BREAKING CHANGE FOR @truffle/config.

So https://github.com/trufflesuite/truffle/pull/5728 brought to our attention the fact that this cache invalidation does not work in the bundle (since `require` in this case is webpack's and not Node's). This PR deletes this line of code until we decide that we actually need to invalidate Node's cache.

### DIFF
--- a/packages/core/test/lib/commands/config.js
+++ b/packages/core/test/lib/commands/config.js
@@ -11,12 +11,9 @@ describe("config", function () {
   let output = "";
   let memStream;
 
-  before(function () {
+  beforeEach(function () {
     config = createTestProject(path.join(__dirname, "../../sources/metacoin"));
     config.logger = { log: val => val && memStream.write(val) };
-  });
-
-  beforeEach(function () {
     memStream = new MemoryStream();
     memStream.on("data", data => {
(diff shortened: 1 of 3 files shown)


haltman-at:
(I changed the labels on this, the "requires-breaking-change" label is for issues that would require a Truffle breaking change; PRs that break a particular package should be marked with the appropriate label for that package.)

packages/core/test/lib/commands/config.js
```diff
@@ -11,12 +11,9 @@ describe("config", function () {
   let output = "";
   let memStream;
 
-  before(function () {
+  beforeEach(function () {
```
cds-amal, line 14: Clearing the cache allowed this to pass before?

cds-amal (approved):
Looks good to me!

packages/core/test/lib/commands/config.js
eggplantzzz: I think so - I think the complications arose from the order in which `Config` merges in changes in `load`

PR #5787, opened 2022-12-15
Change display name of @truffle/error and remove its stack property

cds-amal:
When using the bundle, I noticed `@truffle/error` includes a stack trace, unlike the unbundled behavior. @truffle/error's intent is to give the user an action and does not need a stack trace.

## Error with stack trace
```console
Solidity console.log detected in the following assets:
Telephone.json

ExtendableError: You are trying to deploy contracts that use console.log.
Please fix, or disable this check by setting solidityLog.preventConsoleLogMigration to false

    at module.exports (/home/amal/work/trees/truffle/console.log/packages/truffle/build/webpack:/packages/core/lib/commands/migrate/runMigrations.js:87:1)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at module.exports (/home/amal/work/trees/truffle/console.log/packages/truffle/build/webpack:/packages/core/lib/commands/migrate/setUpDryRunEnvironmentThenRunMigrations.js:37:1)
    at Object.module.exports [as run] (/home/amal/work/trees/truffle/console.log/packages/truffle/build/webpack:/packages/core/lib/commands/migrate/run.js:33:1)
    at runCommand (/home/amal/work/trees/truffle/console.log/packages/truffle/build/webpack:/packages/core/lib/command-utils.js:201:1)
Truffle v5.6.9 (core: 5.6.9)
Node v18.12.1
```

## How to test
1. bootstrap the repo
2. unbox metacoin, config and run migrate with mutually exclusive `--url` and `--network` options with truffle bundle.
   ```console
   reprod
   use-truffle-bundle # if you have the truffle dotfiles setup!
   truffle unbox metacoin
   truffle console --url http://127.0.0.1 --network development
   ```
3. There should be no stack trace! (Make sure you use the bundle)

### DIFF
--- a/packages/error/src/index.ts
+++ b/packages/error/src/index.ts
@@ -1,11 +1,9 @@
-//Note: This class only exists for compatibility with some old Javascript
-//stuff that avoided using Error directly for whatever reason.  Eventually
-//it should be eliminated.
-class ExtendableError extends Error {
+class TruffleError extends Error {
   constructor(message: string) {
     super(message);
     this.name = this.constructor.name;
+    this.stack = "";
   }
 }
 
-export = ExtendableError;
+export = TruffleError;


sukanyaparashar (approved):
This looks good, @cds-amal . Tested it locally. While checking out to your branch, I needed to do `yarn bootstrap` first for changes to reflect in the `error/dist/src` folder. May be you want to mention it in the test instructions too.

dongmingh (commented):
The test result displays a helpful message and no more stack trace: 
```
$ truffle console --url http://127.0.0.1 --network development
Mutually exclusive options, --url and --network detected!
Please use either --url or --network and try again.
See: https://trufflesuite.com/docs/truffle/reference/truffle-commands/#console
```

packages/error/src/index.ts
```diff
@@ -1,11 +1,9 @@
-//Note: This class only exists for compatibility with some old Javascript
-//stuff that avoided using Error directly for whatever reason.  Eventually
-//it should be eliminated.
-class ExtendableError extends Error {
+class TruffleError extends Error {
```
eggplantzzz, line 1: can we just get rid of this package entirely? the comments that you removed seem to imply that we don't need it any more. we don't really use it that much, do we?

cds-amal: I don't think we can remove this because [`@truffle/error` is used](https://github.com/trufflesuite/truffle/blob/develop/packages/core/cli.js#L98) in situations when we know exactly why an error condition exists and should inform the user to take a specific action. For example, if they use `truffle migrate --url xyz --network falafel`, this is a truffle error condition, and we can use `@truffle/error` to let the user know they shouldn't use mutually exclusive options in the `migrate` command. In this situation, showing a stack trace would be confusing.

eggplantzzz: so you are saying you think this package is still useful - I would like to know exactly when to use this type of error in the future. would the `testConnection` method [here](https://github.com/trufflesuite/truffle/blob/develop/packages/provider/index.js#L36-L65) be an appropriate spot to utilize this error type?

cds-amal: That does seem like a good place to utilize `@truffle/error` because there's enough information to give the user a corrective action.

I'm not sure about `ProviderError`, and would have to check how it's used elsewhere.

eggplantzzz: I think we should have some kind of standard policy for `TruffleError` - I never use this at all and have no concept of when it is correct to do so other than what you've written here on this PR

PR #5788, opened 2022-12-15
don't load Truffle's console contract into the console environment

eggplantzzz:
This PR stops Truffle from loading its console contract into the console environment. Previously Truffle would load it into the console context (which users will not need to directly interact with in the repl) which would also cause the namespace error to be printed.

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -263,16 +263,17 @@ class Console extends EventEmitter {
           "utf8"
         );
         const json = JSON.parse(body);
-        // hack to prevent a warning about the console.sol contract which we
-        // don't load into the environment - we only want to do this when it is
-        // Truffle's console.sol, that is why we check the sources keys
+        const metadata = JSON.parse(json.metadata);
+        // filter out Truffle's console.log. We don't want users to interact with in the REPL.
+        // user contracts named console.log will be imported, and a warning will be issued.
         if (
-          !Object.keys(JSON.parse(json.metadata).sources).some(source => {
+          !Object.keys(metadata.sources).some(source => {
             return (
               source === "truffle/console.sol" ||
               source === "truffle/Console.sol"
             );
-          })
+          }) &&
+          Object.keys(metadata.sources).length === 1
         ) {
           jsonBlobs.push(json);
         }
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -262,7 +262,20 @@ class Console extends EventEmitter {
           path.join(this.options.contracts_build_directory, file),
           "utf8"
         );
-        jsonBlobs.push(JSON.parse(body));
+        const json = JSON.parse(body);
+        // hack to prevent a warning about the console.sol contract which we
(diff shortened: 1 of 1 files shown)


packages/core/lib/console.js
```diff
@@ -262,7 +262,20 @@ class Console extends EventEmitter {
           path.join(this.options.contracts_build_directory, file),
           "utf8"
         );
-        jsonBlobs.push(JSON.parse(body));
+        const json = JSON.parse(body);
+        // hack to prevent a warning about the console.sol contract which we
+        // don't load into the environment - we only want to do this when it is
+        // Truffle's console.sol, that is why we check the sources keys
```
cds-amal, line 268: I don't think this is a Hack, it is well reasoned logic :) 

```suggestion
        // Filter out Truffle's console.log. We don't want users to interact with in the REPL.
        // User contracts named console.log will be imported, and a warning will be issued.
```

cds-amal (approved):
I like it! Thanks, @eggplantzzz. I left a suggestion for the comment, which I don't think is a hack, but your call.

packages/core/lib/console.js
```diff
@@ -262,7 +262,20 @@ class Console extends EventEmitter {
...
+        if (
+          !Object.keys(JSON.parse(json.metadata).sources).some(source => {
+            return (
+              source === "truffle/console.sol" ||
+              source === "truffle/Console.sol"
```
cds-amal, line 273: A contract that references `truffle/console.sol` will have it listed in the `metadata.sources` field as well. Maybe this should be changed to sources.length === 1 and it matches "truffle/[cC]onsole.sol"

packages/core/lib/console.js
```diff
@@ -262,7 +262,21 @@ class Console extends EventEmitter {
...
+        // filter out Truffle's console.log. We don't want users to interact with in the REPL.
+        // user contracts named console.log will be imported, and a warning will be issued.
+        if (
+          !Object.keys(metadata.sources).some(source => {
+            return (
+              source === "truffle/console.sol" ||
+              source === "truffle/Console.sol"
+            );
+          }) &&
+          Object.keys(metadata.sources).length === 1
```
cds-amal, line 276: This seems inverted...

```suggestion
          Object.keys(metadata.sources).length > 1
```

cds-amal: Is there an optimization opportunity here? 

```
        const { sources } = JSON.parse(json.metadata);
        const sourcesLength = Object.keys(sources).length;
        if (sourcesLength > 1 || sourcesLength === 1 &&
          !Object.keys(sources).some(source => {
            return (
              source === "truffle/console.sol" ||
              source === "truffle/Console.sol"
            );
          })
        ) {
          jsonBlobs.push(json);
        }
```

eggplantzzz:
ah thanks, yes I inverted the logic trying to get this through quickly

cds-amal (approved):
I like it, really! Thanks, @eggplantzzz

PR #5789, opened 2022-12-15
Update Ganache version to 7.7.0

haltman-at:
Ganache 7.7.0 is out with Paris support, we should update to it!


davidmurdoch:
Maybe let this version sit until the after the new year. We may have discovered an installation issue on Windows.

haltman-at:
Oh hm, looks like there's a CI issue that will need to be sorted out too.

...what do you think of updating to 7.6.0?

davidmurdoch:
7.6.0 should be easy peasy!

PR #5790, opened 2022-12-15
Bump and pin @ganache/console.log

cds-amal:
Bump [`@ganache/console.log` to v0.3.0](https://github.com/trufflesuite/ganache/tree/develop/src/chains/ethereum/console.log#readme). I also pinned the dependency, where it was using `^`, whoops.

### DIFF
--- a/packages/resolver/package.json
+++ b/packages/resolver/package.json
@@ -27,7 +27,7 @@
   },
   "types": "dist/lib/index.d.ts",
   "dependencies": {
-    "@ganache/console.log": "0.2.0",
+    "@ganache/console.log": "0.3.0",
     "@truffle/compile-solidity": "^6.0.54",
     "@truffle/contract": "^4.6.10",
     "@truffle/contract-sources": "^0.2.0",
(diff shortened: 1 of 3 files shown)


haltman-at:
Why does this need to be pinned, I have to ask?

eggplantzzz (approved):
i like

dongmingh (commented):
The changes look good.  I like to know the answer of @haltman-at question: why does this need to be pinned?

cds-amal:
> Why does this need to be pinned, I have to ask?

We already pin `ganache`, and I am under the impression Ganache's release process, auto-patch-bumps `@ganache/console.log`. So, to be safe, I pinned as I would other pre v1.0 release packages. Please fact-check me on the Ganache release process.

haltman-at:
> I pinned as I would other pre v1.0 release package.

Is this something we're doing in general now?

cds-amal:
> > I pinned as I would other pre v1.0 release package.
> 
> Is this something we're doing in general now?

We don't have this general policy, but maybe we should consider it? I do it because of [Semver spec-item-4](https://semver.org/#spec-item-4)

PR #5791, opened 2022-12-15
Update Ganache to 7.6.0

haltman-at:
Since #5789 had problems, doing this (hopefully non-problematic) update first!


eggplantzzz (approved):
berry good

PR #5792, opened 2022-12-16
Add notice for @truffle/config and @truffle/db-loader

cds-amal:
Truffle's monorepo contains several packages that are strictly for internal consumption. We should declare they won't follow semver. This came up when we made a breaking change to `@truffle/config` that doesn't impact production because of webpack + original-require.

Work in progress.


cds-amal:
Closing, until we prioritize our Semver policies.

PR #5794, opened 2022-12-16
bump web3 libs to 1.8.1

eggplantzzz:


packages/hdwallet-provider/test/provider.test.ts
```diff
@@ -49,18 +50,18 @@ describe("HD Wallet Provider", function () {
 
       const mnemonic =
         "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat";
-      provider = new HDWalletProvider(mnemonic, ganacheProvider);
+      hdWalletProvider = new HDWalletProvider(mnemonic, ganacheProvider);
 
-      assert.deepEqual(provider.getAddresses(), truffleDevAccounts);
-      web3.setProvider(provider);
+      assert.deepEqual(hdWalletProvider.getAddresses(), truffleDevAccounts);
+      web3.setProvider(hdWalletProvider as provider);
```
cds-amal, line 56: Can you add a comment explaining why this cast is necessary? Even though the cast is repeated multiple time, I think having one commented would be helpful.

cds-amal (approved):
I like it! I think there should be a comment to explain why we had to cast the `provider` for our future selves, but leave that up to you.

packages/hdwallet-provider/test/provider.test.ts
eggplantzzz: I mean it should be pretty self-explanatory, right? the comment will read "this is necessary because instances of hdwallet provider don't conform to web3's `provider` type - we wouldn't need to do this otherwise...I mean I'll happily add it if you think that it isn't obvious

cds-amal: 🤔, yea, you're right. It's not adding value.

Issue #5812, opened 2022-12-31
Truffle console command throws error trying to read json files built from Vyper source files.


dPreininger:
- [ x] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

Truffle  console command crashes on .json files generated from Vyper source files. Not all versions of Truffle/Vyper compiler seem to be affected.

## Steps to Reproduce

Be sure you are using Ubuntu based OS with Bash terminal.
Clone this repository: https://github.com/dPreininger/vyper-truffle-report
Run `bash run.sh`. This will run included build files and return an error (read README for more details).

## Expected Behavior

Truffle console command should not exit.

## Actual Results

Truffle console exits with this error message:
```console
Unexpected error setting up the environment or provisioning contracts while instantiating the console.
Error: Error parsing or reading VyperStorage.json: Unexpected token u in JSON at position 0
    at {{ MY_HOME_DIR }}/.nvm/versions/node/v18.12.1/lib/node_modules/truffle/build/webpack:/packages/core/lib/console.js:283:1
    at Array.forEach (<anonymous>)
    at Console.provision ({{ MY_HOME_DIR }}/.nvm/versions/node/v18.12.1/lib/node_modules/truffle/build/webpack:/packages/core/lib/console.js:256:1)
    at Console.start ({{ MY_HOME_DIR }}/.nvm/versions/node/v18.12.1/lib/node_modules/truffle/build/webpack:/packages/core/lib/console.js:130:1)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Object.module.exports [as run] ({{ MY_HOME_DIR }}/.nvm/versions/node/v18.12.1/lib/node_modules/truffle/build/webpack:/packages/core/lib/commands/console/run.js:32:1)
    at runCommand ({{ MY_HOME_DIR }}/.nvm/versions/node/v18.12.1/lib/node_modules/truffle/build/webpack:/packages/core/lib/command-utils.js:201:1)
```
The error is caused by console.js file (https://github.com/trufflesuite/truffle/blob/develop/packages/core/lib/console.js), line 266.
The code expects the VyperStorage.json file to include metadata key. However, Vyper/Truffle does not generate this data, hence we are calling JSON.parse(undefined), which throws an error.

I am willing to contribute to this repository to fix this issue, however I am unsure what exactly is the correct behavior here. It seems to me there are two possibilities:

[...]

PR #5819, opened 2023-01-04
Fix: `truffle console` crash when metadata is missing in JSON artifact

dPreininger:
## PR description

As per the issue [#5812 ](https://github.com/trufflesuite/truffle/issues/5812) , `truffle console` command would throw an error and exit, if the metadata field was not present in all JSON blob files generated in the build process. The metadata field is not present in JSON files derived from Vyper source files.

## Testing instructions

1. Either compile a Vyper smart contract or delete metadata field in one of the generated JSON blobs from Solidity smart contracts.
2. Run `truffle migrate`. IMPORTANT: do not run truffle develop, as some configurations (truffle-config.js) were not affected by this bug (see issue for more details).
3. Run `truffle console`. The command executes without errors.

## Documentation, breaking changes and new features

This PR does not introduce any breaking changes or new features. The documentation does not need to be changed either.

### DIFF
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -263,7 +263,8 @@ class Console extends EventEmitter {
           "utf8"
         );
         const json = JSON.parse(body);
-        // Vyper contracts may not have metadata field included, just push them to json blobs
+        // Artifacts may not contain metadata. For example, early Solidity versions as well as
+        // Vyper contracts do not include metadata. Just push them to json blobs.
         if (json.metadata === undefined) {
           jsonBlobs.push(json);
         } else {
--- a/packages/core/lib/console.js
+++ b/packages/core/lib/console.js
@@ -268,10 +268,10 @@ class Console extends EventEmitter {
         if (json.metadata === undefined) {
           jsonBlobs.push(json);
         } else {
-          const metadata = JSON.parse(json.metadata);
-          const sources = Object.keys(metadata.sources);
           // filter out Truffle's console.log. We don't want users to interact with in the REPL.
           // user contracts named console.log will be imported, and a warning will be issued.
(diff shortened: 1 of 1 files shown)


packages/core/lib/console.js
```diff
@@ -263,21 +263,26 @@ class Console extends EventEmitter {
...
-          sources.length > 1 ||
-          (sources.length === 1 &&
-            !sources.some(source => {
-              return (
-                source === "truffle/console.sol" ||
-                source === "truffle/Console.sol"
-              );
-            }))
-        ) {
+        // Vyper contracts may not have metadata field included, just push them to json blobs
```
cds-amal, line 266:
```suggestion
        // Artifacts may not contain metadata. For example, early Solidity versions as well as
        // Vyper contracts do not include metadata. Just push them to json blobs.
```

cds-amal (commented):
Thanks for this, @dPreininger. There are two small comment nits I'd like updated.

dPreininger:
I have committed the suggestions.

cds-amal (approved):
Thanks for finding and fixing this bug, @dPreininger !

eggplantzzz (approved):
this seems like it does the job - I wonder what limitations we have in using these artifacts when they don't contain `metadata`

dPreininger:
@eggplantzzz I have actually looked through the code and found only 2 instances of metadata being used (besides this console.js file):
packages/codec/lib/compilations/utils.ts
packages/compile-solidity/src/run.ts

and 2 more test files:
packages/compile-solidity-tests/test/metadata.ts -> this is a test file
packages/contract-schema/test/solc.js -> this is a test file

I might have missed some more usages, but it seems to me that it was only really used to filter out Console contract.

Issue #5793, opened 2022-12-16
Codec docs no longer build


haltman-at:
## Issue

Codec docs aren't building.  Attempting yields the following error message:
```
Error: The plugin typedoc-plugin-remove-references could not be loaded.
Error: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './dist/lib/converter' is not defined by "exports" in /home/sniffnoy/truffle/truffle/node_modules/typedoc/package.json
    at new NodeError (internal/errors.js:322:7)
    at throwExportsNotFound (internal/modules/esm/resolve.js:332:9)
    at packageExportsResolve (internal/modules/esm/resolve.js:565:3)
    at resolveExports (internal/modules/cjs/loader.js:450:36)
    at Function.Module._findPath (internal/modules/cjs/loader.js:490:31)
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:888:27)
    at Function.Module._load (internal/modules/cjs/loader.js:746:27)
    at Module.require (internal/modules/cjs/loader.js:974:19)
    at require (internal/modules/cjs/helpers.js:101:18)
    at Object.<anonymous> (/home/sniffnoy/truffle/truffle/node_modules/typedoc-plugin-remove-references/dist/index.js:5:21)
error Command failed with exit code 1.
```

No idea what's going wrong.  I tried updating typedoc just in case that would help, it didn't.  Also apparently current typedoc versions don't require an explicit `plugin` line in their config...?  I actually tried removing the `plugin` line to see what would happen, and, indeed, the error was exactly the same!

## Environment

* Truffle version (`truffle version`): 5.7.0

cliffoo:
Fixed in [v5.7.3](https://github.com/trufflesuite/truffle/releases/tag/v5.7.3). Closing.

PR #5820, opened 2023-01-06
Fix codec typedoc build script

cds-amal:
## PR description

This PR makes minimal changes to get the codec docs to build. It now uses the default `typedoc` theme instead of our [fork](https://github.com/trufflesuite/typedoc-default-themes) which should probably incorporate [upstream](https://github.com/TypeStrong/typedoc-default-themes) changes. Upstream has been deprecated and is no longer compatible with `typedoc` v0.22. Truffle is currently using pinned `typedoc` version v0.22.18. 

Issue: #5793

Remove:
  - theme @trufflesuite/typedoc-default-themes
  - typedoc-plugin-remove-references

NOTE: can probably bump `typedoc` to latest, and if we want to keep our customized theme, will have to rework the changes against upstream as `typedoc@0.22.0` broke all hbs custom themes. See: https://github.com/TypeStrong/typedoc/blob/master/internal-docs/custom-themes.md

## Testing instructions

Build the codec docs and view them? 

## Documentation

### DIFF
--- a/packages/codec/docs/typedoc.json
+++ b/packages/codec/docs/typedoc.json
@@ -1,8 +1,6 @@
 {
   "name": "Truffle Decoding and Encoding",
-  "theme": "../../node_modules/@trufflesuite/typedoc-default-themes/bin/default/",
   "excludeExternals": true,
-  "plugin": ["typedoc-plugin-remove-references"],
   "categoryOrder": [
     "Data",
     "ABI data location",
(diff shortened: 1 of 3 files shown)


gnidan (approved):
Tested this out locally and seems to be sufficient for an initial fix! It's a bit annoying that we lose the ability to control how we display things like re-exported symbols, but definitely everything seems fine to get the build working again.

One thought: should we include the docs build as part of the codec `yarn prepare`? That would let us know sooner in the future if docs fail to build.

cds-amal:
> One thought: should we include the docs build as part of the codec yarn prepare? That would let us know sooner in the future if docs fail to build.

I like it

cds-amal:
Weeeeel, still a good idea to generate docs. Unfortunately, will have to be in another PR because of dependency issues.

Issue #5821, opened 2023-01-07
Build typedoc docs as part of prepare lifecycle.


cds-amal:
Codec's typedoc build has failed silently and undetected for a while. We should modify bootstrap to build codec's as well as other generated docs as part of the `prepare` lifecycle.

@gnidan suggested this to make the error detectable in #5820 review.

eggplantzzz:
released in v5.7.3

PR #5823, opened 2023-01-09
Build codec and preserve docs as part of bootstrap

cds-amal:
## PR description

This PR extends the build process to generate documentation for both codec and preserve. If either fails, we will be alerted and can correct at the time of failure. 

First call `yarn prepare` and allow each package to build, then call `yarn build docs` to generate typedocs for built packagers, avoiding dependency build issues.

Motivation: #5821 

## Testing instructions

Build and verify codec and preserve docs are created.

## Documentation

- No label `doc-change-required`

-  No breaking changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "private": true,
   "scripts": {
     "bootstrap": "yarn exec lerna bootstrap",
-    "prepare": "lerna run prepare --stream --concurrency=1 && husky install",
+    "docs": "lerna run docs --stream --concurrency=1",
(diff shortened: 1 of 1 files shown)


cds-amal:
@dongmingh This PR introduces a build step to alert us if doc generation fails. Previously, docs failed to build and we had no way of knowing when. See #5821

lsqproduction:
In the PR descriptions, it said `yarn build docs`, do you mean `yarn run docs`?

eggplantzzz (approved):
i like it

dongmingh (approved):
@cds-amal Thanks for explanations.  The changes look good.

PR #5825, opened 2023-01-09
Remove unused dev dependency for codec docs

haltman-at:
#5820 made this devdep unused but didn't actually remove it, putting up a quick PR to do so.


cds-amal (approved):
Thanks, @haltman-at.  LGTM.

PR #5290, opened 2022-07-13
Dashboard analytics

fainashalts:
This PR creates a frontend interface for asking users whether they are willing to share their usage data with us, and upon confirmation updates their user config appropriately.

Issue #4828, opened 2022-03-09
Truffle Dashboard should politely ask users if they'll graciously share anonymous usage data with us


gnidan:
## Issue

Truffle already provides an opt-in analytics system, but it requires running a specific command to turn it on. We know there are users who would be more than happy to let us collect these anonymous stats, but we haven't wanted to waste precious CLI screen real estate just to pester people about this.

Now, with Truffle Dashboard, there are likely unobtrusive ways to leverage the significantly increased display resolution and put this ask on-screen in a way that encourages users to opt-in without being annoying noodges about it. This would allow us to collect more information about how people use Truffle, what errors they commonly face, etc... ultimately giving us on the team the power to make informed decisions about what to work on next.

Couple thoughts:
- We should only ask the user once, or at most maybe once every six months. Maybe we could even check back with opted-in users, like, "hey, you still cool to share this data with us?"
- If a user declines to opt-in, could we ask a follow-up question, like, "would you mind if we phone home to say you did not opt-in?" I know some applications do telemetry to report this situation without user permission, but that's not really cool... it'd be nice to understand what the opt-in rate is, but not if it means loss of consent in the process.

## Environment

* Operating System: 
* Ethereum client:
* Truffle version (`truffle version`): v5.5.3
* node version (`node --version`):
* npm version (`npm --version`):

MicaiahReid:
Some notes:
 - Would need to add a message type to message bus
 - The front end would need to issue request to get the user config to see if they've responded before
   - (we may need to store user config info in local storage and deicde how to do that)
 - If user hasn't responded, somehow show a prompt asking them (politely) for their data
   - If yes, we issue a request to MB saying they've opted in, MB needs to shell out to tuffle that they've opted in (we're only using truffle's current analytics)
   - If no, there's no current opt-out in the truffle user config (either in or not responded). We need to modify the prompt to actually allow them to opt out

cliffoo:
Done in [v5.7.3](https://github.com/trufflesuite/truffle/releases/tag/v5.7.3). Closing.

PR #5827, opened 2023-01-10
Have Dashboard politely ask users occasionally if they'll share telemetry 

cliffoo:
Addresses: #4828

Use dashboard to prompt our users to graciously enable analytics.

### Behavior on load:
```mermaid
flowchart TD
  A{"Has the user explicitly<br/>opted in/out before?<br />(via the config command,<br />or much less likely, editing<br />the user config manually)"} -->|no| B([Ask to enable analytics])
  A -->|yes| C{Is analytics enabled?}
  C -->|no| D{Has it been a year since<br/>user last said no?}
  C -->|yes| E([OK])
  D -->|no| E
  D -->|yes, ask again| B
```

### How?
`DashboardServer`'s express server has two new endpoints: `POST` and `GET` at `/analytics`. Their corresponding handlers read and write to the user config object, which in turn modifies the config file on disk. Frontend just interacts with `/analytics` and manages notifications as needed.

### Wording
The prompt content (located at `src/utils/notifications/analytics.tsx`) mostly follows that of the previous PR (#5290).

### Testing instructions
0. Locate your Truffle user config file (likely at `~/.config/truffle-nodejs/config.json`)
1. Inside this user config file, remove `enableAnalytics`, `analyticsSet`, `analyticsMessageDateTime` for a clean slate, maybe `uniqueId` also. Continue if these fields are not present.
2. Start and open dashboard. You will be prompted to enable analytics, click `Enable`.
3. Verify that the config file is updated with analytics enabled. Repeat step 1 for clean slate.
4. Refresh dashboard. This time when you're prompted click `Dismiss`.
5. Verify that the config file is updated with analytics disabled. Edit `analyticsMessageDateTime` to `1` (no quotes!).
6. Refresh dashboard. You will be prompted to enable analytics, but with slightly different text.

### Related:
- Our [analytics policy](https://trufflesuite.com/analytics)
- Our [analytics code](https://github.com/trufflesuite/truffle/blob/v5.7.2/packages/core/lib/services/analytics/google.js)

### DIFF
--- a/packages/dashboard/lib/DashboardServer.ts
+++ b/packages/dashboard/lib/DashboardServer.ts
@@ -2,12 +2,13 @@ import express, { Application, NextFunction, Request, Response } from "express";
 import path from "path";
 import getPort from "get-port";
 import open from "open";
+import { v4 as uuid } from "uuid";
+import Config from "@truffle/config";
 import {
   dashboardProviderMessageType,
   LogMessage,
   logMessageType
 } from "@truffle/dashboard-message-bus-common";
-
 import { DashboardMessageBus } from "@truffle/dashboard-message-bus";
 import { DashboardMessageBusClient } from "@truffle/dashboard-message-bus-client";
 import cors from "cors";
@@ -93,6 +94,30 @@ export class DashboardServer {
       await this.connectToMessageBus();
       this.expressApp.post("/rpc", this.postRpc.bind(this));
     }
+
+    this.expressApp.get("/analytics", (_req, res) => {
+      const userConfig = Config.getUserConfig();
+      res.json({
+        enableAnalytics: userConfig.get("enableAnalytics"),
+        analyticsSet: userConfig.get("analyticsSet"),
+        analyticsMessageDateTime: userConfig.get("analyticsMessageDateTime")
+      });
+    });
+
+    this.expressApp.put("/analytics", (req, _res) => {
+      const { enable } = req.body as { enable: boolean };
+
+      const userConfig = Config.getUserConfig();
+
+      const uid = userConfig.get("uniqueId");
+      if (!uid) userConfig.set("uniqueId", uuid());
+
+      userConfig.set({
+        enableAnalytics: !!enable,
+        analyticsSet: true,
+        analyticsMessageDateTime: Date.now()
+      });
+    });
 
     this.expressApp.use(express.static(this.frontendPath));
     this.expressApp.get("*", (_req, res) => {
--- a/packages/dashboard/lib/DashboardServer.ts
+++ b/packages/dashboard/lib/DashboardServer.ts
@@ -105,7 +105,7 @@ export class DashboardServer {
     });
 
     this.expressApp.put("/analytics", (req, _res) => {
-      const { enable } = req.body as { enable: boolean };
+      const { value } = req.body as { value: boolean };
 
       const userConfig = Config.getUserConfig();
 
@@ -113,7 +113,7 @@ export class DashboardServer {
       if (!uid) userConfig.set("uniqueId", uuid());
 
       userConfig.set({
-        enableAnalytics: !!enable,
+        enableAnalytics: !!value,
         analyticsSet: true,
         analyticsMessageDateTime: Date.now()
       });
--- a/packages/dashboard/src/components/composed/Layout.tsx
+++ b/packages/dashboard/src/components/composed/Layout.tsx
@@ -14,7 +14,7 @@ import {
   analyticsNotificationId
 } from "src/utils/notifications";
 
-const ARBITRARY_ANALYTICS_NEXT_ASK_THRESHOLD = 31544444444; // 365 days
+const ARBITRARY_ANALYTICS_NEXT_ASK_THRESHOLD_IN_DAYS = 365;
 
 const useStyles = createStyles((_theme, _params, _getRef) => ({
   main: {
@@ -30,15 +30,26 @@ function Layout(): JSX.Element {
   } = useDash()!;
   const { classes } = useStyles();
 
+  // Analytics notifications
   useEffect(() => {
-    if (notice.show || analyticsConfig.analyticsSet === null) return;
+    // Don't ask if...
+    if (
+      // There's currently a fullscreen notice
+      notice.show ||
+      // Analytics is already enabled
+      analyticsConfig.enableAnalytics ||
+      // Analytics config info is not in state yet
+      analyticsConfig.analyticsMessageDateTime === null
+    )
+      return;
 
-    const shouldAsk =
-      !analyticsConfig.enableAnalytics && !analyticsConfig.analyticsSet;
-    const shouldAskAgain =
-      !analyticsConfig.enableAnalytics &&
-      Date.now() - analyticsConfig.analyticsMessageDateTime! >
-        ARBITRARY_ANALYTICS_NEXT_ASK_THRESHOLD;
+    const askAfter = new Date(analyticsConfig.analyticsMessageDateTime);
+    askAfter.setDate(
+      askAfter.getDate() + ARBITRARY_ANALYTICS_NEXT_ASK_THRESHOLD_IN_DAYS
+    );
+
+    const shouldAsk = !analyticsConfig.analyticsSet;
+    const shouldAskAgain = new Date() > askAfter;
 
     const analyticsNotificationArgs = [
       () => {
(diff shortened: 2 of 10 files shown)


packages/dashboard/lib/DashboardServer.ts
```diff
@@ -94,6 +95,30 @@ export class DashboardServer {
...
+      const userConfig = Config.getUserConfig();
+      res.json({
+        enableAnalytics: userConfig.get("enableAnalytics"),
+        analyticsSet: userConfig.get("analyticsSet"),
+        analyticsMessageDateTime: userConfig.get("analyticsMessageDateTime")
+      });
+    });
+
+    this.expressApp.put("/analytics", (req, _res) => {
+      const { enable } = req.body as { enable: boolean };
```
gnidan, line 108: Could we change this to `value` or something? Or maybe separate things like `{ enable: true } | { disable: true } | {}`?

Maybe this is a nit considering the urgency. Not requesting changes for this one... I just think `{ enable: false }` is like... "so you want me to... do nothing?"

packages/dashboard/src/components/composed/Layout.tsx
```diff
@@ -1,8 +1,20 @@
...
 import { Outlet } from "react-router-dom";
 import { useDash } from "src/hooks";
 import Sidebar from "src/components/composed/Sidebar";
 import Notice from "src/components/composed/Notice";
+import {
+  analyticsNotifications,
+  analyticsNotificationId
+} from "src/utils/notifications";
+
+const ARBITRARY_ANALYTICS_NEXT_ASK_THRESHOLD = 31544444444; // 365 days
```
gnidan, line 17: How'd you do this math?

packages/dashboard/src/components/composed/Layout.tsx
```diff
@@ -13,10 +25,43 @@ const useStyles = createStyles((_theme, _params, _getRef) => ({
...
 
+  useEffect(() => {
+    if (notice.show || analyticsConfig.analyticsSet === null) return;
+
+    const shouldAsk =
+      !analyticsConfig.enableAnalytics && !analyticsConfig.analyticsSet;
+    const shouldAskAgain =
+      !analyticsConfig.enableAnalytics &&
+      Date.now() - analyticsConfig.analyticsMessageDateTime! >
+        ARBITRARY_ANALYTICS_NEXT_ASK_THRESHOLD;
```
gnidan, line 41: Note that this kind of math is rife with edge cases around leap years and whatnot. Better if we can do a proper duration check vs. math and assuming we know how many milliseconds a year is.

Happy to accept an issue for this instead of fixing it now.

gnidan (commented):
This looks good. Happy to approve tomorrow; just want to give others a chance to take a look first.

Thanks @cliffoo!

packages/dashboard/src/components/composed/Layout.tsx
gnidan: (OK, if this triggers a day early, it doesn't really matter. But still... good practice not to assume you can just subtract dates like that)

cds-amal: https://date-fns.org/ may be a viable option

cds-amal: > Happy to accept an issue for this instead of fixing it now.

+1

packages/dashboard/src/utils/notifications/analytics.tsx
```diff
@@ -0,0 +1,71 @@
...
+    ),
+    color: "yellow"
+  }),
+  "ask-again": (enable, disable) => ({
+    ...baseNotification,
+    title: "Telemetry is currently disabled",
+    message: (
+      <Stack>
+        <Text>
+          Would you consider enabling it? It helps us make Truffle a better
```
lsqproduction, line 50: Since this has been a while, should we tack on the policy again?

packages/dashboard/src/utils/notifications/analytics.tsx
```diff
@@ -0,0 +1,71 @@
...
+> = {
+  "ask": (enable, disable) => ({
+    ...baseNotification,
+    title: "Help improve Truffle",
+    message: (
+      <Stack>
+        <Text>
+          You can help us improve Truffle by allowing us to track how you use
+          our tools. Would you like to share anonymous telemetry? View the
+          Truffle Analytics policy&nbsp;
```
lsqproduction, line 26: Truffle Telemetry policy?

packages/dashboard/lib/DashboardServer.ts
cliffoo: 👍 I've renamed it to `value`

packages/dashboard/src/components/composed/Layout.tsx
cliffoo: Uh I took 4 significant digits and just... went with it. This is updated! See below.

packages/dashboard/src/components/composed/Layout.tsx
cliffoo: Good point! I've changed this to compare `Date` objects instead. The logic is as follows:
```ts
const askAfter = new Date(lastAskedInMs);
askAfter.setDate(askAfter.getDate() + 365);

if (new Date() > askAfter) // Ask again!
```
Note that this _is_ affected by leap years. We can fix it by using the pair `setFullYear` and `getFullYear` instead of `setDate` and `getDate`, but I figured days are easier to work with if we later want to change this threshold to 90 days or something.

packages/dashboard/src/utils/notifications/analytics.tsx
cliffoo: I agree! Added.

packages/dashboard/src/utils/notifications/analytics.tsx
cliffoo: I think we're collecting telemetry and it's governed by our analytics policy. Right?

lsqproduction: thank you for clarifying for me

packages/dashboard/src/components/composed/Layout.tsx
gnidan: If we wanted to be super accurate, the way to do this would be to do, like, `add(date, thresholdDuration)`, where `thresholdDuration = { years: 1 }`, then check against that. 

But it's okay. If you say "no" on 2023-12-01 and we ask again on 2024-11-30... kinda trivial.

packages/dashboard/src/contexts/DashContext/DashContext.ts
```diff
@@ -13,6 +13,7 @@ export interface ContextValue {
       lifecycle: ReceivedMessageLifecycle<DashboardProviderMessage>
     ) => any;
     toggleNotice: () => void;
+    updateAnalyticsConfig: (value: boolean) => void;
```
gnidan, line 16: So for this sorta thing, I'll also note that, in situations like this, you might consider getting into the habit of making separate operations for enable vs. disable. It feels a bit more hygienic: `enableAnalytics()` and `disableAnalytics()` are more clear and succinct compared to `updateAnalyticsConfig(true)` and `updateAnalyticsConfig(false)`.

(There are downsides here, of course, like having to pass an extra function around in the context here, but my personal stance is that it's worth it to have operations align with specific user actions)

fainashalts (approved):
LGTM! I think the language is very polite and clear! I agree with @gnidan on separating `enableAnalytics()` and `disableAnalytics()` but that can be done as follow-on work if desired. :)

gnidan (approved):

packages/dashboard/src/utils/notifications/analytics.tsx
```diff
@@ -0,0 +1,75 @@
+import { Stack, Group, Button, Text, Anchor } from "@mantine/core";
+import type { NotificationProps } from "@mantine/notifications";
+
+type NotificationPropsWithId = NotificationProps & { id: string };
+
+export const analyticsNotificationId = "analytics";
```
eggplantzzz, line 6: Just curious, what is this id used for?

cliffoo: Ah this is so that an id can be assigned to a specific notification, so we can reference that notification later and make changes to it.

Here the notification with id `"analytics"` is updated to show _hey thanks for enabling telemetry_ when user clicks enable, and hidden when user click no.

eggplantzzz (approved):
i like it

PR #5829, opened 2023-01-11
Internal improvement: add memoization to `Cache` utility and switch to using `async` fs methods

eggplantzzz:
## PR description

This PR removes one instance of usage of `originalRequire` in favor of reading a cached compiler file directly from disk. This also makes it a bit more consistent with how compilers are loaded when fetching them remotely as it will now use the method [`compilerFromString`](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts#L99-L107)

## Testing instructions

You can test this by running `truffle compile` on a project where you have a compiler cached and verify that it compiles correctly.

## Documentation

### DIFF
--- a/packages/compile-solidity/src/compilerSupplier/Cache.ts
+++ b/packages/compile-solidity/src/compilerSupplier/Cache.ts
@@ -4,6 +4,7 @@ import fs from "fs";
 
 export class Cache {
   private compilerCachePath: string;
+  private memoizedCompilers: Map<string, string>;
 
   constructor() {
     const compilersDir = path.resolve(
@@ -15,6 +16,7 @@ export class Cache {
     if (!fs.existsSync(compilerCachePath)) fs.mkdirSync(compilerCachePath); // for 5.0.8 users
 
     this.compilerCachePath = compilerCachePath;
+    this.memoizedCompilers = new Map();
   }
 
   list() {
@@ -24,11 +26,29 @@ export class Cache {
   add(code: string, fileName: string) {
     const filePath = this.resolve(fileName);
     fs.writeFileSync(filePath, code);
+    this.memoizedCompilers[fileName] = code;
   }
 
   has(fileName: string) {
     const file = this.resolve(fileName);
     return fs.existsSync(file);
+  }
+
+  loadFile(fileName: string): string {
+    if (this.memoizedCompilers.has(fileName)) {
+      return this.memoizedCompilers[fileName];
+    }
+    try {
+      const compiler = fs.readFileSync(fileName).toString();
+      this.memoizedCompilers[fileName] = compiler;
+      return compiler;
+    } catch (error) {
+      if (!error.message.includes("ENOENT: no such file")) {
+        throw error;
+      } else {
+        throw new Error("The file specified could not be found.");
+      }
+    }
   }
 
   resolve(fileName: string) {
--- a/packages/compile-solidity/src/compilerSupplier/Cache.ts
+++ b/packages/compile-solidity/src/compilerSupplier/Cache.ts
@@ -1,6 +1,8 @@
 import Config from "@truffle/config";
 import path from "path";
 import fs from "fs";
+// @ts-ignore
+import * as fsPromises from "fs/promises";
 
 export class Cache {
   private compilerCachePath: string;
@@ -19,28 +21,28 @@ export class Cache {
     this.memoizedCompilers = new Map();
   }
 
-  list() {
-    return fs.readdirSync(this.compilerCachePath);
+  async list() {
+    return await fsPromises.readdir(this.compilerCachePath);
   }
 
-  add(code: string, fileName: string) {
+  async add(code: string, fileName: string) {
     const filePath = this.resolve(fileName);
-    fs.writeFileSync(filePath, code);
+    await fsPromises.writeFile(filePath, code);
     this.memoizedCompilers.set(filePath, code);
   }
 
-  has(fileName: string) {
+  async has(fileName: string) {
     const filePath = this.resolve(fileName);
-    return fs.existsSync(filePath);
+    return await fsPromises.exists(filePath);
   }
 
-  loadFile(fileName: string): string {
+  async loadFile(fileName: string): Promise<string> {
     const filePath = this.resolve(fileName);
     if (this.memoizedCompilers.has(filePath)) {
       return this.memoizedCompilers.get(filePath)!;
     }
     try {
-      const compiler = fs.readFileSync(filePath).toString();
+      const compiler = (await fsPromises.readFile(filePath)).toString();
       this.memoizedCompilers.set(filePath, compiler);
       return compiler;
     } catch (error) {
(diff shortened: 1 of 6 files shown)


packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -2,7 +2,7 @@ import debugModule from "debug";
 const debug = debugModule("compile:compilerSupplier");
 
 import requireFromString from "require-from-string";
-import originalRequire from "original-require";
```
sukanyaparashar, line 5: What was this `originalRequire` module doing previously?

eggplantzzz: So `original-require` is a reference to Node's `require` as opposed to webpack's `require`. It was `require`ing the compiler and wrapping it using `solc/wrapper`. Reading the file from disk does the same thing, but differently.

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
```diff
@@ -120,9 +120,9 @@ export class VersionRange {
     const listeners = observeListeners();
     try {
       const filePath = this.cache.resolve(fileName);
-      const soljson = originalRequire(filePath);
+      const soljson = fs.readFileSync(filePath).toString();
```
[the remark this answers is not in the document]
eggplantzzz: I don't know, do we? Just glancing at the code it looks like the `load` method only gets called once per flow. If it becomes a problem later I suppose we could devise some kind of scheme for memoization maybe?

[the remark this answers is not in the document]
cds-amal: Yea, I like the memos, though I wonder how we'll determine if a problem arises later. 

What do you think of refactoring this functionality to the Cache class and replacing the above with `soljson = this.cache.load(fileName)` ?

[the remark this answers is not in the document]
eggplantzzz: what exactly do you mean by refactoring to the Cache class?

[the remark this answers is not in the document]
cds-amal: Add a method to [this Cache class](https://github.com/trufflesuite/truffle/blob/develop/packages/compile-solidity/src/compilerSupplier/Cache.ts#L5) that loads a file from cache.

```javascript
  loadFile(fileName: string): string | null {
    //resolves and loads file or returns  null
    return null
  }
```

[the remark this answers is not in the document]
eggplantzzz: sure

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
sukanyaparashar: Is this change only for making it consistent with remote fetching of the compilers or there are more advantages of doing this?

packages/compile-solidity/src/compilerSupplier/Cache.ts
```diff
@@ -24,13 +26,31 @@ export class Cache {
   add(code: string, fileName: string) {
     const filePath = this.resolve(fileName);
     fs.writeFileSync(filePath, code);
+    this.memoizedCompilers[fileName] = code;
```
cds-amal, line 29:
```suggestion
    this.memoizedCompilers[filePath] = code;
```
To prevent `fileNames` that would resolve to the same path, I think the map key should be normalized, that is the resolved path in this case.

packages/compile-solidity/src/compilerSupplier/Cache.ts
```diff
@@ -24,13 +26,31 @@ export class Cache {
...
   }
 
   has(fileName: string) {
     const file = this.resolve(fileName);
     return fs.existsSync(file);
   }
 
+  loadFile(fileName: string): string {
+    if (this.memoizedCompilers.has(fileName)) {
+      return this.memoizedCompilers[fileName];
```
cds-amal, line 39: Same consideration as above, the key should be the resolved path

packages/compile-solidity/src/compilerSupplier/loadingStrategies/VersionRange.ts
eggplantzzz: I think we are trying to get rid of usages of `original-require`. This would eliminate one more.

packages/compile-solidity/src/compilerSupplier/Cache.ts
eggplantzzz: I'll change it but currently it doesn't look like two files can have the same name as only one directory is used to cache the compilers

packages/compile-solidity/src/compilerSupplier/Cache.ts
```diff
@@ -24,11 +26,30 @@ export class Cache {
...
+    }
+    try {
+      const compiler = fs.readFileSync(filePath).toString();
+      this.memoizedCompilers[filePath] = compiler;
+      return compiler;
+    } catch (error) {
+      if (!error.message.includes("ENOENT: no such file")) {
+        throw error;
+      } else {
+        throw new Error("The file specified could not be found.");
```
cds-amal, line 50: Would this be a case for `@truffle/error`?

packages/compile-solidity/src/compilerSupplier/Cache.ts
```diff
@@ -24,11 +26,30 @@ export class Cache {
...
   has(fileName: string) {
-    const file = this.resolve(fileName);
-    return fs.existsSync(file);
+    const filePath = this.resolve(fileName);
+    return fs.existsSync(filePath);
+  }
+
+  loadFile(fileName: string): string {
+    const filePath = this.resolve(fileName);
+    if (this.memoizedCompilers.has(filePath)) {
```
cds-amal, line 39: Can you add a test to verify `memoizedCompilers`, please?

cds-amal (commented):
This looks good, @eggplantzzz, though I'd like a test for the memorization.

packages/compile-solidity/src/compilerSupplier/Cache.ts
eggplantzzz: done

packages/compile-solidity/src/compilerSupplier/Cache.ts
eggplantzzz: Hmm probably not, right? This would most likely not be directly exposed to a user.

cds-amal (approved):
Nice work, @eggplantzzz. Love the tests.

Issue #5828, opened 2023-01-11
improve error handling when using `truffle` and you make a typo in the command part


pedro-nonfree:
- [ ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

When try to run `truffle test` and I do something like `truffle tes` or any typo you can think of, I get a very strange error, and I am not thinking that the problem is in my typo in the CLI command but in the tests or in my code.

It happened two times on me, on first time, I was debugging my tests for 30 minutes, and it magically started to work after having a break. On second time I found this problem so that's why I am filling this issue as a Developer eX``perience bug 

## Steps to Reproduce

```
[14:33:02] $ truffle tes
Using network 'development'.


Compiling your contracts...
===========================
> Compiling ./contracts/LoraCoin.sollc-bin. Attempt #1
> Compiling ./test/test_loracoin.sol
> Artifacts written to /tmp/test--132834-2rdUbv0KLOxE
> Compiled successfully using:
   - solc: 0.8.17+commit.8df45f5f.Emscripten.clang
TypeError: Cannot assign to read only property 'undefined' of object '#<Object>'
    at Object.<anonymous> (/usr/lib/node_modules/truffle/build/webpack:/packages/test/dist/src/Test.js:282:1)
    at Generator.next (<anonymous>)
    at /usr/lib/node_modules/truffle/build/webpack:/packages/test/dist/src/Test.js:31:1
    at new Promise (<anonymous>)
    at exports.modules.583725.__awaiter (/usr/lib/node_modules/truffle/build/webpack:/packages/test/dist/src/Test.js:27:1)
    at Object.setJSTestGlobals (/usr/lib/node_modules/truffle/build/webpack:/packages/test/dist/src/Test.js:253:1)
    at Object.<anonymous> (/usr/lib/node_modules/truffle/build/webpack:/packages/test/dist/src/Test.js:149:1)
    at Generator.next (<anonymous>)
    at fulfilled (/usr/lib/node_modules/truffle/build/webpack:/packages/test/dist/src/Test.js:28:43)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
Truffle v5.6.5 (core: 5.6.5)
Node v18.12.1
```

## Expected Behavior

What you expected to happen.

```
truffle tes command not found, run truffle --help to see the available commands for truffle
```

## Actual Results

[...]

PR #5830, opened 2023-01-11
Bug fix: add debugGlobal default to config

eggplantzzz:
Addresses #5828

## PR description

There is a property that gets set on config (during testing) named `debugGlobal`. Currently Truffle relies on the yargs `builder` object to set this property which is jank. It is also causing problems where it doesn't get set when you miskey `truffle test`. There is an issue for this bug [here](https://github.com/trufflesuite/truffle/issues/5828). This PR adds a proper default for this value in @truffle/config.

## Testing instructions

Run `truffle tes` on the command line. Verify that Truffle correctly runs the tests. If you currently try this on published Truffle, it will fail.

## Documentation

### DIFF
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -32,6 +32,7 @@ export const getInitialConfig = ({
       displayPrefix: "",
       preventConsoleLogMigration: false
     },
+    debugGlobal: "debug",
     gas: null,
     gasPrice: null,
     maxFeePerGas: null,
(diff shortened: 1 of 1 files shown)


cds-amal (approved):
Looks good to me! Weird how this bug manifested.

lsqproduction:
What's this debugGlobal?

eggplantzzz:
It looks like it is the variable name used for setting break points when using the debugger. The default case is that it is called "debug".

dongmingh (approved):
With this fix, `truffle tes` succeeds.

haltman-at:
> It looks like it is the variable name used for setting break points when using the debugger. The default case is that it is called "debug". Someone can correct me, of course, if I'm wrong about that.

Well, they're not really break points; it lets you break into the debugger to debug particular transactions, but it doesn't let you set breakpoints inside of them, nor breakpoints outside of them for that matter.

PR #5836, opened 2023-01-13
add blurb in @truffle/contract README about overloaded functions

eggplantzzz:


packages/contract/README.md
```diff
@@ -310,6 +310,52 @@ const val = instance.getValue();
 // since the contract returns that value.
 ```
 
+#### Overloaded functions
+
+You may find yourself in the situation of having a Truffle contract object that has multiple functions with the same name.
+You can call these "overloaded functions" just like you would a normal contract function. Truffle contract instances
```
cds-amal, line 316:
```suggestion
You can invoke these overloaded functions just like you would a normal contract function. Truffle contract instances
```

packages/contract/README.md
```diff
@@ -310,6 +310,52 @@ const val = instance.getValue();
...
+actually wrap web3's contract abstraction (`web3.eth.Contract`) and so when you call an overloaded function, it
+uses the same function resolution that web3 uses. However, we must give a warning that this overloaded function
+resolution is a bit dodgy and can resolve to the wrong function when you call it. So be careful!
```
cds-amal, line 319:
```suggestion
uses the same function resolution that web3 uses. However, we must give a warning that web3's resolution is not always correct and can resolve to the wrong function when invoked. So be careful!
```

packages/contract/README.md
```diff
@@ -310,6 +310,52 @@ const val = instance.getValue();
...
+
+The more unambiguous way of calling these types of functions is to use the `methods` namespace on Truffle contract
```
cds-amal, line 321:
```suggestion
The precice way to invoke these types of functions is to use the `methods` namespace on Truffle contract
```

cds-amal (commented):
I left some suggestions for your consideration.

packages/contract/README.md
sukanyaparashar: Does it mean `precise` @cds-amal ?

packages/contract/README.md
eggplantzzz: I like the change to `invoke` but I prefer to leave overloaded functions to be emphasized in some way to make it clear that this is terminology for people unfamiliar with it.

packages/contract/README.md
eggplantzzz: something is ambiguous when it is unclear - usually the meaning of something is ambiguous when it could mean one of several things

eggplantzzz: for this one I re-wrote it as "The more precise way of referencing these...". Is that good?

packages/contract/README.md
haltman-at: My own editing comment here is that if you're going to call out something, it should be just the word "overloaded", not the whole noun phrase.

packages/contract/README.md
haltman-at: Yeah I think "unambiguous" here is clearer than precise.  I would just say "unambiguous" rather than "more unambiguous", though -- it *is* unambiguous, we don't need to qualify it.

packages/contract/README.md
```diff
@@ -310,6 +310,52 @@ const val = instance.getValue();
...
+
+This Solidity contract contains two functions named `setValue`, each taking a different type of input. In your JavaScript you
+might do something like the following:
+
+```javascript
+await instance.setValue("this is not string cheese");
+```
+
+Truffle contract (internally using web3's overloaded function resolution) will see that you input a string and
+call the second method in the contract to set `myString`. Again however, sometimes when your contract is more
```
haltman-at, line 347: Hahaha you wish!  Web3's overload resolution only works reliably when the functions have different numbers of arguments.  The behavior you describe here is obviously what *ought* to actually happen, but not necessarily what actually will.  In reality there's a decent chance that Web3 tries to go with the integer function *anyway* and then just errors on you because the input wasn't an integer like expected.  This is why I keep saying it's so bad, and the primary reason I created Truffle Encoder with the hope of replacing this!

haltman-at (changes requested):
Gonna have to hit request changes, the example needs to be changed (see my comment).  Might also be good to adjust the language earlier to account for this.

packages/contract/README.md
cds-amal: Wow, that is terrible.

eggplantzzz: Oh I didn't realize it was primarily based on number of arguments. Bummer, I'll update this blurb here. I wonder how easy it would be to create our own shim/adapter-thingy for this...

haltman-at (approved):
Seems OK to me!  There should probably also be a corresponding change to trufflesuite.com.

PR #5841, opened 2023-01-17
FEAT: add a settings menu on dashboard for user to change analytics preference any time

lsqproduction:
## Documentation


packages/dashboard/lib/DashboardServer.ts
```diff
@@ -117,6 +117,7 @@ export class DashboardServer {
         analyticsSet: true,
         analyticsMessageDateTime: Date.now()
       });
+      _res.send();
```
cliffoo, line 120: Two things here:
1. This variable was prepended with `_` because it was unused, so using it here means we should remove `_`.
2. I don't think we need to `send`? Does it work without? In any case this can probably be `res.end`

packages/dashboard/src/App.tsx
```diff
@@ -26,6 +28,7 @@ function App(): JSX.Element {
                     <Route path="/" element={<Layout />}>
                       <Route index element={<Navigate to="/rpcs" replace />} />
                       <Route path="rpcs" element={<RPCs />} />
+                      <Route path="settings" element={<Analytics />} />
```
cliffoo, line 31: It seems to me `/settings` should lead to `<Settings />`, not `<Analytics />`.

packages/dashboard/src/components/composed/Sidebar/Middle/index.tsx
```diff
@@ -40,6 +41,11 @@ function Middle(): JSX.Element {
           </Indicator>
         }
       />
+      <NavButton
+        label="Settings"
+        to="/Settings"
+        icon={<Settings {...featherIconProps} />}
+      />
```
cliffoo, line 48: No problem with this code but I'm thinking settings shouldn't be on the sidebar, along with signature requests and contract interaction and such, as it's more tangential to what dashboard is actually for. Instead of using `<NavButton />` here, do you think the dark/light mode button right now should instead be the settings button? From there you would be able to toggle analytics, dark/light mode, etc.

packages/dashboard/src/components/composed/Analytics.tsx
```diff
@@ -0,0 +1,90 @@
+import {
```
cliffoo, line 1: This component should be renamed. See above.

packages/dashboard/lib/DashboardServer.ts
davidmurdoch, line 120: Since you are now using `_res` the name should be changed to `res`.

packages/dashboard/src/components/composed/Sidebar/Middle/index.tsx
cliffoo: Following on that, maybe settings shouldn't be its own page but a modal instead.

packages/dashboard/src/components/composed/Analytics.tsx
```diff
@@ -0,0 +1,90 @@
...
+                  href="https://trufflesuite.com/analytics/"
+                  target="_blank"
+                  color="truffle-beige"
+                  inherit
+                  className={classes.docsAnchor}
+                >
+                  here
+                  <Space w={3} />
+                  <ExternalLink size={12} />
+                </Anchor>
```
davidmurdoch, line 63: We shouldn't use words like "here" to link places when you can add a link to the thing itself:

```suggestion
                the 
                <Anchor
                  href="https://trufflesuite.com/analytics/"
                  target="_blank"
                  color="truffle-beige"
                  inherit
                  className={classes.docsAnchor}
                >
                  Truffle Analytics policy
                  <Space w={3} />
                  <ExternalLink size={12} />
                </Anchor>
```

note: I don't know what ExternalLink and Space are

packages/dashboard/src/components/composed/Analytics.tsx
```diff
@@ -0,0 +1,90 @@
...
+}));
+
+function Analytics(): JSX.Element {
+  const {
+    state: { analyticsConfig },
+    operations: { updateAnalyticsConfig }
+  } = useDash()!;
+  const analyticsValue = analyticsConfig.enableAnalytics ? true : false;
+
+  const [checked, setChecked] = useState(analyticsValue);
```
cliffoo, line 34: This can just be 
```tsx
const [checked, setChecked] = useState(!!analyticsConfig.enableAnalytics)
```

cliffoo (commented):
Some initial comments :) Thanks @lsqproduction

packages/dashboard/src/components/composed/Analytics.tsx
```diff
@@ -0,0 +1,90 @@
...
+                  <ExternalLink size={12} />
+                </Anchor>
+              </Accordion.Panel>
+            </Grid.Col>
+            <Grid.Col span={1}>
+              <Accordion.Panel>
+                <Switch
+                  size="lg"
+                  onLabel="YES"
+                  offLabel="NO"
```
davidmurdoch, line 71: A suggestion: We shouldn't be use all caps in the literal text, as some screen readers have been reported to pronounce each individual letter instead of the whole word itself (though I doubt any are dumb enough to pronounce "YES", and "NO" as "Y.,E.S.", and "N.O.", it still seems better to avoid all caps here). Instead, use CSS's `text-transform: uppercase` (or just use title case, "Yes", "No").

davidmurdoch (changes requested):

packages/dashboard/lib/DashboardServer.ts
lsqproduction: Woo. Yes let me try those.

packages/dashboard/src/App.tsx
lsqproduction: Good point. Will make the change.

packages/dashboard/src/components/composed/Analytics.tsx
lsqproduction: Nice, thank you!

packages/dashboard/src/components/composed/Analytics.tsx
lsqproduction: Good to know. Thank you for sharing this. I will make the change.

packages/dashboard/src/components/composed/Sidebar/Middle/index.tsx
lsqproduction: What do you think about a settings button next to the dark/light mode button that opens up a modal?

cliffoo: Eh it doesn't agree with me personally.

lsqproduction: yeah, it doesn't look pretty. Any other suggestions?

cliffoo: Yea as I mentioned [above](https://github.com/trufflesuite/truffle/pull/5841#discussion_r1072703074), replacing the dark/light mode button with settings button.

Keen to hear other thoughts though.

packages/dashboard/lib/DashboardServer.ts
eggplantzzz: Does `res.end` send a response to the client?

cliffoo: Yes, client should get a 200.

PR #5849, opened 2023-01-20
bump @ensdomains/ensjs version to 3.0.0-alpha

adonesky1:
## Documentation

### DIFF
--- a/packages/encoder/package.json
+++ b/packages/encoder/package.json
@@ -25,7 +25,7 @@
   },
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@ensdomains/ensjs": "^2.1.0",
+    "@ensdomains/ensjs": "^3.0.0-alpha.41",
     "@ethersproject/address": "^5.0.10",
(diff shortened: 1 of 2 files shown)


eggplantzzz:
Hey @adonesky1, thanks for submitting this! We probably want to wait until they have a stable version of 3 before pushing this through. Do you know anything about their schedule for v3?

haltman-at:
Hm, I notice this only changes it in `encoder`, but we use it in other packages too.  If we're going to change it we should change all of them, right?  (I also have a PR coming up that adds it in some more packages...)

haltman-at:
Hm, this seems to be failing CI, but I don't understand why.  The error that's coming up doesn't seem to have anything to do with ENS, and yet it seems to be coming up repeatedly.  No idea what's going on here.

eggplantzzz:
Closing this until later. We'll either decide to upgrade to a stable version of 3 when it is released or perhaps get rid of the encoder's use of ensjs altogether in favor of our own implementation of whatever we use that library for.

PR #5847, opened 2023-01-19
allow truffle compile behind corporate proxy Issue 4016

joseluu:
## PR description

Did the necessary modifications to the axios calls to allow truffle compile behind corporate proxy

## Testing instructions

truffle compile should be able to download the required compiler version without or with proxy
test by changing the compiler version in truffle-config.js to a version you have never used as to force the download of a new version.
truffle deploy should work without or with proxy

## Documentation

- [ x] I thought about documentation and added the `doc-change-required` label to this PR if documentation updates are required.

to allow pass through corporate proxies
you now can define the environment variable HTTP_PROXY=http://username:password@proxy:port
notice that HTTP and http:// are used rather than HTTPS and https://
you may also define NODE_TLS_REJECT_UNAUTHORIZED=0  or NODE_EXTRA_CA_CERTS=<proxy cert file> if the corporate proxy decripts the https flows

## Breaking changes and new features

- [ x] I have added any needed `breaking-change` and `new-feature` labels for the appropriate packages.
The package compile-solidity has changed, but no breaking change

PR #5854, opened 2023-01-24
Dep update: bump axios to 1.2.4

eggplantzzz:
While reviewing https://github.com/trufflesuite/truffle/pull/5847, I noticed that we are quite behind on the version of axios that Truffle is using. This PR bumps axios to the latest version.


haltman-at (approved):

cds-amal (approved):
Let's give it a go!  Love removing a pre 1.0 SemVer dependency.

Issue #5838, opened 2023-01-13
Truffle REPL does not respect `""` in a string with spaces


sukanyaparashar:
- [ ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

While testing `truffle call` command in `develop` and `console`, I ran into this issue. For example, when a contract function is called with `call <contract_name> <function_name> <string_argument>` which returns the string with spaces, it throws an error that function argument is invalid. The parsing of the string within `""` in the REPL interferes with the spaces and does not treat the whole string as one.

haltman-at:
Could you give an example here?

sukanyaparashar:
For example, when I run `call <contract_name> <function_name> <args>` in the REPL (`develop` or `console`) that takes a string with two words separated by a space, as an argument, the arguments array shows two strings instead of one like the screenshot below-
<img width="469" alt="functionargs" src="https://user-images.githubusercontent.com/25079663/212981120-741fd908-5e20-42b2-948a-4e6834e11ec9.png">

On the other hand, the same call outside of REPL works fine like the screenshot below-
<img width="785" alt="functionargs" src="https://user-images.githubusercontent.com/25079663/212981321-9189fa8f-f383-4987-beea-9da4815644bf.png">

haltman-at:
Oh, yikes!  That's definitely a problem for this...

haltman-at:
OK, I think we can likely handle this with the [shell-quote](https://www.npmjs.com/package/shell-quote) library.

gnidan:
This is now fixed in [Truffle v5.7.6](https://github.com/trufflesuite/truffle/releases/tag/v5.7.6). Thanks all!

PR #5861, opened 2023-01-28
Properly handle quotes, backslashes, etc in commands at console

haltman-at:
## PR description

Addresses #5838.

This PR makes it so that arguments to Truffle commands at the Truffle console are handled like arguments at the actual command line.  This is intended primarily so that `truffle call` (#5647) will work properly at the Truffle console once it is merged, but it also fixes bugs with existing commands.

For instance, suppose you have a test file `test/meta coin.js`, and you are at the Truffle console, and you want to run specifically the tests in this file and no others.  How would you do that?  Without this PR, there's no straightforward way!  With it, you can just do `test 'test/meta coin.js'` or `test test/meta\ coin.js` or similar.

(This PR has been substantially overhauled, original description is in strikeout below.)

I originally did this using the [shell-quote](https://www.npmjs.com/package/shell-quote) package, but it turned out to have substantial problems (see discussion below), so I ended up rolling my own function to handle shell quotes and escapes.  I put it in `command-utils.js`; it's called `parseQuotesAndEscapes`.  I also wrote a bunch of tests for it.

Quoting and escaping works basically like in bash -- backslash is escape and (outside of a quote) causes whatever is after it to be taken literally; double-quotes cause whatever's inside to be taken literally except for the sequence `\"`, which escapes the double-quote; single-quotes are similar but don't allow any escaping at all.  Other than that, arguments are split on (unescaped, unquoted) whitespace.

However the `parseQuotesAndEscapes` function does allow customizing the escape characters; to avoid problems, on Windows, instead of backslash acting as escape, caret and grave will act as escape.  (We check whether we're on Windows by checking `path.sep`, which might not be the proper way to do it, but makes sense for this case -- it's precisely when backslash is the path separator that we don't want it acting as escape!)

Finally, a word on error handling.  The `parseQuotesAndEscapes` function will error if it finds an unmatched quote, or if the line ends with an escape character.  However, erroring inside `console-child.js` doesn't produce useful results, so instead I added a redundant call to it from `console.js` which exists purely for validation purposes.  Yes, this means we call it twice -- ideally we'd call it once, save the results, and pass those to `console-child.js`, but due to the weird way `console-child.js` starts up, I'm not sure that's possible.  IDK, I didn't really try it; honestly, I'm not sure we really need to worry about that, I expect input strings to be pretty short.

~~Specifically:~~
1. ~~Quotes are now handled.~~
2. ~~Backslash-escaping works.~~
3. ~~Environment variable substitution with `$` works?  That's not actually something I wanted, but the package I used to accomplish this, [shell-quote](https://www.npmjs.com/package/shell-quote), has no good way to turn this off, so whatever, let's go with it.~~

~~(I mean, there are ways to *sort* of turn it off, but no *good* ways that fully work.  Like, I think the closest you could come would be passing `x => "$" + x` for the `env` argument.  However, in that case, while `$STUFF` would remain as `$STUFF` as desired, `${STUFF}` would get transformed into `$STUFF`.  Which is kind of odd and warty.  I figured, why not, instead of dealing with that, let's just allow evironment variable substitution, why not.)~~

~~Anyway, yeah, I just replaced the spot where we split the arguments with spaces with a call to `shellQuote.parse`, and it takes care of the rest.  One note though -- `shellQuote.parse` will return an *object* when it encounters a bash operator so that you can handle those specially; that's very nice, but we *don't* want to handle those specially at Truffle console (because it would be quite a bit of work!), so I just converted those back to their underlying strings before passing them on.~~

~~Btw, there's one more spot in `console.js` where we split the arguments on spaces; however, I opted to leave that alone, as that's just for purposes of identifying the command.  Notionally I could have replaced it with `shellQuote.parse`, but because the use of the code is limited, it seems to me that the existing code works fine.  I did however add a comment noting this fact.~~

## Testing instructions

I added a bunch of tests for `parseQuotesAndEscapes`.  As for manual testing, here's one way, which is pretty much what I actually used:

1. Make a new directory and unbox metacoin there.
5. Rename `test/metacoin.js` to `test/meta coin.js`.
6. Open up `truffle develop`.
7. Type `test test/meta\ coin.js`.  Observe that this works.

By comparison, you can try this on the `develop` branch and observe that you get an error.

### DIFF
--- a/packages/core/lib/command-utils.js
+++ b/packages/core/lib/command-utils.js
@@ -13,6 +13,90 @@ const managedGanacheDefaultPort = 9545;
 const managedGanacheDefaultNetworkId = 5777;
 const managedDashboardDefaultPort = 24012;
 
+//takes a string and splits it into arguments, shell-style, while
+//taking account of quotes and escapes; the escape character can be
+//customized (you can also pass in more than one valid escape character)
+function parseQuotesAndEscapes(args, escapeCharacters = "\\") {
+  const quoteCharacters = "\"'"; //note we will handle the two quote types differently
+  let argArray = [];
+  let currentArg = "";
+  let currentQuote = undefined;
+  let currentEscape = undefined;
+  let whitespace = true; //are we currently on whitespace? start this as true to allow whitespace at beginning
+  for (const char of args) {
+    if (currentEscape !== undefined) {
+      //escaped character
+      //note that inside quotes, we don't allow escaping everything;
+      //outside quotes, we allow escaping anything
+      if (currentQuote === '"') {
+        //inside a double-quote case
+        if (char === currentQuote) {
+          currentArg += char; //an escaped quote
+        } else {
+          //attempted to escape something not the current quote;
+          //don't treat it as an escape, include the escape char as well
+          currentArg += currentEscape + char;
+        }
+      } else {
+        //outside a quote case
+        //(note there's no single-quote case because we can't reach here
+        //in that case; currentEscape can't get set inside single quotes)
+        currentArg += char; //just the escaped character
+      }
+      currentEscape = undefined;
+      whitespace = false; //(this is not strictly necessary, but for clarity)
+    } else if (escapeCharacters.includes(char) && currentQuote !== "'") {
+      //(unescaped) escape character
+      //(again, inside single quotes, there is no escaping, so we just treat
+      //as ordinary character in that case)
+      currentEscape = char;
+      whitespace = false;
+    } else if (currentQuote !== undefined) {
+      //quoted character (excluding escape/escaped chars)
+      if (currentQuote === char) {
+        //closing quote
+        currentQuote = undefined;
+      } else {
+        //ordinary quoted character, including quote of non-matching type
+        currentArg += char;
+      }
+      whitespace = false; //again not necessary, included for clarity
+    } else if (quoteCharacters.includes(char)) {
+      //(unescaped) opening quote (closing quotes & quoted quotes handled above)
+      currentQuote = char;
+      whitespace = false;
+    } else if (char.match(/\s/)) {
+      //(unescaped) whitespace
+      if (!whitespace) {
+        //if we're already on whitespace, we don't need
+        //to do anything, this is just more whitespace.
+        //if however we're transitioning to whitespace, that means we need
+        //to split arguments here.
+        argArray.push(currentArg);
+        currentArg = "";
+        whitespace = true;
+      }
+    } else {
+      //default case -- ordinary character
+      currentArg += char;
+      whitespace = false;
+    }
+  }
+  //having reached the end of the string, let's check for unterminated quotes & such
+  if (currentQuote !== undefined) {
+    throw new Error(`Error: quote with ${currentQuote} not terminated`);
+  }
+  if (currentEscape !== undefined) {
+    throw new Error(`Error: line ended with escape character ${currentEscape}`);
+  }
+  //now, we push our final argument,
+  //assuming of course that it's nonempty
+  if (currentArg !== "") {
+    argArray.push(currentArg);
+  }
+  return argArray;
+}
+
 // this function takes an object with an array of input strings, an options
 // object, and a boolean determining whether we allow inexact matches for
 // command names - it returns an object with the command name, the run method,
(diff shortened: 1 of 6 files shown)


cds-amal:
Wow, that library leans into regular expressions! Windows uses`\` as its path separator, and IIRC escapes spaces with the caret in Command shell and with the grave accent in PowerShell (or vice versa, it's been a while). Not a blocker, but I should probably test this on Windows.

eggplantzzz:
Seems like it works!

haltman-at:
> Do we have to worry about handling SHELL variables in truffle's REPL?

I mean, this only lets the user do things they could do anyway, as I see it.

> I think it would be sufficient to handle quotes, which would eliminate the need for shell-quote.

Well, once you have quotes, you have to have escaping, in case you want to put quotes in your quotes, so...

(Also, handling quotes can be a bit tricky!)

(Also, IMO, people should totally be able to escape spaces in the Truffle console.)

haltman-at:
(Oops Amal, I accidentally edited your comment at first...)

haltman-at:
Oh, hm, I also just noticed that this library also provides handling of shell comments.  Notionally we could strip those, but I think I'll just not strip those I guess?

cds-amal:
Looks ok, but I want to test it out on Windows, cmd and PowerShell.

cds-amal:
I couldn't get this to work on Windows 11, cmd.exe shell. If I entered `develop> test test\meta^ coin.js` truffle test, errors, unable to locate `chain.js`. I'll revisit this in the morning.

cds-amal (commented):
I couldn't get this to work on either Windows CMD or PowerShell then wound up testing shell-quote itself. The following snippet yielded the same breaking string behavior on both Windows shells, regardless of the escape string setting ("`" or "^").  

```javascript
const sq = require("shell-quote");
const test = `test^ string`;
const res = sq.parse(test, process.env, { escape:  '^'});
console.log(res);
```

```console
PS C:\Users\cdssu\dev\shellQ> node .\index.js
[ 'test', 'string' ]
```

haltman-at:
Hm, that's odd -- it seems like it ought to work.  I'll take a look?

Incidentally, it turns out that since I put up this PR, shell-quote released a new version.  I'll update this PR to that version, I suppose, although it doesn't look to me like the changes should affect the escaping behavior.

haltman-at:
Yeah something's definitely buggy here, though I'm having trouble figuring out why.  Annoying.  May file an issue.

Also: This thing has even more undocumented behavior, sometimes I have to check the `pattern` field, argh.  I'll have to change that (assuming we are indeed sticking with this library).

haltman-at:
OK, no, testing it out, the new release definitely did not fix the problem.

haltman-at:
OK, I submitted three issues ([1](https://github.com/ljharb/shell-quote/issues/5), [2](https://github.com/ljharb/shell-quote/issues/6), [3](https://github.com/ljharb/shell-quote/issues/7)) to the library maintainer; I think for now though I'm just going to tear this out and roll my own.  Oh well!

haltman-at:
OK, I've pushed a commit to replace the library with a hand-rolled function.  Not quite ready for rereview yet, I think I'm definitely need to have to add tests for this function!

haltman-at:
OK yes the error-handling here needs fixing at the least!

haltman-at:
OK, updated this!

cds-amal (commented):
I'll be able to do a proper review tomorrow, but want to mention [c-escape sequences, like `\n`, `\r`, and others](https://en.wikipedia.org/wiki/Escape_sequences_in_C#Table_of_escape_sequences). Like transforming `\\n` -> `\n`. Should we care about this, for a REPL?

haltman-at:
Well, I'm basically trying to mimic Bash, and it doesn't do that.  So to my mind the answer is no.  Although yeah that does mean entering newlines isn't really possible.  Oh well? :shrug:

cds-amal:
Ok, finally got around to testing this on Windows in CMD and PowerShell, using `test\meta coin.js` as the target file. It worked for both `test test\meta^ coin.js` (caret) and `test test\meta`` coin.js` (grave). Note: looks weird because I don't know how to escape the grave accent in GH markdown.

It is unclear to me why it works for the grave accent char.

haltman-at:
I mean it works that way on Windows because I made it work that way; have you seen the changes to the code since you last reviewed it?

cds-amal (approved):
Oops, missed the grave accent char in the ternary expression!

cds-amal:
> I mean it works that way on Windows because I made it work that way; have you seen the changes to the code since you last reviewed it?

Oops, missed the little guy next to the caret.

PR #5864, opened 2023-02-02
Add appendCBOR field in metadata settings type in source fetcher

haltman-at:
Just keeping this type up to date with the latest Solidity compiler options.  This change has no actual effect at the moment.


cds-amal (approved):
Nice.

PR #5866, opened 2023-02-02
Add `block.prevrandao` as globally available variable in debugger

haltman-at:
Solidity 0.8.18 introduces `block.prevrandao` as an alias for `block.difficulty`, so this PR adds support for that in the debugger.  Or rather in codec, because that's actually where this is handled.

Note that we do a version check, so `block.prevrandao` will only be present on versions 0.8.18 and later; that means that 0.8.18 is another one of those versions we care about.  Strictly speaking, we arguably should also check the version of the EVM that's being targeted, but that potentially opens some cans of worms I didn't want to deal with, so I left that out for now; I figure it can be added in if people really think it's necessary or we decide the resulting worms aren't a big deal.

I also updated the globally-available variable tests in the debugger to include it.

## Testing instructions

I updated the debugger tests, but you might also want to test this manually.  I tested this manually with the `solc8-magic-square` project in the Solidity test examples repo.   I compiled it both with Solidity 0.8.0, ran a transaction, debugged it, verified that `block.prevrandao` is not there.  Then I recompiled with Solidity 0.8.18, did the same thing, and verified that it is there and equal to `block.difficulty`.

### DIFF
--- a/packages/codec/lib/special/decode/index.ts
+++ b/packages/codec/lib/special/decode/index.ts
@@ -103,8 +103,21 @@ export function* decodeMagic(
           info
         )
       };
-      //the other ones are all uint's, so let's handle them all at once; due to
-      //the lack of generator arrow functions, we do it by mutating block
+      //now we handle prevrandao. this one gets special handling
+      //because the name isn't equal to the location name (it's an alias of difficulty)
+      if (solidityVersionHasPrevrandao(info.currentContext.compiler)) {
+        block.prevrandao = yield* Basic.Decode.decodeBasic(
+          {
+            typeClass: "uint" as const,
+            bits: 256
+          },
+          { location: "special" as const, special: "difficulty" },
+          info
+        );
+      }
+      //the other ones are all uint's and all work the same, so let's handle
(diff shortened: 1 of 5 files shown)


packages/codec/lib/special/decode/index.ts
```diff
@@ -203,3 +217,19 @@ function solidityVersionHasBaseFee(
       return true;
   }
 }
+
+function solidityVersionHasPrevrandao(
+  compiler: Compiler.CompilerVersion
+): boolean {
+  switch (Compiler.Utils.solidityFamily(compiler)) {
```
eggplantzzz, line 224: do you think it may be useful to just add a comment here that Solidity versions above 0.8.18 have `block.prevrandao`?

haltman-at: Hm, might be nice, yeah -- I could put it on all of these functions.  Not sure it's essential, might skip doing it unless I'm making some other change?

haltman-at:
Alright, going ahead and merging this because I really want this to make it into the release and @cds-amal and @gnidan are seemingly too sick to look it over. :P

Issue #2041, opened 2019-05-21
Running "help" inside truffle develop run incorrectly


eggplantzzz:
When using the Truffle development console, if you run `help`, it will output the help for the help command and not the general help that you would get from running `truffle help` at the command line.

fainashalts:
@CruzMolina this is assigned to you, we're doing issue maintenance so I just wanted to see if it went anywhere!

CruzMolina:
Woops. Completely forgot about this! Still open for anyone to take!

cds-amal:
v5.5.11 displays help and exits the develop-console!

```console

+ truffle@5.5.11
added 1245 packages from 897 contributors in 105.796s

$ truffle develop
Truffle Develop started at http://127.0.0.1:9545/

<...SNIP ACCOUNT DETAILS>

truffle(develop)> help
truffle [command]

Commands:
  truffle build      Execute build pipeline (if configuration present)
  truffle compile    Compile contract source files
  truffle config     Set user-level configuration options
  truffle create     Helper to create new contracts, migrations and tests
  truffle dashboard  Start Truffle Dashboard to sign development transactions
                     using browser wallet
  truffle db         Database interface commands
  truffle debug      Interactively debug any transaction on the blockchain
  truffle deploy     (alias for migrate)
  truffle exec       Execute a JS module within this Truffle environment
  truffle help       List all commands or provide information about a specific
                     command
  truffle install    Install a package from the Ethereum Package Registry
  truffle migrate    Run migrations to deploy contracts
  truffle networks   Show addresses for deployed contracts on each network
  truffle obtain     Fetch and cache a specified compiler
  truffle opcode     Print the compiled opcodes for a given contract
  truffle preserve   Save data to decentralized storage platforms like IPFS and
                     Filecoin
  truffle publish    Publish a package to the Ethereum Package Registry
  truffle run        Run a third-party command
  truffle test       Run JavaScript and Solidity tests
  truffle version    Show version number and exit
  truffle watch      Watch filesystem for changes and rebuild the project
                     automatically

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
$  
```

eggplantzzz:
Oh, that looks mostly correct other than the exiting of the console.

gnidan:
The fix for this went out in Truffle v5.7.6 and I noticed that this isn't working right (`truffle` is getting listed as `consoleChild.js`). Holding off on closing this until we can investigate.
[...]

PR #5871, opened 2023-02-06
Bug fix: create utility for dealing with help input

eggplantzzz:
## PR description

addresses https://github.com/trufflesuite/truffle/issues/2041

This PR creates a utility method for parsing help input. It then uses that same logic in console-child.js. This is so that we use the same logic for parsing help input from both the command line and the development console (repl).

I also found out that yargs uses `console.error` for the `showHelp` method which we use to display general help in Truffle. This was causing a bug where general help would not print in the development console as the stderr was not "hooked up". So this PR also passes the "log" argument to this method to force it to use `console.log`.

## Testing instructions

To test this, checkout this branch and run the development console. Enter "help" and you can see that the general help is printed to the screen.

## Documentation

### DIFF
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -39,8 +39,6 @@ if (!semver.gte(process.version, minimumNodeVersion)) {
 const listeners = process.listeners("warning");
 listeners.forEach(listener => process.removeListener("warning", listener));
 
-const inputStrings = process.argv.slice(2);
-
 const { handleHelpInput } = require("./lib/cliHelp");
 const {
   getCommand,
@@ -49,8 +47,9 @@ const {
   displayGeneralHelp
 } = require("./lib/command-utils");
 
+const inputArguments = process.argv.slice(2);
 // handle cases where input indicates the user wants to access Truffle's help
-const { displayHelp } = handleHelpInput({ inputStrings });
+const { displayHelp, inputStrings } = handleHelpInput({ inputArguments });
 if (displayHelp) {
   displayGeneralHelp();
   process.exit();
--- a/packages/core/cli.js
+++ b/packages/core/cli.js
@@ -41,6 +41,7 @@ listeners.forEach(listener => process.removeListener("warning", listener));
 
 const inputStrings = process.argv.slice(2);
 
+const { handleHelpInput } = require("./lib/cliHelp");
 const {
   getCommand,
   prepareOptions,
@@ -48,31 +49,11 @@ const {
   displayGeneralHelp
 } = require("./lib/command-utils");
 
-//User only enter truffle with no commands, let's show them what's available.
-if (inputStrings.length === 0) {
+// handle cases where input indicates the user wants to access Truffle's help
+const { displayHelp } = handleHelpInput({ inputStrings });
+if (displayHelp) {
   displayGeneralHelp();
   process.exit();
-}
-
-//if `help` or `--help` is in the command, validate and transform the input argument for help
-if (
-  inputStrings.some(inputString => ["help", "--help"].includes(inputString))
-) {
-  //when user wants general help
-  if (inputStrings.length === 1) {
-    displayGeneralHelp();
-    process.exit();
-  }
-
-  //check where is --help used, mutate argument into a proper help command
-  const helpIndex = inputStrings.indexOf("--help");
-
-  if (helpIndex !== -1) {
-    //remove `--help` from array
-    inputStrings.splice(helpIndex, 1);
-    //insert `help` in first position
-    inputStrings.unshift("help");
-  }
 }
 
 const command = getCommand({
--- a/packages/core/lib/cliHelp.js
+++ b/packages/core/lib/cliHelp.js
@@ -1,29 +1,37 @@
-const handleHelpInput = function ({ inputStrings }) {
+const handleHelpInput = function ({ inputArguments }) {
+  // we call slice to make a copy of the array so as not to mutate it
+  const args = inputArguments.slice();
   // if no command was provided, display all available commands
-  if (inputStrings.length === 0) {
-    return { displayHelp: true };
+  if (args.length === 0) {
+    return {
+      displayHelp: true,
+      inputStrings: args
+    };
   }
   // if only 'help' or '--help' was entered, display all available commands
-  if (
-    inputStrings.length === 1 &&
-    (inputStrings[0] === "help" || inputStrings[0] === "--help")
-  ) {
-    return { displayHelp: true };
+  if (args.length === 1 && (args[0] === "help" || args[0] === "--help")) {
+    return {
+      displayHelp: true,
+      inputStrings: args
+    };
   }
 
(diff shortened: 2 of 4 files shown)


packages/core/lib/cliHelp.js
```diff
@@ -0,0 +1,33 @@
+const handleHelpInput = function ({ inputStrings }) {
+  let displayHelp = false;
+  //User only enter truffle with no commands, let's show them what's available.
+  if (inputStrings.length === 0) {
+    displayHelp = true;
```
cds-amal, line 5: The code can return `true` here.

cds-amal (changes requested):
Some small nits.

packages/core/lib/command-utils.js
```diff
@@ -213,7 +213,8 @@ const displayGeneralHelp = () => {
     // Exclude "install" and "publish" commands from the generated help list
     // because they have been deprecated/removed.
     if (command !== "install" && command !== "publish") {
-      yargs.command(require(`./commands/${command}/meta`));
+      const thing = require(`./commands/${command}/meta`);
+      yargs.command(thing);
```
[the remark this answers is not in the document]
eggplantzzz: yup, whoops

packages/core/lib/cliHelp.js
eggplantzzz: No it can't, that does not conform to the method signature as I wrote it :)

packages/core/lib/cliHelp.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+  // handle 'truffle help' and 'truffle --help'
+  if (
+    inputStrings.length === 1 &&
+    (inputStrings[0] === "help" || inputStrings[0] === "--help")
+  ) {
+    displayHelp = true;
```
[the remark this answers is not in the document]
eggplantzzz: same as the reply to your comment above

packages/core/lib/cliHelp.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  //if `--help` is in the input, validate and transform the input
+  //in order to give the user help
+  if (inputStrings.some(inputString => ["--help"].includes(inputString))) {
+    //check where --help is used, mutate argument into something the rest
+    //of Truffle can deal with
+    const helpIndex = inputStrings.indexOf("--help");
+
+    if (helpIndex !== -1) {
```
cds-amal, line 22: This logic can be simplified if you rely on `string.indexOf` to determine if `--help` exists as well as its index
```
  const helpIndex = inputStrings.indexOf("--help");
  if (helpIndex !== -1) {
    // ...
  }
```

packages/core/lib/cliHelp.js
```diff
@@ -0,0 +1,33 @@
...
+    inputStrings.length === 1 &&
+    (inputStrings[0] === "help" || inputStrings[0] === "--help")
+  ) {
+    displayHelp = true;
+  }
+
+  //if `--help` is in the input, validate and transform the input
+  //in order to give the user help
+  if (inputStrings.some(inputString => ["--help"].includes(inputString))) {
+    //check where --help is used, mutate argument into something the rest
```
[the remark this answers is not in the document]
eggplantzzz: No, that is only when we want to display general help to the user. If this condition is true, it means that what they entered is of the following form:
```
truffle <someString>(x times) --help <someOtherString>(n times)
```
This is where `n` is an integer greater than or equal to 0.

packages/core/lib/cliHelp.js
eggplantzzz: I'm not sure I understand what you want to change here. You wrote the exact same thing as the code itself.

packages/core/lib/cliHelp.js
eggplantzzz: oh do you mean to return `{ displayHelp: true }` in both of these cases?

cds-amal: Yyes, `{ displayHelp: true } ` in both of those cases.

packages/core/lib/cliHelp.js
cds-amal: Removing the condition: `if (inputStrings.some(inputString => ["--help"].includes(inputString)))` simplifies the logic. This works because `Array.some()` and `Array.indexOf()` can both test if an item exists in the collection. 

```javascript
    const helpIndex = inputStrings.indexOf("--help");
    if (helpIndex !== -1) {
      inputStrings.splice(helpIndex, 1);
      inputStrings.unshift("help");
    }
    return { displayHelp };
```

packages/core/lib/cliHelp.js
eggplantzzz: Sure but that is more a stylistic thing than anything

packages/core/lib/cliHelp.js
eggplantzzz: Ah yes, that makes sense. We should only check once. Gotcha!

packages/core/cli.js
```diff
@@ -41,40 +41,21 @@ listeners.forEach(listener => process.removeListener("warning", listener));
...
   getCommand,
   prepareOptions,
   runCommand,
   displayGeneralHelp
 } = require("./lib/command-utils");
 
-//User only enter truffle with no commands, let's show them what's available.
-if (inputStrings.length === 0) {
+// handle cases where input indicates the user wants to access Truffle's help
+const { displayHelp } = handleHelpInput({ inputStrings });
```
cds-amal, line 53: `handleHelpOutput` is too innocent a name for what the function does. It returns a boolean and mutates `inputString`.  What do you think about different name, and return type, so there's no side effect? (`line 42: const inputStrings = ...` would have to be deleted, though.)

```
const { displayName, inputStrings } = parseArguments(proces.argv.slice(2));
```

packages/core/lib/cliHelp.js
cds-amal: I consider it to be a readability refactor. The current iteration makes it easier to understand the success path of the function. It reduces the need to track `displayHelp`'s state throughout the function, as well as clearly outlining the exceptional, failure, cases with the early returns.

packages/core/cli.js
eggplantzzz: I don't know, I kind of feel like this is not worth agonizing over. `handleHelpInput` doesn't bother me, maybe a comment about what it does might suffice?

cds-amal: I'd rather refactor this utility/helper function to be a pure function because it's now a module, and it adds another bit of technical debt. If others think a comment would suffice, I'll be ok with that.  @fainashalts @haltman-at, @gnidan, thoughts?

eggplantzzz: I guess what I'm really saying is that in general I think our time would be better spent fixing technical debt on much more important and complicated features rather than agonizing over help display. However, I'm not going to die on this hill so will rewrite it to be pure.

PR #5883, opened 2023-02-11
Set yargs help to always use truffle instead of script name

cds-amal:
This PR sets the script name to always use `truffle` instead of the default, `process.argv[0]`.  This is related to #2041 

Test this by running help in `truffle develop`, it should display commands as `truffle <cmd>` instead of `console-child.js <cmd>`


haltman-at (approved):
Makes sense!

Issue #5312, opened 2022-07-15
`help` in Truffle console is not useful


cliffoo:
## Steps to Reproduce

- `truffle console`
- `help` in the Truffle console
- Get this:
```
$ truffle console     
truffle(development)> help
consoleChild.bundled.js [command]

Commands:
  consoleChild.bundled.js help  List all commands or provide information about a
                                specific command

Options:
  --help     Show help                                                 [boolean]
  --version  Show version number                                       [boolean]
```

## Expected Behavior

Give the same output as `truffle help`

cliffoo:
Related to [issue#2041](https://github.com/trufflesuite/truffle/issues/2041)

PR #5885, opened 2023-02-11
Restrict general help content in REPL context

cds-amal:
## PR description

This PR limits the general help output for `truffle console` and `truffle develop` so help for `console`, `dashboard`, `db`, `develop`, `init` and `watch` won't be shown. It achieves this by defining a single source of truth for valid CLI and Console commands. The `commands/commands.js` module defines this state, and other parts of the codebase have been refactored to import and rely on this information. The Console class no longer needs to be passed `allowedCommands` and has removed the maintenance of `excludedCommands`. Instead, the signature for the help function has changed to indicate whether it is displaying help for the CLI or the REPL.  Tests and webpack concerns have been updated to reflect these changes, and the source of truth for valid Truffle commands has been updated in various files. 

Contributes to: #5312

## Testing instructions

- Try `help` in truffle develop, and confirm it doesn't show help for `console`, `dashboard`, `db`, `develop`, `init` and `watch`
- Verify help isn't broken :/

## Documentation & breaking changes concerns
- No docs required.
- Error text changed in REPL output, so it's arguable this might be a breaking change. Thoughts?

### DIFF
--- a/packages/core/lib/command-utils.js
+++ b/packages/core/lib/command-utils.js
@@ -5,7 +5,10 @@ const { extractFlags } = require("./utils/utils"); // contains utility methods
 const globalCommandOptions = require("./global-command-options");
 const debugModule = require("debug");
 const debug = debugModule("core:command:run");
-const { validTruffleCommands } = require("./commands/commands");
+const {
+  validTruffleCommands,
+  validTruffleConsoleCommands
+} = require("./commands/commands");
 const Web3 = require("web3");
 const TruffleError = require("@truffle/error");
 
@@ -294,8 +297,16 @@ const runCommand = async function (command, options) {
   return await command.run(options);
 };
 
-const displayGeneralHelp = ({ commands = validTruffleCommands }) => {
+/**
+ * Display general help for Truffle commands
+ * @param {Object} options - options object
+ * @param {Boolean} options.isREPL - whether or not the help is being displayed in a REPL
+ * @returns {void}
+ */
+const displayGeneralHelp = options => {
   const yargs = require("yargs/yargs")();
+  const isREPL = options?.isREPL ?? false; //default to not displaying REPL commands
+  const commands = isREPL ? validTruffleConsoleCommands : validTruffleCommands;
   commands.forEach(command => {
     // Exclude "install" and "publish" commands from the generated help list
     // because they have been deprecated/removed.
--- a/packages/core/lib/command-utils.js
+++ b/packages/core/lib/command-utils.js
@@ -5,7 +5,7 @@ const { extractFlags } = require("./utils/utils"); // contains utility methods
 const globalCommandOptions = require("./global-command-options");
 const debugModule = require("debug");
 const debug = debugModule("core:command:run");
-const commands = require("./commands/commands");
+const { validTruffleCommands } = require("./commands/commands");
 const Web3 = require("web3");
 const TruffleError = require("@truffle/error");
 
@@ -116,11 +116,11 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
   // for inferring the command.
   if (firstInputString === "-v" || firstInputString === "--version") {
     chosenCommand = "version";
-  } else if (commands.includes(firstInputString)) {
+  } else if (validTruffleCommands.includes(firstInputString)) {
     chosenCommand = firstInputString;
   } else if (noAliases !== true) {
     let currentLength = 1;
-    const availableCommandNames = commands;
+    const availableCommandNames = validTruffleCommands;
 
     // Loop through each letter of the input until we find a command
     // that uniquely matches.
@@ -294,7 +294,7 @@ const runCommand = async function (command, options) {
   return await command.run(options);
 };
 
-const displayGeneralHelp = () => {
+const displayGeneralHelp = ({ commands = validTruffleCommands }) => {
   const yargs = require("yargs/yargs")();
   commands.forEach(command => {
     // Exclude "install" and "publish" commands from the generated help list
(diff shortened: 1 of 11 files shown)


packages/core/lib/console.js
```diff
@@ -16,13 +16,13 @@ const { spawn } = require("child_process");
...
 const { getCommand, parseQuotesAndEscapes } = require("./command-utils");
-const validTruffleCommands = require("./commands/commands");
+const { validTruffleConsoleCommands } = require("./commands/commands");
 
 // Create an expression that returns a string when evaluated
 // by the REPL
 const makeIIFE = str => `(() => "${str}")()`;
 
-const processInput = (input, allowedCommands) => {
+const processInput = input => {
```
cds-amal, line 25: While working this PR I was tempted to refactor REPL input processing because it is all over the place, misnamed and should be consolidated. I'll add an issue for that.

haltman-at:
> * Error text changed in REPL output, so it's arguable this might be a breaking change. Thoughts?

I have to say I don't see how that's a breaking change.

packages/core/lib/command-utils.js
```diff
@@ -116,11 +119,11 @@ const getCommand = ({ inputStrings, options, noAliases }) => {
   // for inferring the command.
   if (firstInputString === "-v" || firstInputString === "--version") {
     chosenCommand = "version";
-  } else if (commands.includes(firstInputString)) {
+  } else if (validTruffleCommands.includes(firstInputString)) {
```
haltman-at, line 122: Isn't this function called in both shell and REPL contexts?

packages/core/lib/commands/commands.js
```diff
@@ -26,3 +25,31 @@ module.exports = [
   "version",
   "watch"
 ];
+
+//Subset of truffle commands that are allowed to run in console REPLS.
+//Excluded commands are:
+//  console, dashboard, db, develop, init and watch
+const validTruffleConsoleCommands = [
```
haltman-at, line 32: Not sure it matters much but IMO it would be better to have an excluded commands list (which this could then perhaps be built from) because consider what happens when someone needs to add something.  Excluded commands are the exceptional case, right?  So they're the ones that should get "you need to update two things for this".

packages/core/lib/console.js
```diff
@@ -39,20 +39,19 @@ const processInput = (input, allowedCommands) => {
...
-    if (validTruffleCommands.includes(normalizedCommand)) {
-      return allowedCommands.includes(normalizedCommand)
-        ? words.slice(1).join(" ")
-        : makeIIFE(`ℹ️ : '${cmd}' is not allowed within Truffle REPL`);
-    }
-    return makeIIFE(`ℹ️ : '${cmd}' is not a valid Truffle command`);
+    return validTruffleConsoleCommands.includes(normalizedCommand)
+      ? words.slice(1).join(" ")
+      : makeIIFE(
+          `ℹ️ : '${words[0]} ${cmd}' is not valid in Console environment.`
```
haltman-at, line 45: You've combined these two error messages into one, but I feel like the second is worth keeping...

...except, does the second ever come up?  I don't think it does, right?  Because it'll get interpreted as JS and you'll get an error from that instead, right?

...OK, I guess maybe that error message can go after all!  This might be worth a comment though?

packages/truffle/test/scenarios/commands/develop.js
```diff
@@ -78,19 +78,19 @@ describe("truffle develop", function () {
...
       },
       {
         cmd: "develop",
-        expectedError: `ℹ️ : 'develop' is not allowed within Truffle REPL`
+        expectedError: `ℹ️ : 'truffle develop' is not valid in Console environment.`
       },
       {
         cmd: "alakazam",
-        expectedError: `ℹ️ : 'alakazam' is not a valid Truffle command`
+        expectedError: `ℹ️ : 'truffle alakazam' is not valid in Console environment`
```
haltman-at, line 93: Oh, here's that error message though...?  If this error message ever actually does come up, I feel like it's misleading to say it's "not valid in a console environment" when in fact it's not valid at all!

packages/core/lib/command-utils.js
cds-amal: I missed that, yuck! The REPL's use is [guarded by `validTruffleConsoleCommands`](https://github.com/trufflesuite/truffle/blob/help-less/packages/core/lib/console.js#L401-L406).  It seems redundant to invoke `getCommand()` if it is already a valid console command. I'm going to remove that.

packages/core/lib/commands/commands.js
cds-amal: Good point. I wanted to remove duplicated derivation logic with this PR. I could see maintaining an exclusion list in this file, using it to derive `validTruffleConsoleCommands` once. Would that be ok?

packages/core/lib/console.js
cds-amal: Another good point. I'll think this one over and get back tomorrow.

packages/truffle/test/scenarios/commands/develop.js
cds-amal: Yea, that is misleading. :thinking:

packages/core/lib/commands/commands.js
haltman-at: Sounds good to me.

packages/core/lib/console.js
cds-amal: The latest changes distinguishes between excluded Truffle Commands, and invalid Truffle commands. Thanks for suggesting maintaining an exception list.

packages/truffle/test/scenarios/commands/develop.js
cds-amal: Addressed in prior comment.

haltman-at (approved):
Makes sense to me.  What *was* that `getCommand` doing there, I wonder?  (Actually, I'd bet it was the older code and then the guard was added before it.  Hm...)

Issue #3839, opened 2021-02-20
Support ENS reverse resolution when decoding addresses in Truffle Codec/Decoder


haltman-at:
## Issue

Truffle Codec, when decoding variables of address or contract type, should support ENS reverse resolution.  It would of course need to make a request for this rather than contacting ENS itself.  (So, Truffle Decoder would have to handle that.  It would have to include a cache of course.)

Note that ENS names can be malformed UTF-8, so we have to account for that possibility.  To keep this from being too much of a breaking change, we can have the response type be `Uint8Array | null` (`null` meaning that it's not on ENS).

haltman-at:
Released in 5.9.0!

PR #5895, opened 2023-02-15
Reverse ENS resolution support in debugger and decoder

haltman-at:
Addresses #3839.  **Breaking change to `@truffle/codec`.**

This PR adds support for reverse ENS resolution in Truffle Codec, and in Truffle Decoder and Truffle Debugger that use it.  This means that, if enabled, values of address or contract type that have an ENS name will have an `ensName` interpretation in their (newly-added) `interpretations` field, and values of external function type will have something similar.

(Apologies for this PR being a big ball of messy commits.  You know how it goes with big PRs...)

In order to accomplish this, a third type of decoding request was added to Truffle Codec, the ENS reverse resolution request.  This request will be made for all values of address, contract, or external function type, and it is up to the caller to handle them.  **That makes this a breaking change to `@truffle/codec`**, as callers *must* deal with this new request type.  They can respond with `null` if they don't support ENS resolution, but they must respond.

(Notionally, this could have been done in a nonbreaking way, by not making such requests unless an appropriate option was passed.  I didn't see any particular reason to do this, however.)

So, what's new here?

Well, first off, value results (not errors) now include an `interpretations` field like decodings do.  For most types this is at present always empty.  However, for addresses and contracts there is now the `ensName` interpretation, and for external functions there's the `contractEnsName` interpretation.  Like all interpretations, it's optional and will only be present when relevant.  Note that the value of this interpretation is a `StringValueInfo`, rather than a `string`, in order to account for the possibility that the ENS name consists of malformed UTF-8 (even though in reality that won't come up in this implementation).

Now you may wonder, why is this in `interpretations` rather than `value`?  And if this *is* in `interpretations` rather than `value`, shouldn't that also be true of a bunch of the other stuff currently in `value`?  Well... yes!  But it's not our intent to just go breaking everything, making breaking decoder or debugger changes (just codec), so basically stuff already in `value` is grandfathered in, and new stuff will generally go under `interpretations`.

You might also wonder: If we're doing this, shouldn't address values also get an interpretation with their contract type when available, effectively unifying address values with contract values (in information content if not in format)?  Yes!  We should also do that!  But in order to reduce complication, that is being pushed off to a future PR.

Then, as mentioned, there's the new request type.  I didn't want to go fundamentally changing how decoding requests work, so the response type here is still `Uint8Array | null` rather than anything more sensible.  But you respond with a `Uint8Array` (in UTF-8) if there is an ENS name, and `null` if there isn't or ENS reverse resolution isn't turned on.

Note that it is up to the caller to perform a forward-resolution check!  There is no separate request type for forward resolution (well, that exists as an *encoding* request type, but not decoding); the caller simply has to know that as part of reverse resolution you must confirm with a forward check.  The debugger and decoder will both respond with `null` should the forward check fail.

Of course, in addition to the updates to the debugger and the decoder to handle this new request type -- which I'll get back to in a moment -- we've also got to update the inspectors (as well as `abify.ts`).  So, `ResultInspector` will now print out an ENS names, colored light blue, rather than an address, when it's available.  However in some contexts (such as the debugger) we don't want to hide the address, so, the various inspector constructors now take an optional `options` argument, which allows one to pass options such as `noHideAddress` (the only option at present).  I believe @cds-amal intends to do more with these options in the future, however. :)

(Thought: Should I have used subclasses instead of flags?  Oh well.)

Note there are more places we could put reverse ENS resolution in the debugger -- for instance, in the affected instances display, or stacktraces, or etc.  So the reverse ENS resolution support in the debugger is rater partial as it pertains to decoded values only.  But for obvious reasons, those other places we could support it were considered out of scope for this PR.

Also, I left `interpretations` blank for all values created by `wrap`.  I thought about including an `ensName` interpretation when the input was an ENS name, but 1. this is actually incompatible with the current architecture, and 2. it's not useful anyway.  So I didn't do it.

With all that out of the way, let's discuss the changes to the decoder and the debugger!  (Including the debugger CLI.)

Let's start with the decoder because it's simpler.  The `ProjectDecoder` now gains an `init` method where ENS is set up, and the methods that create `ProjectDecoder`s are updated to call it beofre returning the decoder to you.  Also, decoding methods now handle `"ens"` requests using ENSJS, on top of which I've stuck a cache.

(I didn't bother with a separate cache for forward requests, I only put a combined one for reverse-requests-together-with-forward-requests, but I could add one for forward requests too?)

Note that reverse ENS resolution is turned on by default; if you want to turn it off, you have to do so manually by specifying `provider: null` in the ENS options.  In the debugger CLI, I added two new command-line options; `--no-ens` to turn off ENS support, and `--registry` to specify a registry address other than the default.

Btw, you'll notice some code repeated here from the encoder related to determining the default registry address based on the network ID (and you'll also see this in debugger).  ENSJS is *supposed* to do this automatically, but in actual fact, it does not, and so we have to work around this bug.  This workaround didn't seem worth factoring, so it's just repeated several times.  Sorry.

In the debugger, things are more complicated.  The basic idea is the same as in the decoder, but the details are a bit messier.  The debugger now has an `ensRegistryAddress` option, alongside the light mode option, to set the ENS registry address; you can also set it to `null` to turn off reverse ENS resolution.

Adding reverse ENS resolution meant adding not only a new adapter method, but also all the machinery in the `web3` saga to interface with that method.  I'll skip going into the details here.  One thing worth noting though: You'll note that the web3 saga now ends with putting an action signalling that the web3 submodule is ready, which it didn't before.  Turns out we had a synchronization bug!  This ENS work exposed it and so I had to fix it.

Also, there was the question of where in the state to put the ENS cache (again, I did the cache same way as in the decoder, I didn't bother separately caching forward requests).  After talking to @gnidan we decided that the appropriate location was a new submodule, `ens`.  This submodule doesn't have a listener though.  It only does things when `data` asks it to.  So it's not adding a lot more complication.

Finally, let's talk about the tests, which I think I have to due to some annoying problems with them.  Now, these are not the first tests that involve setting up ENS resolution -- encoder tests already did that.  But I did it a different way for the decoder and debugger tests, and it seems to have caused a problem.

For the debugger tests, the tests would not pass unless Ganache was set to eager mode.  (Thanks to @cds-amal for finding this.)  So, to get this up, that's what I did.  I hope that's OK.  @cds-amal thinks there may be a Ganache issue?  I didn't follow this.  (Note that my manual tests pass without problem!)

For the decoder tests -- which I put in a separate folder, `ens`, separate from `current`, to avoid any interference -- not only did we have to set eager mode, but also, to prevent the tests from hanging after finishing, we had to stick a `--exit` on the Mocha invocation.  (Again, thanks to @cds-amal for finding this.)  That's a bit worrying, I have to admit; but I really want to get this PR up, so, whatever, that's what I'm going with.  Hopefully we can figure out what's going on with that and find a better way?

One final minor note -- because we're now lossily converting a `StringValueInfo` to a `string` in more places, I factored out a function for doing that. It's also exported from Codec if consumers of the library want to use it!

## Testing instructions

There are new automated tests in both the decoder and debugger packages (about which see above).  But you'll likely want to test this manually too.  Here's how I did it:

1. Go to `ens-test` in the `solidity-test-cases` repo.
2. Compile, migrate.
3. Open up the Truffle console, get the `EnsTest` instance, and call `instance.registryAddress()` to get the registry address.
4. Run `instance.run()`.
5. Debug the resulting transaction, using `--registry` to set the registry address to the address you got above.
6. Step through, decode variables, and observe! :)


gnidan:
Quick comment here before I get you a full review: MetaMask has raised the concern that @truffle/decoder is too big with all its dependencies (something like 15% of total MM bundle size), so we **must not** increase the number of dependencies at this time.

This PR might not add net new dependencies at all? We'll have to verify that. But if this does add stuff, then we'll need to rework this to use ethers.js directly and not add ens.js or anything else.

packages/codec/lib/format/values.ts
```diff
@@ -439,6 +449,9 @@ export interface FunctionExternalValue {
   type: Types.FunctionExternalType;
   kind: "value";
   value: FunctionExternalValueInfo;
+  interpretations: {
+    contractEnsName?: StringValueInfo;
```
gnidan, line 453: It's weird that this is an interpretation here. I guess it's too much trouble to move this closer to where the address field appears?

packages/codec/lib/format/index.ts
```diff
@@ -37,8 +37,8 @@
  * types.
  *
  * This format targets types and values understood by the
- * [Solidity programming language](https://solidity.readthedocs.io) and
- * the [Contract ABI specification](https://solidity.readthedocs.io/en/v0.5.3/abi-spec.html),
+ * [Solidity programming language](https://docs.soliditylang.org/) and
+ * the [Contract ABI specification](https://docs.soliditylang.org/en/v0.8.17/abi-spec.html),
```
gnidan, line 41: Of course you probably now want to update this to .19 :)

packages/codec/lib/format/index.ts
```diff
@@ -80,7 +80,18 @@
...
  *     more information.
  *
- *   5. `reference`: This field is a debugger-only feature and does not
+ *   5. `interpretations`: This field will also be present when `kind` is equal
+ *     to `"value"`; it is an object that may contain additional information
+ *     about the decoded value beyond what is found in `value`.  (All fields in
+ *     `interpretations` are always optional and will be included only when
+ *     applicable.)  You may wonder, what is the difference between `value`
+ *     and `interpretations`?  The answer is that the distinction is largely
+ *     historical; `interpretations` was not originally part of the format, and
```
gnidan, line 89: I'm not sure this is the best place to include this note about historical reasoning... is there somewhere else we can note it? Since it's not immediately relevant here, I'd wager.

packages/codec/lib/format/utils/inspect.ts
```diff
@@ -23,6 +23,10 @@ function cleanStylize(options: InspectOptions): InspectOptions {
   return clonedOptions;
 }
 
+export interface ResultInspectorOptions {
+  noHideAddress?: boolean;
```
gnidan, line 27: If this appears in the docs, we might want to add a doc comment about this.

packages/codec/lib/types.ts
```diff
@@ -685,6 +685,16 @@ export interface CodeRequest {
   address: string;
 }
 
+/**
+ * A request for an ENS name
+ *
+ * @Category Requests
+ */
+export interface EnsRequest {
```
gnidan, line 693: I think this should include something about reverse resolution or "primary name" in the identifier here, just so it's more clear. Maybe `EnsPrimaryNameRequest`? I think I prefer that over `EnsReverseResolutionRequest`, but maybe go with whatever terminology you use elsewhere

packages/core/lib/debug/cli.js
```diff
@@ -133,13 +133,15 @@ class CLIDebugger {
     const startMessage = DebugUtils.formatStartMessage(
       this.txHash !== undefined
     );
+    const registry = this.config.noEns ? null : this.config.registry;
```
gnidan, line 136: We have ENS registry configuration options elsewhere in truffle-config.js. Is that scoped specifically to the deployer? I'm not opposed to merging this with the duplication, but could you please at least open an issue to note this as something we'll want to resolve?

@eggplantzzz might be the person to ask about this... it doesn't look like it's in our configuration docs.

gnidan: (Oh, that is to say... keeping `--registry` is fine as a CLI option, but if the user doesn't specify this, then `truffle debug` should first look to see if it can find the setting in the config)

packages/debugger/lib/data/sagas/index.js
```diff
@@ -1082,6 +1083,8 @@ export function* decode(definition, ref, compilationId) {
       case "code":
         response = yield* evm.requestCode(request.address);
         break;
+      case "ens":
```
gnidan, line 1086: Similarly to my point about the identifier name, I think this key should be more explicit, like `"ens-primary-name"` or such. (Yes, I prefer kebab-case ~~everywhere I can get away with it~~ here.)

packages/debugger/lib/debugger.js
```diff
@@ -28,15 +29,22 @@ const Debugger = {
...
-    let { contracts, files, provider, compilations, lightMode, storageLookup } =
-      options;
+    const {
+      contracts,
+      files,
+      provider,
+      compilations,
+      lightMode,
+      storageLookup,
+      ensRegistryAddress
```
gnidan, line 39: Can we make an `ens` namespace here and make `registryAddress` a sub-field of that? I have a strong feeling we'll want to add things here later.

packages/debugger/lib/debugger.js
```diff
@@ -52,12 +60,20 @@ const Debugger = {
...
    */
   forProject: async function (options = {}) {
-    let { contracts, files, provider, compilations, lightMode } = options;
+    const {
+      contracts,
+      files,
+      provider,
+      compilations,
+      lightMode,
+      ensRegistryAddress
```
gnidan, line 69: (Same request for `ens` namespace here)

packages/debugger/lib/ens/actions/index.js
```diff
@@ -0,0 +1,8 @@
+export const RECORD = "ENS_RECORD";
+export function record(address, name) {
+  return {
+    type: RECORD,
+    address,
+    name //may be null
+  };
+}
```
gnidan, line 8: Does this need to be renamed? The asymmetric `null`-ness here suggests that we're specifically dealing with primary name records, so I think this action should reflect that.

packages/debugger/lib/session/actions/index.js
```diff
@@ -1,9 +1,10 @@
 export const START = "SESSION_START";
-export function start(provider, txHash) {
+export function start(provider, txHash, ensRegistryAddress) {
   return {
     type: START,
     provider,
-    txHash //OPTIONAL
+    txHash, //OPTIONAL
+    ensRegistryAddress //OPTIONAL
```
gnidan, line 7: I think we're better-off switching this to an `ens` namespace also, following that change I suggest for `forTx` and `forProject`.

Sure, this is an internal concern, but seems like you might want to be able to add properties here just the same as you would anywhere.

packages/debugger/lib/web3/actions/index.js
```diff
@@ -1,8 +1,16 @@
 export const INIT_WEB3 = "INIT_WEB3";
-export function init(provider) {
+export function init(provider, ensRegistryAddress) {
```
gnidan, line 2: I see there's a bunch of these `ensRegistryAddress`. You might be a bit concerned by my suggesting that we'll have other `ens` settings in the future... happy to get on a call about it, but I'll just suggest that it's pretty common these days for applications to display mainnet ENS names even when connected to other networks. We'll undoubtedly want to follow best practices here.

packages/debugger/lib/web3/adapter.js
```diff
@@ -3,11 +3,29 @@ const debug = debugModule("debugger:web3:adapter");
...
 
 export default class Web3Adapter {
-  constructor(provider) {
+  constructor(provider, ensRegistryAddress) {
     this.web3 = new Web3(provider);
+    this.ensRegistryAddress = ensRegistryAddress; //note: may be null to turn off resolution
+  }
+
+  async init() {
+    debug("initting; registry = %o", this.ensRegistryAddress);
```
gnidan, line 16: i have never seen this spelled that way

packages/debugger/lib/web3/adapter.js
```diff
@@ -104,4 +122,27 @@ export default class Web3Adapter {
     let code = await this.web3.eth.getCode(address, block);
     return code === "0x0" ? "0x" : code;
   }
+
+  async reverseEnsResolve(address) {
+    if (!this.ens) {
+      debug("no ens you fool!");
```
gnidan, line 128: i have a unique appreciation for the kind of software that insults those who diagnose its internal behavior

packages/decoder/lib/decoders.ts
```diff
@@ -89,7 +92,11 @@ export class ProjectDecoder {
     }
     this.providerAdapter = new ProviderAdapter(provider);
     this.compilations = compilations;
-    this.ensSettings = ensSettings || {};
+    this.ensSettings = {
+      provider:
+        ensSettings?.provider !== undefined ? ensSettings?.provider : provider, //importantly, null does not trigger this case
```
gnidan, line 97: Oh you already knew about this!

packages/decoder/package.json
```diff
@@ -21,10 +21,11 @@
     "build": "tsc",
     "prepare": "yarn build",
     "start": "tsc --watch",
-    "test": "mocha ./test/current/test/* ./test/legacy/test/*"
+    "test": "mocha --exit ./test/current/test/* ./test/legacy/test/* ./test/ens/test/*"
   },
   "types": "dist/index.d.ts",
   "dependencies": {
+    "@ensdomains/ensjs": "^2.1.0",
```
gnidan, line 28: We must figure out what impact this has on total package install size!

gnidan (changes requested):
Generally looks good!

Just a few naming requests, and a note that we can't merge this until we know what impact this has on the total install size for @truffle/decoder.

packages/codec/lib/format/values.ts
haltman-at: Yeah I mean where else would it go?  I'm not looking to break things too hard here.

packages/codec/lib/format/index.ts
haltman-at: Ha, I could!

packages/codec/lib/format/index.ts
haltman-at: Where else would it go?  This sort of overview seems like the most sensible place to me...

packages/codec/lib/format/utils/inspect.ts
haltman-at: Ooh, good point.

packages/decoder/package.json
haltman-at: Yeah this is already a transitive dependency via `@truffle/encoder` (I just checked to be certain :P ), so there should be no problem here.

packages/codec/lib/types.ts
haltman-at: Sure, I could make that change.

packages/core/lib/debug/cli.js
haltman-at: Oh, huh -- yeah I hadn't considered that.  Good idea to check that, will have to take a look at that.

packages/debugger/lib/debugger.js
haltman-at: Sure, that's entirely doable.

packages/debugger/lib/ens/actions/index.js
haltman-at: I mean it doesn't *need* to be renamed, but it could be...

packages/debugger/lib/session/actions/index.js
haltman-at: Doing that in an action seems a little awkward, but I guess so?

packages/debugger/lib/web3/actions/index.js
haltman-at: I mean I guess for consistency we can also do that here?

packages/core/lib/debug/cli.js
haltman-at: Oh dear, on taking a look at this, I ended up having to write up #5997!

haltman-at: OK, I've merged in #6005 and updated this PR appropriately!

packages/codec/lib/format/index.ts
gnidan: It just seems like a detail that should come later. But I don't feel strongly enough to keep this comment unresolved.

packages/decoder/package.json
gnidan: Sweet!

packages/codec/lib/format/values.ts
gnidan: Yeah, ok, that's reasonable.

packages/codec/lib/types.ts
gnidan: Thanks!

packages/core/lib/debug/cli.js
gnidan: Awesome!

packages/debugger/lib/data/sagas/index.js
gnidan: If we ever need to do forward resolution here then you'll thank me ;)

gnidan (approved):
Looks good @haltman-at!

PR #5898, opened 2023-02-15
Add support for PUSH0 instruction to debugger & disassembler

haltman-at:
This PR adds support for the upcoming PUSH0 instruction to Truffle's debugger and disassembler.  I'm doing this now because I'm going to be gone in March and this sort of thing usually falls to me, so, why not get it done before I leave.

This PR has 3 commits.  The first commit adds support for the opcode.  Note I grouped it with the other PUSHes rather than with the rest of the 0x50's.

The second commit makes it so that, when printing out PUSH0 in the debugger or disassembler, we just print out `PUSH0`, not `PUSH0 0x`, because the latter would look ugly.

The third commit is purely a code style matter -- I rewrote some calculations that previously used 0x60 (`PUSH1`) as a base to now use 0x5f (`PUSH0`) as a base.

NOTE: I didn't really test this (beyond existing CI) because Shanghai's not out yet...

### DIFF
--- a/packages/code-utils/src/opcodes.ts
+++ b/packages/code-utils/src/opcodes.ts
@@ -75,6 +75,7 @@ const codes = {
   0x5b: "JUMPDEST",
 
   // 0x60 & 0x70 range - pushes
+  0x5f: "PUSH0",
   0x60: "PUSH1",
   0x61: "PUSH2",
   0x62: "PUSH3",
--- a/packages/codec/lib/abi-data/allocate/index.ts
+++ b/packages/codec/lib/abi-data/allocate/index.ts
@@ -1014,9 +1014,7 @@ function constructorOutputAllocation(
   if (contractKind === "library") {
     //note: I am relying on this being present!
     //(also this part is a bit HACKy)
-    const pushAddressInstruction = (0x60 + Evm.Utils.ADDRESS_SIZE - 1).toString(
-      16
-    ); //"73"
+    const pushAddressInstruction = (0x5f + Evm.Utils.ADDRESS_SIZE).toString(16); //"73"
     const delegateCallGuardString =
       "0x" + pushAddressInstruction + "..".repeat(Evm.Utils.ADDRESS_SIZE);
     if (binary.startsWith(delegateCallGuardString)) {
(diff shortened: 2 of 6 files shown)


packages/code-utils/src/opcodes.ts
```diff
@@ -75,6 +75,7 @@ const codes = {
   0x5b: "JUMPDEST",
 
   // 0x60 & 0x70 range - pushes
+  0x5f: "PUSH0",
```
cds-amal, line 78: The comment above is confusing.

packages/codec/lib/abi-data/allocate/index.ts
```diff
@@ -1014,9 +1014,7 @@ function constructorOutputAllocation(
   if (contractKind === "library") {
     //note: I am relying on this being present!
     //(also this part is a bit HACKy)
-    const pushAddressInstruction = (0x60 + Evm.Utils.ADDRESS_SIZE - 1).toString(
-      16
-    ); //"73"
+    const pushAddressInstruction = (0x5f + Evm.Utils.ADDRESS_SIZE).toString(16); //"73"
```
cds-amal, line 1017: Note, this calculation is used in three places across two modules, adding error-prone maintenance complexity.

cds-amal (approved):
I left a nit and an observation, but nothing to prevent me from hitting approval.

packages/code-utils/src/opcodes.ts
haltman-at: Yeah, didn't really know what to change it to though... (thoughts?)

packages/codec/lib/abi-data/allocate/index.ts
haltman-at: There's not really any maintenance complexity here?  Like there's nothing here that needs to be updated.  I just changed the base of the calculation from `PUSH1` to `PUSH0` now that the latter exists... what would change it further in the future?

Note of course I could have just used a literal `"73"` but I figured that doing the calculation made the intent clearer.

haltman-at:
Just going to go ahead and merge this to make sure it makes it into the release... @cds-amal, if you want to change the comments to be more accurate, well, uh, you can do that? :P

packages/code-utils/src/opcodes.ts
cds-amal: `Beginning of push op codes` maybe?

packages/codec/lib/abi-data/allocate/index.ts
cds-amal: I meant that if a new opcode is updated to be base, future maintainers would have three places to update.

Issue #5881, opened 2023-02-10
"Custom error (could not decode)" for tx rejection


wbt:
## Issue
When a user uses a Truffle contract object to trigger a transaction in a browser (tested using MetaMask), and the user rejects the transaction, the error coming back has members
```
message: "MetaMask Tx Signature: User denied transaction signature. -- Reason given: Custom error (could not decode)."
reason: "Custom error (could not decode)"
```
likely due to [this line](https://github.com/trufflesuite/truffle/blob/19b421c987592d75772883849950e46402386ed3/packages/contract/lib/reason.js#L86).  

It also has `code: 4001` which could be used to figure out that this is a user rejecting the transaction.

That seems like a common enough issue that it should probably be handled well. It might also lead to different decisions about what to do in the UI. 

## Steps to Reproduce
Use a Truffle contract method to trigger a transaction in a browser with MetaMask, and reject the transaction.
Have code that listens for and displays errors, in the console and/or on-screen.

## Expected Behavior
A notice that the user rejected the transaction. 

## Actual Results
A notice that the issue was a custom error Truffle couldn't decode. 
The numeric code is present and can be caught/decoded in the application code, so this is not a major blocker, but it is a common enough use case that "could not decode" suggests Truffle immaturity as a tool for smart contract interactions. 

## Environment

* Operating System: Windows 10
* Ethereum client: Doesn't seem relevant
* Truffle version: @truffle/contract@4.6.12

haltman-at:
Hm, I'm confused why there was data attached if the transaction was rejected by the user.  I would expect that to hit either line 58 or line 61, so you wouldn't get this message.  In order to hit line 86 (which is obviously what it's hitting), there has to be data attached.  What is this data and why...?

wbt:
I'm seeing a mix.  For example, in a recent repro, after a tx was rejected the following was shown as originating from Metamask's `inpage.js:1` (i.e. before Truffle tries to decode it): 
```
{
code: 4001
message: "MetaMask Tx Signature: User denied transaction signature."
}
```
Approximately 0.2 seconds later, the error was output to console by application code after Truffle's decoding attempt, with a few more keys (modifications made to `stack` value only):
```
{
code: 4001
[...]

PR #5903, opened 2023-02-16
Fix (most) incorrect reports of custom errors

haltman-at:
Partially addresses #5881.  This is not a total fix, but it fixes the bulk of the problem.

Let's suppose you send a transaction, and it fails due to OOG, or, say, never even happens in the first place (like, intrinsic gas too low, or refused by the user in MetaMask).  In this case, the eth_call to determine the revert reason can *succeed* and get a return value instead of the intended error.  Since we can't decode it as an error, we treat it as a custom error.  Oops!

This PR fixes this by checking the length; if it's 4 mod 32, we treat it as an error, if it's not, we don't.  Assuming the Solidity ABI is being used, then error responses will have length 4 mod 32 (unless they're empty), while return value responses will have length 0 mod 32.

(Well, except for when we're dealing with contract creations, in which case successful responses can have any length, but, uh, I'm just going to not worry about that...)

This is arguably not a total fix for two reasons.  Firstly, there's the case of contract creations as mentioned above; in that case we could still hit the bug if the length of the returned contract just happends to be 4 mod 32.  Secondly, what if the transaction never happens -- e.g. it was refused by the user in MetaMask -- but the subsequent `eth_call` reverts with a valid error?  In this case, we'll append a revert string where it doesn't make sense, and it will be confusing for the user.

In order to address these cases, we should probably go, like, actually checking error codes and such.  However, that would take some actual research, and I thought I should at least start by putting up this quick fix that handles the most frequent cases.  We can do a more comprehensive fix later.

## Testing instructions

I added two tests in `contract-tests` to test this PR.  You might also want to see the original issue if you want to test it manually.

### DIFF
--- a/packages/contract-tests/test/errors.js
+++ b/packages/contract-tests/test/errors.js
@@ -327,6 +327,30 @@ describe("Client appends errors (vmErrorsOnRPCResponse)", function () {
       }
     });
 
+    it("Reports that a custom error occurred when one did", async function () {
+      const example = await Example.new(1);
+      try {
+        await example.triggerCustomError();
+        assert.fail();
+      } catch (e) {
+        assert.include(e.reason, "Custom error");
+        assert.include(e.message, "Custom error");
+      }
+    });
+
+    it("Does not report a custom error when there is none", async function () {
+      const example = await Example.new(1);
+      //there was a bug where custom errors were incorrectly reported when
+      //a function that returns a value failed due to OOG or due to not occurring
+      //(e.g. refused in MetaMask).  This test is meant to check that case.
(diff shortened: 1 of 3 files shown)


cds-amal (approved):
Looks good. I tested this with a too-little-gas tx and it behaved as predicted.

haltman-at:
Note: I was thinking of going back and changing this to have the 4-mod-32 check also apply to Error strings, but the chances of that ever being relevant are so low, because it would have to (very nearly) be a correctly-formatted string otherwise... we can change that on returning to this if we really care. :P

PR #5907, opened 2023-02-16
Increase timeout on migration out of gas test

haltman-at:
I believe this test has been timing out a bunch recently, upping it.  (It's not the only one, but one at a time...)


cds-amal (approved):
Thanks for getting this.

haltman-at:
I was kind of hoping this would fail so I'd know what other tests to turn up and put those in this PR, but, oh well. :P

PR #5909, opened 2023-02-17
Update Sourcify networks for Sourcify 1.2.2

haltman-at:
Unusually, Sourcify has made two network updates in rapid succession.  So, here's a second PR to update source-fetcher correspondingly!  This update adds only one network, Oasis Sapphire Mainnet.  (I already had this in networks.ts, but of course not in sourcify.ts).

If you want to test it, you can use address 0xFBcb580DD6D64fbF7caF57FB0439502412324179.


cds-amal (approved):
LGTM

PR #5919, opened 2023-02-21
Bug fix: only edit state in txLog when we have an initial state

eggplantzzz:
Attempting to use [this utility](https://github.com/trufflesuite/truffle/blob/develop/packages/debug-utils/index.js#L882) with the debugger in light mode resulted in a crash. We tracked this down in the debugger, it crashed at [this line](https://github.com/trufflesuite/truffle/blob/develop/packages/debugger/lib/txlog/sagas/index.js#L265). Apparently in light mode, this tx log tracking is not enabled and will make `initialState` `null` which in turn will cause things to crash. This PR fixes this by only updating state when we receive a value for `initialState`.


gnidan (approved):

cds-amal (approved):
Looks good, no breaking according to some manual resetting tests.

Issue #5441, opened 2022-08-16
Truffle's default optimizer settings actually have an effect


haltman-at:
## Issue

When Truffle compiles Solidity, it merges the settings from the Truffle Config into some default settings, those default settings being:

```javascript
settings: {
  optimizer: {
    enabled: false
    runs: 200
  }
}
```

Seems harmless, right?  Well... turns out that on Solidity versions prior to 0.8.6, having `enabled` set to `false`, and having it absent entirely, had different results.  Having it set to `false`, as opposed to absent entirely, actually turned on some optimizations (though evidently not much).

This poses a problem when it comes to contract verification; see discussion on [this Blockscout issue](https://github.com/blockscout/blockscout/issues/5748).  Unfortunately I don't think we can just get rid of this default without potentially breaking things.  But we can perhaps discard it in `fetch-and-compile` specifically.  However if we do so, we'll have to update the Etherscan fetcher (and possibly the Sourcify fetcher?) per the linked discussion (like, the Etherscan fetcher will have to convert `enabled` being absent to it being `false`; the `runs: 200` default may also be necessary).

## Environment

* Truffle version (`truffle version`): 5.5.26

Issue #2768, opened 2020-01-18
Truffle fails to resolve solc remapping


pik694:
- [x] I've asked for help in the [Truffle Gitter](http://gitter.im/Consensys/truffle) before filing this issue. []()

---------------------------

## Issue

I tried to use `solc` `remapping` by adding the proper option to `truffle-config.js`, but `Resolver` fires an exception before the flow even reaches solc.

`solc remapping` has been enabled (the option is passed from config to `solc`) as a fix to #2355, and it passes unit tests, but there is no integration test that checks whether it is actually possible to compile contracts using `remappings` option.

## Steps to Reproduce

Follow README in [https://github.com/pik694/truffle-remapping-issue.git](https://github.com/pik694/truffle-remapping-issue.git). 

The exception seems to be thrown in this line [https://github.com/trufflesuite/truffle/blob/7c19713c22a512b6a7f936a6927ddb553836d56e/packages/resolver/index.js#L75](https://github.com/trufflesuite/truffle/blob/7c19713c22a512b6a7f936a6927ddb553836d56e/packages/resolver/index.js#L75).

## Expected Behavior

Contracts should compile without errors.

## Actual Results

```
Compiling your contracts...
===========================
Error: Could not find remapped/C.sol from any sources; imported from /***/truffle-remapping-issue/contracts/A.sol
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/resolver/index.js:82:1
    at /usr/local/lib/node_modules/truffle/build/webpack:/node_modules/async/internal/onlyOnce.js:12:1
    at next (/usr/local/lib/node_modules/truffle/build/webpack:/node_modules/async/whilst.js:68:1)
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/resolver/index.js:69:1
    at FS.resolve (/usr/local/lib/node_modules/truffle/build/webpack:/packages/resolver/fs.js:69:1)
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/resolver/index.js:60:1
    at next (/usr/local/lib/node_modules/truffle/build/webpack:/node_modules/async/whilst.js:66:1)
    at /usr/local/lib/node_modules/truffle/build/webpack:/packages/resolver/index.js:69:1
    at GlobalNPM.resolve (/usr/local/lib/node_modules/truffle/build/webpack:/packages/resolver/globalnpm.js:50:1)
[...]

PR #5935, opened 2023-02-25
Allow disabling Truffle's default solc settings

fainashalts:
## PR description

This PR adds a `disableDefaults` field to `config.compilers.solc`. Additionally, it adds a check to the compile functions in `compile-solidity` to refrain from using the default  optimizer settings if `disableDefaults` is true. This is the first step toward Blockscout support in our source-fetcher. See [this issue](https://github.com/trufflesuite/truffle/issues/5441) for more information on why this is necessary.

### DIFF
--- a/packages/compile-solidity-tests/test/fixture/default-box/truffle-config.js
+++ b/packages/compile-solidity-tests/test/fixture/default-box/truffle-config.js
@@ -11,8 +11,7 @@ module.exports = {
     solc: {
       settings: {
         optimizer: {
-          enabled: false, // Default: false
-          runs: 200 // Default: 200
+          enabled: false // Default: false
         }
       }
     }
--- a/packages/compile-solidity-tests/test/fixture/default-box/truffle-config.js
+++ b/packages/compile-solidity-tests/test/fixture/default-box/truffle-config.js
@@ -9,7 +9,6 @@ module.exports = {
   },
   compilers: {
     solc: {
-      version: "0.6.9",
       settings: {
         optimizer: {
           enabled: false, // Default: false
--- a/packages/compile-solidity-tests/test/fixture/default-box/truffle-config.js
+++ b/packages/compile-solidity-tests/test/fixture/default-box/truffle-config.js
@@ -9,9 +9,10 @@ module.exports = {
   },
   compilers: {
     solc: {
+      version: "0.6.9",
       settings: {
         optimizer: {
-          enabled: true, // Default: false
+          enabled: false, // Default: false
           runs: 200 // Default: 200
         }
       }
--- a/packages/compile-solidity-tests/test/settings.ts
+++ b/packages/compile-solidity-tests/test/settings.ts
@@ -17,7 +17,7 @@ before("Set up config", () => {
   Environment.detect(config);
 });
 describe("Compile with disableDefaults flag", () => {
-  it("should not add the optimizer when missing if disableDefaults is true", async () => {
+  it("should not add missing optimizer when `disableDefaults` is true", async () => {
     //compile when optimizer is set to false
     const compilationWithOptimizer = await Compile.all(config);
 
--- a/packages/compile-solidity-tests/test/settings.ts
+++ b/packages/compile-solidity-tests/test/settings.ts
@@ -42,7 +42,8 @@ describe("Compile with disableDefaults flag", () => {
     // Verify that the bytecode is different
     assert.notStrictEqual(
       bytecodeWithDefaultsDisabled,
-      bytecodeWithOptimizerSetting
+      bytecodeWithOptimizerSetting,
+      "Bytecode should be different when disableDefaults is true and optimizer is missing"
     );
   });
 
@@ -71,7 +72,8 @@ describe("Compile with disableDefaults flag", () => {
     // Verify that the bytecode is different
     assert.strictEqual(
       bytecodeWithDefaultsEnabled,
-      bytecodeWithOptimizerSetting
+      bytecodeWithOptimizerSetting,
+      "Bytecode should be the same when disableDefaults is false and optimizer is missing"
     );
   });
 });
--- a/packages/compile-solidity-tests/test/settings.ts
+++ b/packages/compile-solidity-tests/test/settings.ts
@@ -1,26 +1,28 @@
 import { assert } from "chai";
 import Config from "@truffle/config";
 import { Compile } from "@truffle/compile-solidity";
-import { describe, it } from "mocha";
+import { describe, it, before } from "mocha";
 import path from "path";
 import { Environment } from "@truffle/environment";
 
+let config;
+before("Set up config", () => {
+  // Load configuration from the fixture file
+  config = Config.detect(
+    { workingDirectory: path.join(__dirname, "fixture/default-box") },
+    "./truffle-config.js"
+  );
+
+  // Environment.detect() so we get all the other config goodies
+  Environment.detect(config);
+});
 describe("Compile with disableDefaults flag", () => {
   it("should not add the optimizer when missing if disableDefaults is true", async () => {
-    // Load configuration from the fixture file
-    const config = Config.detect(
-      { workingDirectory: path.join(__dirname, "fixture/default-box") },
-      "./truffle-config.js"
-    );
-
-    // Environment.detect() so we get all the other config goodies
-    Environment.detect(config);
-
     //compile when optimizer is set to false
     const compilationWithOptimizer = await Compile.all(config);
 
     //get the bytecode for the `optimizer: false` case
-    const bytecodeOptimizer =
+    const bytecodeWithOptimizerSetting =
       compilationWithOptimizer.compilations[0].contracts[0].bytecode.bytes;
 
     //remove the optimizer completely from the settings
@@ -34,10 +36,42 @@ describe("Compile with disableDefaults flag", () => {
     const compilationWithoutOptimizer = await Compile.all(config);
 
     // Get the bytecode for the compiled contract
-    const bytecodeNoOptimizer =
+    const bytecodeWithDefaultsDisabled =
       compilationWithoutOptimizer.compilations[0].contracts[0].bytecode.bytes;
 
     // Verify that the bytecode is different
-    assert.notStrictEqual(bytecodeNoOptimizer, bytecodeOptimizer);
+    assert.notStrictEqual(
+      bytecodeWithDefaultsDisabled,
+      bytecodeWithOptimizerSetting
+    );
+  });
+
+  it("should add the optimizer when missing if disableDefaults is false", async () => {
+    //compile when optimizer is set to false
+    const compilationWithOptimizer = await Compile.all(config);
+
+    //get the bytecode for the `optimizer: false` case
+    const bytecodeWithOptimizerSetting =
+      compilationWithOptimizer.compilations[0].contracts[0].bytecode.bytes;
+
+    //remove the optimizer completely from the settings
+    config.compilers.solc.settings = {};
+
+    // Set disableDefaults flag to true in the configuration
+    // now Truffle should compile without any optimizer settings
+    config.compilers.solc.disableDefaults = false;
+
+    // Compile the contract
+    const compilationWithoutOptimizer = await Compile.all(config);
+
+    // Get the bytecode for the compiled contract
+    const bytecodeWithDefaultsEnabled =
+      compilationWithoutOptimizer.compilations[0].contracts[0].bytecode.bytes;
+
+    // Verify that the bytecode is different
+    assert.strictEqual(
+      bytecodeWithDefaultsEnabled,
+      bytecodeWithOptimizerSetting
+    );
   });
 });
--- a/packages/compile-solidity-tests/test/settings.ts
+++ b/packages/compile-solidity-tests/test/settings.ts
@@ -1,0 +1,43 @@
+import { assert } from "chai";
+import Config from "@truffle/config";
+import { Compile } from "@truffle/compile-solidity";
+import { describe, it } from "mocha";
+import path from "path";
+import { Environment } from "@truffle/environment";
+
+describe("Compile with disableDefaults flag", () => {
+  it("should not add the optimizer when missing if disableDefaults is true", async () => {
+    // Load configuration from the fixture file
+    const config = Config.detect(
+      { workingDirectory: path.join(__dirname, "fixture/default-box") },
+      "./truffle-config.js"
+    );
+
+    // Environment.detect() so we get all the other config goodies
+    Environment.detect(config);
+
+    //compile when optimizer is set to false
+    const compilationWithOptimizer = await Compile.all(config);
+
+    //get the bytecode for the `optimizer: false` case
+    const bytecodeOptimizer =
+      compilationWithOptimizer.compilations[0].contracts[0].bytecode.bytes;
+
+    //remove the optimizer completely from the settings
+    config.compilers.solc.settings = {};
+
+    // Set disableDefaults flag to true in the configuration
+    // now Truffle should compile without any optimizer settings
+    config.compilers.solc.disableDefaults = true;
+
+    // Compile the contract
+    const compilationWithoutOptimizer = await Compile.all(config);
+
+    // Get the bytecode for the compiled contract
+    const bytecodeNoOptimizer =
+      compilationWithoutOptimizer.compilations[0].contracts[0].bytecode.bytes;
+
+    // Verify that the bytecode is different
+    assert.notStrictEqual(bytecodeNoOptimizer, bytecodeOptimizer);
+  });
+});
--- a/packages/compile-solidity/src/index.ts
+++ b/packages/compile-solidity/src/index.ts
@@ -59,8 +59,16 @@ export const Compile = {
   // NOTE: this function does *not* transform the source path prefix to
   // "project:/" before passing to the compiler!
   async sources({ sources, options }: SourcesArgs) {
+    //this inserts optimizer settings
+    // TODO: do we need a deep copy here?
+    let originalSettings = options.compilers.solc.settings;
     options = Config.default().merge(options);
     options = normalizeOptions(options);
+    // if disableDefaults is set, we don't want to insert
+    // the default optimizer settings
+    if (options.compilers.solc.disableDefaults) {
+      options.compilers.solc.settings = originalSettings;
+    }
     //note: "solidity" here includes JSON as well!
     const [yulNames, solidityNames] = partition(Object.keys(sources), name =>
       name.endsWith(".yul")
--- a/packages/compile-solidity/src/index.ts
+++ b/packages/compile-solidity/src/index.ts
@@ -59,14 +59,9 @@ export const Compile = {
   // NOTE: this function does *not* transform the source path prefix to
   // "project:/" before passing to the compiler!
   async sources({ sources, options }: SourcesArgs) {
-    let originalSettings = options.compilers.solc.settings;
     options = Config.default().merge(options);
     options = normalizeOptions(options);
-    // if disableDefaults is set, we don't want to insert
-    // the default optimizer settings
-    if (options.compilers.solc.disableDefaults) {
-      options.compilers.solc.settings = originalSettings;
-    }
+
     //note: "solidity" here includes JSON as well!
     const [yulNames, solidityNames] = partition(Object.keys(sources), name =>
       name.endsWith(".yul")
@@ -146,14 +141,8 @@ export const Compile = {
       "resolver"
     ]);
 
-    let originalSettings = options.compilers.solc.settings;
     options = Config.default().merge(options);
     options = normalizeOptions(options);
-    // if disableDefaults is set, we don't want to insert
-    // the default optimizer settings
-    if (options.compilers.solc.disableDefaults) {
-      options.compilers.solc.settings = originalSettings;
-    }
 
     const supplier = new CompilerSupplier({
       events: options.events,
--- a/packages/compile-solidity/src/index.ts
+++ b/packages/compile-solidity/src/index.ts
@@ -59,8 +59,6 @@ export const Compile = {
   // NOTE: this function does *not* transform the source path prefix to
   // "project:/" before passing to the compiler!
   async sources({ sources, options }: SourcesArgs) {
-    //this inserts optimizer settings
-    // TODO: do we need a deep copy here?
     let originalSettings = options.compilers.solc.settings;
     options = Config.default().merge(options);
     options = normalizeOptions(options);
--- a/packages/compile-solidity/src/index.ts
+++ b/packages/compile-solidity/src/index.ts
@@ -148,8 +148,14 @@ export const Compile = {
       "resolver"
     ]);
 
+    let originalSettings = options.compilers.solc.settings;
     options = Config.default().merge(options);
     options = normalizeOptions(options);
+    // if disableDefaults is set, we don't want to insert
+    // the default optimizer settings
+    if (options.compilers.solc.disableDefaults) {
+      options.compilers.solc.settings = originalSettings;
+    }
 
     const supplier = new CompilerSupplier({
       events: options.events,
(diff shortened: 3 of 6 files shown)


packages/compile-solidity-tests/test/settings.ts
```diff
@@ -0,0 +1,43 @@
...
+    );
+
+    // Environment.detect() so we get all the other config goodies
+    Environment.detect(config);
+
+    //compile when optimizer is set to false
+    const compilationWithOptimizer = await Compile.all(config);
+
+    //get the bytecode for the `optimizer: false` case
+    const bytecodeOptimizer =
```
gnidan, line 23: Naming this something like `bytecodeUsingDefaults` might make the flow of the test here more clear

packages/compile-solidity-tests/test/settings.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    // Set disableDefaults flag to true in the configuration
+    // now Truffle should compile without any optimizer settings
+    config.compilers.solc.disableDefaults = true;
+
+    // Compile the contract
+    const compilationWithoutOptimizer = await Compile.all(config);
+
+    // Get the bytecode for the compiled contract
+    const bytecodeNoOptimizer =
```
gnidan, line 37: (and similarly this would be `bytecodeWithDefaultsDisabled` or something similar)

packages/compile-solidity/src/index.ts
```diff
@@ -59,8 +59,14 @@ export const Compile = {
...
   // "project:/" before passing to the compiler!
   async sources({ sources, options }: SourcesArgs) {
+    let originalSettings = options.compilers.solc.settings;
     options = Config.default().merge(options);
     options = normalizeOptions(options);
+    // if disableDefaults is set, we don't want to insert
+    // the default optimizer settings
+    if (options.compilers.solc.disableDefaults) {
+      options.compilers.solc.settings = originalSettings;
+    }
```
gnidan, line 69: Do we want to extract these lines into a separate function? Or can we borrow `normalizeOptions` for this and move the defaulting there?

(Curious what everyone thinks of this)

gnidan (commented):
Couple quick comments

packages/compile-solidity/src/index.ts
```diff
@@ -140,8 +146,14 @@ export const Compile = {
...
     ]);
 
+    let originalSettings = options.compilers.solc.settings;
     options = Config.default().merge(options);
     options = normalizeOptions(options);
+    // if disableDefaults is set, we don't want to insert
+    // the default optimizer settings
+    if (options.compilers.solc.disableDefaults) {
+      options.compilers.solc.settings = originalSettings;
+    }
```
cds-amal, line 156: This wouldn't be necessary if this logic is refactored to `normalizeOptions`

cds-amal (commented):
Some small nits.

packages/compile-solidity-tests/test/fixture/default-box/truffle-config.js
```diff
@@ -9,9 +9,10 @@ module.exports = {
   },
   compilers: {
     solc: {
+      version: "0.6.9",
```
cds-amal, line 12: Does this have implications for other tests?

gnidan: Probably best to update to latest whenever we can, also

packages/compile-solidity/src/index.ts
fainashalts: @gnidan when we talked yesterday I thought you said you'd rather not add this logic to `normalizeOptions` but perhaps I remembered wrong. I certainly can do that!

packages/compile-solidity-tests/test/fixture/default-box/truffle-config.js
fainashalts: Good catch y'all. This actually doesn't need to be there at all. I thought Harry said we needed a compiler version below 0.8.6 to see this issue but reading his notes, we just need a Solidity version below that. For reference though, only the profiler test is using this fixture so solc version wouldn't have mattered. Removing!

packages/compile-solidity-tests/test/settings.ts
fainashalts: Omg thank you for saving me from my "unable to properly name things" weakness. 😄 I'll update these.

fainashalts: Although, "using defaults" seems too broad? We really are only testing the optimizer. I went with `bytecodeUsingDefaultOptimizer` and `bytecodeWithoutDefaultOptimizer` but if you feel strongly I'll change it to what you originally suggested!

packages/compile-solidity/src/normalizeOptions.ts
```diff
@@ -10,10 +10,16 @@ export const normalizeOptions = (options: Config) => {
...
+
+  if (options.compilers.solc.disableDefaults) {
+    options.compilers.solc.settings.optimizer =
+      options.compilers.solc.settings.optimizer || {};
+  } else {
+    options.compilers.solc.settings.optimizer =
+      options.compilers.solc.settings.optimizer ||
+      options.compilers.solc.optimizer ||
+      {};
+  }
```
cds-amal, line 22: This requires a comment and possibly a refactor to help the reader track those object chains in this stanza. Maybe something of this form.

```suggestion

  // Helpful comment...
  const solcOptimizerDefaults = options.compilers.solc.disableDefaults
    ? {}
    : options.compilers.solc.optimizer || {};
  options.compilers.solc.settings.optimizer = options.compilers.solc.settings.optimizer ?? solcOptimizerDefaults;
```

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -113,8 +113,9 @@ module.exports = {
       //    enabled: false,
       //    runs: 200
       //  },
-      //  evmVersion: "byzantium"
-      // }
+      //  evmVersion: "byzantium",
+      // },
+      // disableDefaults: false, // Disable the default solc optimizations
```
cds-amal, line 118: This comment should explain the nuance of this setting, or why a user may need it.

packages/compile-solidity/src/index.ts
gnidan: Ah, I just meant that we should check to make sure that every use of `normalizeOptions` is preceded by that `Config.default()...`) business. If so, then there's no reason to do that first statement separately.

packages/compile-solidity-tests/test/settings.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    // Get the bytecode for the compiled contract
+    const bytecodeWithDefaultsDisabled =
+      compilationWithoutOptimizer.compilations[0].contracts[0].bytecode.bytes;
+
+    // Verify that the bytecode is different
+    assert.notStrictEqual(
+      bytecodeWithDefaultsDisabled,
+      bytecodeWithOptimizerSetting
+    );
```
cds-amal, line 46: Can you add a message to this assert for when it fails? The extra context would help tracking down the issue in CI.

packages/compile-solidity-tests/test/settings.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    // Get the bytecode for the compiled contract
+    const bytecodeWithDefaultsEnabled =
+      compilationWithoutOptimizer.compilations[0].contracts[0].bytecode.bytes;
+
+    // Verify that the bytecode is different
+    assert.strictEqual(
+      bytecodeWithDefaultsEnabled,
+      bytecodeWithOptimizerSetting
+    );
```
cds-amal, line 75: Ditto on using a message in this assert.

packages/compile-solidity/src/normalizeOptions.ts
fainashalts: I actually think the above is significantly less readable. I am adding a comment though, thank you for the suggestion!

packages/compile-solidity/src/normalizeOptions.ts
```diff
@@ -10,10 +10,17 @@ export const normalizeOptions = (options: Config) => {
...
+  if (options.compilers.solc.disableDefaults) {
+    // if the user has disabled defaults, we don't want to set optimizer defaults; Solidity versions prior to 0.8.6 will produce different bytecodes if optimizer settings are missing vs. set to false; this matters for how Blockscout does verification
+    options.compilers.solc.settings.optimizer =
+      options.compilers.solc.settings.optimizer || {};
+  } else {
+    options.compilers.solc.settings.optimizer =
+      options.compilers.solc.settings.optimizer ||
+      options.compilers.solc.optimizer ||
+      {};
+  }
```
cds-amal, line 23: The grandfathered check below also sets the optimizer without the disableDefaults guard. Is this the desired behavior?

fainashalts: I considered this and decided that this is a very specific edge case, using a previous way we configured compilers and shouldn't really be relevant here. If you or @gnidan think otherwise, let me know how you'd like it handled.

gnidan: Seems worth having the old-style config respect the new `disableDefaults` flag, since we don't know how prevalent this old-style is, and since it's not really that much of a lift to make it respect this new flag.

packages/core/lib/commands/init/initSource/truffle-config.js
```diff
@@ -113,8 +113,12 @@ module.exports = {
       //    enabled: false,
       //    runs: 200
       //  },
-      //  evmVersion: "byzantium"
-      // }
+      //  evmVersion: "byzantium",
+      // },
+      // disableDefaults: false, // Disable the default solc optimizer settings; if true,
+      //                         // whatever setting is passed in settings.optimizer will be respected;
+      //                         // if false, the default when settings.optimizer is missing is
```
gnidan, line 120: This doesn't _only_ apply to the optimizer settings, right? We're disabling defaults for all solc settings?

fainashalts: The issue was specifically the optimizer. I actually was torn on this and looked at my notes from our call with Harry. Specifically optimizer being empty was being replaced with { enabled: false } and this was a problem pre-Solidity 0.8.6. I think it would be a mistake to disable all default settings if this flag is set. Like, I'm not sure what the evmVersion stuff is about and didn't think we should mess it up. 

Perhaps I should rename this to `disableOptimizerDefaults` to be more clear?

gnidan: Why don't we just make it about all default solc settings? Seems silly to scope this to just that one field.

gnidan: (I know that right now, the optimizer defaults are what's causing problems, but I could see other defaults being problematic in the future)

packages/compile-solidity/src/normalizeOptions.ts
fainashalts: Well, looking at it -- the optimizer setting _already_ doesn't get set to a default, right? It will always be `options.solc.optimizer` for the grandfathered case. Does that require a change?

packages/core/lib/commands/init/initSource/truffle-config.js
gnidan: If we want to scope this more narrowly, then we should definitely rename it to reflect this. This is not my vote, but if others are cool with having a separate `disableOptimizerDefaults` (understanding that we might also need a broader flag in the future)

fainashalts: The optimizer default is a bug and this fix is meant to address it in a non-breaking way. In general I would not expect us to provide defaults when they are not necessary to the proper running of our code and thus chose this narrow approach.

For instance, having `remappings: []` as a default likely makes a difference. It certainly affects behavior, I believe. If you set `remappings` to an empty array, you are explicitly telling the compiler to use only the default remappings and disallowing the use of any custom mappings. If you leave out the remappings setting entirely, you are still using the default remappings, but leaving open the possibility of using custom mappings in the future. I think it is tricky!

gnidan: I'm just thinking, like, if we need to do another one of these in the future, it's not like the `if` statements we're leaving in place here for `disableOptimizerDefaults` will ever go away. At best, we're looking at a field we'll want to deprecate, at worst, a permanent logical complexity for future contributors.

vs. `disableDefaults`, which just turns off all the defaults Truffle provides... a simple notion!

fainashalts: Sure, but disabling all default settings could be a breaking change in some circumstances. Well, _would_ be for someone in the scenario I wrote out above.

gnidan: Yeah, but it's inherently "breaking" if you tell Truffle you don't want the values it's been giving you. All it means is that we'll need to add `remappings: []` ourselves inside the Blockscout fetcher.

fainashalts: We also should not need to do this again. It is a hack, and I'm not sure we should code toward needing similar hacks in the future... We should not add defaults that we then need to disable, I think.

gnidan: I see your argument though that `disableDefaults` introduces this caveat-emptor can of worms. But I don't think it's a big deal because people looking to disable defaults will almost certainly have their fully-populated settings.

fainashalts: Well, this isn't just about that though. The main reason we are doing this is for _already deployed_ contracts we are fetching from Blockscout. We don't know what the preferred configuration was in that setting and we could be introducing a new bug when trying to address the optimizer bug.

fainashalts: Making the setting user-facing was a secondary decision to the main purpose of this code change. In fact, I'd probably not make it user-facing, since defaults only happen when you don't set it yourself. Not sure why someone would want optimizer to be `{}` instead of `enabled: false` and it just sort of pollutes the config. 🤷‍♀️

gnidan: There's been this long outstanding issue with Truffle that we don't let users just wholesale tell us what solc settings to use. This optimizer thing is only the latest of these, so that's why I thought we were talking about the problem generally, since that's a similar diff and solves a categorical concern. Solving this problem one at a time just adds technical debt in the form of `if` statements and the need to maintain these nuanced configuration params.

Anyway, hopefully my argument on the matter is clear. I'll abstain further to let others comment if they have thoughts.

fainashalts: I'd be happy to hear other perspectives of course, but just wanted to address the previous Truffle issues linked here because I do think they actually support my points! 😄 

[#2355](https://github.com/trufflesuite/truffle/issues/2355) actually led to [this PR](https://github.com/trufflesuite/truffle/pull/2358) to _add_ the default `remappings` setting to `configDefaults` and pass it along the flow for Truffle's compilation. But wait! [This issue](https://github.com/trufflesuite/truffle/issues/2768) was created the following year, pointing out that Truffle's resolver is possibly _unable_ to resolve remappings passed through the settings! 🤯 Thus, _the default setting of `remappings: []` may be necessary_. AFAIK, nobody ever fixed the remappings issue with the resolver. And I don't think that (priority 4) issue should block this PR, which is attempting to fix a very specific defaults bug. 🤷‍♀️ 

As for the other two issues shared, the cause was never completely sorted out for the first, and the third is what this PR is attempting to address. 🎉 

I hope that is some helpful history on this rather more complicated than expected default settings issue.

fainashalts: Oh also, happy with the option of just disabling the defaults altogether, but that would be a breaking change and we need this fix short-term so we can enable Blockscout support. 🤷‍♀️

packages/compile-solidity/src/normalizeOptions.ts
gnidan: Hm you might be right! Guess the workaround this whole time was to use the old-style solc config!

cds-amal: Are old-style solc config options a viable work-around for blockscout?

gnidan: Could be wrong but it does seem that way. Not that I think we should lean on that since this is already so far along

fainashalts: One _could_ probably use this as a workaround. Just set `options.solc.optimizer` to `options.solc.settings.optimizer` when fetching from Blockscout? Wouldn't require this PR at all in that case, right. But I think that is a less straightforward hack than the one proposed here. Not to mention that adding default optimizer settings is itself incorrect, so this PR would allow us to rectify it for Blockscout contracts without breaking anything.

cds-amal: is there a way to have a `normalizeForBlockscout` function that is only invoked for blockscout needed resolution?

fainashalts: I'd be fine with that! When I asked @gnidan he wanted this to be a user-facing setting, though I can see now that may have been due to our misunderstanding around how broad/narrow it would be. If he's ok with it, I'd be happy to refactor in this way.

gnidan: > is there a way to have a `normalizeForBlockscout` function that is only invoked for blockscout needed resolution?

Ugh please no. Everyone's always talking about the amount of tech. debt we have these days. We must be mindful of these sorts of hundred-line PRs... they should not introduce new debt!

haltman-at:
Hey, apologies for jumping in so late with a large structural comment like this, but I said I'd check back in occasionally while on vacation and uh I'm getting to this now!

But, well, I have to say that the structure here seems a bit off.  I don't understand why `disableDefaults` is being added to the Truffle Config; that doesn't seem like an appropriate place to me.  Rather I would have expected this to be something like one of the `internalOptions` that's passed to `run` (that obviously doesn't work per se in this case, but someting like that).  I guess really what I'm saying is that, wouldn't it make more sense to modify `sources`, rather than adding anything to the Truffle Config?

Like right now, when fetch-and-compile goes to compile things, it calls `sources`.  You'll notice that `sources` sets the `noTransform` internal option, and the reason for that is that we don't want to apply our path transformations when doing fetch-and-compile.  Basically, `sources` is specialized for fetch-and-compile at the moment.  So I'm just thinking, rather than doing anything to the Truffle config, wouldn't it make more sense to just modify `sources` so that it doesn't apply the defaults?  (Since I believe that's where it happens, rather than in `run`.)  And then the other functions could be left alone.

I don't know, maybe you all discussed this all on Slack, I haven't checked that, I'm on vacation!  But I thought I should at least quickly weigh in.

(Don't let this hold things up if it's a big problem!  Obviously I can't go consistently responding to things or doing approvals while away, so just take this as a bit of outside opinion that you may just end up ignoring. :P )

fainashalts:
> Hey, apologies for jumping in so late with a large structural comment like this, but I said I'd check back in occasionally while on vacation and uh I'm getting to this now!
> 
> But, well, I have to say that the structure here seems a bit off. I don't understand why `disableDefaults` is being added to the Truffle Config; that doesn't seem like an appropriate place to me. Rather I would have expected this to be something like one of the `internalOptions` that's passed to `run` (that obviously doesn't work per se in this case, but someting like that). I guess really what I'm saying is that, wouldn't it make more sense to modify `sources`, rather than adding anything to the Truffle Config?
> 
> Like right now, when fetch-and-compile goes to compile things, it calls `sources`. You'll notice that `sources` sets the `noTransform` internal option, and the reason for that is that we don't want to apply our path transformations when doing fetch-and-compile. Basically, `sources` is specialized for fetch-and-compile at the moment. So I'm just thinking, rather than doing anything to the Truffle config, wouldn't it make more sense to just modify `sources` so that it doesn't apply the defaults? (Since I believe that's where it happens, rather than in `run`.) And then the other functions could be left alone.
> 
> I don't know, maybe you all discussed this all on Slack, I haven't checked that, I'm on vacation! But I thought I should at least quickly weigh in.
> 
> (Don't let this hold things up if it's a big problem! Obviously I can't go consistently responding to things or doing approvals while away, so just take this as a bit of outside opinion that you may just end up ignoring. :P )

Thanks for the input Harry! I think this actually would make a lot more sense, and I had only put this in TruffleConfig because @gnidan wanted disabling defaults to be a user-facing setting. My initial thought was to keep it internal as well.

But also, in testing this I realized I can't make Truffle actually put in the compiler defaults at all? Like, the merge of the configDefaults with the options given doesn't seem to be a deep merge so if I set `compilers.solc.settings = {}` and compile, it doesn't get filled in (the tests here pass because I accidentally used a `before` rather than a `beforeEach` -- changing it makes the second test fail)! So what I'm saying is I'm not sure this is necessary at all? Please correct me if I'm missing something!

haltman-at:
> Thanks for the input Harry! I think this actually would make a lot more sense, and I had only put this in TruffleConfig because @gnidan wanted disabling defaults to be a user-facing setting. My initial thought was to keep it internal as well.

Oh, ha!  This is what happens when I'm off Slack for weeks, I guess!

> But also, in testing this I realized I can't make Truffle actually put in the compiler defaults at all? Like, the merge of the configDefaults with the options given doesn't seem to be a deep merge so if I set compilers.solc.settings = {} and compile, it doesn't get filled in (the tests here pass because I accidentally used a before rather than a beforeEach -- changing it makes the second test fail)! So what I'm saying is I'm not sure this is necessary at all? Please correct me if I'm missing something!

It is a deep merge, as far as I've always seen?  Like, we use `config.merge`, not `config.with`, right?  Now -- going by memory, I'm not checking the code as I say this -- `config.merge` doesn't perform a deep merge on *every* field, but it does on some *particular* fields, and `compilers` is one of those IIRC.  Also, if the defaults weren't getting set, I wouldn't have run into those problems with Blockscout earlier!  So I'm pretty sure it is in fact a deep merge for that field?  Again, I'm saying this without checking the code.

fainashalts:
> It is a deep merge, as far as I've always seen? Like, we use config.merge, not config.with, right? Now -- going by memory, I'm not checking the code as I say this -- config.merge doesn't perform a deep merge on every field, but it does on some particular fields, and compilers is one of those IIRC. Also, if the defaults weren't getting set, I wouldn't have run into those problems with Blockscout earlier! So I'm pretty sure it is in fact a deep merge for that field? Again, I'm saying this without checking the code.

So I thought the same about the deep merge, but I couldn't actually get it to merge in those settings! When I debugged the flow I watched the merge function skip over compilers.solc.settings but I may be missing something! I'm actually about to be OOO for a couple of weeks, and it looks like we'll both be back the first week of April. Let's meet when you're back and sort it out?

haltman-at:
> So I thought the same about the deep merge, but I couldn't actually get it to merge in those settings! When I debugged the flow I watched the merge function skip over compilers.solc.settings but I may be missing something! I'm actually about to be OOO for a couple of weeks, and it looks like we'll both be back the first week of April.

Which merge function, `config.merge` or `lodash.merge`?  The former never explicitly touches `compilers.solc.settings`, only `compilers`; it lets `lodash.merge` handle the latter.  Are you saying you debugged `lodash.merge`?  I'm confused.

> Let's meet when you're back and sort it out?

Sure, I guess!

haltman-at:
OK, after talking to @gnidan, we've decided to close this.  The problem is that the whole idea of making this a user-setting has more problems than we realized.

See, in fetch-and-compile, the merge-in of defaults only happens once, so it's easy to account for.  But for `truffle compile`, the defaults are merged in *three times*.  That makes things a bit more of a problem.  Those three times are:
1. The obvious place, in `compile-solidity`
2. Prior to that, in `workflow-compile`
3. Prior to *that*, it happens in `Config.detect` called from `commands/compile`.

So how do we handle this?  We could put a check on all of these, but the fact that it happens all the way back in `Config.detect` is a bit of a problem.  One could alter `config.merge`, but that raises the specter of all sorts of hairy side effects.

So, @gnidan agreed to take the route that, while this should be a user-setting ideally, it won't be right now.  For now we're going to take the route of just fixing this for fetch-and-compile, and worry about making it a user setting later.

As such, I'm closing this PR and #5957, and I'm just going to do these from scratch in a single PR that doesn't expose this as a user setting... :-/

PR #5946, opened 2023-03-02
Ensure resolver.resolve()'s type only needs 1 arg.

gnidan:
... since only the first argument is required and the others are optional, but the type incorrectly specifies that 2 arguments are required

This was raised [here](https://github.com/rkalis/truffle-plugin-verify/pull/196#discussion_r1123032692)


cds-amal (approved):
Looks good!

PR #5957, opened 2023-03-08
Support blockscout

fainashalts:
## PR description

This PR introduces blockscout support to Truffle's `source-fetcher` and `fetch-and-compile` functionality. In tandem with Ganache forking of the ConsenSys zkEVM, this support should allow for local debugging of transactions retrieved from the blockscout explorer.

## Testing instructions

To test this PR, fork ganache using the ConsenSys zkEVM testnet (this is only available on blockscout's explorer but you could also use mainnet or goerli if you use a verified contract that isn't also verified on etherscan or sourcify, I suppose). Attempt to debug a transaction you find on blockscout.

## Documentation

- [ x] I thought about documentation and added the `doc-change-required` label to this PR if documentation updates are required.

- [ x] I have added any needed `breaking-change` and `new-feature` labels for the appropriate packages.

** This PR is still a work in progress **
## To Do

- [ ] Determine whether fetching via other means (multi-file, json input) is necessary or if single file is sufficient
- [ ] Add multi-source tests if needed
- [ ] Add vyper test, to ensure `processVyperResult` works correctly
- [ ] Ensure handling of libraries is correct; ensure that `processLibraries` can return the appropriate information gleaned from blockscout's response
- [ ] Ensure constructor arguments are handled correctly
- [x] Currently defaults to 200 runs, make sure that doesn't mess with bytecode
- [x] Determine whether the optimization defaults are still an issue for contracts below Solidity 0.8.6

### DIFF
--- a/packages/fetch-and-compile/test/fetch.test.ts
+++ b/packages/fetch-and-compile/test/fetch.test.ts
@@ -95,7 +95,7 @@ describe("Supported networks", function () {
       name: "mainnet",
       networkId: 1,
       chainId: 1,
-      fetchers: ["etherscan", "sourcify"]
+      fetchers: ["etherscan", "sourcify", "blockscout"]
     });
   });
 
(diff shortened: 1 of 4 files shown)


fainashalts:
Just a note from my research: I have been unable to reproduce the optimizer settings issue brought up by Harry [here](https://github.com/trufflesuite/truffle/issues/5441). As far as I can tell, a verified contract(Solidity 0.8.5 in my tests, as I was told before 0.8.6 we see this problem) on Blockscout that left out `optimizer: { enabled: false}` can still be verified by Truffle even though we add that setting in our defaults. I have asked in the [Blockscout issue](https://github.com/blockscout/blockscout/issues/5748) originally opened by Harry but it is closed so I'm not sure if I'll get a response. It is possible the Blockscout folks remedied this on their end to ensure compatibility with the Etherscan and Sourcify's verifcation. 

I have a PR out for the optimizer defaults work, so I'll just proceed without it here for now and will update with any new info.

haltman-at:
Closing this for the reasons discussed on #5935.

Issue #5967, opened 2023-03-11
The new plugin `@truffle/dashboard-hardhat-plugin` breaks with `Error: spawn npx ENOENT`


pcaversaccio:
## Issue

I wanted to add the new Truffle plugin `@truffle/dashboard-hardhat-plugin` to my Hardhat project template [here](https://github.com/pcaversaccio/hardhat-project-template-ts). However, this will break all my scripts with the following error:

```bash
$ npx hardhat run --network truffle-dashboard scripts/deploy.ts
Error: spawn npx ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:282:19)
    at onErrorNT (node:internal/child_process:480:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

---

$ npx hardhat test
Error: spawn npx ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:282:19)
    at onErrorNT (node:internal/child_process:480:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

---

$ npx hardhat run --network hardhat scripts/deploy.ts
Error: spawn npx ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:282:19)
    at onErrorNT (node:internal/child_process:480:16)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

## Steps to Reproduce

Clone my repository [here](https://github.com/pcaversaccio/hardhat-project-template-ts/tree/main) and enable `import "@truffle/dashboard-hardhat-plugin";` in my `hardhat.config.ts`. Thereafter, run e.g. `yarn test:hh` or `yarn deploy:hh`.

## Expected Behavior

It should not break and work smoothly. To be clear, I can use smoothly the Truffle dashboard without the plugin, so that works as intended. Just importing the new plugin breaks it, unfortunately.

## Actual Results

See above in the summary.

## Environment

* Operating System: Windows
* Ethereum client:
* Truffle version (`truffle version`): `v5.8.0 (core: 5.8.0)`
* node version (`node --version`): `v16.0.0`
[...]

PR #5968, opened 2023-03-13
Enable shell option for spawn

cds-amal:
Addresses #5967. This is needed for windows to spawn npx correctly

## Documentation

### DIFF
--- a/packages/from-hardhat/src/ask-hardhat.ts
+++ b/packages/from-hardhat/src/ask-hardhat.ts
@@ -35,7 +35,8 @@ export const askHardhatVersion = async (
 
     const hardhat = spawn(`npx`, ["hardhat", "--version"], {
       stdio: ["pipe", "pipe", "inherit"],
-      cwd: workingDirectory
+      cwd: workingDirectory,
+      shell: true
     });
 
     let output = "";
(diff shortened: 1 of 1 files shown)


cds-amal:
@gnidan,

Do we want to use shell for all OSes, or, only for Win32? This change is for the former.

gnidan:
> @gnidan,
> 
> Do we want to use shell for all OSes, or, only for Win32? This change is for the former.

I think for all OSes... should be fine, in fact preferable (in case the `shell` environment uses a different path for `npx`, we should respect that and not fallback to whatever system default you get without `shell: true`)

gnidan (approved):
OK this isn't ready for review yet but have an approve anyway. Thanks for grabbing this!

cds-amal:
The shell property type is `boolean | string`. If boolean:true, spawn will run the command w/in the OS's default shell, otherwise the string specifies the shell (unsure it's a fullpath or not).

gnidan:
> The shell property type is `boolean | string`. If boolean:true, spawn will run the command w/in the OS's default shell, otherwise the string specifies the shell (unsure it's a fullpath or not).

I think we do want this as `true`... at least it seems correct right now. If users want to override this in the future, I could see that, but let's wait for people to ask before we expand the scope here. `shell: true` just seems like something we should've had here this whole time.

Anyway, LGTM!

PR #5974, opened 2023-03-17
Allow to override chai from the config

farcaller:
## PR description

Allow to override the chai instance from the config. This allows one to reduce the test boilerplate when chai plugins are used, e.g. this snippet could move to the config, instead of being repeated in every test:

```js
import chai from 'chai';
import chaiAsPromised from 'chai-as-promised';

chai.use(chaiAsPromised);

const { assert } = chai;
```

## Testing instructions

require chai from the config and point config's `chaiOverride` to it. No default value is offered as the default is undefined.

## Documentation

Trivial change, no docs required.

Non-breaking.

### DIFF
--- a/packages/test/src/Test.ts
+++ b/packages/test/src/Test.ts
@@ -320,7 +320,7 @@ export const Test = {
     global.interfaceAdapter = interfaceAdapter;
     // @ts-ignore
     global.web3 = web3;
-    const resolvedChai = config.chaiOverride ? config.chaiOverride : chai;
+    const resolvedChai = config.chai?.package ?? chai;
     // @ts-ignore
     global.assert = resolvedChai.assert;
     // @ts-ignore
--- a/packages/test/src/Test.ts
+++ b/packages/test/src/Test.ts
@@ -320,10 +320,11 @@ export const Test = {
     global.interfaceAdapter = interfaceAdapter;
     // @ts-ignore
     global.web3 = web3;
+    const resolvedChai = config.chaiOverride ? config.chaiOverride : chai;
     // @ts-ignore
-    global.assert = chai.assert;
+    global.assert = resolvedChai.assert;
     // @ts-ignore
-    global.expect = chai.expect;
+    global.expect = resolvedChai.expect;
     // @ts-ignore
     global.artifacts = {
       require: (importPath: string) => {


cds-amal (commented):
Hi @farcaller , 

Thanks for the PR. I have a question.

> Test instructions
> require chai from the config and point config's chaiOverride to it. No default value is offered as the default is undefined.

Can you explain what `chaiOverride` is, and which config it resides in?

farcaller:
my understanding is that `config` is resolved to the `truffle-config.js`, so you can set up the chai import in there.

cds-amal (commented):
> my understanding is that config is resolved to the truffle-config.js, so you can set up the chai import in there.

This PR introduces behavior based on a currently undocumented config setting. I think this PR should include the new setting, along with its a default value, so the complete feature implementation is captured in one place.

Thoughts, @eggplantzzz, @gnidan?

gnidan:
Sorry for the delay. Just getting back to this.

I think this approach generally is fine, i.e., allowing users to specify a different package name in the truffle-config.

I think we should do something like this, though, instead:
```javascript
module.exports = {
  // ... rest of truffle-config.js

  chai: {
    package: "chai-as-promised"
  }
}
```

I think this might be the sort of thing we need to add to @truffle/config itself, also, but that sounds like more of an @eggplantzzz question.

eggplantzzz:
Hey @farcaller, so if you can make the changes that @gnidan suggested we can move forward on this. The important part here is that the property in the `truffle-config.js` should be `chai.package` instead of `chaiOverride`. It seems like this should probably be enough for now. Currently there really isn't a need to have default values for this in @truffle/config, unless there is something I missed @gnidan.

Let us know if you need anything else! Also, if you feel like you won't have time to finish this, @cds-amal says he can bring it across the finish line. Just let us know :)

eggplantzzz:
So @farcaller, does the thumbs up mean you'd like to bring this across the finish line? Let us know either way. Like I said, we can take it up if you don't have the time. Thanks!

farcaller:
The thumbs up is my seeing the feedback :) I’m out in holidays this whole month and will get back to coding in May. Sorry for the delays!

eggplantzzz:
Ok great! Enjoy your vacation!

cliffoo:
Hey @farcaller are you still interested in putting the rest of the PR together? Let us know, thanks!

farcaller:
Sorry for the [extreme] delay. The PR is updated as per the feedback.

haltman-at (commented):
Huh, so as this currently is it actually has the user doing the `require` in the config rather than just supplying the package name... is that what we want?

gnidan:
> Huh, so as this currently is it actually has the user doing the `require` in the config rather than just supplying the package name... is that what we want?

This is definitely more ideal, where @truffle/test's internals should do the `require()` for the user.

But this might run into problems... if @truffle/test does `require("chai")`, it will likely use `./node_modules/@truffle/test/node_modules/chai`, when the user probably wants `./node_modules/chai`.

So if we want to avoid this precedence concern, the way to do that is to put the `require("chai")` inside `truffle-config.js`.

I think this might be what we're stuck with. Thanks for raising this concern!

gnidan (approved):
Thanks for following through on this @farcaller!

haltman-at (approved):
Well, if this new way is good, then I'll approve it too!

Issue #5958, opened 2023-03-08
Allow the Dashboard host and port to be configurable when using the Dashboard Hardhat Plugin


kevinbluer:
## Issue

When using the Truffle Dashboard Hardhat Plugin, the target [host and port are currently hardcoded](https://github.com/trufflesuite/truffle/blob/develop/packages/dashboard-hardhat-plugin/src/index.ts#L23-L26) to the defaults (`localhost` and `24012` respectively). As highlighted in [this conversation](https://github.com/trufflesuite/truffle/pull/5927#discussion_r1122179058) this should be made user configurable by allowing users to specifying them in the Hardhat config.

gnidan:
You know what I just realized... maybe there's a way for this plugin to remove the need for the user to define the dashboard network themselves. Can HH plugins define networks... is that kosher?

Ooh yeah OK I think we should probably read up on how [hardhat-foundry](https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-foundry) works and just emulate that.

gnidan:
Did some more digging on the above and chatted with @alcuadrado, who confirms the limitation I describe here.

So I really want this sort of thing:

<details>
<summary>Example <code>hardhat.config.ts</code></summary>

```typescript
import { HardhatUserConfig } from "hardhat/config";
import "@nomicfoundation/hardhat-toolbox";
import "@truffle/dashboard-hardhat-plugin";

const config: HardhatUserConfig = {
  networks: {
    "truffle-dashboard": {
      timeout: 60 // override **only** the default timeout in seconds
    },

    // ... rest of HardhatNetworksUserConfig
  },

  truffle: {
    dashboardNetworkName: "truffle-dashboard"
  },

  // ... rest of HardhatUserConfig ...
};

export default config;
```

</details>

i.e., the plugin should minimize the configuration options it adds in that `truffle` namespace. Instead, we should allow people to override the Dashboard network the same way they would override any HTTP network in their Hardhat config. This plugin should then fill in the gaps between Hardhat's defaults and the user-specified network params.

It's worth noting here that the way Hardhat does config is by separating user config from complete config with defaults. It defines types for the `HardhatUserConfig` (e.g. `HttpNetworkUserConfig`), where most fields are optional, and it has separate types for the `HardhatConfig`, where everything is filled in. Key points here:

- Plugins have the chance to read the user-config and modify the config however they want. So we can add a network just fine.
[...]

PR #5982, opened 2023-03-30
Make @truffle/dashboard-hardhat-plugin automatically manage the Dashboard Hardhat network configuration (while still allowing customization)

gnidan:
## PR description

Addresses #5958 

This PR modifies the @truffle/dashboard-hardhat-plugin to respect custom Dashboard configuration inside truffle-config.js.

At a high-level, this does a couple things:
- Automatically add the `truffleDashboard` network to the Hardhat networks configuration based on the Dashboard host/port configuration found via `TruffleConfig.detect()`
- Allow users to disable this network management if they want to configure this network themselves.
- Whichever way the network is configured, connect to the dashboard-message-bus with the configured host/port.

This adds a new `truffle` section to `hardhat.config.ts` with a few settings:
- `dashboardNetworkName` (default `"truffleDashboard"`), to specify the name of the Truffle Dashboard network
- `dashboardNetworkConfig` (default `{}`), to override network settings (such as `timeout`, etc.)

(This last field (`dashboardNetworkConfig`) is necessary because of reasons stated [here](https://github.com/trufflesuite/truffle/issues/5958#issuecomment-1474552333).)

### plugin setup

`yarn link`, for some reason, breaks the LSP with the new hardhat project. To get around this:
  1. `cd TRUFFLE_REPO/projects/dashboard-hardhat-plugin`
  1. `yarn pack`
  1. From inside Hardhat project, `yarn add full_path_to_generated_truffle-dashboard-hardhat-plugin-v0.1.1.tgz`
  
### test scenarios
~Set up this plugin via `yarn link` inside a Hardhat project and test the following cases:~
- `config.networks[dashboardNetworkName]` should throw an error if it exists
- `config.networks[dashboardNetworkName]` should connect to Dashboard with the settings obtained via truffle-config.js and via `config.truffle.dashboardNetworkConfig`

## Documentation

### DIFF
--- a/packages/dashboard-hardhat-plugin/README.md
+++ b/packages/dashboard-hardhat-plugin/README.md
@@ -1,10 +1,18 @@
 # @truffle/dashboard-hardhat-plugin
 
-Enable decoded results when using
-[Truffle Dashboard](https://trufflesuite.com/docs/truffle/how-to/use-the-truffle-dashboard/)
-with your Hardhat projects.
+> No more private keys inside `.env` files! Truffle Dashboard lets you sign
+> messages with MetaMask and provides insights for better dapp development.
+> :lock:
 
-## What
+This plugin makes it easy to access all the features of
+[Truffle Dashboard](https://trufflesuite.com/docs/truffle/how-to/use-the-truffle-dashboard/),
+including better security and high-quality transaction decoding.
+
+## Summary
+
+This plugin automatically adds a `--network truffleDashboard` to your Hardhat
+project and tells Truffle Dashboard all the information it needs about your
+smart contracts.
 
 This plugin enables you to see decoded transaction information (both the
 function signature and the values of any arguments passed) when using
@@ -23,35 +31,39 @@ transactions with your browser-based wallet.
 
 You can install this plugin with `npm` (or `yarn`) by running:
 
-```bash
-npm i @truffle/dashboard-hardhat-plugin -D
+```console
+npm install --save-dev @truffle/dashboard-hardhat-plugin
 ```
 
-Beyond that, simply import the plugin in your `hardhat.config.js`:
+Then simply import the plugin in your `hardhat.config.ts`:
 
-```ts
+```typescript
 import "@truffle/dashboard-hardhat-plugin";
 ```
 
 ## Setup
 
-This extension assumes you have `truffle` installed (either globally or in a
-local project context). If not, you can install it with `npm i -g truffle`.
-Beyond this you'll be able to start Truffle Dashboard with `truffle dashboard`.
-For reference, more information on using the Truffle Dashboard can be found
+This extension assumes you have Truffle installed (either globally or in a local
+project context). If not, you can install it with `npm i -g truffle`. Beyond
+this you'll be able to start Truffle Dashboard with `truffle dashboard`. For
+reference, more information on using the Truffle Dashboard can be found
 [here](https://trufflesuite.com/docs/truffle/how-to/use-the-truffle-dashboard/).
 
 Alternatively, you can skip the installation and fetch (and run) it remotely
 with `npx truffle dashboard`.
 
-This plugin automatically tells Hardhat about the `"truffle-dashboard"` network,
-so you don't need to modify your `hardhat.config.ts` to get started! Just
-specify `--network truffle-dashboard` when running your usual Hardhat commands
-(e.g.
-`npx hardhat run ./scripts/deploy-contracts.ts --network truffle-dashboard`).
+This plugin automatically tells Hardhat about the `truffleDashboard` network, so
+you don't need to modify your `hardhat.config.ts` to get started. Just specify
+`--network truffleDashboard` when running your usual Hardhat commands (e.g.
+`npx hardhat run ./scripts/deploy-contracts.ts --network truffleDashboard`).
 This managed network will include the `url` property (default
 `"http://localhost:24012/rpc"`) and a few other sensible defaults for this
 workflow.
+
+**Note on request timeouts**: Signing transactions with MetaMask is slower than
+Hardhat computing signatures automatically. For this reason,
+@truffle/dashboard-hardhat-plugin disables request timeouts for the
+`truffleDashboard` network. See below if you'd like to change this behavior.
 
 ### Configuration
 
@@ -62,24 +74,25 @@ workflow.
 > reference documentation to learn more about this; this section pertains only
 > to configuring @truffle/dashboard-hardhat-plugin itself.
 
-To configure this plugin, add a `truffleDashboard` section to your Hardhat
-config. This namespace can contain the following fields (all are optional):
+To configure this plugin, add a `truffle` section to your Hardhat config. This
+namespace can contain the following fields (all are optional):
 
-- **`networkName`** _(default: `"truffle-dashboard"`)_: Specify the name of the
-  Truffle Dashboard network, e.g. via the `--network <...>` command-line option.
+- **`dashboardNetworkName`** _(default: `"truffleDashboard"`)_: Specify the name
+  of the Truffle Dashboard network, e.g. via the `--network <...>` command-line
+  option.
 
-- **`networkConfig`**: A subset of the usual Hardhat network configuration
-  settings.
+- **`dashboardNetworkConfig`**: A subset of the usual Hardhat network
+  configuration settings.
 
   You can use this to override managed network behavior for any fields
   documented by Hardhat's
   [JSON-RPC based networks](https://hardhat.org/hardhat-runner/docs/config#json-rpc-based-networks)
   reference **except** `chainId`, `from`, `accounts`, and `url`. (This plugin
-  supplies `url`, and the other three fields are meaningless in the Truffle
+  supplies `url`; the other three fields are meaningless in the Truffle
   Dashboard signing workflow.)
 
-  All `networkConfig` properties are optional. The following field has a
-  different default than what Hardhat normally provides:
+  All `dashboardNetworkConfig` properties are optional. The following field has
+  a different default than what Hardhat normally provides:
 
   - **`timeout`** _(default: `0`; i.e., no timeout)_: Hardhat's normal default
     value for this is `40000` (40 seconds)
@@ -89,8 +102,8 @@ config. This namespace can contain the following fields (all are optional):
 Assuming the above is all looking good, any subsequent compilations via
 `npx hardhat compile` will send the compiled artifacts to Truffle Dashboard.
 Subsequently, any deployments (or transactions via scripts) that target the
-`"truffle-dashboard"` network will now send them to Truffle Dashboard for
-signing via your browser-based wallet.
+`truffleDashboard` network will now send them to Truffle Dashboard for signing
+via your browser-based wallet.
 
 > Note you'll need to ensure your browser-based wallet is unlocked for
 > transactions to be received (otherwise you might see a
@@ -98,7 +111,7 @@ signing via your browser-based wallet.
 > command below).
 
 ```console
-npx hardhat run scripts/deploy.ts --network truffle-dashboard
+npx hardhat run scripts/deploy.ts --network truffleDashboard
 ```
 
 ![Truffle Dashboard](./assets/truffle-dashboard-screenshot.jpg)
--- a/packages/dashboard-hardhat-plugin/README.md
+++ b/packages/dashboard-hardhat-plugin/README.md
@@ -49,34 +49,27 @@ so you don't need to anything to your `hardhat.config.ts` to get started! Just
 specify `--network truffle-dashboard` when running your usual Hardhat commands
 (e.g.
 `npx hardhat run ./scripts/deploy-contracts.ts --network truffle-dashboard`).
+This managed network will include the `url` property (default
+`"http://localhost:24012/rpc"`) and a few other sensible defaults for this
+workflow.
 
 ### Configuration
 
-This plugin's defaults to injecting `url` and other network properties into your
-project's `networks` configuration automatically.
+> ℹ️ If you want to run Truffle Dashboard on a custom host or port, please
+> configure that inside a `truffle-config.js` file in your Hardhat project
+> directory. See the
+> [truffle-config.js#dashboard](https://trufflesuite.com/docs/truffle/reference/configuration/#dashboard)
+> reference documentation to learn more about this; this section pertains only
+> to configuring @truffle/dashboard-hardhat-plugin itself.
 
-This `url` property respects any local `truffle-config.js`, so if you'd like to
-run Truffle Dashboard on a custom `host` or `port`, please make that file inside
-your Hardhat project directory and see the
-[truffle-config.js#dashboard](https://trufflesuite.com/docs/truffle/reference/configuration/#dashboard)
-reference documentation to learn how to configure this. The rest of this section
-of this README covers how to configure @truffle/dashboard-hardhat-plugin itself.
-
-#### Configure managed network
-
-This plugin defines the `truffleDashboard` namespace within the Hardhat config.
-
-Please see the following configuration settings (all of these are optional):
-
-- **`disableManagedNetwork`** _(default: `false`)_: Ignore `networkConfig` and
-  don't add anything to `config.networks`. See below to learn how to
-  [turn off automatic network management](#turn-off-automatic-network-management).
+To configure this plugin, add a `truffleDashboard` section to your Hardhat
+config. This namespace can contain the following fields (all are optional):
 
 - **`networkName`** _(default: `"truffle-dashboard"`)_: Specify the name of the
   Truffle Dashboard network, e.g. via the `--network <...>` command-line option.
 
 - **`networkConfig`**: A subset of the usual Hardhat network configuration
-  settings. This field is ignored if `disableManagedNetwork` is set to `true`.
+  settings.
 
   You can use this to override managed network behavior for any fields
   documented by Hardhat's
@@ -90,42 +83,6 @@ Please see the following configuration settings (all of these are optional):
 
   - **`timeout`** _(default: `0`; i.e., no timeout)_: Hardhat's normal default
     value for this is `40000` (40 seconds)
-
-<h4 id="turn-off-automatic-network-management">Turn off automatic network management</h4>
-
-If you'd rather this plugin not get fancy, you might want to just define the
-network yourself.
-
-To do this, make sure you specify `disableManagedNetwork: true`. This plugin
-will need to read from your configuration, so if you give your network a name
-different than the default `"truffle-dashboard"`, you'll also need to specify
-`networkName`.
-
-See this example `hardhat.config.ts`:
-
-```typescript
-import { HardhatUserConfig } from "hardhat/config";
-import "@nomicfoundation/hardhat-toolbox";
-import "@truffle/dashboard-hardhat-plugin";
-
-const config: HardhatUserConfig = {
-  networks: {
-    // ... other networks ...
-
-    dashboard: {
-      url: "http://localhost:24012/rpc",
-      timeout: 0 // disable timeout to have enough time for MetaMask pop-ups
-    }
-  },
-
-  // ... rest of config ...
-
-  truffleDashboard: {
-    networkName: "dashboard",
-    disableManagedNetwork: true
-  }
-};
-```
 
 ## Usage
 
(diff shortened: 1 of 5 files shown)


gnidan:
Oh shoot forgot the README change. I'll knock that out after some food.

gnidan:
cc @pcaversaccio I'm curious what you think of this change to @truffle/dashboard-hardhat-plugin, if you have a chance to take a quick look. 🙏 This change will have to go out in a breaking minor release (v0.2.0), so I'm hoping to avoid unpleasant surprises. I did a bunch of investigation at the bottom of #5958 that can explain some background, but tl;dr just skip to the Setup and Configuration sections of the [modified package README](https://github.com/trufflesuite/truffle/blob/5b5ef111fe6748b0b28fd72b23c8d8667f729f55/packages/dashboard-hardhat-plugin/README.md#setup).

pcaversaccio:
@gnidan I like how you extend the HH config - three (general) questions/feedback:
- Why is the default name `"truffle-dashboard"` and not e.g. `truffleDashboard` (which would make it more consistent with `TS` camelCase)? I guess if we want to change the name, now would be good timing.
- You should add as a note to the `README` that if you set `disableManagedNetwork == false` (i.e. the default behavior), the `url` will be `http://localhost:24012/rpc`. When I read it, I was not sure what it defaults to tbh. The same applies to `timeout`. I know that you mention it further below, but the context was not clear to me at first sight.
- Am wondering whether further HH `networkConfig` properties, besides the `timeout`, would benefit from another default value for increased devex. Have you investigated this already?

kevinbluer (commented):
Apologies in advance if this is a bit tricky to read / parse 😄 Happy to jump on a quick call if easier to walk through.

- *`disableManagedNetwork: true` and `config.networks[networkName]` missing is an error* - results in `Error HH100: Network dashboard doesn't exist` (running `npx hardhat run scripts/deployTestToken.ts --network dashboard`)
- *`disableManagedNetwork: false` and `config.networks[networkName]` present is an error* - this actually deploys for me (same deployment command as above)
- *`disableManagedNetwork: true` with `config.networks[networkName]` should connect to Dashboard with the settings listed in `config.networks[networkName]`* - This worked when Truffle Dashboard was running on the default port, but when I changed it to `24011` (via `truffle-config.js`) while the transaction reached dashboard, it resulted in the following (even after doing a recompilation): `Cannot decode transaction / Try running `truffle compile --all` in your Truffle project`
- *`disableManagedNetwork: false` without `config.networks[networkName]` should connect to Dashboard with the settings obtained via truffle-config.js and via `config.truffleDashboard.networkConfig`* - resulted in `Error HH100: Network dashboard doesn't exist` (running the same deployment command and with a `truffle-config.js` present with `24011` as the port)

Note that I'd been updating the network config in `network: { ... }` vs  `truffleDashboard.networkConfig: { ... }` which after re-reading the `README` may have been wrong?

packages/dashboard-hardhat-plugin/README.md
```diff
@@ -26,37 +35,120 @@ import "@truffle/dashboard-hardhat-plugin";
...
 
-Alternatively, you can skip the installation and fetch (and run) it remotely with `npx truffle dashboard`.
+Alternatively, you can skip the installation and fetch (and run) it remotely
+with `npx truffle dashboard`.
+
+This plugin automatically tells Hardhat about the `"truffle-dashboard"` network,
+so you don't need to anything to your `hardhat.config.ts` to get started! Just
+specify `--network truffle-dashboard` when running your usual Hardhat commands
+(e.g.
+`npx hardhat run ./scripts/deploy-contracts.ts --network truffle-dashboard`).
```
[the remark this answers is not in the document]
cds-amal: You have to add `import "@truffle/dashboard-hardhat-plugin";` to `hardhat-config.ts`. Credit to @cliffoo :)

cds-amal (commented):
First pass, will continue evaluating today.

cds-amal (changes requested):
Hey @gnidan ,

After meeting with you, I think we should drop the `disableManagedNetwork` setting because it's confusing, and isn't enough bang for the buck :)

gnidan:
> @gnidan I like how you extend the HH config

Thanks for taking a look @pcaversaccio, and I'm glad you like it!

> * Why is the default name `"truffle-dashboard"` and not e.g. `truffleDashboard` (which would make it more consistent with `TS` camelCase)? I guess if we want to change the name, now would be good timing.

Since this name gets used on the command line, I feel like `--network truffle-dashboard` is more stylistically correct than `--network truffleDashboard` (but maybe I'm just biased). I'm not happy about the discrepancy with the config, though.

What do you think of something like this:
```typescript
module.exports = {
  // ... rest of hardhat.config.ts
  truffle: {
    dashboard: {
      networkName: "truffle-dashboard",
      networkConfig: { /* ... */ }
    }
  }
};
```

Another approach might be this:
```typescript
module.exports = {
  // ... rest of hardhat.config.ts
  truffle: {
    dashboardNetworkName: "truffle-dashboard",
    dashboardNetworkConfig: { /* ... */ }
  }
};
```

I might prefer the second approach, just because I don't foresee a big need to sub-namespace `truffle` in the future. Curious what you think.

> * You should add as a note to the `README` that if you set `disableManagedNetwork == false` (i.e. the default behavior), the `url` will be `http://localhost:24012/rpc`. When I read it, I was not sure what it defaults to tbh. The same applies to `timeout`. I know that you mention it further below, but the context was not clear to me at first sight.

Looks like `disableManagedNetwork` is more confusing than it's worth, so I got rid of the option altogether. This change now means the plugin will force you NOT to include your own `config.networks["truffle-dashboard"]` _at all_.

Hopefully this makes the whole README a lot more clear, but maybe it's still worth highlighting that the defaults here mean "Hardhat will never timeout", and this might be behavior you don't want (you might want a 3 minute timeout or something)

> * Am wondering whether further HH `networkConfig` properties, besides the `timeout`, would benefit from another default value for increased devex. Have you investigated this already?

I gave this a bunch of thought when I went through all the Hardhat network settings... I could see `gasMultiplier` being a common one that people might want to mess with (maybe you want 1.2x buffer over gas estimates!), and `httpHeaders` is useful for people behind corporate proxies. But I don't know if this is something to mention as a concern for this plugin? I linked to the Hardhat configuration docs in the plugin's README, figuring that Hardhat itself is the best place to look for information on all these settings.

`timeout` is definitely a bit of an outlier, though, because this plugin supplies a different default than Hardhat's default.

I'll give this all another review on Monday after I'm done making a few more adjustments. Hope to get usage and configuration here to be very straightforward!

pcaversaccio:
> Since this name gets used on the command line, I feel like --network truffle-dashboard is more stylistically correct than --network truffleDashboard (but maybe I'm just biased). I'm not happy about the discrepancy with the config, though.

Well actually it's the opposite in Hardhat (and how I personally use it). If you for instance run `npx hardhat verify --list-networks` you will see how Hardhat has defined the network names for their Etherscan plugin: 

```bash
Networks supported by hardhat-etherscan:

╔══════════════════════╤════════════╗
║ network              │ chain id   ║
╟──────────────────────┼────────────╢
║ mainnet              │ 1          ║
╟──────────────────────┼────────────╢
║ ropsten              │ 3          ║
╟──────────────────────┼────────────╢
║ rinkeby              │ 4          ║
╟──────────────────────┼────────────╢
║ goerli               │ 5          ║
╟──────────────────────┼────────────╢
║ optimisticEthereum   │ 10         ║
╟──────────────────────┼────────────╢
║ kovan                │ 42         ║
╟──────────────────────┼────────────╢
║ bsc                  │ 56         ║
╟──────────────────────┼────────────╢
║ sokol                │ 77         ║
╟──────────────────────┼────────────╢
║ bscTestnet           │ 97         ║
╟──────────────────────┼────────────╢
║ xdai                 │ 100        ║
╟──────────────────────┼────────────╢
║ gnosis               │ 100        ║
╟──────────────────────┼────────────╢
║ heco                 │ 128        ║
╟──────────────────────┼────────────╢
║ polygon              │ 137        ║
╟──────────────────────┼────────────╢
║ opera                │ 250        ║
╟──────────────────────┼────────────╢
║ hecoTestnet          │ 256        ║
╟──────────────────────┼────────────╢
║ optimisticGoerli     │ 420        ║
╟──────────────────────┼────────────╢
║ moonbeam             │ 1284       ║
╟──────────────────────┼────────────╢
║ moonriver            │ 1285       ║
╟──────────────────────┼────────────╢
║ moonbaseAlpha        │ 1287       ║
╟──────────────────────┼────────────╢
║ ftmTestnet           │ 4002       ║
╟──────────────────────┼────────────╢
║ chiado               │ 10200      ║
╟──────────────────────┼────────────╢
║ arbitrumOne          │ 42161      ║
╟──────────────────────┼────────────╢
║ avalancheFujiTestnet │ 43113      ║
╟──────────────────────┼────────────╢
║ avalanche            │ 43114      ║
╟──────────────────────┼────────────╢
║ polygonMumbai        │ 80001      ║
╟──────────────────────┼────────────╢
║ arbitrumTestnet      │ 421611     ║
╟──────────────────────┼────────────╢
║ arbitrumGoerli       │ 421613     ║
╟──────────────────────┼────────────╢
║ sepolia              │ 11155111   ║
╟──────────────────────┼────────────╢
║ aurora               │ 1313161554 ║
╟──────────────────────┼────────────╢
║ auroraTestnet        │ 1313161555 ║
╟──────────────────────┼────────────╢
║ harmony              │ 1666600000 ║
╟──────────────────────┼────────────╢
║ harmonyTest          │ 1666700000 ║
╚══════════════════════╧════════════╝
```

These networks are also used via the command line like that:

```bash
npx hardhat verify --network networkName --constructor-args arguments.js <YOUR_CONTRACT_ADDRESS>
```

But maybe I'm biased here as well :) - maybe it's more consistent with Truffle internals and that would make sense ofc.

> I might prefer the second approach, just because I don't foresee a big need to sub-namespace truffle in the future

So regarding the config - what if you add for instance some compiler specifics in the future to `truffle` or other possible configs? E.g.:

```js
module.exports = {
  // ... rest of hardhat.config.ts
  truffle: {
    dashboard: {
      networkName: "truffle-dashboard",
      networkConfig: { /* ... */ },
    },
    solidity: {
      version: "0.8.19",
      settings: {
        optimizer: {
          enabled: true,
          runs: 999999,
        },
      },
    },
  },
};
```

I feel like the first option is more "future-proof" and offers more flexibility overall.

> Looks like disableManagedNetwork is more confusing than it's worth, so I got rid of the option altogether. This change now means the plugin will force you NOT to include your own config.networks["truffle-dashboard"] at all.

Gotcha.

> "Hardhat will never timeout", and this might be behavior you don't want (you might want a 3 minute timeout or something)

I would definitely document this behavior and blame me for that if users don't like lol ;-).

> But I don't know if this is something to mention as a concern for this plugin?

I guess if there are potential improvements on the devex part for the dashboard usage, we will figure out through new issues opened by people facing problems. I also quickly went through the HH properties again and don't see any default override need (except ofc for the `timeout`).

packages/dashboard-hardhat-plugin/README.md
```diff
@@ -26,37 +35,120 @@ import "@truffle/dashboard-hardhat-plugin";
...
+      url: "http://localhost:24012/rpc",
+      timeout: 0 // disable timeout to have enough time for MetaMask pop-ups
     }
   },
-  ...
-}
+
+  // ... rest of config ...
+
+  truffleDashboard: {
```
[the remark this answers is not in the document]
gnidan: This seems to be an issue with `yarn link`. See the new testing instructions in the writeup above!

gnidan:
> > Since this name gets used on the command line, I feel like --network truffle-dashboard is more stylistically correct than --network truffleDashboard (but maybe I'm just biased). I'm not happy about the discrepancy with the config, though.
> 
> Well actually it's the opposite in Hardhat (and how I personally use it). If you for instance run `npx hardhat verify --list-networks` you will see how Hardhat has defined the network names for their Etherscan plugin:
> 
> ```shell
> Networks supported by hardhat-etherscan:
> 
> ╔══════════════════════╤════════════╗
> ║ network              │ chain id   ║
> ╟──────────────────────┼────────────╢
> ║ mainnet              │ 1          ║
> ╟──────────────────────┼────────────╢
>                   .
>                   .
>                   .
> ╟──────────────────────┼────────────╢
> ║ optimisticEthereum   │ 10         ║
> ╟──────────────────────┼────────────╢
>                   .
>                   .
>                   .
> ╚══════════════════════╧════════════╝
> ```
> 
> These networks are also used via the command line like that:
> 
> ```shell
> npx hardhat verify --network networkName --constructor-args arguments.js <YOUR_CONTRACT_ADDRESS>
> ```
> 
> But maybe I'm biased here as well :) - maybe it's more consistent with Truffle internals and that would make sense ofc.

Hm, this is a convincing argument. I also pinged @alcuadrado in case he has an official recommendation. I'll tentatively plan to switch to camel-case.

> > I might prefer the second approach, just because I don't foresee a big need to sub-namespace truffle in the future
> 
> So regarding the config - what if you add for instance some compiler specifics in the future to `truffle` (i.e. you only want this compiler version if run via `truffle`) or other possible configs? E.g.:
> 
> ```js
> module.exports = {
>   // ... rest of hardhat.config.ts
>   truffle: {
>     dashboard: {
>       networkName: "truffle-dashboard",
>       networkConfig: { /* ... */ },
>     },
>     solidity: {
>       version: "0.8.19",
>       settings: {
>         optimizer: {
>           enabled: true,
>           runs: 999999,
>         },
>       },
>     },
>   },
> };
> ```
> 
> I feel like the first option is more "future-proof" and offers more flexibility overall.

I don't foresee needing to add Truffle-specific configuration settings to the Hardhat config. I don't really want Hardhat users to have to concern themselves with configuring Truffle... like in your example, if Truffle needs those Solidity settings, it should get them from the normal places inside the Hardhat config, not in an extra JSON blob.

This strategy won't always work, e.g. with configuring Truffle Dashboard itself, but I think we can just lean on the existing truffle-config.js as the place to customize that. But hopefully not many users will need to do this.

After thinking about this more, I'm going to go with the second approach, since `dashboardNetworkName` and `dashboardNetworkConfig` are only two fields. If I'm wrong, it won't be bad to deprecate those and namespace further in the future.

> > "Hardhat will never timeout", and this might be behavior you don't want (you might want a 3 minute timeout or something)
> 
> I would definitely document this behavior and blame me for that if users don't like lol ;-).
> 
> > But I don't know if this is something to mention as a concern for this plugin?
> 
> I guess if there are potential improvements on the devex part for the dashboard usage, we will figure out through new issues opened by people facing problems. I also quickly went through the HH properties again and don't see any default override need (except ofc for the `timeout`).

Sounds good. I'm still planning to do another pass on the README to make sure things like `timeout` are clear, and then yeah, I agree that we can just see if people open issues if anything's confusing or weird.

Thanks for this feedback!

packages/dashboard-hardhat-plugin/README.md
```diff
@@ -26,37 +35,120 @@ import "@truffle/dashboard-hardhat-plugin";
...
+Alternatively, you can skip the installation and fetch (and run) it remotely
+with `npx truffle dashboard`.
+
+This plugin automatically tells Hardhat about the `"truffle-dashboard"` network,
+so you don't need to anything to your `hardhat.config.ts` to get started! Just
+specify `--network truffle-dashboard` when running your usual Hardhat commands
+(e.g.
+`npx hardhat run ./scripts/deploy-contracts.ts --network truffle-dashboard`).
 
 ### Configuration
```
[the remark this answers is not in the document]
gnidan: Seems like `truffleDashboard` is to become the name of the network

pcaversaccio:
@gnidan sounds like you/we have a plan 😁. Thanks for the detailed answers!

packages/dashboard-hardhat-plugin/README.md
```diff
@@ -1,62 +1,124 @@
...
 
-This plugin enables you to see decoded transaction information (both the function signature and the values of any arguments passed) when using [Truffle Dashboard](https://trufflesuite.com/docs/truffle/how-to/use-the-truffle-dashboard/) with your Hardhat projects.
+## Summary
 
-It extends `npx hardhat compile` by sending the compiled artifacts to Truffle Dashboard, which in turn uses [@truffle/decoder](https://trufflesuite.com/docs/truffle/codec/modules/_truffle_decoder.html) for decoding, giving you that extra degree of visibility before signing the transactions with your browser-based wallet.
+This plugin automatically adds a `--network truffleDashboard` to your Hardhat
+project and tells Truffle Dashboard all the information it needs about your
+smart contracts.
+
+This plugin enables you to see decoded transaction information (both the
```
kevinbluer, line 17: Extremely minor nitpick with the dupicate opening to the paragraph. Wonder if "It also enables you..." might read a little better?

packages/dashboard-hardhat-plugin/README.md
```diff
@@ -1,62 +1,124 @@
...
 
-```ts
+```typescript
 import "@truffle/dashboard-hardhat-plugin";
 ```
 
 ## Setup
 
-This extension assumes you have `truffle` installed (either globally or in a local project context). If not, you can install it with `npm i -g truffle`. Beyond this you'll be able to start Truffle Dashboard with `truffle dashboard`. For reference, more information on using the Truffle Dashboard can be found [here](https://trufflesuite.com/docs/truffle/how-to/use-the-truffle-dashboard/).
+This extension assumes you have Truffle installed (either globally or in a local
```
kevinbluer, line 46: "This extension" should probably be changed to "This plugin" for consistency.

packages/dashboard-hardhat-plugin/README.md
```diff
@@ -1,62 +1,124 @@
...
+project context). If not, you can install it with `npm i -g truffle`. Beyond
+this you'll be able to start Truffle Dashboard with `truffle dashboard`. For
+reference, more information on using the Truffle Dashboard can be found
+[here](https://trufflesuite.com/docs/truffle/how-to/use-the-truffle-dashboard/).
+
+Alternatively, you can skip the installation and fetch (and run) it remotely
+with `npx truffle dashboard`.
 
-Alternatively, you can skip the installation and fetch (and run) it remotely with `npx truffle dashboard`.
+This plugin automatically tells Hardhat about the `truffleDashboard` network, so
```
kevinbluer, line 55: "...is slower than Hardhat _at_ computing signatures automatically"?

kevinbluer (approved):
Tested locally and all good! Few very minor copy suggestions in the README. Otherwise good to go from my side 👍

packages/dashboard-hardhat-plugin/README.md
gnidan: I'll try that. I know what you mean though.

packages/dashboard-hardhat-plugin/src/index.ts
```diff
@@ -1,61 +1,191 @@
...
+        `        dashboardNetworkConfig: {\n` +
+        Object.entries(dashboardNetworkConfig)
+          .map(([name, value]) => `          ${name}: ${util.inspect(value)}`)
+          .join(",\n") +
+        `\n` +
+        `        }\n` +
+        `      }\n` +
+        `    }\n\n` +
+        `Please see the README for more details about how to configure this plugin:\n` +
+        `  https://www.npmjs.com/package/@truffle/dashboard-hardhat-plugin`
```
cds-amal, line 183:
```suggestion
      `Manual network config disallowed.\n\n` +
        `This plugin manages your Truffle Dashboard network config for you,\n` +
        `but your Hardhat config also contains \`config.networks["${dashboardNetworkName}"]\`.\n\n` +
        `You can fix this error by removing \`config.networks["${dashboardNetworkName}"]\`.\n\n`  +
        `Please see the README for more details about how to configure this plugin:\n` +
        `  https://www.npmjs.com/package/@truffle/dashboard-hardhat-plugin`
```

cds-amal (commented):
Hi @gnidan, 

Tests look ok but the Errors call to action is still confusing. I changed it to refer to what caused the error in the tests, removing the wall of text about overriding configuration. That can be retrieved from the website/help URL. 

It's a suggestion, but please, wordsmith.

packages/dashboard-hardhat-plugin/src/index.ts
gnidan: You think that the error message shouldn't include the example configuration with all the defaults?

packages/dashboard-hardhat-plugin/README.md
gnidan: I did another pass over the whole README. Hopefully this is better now!

gnidan:
I did another editing pass over the whole README. Hopefully it reads a lot better now.

kevinbluer:
Awesome, looking great from my side @gnidan!

PR #5986, opened 2023-04-03
Add support for Polygon ZkEVM to Etherscan fetcher

haltman-at:
Etherscan has added support for Polygon ZKEVM!  Both mainnet and testnet.  I've added support to the Etherscan fetcher for these.

As usual, I didn't actually bother testing this PR, but if you want to, you could try it with the following addresses:

* Polygon ZKEVM mainnet (network ID 1101): 0xee1727f5074E747716637e1776B7F7C7133f16b
* Polygon ZKEVM testnet (network ID 1442): 0x1dAC955a58f292b8d95c6EBc79d14D3E618971b2


cliffoo (approved):
👍

Issue #5914, opened 2023-02-20
Debugger got wrong source location highlighted


lwedge99:
- [x] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

Debugger got wrong source location highlighted

## Steps to Reproduce

1. `truffle debug 0xfb6027725ccff056d8fe4065227fecd335fe123e8ab4da3a85a71377f7e9c6f9 --fetch-external`

2. step instructions `;3334`

## Expected Behavior

Debugger should highlight current source location

## Actual Results

<img width="400" alt="Snipaste_2023-02-20_15-30-38" src="https://user-images.githubusercontent.com/116623885/220040671-d07edb17-3eb2-41bf-b6bb-4c7a00971fa6.png">

The highlighting seems weird, and it persists if I keep stepping

<img width="356" alt="Snipaste_2023-02-20_15-44-05" src="https://user-images.githubusercontent.com/116623885/220043455-8b15df06-e057-4e95-8657-d8cedf9a00ef.png">

## Environment

* Operating System: MacOS 13.1
* Ethereum client:
* Truffle version (`truffle version`): 5.7.1
* node version (`node --version`): 18.14.0
* npm version (`npm --version`): 9.3.1

cliffoo:
Hey @lwedge99 thanks for opening this issue! I am able to reproduce this and it seems like the highlighting is indeed off.

@haltman-at fyi this is a mainnet transaction.

haltman-at:
Unfortunately this is likely due to a bad sourcemap; if it is we won't really be able to do anything.  I guess I can take a look and see sometime...?

lwedge99:
Hello, I have submitted a PR to fix this. Maybe you can check it out.

eggplantzzz:
looks like this should be fixed in version 5.8.3!

PR #5988, opened 2023-04-10
fix source map with multibyte characters

lwedge99:
## PR description

Make `getCharacterOffsetToLineAndColumnMapping()` handle multibyte characters properly. Previously, it counts all characters as 1 byte.

Addresses https://github.com/trufflesuite/truffle/issues/5914

## Testing instructions

Run `truffle debug 0xfb6027725ccff056d8fe4065227fecd335fe123e8ab4da3a85a71377f7e9c6f9 --fetch-external`. With this bugfix, source mapping seems correct

### Steps
1. Start ganache in terminal 1
  ```console
   ganache --fork
  ```
2. In terminal 2 
  ```console
  truffle debug 0xfb6027725ccff056d8fe4065227fecd335fe123e8ab4da3a85a71377f7e9c6f9 --fetch-external --url http://localhost:8545
  ```
3.  In Truffle debugger
  ```console 
  debug(localhost:8545:0xfb602772...)> b AggregationRouterV5.sol:2849
  debug(localhost:8545:0xfb602772...)>  c
  ```
4. observe source misalignment w/ highlight

### DIFF
--- a/packages/source-map-utils/index.js
+++ b/packages/source-map-utils/index.js
@@ -29,6 +29,9 @@ var SourceMapUtils = {
       }
 
       for (let i = 0; i < Buffer.from(character).length; i++) {
+        //because our character offsets here are in bytes, we need to
+        //pad out the line/column map as per the UTF-8 length of the
+        //characters so we're mapping *bytes* to line/columns
         mapping.push(loc);
       }
       column += 1;
--- a/packages/source-map-utils/index.js
+++ b/packages/source-map-utils/index.js
@@ -17,21 +17,25 @@ var SourceMapUtils = {
     var column = 0;
 
     source.forEach(function (character) {
+      let loc;
       if (character === "\n") {
         line += 1;
         column = -1;
 
-        mapping.push({
+        loc = {
           line: line,
           column: 0
-        });
+        };
       } else {
-        mapping.push({
+        loc = {
           line: line,
           column: column
-        });
+        };
       }
 
+      for (let i = 0; i < Buffer.from(character).length; i++) {
+        mapping.push(loc);
+      }
       column += 1;
     });
 
--- a/packages/source-map-utils/index.js
+++ b/packages/source-map-utils/index.js
@@ -17,7 +17,7 @@ var SourceMapUtils = {
     var column = 0;
 
     source.forEach(function (character) {
-      let loc;
+      let loc = { line, column };
       if (character === "\n") {
         line += 1;
         column = -1;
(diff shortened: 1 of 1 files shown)


gnidan:
😳 has this been the problem with misaligned source maps the whole time?!

awesome find... thank you for figuring this out!

gnidan (approved):

packages/source-map-utils/index.js
```diff
@@ -17,21 +17,25 @@ var SourceMapUtils = {
     var column = 0;
 
     source.forEach(function (character) {
+      let loc;
```
cds-amal, line 20: If `loc` is declared and initited, there's no need for the else clause below. 
```javascript
      let loc = { line, column };
```

cds-amal (approved):
Tested and looks good to me! I'd recommend waiting for @haltman-at to comment before merging.

## error - observed

```console
debug(localhost:8545:0xfb602772...)> b AggregationRouterV5.sol:2849
Breakpoint added at line 2849 in AggregationRouterV5.sol.
debug(localhost:8545:0xfb602772...)> c

AggregationRouterV5.sol:

2847:
2848:     /// @inheritdoc IUniswapV3SwapCallback
2849:     function uniswapV3SwapCallback(
               ^^^^^^^^^^
```

## fix - observed

```console
debug(localhost:8545:0xfb602772...)> b AggregationRouterV5.sol:2849
Breakpoint added at line 2849 in AggregationRouterV5.sol.
debug(localhost:8545:0xfb602772...)> c

AggregationRouterV5.sol:

2847:
2848:     /// @inheritdoc IUniswapV3SwapCallback
2849:     function uniswapV3SwapCallback(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

debug(localhost:8545:0xfb602772...)>
```

packages/source-map-utils/index.js
```diff
@@ -17,21 +17,20 @@ var SourceMapUtils = {
...
-        });
-      } else {
-        mapping.push({
-          line: line,
-          column: column
-        });
+        };
       }
 
+      for (let i = 0; i < Buffer.from(character).length; i++) {
```
haltman-at, line 31: I'd suggest adding a comment on this loop.  Something like, "because our offsets are given in bytes rather than in characters, we pad out the line/column map as per the UTF-8 length of the character".  I think that will make this clearer for future maintainers.

haltman-at (commented):
Thanks for catching this!  My only note is to suggest adding an explanatory comment.  On first read I wasn't convinced this was an appropriate solution.  I had to look it over a bit more to realize that yes this was actually a good way of handling this.  So, I'd like to request adding a comment to explain.

haltman-at:
Actually, I just added the comment myself, so you don't need to do anything.  I'll merge as soon as this passes CI!

Issue #5997, opened 2023-04-14
The `contract` and `deployer` packages read different settings to get a user-specified registry address


haltman-at:
## Issue

See also: https://github.com/trufflesuite/trufflesuite.com/issues/1420

When looking for a user-specified registry adress, the `deployer` and `contract` packages check different locations.  Truffle Contract checks `config.ens.registryAddress`.  However, the deployer checks `config.network_config.registry.address` (i.e., `config.networks[config.network].registry.address`).

This should be made consistent.  In order to do this in a nonbreaking way, presumably both packages should check both locations, with one of them given priority over the other.

(Potentially also, `registry` and `registryAddress` could be made synonyms in both locations?)

## Environment

* Truffle version (`truffle version`): 5.8.2

haltman-at:
Fixed in 5.9.0!

PR #6005, opened 2023-04-21
Consistify location of user ENS registry address setting

haltman-at:
## PR description

This PR addresses #5997. I'm not entirely happy with this one, but, well, I figured I ought to put it up as it is and we can iron out the problems in review.

Basically, I altered both Truffle Contract and Truffle Deployer to check 4 places for the ens registry address:
1. `config.network_config.registryAddress`
2. `config.network_config.registry.address`
3. `config.ens.registryAddress`
4. `config.ens.registry.address`

Note that specific network configs are checked before the overall ENS config, so that a specific network config can override the overall ENS config.

Now I say I altered Truffle Contract, but actually I didn't -- the easiest place to make these changes was actually the provisioner, rather than Contract itself.  Not sure if that's appropriate, but I didn't want to deal with the complication of doing it a different way.

I say I'm not entirely happy with this one, and there are two reasons for that:

1. Right now, since values are just combined with `??`, if you set `config.ens.registryAddress`, it's not possible to, say, set an explicit `null` in a network config to say that you shouldn't use the global registry for that specific network.  I can imagine people might want this in `deployer`.  So likely only `undefined` should defer to the next one, rather than `null`.
2. This PR is repeating the same code twice -- and if we add similar code to #5895 as @gnidan has suggested, that'll make 3 copies.  Moreover, if we account for point (1) just above, the code will get more complicated in all these locations.  That suggests it should be factored.  The question then is, where should such the factored version live?  One possibility would be to add a getter for this in `config` like we have for various other properties of the config; it could be called `config.ensRegistryAddress` or something (`config.networkRegistryAddress`? IDK).  The naming's a little ugly, but it seems probably better than the alternative?

Anyway, let me know what you think and what changes should be made to this.

## Testing instructions

I added tests to `deployer` to test the code changes there.  I didn't really directly test the `provisioner` changes, hm, maybe I should go back and do that... (adding tests there seemed pretty messy)

### DIFF
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -345,15 +345,64 @@ export const configProps = ({
         } catch {
           //if this throws, then there's no network config, whatever
         }
-        //NOTE: in what follows I'm not just using || or ?? because I want
-        //null to be respected but undefined not to be
-        const address = [
-          //note that network-specific locations go first
-          networkConfig?.registry?.address,
-          networkConfig?.registryAddress,
-          configObject.ens?.registry?.address,
-          configObject.ens?.registryAddress
-        ].find(x => x !== undefined);
+        //prefer specific over generic.  if there are two specific and they conflict,
+        //or two generic and they conflict (w/o specific to shadow), throw error.
+        //note: we treat null as a legitimate value here.
+        let address;
+        if (
+          networkConfig?.registry?.address !== undefined ||
+          networkConfig?.registryAddress !== undefined
+        ) {
+          if (
+            networkConfig?.registry?.address !== undefined &&
+            networkConfig?.registryAddress !== undefined
+          ) {
+            if (
+              networkConfig?.registry?.address ===
+              networkConfig?.registryAddress
+            ) {
+              //if both are defined and they're equal, use either one
+              address = networkConfig?.registry?.address;
+            } else {
+              //if both are defined but they're unequal, throw an error
+              throw new Error(
+                "Error: Conflicting values for registry address found in network config"
+              );
+            }
+          } else {
+            //if only one is defined, use that one
+            address =
+              networkConfig?.registry?.address ||
+              networkConfig?.registryAddress;
+          }
+        } else if (
+          configObject.ens?.registry?.address !== undefined ||
+          configObject.ens?.registryAddress !== undefined
+        ) {
+          if (
+            configObject.ens?.registry?.address !== undefined &&
+            configObject.ens?.registryAddress !== undefined
+          ) {
+            if (
+              configObject.ens?.registry?.address ===
+              configObject.ens?.registryAddress
+            ) {
+              //if both are defined and they're equal, use either one
+              address = configObject.ens?.registry?.address;
+            } else {
+              //if both are defined but they're unequal, throw an error
+              throw new Error(
+                "Error: Conflicting values for registry address found in project ens config"
+              );
+            }
+          } else {
+            //if only one is defined, use that one
+            address =
+              configObject.ens?.registry?.address ||
+              configObject.ens?.registryAddress;
+          }
+        }
+        //otherwise, address is just undefined
         return { address };
       },
       set() {
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -336,6 +336,31 @@ export const configProps = ({
           "Don't set config.timeoutBlocks directly. Instead, set config.networks and then config.networks[<network name>].timeoutBlocks"
         );
       }
+    },
+    ensRegistry: {
+      get() {
+        let networkConfig;
+        try {
+          networkConfig = configObject.network_config;
+        } catch {
+          //if this throws, then there's no network config, whatever
+        }
+        //NOTE: in what follows I'm not just using || or ?? because I want
+        //null to be respected but undefined not to be
+        const address = [
+          //note that network-specific locations go first
+          networkConfig?.registry?.address,
+          networkConfig?.registryAddress,
+          configObject.ens?.registry?.address,
+          configObject.ens?.registryAddress
+        ].find(x => x !== undefined);
+        return { address };
+      },
+      set() {
+        throw new Error(
+          "Don't set config.ensRegistry directly. Instead, set config.ens.registry, or config.ens.registryAddress, or config.networks[<network name>].registry, or config.networks[<network name>].registryAddress."
+        );
+      }
     }
   };
 };
--- a/packages/config/src/configDefaults.ts
+++ b/packages/config/src/configDefaults.ts
@@ -358,7 +358,7 @@ export const configProps = ({
       },
       set() {
         throw new Error(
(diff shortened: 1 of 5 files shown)


haltman-at:
Hm, this seems to be failing CI for a reason I don't understand.  Will have to take a look.

packages/deployer/index.js
```diff
@@ -18,9 +18,11 @@ class Deployer extends Deployment {
...
     this.known_contracts = {};
     if (options.ens && options.ens.enabled) {
-      options.ens.registryAddress = this.networks[this.network].registry
-        ? this.networks[this.network].registry.address
-        : null;
+      options.ens.registryAddress =
+        this.networks[this.network].registryAddress ??
+        this.networks[this.network].registry?.address ??
+        options.ens.registryAddress ??
+        options.ens.registry?.address;
```
cds-amal, line 25: As you mentioned, this is not ideal. I think it would be better to have Config resolver fn `ensRegistryAddress(networkName)`

packages/provisioner/src/index.ts
```diff
@@ -16,12 +16,20 @@ const provision = (contractAbstraction: any, truffleConfig: TruffleConfig) => {
...
     }
   }
 
   contractAbstraction.ens = truffleConfig.ens;
+  //HACK
+  contractAbstraction.ens.registryAddress =
+    truffleConfig.network_config.registryAddress ??
+    truffleConfig.network_config.registry?.address ??
+    truffleConfig.ens.registryAddress ??
+    truffleConfig.ens.registry?.address;
```
cds-amal, line 32: Same comment as above.

packages/deployer/index.js
haltman-at: A config resolver function?  What does that mean?  (Just to be clear what I was talking about above having just a getter that returns the *current* registry address, it wouldn't take any arguments.)

cds-amal: ah shoot, **_CURRENT!_** I forgot it gets set, and the getter won't need a network name.

gnidan:
> 2\. The question then is, where should such the factored version live?  One possibility would be to add a getter for this in `config` like we have for various other properties of the config; it could be called `config.ensRegistryAddress` or something (`config.networkRegistryAddress`? IDK).  The naming's a little ugly, but it seems probably better than the alternative?

I think we should do `.registry.address`, because I don't want to preclude other registry settings in the future (besides `address`), and it seems like a `registry` namespace would be useful for that.

It's a bit tricky though... I think the right place for this ENS config is per-network (i.e., `config.networks[config.network].ens`), but does that pose other problems? Can we get away with only configuring ENS at the individual network level?

haltman-at:
These are interesting thoughts on where the user settings in the config should live, but this doesn't answer the question of how to name the getter so as to avoid a naming conflict!

gnidan:
After consideration and discussion, let's just make whatever getter that doesn't conflict with user settings. Maybe `ensRegistry`, or heck, I'm not even altogether that opposed to `ensRegistryAddress`

haltman-at:
OK, I've updated this to use a config getter, even though it's ugly.  I also did some manual testing of the `contract` parts.

Really this is still not the cleanest even with the factoring, because I'm still just plugging the getter in to the particular places early on, converting it to a value, rather than letting config pass through and usign the getter as late as possible.  But whatever -- that can be a thing for the future, likely.

packages/config/src/configDefaults.ts
```diff
@@ -336,6 +336,31 @@ export const configProps = ({
...
+          networkConfig?.registry?.address,
+          networkConfig?.registryAddress,
+          configObject.ens?.registry?.address,
+          configObject.ens?.registryAddress
+        ].find(x => x !== undefined);
+        return { address };
+      },
+      set() {
+        throw new Error(
+          "Don't set config.ensRegistry directly. Instead, set config.ens.registry, or config.ens.registryAddress, or config.networks[<network name>].registry, or config.networks[<network name>].registryAddress."
```
cds-amal, line 361: This message doesn't match the selection  priority order. I think it would be helpful to include a link to the settings/reference docs entry which explains the nuances of these settings.

packages/config/src/configDefaults.ts
```diff
@@ -336,6 +336,31 @@ export const configProps = ({
...
+    ensRegistry: {
+      get() {
+        let networkConfig;
+        try {
+          networkConfig = configObject.network_config;
+        } catch {
+          //if this throws, then there's no network config, whatever
+        }
+        //NOTE: in what follows I'm not just using || or ?? because I want
+        //null to be respected but undefined not to be
```
cds-amal, line 349: Should the resolution order be called out?

cds-amal (approved):
Looks good @haltman-at, though I have a few comments. I think the new settings behavior should be documented on the website, and, imho, linked to by the new exception error the code can produce.

packages/config/src/configDefaults.ts
```diff
@@ -336,6 +336,31 @@ export const configProps = ({
...
+        const address = [
+          //note that network-specific locations go first
+          networkConfig?.registry?.address,
+          networkConfig?.registryAddress,
+          configObject.ens?.registry?.address,
+          configObject.ens?.registryAddress
```
cds-amal, line 355: This precedence resolution locks the API behavior and should be guarded by unit tests to maintain behavior to prevent introducing a subtle bug in future maintenances.

cds-amal (commented):
I think there should be some unit tests to guard ens precedence resolution.

packages/config/src/configDefaults.ts
haltman-at: You mean once such documentation exists as per https://github.com/trufflesuite/trufflesuite.com/issues/1420?

haltman-at: But sure I can change the message to match the priority order if that's relevant, I guess?

packages/config/src/configDefaults.ts
haltman-at: I don't think the particular order is important as long as network-specific is placed above generic.  You shouldn't be setting the same thing twice anyway, so to my mind there's no need to guarantee an order between them; if you set both and get an unpredictable result, that's your own problem for setting the same thing twice.

(To be clear I consider network specific and generic settings to be different -- that's not setting the same thing twice, and there guaranteeing the order matters.)

packages/config/src/configDefaults.ts
haltman-at: Again, as long as specific is before generic, I don't think there's a problem (see comments earlier).  In any case I *did* add a test that tests the specific-before-generic behavior; I don't see a need to add a test for the remainder of the order and lock it in.

cds-amal: That makes sense, though It would be helpful to inform the user they're in undefined territory if they set both network specific and generic ens settings.

haltman-at: Oh, hm, yeah, we could throw an error in that case!  I guess I'll go do that.

haltman-at: OK, I reworked the logic here to throw an error in that case.  I did use copy/paste but I think it's fine, it's all right next to each other and it's just two.

cds-amal (approved):
Looks good! Thanks, @haltman-at

Issue #4499, opened 2021-12-07
`Error: spawnSync /bin/sh ENOBUFS` when compiling Colony with native compiler


cameel:
## Issue

I'm trying to compile the [Colony](https://github.com/JoinColony/colonyNetwork) contracts using the native compiler and I'm running into an error:
```
Error: spawnSync /bin/sh ENOBUFS
```
The same contracts compile just fine when using an emscripten binary.

This seems to have already happened before (#3140) and the buffer size was increased to 10 MB in #3002. Something must have changed since then because this size seems insufficient now.

## Steps to Reproduce
```bash
git clone https://github.com/JoinColony/colonyNetwork.git
cd colonyNetwork/
git submodule update --init

curl -OL https://binaries.soliditylang.org/linux-amd64/solc-linux-amd64-v0.7.3+commit.9bfce1f6
chmod +x solc-linux-amd64-v0.7.3+commit.9bfce1f6

# NOTE: With https://github.com/trufflesuite/truffle/issues/4490 we could avoid the shell magic here
function solc { "$__SOLC_NATIVE_BINARY" "$@"; }
export -f solc
export __SOLC_NATIVE_BINARY=$(realpath solc-linux-amd64-v0.7.3+commit.9bfce1f6)

cat <<EOF >> truffle.js
module.exports['compilers']['solc']['version'] = "native";
module.exports['compilers']['solc']['docker'] = false;
module.exports['compilers']['solc']['parser'] = null;
EOF

npm install --legacy-peer-deps
npm update truffle
npx truffle compile
```

For the record, the top the main branch in the Colony repo while I'm testing it is at [`e06f7815902801c6b9cc66da8dbfa2aee2b9f874`](https://github.com/JoinColony/colonyNetwork/commit/e06f7815902801c6b9cc66da8dbfa2aee2b9f874).

## Expected Behavior

Compilation finishes without errors.

## Actual Results
```
Error: spawnSync /bin/sh ENOBUFS
    at Object.spawnSync (node:internal/child_process:1081:20)
    at spawnSync (node:child_process:812:24)
    at execSync (node:child_process:895:15)
    at Object.compile (/tmp/colonyNetwork/node_modules/truffle/build/webpack:/packages/compile-solidity/dist/compilerSupplier/loadingStrategies/Native.js:14:1)
    at /tmp/colonyNetwork/node_modules/truffle/build/webpack:/packages/compile-solidity/dist/run.js:200:1
[...]

PR #6008, opened 2023-04-25
Increase maxBuffer to 50MB when compiling using native compiler

d10r:
## PR description

Same issue as already tackled in https://github.com/trufflesuite/truffle/pull/3002, but which keeps getting up for more complex projects. Also see https://github.com/trufflesuite/truffle/issues/4499

## Testing instructions

I only can tell how to reproduce the issue before the change:
```
git clone https://github.com/superfluid-finance/protocol-monorepo
cd protocol-monorepo
yarn install
cd packages/ethereum-contracts
sed -i 's/version: "0.8.19",/version: "native",/g' truffle-config.js
npx truffle compile
```

You need to have solc version 0.8.19 in PATH.

### DIFF
--- a/packages/compile-solidity/src/compilerSupplier/loadingStrategies/Native.ts
+++ b/packages/compile-solidity/src/compilerSupplier/loadingStrategies/Native.ts
@@ -6,7 +6,7 @@ export class Native {
   load() {
     const versionString = this.validateAndGetSolcVersion();
     const command = "solc --standard-json";
-    const maxBuffer = 1024 * 1024 * 10;
+    const maxBuffer = 1024 * 1024 * 50;
 
     try {
       return {


hellwolf:
yes please.

gnidan (approved):
Seems like this makes sense! Here's one approval of two (and I'll add a few reviewers)

gnidan:
(Thanks for this, by the way! Appreciate the digging :)

packages/compile-solidity/src/compilerSupplier/loadingStrategies/Native.ts
```diff
@@ -6,7 +6,7 @@ export class Native {
   load() {
     const versionString = this.validateAndGetSolcVersion();
     const command = "solc --standard-json";
-    const maxBuffer = 1024 * 1024 * 10;
+    const maxBuffer = 1024 * 1024 * 50;
```
cds-amal, line 9: we can has more memory!

cds-amal (approved):
Thanks @d10r! Since @gnidan's ok with this, I'll throw another approval with a caveat. 

This solution doesn't solve the general case as there could be another complex contract requiring even more memory! An alternative fix would be to use spawn and collect the child's output via its event system to bypass buffer memory constraints. I'll open a PR for it.

PR #6019, opened 2023-05-05
Add even more ways to specify ens registry address so as to be consistent with docs

haltman-at:
The docs say you can do `config.network_config.ens.registry.address`, so, guess I'd better implement that.  I also added a test.

Will have to update the docs after this is merged...


gnidan (approved):
Thanks!

cds-amal (approved):
Looks good 👍 pending CI

PR #6034, opened 2023-05-11
Change default registry in config from null to undefined

haltman-at:
This `null` was screwing things up, I was using `null` to mean *no* registry, not *default* registry >_>

I guess likely I should have done things differently but oh well this is a simple fix and shouldn't break anything...


cds-amal (approved):
Lgtm

gnidan (approved):
Tested and works!

PR #6037, opened 2023-05-11
Upgrade solc dependencies to v0.8.20

sukanyaparashar:
## PR description

This PR upgrades the `solc` version to the latest **v0.8.20** in the `contract-schema`, `compile-solidity` and `compile-solidity-tests` packages.


haltman-at (approved):
Good catch, thank you!

PR #6041, opened 2023-05-11
Update typescript to ^4.9.5 and and ttypescript to 1.5.15

cliffoo:
In trying to update typescript for all packages to 5.x I found out that the latest ttypescript [does not support typescript 5.x](https://github.com/cevek/ttypescript/issues/140), and it appears to be out of maintenance. We should consider forgoing the use of this package, or at least swapping it for an equivalent.

Here are the changes:
- Updated ttypescript to latest for all packages that use it (codec, compile-solidity, compile-solidity-tests, db, db-kit, expect, from-hardhat, hdwallet), so that they can use the latest typescript 4.x, which I then updated them to.
- Updated remaining packages to typescript 5.x.

Basically all ts packages use 5.x (^5.0.4) if they can, 4.x (^4.9.5) otherwise. Hopefully ci passes.

---

**Change of plans**: Only updating all to the latest 4.x ts, and we'll deal with ttypescript and updating ts to 5.x in the future. Still updating ttypescript.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "nyc": "15.1.0",
     "prettier": "^2.5.1",
     "prs-merged-since": "^1.1.0",
-    "typescript": "^4.7.4"
+    "typescript": "^4.9.5"
   },
   "workspaces": {
     "packages": [
--- a/packages/abi-utils/package.json
+++ b/packages/abi-utils/package.json
@@ -39,7 +39,7 @@
     "jest-json-schema": "^2.1.0",
     "ts-jest": "29.0.3",
     "ts-node": "10.7.0",
-    "typescript": "^4.7.4"
+    "typescript": "^4.9.5"
   },
   "keywords": [
     "abi",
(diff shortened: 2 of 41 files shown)


haltman-at:
Huh, do we want to be using multiple versions of typescript like that?

cliffoo:
Ideally no but we currently have 4 versions 🤨 this cuts it to 2. I'm thinking packages stuck on 4.x can be transitioned over time to move on from ttypescript, and all catch up to using typescript 5.x at some point.

haltman-at:
Oh, wow!  I thought we'd unified it some time ago but I guess not...

haltman-at:
Hm, so looks like most things were on ^4.7.4, but dashboard-hardhat-plugin was on ^4.0.3, and from-hardhat was on ^4.3.5.  Wait, that's only three -- did I miss one?

cliffoo:
I think preserve is locked at 4.7.4 without caret, lerna-update-wizard counted that as an additional version!

eggplantzzz:
@gnidan can you weigh in on this??

gnidan (approved):
Good call to stick with ^4.9.5 until we can consider the ttypescript implications

Issue #6028, opened 2023-05-09
Specificity order for strings and bools in Encoder should swap when `strictBooleans` is enabled


haltman-at:
Right now, strings are more specific than bools.  If `strictBooleans` is on, this makes no sense; bools should be more specific.  This is relevant because it could pose problems for overload resolution in Truffle Call, e.g.  (Imagine if you wrote `call C f true` and got `f(string)` when you wanted `f(bool)`!)

Note that Encoder can't have specificity of types depend on the particular input, but there's no reason it can't have it depend on a flag like this.  It should.

## Environment

* Truffle version (`truffle version`): 5.8.4

haltman-at:
Released in 5.9.4!

PR #6044, opened 2023-05-13
Swap priority order of bools/strings if strictBooleans is on

haltman-at:
Addresses #6028.

The point of this is that, if there's an overloaded function with a bool overload and a string overload, then `truffle call Contract f true` ought to hit the bool overload, not the string overload.  Of course this only makes sense if we're restricted to string input, which is more or less what the `strictBooleans` flag means.

I added tests also.  You could also test it manually by making a function with a `string` overload and a `bool` overload and try calling it with `truffle call`.

*Arguably* this is a breaking change (to codec and encoder), but I'd consider it a bug fix.  Overload resolution is never guaranteed anyway! :)

### DIFF
--- a/packages/codec/lib/wrap/priority.ts
+++ b/packages/codec/lib/wrap/priority.ts
@@ -48,7 +48,7 @@ export function isMoreSpecific(
   if (type2.typeClass === "userDefinedValueType") {
     type2 = getUnderlyingType(type2, userDefinedTypes);
   }
-  let typeClasses = [
+  const typeClassesMinusStringAndBool = [
     ["options"],
     ["array"],
     ["struct", "tuple"],
@@ -56,17 +56,11 @@ export function isMoreSpecific(
     ["bytes"],
     ["function"],
     ["uint", "int", "fixed", "ufixed"],
-    ["enum"],
-    ["string"],
-    ["bool"]
+    ["enum"]
   ];
-  if (strictBooleans) {
-    //swap priority of strings and bools
-    typeClasses.pop();
-    typeClasses.pop();
-    typeClasses.push(["bool"]);
-    typeClasses.push(["string"]);
-  }
+  const typeClasses = typeClassesMinusStringAndBool.concat(
+    strictBooleans ? [["bool"], ["string"]] : [["string"], ["bool"]]
+  );
   //for each type, what's the first one it counts as?
   const index1 = typeClasses.findIndex(classes =>
     classes.includes(type1.typeClass)
--- a/packages/codec/lib/wrap/priority.ts
+++ b/packages/codec/lib/wrap/priority.ts
@@ -10,6 +10,7 @@ import { maxValue, minValue, places } from "./utils";
 export function isMoreSpecificMultiple(
   types1: Format.Types.OptionallyNamedType[],
   types2: Format.Types.OptionallyNamedType[],
+  strictBooleans: boolean,
   userDefinedTypes: Format.Types.TypesById
 ): boolean {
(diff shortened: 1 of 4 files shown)


haltman-at:
Oh wait, I should update the docs in this PR as well... will have to go back and add that...

haltman-at:
OK, updated this with documentation!

packages/codec/lib/wrap/priority.ts
```diff
@@ -10,6 +10,7 @@ import { maxValue, minValue, places } from "./utils";
 export function isMoreSpecificMultiple(
   types1: Format.Types.OptionallyNamedType[],
   types2: Format.Types.OptionallyNamedType[],
+  strictBooleans: boolean,
```
gnidan, line 13: I think this should be in an `options` style argument

packages/codec/lib/wrap/priority.ts
```diff
@@ -52,6 +60,13 @@ export function isMoreSpecific(
     ["string"],
     ["bool"]
   ];
+  if (strictBooleans) {
+    //swap priority of strings and bools
+    typeClasses.pop();
+    typeClasses.pop();
```
gnidan, line 66: This looks brittle; can we do it another way?

gnidan (commented):
Couple things I noticed (I won't request changes because I might not be around to dismiss the review for the next couple days)

packages/codec/lib/wrap/priority.ts
haltman-at: It could be if you really want, but this is sufficiently internal that I'm not really worried about that sort of thing -- there isn't a need to maintain a stable interface here, we can just change it as needed.  Adding yet another `options` here seems more confusing than just adding a flag.

packages/codec/lib/wrap/priority.ts
haltman-at: Sure, there's any number of alternative ways to do it, the question is what's good without being too repetitive.  I guess we could have everything but the last two, then tack on the last two depending on the flag?  Would that be less brittle?

haltman-at: (I've made this change, let me know if you think a different way would be better.)

Leena8686 (approved):

packages/encoder/test/overload.test.ts
```diff
@@ -279,6 +279,39 @@ describe("Overload resolution", () => {
...
+      assert.strictEqual(abi.inputs[0].type, "bool");
+      const selector = Codec.AbiData.Utils.abiSelector(abi);
+      assert.strictEqual(
+        tx.data,
+        selector +
+          "0000000000000000000000000000000000000000000000000000000000000001"
+      );
+    });
+
+    it("Encodes as string as last resort with strict booleans on", async () => {
```
eggplantzzz, line 301: is the `strictBooleans` option actually on here? I don't see it in the options

eggplantzzz: but that's for the previous test, shouldn't it be on line ~303?

haltman-at: ....oh wow you're absolutely right.  Oops.  Better fix that, then!

Leena8686 (approved):

packages/codec/lib/wrap/priority.ts
gnidan: I really think we should always lean in favor of options-style arguments, but I won't press the issue here

gnidan (approved):
Looks good!

Issue #6029, opened 2023-05-09
Allow decoder to use `4byte.directory`


haltman-at:
Related to and partially supplants #3941.

Right now, Decoder can't decode calls to proxy contracts.  There's a whole issue about this, #3941.  However, addressing that via recognition of proxy contracts may not be so easy, and so a quicker, simpler approach may be warranted.

@gnidan and @OnlyOneJMJQ have suggested that Decoder ought to be able to use `4byte.directory` to decode calls that otherwise can't be decoded.  These decodings could go in an interpretation on the `MessageDecoding`.  (There could be more than one, `4byte.directory` does handle collisions.)  Probably this would be done in Decoder rather than Codec.  However modifications would be necessary to allow decoding by ABI.  Again, that might all happen in Decoder rather than in Codec.  We'll see.

Anyway, this should be faster than the original plan for #3941!

## Environment

* Truffle version (`truffle version`): 5.8.4

haltman-at:
Released now in Truffle 5.10.0!

PR #6052, opened 2023-05-18
Add `parseSignature` functions to ABI utils

haltman-at:
## PR description

This PR adds functions to ABI utils for parsing signatures.  There's `parseFunctionSignature`, `parseEventSignature`, and `parseErrorSignature`, depending on the entry type you want out.  There's also just a bare-bones `parseSignature` that doesn't return a valid ABI entry, but which is used to back the other three.

This is intended as precursor work for our upcoming solution to #6029.

Note that the ABI entries generated by these functions will have default values for lots of fields... so e.g. `anonymous` will always be `false`, `outputs` will always be `[]`, etc.  Best we can do in the situation, seems to me.  It won't be relevant in the contexts we need this (#6029) anyway.

Also note that the parser doesn't fully validate the input... when it came to primitive types, I just checked that the types looked vaguely like primitive types, not that they're actual types.  (So e.g. `kafoo2e3` will pass validation.)  I could tighten this up if you think it's warranted.

(Also note that on the other hand, if you're dealing with a weird library signature that has like periods or spaces in it, that will *not* pass validation.  I'd consider that to be out of scope for this though... `abi-utils` deals in valid ABIs, not the weird stuff libraries sometimes do.  And of course for #6029 we should only need valid signatures.)

Anyway, I'm not going to bother trying to explain here how the parser works... just go read the code if you want to know that. :P

Also I just used ordinary `Error`s for when something went wrong, I didn't see any point in using like a custom `Error` class here, but again let me know if you think that should be changed.

## Testing instructions

I think the tests I added -- particularly the property-based tests -- should do the job pretty well! :)

Note: Turns out that TS doesn't recognize the arbitraries for `FunctionEntry`, `EventEntry`, etc, as being of the appropriate types. :-/  Maybe there's a way to fix that, but I don't really know much about arbitraries, so I just tossed in some coercions instead...

### DIFF
--- a/packages/abi-utils/lib/index.ts
+++ b/packages/abi-utils/lib/index.ts
@@ -1,5 +1,6 @@
 export * from "./types";
 export * from "./normalize";
 export * from "./signature";
+export * from "./parse";
 import * as Arbitrary from "./arbitrary";
 export { Arbitrary };
--- a/packages/abi-utils/lib/parse.ts
+++ b/packages/abi-utils/lib/parse.ts
@@ -1,0 +1,138 @@
+import { Parameter, FunctionEntry, ErrorEntry, EventEntry } from "./types";
+
+export function parseFunctionSignature(signature: string): FunctionEntry {
+  const { name, inputs } = parseSignature(signature);
+  return {
+    type: "function" as const,
+    name,
+    inputs,
+    outputs: [] as Parameter[],
+    stateMutability: "nonpayable" as const
+  };
+}
+
+export function parseErrorSignature(signature: string): ErrorEntry {
+  const { name, inputs } = parseSignature(signature);
+  return {
+    type: "error" as const,
+    name,
+    inputs
+  };
+}
+
+export function parseEventSignature(signature: string): EventEntry {
+  const { name, inputs } = parseSignature(signature);
+  return {
+    type: "event" as const,
+    name,
+    inputs: inputs.map(parameter => ({ ...parameter, indexed: false })),
+    anonymous: false
+  };
+}
+
+export function parseSignature(signature: string): {
+  name: string;
+  inputs: Parameter[];
+} {
+  let openParenIndex = signature.indexOf("(");
+  if (openParenIndex === -1) {
+    openParenIndex = signature.length; //set to end of string if not found
+  }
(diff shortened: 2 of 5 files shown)


packages/abi-utils/lib/parse.ts
```diff
@@ -0,0 +1,138 @@
...
+  const parameterList = signature.slice(openParenIndex);
+  const name = signature.slice(0, openParenIndex);
+  const inputs = parseParameterList(parameterList);
+  return {
+    name,
+    inputs
+  };
+}
+
+export function parseParameterList(parameterList: string): Parameter[] {
```
gnidan, line 50: Does this (and the other parameter list one) need to be exported? Or are these more internal concerns?

packages/abi-utils/lib/index.ts
```diff
@@ -1,5 +1,6 @@
 export * from "./types";
 export * from "./normalize";
 export * from "./signature";
+export * from "./fromSignature";
```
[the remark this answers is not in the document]
haltman-at: Ooh good point.  (Also ha you started this review before I fixed that mistake huh? :P )

packages/abi-utils/lib/parse.ts
```diff
@@ -0,0 +1,138 @@
...
+  const { parameters, remaining } =
+    parseParameterListWithRemainder(parameterList);
+  if (remaining !== "") {
```
cds-amal, line 53: Is there a case to allow trailing spaces?

haltman-at: I don't see any reason to.

packages/abi-utils/lib/index.ts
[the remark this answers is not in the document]
haltman-at: OK, did this!

packages/abi-utils/lib/parse.ts
haltman-at: Removed export on parse-with-remainder; left it on on the others.

gnidan:
Note we could probably use the ABI schema inside contract-schema to do the validation pretty easily! (Should probably open an issue for this also)

PR #6058, opened 2023-05-25
New feature: Visual debugger for Truffle Dashboard

eggplantzzz:
This PR implements a visual debugger in Truffle Dashboard.

To test this, check out the branch and build everything with `yarn bootstrap`. In `packages/dashboard` run `yarn start:server` and `yarn start:react` in two separate terminals. In a browser navigate to `localhost:3000`. Then click on the Debugger tab on the left to access it.

You can then connect MetaMask to the network where the tx you want to debug exists. Ensure you have the necessary compilations saved in your browser (see the Gist linked below for more details on this) and paste the tx hash in the input field.

See [this Gist](https://gist.github.com/eggplantzzz/9a88903c1dbbda5995c533a2d9823ee1) on getting compilations saved in the browser.


packages/codec/lib/format/utils/inspect.ts
```diff
@@ -30,6 +30,12 @@ export interface ResultInspectorOptions {
    * the address.  (By default it displays only the ENS name.)
    */
   noHideAddress?: boolean;
+   /**
+   * This flag, if set, causes mappings to be rendered via objects
+   * rather than Maps.  This is intended for compatibility and not
+   * recommended for normal use.
+   */
+  renderMappingsViaObjects?: boolean;
```
gnidan, line 38: Should this be part of this PR? Is this a merge remnant?

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -95,6 +98,62 @@ export class DashboardServer {
...
+        });
+        res.json({
+          hashes,
+          compilations: result.compilations
+        });
+      } else {
+        res.json({ compilations: [] });
+      }
+    });
+
```
gnidan, line 156: If you open this as a separate PR, we can approve+merge it real fast and then this PR's diff will be like 60 lines easier to read!

packages/codec/lib/format/utils/inspect.ts
haltman-at: This was that bit of code I added so that the visual debugger could display mappings.  Possibly Tyler could pull it out into a separate PR or something, but it is here deliberately, its's not some artifact or something.

packages/dashboard/lib/DashboardServer.ts
eggplantzzz: I don't understand, why would you pull this particular code out into its own PR?

packages/codec/lib/format/utils/inspect.ts
eggplantzzz: I pulled Harry's work out into https://github.com/trufflesuite/truffle/pull/6074

packages/dashboard/lib/DashboardServer.ts
eggplantzzz: Ok, I can do that. If it is helpful I can give people an overview-type tour of this PR. Just let me know!

eggplantzzz: Pulled this stuff out into https://github.com/trufflesuite/truffle/pull/6075

packages/dashboard/src/components/composed/Debugger/Stack.tsx
```diff
@@ -0,0 +1,98 @@
...
+        } else if (functionName) {
+          name = functionName;
+        } else {
+          name = "unknown function";
+        }
+        const displayAddress =
+          address === undefined ? "unknown address" : address;
+        const stackDisplay = `at ${name} [address ${displayAddress}]`;
+        return (
+          <div className="truffle-debugger-stack-item" key={index}>
```
cliffoo, line 78: Reference comment about `truffle-debugger-*` class names in src/components/MantineGlobal.tsx

packages/dashboard/src/components/composed/Debugger/Stack.tsx
```diff
@@ -0,0 +1,98 @@
...
+            {stackDisplay}
+          </div>
+        );
+      });
+      setOutput(entries);
```
cliffoo, line 83: This looks like the same imperative approach with an `output` state used in src/components/composed/Debugger/Variables.tsx. Reference the comment there.

packages/dashboard/src/components/composed/Debugger/Sources/index.tsx
```diff
@@ -0,0 +1,173 @@
...
+}: SourcesProps): JSX.Element {
+  const { classes } = useStyles();
+  const currentSourceIdRef = useRef(currentSourceId);
+  currentSourceIdRef.current = currentSourceId;
+
+  // display the first source when currentSourceRange is `undefined`
+  useEffect(() => {
+    if (currentSourceId === undefined) {
+      const sourceIds = sources.map(({ id }) => id);
+      setCurrentSourceId(sourceIds[0]);
```
cliffoo, line 93:
```suggestion
      setCurrentSourceId(sources[0].id);
```

packages/dashboard/src/components/composed/Debugger/Home/EtherscanApiKeyPrompt.tsx
```diff
@@ -0,0 +1,111 @@
...
+  reset: {
+    "&:hover": {
+      cursor: "pointer"
+    },
+    "color": "#E03131",
+    "fontSize": 14
+  },
+  halfWidth: {
+    width: "50%"
+  },
```
cliffoo, line 31: This isn't used.

packages/dashboard/src/components/composed/Debugger/Home/EtherscanApiKeyPrompt.tsx
```diff
@@ -0,0 +1,111 @@
...
+    backgroundColor:
+      theme.colorScheme === "dark"
+        ? "#082720"
+        : theme.colors["truffle-teal"][0],
+    border: `solid ${theme.colors["truffle-teal"][7]}`,
+    padding: 15
+  },
+  etherscanMessageContainer: {
+    width: "100%"
+  }
```
cliffoo, line 44: This also isn't used.

packages/dashboard/src/components/composed/Debugger/Home/EtherscanApiKeyPrompt.tsx
```diff
@@ -0,0 +1,111 @@
...
+}));
+
+function EtherScanApiKeyPrompt() {
+  const { classes } = useStyles();
+
+  const [etherscanApiKey, setEtherscanApiKey, removeEtherscanApiKey] =
+    useLocalStorage({
+      key: "etherscan-api-key"
+    });
```
cliffoo, line 53: Reference comment in src/components/composed/Debugger/index.tsx on naming the key.

packages/dashboard/src/components/composed/Debugger/Home/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+      justify="center"
+      pb="4%"
+      className={classes.welcomeMessage}
+    >
+      <Text size="lg" className={classes.halfWidth}>
+        Welcome!
+      </Text>
+      <Text size="sm" className={classes.halfWidth}>
+        Here you can paste a transaction hash to begin debugging it. You can
+        also start the debugger by clicking the Debug button on a signature
```
cliffoo, line 34: `... Simulate button ...`

packages/dashboard/src/components/composed/Debugger/PreparingSession.tsx
```diff
@@ -0,0 +1,50 @@
...
+  title: {
+    fontSize: 18,
+    marginLeft: 20
+  },
+  container: {
+    height: "40%"
+  }
+}));
+
+function PreparingSession({ ganacheLoggingOutput }: any): JSX.Element {
```
cliffoo, line 14: I think we know the prop type here.

packages/dashboard/src/components/composed/Debugger/PreparingSession.tsx
```diff
@@ -0,0 +1,50 @@
...
+          <Loader />
+          <div className={classes.title}>
+            Preparing your debugger session, this may take some time.
+          </div>
+        </Flex>
+        {ganacheLoggingOutput.length > 0 ? (
+          <pre
+            dangerouslySetInnerHTML={{ __html: ganacheOutput.join("<br>") }}
+          />
+        ) : null}
```
cliffoo, line 44:
```suggestion
        {ganacheLoggingOutput.length > 0 && (
          <pre>{ganacheOutput.join("\n")}</pre>
        )}
```

packages/dashboard/src/components/composed/Debugger/Sources/Source/LineNumber.tsx
```diff
@@ -0,0 +1,95 @@
...
+      highlightedTextTag +
+      lineNumber +
+      closingHighlightedTextTag +
+      "&nbsp;&nbsp;" +
+      `</span>`
+    : `<span class="${classes.lineNumber}">` +
+      "&nbsp;".repeat(lineNumberGutterWidth - lineNumber.toString().length) +
+      lineNumber +
+      "&nbsp;&nbsp;" +
+      `</span>`;
```
cliffoo, line 70: Can we move this to the part where the component returns? So we can avoid `dangerouslySetInnerHTML`.

packages/dashboard/src/components/composed/Debugger/Sources/Source/SourceLine.tsx
```diff
@@ -0,0 +1,51 @@
+import { createStyles } from "@mantine/core";
+
+const useStyles = createStyles((theme, _params, _getRef) => ({
+  lineNumber: {
+    color: theme.colorScheme === "dark" ? "#59534c" : "#c4b08d"
+  },
+  selected: {
+    backgroundColor: theme.colorScheme === "dark" ? "#4444aa60" : "#ffff0050"
+  },
```
cliffoo, line 9: These classes aren't used.

packages/dashboard/src/components/composed/Debugger/Sources/Source/SourceLine.tsx
```diff
@@ -0,0 +1,51 @@
...
+  const lineId = `${sourceId.slice(-10)}-${lineNumber}`;
+
+  return (
+    <div className={classes.sourceLine}>
+      <div
+        className={classes.content}
+        id={lineId}
+        dangerouslySetInnerHTML={{ __html: line }}
+      />
+    </div>
```
cliffoo, line 47: This nesting of a flexbox in a flexbox seems unnecessary.

cliffoo (changes requested):
First pass!

Two things that stood out to me as blocking:
- Vscode ts server emits type errors everywhere mantine components are used. This doesn't happen on develop. I'm not sure about the root cause but I suppose it must be related to upgrading mantine.
- A regression: I ran into this issue a couple of times where switching network doesn't prompt the `Confirm this is the desired network...` overlay. I'm not able to reliably reproduce this, but if this really is a problem then the cause can likely be located in `DashProvider`.

I can't really evaluate the correctness of programmatic debugger usage... But I didn't run into anything weird from what I've tried!

Some minor things:
- There are some eslint errors that need to be addressed (`react-hooks/exhaustive-deps`, `no-useless-escape`).
- There seems to be a line under the control buttons and input section. It's more visible in dark mode.
- Without the contrast of a control button that is brighter, the control buttons look like they can be clicked. Maybe fade them when there isn't an active debugger session.

Oh and this should probably be squashed when merging ;)

:tada:

packages/dashboard/package.json
```diff
@@ -33,31 +33,39 @@
...
     "ethers": "^5.6.9",
     "express": "^4.17.1",
+    "ganache": "7.9.0",
     "get-port": "^5.1.1",
+    "highlightjs-solidity": "^2.0.6",
     "idb": "^7.1.1",
-    "isomorphic-ws": "^4.0.1",
+    "lowlight": "^2.8.1",
     "object-hash": "^3.0.0",
-    "object-inspect": "^1.12.2",
```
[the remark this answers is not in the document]
eggplantzzz: Ah you're right, good catch!

packages/dashboard/src/components/MantineGlobal.tsx
```diff
@@ -42,6 +48,33 @@ function MantineGlobal(): JSX.Element {
             [`${truffleBgColorSelectors}, ${truffleOffBgColorSelectors}`]: {
               transition: "background-color 0.1s"
             },
+            ".trfl-AppShell-main": {
+              width: "auto"
+            },
```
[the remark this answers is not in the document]
eggplantzzz: Ok, I think this might have been left over from a css session with @OnlyOneJMJQ. I'll get rid of it!

packages/dashboard/src/components/composed/Debugger/Sources/Source/SourceLine.tsx
eggplantzzz: true enough!

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -98,41 +98,41 @@ export class DashboardServer {
...
+          etherscan: {
+            apiKey: etherscanKey
+          }
+        });
+
         result = (await fetchAndCompile(address, config)).compileResult;
       } catch (error) {
         if (!error.message.includes("No verified sources")) {
-          throw error;
+          return next(error);
```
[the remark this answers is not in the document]
eggplantzzz: Ah, yes. I think I misunderstood the express docs - looks like you need to pass the error to `next` only when inside a callback after something asynchronous.

packages/dashboard/src/components/composed/Debugger/Home/index.tsx
eggplantzzz: Oh yeah these should be consistent - the designs name the button "Debug" so I'll change the button so it has that text.

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -98,41 +98,41 @@ export class DashboardServer {
...
     }
 
-    this.expressApp.get("/fetch-and-compile", async (req, res) => {
-      const { address, networkId, etherscanApiKey } = req.query as Record<
-        string,
-        string
-      >;
-      let config;
+    this.expressApp.get("/fetch-and-compile", async (req, res, next) => {
+      let result;
```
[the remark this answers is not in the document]
eggplantzzz: actually a bit of code just gets moved inside the `try/catch` and a missing dep is added.

packages/dashboard/src/components/composed/Debugger/Sources/Source/LineNumber.tsx
eggplantzzz: I don't think we can since we need to generate raw html. Notice that we have a static class to designate text as highlighted. We generate the html for the source separately from this and then highlight the number in this step. I suppose we could just use a separate class...but we also need to be able to set whitespace as well for source formatting.

packages/dashboard/src/components/composed/Debugger/Sources/Source/SourceLine.tsx
eggplantzzz: Actually if you remove one, it breaks the alignment between the line number and the source material. There may be a better way to do it but...

eggplantzzz: although it looks like the `flex` property can be removed

eggplantzzz: oh if we remove `flex` then the highlighting gets messed up as there ends up being horizontal spaces between lines - I'm going to re-add it

packages/dashboard/lib/DashboardServer.ts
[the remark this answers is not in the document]
cliffoo: Missing dep?

packages/dashboard/src/components/composed/Debugger/Sources/Source/LineNumber.tsx
cliffoo: There are two things here:

The first thing is related to my comment [here](https://github.com/trufflesuite/truffle/pull/6058#discussion_r1254978658) about not putting debugger-related styles in `<Global />`, so the class that does the highlight belongs here.

The second thing is that even if the class name _has_ to be static, we still don't need to write markup in string and then `dangerouslySetInnerHtml`. It would just be

```tsx
return (
  ...
  <div> { /* Don't dangerouslySetInnerHtml */ }
    <span class={`${classes.lineNumber} ${highlighted ? "your-static-class" : undefined}`}>
      {lineNumber}
    </span>
  </div>
  ...
)
```
But again, replace `"your-static-class"` with a new `classes.` class that you define here, not globally.

packages/dashboard/src/components/MantineGlobal.tsx
```diff
@@ -42,6 +48,33 @@ function MantineGlobal(): JSX.Element {
...
+            ".truffle-debugger-input-and-button > div > input": {
+              borderTopRightRadius: 0,
+              borderBottomRightRadius: 0,
+              border: 0,
+              height: 42,
+              backgroundColor:
+                theme.colorScheme === "dark"
+                  ? theme.colors["truffle-brown"][5]
+                  : "#FFFFFF"
+            },
```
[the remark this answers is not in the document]
eggplantzzz: I moved ".truffle-debugger-input-and-button" and the class below into the component file. The ".truffle-debugger-text-highlighted" is currently global because those classes are attached in the utilities before they get to the component. Should we architect this differently?

packages/dashboard/src/components/composed/Debugger/Sources/Source/LineNumber.tsx
eggplantzzz: I made a comment on the comment on `MantineGlobal` around line 54. Currently the ".truffle-debugger-text-highlighted" class gets applied in about 3 different places and so it is a global. In my reply I ask if we should consider architecting this differently. The static class doesn't apply to everyting in `lineNumber`, only a portion so I don't think we can do exactly what you suggest unless I'm being dense and not fully understanding right now.

packages/dashboard/src/components/composed/Debugger/Stack.tsx
eggplantzzz: this class turns out to be unused - I'll remove it

packages/dashboard/src/components/composed/Debugger/index.tsx
```diff
@@ -0,0 +1,328 @@
...
+    if (e.keyCode === 13) {
+      startDebugger();
+    }
+  };
+
+  const buttonStyles = {
+    height: "42px",
+    borderTopLeftRadius: "0px",
+    borderBottomLeftRadius: "0px"
+  };
```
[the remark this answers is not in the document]
eggplantzzz: The problem here is that this gets applied to a Mantine component. It doesn't accept a `className` property. Is there a way to do this? Currently it's getting passed as `style` which Mantine components accept for custom styling I guess.

packages/dashboard/src/components/composed/Debugger/index.tsx
```diff
@@ -0,0 +1,328 @@
...
+    }
+  }, [currentStep, currentSourceId]);
+
+  // check whether we need to scroll to a breakpoint
+  // this is to ensure the source has fully rendered before scrolling
+  useEffect(() => {
+    if (goToBreakpoint !== null) {
+      const { sourceId, line } = goToBreakpoint;
+      // @ts-ignore
+      scrollToLine({ sourceId, line });
```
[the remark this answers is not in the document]
eggplantzzz: Hmm weird, I wonder why this isn't causing problems...or maybe it is and I haven't noticed? The scrolling worked, maybe there is some coercion somewhere? I'll correct this.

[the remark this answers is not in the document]
eggplantzzz: Yeah looks like there is some coercion: testing in `scrollToLine` it is a number for whatever reason.

packages/dashboard/src/components/composed/Debugger/utils/source/source.ts
```diff
@@ -0,0 +1,230 @@
...
+      // line where a multiline comment begins
+      sourceWithSpans.push(lowlightCommentSpan + line);
+      inMultilineComment = false;
+    } else {
+      // line of code not in a multiline comment
+      sourceWithSpans.push(line);
+    }
+  }
+  return sourceWithSpans;
+}
```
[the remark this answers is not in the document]
eggplantzzz: Ah yes, this got replaced by `completeMultilineSpans` and never got removed!

packages/dashboard/src/components/composed/Sidebar/Middle/NavButton.tsx
```diff
@@ -1,28 +1,27 @@
...
       borderRadius: radius.sm
     },
-    enabled: {
-      "&:hover": {
-        backgroundColor:
-          colorScheme === "dark"
-            ? fn.darken(colors["truffle-brown"][8], 0.08)
-            : fn.darken(colors["truffle-beige"][4], 0.08)
-      }
-    },
```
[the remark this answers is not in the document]
eggplantzzz: Yep, you got it!

packages/dashboard/src/components/composed/Debugger/Variables.tsx
```diff
@@ -0,0 +1,129 @@
...
+          })
+          .filter((item: JSX.Element | undefined) => item);
+        if (variableValues.length > 0) {
+          entries.push(
+            <div key={section}>
+              <div className={classes.variablesTypes}>{section}</div>
+              <ul className={classes.variablesSection}>{...variableValues}</ul>
+            </div>
+          );
+        }
```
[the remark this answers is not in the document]
eggplantzzz: Let's talk about this when we meet today.

packages/dashboard/src/components/composed/Debugger/index.tsx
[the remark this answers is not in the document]
eggplantzzz: The `Button` component takes a `classNames` prop which you pass an object with a `root` property. This has been updated.

packages/dashboard/src/components/composed/Debugger/index.tsx
```diff
@@ -0,0 +1,328 @@
...
+  const [inputValue, setInputValue] = useInputState("");
+  const [sessionUpdated, { increment: sessionTick }] = useCounter();
+  const {
+    operations,
+    state: {
+      debugger: { sources, unknownAddresses, session, txToRun }
+    }
+  } = useDash()!;
+
+  const [etherscanApiKey] = useLocalStorage({ key: "etherscan-api-key" });
```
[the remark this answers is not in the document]
eggplantzzz: done

packages/dashboard/src/components/composed/Sidebar/index.tsx
```diff
@@ -4,11 +4,15 @@ import Middle from "src/components/composed/Sidebar/Middle";
...
-    minWidth: 370,
-    maxWidth: 370,
-    borderRight: "none"
+    height: "100vh",
+    width: 370,
+    borderRight: "none",
+    backgroundColor:
+      theme.colorScheme === "dark"
+        ? theme.colors["truffle-brown"][8]
+        : theme.colors["truffle-beige"][4]
```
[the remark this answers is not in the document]
eggplantzzz: Ok, I got rid of this and a couple of other instances of this.

packages/dashboard/src/components/composed/Debugger/Sources/Source/LineNumber.tsx
eggplantzzz: updated!

cliffoo (approved):
An approve with two asides.

First thing we can fix later: Switching network should void the debugger session, but it doesn't.

Second thing I considered blocking so I pushed a fix: Ganache indexeddb usage isn't cleaned up and builds up over time. For the sake of simplicity I just added a clean-up on start.

Good stuff!

cliffoo:
(reminder to squash then merge)

PR #6063, opened 2023-06-01
pass solc when available to profiler to avoid fetching the solc list again

eggplantzzz:
This PR is an optimization. In certain cases during the compilation flow, the CompilerSupplier is instantiated twice and it fetches the list of Solidity compiler versions twice. This is because it needs to see what compiler versions are available when making sure it has the right version and because we need the parser to determine what a sources dependencies are for compilation. So this PR passes the compiler to the Profiler when it has it so that it doesn't need to be fetched again.


haltman-at (approved):
Makes sense to me!

PR #6071, opened 2023-06-05
Dashboard notify user if no Frame account is connected

cliffoo:
Thanks to curiousapple.eth and @pcaversaccio for bring up the underlying issue!

When using Frame with Dashboard, Dashboard doesn't know what to do if no Frame account is connected. This PR sidesteps wagmi and finds out if that's the case, provides a helpful notification asking the user to connect a Frame account, and hides the notification when Dashboard is ready.

Environment: Frame v0.6.6 (amd64, deb), Frame Companion v0.10.2 (Chrome).

Tested with Frame Companion's two options: Injecting as Frame / MetaMask.


gnidan (approved):
Oh nice, so that's what was wrong! 

Does everything else work when the account is connected properly?

cliffoo:
From a few test transactions it looks like it. Though I did just discover an unhandled case: If you deny connection in Frame, Dashboard does nothing again because Frame remembers that you denied connection to this host.

The fix would be the same. @gnidan Should we generalize this instead of having checking for more regex? The notification can just be titled "Wallet error" and we'll put `err.message` in notification body.

In this case `err.message` is `"Permission denied, approve localhost:3000 in Frame to continue"`

gnidan:
> Should we generalize this instead of having checking for more regex?

I'm not sure we should generalize this... it might be alright to generalize as "something is wrong with Frame, go check", but I think it's just nicer to specify what is wrong when we know.

PR #6075, opened 2023-06-07
add new route in preparation for visual debugger PR

eggplantzzz:
This PR adds a route in preparation for getting https://github.com/trufflesuite/truffle/pull/6058 merged.

The visual debugger will be fetching verified sources externally in the case that it doesn't have compilations for some of the contracts the tx touches. In order to implement this, this route is added. It allows dashboard to give the server a list of addresses and a network to use to attempt to fetch the source material. This route fetches what it can, compiles everything, and then returns it to dashboard.

### DIFF
--- a/packages/dashboard/lib/DashboardServer.ts
+++ b/packages/dashboard/lib/DashboardServer.ts
@@ -106,12 +106,8 @@ export class DashboardServer {
       let config;
       try {
         config = Config.detect();
-      } catch (error) {
-        const notFound = "Could not find suitable configuration file.";
-        if (!error.message.includes(notFound)) {
-          throw error;
-        }
-      }
+        // we'll ignore errors as we only get the config for the api key
+      } catch {}
 
       // a key provided in the browser takes precedence over on in the config
       let etherscanKey: undefined | string;
--- a/packages/dashboard/lib/DashboardServer.ts
+++ b/packages/dashboard/lib/DashboardServer.ts
@@ -2,6 +2,8 @@ import express, { Application, NextFunction, Request, Response } from "express";
 import path from "path";
 import getPort from "get-port";
 import open from "open";
+import { fetchAndCompile } from "@truffle/fetch-and-compile";
+import { sha1 } from "object-hash";
 import { v4 as uuid } from "uuid";
 import Config from "@truffle/config";
 import {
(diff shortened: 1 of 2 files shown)


gnidan (approved):

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -95,6 +98,62 @@ export class DashboardServer {
...
+      >;
+      let config;
+      try {
+        config = Config.detect();
+      } catch (error) {
+        const notFound = "Could not find suitable configuration file.";
+        if (!error.message.includes(notFound)) {
+          throw error;
+        }
+      }
```
cliffoo, line 114: I don't think this should throw? Since this config isn't actually used for fetch and compile, only to grab the etherscan key if it's there.

packages/dashboard/lib/DashboardServer.ts
```diff
@@ -95,6 +98,62 @@ export class DashboardServer {
...
+      });
+
+      let result;
+      try {
+        result = (await fetchAndCompile(address, config)).compileResult;
+      } catch (error) {
+        if (!error.message.includes("No verified sources")) {
+          throw error;
+        }
+      }
```
cliffoo, line 141: Also don't throw if no sources? It can be handled like how `!result` is handled below with `res.json({ compilations: [] })`

eggplantzzz: The point of this is that `fetchAndCompile` throws with that error message in the case that it can't find verified sources. We want to not throw in that case, but rather throw when it fails in some other unexpected way. Does that make sense?

packages/dashboard/lib/DashboardServer.ts
eggplantzzz: Hmm, I don't know. The point here is the same with the other `try/catch`. If we can't find a config, we want to move on and not care. However, if it throws in some unexpected way we want to know about it. In this case though it might not matter...maybe we should just leave the `catch` empty? Do we want to know when it fails for some other reason than not being able to find it?

packages/dashboard/lib/DashboardServer.ts
cliffoo: Yes. I was just thinking it'd be less work to write a case to receive a 500 in react? Maybe we can `console.error` it, which should show in the Dashboard terminal, and still send the empty compilations response?

packages/dashboard/lib/DashboardServer.ts
cliffoo: Do nothing/log, and then move on to fetch and compile seems reasonable to me. But I understand your point with this and the try/catch block below though. If in react we also wrap the request in try/catch then my concerns are not  important here.

eggplantzzz: Ok, in this case I think it makes sense to just ignore errors with `Config.detect` - the other one below, instead of throwing should probably 500 like you said

cliffoo (approved):

PR #6090, opened 2023-06-13
Replace bootstrap with install in scripts/workflows

haltman-at:
Since `yarn bootstrap` seems to be causing such problems, and I can't figure out the root cause, here's a simple workaround... replacing `yarn bootstrap` with `yarn install`.   There used to be a difference between these, but due to every package using `prepare` now, there no longer is.  Hopefully this is all OK and solves the problem?

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -45,7 +45,7 @@ jobs:
 
     - run: npm install -g yarn
 
-    - run: yarn bootstrap
+    - run: yarn install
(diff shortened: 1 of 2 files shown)


cliffoo (approved):
We should probably also update readme (`under ###Development`) and maybe remove the `bootstrap` command in package.json?

haltman-at:
> We should probably also update readme (under `### Development`) and maybe remove the `bootstrap` command in package.json?

Hm, possibly -- right now the problem is only occurring on CI.  @gnidan, any opinion on all this?  I'm going to just merge this so that CI can be less of a disaster, and we can figure this out after I figure...

Issue #6095, opened 2023-06-15
An error occurred in the command `truffle dashboard` when using vscode remote development


fsrm365:
- [ x ] I've [opened a support ticket](https://trufflesuite.zendesk.com/hc/en-us/requests/new) before filing this issue.

---------------------------

## Issue

/code/coins/MetaCoin$ truffle dashboard
Truffle Dashboard running at http://localhost:24012
DashboardProvider RPC endpoint running at http://localhost:24012/rpc
Error: spawn xdg-open ENOENT
    at Process.ChildProcess._handle.onexit (node:internal/child_process:283:19)
    at onErrorNT (node:internal/child_process:476:16)
    at processTicksAndRejections (node:internal/process/task_queues:82:21)

## Steps to Reproduce

I use the remote development of vscode. The remote server is a linux system. No browser installed. lead to errors. 

## Expected Behavior

Is it possible to add an option to the command `truffle dashboard` to decide whether to open the browser automatically?

## Environment

* Operating System:  local:mac os 13.4 remote: ubuntu 22.04
* Ethereum client: None
* Truffle version (`truffle version`): Truffle v5.9.4 (core: 5.9.4)
* node version (`node --version`): Node v18.16.0
* npm version (`npm --version`): 9.7.1

gnidan:
Good catch! Thanks for raising this. We're implementing this for you over in #6096.

Something else you might want to be aware of: if you're trying to use `truffle dashboard` via a different machine, you might want to bind the hostname to 0.0.0.0 (as opposed to 127.0.0.1), since you won't be able to access the latter from your terminal machine.

haltman-at:
This is released now in Truffle 5.10.0!

fsrm365:
@gnidan @haltman-at thank you

PR #6096, opened 2023-06-15
Allow Truffle Dashboard to run without launching a browser window automatically

cliffoo:
Addresses #6095

It can be useful in some environments to not want Dashboard to open in browser on start.

To test:
- `truffle dashboard --no-auto-open`
- `truffle dashboard` in a Truffle project with `dashboard.autoOpen` set in truffle-config.js

CLI value takes precedence.


gnidan:
For the CLI, we should just call this `--open`. Or at least make it kebab-case!

gnidan (approved):
Thanks @cliffoo!

PR #6097, opened 2023-06-15
Add missing codec component for new error type

haltman-at:
I merged #6057 while ignoring that it adds a new codec error type which needs a component, crap!  So uh I added that quickly.

I didn't actually test this component because this error should never actually appear onscreen and also I just want develop not broken.  It builds, that's good enough for me. :P  Next time I write a component, which will be something that could appear onscreen, I will bother to test it. :P


gnidan (approved):
Maybe wait for @cliffoo on this one though

cliffoo (approved):
# 👍

PR #6101, opened 2023-06-18
Update README.md

menterche:
Clarified that the transition goes to the MIT License page


eggplantzzz:
Thanks for the proposed change @menterche! However, I think it is a bit redundant and unnecessary. We're going to close this for now.

PR #6107, opened 2023-06-19
Update README.md

Valdemaroff:
I have added coverage


eggplantzzz:
Thanks for the proposed change! We used to have code coverage but removed that badge after it broke. We never ended up getting it fixed so I suppose we'll want to leave the badge off until we get it working again. Feel free to take a stab at fixing it of course :) Closing this for now.

PR #6116, opened 2023-06-22
Specify "magic" variables' type only in component tooltip, not via prefix

gnidan:
## PR description

This updates the `MagicValue` component to omit the prefix specifying which kind of magic variable it is, instead putting that information in the tooltip.

See: 

Screenshot 2023-06-21 at 11 06 17 PM

### DIFF
--- a/packages/codec-components/src/react/components/codec/format.values.magic-value.tsx
+++ b/packages/codec-components/src/react/components/codec/format.values.magic-value.tsx
@@ -4,23 +4,31 @@ import { createCodecComponent } from "../../utils/create-codec-component";
 import { Result } from "./format.values.result";
 import { NextBracketDepth } from "../providers/next-bracket-depth";
 import { InjectedNode } from "../providers/injected-node";
+import { useInjectedNode } from "../../contexts/injected-node";
 import { Container } from "../common/container";
 import { Code } from "../common/code";
 
 export const { MagicValue } = createCodecComponent(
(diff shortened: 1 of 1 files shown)


gnidan:
cc @eggplantzzz so you're aware

haltman-at:
Seems sensible to me!  Although I'm a bit wary to approve it without @cliffoo. :P

packages/codec-components/src/react/components/codec/format.values.magic-value.tsx
```diff
@@ -15,9 +15,9 @@ export const { MagicValue } = createCodecComponent(
       <Container
         prefix={
           <>
-            <Code title="type: magic (debugger-only)">{type.variable}</Code>
-            <Code type="colon">:</Code>
-            <Code type="bracket">{"{"}</Code>
+            <Code title={`type: ${type.variable}`} type="bracket">
+              {"{"}
+            </Code>
           </>
```
cliffoo, line 21: No need for the `<></>` fragment anymore!

cliffoo (approved):
👍

cliffoo (changes requested):
Oops almost forgot, need injected nodes! See [here](https://github.com/trufflesuite/truffle/commit/84225a69ad86a56c7b89357a36f793e137c56898).

cliffoo (approved):

Issue #5913, opened 2023-02-17
Address values should have contract interpretation


haltman-at:
Well, not yet they shouldn't.  But once #5895 is merged they should, for consistency.  See #5895 for more about this.

## Environment

* Truffle version (`truffle version`): 5.7.7

cliffoo:
Addressed in [v5.10.2](https://github.com/trufflesuite/truffle/releases/tag/v5.10.2)

PR #6130, opened 2023-06-28
Add `contractClass` interpretation to address values

haltman-at:
## PR description

Addresses #5913.  Adds a `contractClass` interpretation to address values.  Basically, if we can identify what sort of contract a given address is, it will have this interpretation with more information; otherwise it won't.

Yes, this means that the address/contract distinction is now pretty silly -- they now contain the same information but formatted differently.  Oh well.  I don't want to make a breaking change to make them more unified.

The code is pretty straightforward, not much to explain.  I could have further eliminated some minor recomputation by just using `decodeContract` and then doing everything based on the results of that, but I decided it was clearer to do it this way, even if that means calling `Evm.Utils.toAddress` twice. :P

I also updated the inspector so that it now basically just treats addresses like contracts.  Could have distinguished them some but that would have been more work.

I did *not* update `codec-components`.  I figure I can worry about that later; there are likely multiple changes I want to make to address and contracts there.

## Testing instructions

I added a debugger test for it, I figured that should be enough.

### DIFF
--- a/packages/codec/lib/basic/decode/index.ts
+++ b/packages/codec/lib/basic/decode/index.ts
@@ -222,7 +222,7 @@ export function* decodeBasic(
       }
       bytes = removePadding(bytes, dataType, paddingMode);
       const address = Evm.Utils.toAddress(bytes);
-      let decoded = {
+      let decoded: Format.Values.AddressValue = {
         type: dataType,
         kind: "value" as const,
         value: {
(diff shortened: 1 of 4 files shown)


cliffoo (approved):
Makes sense! Although maybe someone else should take a look also...

gnidan (approved):

Issue #6080, opened 2023-06-09
Decoder should allow `bigint` or `string` in place of `number` in input fields for logs & transactions


haltman-at:
## Issue

Right now these are assumed to be numbers, bigints should be allowed as well.

Edit: Strings as well, apparently.

## Environment

* Truffle version (`truffle version`): 5.9.4

cliffoo:
Addressed in [v5.10.2](https://github.com/trufflesuite/truffle/releases/tag/v5.10.2)

PR #6131, opened 2023-06-28
Allow strings and bigints instead of numbers in decoder input fields

haltman-at:
## PR description

Addresses #6080.  This makes it so that fields in `Log` or `Transaction` that previously were `number` are now `number | string | bigint`, and fields that were `number | null` are now `number | string | bigint | null`.

Most of these fields are ignored, so the type is changed and that's all.

The one changed field that's actually used is `blockNumber`, so I just changed things so that if we get that and it's not null, we convert it to a number.

I do have to also discuss what happened to the `DecodedLog` type.  For simplicity, `DecodedLog` extends `Log`.  However, `Log` has just had its type broadened.  But while that's fine for `Log`, because it's an input type, `DecodedLog` is an output type.  So to broaden its type would be a breaking change.  So I re-narrowed the types of the appropriate fields to prevent that.

Anyway that's it (aside from tests); this is pretty simple.

## Testing instructions

I altered some decoder tests so that they pass in strings for the block number (the only part that matters), I think that should be sufficient.

### DIFF
--- a/packages/decoder/lib/types.ts
+++ b/packages/decoder/lib/types.ts
@@ -125,11 +125,28 @@ export interface ContractState {
 
 /**
  * This type represents a web3 Log object that has been decoded.
- * Note that it extends the Log type and just adds an additional field
- * with the decoding.
+ * Note that it extends the Log type.   The differences are:
+ *
+ *   1. It adds an additional field with the decodings;
+ *   2. Some fields that in Log are allowed to be strings,
+ *     here are specified as numbers.  (Because this type is
+ *     for output and that type is for input.)
  * @category Results
  */
 export interface DecodedLog extends Log {
+  /**
+   * Index of the log within the block.
+   */
+  logIndex?: number;
+  /**
+   * Index within the block of the emitting transaction; null if
+   * block is pending.
+   */
+  transactionIndex?: number | null;
+  /**
+   * The block number.  Null if pending.
+   */
+  blockNumber: number | null;
   /**
    * An array of possible decodings of the given log -- it's an array because logs can be ambiguous.
    *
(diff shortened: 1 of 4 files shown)


packages/decoder/lib/types.ts
```diff
@@ -125,11 +125,28 @@ export interface ContractState {
...
 export interface DecodedLog extends Log {
+  /**
+   * Index of the log within the block.
+   */
+  logIndex?: number;
+  /**
+   * Index within the block of the emitting transaction; null if
+   * block is pending.
+   */
+  transactionIndex?: number | null;
```
gnidan, line 145: I think these optional fields need explanation on why they might be `undefined`. Probably warrants a note in the `DecodedLog` docstring, to say that @truffle/decoder _re-emits_ everything given, and unspecified optional input fields will thus not appear in output

gnidan (changes requested):
Small docs thing

packages/decoder/lib/types.ts
haltman-at: Actually, that's not how this works!  Which raises a good point: These fields shouldn't be optional.  I guess I should just not make this an extension of `Log`, huh?  Because it doesn't save me any work if everything is mandatory. :-/

To explain: If you use `decodeLog`, passing in your own logs, you just get back the decodings, you don't get `DecodedLog`.  You only get `DecodedLog` if you call `events()`, to which one does not pass in their own logs.  So actually these fields will always be present in this type.

...so do you agree that's what I should do?

gnidan: Can this emit the same info in both situations? It's weird that passing in a `Log` will remove fields, while `events()` will return everything log-related

haltman-at: What?  Passing in a `Log` doesn't remove fields.  What I'm saying is that `decodeLog` doesn't use the `DecodedLog` type at all.  It returns the decodings, they're not inside a `DecodedLog` container.  Only `events` uses that.

gnidan: Ohh, sorry to misunderstand. Yeah, let's not extend `Log`

gnidan (approved):

PR #6132, opened 2023-06-30
replace google analytics with mixpanel

eggplantzzz:
Google analytics moved to a new format which caused our analytics system to stop working. Google analytics is not really made for the kind of thing we have implemented here and so this PR drops in Mixpanel as a replacement.


packages/core/lib/services/analytics/google.js
```diff
@@ -1,187 +0,0 @@
...
-const analyticsEnable = [
-  {
-    type: "confirm",
-    name: "analyticsEnable",
-    message: "Analytics are currently disabled. Would you like to enable them?",
-    default: false
-  }
-];
-
-const googleAnalytics = {
```
cliffoo, line 48: It looks like the analytics object is still named `googleAnalytics` elsewhere (core/lib/commands/config/run.js), and some of these functions are not copied over to core/lib/services/analytics/mixpanel.js but are in use.

packages/core/package.json
```diff
@@ -73,6 +73,7 @@
     "iter-tools": "^7.5.0",
     "js-interpreter": "2.2.0",
     "lodash": "^4.17.21",
+    "mixpanel": "0.17.0",
```
cliffoo, line 76: Caret?

packages/core/lib/services/analytics/mixpanel.js
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+      sendObject["ec"] = "Error";
+      sendObject["ea"] = "nonzero exit code";
+      sendObject["el"] =
+        eventObject["version"] + " " + eventObject["exception"];
+      sendObject["dp"] = "/error";
+    }
+
+    if (userId) {
+      // eslint-disable-next-line no-unused-vars
```
cliffoo, line 28: This isn't necessary I think.

packages/core/lib/services/analytics/mixpanel.js
```diff
@@ -0,0 +1,41 @@
...
+      sendObject["ea"] = JSON.stringify(eventObject["args"]);
+      sendObject["el"] = eventObject["version"];
+      sendObject["dp"] = "/" + eventObject["command"];
+    } else {
+      sendObject["ec"] = "Error";
+      sendObject["ea"] = "nonzero exit code";
+      sendObject["el"] =
+        eventObject["version"] + " " + eventObject["exception"];
+      sendObject["dp"] = "/error";
+    }
```
cliffoo, line 25: Does mixpanel use the same measurement protocol params [as GA](https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters)? I'm not familiar.

eggplantzzz: You can just toss in arbitrary properties. I left them the same for continuity.

packages/core/package.json
eggplantzzz: Ech, I don't like adding the caret so that it won't accidentally get upgraded to some other version. Should I add it back in?

packages/core/lib/services/analytics/google.js
eggplantzzz: This should no longer be a problem.

packages/core/package.json
haltman-at: I mean, preventing accidental upgrades to another version is what `yarn.lock` is for, right?  Generally our practice has been not to pin unless there's a specific reason to.  Makes things harder for Dependabot, after all! :)

PR #6136, opened 2023-07-07
Ensure CalldataDecodingInspector always passes through inspector options

haltman-at:
## PR description

I noticed tonight that the `CalldataDecodingInspector` doesn't actually pass its inspector options through when inspecting intepretations as multicalls or etc.  Oops.  Changed that to pass that through.  Also turned on the `noHideAddress` option in debug-utils, because we always use that option in the debugger.

## Testing instructions

I'm going to be honest, I didn't really test this one, I wanted to get it up quick...

Because I changed the argument order of `formatFunctionLike`, which is exported, this is technically a breaking change to Codec (and indeed, I had to adjust the use in `debug-utils` appropriately).

It could be done in a non-breaking way, but that would be ugly.  Arguably we should switch from positional params to named params, but that would also be breaking, and frankly I don't think it's worth the effort for something this small.

### DIFF
--- a/packages/codec/lib/export.ts
+++ b/packages/codec/lib/export.ts
@@ -312,13 +312,15 @@ export class CalldataDecodingInspector {
           return formatMulticall(
             fullName,
             this.decoding.interpretations.multicall,
-            options
+            options,
+            this.options
           );
         } else if (this.decoding.interpretations.aggregate) {
           return formatAggregate(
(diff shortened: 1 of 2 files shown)


gnidan:
(This is why you should always used named options for everything all the time ;)

gnidan (approved):

PR #6169, opened 2023-08-22
Update Ganache to 7.9.1; turn off Node 14 and turn on Node 20 (and fix a bad type)

haltman-at:
As described!

I noticed some warnings in dashboard when bootstrapping this one that I hadn't seen before?  But that may possibly be due to me switching to Node 16 rather than using Node 14 like I had been.


haltman-at:
OK, added a commit to update hardhat, let's see how that goes...

haltman-at:
OK, going with latest hardhat, but I fixed the bad type in compile-solidity.

haltman-at:
OK, I also had to update how Ganache is initialized in one place; we'll see if there are any others.

haltman-at:
OK, I have no idea why this test of `truffle networks --clean` keeps failing.  It passes locally, including on Node 20...

PR #6171, opened 2023-08-25
Remove invalid networks check from Sourcify fetcher

haltman-at:
As described!

### DIFF
--- a/packages/fetch-and-compile/README.md
+++ b/packages/fetch-and-compile/README.md
@@ -124,7 +124,7 @@ export interface FetchAndCompileOptions {
 
 ### `getSupportedNetworks`
 
-If you want a list of supported networks, you can call `getSupportedNetworks`:
+If you want a (potentially partial) list of supported networks, you can call `getSupportedNetworks`:
 
 ```ts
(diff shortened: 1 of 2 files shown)


gnidan (commented):
How do we want to deal with the list of networks possibly becoming out of date?

haltman-at:
Well, as mentioned, we'll no longer be checking it, so the list will only even be relevant if you explicitly ask for supported networks.  As for how we stop that from going out of date, outside of constant manual updating... well, there are potential ways if, like, we import sourcify as a dependency?  I think we don't want to do that though.

gnidan:
> As for how we stop that from going out of date, outside of constant manual updating... well, there are potential ways if, like, we import sourcify as a dependency? I think we don't want to do that though.

Moreso I'm asking about how to ensure users are aware that this list may not be up-to-date

haltman-at:
Oh, yeah, I have no ideas for that offhand.

haltman-at:
OK, I added a note to the README per @gnidan's suggestion elsewhere.

gnidan (approved):

PR #6177, opened 2023-09-03
deps(core): semver@7.5.2->7.5.4

legobeat:
## PR description

- bump direct dependency `semver` to latest `7.5.4`.
- bump direct dependency `@types/semver` to latest `7.5.1`.
  - add assertion for type update
- dedupe `semver` to address ReDoS in transitive deps.
- dedupe `@types/semver`

Note that this does not completely remove all dependencies on broken versions of `semver`. It's still being pulled in via `nx` and `[ethereumjs-block,ethereumjs-vm] > merkle-patricia-tree > levelup@1`.
 
## Documentation


gnidan:
> Maintainers: This PR still keeps the version pinned to conform with existing approach but I'd like to propose changing to caret range `^7.5.4` for direct dependencies on `semver` to reduce further churn for you and downstreams for any future bugfixes.

I don't think we pin semver for any particular reason, so we can probably just switch it to caret. I'll check the history to confirm next week, to see if we pinned it on purpose at some point.

haltman-at (approved):
Seems good to me!

haltman-at:
Also, I just looked through the history and it looks to me like @eggplantzzz just pinned it mistakenly.  I don't see any particular reason it got pinned.  I'm OK with this being merged in its current form, but I agree it would be preferable for them to be unpinned.

gnidan (approved):
Confirmed that pinning seems accidental. Thanks for making this change!

PR #6180, opened 2023-09-03
Indicate Node.js supported versions

legobeat:
## PR description

- Add `engines.node` to indicate supported Node.js version range
  - Assuming minimum 16 since #6169 removed v14 from CI
- Add `.nvmrc` set to `lts/*`
- ci: Read version for `yarncheck` job from `.nvmrc` instead of using harcoded vesion

## Documentation


haltman-at:
Hm, do we want to set `.nvmrc` to `lts/*`?  That's what it is now, but we've often lagged behind a bit.

haltman-at:
Oh, well right now we're still supporting 16, so...?

haltman-at (approved):
Well, if Nick says it's good, then I'm also OK with it!

PR #6185, opened 2023-09-05
devDeps: deduplicate babel packages

legobeat:
## PR description

Currently `yarn.lock` pins multiple different versions of the same package, even if a single version would satisfy all the constraints. This wastes disk space, computational time, human time, and environmental resources ;)

This PR deduplicates packages with the `@babel/` and `babel-` prefixes, saving ~400MB.

```
#Before
$ du -hs node_modules
2.1G    node_modules

# After
$ du -hs node_modules
1.7G    node_modules
```

Based on #6177 as they touch the same lines and that one seems higher priority.

## Documentation

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -32,7 +32,8 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
-        node-version: [16.x, 18.x, 20.x]
+        # TODO: unpin to 20.x once 20.7 has ben released
+        node-version: [16.x, 18.x, 20.5]
         env: [GETH=true, PACKAGES=true, INTEGRATION=true]
     runs-on: ${{ matrix.platform }}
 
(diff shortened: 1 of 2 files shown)


haltman-at:
Did you PR the right thing? I'm only changes to semver packages in this PR.

haltman-at:
Oh, I see!

haltman-at:
Well, now this has a `yarn.lock` conflict...

haltman-at:
Hm, something seems to be going wrong.  A test in `codec-components` seems to be consistently failing.  @cliffoo, any ideas by any chance?

haltman-at:
Oh, ick -- it looks like this failure is happening on `develop` as well.  So it's not due to this PR.  But that leaves the question of why something is going wrong on `develop`...

haltman-at (approved):
Seems fine now that it's passing?

gnidan (approved):
Thanks for this!

PR #6193, opened 2023-09-12
ci: enforce deduped lockfile when linting dependencies

legobeat:
## PR description

- New workspace package scripts
  - `deduplicate:check`: exits with error if any redundancies in `yarn.lock` are detected
  - `deduplicate`: performs a best-effort attempt to resolve such duplicates
  - `lint:dependencies`: `deduplicate:check && depcheck`
  - `lint:dependencies:fix`: `deduplicate && depcheck`
- ci: fail if `yarn lint:dependencies` errors
- new devDependency [`yarn-deduplicate`](https://www.npmjs.com/package/yarn-deduplicate)
  - see this package for details about how the dedupe is done and possible configuration (the main one being if `highest` or `fewer` should be used - this uses `fewer` but it's a tradeoff and up to maintainer to decide if `fewer` or `highest` is preferred)

#### Blocked by
- #6194 
  - though the scripts could be broken out separately without the ci checks to make those not dependent, if this is desired 
  
## Documentation

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -49,7 +49,7 @@ jobs:
     - run: yarn install
     - run: test -z "$(git diff)" || (echo 'Did you check in a generated file to source control?  Please remove it if so'; false)
 
-    - run: yarn depcheck
+    - run: yarn lint:dependencies
 
     - run: ${{ matrix.env }} yarn ci
       env:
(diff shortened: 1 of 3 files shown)


haltman-at:
Nice!  The one thing I'm wondering about here is -- is there any possibility that `yarn-deduplicate` can fail, in such a way that the test would not pass without manual intervention?  You describe it above as "best-effort".  I'm wondering that means it can get stuck in a fail state like that.

haltman-at (approved):
Well, seems good to me, I guess.  We'll see what others think!

gnidan (approved):

